PR #10322, opened 2014-12-04
fix($parse): allow assignment on objects in locals

### LINKED ISSUE #4664: `$eval` assignment expression and locals
We can't use a `local` variable in the left side of an assignment without enclosing it in parenthesis to ensure it to be resolved first.

I tried to raise a discussion [here](https://groups.google.com/forum/#!searchin/angular/eval$20locals/angular/rEvNHJ9x5jI/CraUhwCMOsMJ) in order to understand whether it's a bug or not but had not much success. 

Basically, this test case won't succeed:

```javascript
it('should allow passing object as locals to the left part of an assignment expression', inject(function($rootScope) {
  var ob = {};
  $rootScope.scopedValue = 1;
  $rootScope.$eval('a.value = scopedValue', {a: ob});
  expect(ob.value).toBe(1);
  expect($rootScope.a).toBeUndefined();
}));
```

While changing the expression to `(a).value = scopedValue` the test will succeed. Assembled a [Plnkr here](http://plnkr.co/edit/OymAjMSdGXjwqifesxKD?p=preview) to illustrate it.

As I said before, I'm not sure if this is a bug or a limitation. If it's the later, then I think I could change the documentation to exemplify it. So, is this a bug or a expected limitation?

jbedard:
Fixes #4664 

I saw this issue on the 1.4 list but the fix seems simple...

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -489,19 +489,19 @@ var formDirectiveFactory = function(isNgForm) {
                 alias = controller.$name;
 
             if (alias) {
-              setter(scope, alias, controller, alias);
+              setter(scope, null, alias, controller, alias);
               attr.$observe(attr.name ? 'name' : 'ngForm', function(newValue) {
                 if (alias === newValue) return;
-                setter(scope, alias, undefined, alias);
+                setter(scope, null, alias, undefined, alias);
                 alias = newValue;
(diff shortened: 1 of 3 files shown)


petebacondarwin:
Landed as 869008140a96e0e9e0d9774cc2e5fdd66ada7ba9
Thanks @jbedard et al

PR #10327, opened 2014-12-04
fix(input): create max and/or min validation whatever the initial value is

### LINKED ISSUE #10307: max validator of input element do not work as expected on some cases
Demo Html, or view it on [Plunker](http://plnkr.co/edit/Uq0rACDTbkl8Fokaizmo?p=preview)
````html
<!DOCTYPE html>
<html lang="en">
<head>
    <title></title>
    <meta http-equiv="content-type" content="text/html;charset=UTF-8" />
    <script src="//ajax.googleapis.com/ajax/libs/angularjs/1.3.5/angular.min.js"></script>
    <script>
        angular.module("foo", []).controller("fooCtrl", function ($scope) {
            $scope.limit3 = 2;
        });
    </script>
    <style>
        input.ng-invalid {
            background-color: red;
        }
    </style>
</head>
<body ng-app="foo">
    <label>Form 1</label>
    <form action="/" method="post">
        <!--no controller-->
        <input type="number" ng-model="limit1" placeholder="limit" />
        <input type="number" max="{{limit1}}" ng-model="used1" placeholder="used" />
    </form>
    <br />
    <label>Form 2</label>
    <form action="/" method="post">
        <!--no controller, use ng-init directive-->
        <input type="number" ng-init="limit2=2" ng-model="limit2" placeholder="limit" />
        <input type="number" max="{{limit2}}" ng-model="used2" placeholder="used" />
    </form>
    <br />
    <label>Form 3</label>
    <form action="/" method="post" ng-controller="fooCtrl">
        <!--has controller-->
        <input type="number" ng-model="limit3" placeholder="limit" />
        <input type="number" max="{{limit3}}" ng-model="used3" placeholder="used" />
    </form>
</body>
</html>
````

Result:

rtx

The problem if that form 1 can not detect the max error.
This issue was found in both 1.2.x and 1.3.x.
I haven't test the other validators, such as min, max-length, may be there are some problem.

tobyee:
- fix issue #10307
- change tests to corresponding changes
- also change tests for ngmax and ngmin (though they have no some issue)

Closes #10307

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1230,7 +1230,7 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     return value;
   });
 
-  if (attr.min || attr.ngMin) {
+  if (isDefined(attr.min) || attr.ngMin) {
     var minVal;
     ctrl.$validators.min = function(value) {
       return ctrl.$isEmpty(value) || isUndefined(minVal) || value >= minVal;
@@ -1246,7 +1246,7 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     });
   }
 
-  if (attr.max || attr.ngMax) {
+  if (isDefined(attr.max) || attr.ngMax) {
     var maxVal;
     ctrl.$validators.max = function(value) {
       return ctrl.$isEmpty(value) || isUndefined(maxVal) || value <= maxVal;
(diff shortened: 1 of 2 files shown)


Narretz:
Hi @Tobyee, sorry for the late response. 1.2.x only receives security fixes, so we won't backport it.

In the commit message it says:
> also change tests for ngmax and ngmin (though they have no some issue)
 is this statement still relevant? If so, to what does "they have no some issue" mean? You mean that they are correctly created even when the value is `undefined` ?

Tobyee:
Yes, I
>mean that they are correctly created even when the value is undefined

PR #10348, opened 2014-12-06
WIP: add karma-benchpress support for all benchmarks

jeffbcross:
With this PR, there is a karma test that will run the compiled largetable-bp benchmark in `build/benchmarks` with 1 sample for each variable. The results are just logged to the console with some basic tabbed formatting. Anyone should be able to make this work by the following steps (requires Chrome Canary):

``` shell
$ npm install . (may require removing npm-shrinkwrap.json to get latest benchpress & karma-benchpress dependencies)
$ grunt bp_build
$ karma start karma-benchpress.conf.js
```

Karma should start up, and should automatically run each benchmark/variable combination in a new incognito window (with other flags set for memory reporting), and should log the results in the process console.

TODOS for this PR:
- [x] Update shrinkwrap (having trouble figuring out which version of node I should use to generate this properly. Will probably end up hand-coding the changed dependencies.)
- [x] Update remaining benchmarks to new Variables and ready() APIs
- [x] Create tests to run remaining benchmarks with each variable with 25 iterations & 20 samples
- [x] Create jenkins job to run benchmarks on each SHA
- [ ] Squash commits

TODOS for later:
- Report standard deviation and coefficient of variation back from karma-benchpress. This needs to be fixed in the karma-benchpress framework.
- Push the results to a logging service, along with environment information and other metadata

Whatever CI server this ends up running on (Jenkins, my workstation), I don't think it should be part of the regular angular.js job given how long it would take to run all benchmarks with all variables n times. Ie if the average sample of all steps takes 750ms (made up number, probably not far off), multiply that by number of iterations (25) by number of benchmark variables (33 at present). This would take about 10 minutes for each revision.

### DIFF
--- a/benchmarks/event-delegation-bp/app.js
+++ b/benchmarks/event-delegation-bp/app.js
@@ -54,4 +54,41 @@ app.controller('DataController', function($rootScope) {
       $rootScope.$apply();
     }
   });
+
+  bp.variables.addMany([
+    {
+      value: 'ngClick',
+      label: 'ngClick'
+    },
+    {
+      value: 'ngClickNoJqLite',
+      label: 'ngClick without jqLite'
+    },
+    {
+      value: 'ngShow',
+      label: 'baseline: ng-show'
+    },
+    {
+      value: 'textInterpolation',
+      label: 'baseline: text interpolation'
+    },
+    {
+      value: 'dlgtClick',
+      label: 'delegate event directive (only compile)'
+    },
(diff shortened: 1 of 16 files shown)


petebacondarwin:
@jeffbcross where did you get to on this PR? Is it still valid / useful / worth merging?

petebacondarwin:
Closing on the command of @jeffbcross

PR #10352, opened 2014-12-06
fix(filter): throw error if not used with an array

### LINKED ISSUE #9992: Generate error when attempting to use filter on objects
Currently it fails silently if you try to use `filter` on objects, as in the example below, because it only supports arrays.

````
<div ng-repeat="item in object | filter:{display:true}">stuff</div>
````

I blew a ton of time on this because it's not very intuitive since `ng-repeat` supports objects. I fully expect to make the same mistake again. A simple error message would save a world of hurt.

### PR DESCRIPTION
**Pull request by Puigcerber**
Throw error if filter is not used with an array.

Closes #9992

BREAKING CHANGE: Previously, the filter was not applied if used with a non array.
Now, it throws an error.

### DIFF
--- a/docs/content/error/filter/notarray.ngdoc
+++ b/docs/content/error/filter/notarray.ngdoc
@@ -1,0 +1,8 @@
+@ngdoc error
+@name filter:notarray
+@fullName Not an array
+@description
+
+This error occurs when {@link ng.filter filter} is not used with an array.
+Filter must be used with an array so a subset of items can be returned.
+The array can be initialized asynchronously so null or undefined won't throw this error.
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -117,7 +117,13 @@
  */
 function filterFilter() {
   return function(array, expression, comparator) {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: docs/content/error/filter/notarray.ngdoc [hunk 1]
```diff
@@ -0,0 +1,8 @@
+@ngdoc error
+@name filter:notarray
+@fullName Not an array
+@description
+
+This error occurs when {@link ng.filter filter} is not used with an array.
+Filter must be used with an array so a subset of items can be returned.
+The array can be initialized asynchronously so null or undefined won't throw this error.
```
caitp: personally I think it's good to show causes of the error in code snippets, as well as suggested fixes.

caitp:
lgtm, will land today

caitp:
a good followup bug would be to improve the error document a bit

docs/content/error/filter/notarray.ngdoc
Puigcerber: Ok. I'll try to add it when I find time this week. Thanks!

PR #10355, opened 2014-12-06
fix($parse): ensuring shared sub expressions do not use full expression text

jbedard:
This depends on #10351 otherwise some of the changed error messages would be even shorter (tests will probably fail for now showing why...).

I guess this is debatable but I don't think anything shared (just `getterFn`s really...) should reference the `fullExp` which created it when it might be used by other expressions. For example (without this patch):

```
$rootScope.$eval('foo.constructor + 123')  => 123
$rootScope.$eval('foo.constructor.foo()', {foo: Function}) => 
   [$parse:isecfn] Referencing Function in Angular expressions is disallowed! Expression: foo.constructor + 123
```

This patch changes it so the shared `getterFn` only references the shared path itself. While seeing the full expression would be nicer at least for now it won't be wrong.

This was caused by e676d642f5feb8d3ba88944634afb479ba525c36 and also exists in 1.2 from the same security fix.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -830,14 +830,14 @@ function isPossiblyDangerousMemberName(name) {
  * - http://jsperf.com/angularjs-parse-getter/4
  * - http://jsperf.com/path-evaluation-simplified/7
  */
-function cspSafeGetterFn(key0, key1, key2, key3, key4, fullExp, expensiveChecks) {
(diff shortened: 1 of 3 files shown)


jbedard:
Well I tried what I mentioned a while back and it basically requires adding an argument to 23432 functions so I don't think it's worth it. I still think this PR is an improvement but I'll just close this for now since it should be fixed in 14...

PR #10368, opened 2014-12-08
feat(orderBy): make orderBy filter accept a custom compare function.

Rowanto:
Expose the inner compare function used by orderBy to the outside world.
This will make it possible to use orderBy with a very specific sort requirement like sorting using the localeCompare or making a few entries appear first in the list.

### DIFF
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -32,6 +32,10 @@
  *    If the predicate is missing or empty then it defaults to `'+'`.
  *
  * @param {boolean=} reverse Reverse the order of the array.
+ * @param {function(object1, object2, defaultCompare)} compare The custom compare function to use.
+ *    Should return -1 if object2 is bigger than object1. 0 if the same. 1 if object2 is smaller than
+ *    object1. The third variable is the default compare function which can be reused. It accepts
+ *    object1, and object2 as parameter.
  * @returns {Array} Sorted copy of the source array.
  *
  * @example
@@ -47,6 +51,16 @@
                   {name:'Adam', phone:'555-5678', age:35},
                   {name:'Julie', phone:'555-8765', age:29}];
              $scope.predicate = '-age';
+             $scope.ladiesFirst = function(name1, name2, defaultComparator) {
+               if (name1 === 'Julie' || name1 === 'Mary') {
+                 return -1;
+               }
+               if (name2 === 'Julie' || name2 === 'Mary') {
+                 return 1;
+               }
+               return defaultComparator(name1, name2);
+             };
+             $scope.customSort;
            }]);
        </script>
        <div ng-controller="ExampleController">
@@ -55,12 +69,13 @@
          [ <a href="" ng-click="predicate=''">unsorted</a> ]
          <table class="friend">
            <tr>
-             <th><a href="" ng-click="predicate = 'name'; reverse=false">Name</a>
-                 (<a href="" ng-click="predicate = '-name'; reverse=false">^</a>)</th>
-             <th><a href="" ng-click="predicate = 'phone'; reverse=!reverse">Phone Number</a></th>
-             <th><a href="" ng-click="predicate = 'age'; reverse=!reverse">Age</a></th>
+             <th><a href="" ng-click="predicate = 'name'; reverse=false; customSort = null">Name</a>
+                 (<a href="" ng-click="predicate = '-name'; reverse=false; customSort = null">^</a>)</th>
+             <th><a href="" ng-click="predicate = 'phone'; reverse=!reverse; customSort = null">Phone Number</a></th>
+             <th><a href="" ng-click="predicate = 'age'; reverse=!reverse; customSort = null">Age</a></th>
+             <th><a href="" ng-click="predicate = 'name'; reverse=false; customSort = ladiesFirst">Ladies First</a></th>
            </tr>
-           <tr ng-repeat="friend in friends | orderBy:predicate:reverse">
+           <tr ng-repeat="friend in friends | orderBy:predicate:reverse:customSort">
              <td>{{friend.name}}</td>
              <td>{{friend.phone}}</td>
              <td>{{friend.age}}</td>
@@ -86,6 +101,7 @@
               (<a href="" ng-click="order('-name',false)">^</a>)</th>
             <th><a href="" ng-click="reverse=!reverse;order('phone', reverse)">Phone Number</a></th>
             <th><a href="" ng-click="reverse=!reverse;order('age',reverse)">Age</a></th>
+            <th><a href="" ng-click="reverse=false;order('name',reverse, ladiesFirst)">Ladies First</a></th>
           </tr>
           <tr ng-repeat="friend in friends">
             <td>{{friend.name}}</td>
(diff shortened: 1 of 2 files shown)


elad:
Yes please, this would be great for i18n support.

elad:
Okay, in case it helps anyone, here's a quick and dirty solution. I took the orderBy code and slightly modified it to support locale-aware string sorting. I expose a method to let the filter know which locale to use. If the method's unused, it should fallback to the browser's locale. I also introduced a new notation, the `@` symbol, that indicates "this string is to be sorted by locale." (Open to alternatives...)

For brevity, controller usage:

```js
angular.module('exampleApp', ['orderByLocaleAware'])
.controller('exampleController', function ($scope, orderByLocaleAwareId) {
  $scope.setLocaleId('he-IL');

  $scope.sortColumns = ['@+firstName', '@-lastName'];

  $scope.data = [
    { firstName: 'Elad', lastName: 'Efrat' },
    { firstName: 'אלעד', lastName: 'אפרת' }
  ];
});
```

and filter usage:

```html
<div ng-controller="exampleController">
  <table>       
    <tr ng-repeat="record in data | orderByLocaleAware:sortColumns">
      <td>{{ record.firstName }}</td>
      <td>{{ record.lastName }}</td>
    </tr>
  </table>
</div>
```

I'm going to use it in my Angular i18n/l10n helper module, [angular-kencale](https://github.com/elad/angular-kencale), but I can release it as a standalone module as well.

Here's a full example you can play with, just copy/paste to (say) `example.html` and open in a browser:

```html
<!doctype html>
<html ng-app="exampleApp">
  <head>
    <meta charset="utf-8">

    <style>
      .strong { font-weight: bold };
    </style>

    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.3/angular.min.js"></script>

      angular.module('exampleApp', ['orderByLocaleAware'])
      .controller('exampleController', function ($scope, orderByLocaleAwareId) {
        $scope.setLocaleId = function(localeId) {
          orderByLocaleAwareId.localeId = localeId;
        }

        $scope.getLocaleId = function() {
          return orderByLocaleAwareId.localeId;
        }

        // Set an initial locale.
        $scope.setLocaleId('he-IL');

        // The '@' symbol indicates the string is to be ordered by locale.
        $scope.sortColumns = ['@firstName', '@lastName'];

        // Sample data. For he-IL, Hebrew names should appear first. For en-US, English names.
        $scope.data = [
          { firstName: 'Elad', lastName: 'Efrat' },
          { firstName: 'אלעד', lastName: 'אפרת' }
        ];
      });
    </script>
  </head>

  <body>
    <div ng-controller="exampleController">
      <p>
        Locale:
        <a href="#" ng-click="setLocaleId('en-US')" ng-class="{ strong: getLocaleId() === 'en-US' }">en-US</a>
        <a href="#" ng-click="setLocaleId('he-IL')" ng-class="{ strong: getLocaleId() === 'he-IL' }">he-IL</a>
      </p>

      <table>       
        <tr ng-repeat="record in data | orderByLocaleAware:sortColumns">
          <td>{{ record.firstName }}</td>
          <td>{{ record.lastName }}</td>
        </tr>
      </table>
    </div>

    <!-- The module, copied almost verbatim from AngularJS. -->
      angular.module('orderByLocaleAware', [])
      .service('orderByLocaleAwareId', function () {
        this.localeId = null;
      })
      .filter('orderByLocaleAware', function ($parse, orderByLocaleAwareId) {
        return function(array, sortPredicate, reverseOrder) {
          if (!array || !angular.isArray(array)) return array;

          if (!angular.isArray(sortPredicate)) { sortPredicate = [sortPredicate]; }

          var predicates = processPredicates(sortPredicate, reverseOrder);
          // Add a predicate at the end that evaluates to the element index. This makes the
          // sort stable as it works as a tie-breaker when all the input predicates cannot
          // distinguish between two elements.
          predicates.push({ get: function() { return {}; }, descending: reverseOrder ? -1 : 1});

          // The next three lines are a version of a Swartzian Transform idiom from Perl
          // (sometimes called the Decorate-Sort-Undecorate idiom)
          // See https://en.wikipedia.org/wiki/Schwartzian_transform
          var compareValues = Array.prototype.map.call(array, getComparisonObject);
          compareValues.sort(doComparison);
          array = compareValues.map(function(item) { return item.value; });

          return array;

          function getComparisonObject(value, index) {
            return {
              value: value,
              predicateValues: predicates.map(function(predicate) {
                return getPredicateValue(predicate.get(value), index);
              })
            };
          }

          function doComparison(v1, v2) {
            var result = 0;
            for (var index=0, length = predicates.length; index < length; ++index) {
              result = compare(v1.predicateValues[index], v2.predicateValues[index], predicates[index].localeAware) * predicates[index].descending;
              if (result) break;
            }
            return result;
          }
        };

        function processPredicates(sortPredicate, reverseOrder) {
          reverseOrder = reverseOrder ? -1 : 1;
          return sortPredicate.map(function(predicate) {
            var descending = 1, get = angular.identity, localeAware = false;

            if (angular.isFunction(predicate)) {
              get = predicate;
            } else if (angular.isString(predicate)) {
              if (predicate.charAt(0) == '@') {
                localeAware = true;
                predicate = predicate.substring(1);
              }
              if ((predicate.charAt(0) == '+' || predicate.charAt(0) == '-')) {
                descending = predicate.charAt(0) == '-' ? -1 : 1;
                predicate = predicate.substring(1);
              }
              if (predicate !== '') {
                get = $parse(predicate);
                if (get.constant) {
                  var key = get();
                  get = function(value) { return value[key]; };
                }
              }
            }
            return { get: get, descending: descending * reverseOrder, localeAware: localeAware };
          });
        }

        function isPrimitive(value) {
          switch (typeof value) {
            case 'number': /* falls through */
            case 'boolean': /* falls through */
            case 'string':
              return true;
            default:
              return false;
          }
        }

        function objectValue(value, index) {
          // If `valueOf` is a valid function use that
          if (typeof value.valueOf === 'function') {
            value = value.valueOf();
            if (isPrimitive(value)) return value;
          }
          // If `toString` is a valid function and not the one from `Object.prototype` use that
          if (hasCustomToString(value)) {
            value = value.toString();
            if (isPrimitive(value)) return value;
          }
          // We have a basic object so we use the position of the object in the collection
          return index;
        }

        function getPredicateValue(value, index) {
          var type = typeof value;
          if (value === null) {
            type = 'string';
            value = 'null';
          } else if (type === 'string') {
            value = value.toLowerCase();
          } else if (type === 'object') {
            value = objectValue(value, index);
          }
          return { value: value, type: type };
        }

        function compare(v1, v2, localeAware) {
          var result = 0;
          if (v1.type === v2.type) {
            if (v1.value !== v2.value) {
              result = (localeAware && v1.type === 'string') ? v1.value.localeCompare(v2.value, orderByLocaleAwareId.localeId) : (v1.value < v2.value ? -1 : 1);
            }
          } else {
            result = v1.type < v2.type ? -1 : 1;
          }
          return result;
        }

        function hasCustomToString(obj) {
          return angular.isFunction(obj.toString) && obj.toString !== Object.prototype.toString;
        }
      });
    </script>
  </body>
</html>
```

jrencz:
:+1:

jrencz:
I think that in addition to being able to set compare function as another filter argument there should be a possibility to set a default compare function with a provider (optimally: also to pass a service or a service name just as we can do with `$http` interceptors).

I'm not convinced compare function should be another argument.
It should probably extend the 2nd argument by either replacing the `reverse` argument (which is backwards incompatible) or allowing 2nd argument to be a function. `reverse` is just a special case of comparison where returned values are multiplied by `-1`. I suggest that if a function is passed as the 2nd argument it should be treated as a compare function and if a boolean is passed then it will be treated as `reverse` flag to maintain backwards compatibility.
Support for boolean flag should probably be marked as deprecated and eventually removed for simplicity.

Maaartinus:
I've just tried it and it works well.... but very slow: a few seconds for 2000 elements. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#Performance)  `Intl.Collator` should be used. I tried it and it helped a lot (nearly back to always instant rendering).

PR #10369, opened 2014-12-08
refactor($browser): wrap Get(history.state) in try/catch

### LINKED ISSUE #10367: "SCRIPT16389: Unspecified error" from AngularJs in IE if no browser history
**Angular version:** 1.3.4
**Browser version:** Internet Explorer 10

In IE10 (may also occur in other versions of IE as well but haven't got easy access to other versions to test) then if a web application using AngularJs is the very first page loaded in the browser the following JavaScript error may occur:

SCRIPT16389: Unspecified error. 
angular.js, line 4116 character 11

This is coming from the call to factory() in the following:

```javascript
    function getService(serviceName) {
      if (cache.hasOwnProperty(serviceName)) {
        if (cache[serviceName] === INSTANTIATING) {
          throw $injectorMinErr('cdep', 'Circular dependency found: {0}',
                    serviceName + ' <- ' + path.join(' <- '));
        }
        return cache[serviceName];
      } else {
        try {
          path.unshift(serviceName);
          cache[serviceName] = INSTANTIATING;
          return cache[serviceName] = factory(serviceName);
        } catch (err) {
          if (cache[serviceName] === INSTANTIATING) {
            delete cache[serviceName];
          }
          throw err;
        } finally {
          path.shift();
        }
      }
    }
```

After further investigation I've found that this appears to be originating from the "cachedState = window.history.state;" statement in the following function.

```javascript
  function cacheState() {
    // This should be the only place in $browser where `history.state` is read.
    cachedState = window.history.state;
    cachedState = isUndefined(cachedState) ? null : cachedState;

    // Prevent callbacks fo fire twice if both hashchange & popstate were fired.
    if (equals(cachedState, lastCachedState)) {
      cachedState = lastCachedState;
    }
    lastCachedState = cachedState;
  }
```

This error does not occur if any other page has been visited prior to going to the page that uses AngularJs as window.history.state then no longer returns an error in IE.

Note that this error does not appear to occur when using Firefox or Chrome (as neither of these browsers raise an error when accessing window.history.state).

A try catch block on the window.history.state should fix this:

```javascript
  function cacheState() {
      // This should be the only place in $browser where `history.state` is read.
    try {
        cachedState = window.history.state;
    }
    catch (e) {
        cachedState = null;
    }
    cachedState = isUndefined(cachedState) ? null : cachedState;

    // Prevent callbacks fo fire twice if both hashchange & popstate were fired.
    if (equals(cachedState, lastCachedState)) {
      cachedState = lastCachedState;
    }
    lastCachedState = cachedState;
  }
```

caitp:
Reportedly, MSIE can throw under certain conditions when fetching this attribute.

Will confirm before landing.

Fixes #10367

### DIFF
--- a/src/ng/browser.js
+++ b/src/ng/browser.js
@@ -235,10 +235,22 @@ function Browser(window, document, $log, $sniffer) {
 
   // This variable should be used *only* inside the cacheState function.
   var lastCachedState = null;
+  function getCurrentState() {
+    try {
+      var state = history.state;
+      if (state === void 0) {
+        state = null;
+      }
+      return state;
+    } catch (e) {
+      // MSIE can reportedly throw when there is no state (UNCONFIRMED).
+      return null;
+    }
+  }
+
   function cacheState() {
     // This should be the only place in $browser where `history.state` is read.
-    cachedState = window.history.state;
-    cachedState = isUndefined(cachedState) ? null : cachedState;
+    cachedState = getCurrentState();
 
     // Prevent callbacks fo fire twice if both hashchange & popstate were fired.
     if (equals(cachedState, lastCachedState)) {


shahata:
@caitp any progress with this? Apparently we are experiencing this issue in one of our apps as well.

caitp:
try the patch and see if it fixes it, i have no idea how we test this since nobody has produced a repro scenario

shahata:
Yeah, I found out a couple a days ago that we've been using this patch for a while now in some project. @donataswix any chance you can post a reproduce of this issue?

caitp:
> Yeah, I found out a couple of days ago that we've been using this patch for a while now in some project.

does that mean the patch solved some issue, or it's still causing problems?

shahata:
Yeah, the patch solved the issue. Hopefully @donataswix will be able to add more details tomorrow.

donataswix:
The patch solved the issue, I'll try to make reproducible example later today or at least figure out some more details.

petebacondarwin:
@donataswix - that would be extremely helpful. Thanks

donataswix:
For better or worse I can no longer reproduce this no matter what I try. So even though that patch helped at the time, it's no longer necessary for us. Although, I'm not sure I want to remove it from our build...

If anyone's interested our setup has dynamically created iframe where our app is loaded. There is a lot of code that runs before ours, and it seems that that code was responsible for creating preconditions for bug to appear.

src/ng/browser.js
```diff
@@ -235,10 +235,22 @@ function Browser(window, document, $log, $sniffer) {
 
   // This variable should be used *only* inside the cacheState function.
   var lastCachedState = null;
+  function getCurrentState() {
+    try {
+      var state = history.state;
+      if (state === void 0) {
```
petebacondarwin: Perhaps we should change `isDefined()` to use this method of testing for `undefined` and use it here?

petebacondarwin: Since `history.state` is either going to be a valid object, undefined or null, or it is going to throw an error we can simply do:

```js
function getCurrentState() {
  try {
    return history.state || null;
  } catch(e) {
    return null;
  }
}
```

caitp: It's not going to be either a valid object, undefined or null, eg https://dxr.mozilla.org/mozilla-central/source/dom/webidl/History.webidl#18

caitp: could be "false", could be "0", could be the empty string, etc

caitp: technically, it can legally be `undefined`, too... so hmm

petebacondarwin: Looks like we are pretending that it will never be `undefined` : https://github.com/angular/angular.js/blob/master/src/ng/browser.js#L159-L164

petebacondarwin:
Given that whatever we do here it is not going to be **correct**,  we should keep the behaviour as close to what it is currently, so can we simply go with the following?

```js
function getCurrentState() {
  try {
    return history.state;
  } catch (e) {
    // MSIE can reportedly throw when there is no state (UNCONFIRMED).
    return null;
  }
}

function cacheState() {
  cachedState = getCurrentState();
  cachedState = isUndefined(cachedState) ? null : cachedState;
}
```

Is that OK with you @caitp ?

dbabaioff:
We also are experiencing this issue. We tried the patch and didn't work (it didn't throw an exception). To solve we did like this:
```javascript
Object.defineProperty(window.history, 'state', {
    get: function() { return null; }
});
```

petebacondarwin:
@dbabaioff - when you say it didn't throw an exception with this patch, do you mean that you get the error message without an error being thrown? Are you able to create a reproduction of this that we could add to our unit tests?

dbabaioff:
yes - I get the error message ("SCRIPT16389: Unspecified error .") without an error being throw. I'll try to create a reproduction of this.

petebacondarwin:
@dbabaioff - did you ever manage to create a reproduction of this without an error being thrown?

filipposarzana:
@dbabaioff yes! it worked for me like a charm! Thanks!!!

PR #10381, opened 2014-12-09
fix($parse): making unary +/- operators consistent

### LINKED ISSUE #10371: Unary plus in expression does not work
<a href='http://plnkr.co/edit/aoq1JipsmkVdw8OFkNOM?p=preview'>Here is plunker examples</a> of expressions with unary pluses that does not work

jbedard:
Fixes #10371
- makes `+undefined` evaluate as a numeric operator the same as `-undefined`
- allows `+` to be followed by another unary operator like `-` already does
- no longer parses `--i` or `---------i` which were previously parsed as a bunch of negates

That last one is a breaking change but I assume not many people use such things. If anyone it is probably people using `--i` and then spending far too much time trying to figure out why the decrement operator isn't working right...

I'm half tempted to do the opposite change to be more restrictive (make `-` more like `+`) but that would be a bigger breaking change.

EDIT: this has been changed to no longer be breaking and now only makes unary `+` the same as unary `-`. So everything above is out of date...

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -642,9 +642,7 @@ Parser.prototype = {
 
   unary: function() {
     var token;
-    if (this.expect('+')) {
-      return this.primary();
-    } else if ((token = this.expect('-'))) {
+    if ((token = this.expect('-', '+'))) {
       return this.binaryFn(Parser.ZERO, token.text, this.unary());
     } else if ((token = this.expect('!'))) {
       return this.unaryFn(token.text, this.unary());
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -247,7 +247,7 @@ describe('parser', function() {
       }));
 
       it('should parse expressions', function() {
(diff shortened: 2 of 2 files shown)


lgalfaso:
I still think that this patch is not doing the right thing. Eg I would expect `expect(scope.$eval('+"1"')).toEqual(1);`

jbedard:
This patch just makes unary + like - and doesn't really fix anything (just makes them consistent). Do you think fixing them further is worth it in 1.3? Could just make a separate set of `UNARY_OPERATORS` instead of the current approach of throwing a 0 in...

lgalfaso:
> Do you think fixing them further is worth it in 1.3?
I think it is better to focus on 1.4, but will leave that to @petebacondarwin

Just saying that with the patch `scope.$eval('+"1"') === "01"`, that is not what people might expect

jbedard:
I don't think it's worth fixing for 1.3 either so I'm closing this for now. If someone really wants this in 1.3 let me know...

PR #10413, opened 2014-12-11
refactor($interpolate): remove some unnecessary logic, move functions out of closure

jbedard:
Moving the functions out actually decreased memory usage 3% in the benchmark added in #10411 (according to benchpress when using chrome with the memory/gc params).

### DIFF
--- a/src/ng/interpolate.js
+++ b/src/ng/interpolate.js
@@ -88,6 +88,28 @@ function $InterpolateProvider() {
       return '\\\\\\' + ch;
     }
 
+    function unescapeText(text) {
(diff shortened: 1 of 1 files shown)


jbedard:
Anyone taken a look at this? Seemed worthwhile to me...

lgalfaso:
It is odd that stringify(1) is no longer a string, the name of the function
does not match what it does
>
>

jbedard:
That's true. I've removed that commit for now, can always do that later. Now there is just the one commit that moves the functions up...

lgalfaso:
landed as 560951e9881b5f772262804384b4da9f673b925e

PR #10414, opened 2014-12-11
perf(ngModel,form): avoid invoking $interpolate for empty/unset values

jbedard:
This reduces GC (KB) by ~10% in the added benchmark create step.

I'm tempted to add a helper such as `$interpolate.$$eval(exp, scope)` to avoid the ugly duplication. A helper could also be used for spots such as https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L1922 (the helper can pass `mustHaveExpression=true` to avoid creating the interp result function even though it has no interpolation...). Worth it?


Narretz:
@lgalfaso do you want to give this another shot?

lgalfaso:
@Narretz will look at this this WE

jbedard:
This PR is pretty ugly as is. I still think the best solution is to avoid interpolating the same thing per link (7d61a984c23139752f8d778bd2b674856dc5d28e). Although a more general quick-exit in `$interpolate` for strings with no interpolation would still be good as well.

petebacondarwin:
@jbedard, @lgalfaso and @Narretz - do you fancy getting this improvement in this week?

lgalfaso:
@petebacondarwin will take another look this week

PR #10422, opened 2014-12-11
feat($filter): Display Infinity symbol when number is Infinity

### LINKED ISSUE #10421: $filter number: Could we display an Infinity symbol in case of Infinity value?
Could we display an Infinity symbol (∞) instead of an empty string through the "number" filter, when it's applied to `Infinity/-Infinity`?
According to [this thread](http://stackoverflow.com/questions/2254913/infinity-symbol-with-html), we could use `&infin;` or `&#8734;` (decimal) or `&#x221E;` (hex)

I can come with a PR if it's in the framework's scope. I guess we only have to add a case [here](https://github.com/angular/angular.js/blob/master/src/ng/filter/filters.js#L141).

QuentinFchx:
Infinity is a value and should not be treated as an empty string

Closes #10421

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -82,6 +82,8 @@ function currencyFilter($locale) {
  *
  * If the input is not a number an empty string is returned.
  *
+ * If the input is an infinite (Infinity/-Infinity) the Infinity symbol '∞' is returned.
+ *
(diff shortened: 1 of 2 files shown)


lgalfaso:
landed as 51d6774286202b55ade402ca097e417e70fd546b

PR #10424, opened 2014-12-11
fix(filterFilter): Fix filtering using an object expression when the filter value is undefined

caseyhoward:
Fixes #10419

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -196,7 +196,7 @@ function deepCompare(actual, expected, comparator, matchAgainstAnyProp) {
       } else if (expectedType === 'object') {
         for (key in expected) {
           var expectedVal = expected[key];
-          if (isFunction(expectedVal)) {
+          if (isFunction(expectedVal) || isUndefined(expectedVal)) {
             continue;
           }
 
--- a/test/ng/filter/filterSpec.js
+++ b/test/ng/filter/filterSpec.js
@@ -48,6 +48,15 @@ describe('Filter: filter', function() {
   });
 
 
+  it('should filter when a specific property is undefined', function() {
+    var items = [{name: 'a'}, {name: 'abc'}];
+    expect(filter(items, {name: undefined})).toEqual([{name: 'a'}, {name: 'abc'}]);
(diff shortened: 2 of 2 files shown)


src/ng/filter/filter.js
```diff
@@ -196,7 +196,7 @@ function deepCompare(actual, expected, comparator, matchAgainstAnyProp) {
       } else if (expectedType === 'object') {
         for (key in expected) {
           var expectedVal = expected[key];
-          if (isFunction(expectedVal)) {
+          if (isFunction(expectedVal) || isUndefined(expectedVal)) {
```
zuzusik: This broke things in our project. It changed the behaviour of `$filter`. I don't think it is OK to change the behaviour in minor releases.

pkozlowski-opensource: @zuzusik I was under the impression that c62fa6bd898e1048d4690d41034489dc60ba6ac2 was only merged to 1.4.x beta so a release where we've got breaking changes. Do you see it on any of the 1.3.x branches as well?

zuzusik: It was successfully landed to `1.3.15`, look at the [changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#1315-locality-filtration-2015-03-17)

zuzusik: @pkozlowski-opensource here it is: https://github.com/angular/angular.js/commit/63b9956faf4c3679c88a9401b8ccbb111c0294ee

petebacondarwin:
This change was actually fixing a **regression** that occurred between 1.3.5 and 1.3.6 (see https://github.com/angular/angular.js/issues/10419). I am sorry that it took so long for us to fix this regression that you started using the filter in its regressed state.

src/ng/filter/filter.js
petebacondarwin: @zuzusik - This change was actually fixing a regression that occurred between 1.3.5 and 1.3.6 (see #10419). I am sorry that it took so long for us to fix this regression that you started using the filter in its regressed state.

PR #10429, opened 2014-12-12
fix($rootScope): ignore properties beginning with `$` in $watchCollection object mode

### LINKED ISSUE #10426: Align behavior of deep $watch and $watchCollection
Deep $watch'es rely on angular.equals which ignore object properties that begin with '$'.  

Is there any reason $watchCollection can't follow this behavior too?  Currently, watch listeners are triggered when a property of an object is changed that begins w/ '$'.

caitp:
I'm not too keen on this, I think we want to fix angular.equals() to only ignore double-dollar-signs $$, which would make a lot more sense to do in $watchCollection. Up to you @petebacondarwin.

---

Make behaviour of $watchCollection (for non-ArrayLike objects) consistent with angular.equals, used by $watch when objectEquality flag is set to true.

BREAKING CHANGE:

Previously, enumerable properties beginning with `$` would be taken into consideration. Now, they are simply ignored.

Closes #10426

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -599,6 +599,12 @@ function $RootScopeProvider() {
             // copy the items to oldValue and look for changes.
             newLength = 0;
             for (key in newValue) {
+              // Emulate angular.equals(), which ignores keys beginning with
+              // a single dollar sign.
+              if (key.charAt(0) === '$') {
+                continue;
+              }
+
               if (newValue.hasOwnProperty(key)) {
                 newLength++;
                 newItem = newValue[key];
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I agree with @caitp  and @lgalfaso that we should change `angular.equals` not to ignore single `$` but it scares me since there may well be lots of things relying on `$` being "private" values that should be ignored...

petebacondarwin:
Closing this PR - let's have a discussion about the alternative change instead back on #10426

F1LT3R:
This is an interesting idea. I wrote something custom to get around the same issue.

http://plnkr.co/edit/zIUdSoPEtNXMJ29TSBDp?p=preview

PR #10430, opened 2014-12-12
added ng-jq functionality to be able to force jqLite instead of jQuery if present.

mboudreau:
made it very similar to csp() function as it should be global since angular.element is as well.  Added some unit tests around the jq() function, but still missing scenario testing which I need some direction on to complete.  This is a first attempt and hoping @mzgol can help out more to finish the contribution to the angular.js codebase.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -913,25 +913,36 @@ function equals(o1, o2) {
 }
 
 var csp = function() {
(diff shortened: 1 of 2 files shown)


mboudreau:
@mzgol Created a new pull request, so please go ahead and abandon this one:  https://github.com/angular/angular.js/pull/10750

PR #10432, opened 2014-12-12
WIP: perf(ngInclude): cache linkFns for each template

petebacondarwin:
This is probably not the best way to do this but here is a working Proof of Concept.

See http://www.bennadel.com/blog/2738-using-ngrepeat-with-nginclude-hurts-performance-in-angularjs.htm

Here is demo: http://plnkr.co/edit/Pf9DVKxNEAZQavcGm7Bu?p=preview


petebacondarwin:
Pushing to the backlog for now.

petebacondarwin:
I don't think we should do this. It is easy enough to get the caching you need for real application by using a directive with a templateURL instead (IMO).

PR #10467, opened 2014-12-15
feat($compile): allow using bindToController as object, support both new/isolate scopes

### LINKED ISSUE #10420: bindToController should support "scope" syntax/features
currently, if you want to use "bindToController", you have to do this:
```
bindToController: true,
scope: {
message: "@"
}
```

to simplify, bindToController should support the scope syntax and automatically set to true:

```
bindToController: {
message: "@"
}
```

This helps explain what is actually happening and simplifies your code so it should be a simple win/win.

caitp:
bindToController is now able to be specified as a convenient object notation:

```
bindToController: {
  text: '@text',
  obj: '=obj',
  expr: '&expr'
},
scope: {
  notBoundToCtrl: '@someOtherAttr'
}
```

It can also be used in conjunction with new scopes, rather than exclusively isolate scopes:

```
bindToController: {
  text: '@text',
  obj: '=obj',
  expr: '&expr'
},
scope: true
```

PING @petebacondarwin / @johnlindquist

I am not super keen on this, because it makes things much crazier than they used to be. That said:
- Currently, `bindToController` does not automatically imply an isolate scope --- maybe it should (I hate this though)
- It could probably do with more errors being thrown when people use it wrong
- The errors it does throw could be better
- The errors it throws are potentially breaking changes :(

Closes #10420

### DIFF
--- a/docs/content/error/$compile/noctrl.ngdoc
+++ b/docs/content/error/$compile/noctrl.ngdoc
@@ -4,65 +4,9 @@
 @description
 
 When using the `bindToController` feature of AngularJS, a directive is required
-to have a Controller, in addition to a controller identifier.
+to have a Controller. A controller may be specified by adding a "controller"
+property to the directive definition object. Its value should be either a
+string, or an invokable object (a function, or an array whose last element is a
+function).
 
-For example, the following directives are valid:
-
-```js
-// OKAY, because controller is a string with a label component.
-directive("okay", function() {
-  return {
-    bindToController: true,
-    controller: "myCtrl as $ctrl"
-    scope: {
-      text: "@text"
-    }
-  };
-});
-
-
-// OKAY, because the directive uses the controllerAs property to override
-// the controller identifier.
-directive("okay2", function() {
-  return {
-    bindToController: true,
-    controllerAs: "$ctrl",
-    controller: function() {
-
-    },
-    scope: {
-      text: "@text"
-    }
-  };
-});
-```
-
-While the following are invalid:
-
-```js
-// BAD, because the controller property is a string with no identifier.
-directive("bad", function() {
-  return {
-    bindToController: true,
-    controller: "unlabeledCtrl",
-    scope: {
-      text: "@text"
-    }
-  };
-});
-
-
-// BAD because the controller is not a string (therefore has no identifier),
-// and there is no controllerAs property.
-directive("bad2", function() {
-  return {
-    bindToController: true,
-    controller: function noControllerAs() {
-
-    },
-    scope: {
-      text: "@text"
-    }
-  };
-});
-```
+For more information, see the {@link guide/directive directives guide}.
--- a/docs/content/error/$compile/noctrl.ngdoc
+++ b/docs/content/error/$compile/noctrl.ngdoc
@@ -1,0 +1,68 @@
+@ngdoc error
+@name $compile:noctrl
+@fullName Controller is required.
+@description
+
+When using the `bindToController` feature of AngularJS, a directive is required
+to have a Controller, in addition to a controller identifier.
+
+For example, the following directives are valid:
+
+```js
+// OKAY, because controller is a string with a label component.
+directive("okay", function() {
+  return {
+    bindToController: true,
+    controller: "myCtrl as $ctrl"
+    scope: {
+      text: "@text"
+    }
+  };
+});
+
+
+// OKAY, because the directive uses the controllerAs property to override
+// the controller identifier.
+directive("okay2", function() {
+  return {
+    bindToController: true,
+    controllerAs: "$ctrl",
+    controller: function() {
+
+    },
+    scope: {
+      text: "@text"
+    }
+  };
+});
+```
+
+While the following are invalid:
+
+```js
+// BAD, because the controller property is a string with no identifier.
+directive("bad", function() {
+  return {
+    bindToController: true,
+    controller: "unlabeledCtrl",
+    scope: {
+      text: "@text"
+    }
+  };
+});
+
+
+// BAD because the controller is not a string (therefore has no identifier),
+// and there is no controllerAs property.
+directive("bad2", function() {
+  return {
+    bindToController: true,
+    controller: function noControllerAs() {
+
+    },
+    scope: {
+      text: "@text"
+    }
+  };
+});
+```
--- a/docs/content/error/$compile/noident.ngdoc
+++ b/docs/content/error/$compile/noident.ngdoc
@@ -1,0 +1,71 @@
+@ngdoc error
+@name $compile:noident
+@fullName Controller identifier is required.
+@description
+
+When using the `bindToController` feature of AngularJS, a directive is required
+to have a Controller identifier, which is initialized in scope with the value of
+the controller instance. This can be supplied using the "controllerAs" property
+of the directive object, or alternatively by adding " as IDENTIFIER" to the controller
+name.
+
+For example, the following directives are valid:
+
+```js
+// OKAY, because controller is a string with an identifier component.
+directive("okay", function() {
+  return {
+    bindToController: true,
+    controller: "myCtrl as $ctrl"
+    scope: {
+      text: "@text"
+    }
+  };
+});
+
+
+// OKAY, because the directive uses the controllerAs property to override
+// the controller identifier.
+directive("okay2", function() {
+  return {
+    bindToController: true,
+    controllerAs: "$ctrl",
+    controller: function() {
+
+    },
+    scope: {
+      text: "@text"
+    }
+  };
+});
+```
+
+While the following are invalid:
+
+```js
+// BAD, because the controller property is a string with no identifier.
+directive("bad", function() {
+  return {
+    bindToController: true,
+    controller: "noIdentCtrl",
+    scope: {
+      text: "@text"
+    }
+  };
+});
+
+
+// BAD because the controller is not a string (therefore has no identifier),
+// and there is no controllerAs property.
+directive("bad2", function() {
+  return {
+    bindToController: true,
+    controller: function noControllerAs() {
+
+    },
+    scope: {
+      text: "@text"
+    }
+  };
+});
+```
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -152,6 +152,9 @@
     "urlResolve": false,
     "urlIsSameOrigin": false,
 
+    /* ng/controller.js */
+    "identifierForController": false,
+
     /* ng/compile.js */
     "directiveNormalize": false,
 
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1373,15 +1373,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             if (nodeLinkFn.scope) {
               childScope = scope.$new();
               compile.$$addScopeInfo(jqLite(node), childScope);
-              var onDestroyed = nodeLinkFn.$$onScopeDestroyed;
-              if (onDestroyed) {
-                nodeLinkFn.$$onScopeDestroyed = null;
-                childScope.$on('$destroyed', function() {
-                  for (var i=0, ii = onDestroyed.length; i < ii; ++i) {
-                    onDestroyed[i]();
-                  }
-                  onDestroyed = null;
-                });
+              var destroyBindings = nodeLinkFn.$$destroyBindings;
+              if (destroyBindings) {
+                nodeLinkFn.$$destroyBindings = null;
+                childScope.$on('$destroyed', destroyBindings);
               }
             } else {
               childScope = scope;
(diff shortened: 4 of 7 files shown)


docs/content/error/$compile/noctrl.ngdoc
```diff
@@ -0,0 +1,68 @@
+@ngdoc error
+@name $compile:noctrl
+@fullName Controller is required.
```
petebacondarwin: Should this full name be more like: `"controller as" identifier required to bind to controller`?

docs/content/error/$compile/noctrl.ngdoc
```diff
@@ -0,0 +1,68 @@
...
+@description
+
+When using the `bindToController` feature of AngularJS, a directive is required
+to have a Controller, in addition to a controller identifier.
```
petebacondarwin: > When using the `bindToController` feature of AngularJS, a directive is required
to provide the `controller as` identifier, in addition to providing a controller.

src/ng/compile.js
```diff
@@ -738,6 +740,37 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+    if (isObject(directive.bindToController)) {
+      bindings.bindToController =
+          parseIsolateBindings(directive.bindToController, directiveName, true);
+    }
+    if (isObject(bindings.bindToController)) {
+      var controller = directive.controller;
+      var controllerAs = directive.controllerAs;
+      if (!directive.controller || !identifierForController(controller, controllerAs)) {
+        throw $compileMinErr('noctrl',
+              "Cannot bind to controller without directive '{0}'s controller.",
```
petebacondarwin: "Cannot bind to controller without directive '{0}'s controller and `controller as` identifier."

test/ng/compileSpec.js
```diff
@@ -3798,6 +3798,180 @@ describe('$compile', function() {
...
+        element = $compile('<div foo-dir dir-data="remoteData" ' +
+        'dir-str="Hello, {{whom}}!" ' +
+        'dir-fn="fn()"></div>')($rootScope);
+        $rootScope.$digest();
+        expect(controllerCalled).toBe(true);
+      });
+    });
+
+
+    it('should put controller in scope when labelled but not using controllerAs', function() {
```
petebacondarwin: I don't follow the description of this test. The code below does indeed use `controllerAs`.

caitp: It's not using the `controllerAs` property of the DDO.

docs/content/error/$compile/noctrl.ngdoc
gkalpak: If I am not mitaken, this error will also be thrown if there is `bindToController: true` but no `controler:...`. So the literals have to account for all cases (i.e. `bindToController: true` but no `controller` or no controller identifier.

docs/content/error/$compile/noctrl.ngdoc
gkalpak: Same as above.
<sub>
BTW, I like _"controller identifier"_ (or _"controller alias"_) better than _"`controller as` identifier"_, as the latter might be misunderstood as implying that a `controllerAs` property is mandatory (which is not the case).
Not a big issue though, since the examples below make it clear.
</sub>

src/ng/controller.js
```diff
@@ -1,5 +1,15 @@
 'use strict';
 
+var CNTRL_REG = /^(\S+)(\s+as\s+(\w+))?$/;
+function identifierForController(controller, ident) {
+  if (ident && isString(ident)) return ident;
+  if (isString(controller)) {
+    var match = CNTRL_REG.exec(controller);
+    if (match) return match[3];
+  }
+}
```
petebacondarwin: Since this function is only used in compile.js perhaps it should be there and we just allow the CNTRL_REG to be a global instead?

docs/content/error/$compile/noctrl.ngdoc
petebacondarwin: Yep, I see that now. I was thrown by the main body of the error document, which doesn't really mention that aspect of the error at all.
I wonder if we ought to have two separate errors?

docs/content/error/$compile/noctrl.ngdoc
petebacondarwin: I like `controller alias`

src/ng/compile.js
petebacondarwin: I think we probably need two different errors here. One for a missing controller and one for a missing controller alias.

test/ng/compileSpec.js
petebacondarwin: Ah right. I didn't get that "labelled" meant something and that `controllerAs` was referring to the DDO property rather than implying that the controller had an alias.

I am still not sure what is being tested here. Is it that having a `bindToController` property should not prevent the controller from being attached to the scope?

src/ng/compile.js
```diff
@@ -722,9 +722,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       if (!match) {
         throw $compileMinErr('iscp',
-            "Invalid isolate scope definition for directive '{0}'." +
+            "Invalid {3} for directive '{0}'." +
             " Definition: {... {1}: '{2}' ...}",
-            directiveName, scopeName, definition);
+            directiveName, scopeName, definition,
+            (isController ? "controller bindings definition" :
+            "isolate scope definition"));
```
petebacondarwin: nice

src/ng/controller.js
caitp: I think it's better to keep it here, and it's probably better to use it in controller.js too, just to make sure the same logic is used in both places

test/ng/compileSpec.js
caitp: `bindToController` is actually irrelevant here, but previously the controller label in the string was ignored. It's just simpler to support the controller-as label too, so it's testing that it works.

caitp: although I'm not super keen on it because it creates 2 ways to do a thing, but it can be worried about later

petebacondarwin:
Apart from splitting the error message, this looks good to me. @gkalpak do you have any further comments.

gkalpak:
:+1: for separate error messages.
<sub>
In the original PR, @caitp mentioned something about _potential leaks and performance regressions_.
I don't think it would be much of an issue (but I wouldn't know better); do you think it would make sense to run some benchmarks (if not already done so) ?
</sub>
Other than that it LGTM.

caitp:
if you want to release this today, i'll split up the error message and do whatever other cleanup in about 30 minutes.

caitp:
updated to have 2 different errors

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
+@ngdoc error
+@name $compile:noident
+@fullName Controller alias is required.
+@description
+
+When using the `bindToController` feature of AngularJS, a directive is required
+to have a Controller identifier. This can be supplied using the "controllerAs"
+property of the directive object, or alternatively by adding " as ALIAS" to the
```
gkalpak: directive _definition_ object

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
...
+controller name (in either the "controller" property, or where the controller
+was registered).
```
gkalpak: > or where the controller was registered

(With the risk of embarrassing myself :confused:) what do you mean by that ?

petebacondarwin: I am not sure. I don't think you can specify `controllerName as alias` when defining a controller. The only other time would be inside the `ngController` directive, where the controller name is `'@'` and so we get the controller from the `ng-controller` attribute value, which obviously can also have the `controllerName as alias`.

This phrase should either be clarified or removed.

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
+@ngdoc error
+@name $compile:noident
+@fullName Controller alias is required.
+@description
+
+When using the `bindToController` feature of AngularJS, a directive is required
+to have a Controller identifier. This can be supplied using the "controllerAs"
```
petebacondarwin: Let's standardize on always calling this identifier an **alias** throughout the code and the docs.

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
+@ngdoc error
+@name $compile:noident
```
petebacondarwin: `$compile:noalias`

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
...
+When using the `bindToController` feature of AngularJS, a directive is required
+to have a Controller identifier. This can be supplied using the "controllerAs"
+property of the directive object, or alternatively by adding " as ALIAS" to the
+controller name (in either the "controller" property, or where the controller
+was registered).
+
+For example, the following directives are valid:
+
+```js
+// OKAY, because controller is a string with a label component.
```
petebacondarwin:
```js
// OKAY, because controller is a string, with an alias component.
```

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
...
+    controller: "myCtrl as $ctrl"
+    scope: {
+      text: "@text"
+    }
+  };
+});
+
+
+// OKAY, because the directive uses the controllerAs property to override
+// the controller identifier.
```
petebacondarwin:
```js
// the controller alias
```

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
...
+      text: "@text"
+    }
+  };
+});
+```
+
+While the following are invalid:
+
+```js
+// BAD, because the controller property is a string with no identifier.
```
petebacondarwin:
```js
// BAD, because the controller property is a string with no alias.
```

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
...
+directive("bad", function() {
+  return {
+    bindToController: true,
+    controller: "unlabeledCtrl",
```
petebacondarwin:
```js
controller: "unaliasedCtrl",
```

docs/content/error/$compile/noident.ngdoc
```diff
@@ -0,0 +1,71 @@
...
+    scope: {
+      text: "@text"
+    }
+  };
+});
+
+
+// BAD because the controller is not a string (therefore has no identifier),
```
petebacondarwin:
```js
// BAD because the controller is not a string (therefore has no alias),
```

src/.jshintrc
```diff
@@ -153,6 +153,9 @@
     "urlResolve": false,
     "urlIsSameOrigin": false,
 
+    /* ng/controller.js */
+    "identifierForController": false,
```
petebacondarwin:
```js
"aliasForController": false,
```

src/ng/compile.js
```diff
@@ -738,6 +740,43 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+                                                     directiveName, false);
+      }
+    }
+    if (isObject(directive.bindToController)) {
+      bindings.bindToController =
+          parseIsolateBindings(directive.bindToController, directiveName, true);
+    }
+    if (isObject(bindings.bindToController)) {
+      var controller = directive.controller;
+      var controllerAs = directive.controllerAs;
```
petebacondarwin:
```js
var controllerAlias = directive.controllerAs;
```

src/ng/compile.js
```diff
@@ -738,6 +740,43 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+      if (!controller) {
+        // There is no controller, there may or may not be a controllerAs property
+        throw $compileMinErr('noctrl',
+              "Cannot bind to controller without directive '{0}'s controller.",
+              directiveName);
+      } else if (!identifierForController(controller, controllerAs)) {
```
petebacondarwin:
```js
 } else if (!identifierForController(controller, controllerAlias)) {
```

src/ng/controller.js
```diff
@@ -1,5 +1,15 @@
 'use strict';
 
+var CNTRL_REG = /^(\S+)(\s+as\s+(\w+))?$/;
+function identifierForController(controller, ident) {
```
petebacondarwin: `aliasForController`

src/ng/compile.js
gkalpak: Or even go one step further and rename `identifierForController` to `aliasForController` :P

src/ng/controller.js
petebacondarwin: In that case this function needs refactoring to return the controller name as well, since that is what is needed in $controller.

All the references to `label`, `ident` and `identifier` in this file ought to be changed to `alias` for consistency.  If you don't like `alias` then you can choose an alternate and change all the references to `ident`, `identifier`, `alias` and `label` to a single consistent term throughout the codebase.

test/ng/compileSpec.js
petebacondarwin: OK, we need to change `labelled` to `has a controller alias` or something like that for consistency. It is only now that I realise that `label` has been used interchangeably with `alias` and `identity` throughout the source.

src/ng/compile.js
petebacondarwin: :-P - quite right! I knew I would miss one...

petebacondarwin:
@caitp  - can you take a look at the comments here and update this PR? Thanks

caitp:
I don't think we should be calling it "alias", because it's misleading --- so I'm just going to ignore all of the comments about renaming "identifier" to "alias"

petebacondarwin:
@caitp - it is fine for you not to think we should call it alias. As I explained in [this comment](https://github.com/angular/angular.js/pull/10467/files#r23216369), what we do need is consistency.

I assume that your preferred name is `identifier`. In which case please go through and make sure that this is used throughout. In particular remove any use of `alias` and `label`.

frfancha:
@caitp 
You say "the important thing is that people generally like controllerAs and want to use it".
I wonder why. In our experiences directives with controller are mush slower than directives with just link function and scope access, especially used under ng-repeat

caitp:
this is ready to merge

petebacondarwin:
Great! LGTM! Thanks @caitp. I will merge tomorrow unless you want to do it today.

caitp:
I've got it, thanks for reviewing

PR #10470, opened 2014-12-15
fix(form): ignore properties in prototype chain in $error

### LINKED ISSUE #10469: Angular have problems with Object.prototype
The bug stated (and is **not yet fixed** as of 1.3.6) when upgraded from **1.3.0-rc.0** to **1.3.0-rc.1**. Sorry to report this late, but I couldn't figure it out.
The bug is easily reproduced by:
Object.prototype.x = function(){};

I think this is a bug, since prevent the user from using prototypes.

caitp:
Closes #10469

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -263,7 +263,7 @@ function forEach(obj, iterator, context) {
         obj.forEach(iterator, context, obj);
     } else {
       for (key in obj) {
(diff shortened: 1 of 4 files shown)


lgalfaso:
there were some refactors on `input.js` and `inputSpec.js` can you please rebase this?

PR #10502, opened 2014-12-17
fix($compile): respect return value from controller constructor

caitp:
The return value of the controller constructor is now respected in all cases.

If controllerAs is used, the controller will be re-bound to scope. If bindToController is used, the previous binding $watches (if any) will be unwatched, and bindings re-installed on the new controller.

Blocked on #10467 --- This potentially results in leaks and performance regressions, good review is needed here.


petebacondarwin:
@caitp and @gkalpak is this ready to go?

caitp:
Version of this landed in https://github.com/angular/angular.js/commit/62d514b06937cc7dd86e973ea11165c88343b42d already

PR #10507, opened 2014-12-17
feat(angular.extend) - Simple non-breaking change to support deep extend

prettycode:
People use `angular.extend()` as a utility function, whether or not it's designed for that purpose. It can easily be extended to support recursive merging with very very minor additions. Developers coming to Angular expect `extend()` to support this.


petebacondarwin:
Closing in favour of #10519

PR #10519, opened 2014-12-18
feat(extend): optionally deep-extend when last parameter is `true`

caitp:
Same as #10507 but with my review comments addressed and jscs fixed.

Someone quickly review this so we can land it :<


petebacondarwin:
I'll tweak the tests and land this today.

PR #10525, opened 2014-12-19
feat($time): create time service

### LINKED ISSUE #10402: $time service
I would be nice to have a simple ```$time { now(): number }``` service. The now method would be a simple forward to Date.now()
This would allow cleaner and more testable code relying on timeout, calendar, etc. A mockable version would allow to set current time.

marcin-wosinek:
Add simple $time service to allow easier mocking dates in applications.

Closes #10402


lgalfaso:
@manuel-woelker sorry for the late response. Even when the PR looks good if we are just looking at the core, I think it is important that there is a reasonable counterpart for ng-mock. This is going to take a little bit more time to get done right, but I would still would like to aim at 1.4

marcin-wosinek:
Could you provide example of what should this ng-mock counterpart do? I can think only of some $date.setTime(), but this can be acomplished by just mocking the normal fuction

lgalfaso:
Here is the first draft
https://docs.google.com/document/d/1nkTDEd6yaV9KWEqXEuqHjFzt3WtZ_WNVFC1vA1la0ts/edit#heading=h.csrxcq4wx70q
>
>
Note: This is still an early draft so it may change entirely or not be done
at all

marcin-wosinek:
Ah, that looks like a pretty big change. Is there separate ticket for that? There is #10402, but it's defined much simpler.

btw
> Keep a unified queue of elements that are triggered thru $timeout and $interval
should this be something like
0 => now
+100ms => interval 1
+200ms => interval 2
+250ms => timeout

So the queue will know that there should be two intervals fired before timeout?

lgalfaso:
> So the queue will know that there should be two intervals fired before timeout?

yes, and also that `$time.now()` (or however we call it) also moves forward the right number of ms

petebacondarwin:
This should land in 1.5 - @lgalfaso can you manage it?

lgalfaso:
> This should land in 1.5 - @lgalfaso <https://github.com/lgalfaso> can you
> manage it?
>


I will take care

src/AngularPublic.js
```diff
@@ -80,6 +80,7 @@
   $TemplateCacheProvider,
   $TemplateRequestProvider,
   $$TestabilityProvider,
+  $TimeProvider,
```
gkalpak: Shouldn't this be `$DateProvider` ?

gkalpak:
@lgalfaso, I am kind of confused. Is there some followup planned with more features ?

From the discussion above, I got the impression this would be more feature-rich/involved.
In the current state it's essentially a wrapper for `Date()`. What is the benefit of using `$date` over just using `$window.Date` ?

petebacondarwin:
When I said "this should land", I hadn't actually reviewed this, by the way ;-P

lgalfaso:
Let me revert this
>
>

gkalpak:
> When I said "this should land", I hadn't actually reviewed this, by the way ;-P

Wait...whaat ?! We should actually **review** the PRs before LGTM'ing ? :stuck_out_tongue:

petebacondarwin:
'This Should Land In One Point Five' (TSLIOPF) does not start with the letters L.G.T.M, as far as I can tell.
For the avoidance of doubt: where I said such a thing I really meant that we should work on the issue or pull request once we are in 1.5 mode.

petebacondarwin:
@lgalfaso - we probably don't need a revert just a bit of additional polish, right?

lgalfaso:
reverted this with 7dcfe5e03ee0943f92791a0e066ce42636ab1b58

Issue #8324, opened 2014-07-24
Two equal cookies cause of subdomain handling


qstiegler:
On one of our Projects, we´ve got a cookie with the Session ID for the whole domain, so we can also access this id from the subdomains. But if i´m working with $cookies or $cookieStore. It doesn´t look at the domain. So when I change, add or delete another cookie on a subdomain, the function create a duplicate of the Session ID cookie for the subdomain, too.

Please fix this, that if a cookie belongs to another subdomain or like in this case to the main domain, it won´t create a duplicate.

Issue #7631, opened 2014-05-29
$cookies doesn't represent current state of document.cookie


amaltson:
In an `$http` login call that we make, the resulting header has a `Set-Cookie`. While the cookie does get set in the browser, i.e. I can see it in `document.cookie`, it doesn't make it's way to `$cookies`.

Below is the response header setting the cookie:
![response-header-cookie](https://cloud.githubusercontent.com/assets/167443/3122303/ca255f4c-e765-11e3-8248-d459c4e194c0.png)

And here is `document.cookie` indicating the cookie exists but `$cookies` doesn't see it.
![nothing-in-cookies](https://cloud.githubusercontent.com/assets/167443/3122307/d87924d4-e765-11e3-8d62-76ef00cb61c6.png)

The strangest thing is on the next Angular action I take, `$cookies` has the cookie. The example above is already wrapping the `$cookies` check in a `$timeout` to give the browser time to digest.

The workaround right now is to parse `document.cookie` and set `$cookies` manually, but this is super hacky.

amaltson:
If `$browser.cookies()` is the new hotness, I'll close off this issue.

mkonecny:
> If `$browser.cookies()` is the new hotness, I'll close off this issue.

It's not the new hotness. Just upgraded from Angular 1.3 -> 1.4 and the function no longer exists.

Issue #6411, opened 2014-02-23
$cookies need callback once cookie value is set.


manishtpatel:
Right now if you set $cookies values then call $http immediately, cookie is not updated in the request. if $cookies returns before saving the cookie in the browser then we need callback to execute code that relies on cookie.

$cookies.myName = 'foo'

// request headers are missing myName cookie update from above line
$http.get('/api/getmystar')

To workaround this issue i have to use setTimeout to wait for some time before i can proceed.

It either need to not set the cookie asynchronously or provide a call back function, call back is not possible with $cookies but $cookieStore atleast should solve this issue. 

## summary
cookie writes are flushed asynchronously, which makes it hard to predict when the cookie will be picked by other browser apis like XMLHttpRequest used by $http

darkbasic:
The bug has been closed but $cookies.get() on AngularJS 1.4.7 still doesn't return a promise: how am I supposed to do?

ZachMoreno:
@darkbasic +1

```
$cookies.get('authentication').then(function(authCookie) {
    var userID = authCookie.id;
});
```
returns
```
TypeError: $cookies.get(...).then is not a function
```
in AngularJS v1.4.7

lgalfaso:
@ZachMoreno `$cookies.get('authentication')` is a synchronous function

firatkucuk:
I confirm this issue. It must be an async function. It refreshes too late.

asim-qb:
Any update on this? Having a timeout is the only workaround?

gkalpak:
As already mentioned, the `$cookies` methods are synchronous, so you shouldn't need a timeout. If you are running into problems, please submit a new issue providing a reproduction of the problem, so we can investigate).

WolfElkan:
[Here](https://github.com/WolfElkan/AngularUsers/commit/ed2877f5a52b1d4851fa19f40f7774daeddb681c) is a reproduction of the problem.  Loading this server creates a login/registration application but the page must be refreshed after clicking Login or Logout before the new Partial will be loaded.

gkalpak:
@WolfElkan, this does not demonstrate that `$cookies.set()` is asynchronous. The code you linked to doesn't even use `$cookies.set()` afaict.

OrganicCat:
I run into this issue if I have an interceptor. So the basic code goes like so:

```
$cookies.putObject('wowcookie', mycookie);
            
$state.go('home');
```

[...]

Issue #3988, opened 2013-09-12
deleting cookie from $cookies doesn't delete cookie in IE


jnizet:
Here's the situation:

The application is served from the root of a web server.
The index.html doesn't have a `<base>` tag. The server sets a session cookie named XSRF-TOKEN with a path set to `/`. This cookie is visible by the JS code, and is present in the $cookies object. But calling `delete $cookies['XSRF-TOKEN']` or `$cookies['XSRF-TOKEN'] = undefined` does not delete the cookie from the browser.

If, in debug mode, I set the cookiePath variable to '/' (in angular.js, in the cookies API), then the cookie is deleted.

If I add a `<base href="/" />` to index.html, then the cookie is also deleted (but I don't see why I would need a base tag when the app is served from the root).

If I don't set the path of the cookie to `/`, it isn't visible from JavaScript.

If I change the following line in angular.js :

    rawDocument.cookie = escape(name) + "=;path=" + cookiePath + ";expires=Thu, 01 Jan 1970 00:00:00 GMT";

to

    rawDocument.cookie = escape(name) + "=;expires=Thu, 01 Jan 1970 00:00:00 GMT"; 

then the cookie is deleted. But I'm not sure what the goal and importance of the path attribute is in this line. 

This is with angular 1.2.0-rc.1.

PR #10530, opened 2014-12-19
ng-cookies

shahata:
This is an initial suggestion for dealing with the current issues with angular cookies. The main things added by this:
1. An `options` parameter was added to `$cookieStore.put()` which supports passing `path`, `domain`, `expires` and `secure` parameters.
2. `$cookieStore` was refactored so it will use `$browser.cookies()` directly instead of going through `$cookies`. This was in order to allow direct update/read of cookies instead of going through polling/digest mechanism.
3. `$cookies` was deprecated in order to stop using the polling/digest mechanism and instead `getRaw`, `getAll` and `putRaw` methods were added to `$cookieStore` in order to accommodate some of the stuff that were previously available only when using `$cookies`

This is obviously just my take on this and completely open for discussion.

Some open issues:
1. The cookie setting logic is currently located in `$browser` although it is not used internally by angular (angular only uses the cookie reading logic for `XSRF-TOKEN`). It might be better to just move it into `ngCookies`. On the other hand maybe it should be left in `$browser` for balance, I don't know.
2. Theoretically it is possible to leave `$cookies` alive and make its properties getters/setters using `Object.defineProperty` (assuming everyone else agrees that polling/digest mechanism is bad), but we would still need some sort of polling for learning about new cookies. I think we'd better just deprecate it.
3. I'm somewhat uncomfortable with the `getRaw`/`get` and `putRaw`/`put` duplication. Another option can be to use something like `isJsonLike` in `$http` or add a flag to `get`/`put` or use `try`/`catch` or maybe even configure this behavior in a `$cookieStoreProvider`, but I don't like it very much. Since `getRaw`/`putRaw` are consistent with `getAll` (which must be raw imo), I like this option best.
4. Might be good to add a `$cookieStoreProvider` in order to give better defaults for options, I dunno.


shahata:
@IgorMinar did you get a chance to look at the rest of the PR?

lgalfaso:
I wonder if it would be best to have a super-light version that has the minimum to make $http work at $browser, and move everything else to ng-cookies

shahata:
Yeah, I mentioned that in the first "open issue" in my comment above. The cookie writing logic was originally added in `$browser` only for completeness (imo) and can very easily be moved to `ngCookies`. It would be a little strange though that the cookie reading is in `angular.js` and the writing is in `angular-cookies.js`, I'm not sure. Whether the cookie reading logic is in `$browser` or `$http` makes little difference to me, but I think it is better to leave it where it is.

IgorMinar:
@shahata finished the review of code as is now. some more changes to be done:

- [ ] move cookie parsing logic from $browser to ngCookies module
- [ ] remove the generic polling functionality from $browser (it's private api so it can be simply deleted)
- [ ] the refactoring as it stands today, removes the ability to be notified of cookie changes. This is why the polling functionality was added many years ago. We should consider if it's ok to drop this feature (and document why) or alternatively add the feature back in a more lightweight format (as an opt-in or something)

shahata:
@IgorMinar per your comments:
 1. There is the issue of `$http` requiring the cookie parsing logic, do you have any suggestion regarding this?
 2. Gladly :D
 3. Do we have a good use case where someone would want to poll for cookie changes? The way I see it, people were previously able to set a `scope.$watch` on cookies and count on the poller to trigger digest when they change, but actually cookies only change when the application sets them manually or due to http request (on both option a digest will be triggered anyway, so no need for poller to trigger it). Of course there are cases where people will touch the cookies "out of digest", but I think that it is very rare and those people can write their own polling/notification mechanism...

shahata:
Okay, I just talked with @IgorMinar about this, here are the conclusions:

 1. We are going to remove to cookie polling part completely along with the generic polling mechanism in `$browser`. We don't see a lot of value in the current polling mechanism - the reason this was originally added was to allow communication between different tabs, but there are better ways to do this today (for example localStorage). If anyone will still need it I guess it is pretty easy to implement the polling on his own. This change and the reasoning behind it needs to be documented.
 2. We are going to break the cookie logic into two services: `$$cookieReader` and `$$cookieParser`. `$$cookieReader` will remain in angular core (since `$http` needs it) and `$$cookieWriter` will move into ng-cookies.
 3. We are going to move the cookie API to `$cookies` instead of `$cookieStore` and then deprecate `$cookieStore` (it will still have all of its current API which will invoke `$cookies` internally just so that we don't break apps that currently use `$cookieStore`). The new API on $cookies will include the following methods: `get`, `put`, `getObject`, `putObject`, `getAll`. This is obviously a breaking change since `$cookies` will no longer have the cookies as properties like it has today.
 4. We'll add an option to set defaults for cookie setting options (`path`, `domain`, `expires`, `secure`) using `$cookiesProvider`. This will help people who, for example, want to set all of their cookies as `secure`.
 5. The default for the `path` option when setting a cookie is the value you have in your `<base>` tag. This is important since if you load an app at `https://myapp.com/my/subview/234`, you usually want the cookies to be scoped to the whole app and not to `/my/subview/234`. This should be documented.

petebacondarwin:
These conclusions sound about right.

* @lgalfaso suggested that the ngCookies module should be moved into its own file, similar to how ngAnimate works.

shahata:
This PR is now updated with the conclusions listed above

/cc @IgorMinar @petebacondarwin @lgalfaso

src/ngCookies/cookieWriter.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @param {string} name Cookie name
+ * @param {string=} value Cookie value (if undefined, cookie will be deleted)
+ * @param {Object=} options Object with options that need to be stored for the cookie.
+ */
+function $$CookieWriter($document, $log, $browser) {
+  var cookiePath = $browser.baseHref();
+  var rawDocument = $document[0];
+  var lastCookies = {};
+  var lastCookieString = '';
```
gkalpak: These 2 variables seem unused. Could be removed.

src/ngCookies/cookies.js
```diff
@@ -19,180 +19,164 @@
...
-   *     // Removing a cookie
-   *     $cookieStore.remove('myFavorite');
-   *   }]);
-   * ```
-   */
-   factory('$cookieStore', ['$cookies', function($cookies) {
-
+   * Use `$cookiesProvider` to change the default behavior of the {@link ngCookies.$cookies $cookies} service.
+   * */
+   provider('$cookies', [function() {
```
gkalpak: It would be nice to have a named function.

test/ng/cookieReaderSpec.js
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+
+  afterEach(function() {
+    deleteAllCookies();
+    expect(document.cookie).toEqual('');
+  });
+
+
+  describe('get via cookies()[cookieName]', function() {
```
gkalpak: Shouldn't this rather be `$$cookieReader()[cookieName]`

test/ngCookies/cookieWriterSpec.js
```diff
@@ -0,0 +1,194 @@
+/* global $$CookieWriter */
```
gkalpak: Same here: Couldn't this be DI-ed instead ?

test/ngCookies/cookiesSpec.js
```diff
@@ -1,149 +1,130 @@
...
+      expect($$cookieWriter).toHaveBeenCalledWith('name', 'value', {secure: false});
+    });
+  });
 
-  it('should deserialize json to object', inject(function($cookieStore, $browser) {
-    $browser.cookies('objectCookie', '{"id":123,"name":"blah"}');
-    $browser.poll();
-    expect($cookieStore.get('objectCookie')).toEqual({id: 123, name: 'blah'});
+  it('should put cookie value without serializing', inject(function($cookies, $$cookieReader) {
+    $cookies.put('name', 'value');
```
gkalpak: I think it would be safer to have a serialized object as value.

test/ngCookies/cookiesSpec.js
```diff
@@ -1,149 +1,130 @@
...
-    $cookieStore.put('gonner', { "I'll":"Be Back"});
-    $rootScope.$digest(); //force eval in test
-    $browser.poll();
-    expect($browser.cookies()).toEqual({'gonner': '{"I\'ll":"Be Back"}'});
-
-    $cookieStore.remove('gonner');
-    $rootScope.$digest();
-    expect($browser.cookies()).toEqual({});
+  it('should get cookie value without deserializing', inject(function($cookies, $$cookieWriter) {
+    $$cookieWriter('name', 'value');
```
gkalpak: Same here.

gkalpak:
:+1: :+1: @shahata

shahata:
@gkalpak thanks for the comments!
@IgorMinar WDYT?

pkozlowski-opensource:
@shahata @petebacondarwin @IgorMinar  we need to get this in before ng-conf, ideally before the next release. Are there any blocking points left? If so, what are those blocking points and who could help unblocking those?

@shahata the one obvious thing is that the thing doesn't merge cleanly any more. Could you please rebase it? Do we want to squash commits before merging?

shahata:
@pkozlowski-opensource I've rebased this, thanks. Not sure regarding squashing, it is quite a big change and each commit is an incremental step towards it, so I think it is much easier to understand this way, but if anyone has a strong opinion about this, I don't mind squashing.

@petebacondarwin WDYT? I think it is time to land this... (after travis is green obviously)

pkozlowski-opensource:
ng-conf is 5 days, just saying :-)

petebacondarwin:
I'll look at this in the next two days.

src/ngCookies/cookieStore.js
```diff
@@ -0,0 +1,80 @@
...
+   *     // Get cookie
+   *     var favoriteCookie = $cookieStore.get('myFavorite');
+   *     // Removing a cookie
+   *     $cookieStore.remove('myFavorite');
+   *   }]);
+   * ```
+   */
+   factory('$cookieStore', ['$cookies', function($cookies) {
+
+      return {
```
petebacondarwin: I wonder if we should write a deprecation message to the console when this service is instantiated?

src/ngCookies/cookieWriter.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {string} name Cookie name
+ * @param {string=} value Cookie value (if undefined, cookie will be deleted)
+ * @param {Object=} options Object with options that need to be stored for the cookie.
+ */
+function $$CookieWriter($document, $log, $browser) {
+  var cookiePath = $browser.baseHref();
+  var rawDocument = $document[0];
+
+  function buildCookieString(name, value, options) {
+    options = angular.extend({path: cookiePath}, options);
```
petebacondarwin: Rather than use `extend` here we could simply set a `path` variable and use the rest of the `options` object as-is...


```js
var path = angular.isDefined(options.path) ? options.path : cookiePath;
```

src/ngCookies/cookies.js
```diff
@@ -19,180 +19,164 @@
...
+     *
+     * @example
+     *
+     * ```js
+     * angular.module('cookiesExample', ['ngCookies'])
+     *   .controller('ExampleController', ['$cookies', function($cookies) {
+     *     // Retrieving a cookie
+     *     var favoriteCookie = $cookies.myFavorite;
+     *     // Setting a cookie
+     *     $cookies.myFavorite = 'oatmeal';
```
petebacondarwin: This example is now invalid, yes?

src/ngCookies/cookies.js
```diff
@@ -19,180 +19,164 @@
...
+        /**
+         * @ngdoc method
+         * @name $cookies#put
+         *
+         * @description
+         * Sets a value for given cookie key
+         *
+         * @param {string} key Id for the `value`.
+         * @param {string} value Raw value to be stored.
+         * @param {Object=} options Options object.
```
petebacondarwin: Add a link to the `$cookiesProvider` for what `options` are available.

src/ngCookies/cookies.js
```diff
@@ -19,180 +19,164 @@
...
+         *    The object may have following properties:
+         *
+         *    - **path** - `{string}` - The cookie will be available only for this path and its
+         *      sub-paths. By default, this would be the URL that appears in your base tag.
+         *    - **domain** - `{string}` - The cookie will be available only for this domain and
+         *      its sub-domains. For obvious security reasons the user agent will not accept the
+         *      cookie if the current domain is not a sub domain or equals to the requested domain.
+         *    - **expires** - `{string|Date}` - String of the form "Wdy, DD Mon YYYY HH:MM:SS GMT"
+         *      or a Date object indicating the exact date/time this cookie will expire.
+         *    - **secure** - `{boolean}` - The cookie will be available only in secured connection.
```
petebacondarwin: Instead of listing all these properties here again, just add a link to the `$cookiesProvider` for what `options` are available.

src/ngCookies/cookies.js
```diff
@@ -19,180 +19,164 @@
...
         /**
          * @ngdoc method
-         * @name $cookieStore#remove
+         * @name $cookies#remove
          *
          * @description
          * Remove given cookie
          *
          * @param {string} key Id of the key-value pair to delete.
+         * @param {Object=} options Object with options for the cookie to be removed.
```
petebacondarwin: Add a link to the `$cookiesProvider` for what `options` are available.

petebacondarwin:
Nice work @shahata ! Just a bit of tweaking then we can merge this, eh?

**TO DO:**
* [ ] optimize/remove calls to `extend`
* [ ] locally mock `$$cookieReader/Writer` for `$cookie` tests but continue to use the actual `document.cookie` object when testing lower level private `$$cookieReader` and `$$cookieWriter` services
* [ ] locally mock `$$cookieReader` for `$http` tests
* [ ] fix up docs

**COMMITS**
I suggest that we squash this down into the following set of commits, which would make sense to me if viewed in the change log. Did I miss anything?


=============

**refact($cookies): split cookie access into $$cookieReader and $$cookieWriter services**


===================


**fix($cookies): move logic into `$cookies` and deprecate `$cookieStore`**

The new API on `$cookies` includes:

* `get`
* `put`
* `getObject`
* `putObject`
* `getAll`,
* `remove`

The `put`, `putObject` and `remove` methods now take an `options` parameter
where you can provide additional options for the cookie value, such as `expires`,
`path`, `domain` and `secure`.


DEPRECATION NOTICE:

`$cookieStore` is now deprecated as all the useful logic
has been moved to `$cookies`, to which `$cookieStore` now simply
delegates calls.

BREAKING CHANGE:

The `get` method now returns the raw cookie value as a string rather than a
deserialized object. Use `getObject` if this is what you require.

The `put` method now takes a string, which will be set as the raw cookie value,
rather than an object to be serialized. Use `putObject` if this is what you require.

Closes #8324
Closes #3988
Closes #1786
Closes #950

==================


**fix($cookies): stop polling the browser's cookies values**

No longer poll the browser for changes to the cookies and no longer copy
cookie values onto the `$cookies` object.

The polling is expensive and caused issues with the `$cookies` properties not
synchronizing correctly with the actual browser cookie values.

The reason the polling was originally added was to allow communication between
different tabs, but there are better ways to do this today (for example `localStorage`).


BREAKING CHANGE:

`$cookies` no longer exposes properties that represent the current browser cookie
values. Now you must explictly call `get` or `getObject` to access the cookie
values. This also means that you can no longer watch the `$cookies` properties for
changes to the browser's cookies.

This feature is generally only needed if a 3rd party library was programmatically
changing the cookies at runtime. If you rely on this then you must either write code that
can react to the 3rd party library making the changes to cookies or implement your own polling
mechanism.


Closes #6411
Closes #7631

=================


**feat($cookiesProvider): provide `path`, `domain`, `expires` and `secure` options**

This change provides properties on `$cookiesProvider` so that you can set the application
level default options for cookies that are set using the `$cookies` service


=================


**fix($cookies): set the default path for $cookies from the `<base>` tag**

This is important since if you load an app at
`https://myapp.com/my/subview/234`, you usually want the cookies to be
scoped to the whole app and not to `/my/subview/234`.


=================


**refact($http): use $$cookieReader for XSRF support**


=================


**fix($browser): remove private polling mechanism**

The only feature of Angular using this mechanism was `$cookies`,
which no longer mirrors the browser cookie values and so does not
need to poll.

shahata:
@petebacondarwin @IgorMinar - I fixed and pushed all comments. I'll try to finish squashing tomorrow night.

petebacondarwin:
Great! Thanks @shahata - if you don't have time to do the squashing I can do this today.

src/ngCookies/cookieStore.js
```diff
@@ -0,0 +1,80 @@
...
+   * @requires $cookies
+   *
+   * @description
+   * Provides a key-value (string-object) storage, that is backed by session cookies.
+   * Objects put or retrieved from this storage are automatically serialized or
+   * deserialized by angular's toJson/fromJson.
+   *
+   * Requires the {@link ngCookies `ngCookies`} module to be installed.
+   *
+   * <div class="alert alert-error">
```
IgorMinar: I think you want to add `@deprecated` annotation to this service

petebacondarwin:
Closing in favour of #11222

PR #10538, opened 2014-12-20
fix(ngRepeat): do not sort object keys alphabetically

### LINKED ISSUE #6210: ngRepeat iterates over object in alphabetical order
This is very unexpected, all major browsers retain order of inserted keys and it is not going to change in any way in future as it would break a lot of existing code. 

No matter of that, angular does it its own "way".

Basically it makes this feature **completely useless** as you can't use it for example for iterating computed averages, and similar great things.

This is major flaw in design and fixing it involves removing **one line**:

    collectionKeys.sort();

Can we at least discuss this? I am highly convinced this (possibly) breaking change should affect minimum people and if orderBy worked on objects too, fixing may be very simple on their side.

caitp:
BREAKING CHANGE:

Previously, the order of object properties was consistent and simple. Now, it's not guaranteed to
be any particular order at all.

The best approach is to convert Objects into Arrays by a filter such as https://github.com/petebacondarwin/angular-toArrayFilter
or some other mechanism, and then sort them manually in the order you need.

Closes #6210

### DIFF
--- a/src/ng/directive/ngRepeat.js
+++ b/src/ng/directive/ngRepeat.js
@@ -335,14 +335,13 @@ var ngRepeatDirective = ['$parse', '$animate', function($parse, $animate) {
             trackByIdFn = trackByIdExpFn || trackByIdArrayFn;
           } else {
             trackByIdFn = trackByIdExpFn || trackByIdObjFn;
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I added documentation and landed. Thanks to everyone who has had input in this issue.

PR #10548, opened 2014-12-22
feat(filterFilter): compare object with custom `toString()` to primitive

### LINKED ISSUE #10464: filterFilter compare custom objects in 1.3.6
First off all I am a big fan of the changed behavior as I can do finally filtering complex data as I could in angular 1.0.x

Only the default comparator should not return false if  actual or expected is an object.

My use case are mongodb ObjectIds with a custom toString function.
From angular 1.0 to 1.3.5 the filterFilter could compare this custom object with a string or another ObjectId.

I know I can use my one comparator but i think the new comparator is a bit to picky.
A string like `'[object'` is in my opinion a user error.

WDYT?

m7r:
Closes #10464

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -54,6 +54,9 @@
  *   - `false|undefined`: A short hand for a function which will look for a substring match in case
  *     insensitive way.
  *
+ *     Primitive values are converted to strings. Objects are not compared against primitives,
+ *     unless they have a custom `toString` method (e.g. `Date` objects).
+ *
  * @example
    <example>
      <file name="index.html">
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@m7r I've left some additional comments to the tests - not a big deal, I can change those while merging, but would like to have your feedback on those.

m7r:
@pkozlowski-opensource why is the milestone set from 1.3.9 to 1.4?

m7r:
@pkozlowski-opensource what are the next steps?

petebacondarwin:
I tweaked this and landed it. Thanks @m7r

PR #10567, opened 2014-12-24
docs(guide/Directives): data from isolate to parent scope

toastynerd:
It looks like this used to be in the Angular docs as per this thread: https://groups.google.com/d/msg/angular/3CHdR_THaNw/AxqKwUw5t0oJ I recently spent some time trying to get this to work and was very frustrated by lack of documentation. I tried to tie this into the existing &attr example but let me know if that is not desirable.


petebacondarwin:
@toastynerd - please can you sign the CLA? Thanks

PR #10577, opened 2014-12-25
Screening colon in given path.

stervets:
For example path "/:foo/:bar" returns only "bar" variable


lgalfaso:
Hi. Can you please fix the indentation to two spaces? BTW, if you can follow the [commit message format](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format), it would make things easier. Thanks!

PR #10592, opened 2014-12-28
refactor($parse): new and faster $parse

lgalfaso:
Change the way parse works from the old mechanism to a multiple stages
parsing and code generation. The new parse is a four stages parsing
- Lexer
- AST building
- AST processing
- Cacheing, one-time binding and `$watch` optimizations

The Lexer phase remains unchanged.

AST building phase follows Mozilla Parse API [1] and generates an AST that is compatible. The only exception was needed for `filters` as JavaScript does not support filters, in this case, a filter is transformed into a `CallExpression` that has an extra property named `filter` with the value of `true`. This phase is heavily based on the previous implementation of `$parse`.

The AST processing phase transforms the AST into a function that can be executed to evaluate the expression. The logic for expressions remains unchanged. The AST processing phase works in two different ways depending if csp is enabled or disabled. If csp is enabled, the processing phase returns pre-generated function that interpret specific parts of the AST.
When csp is disabled, then the entire expression is compiled into a single function that is later evaluated using `Function`. In both cases, the returning function has the properties `constant`, `literal` and `inputs` as in the previous implementation. These are used in the next phase to perform different optimizations.

The cacheing, one-time binding and `$watch` optimizations phase remains mostly unchanged.

[1] https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1309,7 +1309,10 @@ ASTInterpreter.prototype = {
       };
     case AST.MemberExpression:
       left = this.recurse(ast.object, false, !!create);
-      if (!ast.computed) ensureSafeMemberName(ast.property.name, self.expression);
+      if (!ast.computed) {
+        ensureSafeMemberName(ast.property.name, self.expression);
+        right = ast.property.name;
+      }
       if (ast.computed) right = this.recurse(ast.property);
       return ast.computed ?
         function(scope, locals, assign, inputs) {
@@ -1333,15 +1336,15 @@ ASTInterpreter.prototype = {
         } :
         function(scope, locals, assign, inputs) {
           var lhs = left(scope, locals, assign, inputs);
-          if (create && create !== 1 && lhs && !(ast.property.name in lhs)) {
-            lhs[ast.property.name] = {};
+          if (create && create !== 1 && lhs && !(right in lhs)) {
+            lhs[right] = {};
           }
-          var value = lhs != null ? lhs[ast.property.name] : undefined;
-          if (self.expensiveChecks || isPossiblyDangerousMemberName(ast.property.name)) {
+          var value = lhs != null ? lhs[right] : undefined;
+          if (self.expensiveChecks || isPossiblyDangerousMemberName(right)) {
             ensureSafeObject(value, self.expression);
           }
           if (context) {
-            return {context: lhs, name: ast.property.name, value: value};
+            return {context: lhs, name: right, value: value};
           } else {
             return value;
           }
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1764,21 +1764,23 @@ function $ParseProvider() {
       var lastResult;
 
       if (inputExpressions.length === 1) {
-        var oldInputValue = expressionInputDirtyCheck; // init to something unique so that equals check fails
+        var oldInputValueOf = expressionInputDirtyCheck; // init to something unique so that equals check fails
         inputExpressions = inputExpressions[0];
         return scope.$watch(function expressionInputWatch(scope) {
           var newInputValue = inputExpressions(scope);
-          if (!expressionInputDirtyCheck(newInputValue, oldInputValue)) {
+          if (!expressionInputDirtyCheck(newInputValue, oldInputValueOf)) {
             lastResult = parsedExpression(scope, undefined, undefined, [newInputValue]);
-            oldInputValue = newInputValue && getValueOf(newInputValue);
+            oldInputValueOf = newInputValue && getValueOf(newInputValue);
           }
           return lastResult;
         }, listener, objectEquality, prettyPrintExpression);
       }
 
       var oldInputValueOfValues = [];
+      var oldInputValues = [];
       for (var i = 0, ii = inputExpressions.length; i < ii; i++) {
         oldInputValueOfValues[i] = expressionInputDirtyCheck; // init to something unique so that equals check fails
+        oldInputValues[i] = null;
       }
 
       return scope.$watch(function expressionInputsWatch(scope) {
@@ -1787,12 +1789,13 @@ function $ParseProvider() {
         for (var i = 0, ii = inputExpressions.length; i < ii; i++) {
           var newInputValue = inputExpressions[i](scope);
           if (changed || (changed = !expressionInputDirtyCheck(newInputValue, oldInputValueOfValues[i]))) {
+            oldInputValues[i] = newInputValue;
             oldInputValueOfValues[i] = newInputValue && getValueOf(newInputValue);
           }
         }
 
         if (changed) {
-          lastResult = parsedExpression(scope, undefined, undefined, oldInputValueOfValues);
+          lastResult = parsedExpression(scope, undefined, undefined, oldInputValues);
         }
 
         return lastResult;
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -606,6 +606,11 @@ function findConstantAndWatchExpressions(ast, $filter) {
   var allConstants;
   var argsToWatch;
   switch (ast.type) {
+  case AST.Program:
+    forEach(ast.body, function(expr) {
+      findConstantAndWatchExpressions(expr.expression, $filter);
+    });
+    break;
   case AST.Literal:
     ast.constant = true;
     ast.toWatch = [];
@@ -738,17 +743,7 @@ ASTCompiler.prototype = {
       assign: {vars: [], body: [], own: {}},
       inputs: []
     };
-    var lastExpression;
-    forEach(ast.body, function(expression) {
-      findConstantAndWatchExpressions(expression.expression, self.$filter);
-    });
-    this.state.computing = 'fn';
-    this.stage = 'main';
-    forEach(ast.body, function(expression) {
-      if (lastExpression) self.current().body.push(lastExpression, ';');
-      self.recurse(expression.expression, undefined, undefined, function(expr) { lastExpression = expr; });
-    });
-    if (lastExpression) this.return(lastExpression);
+    findConstantAndWatchExpressions(ast, self.$filter);
     var extra = '';
     var assignable;
     this.stage = 'assign';
@@ -770,6 +765,9 @@ ASTCompiler.prototype = {
       self.state.inputs.push(fnKey);
       watch.watchId = key;
     });
+    this.state.computing = 'fn';
+    this.stage = 'main';
+    this.recurse(ast);
     var fnString =
       // The build and minification steps remove the string "use strict" from the code, but this is done using a regex.
       // This is a workaround for this until we do a better job at only removing the prefix only when we should.
@@ -852,6 +850,16 @@ ASTCompiler.prototype = {
     var left, right, self = this, args, expression;
     recursionFn = recursionFn || noop;
     switch (ast.type) {
+    case AST.Program:
+      forEach(ast.body, function(expression, pos) {
+        self.recurse(expression.expression, undefined, undefined, function(expr) { right = expr; });
+        if (pos !== ast.body.length - 1) {
+          self.current().body.push(right, ';');
+        } else {
+          self.return(right);
+        }
+      });
+      break;
     case AST.Literal:
       expression = this.escape(ast.value);
       this.assign(intoId, expression);
@@ -1213,9 +1221,7 @@ ASTInterpreter.prototype = {
     var ast = this.astBuilder.ast(expression);
     this.expression = expression;
     this.expensiveChecks = expensiveChecks;
-    forEach(ast.body, function(expression) {
-      findConstantAndWatchExpressions(expression.expression, self.$filter);
-    });
+    findConstantAndWatchExpressions(ast, self.$filter);
     var expressions = [];
     forEach(ast.body, function(expression) {
       expressions.push(self.recurse(expression.expression));
@@ -1229,6 +1235,13 @@ ASTInterpreter.prototype = {
                });
                return lastValue;
              };
+    var assignable;
+    if ((assignable = assignableAST(ast))) {
+      var assign = this.recurse(assignable);
+      fn.assign = function(scope, value, locals) {
+        return assign(scope, locals, value);
+      };
+    }
     var toWatch = getInputs(ast.body);
     if (toWatch) {
       var inputs = [];
@@ -1236,13 +1249,6 @@ ASTInterpreter.prototype = {
         inputs.push(self.recurse(watch));
       });
       fn.inputs = inputs;
-    }
-    var assignable;
-    if ((assignable = assignableAST(ast))) {
-      var assign = this.recurse(assignable);
-      fn.assign = function(scope, value, locals) {
-        return assign(scope, locals, value);
-      };
     }
     fn.literal = isLiteral(ast);
     fn.constant = isConstant(ast);
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -818,7 +818,7 @@ ASTCompiler.prototype = {
     var fns = this.state.inputs;
     var self = this;
     forEach(fns, function(name) {
-      result.push('var ' + name + '=' + self.generateFunction(name, 's,l'));
+      result.push('var ' + name + '=' + self.generateFunction(name, 's'));
     });
     if (fns.length) {
       result.push('fn.inputs=[' + fns.join(',') + '];');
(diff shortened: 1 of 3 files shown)


jbedard:
Here are two failing tests: https://github.com/jbedard/angular.js/commit/da2d8b2780a6f85831cbbf0fa0751a0b5ec58eff. These weird cases are the reason I skipped the `inputs` for assignment and multistatement expressions before, didn't seem worth it...

lgalfaso:
@jbedard thanks for the test cases. Updated the PR with your test cases and the fixes for ir.

jbedard:
Here's another case that has changed: https://github.com/jbedard/angular.js/commit/7ae6e69c81b5a249734abb7010326091b7456d08. The "correct" behaviour might be debatable, but it has changed.

Might also be worth adding the tests from #10322 (or maybe that will get merged?)?

lgalfaso:
The new behavior is the correct as it follows #4664

-lucas

jbedard:
Yea, I guess that does make sense. I was thinking variables should only be *read* from locals like #10322. But now that I think of it `(a = 1) && a` should always return 1, weather `a` exists as a local or not. I think tests should be added showing this though...

Previously the "is a local" test used `.hasOwnProperty` (https://github.com/angular/angular.js/blob/master/src/ng/parse.js#L850) where now it uses `in` (https://github.com/lgalfaso/angular.js/commit/c6a09eb755d2672a9748fa08965bd4f31a491412#diff-780de070ede30180f6aedb6c5f7d57caR1072). This means that variables named "watch" (in FF) or "toString" will be treated as locals (if the locals arg is passed). Does that matter? That could also fix some cases though (like passing prototype/inheriting objects, such as scopes, as locals).

lgalfaso:
> Previously the "is a local" test used `.hasOwnProperty` [...]  where now it uses `in`...

This is intentional, we used to have a discrepancy, use `in` when trying to know if an object has a property that we should assign something or create a new instance [1] and when trying to access that instance. Migrating everything to `in` makes the logic simpler to understand


[1] in expressions of the form `foo.bar = 1`, when trying to know if `scope.foo` should be created or not

lgalfaso:
Added a test and fixed an issue as when csp was enabled, the patch was using `hasOwnProperty` and not `in`

jbedard:
How about https://github.com/jbedard/angular.js/commit/5840e10f9a10b2e17c06ef95412117d8bc5bbbb9?w=1 to simplify the `if (!i)` generation? That way only `findConstantAndWatchExpressions` needs to know which expressions can be replaced with inputs. It also changes the generated code from `if (!i) { N } else { Y }` to `if (i) { Y } else { N }` which is a little simpler but should make no difference at runtime.  Might be nicer to add another function arg instead of using delete to know that the if was already generated though...

src/ng/parse.js
```diff
@@ -426,375 +574,1027 @@ Parser.prototype = {
...
+      expression = ast.operator + '(' + this.ifDefined(right, 0) + ')';
+      this.assign(intoId, expression);
+      recursionFn(expression);
+      break;
+    case AST.BinaryExpression:
+      this.recurse(ast.left, undefined, undefined, function(expr) { left = expr; });
+      this.recurse(ast.right, undefined, undefined, function(expr) { right = expr; });
+      if (ast.operator === '+') {
+        expression = this.plus(left, right);
+      } else if (ast.operator === '=') {
```
jbedard: I think this should be `ast.operator === '-'`? Looks like this needs tests as well...

```javascript
expect($rootScope.$eval("a - b")()).toBe(0);
expect($rootScope.$eval("a - b")({a:0})).toBe(0);
expect($rootScope.$eval("a - b")({b:0})).toBe(0);
```

(`+` should also be tested even though it currently passes)

src/ng/parse.js
```diff
@@ -426,375 +574,1027 @@ Parser.prototype = {
...
+  lazyAssign: function(id, value) {
+    var self = this;
+    return function() {
+      self.assign(id, value);
+    };
+  },
 
-    var indexFn = this.expression();
-    this.consume(']');
+  stringEscapeRegex: new RegExp('[^ a-zA-Z0-9]', 'g'),
```
jbedard: Normally regexps are declared using the `/[^ a-zA-Z0-9]/g` style. Any reason not to here?

This will also escape a lot more then necessary, but a bigger/uglier regex might not be worth it? But `/[ -&(-}]/g` (smaller/uglier!) would match all standard ascii except the quote...

src/ng/parse.js
```diff
@@ -426,375 +574,1027 @@ Parser.prototype = {
...
-      return array;
-    }, {
-      literal: true,
-      constant: elementFns.every(isConstant),
-      inputs: elementFns
+    var toWatch = getInputs(ast.body);
+    var inputs;
+    if (toWatch) {
+      inputs = [];
+      forEach(toWatch, function(watch, key) {
```
jbedard: I'd always vote for `inputs = toWatch.map(...)` for cases like this, just because it looks nicer imo.

Could even do `var inputs = toWatch && toWatch.map(...)` and drop the `if`...

src/ng/parse.js
```diff
@@ -426,375 +574,1027 @@ Parser.prototype = {
...
+        this.recurse(ast.test),
+        this.recurse(ast.alternate),
+        this.recurse(ast.consequent),
+        context
+      );
+    case AST.Identifier:
+      ensureSafeMemberName(ast.name);
+      return function(scope, locals, assign, inputs) {
+        var base = locals && (ast.name in locals) ? locals : scope;
+        if (self.expensiveChecks || isPossiblyDangerousMemberName(ast.name)) {
```
jbedard: Referencing `self` and `ast` here in the returned method causes (I assume, haven't actually verified) both the AST and ASTInterpreter to not get GCed (and the ASTInterpreter references the Parser and Lexer). There are a lot of other references in functions return from this method. Do we want to avoid that still?

src/ng/parse.js
```diff
@@ -426,375 +574,1027 @@ Parser.prototype = {
...
+    'true': { type: AST.Literal, value: true },
+    'false': { type: AST.Literal, value: false },
+    'null': { type: AST.Literal, value: null },
+    'undefined': {type: AST.Literal, value: undefined },
+    'this': {type: AST.ThisExpression }
+  }
+};
+
+function ifDefined(v, d) {
+  return typeof v !== 'undefined' ? v : d;
```
jbedard: Most places in the code use `v !== undefined` instead of typeof. Same with the checks below in `plusFn`.

Ironically the one place that typeof is used is in `isDefined` and `isUndefined` though...

src/ng/parse.js
jbedard: There are quite a few places in this PR where `.map` could be used. IMO `.map` is normally nicer, often makes it clearer what the loop is doing, sometimes avoids a closure etc. but that is debatable. What do others think?

src/ng/parse.js
```diff
@@ -426,375 +574,1027 @@ Parser.prototype = {
...
+      extra +
+      this.watchFns() +
+      'return fn;';
+
+    /* jshint -W054 */
+    var fn = (new Function('$filter',
+        'ensureSafeMemberName',
+        'ensureSafeObject',
+        'ensureSafeFunction',
+        'isPossiblyDangerousMemberName',
```
jbedard: This arg isn't used in the generated code.

src/ng/parse.js
```diff
@@ -426,375 +574,1027 @@ Parser.prototype = {
...
+            return value;
+          }
+        } :
+        function(scope, locals, assign, inputs) {
+          var lhs = left(scope, locals, assign, inputs);
+          if (create && create !== 1 && lhs && !(right in lhs)) {
+            lhs[right] = {};
+          }
+          var value = lhs != null ? lhs[right] : undefined;
+          if (self.expensiveChecks || isPossiblyDangerousMemberName(right)) {
```
jbedard: This `expensive || dangerous` check could be done once outside the closure. Not sure if that's worth it for CSP mode though? That would also remove the `self` reference though...

jbedard:
Inputs to interceptor functions needs to change. Previously the inputs were "collected" the first time an expression was watched (`collectExpressionInputs`) and cached in $$inputs. Now this is done at parse time (`findConstantAndWatchExpressions`). This means interceptor function inputs aren't taken into account when collecting those inputs.

This makes all array/object parsed-expressions-bp tests (which use interceptors for unrelated reasons) faster then master instead of slower: https://github.com/jbedard/angular.js/commit/a953285048fb7a49a2b291fc4f6083077e72c3ab. The TODO/test-change is probably not worth fixing...

src/ng/parse.js
lgalfaso: done

src/ng/parse.js
lgalfaso: I think `/[ -&(-}]/g` is too broad, that said, I think this can be extended to other characters later

src/ng/parse.js
lgalfaso: I like using `map` only when there are no side-effects. In this specific case, there are

src/ng/parse.js
lgalfaso: Just for consistency, we should use whatever `isDefined` and `isUndefined` use

src/ng/parse.js
lgalfaso: done

src/ng/parse.js
lgalfaso: thanks!

src/ng/parse.js
lgalfaso: we can keep on in CSP mode, we can end up with 100s of functions for all possible variations, I think the current implementation is a reasonable balance

lgalfaso:
> How about jbedard@5840e10

As in the patch, it will have side-effects as the recursion function should be executed only when not going thru the `inputs` path (and I do not like removing properties from the ast), the new patch fixes these

>  jbedard@a953285

Has a bug as when there are no inputs and the interceptos is stateless, then the generated function should not call the interceptor when not needed, fixed this and make it part of the new patch

jbedard:
> Has a bug as when there are no inputs

That's what the TODO was. Just note that this specific case doesn't use the input value. This is actually equivalent to before the PR so I think it's fine. However that unused input value gets passed into the parsedExpression which is incorrect, but causes no issues (at this time) because this only happens when the parsedExpression has no inputs so it will ignore it.

src/ng/parse.js
jbedard: I think for consistency we should use what the rest of the code uses...

src/ng/parse.js
jbedard: Is it too broad? I thought it was fine, the regexp is just harder to understand without an ascii chart...

src/ng/parse.js
```diff
@@ -1018,54 +1692,34 @@ function $ParseProvider() {
...
 
       switch (typeof exp) {
         case 'string':
-          cacheKey = exp = exp.trim();
+          exp = exp.trim();
+          cacheKey = exp;
 
           var cache = (expensiveChecks ? cacheExpensive : cacheDefault);
           parsedExpression = cache[cacheKey];
+          if (exp.charAt(0) === ':' && exp.charAt(1) === ':') {
```
jbedard: Most of the changes in this method are unnecessary now (other then removing `wrapSharedExpression`) and should probably be reverted?  This `if` especially though...

src/ng/parse.js
jbedard: I think that's fair (although constructing `inputs` is an extra side effect!). What about ~8 lines down? In that case the function itself has no side effects, but it calls `recurse` which does...

src/ng/parse.js
jbedard: I meant avoiding recomputing `expensive || dangerous()` (mainly the method call) which will never change by pulling them out into a var, but leave the `if` in this method... (I agree it wouldn't be worth creating a separate method). This is very minor though...

src/ng/parse.js
lgalfaso: not trying to skip the conversation, but I think this is something that can be tweaked later

jbedard: That's fine, it's minor.  What about `/.../` style instead of `new RegExp(...)` though?

src/ng/parse.js
lgalfaso: do you mean to put this `if` inside `if (!parsedExpression) {`?

jbedard: Yea...

src/ng/parse.js
lgalfaso: done

src/ng/parse.js
lgalfaso: done

src/ng/parse.js
lgalfaso: done

src/ng/parse.js
lgalfaso: I am not saying it is a rule that cannot be bent, but I think we are moving into a personal style arena

jbedard: Yes it is just style and person preference. I'd be curious what others would say. I'm not sure if use of ES5 methods has been discussed much since dropping IE8?

src/ng/parse.js
```diff
@@ -426,375 +574,1024 @@ Parser.prototype = {
...
-        } else if (token.identifier) {
-          keys.push(token.text);
-        } else {
-          this.throwError("invalid key", token);
+        return context ? {value: value} : value;
+      };
+    case AST.ObjectExpression:
+      args = [];
+      forEach(ast.properties, function(property) {
+        args.push({key: property.key.type === AST.Identifier ?
```
LeonardoBraga: What about storing `property.key` in a local variable to avoid the object traversal as well as improve the minified output?

lgalfaso: > What about storing property.key in a local variable to avoid the object
> traversal as well as improve the minified output?
>

 I think it is important that we accept an expression that follows
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API

LeonardoBraga: I'm sorry, I wasn't clear. I meant the following:

```javascript
forEach(ast.properties, function(property) {
  var propertyKey = property.key;
  args.push({key: propertyKey.type === AST.Identifier ? 
                  propertyKey.name : ('' + propertyKey.value),
             value: self.recurse(property.value)
  });
```

Since I'm not familiar with the parser API you referred to, please accept my apologies if this breaks the construction of the expression somehow. Otherwise, if this offers any benefits, there are a few other cases with `.body.length` down below that could leverage the same idea as well.

lgalfaso: this code is executed only during parsing and not during the expression evaluation, so readability is more important than trying to remove one cycle here or there. Eg. a lot more can be saved by using a normal for-loop, but its current form is easier to read

LeonardoBraga: Got it. Thanks for the explanation.

jbedard:
Here's an issue and fix with how inputs/valueOf works: https://github.com/jbedard/angular.js/commit/f9fd9936dbbe2434e3a8c5f5994db12260445507

That commit just modifies a test to reproduce it, but it might be nice to have better tests around that.

jbedard:
Do we want the tests from https://github.com/angular/angular.js/commit/869008140a96e0e9e0d9774cc2e5fdd66ada7ba9 since it was only merged to 1.3?

lgalfaso:
@jbedard will cherry pick these tests later today

lgalfaso:
There is something wrong with 8690081 as we should only throw with
```js
scope.$eval("c.a = 1", {c: Function.prototype.constructor}
```

if expensive checks is enabled

jbedard:
The `setter` method always does expensive checks though. 8690081 didn't change that.

If you do the same thing with `scope.c = Function.prototype.constructor` instead of locals it also throws (with or without 8690081)...

lgalfaso:
Then we need more tests and a clear definition of what "expensive checks"
means, let me ping a few people and then I will create the tests

jbedard:
The way I understood it "expensive checks" means each and every object in a getterFn gets checked. Where non-expensive only checks "suspicious looking" parts of a getterFn (".constructor" is the only suspicious thing today). The `setter` was doing the check on every object before the "expensive" stuff.

lgalfaso:
@jbedard I know how it works, but the fact that it works different for getters and setters is odd (the fact that `a['b']` is also different than `a.b` is also odd, but for other reasons)

lgalfaso:
landed as 0d424263ead16635afb582affab2b147f8e71626

PR #10615, opened 2015-01-01
docs(tutorial/0 - Bootstrapping): Remind users to refresh page

anyong:
On line 32-34 after reverting to step-0 and starting the webserver, the browser may have already cached the master branch of the app and the user will see the master version in their browser. I just added a reminder to tell them to refresh the page if this happens!


anyong:
Just signed the CLA.

PR #10618, opened 2015-01-02
fix(jqLite): properly handle dash-delimited node names in `jqLiteBuildFragment`

### LINKED ISSUE #10617: jqLite dumps <tr-xxx> while running jqLiteBuildFragment
`$compile` service will call jqLite to convert html to JQuery element. It's found that in `jqLiteBuildFragment` function, if a customized directive tag like `<tr-xxx>` will be wrapped to `<table>...` due to this regexp `var TAG_NAME_REGEXP = /<([\w:]+)/;` does not capture 'dash-delimited' tag name. And the fragment `tr` is in `wrapMap` which will be wrapped to a comprehensive `table` html later.
 
As a result, the directive cannot be compiled successfully.

I wonder any reason behind this or it's a bug to be fixed, should I PR for this?

---

https://github.com/angular/angular.js/blob/master/src/jqLite.js#L145

gkalpak:
Closes #10617

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -142,7 +142,7 @@ function camelCase(name) {
 
 var SINGLE_TAG_REGEXP = /^<(\w+)\s*\/?>(?:<\/\1>|)$/;
 var HTML_REGEXP = /<|&#?\w+;/;
-var TAG_NAME_REGEXP = /<([\w:]+)/;
+var TAG_NAME_REGEXP = /<([\w:-]+)/;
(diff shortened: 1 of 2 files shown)


lgalfaso:
LGTM, please also commit this to the 1.3.x branch

gkalpak:
**NOTE:**
This commit fixes the bug for JQLite, but the issue still remains when jQuery is used.
See discussion on #10617 for more details.

lgalfaso:
closing in favor of #10619

PR #10619, opened 2015-01-02
fix(jqLite): Adding more support for custom element tagnames

richardaday:
Before this change, jqLite would allow you to create elements such as
"my-foo" and "div-foo" but would not allow you to create certain
elements such as "tr-foo" and  "td-foo".  This is incorrect as well as
undocumented and unexpected.

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -142,7 +142,7 @@ function camelCase(name) {
 
 var SINGLE_TAG_REGEXP = /^<(\w+)\s*\/?>(?:<\/\1>|)$/;
 var HTML_REGEXP = /<|&#?\w+;/;
-var TAG_NAME_REGEXP = /<([\w:]+)/;
+var TAG_NAME_REGEXP = /<([\w:-]+)/;
 var XHTML_TAG_REGEXP = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:]+)[^>]*)\/>/gi;
 
 var wrapMap = {
--- a/test/jqLiteSpec.js
+++ b/test/jqLiteSpec.js
@@ -67,6 +67,47 @@ describe('jqLite', function() {
     });
 
 
+    it('should allow construction of an html element containing dashes', function() {
+      var nodes;
+      nodes = jqLite('<my-foo></my-foo>');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('MY-FOO');
+
+      nodes = jqLite('<my-foo />');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('MY-FOO');
+    });
+
+
+    it('should allow construction of an html element containing dashes and starting with characters that match existing html elements', function() {
+      var nodes;
+
+      nodes = jqLite('<option-foo></option-foo>');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('OPTION-FOO');
+
+      nodes = jqLite('<thead-foo></thead-foo>');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('THEAD-FOO');
+
+      nodes = jqLite('<col-foo></col-foo>');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('COL-FOO');
+
+      nodes = jqLite('<tr-foo></tr-foo>');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('TR-FOO');
+
+      nodes = jqLite('<td-foo></td-foo>');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('TD-FOO');
+
+      nodes = jqLite('<div-foo></div-foo>');
+      expect(nodes.length).toEqual(1);
+      expect(nodes[0].nodeName).toEqual('DIV-FOO');
+    });
+
+
     it('should allow construction of html with leading whitespace', function() {
       var nodes = jqLite('  \n\r   \r\n<div>1</div><span>2</span>');
       expect(nodes[0].parentNode).toBeDefined();


richardaday:
This bug also exists in jQuery and I hope to find some time soon to send a patch their way as well.

gkalpak:
Same fix as #10618.
(More explicit testing :smiley:).

richardaday:
Wow, both pull requests were created at almost the same time!

What would you like me to do with this pull request?  I can close it if you like.

gkalpak:
I like the fact that there are tests for more elements (it shouldn't make any actual difference, but feels better :smiley:).
Ideally, I would change the tests like this (using `nodeName_` and having also some HTML content just to be sure):

```js
nodes = jqLite('<x>y<x/>');
expect(nodes.length).toBe(1);
expect(nodeName_(nodes)).toBe('x');
expect(nodes.html()).toBe('y');
```

It would also be nice to have similar tests for `:`-delimited tag names (for future-proofing) and some nested elements (e.g. `jqLite('<div><tr-foo>Test</tr-foo></div>')`).

<sub>If you can add these test, I would close my PR in favor of this one :stuck_out_tongue:</sub>

richardaday:
Perfect. Give me a day or two.

gkalpak:
The bug has been [fixed on the jQuery side][1] (thx @LeonardoBraga), but it seems that the fix will be [released in version 3.0][2] (which according to the [roadmap][3] is planned for Q1 2015).

Not sure if/when Angular 1.x will make the switch to jQuery 3.x.
@petebacondarwin: What should we do about this ?

1. Fix it in jqLite, but don't have tests for the fix (since it will break on jQuery) ?
2. Fix it in jqLite and have different tests for jqLite/jQuery (until jQuery gets to 3.0) ?
3. Wait for Angular to upgrade to jQuery 3.x, before fixing jqLite ?

[1]: https://github.com/jquery/jquery/issues/1987
[2]: https://github.com/jquery/jquery/issues/1987#issuecomment-70250751
[3]: https://github.com/jquery/jquery/wiki/Roadmap

lgalfaso:
We should fix this at jqLite (using the same pattern that was used in jQuery) and have tests that only run with jqLite. Do please add a comment in the test pointing to jQuery issue and the fact that this will work as expected in jQuery 3.0

gkalpak:
What needs to be done:

- [ ] Port the changes from the [jQuery patch][1]:
    - [ ] Update `SINGLE_TAG_REGEXP`. (This is not functionaly necessary; just to be in sync with jQuery.)
    - [x] Update `TAG_NAME_REGEXP`.
    - [ ] Update `XHTML_TAG_REGEXP`. (This is not functionaly necessary; just to be in sync with jQuery.)
- [ ] Address [my comment][2] above.
- [ ] Add 
- [ ] Make sure the test are only run with jqLite. The `_jqLiteMode` global can be used for that:  
    `if (_jqLiteMode) { /* run the test */ }`
- [ ] Add an appropriate comment in the test as per @lgalfaso's [comment][3].

@richardaday: Would you like to take care of these in this PR ?

[1]: https://github.com/jquery/jquery/pull/1988/files
[2]: https://github.com/angular/angular.js/pull/10619#issuecomment-68589113
[3]: https://github.com/angular/angular.js/pull/10619#issuecomment-70632767

caitp:
this has been on hold for a while --- do we want this?

Since it won't do anything to help people using jQuery, I'm not sure. Really don't want to monkey patch jQuery again

LeonardoBraga:
I can provide the PR to address @gkalpak's and @lgalfaso's comments and action items. I think we should add these changes. Do you see an downside on this, @caitp ?

caitp:
@LeonardoBraga the downside I see is that it will mean an inconsistency between jqLite and the jQuery version used by most applications, I'm not super keen on that. But I'm undecided, I think @petebacondarwin should make the call on this.

caitp:
that said, by all means send a PR, since this one has not been updated since January

petebacondarwin:
Bumping this to 1.5 as it is not a regression and we need to think about it a little. @mzgol do you have a view on this?

mzgol:
IMO it's not bad that it creates inconsistency with jQuery as it's just fixing a bug. jQuery has fixed this bug as well and it will be included in 3.0.0 when it's released. I don't see it blocking that if Angular is used with jQuery 2.1.1 this bug will appear; we can say it's a jQuery bug and you have to either use jqLite or upgrade to jQuery 3.0.0 or later to have it working.

Review comments still need to be addressed, though, like the @gkalpak's one: https://github.com/angular/angular.js/pull/10619#issuecomment-70640445. I'd just omit the relevant tests in jQuery <3.0.0 and not all jQueries. If the PR is ready before the final 1.4 release then we can land it IMO.

gkalpak:
@LeonardoBraga, do you plan to submit a PR for this ?
(If not, I will; just asking to avoid duplication of effort :smile:)

gkalpak:
@mzgol, what is the best option for verifying the jQuery version used in tests ?

mzgol:
> @mzgol, what is the best option for verifying the jQuery version used in tests?

If you just want to check if it's 3.0.0 or newer, the easiest would be to:
```js
if (window.jQuery && Number(window.jQuery.fn.jquery.split('.')[0]) >= 3) { /* code */ }
```
as `jQuery.fn.jquery` contains the version. Some files have the `jQuery` global set to `undefined` in jqLite mode (no idea where's the code causing that) so some tests check for `jQuery` instead of `window.jQuery`, you may try it.

If we wanted more fine-grained version comparisons (the above code is ugly and doesn't scale behind major checks), it'd be best to use the [semver](https://www.npmjs.com/package/semver) module:
 ```js
if (window.jQuery && semver.satisfies(window.jQuery.fn.jquery, '>=3.0.0')) { /* code */ }
```

I know @IgorMinar has been recently trying to reduce used dependencies instead of adding new ones, though. :-)

gkalpak:
Ckecking for 3.0.0 or newer would suffice, so the former approach is fine. Thx !

gkalpak:
Closing in favor of #12759 (due to inactivity).

PR #10620, opened 2015-01-02
chore(travis): update browsers to the latest version

lgalfaso:
Update the used browsers to the latest versions available

### DIFF
--- a/karma-shared.conf.js
+++ b/karma-shared.conf.js
@@ -35,18 +35,18 @@ module.exports = function(config, specificOptions) {
       'SL_Chrome': {
         base: 'SauceLabs',
         browserName: 'chrome',
-        version: '34'
+        version: '39'
       },
       'SL_Firefox': {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@lgalfaso - there appear to be two fixes in this commit.
One is related to a bug in Safari I believe? (The old 2px thing). I think this should be in a separate commit.

lgalfaso:
Newer versions of Safari do not update the calculated properties (after a programatic css change) synchronously. The change to the test is needed to make it pass in Safari (I can make it 2 commits, but just the browser update without the change on the test will make the test fail in Safari)

petebacondarwin:
You can always have commits that fail followed by commits that fix them as long as they are pushed together. But in this case, it is not needed. We just put the "fix" commit first.

petebacondarwin:
I guess we backport to 1.3.x

PR #10621, opened 2015-01-02
fix(filterFilter): do not filter objects properly in case a property is an array

fealaer:
In JavaScript, an array is a special type of object, therefore typeof [] returns object.
Added corresponding unit tests.
Changed condition for array type to isArray.

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -186,7 +186,7 @@ function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatc
 
   if ((expectedType === 'string') && (expected.charAt(0) === '!')) {
     return !deepCompare(actual, expected.substring(1), comparator, matchAgainstAnyProp);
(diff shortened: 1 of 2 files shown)


src/ng/filter/filter.js
```diff
@@ -186,7 +186,7 @@ function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatc
 
   if ((expectedType === 'string') && (expected.charAt(0) === '!')) {
     return !deepCompare(actual, expected.substring(1), comparator, matchAgainstAnyProp);
-  } else if (actualType === 'array') {
+  } else if (isArray(actual)) {
```
caitp: Jeeze @_@

caitp:
Looks correct to me, but could you please fix the indentation in your tests? 2 space indentation

caitp:
LGTM, just waiting on travis

PR #10622, opened 2015-01-02
feat($http): provide a config object as an argument to header functions

pkozlowski-opensource:
This is a cleaned up / alternative version for #7235

### DIFF
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -781,6 +781,34 @@ describe('$http', function() {
         $httpBackend.flush();
       });
 
+      it('should expose a config object to header functions', function() {
+        var config = {
+          foo: 'Rewritten',
+          headers: {'Accept': function(config) {
+            return config.foo;
+          }}
+        };
+
+        $httpBackend.expect('GET', '/url', undefined, {Accept: 'Rewritten'}).respond('');
+        $http.get('/url', config);
+        $httpBackend.flush();
+      });
+
+      it('should not allow modifications to a config object in header functions', function() {
+        var config = {
+          headers: {'Accept': function(config) {
+            config.foo = 'bar';
+            return 'Rewritten';
+          }}
+        };
+
+        $httpBackend.expect('GET', '/url', undefined, {Accept: 'Rewritten'}).respond('');
+        $http.get('/url', config);
+        $httpBackend.flush();
+
+        expect(config.foo).toBeUndefined();
+      });
+
       it('should check the cache before checking the XSRF cookie', inject(function($browser, $cacheFactory) {
         var testCache = $cacheFactory('testCache'),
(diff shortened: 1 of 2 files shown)


fundon:
+1, Nice!

pkozlowski-opensource:
@caitp / @petebacondarwin / @lgalfaso (or anyone else!) - would you guys mind reviewing this one - it is pretty minimal

caitp:
taking a quick gander

test/ng/httpSpec.js
```diff
@@ -781,6 +781,26 @@ describe('$http', function() {
...
+          return headers['Accept'] == 'Rewritten';
+        }
+
+        $httpBackend.expect('GET', '/url', undefined, checkHeaders).respond('');
+        $http.get('/url', config);
+        $httpBackend.flush();
+
+        expect(config.bar).toBeUndefined();
+      });
+
```
caitp: should we also test the main functionality of this feature? :p

pkozlowski-opensource: @caitp it is tested as otherwise a mocked request would fail, no?

caitp: let me check ngMock one sec... But in spite of that, I think you should have a different spec for that, because it's not explicitly testing that, it's explicitly testing that the passed in config object is a shallow copy

pkozlowski-opensource: Yep, I thought of splitting it into 2 separate specs when I saw your comment. Will create 2 separate specs.

petebacondarwin:
Can you guys triage your own PRs please!

pkozlowski-opensource:
I've rebased this one and split tests based on the review comments. 
@caitp do you want to have another look at this one?

caitp:
I'll take a look on Monday

test/ng/httpSpec.js
```diff
@@ -781,6 +781,34 @@ describe('$http', function() {
...
+            return config.foo;
+          }}
+        };
+
+        $httpBackend.expect('GET', '/url', undefined, {Accept: 'Rewritten'}).respond('');
+        $http.get('/url', config);
+        $httpBackend.flush();
+      });
+
+      it('should not allow modifications to a config object in header functions', function() {
```
caitp: should ignore modifications (but no big deal)

caitp: this will also only "ignore" changes to root properties of the config object, so changes to child object properties won't be ignored :( but I don't think it's worth worrying about it until someone complains

caitp:
still lgtm

test/ng/httpSpec.js
pkozlowski-opensource: yeh... but I agree that we shouldn't worry about it too much right now...

PR #10623, opened 2015-01-02
docs(tutorial/11 - REST and Custom Services): Add a note that new test d...

kenhtanaka:
...ependencies cause a failure that can be cleared by restarting Karma.

### DIFF
--- a/docs/content/tutorial/step_11.ngdoc
+++ b/docs/content/tutorial/step_11.ngdoc
@@ -15,6 +15,10 @@ lower-level {@link ng.$http $http} API, HTTP methods and URLs.
 
 <div doc-tutorial-reset="11"></div>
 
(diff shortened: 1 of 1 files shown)


kenhtanaka:
I've signed the CLA, Contributor License Agreement.

pkozlowski-opensource:
I'm not big fan of this change the way it is written. Maybe we should advise people to stop Karma before going to a next step instead?

kenhtanaka:
I think it would be clear to advise people to stop Karma before going to the next step (and then restart afterward). Do I reword this and then issue another pull request?

petebacondarwin:
Please reword and then add a new commit to this PR. We can merge and rebase etc when it is good.

petebacondarwin:
Closing due to lack of activity.

PR #10631, opened 2015-01-04
feat($timeout): pass arguments in callback

PatrickJS:
PR has updated Docs and Tests
`setTimeout` allows you to pass arguments into the callback function in
the …3rd argument. Since that’s taken I added it to the …4th
http://mdn.io/setTimeout#Syntax

here is an example of why you would need to do this

``` javascript
import {$http} from 'angular.js/modules';
import {setTimeout, Promise} from 'facade/async';
var cache = {};

export class AwesomeModule {
  static getSwag(url:string):Promise {
    return new Promise(function(resolve, reject) {
      var complete = function(data) {
        cache[url] = data;
        resolve(data);
      };

      if (url in cache) {
        setTimeout(complete, 0, cache[url]);
      } else {
        $http.get(url).then(complete, reject);
      } // end url in cache
    });
  } // end getSwag
} // end AwesomeModule
```

similar pull request https://github.com/angular/angular.js/pull/10632

### DIFF
--- a/src/ng/timeout.js
+++ b/src/ng/timeout.js
@@ -32,6 +32,7 @@ function $TimeoutProvider() {
       * @param {number=} [delay=0] Delay in milliseconds.
       * @param {boolean=} [invokeApply=true] If set to `false` skips model dirty checking, otherwise
       *   will invoke `fn` within the {@link ng.$rootScope.Scope#$apply $apply} block.
+      * @param {...*=} Pass additional parameters to the executed function.
       * @returns {Promise} Promise that will be resolved when the timeout is reached. The value this
       *   promise will be resolved with is the return value of the `fn` function.
       *
@@ -43,14 +44,15 @@ function $TimeoutProvider() {
         fn = noop;
       }
 
-      var skipApply = (isDefined(invokeApply) && !invokeApply),
+      var args = sliceArgs(arguments, 3),
+          skipApply = (isDefined(invokeApply) && !invokeApply),
           deferred = (skipApply ? $$q : $q).defer(),
           promise = deferred.promise,
           timeoutId;
 
       timeoutId = $browser.defer(function() {
         try {
-          deferred.resolve(fn());
+          deferred.resolve(fn.apply(null, args));
         } catch (e) {
           deferred.reject(e);
           $exceptionHandler(e);
--- a/test/ng/timeoutSpec.js
+++ b/test/ng/timeoutSpec.js
@@ -22,7 +22,7 @@ describe('$timeout', function() {
   it('should call $apply after each callback is executed', inject(function($timeout, $rootScope) {
     var applySpy = spyOn($rootScope, '$apply').andCallThrough();
 
-    $timeout(function() {});
+    $timeout(noop);
     expect(applySpy).not.toHaveBeenCalled();
 
     $timeout.flush();
@@ -30,8 +30,8 @@ describe('$timeout', function() {
 
     applySpy.reset();
 
-    $timeout(function() {});
-    $timeout(function() {});
+    $timeout(noop);
+    $timeout(noop);
     $timeout.flush();
     expect(applySpy.callCount).toBe(2);
   }));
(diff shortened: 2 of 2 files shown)


gkalpak:
Note that according to MDN, IE9 does not support the additional arguments.

gdi2290:
due to the way angular is handling `setTimeout ` intervally this code works in IE9 even if `setTimeout ` doesn't support it natively

gkalpak:
@gdi2290: Good point ;) BTW, if it gets merged, it would be good to have the docs updatd as well (imo).

gdi2290:
I added docs
```javascipt
* @param {...*=} Pass additional parameters to the executed function.
```

src/ng/timeout.js
```diff
@@ -28,19 +28,21 @@ function $TimeoutProvider() {
...
+      var args = arguments.length > 3 ? sliceArgs(arguments, 3) : [],
+          skipApply = (isDefined(invokeApply) && !invokeApply),
           deferred = (skipApply ? $$q : $q).defer(),
           promise = deferred.promise,
           timeoutId;
 
       timeoutId = $browser.defer(function() {
         try {
-          deferred.resolve(fn());
+          deferred.resolve(fn.apply(null, args));
```
gkalpak: I wonder if making the same optimisation as in `$interval` here would make any tangible difference. E.g.: `deferred.resolve(hasArgs ? fn.apply(null, args) : fn())`

lgalfaso:
this and #10632 look good, but I would like to see a few more tests that make sure that ngMock also works as expected

src/ng/timeout.js
shahata: I'd go the other way and make this change a simple one liner:
```js
deferred.resolve(fn.apply(null, sliceArgs(arguments, 3)));
```
there's no need for this `hasArgs` variable and no need to test the `arguments.length` on your own, `sliceArgs` will take care of it just fine.

gdi2290: I don't have reference to `arguments` in that scope so the best I can do is `sliceArgs` regardless of argument length
```javascript
//...
var args = sliceArgs(arguments, 3)
//...
  //...
  deferred.resolve(fn.apply(null, args));
```

shahata: :+1:

petebacondarwin:
LGTM

petebacondarwin:
Feel free to merge

gdi2290:
I need to update this PR due to the API change of $timeout https://github.com/angular/angular.js/commit/5a60302389162c6ef45f311c1aaa65a00d538c66 but I should be good to go soon

gdi2290:
@petebacondarwin PR updated with rebase from master branch. Resolving merge conflicts are a pain I have to say. 

Since the `$timeout` overload doesn't affect this feature too much so I didn't account for it in the code (no callback so no point giving `noop` it arguments). I could check if they're using method overload and use `$browser.defer(noop, delay)` rather than having it run all the logic needed for the original api. The other PR (https://github.com/angular/angular.js/commit/5a60302389162c6ef45f311c1aaa65a00d538c66) didn't do this optimization so I can assume it doesn't matter

gdi2290:
I pushed up changes to mock to reflect the `$interval` in order to get this merged. There are a few ways this can be solved

* Refactoring to expose private methods `setInterval` and `clearInterval` so we can use a decorator similar to how [ngMocks/$timeout#L1713](https://github.com/angular/angular.js/blob/master/src/ngMock/angular-mocks.js#L1713) is handled in Angular 1.x or Angular 2 with [facade/async.es6#L39](https://github.com/angular/angular/blob/master/modules/facade/src/async.es6#L39)
* Hack Jasmine/tests to repeat the tests for mock/real 
* Update the build system to run tests for mocks/real (not sure if mocks are included into docs tests but again I have Java errors) 

either one that's chosen I think a new PR would make sense for that change while this one should be good

petebacondarwin:
Welcome to our world (rebase hell)!

I think it is fine not to optimize the case where no handler function is passed for now.

Merging

PR #10632, opened 2015-01-04
feat($interval): pass arguments in callback

PatrickJS:
PR has updated Docs and Tests
`setInterval` allows you to pass arguments into the callback function
in
the …3rd argument. Since that’s taken I added it to the …5th
http://mdn.io/setInterval#Syntax

similar pull request https://github.com/angular/angular.js/pull/10631

### DIFF
--- a/test/ng/intervalSpec.js
+++ b/test/ng/intervalSpec.js
@@ -142,6 +142,31 @@ describe('$interval', function() {
   }));
 
 
+  it('should allow you to specify a number of arguments', inject(function($interval, $window) {
+    var task1 = jasmine.createSpy('task1'),
+        task2 = jasmine.createSpy('task2'),
+        task3 = jasmine.createSpy('task3');
+    $interval(task1, 1000, 2, true, 'Task1');
+    $interval(task2, 1000, 2, true, 'Task2');
+    $interval(task3, 1000, 2, true, 'I', 'am', 'a', 'Task3', 'spy');
+
+    $window.flush(1000);
+    expect(task1).toHaveBeenCalledWith('Task1');
+    expect(task2).toHaveBeenCalledWith('Task2');
+    expect(task3).toHaveBeenCalledWith('I', 'am', 'a', 'Task3', 'spy');
+
+    task1.reset();
+    task2.reset();
+    task3.reset();
+
+    $window.flush(1000);
+    expect(task1).toHaveBeenCalledWith('Task1');
+    expect(task2).toHaveBeenCalledWith('Task2');
+    expect(task3).toHaveBeenCalledWith('I', 'am', 'a', 'Task3', 'spy');
+
+  }));
+
+
   it('should return a promise which will be updated with the count on each iteration',
       inject(function($interval, $window) {
     var log = [],
--- a/src/ng/interval.js
+++ b/src/ng/interval.js
@@ -39,6 +39,7 @@ function $IntervalProvider() {
       *   indefinitely.
       * @param {boolean=} [invokeApply=true] If set to `false` skips model dirty checking, otherwise
       *   will invoke `fn` within the {@link ng.$rootScope.Scope#$apply $apply} block.
+      * @param {...*=} Pass additional parameters to the executed function.
       * @returns {promise} A promise which will be notified on each iteration.
       *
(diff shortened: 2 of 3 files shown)


gkalpak:
Note that according to MDN, IE9 does not support the additional arguments.

gdi2290:
due to the way angular is handling `setInterval` intervally this code works in IE9 even if `setInterval` doesn't support it natively

gkalpak:
@gdi2290: Good point ;) BTW, if it gets merged, it would be good to have the docs updatd as well (imo).

gdi2290:
I added docs and optimise the code a bit to check args once
```javascript
* @param {...*=} Pass additional parameters to the executed function.
```

test/ng/intervalSpec.js
```diff
@@ -142,6 +142,23 @@ describe('$interval', function() {
...
+    $interval(task2, 1000, 1, true, 'Task2');
+    $interval(task3, 1000, 1, true, 'I', 'am', 'a', 'Task3', 'spy');
+
+    $window.flush(1000);
+    expect(task1).toHaveBeenCalledWith('Task1');
+    $window.flush(1000);
+    expect(task2).toHaveBeenCalledWith('Task2');
+    $window.flush(1000);
+    expect(task3).toHaveBeenCalledWith('I', 'am', 'a', 'Task3', 'spy');
+  }));
```
petebacondarwin: This sequence of expectations seems not quite right. Since we are using `$interval` then each of the spies should have been called once per flush.

How about something more like:

```js
$window.flush(1000);
expect(task1).toHaveBeenCalledWith('Task1');
expect(task2).toHaveBeenCalledWith('Task2');
expect(task3).toHaveBeenCalledWith('I', 'am', 'a', 'Task3', 'spy');
task1.reset();
task2.reset();
task3.reset();
$window.flush(1000);
expect(task1).toHaveBeenCalledWith('Task1');
expect(task2).toHaveBeenCalledWith('Task2');
expect(task3).toHaveBeenCalledWith('I', 'am', 'a', 'Task3', 'spy');
```

gdi2290: I updated the tests (and rebase master) with an `$interval` count of 2 otherwise the second part of this won't work since the `.reset()` only resets the state of the spy and not `$interval`.

gdi2290:
I have a question about the mocks. https://github.com/angular/angular.js/blob/master/src/ngMock/angular-mocks.js#L456 I'm assuming I need to update the mocks with the same logic but it seems a bit crazy that I can add a feature to `$interval` without this showing up as an error. Then again I didn't run any of the e2e tests since my computer is throwing Java errors (gg Java) if it did account for it I assumed Travis would catch it

petebacondarwin:
@gdi2290 - the tests of the `$interval` service should not be relying on the mock version of `$interval`; so we ought to be able to check that the `$interval` service is correct.
But I agree that keeping mocks in sync with their real life counterparts. One approach, that perhaps you might like to investigate, is to see if we can run the `$interval` unit tests against the mock version of `$interval`.

gdi2290:
I pushed up changes to mock to reflect the `$interval` in order to get this merged. There are a few ways this can be solved

* Refactoring to expose private methods `setInterval` and `clearInterval` so we can use a decorator similar to how [ngMocks/$timeout#L1713](https://github.com/angular/angular.js/blob/master/src/ngMock/angular-mocks.js#L1713) is handled in Angular 1.x or Angular 2 with [facade/async.es6#L39](https://github.com/angular/angular/blob/master/modules/facade/src/async.es6#L39)
* Hack Jasmine/tests to repeat the tests for mock/real 
* Update the build system to run tests for mocks/real (not sure if mocks are included into docs tests but again I have Java errors) 

either one that's chosen I think a new PR would make sense for that change while this one should be good

petebacondarwin:
Actually `setInterval` and `clearInterval` are already exposed because `$interval` gets them from `$window`, which is a service that we can adapt. So perhaps instead of decorating `$interval` what we need to do is decorate or replace `$window` with new versions of these methods?

petebacondarwin:
In the meantime, this PR looks fine from the point of view of this feature, so let's land it too.

PR #10639, opened 2015-01-04
refact(ngOptions): major refactoring...

### LINKED ISSUE #8019: Bad performance when selecting an item the first time in select with many items
When having many items in a <select> (more than a couple of thousand items), selecting an item becomes really really slow the first time. In this example I add 25000 items. Selecting an item the first time takes around 12s.

http://jsfiddle.net/2W7Kx/1/

Seems like "function render()" is responsible for almost all execution time, with these two lines being suspicious:

    // TODO(vojta): can't we optimize this ?
    scope.$watch(render);

I also initially added a stackoverflow question about the issue: http://stackoverflow.com/questions/24470563/angularjs-performance-for-select

### LINKED ISSUE #9714: Regression using ngOptions asynchronously in angular 1.3
Hi guys,

Just wanted to start by saying that Angular is a wonderful project and I'm having a great time working with it :)

I have an example of the issue in this jsFiddle: http://jsfiddle.net/88aw4yyq/

Consider this HTML:

    <select name="test"
            ng-model="test"
            ng-options="opt.id as opt.label for opt in availableOptions">
    </select>

And this controller:

    $scope.test = '';
    $scope.availableOptions = [];
    
    $timeout(function () {
        $scope.availableOptions = [
            {id: '', label: 'Please select...'},
            {id: 1, label: 'opt1'},
            {id: 2, label: 'opt2'}
        ];
    }, 1000);

The `$timeout` simulates any regular AJAX request. `{id: '', label: 'Please select...'}` is added for convenience for the user.

**Expected behavior**: 'Please select...' should be selected once the data is loaded.

**Actual behavior**: An empty option is added on top in the select field.

**Angular versions**: 1.3.0 this is a regression, 1.2.x was working fine.

**Browser versions**: Tested on Fx/Chrome (latest versions) on both OS X & Windows.

If anyone else is encountering the issue, here is a workaround:

    $scope.test = '';
    $scope.availableOptions = [
        {id: '', label: 'Please select...'}
    ];
    
    $timeout(function () {
        $scope.availableOptions = $scope.availableOptions.concat([
            {id: 1, label: 'opt1'},
            {id: 2, label: 'opt2'}
        ]);
    }, 1000);

### LINKED ISSUE #10531: ngOptions: non-grouped options shouldn't float to the top when using group by
Example fiddle where non-grouped options incorrectly always float to the top: http://jsfiddle.net/raxzauyL/  (against 1.3.7)

The expected behavior is that the `options` should be interspersed between the `optgroups`, not bunched at the top.  I've tried setting the `group` by value to '', undefined, null, and even removing the key from the obj completely, but none seem to work.

petebacondarwin:
**Major reworking of select and ngOptions**:
- The `SelectController` is now used as an abstraction for the `select` and `ngOptions` directives
  to override to get their desired behaviour
- The `select` directive is completely oblivious to the ngOptions directive now - the `ngOptions`
  directive could be deleted without having to make any changes to the `select` directive.
- Select related directives (single/multiple/ngOptions) can provide specific versions of
  `SelectController.writeValue` and `SelectController.readValue`, which are responsible for getting
  the `$viewValue` in or out of the actual `<select>` element and its `<option>` children.

BREAKING CHANGE:

When using `ngOptions`: the directive applies a surrogate key as the value of the `<option>` element.
This commit changes the actual string used as the surrogate key. We now store a string that is computed
by calling `hashKey` on the item in the options collection; previously it was the index or key of the
item in the collection.

Before you might have seen:

```
<select ng-model="x" ng-option="i in items">
  <option value="1">a</option>
  <option value="2">b</option>
  <option value="3">c</option>
  <option value="4">d</option>
</select>
```

Now it will be something like:

```
<select ng-model="x" ng-option="i in items">
  <option value="string:a">a</option>
  <option value="string:b">b</option>
  <option value="string:c">c</option>
  <option value="string:d">d</option>
</select>
```

If your application code relied on this value, which it shouldn't, then you will need to modify your
application to accommodate this. You may find that you can use the `track by` feaure of `ngOptions`
as this provides the ability to specify the key that is stored.

BREAKING CHANGE:

When iterating over an object's properties using the `(key, value) in obj` syntax the order of the elements used to be sorted alphabetically. This was an artificial attempt to create a deterministic ordering since browsers don't guarantee the order. But in practice this is not what people want and so this change iterates over properties in the order they are returned by `Object.keys(obj)`, which is almost always the order in which the properties were defined.

Closes #8019
Closes #9714
Closes #10531


lgalfaso:
would it be possible to split this into two commits, one that move things around and another that make the modifications? It would make it a lot easier to review the patch. One PR is ok and so is the order of the commits

petebacondarwin:
Not really. The original file (select.js) is so interwoven it was not really feasible to extract the ngOptions bits without doing a lot of the refactoring. I'll see what I can do...

petebacondarwin:
I guess I could combine these two files into one and then split them in a subsequent commit?

lgalfaso:
the issue with the current patch is that is very hard to know what actually changed and what was only moved. Whatever makes this easy, works for me

petebacondarwin:
Well pretty much everything that was related to ngOptions changed :-)

petebacondarwin:
@lgalfaso  - How's that?

lgalfaso:
what will happen if the same option value shows up several times? This can happen when the two options are in different option groups

petebacondarwin:
>what will happen if the same option value shows up several times? This can happen when the two options are in different option groups

Good question, I have to check the previous implementation but I think it didn't support multiple options with the same value either,

jayoungers:
Even if the previous implementation did support it, is that really a valid requirement?  How would you even select the correct item in the list the next time they view that page?  you'd have to 'track by' group and value.

petebacondarwin:
@jayoungers - thanks for commenting. I agree that this requirement would be tricky. You could certainly create a `track by` expression that was unique for a given group/value combination.

lgalfaso:
I was thinking when there is no `value`, and the text is used

petebacondarwin:
@lgalfaso - so you are referring to the standard `select` and `option` directive combination rather than `ngOptions`? I don't know if this was ever supported either since the `addOption` method always took the value of the option as its parameter, which means that you would only get one option in the `SelectController`'s map

lgalfaso:
@petebacondarwin ok, makes sense

petebacondarwin:
>will using a hashmap impact performance with many options? Generally, extraordinary numbers of options smell of bad ui, but it's possible, see e.g. #8019 - 2500 options is definitely too much, but we should test if there's a visible regression for first render time with say, 200 options.

I need to set up some performance tests for this refactoring to check the impact of these changes. One of the main benefits this change provides is that the option list is no longer touched at all when the view or model are changing - only if the data that defines the options changes.

>In the breaking changes you say that the application shouldn't rely on the value of the option. However if you use ngOptions to render, but submit the form to the server via action="", then your backend will probably rely on the correct select value. I guess the number of people who do this is very low, though.

In `ngOptions` the value of each option is supposed to be a surrogate key - i.e. it should not be relied upon. If people do rely on this then they could recreate it by using a `track by` expression, since the value of each option is the result of evaluating `track by` for each option item.

>now that ngOptions and select are two different directives, could this cause problems of interoperability with other directives? For example, if someone attached a directive to a select that is supposed to run after the select is completely initialized, could this break?

There was actually always an `ngOptions` directive but all it did was terminate the compilation to prevent `<option>` elements from compiling and trying to interact with the `SelectController`. A main feature of this refactoring is that the logic for setting up `ngOption` based selects has moved from the `select` directive to the `ngOptions` directive.

petebacondarwin:
The test case I am using looks like this:

```js
<!DOCTYPE html>
<html>

  <head>
    <script src="path/to/angular.js"></script>
      angular.module('app', [])

      .controller('C', function($scope) {
        $scope.items = [];
        $scope.count = 10000;

        $scope.changeOptions = function() {
          $scope.items = [];
          for(var i=0; i<$scope.count; ++i) {
            $scope.items.push({
              id: i,
              label: 'item-' + i,
              group: 'group-' + i % 100
            });
          }
        };
      });
    </script>
  </head>

  <body ng-app="app" ng-controller="C">
    <h1>refactored</h1>
    <input ng-model="count" type="text"><button ng-click="changeOptions()">Change Options</button>
    <br/>
    <select ng-model="x" ng-options="a as a.label group by a.group for a in items track by a.id"></select>
    {{ x }}
  </body>

</html>
```

src/ng/directive/select.js
```diff
@@ -745,31 +249,30 @@ var optionDirective = ['$interpolate', function($interpolate) {
...
-              selectCtrl.removeOption(oldVal);
-            }
-            selectCtrl.addOption(newVal, element);
+          if (interpolateFn) {
+            scope.$watch(interpolateFn, function interpolateWatchAction(newVal, oldVal) {
+              attr.$set('value', newVal);
+              if (oldVal !== newVal) {
+                selectCtrl.removeOption(oldVal);
+              }
+              selectCtrl.addOption(newVal, element);
```
lgalfaso: two lines up, we are removing an option (and calling $render) and then adding an option (and calling $render again)

Narretz:
@petebacondarwin Do you plan on benchmarking the render time with benchpress? Visually I don't see a huge difference in render time with your template.

src/ng/directive/select.js
Narretz: Does It make sense to move the call to `$render`out of  `add/removeOption`? With many repeated options you have quite a few calls to $render. Imho, it's also a good idea if the add/remove functions don't do any additional work.

src/ng/directive/ngOptions.js
```diff
@@ -0,0 +1,593 @@
...
+
+
+      function skipEmptyAndUnknownOptions(current) {
+        var emptyOption_ = emptyOption && emptyOption[0];
+        var unknownOption_ = unknownOption && unknownOption[0];
+
+        if (emptyOption_ || unknownOption_) {
+          while (current &&
+                current === emptyOption_ ||
+                current === unknownOption_) {
```
gkalpak: Is this really supposed to mean `(current && current === emptyOption) || current === unknownOption` ?

src/ng/directive/select.js
```diff
@@ -1,6 +1,113 @@
...
+    var count = optionsMap.get(value) || 0;
+    optionsMap.put(value, count + 1);
+    self.ngModelCtrl.$render();
+  };
+
+  // Tell the select control that an option, with the given value, has been removed
+  self.removeOption = function(value) {
+    var count = optionsMap.get(value);
+    if (count) {
+      if (count == 1) {
```
gkalpak: Since `optionsMap` only seem to store integer values, shouldn't we use `===` instead ?
Do I miss something ?

src/ng/directive/select.js
```diff
@@ -344,400 +202,46 @@ var selectDirective = ['$compile', '$parse', function($compile,   $parse) {
...
+    // automatically select the new element
+    if (optionElement[0].hasAttribute('selected')) {
+      optionElement[0].selected = true;
+    }
+  }
 
   return {
     restrict: 'E',
     priority: 100,
+    require: '?^select',
```
gkalpak: Minor, but using `^^` (instead of `^`) would save one level of checking.

test/ng/directive/ngOptionsSpec.js
```diff
@@ -0,0 +1,1731 @@
...
+
+      scope.$apply(function() {
+        scope.selected = scope.values[3];
+      });
+
+      expect(element.val()).toEqual('4');
+    });
+
+
+    it('should bind to scope value through experession', function() {
```
gkalpak: Type: _experession_

test/ng/directive/ngOptionsSpec.js
```diff
@@ -0,0 +1,1731 @@
...
+          subItem: {label: 'new twenty', id: 20}
+        }];
+      });
+      expect(element.val()).toEqual(['10']);
+      expect(scope.selected[0].id).toEqual(10);
+      expect(scope.selected.length).toBe(1);
+    });
+
+
+    it('It should use the "value" variable to represent items in the array as well as for the ' +
```
gkalpak: Redundant: `It `

test/ng/directive/ngOptionsSpec.js
```diff
@@ -0,0 +1,1731 @@
...
+        var errors = [];
+
+        if (multiple) {
+          value = value.map(function(val) { return hashKey(val); });
+        } else {
+          value = hashKey(value);
+        }
+
+        if (!equals(this.actual.val(), value)) {
+          errors.push('Expected select value "' + this.actual.val() + '" to equal "' + hashKey(value) + '"');
```
gkalpak: Wouldn't this result in re-hashing the value ? Is this intended ?

src/ng/directive/ngOptions.js
petebacondarwin: No :-)

src/ng/directive/select.js
petebacondarwin: Yes :-)

src/ng/directive/select.js
petebacondarwin: I forgot about that!

src/ng/directive/select.js
petebacondarwin: $render is now pretty cheap but I think you are probably right here.

test/ng/directive/ngOptionsSpec.js
petebacondarwin: Another good spot! Originally I was hashing inside the if statement but not now.

test/ng/directive/ngOptionsSpec.js
petebacondarwin: +1

petebacondarwin:
Awesome bit of review @gkalpak - thank you. I will address those issues over the weekend.

src/ng/directive/select.js
petebacondarwin: Actually Igor originally put in an even more aggressive optimization in the original, which I guess we should use.

src/ng/directive/select.js
petebacondarwin: I fixed this too.

lgalfaso:
LGTM, @petebacondarwin can you land this?

petebacondarwin:
Landing now.

PR #10644, opened 2015-01-05
fix(dateFilter): ignore invalid dates

### LINKED ISSUE #10640: Date filter breaks consistency with Angular
If I have something like this `{{day | date:'mediumDate'}}` where `day` could be an empty string `''`, angular is supposed to produce nothing for invalid values. However, I see `undefined NaN, NaN`, which goes against Angular's consistent "if the input is invalid, Angular won't throw errors, it will just output nothing." Is that not the case for the date filter?

How can I still get an empty string `''` to be outputted? Do I have to add a conditional inside the expression?

pkozlowski-opensource:
Fixes #10640

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -459,7 +459,7 @@ function dateFilter($locale) {
       date = new Date(date);
     }
 
-    if (!isDate(date)) {
+    if (!isDate(date) || !isFinite(date.getTime())) {
       return date;
     }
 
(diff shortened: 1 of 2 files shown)


lgalfaso:
LGTM, but please rebase and do not include 35aec525ef8ac6eb1e274e89627c5c04fbbaa3df into the merge (sorry for my mess)

pkozlowski-opensource:
@lgalfaso rebased

PR #10645, opened 2015-01-05
fix(ngRepeat): Track By function conflict

FrankReDo:
Added 'var' to make sure the Track By function/closure is attached to each instance.
Otherwise all directives of the same type call the same Track By function, with the same arguments/closure. (the last one created).
This problem might affect other types of track by expressions variables, as they are declared in the compile method. (line 275)

### DIFF
--- a/src/ng/directive/ngRepeat.js
+++ b/src/ng/directive/ngRepeat.js
@@ -289,7 +289,7 @@ var ngRepeatDirective = ['$parse', '$animate', function($parse, $animate) {
       return function ngRepeatLink($scope, $element, $attr, ctrl, $transclude) {
 
         if (trackByExpGetter) {
-          trackByIdExpFn = function(key, value, index) {
+          var trackByIdExpFn = function(key, value, index) {
             // assign key, value, and $index to the locals so that they can be used in hash functions
             if (keyIdentifier) hashFnLocals[keyIdentifier] = key;
             hashFnLocals[valueIdentifier] = value;
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -607,8 +607,8 @@ function $RouteProvider() {
               return $q.all(locals);
             }
           }).
-          // after route change
           then(function(locals) {
+            // after route change
             if (nextRoute == $route.current) {
               if (nextRoute) {
                 nextRoute.locals = locals;


FrankReDo:
I've signed the CLA

Narretz:
Okay - would it be possible for you to create an example that shows the bug in a plnkr.co or similar? That would also help.

pacarole:
+1 on a fix for this. it's causing dupe errors on track by expressions. i have in manually patched right now which is making very difficult to keep updating angular.

Narretz:
Unfortunately, I still cannot reproduce this. We really need a reproduction before we can write a test and merge this.

pacarole:
Here's a repro scenario:

Say we have some service tracker. That service generates a unique ID based on a prefix and an ID of some object.

```javascript
// this is exposed as part of the tracker service
function generateID(prefix, item) {
    return prefix + '_' + item.id;
}
```

We have another service called dataManager that has an array of objects called items.

```javascript
// this is exposed as part of the dataManager
var item1 = {
    id: 1,
    value: "Great Job!";
};

var item2 = {
    id: 2,
    value: "Amazing!"
};

var items = [item1, item2];
```

Then we have two separate templates that both refer to the dataManager service. So that we don't get a duplicate error, we make sure to generate unique IDs for the objects:

```
// this appears in one directive called Congrats
<div class="congrats">
    <span ng-repeat="item in dataManager.items track by tracker.generateID("congrats", item)">{{item.value}}</span>
</div>

// this appears in another directive called Phrases
<div class="phrases">
    <span ng-repeat="item in dataManager.items track by tracker.generateID("phrases", item)">{{item.value}}</span>
</div>
```

Although these directives generate separate IDs, the bug in ngRepeat.js means that the track by expression could get overwritten and they could both end up calling the same function, which would cause duplicate issues.

Does that make sense Narretz? It's a simple variable scope issue that needs to be fixed.

Narretz:
Thanks, I was able to reproduce the error. :+1:  I didn't doubt there was an error here, I just couldn't reproduce it. And without reproduction, I cannot write a test, and without a test, we don't commit bug fixes.

pacarole:
:+1: great, thanks!

Narretz:
Hi @pacarole I'm sorry for getting your hopes up, but it turns out I had an error in my test. I couldn't reproduce the issue with your info after all. Please modify the following plnkr, so we can get to the bottom of this:
http://plnkr.co/edit/K786PYka8JiJLhHihUoE?p=preview

Narretz:
I'm closing this because we have reproduce scenario / tes. Feel free to reopen if you can provide one.

PR #10646, opened 2015-01-05
fix($templateRequest): cache downloaded templates as strings

### LINKED ISSUE #10630: Beginning with angular-1.3.6 $templateRequest does not populate $templateCache
I upgraded from angular-1.3.0 to angular-1.3.8 and noticed that $templateCache.get(template) was not consistently returning a template string.  Without getting into the code I determined the issue was introduced in angular-1.3.6.

I use $templateRequest to load templates into the $templateCache. Beginning with angular-1.3.6, the first call to $templateCache.get(template) returns the template string. All subsequent calls return an array in the following form [200, "template string", Object, "OK"].

pkozlowski-opensource:
Fixes #10630

### DIFF
--- a/src/ng/templateRequest.js
+++ b/src/ng/templateRequest.js
@@ -42,6 +42,7 @@ function $TemplateRequestProvider() {
 
       return $http.get(tpl, httpOptions)
         .then(function(response) {
+          $templateCache.put(tpl, response.data);
           self.totalPendingRequests--;
           return response.data;
         }, handleError);
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -607,8 +607,8 @@ function $RouteProvider() {
               return $q.all(locals);
             }
           }).
-          // after route change
           then(function(locals) {
+            // after route change
             if (nextRoute == $route.current) {
               if (nextRoute) {
                 nextRoute.locals = locals;
--- a/test/ng/templateRequestSpec.js
+++ b/test/ng/templateRequestSpec.js
@@ -17,7 +17,7 @@ describe('$templateRequest', function() {
   }));
 
   it('should cache the request to prevent extra downloads',
-    inject(function($rootScope, $templateRequest, $httpBackend) {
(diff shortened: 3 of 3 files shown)


caitp:
FWIW I don't think this is the right fix at all =)

caitp:
@pkozlowski-opensource lets just add a method to `$templateCache` like `getTemplate()` or something, which will dig the string out of the response array.

pkozlowski-opensource:
@caitp I'm also not happy about how this code looks like but I think it is less evil than adding new API and duplicating logic from $http. Actually I was really surprised to see that we've got those checks for an array / string inside $http...

I don't feel strongly about this fix at all (I think it is kind of corner case) but I think that it does what most people would expect...

caitp:
If you don't want to add a new function, then I think we should change the approach taken by $http and store the "full response" cache in a different place somehow.

shahata:
The storing of the promise while the request is still pending should also be in this "different place" then... But what is this different place? A cache key with a special prefix that can't exist in url's? Sounds kinda hackish...

caitp:
a different cache object. it would be best to use WeakMaps to map between http's internal cache and the used cache object, but it's not really an option unfortunately.

pkozlowski-opensource:
Soooooo... @caitp / @petebacondarwin / @shahata / whoever - could one of you guys make a call on this one? Funnily enough I had someone few days back who bumped into this one but the use-case was debatable and I agree that people shouldn't be just using $templateCache as a synchronous substitute of the API that is asynchronous by nature ($templateCache).

Personally I don't feel too strong about this one, let's just make some decision.

petebacondarwin:
OK, so I think we should land this since clearly users are doing things this way. It is kind of reasonable to want to access the cached template from inside the then handler of `$templateRequest` since arguably at that point it should be there.

For the record, the `$templateCache` is not designed to be accessed directly. You should always go through the `$templateRequest` since this the way to ensure that you do get the template you want - for example, something you don't have control over in your app might clear the cache in the background.

We should merge this and add it as a lesson learned to be very careful about what appear to public APIs.

PR #10647, opened 2015-01-05
fix(ngRepeat): don't clone DOM for holes in sparse arrays

### LINKED ISSUE #10544: ngRepeat fails with sparse Arrays
ngRepeat iterates over the spaces in a sparse array. If the array is missing one element this leads to an extra element being created by ngRepeat. If the array has more than one element missing the ngRepeat fails because it sees the missing elements as duplicates. 

See Plunk: http://plnkr.co/edit/uvzROas6NHVzOzmfSoJi?p=preview

This is in contrast to the behavior of for(x in array) which only iterates over the elements that are actually present in the array.

Tested with angularjs 1.3.8.

caitp:
Previously, ngRepeat would clone DOM for holes in sparse arrays. Now, holes are ignored. The $index
is translated to the index of the repeated clone, rather than the original array $index, so that
$even and $odd still work correctly.

Closes #10544

### DIFF
--- a/src/ng/directive/ngRepeat.js
+++ b/src/ng/directive/ngRepeat.js
@@ -348,17 +348,28 @@ var ngRepeatDirective = ['$parse', '$animate', function($parse, $animate) {
           collectionLength = collectionKeys.length;
           nextBlockOrder = new Array(collectionLength);
 
+          var k = 0;
+          var keyIndex;
           // locate existing items
           for (index = 0; index < collectionLength; index++) {
-            key = (collection === collectionKeys) ? index : collectionKeys[index];
+            if (collection === collectionKeys) {
+              // Do not iterate over holes in arrays or array-like values
+              if (!(index in collection)) continue;
+              key = index;
+              keyIndex = k++;
+            } else {
+              key = collectionKeys[index];
+              keyIndex = index;
+            }
+
             value = collection[key];
-            trackById = trackByIdFn(key, value, index);
+            trackById = trackByIdFn(key, value, keyIndex);
             if (lastBlockMap[trackById]) {
               // found previously seen block
               block = lastBlockMap[trackById];
               delete lastBlockMap[trackById];
               nextBlockMap[trackById] = block;
-              nextBlockOrder[index] = block;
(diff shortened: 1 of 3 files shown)


src/ng/directive/ngRepeat.js
```diff
@@ -389,11 +400,22 @@ var ngRepeatDirective = ['$parse', '$animate', function($parse, $animate) {
...
+              keyIndex = k++;
+            } else {
+              key = collectionKeys[index];
+              keyIndex = index;
+            }
+
             value = collection[key];
-            block = nextBlockOrder[index];
+            block = nextBlockOrder[keyIndex];
+            if (!block) continue;
```
lgalfaso: why is this needed?

caitp: leftover

lgalfaso:
I am torn if we should do this, for sure I would not oppose. That said, if we want to move forward, then it LGTM

caitp:
yeah, like I said I'm not sure this is worth supporting. I'm talking to @petebacondarwin about a new process for things like this so we can get more feedback from the community on this stuff and see if people really feel like they need it

lgalfaso:
I thought about this again, now I lean towards not to merge this. The reason being that `item === collection[$index]` would no longer be true (and people that would like to use the previous/next element would not be able to depend on `collection[$index +/- 1]` )

lgalfaso:
one more thing, it looks like the property `$last` is not respected when `collection.length` is bigger than the key for the last element

petebacondarwin:
It looks like there are problems with this approach. We need to change the algorithm or explicitly not support this case: the workaround being that instead of sparse arrays you use an object: http://plnkr.co/edit/ghIfW9n6OD3I9XslrP0A?p=preview

jbedard:
Wouldn't a filter be able to fix this? It wouldn't fix ng-repeat itself, but it would avoid making ng-repeat more complicated and I think handles the issues raised here ($last, $index, etc.).
```html
<element ng-repeat="item in sparse | removeUndefined as nonSparse">
```

PR #10651, opened 2015-01-06
fix(ngClass): handle multi-class definitions as an element of an array

### LINKED ISSUE #8578: ngClass array syntax behavior not according to documentation
Is this a bug or should the documentation be updated?

Documentation:
> If the expression evaluates to an array, each element of the array should be a string that is one or more space-delimited class names.

However if an array element has multiple, space-delimited, classes only the last one appears to be applied. This behavior can be seen in the example in the documentation for ngClass.

Moreover; the example code for ngClass also suggest that an array element can only have one class - the placeholder text is "Type: bold, strike or red".

lgalfaso:
Handles multi-class definition as an element of an array

Closes #8578

### DIFF
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -97,7 +97,7 @@ function classDirective(name, selector) {
 
     function arrayClasses(classVal) {
       if (isArray(classVal)) {
-        return classVal;
+        return classVal.join(' ').split(' ');
       } else if (isString(classVal)) {
         return classVal.split(' ');
       } else if (isObject(classVal)) {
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@lgalfaso hmm, I've got the impression that it won't work in all the case - since you can have expressions as part of the ngClass array you can imagine having spaces being part of those expressions - one contrived example: `<div ng-class="['A', 'B', dynamic('C D')]"></div>`.

Didn't look deeper, but if you believe that this works for this case as well, we should add more tests.

pkozlowski-opensource:
Oh, wait, on the directive level we are going to have already an array of strings, right?

pkozlowski-opensource:
@lgalfaso sorry for the noise. LGTM. Merging.

PR #10652, opened 2015-01-06
Get rid of unused variable

karlcoelho:
It may not cause a large problem for the project, but in the long-term, it might disrupt the ng-messages module, which is quite popular (All my angular apps have them), so I cleaned it up.


karlcoelho:
signed the CLA!

PR #10653, opened 2015-01-06
feat(ngRoute): screening colon in given path

stervets:
Use backslash symbol before colon to prevent it interpretation as variable.
For example: path "/\:foo/:bar" returns only "bar" variable

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -191,35 +191,39 @@ function $RouteProvider() {
     *
     * Inspired by pathRexp in visionmedia/express/lib/utils.js.
     */
-  function pathRegExp(path, opts) {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I think that a better description of this feature is "allow colons to be escaped in route definitions".
That being said, colons are reserved characters in URLs anyway. And so they should be escaped anyway as `%3A`. We should check that this works and then document it.

gkalpak:
According to [RFC2396][1] - although reserved - `:` is allowed (unescaped) in fragments, which is what we use when in HashBang mode (currently the default).
Considering that (and the fact that I couldn't make `%3A` work, I think it makes sense to support escaping `:`.

That said, we need better tests and docs.

[1]: http://tools.ietf.org/html/rfc2396#appendix-A

PR #10656, opened 2015-01-06
docs($window): improve the example

thorn0:


src/ng/window.js
```diff
@@ -22,22 +22,37 @@
          angular.module('windowExample', [])
            .controller('ExampleController', ['$scope', '$window', function($scope, $window) {
              $scope.greeting = 'Hello, World!';
-             $scope.doGreeting = function(greeting) {
-               $window.alert(greeting);
+             $scope.doGreeting = function() {
+               $window.alert($scope.greeting);
+               $scope.greeting = '';
```
gkalpak: What's the purpose of this line ?

src/ng/window.js
```diff
@@ -22,22 +22,37 @@
...
+       beforeEach(module(function($provide) {
+         // Mock $window
+         $provide.decorator('$window', function($delegate) {
+           // Use the original window as a prototype for the mock in end-to-end
+           // tests so that the parts of Angular that interact with the browser
+           // via $window won't break.
+           var decoratedWindow = Object.create($delegate);
+           // The browser's implementation of the alert function blocks the test
+           // runner, so replace it in the mock.
+           decoratedWindow.alert = this.alert = jasmine.createSpy('alert');
```
gkalpak: I am not keen on `this.alert`.

src/ng/window.js
```diff
@@ -22,22 +22,37 @@
...
+           return decoratedWindow;
+         });
+       }));
+       it('should clear the input field after alert', function() {
+         element(by.model('greeting')).sendKeys('Hello, E2E Tests');
+         element(':button').click();
+         expect(this.alert).toHaveBeenCalled();
```
gkalpak: Why not inject `$window` and check up on `$window.alert` instead of having `this.alert`.

src/ng/window.js
thorn0: Oops. Yes, in that place `this` doesn't mean Jasmine's `this`...

src/ng/window.js
thorn0: Just some logic to test it in the test.

PR #10663, opened 2015-01-07
docs(guide/Using $location): update link to HTML5 history API

mkllnk:
The old link http://www.w3.org/TR/html5/introduction.html#history-0 points to the history of HTML5. The new link points to the API of the `history` object. Additionally, this change normalises the upper/lower case of the phrase 'HTML5 history API'.

### DIFF
--- a/docs/content/guide/$location.ngdoc
+++ b/docs/content/guide/$location.ngdoc
@@ -165,7 +165,7 @@ encoded.
 
 `$location` service has two configuration modes which control the format of the URL in the browser
 address bar: **Hashbang mode** (the default) and the **HTML5 mode** which is based on using the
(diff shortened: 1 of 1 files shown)


mkllnk:
@googlebot Sorry to not have a Google account. My change is so tiny that it's public domain.

I don't see a way of agreeing without a Google account. If so, that's my first and last contribution. Thanks.

pkozlowski-opensource:
@mkllnk I'm afraid we can't merge PRs without the associated CLA. While it will be sad to not have your change integrated, I can understand your reasons....

Going to close this PR but would be happy to integrate it if you change your mind....

mkllnk:
And I can't sign without a Google account?

PR #10664, opened 2015-01-07
fix(ngChecked): ensure that ngChecked doesn't interfere with ngModel

### LINKED ISSUE #10662: Bug in checkbox
I use checkbox like this:
```
[label][input type="checkbox" ng-checked="go" ng-true-value="true" ng-false-value="false" ng-model="go"][/input]Send out email notification[/label]
```

the default value of parameter "go" is "true", when I un-checked the checkbox,the value of go is still "true" when I use console.log(go) to trace it. And when I check and then un-check it angain, its value changed.

LeonardoBraga:
Fixes #10662

### DIFF
--- a/src/ng/directive/attrs.js
+++ b/src/ng/directive/attrs.js
@@ -341,22 +341,34 @@
 
 var ngAttributeAliasDirectives = {};
 
-
 // boolean attrs are evaluated
 forEach(BOOLEAN_ATTR, function(propName, attrName) {
   // binding to multiple is not supported
   if (propName == "multiple") return;
 
+  function defaultLinkFn(scope, element, attr) {
+    scope.$watch(attr[normalized], function ngBooleanAttrWatchAction(value) {
+      attr.$set(attrName, !!value);
+    });
+  }
+
   var normalized = directiveNormalize('ng-' + attrName);
+  var linkFn = defaultLinkFn;
+
+  if (propName === 'checked') {
+    linkFn = function(scope, element, attr) {
+      // ensuring ngChecked doesn't interfere with ngModel when both are set on the same input
+      if (attr.ngModel !== attr[normalized]) {
+        defaultLinkFn(scope, element, attr);
+      }
+    };
+  }
+
   ngAttributeAliasDirectives[normalized] = function() {
     return {
       restrict: 'A',
       priority: 100,
-      link: function(scope, element, attr) {
-        scope.$watch(attr[normalized], function ngBooleanAttrWatchAction(value) {
-          attr.$set(attrName, !!value);
-        });
-      }
+      link: linkFn
     };
   };
 });
--- a/test/ng/directive/booleanAttrsSpec.js
+++ b/test/ng/directive/booleanAttrsSpec.js
@@ -42,6 +42,22 @@ describe('boolean attr directives', function() {
   }));
 
 
+  it('should not bind checked when ngModel is present', inject(function($rootScope, $compile) {
+    // test for https://github.com/angular/angular.js/issues/10662
+    element = $compile('<input type="checkbox" ng-model="value" ng-false-value="\'false\'" ' +
+      'ng-true-value="\'true\'" ng-checked="value" />')($rootScope);
+    $rootScope.value = 'true';
+    $rootScope.$digest();
+    expect(element[0].checked).toBe(true);
+    browserTrigger(element, 'click');
+    expect(element[0].checked).toBe(false);
+    expect($rootScope.value).toBe('false');
+    browserTrigger(element, 'click');
+    expect(element[0].checked).toBe(true);
+    expect($rootScope.value).toBe('true');
+  }));
+
+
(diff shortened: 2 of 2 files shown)


src/ng/directive/attrs.js
```diff
@@ -341,22 +341,33 @@
...
+  }
+
+  var normalized = directiveNormalize('ng-' + attrName),
+    linkFn = defaultLinkFn;
+
+  if (propName === 'checked') {
+    linkFn = function(scope, element, attr) {
+      // ensuring ngChecked doesn't interfere with ngModel when both are set on the same input
+      // - https://github.com/angular/angular.js/issues/10662
+      !attr.ngModel && defaultLinkFn(scope, element, attr);
```
caitp: or better, don't even set up a link function if `attr.ngModel === attr[normalized]`

caitp: eh nevermind, I guess that doesn't work :(

caitp: However, for code style, please write it like this:

```js
if (attr.ngModel !== attr[normalized) {
  defaultLinkFn(scope, element, attr);
}
```

LeonardoBraga: @caitp Done. Please let me know if something else requires adjustment.

src/ng/directive/attrs.js
```diff
@@ -341,22 +341,35 @@
...
+  }
+
+  var normalized = directiveNormalize('ng-' + attrName),
+    linkFn = defaultLinkFn;
+
+  if (propName === 'checked') {
+    linkFn = function(scope, element, attr) {
+      // ensuring ngChecked doesn't interfere with ngModel when both are set on the same input
+      // - https://github.com/angular/angular.js/issues/10662
+      if (attr.ngModel !== attr[normalized) {
```
caitp: typo ^^

caitp: kinda feel like special casing different attributes in here is kind of gross, too --- but we can worry about that later

LeonardoBraga: :-$ - sorry, it's fixed now.

LeonardoBraga: > kinda feel like special casing different attributes in here is kind of gross, too --- but we can worry about that later

Sure. Just ping if you want me to consider that in another PR.

caitp:
lgtm

caitp:
@naretz am I missing anything? could you give this a second look before merging thanks

src/ng/directive/attrs.js
```diff
@@ -341,22 +341,35 @@
...
 
-  var normalized = directiveNormalize('ng-' + attrName);
+  function defaultLinkFn(scope, element, attr) {
+    scope.$watch(attr[normalized], function ngBooleanAttrWatchAction(value) {
+      attr.$set(attrName, !!value);
+    });
+  }
+
+  var normalized = directiveNormalize('ng-' + attrName),
+    linkFn = defaultLinkFn;
```
Narretz: Can you explicitly declare the linkFn variable? I think we prefer this now.

src/ng/directive/attrs.js
```diff
@@ -341,22 +341,35 @@
...
+
+  if (propName === 'checked') {
+    linkFn = function(scope, element, attr) {
+      // ensuring ngChecked doesn't interfere with ngModel when both are set on the same input
+      // - https://github.com/angular/angular.js/issues/10662
```
Narretz: You can remove the issue link - we almost never do include this info

src/ng/directive/attrs.js
LeonardoBraga: So `linkFn` would have it's own `var`? I'm asking because it differs with other places I saw in the code. Just let me know if that's indeed the case and I'll change it.

Narretz:
Apart from these small nitpicks, this looks good to me. I also didn't know we had this directive ...

src/ng/directive/attrs.js
Narretz: Yes, that's what I mean. It's a minor thing - but recent code changes favour this version.

caitp: agree with @Narretz, we prefer to give each variable it's own `var`, it's a bit easier than dealing with a huge comma-separated list (see compile.js for examples :x) =)

LeonardoBraga: Not a problem! :-) Thanks for the instructions and I'll be updating the PR soon.

caitp:
lgtm, will merge when green

LeonardoBraga:
@caitp Travis seems flaky again. Should I close + reopen to force a re-test?

caitp:
no need, restarted manually

LeonardoBraga:
Could you please restart again? :-/

caitp:
it's just browserstack having issues, it's not failing in a consistent way, so I think it's fine. chock it up to flakes

caitp:
@petebacondarwin I think this is a good candidate for cherry-picking into 1.3, but can you make the call on that?

PR #10667, opened 2015-01-07
docs(ngModelOptions): Config objects cannot be shared among ngModelOptions instances.

mtr:
Just thought I'd add a note about this issue since it hit me during development and I cannot find any note about it in the documentation.  Looking at the source code, it seems that the changes to the object made in the block of https://github.com/angular/angular.js/blob/master/src/ng/directive/ngModel.js#L1223 are not idempotent, but I guess that's a design choice, not a bug?

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -1082,6 +1082,10 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
  * Any pending changes will take place immediately when an enclosing form is submitted via the
  * `submit` event. Note that `ngClick` events will occur before the model is updated. Use `ngSubmit`
  * to have access to the updated model.
+ * 
+ * Multiple `ngModelOptions` instances cannot share the same options Object. For example, supplying a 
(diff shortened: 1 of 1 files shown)


mtr:
A few moments ago, I signed the CLA.

pkozlowski-opensource:
@mtr let's see what others think but generally speaking I'm against dotting docs with all the possible use-cases that people might find confusing. If this scenario is common we can fix it in the code we should fix the underlying issue. If it is not fixable and very common only then we should add a note to the docs, IMO

mtr:
@pkozlowski-opensource That makes sense.  I think the underlying code should be fixed then, and I think I see how that can be done quite easily, but I'm not sure when I'll get the time to do it.  If I find the time, I'll submit a pull request for fixing the underlying issue.

pkozlowski-opensource:
@mtr ok, send a PR for the code fix when you've got time then. Or at least open an issue with the reproduce scenario so maybe someone else will be able to fix it faster.

Closing this PR for now.

PR #10673, opened 2015-01-07
fix(ngModelOptions): allow sharing options between multiple inputs

shahata:
Closes #10667

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -1215,7 +1215,7 @@ var ngModelOptionsDirective = function() {
     restrict: 'A',
     controller: ['$scope', '$attrs', function($scope, $attrs) {
       var that = this;
-      this.$options = $scope.$eval($attrs.ngModelOptions);
+      this.$options = angular.copy($scope.$eval($attrs.ngModelOptions));
       // Allow adding/overriding bound events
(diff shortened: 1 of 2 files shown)


lgalfaso:
landed as 9c9c6b3fe4edfe78ae275c413ee3eefb81f1ebf6

src/ng/directive/ngModel.js
```diff
@@ -1215,7 +1215,7 @@ var ngModelOptionsDirective = function() {
     restrict: 'A',
     controller: ['$scope', '$attrs', function($scope, $attrs) {
       var that = this;
-      this.$options = $scope.$eval($attrs.ngModelOptions);
+      this.$options = angular.copy($scope.$eval($attrs.ngModelOptions));
```
jbedard: Really minor, and a bit late, but this should just be `copy`...

lgalfaso: You are right, should be able to make the change in about an hour (unless
someone beats me)
>
>

lgalfaso: done and thanks

shahata: Thanks & Sorry. I always keep forgetting not to use `angular.*` :(

PR #10674, opened 2015-01-07
fix(ngModelOptions): Allow sharing ngModelOptions objects between multiple directive instances, even when they specify default behavior.

mtr:
Fixes #10667.


pkozlowski-opensource:
Thnx for this PR @mtr but it looks like the issue was just solved via 9c9c6b3fe4edfe78ae275c413ee3eefb81f1ebf6

PR #10676, opened 2015-01-07
feat(ngMessages): provide support for dynamic message resolution

### LINKED ISSUE #9338: ngMessages blows up when there are duplicate child ngMessage keys
The `ngMessages` directive gets into an invalid state and throws an exception when there are duplicate `ngMessage` child keys specified. For example:

```html
<div ng-messages="form.id.$error">
  <div ng-message="required">
    User ID is required.
  </div>
  <div ng-message="required">
    User ID is required.
  </div>
</div>
```

The issue seems to be a bug in the re-ordering logic in `registerMessage` which is attempting to preserve the ordering from an included template, but in this scenario inserts `null` into `messages`.

Test case here: http://plnkr.co/edit/7u4oFTGbeLqa1tmfMNlJ (look at the console log).

### LINKED ISSUE #10036: Feature Request: Dynamically add ng-message to ng-messages
Add the ability to programaticaly add ngMessage item to an existing ngMessages directive.

I am trying to show the field error returned by an AJAX form submit. This way frontend errors and backend errors displaying would be done the same way.

Do you this is the right approach or you would just create another way to insert the server side error in the DOM?

matsko:
# What's the goal here?
1. To allow multiple message states to use the same message.
2. To allow structural directives such as ngRepeat, ngIf and ngSwitch to add and remove instances of ngMessage within a ngMessage container.
3. To allow form controller code to download and dynamically populate message instances using validations from the server.

-- commit message --
Prior to this fix it was impossible to apply a binding to a the
ngMessage directive to represent the name of the error. It was also
not possible to use ngRepeat or any other structural directive to
dynamically update the list of messages. This fix ensures that both
the ngMessage and ngMessages directives automatically update when
any dynamic message data changes.

BREAKING CHANGE:

The `ngMessagesInclude` attribute is now its own directive and that must
be placed as a **child** element within the element with the ngMessages
directive. (Keep in mind that the former behaviour of the
ngMessageInclude attribute was that all **included** ngMessage template
code was placed at the **bottom** of the element containing the
ngMessages directive; therefore to make this behave in the same way,
place the element containing the ngMessagesInclude directive at the
end of the container containing the ngMessages directive).

``` html
<!-- AngularJS 1.3.x -->
<div ng-messages="model.$error" ng-messages-include="remote.html">
  <div ng-message="required">Your message is required</div>
</div>

<!-- AngularJS 1.4.x -->
<div ng-messages="model.$error">
  <div ng-message="required">Your message is required</div>
  <div ng-messages-include="remote.html"></div>
</div>
```

Closes #10036
Closes #9338

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -1,5 +1,13 @@
 'use strict';
 
+/* jshint ignore:start */
+// this code is in the core, but not in angular-messages.js
+var isArray = angular.isArray;
+var forEach = angular.forEach;
+var isString = angular.isString;
+var jqLite = angular.element;
+/* jshint ignore:end */
+
 /**
  * @ngdoc module
  * @name ngMessages
@@ -12,8 +20,8 @@
  * `ngMessage` directives are designed to handle the complexity, inheritance and priority
  * sequencing based on the order of how the messages are defined in the template.
  *
- * Currently, the ngMessages module only contains the code for the `ngMessages`
- * and `ngMessage` directives.
+ * Currently, the ngMessages module only contains the code for the `ngMessages`, `ngMessagesInclude`
+ * `ngMessage` and `ngMessageExp` directives.
  *
  * # Usage
  * The `ngMessages` directive listens on a key/value collection which is set on the ngMessages attribute.
@@ -26,7 +34,9 @@
  *   <input type="text" ng-model="field" name="myField" required minlength="5" />
  *   <div ng-messages="myForm.myField.$error">
  *     <div ng-message="required">You did not enter a field</div>
- *     <div ng-message="minlength">The value entered is too short</div>
+ *     <div ng-message="minlength, maxlength">
+ *       Your email must be between 5 and 100 characters long
+ *     </div>
  *   </div>
  * </form>
  * ```
@@ -52,7 +62,11 @@
  * ngMessages directive to make this happen.
  *
  * ```html
+ * <!-- attribute-style usage -->
  * <div ng-messages="myForm.myField.$error" ng-messages-multiple>...</div>
+ *
+ * <!-- element-style usage -->
+ * <ng-messages for="myForm.myField.$error" multiple>...</ng-messages>
  * ```
  *
  * ## Reusing and Overriding Messages
@@ -65,7 +79,10 @@
  *   <div ng-message="required">This field is required</div>
  *   <div ng-message="minlength">This field is too short</div>
  * </script>
- * <div ng-messages="myForm.myField.$error" ng-messages-include="error-messages"></div>
+ *
+ * <div ng-messages="myForm.myField.$error">
+ *   <div ng-messages-include="error-messages"></div>
+ * </div>
  * ```
  *
  * However, including generic messages may not be useful enough to match all input fields, therefore,
@@ -87,12 +104,17 @@
  *          minlength="5"
  *          required />
  *
- *   <div ng-messages="myForm.myEmail.$error" ng-messages-include="my-custom-messages">
+ *   <!-- any ng-message elements that appear BEFORE the ng-messages-include will
+ *        override the messages present in the ng-messages-include template -->
+ *   <div ng-messages="myForm.myEmail.$error">
  *     <!-- this required message has overridden the template message -->
  *     <div ng-message="required">You did not enter your email address</div>
  *
  *     <!-- this is a brand new message and will appear last in the prioritization -->
  *     <div ng-message="email">Your email address is invalid</div>
+ *
+ *     <!-- and here are the generic error messages -->
+ *     <div ng-messages-include="error-messages"></div>
  *   </div>
  * </form>
  * ```
@@ -102,10 +124,65 @@
  * email addresses, date fields, autocomplete inputs, etc...), specialized error messages can be applied
  * while more generic messages can be used to handle other, more general input errors.
  *
+ * ## Dynamic Messaging
+ * ngMessages also supports using expressions to dynamically change key values. Using arrays and
+ * repeaters to list messages is also supported. This means that the code below will be able to
+ * fully adapt itself and display the appropriate message when any of the expression data changes:
+ *
+ * ```html
+ * <form name="myForm">
+ *   <input type="email"
+ *          name="myEmail"
+ *          ng-model="email"
+ *          minlength="5"
+ *          required />
+ *
+ *   <div ng-messages="myForm.myEmail.$error">
+ *     <div ng-message="required">You did not enter your email address</div>
+ *     <div ng-repeat="errorMessage in errorMessages">
+ *       <!-- use ng-message-exp for a message whose key is given by an expression -->
+ *       <div ng-message-exp="errorMessage.type">{{ errorMessage.text }}</div>
+ *     </div>
+ *   </div>
+ * </form>
+ * ```
+ *
+ * The `errorMessage.type` expression can be a string value or it can be an array so
+ * that multiple errors can be associated with a single error message:
+ *
+ * ```html
+ *  <input type="email"
+ *         ng-model="data.email"
+ *         name="myEmail"
+ *         ng-minlength="5"
+ *         ng-maxlength="100"
+ *         required />
+ *   <div ng-messages="myForm.myEmail.$error">
+ *     <div ng-message-exp="'required'">You did not enter your email address</div>
+ *     <div ng-message-exp="['minlength', 'maxlength]">
+ *       Your email must be between 5 and 100 characters long
+ *     </div>
+ *   </div>
+ * ```
+ *
+ * Feel free to use other structural directives such as ng-if and ng-switch to further control
+ * what messages are active and when. Be careful, if you place ng-message on the same element
+ * as these structural directives, Angular may not be able to determine if a message is active
+ * or not. Therefore it is best to place the ng-message on a child element of the structural
+ * directive.
+ *
+ * ```html
+ * <div ng-messages="myForm.myEmail.$error">
+ *   <div ng-if="showRequiredError">
+ *     <div ng-message="required">Please enter something</div>
+ *   </div>
+ * </div>
+ * ```
+ *
  * ## Animations
- * If the `ngAnimate` module is active within the application then both the `ngMessages` and
- * `ngMessage` directives will trigger animations whenever any messages are added and removed
- * from the DOM by the `ngMessages` directive.
+ * If the `ngAnimate` module is active within the application then the `ngMessages`, `ngMessage` and
+ * `ngMessageExp` directives will trigger animations whenever any messages are added and removed from
+ * the DOM by the `ngMessages` directive.
  *
  * Whenever the `ngMessages` directive contains one or more visible messages then the `.ng-active` CSS
  * class will be added to the element. The `.ng-inactive` CSS class will be applied when there are no
@@ -171,7 +248,7 @@ angular.module('ngMessages', [])
     * messages use the `ngMessage` directive and will be inserted/removed from the page depending
     * on if they're present within the key/value object. By default, only one message will be displayed
     * at a time and this depends on the prioritization of the messages within the template. (This can
-    * be changed by using the ng-messages-multiple on the directive container.)
+    * be changed by using the `ng-messages-multiple` or `multiple` attribute on the directive container.)
     *
     * A remote template can also be used to promote message reusability and messages can also be
     * overridden.
@@ -182,23 +259,22 @@ angular.module('ngMessages', [])
     * ```html
     * <!-- using attribute directives -->
     * <ANY ng-messages="expression">
-    *   <ANY ng-message="keyValue1">...</ANY>
-    *   <ANY ng-message="keyValue2">...</ANY>
-    *   <ANY ng-message="keyValue3">...</ANY>
+    *   <ANY ng-message="stringValue">...</ANY>
+    *   <ANY ng-message="stringValue1, stringValue2, ...">...</ANY>
+    *   <ANY ng-message-exp="expressionValue">...</ANY>
     * </ANY>
     *
     * <!-- or by using element directives -->
     * <ng-messages for="expression">
-    *   <ng-message when="keyValue1">...</ng-message>
-    *   <ng-message when="keyValue2">...</ng-message>
-    *   <ng-message when="keyValue3">...</ng-message>
+    *   <ng-message when="stringValue">...</ng-message>
+    *   <ng-message when="stringValue1, stringValue2, ...">...</ng-message>
+    *   <ng-message when-exp="expressionValue">...</ng-message>
     * </ng-messages>
     * ```
     *
     * @param {string} ngMessages an angular expression evaluating to a key/value object
     *                 (this is typically the $error object on an ngModel instance).
     * @param {string=} ngMessagesMultiple|multiple when set, all messages will be displayed with true
-    * @param {string=} ngMessagesInclude|include when set, the specified template will be included into the ng-messages container
     *
     * @example
     * <example name="ngMessages-directive" module="ngMessagesExample"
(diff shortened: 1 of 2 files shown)


caitp:
> This breaks a bunch of stuff...

can we expand on this :)

src/ngMessages/messages.js
```diff
@@ -87,12 +95,17 @@
  *          minlength="5"
  *          required />
  *
- *   <div ng-messages="myForm.myEmail.$error" ng-messages-include="my-custom-messages">
+ *   <!-- any ng-message elements that appear BEFORE the ng-message-incldue will
```
petebacondarwin: typo: `ng-message-incldue`

petebacondarwin:
You could add that being able to apply an array of key names to the `ngMessage` directive is one of the drivers behind the major breaking change that all `ngMessage` attributes are now expressions.

Without this requirement I believe that we could have just used interpolation for binding since the attribute type would have only been a string - but I might be wrong.

src/ngMessages/messages.js
```diff
@@ -228,91 +291,217 @@ angular.module('ngMessages', [])
...
+           if (match) {
+             match.next = messageNode.next;
+           } else {
+             ctrl.head = messageNode.next;
+           }
+         }
+       }],
+
+       link: function(scope, element, attrs, ngMessagesCtrl) {
+         ngMessagesCtrl.render = function(collection) {
```
petebacondarwin: Why is this method not in the controller already?

src/ngMessages/messages.js
```diff
@@ -228,91 +291,217 @@ angular.module('ngMessages', [])
...
+
+           // this is true if the attribute is empty or if the attribute value is truthy
+           var showAll = isAttrTruthy(scope, attrs.ngMessagesMultiple) ||
+                         isAttrTruthy(scope, attrs.multiple);
 
+           var matches = [];
+           var nameMatches = [];
+           var item = ngMessagesCtrl.head;
+           var messageExists = false;
+           while (item != null && !messageExists) {
```
petebacondarwin: `item !== null` ?

petebacondarwin:
Should we move `contains()` outside of the link function?

src/ngMessages/messages.js
matsko: I tried that one. It doesn't work since if there are no messages present them item == undefined which !== null and then the loop never breaks.

src/ngMessages/messages.js
matsko: Moved.

src/ngMessages/messages.js
matsko: fixed.

matsko:
@petebacondarwin fixed contains.

@petebacondarwin The goal for array-based expressions is noted at the top of the PR.

matsko:
@petebacondarwin all code pushed.

petebacondarwin:
Great @matsko - What I meant about the array expressions was that it would be good to include the motivation for the breaking change in the commit message.

src/ngMessages/messages.js
```diff
@@ -330,63 +520,80 @@ angular.module('ngMessages', [])
...
-          var node = parentNode.childNodes[i];
-          if (node.nodeType == COMMENT_NODE && node.nodeValue.indexOf('ngMessage') >= 0) {
-            if (node === commentNode) {
-              index = j;
-              break;
-            }
-            j++;
-          }
-        }
+      require: '^ngMessages',
```
petebacondarwin: This can be `'^^ngMessages'`

src/ngMessages/messages.js
```diff
@@ -217,9 +280,9 @@ angular.module('ngMessages', [])
     *       <pre>myForm.myName.$error = {{ myForm.myName.$error | json }}</pre>
     *
     *       <div ng-messages="myForm.myName.$error" style="color:maroon">
-    *         <div ng-message="required">You did not enter a field</div>
-    *         <div ng-message="minlength">Your field is too short</div>
-    *         <div ng-message="maxlength">Your field is too long</div>
+    *         <div ng-message="'required'">You did not enter a field</div>
+    *         <div ng-message="'minlength'">Your field is too short</div>
+    *         <div ng-message="'maxlength'">Your field is too long</div>
```
caitp: It was mentioned previously, but why is interpolation not good enough for this? Using interpolation would mean we could ship it in 1.3

test/ngMessages/messagesSpec.js
```diff
@@ -189,12 +189,71 @@ describe('ngMessages', function() {
...
+    $rootScope.$apply(function() {
+      // remove the age!
+      $rootScope.items.shift();
+    });
+
+    expect(messageChildren().length).toBe(1);
+    expect(trim(element.text())).toEqual("Your hair is too long");
+
+    $rootScope.$apply(function() {
+      // remove the age!
```
petebacondarwin: Should be `// remove the hair!`

src/ngMessages/messages.js
petebacondarwin: Moreover it "should" be since it will then provide an error message if the user puts it in the wrong place.

src/ngMessages/messages.js
```diff
@@ -228,91 +291,218 @@ angular.module('ngMessages', [])
...
+    * {@link module:ngMessages Click here} to learn more about `ngMessages` and `ngMessage`.
+    *
+    * @param {string} ngMessagesInclude|src a string value corresponding to the remote template.
+    */
+   .directive('ngMessagesInclude',
+     ['$templateRequest', '$document', '$compile', function($templateRequest, $document, $compile) {
+
+     return {
+       restrict: 'AE',
+       require: '^ngMessages',
```
petebacondarwin: This should be `'^^ngMessages'` to prevent the user from adding it to the `<ngMessage>` element

src/ngMessages/messages.js
```diff
@@ -228,91 +291,218 @@ angular.module('ngMessages', [])
...
+       compile: function(element, attrs) {
+         var parent = element.parent();
```
petebacondarwin: `parent` is not used

src/ngMessages/messages.js
```diff
@@ -228,91 +291,218 @@ angular.module('ngMessages', [])
...
+         var comment = angular.element($document[0].createComment(' ngMessagesInclude: '));
+         element.after(comment);
+
+         return function($scope, $element, attrs, ngMessagesCtrl) {
+           // we're removing this since we only need access to the newly
+           // created comment node as an anchor.
+           element.remove();
```
petebacondarwin: Is it safe to remove the linked element here? I guess the tests work but I am sure we always discouraged work outside of the children of the element.

src/ngMessages/messages.js
petebacondarwin: Especially as the old syntax was to put it on the ngMessages element

src/ngMessages/messages.js
```diff
@@ -228,91 +291,218 @@ angular.module('ngMessages', [])
...
+
+           $templateRequest(attrs.ngMessagesInclude || attrs.src).then(function(html) {
+             var elements = angular.element('<div></div>').html(html).contents();
```
petebacondarwin: What is the performance cost of creating a DOM element and then destroying it
over reusing the original `element`?

```js
var elements = element.html(html).contents();
```

src/ngMessages/messages.js
```diff
@@ -228,91 +291,218 @@ angular.module('ngMessages', [])
...
-        };
+           // this is true if the attribute is empty or if the attribute value is truthy
+           var showAll = isAttrTruthy($scope, $attrs.ngMessagesMultiple) ||
+                         isAttrTruthy($scope, $attrs.multiple);
 
-        //JavaScript treats empty strings as false, but ng-message-multiple by itself is an empty string
-        var multiple = angular.isString($attrs.ngMessagesMultiple) ||
-                       angular.isString($attrs.multiple);
+           var matches = [];
+           var nameMatches = [];
```
petebacondarwin: Could you improve performance by using angular `HashMap` objects here for `matches` and `nameMatches`?

petebacondarwin:
@matsko - if you can address the comments above this looks good to me.  A really fascinating bit of work resulting in a much more Angular-ish way of working: not only the ability to use structural elements but also being able to insert `ngMessageInclude` directives at will such that the ordering is taken into account. Nice!

src/ngMessages/messages.js
```diff
@@ -228,91 +291,218 @@ angular.module('ngMessages', [])
...
+    * @usage
+    * ```html
+    * <!-- using attribute directives -->
+    * <ANY ng-messages="expression">
+    *   <ANY ng-messages-include="remoteTplString">...</ANY>
+    * </ANY>
+    *
+    * <!-- or by using element directives -->
+    * <ng-messages for="expression">
+    *   <ng-messages-include src="expressionValue1">...</ng-messages-include>
```
Narretz: Shouldn't this be interpolatedValue1 or even {{interpolatedValue}}?

Narretz:
This looks really useful. A few comments:
- Is there a specific reason why overwriting ng-message works by putting them **before** the includes? This seem counterintuitive to me.
Docs:
- (introduction paragraph): should mention `ngMessageInclude` (maybe make all of them links, too)
- (Usage): `By default, ngMessages will only display one error at a time ... ` can't hurt to explicitly mention that the first message will be displayed
- (Reusing and Overriding Messages): in both script ng-templates , the `ng-message`attributes need to be wrapped in single quotes.

src/ngMessages/messages.js
Narretz: Or, since `ng-message` now uses expressions, would it make sense for the include to use them, too? `ng-include`already uses expressions. Now there's another include directive, but this one uses interpolation.

petebacondarwin:
@Narretz - the ngMessages directive works by displaying the first matching message. Its a case of first-in wins. So putting messages earlier overwrites later messages. This should be made clearer in the docs as you say.

petebacondarwin:
@matsko and @caitp, regarding not having the breaking change of making ng-message expect an expression now.

**Benefits:**

* Not a breaking change so can be back-ported to 1.3.x
* Less verbose syntax for the most frequent use case

**Problems:**

* We would need another attribute for expressions (`ng-message-match` / `ng-message-exp` / `ng-message-props`?)
* We would need to find a way to support interpolation in the original `ng-message` attribute since is not supported right now

matsko:
@petebacondarwin why not just place priority on `ng-message-match` (or prop or exp) and if that attribute is undefined then we place a watcher on `ng-message`. This way we can still interpolate the `ng-message` attr without any performance hit.

gkalpak:
Could someone explain briefly what is the benefit of watching the value of an `ng-message` attribute ?
I understand why I might want to add/remove messages dynamically or dynamically update the content of a message, but why change the "key" ?

petebacondarwin:
@gkalpak - it is necessary if you are using ng-repeat:

```html
<div ng-message="{{ error.key }}Error" ng-repeat="error in errors">{{ error.description }}</div>
```

gkalpak:
@petebacondarwin, is it ?

From a quick look it seems that `ngMessage` registers itself with its parent `ngMessages` during the post-linking phase, by which time attribute values (such as `ng-message="{{...}}"`) should have been interpolated.

So, using interpolated values should be a problem, nor should watching expressions be necessary to achieve that. (Again, this was a really quick look, so I might have missed something.)

petebacondarwin:
@matsko - can you take a look at the comments here when you get a chance? Did we decide on a naming convention in the end?

Narretz:
Regarding the naming: I have no problem with two attributes, `ng-message` that takes constant input, and another one that takes an expression.
For the second, I don't like `ng-message-match` because it's too unspecific. If the difference is no-expression / expression, the new attribute should be called `ng-switch-exp(ression)`.

matsko:
I'm going with `ng-message` and `ng-message-exp`.

matsko:
The reasons why we're going with two attributes is because:

1) `{{ curly }}` attributes are evaluated at compile. So if a value doesn't exist within the scope at the right time then the message key won't be intercepted. Watching an attribute directly solves this issue.

2) By having watchers from the start then we can safely stick to that API. If we end up supporting `{}` object keys then we can support that directly without having to use `$parse` each time on the interpolated string.

petebacondarwin:
+1 - go for it @matsko

src/ngMessages/messages.js
matsko: I'll ping Igor about it.

src/ngMessages/messages.js
matsko: I'm not sure, but we're doing a similar thing in ngMessagesInclude on master right now.

src/ngMessages/messages.js
matsko: This is outside of the core so I can't use `HashMap`. I did change `nameMatches` to a hash object to make things better though.

petebacondarwin: `nameMatches` appears to be an array; and you use `indexOf` on it below.

src/ngMessages/messages.js
matsko: Good point, but I think for this PR we'll hold off. The main reason is that ngMessageInclude has to keep track of the old message nodes and then destroy them when the watcher fires. I need some time to figure out a good way to make this work. For now it won't be causing a breaking change.

matsko:
Now that we agreed on having two attributes, I think we should also put this feature into 1.3. The only breaking change is `ngMessagesInclude` code which is a one-line fix.

src/ngMessages/messages.js
```diff
@@ -25,8 +33,10 @@
  * <form name="myForm">
  *   <input type="text" ng-model="field" name="myField" required minlength="5" />
  *   <div ng-messages="myForm.myField.$error">
- *     <div ng-message="required">You did not enter a field</div>
- *     <div ng-message="minlength">The value entered is too short</div>
+ *     <div ng-message="require">You did not enter a field</div>
```
petebacondarwin: `required`

src/ngMessages/messages.js
```diff
@@ -52,7 +62,11 @@
  * ngMessages directive to make this happen.
  *
  * ```html
+ * <!-- standard attribute -->
```
petebacondarwin: `<!-- attribute style usage -->`?

src/ngMessages/messages.js
```diff
@@ -52,7 +62,11 @@
...
  * <div ng-messages="myForm.myField.$error" ng-messages-multiple>...</div>
+ *
+ * <!-- element-style attribute -->
```
petebacondarwin: `<!-- element style usage -->`?

src/ngMessages/messages.js
```diff
@@ -102,6 +125,60 @@
...
+ *   <input type="email"
+ *          name="myEmail"
+ *          ng-model="email"
+ *          minlength="5"
+ *          required />
+ *
+ *   <div ng-messages="myForm.myEmail.$error">
+ *     <div ng-message="required">You did not enter your email address</div>
+ *     <div ng-repeat="errorMessage in errorMessages">
+ *       <!-- use ng-message-exp for a message that is an expression -->
```
petebacondarwin: use ng-message-exp for a message whose key is given by an expression

src/ngMessages/messages.js
```diff
@@ -102,6 +125,60 @@
...
+ *       <div ng-message-exp="errorMessage.type">{{ errorMessage.text }}</div>
+ *     </div>
+ *   </div>
+ * </form>
+ * ```
+ *
+ * The `errorMessage.type` variable can be a string value or it can be an array that can have
```
petebacondarwin: variable -> expression

src/ngMessages/messages.js
```diff
@@ -102,6 +125,60 @@
...
+ * multiple error states be associated with a single error message:
```
petebacondarwin: ... or it can be an array so that multiple error states can be associated with a single error message:

petebacondarwin: states -> keys?

src/ngMessages/messages.js
```diff
@@ -102,6 +125,60 @@
...
+ *     <div ng-message-exp="['minlength', 'maxlength]">
+ *       Your email must be between 5 and 100 characters long
+ *     </div>
+ *   </div>
+ * ```
+ *
+ * Feel free to use other directives such as ng-if and ng-switch to further control what messages are active
+ * and when. This, however, may cause some issues with Angular not being able to determine if a directive
+ * is active or not. Therefore if an error does occur it's best to place the ng-message element on its own
+ * element within the structural directive HTML code:
```
petebacondarwin: This paragraph makes sense when you see the example but I think it could be clearer in itself.
Perhaps something like:

> Feel free to use other structural directives such as ng-if and ng-switch to further control what messages are active and when. Be careful, if you place ng-message on the same element as these structural directives, Angular may not be able to determine if a message is active or not. Therefore it is best to place the ng-message on a child element of the structural directive.

src/ngMessages/messages.js
```diff
@@ -228,91 +304,214 @@ angular.module('ngMessages', [])
...
+           var showAll = isAttrTruthy($scope, $attrs.ngMessagesMultiple) ||
+                         isAttrTruthy($scope, $attrs.multiple);
 
-        //JavaScript treats empty strings as false, but ng-message-multiple by itself is an empty string
-        var multiple = angular.isString($attrs.ngMessagesMultiple) ||
-                       angular.isString($attrs.multiple);
+           var matches = [];
+           var nameMatches = {};
+           var item = ctrl.head;
+           var messageExists = false;
```
petebacondarwin: I think this variable name is a little misleading...

messageExists -> messageFound / messagesFound

src/ngMessages/messages.js
```diff
@@ -228,91 +304,214 @@ angular.module('ngMessages', [])
...
-        $scope.$watchCollection(watchAttr, function(values) {
-          cachedValues = values;
-          ctrl.renderMessages(values, multiple);
-        });
+           // all of the messages that were not true are not to be displayed
+           forEach(messages, function(item) {
+             if (matches.indexOf(item) == -1) {
+               item.message.detach();
+             }
+           });
```
petebacondarwin: Isn't there some way we could incorporate this into the previous loops to save on this extra loop with multiple indexOf calls.

src/ngMessages/messages.js
matsko: ...so that multiple **errors** can be associated with a single error message...

src/ngMessages/messages.js
matsko: Looks good. Updated.

src/ngMessages/messages.js
IgorMinar: it should be safe as long as nobody attaches any data to it.

from the code it looks like that's the case, so you should be good to go.

matsko:
Merged as https://github.com/angular/angular.js/commit/c9a4421fc3c97448527eadef1f42eb2f487ec2e0

Tobbe:
I noticed that this is marked with milestone 1.3.14, but no traces of this can be found even in 1.3.15. What is holding this back from being introduced in the 1.3 branch?

I cherry-picked this commit on top of the v1.3.x branch and I had no problems applying it, and from my very quick test everything looks to be working as it should.

Tobbe:
(I also needed to apply this: https://github.com/angular/angular.js/pull/11199 for my usecase to work, but I had no problem with that one either on my 1.3 branch)

Narretz:
We had / have double milestones for 1.4 beta and 1.3 releases because we try to release simultaneously. However, only the next minor release receives features, the current stable release only receives bug fixes.

Tobbe:
So no chance of this going in to 1.3 then I take it. We'll see if I'll maintain my own fork, or if we'll just have to live without this feature...

In any case, thanks for getting back to me :)

PR #10678, opened 2015-01-08
fix(minErr.js): delete unused variable `arg` in `template.replace`

cookfront:


lgalfaso:
Hi, thanks for the PR, but this was fixed with b146cae02c6345d324b1c53845fb01b783226669, so this PR is no longer necessary. Please keep on contributing!

PR #10680, opened 2015-01-08
docs(date): Change milliseconds pagging syntax

mmouterde:
Change millisecond pagging  syntax : no need to prefix with ',' or '.'.


mmouterde:
Contributor License Agreement Signed !

mmouterde:
Thanks @gkalpak.

To complete the reason why `.` and `,` are not required,  look at code comments on line 321-322 :
~~~
    // while ISO 8601 requires fractions to be prefixed with `.` or `,`
    // we can be just safely rely on using `sss` since we currently don't support single or two digit fractions
~~~

PR #10689, opened 2015-01-08
fix(ngRoute): dont duplicate optional params into query

omsmith:
When calling updateParams with properties which were optional, but
previously undefined, they would be duplicated into the query params as
well as into the path.

[jsfiddle reproduction](http://jsfiddle.net/cryqss2z/)

From `#updateParams` docs:

> Provided property names that match the route's path segment definitions will be interpolated into the location's path, while remaining properties will be treated as query params.

In the actual [implementation of updateParams](https://github.com/angular/angular.js/blob/e24d96827628285976526c7e8d91090a90e36322/src/ngRoute/route.js#L487) the "remaining properties" are determined prior to interpolation by checking if the `pathParam` for the given `key` has a value; however, optional parameters may not exist in `pathParams`.

Alternative fixes might include
-  (probably better) running `interpolate` before separating `searchParams`, since it will remove matched keys from the passed in `newParams` object.
-  (maybe more correct?) include undefined optional params in the `pathParams` object (switchRouteMatcher L536, `if (key && val)` -> `if(key)`) and then check for the property as opposed to truthiness

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -481,15 +481,10 @@ function $RouteProvider() {
            */
           updateParams: function(newParams) {
             if (this.current && this.current.$$route) {
-              var searchParams = {}, self=this;
-
-              angular.forEach(Object.keys(newParams), function(key) {
-                if (!self.current.pathParams[key]) searchParams[key] = newParams[key];
-              });
-
               newParams = angular.extend({}, this.current.params, newParams);
               $location.path(interpolate(this.current.$$route.originalPath, newParams));
-              $location.search(angular.extend({}, $location.search(), searchParams));
+              // interpolate modifies newParams, only query params are left
+              $location.search(newParams);
             }
             else {
               throw $routeMinErr('norout', 'Tried updating route when with no current route');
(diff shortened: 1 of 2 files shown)


shahata:
This sounds reasonable to be but I think the implementation can be much more simple:
```js
updateParams: function(newParams) {
  if (this.current && this.current.$$route) {
    newParams = angular.extend({}, this.current.params, newParams);
    $location.path(interpolate(this.current.$$route.originalPath, newParams))
      .search(newParams);
  }
  else {
    throw $routeMinErr('norout', 'Tried updating route when with no current route');
  }
}
```

omsmith:
Corporate CLA just sent over

omsmith:
@shahata Yeah that was my first alternative suggestion - I'll push the change

omsmith:
We sent over details about the new Google Group based CLA authorization on Monday. Haven't heard back, but maybe we're good to go?

petebacondarwin:
@shahata - can you check with the guys in the office that the CLA is good then look at merging this when you are happy with it?

omsmith:
@shahata @petebacondarwin  Alright, actually got the e-mail back now, CLA should be good to go

PR #10697, opened 2015-01-09
fix($q): correctly set constructor on prototypes

bluepnume:
This patch will correctly set `Promise.prototype.constructor` and `Deferred.prototype.constructor`.

When prototype is set in the form `Foo.prototype = {x: y};`, the constructor will not be set correctly (as opposed to in the form `Foo.prototype.x = y;`.

As such, it is necessary to manually add the constructor to the prototype in these cases.

### DIFF
--- a/lib/promises-aplus/promises-aplus-test-adapter.js
+++ b/lib/promises-aplus/promises-aplus-test-adapter.js
@@ -10,6 +10,31 @@ var minErr = function minErr (module, constructor) {
     throw new ErrorConstructor(module + arguments[0] + arguments[1]);
   };
 };
+var setHashKey = function setHashKey(obj, h) {
+  if (h) {
+    obj.$$hashKey = h;
+  }
+  else {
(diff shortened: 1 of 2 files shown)


lgalfaso:
Why not just
>
>
extend(Promise.prototype, {...});

?

bluepnume:
Good suggestion! Done

lgalfaso:
@bluepnume can you look into the Travis failure, it looks like a valid failure

bluepnume:
I reset the previous commit. It looks like `extend` is not available at this point in the execution.

lgalfaso:
@bluepnume look at `lib/promises-aplus/promises-aplus-test-adapter.js`, some copy & paste should be able to handle the rest

lgalfaso:
Hi,
  This needs a little more work before it can land:
* add some tests
* rebase on top of master
* squash everything to a single commit

Thanks!

PR #10705, opened 2015-01-10
feat($http): add X-Forwarded-* headers to requests

alfonso-presa:
.... for compatibility with RESTful backends (i.e. Spring MVC + HATEOAS) that are not beeing reversed proxied, and don't know for sure wich protocol they are beeing used.

I found that this could be an usefull thing to be done automatically by Angular as some frameworks (like Spring HATEOAS) rely in this information to generate absolute URLs (cause they expect a reverse proxy to set this headers for them). In a cloud environment (i.e. PaaS), or local environment, this frameworks are not likely to be reverse proxied (although they could) so currently there is no way other than hacking the framework itself, or creating a custom request interceptor in angular to provide this two headers so everything works .... 

I don't see any drawbacks (other than the little bit increased request size) in having this set automatically by angularjs. 

Also I have to change the 'should allow replacement of the headers object' spec as AFAIK $http could add other headers after the interceptor (as the ones I added or others like XSRF).

Would be glad to hear any feedbacks about the changes I made :-).

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1056,14 +1056,25 @@ function $HttpProvider() {
       }
 
 
-      // if we won't have the response in cache, set the xsrf headers and
+      // if we won't have the response in cache, set the convenient headers and
       // send the request to the backend
       if (isUndefined(cachedResp)) {
         var xsrfValue = urlIsSameOrigin(config.url)
             ? $browser.cookies()[config.xsrfCookieName || defaults.xsrfCookieName]
             : undefined;
+
+           //TODO
         if (xsrfValue) {
           reqHeaders[(config.xsrfHeaderName || defaults.xsrfHeaderName)] = xsrfValue;
+        }
+
+        var target = document.createElement('a');
+        target.href = url;
+
+        var protocol = target.protocol.substring(0, target.protocol.length - 1);
+        reqHeaders['X-Forwarded-Proto'] = protocol;
+        if (protocol == 'https') {
+          reqHeaders['X-Forwarded-Ssl'] = 'on';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@alfonso-presa I'm not HTTP protocol expert by from the quick reading it seems like those headers are meant to be added by proxies, so I'm not sure why AngularJS should be sending those. On top of this your change makes it so those headers are sent on each and every request for applications which we definitively don't want as it could be triggering CORS pre-flight requests in situations when those are not triggered today.

Given all of this I don't think this change belongs to the core, especially that you can easily add those in a dedicated interceptors if you really think that headers like this needs to be sent from an application.

alfonso-presa:
Yes. You could definitely solve this with an interceptor (in fact we did it this way in the project I'm on). I also agree that this doesn't seem to belong to the angular core, but truth is that developers would expect that when you have a framework capable of exposing a RESTful API with affordances (HATEOAS) included and you consume it with $resource in angular, everything in between those two frameworks should be solved by them somehow. The server frameworks rely on those headers to know how to generate absolute urls for HATEOAS. 

What I mean is that we developers expect RESTful APIs to be easy to consume and having to add an interceptor to AngularJS or forcing a reverse proxy to be present doesn't seem like the proper solution for this. I'm not blaming AngularJS on this problem, but I think it's the only place we can solve the problem in a cross way. But you're completely right that I may have solved this by adding changes too deep inside the core of AngularJS, and that could cause side effects. I will rethink the proposal and come back to you if I find a better, less intrusive way to handle this problem.

Thank you very much.

PR #10708, opened 2015-01-11
docs(tutorial/4 - Two-way Data Binding): describe your change...

awesomebjt:
I have Nodejs v0.10.33, AngularJS v1.3.8, and protractor 1.0.0. The end-to-end test as written always failed with the following messages:

```
Expected [  ] to equal [ 'Motorola XOOM™ with Wi-Fi', 'MOTOROLA XOOM™' ].
```

After a great deal of head-scratching I discovered that for whatever reason, element.all(by.repeater('phone in phones').column('phone.name')) was returning this:

```
{"locator_":{},"parentElementFinder_":null}
```

However, if I added classes to the items I was going to be testing for, and built the array of names from those instead, my test would pass.

I think something must be wrong either with Angular or Protractor. It would make more sense for the code that's already in the tutorial to work.

### DIFF
--- a/docs/content/tutorial/step_04.ngdoc
+++ b/docs/content/tutorial/step_04.ngdoc
@@ -32,7 +32,7 @@ __`app/index.html`:__
 
   <ul class="phones">
     <li ng-repeat="phone in phones | filter:query | orderBy:orderProp">
-      <span>{{phone.name}}</span>
+      <span class="name">{{phone.name}}</span>
       <p>{{phone.snippet}}</p>
     </li>
   </ul>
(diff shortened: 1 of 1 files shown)


petebacondarwin:
Can you try upgrading protractor? I am using 1.4.0

praxeo:
Sure, I'll give that a try. Thanks for the response!

petebacondarwin:
@praxeo - did that help?

petebacondarwin:
OK, so the actual problem is that a newer of version of Protractor is being used, which doesn't expect the `{{  }}` curly braces in the call to `column()`.  #10843 - fixes this.

PR #10710, opened 2015-01-11
Grammar change

jDeppen:
Not sure if this is a worthy change but it confused me when I read it.

"In practice, different from is by far the most common of the three, in both British and American English:"
http://www.oxforddictionaries.com/us/words/different-from-than-or-to

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -285,10 +285,10 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
    * * `$rollbackViewValue()` is called.  If we are rolling back the view value to the last
    *   committed value then `$render()` is called to update the input control.
    * * The value referenced by `ng-model` is changed programmatically and both the `$modelValue` and
-   *   the `$viewValue` are different to last time.
+   *   the `$viewValue` are different from last time.
    *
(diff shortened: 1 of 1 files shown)


src/ng/directive/ngModel.js
```diff
@@ -285,10 +285,10 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
    * * `$rollbackViewValue()` is called.  If we are rolling back the view value to the last
    *   committed value then `$render()` is called to update the input control.
    * * The value referenced by `ng-model` is changed programmatically and both the `$modelValue` and
-   *   the `$viewValue` are different to last time.
+   *   the `$viewValue` are different from last time.
    *
    * Since `ng-model` does not do a deep watch, `$render()` is only invoked if the values of
-   * `$modelValue` and `$viewValue` are actually different to their previous value. If `$modelValue`
+   * `$modelValue` and `$viewValue` are actually from their previous value. If `$modelValue`
```
caitp: i think you want to keep the word "different" here

caitp:
other than that it looks ok, could you please update it and sign the CLA so we can merge it? =]

jDeppen:
Done

caitp:
looks ok to me, should do another one o try to fix the run-on sentences though

PR #10711, opened 2015-01-11
fix(input:number): min/max not work when binded variable initial as false value

kltk:
fix(input:number): min/max not work when binded variable initial as false value


gkalpak:
This is a duplicate of #10327 (but #10327 also has tests).

PR #10716, opened 2015-01-11
docs(error/nobase): Added trailing slash for base path

Elijen:
Trailing slash seems to be necessary, otherwise `$routeProvider` does not match routes correctly. Following is not matched:

URL: http://www.example.com/b/foo/1234

``` html
<base href="/b/foo">
```

``` javascript
$routeProvider.when('/:id', {
   templateUrl: '/view/path.html',
   controller: 'MyCtrl',
   reloadOnSearch: false
});
```

### DIFF
--- a/docs/content/error/$location/nobase.ngdoc
+++ b/docs/content/error/$location/nobase.ngdoc
@@ -35,7 +35,7 @@ URL of the subcontext:
 
 ```html
 <head>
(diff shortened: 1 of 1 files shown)


Elijen:
Singed CLA

lgalfaso:
landed as 034fade3e8b6448231b640948995960db75d8605

PR #10717, opened 2015-01-11
Fix memory leak with $setValidity

Narretz:
This PR includes:
- commit that correctly cleans up references to input controls in forms after they have been removed
- commit that renames an argument to better reflect its purpose
- commit which tests the form clean-up behavior for $pending inputs

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -163,6 +163,9 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
     forEach(form.$error, function(value, name) {
       form.$setValidity(name, null, control);
     });
+    forEach(form.$$success, function(value, name) {
+      form.$setValidity(name, null, control);
+    });
 
     arrayRemove(controls, control);
   };
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -1247,22 +1247,22 @@ function addSetValidityMethod(context) {
 
   ctrl.$setValidity = setValidity;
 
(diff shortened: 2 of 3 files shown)


Narretz:
@pkozlowski-opensource you commented on the original issue, can you take a look at this?

lgalfaso:
overall, it looks good. Can you squash this into one commit?

Narretz:
I don't know ... it's three different things, I'd like to keep them separate.

lgalfaso:
if it these are 3 different things, then they should be 3 different PRs...

Narretz:
Why? They all affect the same area (I noticed the two other things while working on the fix). They are also not huge changes that would make reviewing difficult - which is one reason for having separate PRs. I do not want to squash them, because then you'll have on commit that says "fix..." but does some additional (though related) things, which makes it more difficult for someone to  comprehend the changes later on.

petebacondarwin:
I agree with @Narretz that this is three commits but I think I agree with @lgalfaso that these should normally be in three PRs since they are basically unrelated - even though they are small. But apart from coming up with an even better name for `control` it looks good enough to merge all three.

Narretz:
What about `controller`. That's actually what it is.
I'll also cherry-pick this to 1.3.x

Narretz:
Landed as 2408f2ded5ead6e678c241e38ef474c1fadff92b and cdc7280dd3d5a2ded784c06dd55fe36c2053fb6f (1.3)

Next time I'll split the commits into PRs

PR #10719, opened 2015-01-12
docs(tutorial/6 - Templating Links): insert comma

kchia:
The file change inserts a missing comma into the document.


btford:
Thanks for the fix, but I think this reads fine both with and without the comma. I'm going to close this, but thank you for taking the time to submit it!

PR #10720, opened 2015-01-12
Docs (index.ngdoc) change: A potential compatibility issue with Firefox OS

acegautam:
An update on Firefox OS ver 2.0 seems to cause an issue with dependency injection using Angular JS. Modified the docs to include a link related to Firefox OS.

### DIFF
--- a/docs/content/guide/index.ngdoc
+++ b/docs/content/guide/index.ngdoc
@@ -59,7 +59,7 @@ In Angular applications, you move the job of filling page templates with data fr
 * **Realtime: **[Socket.io](http://www.creativebloq.com/javascript/angularjs-collaboration-board-socketio-2132885), [OmniBinder](https://github.com/jeffbcross/omnibinder)
 * **Visualization:** [SVG](http://gaslight.co/blog/angular-backed-svgs), [D3.js](http://www.ng-newsletter.com/posts/d3-on-angular.html)
 * **Local Storage and session:** [ngStorage](https://github.com/gsklee/ngStorage)
-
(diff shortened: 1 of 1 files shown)


acegautam:
I just sent out a request for CLA. Got a confirmation, the request has been submitted.

btford:
I'm not entirely sure that your blog is correct; AFAIK, there's no reason that a browser would need Angular's strict mode.

caitp:
it's possible that Function.prototype.toString does the wrong thing in FFOS, so it's not totally unreasonable. I don't care to verify it though

acegautam:
That's correct - A browser wouldn't necessarily need a strict DI. However,
we have experienced while working on FFOS applications, that on the Flame
2.0 devices, the application stopped working after a possible update to the
OS (OTA). It was working on the factory 2.0 version. Also some end users on
Firefox Marketplace reported the same issue with 2.x OS versions.

The error that we were seeing was - "$scope" is undefined. And I found that
the dependency was implicitly being injected into the controller (only as a
function argument). The moment I did an explicit annotation using the array
syntax, issue was mitigated and app started working fine.

I am not sure of the exact reason what was the change in the OS update that
cause this, but I just thought developers shouldn't waste efforts and time
figuring out this issue, which is not so obvious. And Angular being one of
the most loved MVP frameworks, there are more probabilities of devs using
it for FFOS app development. Hence was trying to contribute to the cause.
That's okay if you don't feel this is accurate.

Narretz:
As this bug is pretty specific (andmight have been fixed in the meantime), I don't think it adds a lot of value to the guide.

PR #10723, opened 2015-01-12
fix(ngScenario): Allow ngScenario to handle lazy-loaded and manually bootstrapped applications

marcfallows:
I know protractor is preferred, and ngScenario is only in maintenance mode. But, we are limited to
ngScenario based on the devices/browsers we are targeting (no web-driver available). So, we need
to address the bug where ngScenario does not work with manual bootstrap and also has issues if
angular.resumeBootstrap is not yet defined (race condition when lazy-loading).

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1408,8 +1408,12 @@ function bootstrap(element, modules, config) {
     forEach(extraModules, function(module) {
       modules.push(module);
     });
-    doBootstrap();
+    return doBootstrap();
   };
+
+  if(isFunction(angular.resumeDeferredBootstrap)){
+    angular.resumeDeferredBootstrap();
+  }
 }
 
 /**
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1411,7 +1411,7 @@ function bootstrap(element, modules, config) {
     return doBootstrap();
   };
 
-  if(isFunction(angular.resumeDeferredBootstrap)){
+  if (isFunction(angular.resumeDeferredBootstrap)) {
     angular.resumeDeferredBootstrap();
   }
 }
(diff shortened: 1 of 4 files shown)


marcfallows:
Signed CLA

caitp:
ngScenario is deprecated, I'm not sure anyone maintaining 1.x can properly review this

marcfallows:
@caitp Even if deprecated, I don't think ngScenario should exist in angular with bugs. At the moment ngScenario does not support apps which manual bootstrap or are lazy-loaded (Googling "ngscenario manual bootstrap" highlights that this is causing problems for more than just me).

You'll notice that no tests were changed or removed, only added. The existing functionality remains unchanged and I only fix the broken cases.

If well covered bug fixes won't be considered it should really be removed from the project.

petebacondarwin:
Given that we are in maintenance mode, we should probably just land this - but we should remove ngScenario (perhaps to its own project where it can be taken on by others) before 1.5.0

@marcfallows - can you please run `grunt test` and fix the jshint and jscs errors in this?

marcfallows:
Thanks @petebacondarwin. I've updated the jshint and jscs errors in my changes, sorry I didn't catch that on the first go-around.

It would certainly be useful for us that if ngScenario were removed it was still available (as it's own project, per your suggestion). Unfortunately we are currently tied to using it, but there's not really any reason it needs to be part of angular's core if the appropriate hooks are available in angular for it to sit a separate project.

petebacondarwin:
@marcfallows - I have merged this. Would you be interested in taking more of an active role in supporting this module if we separated it from the angular.js repository?

marcfallows:
Nice! @petebacondarwin I would definitely like to be involved if it became a standalone project, and assist in decoupling from the angular project to support that move.

petebacondarwin:
@marcfallows - I have created a repository https://github.com/angular/ngScenario and given you write access. Are there any other people that should be invited?
I have created a bunch of issues to help move things forward. Feel free to take on as many as you are able...

PR #10731, opened 2015-01-12
chore(npm): Make require()-able as part of publish script

bclinkinbeard:
Deletes main field of package.json and creates index.js that is CommonJS compatible

/cc @caitp

### DIFF
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -75,6 +75,18 @@ function prepare {
     replaceJsonProp "bower.json" "angular.*" ".*" "$NEW_VERSION"
     replaceJsonProp "package.json" "version" ".*" "$NEW_VERSION"
     replaceJsonProp "package.json" "angular.*" ".*" "$NEW_VERSION"
+    deleteJsonProp "package.json" "main"
+
+    echo "-- Adding CommonJS index file"
+    if [ -f "index.js" ]
+    then
+      rm index.js
+    fi
+
+    touch index.js
+    echo "require('./$repo');" >> index.js
+    echo "" >> index.js
+    echo "module.exports = $repo;" >> index.js
 
     git add -A
 
--- a/scripts/utils.inc
+++ b/scripts/utils.inc
@@ -207,6 +207,12 @@ function replaceJsonProp {
   replaceInFile $1 '"('$2')"[ ]*:[ ]*"'$3'"' '"\1": "'$4'"'
 }
 
+# deleteJsonProp(jsonFile, property)
+# - restriction: property needs to be on a single line!
+function deleteJsonProp {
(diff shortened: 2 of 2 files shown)


caitp:
@petebacondarwin I'd like to get this into the release today. Just going to test it real quick. That cool with you?

caitp:
so, on my system, the `sed` call on package.json is getting `-e` passed as the value to `-i`, which is weird. I'm not sure if that's happening for you or not, but we might need to add a safety check just to prevent that from happening

bclinkinbeard:
Not sure I follow. I'm not super familiar with sed tbh, but the deleteJsonProp function is using `sed -i -e`, which could very well be wrong. Is that the call you are referring to?

caitp:
yes, I thought you added it, didn't check the diff =) @IgorMinar do you know if we were using that utility before and it worked right on jenkins?

caitp:
wait, it is added in the diff :< hmmm

bclinkinbeard:
Oh, I should probably just remove -e, no?

bclinkinbeard:
Just pushed another commit combining the flags. I think that is right...

caitp:
no, you definitely don't want to combine the flags... here's an idea, just leave it as it was before and delete the package.json-e if it exists, that probably is fine

bclinkinbeard:
> delete the package.json-e if it exists

Sorry, I don't know what you mean by that.

caitp:
.. if the file exists (`package.json-e`), remove it

caitp:
well, I guess this won't make it for 1.3.9 :( sorry man.

caitp:
...or maybe it will, we'll see :>

bclinkinbeard:
I think this should be correct. Apparently some versions of sed require a suffix be passed to `-i`, so I've set it to use a blank string.

Fingers crossed! :)

petebacondarwin:
sed is well outside of my comfort zone. If you two are happy with that then so am I.

caitp:
just tested with the latest commit, looks good here

caitp:
lgtm~

caitp:
eh pete, should we backport this? I guess future 1.3 releases would want it too

bclinkinbeard:
Sweet!

petebacondarwin:
@caitp - let's put it in 1.4 and 1.3 but not 1.2 OK?
Feel free to merge then we have our release SHAs.

caitp:
it is done~

spacepluk:
thank you guys :)

petebacondarwin:
I understand that there might be a problem with this?

bclinkinbeard:
@petebacondarwin Yea we realized some shortcomings. https://github.com/angular/angular.js/pull/10732 now has the correct, improved version.

PR #10732, opened 2015-01-13
chore(npm): Make modules Browserify compatible

bclinkinbeard:
Better version of earlier PR.

index.js for angular:

``` js
require('./angular');

module.exports = angular;
```

index.js for other modules:

``` js
require('./angular-animate');

module.exports = angular.module('ngAnimate');
```

Still working on adding a dependency to angular in package.json of other modules.

### DIFF
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -78,13 +78,7 @@ function prepare {
     deleteJsonProp "package.json" "main"
 
     echo "-- Adding CommonJS index file"
-    if [ -f "index.js" ]
-    then
-      rm index.js
-    fi
-
-    touch index.js
-    echo "require('./$repo');" >> index.js
+    echo "require('./$repo');" > index.js
     echo "" >> index.js
     echo "module.exports = $repo;" >> index.js
     if [ $repo == "angular" ]
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -75,6 +75,28 @@ function prepare {
     replaceJsonProp "bower.json" "angular.*" ".*" "$NEW_VERSION"
     replaceJsonProp "package.json" "version" ".*" "$NEW_VERSION"
     replaceJsonProp "package.json" "angular.*" ".*" "$NEW_VERSION"
+    deleteJsonProp "package.json" "main"
+
+    echo "-- Adding CommonJS index file"
+    if [ -f "index.js" ]
+    then
+      rm index.js
+    fi
+
+    touch index.js
+    echo "require('./$repo');" >> index.js
+    echo "" >> index.js
+    echo "module.exports = $repo;" >> index.js
+    if [ $repo == "angular" ]
+    then
+      echo "module.exports = angular;"
+    else
+      suffix=`echo $repo | cut -c9-`
+      first=`echo $suffix | cut -c1 | sed -e 'y/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/'`
+      tail=`echo $suffix | cut -c2-`
+
+      echo "module.exports = angular.module('ng$first$tail');"
+    fi
 
     git add -A
 
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -80,7 +80,6 @@ function prepare {
     echo "-- Adding CommonJS index file"
     echo "require('./$repo');" > index.js
     echo "" >> index.js
-    echo "module.exports = $repo;" >> index.js
     if [ $repo == "angular" ]
     then
       echo "module.exports = angular;" >> index.js
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -91,6 +91,14 @@ function prepare {
       tail=`echo $suffix | cut -c2-`
 
       echo "module.exports = 'ng$first$tail';" >> index.js
+
+      # add angular as a peer dependency
+      deleteJsonProp "package.json" "peerDependencies"
+      replaceInFile "package.json" "homepage\"\: \"http\:\/\/angularjs\.org\"$" "homepage\"\: \"http\:\/\/angularjs\.org\","
+      sed -i '' -e /^}/d "package.json"
+      # have to use single line form so deleteJsonProp will work
+      echo "\t\"peerDependencies\": { \"angular\": \"$NEW_VERSION\" }" >> package.json
+      echo '}' >> package.json
     fi
 
     git add -A
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -89,14 +89,14 @@ function prepare {
     echo "module.exports = $repo;" >> index.js
     if [ $repo == "angular" ]
     then
-      echo "module.exports = angular;"
+      echo "module.exports = angular;" >> index.js
     else
       # convert to module names (angular-animate >> ngAnimate)
       suffix=`echo $repo | cut -c9-`
       first=`echo $suffix | cut -c1 | sed -e 'y/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/'`
       tail=`echo $suffix | cut -c2-`
 
-      echo "module.exports = 'ng$first$tail';"
+      echo "module.exports = 'ng$first$tail';" >> index.js
     fi
 
     git add -A
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -100,14 +100,14 @@ function prepare {
 
       if [ $repo != "angular-mocks" ]
       then
-        echo "module.exports = angular.module('ng$first$tail');" >> index.js
+        echo "module.exports = 'ng$first$tail';" >> index.js
       else
         for mock in "${MOCK_MODULES[@]}"
         do
           echo "-- Adding $repo/$mock CommonJS file"
           echo "require('./$repo');" > $mock.js
           echo "" >> $mock.js
-          echo "module.exports = angular.module('$mock');" >> $mock.js
+          echo "module.exports = '$mock';" >> $mock.js
         done
       fi
     fi
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -90,7 +90,14 @@ function prepare {
       first=`echo $suffix | cut -c1 | sed -e 'y/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/'`
       tail=`echo $suffix | cut -c2-`
 
-      echo "module.exports = 'ng$first$tail';" >> index.js
+      if [ $repo != "angular-mocks" ]
+      then
+        echo "module.exports = 'ng$first$tail';" >> index.js
+      else
+        echo "exports.ngMock = 'ngMock';" >> index.js
+        echo "exports.ngMockE2E = 'ngMockE2E';" >> index.js
+        echo "exports.ngAnimateMock = 'ngAnimateMock';" >> index.js
+      fi
 
       # add angular as a peer dependency
       deleteJsonProp "package.json" "peerDependencies"
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -91,11 +91,12 @@ function prepare {
     then
       echo "module.exports = angular;"
     else
+      # convert to module names (angular-animate >> ngAnimate)
       suffix=`echo $repo | cut -c9-`
       first=`echo $suffix | cut -c1 | sed -e 'y/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/'`
       tail=`echo $suffix | cut -c2-`
 
-      echo "module.exports = angular.module('ng$first$tail');"
+      echo "module.exports = 'ng$first$tail';"
     fi
 
     git add -A
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -103,7 +103,7 @@ function prepare {
       replaceInFile "package.json" "homepage\"\: \"http\:\/\/angularjs\.org\"$" "homepage\"\: \"http\:\/\/angularjs\.org\","
       sed -i '' -e /^}/d "package.json"
       # have to use single line form so deleteJsonProp will work
-      echo "\t\"peerDependencies\": { \"angular\": \"$NEW_VERSION\" }" >> package.json
+      echo "  \"peerDependencies\": { \"angular\": \"$NEW_VERSION\" }" >> package.json
       echo '}' >> package.json
     fi
 
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -97,14 +97,6 @@ function prepare {
         echo "exports.ngMockE2E = 'ngMockE2E';" >> index.js
         echo "exports.ngAnimateMock = 'ngAnimateMock';" >> index.js
       fi
-
-      # add angular as a peer dependency
-      deleteJsonProp "package.json" "peerDependencies"
-      replaceInFile "package.json" "homepage\"\: \"http\:\/\/angularjs\.org\"$" "homepage\"\: \"http\:\/\/angularjs\.org\","
-      sed -i '' -e /^}/d "package.json"
-      # have to use single line form so deleteJsonProp will work
-      echo "  \"peerDependencies\": { \"angular\": \"$NEW_VERSION\" }" >> package.json
-      echo '}' >> package.json
     fi
 
     git add -A
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -29,6 +29,11 @@ function init {
     angular-touch
     angular-messages
   )
+  MOCK_MODULES=(
+    ngMock
+    ngMockE2E
+    ngAnimateMock
+  )
 }
 
 
@@ -77,9 +82,13 @@ function prepare {
     replaceJsonProp "package.json" "angular.*" ".*" "$NEW_VERSION"
     deleteJsonProp "package.json" "main"
 
-    echo "-- Adding CommonJS index file"
-    echo "require('./$repo');" > index.js
-    echo "" >> index.js
+    if [ $repo != "angular-mocks" ]
+    then
+      echo "-- Adding $repo CommonJS index file"
+      echo "require('./$repo');" > index.js
+      echo "" >> index.js
+    fi
+    
     if [ $repo == "angular" ]
     then
       echo "module.exports = angular;" >> index.js
@@ -93,9 +102,13 @@ function prepare {
       then
         echo "module.exports = angular.module('ng$first$tail');" >> index.js
       else
-        echo "exports.ngMock = 'ngMock';" >> index.js
-        echo "exports.ngMockE2E = 'ngMockE2E';" >> index.js
-        echo "exports.ngAnimateMock = 'ngAnimateMock';" >> index.js
+        for mock in "${MOCK_MODULES[@]}"
+        do
+          echo "-- Adding $repo/$mock CommonJS file"
+          echo "require('./$repo');" > $mock.js
+          echo "" >> $mock.js
+          echo "module.exports = angular.module('$mock');" >> $mock.js
+        done
       fi
     fi
 
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -91,7 +91,7 @@ function prepare {
 
       if [ $repo != "angular-mocks" ]
       then
-        echo "module.exports = 'ng$first$tail';" >> index.js
+        echo "module.exports = angular.module('ng$first$tail');" >> index.js
       else
         echo "exports.ngMock = 'ngMock';" >> index.js
         echo "exports.ngMockE2E = 'ngMockE2E';" >> index.js
--- a/scripts/utils.inc
+++ b/scripts/utils.inc
@@ -207,6 +207,12 @@ function replaceJsonProp {
   replaceInFile $1 '"('$2')"[ ]*:[ ]*"'$3'"' '"\1": "'$4'"'
 }
 
+# deleteJsonProp(jsonFile, property)
+# - restriction: property needs to be on a single line!
+function deleteJsonProp {
+  sed -i '' -e /\"$2\"\:/d $1
+}
+
 # replaceInFile(file, findPattern, replacePattern)
 function replaceInFile {
   sed -i .tmp -E "s/$2/$3/" $1
--- a/scripts/utils.inc
+++ b/scripts/utils.inc
@@ -46,6 +46,7 @@
 # Utility functions:
 # - readJsonProp
 # - replaceJsonProp
+# - deleteJsonProp
 # - resolveDir
 # - getVar
 # - serVar


caitp:
i don't really see the benefit if exporting the module object. seems like it would just encourage people to add things to existing modules :\

scripts/bower/publish.sh
```diff
@@ -75,6 +75,28 @@ function prepare {
...
+    echo "require('./$repo');" >> index.js
+    echo "" >> index.js
+    echo "module.exports = $repo;" >> index.js
+    if [ $repo == "angular" ]
+    then
+      echo "module.exports = angular;"
+    else
+      suffix=`echo $repo | cut -c9-`
+      first=`echo $suffix | cut -c1 | sed -e 'y/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/'`
+      tail=`echo $suffix | cut -c2-`
```
caitp: make sure to comment these lines clearly to indicate that you're transforming the repo name into a module name, alright?

bclinkinbeard:
No, it allows you to pull them in the way you'd expect in a CommonJS workflow. When I write my own modules, I usually create my app structure like this:

```js
angular.module('app', [
   require('./someModule').name,
   require('./otherModule').name
]);
```

Copying @btford as he recommended the format.

scripts/bower/publish.sh
bclinkinbeard: Sure, good idea.

caitp:
you could just export the name of the module instead, I don't think this is that great

bclinkinbeard:
The module is what the file is actually creating though, so in idiomatic CommonJS a developer would expect the module to be returned when requiring it.

caitp:
I don't really see it this way.

I think a developer would be more likely to expect the `angular` object to be exported from the angular module, not the module itself. You can do things with the module, but you should not do anything with the module, it doesn't really serve any purpose to you as a developer. The name of the module is all that really matters there

bclinkinbeard:
Fair enough, updated.

scripts/bower/publish.sh
```diff
@@ -75,6 +75,29 @@ function prepare {
...
+      rm index.js
+    fi
+
+    touch index.js
+    echo "require('./$repo');" >> index.js
+    echo "" >> index.js
+    echo "module.exports = $repo;" >> index.js
+    if [ $repo == "angular" ]
+    then
+      echo "module.exports = angular;"
```
caitp: don't you need to pipe these echos to index.js with `>>`?

bclinkinbeard:
Thinking about the dependency more, `angular` would actually be a `peerDependency` of the other modules. That said, peer dependencies are notoriously error prone, and may even be deprecated at some point in the future.

I think we should call this PR done, and people will just need to realize they can't use things like `ngAnimate` without separately having `angular`. That seems pretty reasonable to me.

scripts/bower/publish.sh
bclinkinbeard: Good catch, thanks! Fixed.

scripts/bower/publish.sh
```diff
@@ -75,6 +75,29 @@ function prepare {
...
+    deleteJsonProp "package.json" "main"
+
+    echo "-- Adding CommonJS index file"
+    if [ -f "index.js" ]
+    then
+      rm index.js
+    fi
+
+    touch index.js
+    echo "require('./$repo');" >> index.js
```
caitp: instead of removing index.js, touching it and then appending to it, you could just do this:

`echo "require('./$repo');" > index.js` --- that would shave off a few lines of code and be a bit easier to read

caitp:
>Thinking about the dependency more, angular would actually be a peerDependency of the other modules. That said, peer dependencies are notoriously error prone, and may even be deprecated at some point in the future.

In practice, not so much. Angular core modules aren't very good at mixing/matching versions, it usually causes bugs. So depending on the same version is fine

scripts/bower/publish.sh
bclinkinbeard: Very nice, done.

bclinkinbeard:
So you think we do need to add a peerDependency to the modules?

scripts/bower/publish.sh
```diff
@@ -75,6 +75,23 @@ function prepare {
...
+    if [ $repo == "angular" ]
+    then
+      echo "module.exports = angular;" >> index.js
+    else
+      # convert to module names (angular-animate >> ngAnimate)
+      suffix=`echo $repo | cut -c9-`
+      first=`echo $suffix | cut -c1 | sed -e 'y/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/'`
+      tail=`echo $suffix | cut -c2-`
+
+      echo "module.exports = 'ng$first$tail';" >> index.js
```
caitp: I'll give this a few other comments in the morning, but I guess ngMock is tricky, because it exports 3 different modules, and the module names don't end with an `s` like the repo name. Worry about that later though. Good work so far!

bclinkinbeard: Not making it in for 1.3.9, eh? :(

Thanks for all your help!

bclinkinbeard:
OK, peerDependencies added to modules, multiple exports for ngMocks, and a bit of clean up. I've tested it a few times now by running the prepare action, and all the generated code looks correct to me.

I am seeing a problem trying to use the modules though. Given the following code, I get the error `Uncaught Error: [$injector:unpr] Unknown provider: ngRouteProvider <- ngRoute`. 

Any ideas?

```js
var angular = require('angular');
var mocks = require('angular-mocks');

console.log(angular);

angular.module('app', [
  require('angular-route'),
  mocks.ngMock,
  mocks.ngMockE2E,
  mocks.ngAnimateMock
])
  .run(function (ngRoute, ngMock, ngMockE2E, ngAnimateMock) {
    console.log(ngRoute);
    console.log(ngMock);
    console.log(ngMockE2E);
    console.log(ngAnimateMock);
  });
```

bclinkinbeard:
OK, duh, in my rush to prove this out I was using the wrong injection names. This code works as expected!

```js
var angular = require('angular');
var mocks = require('angular-mocks');

console.log(angular);

angular.module('app', [
  require('angular-route'),
  //mocks.ngMock,
  mocks.ngMockE2E,
  mocks.ngAnimateMock
])
  .run(function ($route, $httpBackend, $animate) {
    console.log($route);
    console.log($httpBackend);
    console.log($animate);
  });
```

Let me know if you'd like me to share my test app.

caitp:
> Let me know if you'd like me to share my test app.

Please do =)

bclinkinbeard:
[Here you go.](https://dl.dropboxusercontent.com/u/423061/ng-ify.zip)

It has angular, angular-animate, angular-mocks, and angular-route in the node_modules folder but you can add any of the modules from the tmp folder after running the prepare action from publish.sh.

`npm run bundle` will generate the bundle, then start a server and open `index.html`.

scripts/bower/publish.sh
```diff
@@ -75,6 +75,37 @@ function prepare {
...
+        echo "exports.ngMockE2E = 'ngMockE2E';" >> index.js
+        echo "exports.ngAnimateMock = 'ngAnimateMock';" >> index.js
+      fi
+
+      # add angular as a peer dependency
+      deleteJsonProp "package.json" "peerDependencies"
+      replaceInFile "package.json" "homepage\"\: \"http\:\/\/angularjs\.org\"$" "homepage\"\: \"http\:\/\/angularjs\.org\","
+      sed -i '' -e /^}/d "package.json"
+      # have to use single line form so deleteJsonProp will work
+      echo "  \"peerDependencies\": { \"angular\": \"$NEW_VERSION\" }" >> package.json
```
btford: I highly recommend against using peerDependencies – https://github.com/npm/npm/issues/5080#issuecomment-40545599

bclinkinbeard: Yea, I mention their problems and uncertain future, but @caitp pointed out that since these would be locked to a specific version we shouldn't have any issues. I'm happy to remove it, but that means devs will need to explicitly install angular itself, along with the other modules they need. (Which seems fine, just pointing out.)

caitp: well, no, I wrote a comment originally but removed it because of issues with npm that kind of get in the way of it =)

Basically, you need all the angular modules to be for the same version --- buuuuut, because npm does its own thing, every module can have its own versions of the same dependencies... so it doesn't really solve anything :(

peer deps don't really help either =\ it's just one of those things ._.

bclinkinbeard: Using `peerDependencies` like this _would_ actually help, because if one were to `npm i angular-route`, you would get both `angular` and `angular-route` at the top level of your `node_modules` directory.

That being said, say the word and I will delete that line. :)

caitp:
@bclinkinbeard so, I think we're definitely going to want to get some feedback from the community on what gets exported, and the sort of conventions that go along with it.

So, we're sort of trying out a new thing to get feedback from the (huge) community of angular users, could I get you to try and follow [this document](https://docs.google.com/document/d/16UhnPdpYG4tvcFHKYCI9hMSCWwD7F0x3wzeCe4YHLKI/edit#heading=h.kxdf5zr0fyr3) by sending an intent-to-implement message to the mailing list?

The goal of this would be:

1. inform people that CJS support is planned
2. figure out what CJS support should look like (what should be exported, etc)
3. get approval from owners to ship CJS support

Sound good?

bclinkinbeard:
Sure, I will do that this evening.

btford:
Thanks for your tireless work, Ben!

bclinkinbeard:
https://groups.google.com/forum/#!topic/angular/i5by29Okh48

I wasn't sure if I should link to this PR, so I didn't. The template mentioned tweeting out a link, but I wasn't sure if that meant me, or the team, or what. Anywho, let's get this landed! :)

bclinkinbeard:
I decided to remove the peerDependencies stuff. I conferred with an npm employee and heard again that they may very well be deprecated down the road. What finally convinced me though, was realizing that even with peerDependencies specified, the developer still needs to know to `require('angular')`, as all peerDependencies does is install it for them.

I don't think omitting this will cause any real issues.

caitp:
> I wasn't sure if I should link to this PR, so I didn't. The template mentioned tweeting out a link, but I wasn't sure if that meant me, or the team, or what. Anywho, let's get this landed! :)

I've asked on the team chat for someone to tweet it from the angular account, but it's of course better if you tweet it yourself and we retweet it =)

bclinkinbeard:
https://twitter.com/bclinkinbeard/status/555201780007837697

:)

bclinkinbeard:
I updated the script to return the module for everything but `angular` and `angular-mocks`. Will update `angular-mocks` pending more feedback.

filipedeschamps:
Hi @bclinkinbeard !

This PR is related to the fact I can't require `ng-animate` like this?

Require

ngAnimate Error

It's working for everything else, but Angular's submodules. Is there any workaround better then:
```js
require('angular-animate/angular-animate.js');
```

Best regards.

caitp:
@filipedeschamps the brokenness right now was reverted, but unfortunately the revert didn't make it into the release --- so no, there's no real workaround other than relying on 1.3.8 or doing what you're doing

filipedeschamps:
@caitp ah great! Downgraded to 1.3.8 and worked perfectly. Thank you for pointing it out.

kentcdodds:
Originally commented on #10785, but moving this over here as it seems more relevant.

For projects using UMD, some of them look like this:

```js
(function (root, factory) {
  if (typeof define === 'function' && define.amd) {
    // AMD. Register as an anonymous module.
    define(['angular'], function (angular) {
      return (root.returnExportsGlobal = factory(angular));
    });
  } else if (typeof exports === 'object') {
    // Node. Does not work with strict CommonJS, but
    // only CommonJS-like enviroments that support module.exports,
    // like Node.
    module.exports = factory(require('angular'));
  } else {
    factory(root.angular);
  }
}(this, function (angular) {
  // create angular module here and do stuff
});
```

I just updated to 1.3.10 and it broke one of my dependencies because `require('angular')` no longer exports `angular`. I am in favor of `require('angular')` exporting the `angular` variable. [I personally prefer](https://medium.com/@kentcdodds/how-to-distribute-your-angularjs-module-e04d4dd58ddc) that modules export the name only as it indicates to users that they shouldn't be messing with the module and should only depend on the name. But I'm fine either way. I would just prefer that the community pick one and everyone stick with it.

petebacondarwin:
@caitp - looks like there might still be an issue with 1.3.10 and npm. Can we get all this ironed out before Sunday so that it can go in 1.3.11?

caitp:
the code on the bower repository tagged for v1.3.10 appears to be correct, and if it's not correct, 1.4 is broken too

bclinkinbeard:
OK, I've now enabled consistent use of the `angular-mocks` module as well. You can check out [this project](https://dl.dropboxusercontent.com/u/423061/ng-cjs.zip) to play with it. I'm including the code of the main file below for convenience.

@caitp @btford @IgorMinar @petebacondarwin it would be **awesome** if we could get consensus on this for the next release.

```js
var angular = require('angular');

angular.module('app', [
  require('angular-animate').name,
  require('angular-aria').name,
  require('angular-cookies').name,
  require('angular-messages').name,
  require('angular-mocks/ngMock').name,
  require('angular-mocks/ngAnimateMock').name,
  require('angular-resource').name,
  require('angular-route').name,
  require('angular-sanitize').name,
  require('angular-touch').name
])
  .run(function ($animate, $aria, $cookies, $resource, $route, $sanitize) {
  console.log($animate);
  console.log($aria);
  console.log($cookies);
  console.log($resource);
  console.log($route);
  console.log($sanitize);
});
```

caitp:
I'm still really skeptical about the "export the whole module" thing, I want someone else to sign off on that before landing it. Admittedly yes, you can still shoot yourself in the foot without that, but it's weird

bclinkinbeard:
I think it's kind of unnecessary too, but it seemed to be the prevailing preference on the mailing list since it would require the least amount of changes for people currently shimming the modules.

kentcdodds:
@bclinkinbeard, I don't know how much work it would take for people shimming modules. I haven't done that before. But my knee jerk reaction is that we should prefer the "right" way over the "easier" way. I personally feel like the "right" way is to encourage the use of these modules as angular dependencies (as in, the string array of module names) and make it more difficult to shoot yourself in the foot by manipulating those modules.

petebacondarwin:
I think we should return the module but provide a mechanism for "locking" a module, which we use on Angular modules.  I think @lgalfaso  is looking at that

caitp:
the locking suggestion was from me, and it's trivial to implement. but it's quite likely a very breaking change, which means we'd not be able to put it in 1.3 too

kentcdodds:
It would be a breaking change to actually lock the module, but it wouldn't be a breaking change for the feature to be implemented. If implemented, other library authors could add it and it would be a breaking change for them, but that's their prerogative. I would love to see this feature implemented.

bclinkinbeard:
What's the point of returning a module if it's locked?

caitp:
whats the point in returning a module at all?

kentcdodds:
:smile_cat:

bclinkinbeard:
The only thing I can think of is for monkey patching. I'm happy to switch it back to returning the string, I was just trying to accommodate the feedback you requested we gather. :)

caitp:
i don't have seriously strong opinions on it, it seems clunky either way

kentcdodds:
Hmmm... So my library has a use case for exporting more than just a string... Just thought I'd mention this because I think that the community should be consistent with how angular modules are created and consumed in this context. There is pre-config stuff that people need to set up prior to the bootstrap of the app. So I'm going to be exporting an object that has the properties I need along with `ngModuleName`. So I'm still not exporting the entire module, but I'm not just exporting a string.

So a consumer would use it like so:

```javascript
angular.module('theirModule', [
  require('myModule').ngModuleName
]);
```

I'm in favor of community consistency, so I would really like feedback on this idea so my library follows the standard set by the angular team's module. But I think that it's valuable to export more than just a string... And I think it's valuable to not just export the angular module because I don't want to monkey-punch stuff on my module for fear of future angular changes breaking my module.

bendrucker:
I've released a lot of open source Angular modules written with CommonJS (my quick count says 14). I write the internals of my applications the same way. And I export the module names only. From an appearance standpoint, I'd prefer exporting strings. If the convention goes the other way I have more than a dozen modules to go update code and docs. 

I also try to be good about doing UMD builds for people still using Bower/concatenating their scripts. I wrote [publicist](https://github.com/bendrucker/publicist) to handle that process gracefully. 

Here's an example of how I typically document the npm vs. bower setup process:

https://github.com/bendrucker/angular-stripe#setup

I'm also a maintainer over at [thlorenz/browserify-shim](https://github.com/thlorenz/browserify-shim) which is basically a required build tool for anyone using Angular + CommonJS. The example for Angular use is a bit out of date, but it's telling people to configure the shim to export a string:

https://github.com/thlorenz/browserify-shim/wiki/browserify-shim-recipes#angular-which-depends-on-jquery--angular-modules-installation-via-npm-and-napa

We're working with very incomplete data here, based on everything I've seen, there's way more cases out there where people are exporting strings. I don't think that the existing convention is arbitrary. The need to get at the module object directly is pretty rare. It's super verbose when you do need it, but I think that's far outweighed by the concision of being able to `require` the name in your module array for the vast majority of the cases when that's all you want.

mtr:
I am in favor of ```require('angular')``` exporting the ```angular``` variable, so that the following would work again:
```javascript
var angular = require('angular');

angular.module('foo', …);
```
without any need for an additional browserify-shim specification for the Angular package.

bendrucker:
Wanted to add a note specifically on the idea of holding off on this since I think I said enough above about why I want exported strings. 

The need to shim Angular negates a lot of the benefits of using npm, including `npm dedupe` and not being couple to the internal structure of your packages.

Browserify transforms don't run on any path that's not absolute and doesn't begin with a dot. This means stuff from `node_modules. That means that in order to shim Angular with browserify-shim, you need to add an entry in `browser` in the package to map `require('angular')` to a relative path:

https://github.com/thlorenz/browserify-shim/commit/7c63b7794ceaf35e3a6407e896671647beb05098

At that point npm is just a downloader with a registry and not much of a package manager anymore.

spacepluk:
@bendrucker I think you can use global transforms in browserify if you want them to affect the stuff in `node_modules`.  It seems to be working for me if you shim it like this:

````json
"browserify-shim": {
    "./node_modules/angular/angular.js": {
      "exports": "angular"
    }
  }
````

I also prefer the angular modules to export strings except for `angular`, so I can require them straight into the dependencies array.  The other style (module+name) is probably more consistent... but honestly I don't think it's such a big deal.

**Comment by bclinkinbeard** [comment]
To be clear, nobody (that I am aware of) is advocating for `require('angular')` to return anything besides the `angular` object. That has been the case in every iteration of this PR.

The only thing to decide on is what the supporting modules like `angular-resource` export. It sounds like exporting strings is preferred, which I am fine with.

If the core team can confirm it will be accepted, I will make the change immediately.

**Comment by bendrucker** [comment]
@spacepluk You can always use a relative path to `node_modules` and it'll get picked up. That's pretty brittle though. If you're creating a reusable module and you call `npm dedupe` from a parent, the relative path hack breaks.

**Comment by stryju** [comment]
+1 for exporting module name only, not the module (been doing both and the "name-only" approach feels cleaner and, like @caitp mentioned, discourages the urge to add something to the module)

**Comment by cesarandreu** [comment]
+1 to exporting the module, I've been using the same workflow mentioned above.

**Comment by btford** [comment]
I prefer that it export the name. I think it's better to encourage
workflows that don't depend on messing with core modules.

**Comment by bclinkinbeard** [comment]
> I prefer that it export the name.

@btford: [Done.](https://github.com/bclinkinbeard/angular.js/commit/ed8489a055e81ff74fa4a985292afd2f8a1d8087)

Module name is returned for all but angular and angular-mocks. angular returns the angular object, angular-mocks provides ngAnimateMock.js, ngMock.js and ngMockE2E.js which all return their respective module names via code like `require('angular-mocks/ngMock')`.

Can we pretty please get this merged in now?

Regarding "Default npm packages vs alternative package or tagged release", there is no reason to do anything besides default packages because the existing packages are not usable via CommonJS. **This PR does not break any existing workflows. It only _adds_ proper CommonJS support.**

**Comment by cesarandreu** [comment]
Is returning the module name gonna be the blessed approach by angular?
I can live with whatever choice gets made, as long as there's consistency.

**Comment by caitp** [comment]
Just for the record:

I would personally prefer (for the same reasons Brian mentioned) just exporting the name. That said, it doesn't really prevent people from messing with core modules (although you could argue it encourages it :(). In addition to this, I agree that consistency with what people are already doing is a big deal. So if people are saying "export the whole module", I have to accept that that's what needs to happen. It definitely needs to be as consistent as possible.

**Comment by kentcdodds** [comment]
Based on my experience, I would say that the community is anything but consistent on this matter, so I think that we can safely set a trend here.

**Comment by bclinkinbeard** [comment]
> Based on my experience, I would say that the community is anything but consistent on this matter, so I think that we can safely set a trend here.

Agreed. Returning the name was more popular based on the responses to the mailing list thread, but neither approach is going to be unanimous.

**Comment by caitp** [comment]
> Returning the name was more popular based on the responses to the mailing list thread, but neither approach is going to be unanimous.

I didn't actually get that impression from the thread. I think people seemed to prefer the "whole module" thing. But I agree it's not going to be unanimous one way or the other.

**Comment by petebacondarwin** [comment]
We will announce a final decision on this before the end of this week.

Thanks for everyone's input into this. It has been very constructive.

Your patience will be rewarded :-)

**Comment by othiym23** [comment]
Although it's pretty late in the game for this PR and I don't want to delay your resolution, I do want to add some clarity to the discussion about the future of `peerDependencies` (which is based in part on a discussion npm had with the Angular team last week): https://github.com/npm/npm/issues/6565#issuecomment-74971689

The short version is that `peerDependencies` will be less awkward, but also less powerful, in the next major version of npm, which will be out in the near term.

**Comment by btford** [comment]
Thanks @othiym23! I think how these modules interact with peerDeps has mostly been decided, but good to know. :+1: 

I still feel strongly that `require` should just return the name. I'm going to recap my argument:

---

I think `require` should have the sole responsibility of being the way to consume modules, and `angular.module` be the way to author them.

```js
angular.module('foo', [
  require('bar')
]);
```

Is much more DRY than:

```js
require('bar');
angular.module('foo', [
  'whateverBarIsCalled'
]);
```

It also avoids you having to look inside of `bar` to figure out what the module is called.

---

If you're building up a module with parts defined in different files, you can do this to add to the module:

```js
angular.module(require('./foo')).whatever
```

This API is just slightly more cumbersome compared to:

```js
require('./foo').whatever
```

This is an acceptable tradeoff, because modules almost always have more consumers than authors. We should prefer a tradeoff that makes it easier for consumers at a slight expense of authors.

petebacondarwin:
So the choice is between:

**For consuming modules:**
```js
angular.module('myApp', [require('someDep')]);
```
compared to
```js
angular.module('myApp', [require('someDep').name]);
```

**For extending modules when authoring:**
```js
var myMod = angular.module(require('myMod'));
```
compared to
```js
var myMod = require('myMod');
```

I think the AngularJS team is generally in favour of the former style, for the reasons given in @btford's [comment](https://github.com/angular/angular.js/pull/10732#issuecomment-74986139).

The comments here and in the mailing list have a marginal leaning towards the latter style, but I feel that this is partially because that is just the way that they have been doing it up till now or that it feels right to return some "thing" from a call to require, even though, in reality the AngularJS module system is completely subverting the CJS system here, through its own registration system.

Can we have any final comments here in the next 24 hours so that we can make a decision and get this resolved?

bendrucker:
> The comments here and in the mailing list have a marginal leaning towards the latter style

I think this can be attributing to sampling constraints more than anything. Most of the comments in the Google Group focused more on what people thought other people thought was right, versus concrete reasons to go one way or the other. 

I'm with Brian 100% here. I've been using the name export pattern for a while and have never once needed to modify a module. For the ultra-rare cases in which people want to do that, wrapping the export in `angular.module` is a very acceptable trade off. I've published more than a dozen modules written this way with a decent number of users on some of them. I haven't gotten any issues or complaints about the export convention.

<hr>

Also, from a usability perspective, encouraging people to export strings is a more robust convention given the existing splits between exporting a module and a string. If you tell people that exporting a module is now the blessed way, everything that exports a string now creates an empty error if consumed improperly:

```js
angular.module('myApp', [
  require('module-that-exports-string').name
]);
```

The exception that `undefined` (i.e. `module.name.name`) is not a valid module name is hard to understand and correct. Reverse the situation and its much easier to understand what's going on:

```js
angular.module('myApp', [
  require('module-that-exports-module')
]);
```

Now you get a clear error where you can see that where you should have had the module name you had a module object instead and need to add `.name`. 

So while people migrate to the string convention from exporting a module, there's clear feedback for users. If the reverse had to happen, there would be a lot more confusion.

stryju:
please correct me if i'm wrong, but if my angular module depends on, let's say, `module X`, i don't intend to add anything to the `module X`, just to my module.

if I want to use a component (be that provider, controller, etc.) from `module X`, angular's `DI` is my "friend", no?

---

hence I'm more for getting back the module's name only, since there's no need to expose full module

bendrucker:
> if I want to use a component (be that provider, controller, etc.) from module X, angular's DI is my "friend", no?

That's correct.

petebacondarwin:
Thank you everyone for your constructive input in this feature. It has been good to allow everyone to put forward their views and experience.

We are going to go with the "exporting a string" version.  We were never going to get unanimous agreement on the final solution but we feel that this is the right way to go for Angular 1.x, for the reasons already laid out by @btford earlier in this thread.

Here is the breakdown of the angular project's exports:

## angular package
index.js
```js
require('./angular');
module.exports = angular;
```

## other single module packages:
index.js
```js
require('./angular-animate');
module.exports = 'ngAnimate';
```

## multi module packages

There is no index.js for the multi module packages

ngMock.js
```js
require('./angular-mocks');
module.exports = 'ngMock';
```

ngMockE2E.js
```js
require('./angular-mocks');
module.exports = 'ngMockE2E';
```

ngAnimateMock.js
```js
require('./angular-mocks');
module.exports = 'ngAnimateMock';
```

## Usage

```js
var angular = require('angular');
angular.module('myMod', [
  require('angular-animate'),
  require('angular-mocks/ngMock')
  require('angular-mocks/ngAnimateMock')
]);
```

## Implementation

The only thing to say now is that I am not happy making these updates in the build scripts.

Instead I would like to simply update each of the `bower_angular...` repositories with the correct `package.json`, `index.js` and other appropriate files.  This removes this complexity from the build scripts and puts the export information directly into the repositories.

The overhead of this is that we need to remember to fix up these files when we create new angular modules, but there is already a number of manual steps required in these cases to create these repositories and setup the npm and bower publication mechanisms.

@bclinkinbeard  - would you like to put together PRs for each of these repositories? Once they have been updated, we can then close this PR.

bclinkinbeard:
My first PR made these changes directly to the bower-angular repository, but @btford and @caitp felt the manual approach was not right and pointed me to the build script. If the team has come to a different decision and would now rather manually update 13 repositories instead of adding 35 lines to the script that is obviously their right, but somebody else will need to make the changes.

caitp:
yeah, pete wants to just manually set up the bower repos. i'm not keen on it, especially if we have to undo it and redo it. but, it's all good

bendrucker:
Alrighty, so I stepped up to the plate here and sent PRs to update the bower-* repo. This is important enough to me to burn an hour getting that done. 

I left out ngLocale because the same index.js method can't be applied. We'd have to go the ngMock route but with dozens of files. I can write up a script to generate those if we can agree on a file naming convention.

caitp:
I wouldn't worry about ngLocale right away. The naming convention is easy (e.g. "angular-i18n/en-us") but the challenge is that the actual set of locale modules changes each time closure is upgraded. New ones added, some removed. Do I don't think it makes a lot of sense to do that one manually

petebacondarwin:
Thanks @bendrucker - I didn't think about ngLocale. My reasoning for putting the changes in the bower repositories is that the changes are close to where the impact is, rather than hidden inside some build script somewhere. It is more obvious where to look if you want to change it.

Perhaps we could have a script in the actual `bower-angular-i18n` repository that does the generation?

bendrucker:
> Perhaps we could have a script in the actual bower-angular-i18n repository that does the generation?

That'll work only if the language set of ngLocale is never expected to change. If it does, someone needs to know to go in to bower-angular-i18n and run the script which doesn't make much sense. 

Seems like it would be much better to drop `main` from the package and then write the files with the require statement as `${correctedLocaleId}.js` from the build script here: https://github.com/angular/angular.js/blob/3bc429ad9f347139d09dceae6aeaaf7f16e08ddd/i18n/src/closureSlurper.js#L76-L89

petebacondarwin:
Yes, @bendrucker, I agree that in that case the best place for this is in the `closureSlurper`

bendrucker:
@petebacondarwin Opened #11161 to add that and then (hopefully) we're good to go.

petebacondarwin:
Actually, I take that back. It is not the responsibility of that script to create CJS compatible files, which would pollute our build folders.  I think we should have a script inside the `bower-angular-i18n` repository, which is called by the release script whenever we are releasing a new version.  In this way, we are locating the code to generate these files local to where they are found but controlling it from the release script, so it is still done automatically.

@bendrucker  - ah! You are too fast, I didn't manage to write my new comment before I saw that you implemented it already.

bendrucker:
A quick update for the folks at home. I worked with @petebacondarwin to get all these ready this morning, including `ngLocale` such that you can do `angular.module('app', [require('angular-i18n/aa')])`. There's a small change to land (#11164) to support that and then we're good to go. All future builds will be CommonJS compatible! :tada:

spacepluk:
Thank you! :)

kentcdodds:
:clap:

jgoux:
Awesome work, thanks a lot !

petebacondarwin:
Thanks to all you guys who have contributed to this effort and discussion, especially @bclinkinbeard, @bendrucker, @kentcdodds, @caitp and @btford!

snapwich:
Good work guys, I think this is great for Angular community.

I do have some questions regarding CommonJS support and Angular's global export and it seems like this would be the best place to voice them; @caitp let me know if this should be placed somewhere else.

Are there any plans, or could there be plans, to prevent Angular's global export along with this effort for CommonJS compatibility?  Most CommonJS libraries do not usually export to window.  Also, as is common in the UMD pattern, modules are only exported to window if the CommonJS exports is not detected.  So while the way CommonJS is implemented here will make require'ing angular a possibility (awesome, thanks again guys), it won't provide the normal encapsulation one would expect from using a CommonJS library.

This primarily seems to be a concern in the situation of multiple angular's being present on the page.  In my situation at @needle our application is injected into third-party pages which may or may not already contain an incompatible version of angular or no angular at all.  It's probably not a common use-case, but an important one for us as well as some other companies I have talked to while attending ng-conf.

To fix this problem personally for our application I use a custom version of angular bundled with our application and bootstrapped manually when it's injected.  The changes I made can be seen here: https://github.com/snapwich/angular.js/commit/00d26681164b47b7c556553ce39e3860528986eb

Does anyone else find disabling global export as a useful use-case (either as I did above, or in some other way), and is it something we could consider adding to the master branch?

linclark:
I've written up some [documentation](http://blog.npmjs.org/post/114584444410/using-angulars-new-improved-browserify-support) on how to make use of this.

btford:
Thanks, Lin!

micahblu:
Any plans to support AMD, most specifically requireJS?

stryju:
@micahblu it's been "supported" for a long time, with shims.

also, you can easily load common.js modules in requirejs: http://requirejs.org/docs/api.html#packages

PR #10735, opened 2015-01-13
Changes at controller's method $scope.reset

luisgallon:
I changed the controller's method $scope.reset because it seems that is not reseting the input fields the user filled out, and neither is restoring to their initial state both model objects: $scope.master and $scope.user.

I hope this help to improve the sample.

### DIFF
--- a/docs/content/guide/forms.ngdoc
+++ b/docs/content/guide/forms.ngdoc
@@ -44,7 +44,9 @@ for other directives to augment its behavior.
           };
 
           $scope.reset = function() {
-            $scope.user = angular.copy($scope.master);
+            //$scope.user = angular.copy($scope.master);
+            $scope.master = {};
+            $scope.user = {};
           };
 
           $scope.reset();


luisgallon:
hi,

I just accepted a Contributor License Agreement (CLA) at https://cla.developers.google.com/.

Please let me know if I should take further steps to send my contribution.


Thanks a lot,
Luis Fdo.

docs/content/guide/forms.ngdoc
```diff
@@ -44,7 +44,9 @@ for other directives to augment its behavior.
           };
 
           $scope.reset = function() {
-            $scope.user = angular.copy($scope.master);
+            //$scope.user = angular.copy($scope.master);
+            $scope.master = {};
+            $scope.user = {};
```
gkalpak: I don't think the point of resetting is to clear the fields/data. The point is to undo any (non-committed) changes, i.e. make `user` a copy of `master`.

petebacondarwin: I agree. The idea is that the `$scope.user` object is modified when the input control values are changed. Clicking reset returns it to the "saved" value, which is in `$scope.master`.

petebacondarwin:
@gkalpak - thanks for commenting. Can you get into the habit of assigning labels when you are triaging too? Great!

PR #10745, opened 2015-01-14
fix(angular.copy): should be able to angular.copy Float32Array

### LINKED ISSUE #10210: angular.copy doesn't work with typed arrays
Hello, please consider the following plunker:

http://plnkr.co/edit/AgjiBp2YtFKt8VfqBosT?p=preview

The call to angular.copy() passing a typed array as parameter throws exception:  length method called on incompatible Object at isArrayLike.
Using Firefox 33 on Linux

Thanks for the support

DropsOfSerenity:
angular.copy can now copy a Float32Array when only provided src. I added some tests to confirm this.

fixes #10210

### DIFF
--- a/docs/content/error/ng/cpta.ngdoc
+++ b/docs/content/error/ng/cpta.ngdoc
@@ -1,0 +1,7 @@
+@ngdoc error
+@name ng:cpta
+@fullName Copying TypedArray
+@description
+
+Copying TypedArray's with a destination is not supported because TypedArray
+objects can not be mutated, they are fixed length.
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -500,7 +500,6 @@ function isDate(value) {
   return toString.call(value) === '[object Date]';
 }
 
-
 /**
  * @ngdoc function
  * @name angular.isArray
@@ -583,6 +582,10 @@ function isPromiseLike(obj) {
   return obj && isFunction(obj.then);
 }
 
+var TYPED_ARRAY_REGEXP = /^\[object (Uint8(Clamped)?)|(Uint16)|(Uint32)|(Int8)|(Int16)|(Int32)|(Float(32)|(64))Array\]$/;
+function isTypedArray(value) {
+  return TYPED_ARRAY_REGEXP.test(toString.call(value));
+}
 
 var trim = function(value) {
   return isString(value) ? value.trim() : value;
@@ -705,12 +708,18 @@ function copy(source, destination, stackSource, stackDest) {
     throw ngMinErr('cpws',
       "Can't copy! Making copies of Window or Scope instances is not supported.");
   }
+  if (isTypedArray(destination)) {
+    throw ngMinErr('cpta',
+      "Can't copy! TypedArray destination cannot be mutated.");
+  }
 
   if (!destination) {
     destination = source;
     if (source) {
       if (isArray(source)) {
         destination = copy(source, [], stackSource, stackDest);
+      } else if (isTypedArray(source)) {
+        destination = new source.constructor(source);
       } else if (isDate(source)) {
         destination = new Date(source.getTime());
       } else if (isRegExp(source)) {
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -78,6 +78,186 @@ describe('angular', function() {
       expect(copy(objWithRegExp.re) === objWithRegExp.re).toBeFalsy();
     });
 
+    it("should copy a Uint8Array with no destination", function() {
+      if (typeof Uint8Array !== 'undefined') {
+        var src = new Uint8Array(2);
+        src[1] = 1;
+        var dst = copy(src);
+        expect(copy(src) instanceof Uint8Array).toBeTruthy();
+        expect(dst).toEqual(src);
+        expect(dst).not.toBe(src);
+      }
+    });
+
+    it("should copy a Uint8ClampedArray with no destination", function() {
+      if (typeof Uint8ClampedArray !== 'undefined') {
+        var src = new Uint8ClampedArray(2);
+        src[1] = 1;
+        var dst = copy(src);
+        expect(copy(src) instanceof Uint8ClampedArray).toBeTruthy();
+        expect(dst).toEqual(src);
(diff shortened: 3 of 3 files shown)


DropsOfSerenity:
@googlebot done.

docs/content/error/ng/cpfad.ngdoc
```diff
@@ -0,0 +1,7 @@
+@ngdoc error
+@name ng:cpfad
+@fullName Copying Float32Array
```
caitp: s/Float32Array/%TypedArray%/g

src/Angular.js
```diff
@@ -583,6 +582,9 @@ function isPromiseLike(obj) {
   return obj && isFunction(obj.then);
 }
 
+function isFloat32Array(value) {
+  return toString.call(value) === '[object Float32Array]';
```
caitp: should be `isTypedArray()`, tbh. Using O.p.toString isn't very reliable with ES6 anymore, but we should be able to test it against a regexp like `/^[object (Uint8(Clamped)?)|(Uint16)|(Uint32)|(Int8)|(Int16)|(Int32)|(Float(32)|(64)Array]$`, I think

src/Angular.js
```diff
@@ -705,12 +707,20 @@ function copy(source, destination, stackSource, stackDest) {
...
 
   if (!destination) {
     destination = source;
     if (source) {
       if (isArray(source)) {
         destination = copy(source, [], stackSource, stackDest);
+      } else if (isFloat32Array(source)) {
+        destination = new Float32Array(source.length);
+        for (var j = 0; j < source.length; j++) {
+          destination[j] = source[j];
```
caitp: this is not actually necessary, you can just do `destination = new source.constructor(source)` (or `new <specific constructor>(source)` if worried about `constructor` being overwritten)

caitp: (typed array support is great, but this should not be Float32-specific)

petebacondarwin:
+1 @caitp 

I was just looking at this too and this is just what I was thinking.
If we are going to support copying Float32Array then we should support all type arrays. Unfortunately IE9 doesn't support them so we need to tread carefully but @caitp's ideas are the direction to go.

@DropsOfSerenity - if you would like to update your PR accordingly that would be great.

- [ ] Update the error doc to be more generic
- [ ] Create a `isTypedArray()` method instead of specific one. Perhaps we could compare `src.constuctor.name` with /(Uint8(Clamped)?)|(Uint16)|(Uint32)|(Int8)|(Int16)|(Int32)|(Float(32)|(64))Array/`
- Update the `copy` method accordingly
- Update the docs

DropsOfSerenity:
Excellent, I'll get to it, on a side-note I don't know why the CI build keeps failing, all the tests pass over here.

petebacondarwin:
Don't worry. We are having flakiness issues connecting to SauceLabs and BrowserStack when running tests on Travis

DropsOfSerenity:
@caitp @petebacondarwin 

Alright thanks for the great advice, I've added a new commit that supports all TypedArrays. I did try using constructor.name instead however that resulted in a memory leak for reasons I don't understand.

One other approach I was thinking of could be to use instanceOf, but then I'd have to add a check to make sure the Type being compared to actually exists (in the case of ie9- it doesn't)

Either way this solution seems to be working well, updated the tests to check all TypedArrays.

DropsOfSerenity:
@caitp sweet looks like CI finally passed :)

caitp:
party =) I'll take a look at this later today, gotta finish up a quick patch I'm writing first

docs/content/error/ng/cpta.ngdoc
```diff
@@ -0,0 +1,7 @@
+@ngdoc error
+@name ng:cpfad
```
caitp: should be `@name ng:cpta`

src/Angular.js
```diff
@@ -583,6 +582,9 @@ function isPromiseLike(obj) {
   return obj && isFunction(obj.then);
 }
 
+function isTypedArray(value) {
+  return /^\[object (Uint8(Clamped)?)|(Uint16)|(Uint32)|(Int8)|(Int16)|(Int32)|(Float(32)|(64))Array\]$/.test(toString.call(value));
```
caitp: It's not going to matter to code-gen (at least in v8), but we should probably declare the regexp as a variable outside of the function, so that it's more readable

test/AngularSpec.js
```diff
@@ -78,6 +78,42 @@ describe('angular', function() {
...
+          expect(dst).toEqual(src);
+          expect(dst).not.toBe(src);
+        }
+      }
+    });
+
+    it("should throw an exception if a TypedArray is the destination", function() {
+      if (typeof Float32Array !== 'undefined') {
+        var arraysToCheck = [
+          Uint8Array, Uint16Array, Uint32Array, Int8Array, Int16Array,
```
caitp: we should actually split these into multiple `it()` blocks... @matsko has a tool for doing this in https://github.com/angular/angular.js/pull/10864 but it has not landed yet. Matias do you think you could get that in soon or else submit a followup PR to cleanup these tests if that lands first?

caitp:
I lied, I guess I did review the PR before finishing my patch :3 It looks good to me, just fix up the nits and squash the commits

test/AngularSpec.js
DropsOfSerenity: @caitp I can split these, it's not big deal, don't wanna commit code that would have to be cleaned up later :)

DropsOfSerenity:
@caitp Ok i squashed with my changes. :3

caitp:
good stuff, looks ready to merge to me

caitp:
i guess this was really a feature rather than a fix >_- oh well, not changing that message :>

Hypercubed:
This change is great... have you considered exposeding `istypedArray` as `angular.isTypedArray`?

caitp:
glad you like it @Hypercubed! but... exposing these utilities is problematic in many ways =\ i don't think it would be smart... Besides, `@@toStringTag` makes it unreliable anyhow

PR #10747, opened 2015-01-14
fix($location): strip off empty hash segments when comparing

tleruitte:
Backported from e93710fe0e4fb05ceee59a04f290692a5bec5d20

The url is the same whether or not there is an empty `#` marker at the end.
This prevents unwanted calls to update the browser, since the browser is
automatically applying an empty hash if necessary to prevent page reloads.

Closes #9635


tleruitte:
Close in favour of #10748 . Sorry for the disagreement.

PR #10748, opened 2015-01-14
fix($location): strip off empty hash segments when comparing

tleruitte:
Backported from e93710fe0e4fb05ceee59a04f290692a5bec5d20

The url is the same whether or not there is an empty `#` marker at the end.
This prevents unwanted calls to update the browser, since the browser is
automatically applying an empty hash if necessary to prevent page reloads.

Closes #9635

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -68,6 +68,10 @@ function stripHash(url) {
   return index == -1 ? url : url.substr(0, index);
 }
 
+function trimEmptyHash(url) {
+  return url.replace(/(#.+)|#$/, '$1');
+}
+
 
 function stripFile(url) {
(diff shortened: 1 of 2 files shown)


tleruitte:
I've already signed the CLA as a corporation. The corporation name is 'Whatever SA'. Should I resend the CLA to cla-submissions@google.com?

tleruitte:
I've force-pushed the typo that caused travis to fail.

tleruitte:
So the tests fail on travis because of a timeout, but everything works fine when I run the tests locally. Is there anything I can do to fix that?

petebacondarwin:
Landed in 1.2.x as e81b2f7.
Not sure when the next release will be...

PR #10750, opened 2015-01-14
Adding ng-jq functionality, the ability to choose your angular.element l...

mboudreau:
...ibrary or force jqLite even if jquery is available.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -767,6 +767,36 @@ function equals(o1, o2) {
   return false;
 }
 
+/**
+ * @ngdoc directive
+ * @name ng.directive:ngJq
+ *
+ * @element html
+ * @param {string} (optional) the name of the library available under `window` to be used for angular.element
+ *
+ * @description
+ *
+ * Use this directive to force the angular.element library.  This should be used to force
+ * either jqLite by leaving ng-jq blank or setting the name of the jquery variable under window
+ * (eg. jQuery).
+ *
+ * This directive is global for the whole of the angular library, hence the directive can only
+ * be added to the top-most HTML element.
+ *
+ */
+var jq = function () {
+  if (isDefined(jq.name_)) return jq.name_;
+
+  var el = document.querySelector('[ng-jq]') || document.querySelector('[data-ng-jq]');
+  var name = undefined;
+
+  if (el) {
+    name = el.getAttribute('ng-jq') || el.getAttribute('data-ng-jq') || '';
+  }
+
+  return (jq.name_ = name);
+};
+
 
 function concat(array1, array2, index) {
   return array1.concat(slice.call(array2, index));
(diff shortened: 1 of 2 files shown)


src/Angular.js
```diff
@@ -1147,7 +1177,7 @@ function snake_case(name, separator){
 
 function bindJQuery() {
   // bind to jQuery if present;
-  jQuery = window.jQuery;
+  jQuery = jq() ? window[jq()] : window.jQuery;
```
lgalfaso: I think this will not do what you are expecting as `''` is falsy

lgalfaso: BTW, please do not call `jq()` twice

src/Angular.js
```diff
@@ -767,6 +767,36 @@ function equals(o1, o2) {
...
+ * (eg. jQuery).
+ *
+ * This directive is global for the whole of the angular library, hence the directive can only
+ * be added to the top-most HTML element.
+ *
+ */
+var jq = function () {
+  if (isDefined(jq.name_)) return jq.name_;
+
+  var el = document.querySelector('[ng-jq]') || document.querySelector('[data-ng-jq]');
```
lgalfaso: we should follow the same mechanism that we have for `ng-app`

caitp: @lgalfaso this branch looks like it's very old, a lot of things are missing. In current branch, it's possible to use `getNgAttribute(rootElement, 'jq')` to check for these bootstrap parameters

lgalfaso: you are right, the branch is really old

src/Angular.js
mboudreau: @lgalfaso hm, guess I need to change my unit tests then.  Thanks for the heads up.  Will change right now.

src/Angular.js
mboudreau: @lgalfaso I'm going on @mzgol suggestions on how to implement this, which was based on ng-csp since angular.element is a global variable, unlike ng-app which you can have multiple in a page.

As for the branch itself, this was the latest that I could get from github, fetching the latest from upstream and making sure that my master has the same last commit than the angular one.  If there's any branch 'weirdness', I can't comment why that would be.

lgalfaso: The parent of 9b7b0fc7dbabab41313c17acf8642e69efccce2a is 28fe4464d8f40409dc05a290224cc06883aed2f6

28fe4464d8f40409dc05a290224cc06883aed2f6 was committed on Oct 6, 2013

mboudreau: Crap.  That's weird.  Something must of messed up somewhere.

Let me add some of your suggestions and I'll create a new pull request based on the right parent.

Cheers.

mboudreau:
@lgalfaso 

Added new pull request: https://github.com/angular/angular.js/pull/10761

PR #10757, opened 2015-01-14
Remove sortedKeys()

realityking:
Since #10639 there's only one caller to `sortedKeys()` left, so it can be inlined to save some bytes.

Second commit applies #9704 and this change to `testabilityPatch.js` to keep it in sync.


pkozlowski-opensource:
LGTM, restarted Travis job to see what is has to say :-)

PR #10761, opened 2015-01-15
Adding ng-jq functionality, the ability to force jqLite or any other lib...

mboudreau:
...rary by name to be used for angular.element

Added unit tests to check that jq function returns the right result, but couldn't test adding a fake jQuery object without failing many other tests.  Maybe this is better for e2e tests?  I would add some, but couldn't figure out the structure.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -58,6 +58,7 @@
   shallowCopy: true,
   equals: true,
   csp: true,
+  jq: true,
   concat: true,
   sliceArgs: true,
   bind: true,
@@ -899,7 +900,61 @@ var csp = function() {
   return (csp.isActive_ = active);
 };
 
+/**
+ * @ngdoc directive
+ * @module ng
+ * @name ngJq
+ *
+ * @element ANY
+ * @param {string=} the name of the library available under `window`
+ * to be used for angular.element
+ * @description
+ * Use this directive to force the angular.element library.  This should be
+ * used to force either jqLite by leaving ng-jq blank or setting the name of
+ * the jquery variable under window (eg. jQuery).
+ *
+ * Since this directive is global for the angular library, it is recommended
+ * that it's added to the same element as ng-app or the HTML element, but it is not mandatory.
+ * It needs to be noted that only the first instance of `ng-jq` will be used and all others
+ * ignored.
+ *
+ * @example
+ * This example shows how to force jqLite using the `ngJq` directive to the `html` tag.
+ ```html
+ <!doctype html>
+ <html ng-app ng-jq>
+ ...
+ ...
+ </html>
+ ```
+ * @example
+ * This example shows how to use a jQuery based library of a different name.
+ * The library name must be available at the top most 'window'.
+ ```html
+ <!doctype html>
+ <html ng-app ng-jq="jQueryLib">
+ ...
+ ...
+ </html>
+ ```
+ */
+var jq = function() {
+  if (isDefined(jq.name_)) return jq.name_;
+  var el;
+  var i, ii = ngAttrPrefixes.length;
+  for (i = 0; i < ii; ++i) {
+    if (el = document.querySelector('[' + ngAttrPrefixes[i].replace(':', '\\:') + 'jq]')) {
+      break;
+    }
+  }
 
+  var name;
+  if (el) {
+    name = getNgAttribute(el, "jq");
+  }
+
+  return (jq.name_ = name);
+};
 
 function concat(array1, array2, index) {
   return array1.concat(slice.call(array2, index));
@@ -1472,7 +1527,12 @@ function bindJQuery() {
   }
 
   // bind to jQuery if present;
-  jQuery = window.jQuery;
+  var jqName = jq();
+  jQuery = window.jQuery; // use default jQuery.
+  if (isDefined(jqName)) { // `ngJq` present
+    jQuery = jqName === null ? undefined : window[jqName]; // if empty; use jqLite. if not empty, use jQuery specified by `ngJq`.
+  }
+
   // Use jQuery if it exists with proper functionality, otherwise default to us.
   // Angular 1.2+ requires jQuery 1.7+ for on()/off() support.
   // Angular 1.3+ technically requires at least jQuery 2.1+ but it may work with older
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -615,6 +615,94 @@ describe('angular', function() {
   });
 
 
+  describe('jq', function() {
+    var element;
+
+    beforeEach(function() {
+      element = document.createElement('html');
+    });
+
+    afterEach(function() {
+      delete jq.name_;
+    });
+
+    it('should return undefined when jq is not set, no jQuery found (the default)', function() {
+      expect(jq()).toBe(undefined);
+    });
+
+    it('should return empty string when jq is enabled manually via [ng-jq] with empty string', function() {
+      element.setAttribute('ng-jq', '');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[ng-jq]') return element;
+      });
+      expect(jq()).toBe('');
+    });
+
+    it('should return empty string when jq is enabled manually via [data-ng-jq] with empty string', function() {
+      element.setAttribute('data-ng-jq', '');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[data-ng-jq]') return element;
+      });
+      expect(jq()).toBe('');
+      expect(document.querySelector).toHaveBeenCalledWith('[data-ng-jq]');
+    });
+
+    it('should return empty string when jq is enabled manually via [x-ng-jq] with empty string', function() {
+      element.setAttribute('x-ng-jq', '');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[x-ng-jq]') return element;
+      });
+      expect(jq()).toBe('');
+      expect(document.querySelector).toHaveBeenCalledWith('[x-ng-jq]');
+    });
+
+    it('should return empty string when jq is enabled manually via [ng:jq] with empty string', function() {
+      element.setAttribute('ng:jq', '');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[ng\\:jq]') return element;
+      });
+      expect(jq()).toBe('');
+      expect(document.querySelector).toHaveBeenCalledWith('[ng\\:jq]');
+    });
+
+    it('should return "jQuery" when jq is enabled manually via [ng-jq] with value "jQuery"', function() {
+      element.setAttribute('ng-jq', 'jQuery');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[ng-jq]') return element;
+      });
+      expect(jq()).toBe('jQuery');
+      expect(document.querySelector).toHaveBeenCalledWith('[ng-jq]');
+    });
+
+    it('should return "jQuery" when jq is enabled manually via [data-ng-jq] with value "jQuery"', function() {
+      element.setAttribute('data-ng-jq', 'jQuery');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[data-ng-jq]') return element;
+      });
+      expect(jq()).toBe('jQuery');
+      expect(document.querySelector).toHaveBeenCalledWith('[data-ng-jq]');
+    });
+
+    it('should return "jQuery" when jq is enabled manually via [x-ng-jq] with value "jQuery"', function() {
+      element.setAttribute('x-ng-jq', 'jQuery');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[x-ng-jq]') return element;
+      });
+      expect(jq()).toBe('jQuery');
+      expect(document.querySelector).toHaveBeenCalledWith('[x-ng-jq]');
+    });
+
+    it('should return "jQuery" when jq is enabled manually via [ng:jq] with value "jQuery"', function() {
+      element.setAttribute('ng:jq', 'jQuery');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector === '[ng\\:jq]') return element;
+      });
+      expect(jq()).toBe('jQuery');
+      expect(document.querySelector).toHaveBeenCalledWith('[ng\\:jq]');
+    });
+  });
+
+
   describe('parseKeyValue', function() {
     it('should parse a string into key-value pairs', function() {
       expect(parseKeyValue('')).toEqual({});
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -60,6 +60,7 @@
     "shallowCopy": false,
     "equals": false,
     "csp": false,
+    "jq": false,
     "concat": false,
     "sliceArgs": false,
     "bind": false,


mboudreau:
@mzgol @lgalfaso 
Here's the new pull request that should fix ng-jq for angular master branch.

mzgol:
@mboudreau Thanks, I'll review it later today.

Note for the future - never commit anything to the master branch, this is probably the source of your problems. master should be used exclusively to pull from the upstream master so that you can easily have the same master locally as the latest master in the Angular repository.

Your changes should be made on a different branch that you can then easily rebase on top of master.

Another thing: if sth goes wrong it's not necessary to open a new PR; it's enough if you amend the commit, introduce changes, rebase on top of latest master and do "git push --force". GitHub will pick it up as long as the branch name doesn't change.

You can read this guide: http://contribute.jquery.org/commits-and-pull-requests/
This is for jQuery but things not related to commit message format are universal to working with most other open source projects, including Angular.

mboudreau:
@mzgol Thanks for the tips.  I already knew about them, I just didn't want to mess up things for the Angular team and I figured a redo would of been simpler.  Maybe not.

Either way, it's done now :P 

@petebacondarwin Is there anything else you want me to do?  Like add that public api?

mzgol:
@mboudreau Fire `grunt test` in terminal, you'll see a couple of linting errors, e.g. not exporting the `jq` variable (it needs to be added to the beginning of the file in the `/* global` block.

src/Angular.js
```diff
@@ -893,7 +893,35 @@ var csp = function() {
...
+ * @param {string} (optional) the name of the library available under `window` to be used for angular.element
+ *
+ * @description
+ *
+ * Use this directive to force the angular.element library.  This should be used to force
+ * either jqLite by leaving ng-jq blank or setting the name of the jquery variable under window
+ * (eg. jQuery).
+ *
+ * This directive is global for the whole of the angular library, hence the directive can only
+ * be added to the top-most HTML element.
```
mzgol: This is not true as you're catching `ng-jq` anywhere in the document. I'd rather copy the text from `ngCsp.js`:

"put the `ngJq` directive on the root element of the application or on the `angular.js` script tag, whichever appears first in the html document."

so that it's advised to put it there but not claim it won't work otherwise.

src/Angular.js
```diff
@@ -893,7 +893,35 @@ var csp = function() {
   return (csp.isActive_ = active);
 };
 
+/**
+ * @ngdoc directive
+ * @name ng.directive:ngJq
+ *
+ * @element html
+ * @param {string} (optional) the name of the library available under `window` to be used for angular.element
```
mzgol: optional params are marked with `=` at the end of the type, e.g. here `@param {string=}`; remove `(optional)`.

src/Angular.js
```diff
@@ -893,7 +893,35 @@ var csp = function() {
...
+ *
+ * This directive is global for the whole of the angular library, hence the directive can only
+ * be added to the top-most HTML element.
+ *
+ */
+var jq = function() {
+  if (isDefined(jq.name_)) return jq.name_;
 
+  var el = document.querySelector('[ng-jq]') || document.querySelector('[data-ng-jq]');
+  var name = undefined;
```
mzgol: No need to initialize to `undefined` (this & some other mistakes are catched by `grunt test`; linting errors in particular can be checked via `grunt jshint`).

src/Angular.js
```diff
@@ -1462,7 +1490,9 @@ function bindJQuery() {
   }
 
   // bind to jQuery if present;
-  jQuery = window.jQuery;
+  var jqName = jq();
+  jQuery = jqName !== undefined ? window[jqName] : window.jQuery;
```
mzgol: `jqName != null` is enough

src/Angular.js
```diff
@@ -893,7 +893,35 @@ var csp = function() {
...
+ *
+ * Use this directive to force the angular.element library.  This should be used to force
+ * either jqLite by leaving ng-jq blank or setting the name of the jquery variable under window
+ * (eg. jQuery).
+ *
+ * This directive is global for the whole of the angular library, hence the directive can only
+ * be added to the top-most HTML element.
+ *
+ */
+var jq = function() {
```
mzgol: jsHint will catch it (see my other comments), you can search for `csp` and see where it needs to be added to pass jsHint tests.

mzgol:
> @petebacondarwin Is there anything else you want me to do? Like add that public api?

`csp` needed to be exposed to `$sniffer` as it's used later in the code. `jq` is needed only during bootstrap, the rest is left to the chosen library so I think we're good.

mzgol:
@mboudreau This looks mostly good, only a couple of remarks. :+1:

mboudreau:
@mzgol Added the fixes from your comments. Cheers.

src/Angular.js
petebacondarwin: @mboudreau - don't forget this change!

src/Angular.js
petebacondarwin: +1

mzgol: @mboudreau If you could rebase your PR over latest `master` after introducing remaining changes, that'd be great. It seems you branched at a commit that fails tests so it's impossible for your PR to pass tests because of that. It should be fine on latest `master` (am I right, @petebacondarwin?).

petebacondarwin: I would expect so @mzgol. Thanks for reviewing this one!

src/Angular.js
mboudreau: @petebacondarwin Already done.

mzgol: @mboudreau What do you mean? The diff shows nothing changed here.

mboudreau: @mzgol Look again, it's already there.  The inline comment seems to be referencing the old diff.  If you look at the file directly from the PR, you'll see that the optional flag is there.

src/Angular.js
mboudreau: @mzgol @petebacondarwin I just tried merging the upstream master to my local without issue, however, the build is still failing even after an npm/bower install/update.  Same error as the CI build (unless there's something different on the build server that I'm missing locally):

```
error:   TypeError: Cannot read property '0' of undefined
    at _.groupBy.tap.tap.docTypes.directive (/home/michelb/Work/angular.js/docs/config/processors/pages-data.js:65:43)
    at Function.tap (/home/michelb/Work/angular.js/node_modules/lodash/dist/lodash.js:6448:7)
    at lodash.ctor.(anonymous function) [as tap] (/home/michelb/Work/angular.js/node_modules/lodash/dist/lodash.js:5924:31)
    at /home/michelb/Work/angular.js/docs/config/processors/pages-data.js:62:14
    at /home/michelb/Work/angular.js/node_modules/lodash/dist/lodash.js:3513:29
    at forOwn (/home/michelb/Work/angular.js/node_modules/lodash/dist/lodash.js:2105:15)
    at Function.map (/home/michelb/Work/angular.js/node_modules/lodash/dist/lodash.js:3512:9)
    at lodash.ctor.(anonymous function) [as map] (/home/michelb/Work/angular.js/node_modules/lodash/dist/lodash.js:5924:31)
    at navGroupMappers.api (/home/michelb/Work/angular.js/docs/config/processors/pages-data.js:53:10)
    at /home/michelb/Work/angular.js/docs/config/processors/pages-data.js:192:28
```

src/Angular.js
mzgol: @mboudreau It's not there; I'm looking at the "Files changed" tab, not the inline comments view. I've also re-downloaded the PR and the change is not there.

Maybe you forgot to push the changes but, believe me, they're not here.

Also, would you be able to rebase the PR over latest master so that we can see the tests run?

src/Angular.js
mzgol: Are you sure you're on a latest `master`? If you're on Linux/OS X, could you invoke:
```
git show -s --format=%ci `git merge-base HEAD master` | head
```
? This should show you a date from today or yesterday; if it's older, it means you didn't rebase over latest changes.

mboudreau: @mzgol `2015-01-21 09:53:33 +1100` is the output.

Everything looks good to me, I got all the changes.  Dunno what I'm missing to build...

Apparently, Travis is having issues too, so I don't know what's going on.  Weird stale files somewhere?

src/Angular.js
mboudreau: @mzgol Okay, this is starting to get weird.  If you look at my master right now, the changes are there: https://github.com/mboudreau/angular.js/blob/master/src/Angular.js

mzgol: This line:
https://github.com/mboudreau/angular.js/blob/fc0ddbf75611dd79da2a19a400dd4e747c290ea6/src/Angular.js#L902
from the top of your `master` branch is still what was here before.

To make sure we're on the same page. It is:
```
 * @param {string} (optional) the name of the library available under `window` to be used for angular.element
```
it should be:
```
 * @param {string=} the name of the library available under `window` to be used for angular.element
```

mboudreau: Weird, apparently, something got duplicated in the docs.  Some weird merge somewhere?  Anywho, that's been fixed and I've pushed again.  I'm still seeing the `grunt package` failing on doc generation though.

mboudreau:
@mzgol @petebacondarwin Build still failing after merge from upstream.  Anyone care to help me debug?  I'm trying to understand this document processor code, but there's a lot of abstraction...

gkalpak:
I don't know for sure, but adding `@module ng` in the directive's ngdoc might help.

src/Angular.js
```diff
@@ -893,7 +894,53 @@ var csp = function() {
   return (csp.isActive_ = active);
 };
 
+/**
+ * @ngdoc directive
+ * @name ngJq
+ *
+ * @element html
```
gkalpak: Why specifically `html` ?

src/Angular.js
```diff
@@ -893,7 +894,53 @@ var csp = function() {
...
+ ```
+ */
+var jq = function() {
+  if (isDefined(jq.name_)) return jq.name_;
 
+  var el = document.querySelector('[ng-jq]') || document.querySelector('[data-ng-jq]');
+  var name;
+
+  if (el) {
+    name = el.getAttribute('ng-jq') || el.getAttribute('data-ng-jq') || '';
```
gkalpak: You are missing several "acceptable" forms (`x`-prefixed, `:`-delimited etc).
You could use `ngAttrPrefixes` (as does `ngApp`), but even that leaves a few prefixes out.

@petebacondarwin: should we care to recognize all prefixes according to directives' normalization rules ?

mzgol: > You are missing several "acceptable" forms (x-prefixed, :-delimited etc).
> You could use ngAttrPrefixes (as does ngApp), but even that leaves a few prefixes out.

csp [is missing them as well](https://github.com/angular/angular.js/blob/331cac233f56dc7f63ab8ac48de0d40fc7256047/src/Angular.js#L877-L894). They're both not regular directives.

gkalpak: I know (`ngApp` isn't a regular directive either). Yet, `ngCsp` shouldn't be that common; I would expect `ngJq` to be more popular.

In any case, since they are presented as directives in the docs, I would either expect them to conform to the directive normalization rules or have an explicit notice about the recognized forms.

<sub>(This is also true for `ngApp`/`ngCsp`, not only `ngJq`.)</sub>

src/Angular.js
petebacondarwin: Should be ANY I think

src/Angular.js
petebacondarwin: I think this should use `ngAttrPrefixes` and then we should look at whether that needs to be made better.

mboudreau:
@gkalpak Thanks for that.  That indeed fixed it.  

I also added all prefixes to the directive.  Thanks for your direction :)

mboudreau:
@mzgol CI still failing, but not because of my code:
```
ScriptTimeoutError: Timed out awaiting response to command "switchToFrame" after 30000 ms (WARNING: The server did not provide any stacktrace information)
```
Is there a way to restart the build?  I'm fairly sure it's just me being unlucky :P

src/Angular.js
```diff
@@ -888,7 +889,59 @@ var csp = function() {
...
+ <!doctype html>
+ <html ng-app ng-jq="jQueryLib">
+ ...
+ ...
+ </html>
+ ```
+ */
+var jq = function() {
+  if (isDefined(jq.name_)) return jq.name_;
+	var el;
```
gkalpak: Indentation should be 2 spaces (not tabs).

src/Angular.js
```diff
@@ -888,7 +889,59 @@ var csp = function() {
...
+	var i, ii = ngAttrPrefixes.length;
+	for (i = 0; i < ii; ++i) {
+		if (el = document.querySelector('[' + ngAttrPrefixes[i].replace(':', '\\:') + 'jq]')) {
+			break;
+		}
+	}
+
+  var name;
+  if (el) {
+    name = getNgAttribute(el, "jq") || '';
```
gkalpak: What is the purpose of `|| ''` ? It will result in `jQuery = window['']` on [line 1514][1] (without any obvious benefit).
Do I miss something (it's happened before :stuck_out_tongue:) ?

[1]: https://github.com/angular/angular.js/pull/10761/files#diff-1d54c5f722aebc473dbe96f836ddf974R1514

test/AngularSpec.js
```diff
@@ -435,6 +435,95 @@ describe('angular', function() {
...
+  describe('jq', function() {
+    var element;
+
+    beforeEach(function() {
+      element = document.createElement('html');
+    });
+
+    afterEach(function() {
+      jq.name_ = undefined;
+      delete jq.name_;
```
gkalpak: Just being curious: Why first set it to undefined and _then_ delete it ?

test/AngularSpec.js
```diff
@@ -435,6 +435,95 @@ describe('angular', function() {
...
+    });
+
+    it('should return undefined when jq is not set, no jquery found (the default)', function() {
+      expect(jq()).toBe(undefined);
+    });
+
+    it('should return empty string when jq is enabled manually via [ng-jq] with empty string', function() {
+      element.setAttribute('ng-jq', '');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector == '[ng-jq]') return element;
```
gkalpak: Better use strict comparison (`===`).

test/AngularSpec.js
```diff
@@ -435,6 +435,95 @@ describe('angular', function() {
...
+    afterEach(function() {
+      jq.name_ = undefined;
+      delete jq.name_;
+    });
+
+    it('should return undefined when jq is not set, no jquery found (the default)', function() {
+      expect(jq()).toBe(undefined);
+    });
+
+    it('should return empty string when jq is enabled manually via [ng-jq] with empty string', function() {
```
gkalpak: I don't get the purpose of the functionality tested here :(

test/AngularSpec.js
```diff
@@ -435,6 +435,95 @@ describe('angular', function() {
...
+    it('should return empty string when jq is enabled manually via [ng:jq] with empty string', function() {
+      element.setAttribute('ng:jq', '');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector == '[ng\\:jq]') return element;
+      });
+      expect(jq()).toBe('');
+      expect(document.querySelector).toHaveBeenCalledWith('[ng\\:jq]');
+    });
+
+    it('should return "jquery" when jq is enabled manually via [ng-jq] with value "jQuery"', function() {
```
gkalpak: _should return "jQuery"_ (capital 'Q')

test/AngularSpec.js
```diff
@@ -435,6 +435,95 @@ describe('angular', function() {
...
+
+    it('should return "jquery" when jq is enabled manually via [ng:jq] with value "jQuery"', function() {
+      element.setAttribute('ng:jq', 'jQuery');
+      spyOn(document, 'querySelector').andCallFake(function(selector) {
+        if (selector == '[ng\\:jq]') return element;
+      });
+      expect(jq()).toBe('jQuery');
+      expect(document.querySelector).toHaveBeenCalledWith('[ng\\:jq]');
+    });
+  });
```
gkalpak: I believe there should be also some tests verifying the correct behaviour when `jQuery` is set on `window` (either with or without `ngJq`).

src/Angular.js
```diff
@@ -888,7 +889,59 @@ var csp = function() {
...
+ ...
+ </html>
+ ```
+ */
+var jq = function() {
+  if (isDefined(jq.name_)) return jq.name_;
+	var el;
+	var i, ii = ngAttrPrefixes.length;
+	for (i = 0; i < ii; ++i) {
+		if (el = document.querySelector('[' + ngAttrPrefixes[i].replace(':', '\\:') + 'jq]')) {
```
gkalpak: There is one possible issue when having multiple manually bootstraped Angular apps on the same page and each one whats to use (or not to use) jQuery (and possibly different versions).
Calling `querySelector` on `document` will always return the first occurence.
Could/Should we somehow associate this with the app's `$rootElement` ?

If not, I believe this should be documented.

gkalpak:
I restarted the flakey build and all is green now.
But I have left some inline comments... :laughing:

src/Angular.js
mzgol: @gkalpak That scenario is intentionally not supported. Note that `angular.element` is set to the chosen jQuery version so it's impossible for two apps running on the same Angular instance to work with two different jQueries. The restriction is similar to `ng-csp`.

This *could* be made possible if bootstrapped apps didn't use `angular.element` but their own references to jQueries used but that's a big change to make, far beyond the scope of this PR.

test/AngularSpec.js
mzgol: It's to differentiate between an `ng-jq` attribute without value that means "just use `jqLite` and not setting it at all which means "use the native detection mechanism".

Not the prettiest APIs of all, I agree, but it's used just in one place really so I didn't picked at it.

test/AngularSpec.js
mzgol: Doesn't jsHint catch this? :/

src/Angular.js
```diff
@@ -1457,7 +1510,9 @@ function bindJQuery() {
   }
 
   // bind to jQuery if present;
-  jQuery = window.jQuery;
+  var jqName = jq();
+  jQuery = jqName != null ? window[jqName] : window.jQuery;
```
mzgol: I've missed it before: an empty `ng-jq` attribute should mean "use jqLite" and not "use the native detection algorithm" as it's done now. Otherwise it'll be impossible for people to opt in to jqLite if jQuery is included on the site which is one of the major points of this functionality. :)

Refs the [above comment](https://github.com/angular/angular.js/pull/10761/files#r23366361) by @gkalpak.

src/Angular.js
mzgol: Nice catch, I've missed what's really happening here... See https://github.com/angular/angular.js/pull/10761/files#r23376730.

src/Angular.js
gkalpak: True. It would be nice to have a short notice that only the 1st occurrence is aken into account.

src/Angular.js
gkalpak: So, the point is that `ng-jq` without a value should mean use jqLite, which is achieved by setting `jQuery = window['']` (which will hopefully be undefined). It kind of makes sense, but I wonder if there is a more robust way.

mboudreau: @gkalpak I'm all ears.  This is the simplest approach I could think of without adding a lot of conditionals.

test/AngularSpec.js
mboudreau: @gkalpak Sorry, force of habit from a time when delete was buggy in javascript.  I guess that's fixed now since Angular supports IE 9 and above.

test/AngularSpec.js
mboudreau: @gkalpak @mzgol Again, open to suggestions.  This is my first time coding angular itself and I'm not sure what would be the best approach.

test/AngularSpec.js
mboudreau: @gkalpak Somebody wanted me to be thorough with all supported delimiters.  This includes tests for all of them even if verbose.

test/AngularSpec.js
mboudreau: @gkalpak I agree, but when I tried to do said test, it would fail all other tests since the real jQuery was being disrupted, which is used in other tests.  I tried using the angularInit tests as an example since I need to test out the bindJquery function.  Any direction would be appreciated.

gkalpak: @mboudreau: I was thinking something "non-disruptive". Maybe verifying what `jqLite` is bound to, both with and without jQuery being loaded (you can differentiate by checking the `_jqLiteMode` global.

But I don't feel too strongly about it :)

src/Angular.js
gkalpak:
```js
var name;
if (el) {
  name = getNgAttribute(el, "jq");
}
```
This would result in `name` being `undefined` if there is no `ngJq` and `null` if `ngJq` is set but empty.

Then on [line 1516][1]: 

```js
var jqName = jq();
jQuery = isUndefined(jqName) ? window.jQuery :   // No `ngJq`; use default jQuery.
    ((jqName === null) ? undefined :             // `ngJq` present, but empty; use jqLite.
                         window[jqName]);        // Use jQuery specified by `ngJq`.
```

[1]: https://github.com/angular/angular.js/pull/10761/files#diff-1d54c5f722aebc473dbe96f836ddf974R1516

mboudreau:
@gkalpak @mzgol Build's still failing.  Getting script timeouts.

src/Angular.js
mboudreau: Yeah, that doesn't sound bad.  Actually, what if we did this? I've only done it for readability.  I don't like going 2 levels in for ternary operators.

```
jQuery = window.jQuery; // use default jQuery.
	if(isDefined(jqName)) { // `ngJq` present
		jQuery = jqName === null ? undefined : window[jqName]; // if empty; use jqLite. if not empty, use jQuery specified by `ngJq`.
	}
```

mboudreau:
@gkalpak @mzgol What's next gents? How do I get my little baby into Angular master?

src/Angular.js
```diff
@@ -1457,7 +1512,12 @@ function bindJQuery() {
   }
 
   // bind to jQuery if present;
-  jQuery = window.jQuery;
+  var jqName = jq();
+	jQuery = window.jQuery; // use default jQuery.
+	if (isDefined(jqName)) { // `ngJq` present
+		jQuery = jqName === null ? undefined : window[jqName]; // if empty; use jqLite. if not empty, use jQuery specified by `ngJq`.
+	}
```
gkalpak: Hm...I still see some tabs here :confused:

mboudreau: @gkalpak  God damnit.  Fixed now.

lgalfaso:
@mboudreau overall looks good, but can you please squash this into one commit?

mboudreau:
@lgalfaso Done :)

lgalfaso:
landed as 09ee82d84dcbea4a6e8d85903af82dcd087a78a7

mboudreau:
Woo! Thanks guys! @lgalfaso @gkalpak @mzgol @petebacondarwin

PR #10763, opened 2015-01-15
optimization in isBoolean

zaheerehmad:
typeof on value and then comparison with string is slower then comparing value directly with boolean ie value === true. It will optimized the performance.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -505,7 +505,6 @@ function isDate(value) {
   return toString.call(value) === '[object Date]';
 }
 
-
 /**
  * @ngdoc function
  * @name angular.isArray
@@ -580,7 +579,7 @@ function isBlob(obj) {
 
 
 function isBoolean(value) {
-  return typeof value === 'boolean';
+  return value === true;
 }
 
 


src/Angular.js
```diff
@@ -580,7 +579,7 @@ function isBlob(obj) {
 
 
 function isBoolean(value) {
-  return typeof value === 'boolean';
+  return value === true;
```
caitp: this is semantically different from what was happening before.

caitp:
I'd be interested in seeing profiling of an application which shows this is worth fixing in this way, because right now this does not lgtm

lgalfaso:
This PR changes the semantics of the function, without fixing this and some numbers that show that it makes a difference, then this is not good for landing. Closing, but if you provide this information, feel free to reopen the PR

src/Angular.js
Burgov: shouldn't it be `value === true || value === false`?

zaheer-ahmed: yes it should be. thanks

PR #10766, opened 2015-01-15
bug(ngAria): native controls fire a single click event

### LINKED ISSUE #10388: bug(ngAria): ng-click adds keypress event to buttons which already fire a click event on "enter"
Originally discussed [here](https://github.com/angular/angular.js/pull/10288#issuecomment-66345338) in pull #10288 

 > Hmmmm. Looks like this introduces [a bug](http://jsbin.com/relejeguke/2/edit)? Looks like this is a problem in Chrome, Firefox, and Safari. Tab to the button (turns red) and hit "Enter" and the alert pops up twice. This is because (I believe) that the aforementioned browsers will fire a "clicked" event when the "enter" key is pressed.

Fix suggested by @marcysutton:

> Ahh, I see that. Good catch! It should only bind on a custom control that doesn't already have keyboard support. The approach used for adding roles in this PR would work:
> https://github.com/angular/angular.js/pull/10318/files#diff-91e9129201746ba107348c9a0a7735edR333

marcysutton:
To fix the widespread accessibility issue of `ngClick` on divs, ngAria dynamically binds `keypress` in addition to `tabIndex`. As first implemented, this introduced a bug where `<button>` elements would have both events bound, causing multiple callbacks to fire. This PR makes limits the addition of keypress to `<div>` and `<li>`, which can be overridden with the `bindKeypress` config option.

Closes https://github.com/angular/angular.js/issues/10388. Thanks to @caitp for the help & unit test on native elements.

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -100,7 +100,8 @@ function $AriaProvider() {
    *  - **ariaMultiline** – `{boolean}` – Enables/disables aria-multiline tags
    *  - **ariaValue** – `{boolean}` – Enables/disables aria-valuemin, aria-valuemax and aria-valuenow tags
    *  - **tabindex** – `{boolean}` – Enables/disables tabindex tags
-   *  - **bindKeypress** – `{boolean}` – Enables/disables keypress event binding on ng-click
+   *  - **bindKeypress** – `{boolean}` – Enables/disables keypress event binding on `&lt;div&gt;` and
+   *    `&lt;li&gt;` elements with ng-click
    *
    * @description
    * Enables/disables various ARIA attributes
@@ -303,11 +304,18 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
     compile: function(elem, attr) {
       var fn = $parse(attr.ngClick, /* interceptorFn */ null, /* expensiveChecks */ true);
       return function(scope, elem, attr) {
+
+        function isNodeOneOf(elem, nodeTypeArray) {
+          if (nodeTypeArray.indexOf(elem[0].nodeName) !== -1) {
+            return true;
+          }
+        }
+
         if ($aria.config('tabindex') && !elem.attr('tabindex')) {
           elem.attr('tabindex', 0);
         }
 
-        if ($aria.config('bindKeypress') && !attr.ngKeypress) {
+        if ($aria.config('bindKeypress') && !attr.ngKeypress && isNodeOneOf(elem, ['DIV', 'LI'])) {
           elem.on('keypress', function(event) {
             if (event.keyCode === 32 || event.keyCode === 13) {
               scope.$apply(callback);
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -488,12 +488,24 @@ describe('$aria', function() {
 
     it('should a trigger click from the keyboard', function() {
       scope.someAction = function() {};
-      compileInput('<div ng-click="someAction()" tabindex="0"></div>');
+
+      var elements = $compile('<section>' +
+                  '<div class="div-click" ng-click="someAction(\'div\')" tabindex="0"></div>' +
+                  '<ul><li ng-click="someAction( \'li\')" tabindex="0"></li></ul>' +
+                  '</section>')(scope);
+
+      scope.$digest();
+
       clickFn = spyOn(scope, 'someAction');
 
-      element.triggerHandler({type: 'keypress', keyCode: 32});
+      var divElement = elements.find('div');
+      var liElement = elements.find('li');
 
-      expect(clickFn).toHaveBeenCalled();
+      divElement.triggerHandler({type: 'keypress', keyCode: 32});
+      liElement.triggerHandler({type: 'keypress', keyCode: 32});
+
+      expect(clickFn).toHaveBeenCalledWith('div');
+      expect(clickFn).toHaveBeenCalledWith('li');
     });
 
     it('should not override existing ng-keypress', function() {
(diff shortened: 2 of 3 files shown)


src/ngAria/aria.js
```diff
@@ -83,7 +83,8 @@ function $AriaProvider() {
     ariaMultiline: true,
     ariaValue: true,
     tabindex: true,
-    bindKeypress: true
+    bindKeypress: true,
+    keypressEls: ['DIV', 'LI']
```
gkalpak: Rather trivial and probably a matter of personal style, but I would prefer a non- abbreviated name (`keypressElements`, `keypressNodeNames` etc).

src/ngAria/aria.js
```diff
@@ -303,11 +304,20 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
     compile: function(elem, attr) {
       var fn = $parse(attr.ngClick, /* interceptorFn */ null, /* expensiveChecks */ true);
       return function(scope, elem, attr) {
+
+        function isNodeOneOf(elem, nodeTypeArray) {
+          for (var i=0; i < nodeTypeArray.length; i++) {
```
gkalpak: Wouldn't the plain old `indexOf` work as well ?

<sub>Also, having spaces around `=` would make jscs happier I believe :)</sub>

test/ngAria/ariaSpec.js
```diff
@@ -496,6 +496,16 @@ describe('$aria', function() {
       expect(clickFn).toHaveBeenCalled();
     });
 
+    it('should a trigger click from the keyboard', function() {
```
gkalpak: Did you mean _trigger a click_ ?

test/ngAria/ariaSpec.js
```diff
@@ -526,6 +536,31 @@ describe('$aria', function() {
...
+    });
+  });
+
+  describe('actions for different configured elements', function() {
+    beforeEach(configAriaProvider({
+      keypressEls: ['DIV']
+    }));
+    beforeEach(injectScopeAndCompiler);
+
+    it('should not a trigger click from the keyboard', function() {
```
gkalpak: _trigger a click_ ?

src/ngAria/aria.js
marcysutton: Yes! Great idea.

src/ngAria/aria.js
marcysutton: I like `keypressNodeNames`. I've changed it to that. Thanks for the suggestion!

marcysutton:
@caitp this is ready for another review.

test/ngAria/ariaSpec.js
```diff
@@ -526,6 +538,31 @@ describe('$aria', function() {
       element.triggerHandler({ type: 'keypress', keyCode: 13 });
       expect(element.text()).toBe('keypress13');
     });
+
+    it('should not bind keypress to non-div elements', function() {
```
caitp: Might be good to note in the spec title that this is a symptom of using the default config. (This might be spelled out clearly higher up, can't tell on mobile view)

caitp:
Still LGTM

test/ngAria/ariaSpec.js
marcysutton: Good call. I've updated it.

src/ngAria/aria.js
```diff
@@ -83,7 +83,8 @@ function $AriaProvider() {
     ariaMultiline: true,
     ariaValue: true,
     tabindex: true,
-    bindKeypress: true
+    bindKeypress: true,
+    keypressNodeNames: ['DIV', 'LI']
```
caitp: I still kind of feel like this is a really weird API. Chances are, you're going to want emulated clicks for any element that doesn't produce them on their own, right? if this array gets out there and people start using it, we won't be able to get rid of it later. Lets make sure this is needed before doing it

marcysutton: Fair point. `<button>` is the biggest issue we had, since it gets both click and keypress. `<a>` should not be used with `ng-click` (that's what buttons are for) but rather `ng-href` and routing...although I suppose that could be abused. There's also the `<details>`/`<summary>` elements, `<dialog>`, form controls, and [a few more](http://www.w3.org/html/wg/drafts/html/master/interactive-elements.html). An interesting thing to consider, as well, is the work being done on Custom Elements and tabindex: https://docs.google.com/document/d/1k93Ez6yNSyWQDtGjdJJqTBPmljk9l2WS3JTe5OHHB50/edit#heading=h.tf9twftfr65h

So, we need some mechanism of checking whether or not to apply keypress–I'm personally in favor of opt-in for `<div>` and `<li>` so that it limits the scope of this behavior. It's just a question of whether we want that list to be configurable.

caitp: configurable is good, but I think we shouldn't expose an array for it at least.

If we decide we don't want it to be configurable and just "do the right thing always" later, it's trivial to replace configuration methods with `noop`, and not have anything unnecessary taking up space in the framework. So I see that as being preferable, tbh

caitp: Something like

```js
// enable keypress on all elements!
$ariaProvider.keypressClickElements('*');
```

or something like that.

marcysutton: Would that allow all elements by default? How would we discern between `<div>` and `<button>`?

marcysutton: @caitp revisiting this so we can get a fix in to the next release. I'm still unclear how we would fix the bug with a wildcard for all elements.

caitp: wildcard case would just not filter, or only blacklist known-issue elements, I guess.

Maybe it's not worth worrying about it and just land it, @lgalfaso what do you think?

marcysutton: What if we just do away with the config for elements and limit dynamic binding of keypress to `div` and `li`? We could just wait and see if anyone actually needs it to be configurable.

marcysutton:
@caitp I simplified the code to just Do the Right Thing&trade; without a configuration array. I also updated the docs.

caitp:
okay, i think a blacklist would be better than a whitelist, but this is fine for now... @petebacondarwin / @Narretz / @lgalfaso could you merge this before monday?

petebacondarwin:
Will do

marcysutton:
Thanks @petebacondarwin!

Ashutosh7376:
try this .. its 100% works for me ...

.config(function( $mdGestureProvider ) {
          $mdGestureProvider.skipClickHijack();
  })

source..
https://rubberchickin.com/issue-with-ng-click-on-ios-firing-twice/

marcysutton:
@Ashutosh7376 that is unrelated to this change, and only applicable to Angular Material.

PR #10770, opened 2015-01-16
fix($animate): ensure no transitions are applied when an empty inline style object is provided

### LINKED ISSUE #10613: animate.css animation within angularjs 1.3 using ngAnimate gives different animations in Firefox
There are differences between Firefox and Chrome/IE when animating.
The difference, in this case is that IE/Chrome do show an bounce effect when showing a message.

The source looks like this:

    <!DOCTYPE html>
    <html ng-app="MyApp" >
    
    <head>
      <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/animate.css/3.2.0/animate.min.css" />
      <script src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.3.7/angular.js"></script>
      <script src="https://code.angularjs.org/1.3.5/angular-animate.js"></script>
      <style>
        .welcome.ng-hide-remove {
          -webkit-animation: bounceInLeft 1s;
          -moz-animation: bounceInLeft 1s;
          -o-animation: bounceInLeft 1s;
          animation: bounceInLeft 1s;
        }
        .welcome.ng-hide-add {
          -webkit-animation: bounceOutRight 1s;
          -moz-animation: bounceOutRight 1s;
          -o-animation: bounceOutRight 1s;
          animation: bounceOutRight 1s;
        }
      </style>
      <script>
        (function() {
          var angularApp = angular.module("MyApp", ['ngAnimate']);
    
          var mainCtrl = function($scope, $timeout, $interval) {
            $scope.welcomeBool = false;
            $interval(function() {
              $scope.welcomeBool = !$scope.welcomeBool;
            }, 1500);
          };
          angularApp.controller("MainCtrl", 
             ["$scope", "$timeout", "$interval", mainCtrl]
          );
        }());
      </script>
    </head>
    
    <body ng-controller="MainCtrl">
      <h1 ng-show="welcomeBool" class="welcome">
         Hello 'non-bouncy' and fading Firefox animation!
      </h1>
    </body>
    
    </html>

This is the plunk: http://plnkr.co/edit/Srb5jY0LnBp25QJqSxFs?p=preview

Is seem like an ngAnimate bug...

I also posted it on StackOverflow, here: http://stackoverflow.com/questions/27714749/animate-css-animation-within-angularjs-1-3-using-nganimate-gives-different-anima

matsko:
Closes #10613

### DIFF
--- a/src/ngAnimate/animate.js
+++ b/src/ngAnimate/animate.js
@@ -1869,7 +1869,7 @@ angular.module('ngAnimate', ['ng'])
           return;
         }
 
-        if (!staggerTime && styles) {
+        if (!staggerTime && styles && Object.keys(styles).length > 0) {
           if (!timings.transitionDuration) {
             element.css('transition', timings.animationDuration + 's linear all');
             appliedStyles.push('transition');
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -1381,6 +1381,25 @@ describe("ngAnimate", function() {
             expect(element.attr('style')).toContain('border-color: blue');
           }));
 
+          it("should not apply a piggy-back-transition if the styles object contains no styles",
+            inject(function($compile, $animate, $rootScope, $sniffer) {
+
+            if (!$sniffer.animations) return;
+
+            $animate.enabled(true);
+            ss.addRule('.on', '-webkit-animation: 1s super-animation; animation: 1s super-animation;');
+
+            element = $compile(html('<div>1</div>'))($rootScope);
+
+            $animate.addClass(element, 'on', {
+              to: {}
+            });
+
+            $rootScope.$digest();
+            $animate.triggerReflow();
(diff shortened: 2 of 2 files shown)


src/ngAnimate/animate.js
```diff
@@ -1869,7 +1869,7 @@ angular.module('ngAnimate', ['ng'])
           return;
         }
 
-        if (!staggerTime && styles) {
+        if (!staggerTime && styles && Object.keys(styles).length > 0) {
```
petebacondarwin: Is it possible that styles could be a string? If so then calling `Object.keys` would throw an exception.

petebacondarwin: Oh I think it cannot since in `$animate.animate(element, from, to, ...)` `from` and `to` are both supposed to be objects.

petebacondarwin:
LGTM

PR #10772, opened 2015-01-16
Update jscs

hzoo:
Updates jscs to latest (1.10.0) by updating grunt-jscs to 1.2.0

add rule disallowKeywordsOnNewLine: "else" ~ 10 changes

``` js
// Valid
if (x) {
} else {
}
// Invalid
if (x) {
}
else {
}
```

add rule disallowSpacesInCallExpression ~ 19 changes

``` js
// Valid
functionCall()
// Invalid
functionCall ()
```

add rule requireSpacesInForStatement ~ 1 change

``` js
// Valid
for (var i=0; i < length; i++)
// Invalid
for (var i=0;i < length;i++)
for (var i=0;  i < length;  i++)
```

add rule requireSpaceBeforeKeywords ~ 0 changes

``` js
// Valid
} else {
// Invalid
}else {
```

Others that could be added:
~32 changes: "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
"requireSpaceBetweenArguments": true,
"disallowMultipleLineBreaks": true

@caitp

### DIFF
--- a/.jscs.json
+++ b/.jscs.json
@@ -20,6 +20,11 @@
     "beforeOpeningRoundBrace": true
   },
   "disallowSpacesInsideArrayBrackets": true,
+  "requireSpaceBeforeKeywords": [
+    "else",
+    "while",
+    "catch"
+  ],
   "disallowSpacesInsideParentheses": true,
   "disallowTrailingComma": true,
   "disallowTrailingWhitespace": true,
--- a/.jscs.json
+++ b/.jscs.json
@@ -12,6 +12,7 @@
   "disallowSpacesInAnonymousFunctionExpression": {
     "beforeOpeningRoundBrace": true
   },
+  "disallowSpacesInCallExpression": true,
   "disallowSpacesInFunctionDeclaration": {
     "beforeOpeningRoundBrace": true
   },
--- a/.jscs.json
+++ b/.jscs.json
@@ -35,6 +35,7 @@
     "afterConsequent": true,
     "beforeAlternate": true
   },
+  "requireSpacesInForStatement": true,
   "requireSpacesInFunction": {
     "beforeOpeningCurlyBrace": true
   },
--- a/.jscs.json
+++ b/.jscs.json
@@ -1,6 +1,7 @@
 {
   "excludeFiles": ["src/ngLocale/**"],
   "disallowKeywords": ["with"],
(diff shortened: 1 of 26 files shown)


caitp:
hmmm, not sure how i feel about making these rules even more strict. @petebacondarwin make the call on this one

petebacondarwin:
@hzoo - Thanks for the update.
This seems weird that updating broke a load of the code validation without changing the rules.
For instance my understanding was that `validateParameterSeparator : ", "` allowed any whitespace after the comma. Even the error message indicates that this is not working properly, when you have:

```
Missing space before function parameter '$q' at src/ngMock/angular-mocks.js :
   456 |angular.mock.$IntervalProvider = function() {
   457 |  this.$get = ['$browser', '$rootScope', '$q', '$$q',
   458 |       function($browser,   $rootScope,   $q,   $$q) {
-----------------------------------------------^
```

Can we look into the causes of this before we do this update?

hzoo:
Oh it got updated to only allow a single space. Actually I realize I added that change... https://github.com/jscs-dev/node-jscs/issues/701. Now the rule enforces the value in quotes so it has to be exactly `", "` which is strict. At the time I thought it was ok to enforce consistency since there isn't an align parameters style. (The error message wasn't updated either, so I'l have to change that as well).

There will need to be an option to specify > 1 space to allow the code to have aligned parameters if people want.

I can update to remove some rules - and we can temporarily remove the `validateParameterSeparator` until another option is available?

petebacondarwin:
>Also we can probably remove .jscs.json.todo now - not sure if we want any of those rules still.

> requireCurlyBraces is for single lines
> disallowImplicitTypeConversion prevents '' + variable but i'm sure most people like that.
> disallowMultipleLineBreaks - I think we have 2 line breaks for some tests?
> validateJSDoc already discussed in the other PR, depends.

Yes, I agree that we don't these rules

petebacondarwin:
OK, so I am going to merge this - but turn off the `validateParameterSeparator`

hzoo:
Alright that sounds good - should I make the changes and rebase or you got it? I would fixup the first commit and then remove the `jscs.json.todo`.

petebacondarwin:
I am doing it now. Thanks

PR #10778, opened 2015-01-16
Fix(ngRepeat) do not allow `$id` and `$root` as aliases

KELiON:
Currently user can use `$id` or `$root` as alias in ng-repeat directive that leads to rewriting these scope-related variables. This commit fixes this behavior by throwing an error when user try to use these values.

### DIFF
--- a/docs/content/error/ngRepeat/badident.ngdoc
+++ b/docs/content/error/ngRepeat/badident.ngdoc
@@ -16,6 +16,8 @@ Reserved names include:
   - `this`
   - `undefined`
   - `$parent`
+  - `$id`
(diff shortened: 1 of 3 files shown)


Narretz:
LGTM, but just to be sure, @caitp can you take a look at this?

caitp:
lgtm

Narretz:
Thanks! Should we cherry-pick this to 1.3 too?

pkozlowski-opensource:
@Narretz it was picked as d01cae2a0dd8bae3c8b8bd5792816d0c466355d1

PR #10786, opened 2015-01-17
docs(app): fixes material build

jesselpalmer:
The theme css files were reorganized at some point in Angular Material 0.7.x and the build is now failing because the files are missing. The styles on the site do not work correctly now with 0.7.x because the css is missing. This change will keep us on 0.6.x until we can update the theme issue.

### DIFF
--- a/docs/bower.json
+++ b/docs/bower.json
@@ -4,7 +4,7 @@
     "jquery": "2.1.1",
     "lunr.js": "0.4.3",
     "google-code-prettify": "1.0.1",
-    "angular-material": "master",
+    "angular-material": "0.6.x",
     "angular-sanitize": "1.3.x"
   }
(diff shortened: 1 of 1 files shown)


gkalpak:
Indeed, the theming system has been majorly refactored. There are no CSS files for the themes now; themes are built dynamically by JS.

(Speaking of changes, another major one (not related to theming) is that `md-text-float` has been deprecated in favor of the more powerful `md-input-container`.
Also, the way the `show/hide-{size}` attributes work has changed a bit (not sure if it actually affects as though).)

I'll try to take care of those issues some time next week (unless someone beats me to it :smile:).

jesselpalmer:
That clears things up. Thank you @gkalpak!

PR #10789, opened 2015-01-17
docs(angular.bootstrap): fns passed as modules are called on config stag...

thorn0:
...e


pkozlowski-opensource:
Indeed, there was a mistake! Thnx @thorn0 !

PR #10790, opened 2015-01-17
fix(ngController): allow bound constructor fns as controllers

### LINKED ISSUE #10784: Bound constructor functions can't be used as controllers
```js
var app = angular.module('nojs', []);

app.controller('nojs', [controller.bind(null)]); // not working
app.controller('nojs', [controller]); // working

function controller(){};
```

pkozlowski-opensource:
Fixes #10784

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -829,7 +829,7 @@ function createInjector(modulesToLoad, strictDi) {
       // Check if Type is annotated and use just the given function at n-1 as parameter
       // e.g. someModule.factory('greeter', ['$window', function(renamed$window) {}]);
       // Object creation: http://jsperf.com/create-constructor/2
-      var instance = Object.create((isArray(Type) ? Type[Type.length - 1] : Type).prototype);
+      var instance = Object.create((isArray(Type) ? Type[Type.length - 1] : Type).prototype || null);
       var returnedValue = invoke(Type, instance, locals, serviceName);
 
       return isObject(returnedValue) || isFunction(returnedValue) ? returnedValue : instance;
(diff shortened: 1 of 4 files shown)


caitp:
hmm, it's really too bad we don't have `Reflect.IsConstructor()` yet, that would make this a lot easier.

caitp:
actually you know what, I think this is still a kind of dumb idea, since the bound function won't have the expected prototype. It would be like giving people a footgun to shoot themselves with.

I apologize for changing my mind about this twice today, but unfortunately we have no way of getting at the "real" prototype from JS :(

pkozlowski-opensource:
@caitp it is OK to change ones mind if in the end it leads to better framework! I understand your position and I don't have any more arguments apart from the ones presented already:
* bound constructor functions are special animals that are half-constructors (you can `new` them but those won't have a prototype) - so it is hard to arbitrary say if it is a proper constructor or not - we are in the grey are here...
* people using bound constructor functions should be aware of the fact that prototypes are not available - this is how it works in vanilla JS
* I'm not sure how common it is to use both prototypal inheritance for controllers
* it used to work in 1.2.x
* it is a trivial fix

But as I've stated before I don't feel strongly about this one as I don't need it personally. I also understand your will to "protect the innocent" and don't give people tools that they can use to hurt themselves. 

To sum up: I've got nothing more to say / do when it comes to this issue so if you @caitp or @petebacondarwin got a definitive opinion on this one please close / merge this PR / associated issue.

petebacondarwin:
We should land this, since it was available throughout for 1.2 and no one managed to shoot themselves in the foot so far; so it is probably a very unlikely thing to happen. In the meantime people do want to be able to create bound controllers in their applications and the use cases provided are reasonable.

We should put a disclaimer in the documentation for `ngController` and `module.controller` to highlight that bound controllers will not have the expected prototype.

Can you do that and then merge @pkozlowski-opensource ?

thinkingmedia:
-1 for this change. I can not think of one case where a programmer would need to do this. It's more likely they're doing something they shouldn't. I would hate to see this introduce modules that become popular but are full of odd controllers.

PR #10793, opened 2015-01-18
docs(ngClass): fix formatting

espiler:
adds line break between add/remove in animations. Previously showing up on one line like this: "add - happens just before the class is applied to the element remove - happens just before the class is removed from the element"


espiler:
CLA is signed. Thanks! -Evan

PR #10794, opened 2015-01-18
docs($cookieStore#get): Clarified that undefined is returned if no cookie exists

robinwassen:
Simply clarified that `undefined` is the return value if no cookie exists.


robinwassen:
Signed the CLA.

PR #8584, opened 2014-08-12
Allowing angular-touch to work with jQuery.

mcmar:
Angular includes a small subset of jQuery out of the box called jqLite. Normally, angular wraps all the elements it passes to its directives in jqLite. When jQuery is loaded ahead of angular, angular instead wraps its elements in the full implementation of jQuery. Additionally, angular wraps its events in jQuery. These jQuery events work throughout most of angular. Angular-touch currently was not working with those jQuery-wrapped events. Touches were registering inconsistently and rarely when using the ngClick directive. This commit should fix that. I took it from another person I found on github. I'm not sure why his changes were never merged into the main angular.js repo, but I've verified that it works and I decided to submit a pull request for it. There is a bug/issue thread I found relating to this. It is linked below.
https://github.com/angular/angular.js/issues/4001

PR #10797, opened 2015-01-18
Fix/jquery touch

pomerantsev:
This is an addition to PR #8584. The discussion is also there.

### DIFF
--- a/src/ngTouch/directive/ngClick.js
+++ b/src/ngTouch/directive/ngClick.js
@@ -123,6 +123,8 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
       return; // Too old.
     }
 
+    // Use JQuery originalEvent
+    event = event.originalEvent || event;
     var touches = event.touches && event.touches.length ? event.touches : [event];
     var x = touches[0].clientX;
     var y = touches[0].clientY;
@@ -165,6 +167,8 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
   // Global touchstart handler that creates an allowable region for a click event.
   // This allowable region can be removed by preventGhostClick if we want to bust it.
   function onTouchStart(event) {
+    // Use JQuery originalEvent
+    event = event.originalEvent || event;
     var touches = event.touches && event.touches.length ? event.touches : [event];
     var x = touches[0].clientX;
     var y = touches[0].clientY;
@@ -210,6 +214,8 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
     }
 
     element.on('touchstart', function(event) {
+      // Use JQuery originalEvent
+      event = event.originalEvent || event;
       tapping = true;
       tapElement = event.target ? event.target : event.srcElement; // IE uses srcElement.
       // Hack for Safari, which can target text nodes instead of containers.
@@ -237,7 +243,9 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
 
     element.on('touchend', function(event) {
       var diff = Date.now() - startTime;
-
+      
+      // Use JQuery originalEvent
+      event = event.originalEvent || event;
       var touches = (event.changedTouches && event.changedTouches.length) ? event.changedTouches :
           ((event.touches && event.touches.length) ? event.touches : [event]);
       var e = touches[0].originalEvent || touches[0];
--- a/src/ngTouch/directive/ngClick.js
+++ b/src/ngTouch/directive/ngClick.js
@@ -283,6 +283,9 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
     // - But the browser's follow-up slow click will be "busted" before it reaches this handler.
     // Therefore it's safe to use this directive on both mobile and desktop.
     element.on('click', function(event, touchend) {
+      // Use JQuery originalEvent
+      event = event.originalEvent || event;
+
       scope.$apply(function() {
         clickHandler(scope, {$event: (touchend || event)});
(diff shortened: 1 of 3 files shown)


lgalfaso:
the patch looks fine, but please
* it looks like the new tests only check 2 of the 3 changes
* rebase this into a single commit
* follow the commit message conventions

@mzgol can you double check that everything is fine?

test/ngTouch/directive/ngClickSpec.js
```diff
@@ -49,6 +49,16 @@ describe('ngClick (touch)', function() {
   }));
 
 
+  it('should unwrap a jQuery-wrapped event object', inject(function($rootScope, $compile) {
+    element = $compile('<div ng-click="event = $event"></div>')($rootScope);
+    $rootScope.$digest();
+
+    browserTrigger(element, 'touchstart');
+    browserTrigger(element, 'touchend');
+    expect($rootScope.event.originalEvent).toBeUndefined();
```
mzgol: I'd rather you check here for `$rootScope.event.{pageX,pageY}` as this is what's actually important here. This should still fail without the patch.

test/ngTouch/directive/ngClickSpec.js
```diff
@@ -49,6 +49,16 @@ describe('ngClick (touch)', function() {
   }));
 
 
+  it('should unwrap a jQuery-wrapped event object', inject(function($rootScope, $compile) {
+    element = $compile('<div ng-click="event = $event"></div>')($rootScope);
+    $rootScope.$digest();
+
+    browserTrigger(element, 'touchstart');
```
mzgol: You should check if `$rootScope.event` is correct after this line as well; otherwise the `touchstart` part of the patch is not tested.

test/ngTouch/directive/ngClickSpec.js
mzgol: BTW, this test will pass even if only `touchend` is patched. It'd be good to test them both.

mzgol: > I'd rather you check here for $rootScope.event.{clientX,clientY} as this is what's actually important here.

This would be a false positive on `browserTrigger(element, 'click')`, though (which you should test as well as it's patched, too).

mzgol: @lgalfaso We should most likely just create a Protractor test for such stuff, too much hacking must happen here to test anything.

pomerantsev:
@mzgol - sorry, I'm not following.
> This would be a false positive on browserTrigger(element, 'click'), though (which you should test as well as it's patched, too).
What exactly should I test in `browserTrigger`? You're absolutely right about the false positive - `clientX` and `clientY` are always there on the `event` object in tests.
> BTW, this test will pass even if only `touchend` is patched. It'd be good to test them both.
Right, I couldn't find a way to unit-test the rest.

I'll be happy to write a protractor test for this change. Should I put it in test/e2e/tests?

mzgol:
> I'll be happy to write a protractor test for this change. Should I put it in test/e2e/tests?

I wrote it too fast. We run Protractor tests on Chrome, Firefox & Safari so touch events won't work there, we probably need to stay with just unit tests for now.

>> This would be a false positive on browserTrigger(element, 'click'), though (which you should test as well as it's patched, too).
>
> What exactly should I test in browserTrigger? You're absolutely right about the false positive - clientX and clientY are always there on the event object in tests.

I've applied your test on `master` and just changed it to check for `clientX`/`clientY` and the test failed as expected in jQuery tests (it didn't fail on jqLite as expected).

How about you add those two `clientX`/`clientY` checks without removing your `originalEvent` one and create a separate test for the `click` event where you only check for the missing `originalEvent` property?

pomerantsev:
I've updated tests and I actually had to include the same transformation to the `click` handler.
The failing test is most probably a flake - some WebDriver stuff.
I will squash / rebase commits if needed, but I feel slightly bad about squashing someone else's commits with mine.

petebacondarwin:
@lgalfaso and @mzgol can you decide who should own this and then see if we can get it into 1.4.x?

mzgol:
@petebacondarwin I can take it.

petebacondarwin:
Magic! Thanks @mzgol

pomerantsev:
@mzgol - like I said, I'll be happy to help to get it merged.

src/ngTouch/directive/ngClick.js
```diff
@@ -238,6 +244,8 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
     element.on('touchend', function(event) {
       var diff = Date.now() - startTime;
 
+      // Use JQuery originalEvent
+      event = event.originalEvent || event;
       var touches = (event.changedTouches && event.changedTouches.length) ? event.changedTouches :
           ((event.touches && event.touches.length) ? event.touches : [event]);
       var e = touches[0].originalEvent || touches[0];
```
LeonardoBraga: Just a suggestion: as far as I know, there's no "originalEvent" in `touches[0]` - only in the event itself. This would allow you to reduce these two variables to:

```javascript
var e = (event.changedTouches && event.changedTouches.length) ? event.changedTouches[0] :
    ((event.touches && event.touches.length) ? event.touches[0] : event;
```

pomerantsev: You're right - thanks for noticing it. I guess I'd just like to get it merged sooner rather than later, and then we can clean it up. Let's see what @mzgol says.

mzgol:
I went in another direction as just assigning `$event.originalEvent` to `$event` breaks the general assumption that `$event` is a jQuery/jqLite event object, not a native one. Please see #11471.

PR #3516, opened 2013-08-09
feat(ngSwitch): Allow multiple case matches for one element

lrlopez:
Now you can also use an array in `ng-switch-when`, which allows specifying multiple values for the same element. Values will be evaluated within the scope when using the array notation.

I.e. `ng-switch-when="[1, 3, 'hello', 5-3]"`

Each element will switch only once even if there are multiple matches in the same element.

**BREAKING CHANGE:** Previously `ng-switch-when` expressions that matched against square brackets strings won't work anymore. The fix is encapsulate the string into a one-element array.
I.e. before `ng-switch-when='[hi]'`, now `ng-switch-when="['[hi]']"`

Closes #3410

PR #10798, opened 2015-01-18
feat(ngSwitch): add an optional attribute ngSwitchWhenSeparator

### LINKED ISSUE #3410: make ng-switch-when more powerful
Programming languages with switch statements allow associating several cases with the same block, e.g.:
```javascript
switch (value) {
case 0:
  foo();
  break;
case 1: // fall through
case 2: // fall through
case 3:
  bar();
  break;
}
```

It would be useful if ng-switch-when similarly allowed matching on multiple values, e.g.
```html
<div ng-switch-on="value">
  <div ng-switch-when="foo">foo</div>
  <div ng-switch-when="/bar|baz/">bar or baz</div>
</div>
```

lgalfaso:
Adds an optional attribute `ngSwitchWhenSeparator` that allows multiple tokens to match a given `ngSwitchWhen`.

Closes #3410
Closes #3516

### DIFF
--- a/src/ng/directive/ngSwitch.js
+++ b/src/ng/directive/ngSwitch.js
@@ -48,7 +48,11 @@
  *
  * * `ngSwitchWhen`: the case statement to match against. If match then this
  *   case will be displayed. If the same match appears multiple times, all the
- *   elements will be displayed.
+ *   elements will be displayed. It is possible to associate mutiple values to
+ *   the same `ngSwitchWhen` by defining the optional attribute
+ *   `ngSwitchWhenSeparator`. The separator will be used to split the value of
+ *   the `ngSwitchWhen` attribute into multiple tokens, and the element will show
+ *   if any of the `ngSwitch` evaluates to any of these tokens.
  * * `ngSwitchDefault`: the default case when no other case match. If there
  *   are multiple default cases, all of them will be displayed when no other
  *   case match.
@@ -189,8 +193,13 @@ var ngSwitchWhenDirective = ngDirective({
   require: '^ngSwitch',
   multiElement: true,
   link: function(scope, element, attrs, ctrl, $transclude) {
-    ctrl.cases['!' + attrs.ngSwitchWhen] = (ctrl.cases['!' + attrs.ngSwitchWhen] || []);
-    ctrl.cases['!' + attrs.ngSwitchWhen].push({ transclude: $transclude, element: element });
+    var cases = attrs.ngSwitchWhen.split(attrs.ngSwitchWhenSeparator).sort().filter(
+      function(el, ix, ar) { return ar[ix - 1] !== el;}
+    );
+    forEach(cases, function(whenCase) {
+      ctrl.cases['!' + whenCase] = (ctrl.cases['!' + whenCase] || []);
+      ctrl.cases['!' + whenCase].push({ transclude: $transclude, element: element });
+    });
   }
 });
 
--- a/test/ng/directive/ngSwitchSpec.js
+++ b/test/ng/directive/ngSwitchSpec.js
@@ -299,6 +299,113 @@ describe('ngSwitch', function() {
     $rootScope.$apply('mode = "b"');
     expect(element.children().length).toBe(1);
   }));
+
+
+  describe('ngSwitchWhen separator', function() {
+    it('should be possible to define a separator', inject(function($rootScope, $compile) {
+      element = $compile(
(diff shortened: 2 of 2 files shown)


Narretz:
Does the separator need to be customized? Why not settle on comma?

test/ng/directive/ngSwitchSpec.js
```diff
@@ -299,6 +299,113 @@ describe('ngSwitch', function() {
...
+      $rootScope.$apply('mode = ""');
+      expect(element.children().length).toBe(1);
+      expect(element.text()).toBe('Block1|');
+      $rootScope.$apply('mode = "c"');
+      expect(element.children().length).toBe(1);
+      expect(element.text()).toBe('Block3|');
+    }));
+
+
+    it('should be possible to use the empty string as a separators', inject(function($rootScope, $compile) {
```
gkalpak: Typo: _separator_

gkalpak:
> Does the separator need to be customizable? Why not settle on comma?

@Narretz: I guess it is to enable using `,` in the value (not as a separator).

--
@lgalfaso: This might be a little farfetched, but it would be nice to allow setting the "default" separator on `ng-switch` (instead of having to use it on every `ng-switch-when`). E.g.:

```html
<ng-switch ng-switch-separator="|">
    <ng-switch-when="a|b">A or B</ng-switch-when>
    <ng-switch-when="c|d">C or D</ng-switch-when>
    <ng-switch-when="|+$" ng-switch-separator="+">
        | or $ (local separator overwrites default)
    </ng-switch-when>
</ng-switch>
```

test/ng/directive/ngSwitchSpec.js
lgalfaso: thx

lgalfaso:
> Does the separator need to be customizable? Why not settle on comma?

this would be a breaking change (that I am trying to avoid)

> This might be a little farfetched, but it would be nice to allow setting the "default" separator

if there is some interest in this PR, then we can think about adding a default separator... doing so would be relatively easy

bricss:
Very needed feature, please merge it down.

petebacondarwin:
What if we made `ngSwitchWhenSeparator` a directive of its own that `ngSwitch` could "require". This way you could specify it as an ancestor and it would affect all descendants (similar to how `ngModelOptions` works)

Narretz:
This would be pretty cool! A global configuration is probably not a good idea, because you cannot control what templates other modules might use.

lgalfaso:
@petebacondarwin making it another directive could give some unexpected results. Eg.
```html
<div ng-switch="mode">
  <p ng-switch-when="a|b" ng-switch-when-separator="|"><span other-directive/></p>
  <p ng-switch-default>X</div>
</div>
```

if `other-directive` has a `ngSwitch`, then it could end up using an unexpected `ngSwitchWhenSeparator`.

petebacondarwin:
@lgalfaso OK, let's leave it as it is. Shame that we traverse up past isolated scope boundaries when doing require :-)

src/ng/directive/ngSwitch.js
```diff
@@ -189,8 +193,13 @@ var ngSwitchWhenDirective = ngDirective({
   require: '^ngSwitch',
   multiElement: true,
   link: function(scope, element, attrs, ctrl, $transclude) {
-    ctrl.cases['!' + attrs.ngSwitchWhen] = (ctrl.cases['!' + attrs.ngSwitchWhen] || []);
-    ctrl.cases['!' + attrs.ngSwitchWhen].push({ transclude: $transclude, element: element });
+    var cases = attrs.ngSwitchWhen.split(attrs.ngSwitchWhenSeparator).sort().filter(
+      function(el, ix, ar) { return ar[ix - 1] !== el;}
```
gkalpak: What is the purpose of this?

gkalpak:
I still believe that moving `ngSwitchWhenSeparator` to `ngSwitch` (https://github.com/angular/angular.js/pull/10798#issuecomment-70651897) is better,but it LGTM.

Narretz:
@gkalpak Now  that you say it, I think it makes sense. Although I guess that custom separators will be used quite rarely anyway (only when the default operator is a value).

gkalpak:
What default operator?

Narretz:
Yeah, there's no default value of course. Merged to both master and 1.5.x. We can put this attribute on the ngSwitch later on.

PR #10809, opened 2015-01-19
docs(guide/Controllers): describe your change...

ghost:
dont' show when num is NaN


Narretz:
This is merely an example of how to define a function on the scope and how to call this function from the template. Your change adds 3 different directives to the template without explaining why.
This is essentially making the whole example more confusing -  I'll have to close this PR.

PR #10810, opened 2015-01-20
docs($routeProvider): describe your change...

prashantma:
Purpose of "templateUrl" option (as a string value) within "param" parameter when there also exists "path" parameter "in the API": 

> > > "path" allows scenarios where same route URL is used with different "controller"s
> > > "templateURL" allows scenarios where route URL is different for same "controller".

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -139,6 +139,11 @@ function $RouteProvider() {
    *
    * @description
    * Adds a new route definition to the `$route` service.
+   * 
+   * Note:
+   * Purpose of "templateURL" option (as a string value) within "param" parameter when there also exists "path" parameter "in the API": 
+   * >>> "path" allows scenarios where same route URL is used with different "controller"s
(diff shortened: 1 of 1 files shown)


src/ngRoute/route.js
```diff
@@ -139,6 +139,11 @@ function $RouteProvider() {
    *
    * @description
    * Adds a new route definition to the `$route` service.
+   * 
+   * Note:
+   * Purpose of `templateUrl` option (as a string value) within `param` parameter when there also exists `path` parameter: 
+   *    - `path` allows scenarios where same route URL is used with different controllers
+   *    - `templateUrl` allows scenarios where route URL is different for same controller.
```
gkalpak: I don't know if it's me, but I don't understand what you mean...

Narretz: Me neither. :|

prashant-gcm:
Closing. Now I understand the difference in "path" and "templateUrl".

PR #10818, opened 2015-01-20
Adds missing $setUntouched method to nullFormCtrl.

firletka:


firletka:
I'm contributing as a Google employee.

PR #10820, opened 2015-01-20

Narretz:
When a parser has the same name as a validator key, and a parse error
occurs, set validity for parser name after setting validity for the
validator key.

Otherwise, the parse key is set, and then immediately afterwards the
validator key is unset (because parse errors remove all other
validations).


Narretz:
@shahata can you take a look at this?

Narretz:
Actually, this doesn't completely fix this. Closing for now.

PR #10823, opened 2015-01-21
fix($location): Allow urlencoded slashes to remain encoded

kaitrin:
Encoded slashes passed to $location.url() may now stay encoded if
$locationProvider.leaveSlashesEncoded(true) is called in the config
block. Other encoded entities continue to be decoded as before,
regardless of leaveSlashesEncoded value. Previously all entities,
including slashes, were decoded, which caused a problem when trying
to capture route params, which are slash-delimited.

Closes #10479

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -1,9 +1,10 @@
 'use strict';
 
 var PATH_MATCH = /^([^\?#]*)(\?([^#]*))?(#(.*))?$/,
-    DEFAULT_PORTS = {'http': 80, 'https': 443, 'ftp': 21};
+    DEFAULT_PORTS = {'http': 80, 'https': 443, 'ftp': 21},
+    ENCODED_SLASH = '%2F',
+    DOUBLE_ENCODED_SLASH = '%252F';
 var $locationMinErr = minErr('$location');
-
 
 /**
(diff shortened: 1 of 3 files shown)


kaitrin:
I authored this commit and have added my author email address as a secondary email to this account as instructed. I'm uncertain why the googlebot isn't picking up the secondary email.

pkozlowski-opensource:
@kaitrin I need to spend some more time on the underlying issue as I think that it might be potentially solved by using * params in a route so I can't guarantee this one will land (but at least your PR seems to be clean and has tests so it definitively helps!).

In the meantime, could you please sign CLA so we can merge it if this one turns out to be good to go? Thnx!

kaitrin:
Thanks for checking it out! I have signed the CLA, but googlebot is still flagging me as not being the sole author (I am), since my gmail address is not the same as my non-gmail committer email address. After it flagged me the first time, I added my committer email address as a secondary email address. It didn't seem to find it this time, either, so I have resubmitted the CLA now that kate@kaitrin.me has been added as a secondary. Hopefully that will resolve the CLA/authors issue.

lgalfaso:
@btford can you comment on this? Whatever we do, it needs to be handled by the new router.

Issue #10641, opened 2015-01-05
When we click Ctrl+"Edit in Plunker" button, the Plunker is not opening in new tab.


rajeshwarpatlolla:
When we click Ctrl+"Edit in Plunker" button, the plunker is not opening in new tab, instead it is  opening the plunker in the same tab. But when we click Ctrl+edit in plunker button, it should open in the new tab, which is not happening as of now. And because of it we are unable to stay in the same page to refer the docs.

arpit2438735:
It's very specific to Browser.It's not a issue with Angular website,can you retest the same scenario with different site probably [Stackoverflow](http://stackoverflow.com)

rajeshwarpatlolla:
When i click Ctrl+"Edit in Plunker" in  "angular-ui.github.io/bootstrap/", the plunker is opening in a new tab. But this is not happenning in "angularjs.org". In "angularjs.org", when i do the same the plunker is opening in the same tab.

arpit2438735:
Ok that fine but this forum is not for website bug it is specific to angular.js code.You can post the bug to https://github.com/angular/angularjs.org

pkozlowski-opensource:
Closing as duplicate of https://github.com/angular/angularjs.org/issues/149

jeffbcross:
Reopening, as the code for this is actually in angular.js, not angularjs.org: https://github.com/angular/angular.js/blob/master/docs/config/templates/runnableExample.template.html

ghoti143:
Looks like the plunkr form used to open in a new window, but due to popup blockers getting in the way the behavior was changed to it's current state.  https://github.com/angular/angular.js/commit/925b2080a0341d9348feeb4f492957a2e2c80082

PR #10826, opened 2015-01-21
feat(openPlunkr): enable ctrl+click

ghoti143:
This is an initial suggestion for dealing with the problem described in #10641.  To summarize, the problem is:
- The "Edit in Plunkr" anchor tag has an ng-click attribute
- The `openPlunkr` function is fired on click which calls `formPostData`
- `formPostData` creates a new form element, adds it to the page, posts it, and removes the form element
- The form, when posted, opens in the same page (`target="_self"`)
- This can be unhelpful for people who want to stay on the AngularJS documentation site.
- Changing the target attribute to `"_blank"`, however, can trigger the popup blocker.
- If a user chose to bypass pop-up blocker one time and click the link, they would arrive at a new default plnkr, not a plnkr with the desired template.

To resolve, the following steps were taken:
- Pass the `$event` object into the `openPlunkr` function
- Check if the ctrl button was pressed during click
  - If true, set `target="_blank"`
  - If false, set `target="_self"`

This enables a user to "opt-in" to the new window functionality, and the assumption is that they will whitelist the AngularJS docs page so their popup blocker doesn't block the popup.  However, for the users who do not ctrl+click, they will continue experiencing the default functionality and not experience the undesired behavior described above.

### DIFF
--- a/docs/app/src/examples.js
+++ b/docs/app/src/examples.js
@@ -1,13 +1,16 @@
 angular.module('examples', [])
 
 .factory('formPostData', ['$document', function($document) {
-  return function(url, fields) {
+  return function(url, newWindow, fields) {
     /**
-     * Form previously posted to target="_blank", but pop-up blockers were causing this to not work.
-     * If a user chose to bypass pop-up blocker one time and click the link, they would arrive at
-     * a new default plnkr, not a plnkr with the desired template.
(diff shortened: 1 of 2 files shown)


caitp:
it looks good to me, tbh. *tries it*

caitp:
added one more commit to it to work better on macs, since by default ctrl+click is right-click

PR #10827, opened 2015-01-21
docs(ngIf): Fixed typo

MetalMichael:
"falsy" -> "false"

### DIFF
--- a/src/ng/directive/ngIf.js
+++ b/src/ng/directive/ngIf.js
@@ -39,7 +39,7 @@
  * @element ANY
  * @scope
  * @priority 600
- * @param {expression} ngIf If the {@link guide/expression expression} is falsy then
+ * @param {expression} ngIf If the {@link guide/expression expression} is false then
  *     the element is removed from the DOM tree. If it is truthy a copy of the compiled
  *     element is added to the DOM tree.
  *


MetalMichael:
Submitted CLA

src/ng/directive/ngIf.js
```diff
@@ -39,7 +39,7 @@
  * @element ANY
  * @scope
  * @priority 600
- * @param {expression} ngIf If the {@link guide/expression expression} is falsy then
+ * @param {expression} ngIf If the {@link guide/expression expression} is false then
```
caitp: it's not really a typo -- see https://developer.mozilla.org/en-US/docs/Glossary/Falsy

pkozlowski-opensource:
Yup, @caitp is right! We had similar PRs in the past so maybe we should link the `falsy` word to https://developer.mozilla.org/en-US/docs/Glossary/Falsy so people are less likely to be confused by it?

pkozlowski-opensource:
Bah, we've got 7 references to the `falsy` word in docs, not sure if it makes sense to link them all to the definition...

Going to close this one as the docs is correct. If anyone is up to sending PR to link those words to a definition that PRs welcomed!

PR #10832, opened 2015-01-22
link to Google CDN is wrong

Starojitski:
it points to https://developers.google.com/speed/libraries/devguide#angularjs

instead of http://cdnjs.com/libraries/angular.js/

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -1,4 +1,10 @@
 'use strict';
+// I have no idea how to change the common header that 
+// tells people that they can download some module from different places
+// that's why i do this comment here . maybe you can figure it out
+// the link to Google CDN is wrong 
+// it points to https://developers.google.com/speed/libraries/devguide#angularjs
+// instead of http://cdnjs.com/libraries/angular.js/ 
 
 var $sanitizeMinErr = angular.$$minErr('$sanitize');
 


Narretz:
I have actually no idea what this is about.

Starojitski:
if you go to https://docs.angularjs.org/api/ngSanitize 

image

Now try to click on the 'Google CDN' link. You don't end up at Google CDN but at https://developers.google.com/speed/libraries/devguide#angularjs

caitp:
I mean, the cdn link shows you how to load angular from cdn, and the text underneath shows you how to load angular-sanitize. is this confusing? can't really link to the cdn itself since there's no directory listing, afaik

Narretz:
I agree ith @caitp , I don't think there's anything actionable about this. Google ajax.googleapis does not support directory listings, and the link points exactly to where you can find, plus, the code block gives you the right address.

Starojitski:
You are right. I actually took time to read what is written and everything is kind of correct. 

On one hand i feel stupid for not reading everything the first time and on the the other i find the way it's written is stupid. 

When i read ```You can download this file from the following places:``` i immediately scanned all three options and selected the first one. My eyes did not even register the line ```where X.Y.Z is the AngularJS version you are running.``` It is essentially hidden below 9 lines of different sizes and colors.

So the first thing i did was click on "Google CDN" and there i saw only link to angular itself...

PR #10839, opened 2015-01-22
fix(ngPluralize): cap value - offset to 0

### LINKED ISSUE #10836: ngPluralize example has problem when a value >3 is deleted
In this page
https://code.angularjs.org/1.3.10/docs/api/ng/directive/ngPluralize

Change the number of peoples to 3 or more and then clear the field.

typing 3 and cleaning get -2 other peoples...

caitp:
BREAKING CHANGE:

ngPluralize won't produce messages with signed (negative) values anymore.

Closes #10836

---

maybe this should be optional, but idk


caitp:
blehhh this doesn't work, and looking at the test failures i'm not sure how it worked before, either.

PR #10841, opened 2015-01-22
fix(ngPluralize): fix wrong text content when count is null/undefined

### LINKED ISSUE #10836: ngPluralize example has problem when a value >3 is deleted

gkalpak:
When `lastCount` was evaluated to an non-numeric value (e.g. "other") and
`count` was evaluated to `NaN` (e.g. `null`/`undefined`), the text content
would be (wrongly) based on the previous template.
This commits makes sure the text content is updated correctly.

In order to customize the message shown upon `null`/`undefined` one can
specify a `'NaN'` property on the `when` exression object.

Closes #10836

### DIFF
--- a/src/ng/directive/ngPluralize.js
+++ b/src/ng/directive/ngPluralize.js
@@ -217,11 +217,13 @@ var ngPluralizeDirective = ['$locale', '$interpolate', '$log', function($locale,
 
         // If both `count` and `lastCount` are NaN, we don't need to re-register a watch.
         // In JS `NaN !== NaN`, so we have to exlicitly check.
-        if ((count !== lastCount) && !(countIsNaN && isNaN(lastCount))) {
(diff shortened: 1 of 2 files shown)


gkalpak:
With this PR, the warning about no matching when expression is only logged if the new value is not `null`/`undefined`. I thought logging even for `null`/`undefined` would produce too many (possibly undesirable) warnings.
But I don't feel to strongly about it though.

/ping @lgalfaso <sub>(because if I recall correctly you were "mentoring" the relevant PRs)</sub>

petebacondarwin:
LGTM

PR #10843, opened 2015-01-23
docs(tutorial/4 - Two-way Data Binding): Variable name typo

dnicolson:
This appears to be a typo in the tutorial as the protractor build fails when using the `{{phone.name}}` variable name.

Additionally, performing a `git checkout -f step-4` results in the `scenario.js` file using the `phone.name` variable name.


petebacondarwin:
Yes I think you are right. We updated protractor but didn't update the docs.

petebacondarwin:
This was already updated in October: https://github.com/angular/angular.js/commit/43b1a3739a3cb36bf9731d3bbd444c13f0480d42

PR #10849, opened 2015-01-23
fix(htmlAnchorDirective): remove event listener if target is not element

### LINKED ISSUE #4262: Broken anchor link when directive replaces its markup
A simple directive that replaces markup with an anchor surrounded by a div. This is broken when the markup being replaced is an anchor itself.

There are 2 workarounds:

* don't use an anchor tag to be replaced
* don't surround the replacing anchor with a div

See here for a reproduction of the issue: http://plnkr.co/edit/sYqXhe3gcBicmR1tSECN

caitp:
Previously, when an `a` tag element used a directive with a replacing template, and did not include an `href` or `name` attribute before linkage, the anchor directive would always prevent default.

Now, the anchor directive will cancel its event listener if the linked element is not the same as the target element of the event.

Closes #4262

### DIFF
--- a/src/ng/directive/a.js
+++ b/src/ng/directive/a.js
@@ -18,10 +18,16 @@ var htmlAnchorDirective = valueFn({
   compile: function(element, attr) {
     if (!attr.href && !attr.xlinkHref && !attr.name) {
       return function(scope, element) {
+        // If the linked element is not an anchor tag anymore, do nothing
+        if (element[0].nodeName.toLowerCase() !== 'a') return;
+
         // SVGAElement does not use the href attribute, but rather the 'xlinkHref' attribute.
         var href = toString.call(element.prop('href')) === '[object SVGAnimatedString]' ?
                    'xlink:href' : 'href';
         element.on('click', function(event) {
+          // If a different element was clicked, ignore it.
+          if (element[0] !== event.target) return;
+
           // if we have no href url, then don't navigate anywhere.
           if (!element.attr(href)) {
             event.preventDefault();
--- a/test/ng/directive/aSpec.js
+++ b/test/ng/directive/aSpec.js
@@ -3,6 +3,25 @@
 describe('a', function() {
   var element, $compile, $rootScope;
 
+  beforeEach(module(function($compileProvider) {
+    $compileProvider.
+      directive('linkTo', valueFn({
+        restrict: 'A',
+        template: '<div class="my-link"><a href="{{destination}}">{{destination}}</a></div>',
+        replace: true,
+        scope: {
+          destination: '@linkTo'
+        }
+      })).
+      directive('linkNot', valueFn({
+        restrict: 'A',
+        template: '<div class="my-link"><a href>{{destination}}</a></div>',
+        replace: true,
+        scope: {
+          destination: '@linkNot'
+        }
+      }));
+  }));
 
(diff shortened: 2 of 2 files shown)


caitp:
self-reviewed, minimal perf impact, doesn't seem to introduce any new issues

b1rdex:
This causes regression: previously `<a href="" ng-click="doSomething()">test</a>` just called `doSomething()` and click was prevented.
Now if I click such link, `doSomething()` is called, but `click` event is not prevented for `<a>`.

caitp:
please produce a reproduction

b1rdex:
Can't include all my project code, but here is link example.

```html
<div class="actions">
        <a href="" class="remover" ng-click="removeAttachment(attachment)"><span class="glyphicon glyphicon-remove"></span></a>
      </div>
```

```coffeescript
$scope.removeAttachment = (attachment) ->
      log 'remove called'
      index = $scope.attachments.indexOf(attachment)
      link = $scope.attachments[index]
      $scope.attachments.splice index, 1
      link.canceler.resolve "aborted"

      $http.get(Data.storage.delete + link.id);

      link = null
      log 'remove completed'

      return false
```

caitp:
so your issue is that the target element of the click is not the link itself, it's the span/icon. i'll try to produce a fix for that.

src/ng/directive/a.js
```diff
@@ -18,10 +18,16 @@ var htmlAnchorDirective = valueFn({
...
       return function(scope, element) {
+        // If the linked element is not an anchor tag anymore, do nothing
+        if (element[0].nodeName.toLowerCase() !== 'a') return;
+
         // SVGAElement does not use the href attribute, but rather the 'xlinkHref' attribute.
         var href = toString.call(element.prop('href')) === '[object SVGAnimatedString]' ?
                    'xlink:href' : 'href';
         element.on('click', function(event) {
+          // If a different element was clicked, ignore it.
+          if (element[0] !== event.target) return;
```
b1rdex: I think this line caused regression. `preventDefault()` is below and called anymore.

caitp: this code hasn't been released yet, but thanks for testing out the cutting edge, i'll push a fix tonight

caitp: but just to make sure, you aren't using "cutting edge" in production right? that would be a bad idea.

b1rdex: You are right, I'm not using edge. Just tested my PR
https://github.com/angular/angular.js/pull/7995 to be working after rebase.

2015-01-26 14:37 GMT+10:00 ⭐caitp⭐ <notifications@github.com>:

> In src/ng/directive/a.js
> <https://github.com/angular/angular.js/pull/10849#discussion_r23512554>:
>
> >          // SVGAElement does not use the href attribute, but rather the 'xlinkHref' attribute.
> >          var href = toString.call(element.prop('href')) === '[object SVGAnimatedString]' ?
> >                     'xlink:href' : 'href';
> >          element.on('click', function(event) {
> > +          // If a different element was clicked, ignore it.
> > +          if (element[0] !== event.target) return;
>
> but just to make sure, you aren't using "cutting edge" in production
> right? that would be a bad idea.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/angular/angular.js/pull/10849/files#r23512554>.
>



-- 
Пашин Анатолий,
эникейщик.

PR #10850, opened 2015-01-23

### LINKED ISSUE #10698: ng-required overrides type=email initial validation in chrome
It seems since 1.3.4, ngmodel set from controller, will not get ng-invalid initially if I use both type=email and ng-required in latest chrome. works fine in firefox, or with just required attribute.

Reproducable: always
Browsers: Chrome 39.0.2171.95 m
Operating system: Windows 7

Details in the plnkr:

http://plnkr.co/edit/1iphMDUloBtUQJ4vHxdz?p=preview

### LINKED ISSUE #10828: docs: step 6 of tutorial use a href rather than an ng-href
[step_06.ngdoc](https://github.com/angular/angular.js/blob/master/docs/content/tutorial/step_06.ngdoc) uses a dynamically generated link with the href attribute rather than the [ng-href](https://docs.angularjs.org/api/ng/directive/ngHref) attribute.

Narretz:
For $validate(), it is necessary to store the parseError state
in the controller. Otherwise, if the parser name equals a validator
key, $validate() will assume a parse error occured if the validator
is invalid.

Also, setting the validity for the parser now happens after setting
validity for the validator key. Otherwise, the parse key is set,
and then immediately afterwards the validator key is unset
(because parse errors remove all other validations).

Fixes #10698
Closes #10828

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -233,6 +233,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
   this.$dirty = false;
   this.$valid = true;
   this.$invalid = false;
+  this.$$parseError = false;
   this.$error = {}; // keep invalid keys here
   this.$$success = {}; // keep valid keys here
   this.$pending = undefined; // keep pending keys here
@@ -518,7 +519,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
 
     // Check if the there's a parse error, so we don't unset it accidentially
     var parserName = ctrl.$$parserName || 'parse';
-    var parserValid = ctrl.$error[parserName] ? false : undefined;
+    var parserValid = ctrl.$$parseError ? false : undefined;
 
     var prevValid = ctrl.$valid;
     var prevModelValue = ctrl.$modelValue;
@@ -563,7 +564,6 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
       if (parseValid === undefined) {
         setValidity(errorKey, null);
       } else {
-        setValidity(errorKey, parseValid);
         if (!parseValid) {
           forEach(ctrl.$validators, function(v, name) {
             setValidity(name, null);
@@ -571,9 +571,14 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
           forEach(ctrl.$asyncValidators, function(v, name) {
             setValidity(name, null);
           });
-          return false;
+          ctrl.$$parseError = true;
         }
+        // Set the parse error last, to prevent unsetting it, should a $validators key == parserName
+        setValidity(errorKey, parseValid);
+        return parseValid;
       }
+
+      ctrl.$$parseError = false;
       return true;
     }
 
@@ -814,6 +819,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
     // TODO(perf): why not move this to the action fn?
     if (modelValue !== ctrl.$modelValue) {
       ctrl.$modelValue = ctrl.$$rawModelValue = modelValue;
+      ctrl.$$parseError = false;
 
       var formatters = ctrl.$formatters,
           idx = formatters.length;
--- a/test/ng/directive/ngModelSpec.js
+++ b/test/ng/directive/ngModelSpec.js
@@ -1222,77 +1222,93 @@ describe('ngModel', function() {
         expect(ctrl.$validators.mock.calls.length).toEqual(2);
       });
 
-      it('should validate correctly when $parser name equals $validator key', function() {
+      iit('should validate correctly when $parser name equals $validator key', function() {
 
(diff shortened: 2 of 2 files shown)


src/ng/directive/ngModel.js
```diff
@@ -518,7 +519,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
 
     // Check if the there's a parse error, so we don't unset it accidentially
     var parserName = ctrl.$$parserName || 'parse';
```
petebacondarwin: This variable doesn't seem to be used now.  What does this mean for the default of 'parse'?

petebacondarwin: I see that the `errorKey` variable sets the default below on line 563

src/ng/directive/ngModel.js
```diff
@@ -571,9 +571,14 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
           forEach(ctrl.$asyncValidators, function(v, name) {
             setValidity(name, null);
           });
-          return false;
+          ctrl.$$parseError = true;
         }
+        // Set the parse error last, to prevent unsetting it, should a $validators key == parserName
```
petebacondarwin: This comment doesn't read very well. Also if we remove the unused `parserName` variable then this makes less sense.

test/ng/directive/ngModelSpec.js
```diff
@@ -1221,6 +1221,80 @@ describe('ngModel', function() {
...
+
+        ctrl.$validators.test = function(value) {
+          return testValid;
+        };
+
+        ctrl.$validators.other = function(value) {
+          return otherValid;
+        };
+
+        ctrl.$$parserName = 'test';
```
petebacondarwin: I think `ctrl.$$parserName` was a bad idea in the first place. We ought to refactor this in general if possible.

test/ng/directive/ngModelSpec.js
```diff
@@ -1221,6 +1221,80 @@ describe('ngModel', function() {
...
+
+        ctrl.$setViewValue('stillParseInvalidAndValidatorValid');
+        expect(scope.value).toBeUndefined();
+        expect(ctrl.$error).toEqual({test: true});
+
+        ctrl.$validate();
+        expect(scope.value).toBeUndefined();
+        expect(ctrl.$error).toEqual({test: true});
+      });
+
```
petebacondarwin: This test seems overly complicated. Couldn't we replace the parser and validators with simple functions, like:

* parser -> `isEven(value)`
* validator1 -> `isOdd(value)`
* validator2 -> `isGreaterThanTen(value)`

Then we don't have to grok as many variables - just simply send in the following values:

* `1`
* `2`
* `11`
* `12`

And check that we get the correct validation?

test/ng/directive/ngModelSpec.js
Narretz: Any ideas? It's currently a private API, but we rely on it for the name of the error key. A separate function, $setParseError(name) might be possible that is called inside the parse error. Although I think we have to stick with `return undefined` denoting a parse error, bad as it is.

Narretz: We could also allow returning an object that contains parser name, error state, and return value. This would allow returning undefined as a valid value, as well as get rid of $$parserName and expose an API to actually set the parser name.

petebacondarwin: How about we use the `name` on the parser function? This would let us provide a named function for a parse and get a free name, or we can specify it explicitly.

Narretz: I just noticed that `name` is not supported in IE9+ - we'd have to polyfill it.

petebacondarwin: Hmm, how about we do something like:  `parser.name || parser.$$name` or something to allow people to specify the name explicitly if they need to support IE9

Narretz: It's all IEs, up to IE11, so I don't think it's an option.

Narretz:
I updated the test. I removed the variables that handle the validity of parser / validators. Instead, they now directly react to the input.
I didn't use the even / odd / largerThanTen scheme, because the parser and one validator need to have the same name, but would have validated different things. This is imo even more confusing. Now it's very verbose, but also quite clear what is tested.

If you're okay with the test, can we merge this, and push the parseName refactor back? The bug is pretty gnarly and we can fix the parser name later.

test/ng/directive/ngModelSpec.js
```diff
@@ -1221,6 +1221,96 @@ describe('ngModel', function() {
         expect(ctrl.$validators.mock).toHaveBeenCalledWith('a', 'ab');
         expect(ctrl.$validators.mock.calls.length).toEqual(2);
       });
+
+      iit('should validate correctly when $parser name equals $validator key', function() {
```
petebacondarwin: iit

petebacondarwin:
Oh! Sorry @Narretz - when I squashed I attributed it to me instead of you. I am sorry about that - I will not do it again. Great work.

PR #10851, opened 2015-01-24
docs(input[checkbox]): Update example to best practice

mhoffmeyerDC:
It's not required for the example to function, but it prevents scope weirdness/unexpected behavior when using directives (especially with ngTransclude!). I think it's a good pattern to encourage and might prevent a bug down the road for for people who just scan for the monospace font. See [Understanding Scopes](https://github.com/angular/angular.js/wiki/Understanding-Scopes) for mgModel best practices.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -902,16 +902,18 @@ var inputType = {
          <script>
            angular.module('checkboxExample', [])
              .controller('ExampleController', ['$scope', function($scope) {
-               $scope.value1 = true;
-               $scope.value2 = 'YES'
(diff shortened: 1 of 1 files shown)


petebacondarwin:
Yes, fair enough.

petebacondarwin:
But we should also fix up the rest of the examples in that file. I'll do that and merge.

mhoffmeyerDC:
Thanks, Peter. Think we should update the first example in [ngModel](https://docs.angularjs.org/api/ng/directive/ngModel) as well?

PR #10852, opened 2015-01-24
docs($httpBackend): enhance readability

Yahkob:
In certain parts of the $httpBackend documentation there is a recurring description for $httpBackend methods that I thought reads a bit strange: 
 "...function that receives the url and returns true if the url match the current definition.". 

This is easier to read by changing 'the' to 'a' and 'match' to 'matches':
"...function that receives ~~the~~ a url and returns true if the url ~~match~~ matches the current definition.".


pkozlowski-opensource:
Very nice patch @Yahkob - good change, good commit message - keep them coming!

Yahkob:
Thanks @pkozlowski-opensource , will do!

PR #10853, opened 2015-01-24
feat($httpUrlParams): introduce new service abstracting params serialization

### LINKED ISSUE #7429: feat(http): Please make buildUrl public
Feature Request:
It would be great if [`buildUrl`](https://github.com/angular/angular.js/blob/ad4336f9359a073e272930f8f9bcd36587a8648f/src/ng/http.js#L942) would be public. It is a function which was copied nearly identical into [Restangular](https://github.com/mgonto/restangular/blob/master/src/restangular.js#L694) and I had to do the same for a private module.

### LINKED ISSUE #9224: Passing a string that contains a semi-colon doesn't properly get encoded
Discovered this today in a rails app as the backend (since it totally barfs on unescaped semis)

I had a string such as "test;test"

passed it to params in $http post() and put() and got 

```
test;test instead of 
```

```
%3B
```

Shouldn't it encode?

pkozlowski-opensource:
This is the first step of the changes [planned]() for $http's params serialization. For now a simple refactoring that moves serialization logic to a dedicated service. Even if simple, this first moves already solves some long-opened issues:

Closes #7429
Closes #9224


shahata:
maybe we should also move the `parseKeyValue` method into this service as a `deserialize` function?

src/ng/http.js
```diff
@@ -1130,28 +1130,61 @@ function $HttpProvider() {
...
+  }];
+}
+
+/**
+ * @ngdoc provider
+ * @name $httpUrlParamsProvider
+ * @description
+ * Use `$httpUrlParamsProvider` to change the default behavior of the {@link ng.$httpUrlParams $httpUrlParams} service.
+ * */
+function $HttpUrlParamsProvider() {
```
shahata: Maybe it is better to move this service (and maybe also the relevant unit tests) to a different file?

pkozlowski-opensource:
@shahata yes, my bigger plan is to expand this service to handle deserialization in the future and - why not - turn it into a generic URLs params handling service. But I'm taking baby steps with those PRs so those are easier to review and getting merge faster.

In short: yes, definitively you remark make sense - it is mid-term plan, but doesn't have to be handled in this PR (I'm planning to send several focused PRs like this one in the coming days)

src/ng/http.js
pkozlowski-opensource: Yup, this is the plan as soon as I'm will expand on this service :-)

pkozlowski-opensource:
Thnx @caitp !

src/ng/http.js
```diff
@@ -1130,28 +1130,61 @@ function $HttpProvider() {
...
+  * It abstracts the way in which request params are transformed, grouped, encoded etc.
+  *
+  * Normally you wouldn't use this service directly in the application's code but rather override this service
+  * to enable custom serialization schemas for URL parameters.
+  *
+  * The `$httpUrlParams` service comes handy while unit testing with {@link ngMock.$httpBackend $httpBackend mock},
+  * as one can inject `$httpUrlParams` into a test and serialize URL params as {@link ng.$http $http} service.
+  */
+  this.$get = function() {
+    var HttpUrlParams = {};
```
caitp: nit: make the object literal constant

caitp:
things that I immediately don't like about this:

1. This makes it difficult to use multiple query string serialization techniques in a single app --- sometimes apps need to talk to multiple backends.
1. I think jQuery's strategy is cleaner, and allows multiple serialization strategies with less code.

pkozlowski-opensource:
I guess that "jQuery" strategy you mean their `traditional` switch, right? Would you add it to the config object? If so, this approach doesn't help with testing, does it?

Regarding support for  multiple query string serialization techniques in a single app - it is true, this proposal doesn't address this issue (but also doesn't change anything as compared to the today's situation).

Could iterate on this and [your proposal](https://gist.github.com/pkozlowski-opensource/5a57d28ccfeacaba7661#comment-1380670) to work out something that covers most of the problem people face? Or decide which problem(s) are more important (for now it seems like there is a tension between making testing easier vs. having multiple strategies in one app).

I will think of something that tries to cover both testing and multiple strategies in one app...

pkozlowski-opensource:
I'm going to close this one for now and try to implement @caitp [suggestion](https://gist.github.com/pkozlowski-opensource/5a57d28ccfeacaba7661#comment-1380670). Will open a new PR when ready.

caitp:
I've had some ideas, but most likely the simplest / most performant approach would be selecting which method to use in the http config object (using either a boolean flag to flip between "traditional" mode and not, or using a closure or a string to lookup a pre-registered serializer). The boolean flag is the least flexible but by far the simplest and fastest.

A slower approach would be having a good "default", and being able to select an alternative strategy based on the domain name of the host you're sending the request to.

Yet another one (which is probably not tenable in 1.x, but could be possible in 2.x) is just subclassing Http and setting a new default in the child class.

But yeah, I tend to prefer the boolean switch

pkozlowski-opensource:
@caitp thnx for the additional input. Personally I'm leaning towards string flag as I think a boolean one - although the simplest, is probably too restrictive. But yeh, we need to have some default + ability to select an alternative in the config object.

PR #10858, opened 2015-01-24
feat(filter): support conversion to timezone other than UTC

shahata:


shahata:
If this is something we want, I can do it also for `ng-model-options`, but let's get this in first

test/ng/filter/filtersSpec.js
```diff
@@ -457,5 +457,14 @@ describe('filters', function() {
       expect(date(new Date(2003, 8, 10, 3, 2, 4), 'yyyy-MM-dd HH-mm-ss')).toEqual('2003-09-10 03-02-04');
       expect(date(new Date(Date.UTC(2003, 8, 10, 3, 2, 4)), 'yyyy-MM-dd HH-mm-ss', 'UTC')).toEqual('2003-09-10 03-02-04');
     });
+
+    it('should support conversion to any timezone', function() {
+      expect(date(new Date(Date.UTC(2003, 8, 10, 3, 2, 4)), 'yyyy-MM-dd HH-mm-ssZ', 'GMT+0500')).toEqual('2003-09-10 08-02-04+0500');
+    });
+
+    it('should fallback to default timezone in case an unknown timezone was passed', function() {
```
petebacondarwin: Question: Fallback to **local** or to **UTC**?

petebacondarwin: Or throw?

petebacondarwin: I guess the current behaviour (before this PR) is to fallback to local time.

shahata: I currently fallback to local tz, but I can change it if you think UTC is preferred.

shahata: hmmm, yes, throwing is also a viable option. we could also have the filter output something that indicates a problem. not sure. I don't think we have any other filters that can fail, I'll look into it.

petebacondarwin: Let's go with what you have - it is a reasonable default and it is what happens now.

petebacondarwin:
LGTM - let's merge and then look at the date input controls

PR #10862, opened 2015-01-25
style(*) add rule disallowImplicitTypeConversion

rzschech:
Note in route.js the variable slash is already a string
so no conversion is required.

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -206,8 +206,7 @@ function $RouteProvider() {
         var star = option === '*' ? option : null;
         keys.push({ name: key, optional: !!optional });
         slash = slash || '';
-        return ''
-          + (optional ? '' : slash)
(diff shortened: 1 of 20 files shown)


src/ngRoute/route.js
```diff
@@ -206,8 +206,7 @@ function $RouteProvider() {
         var star = option === '*' ? option : null;
         keys.push({ name: key, optional: !!optional });
         slash = slash || '';
-        return ''
-          + (optional ? '' : slash)
+        return (optional ? '' : slash)
```
rzschech: `slash` is always a string, therefore  `(optional ? '' : slash)` is always a string and no string conversion is required.

petebacondarwin:
This is a commonly used method in the code, which I don't think we are too worried about. So I don't think we should land this rule. Sorry @rzschech

PR #10864, opened 2015-01-25
chore(ngMock): add `they` helper for testing multiple specs

matsko:


matsko:
This doesn't work yet with `inject` when additional modules are used and new services are injected in.

matsko:
Nevermind. This stuff works great. Please merge.

petebacondarwin:
I believe that Jasmine sets `this` when calling the spec function so you should too to be consistent

test/helpers/privateMocks.js
```diff
@@ -1,5 +1,37 @@
...
+    switch(specialState) {
+      case 'focus':
+        method = iit;
+        break;
+      case 'skip':
+        method = xit;
+        break;
+      default:
+        method = it;
+    }
```
shahata: How about replacing this `switch` with something like:
```js
var methods = {focus: iit, skip: xit, default: it};
var method = methods[specialState || 'default'];
```

lgalfaso:
I think this will make it harder to move to Jasmine 2.x, as in Jasmine 2.x, the function takes an optional `done` parameter

caitp:
we can handle the `done` parameter in the wrapper --- but who cares about jasmine 2.x anyways?

caitp:
also note: commit message is wrong, this is not an ngMock change

lgalfaso:
> we can handle the done parameter in the wrapper --- but who cares about
> jasmine 2.x anyways?
>

Angular/ngMock is also used by developers to test their modules, not only
to test angular itself

petebacondarwin:
For Jasmine 2, the spec functions can take an optional `done` fn parameter, yes?

If we were to migrate to supporing Jasmine 2 (which is not trivial for `ngMock`) then we would define the spec function for `they` blocks would take an optional second parameter:

```js
describe( ... ) {
  it('should do something async', function(done) {
    ...
    done();
  };

  var vals = [ ... ];
  they('should do something async', function(vals, function(val, done) {
    ...
    done();
  });
```

Is there anything else we are missing here?  If not then we should merge this once the other comments are addressed.

lgalfaso:
@petebacondarwin yes, and ok

test/helpers/privateMocks.js
petebacondarwin: +1

PR #10865, opened 2015-01-26
Add jscs rule requireCurlyBraces and disallowKeywordsOnNewLine

rzschech:
I split the `requireCurlyBraces` changes into three commits to make it easer to review the code. They are also ordered from the least contentious (adding curly braces to statements that are already multiline) to most contentious (single line statements that return, break or throw) which makes the code more verbose.

The `disallowKeywordsOnNewLine` changes are unfortunately dependent on the `requireCurlyBraces` so can't be done as an independent pull request.

### DIFF
--- a/.jscs.json
+++ b/.jscs.json
@@ -1,6 +1,7 @@
 {
   "excludeFiles": ["src/ngLocale/**"],
   "disallowKeywords": ["with"],
+  "disallowKeywordsOnNewLine": ["else"],
   "disallowMixedSpacesAndTabs": true,
   "disallowMultipleLineStrings": true,
   "disallowNewlineBeforeBlockStatements": true,
--- a/.jscs.json
+++ b/.jscs.json
@@ -22,6 +22,7 @@
   "disallowTrailingComma": true,
   "disallowTrailingWhitespace": true,
   "requireCommaBeforeLineBreak": true,
+  "requireCurlyBraces": ["if", "else", "for", "while", "do", "try", "catch"],
   "requireLineFeedAtFileEnd": true,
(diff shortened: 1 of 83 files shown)


caitp:
i think we should stop adding new style rules to enforce :| this is getting too restrictive

lgalfaso:
I am ok with these changes, but I am ok with what the majority thinks is best

hzoo:
I was thinking maybe we remove the remaining `.jscs.json.todo` file since those rules were only initially there as suggestions. Maybe even remove rules if a certain rule keeps being committed in PRs?

Although they should be for the most part pretty good - I tried opposite rules to see how much more percentage a style was being used.

petebacondarwin:
I would be quite keen to clean out the cases where we have an `if` with a line break but no curly braces

But we generally accept in our code base that a simple `if` statement can have a single statement that appears on the same single line as the `if`.  Since there is no rule for this specific scenario that doesn't also prevent the single line scenario above, I don't think we can include the `requireCurlyBraces` rule.

And so I don't think we can land this PR. Sorry @rzschech. Thank you for looking into it.

hzoo:
There is an issue for an exception to allow single statements (like returns) so maybe in the future https://github.com/jscs-dev/node-jscs/issues/244

petebacondarwin:
@hzoo - yes, I think we are about there on the JSCS rules for now.

petebacondarwin:
@rzschech - I did land the first commit though, since I like that style even if it is not enforceable right now.

rzschech:
@petebacondarwin thanks, I thought this might be contentious hence splitting the pull request into a few commits.

PR #10871, opened 2015-01-26
docs(angular-mocks): replace $rootScope with scope

marcin-wosinek:
use scope instead of $rootScope in the code example

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1030,7 +1030,7 @@ angular.mock.dump = function(object) {
   ```js
     // testing controller
     describe('MyController', function() {
-       var $httpBackend, $rootScope, createController, authRequestHandler;
+       var $httpBackend, scope, createController, authRequestHandler;
 
        // Set up the module
        beforeEach(module('MyApp'));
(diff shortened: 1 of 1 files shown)


Narretz:
Why not `$rootScope.$new()`? then you don't have to use the injector.

marcin-wosinek:
I need to get the $rootScope in some way. So the other option would be:
```js
var scope, $rootScope;
/* ... */
$rootScope = $injector.get('$rootScope');
scope = $rootScope.$new();
```

Narretz:
Yeah, sorry I didn't see that the $rootScope was not available before.

caitp:
i'm not sure this is that useful, I think for documentation purposes, it's better to clearly spell out the services being used. In real apps, you can test it however you want of course

caitp:
@Narretz if you agree please close it, otherwise land at will

Narretz:
@marcin-wosinek What was your goal in changing this? Do you think this reflects testing reality more? Angular core uses both in tests, $rootScope and a derived scope.

marcin-wosinek:
I was confused by the $rootScope in test code. The controller doesn't use any $methods from scope, and even empty object would be enough for the test to work. Because of $rootScope in the test code I was searching for some hacks related to promise resolution - as $rootScope.$digest() when testing code that uses $q

Narretz:
I thought about this again and I have agree with @caitp that replacing rootScope with scope doesn't add much clarity to the test. Nevertheless, thanks for your contribution!

PR #10880, opened 2015-01-27
fix(a): don't reload if there is only a name attribute

### LINKED ISSUE #6273: regression: ngHref -
This commit (https://github.com/angular/angular.js/commit/f3de5b6eac90baf649506072162f36dbc6d2f028) caused an old e2e test to fail (See https://github.com/angular/dgeni/issues/44).

See https://github.com/angular/angular.js/blob/master/src/ng/directive/booleanAttrs.js#L72

```
<a id="link-4" href="" name="xx" ng-click="value = 4">anchor</a> (link, don't reload)<br />
```

```
        it('should execute ng-click but not reload when href empty string and name specified', function() {
          element(by.id('link-4')).click();
          expect(element(by.model('value')).getAttribute('value')).toEqual('4');
          expect(element(by.id('link-4')).getAttribute('href')).toBe('');
        });
```

Previously, if you had a `name` but no `href` it would still prevent the click from propagating.  The performance refactor commit, referenced above, changed the `a` directive so that if there is neither `href` or `name` then the click will propagate and so clicking the link would cause the page to reload.

We need to decide if this previous functionality is important and if so modify the `a` directive accordingly and potentially re-activate the protractor test that was broken. If not, then we should just remove the failing protractor test altogether.

FAO: @IgorMinar and @kseamon

petebacondarwin:
Closes #6273

### DIFF
--- a/src/ng/directive/a.js
+++ b/src/ng/directive/a.js
@@ -16,7 +16,7 @@
 var htmlAnchorDirective = valueFn({
   restrict: 'E',
   compile: function(element, attr) {
-    if (!attr.href && !attr.xlinkHref && !attr.name) {
+    if (!attr.href && !attr.xlinkHref) {
       return function(scope, element) {
         // If the linked element is not an anchor tag anymore, do nothing
         if (element[0].nodeName.toLowerCase() !== 'a') return;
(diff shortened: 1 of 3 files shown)


caitp:
ping @kseamon about this one I guess

petebacondarwin:
@kseamon did say it was OK to remove the name check in [this comment](https://github.com/angular/angular.js/issues/6273#issuecomment-57448466)

kseamon:
Yup

PR #10883, opened 2015-01-27
fix($http): properly access request headers with mixed case

### LINKED ISSUE #10881: headersGetter alway return null on transformRequest function in $http
Because headers object from request has upper case name,
but when headersGetter function try to retrieve it convert lookup name to lower case.

```javascript
function headersGetter(headers) {
  var headersObj = isObject(headers) ? headers : undefined;

  return function(name) {
    if (!headersObj) headersObj =  parseHeaders(headers);

    if (name) {
      var value = headersObj[lowercase(name)];
      if (value === void 0) {
        value = null;
      }
      return value;
    }

    return headersObj;
  };
}
```

pkozlowski-opensource:
Fixes #10881

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -37,19 +37,24 @@ function isJsonLike(str) {
  * @returns {Object} Parsed headers as key value object
  */
 function parseHeaders(headers) {
-  var parsed = createMap(), key, val, i;
+  var parsed = createMap(), i;
 
-  if (!headers) return parsed;
-
-  forEach(headers.split('\n'), function(line) {
-    i = line.indexOf(':');
-    key = lowercase(trim(line.substr(0, i)));
-    val = trim(line.substr(i + 1));
-
+  function fillInParsed(key, val) {
     if (key) {
       parsed[key] = parsed[key] ? parsed[key] + ', ' + val : val;
     }
-  });
+  }
+
+  if (isString(headers)) {
+    forEach(headers.split('\n'), function(line) {
+      i = line.indexOf(':');
+      fillInParsed(lowercase(trim(line.substr(0, i))), trim(line.substr(i + 1)));
+    });
+  } else if (isObject(headers)) {
+    forEach(headers, function(headerVal, headerKey) {
+      fillInParsed(lowercase(headerKey), trim(headerVal));
+    });
+  }
 
   return parsed;
 }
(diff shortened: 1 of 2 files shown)


src/ng/http.js
```diff
@@ -71,14 +71,16 @@ function headersGetter(headers) {
   var headersObj = isObject(headers) ? headers : undefined;
 
   return function(name) {
+    var existingHeaderName, lowercaseName = lowercase(name);
```
Narretz: Can you please move this directly above the `for .. in`? Or even inside, browsers should be able to optimize this.

pkozlowski-opensource: Hmm, I would rather not as if / for don't create lexical scope in JS. I know that browsers will do it for you but I think that most of the code in AngularJS is declaring variables at the lexical scope, no?

Anyway, I don't feel strong about it, let's just be consistent.

Narretz: I hadn't thought about that actually. I just find it less unreadable - as the variable is only used in the loop.
Anyway, it's your call. However, if you don't change it, can you give the variable its own `var`? I think that most new code follows this pattern, and I think at least @caitp also prefers it.

pkozlowski-opensource: @Narretz please have a look at the existing in methods in $http - as far as I can see there is not consistent style for declaring vars but many methods declare multiple vars using one `var` keyword ([example](https://github.com/angular/angular.js/blob/d729fcf030be1d3ef37196d36ea3bf3249ee3318/src/ng/http.js#L855-L857)). One again, I'm all for consistency and clear code standards, but given lack of consistency in the current code base, I don't think style discussion should be happening to in this PR.

My point is: I'm all for common coding standards and consistent coding patterns and we can raise it on the team meeting. But given that this PR doesn't introduce different coding style, as compared to the existing code in this file, I don't think that this PR is the best place to discuss personal coding preferences... At the end of the day I don't care either way as long we all on the same page.

Narretz: Fair enough. Let's just leave it as is.

pkozlowski-opensource:
@petebacondarwin would you mind taking a quick look at this one?

petebacondarwin:
Style discussions aside - Wouldn't it be more performance to simply lower case the header names when parsing them in the first place?

pkozlowski-opensource:
@petebacondarwin here we are talking about _request_ headers so the ones provided by a user. In this sense we are not parsing anything.

pkozlowski-opensource:
@petebacondarwin I did the changes we've discussed during the hangout review. Mind taking another look?

petebacondarwin:
LGTM

PR #10893, opened 2015-01-28
fix(ngOptions): fix model<->option interaction when using track by

### LINKED ISSUE #10869: ng-select/ng-options causes underlying data to be changed, when assiging properties on ng-model
Hello!

we encountered some odd behaviour, when using ng-select and changing the ng-model _not_ through a selection.

http://jsfiddle.net/r5xanz0s/1/

There are 2 different scenarios (starting after a reload):

1. When assigning new properties for ng-model (via "Assign 'someName3' properties only") the ng-select is not updated at all.
2. When selecting 'someName2' and then assigning new properties for ng-model (via "Assign 'someName3' properties only") the entry for 'someName2' within the select is changed to data from 'someName3'. This results in a duplicate entry.

This behaviour does not occur, when assigning a new object to ng-model.

Any ideas?

Best,
Gregor

petebacondarwin:
Closes #10869

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -335,6 +335,11 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
           selectValueMap: selectValueMap,
           getOptionFromViewValue: function(value) {
             return selectValueMap[getTrackByValue(value, getLocals(value))];
+          },
+          getViewValueFromOption: function(option) {
+            // If the viewValue could be an object that may be mutated by the application,
+            // we need to make a copy and not return the reference to the value on the option.
+            return trackBy ? angular.copy(option.viewValue) : option.viewValue;
           }
         };
       }
@@ -428,7 +433,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         if (selectedOption && !selectedOption.disabled) {
           removeEmptyOption();
           removeUnknownOption();
-          return selectedOption.viewValue;
+          return options.getViewValueFromOption(selectedOption);
         }
         return null;
       };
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@petebacondarwin I don't know the code of select / ngOptions well enough to give you thumbs up but I would feel much better if tests could verify the impact of changes in the object on what is displayed by select (if possible).

I guess this PR will need another pair of eyes or we can hangout together so you can walk me through the code / design.

petebacondarwin:
@pkozlowski-opensource - I updated the test. Let me know if this is still missing something...

src/ng/directive/ngOptions.js
```diff
@@ -492,6 +495,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
 
       // We will re-render the option elements if the option values or labels change
       scope.$watchCollection(ngOptions.getWatchables, updateOptions);
+      scope.$watch(attr.ngModel, function() { ngModelCtrl.$render(); }, true);
```
petebacondarwin: The reason we now need this is that we can't assume that the selected model value is not a complex object, in which case the normal `ngModel` checking would fail to notice the change here.

src/ng/directive/ngOptions.js
```diff
@@ -335,6 +335,9 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
           selectValueMap: selectValueMap,
           getOptionFromViewValue: function(value) {
             return selectValueMap[getTrackByValue(value, getLocals(value))];
+          },
+          getViewValueFromOption: function(option) {
+            return trackBy ? angular.copy(option.viewValue) : option.viewValue;
```
petebacondarwin: Since the viewValue might now be an object that could be mutated by the application, we need to make a copy and not return the reference to the value on the option.

Narretz:
As far as I can tell, this looks good. Your explanation in https://github.com/angular/angular.js/issues/10869#issuecomment-71827584 was very helpful. I think it would make sense to add a condensed version of the explanation to the commit message or add your comments as code comments.

petebacondarwin:
Great! Thanks @Narretz - I will add that info to the commit and possibly as comments in code.

PR #10894, opened 2015-01-28
fix($parse): remove references to last arguments to a fn call

vojtajina:
This can be an issue if running (and killing) multiple apps/injectors on the same page. The `args` array holds references to all previous arguments to a function call and thus they cannot be garbage-collected.

In a regular (one app/injector on a page) app, this is not an issue.

cc @aaronfrost

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -728,6 +728,11 @@ Parser.prototype = {
             ? fn.apply(context, args)
             : fn(args[0], args[1], args[2], args[3], args[4]);
 
+      if (args) {
+        // Free-up the memory (arguments of the last function call).
+        args.length = 0;
+      }
+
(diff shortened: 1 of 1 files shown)


petebacondarwin:
If the `$parse` service was destroyed and rebuilt would that not free up these objects?

jbedard:
Is there any point keeping the array around if it gets set to a length of 0 each time? Does that still avoid reallocating each time?

LorenzoDV:
Just a note: this issue could have still created memory leaks in a single-injector app when using `ng-view`.

Passing a function argument refering to the DOM (such as `$event`) would have caused the DOM reference to be indirectly cached. After switching the view away, its entire DOM would have been leaked, including directives, scopes, etc.

PR #10898, opened 2015-01-28
fix(validators): check if viewValue is empty for maxlength

Narretz:
@petebacondarwin This is the bug you found where $isEmpty is called with the modelValue. To make this consistent, we should also add this test to all other validators. Does this make sense to you?


petebacondarwin:
LGTM - feel free to merge. You can add the extra tests if you like.

PR #10899, opened 2015-01-29
feat(limitTo): extend the filter to take a beginning index argument

### LINKED ISSUE #5355: limitTo filter should have a left and right bound.
using like this:
`{{'abcdefg' | limitTo:2:5}}`
output:
`cde`
useful when paging sth..

tameraydin:
Extend the limitTo filter to take an optional argument for beginning index. It provides a slice-alike functionality to manipulate the input.

Closes #5355

### DIFF
--- a/src/ng/filter/limitTo.js
+++ b/src/ng/filter/limitTo.js
@@ -17,6 +17,8 @@
  *     If the number is negative, `limit` number  of items from the end of the source array/string
  *     are copied. The `limit` will be trimmed if it exceeds `array.length`. If `limit` is undefined,
  *     the input will be returned unchanged.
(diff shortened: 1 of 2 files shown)


gkalpak:
Now that there is an `offset` parameter, it might make sense to disallow negative `limit`, as it leads to weird semantics (and code).

gkalpak:
<sub> But it **is**going to be a breaking change :(</sub>

pkozlowski-opensource:
This PR seems good after a quick look - that is - it does what it announces. Having said this the resulting API is a bit odd... but I guess this is the best we can do without API breaking changes... 

Let's reach out to the community to see what other think...

revolunet:
very cool thanks

PR #10910, opened 2015-01-29
fix($controller): throw better error when controller expression is bad

### LINKED ISSUE #10875: angular cannot read property '1' of null  is not a useful error message.
What does this error even mean.  Error message is not actionable.

caitp:
Previously, the error was a JS runtime error when trying to access a property of `null`. But, it's
a bit nicer to throw a real error and provide a description of how to fix it. Developer ergonomics
and all that.

Closes #10875


steowens:
This is first rate code.  Impressive.

PR #10913, opened 2015-01-29
docs(ngHide): Use proper selector for CSS override

### LINKED ISSUE #10902: ngHide CSS override doesn't work as documented
There seems to be a disconnect between the docs for ngHide and how the code behaves (this is present in at least Angular 1.3 and 1.4). According to the docs:

> By default, the `.ng-hide` class will style the element with `display: none!important`. If you wish to change
> the hide behavior with ngShow/ngHide then this can be achieved by restating the styles for the `.ng-hide`
> class in CSS:
> 
> .ng-hide { … }

However, in practice that CSS override doesn't take effect. It looks like it's losing out to the last clause of this rule, which is being inlined somehow:

```css
[ng\:cloak],[ng-cloak],[data-ng-cloak],[x-ng-cloak],.ng-cloak,.x-ng-cloak,.ng-hide:not(.ng-hide-animate){
	display:none !important;
}
```

[Here's a simple example that illustrates the problem](http://plnkr.co/edit/yFhKgMSIM4PwCoEELmg1). We should be overriding the `.ng-hide` class to show hidden items, but they are still hidden.

I'm not sure if the code needs to move towards the docs, or the docs toward the code. Ideally I'd like to have it work as documented, since overriding the one class rule is easier than overriding the other complicated selectors. But I'm not sure what's going on underneath the hood here, so I don't feel qualified to propose the solution.

iangreenleaf:
The animation selector gives the default styles greater specificity that should be matched when overriding.

Closes #10902

### DIFF
--- a/src/ng/directive/ngShowHide.js
+++ b/src/ng/directive/ngShowHide.js
@@ -40,10 +40,10 @@ var NG_HIDE_IN_PROGRESS_CLASS = 'ng-hide-animate';
  *
  * By default, the `.ng-hide` class will style the element with `display: none!important`. If you wish to change
  * the hide behavior with ngShow/ngHide then this can be achieved by restating the styles for the `.ng-hide`
- * class in CSS:
+ * class (plus a psuedo-selector for animation) in CSS:
  *
  * ```css
- * .ng-hide {
+ * .ng-hide:not(.ng-hide-animate) {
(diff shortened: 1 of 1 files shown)


petebacondarwin:
Tweaked and merge. Thanks @iangreenleaf

PR #10919, opened 2015-01-30
feat(ngLocale): Add FIRSTDAYOFWEEK and WEEKENDRANGE from google data

santialbo:
This data is available on the Google closure localization data used. This pull request adds `FIRSTDAYOFWEEK` and `WEEKENDRANGE` into the `DATETIME_FORMATS` property of $locale.

### DIFF
--- a/i18n/spec/closureI18nExtractorSpec.js
+++ b/i18n/spec/closureI18nExtractorSpec.js
@@ -184,6 +184,8 @@ describe("extractDateTimeSymbols", function() {
                       'nov.', 'déc.'],
                   DAY: ['dimanche', 'lundi', 'mardi', 'mercredi', 'jeudi', 'vendredi', 'samedi'],
                   SHORTDAY: ['dim.', 'lun.', 'mar.', 'mer.', 'jeu.', 'ven.', 'sam.'],
+                  FIRSTDAYOFWEEK: 6,
+                  WEEKENDRANGE: [5, 6],
                   AMPMS: ['AM', 'PM'],
                   medium: 'yyyy-MM-dd HH:mm:ss',
                   short: 'yy-MM-dd HH:mm',
--- a/i18n/src/converter.js
+++ b/i18n/src/converter.js
@@ -42,6 +42,8 @@ function convertDatetimeData(dataObj) {
   datetimeFormats.DAY = dataObj.WEEKDAYS;
   datetimeFormats.SHORTDAY = dataObj.SHORTWEEKDAYS;
   datetimeFormats.AMPMS = dataObj.AMPMS;
+  datetimeFormats.FIRSTDAYOFWEEK = dataObj.FIRSTDAYOFWEEK;
+  datetimeFormats.WEEKENDRANGE = dataObj.WEEKENDRANGE;
 
 
   datetimeFormats.medium      = dataObj.DATEFORMATS[2] + ' ' + dataObj.TIMEFORMATS[2];


santialbo:
CLA signed.

santialbo:
If you need me to commit the generated $locale files please say so.
regards

caitp:
It looks okay to me, but we don't really have a translation from the monday-sunday format and the sunday-saturday format, does that matter?

caitp:
(i'm looking at the closure code which uses this and it's not very clear on how exactly they're using the monday-sunday format)

santialbo:
I'm not sure I understand what you mean @caitp, the numbers they use go from 0-Monday to 6-Sunday (I looked at a few $locales that I know).

caitp:
yes, I know --- I call this `monday-sunday` --- but the serialized date values are sunday-saturday in ngLocale, so it seems hard to get the text value associated (eg if you were trying to create a calendar widget, you'd want to put the right text in the right column...) I think you need a few more variables to do that

santialbo:
I see, I think you should be fine doing
    SHORTDAY[(i + 1) % 7]
where i is the number in the `moday-sunday` format.

caitp:
that depends on what the first day of the week and weekend range values are.

caitp:
either way i'm fine to land it, @petebacondarwin is this alright with you?

petebacondarwin:
We can add in these items but I think that they are misleading.

`FIRSTDAYOFWEEK` is only used in Google Closure library for computing the week number of a day in a year.

> Week numbers are defined in this same ISO 8601 standard. Each year has 52 or 53 weeks and weeks always start on a Monday. Week number 1 of each year is the first week in a year that has the first Thursday of the year, or in other words, the week containing January 4th.

See http://docs.closure-library.googlecode.com/git/local_closure_goog_date_date.js.source.html#line234

I can't find any use of the `WEEKENDRANGE`, in Google Closure. It also seems to be based on Monday being the first day of the week. So one is going to use it to get locale specific weekend day names then you would need to do the mapping between the `[MON, SUN]` of `WEEKENDRANGE` and the `[SUN,SAT]` of the `WEEKDAYS` array.

@santialbo - is this really beneficial to you?

pkozlowski-opensource:
@petebacondarwin Personally I've got a use-case for locale-dependent week no computation (date-picker for example) and the ISO 8601 logic is completely crazy to the point of not being that useful in practice. 

I think that it would be cool if we could have locale-dependent first-day of a week as currently it is hard-code to Sun (works for US but not for Europe, for example)

santialbo:
@petebacondarwin Yes, it's confusing that the `SHORTDAY` array starts on Sundays but these two properties are set on a 0-Monday scale. A mapping would be needed, but it's straighforward, i.e. `SHORTDAY[(i + 1) % 7]` to get the name of the day.

My usecase is a dayparting directive, something similar to the following image I just found.
dayparting
I would like to be able to order days depending on the locale (e.g. Sunday first on the US, Monday first on the UK) and also highlight weekends in a different colour.

petebacondarwin:
OK.  Sounds fair enough to land

santialbo:
Thanks @petebacondarwin. Do I need to commit the updated generated files or are they generated during the main build?

petebacondarwin:
We can generate the new files when we merge

petebacondarwin:
Landed as 3d149c7f20ffabab5a635af9ddcfc7105112ab4a and 80f139b8607ce8ef066f0a23142cfb0aebbdc0c2. Thanks

PR #10922, opened 2015-01-31
feat(ngModelOptions): allow options to be inherited from ancestors

petebacondarwin:
Previously, you had to apply the complete set of ngModelOptions at every point where
you might want to modify just one or two settings.

This change allows more general settings to be applied at nearer the top of the DOM
and then for more specific settings to override those general settings further down
in the DOM.

_It should also be noted that this change means that `ctrl.$options` will always reference a valid object
and so a large part of this PR is removing the unnecessary checks for existence of that object._

BREAKING CHANGE:

Previously, if a setting was not applied on ngModelOptions, then it would default
to undefined. Now the setting will be inherited from the nearest ngModelOptions
ancestor.

It is possible that an ngModelOptions directive that does not set a property,
has an ancestor ngModelOptions that does set this property to a value other than
undefined. This would cause the ngModel and input controls below this ngModelOptions
directive to display different behaviour. This is fixed by explictly setting the
property in the ngModelOptions to prevent it from inheriting from the ancestor.


src/ng/directive/ngModel.js
```diff
@@ -1065,9 +1072,41 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
+ *
+ * ```html
+ * <div ng-model-options="{ allowInvalid: true }">
+ *   <form ng-model-options="{ updateOn: \'blur\' }">
+ *     <input ng-model-options="{ updateOn: \'default\' }">
+ *   </form>
+ * </div>
+ * ```
+ *
+ * ... the `input` element will effective have the follow settings ...
```
gkalpak: _effectively_ ?
_following_ ?

src/ng/directive/ngModel.js
```diff
@@ -1337,3 +1403,45 @@ function isObjectEmpty(obj) {
...
+/**
+ * @ngdoc service
+ * @name $modelOptions
+ * @description
+ * This service provides support to the ngModelOptions directive.
+ *
+ * You can register default options that are used application wide if no ngModelOptions
+ * directive is found.
+ *
+ * You can register model adaptors that can selected by ngModelOptions settings to modify
```
gkalpak: can _be_ selected

src/ng/directive/ngModel.js
```diff
@@ -1337,3 +1403,45 @@ function isObjectEmpty(obj) {
...
+          updateOnDefault: true
+        },
+        /**
+         * @ngdoc property
+         * @name $modelOptions#modelAdaptors
+         * @type {Object<string, Function>}
+         * @description
+         * The functions that adapt the behaviour of ngModel
+         */
+        modelAdaptors: {}
```
petebacondarwin: This modelAdaptors bit should not be here yet. That is for a future PR :-)

src/ng/directive/ngModel.js
petebacondarwin: Yes, "following" thanks @gkalpak

src/ng/directive/ngModel.js
gkalpak: What is this ? It doesn't seem to be used anywhere. Is it a WIP ?

src/ng/directive/ngModel.js
gkalpak: Not a big deal, but shouldn't `effective` be `effectively` ?
(I am not insisting or anything, just bringing it up again in case you overlooked it; you obviously know better (being a native speaker :smiley:).)

petebacondarwin: :-) yes. Sorry about that

src/ng/directive/ngModel.js
```diff
@@ -1016,7 +1019,11 @@ var ngModelDirective = ['$rootScope', function($rootScope) {
           var modelCtrl = ctrls[0],
               formCtrl = ctrls[1] || nullFormCtrl;
 
-          modelCtrl.$$setOptions(ctrls[2] && ctrls[2].$options);
```
Narretz: For readability, could you assign ctrl[2] to a variable?

src/ng/directive/ngModel.js
```diff
@@ -1083,7 +1122,130 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
+ *
+ * The default behaviour in `ngModel` is that the model value is set to `null` when the validation
+ * determines that the value is invalid. By setting the `allowInvalid` property to true, the model
+ * will still be updated even if the value is invalid.
+ *
+ *
+ * ## Connecting to the scope
+ *
+ * By setting the `getterSetter` property to true you are telling ngModel that the `ngModel` expression
+ * on the scope actually refers to a "getter/setter" function rather than the actual value itself.
```
Narretz: kind of awkward duplication of "actually" and "actual". Maybe just get rid of "actually"?

src/ng/directive/ngModel.js
```diff
@@ -1337,3 +1403,35 @@ function isObjectEmpty(obj) {
...
+
+
+/**
+ * @ngdoc service
+ * @name $modelOptions
+ * @description
+ * This service provides support to the {@link ngModelOptions} directive.
+ *
+ * Here, you can register default options that are used application wide if no {@link ngModelOptions}
+ * directive is found.
```
Narretz: "See ngModelOptions for a list of available options" And should we mention what exactly the default options are?

Narretz:
Looks good so far, but could you mention the modelOptionsProvider in the commit message?

Narretz:
Btw, in the last paragraph, I am not completely sure what "This would cause the ngModel and input controls below this ngModelOptions directive to display different behaviour." means. Different as in unexpected? Because the option is applied (from ancestor) even though you did not specify it on the current element?

petebacondarwin:
@gkalpak and @Narretz - thanks for the reviews. I will update now.

src/ng/directive/ngModel.js
```diff
@@ -1096,138 +1258,42 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
+    priority: 10,
+    require: ['ngModelOptions', '?^^ngModelOptions'],
+    controller: function() {},
+    link: {
+      pre: function ngModelOptionsPreLinkFn(scope, element, attrs, ctrls) {
+        var optionsCtrl = ctrls[0];
+        var parentOptions = ctrls[1] ? ctrls[1].$localOptions : {};
+
+        // Store the raw options taken from the attributes (after inheriting parent options)
+        optionsCtrl.$localOptions = extend({}, scope.$eval(attrs.ngModelOptions), parentOptions);
```
petebacondarwin: Oops, we all missed that I had the parameters in the wrong order here!!

src/ng/directive/ngModel.js
petebacondarwin: Will do

src/ng/directive/ngModel.js
petebacondarwin: Will get rid of both "actually" :-)

src/ng/directive/ngModel.js
petebacondarwin: Yep

src/ng/directive/ngModel.js
Narretz: Oh my ... well you didn't miss it. ;) Should the merging be tested?

petebacondarwin: On it

petebacondarwin:
My main concern about the design of this is that

a) ngModelOptions has to do some post processing on the options which adds some complexity. We could get rid of this with a breaking change to the `updateOn` option property; but actually it might be a nice feature since we might want to do some processing to optimize model "adaptors" when/if they get implemented.

b) if an ngModel directive has no ancestor ngModelOptions then it needs to get its defaults from the `$modelOptions` service, but this won't have done the post processing (or we would need to replicate the functionality in both the `ngModelOptions` directive and the `$modelOptions` service.

One solution to both of these is to put the post processing into the `$modelOptions` and then the ngModel and ngModelOptions directives could pass the options to a method to be processed (maybe cached somehow). Then we could be really flexible and allow things to register specific post processors on the options that would be run on all instances of options.

Narretz:
Doing the post processing in the $modelOptions service sounds like a good idea to me.

src/ng/directive/ngModel.js
```diff
@@ -1095,141 +1251,97 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
+   * directives as ancestors
+   *
+   * The initial default options are:
+   *
+   * * `updateOn`: `default`
+   * * `debounce`: `0`
+   */
+  var defaultOptions = {
+    updateOn: 'default',
+    debounce: 0
```
shahata: maybe add the rest of the options just so it will be a bit more explicit?

src/ng/directive/ngModel.js
```diff
@@ -1095,141 +1251,97 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
-        input.click();
-        expect(model.getText()).toEqual('say');
-        other.click();
-        expect(model.getText()).toEqual('say hello');
-      });
+  return {
+    defaultOptions: defaultOptions,
+    $get: function() {
+      return {
+        defaultOptions: new ModelOptions(defaultOptions)
```
shahata: I think that this will not work if someone will do:
```js
modelOptionsProvider.defaultOptions = {updateOn: 'blur'};
```
rather then
```js
modelOptionsProvider.defaultOptions.updateOn = 'blur';
```

src/ng/directive/ngModel.js
```diff
@@ -1095,141 +1251,97 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
+    // extract "default" pseudo-event from list of events that can trigger a model update
+    this.updateOn = trim(options.updateOn.replace(DEFAULT_REGEXP, function() {
+      self.updateOnDefault = true;
+      return ' ';
+    }));
+  } else if (parent) {
+    this.updateOn = parent.updateOn;
+    this.updateOnDefault = parent.updateOnDefault;
+  } else {
+    this.updateOnDefault = true;
```
shahata: I think there's a missing test for this case (assuming I understand the case correctly):
```js
it('should update on default in case default options do not include updateOn', function () {
  module(function ($modelOptionsProvider) {
    delete $modelOptionsProvider.defaultOptions.updateOn;
  });
  inject(function ($rootScope) {
    var inputElm = helper.compileInput('<input type="text" ng-model="name" name="alias" />');
    var inputOptions = $rootScope.form.alias.$options;
    expect(inputOptions.updateOnDefault).toEqual(true);
  });
});
```

src/ng/directive/ngModel.js
```diff
@@ -1095,141 +1251,97 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
+        var parentOptions = ctrls[1] ? ctrls[1].$options : $modelOptions.defaultOptions;
+        optionsCtrl.$options = new ModelOptions(scope.$eval(attrs.ngModelOptions), parentOptions);
+      }
+    }
+  };
+}];
+
+
+/**
+ * @ngdoc service
```
shahata: provider docs are missing?

src/ng/directive/ngModel.js
petebacondarwin: Yeah, that is a problem... I will have to think about that.

src/ng/directive/ngModel.js
petebacondarwin: Right!

src/ng/directive/ngModel.js
```diff
@@ -1095,141 +1251,97 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
  *   - `timezone`: Defines the timezone to be used to read/write the `Date` instance in the model for
  *     `<input type="date">`, `<input type="time">`, ... . Right now, the only supported value is `'UTC'`,
  *     otherwise the default timezone of the browser will be used.
+ */
+var ngModelOptionsDirective = ['$modelOptions', function($modelOptions) {
+  return {
+    restrict: 'A',
+    // ngModelOptions needs to run before ngModel and input directives
+    priority: 10,
+    require: ['ngModelOptions', '?^^ngModelOptions'],
```
Narretz: This means every `ngModelOptions` directive needs to search for a parent `ngModelOptions` up to the app root element. Isn't that a bit excessive? I think we could instead allow `ngModelOptions` on form / ngForm directives, which also seems more semantic to me.

Narretz: Hm, thinking again, since it only applies if you set ngModelOptions, it should actually be okay.

Narretz:
There was just a new issue opened for that: https://github.com/angular/angular.js/issues/12644
So let's revisit this:
Looks like the post-processing is extracted!
Now what's left, supporting both setting / replacing the options in the provider?
And a rebase.

src/ng/directive/ngModel.js
petebacondarwin: I have fixed it so that you can replace the `defaultOptions` object on the provider.
But if you start mucking about with that object in the service then it will still break.

src/ng/directive/ngModel.js
petebacondarwin: Added a relevant test

src/ng/directive/ngModel.js
petebacondarwin: Added.

src/ng/directive/ngModel.js
petebacondarwin: And we actually do this sort of thing for every `ngModel` anyway looking for `form` or `ngForm` directives already.

petebacondarwin:
So I added a bunch more code to this PR.
Given that it is a breaking change anyway, I went all out and broke it properly.
It shouldn't actually break for most users, only those who are writing directives that read from the model options directly.

The upside is that it will allow programmatic changes to model options as well as making it harder to wipe out the options by dodgy overwriting of the `ModelOptions` object with a POJO.

test/ng/directive/formSpec.js
```diff
@@ -290,7 +290,7 @@ describe('form', function() {
   describe('triggering commit value on submit', function() {
     it('should trigger update on form submit', function() {
       var form = $compile(
-          '<form name="test" ng-model-options="{ updateOn: \'\' }" >' +
+          '<form name="test" ng-model-options="{ updateOn: \'submit\' }" >' +
```
Narretz: Why this change?

test/ng/directive/ngModelSpec.js
```diff
@@ -1784,6 +1807,63 @@ describe('ngModelOptions attributes', function() {
   }));
 
 
+  it('should inherit options from ngModelOptions directives declared on ancestor elements', function() {
+    var container = $compile('<div ng-model-options="{ allowInvalid: true }">' +
+               '<form ng-model-options="{ updateOn: \'blur\' }">' +
+                 '<input ng-model-options="{ updateOn: \'default\' }">' +
```
Narretz: Did you want to check the value of this one, too or are they there for a particular reason?

Narretz:
When you broke it good and proper, did you introduce new BC that aren't yet included in the BC notice?

test/ng/directive/formSpec.js
petebacondarwin: Good question. It seems that my commits caused a change to the semantic of providing an empty string to the `updateOn` property. We never documented what an empty string should do, and I honestly don't know why these tests used an empty string. I think they are trying to "never" update the model, right?

test/ng/directive/ngModelSpec.js
```diff
@@ -1784,6 +1807,63 @@ describe('ngModelOptions attributes', function() {
...
+              '</div>')($rootScope);
+
+    var form = container.find('form');
+    var input = container.find('input');
+
+    var containerOptions = container.controller('ngModelOptions').$options;
+    var formOptions = form.controller('ngModelOptions').$options;
+    var inputOptions = input.controller('ngModelOptions').$options;
+
+    expect(inputOptions.getOption('allowInvalid')).toEqual(true);
```
petebacondarwin: this should be `containerOptions`

test/ng/directive/ngModelSpec.js
petebacondarwin: The idea is that we have three levels of options: container, form and input.
We want to check that the `allowInvalid` option is getting inherited all the way down. But I agree that we might not be creating a new `ModelOptions` object if this `<input>` does not contain an `ng-model` directive. So we should indeed add that.

petebacondarwin:
@Narretz - I still need to update the commit message with the new BC, thanks

test/ng/directive/ngModelSpec.js
Narretz: Still needs to be changed

test/ng/directive/ngModelSpec.js
```diff
@@ -1771,6 +1771,29 @@ describe('ngModel', function() {
 });
 
 
+describe('$modelOptions', function() {
```
Narretz: Can you add a test the checks the default values?

src/ng/directive/ngModel.js
```diff
@@ -1163,7 +1196,130 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
  * `submit` event. Note that `ngClick` events will occur before the model is updated. Use `ngSubmit`
  * to have access to the updated model.
  *
- * `ngModelOptions` has an effect on the element it's declared on and its descendants.
+ * The following example shows how to override immediate updates. Changes on the inputs within the
```
Narretz: The original examples for immediate updates, debouncing and getterSetter are still in the file. Can you remove them?

Narretz:
LGTM, except some minor issues. And in the commit message there could be an example of how to migrate if a setting that was set to undefined must now be explicitly set to avoid override from a parent / provider setting.

src/ng/directive/ngModel.js
petebacondarwin: Done

test/ng/directive/ngModelSpec.js
petebacondarwin: Done

test/ng/directive/ngModelSpec.js
petebacondarwin: Fixed

petebacondarwin:
I have updated the commit message. @Narretz PTAL

Narretz:
LGTM

fenduru:
What is the motivation for this change? From the docs:

> Although we try to avoid breaking changes, there are some cases where it is unavoidable:
>
> * AngularJS has undergone thorough security reviews to make applications safer by default, which drives many of these changes.
> * Several new features, especially animations, would not be possible without a few changes.
> * Finally, some outstanding bugs were best fixed by changing an existing API.

This does not seem to fit into any of these categories of unavoidable breaking changes.

As the author of an Angular component library, this is massively breaking due to the fact that we export controls that support `ng-model` as part of their API, but also consume smaller components in their templates. We have already had to guard against the fact that `ng-model` does `require: '^^ngModelOptions'`.

In all the places that we guard against this, we will now also have to add every possible `ngModelOptions` key explicitly. This has the probably unintentional side effect of making any addition to `ngModelOptions` (for instance, introducing a new option in Angular 1.7) a breaking change.

This seems like something that has small enough value that it should not have been deemed necessary to make in a minor release.

Narretz:
The main value of this is that you don't have to set your model options for every single input, but can set it either in a parent or in the provider. It's very common that you want the same input behavior for parts of your application. And you can still customize your ngModels per instance. You can even "reset" a sub-part of your application (if you don't know what kind of inputs appear in it for example). At the moment, you need to know the defaults for that, that's true. We could maybe introduce a special value that "resets" the options to the default.
Can you show an example of your component and why it breaks? Is it because with this change you possibly don't know what ngModelOptions are set for the application?

fenduru:
> The main value of this is that you don't have to set your model options for every single input, but can set it either in a parent or in the provider.

I personally do not feel that this convenience is enough justification for a breaking change.

> Can you show an example of your component and why it breaks? Is it because with this change you possibly don't know what ngModelOptions are set for the application?

This breaks my component because the user was able to inject different behavior past my API boundary with that user. We have this all over our library, but for a simple contrived (though realistic) example:

```html
<!-- myReusableComponent template -->
Enter text: <input ng-model="internalModel" ng-model-options="{ updateOn: 'keypress' }"></input>
This should always update instantly: {{ internalModel}}
<button ng-click="writeToModel(internalModel)"></button>

<!-- Application template -->
<my-reusable-component ng-model="userModel" ng-model-options="{ debounce: 100 }"></my-reusable-component>
```

The user's intent was to debounce how often that `userModel` gets written to, however because of the inheritance now they are also affecting how often the `internalModel` inside of `myReusableComponent` updates.

So basically I'm (as a library author) forced to guard against an Angular "feature", while at the same time my user (app author) needs to be aware of whether or not the libraries they are using properly guard against this (in order to satisfy `users must make sure their apps (and that includes 3rd party code) are compatible with it when they use it` - I don't think it is reasonable for users to have to understand the implementation details of the libraries they are using. Seems weird to me if the answer to "can we upgrade to 1.6 safely" includes the statement "I will need to look at ngMaterial's source code to be sure".

> I'm also not sure introducing a new option will introduce a BC. A new option will be`optional, so users must make sure their apps (and that includes 3rd party code) are compatible with it when they use it.

If a framework (Angular) changes code that causes my project (library) to no longer work correctly without me doing work, that is by definition a breaking change. If one of my users starts to use a new option somewhere higher up in the DOM, that option will cause my implementation details to behave differently, causing my library to need code changes.

petebacondarwin:
This is an interesting use case that we had not thought about. We should definitely work something out here before we release 1.6.0.

Can I be devil's advocate for a moment...

> You say, @fenduru, that this PR would make you have to set ngModelOptions for all possible values in every directive that you create. Strictly you only need to set the values of the ones that you care about. In the example case, `debounce`.
> 
> With the behaviour that this PR introduces, any new options added to ngModelOptions would most likely need to be considered as a breaking change since they could affect the behaviour of applications that are already in production. This is the same situation as if we changed the default behaviour of an input directive directly.
> 
> So if we followed this approach you would only need to update the ngModelOptions values on your own directives at breaking change versions, e.g. 1.7.0, 1.8.0, etc.
 

I wonder if this restriction on adding values to ngModelOptions would be too restrictive in any case?

I don't like the idea of adding a "reset" option, since it just adds to the API surface area to work around something in a way that is not intuitive without reading the docs.

One option could be that we, somehow, break the inheritance at isolated scope boundaries. Which means that the ngModelOptions will not impact on components that will have complete control over their ngModelOptions values in their templates. I think this could work, although if users could still update values in the global `$modelOptions` service then we would potentially have the same problem.

If we cannot find a reasonable solution to this then I agree that we should consider reverting for 1.6.0.

gkalpak:
Another idea would be making the inheritance opt-in. E.g. keep `ng-model-options` non-inheriting (as it was before 1.6.0-rc.0) and add a new `ng-extend-model-options` (or something)  that does exactly the same except for inheriting (i.e. as it is in 1.6.0-rc.0).

Internally, we will be using the same code with a flag for inheriting vs not inheriting.

petebacondarwin:
Hmm, @gkalpak, I kind of like it but it makes me feel bit queasy...
Perhaps rather than a new directive we just added a property to the ngModelOptions bag, such as `inherit: true`?

So then you would have:

```
<form ng-model-options="{ debounce: true }">
  <div ng-model-options="{ inherit:true, allowInvalid: true }">
    <input ng-model="o.x"> <!-- this will have { debounce:  true, allowInvalid: true } -->
    <input ng-model-options="{ updateOn: 'keypress' }" ng-model="o.y"> <!-- this will only have { updateOn: 'keypress' } -->
  </div>
</form>
```

Would that work??

Narretz:
I'm not a fan of putting another option in the bag. The options affects the ngModel directive, not how the ngModelOptions directive works. It also doesn't make much sense to me that you would have to opt in into the inheritance, because that would defeat the purpose of having to write *less* code.

gkalpak:
Yeah, I am not a big fan of the extra option either. All other options refer to `ngModel`, but `inherit` would refer to `ngModelOptions`. Also what would happen when you inherit `inherit: false` 😛 

> It also doesn't make much sense to me that you would have to opt in into the inheritance, because that would defeat the purpose of having to write less code.

Typically, you can save several characters via inheritance, so an extra few character (like `-extend` or `-ext`) doesn't defeat the purpose imo. Most importantly, inheritance comes in handy when you want to have consistent behavior across several instances without having to repeat the same code, which makes it easier to make changes in one place and have them propagate to children (via inheritance). This attribute of the feature remains unaffected: You only write `-extend` once and then any change you make to an ancestor `ngModeOptions` propagates automatically.

petebacondarwin:
`inherit: false` would not inherit :-P

There was a request to be able to set defaults for `ngModelOptions` at an application level - but this would also break the use case given by @fenduru.

The original motivation behind making `ngModelOptions` inheritable was to support potential changes to `ngModel` regarding how validators and parsers/formatters work. I believe that it is unlikely that we will implement these changes. (Instead it is more likely that we could offer a completely new forms library that people could opt into).

So I think it is simplest just to revert this feature.

WDYT?

fenduru:
> One option could be that we, somehow, break the inheritance at isolated scope boundaries.

This is something that I have desired in the past to prevent transcluded content from being able to require controllers beyond a boundary (I forget the exact use case now though). Even if it is possible, this sort of things seems too high risk to be worth it IMO.

> Another idea would be making the inheritance opt-in. E.g. keep ng-model-options non-inheriting (as it was before 1.6.0-rc.0) and add a new ng-extend-model-options (or something)

This seems like probably the best way of accomplishing this goal. I agree it feels weird, but it also seems the most technically correct / least risky way of doing this.

> I'm not a fan of putting another option in the bag. The options affects the ngModel directive, not how the ngModelOptions directive works.

I agree philosophically here, but wouldn't hate this if it made the ergonomics better - though I think `ng-extend-model-options` is probably just as ergonomic

I'm a big fan of opt in solutions here. Currently a user can accomplish something very similar by putting a function on their scope and writing `ng-model-options="extendDefaults({ debounce: 100 })"` - this might be slightly annoying to do if you have many different scope that want this, but perhaps `ng-model-options` could pass a local function to the expression? So `ng-model-options="$extendDefaults({ debounce: 100 })"` (or some better name - naming is hard).

Narretz:
There's a valid use-case for this feature, so I'm not happy with simply reverting it. Introducing a new directive / behavior that achieves inheritance is fine with me. It should be easy to replace ng-model-options with it, so maybe `ng-model-options-extend` or `ng-model-options-inherit`

PR #10926, opened 2015-02-01
fix ($compile): keeps prototype properties for template URL directives

hannahhoward:
currently, if a directive definition objects has prototype properties, such as a compile function, those are lost when compiled if the directive has a templateUrl function. Steps to reproduce:

``` js
directive("templateUrlDoesntWork", function() {
   var base = {
      compile: function() { 
         return function(scope, element, attrs) {
            scope.value = "Test"
         }
     }
  }
  inherited = Object.create(base)
  inherited.templateUrl = "test.html"
  return inherited;
}
```

The DDO's compile function will never be called. On the other hand the same version with a plain template works:

``` js
directive("templateWorks", function() {
   var base = {
      compile: function() { 
         return function(scope, element, attrs) {
            scope.value = "Test"
         }
     }
  }
  inherited = Object.create(base)
  inherited.template = "<p>{{value}}<p>"
  return inherited;
}
```

The DDO's compile function will be called.

This makes a difference for an increasing number of people particularly who are experimenting with ES6 classes in Angular 1.x.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2155,7 +2155,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           afterTemplateChildLinkFn,
           beforeTemplateCompileNode = $compileNode[0],
           origAsyncDirective = directives.shift(),
-          // The fact that we have to copy and patch the directive seems wrong!
           derivedSyncDirective = inherit(origAsyncDirective, {
             templateUrl: null, transclude: null, replace: null, $$originalDirective: origAsyncDirective
           }),
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2156,7 +2156,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           beforeTemplateCompileNode = $compileNode[0],
           origAsyncDirective = directives.shift(),
           // The fact that we have to copy and patch the directive seems wrong!
-          derivedSyncDirective = extend({}, origAsyncDirective, {
+          derivedSyncDirective = inherit(origAsyncDirective, {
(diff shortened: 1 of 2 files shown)


caitp:
i could see this being useful for typescript/coffeescript users, and it is a pretty small fix. lgtm, but wait for pete or someone else to signoff on it

caitp:
Marking this for beta 3, but this should probably wait for beta 4. /cc @petebacondarwin

petebacondarwin:
LGTM - let's get this in next week.  Thanks for taking a look @caitp and pinging me.

lgalfaso:
looks good, the only thing I would like to see is the same test using `template`. I know that this test would pass even without the PR, but having a test in place would reassure that this does not break in the future

petebacondarwin:
Quick work @hannahhoward - nicely done and nice PR. Thanks

hannahhoward:
added second test per request. and removed the accidental leftover console.log. I'm not sure why the build is failing on Firefox 31.0 on Linux

src/ng/compile.js
```diff
@@ -2156,7 +2156,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           beforeTemplateCompileNode = $compileNode[0],
           origAsyncDirective = directives.shift(),
           // The fact that we have to copy and patch the directive seems wrong!
```
realityking: This comment doesn't reflect the code below it anymore.

hannahhoward:
removed the outdated comment

lgalfaso:
Looks very good. Just one more thing @hannahhoward can you please squash everything into one commit? Thanks!

petebacondarwin:
Actually @hannahhoward don't worry, we can squash when we merge.

PR #10941, opened 2015-02-03
perf(filterFilter): change loop validation

HenriqueLimas:
The filterFilter uses the filter method from Array but it seems that the forLoop is more performatic than the native methods.

jsPerf: http://jsperf.com/aarontgrogg-array-filter-vs-for-loop

### DIFF
--- a/src/ng/controller.js
+++ b/src/ng/controller.js
@@ -69,6 +69,8 @@ function $ControllerProvider() {
      *    * check if evaluating the string on the current scope returns a constructor
      *    * if $controllerProvider#allowGlobals, check `window[constructor]` on the global
      *      `window` object (not recommended)
+     *    * if use controllerAs syntax, the controller instance is published into a scope property.
+     *      (scope must be injected into locals param)
      *
      *    The string can use the `controller as property` syntax, where the controller instance is published
      *    as the specified property on the `scope`; the `scope` must be injected into `locals` param for this
--- a/src/ng/controller.js
+++ b/src/ng/controller.js
@@ -70,7 +70,7 @@ function $ControllerProvider() {
      *    * if $controllerProvider#allowGlobals, check `window[constructor]` on the global
      *      `window` object (not recommended)
      *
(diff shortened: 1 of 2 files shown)


lgalfaso:
Hi, thanks for the PR. Can you please
- [ ] Create a test that shows some number with the improvement
- [ ] Squash everything into one commit

jbedard:
I've always wondered if it would be worth avoiding the creation of the array if nothing is filtered out (and return the original). Since this is often done per digest it might be worth it...

HenriqueLimas:
@lgalfaso how can I create perfomance test in Angular? With "benchpress"?

HenriqueLimas:
@lgalfaso, I made another JsPerf to demonstrate the difference. http://jsperf.com/native-array-filter-vs-forloop

@jbedard the methods filter, create a new array and return that new array under the hood. You can see that in http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.20

jbedard:
@HenriqueLimas that's how the ES5 array filter works, but I'm saying we could do it differently if we no longer use it.  I assume the filterFilter often does nothing but clone the (potentially large) array. When done every digest this might cause a lot of unnecessary GC. Just an idea though...

For a perf test I think you'd want to test the full filter function, not only ES5 filter vs for-loops. Everyone knows for-loops are faster then ES5 methods, but is it noticeable in this case? In many use cases `deepCompare` will be far more expensive then the ES5 filter function (especially when the ES5 `some` is called *per* array entry).

HenriqueLimas:
@jbedard, yeah, it's correct. I made this test http://jsperf.com/angular-filter-arrayfilter-vs-forloop and it's seems some times for loop it's better and sometimes native filter it's better.

jbedard:
Seems like it (relatively) made no difference then, so I'd say such a change isn't worth the extra code at this point...

HenriqueLimas:
Yes @jbedard , it's better and more easy to read than the forLoop. I think we can cancel this pull request. Thanks.

PR #10948, opened 2015-02-03
fix($parse): handle null targets at assign

lgalfaso:
When assigning to a sub property of a property that its value is `null` then write an empty object to the property that will contain the sub-property

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -926,7 +926,7 @@ ASTCompiler.prototype = {
           self.if(self.stage === 'inputs' || 's', function() {
             if (create && create !== 1) {
               self.if(
-                self.not(self.getHasOwnProperty('s', ast.name)),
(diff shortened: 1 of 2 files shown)


jbedard:
Just checking that the value is falsy will also cause false/0/''/... to be replaced with `{}`. Is that correct? Seems weird when `0` will be replaced but `1` wouldn't...

lgalfaso:
@jbedard this follows the same logic that we had before, and the change in the logic is causing some regressions. We need this as this regression is blocking the next beta. Later we can tweak the logic to whatever we want.

jbedard:
Yea it is the same as before, so lgtm. Might be something worth changing in 1.4 though...

lgalfaso:
landed as 2e5a7e52a0385575bbb55a801471b009afafeca3

PR #10950, opened 2015-02-03
docs(misc-core): Add `undefined` as in/outputs of to/fromJson

wesalvaro:
Request Type: docs

Component(s): misc-core

Impact: small

Complexity: small

Detailed Description:
It is acceptable to pass and possible to receive `undefined` when using the `angular.toJson` and `angular.fromJson` methods.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -979,7 +979,7 @@ function toJsonReplacer(key, value) {
  * Serializes input into a JSON-formatted string. Properties with leading $$ characters will be
  * stripped since angular uses this notation internally.
  *
- * @param {Object|Array|Date|string|number} obj Input to be serialized into JSON.
+ * @param {Object|Array|Date|string|number|undefined} obj Input to be serialized into JSON.
  * @param {boolean|number=} pretty If set to true, the JSON output will contain newlines and whitespace.
  *    If set to an integer, the JSON output will contain that many spaces per indentation (the default is 2).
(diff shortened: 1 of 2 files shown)


lgalfaso:
Hi, this might be a little bit personal, but instead of saying that you can pass `undefined` I would say that the parameter is optional. One way or another, if we are going to relax this, then we need some tests

wesalvaro:
I don't think I would change it to an optional parameter. If you call it without an argument, nothing will ever happen; that's the wrong way to call it. However, you can certainly pass in `undefined`, in which case you'll get `undefined` back. There are a variety of scenarios where that would be reasonable. I'll add some tests if this will be merged, no problem.

lgalfaso:
> I don't think I would change it to an optional parameter. If you call it without an argument, nothing will ever happen

that is fair

> I'll add some tests if this will be merged, no problem.

sounds good

wesalvaro:
Added test for `fromJson`, there was already a test in place for `toJson`.

mprobst:
YMMV, but I'd rather consider this a bug in the implementation - there's no point in calling the method with an undefined argument, as nothing is going to happen. So the user code calling you is most certainly wrong in some way (it's most likely expecting to parse some JSON).

So the best behaviour would be to fail early instead of silently trucking along.

wesalvaro:
Good point, Martin. This should be turned into a bug that is fixed.

After all, JSON.parse(undefined) is a SyntaxError.

PR #10952, opened 2015-02-03
fix(sanitize): handle newline characters inside special tags

### LINKED ISSUE #10943: $sanitize Bug
I have found a bug whereby script and style tags with newlines in them are not correctly removed.
Demo here: http://codepen.io/anon/pen/pvWWjz

Example Snippet:

```html
<style type="text/css">
p.p1 {margin: 0.0px 0.0px 0.0px 0.0px; font: 13.0px Times; color: #232323; -webkit-text-stroke: #232323}
</style>
<script type="script/javascript">
function bad(){return false;}
</script>

<p>This Should be the first content!!</p>
```

First discovered over at textAngular: https://github.com/fraywing/textAngular/issues/533

I'm going to investigate a bit more and I'll post up if I can figure a solution.

lgalfaso:
Handle newlines characters when they are present inside `<script>` and `<style>` tags.

Closes #10943

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -336,7 +336,7 @@ function htmlParser(html, handler) {
       }
 
     } else {
-      html = html.replace(new RegExp("(.*)<\\s*\\/\\s*" + stack.last() + "[^>]*>", 'i'),
+      html = html.replace(new RegExp("([^]*)<\\s*\\/\\s*" + stack.last() + "[^>]*>", 'i'),
         function(all, text) {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I would feel more comfortable if we added two new (with newlines) tests rather than modifying the current ones.  Other than that LGTM.

lgalfaso:
Split the checks into new tests and landed as
1.4: cc8755cda6efda0b52954388e8a8d5306e4bfbca
1.3: 11aedbd741ccddba060a9805adba1779391731da

shahata:
@lgalfaso Looking at Travis, I think this broke the build on IE9 and IE10

lgalfaso:
this should be fixed as #10976 landed

PR #10970, opened 2015-02-04
fix($parse): Initialize elements in an array from left to right

### LINKED ISSUE #10968: $parse regression in 1.4.0-beta.3
```
var s = {c:function() {return {b:1}}};

// this works as before
$parse("[a=c(),d=a.b+1]")(s)
// [ { b: 1 }, 2 ]
s
// { c: [Function], a: { b: 1 }, d: 2 }

// this used to give the same result as above
$parse("e=1;[a=c(),d=a.b+1]")(s)
// [ { b: 1 }, 1 ]
s
// { c: [Function], e: 1, a: { b: 1 }, d: 1 }
```

lgalfaso:
When constructing an array, never lazy initialize the elements and build the array strictly from left to right.

Closes: #10968

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -996,7 +996,7 @@ ASTCompiler.prototype = {
           self.if(self.notNull(right), function() {
             self.addEnsureSafeFunction(right);
             forEach(ast.arguments, function(expr) {
-              self.recurse(expr, undefined, undefined, function(argument) {
+              self.recurse(expr, self.nextId(), undefined, function(argument) {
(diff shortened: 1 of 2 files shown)


jbedard:
LGTM, although I think it would be worth adding tests for all 3 cases (func args, arrays, objects)...

lgalfaso:
@jbedard even when there is no such constraint for object literals (there is for argument evaluation in function calls), I added the tests and landed this as 966f6d831f9469a917601f9a10604612cd7bd792

PR #10976, opened 2015-02-05
fix(sanitize): handle newline characters inside <script> for IE

lgalfaso:
Tweak the regex used match characters inside <script> and <style> to a IE compatible regex.

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -336,7 +336,8 @@ function htmlParser(html, handler) {
       }
 
     } else {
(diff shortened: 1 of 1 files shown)


petebacondarwin:
I thought the normal way to match anything is `[\W\w]*`. Does that work with IE <11?
I would rather not be sniffing this.

petebacondarwin:
Or I think I meant `[\s\S]`

lgalfaso:
@petebacondarwin let me try

lgalfaso:
@petebacondarwin Travis failed as there was some issue with SourceLabs, but the relevant test now passed

lgalfaso:
this is breaking Travis, so landing this 030a42e79dec8a4bb73053762f7a54d797a058f6

PR #10983, opened 2015-02-05
Watch the ngInit so that if it changes it can be re-evaluated.

rwlogel:
If `ng-init` is used for creating a variable alias it currently only evaluates that alias when it is created.  For example:

``` html
<div ng-init="model=some.long.variable.name.object">
    <input ng-model="model.foo">
    <input ng-model="model.bar">
</div>
```

If `some.long.variable.name.object` changes then model won't be updated.  This change will ensure that if it changes it will be re-evaluated.

### DIFF
--- a/src/ng/directive/ngInit.js
+++ b/src/ng/directive/ngInit.js
@@ -62,6 +62,11 @@ var ngInitDirective = ngDirective({
     return {
       pre: function(scope, element, attrs) {
         scope.$eval(attrs.ngInit);
+      },
+      post: function(scope, element, attrs) {
+        scope.$watch(attrs.ngInit, function() {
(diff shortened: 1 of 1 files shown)


rwlogel:
CLA Signed

rwlogel:
This feature isn't going to work as I expected because something like this triggers an infinite loop:

```html
<div ng-init="model= {foo: 1, bar: 2}">
    <input ng-model="model.foo">
    <input ng-model="model.bar">
</div>
```
Everytime it is evaluated the reference for scope.model changes, which triggers another evaluate.

In our project we ended up just creating a directive we called `gsAlias` which works like ng-init except only for reference assignment.

PR #10985, opened 2015-02-05
fix(ngTouch): don't prevent the click event after a touchmove

mikec:
Remove the touchmove handler so that resetState is not called on touchmove. The touchend event handler already prevents the click from being triggered if the distance moved exceeds the MOVE_TOLERANCE, so detection of touchmove is not needed. Previously, because of resetState on touchmove, the click would not be triggered even if the event coordinates changed by only 1px or 2px, which seems to be very common for taps on mobile browsers.


pomerantsev:
+1

dheidebrecht:
:+1: we had to patch a local version of this to fix issues on iOS.

PR #10996, opened 2015-02-06
fix(ngTouch): check undefined tagName for SVG event target

mikec:
When target click element is an SVG tag, event.target.tagName and event.target.blur are undefined on Chrome v40 and iOS 8.1.3

### DIFF
--- a/src/ngTouch/directive/ngClick.js
+++ b/src/ngTouch/directive/ngClick.js
@@ -1,6 +1,8 @@
 'use strict';
 
-/* global ngTouch: false */
+/* global ngTouch: false,
+  nodeName_: false
+*/
 
 /**
  * @ngdoc directive
@@ -142,7 +144,7 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
       lastLabelClickCoordinates = null;
     }
     // remember label click coordinates to prevent click busting of trigger click event on input
-    if (event.target.tagName.toLowerCase() === 'label') {
+    if (nodeName_(event.target) === 'label') {
       lastLabelClickCoordinates = [x, y];
     }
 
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
Thnx for the PR @mikec - it looks good on the first look. I was wondering, though, if there would be a way to simplify a test  and or move some functionality to a helper. In the essence this issue is the same as the one fixed in 09a9832358960c98392c9df1a9fd9592f59bc844 so maybe we should be using the same helper? WDYT?

mikec:
@pkozlowski-opensource: I simplified the test. It still triggers the click busting event handler, and should fail if tagName is undefined.

I like the idea of reusing the `nodeName_` helper but I'm not sure that function can be used from the `ngTouch` module unless we expose it in `AngularPublic.js`. Is there another way to reuse this code, or should the helper be exposed to the public API?

I could also copy the code from `nodeName_` in place of the current tagName check, or copy it into a helper function within the `ngTouch` module. What do you think is the best approach?

pkozlowski-opensource:
@mikec thnx for updating this PR. Yeh, with `nodeName_` being non-public creating a separate method inside `ngTouch` module sounds like a best option...

I'm leaving one more comment in the PR, would be awesome to have your input. Otherwise this one is almost good to go in.

mikec:
@pkozlowski-opensource: updated with `nodeName_` function added to the touch module.

src/ngTouch/directive/ngClick.js
```diff
@@ -158,7 +160,7 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
     event.preventDefault();
 
     // Blur focused form elements
-    event.target && event.target.blur();
+    event.target && event.target.blur && event.target.blur();
```
mikec: @pkozlowski-opensource, this is related to the SVG issue. On Chrome v40 and iOS 8.1.3, a lot of properties on the SVG element are undefined, including `tagName

petebacondarwin:
Landed as 74eb17d7c8232f72f134bf2546f10fed7234d276

PR #11000, opened 2015-02-06
Update controller.ngdoc fixes grammar

rwuebker:
Improper use of the word 'which'.  

BEFORE: The following example uses a Controller to add a method to the scope, which doubles a number:
AFTER: The following example uses a Controller to add a method, which doubles a number, to the scope:


rwuebker:
Hello,

I have signed the agreement.

Thanks,
Rick

PR #11005, opened 2015-02-07
feat(ngModel): support conversion to timezone other than UTC

shahata:


petebacondarwin:
Looks good. We need to update the ngModelOptions docs too.

See https://docs.angularjs.org/api/ng/directive/ngModelOptions

shahata:
Sure. Docs added.

petebacondarwin:
LGTM - I am going to combine `fromTimezoneToLocal` with `fromLocalToTimezone` into a single `convertTimezoneToLocal(dateTime, timezone, reverse)` and merge

PR #11006, opened 2015-02-07
updated copy to throw an error if attempting to copy an element. if copy...

micah-williamson:
...ing an object that contains an element, the element isnt copied recursively. watch does not copy an object for the last value if it is an element.


iamchairs:
Updated CLA.

gkalpak:
This has been fixed on 1.5. `copy()` now uses `.cloneNode(true)`, so elements are copied correctly.

PR #11007, opened 2015-02-08
fix(ngAria): correctly set "checked" attr for checkboxes and radios

### LINKED ISSUE #10212: ngAria - incorrect "aria-checked" and "tabindex" applied to checkbox/radio
Hello angular team!

**Overview of the Issue** - incorrect attributes `aria-checked` and `tabindex` are applied to the input of type `radio`: when radio is selected, the following attributes are applied: `aria-checked="false"`, `tabindex="-1"`.
**Motivation for or Use Case** - radio button never gets selected when pressing *Tab*
**Angular Version(s)** - 1.3.1 for all modules
**Browsers and Operating System** - Chrome 38.0.2125.111 m. Haven't tested in other browsers, but pretty sure it'll be the same.
**Reproduce the Error** - see [Plunker](http://plnkr.co/edit/liOsgkByxFgPeqF7iuyc?p=preview).
**Related Issues** - not found
**Suggest a Fix** - Switching attribute `value` to `ng-value` fixes the issue, but it is inconsistent through out the angular's logic - it was able to figure out that the radio should be selected with the regular `value` attribute, but it could not figure out correct accessibility attributes. It appears the issue has to do with the following [lines of code](https://github.com/angular/angular.js/blob/5b23bc9b07f8a4363c9c549a248f690b1a31b7be/src/ngAria/aria.js#L232-L233)

### LINKED ISSUE #10389: bug(ngAria): Use of ng-true-value and ng-false-value not updating aria-checked
Checkbox components using `ng-true-value` and `ng-false-value` are not computing the correct `aria-checked` value. 

See this thread from Angular Material: https://github.com/angular/material/pull/882#issuecomment-66353514

Example code:
```
<md-switch ng-model="data.cb2" aria-label="Switch 2" ng-true-value="'yup'" 
ng-false-value="'nope'" role="checkbox" tabindex="0" 
aria-checked="true" aria-invalid="false">
</md-switch>
```

When the checkbox is toggled to the off position using these attributes, the `aria-checked` property is set to `true`. It should be false.

Narretz:
Make sure the checked attribute is set correctly for:
- checkboxes with string and integer models using ngTrueValue /
  ngFalseValue
- radios with integer models
- radios with boolean models using ngValue

Fixes #10389
Fixes #10212

@marcysutton can you take a look at this?

### DIFF
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -85,6 +85,28 @@ describe('$aria', function() {
       expect(element.attr('aria-checked')).toBe('false');
     });
 
+    it('should handle checkbox with string model values using ng(True|False)Value', function() {
+      var element = $compile('<input type="checkbox" ng-model="val" ng-true-value="\'yes\'" ' +
+        'ng-false-value="\'no\'">'
+      )(scope);
+
+      scope.$apply('val="yes"');
+      expect(element.eq(0).attr('aria-checked')).toBe('true');
+
+      scope.$apply('val="no"');
+      expect(element.eq(0).attr('aria-checked')).toBe('false');
+    });
+
+    it('should handle checkbox with integer model values using ngTrueValue', function() {
+      var element = $compile('<input type="checkbox" ng-model="val" ng-true-value="0">')(scope);
+
+      scope.$apply('val=0');
+      expect(element.eq(0).attr('aria-checked')).toBe('true');
+
+      scope.$apply('val=1');
+      expect(element.eq(0).attr('aria-checked')).toBe('false');
+    });
+
     it('should attach itself to input type="radio"', function() {
       var element = $compile('<input type="radio" ng-model="val" value="one">' +
           '<input type="radio" ng-model="val" value="two">')(scope);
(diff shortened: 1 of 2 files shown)


test/ngAria/ariaSpec.js
```diff
@@ -85,6 +85,28 @@ describe('$aria', function() {
       expect(element.attr('aria-checked')).toBe('false');
     });
 
+    it('should handle checkbox with string model values using ng(True|False)Value', function() {
+      var element = $compile('<input type="checkbox" ng-model="val" ng-true-value="\'yes\'">' +
+        'ng-false-value="\'no\'"'
```
marcysutton: Should this attribute be part of the input element?

Narretz: Yes, definitely! although it doesn't make a difference, since we don't test if the correct ng-false-value is set.

marcysutton:
This looks good to me. Nice work on the test coverage! :+1:

Narretz:
Landed as d6eba21733c6e67e90e3a4763d8d41ad89a73a0c and 944c150e6c3001e51d4bf5e2d8149ae4c565d1e3!

PR #11012, opened 2015-02-09
feat(ngModel, form): add 'reset all errors' functionality

vvasianovych:
This PR contains new method for form and inputs (controllers) that allows to reset all errors added with $setValidity method.
Use case:
1. User fills form and click submit button. Application sends request to the server. Server validates this form somehow. Server decided that field 'A' is invalid.
2. Application calls $setValidity method for the input 'A' to mark it as ng-invalid-*
3. User changes field 'A' and click submit button.
4. At this time server decided that in this condition user specified wrong value in the field 'B'.
At this time, application can't reset fields marked with $setValidity all together.

That's the functionality I tried to cover with this PR.

plunker: http://plnkr.co/edit/aev16Y3cAxTljhctUDMI

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -7,6 +7,7 @@ var nullFormCtrl = {
   $$renameControl: nullFormRenameControl,
   $removeControl: noop,
   $setValidity: noop,
+  $resetErrors: noop,
   $setDirty: noop,
   $setPristine: noop,
   $setSubmitted: noop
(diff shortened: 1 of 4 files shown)


neoshka:
:+1:

petebacondarwin:
Can you clarify what the motivation for clearing **all** the errors on the form is?

If you have an external process specifying validation errors on controls then shouldn't that same process be responsible for resetting the errors that it sets?

Clearing all the errors in a form seems like a heavy handed approach - what if there are other external processes at work here?

taqi:
@petebacondarwin 
It is something like 'restore form and controlls state to state before validation'. (Yes, I tried $setPristine)
In my usecase it is usefull in situation, when I have popup window with some form, and this popup will be closed only when object will be saved successfully. When user clicks 'Submit' - server validates it, and sends message about errors. 
When user refills data and clicks 'submit' again, and server found errors again, I have to reset all previous errors and ng-invalid-* classes on inputs before I set new errors. 

Does this make sense?

PR #9854, opened 2014-10-31
feat(select): ngOptions support disabling options

cades:
Extend the ngOptions DSL to support disabling options. Since <option>
elements are generated by ngOptions, we cannot use ngDisabled in this
scenario. Via extend the DSL, user have control over if a option
is diabled on arbitrary condition.

This commit insert a `disable if cond` match syntax into
NG_OPTIONS_REGEXP, change other match index accordingly, add disabled
attribute to null option element and newly cloned option elements,
add 2 new unit tests, and change docs as well.

Closes #638

PR #11017, opened 2015-02-09
feat(ngOptions): add support for disabling an option

### LINKED ISSUE #638: Support disabling options in select with ng:options
Need a way to construct

    <select>
      <option disabled>A</option>
      <option>B</option>
      <optionC</option>
    </select>

There's no way to express the disabled state of a specific option with ng:options

sjbarker:
This patch adds support for disabling options based on model values. The
"disable when" syntax allows for listening to changes on those model values,
in order to dynamically enable and disable the options.

The changes prevent disabled options from being written to the selectCtrl
from the model. If a disabled selection is present on the model, normal
unknown or empty functionality kicks in.

closes #638

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -94,14 +94,19 @@ var ngOptionsMinErr = minErr('ngOptions');
  *     * `label` **`for`** `value` **`in`** `array`
  *     * `select` **`as`** `label` **`for`** `value` **`in`** `array`
  *     * `label` **`group by`** `group` **`for`** `value` **`in`** `array`
+ *     * `label` **`disable when`** `disable` **`for`** `value` **`in`** `array`
  *     * `label` **`group by`** `group` **`for`** `value` **`in`** `array` **`track by`** `trackexpr`
+ *     * `label` **`disable when`** `disable` **`for`** `value` **`in`** `array` **`track by`** `trackexpr`
  *     * `label` **`for`** `value` **`in`** `array` | orderBy:`orderexpr` **`track by`** `trackexpr`
  *        (for including a filter with `track by`)
  *   * for object data sources:
  *     * `label` **`for (`**`key` **`,`** `value`**`) in`** `object`
  *     * `select` **`as`** `label` **`for (`**`key` **`,`** `value`**`) in`** `object`
  *     * `label` **`group by`** `group` **`for (`**`key`**`,`** `value`**`) in`** `object`
+ *     * `label` **`disable when`** `disable` **`for (`**`key`**`,`** `value`**`) in`** `object`
  *     * `select` **`as`** `label` **`group by`** `group`
+ *         **`for` `(`**`key`**`,`** `value`**`) in`** `object`
+ *     * `select` **`as`** `label` **`disable when`** `disable`
  *         **`for` `(`**`key`**`,`** `value`**`) in`** `object`
  *
  * Where:
@@ -116,6 +121,8 @@ var ngOptionsMinErr = minErr('ngOptions');
  *      element. If not specified, `select` expression will default to `value`.
  *   * `group`: The result of this expression will be used to group options using the `<optgroup>`
  *      DOM element.
+ *   * `disable`: The result of this expression will be used to disable the rendered `<option>`
+ *      element. Return `true` to disable.
  *   * `trackexpr`: Used when working with an array of objects. The result of this expression will be
  *      used to identify the objects in the array. The `trackexpr` will most likely refer to the
  *     `value` variable (e.g. `value.propertyName`). With this the selection is preserved
@@ -129,10 +136,10 @@ var ngOptionsMinErr = minErr('ngOptions');
           .controller('ExampleController', ['$scope', function($scope) {
             $scope.colors = [
               {name:'black', shade:'dark'},
-              {name:'white', shade:'light'},
+              {name:'white', shade:'light', notAnOption: true},
               {name:'red', shade:'dark'},
-              {name:'blue', shade:'dark'},
-              {name:'yellow', shade:'light'}
+              {name:'blue', shade:'dark', notAnOption: true},
+              {name:'yellow', shade:'light', notAnOption: false}
             ];
             $scope.myColor = $scope.colors[2]; // red
           }]);
@@ -141,6 +148,7 @@ var ngOptionsMinErr = minErr('ngOptions');
           <ul>
             <li ng-repeat="color in colors">
               Name: <input ng-model="color.name">
+              <input type="checkbox" ng-model="color.notAnOption"> Disabled?
               [<a href ng-click="colors.splice($index, 1)">X</a>]
             </li>
             <li>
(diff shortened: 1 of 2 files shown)


sjbarker:
Explanation
===========
The changes in this pull request reflect a long-time reoccurring and widely popular request for the ability to disable options dynamically in `ngOptions`. This solution accomplishes this feature by adding the `disable by` syntax to the ng-options directive. All current functionality is retained in select controls.

The solution for model changes reflecting disabled options is to keep the current functionality of "unknown" or "empty" options. Tests have been written to reflect this feature. Consider the following:
```javascript
scope.selected = '';
scope.options = [ 'a', 'b', 'c', 'd'];
scope.disableOption = function(option) {
  return option === 'b';
}
```
```html
<select ng-model="selected"
             ng-options="o disable by disableOption(o) for o in options"></select>
```
**-OR-**
```javascript
scope.selected = '';
scope.options = [ 
  {
    name: 'a',
    disabled: false
  },
  {
    name: 'b',
    disabled: false
  },
  {
    name: 'c',
    disabled: false
  },
  {
    name: 'd',
    disabled: false
  }
];
```
```html
<select ng-model="selected"
             ng-options="o.name disable by o.disabled for o in options"></select>
```
In either case, if selected becomes a disabled option, then select will render the empty or unknown option. This concurrent with any unknown or invalid option.

Current Issues
============
Issue #638 has been open for over three years now and is currently on backlog.

Current PR's
============
There is one current PR (#9854) by @cades that has been rendered obsolete by @petebacondarwin's `select.js` refactor (408f89d8e6b8092d4abdc0cf617487f5cbb06797).

sjbarker:
@petebacondarwin - Since you recently did the select directive refactor, will you take a quick look at this and pass it around? We have been sitting on this for a really long time and this would provide the functionality without breaking anything; and with room to grow in the future. Thanks!

cades:
Great job, thanks to your effort and time!
Hope to merge this PR as soon as possible.

sjbarker:
Thanks @cades. You were there first until they refactored the select directive. And thank you for your time and effort as well.

sjbarker:
@pkozlowski-opensource - I figured I would tag you in this since you were attending to the #9854 PR. When you have a moment, will you take a look as well? Thanks.

petebacondarwin:
OK, I like this idea, although I think that the syntax sound better with `disable when` rather than `disable by`.

Also I am concerned about add to the `getWatchables` array, which is already a bit of a pain point.

I realised that we can actually optimize this by only including things that are being specified. So for instance, if no `disable when` clause was provided we would not need to watch this expression. The same for the `displayFn` which computes the label.

sjbarker:
No problem, @petebacondarwin. Let me get those changes in real quick. I will open an issue for optimizing the `getWatchables` array and cover it in this commit as well.

src/ng/directive/ngOptions.js
```diff
@@ -186,16 +200,17 @@ var ngOptionsMinErr = minErr('ngOptions');
  */
 
 // jshint maxlen: false
-                         //000011111111110000000000022222222220000000000000000000003333333333000000000000004444444444444440000000005555555555555550000000666666666666666000000000000000777777777700000000000000000008888888888
-var NG_OPTIONS_REGEXP = /^\s*([\s\S]+?)(?:\s+as\s+([\s\S]+?))?(?:\s+group\s+by\s+([\s\S]+?))?\s+for\s+(?:([\$\w][\$\w]*)|(?:\(\s*([\$\w][\$\w]*)\s*,\s*([\$\w][\$\w]*)\s*\)))\s+in\s+([\s\S]+?)(?:\s+track\s+by\s+([\s\S]+?))?$/;
+//                     //00001111111111000000000002222222222000000000000000000000333333333300000000000000000000000004444444444400000000000005555555555555550000000006666666666666660000000777777777777777000000000000000888888888800000000000000000009999999999
+var NG_OPTIONS_REGEXP = /^\s*([\s\S]+?)(?:\s+as\s+([\s\S]+?))?(?:\s+group\s+by\s+([\s\S]+?))?(?:\s+disable\s+when\s+([\s\S]+?))?\s+for\s+(?:([\$\w][\$\w]*)|(?:\(\s*([\$\w][\$\w]*)\s*,\s*([\$\w][\$\w]*)\s*\)))\s+in\s+([\s\S]+?)(?:\s+track\s+by\s+([\s\S]+?))?$/;
```
sjbarker: @petebacondarwin Here is the syntax change.

src/ng/directive/ngOptions.js
```diff
@@ -269,8 +286,11 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
           var locals = getLocals(values[key], key);
           var label = displayFn(scope, locals);
           var selectValue = getTrackByValue(values[key], locals);
+          var disableWhen = disableWhenFn(scope, locals);
           watchedArray.push(selectValue);
-          watchedArray.push(label);
+          // for optimization, do not watch label or disableWhen if they weren't defined
```
sjbarker: @petebacondarwin and here is the check for expression definition prior to adding watchers

src/ng/directive/ngOptions.js
```diff
@@ -269,8 +286,11 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+          if (match[2]) watchedArray.push(label);
+          if (match[4]) watchedArray.push(disableWhen);
```
sjbarker: @petebacondarwin and here is the check for expression definition prior to adding watchers

petebacondarwin: We can do more here. We don't need to even compute the `label` or `displayWhen` values if they are not specified. See https://github.com/angular/angular.js/pull/11052/files

sjbarker: Agreed. Let me push that.

src/ng/directive/ngOptions.js
```diff
@@ -275,6 +292,12 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
             var label = displayFn(scope, locals);
             watchedArray.push(label);
           }
+
+          // Only need to watch the disableWhenFn if there is a specific disable expression
+          if (match[4]) {
+            var disableWhen = disableWhenFn(scope, locals);
+            watchedArray.push(disableWhen);
+          }
```
sjbarker: @petebacondarwin - I rebased your perf and mimicked it with the disableWhenFn

src/ng/directive/ngOptions.js
```diff
@@ -326,7 +350,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
   return {
     restrict: 'A',
     terminal: true,
-    require: ['select', '?ngModel'],
+    require: ['select', 'ngModel'],
```
petebacondarwin: ngModel is still optional here.

sjbarker:
@petebacondarwin I returned the `ngModel` requirement to optional.

petebacondarwin:
I added a few extra tests and merged into master - thanks @sjbarker and @cades !

sjbarker:
Thanks @petebacondarwin! I didn't see that `ngModel` optional requirement make it in though. Is that ok?

petebacondarwin:
Oops

sjbarker:
@petebacondarwin - f3ae91dbd2c5813b9c2c3837bb39543b53bf3d9d here has the commit you just merged with the `ngModel` require change.

petebacondarwin:
Thanks - I fixed that here: https://github.com/angular/angular.js/commit/ef894c87eaead76d90169113ab6acc9287654ea3

sjbarker:
Ok perfect

IngoVals:
If the disabling requirements are dynamic and something that was selected becomes disable the value is set to null, could we intercept this event and change it to something else?

sjbarker:
@IngoVals when you say change it to something else, what do you mean?

PR #11018, opened 2015-02-10
remove scare quotes and so-called

chrisranderson:
So-called is defined as "commonly named" or "falsely or improperly so named". The scare quotes are definitely unnecessary. It makes it sound like things aren't actually called that, or it hints at sarcasm: "He's the so-called 'mayor', but he never does anything!" 

http://en.wikipedia.org/wiki/Scare_quotes


chrisranderson:
I took care of the CLA.

PR #11021, opened 2015-02-10
refactor(interpolate): remove redundant $watchGroup param

marcin-wosinek:


pkozlowski-opensource:
@marcin-wosinek thnx, this looks like a good move but I think that we can go a step further - left a comment in the code. WDYT?

PR #11023, opened 2015-02-10
fix(form): set $submitted to true on child forms when parent is submitted

### LINKED ISSUE #10071: $submitted is not set on sub ng-form
Hi,

I have &lt;form&gt; with inner &lt;ng-form&gt; 
when form is submitted inner &lt;ng-form&gt;.$submitted is not set to true only parent

ckniffen:
FormController will now call $setSubmitted on its child forms when it is submitted.

Closes #10071

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -295,10 +295,17 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
    * @description
    * Sets the form to its submitted state.
    */
-  form.$setSubmitted = function() {
+  form.$setSubmitted = function(setOnChildren) {
     $animate.addClass(element, SUBMITTED_CLASS);
     form.$submitted = true;
-    form.$$parentForm.$setSubmitted();
+    if (!setOnChildren) {
+      form.$$parentForm.$setSubmitted();
+    }
+    forEach(controls, function(control) {
+      if (control.$setSubmitted) {
+        control.$setSubmitted(true);
+      }
+    });
   };
 }
 
--- a/test/ng/directive/formSpec.js
+++ b/test/ng/directive/formSpec.js
@@ -523,6 +523,22 @@ describe('form', function() {
       expect(parent.$submitted).toBeTruthy();
     });
 
+    it('should set $submitted equal to true on child forms when parent is submitted', function() {
+      doc = jqLite(
+          '<ng-form name="parent">' +
+            '<ng-form name="child">' +
+              '<input ng:model="modelA" name="inputA">' +
+              '<input ng:model="modelB" name="inputB">' +
+            '</ng-form>' +
+          '</ng-form>');
+      $compile(doc)(scope);
+
+      var parent = scope.parent,
+          child = scope.child;
+
+      parent.$setSubmitted();
+      expect(child.$submitted).toBeTruthy();
+    });
 
     it('should deregister a child form when its DOM is removed', function() {
       doc = jqLite(


src/ng/directive/form.js
```diff
@@ -276,10 +276,17 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
    * @description
    * Sets the form to its submitted state.
    */
-  form.$setSubmitted = function() {
+  form.$setSubmitted = function(reverse) {
```
Narretz: `reverse` is not really a good name. How about `setOnChild` or something similar?

ckniffen:
Replying because googlebot wants me to confirm I am indeed the author.

scottty881:
+1

ckniffen:
@Narretz what is the status of this issue?  Is there anything I can do to move this along?

kentcdodds:
:+1: any chance this could be made available in 1.3.x? Or at least 1.4.x (rather than just 1.5.x) :-)

benoror:
:+1:

Narretz:
1.4.x yes, 1.3.x no. I just wonder if it can be considered a breaking change

ckniffen:
One possible middle ground would be exposing the array of controls.  That way they can be iterated over in a third party plugin.

test/ng/directive/formSpec.js
```diff
@@ -523,6 +523,22 @@ describe('form', function() {
       expect(parent.$submitted).toBeTruthy();
     });
 
+    it('should set $submitted equal to true on child forms when parent is submitted', function() {
+      doc = jqLite(
+          '<ng:form name="parent">' +
```
Narretz: Can you please use `ng-form` et. al.? I think these tests are very old, that's why they use `ng:form`.

src/ng/directive/form.js
```diff
@@ -295,10 +295,17 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
    * @description
    * Sets the form to its submitted state.
    */
-  form.$setSubmitted = function() {
+  form.$setSubmitted = function(setOnChildren) {
```
gkalpak: I'd name it something more descriptive (e.g. `propagateToChildrenOnly` or `skipPropagationToParent` or something).
(`setOnChildren` made me think we'll always update the parent and we're wondering about the children.)

We should also improve the description, mentioning that it will propagate to parent and children.


Should the arg be public (aka documented).

ckniffen:
@gkalpak I looked into the the docs for the similar methods on a FormController.  The current documentation for $setSubmitted says nothing about propagating.  I think there is some confusion, at least for me, around the reasoning for why different methods propagate different.

$setDirty - goes up 
$setSubmitted - goes up
$setPristine - goes down
$setUntouched - goes down

To me $setDirty makes sense to propagate up because if one child control or form is dirty then the parents are too.  $setSubmitted has a similar use case as to $setPristine and $setUntouched because often times a form is made up of many nested forms which submit as one.

One alternative to making this change as I have mentioned above is to expose the `controls` array as `$controls` so that consumers of this api can write directives that are able to safely iterate over a FormController's child controls.  Such a change could be leveraged by formly for example.

test/ng/directive/formSpec.js
ckniffen: @Narretz Should I update all the tests in this file to use `ng-form`?

gkalpak:
@ckniffen, I wonder myself why `$setSubmitted` propagates the way it does (I don't find it intuitive, but I rarely make use of it anyway). I would be open to changing its behaviour, but it will be a breaking change so it has to be made before 1.5.1 (if at all).
Feel free to open an issue about it and see what people think.

Exposing `$controls` should definitely be helpful in certain scenarios, but could also be _abused_ leading to weird and hard to debug bugs. I'm undecided about this, but would happily review a PR :)

Narretz:
There's no (big) rush to get this in. We'll have a few 1.5.0-beta.x releases before the 1.5.0 release (which should be the first not to contain any breaking changes).

pbrain19:
will we be able to merge this before Obama leaves office?

Narretz:
Let's continue discussion in #15778. I'm confident we'll resolve this before Trump is impeached.

Issue #11016, opened 2015-02-09
ng-jq on <html> gives error


alexlamsl:
So I follow the example in the documentation to try `ng-jq` out, and as far as I can tell, `getNgAttribute()` is called within `jq()`, which in turn uses `jqLite`, which hasn't been defined yet.

Narretz:
Hi, what kind of error is thrown?

alexlamsl:
Tested on `angular.js 1.4.0-beta.4` and IE11, using example HTML:
```
<html ng-app="app" ng-jq>
...
</html>
```
gives the following within `Developer Tools`:
```
'jqLite' is undefined
```
showing the reference inside `function getNgAttribute()`

Narretz:
Yep, that's broken: http://plnkr.co/edit/dRlqiPdHKU6xchecBVnE?p=preview

@mboudreau @lgalfaso could you look into this?

mboudreau:
@alexlamsl @Narretz Yeah, seems broken.  The getNgAttribute was added later as per comments on the pull request.  What's weird is that this isn't caught in the unit tests.  I would of imagined they would of failed entirely in the unit tests, but I guess by that point, Angular is already loaded and jqLite already initialized.

Let me fix this and put in a new pull request ASAP.

mboudreau:
@alexlamsl @Narretz It's been taken cared of, shouldn't cause any issues, but definitely needs to add better tests around it.  Just waiting for the pull request to be merged in and hopefully go out in the next beta revision.

Thanks for letting me know on the issue.

petebacondarwin:
This fix was landed via #11044 - see http://plnkr.co/edit/Xa7P4JI0EAJtnVwDS5mX?p=preview

mboudreau:
@petebacondarwin You da man.

PR #11027, opened 2015-02-10
Fixing ng-jq bug with using getNgAttribute before jqLite variable is set

mboudreau:
The unit tests wasn't able to catch this unless it's on a 'fresh' angular page with ng-jq already added before angular bootstrapping.  I've removed the need for jq() to rely on getNgAttribute and instead using element.getAttribute function with the found prefix.

Could use some direction on how to test ng-jq more thoroughly so this doesn't happen again in the future.

This fixes issue #11016.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1250,7 +1250,7 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     return value;
   });
 
-  if (attr.min || attr.ngMin) {
+  if (isDefined(attr.min) || attr.ngMin) {
     var minVal;
     ctrl.$validators.min = function(value) {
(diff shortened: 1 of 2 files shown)


mboudreau:
Lots of timed out tests again.  Can someone see what's going on please?

Narretz:
I don't know about the timed out tests, but maybe it helps if you clean up the PR (put the fix on a new branch or make sure it sits exactly on top of master, so these other commits don't get in the way.

Narretz:
And I think if you force push the branch, it will trigger another travis run.

For the tests, maybe we should do an end2end test?

mboudreau:
Sorry, I resetted my fork to the upstream/master and added the change to it, but in the process it closed this PR.  New PR was created: #11044

PR #9964, opened 2014-11-07
fix(rootScope, ngRepeat): add support for objects created with Object.create(null) for ngRepeat and $watchCollection

HeberLZ:
Basically, the hasOwnProperty method was being called from an object gotten from the user. When that object was initially created with Object.create(null), it didn't inherit from Object so it didn't have the hasOwnProperty method throwing an "undefined is not a function" error. 

This was fixed by executing hasOwnProperty.call(object, key) instead of object.hasOwnProperty(key)

PR #11028, opened 2015-02-11
fix($rootScope, ngRepeat): accomodate objects resulting from Object.create(null)

HandyAndyShortStack:
Use Object.hasOwnProperty.call when iterating over properties of user-supplied objects.
Allow for objects not inheriting from Object.prototype.

### DIFF
--- a/src/ng/directive/ngRepeat.js
+++ b/src/ng/directive/ngRepeat.js
@@ -363,7 +363,7 @@ var ngRepeatDirective = ['$parse', '$animate', function($parse, $animate) {
             // if object, extract keys, in enumeration order, unsorted
             collectionKeys = [];
             for (var itemKey in collection) {
-              if (collection.hasOwnProperty(itemKey) && itemKey.charAt(0) !== '$') {
+              if (hasOwnProperty.call(collection, itemKey) && itemKey.charAt(0) !== '$') {
                 collectionKeys.push(itemKey);
               }
             }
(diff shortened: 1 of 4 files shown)


lgalfaso:
If we want to support this, then the patch looks good, but every single time we change a function call to a `call` there is a performance implication. Given that this change is in code that is very hot, then I would like to know the opinion of other

HandyAndyShortStack:
If we decide that using `call` is too expensive, we should definitely update documentation to make it clear that users are expected to supply a `hasOwnProperty` property for any object watched in scopes or iterated over by **ngRepeat**.

lgalfaso:
Hi @HandyAndyShortStack I think something needs to be done. The main issue is that the two places that this PR modifies are super critical, and a minor change can have a big performance change.
Once again, if objects without a prototype are going to be supported, then this looks like the right fix, I just would like other opinions

Narretz:
@petebacondarwin landed a commit which implements this for ngRepeat (https://github.com/angular/angular.js/commit/7ea2c7f36ef854391df3f6b127ad42a2d5cbf1a3), but the didn't land the change for $rootScope(which was also in https://github.com/angular/angular.js/pull/9964.
Afaik, the code in ngRepeat isn't actually that hot, as it is only used for objects. But putting it in the rootScope code is probably a bad idea.

gkalpak:
@Narretz, I think @petebacondarwin landed it separately in 20fb626.

Closing as duplicate of #9964 (which has already been "taken care of").

petebacondarwin:
I felt that it was worth the performance impact.

PR #11046, opened 2015-02-12

petebacondarwin:


Narretz:
Looks good, thanks for cleaning this up!

PR #11051, opened 2015-02-13
fix(syntax): `.finally` syntax error in Android 2.3

adamdbradley:
Android 2.3 throws an `Uncaught SyntaxError: Unexpected token finally`
pointing at this line. Change `.finally` to bracket notation.


adamdbradley:
Reported here: https://github.com/driftyco/ionic/issues/3070

PR #9772, opened 2014-10-24
perf(ngmodel/compile): improve ngmodel and general compile/controller performance (#9609)

jbedard:
This was mainly focused around #9609 - bringing 1.3 up to speed with 1.2 with an ng-repeated ng-model, at least when not using jQuery.

Overall these changes make the added benchmark (and the jsfiddle in #9609) 2-3x faster and seem to be faster then 1.2 now.

The last 3 commits (refactoring controller stuff in $compile) can probably be squashed but I left them separate for now for nicer diffs.

The only functional change (that I know of) is that the NgModelController `$name` is now initialized by the directive (pre-link) instead of within the controller. I'm not sure if this an issue or considered breaking? This was already the case for the `$options` field...

PR #11059, opened 2015-02-14
perf($compile): avoid .data when fetching required controllers

jbedard:
This changes `getControllers` to try fetching controllers from `elementControllers` which in many cases can avoid calling `.data`. For a largetable-bp benchmark of `<input ng-model="foo">` elements this reduces the create step GC about 50% (which seems crazy, but that's what chrome keeps claiming...) and improves the time 5-10%.

(this was originally part of #9772)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1840,49 +1840,42 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
 
       function getControllers(directiveName, require, $element, elementControllers) {
-        var value, retrievalMethod = 'data', optional = false;
-        var $searchElement = $element;
-        var match;
+        var value;
+
         if (isString(require)) {
-          match = require.match(REQUIRE_PREFIX_REGEXP);
-          require = require.substring(match[0].length);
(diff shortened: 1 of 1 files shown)


src/ng/compile.js
```diff
@@ -1840,49 +1840,42 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+                name, directiveName);
           }
-          return value || null;
         } else if (isArray(require)) {
-          value = [];
-          forEach(require, function(require) {
-            value.push(getControllers(directiveName, require, $element, elementControllers));
-          });
+          value = new Array(i = require.length);
+          while (i--) {
```
lgalfaso: nit: I find
```js
for (i=0, ii = requite.length; i < ii; ++i) {
  ...
}
```
much easier to read (unless that for some specific reason we want to run this in reverse order)

lgalfaso:
otherwise, LGTM

jbedard:
Updated...

lgalfaso:
landed as fa0aa83937378cf8fc720c38bcc5c78fc923624e

PR #11062, opened 2015-02-15
Update broken e2e tests.

benedictchen:
The e2e tests now don't work :-(.  This is an update to that.


benedictchen:
CLA should be there right now...the tag is wrong.

benedictchen:
Googlebot is frighteningly human.

gkalpak:
Bah, I miss @mary-poppins :disappointed:

benedictchen:
**nudge**

petebacondarwin:
@benedictchen - is this actually fixing broken e2e tests or adding new tests that you think would be useful? I think that perhaps it is the latter, in which case this should appear in the **Experiments** section, yes?

petebacondarwin:
Closing due to lack of activity. Please re-comment if you want to progress this PR further.

PR #11065, opened 2015-02-15
docs(tutorial/8 - More Templating): describe your change...

benedictchen:
Fixing missing route params being `undefined` bug.

### DIFF
--- a/docs/content/tutorial/step_08.ngdoc
+++ b/docs/content/tutorial/step_08.ngdoc
@@ -138,7 +138,7 @@ __`test/unit/controllersSpec.js`:__
 
       $routeParams.phoneId = 'xyz';
       scope = $rootScope.$new();
-      ctrl = $controller('PhoneDetailCtrl', {$scope: scope});
+      ctrl = $controller('PhoneDetailCtrl', {$scope: scope, $routeParams: $routeParams});
     }));
 
 


gkalpak:
I don't think this change is necessary. `$routeParams` is a service and shouldn't need to be passed via `locals`. [It seems to work as expected][1] without this change.

@benedictchen, did you actually get an `$routeParams being undefined` error ?

[1]: http://jsfiddle.net/ExpertSystem/bob5gdeL/1/

benedictchen:
Hi, @gkalpak.  It's always possible that it was human error on my part, but I when I performed `console.log($routeParams)` in the unit test on my end, it ended up being `undefined`, so I explicitly had to pass it in.  Perhaps It's just a weird glitch on my end?  You can close this PR if it makes no sense.

gkalpak:
> It's always possible that it was human error on my part

But the question is "was it ?". I don't want to close this until we can confirm that this works as expected (not that it **seems** to work as expected).

Can you try to reproduce the error again ? (E.g. you might have misspelled `$routeParams` or something). If you are unable to reproduce, you can close this PR. If you can still see the error, sharing more code should help pinpoint the issue (and fix it).

Thx for taking the time to look into this :)

benedictchen:
Hi @gkalpak I pulled fresh from the repository, so most likely it must've been human error since it seems to work okay and resolves the params.  There is no way to know for sure what happened since I had moved along through the tutorial already, but a fresh pull indicates that all tests pass. 
Thanks and sorry for the trouble!

PR #11074, opened 2015-02-16
feat(annotate): support injection of `@` function parameters in CoffeeScript 1.9

NickClark:
With the release of CoffeeScript 1.9, they changed the strategy for the
generation of internal compiler variable names. This means using
function parameters prefixed with an `@` no longer are injected
properly due to CoffeeScript converting them to an `_at_`. (ie. `@myVal`
becomes `_at_myVal`). This commit adds support to the Angular
annotator for this.

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -64,7 +64,7 @@
 
 var FN_ARGS = /^function\s*[^\(]*\(\s*([^\)]*)\)/m;
 var FN_ARG_SPLIT = /,/;
-var FN_ARG = /^\s*(_?)(\S+?)\1\s*$/;
+var FN_ARG = /^\s*(?:_at_)?(_?)(\S+?)\1\s*$/;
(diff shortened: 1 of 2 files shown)


Narretz:
Huh, I am actually not a big fan of angular accomodating compile to js languages. Isn't there a way to escape this inside Coffescript?

NickClark:
Ya, I'm not a fan either. It was a very nice convenience to have this shortcut for testing. At this point, I'll probably just have to lock it down to CoffeeScript 1.8 for existing code and quit using @vars in injected functions on future tests. I didn't see a way to configure or escape this in CoffeeScript itself.

NickClark:
Made an issue. (jashkenas/coffeescript/issues/3856)

NickClark:
Ok, I'm blind. Found this: https://github.com/jashkenas/coffeescript/issues/3816
Turns out a future release should fix our issues.

PR #11079, opened 2015-02-17
docs(input): better accessibility in examples

marcysutton:
The input and ngModel docs examples have no labels or accessible error handling. They could do more to encourage accessibility best practices. This change does two things:
1. Adds labels to each input.
2. Wraps error messages with `role="alert" so they will announce in a screen reader.

Even though Angular 2.0 is coming soon, these docs will live on with the 1.0 project so it makes sense to show more accessible examples. (Better late than never.)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -621,8 +621,8 @@
       }]);
     </script>
     <div ng-controller="GreeterController">
-      <input ng-model="name"> <br>
-      <textarea ng-model="html"></textarea> <br>
+      <input ng-model="name"> <br/>
+      <textarea ng-model="html"></textarea> <br/>
       <div compile="html"></div>
     </div>
    </file>
@@ -659,7 +659,7 @@
  *  * `cloneAttachFn` - If `cloneAttachFn` is provided, then the link function will clone the
  *  `template` and call the `cloneAttachFn` function allowing the caller to attach the
  *  cloned elements to the DOM document at the appropriate place. The `cloneAttachFn` is
- *  called as: <br> `cloneAttachFn(clonedElement, scope)` where:
+ *  called as: <br/> `cloneAttachFn(clonedElement, scope)` where:
  *
  *      * `clonedElement` - is a clone of the original `element` passed into the compiler.
  *      * `scope` - is the current scope with which the linking function is working with.
--- a/src/ng/directive/attrs.js
+++ b/src/ng/directive/attrs.js
@@ -113,12 +113,12 @@
  *
  * The buggy way to write it:
  * ```html
- * <img src="http://www.gravatar.com/avatar/{{hash}}"/>
+ * <img src="http://www.gravatar.com/avatar/{{hash}}" alt="Description"/>
  * ```
  *
  * The correct way to write it:
  * ```html
- * <img ng-src="http://www.gravatar.com/avatar/{{hash}}"/>
+ * <img ng-src="http://www.gravatar.com/avatar/{{hash}}" alt="Description" />
  * ```
  *
  * @element IMG
@@ -139,12 +139,12 @@
  *
  * The buggy way to write it:
  * ```html
- * <img srcset="http://www.gravatar.com/avatar/{{hash}} 2x"/>
+ * <img srcset="http://www.gravatar.com/avatar/{{hash}} 2x" alt="Description"/>
  * ```
  *
  * The correct way to write it:
  * ```html
- * <img ng-srcset="http://www.gravatar.com/avatar/{{hash}} 2x"/>
+ * <img ng-srcset="http://www.gravatar.com/avatar/{{hash}} 2x" alt="Description" />
  * ```
  *
  * @element IMG
@@ -181,7 +181,7 @@
  * @example
     <example>
       <file name="index.html">
-        Click me to toggle: <input type="checkbox" ng-model="checked"><br/>
+        <label>Click me to toggle: <input type="checkbox" ng-model="checked"></label><br/>
         <button ng-model="button" ng-disabled="checked">Button</button>
       </file>
       <file name="protractor.js" type="protractor">
@@ -216,8 +216,8 @@
  * @example
     <example>
       <file name="index.html">
-        Check me to check both: <input type="checkbox" ng-model="master"><br/>
-        <input id="checkSlave" type="checkbox" ng-checked="master">
+        <label>Check me to check both: <input type="checkbox" ng-model="master"></label><br/>
+        <input id="checkSlave" type="checkbox" ng-checked="master" aria-label="Slave input">
       </file>
       <file name="protractor.js" type="protractor">
         it('should check both checkBoxes', function() {
@@ -251,8 +251,8 @@
  * @example
     <example>
       <file name="index.html">
-        Check me to make text readonly: <input type="checkbox" ng-model="checked"><br/>
-        <input type="text" ng-readonly="checked" value="I'm Angular"/>
+        <label>Check me to make text readonly: <input type="checkbox" ng-model="checked"></label><br/>
+        <input type="text" ng-readonly="checked" value="I'm Angular" aria-label="Readonly field" />
       </file>
       <file name="protractor.js" type="protractor">
         it('should toggle readonly attr', function() {
@@ -287,8 +287,8 @@
  * @example
     <example>
       <file name="index.html">
-        Check me to select: <input type="checkbox" ng-model="selected"><br/>
-        <select>
+        <label>Check me to select: <input type="checkbox" ng-model="selected"></label><br/>
+        <select aria-label="ngSelected demo">
           <option>Hello!</option>
           <option id="greet" ng-selected="selected">Greetings!</option>
         </select>
@@ -324,7 +324,7 @@
  * @example
      <example>
        <file name="index.html">
-         Check me check multiple: <input type="checkbox" ng-model="open"><br/>
+         <label>Check me check multiple: <input type="checkbox" ng-model="open"></label><br/>
          <details id="details" ng-open="open">
             <summary>Show/Hide me</summary>
          </details>
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -68,13 +68,16 @@ var inputType = {
              }]);
          </script>
          <form name="myForm" ng-controller="ExampleController">
-           Single word: <input type="text" name="input" ng-model="example.text"
-                               ng-pattern="example.word" required ng-trim="false">
-           <span class="error" ng-show="myForm.input.$error.required">
-             Required!</span>
-           <span class="error" ng-show="myForm.input.$error.pattern">
-             Single word only!</span>
-
+           <label>Single word:
+             <input type="text" name="input" ng-model="example.text"
+                  ng-pattern="example.word" required ng-trim="false">
+           </label>
+           <div role="alert">
+             <span class="error" ng-show="myForm.input.$error.required">
+               Required!</span>
+             <span class="error" ng-show="myForm.input.$error.pattern">
+               Single word only!</span>
+           </div>
            <tt>text = {{example.text}}</tt><br/>
            <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
            <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -153,13 +156,15 @@ var inputType = {
             }]);
        </script>
        <form name="myForm" ng-controller="DateController as dateCtrl">
-          Pick a date in 2013:
+          <label for="exampleInput">Pick a date in 2013:</label>
           <input type="date" id="exampleInput" name="input" ng-model="example.value"
               placeholder="yyyy-MM-dd" min="2013-01-01" max="2013-12-31" required />
-          <span class="error" ng-show="myForm.input.$error.required">
-              Required!</span>
-          <span class="error" ng-show="myForm.input.$error.date">
-              Not a valid date!</span>
+          <div role="alert">
+            <span class="error" ng-show="myForm.input.$error.required">
+                Required!</span>
+            <span class="error" ng-show="myForm.input.$error.date">
+                Not a valid date!</span>
+           </div>
            <tt>value = {{example.value | date: "yyyy-MM-dd"}}</tt><br/>
            <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
            <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -246,13 +251,15 @@ var inputType = {
           }]);
       </script>
       <form name="myForm" ng-controller="DateController as dateCtrl">
-        Pick a date between in 2013:
+        <label for="exampleInput">Pick a date between in 2013:</label>
         <input type="datetime-local" id="exampleInput" name="input" ng-model="example.value"
             placeholder="yyyy-MM-ddTHH:mm:ss" min="2001-01-01T00:00:00" max="2013-12-31T00:00:00" required />
-        <span class="error" ng-show="myForm.input.$error.required">
-            Required!</span>
-        <span class="error" ng-show="myForm.input.$error.datetimelocal">
-            Not a valid date!</span>
+        <div role="alert">
+          <span class="error" ng-show="myForm.input.$error.required">
+              Required!</span>
+          <span class="error" ng-show="myForm.input.$error.datetimelocal">
+              Not a valid date!</span>
+        </div>
         <tt>value = {{example.value | date: "yyyy-MM-ddTHH:mm:ss"}}</tt><br/>
         <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
         <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -340,13 +347,15 @@ var inputType = {
         }]);
      </script>
      <form name="myForm" ng-controller="DateController as dateCtrl">
-        Pick a between 8am and 5pm:
+        <label for="exampleInput">Pick a between 8am and 5pm:</label>
         <input type="time" id="exampleInput" name="input" ng-model="example.value"
             placeholder="HH:mm:ss" min="08:00:00" max="17:00:00" required />
-        <span class="error" ng-show="myForm.input.$error.required">
-            Required!</span>
-        <span class="error" ng-show="myForm.input.$error.time">
-            Not a valid date!</span>
+        <div role="alert">
+          <span class="error" ng-show="myForm.input.$error.required">
+              Required!</span>
+          <span class="error" ng-show="myForm.input.$error.time">
+              Not a valid date!</span>
+        </div>
         <tt>value = {{example.value | date: "HH:mm:ss"}}</tt><br/>
         <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
         <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -433,13 +442,16 @@ var inputType = {
         }]);
       </script>
       <form name="myForm" ng-controller="DateController as dateCtrl">
-        Pick a date between in 2013:
-        <input id="exampleInput" type="week" name="input" ng-model="example.value"
+        <label>Pick a date between in 2013:
+          <input type="week" name="input" ng-model="example.value"
             placeholder="YYYY-W##" min="2012-W32" max="2013-W52" required />
-        <span class="error" ng-show="myForm.input.$error.required">
-            Required!</span>
-        <span class="error" ng-show="myForm.input.$error.week">
-            Not a valid date!</span>
+        </label>
+        <div role="alert">
+          <span class="error" ng-show="myForm.input.$error.required">
+              Required!</span>
+          <span class="error" ng-show="myForm.input.$error.week">
+              Not a valid date!</span>
+        </div>
         <tt>value = {{example.value | date: "yyyy-Www"}}</tt><br/>
         <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
         <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -526,13 +538,15 @@ var inputType = {
         }]);
      </script>
      <form name="myForm" ng-controller="DateController as dateCtrl">
-       Pick a month in 2013:
+       <label for="exampleInput">Pick a month in 2013:</label>
        <input id="exampleInput" type="month" name="input" ng-model="example.value"
           placeholder="yyyy-MM" min="2013-01" max="2013-12" required />
-       <span class="error" ng-show="myForm.input.$error.required">
-          Required!</span>
-       <span class="error" ng-show="myForm.input.$error.month">
-          Not a valid month!</span>
+       <div role="alert">
+         <span class="error" ng-show="myForm.input.$error.required">
+            Required!</span>
+         <span class="error" ng-show="myForm.input.$error.month">
+            Not a valid month!</span>
+       </div>
        <tt>value = {{example.value | date: "yyyy-MM"}}</tt><br/>
        <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
        <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -629,12 +643,16 @@ var inputType = {
              }]);
          </script>
          <form name="myForm" ng-controller="ExampleController">
-           Number: <input type="number" name="input" ng-model="example.value"
-                          min="0" max="99" required>
-           <span class="error" ng-show="myForm.input.$error.required">
-             Required!</span>
-           <span class="error" ng-show="myForm.input.$error.number">
-             Not valid number!</span>
+           <label>Number:
+             <input type="number" name="input" ng-model="example.value"
+                  min="0" max="99" required>
+          </label>
+           <div role="alert">
+             <span class="error" ng-show="myForm.input.$error.required">
+               Required!</span>
+             <span class="error" ng-show="myForm.input.$error.number">
+               Not valid number!</span>
+           </div>
            <tt>value = {{example.value}}</tt><br/>
            <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
            <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -719,11 +737,15 @@ var inputType = {
              }]);
          </script>
          <form name="myForm" ng-controller="ExampleController">
-           URL: <input type="url" name="input" ng-model="url.text" required>
-           <span class="error" ng-show="myForm.input.$error.required">
-             Required!</span>
-           <span class="error" ng-show="myForm.input.$error.url">
-             Not valid url!</span>
+           <label>URL:
+             <input type="url" name="input" ng-model="url.text" required>
+           <label>
+           <div role="alert">
+             <span class="error" ng-show="myForm.input.$error.required">
+               Required!</span>
+             <span class="error" ng-show="myForm.input.$error.url">
+               Not valid url!</span>
+           </div>
            <tt>text = {{url.text}}</tt><br/>
            <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
            <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -810,11 +832,15 @@ var inputType = {
              }]);
          </script>
            <form name="myForm" ng-controller="ExampleController">
-             Email: <input type="email" name="input" ng-model="email.text" required>
-             <span class="error" ng-show="myForm.input.$error.required">
-               Required!</span>
-             <span class="error" ng-show="myForm.input.$error.email">
-               Not valid email!</span>
+             <label>Email:
+               <input type="email" name="input" ng-model="email.text" required>
+             </label>
+             <div role="alert">
+               <span class="error" ng-show="myForm.input.$error.required">
+                 Required!</span>
+               <span class="error" ng-show="myForm.input.$error.email">
+                 Not valid email!</span>
+             </div>
              <tt>text = {{email.text}}</tt><br/>
              <tt>myForm.input.$valid = {{myForm.input.$valid}}</tt><br/>
              <tt>myForm.input.$error = {{myForm.input.$error}}</tt><br/>
@@ -883,9 +909,18 @@ var inputType = {
              }]);
          </script>
          <form name="myForm" ng-controller="ExampleController">
-           <input type="radio" ng-model="color.name" value="red">  Red <br/>
-           <input type="radio" ng-model="color.name" ng-value="specialValue"> Green <br/>
-           <input type="radio" ng-model="color.name" value="blue"> Blue <br/>
+           <label>
+             <input id="red" type="radio" ng-model="color.name" value="red">
+             Red
+           </label><br/>
+           <label>
+             <input id="green" type="radio" ng-model="color.name" ng-value="specialValue">
+             Green
+           </label><br/>
+           <label>
+             <input id="blue" type="radio" ng-model="color.name" value="blue">
+             Blue
+           </label><br/>
            <tt>color = {{color.name | json}}</tt><br/>
           </form>
           Note that `ng-value="specialValue"` sets radio item's value to be the value of `$scope.specialValue`.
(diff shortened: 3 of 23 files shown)


gkalpak:
I wonder if it is a good idea to use the `<label>...<input .../></label>` form (see below) in order to avoid clattering the views (and avoid having to add `id` attributes to inputs that don't need it).

```html
<label>
  Some label:
  <input type="text" name="input" ng-model="example.text" />
</label>
```

marcysutton:
@gkalpak because this is primarily for developer education, I opted for explicit labels on purpose. A common accessibility problem is to leave off the `for=""` and not properly link the `<label>` and the `<input>` elements (or to leave off `<label>` entirely). Showing how they are linked together does more for developer education, which has been sorely lacking in Angular in the past. 

You can read about implicit vs. explicit labeling vs. aria-labelledby here: http://www.paciellogroup.com/blog/2011/07/html5-accessibility-chops-form-control-labeling/

gkalpak:
Interestingly enough, although the author seems to suggest using `for / id` instead of `wrapping`, the [test results][1] seem to indicate that `wrapping` is more reliable than `for / id` ! Do I miss something ?

[1]: http://www.html5accessibility.com/tests/form-labels.html

marcysutton:
The support differences look pretty minimal to me. This post is also from 2012, so take it with a grain of salt.

zamith:
:+1:

pauljadam:
+1 please fix a11y in the docs it's very very bad, all the angularJS books have not a11y just like the docs, <label> wrapped labels <input> </label> are fine in my testing, yes it's less code and will look more readable, any a11y we can get the better!

src/ng/directive/input.js
```diff
@@ -68,13 +68,15 @@ var inputType = {
...
-                               ng-pattern="example.word" required ng-trim="false">
-           <span class="error" ng-show="myForm.input.$error.required">
-             Required!</span>
-           <span class="error" ng-show="myForm.input.$error.pattern">
-             Single word only!</span>
-
+           <label for="input">Single word:</label>
+           <input id="input" type="text" name="input" ng-model="example.text"
+                  ng-pattern="example.word" required ng-trim="false">
+           <span role="alert">
```
ckundo: Does `role=alert` announce assertively/immediately? Is that the desired experience for a form validation?

src/ng/directive/input.js
```diff
@@ -68,13 +68,15 @@ var inputType = {
...
          <form name="myForm" ng-controller="ExampleController">
-           Single word: <input type="text" name="input" ng-model="example.text"
-                               ng-pattern="example.word" required ng-trim="false">
-           <span class="error" ng-show="myForm.input.$error.required">
-             Required!</span>
-           <span class="error" ng-show="myForm.input.$error.pattern">
-             Single word only!</span>
-
+           <label for="input">Single word:</label>
+           <input id="input" type="text" name="input" ng-model="example.text"
```
ckundo: Is it confusing to have the `id` value be the same as the element name and `name` attribute?

src/ng/directive/input.js
```diff
@@ -153,13 +155,15 @@ var inputType = {
...
-          <span class="error" ng-show="myForm.input.$error.required">
-              Required!</span>
-          <span class="error" ng-show="myForm.input.$error.date">
-              Not a valid date!</span>
+          <span role="alert">
+            <span class="error" ng-show="myForm.input.$error.required">
+                Required!</span>
+            <span class="error" ng-show="myForm.input.$error.date">
+                Not a valid date!</span>
+           </span><br>
```
ckundo: Maybe worth making this `<br>` self closing, to match other break tags in the document.

src/ng/directive/input.js
marcysutton: When I tested it, it matched the speed that the visual message would come up in context to what you're typing...it was especially useful for validations related to input length, so you'd know exactly when the error occurred.

src/ng/directive/input.js
```diff
@@ -153,13 +155,15 @@ var inputType = {
             }]);
        </script>
        <form name="myForm" ng-controller="DateController as dateCtrl">
-          Pick a date in 2013:
+          <label for="exampleInput">Pick a date in 2013:</label>
```
ckundo: It looks like `exampleInput` is a duplicate `id`. Will that impact focus and AT behavior?

src/ng/directive/input.js
marcysutton: Good point. Or it could just become a `<div>`!

src/ng/directive/ngModel.js
```diff
@@ -943,7 +945,8 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
        Update input to see transitions when valid/invalid.
        Integer is a valid value.
        <form name="testForm" ng-controller="ExampleController">
-         <input ng-model="val" ng-pattern="/^\d+$/" name="anim" class="my-input" />
+         <label for="my-input">My input</label>
+         <input id="my-input" ng-model="val" ng-pattern="/^\d+$/" name="anim" class="my-input" />
```
ckundo: `id` values are camelcase elsewhere. Should that be true here too?

src/ng/directive/ngModel.js
```diff
@@ -1110,14 +1113,14 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
     <file name="index.html">
       <div ng-controller="ExampleController">
         <form name="userForm">
-          Name:
-          <input type="text" name="userName"
+          <label for="userName">Name:</label>
+          <input type="text" name="userName" id="userName"
```
ckundo: Seems like it'd be more consistent with the document to put the `id` attribute first.

src/ng/directive/input.js
ckundo: That's really cool! Good to know.

src/ng/directive/input.js
marcysutton: Probably. I'll simplify these based on all the great feedback so far and just use implicit labeling. Thanks!

ckundo:
Huge +1 in general. Great to see improved accessibility in the docs, and the `role=alert` pattern on form validations is super cool and a unique accessibility feature to Angular. The duplicate `id`s might be a blocker until resolved, but otherwise, I vote ship it :rocket:!

src/ng/directive/input.js
marcysutton: It normally would, but it's isolated to this one demo.

marcysutton:
Thanks for the feedback! I've simplified a lot of the labeling based on comments in this PR.

marcysutton:
@ckundo @gkalpak I did a search for `<input>` and found a bunch more places that needed accessibility help, including stuff like this, where the "good" example is missing an alt attribute:

 > The buggy way to write it:
 ```html
 <img src="http://www.gravatar.com/avatar/{{hash}}"/>
 ```
 > The correct way to write it:
 ```html
 <img ng-src="http://www.gravatar.com/avatar/{{hash}}"/>
 ```
Can you take another look if you get a chance? Thanks so much.

src/ng/directive/input.js
```diff
@@ -68,13 +68,16 @@ var inputType = {
...
-           Single word: <input type="text" name="input" ng-model="example.text"
-                               ng-pattern="example.word" required ng-trim="false">
-           <span class="error" ng-show="myForm.input.$error.required">
-             Required!</span>
-           <span class="error" ng-show="myForm.input.$error.pattern">
-             Single word only!</span>
-
+           <label>Single word:
+             <input type="text" name="input" ng-model="example.text"
+                  ng-pattern="example.word" required ng-trim="false">
```
ckundo: Alignment with the attributes above is not quite right.

src/ng/directive/input.js
```diff
@@ -433,13 +442,16 @@ var inputType = {
         }]);
       </script>
       <form name="myForm" ng-controller="DateController as dateCtrl">
-        Pick a date between in 2013:
-        <input id="exampleInput" type="week" name="input" ng-model="example.value"
+        <label>Pick a date between in 2013:
+          <input type="week" name="input" ng-model="example.value"
             placeholder="YYYY-W##" min="2012-W32" max="2013-W52" required />
```
ckundo: Attributes don't line up here (`placeholder` with `type`)

src/ng/directive/input.js
```diff
@@ -625,12 +639,16 @@ var inputType = {
...
          <form name="myForm" ng-controller="ExampleController">
-           Number: <input type="number" name="input" ng-model="example.value"
-                          min="0" max="99" required>
-           <span class="error" ng-show="myForm.input.$error.required">
-             Required!</span>
-           <span class="error" ng-show="myForm.input.$error.number">
-             Not valid number!</span>
+           <label>Number:
+             <input type="number" name="input" ng-model="example.value"
+                  min="0" max="99" required>
```
ckundo: Attributes don't line up here.

src/ng/directive/input.js
```diff
@@ -929,9 +964,13 @@ var inputType = {
...
          <form name="myForm" ng-controller="ExampleController">
-           Value1: <input type="checkbox" ng-model="checkboxModel.value1"> <br/>
-           Value2: <input type="checkbox" ng-model="checkboxModel.value2"
-                          ng-true-value="'YES'" ng-false-value="'NO'"> <br/>
+           <label>Value1:
+             <input type="checkbox" ng-model="checkboxModel.value1">
+           </label><br/>
+           <label>Value2:
+             <input type="checkbox" ng-model="checkboxModel.value2"
+                ng-true-value="'YES'" ng-false-value="'NO'">
```
ckundo: Another attribute alignment issue.

ckundo:
Is there some guideline around line length? 80 characters is a sane guideline, but I don't know if the project adheres to that. @gkalpak do you know?

ckundo:
@marcysutton it occurred to me that the `role=alert` changes might be better off in a separate PR. Do you think that change can be isolated?

src/ng/directive/ngClass.js
```diff
@@ -156,17 +156,17 @@ function classDirective(name, selector) {
...
-       <input type="checkbox" ng-model="deleted"> deleted (apply "strike" class)<br>
-       <input type="checkbox" ng-model="important"> important (apply "bold" class)<br>
-       <input type="checkbox" ng-model="error"> error (apply "red" class)
+       <label><input type="checkbox" ng-model="deleted"> deleted (apply "strike" class)</label><br>
+       <label><input type="checkbox" ng-model="important"> important (apply "bold" class)</label><br>
+       <label><input type="checkbox" ng-model="error"> error (apply "red" class)</label>
        <hr>
        <p ng-class="style">Using String Syntax</p>
-       <input type="text" ng-model="style" placeholder="Type: bold strike red">
+       <label><input type="text" ng-model="style" placeholder="Type: bold strike red">
```
ckundo: Is this missing a closing tag?

src/ng/directive/ngClass.js
```diff
@@ -156,17 +156,17 @@ function classDirective(name, selector) {
...
        <hr>
        <p ng-class="[style1, style2, style3]">Using Array Syntax</p>
-       <input ng-model="style1" placeholder="Type: bold, strike or red"><br>
-       <input ng-model="style2" placeholder="Type: bold, strike or red"><br>
-       <input ng-model="style3" placeholder="Type: bold, strike or red"><br>
+       <input ng-model="style1" placeholder="Type: bold, strike or red" aria-label="Type: bold, strike or red"><br>
```
ckundo: This is where a line length guideline might be nice.

src/ng/directive/ngController.js
```diff
@@ -64,20 +64,20 @@
  * <example name="ngControllerAs" module="controllerAsExample">
  *   <file name="index.html">
  *    <div id="ctrl-as-exmpl" ng-controller="SettingsController1 as settings">
- *      Name: <input type="text" ng-model="settings.name"/>
- *      [ <a href="" ng-click="settings.greet()">greet</a> ]<br/>
+ *      <label>Name: <input type="text" ng-model="settings.name"/></label>
+ *      <button ng-click="settings.greet()">greet</button><br/>
```
ckundo: Nice addition! This seems though like another instance of something that could be included in a separate PR, one related specifically to semantic use of link vs. button. Same applies for the additions below.

src/ng/directive/ngController.js
```diff
@@ -64,20 +64,20 @@
...
  *      Contact:
  *      <ul>
  *        <li ng-repeat="contact in settings.contacts">
- *          <select ng-model="contact.type">
+ *          <select ng-model="contact.type" aria-label="contact method" id="select_{{$index}}">
```
ckundo: Does capitalization matter here for the `aria-label`? I noticed that elsewhere it's capitalized.

src/ng/directive/ngController.js
```diff
@@ -147,20 +147,20 @@
...
  *   <div id="ctrl-exmpl" ng-controller="SettingsController2">
- *     Name: <input type="text" ng-model="name"/>
- *     [ <a href="" ng-click="greet()">greet</a> ]<br/>
+ *     <label>Name: <input type="text" ng-model="name"/></label>
+ *     <button ng-click="greet()">greet</button><br/>
  *     Contact:
  *     <ul>
  *       <li ng-repeat="contact in contacts">
- *         <select ng-model="contact.type">
+ *         <select ng-model="contact.type" id="select_{{$index}}">
```
ckundo: Cool, good call with the `aria-labelledby`! I'll defer to @gkalpak on this approach to interpolating the `$index` value, I'm not sure if there are other approaches to getting the index of the chosen option in the `select`.

src/ng/directive/ngModel.js
```diff
@@ -474,12 +474,14 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
    *        <p>Now see what happens if you start typing then press the Escape key</p>
    *
    *       <form name="myForm" ng-model-options="{ updateOn: 'blur' }">
-   *         <p>With $rollbackViewValue()</p>
-   *         <input name="myInput1" ng-model="myValue" ng-keydown="resetWithCancel($event)"><br/>
+   *         <p id="inputLabel1">With $rollbackViewValue()</p>
```
ckundo: Can't this just be a `label` instead of a `p`?

gkalpak:
100 characters per line is the limit.
(I believe it's somewhere in contributing.md)

src/ng/directive/ngModel.js
```diff
@@ -474,12 +474,14 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
...
+   *         <input name="myInput1" aria-labelledby="inputLabel1" ng-model="myValue"
```
ckundo: If the `p` is in fact a description, should this be `aria-describedby`?

src/ng/directive/ngModel.js
```diff
@@ -474,12 +474,14 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
...
+   *                ng-keydown="resetWithCancel($event)"><br/>
    *         myValue: "{{ myValue }}"
    *
-   *         <p>Without $rollbackViewValue()</p>
-   *         <input name="myInput2" ng-model="myValue" ng-keydown="resetWithoutCancel($event)"><br/>
+   *         <p id="inputLabel2">Without $rollbackViewValue()</p>
```
ckundo: Same question about `p` vs `label` and `labelledby` vs `describedby`

src/ng/directive/ngModel.js
```diff
@@ -940,10 +942,13 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
...
-       Update input to see transitions when valid/invalid.
-       Integer is a valid value.
+       <p id="inputDescription">
+        Update input to see transitions when valid/invalid.
+        Integer is a valid value.
+       </p>
        <form name="testForm" ng-controller="ExampleController">
-         <input ng-model="val" ng-pattern="/^\d+$/" name="anim" class="my-input" />
+         <input ng-model="val" ng-pattern="/^\d+$/" name="anim" class="my-input"
+            aria-labelledby="inputDescription" />
```
ckundo: Attribute alignment is off, and same question about label association vs description.

src/ng/directive/ngModel.js
```diff
@@ -973,10 +978,11 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
      <file name="index.html">
        <div ng-controller="ExampleController">
          <form name="userForm">
-           Name:
-           <input type="text" name="userName"
+           <label>Name:
+             <input type="text" name="userName"
                   ng-model="user.name"
```
ckundo: Attribute alignment is off here.

src/ng/directive/ngModel.js
```diff
@@ -1110,14 +1116,15 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
     <file name="index.html">
       <div ng-controller="ExampleController">
         <form name="userForm">
-          Name:
-          <input type="text" name="userName"
+          <label>Name:
+            <input type="text" name="userName"
                  ng-model="user.name"
```
ckundo: Attribute alignment is off here.

src/ng/directive/ngModel.js
```diff
@@ -1110,14 +1116,15 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
                  ng-model-options="{ updateOn: 'blur' }"
-                 ng-keyup="cancel($event)" /><br />
-
-          Other data:
-          <input type="text" ng-model="user.data" /><br />
+                 ng-keyup="cancel($event)" />
+          </label><br />
```
ckundo: `<br>` with no closing / would be more consistent with surrounding code.

src/ng/directive/ngModel.js
```diff
@@ -1110,14 +1116,15 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
+          <label>Other data:
+            <input type="text" ng-model="user.data" />
+          </label><br />
```
ckundo: Sans `/` s'il vous plait

src/ng/directive/ngModel.js
```diff
@@ -1165,10 +1172,11 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
     <file name="index.html">
       <div ng-controller="ExampleController">
         <form name="userForm">
-          Name:
-          <input type="text" name="userName"
+          <label>Name:
+            <input type="text" name="userName"
                  ng-model="user.name"
```
ckundo: Attribute alignment.

src/ng/directive/ngModel.js
```diff
@@ -1188,10 +1196,11 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
...
```
ckundo: Attribute alignment

src/ng/directive/attrs.js
```diff
@@ -144,7 +144,7 @@
  *
  * The correct way to write it:
  * ```html
- * <img ng-srcset="http://www.gravatar.com/avatar/{{hash}} 2x"/>
+ * <img ng-srcset="http://www.gravatar.com/avatar/{{hash}} 2x" alt="Description" />
```
gkalpak: I suggest the same change is made to the buggy way above.
Since this example is about `src` vs `ngSrc`, we don't want to distract the reader into thinking `alt` has anything to do with that.

src/ng/directive/ngClass.js
ckundo: Per the 100 character max guideline, this should be wrapped (right now it's 115 characters)

src/ng/directive/attrs.js
```diff
@@ -118,7 +118,7 @@
  *
  * The correct way to write it:
  * ```html
- * <img ng-src="http://www.gravatar.com/avatar/{{hash}}"/>
+ * <img ng-src="http://www.gravatar.com/avatar/{{hash}}" alt="Description" />
```
gkalpak: See https://github.com/angular/angular.js/pull/11079/files?diff=unified#r25614911.

src/ng/directive/ngOptions.js
```diff
@@ -147,37 +147,39 @@ var ngOptionsMinErr = minErr('ngOptions');
         <div ng-controller="ExampleController">
           <ul>
             <li ng-repeat="color in colors">
-              Name: <input ng-model="color.name">
-              <input type="checkbox" ng-model="color.notAnOption"> Disabled?
-              [<a href ng-click="colors.splice($index, 1)">X</a>]
+              <label>Name: <input ng-model="color.name"></label>
+              <label><input type="checkbox" ng-model="color.notAnOption"> Disabled?</label>
+              [<button ng-click="colors.splice($index, 1)" aria-label="remove">X</button>]
```
ckundo: Maybe a separate PR. Also, in a similar change earlier in these changes the brackets were dropped. That's inconsistent with this change. Not sure which you prefer.

src/ng/directive/ngOptions.js
```diff
@@ -147,37 +147,39 @@ var ngOptionsMinErr = minErr('ngOptions');
...
             </li>
           </ul>
           <hr/>
-          Color (null not allowed):
-          <select ng-model="myColor" ng-options="color.name for color in colors"></select><br>
-
-          Color (null allowed):
+          <label>Color (null not allowed):
+            <select ng-model="myColor" ng-options="color.name for color in colors"></select><br>
+          </label>
```
ckundo: Something is wonky with this whitespace.

src/ng/directive/ngOptions.js
```diff
@@ -147,37 +147,39 @@ var ngOptionsMinErr = minErr('ngOptions');
...
+            </select>
+          </label><br/>
 
-          Color grouped by shade, with some disabled:
-          <select ng-model="myColor"
-                  ng-options="color.name group by color.shade disable when color.notAnOption for color in colors">
-          </select><br/>
+          <label>Color grouped by shade, with some disabled:
+            <select ng-model="myColor"
+                    ng-options="color.name group by color.shade disable when color.notAnOption for color in colors">
```
ckundo: This looks like it might be another line length concern.

src/ng/directive/ngShowHide.js
```diff
@@ -98,7 +98,7 @@ var NG_HIDE_IN_PROGRESS_CLASS = 'ng-hide-animate';
  * @example
   <example module="ngAnimate" deps="angular-animate.js" animations="true">
     <file name="index.html">
-      Click me: <input type="checkbox" ng-model="checked"><br/>
+      Click me: <input type="checkbox" ng-model="checked" aria-label="toggle ngHide and ngShow"><br/>
```
ckundo: Is this actually what's happening ("toggle ngHide and ngShow")? It seems like this is actually toggling `.ng-hide`. @gkalpak can you weigh in?

src/ng/directive/ngShowHide.js
```diff
@@ -263,7 +263,7 @@ var ngShowDirective = ['$animate', function($animate) {
  * @example
   <example module="ngAnimate" deps="angular-animate.js" animations="true">
     <file name="index.html">
-      Click me: <input type="checkbox" ng-model="checked"><br/>
+      Click me: <input type="checkbox" ng-model="checked" aria-label="toggle ngShow and ngHide"><br/>
```
ckundo: Same comment, what are we actually toggling here?

src/ng/directive/input.js
```diff
@@ -879,9 +905,18 @@ var inputType = {
...
+             Red
+           </label><br/>
+           <label>
+             <input id="green" type="radio" ng-model="color.name" ng-value="specialValue">
+             Green
+           </label><br/>
+           <label>
+             <input id="blue" type="radio" ng-model="color.name" value="blue">
+             Blue
+           </label><br/>
```
gkalpak: Why add ids, if are nesting them into labels ?

src/ng/filter/limitTo.js
```diff
@@ -37,11 +37,11 @@
...
        </script>
        <div ng-controller="ExampleController">
-         Limit {{numbers}} to: <input type="number" step="1" ng-model="numLimit">
+         <label>Limit {{numbers}} to: <input type="number" step="1" ng-model="numLimit"></label>
          <p>Output numbers: {{ numbers | limitTo:numLimit }}</p>
-         Limit {{letters}} to: <input type="number" step="1" ng-model="letterLimit">
+         <label>Limit {{letters}} to: <input type="number" step="1" ng-model="letterLimit"></label>
          <p>Output letters: {{ letters | limitTo:letterLimit }}</p>
-         Limit {{longNumber}} to: <input type="number" step="1" ng-model="longNumberLimit">
+         <label>Limit {{longNumber}} to: <input type="number" step="1" ng-model="longNumberLimit"></label>
```
ckundo: Line length?

src/ng/http.js
```diff
@@ -671,11 +671,11 @@ function $HttpProvider() {
 <example module="httpExample">
 <file name="index.html">
   <div ng-controller="FetchController">
-    <select ng-model="method">
+    <select ng-model="method" aria-label="HTTP method">
```
ckundo: I don't think JSONP is an HTTP method, maybe the label should be "Request method"

src/ng/directive/input.js
```diff
@@ -1449,15 +1488,23 @@ function checkboxInputType(scope, element, attr, ctrl, $sniffer, $browser, $filt
...
+           <label>User name:
+             <input type="text" name="userName" ng-model="user.name" required>
+           </label>
+           <div role="alert">
+             <span class="error" ng-show="myForm.userName.$error.required">
+              Required!</span>
+           </div>
+           <label>Last name:
+             <input type="text" name="lastName" ng-model="user.last"
              ng-minlength="3" ng-maxlength="10">
```
gkalpak: This seems misaligned...

src/ng/sce.js
```diff
@@ -413,7 +413,10 @@ function $SceDelegateProvider() {
  * Here's an example of a binding in a privileged context:
  *
  * ```
- * <input ng-model="userHtml">
+ * <label>
+ *   User input
```
ckundo: This seems to be changing the UI by adding additional content. Seems like that should be fine, but thought I'd call it out since so far that has not been the case.

src/ngMessages/messages.js
```diff
@@ -31,7 +31,9 @@ var jqLite = angular.element;
  *
  * ```html
  * <form name="myForm">
- *   <input type="text" ng-model="field" name="myField" required minlength="5" />
+ *   <label>Enter text:
```
ckundo: Same question, is adding this new label text in scope?

src/ng/directive/ngController.js
gkalpak: I think the `$index` approach is as good as any (as long as screen readers can cope with the id changing dynamically).

src/ng/directive/ngList.js
```diff
@@ -30,9 +30,11 @@
  *   </file>
  *   <file name="index.html">
  *    <form name="myForm" ng-controller="ExampleController">
- *      List: <input name="namesInput" ng-model="names" ng-list required>
- *      <span class="error" ng-show="myForm.namesInput.$error.required">
+ *      <label>List: <input name="namesInput" ng-model="names" ng-list required></label>
+ *      <span role="alert">
+ *        <span class="error" ng-show="myForm.namesInput.$error.required">
  *        Required!</span>
```
gkalpak: Indentation...

src/ng/directive/ngModel.js
gkalpak: I would happier if there were `/`s everywhere :)
<sub>(It's a matter of personal reference, so I don't think it's a big deal. The docs are full of such little inconsistencies.)</sub>

src/ngMessages/messages.js
```diff
@@ -31,7 +31,9 @@ var jqLite = angular.element;
...
+ *     <input type="text" ng-model="field" name="myField" required minlength="5" />
+ *   </label>
  *   <div ng-messages="myForm.myField.$error">
```
gkalpak: Could this use a `role="alert"` attrbute ?

ckundo:
Besides the formatting fixes, I think the `role=alert` and `a` vs `button` concerns might be better in a separate PR, but that may be a little fussy. Also noted that the last file in the changes has new content, which was different. All the other changes should have no effect on the presentation of the page, so thought I'd raise it.

@marcysutton thanks a million for this work. Keep it up!

src/ng/directive/ngModel.js
ckundo: @gkalpak I agree, erring on the side of consistency with the surrounding code, to avoid complicated diffs/unrelated changes.

src/ng/http.js
gkalpak: :+1: Good catch

src/ng/directive/ngShowHide.js
gkalpak: Toggling `.ng-hide` is an implementation detail that is of no interest to the user.
Maybe "toggle thumbs-up thumbs-down icons" ?

<sub>(Again, this is not a real app, so it isn't all that significant. The important thing here is to demonstrate proper use of aria attributes.)</sub>

gkalpak:
It generally LGTM. Really great work ! :+1: @marcysutton 

--
A couple of notes:

1. Alignment is not super-correct. It would nice to have everything align/indented consistently of course, but I can leave without as well :)

2. I prefer closing `/>` (**with** the `/`), but that seems to not be directly related to this PR (plus is a matter of personal preference anyway).

3. Although I generally agree that `button` is preferable to `a` (when used as a button), `a`s do look better in certain cases. Couldn't we go with the `a`s (e.g. for `X` "buttons") and put a `role` attribute or something. Since many people do use `a`s anyway, it would be good to raise awareness around "doing it right" (if you have to).

src/ng/directive/ngClass.js
marcysutton: It is indeed! Good catch.

src/ng/directive/ngController.js
marcysutton: It doesn't matter since it will be read aloud the same way, but I'll update it for consistency.

src/ng/directive/ngModel.js
marcysutton: Yeah this was a weird case where the text content didn't really work as a label for the input, but it still needed to be described by something. That's why I did it this way.

src/ng/directive/ngModel.js
marcysutton: I've seen examples with both. I can certainly make them more consistent, but it's hard to tell which convention to use :)

src/ng/directive/attrs.js
marcysutton: Gotcha. That makes sense to me.

src/ng/directive/ngModel.js
ckundo: See Georgios comment below, looks like keeping the trailing slashes is preferred.

src/ng/directive/ngOptions.js
marcysutton: Good catch. Without brackets works best with buttons.

src/ng/directive/attrs.js
ckundo: @gkalpak @marcysutton why not add the `alt` to both?

src/ng/directive/ngOptions.js
marcysutton: Can we break the `ng-options` value string up at all? I think it was over the line length already before I made the edit, but if there's a way to make it fit better, I'll do it!

src/ng/directive/input.js
marcysutton: Good question, that was leftover from my first pass on the code. They are no longer necessary!

src/ng/sce.js
marcysutton: I will switch it to `aria-label` so that we don't have to add new content.

src/ngMessages/messages.js
marcysutton: `ngAria` actually adds them to `ngMessages`, but it can't hurt to include the role here.

src/ngMessages/messages.js
marcysutton: This is pretty basic form label stuff, in this kind of an example to me it makes sense to add label text.

src/ng/directive/attrs.js
marcysutton: Done! Great suggestion.

marcysutton:
Thanks again for your thorough review, this is awesome!

1. Edited alignment to be more consistent.

2. I did a find and replace for `<br>` in the files I had edited and changed any occurrences to `<br/>` since it was pretty easy.

3. The reason I changed the `<a>`s to `<button>`s is because they all had empty `href`s and were essentially doing the work of buttons (ng-click only, no routing). Adding `role=button` wouldn't negate the need for an empty `href` to make it a valid element. It's just easier to make them buttons. :)

If it's okay to leave this all in one branch, it will save me from having to do the work of splitting it up and creating additional PRs. In the future I'll try to break it up though. I know it's a big one! Thanks again for all your help.

ckundo:
@marcysutton thanks, LGTM! :+1:

src/ng/directive/ngOptions.js
gkalpak: At least in Chrome, you can break strings up with no problem. But since this is one long string I believe it's OK even if it's a little longer than 100 chars.

src/ng/directive/ngModel.js
gkalpak: Just to be clear: leaving the slashes in MY personal preference. My sense is that in most instances in the docs omitted the trailing slashes, but as mentioned already, either way is fine :)

src/ng/compile.js
```diff
@@ -621,8 +621,8 @@
       }]);
     </script>
     <div ng-controller="GreeterController">
-      <input ng-model="name"> <br>
-      <textarea ng-model="html"></textarea> <br>
+      <input ng-model="name"> <br/>
+      <textarea ng-model="html"></textarea> <br/>
```
marcysutton: These should have labels too. Oops! I'll fix them up later.

PR #11084, opened 2015-02-17
perf($compile): replace forEach(controller) with plain loops

jbedard:
Pretty minor but it is often a couple % spent just within these forEach(controller) calls. Minor perf aside, imo moving this chunk of code to it's own method is nicer anyway to reduce the size of `nodeLinkFn` a bit... 

(this was originally part of #9772)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1671,7 +1671,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
         if (!directive.templateUrl && directive.controller) {
           directiveValue = directive.controller;
-          controllerDirectives = controllerDirectives || {};
+          controllerDirectives = controllerDirectives || createMap();
           assertNoDuplicate("'" + directiveName + "' controller",
               controllerDirectives[directiveName], directive, $compileNode);
           controllerDirectives[directiveName] = directive;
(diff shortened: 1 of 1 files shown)


src/ng/compile.js
```diff
@@ -1877,6 +1877,36 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         return value || null;
       }
 
+      function setupControllers($element, attrs, transcludeFn, controllerDirectives, isolateScope, scope) {
```
lgalfaso: by adding `newIsolateScopeDirective` as a parameter, this can be moved out of the closure (or I am missing something?)

src/ng/compile.js
```diff
@@ -1877,6 +1877,36 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+        var elementControllers = createMap();
+        for (var controllerKey in controllerDirectives) {
```
lgalfaso: I am not a big fan of `for ... in`, would like to know more opinions

petebacondarwin: Well, `angular.forEach` ends up using `for ... in` in any case, except that it checks the `hasOwnProperty` first.  In this case, I am pretty confident that `controllerDirectives` will not contain any dodgy properties, especially since @jbedard has used `createMap()` to ensure that it doesn't derive from `Object`.

src/ng/compile.js
petebacondarwin: Actually, if `newIsolateScopeDirective` is not truthy then `isolateScope` will be null, in which case we don't need `newIsolateScopeDirective` to be passed through at all, via param or closure.

src/ng/compile.js
```diff
@@ -1877,6 +1877,36 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+          var directive = controllerDirectives[controllerKey];
+          var locals = {
+            $scope: directive === newIsolateScopeDirective || directive.$$isolateScope ? isolateScope : scope,
```
petebacondarwin: This can just be

```js
$scope: isolateScope || scope,
```

right?

jbedard: I don't think so... we only want the isolated scope if this controller is for the isolated scope directive. There might be other controllers that need to be initialized with the normal scope...

src/ng/compile.js
jbedard: Personally I often prefer `for ... in` over `forEach`. That's all `forEach` does anyway.

Could use `hasOwnProperty` instead of `createMap` but that seems to be slower in many cases.  Although that would actually make it more like previously (because that's what `forEach` does) which would reduce the changes in this commit.

The other option is `Object.keys() + for (0 to len)`. Or maybe `controllerDirectives` could actually be an array? Looks like the only place `controllerDirectives` is used other then looping over the values is for the `assertNoDuplicate` check...

src/ng/compile.js
petebacondarwin: But the only place that `isolateScope` gets assigned is line 1925 (https://github.com/angular/angular.js/pull/11084/files#diff-a732922b631efed1b9f33a24082ae0dbR1924), which only happens if `newIsolateScopeDirective` is truthy.

jbedard: But that doesn't mean `directive === newIsolateScopeDirective`, there might be other controllers that don't want the isolated scope.  Unless `directive.$$isolateScope` will always be true on that case? (which I'm assuming it is not)

src/ng/compile.js
petebacondarwin: I am happy with `for ... in` in this case where we have control over the object being iterated over.

src/ng/compile.js
petebacondarwin: Doh! Sorry. Yes, you are right - I didn't notice the comparison with `directive`.

src/ng/compile.js
petebacondarwin: We would also need `directive` and possibly other stuff. Let's not go there for now.

petebacondarwin: We would also need access to `hasElementTranscludeDirective`. Let's not go there for now...

PR #11087, opened 2015-02-17
docs(ngRepeat): there is no clarification about 'as'

ghost:
as [variable name] is not an operator but rather a part of ngRepeat micro-syntax so it can be used only at the end (and not as operator, inside an expression)

There is no such clarification. It would be better to make such addition in it.


pkozlowski-opensource:
@FlamesoFF thnx, it makes sense to update docs like this. Could you please sign the CLA so we can merge it? Thnx!

FlamesoFF:
Ok, I did.

PR #11088, opened 2015-02-17
refactor($templateRequest): Usage of .finally is throwing errors in And...

dbabaioff:


campersau:
Just for reference: https://github.com/angular/angular.js/pull/11051 which includes the same change was opened 4 days ago

pkozlowski-opensource:
@dbabaioff thnx for the PR but I've ended up merging #11051 as it is the same patch but was opened earlier. Thnx for your work, though!

PR #11092, opened 2015-02-17
fix(select): remove unknown option when model is undefined and empty option is available

### LINKED ISSUE #11078: bug(select): switching from an invalid to undefined model value keeps invalid option in list
In 1.4 snapshot, given a select box without ng-options, switching from an invalid to undefined model value doesn't remove the '?' option (even though the value is correctly selected to the empty option): http://jsfiddle.net/xek8wqk1/

Resetting back to empty string is fine, but resetting to undefined doesn't remove the "?" value.

petebacondarwin:
Closes #11078

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -73,6 +73,7 @@ var SelectController =
       if (value === '') self.emptyOption.prop('selected', true); // to make IE9 happy
     } else {
       if (isUndefined(value) && self.emptyOption) {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@lgalfaso or @Narretz - if you have a moment can you just LGTM this PR?

Narretz:
LGTM!

petebacondarwin:
Thanks @Narretz

PR #11096, opened 2015-02-18
fix(form): allow dynamic form names which initially evaluate to blank

jbedard:
Previously the form name watching was initialized based on the initial name _value_ instead of the attribute being set.

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -454,9 +454,11 @@ var formDirectiveFactory = function(isNgForm) {
       name: 'form',
       restrict: isNgForm ? 'EAC' : 'E',
       controller: FormController,
-      compile: function ngFormCompile(formElement) {
+      compile: function ngFormCompile(formElement, attr) {
         // Setup initial state of the control
         formElement.addClass(PRISTINE_CLASS).addClass(VALID_CLASS);
+
+        var nameAttr = attr.name ? 'name' : (isNgForm && attr.ngForm ? 'ngForm' : false);
 
         return {
           pre: function ngFormPreLink(scope, formElement, attr, controller) {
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
LGTM after taking a quick look. Would be cool if someone more familiar with the form stuff have another look, though. @Narretz  / @petebacondarwin ?

src/ng/directive/form.js
```diff
@@ -488,23 +490,21 @@ var formDirectiveFactory = function(isNgForm) {
...
-                alias = newValue;
-                setter(scope, alias, controller, alias);
-                parentFormCtrl.$$renameControl(controller, alias);
+            var parentFormCtrl = controller.$$parentForm;
+
+            if (nameAttr) {
+              setter(scope, controller.$name, controller, controller.$name);
+              attr.$observe(nameAttr, function(newValue) {
+                if (controller.$name === newValue) return;
+                setter(scope, controller.$name, undefined, controller.$name);
```
Narretz: I realize this line was there before - but as far as I can tell, it doesn't do anything, because two lines after the setter is called again, this time with the correct value (also, all tests pass if I remove this)

jbedard: It is to remove the controller from the scope under the old name (although it doesn't actually remove it but sets it to undefined). I've added this expectation to the form renaming tests.

Narretz: Ah, yeah, that makes a lot of sense.

jbedard:
Closing this since it was merged.

petebacondarwin:
Was this supposed to go into master too? It looks like @Narretz only merged it into 1.3..x

Narretz:
It's here: https://github.com/angular/angular.js/commit/410f7c682633c681be641cd2a321f9e51671d474 I had to make adjustments for 1.3.x, as the setter takes different arguments. Not sure why the PR wasn't closed from the commit, maybe because the branches didn't match?

petebacondarwin:
It didn't close it because the commit doesn't have a `Closes #...` mention. :-)

Narretz:
But this one has :) https://github.com/angular/angular.js/commit/190ea883c588d63f8b900a8de1d45c6c9ebb01ec

petebacondarwin:
Yeah but that one wasn't merged to master (the default branch for this repo) :-P

Issue #10508, opened 2014-12-17
$http .success/.error return original promise, dissimilar from how .then works. Clarify in docs?


glebec:
Hello. `$http` returns a normal promise for a server response object (with `data`, `status`, `headers` etc.). As a normal promise, returning values from a handler in `.then` results in a new exported promise for the value (or assimilation, if the return value is a promise, or rejection in case of error, etc.). That is all well and good.

For convenience/abstraction, AngularJS provides two custom methods on promises returned by `$http`: `.success` and `.error`. Each takes a single handler with the `data`, `status` etc. as parameters.

The problem is that people familiar with promises will likely try to chain off of `.success`, perhaps by transforming and returning new values or new promises. However, `.success` does not return a new promise; rather, it returns the original promise:

```javascript
angular.module('promisesApp')
  .controller('MainCtrl', function ($http, $log) {

    $http.get('/api/things')
      .success(function (data) {
        $log.debug('in success:', data); // these are the things from the API
        return 'a new value';
      })
      .then(function (data) {
        $log.debug('in chained .then:', data);
        // data is the original server response object, NOT 'a new value'
        // data.data are the things from the API
      });

  });
```

If you were trying to make `.success` chainable in the same way `.then` is, it would be easy to change in the AngularJS source, and indeed I had started work on a pull request:

```javascript
// in theory, could change this:
promise.success = function(fn) {
  promise.then(function(response) {
    fn(response.data, response.status, response.headers, config);
  });
  return promise;
};
// to this:
promise.success = function(fn) {
  return promise.then(function(response) {
    return fn(response.data, response.status, response.headers, config);
  });
};
```

*However,* I then realized that this would mean you could not attach both `.success` and `.error` to a single promise. The reason you can write this code:

```javascript
$http.get('api/things')
  .success( successHandler )
  .error( errorHandler )
```

…is because `.success` and `.error` return the original promise, not a new promise representing the result of the previous method.

This leaves me in a bit of a quandary. The recommendation must therefore be that `.success` and `.error` are only suitable as "end of the road" promise consumers, when you do not intend to do any more post-processing / chaining. If that is the case, I think the AngularJS docs should reflect this, and make it explicitly clear that chaining off of `.success`/`.error` does not work in the normal promises fashion, since the original promise is returned.

What do you think?

PinkyJie:
Impressive!

likaiwalkman:
[...]

PR #11097, opened 2015-02-18
fix($http): promise chain for helper methods

PatrickJS:
can has less promise chain breaking

``` javascript
angular.module('app', [])
.factory('Auth', function($http) {
  return {
    login: function(data) {
      return $http.post('/signup', data)
        .success(function(data) {
          return data;
        })
        .error(function(err) {
          return $q.reject(err);
        });
    } // end login
  }; // end module
}) // end factory
.controller('MainCtrl', function($scope, $state, Auth) {

  $scope.login = function(user) {
    Auth.login(user).then(function(data) {
      $state.go('dashboard');
    })
    .catch(function(err) {
      alert(err);
    });
  }; // end login
}); // end ctrl
```

since the promise chain is broken when using `Auth.login().then(otherStuff).catch(errorStuff)` you get the result from the http and not the error from `.error`.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -807,17 +807,15 @@ function $HttpProvider() {
       }
 
       promise.success = function(fn) {
-        promise.then(function(response) {
-          fn(response.data, response.status, response.headers, config);
+        return promise.then(function(response) {
+          return fn(response.data, response.status, response.headers, config);
         });
-        return promise;
       };
 
       promise.error = function(fn) {
-        promise.then(null, function(response) {
-          fn(response.data, response.status, response.headers, config);
+        return promise.then(null, function(response) {
+          return fn(response.data, response.status, response.headers, config);
         });
-        return promise;
       };
 
       return promise;


gkalpak:
Based on the discussion at #10508 (especially https://github.com/angular/angular.js/issues/10508#issuecomment-67734857), it seems like `success()` and `error()` are heading for deprecation.

Note that fixing their "brokenness" would be quite of a breaking change.

pkozlowski-opensource:
I'm not big fan of this change - for a start it breaks the existing tests, so this would have to be looked into. Besides it tries to pretend that `success` and `error` are promises (or promise-like) while those are ordinary callbacks.

I think that the current situation - although broken - is easier to explain: `then` is for promises while `error` / `success` calbacks are ... plain old callbacks that can be used by people who don't want mess with promises.

My best advice - let go of `error` / `success`....

Narretz:
I agree with @pkozlowski-opensource . Let's not shoehorn promise behavior onto callbacks.

pkozlowski-opensource:
Yeh, thnx for the suggestion @gdi2290 but we are going to pass on this one...

PR #11104, opened 2015-02-19
doc(ngIf): mention multiElement in documentation

marcin-wosinek:


src/ng/directive/ngIf.js
```diff
@@ -32,6 +32,9 @@
  * Additionally, you can provide animations via the `ngAnimate` module to animate the `enter`
  * and `leave` effects.
  *
+ * You can use ng-if across few dom elements. First element needs `ng-if-start="expresion"`, and the
+ * last one `ng-if-end`. Those two and all elements between will be hide/appear together.
```
gkalpak: _will hide_ (without _be_)

gkalpak:
I am not sure this should be documentd separately on every `multiElement` directive (currently `ngRepeat`, `ngShow`, `ngHide`, `ngIf`, `ngSwitchWhen`, `ngSwitchDefault`).

But it might be a good idea to introduce a `@multiElement` ngdoc annotation and have dgeni automatically add a note on such directives (maybe with a link to the https://docs.angularjs.org/api/ng/service/$compile#-multielement-).

marcin-wosinek:
while `@multiElement` seems to be reasonable, I wonder if all use cases are similar enough to be covered by one template.

gkalpak:
It could be a minimal template pointing to the more detailed explanation in [`$compile`'s docs][1].

[1]: https://docs.angularjs.org/api/ng/service/%24compile#-multielement-

Narretz:
I agree with @gkalpak, an annotation seems like a good idea.

marcin-wosinek:
Do I get it right that this annotation should be implemented in https://github.com/angular/dgeni-packages, by a new module similar to `ngdoc/tag-defs/priority.js` & a bit of template in `ngdoc/templates/api/directive.template.html`

Narretz:
Yes, that is correct! Though I'd like to know what @petebacondarwin thinks of introducing a tag for `multiElement`.

petebacondarwin:
Sounds good to me

petebacondarwin:
The change has landed on dgeni-packages master, just need to run a release then we will need a docs PR to update the directives that support multi-element

petebacondarwin:
The only problem left is that `ngSwitchWhen` and `ngSwitchDefault` are not documented as separate directives to `ngSwitch` so it is not possible to apply the `@multiElement` to them. We should probably just bite the bullet and put a block of text into `ngSwitch` explaining this case.

marcin-wosinek:
Cool!

Later I've realized that multiElement can be the better shown as part of **Usage** section as:

```html
<ANY
  ng-repeat-start="">
</ANY>
...
<ANY
  ng-repeat-end>
</ANY>
```

But with directive in place, this can be changed later.

petebacondarwin:
@marcin-wosinek - nice idea. If you fancied trying to do a PR with that change that would be cool

PR #11106, opened 2015-02-19
using ng-checked="expression" will never update the respective ng-model value.

pankajparkar:
I'm adding description so that everyone know that updating ng-checked for checked will never update the binding between ng-model variable.

Plunkr
http://plnkr.co/edit/vdVMmhe0rY1aGzX4koyM?p=preview


Narretz:
I don't understand what you mean. `ngChecked` will only set the `checked` attribute, it does not work together with ngModel in any way. Is that what you are trying to say? I#m particularly confused about the `ng-changed` part of the sentence.

pankajparkar:
@Narretz Ya, that's exactly i wanted to say..

PR #11116, opened 2015-02-20
fix(filterFilter): solve issue #10991 with null property value when using objects with filter 

ElephantRose:
fix(filterFilter): solve issue #10991 with null property value when using objects with filter 

 Regarding the issue #10991 and the following discussion #10992

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -187,7 +187,7 @@ function createPredicateFn(expression, comparator, matchAgainstAnyProp) {
 }
 
 function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatchWholeObject) {
-  var actualType = (actual !== null) ? typeof actual : 'null'; 
+  var actualType = (actual !== null) ? typeof actual : 'null';
   var expectedType = typeof expected;
 
   if ((expectedType === 'string') && (expected.charAt(0) === '!')) {
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -187,7 +187,7 @@ function createPredicateFn(expression, comparator, matchAgainstAnyProp) {
 }
 
 function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatchWholeObject) {
-  var actualType = typeof actual;
+  var actualType = (actual !== null) ? typeof actual : 'null'; 
   var expectedType = typeof expected;
 
   if ((expectedType === 'string') && (expected.charAt(0) === '!')) {
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -188,7 +188,7 @@ function createPredicateFn(expression, comparator, matchAgainstAnyProp) {
 
 function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatchWholeObject) {
   var actualType = (actual !== null) ? typeof actual : 'null';
-  var expectedType = typeof expected;
+  var expectedType = (expected !== null) ? typeof expected : 'null';
 
   if ((expectedType === 'string') && (expected.charAt(0) === '!')) {
     return !deepCompare(actual, expected.substring(1), comparator, matchAgainstAnyProp);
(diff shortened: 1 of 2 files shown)


test/ng/filter/filterSpec.js
```diff
@@ -428,6 +428,32 @@ describe('Filter: filter', function() {
...
+      var items = [
+          { office:1, people: {name:'john'}},
+          { office:2, people: {name:'jane'}},
+          { office:3, people: null}
+      ];
+      var f = { };
+      expect(filter(items, f).length).toBe(3);
+
+      f = { people:null };
+      expect(filter(items, f).length).toBe(3);
```
gkalpak: Hm...this test shouldn't pass (the length should be 1). But it does pass since version 1.3.6.
This made me realize that there is a similar bug with `expectedType` (see above).

src/ng/filter/filter.js
```diff
@@ -187,7 +187,7 @@ function createPredicateFn(expression, comparator, matchAgainstAnyProp) {
 }
 
 function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatchWholeObject) {
-  var actualType = typeof actual;
+  var actualType = (actual !== null) ? typeof actual : 'null';
   var expectedType = typeof expected;
```
gkalpak: The test made me realize that a similar "fix" is necessary for `expectedType`.

```js
var expectedType = (expected !== null) ? typeof expected : 'null';
```

test/ng/filter/filterSpec.js
```diff
@@ -428,6 +428,32 @@ describe('Filter: filter', function() {
...
+
+      //should throw an error in 1.3.12
```
gkalpak: These comments are not necessary (read "should be removed" :smile:).

gkalpak:
I left a couple of inline comments (seems that there is one more small fix to be done).
Also, the commits should be squashed into 1.

--
<sub>@ElephantRose: I understand that your time might be limited at the moment and trying to figure out all the necessary bits for a first contribution (git stuff (forking, branching, rebasing, squashing, force-pushing), github stuff (submitting PR, updating PR), angular conventions (code style, file/folder layout, commit message formats, build/testing process), tooling (npm, grunt, gulp) etc) can be time-consuming (and frustrating at times) - although I promise it is going to be a no-brainer after a few contributions.</sub>

<sub>So, there are a few things that need to be done in order for this PR to be 100% complete, but if you don't want to spend any more time on it right now, it is good enough for merging (it does solve part of the problem and squashing can be done while merging). The rest can be taken care of in subsequent PRs (I'll take care of that).
What I am trying to say is, if you are willing/able to put in the extra cycles needed to make this fix 100% complete you are more than welcome to go for it. If not, we can merge it as is and someone else will take it from there.</sub>

<sub>In any case, thx for all the work so far :+1:</sub>

gkalpak:
@caitp or @petebacondarwin or @anyone, should this be backported to 1.3.x (since it was working correclty up until 1.3.5 and broke in 1.3.6) ?

caitp:
yes, backport to 1.3

pkozlowski-opensource:
@gkalpak what is the milestone in which you want to merge this?

gkalpak:
@pkozlowski-opensource, I don't really know what's the deal with milestones (nor do I know when the next release is happening (suspecting on Monday)) and how I should determine which one I want t merge this in :)

If it helps answer the question, I would like to merge this as soon as possible (since it is a regression), preferrably with the additional fix mentioned above, but if not possible then as is (and submitting a follow-up PR).

ElephantRose:
@gkalpak, Sorry for the late reply. I was a bit busy.
I changed the null test for expected value as you mentionned and updated the tests as well (comments have been deleted too btw)

gkalpak:
@ElephantRose, thx ! Apart for the commits' needing to be squashed into one, it LGTM.

gkalpak:
@petebacondarwin, according to https://github.com/angular/angular.js/pull/11116#issuecomment-75253772 this should be also backported to 1.3.x

pkozlowski-opensource:
@gkalpak wasn't it done via 01161a0e9fb1af93e9f06535aed8392ed7f116a4 ?

gkalpak:
@pkozlowski-opensource, @petebacondarwin, ooops ! Seems like I missed that. All good then :)

PR #11119, opened 2015-02-20
refactor(ngSwitch):delete default directive restriction

pkozlowski-opensource:
delete the directive restriction to avoid code duplication since the default is already applied

Closes #10987


pkozlowski-opensource:
This is rebased #10987 to check on TravisCI

PR #11132, opened 2015-02-22
docs(versioning): modify versioning for < v1.0.1

t-jindai:
less than angular v1.0.2 dose not support Semantic Versioning. so You can not access api docs because of difference of file path. fix it.

If merge this PR You can access less than v1.0.2 api docs from versioning select menu on api doc.


petebacondarwin:
Nice one @kashiro - I will review.

PR #11133, opened 2015-02-22
docs(ngDisabled): improve wording

### LINKED ISSUE #11032: Did someone make a mistake editing the docs for ng-disabled
Just noticed in large bold letters at the top:

"We shouldn't do this, because it will make the button enabled on Chrome/Firefox but not on IE8 and older IEs:"

Looks like this was added as a passing comment by someone and not part of the doc page.

pkozlowski-opensource:
Fixes #11032

### DIFF
--- a/src/ng/directive/attrs.js
+++ b/src/ng/directive/attrs.js
@@ -159,17 +159,18 @@
  *
  * @description
  *
- * We shouldn't do this, because it will make the button enabled on Chrome/Firefox but not on IE8 and older IEs:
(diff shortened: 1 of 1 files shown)


Narretz:
I just did the same: https://github.com/angular/angular.js/pull/11138 (: I cleaned it up some more, if you're okay with it I'll merge mine.

pkozlowski-opensource:
@Narretz sure, whatever you feel is better!

PR #11144, opened 2015-02-23
docs(orderBy): Start with a simpler example.

edthedev:
Per the question raised at this Stack Overflow question:
http://stackoverflow.com/questions/24048590/angularjs-ng-repeat-orderby-date-not-working

The first example on this page is too complex to convey the simplest possible case for using this function.

### DIFF
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -34,6 +34,41 @@
  * @param {boolean=} reverse Reverse the order of the array.
  * @returns {Array} Sorted copy of the source array.
  *
+ * To make AngularJS perform a sort, add orderBy to the ng-repeat. 
+ * The example below demonstrates setting the predicate to '-age' and reverse * to false.
+ *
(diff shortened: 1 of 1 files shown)


edthedev:
cla:yes.

Narretz:
This is a good idea. However, I think it would be better if the new example 1 and example 2 had an introductory sentence, explaining which features of order By are used in the example. Could you add this?

edthedev:
Sounds good. Thanks.

I have updated the commit (with a rebase / squash) to add a couple of sentences describing each example.

Narretz:
Thanks, landed! I had to make a few adjustments - the new example didn't actually work: if you don't use a scope property for your predicate, you need to make it a string literal: `'-age'`

edthedev:
You're very welcome. I'm glad you caught my mistake, too.

 The updated example will be very helpful for programmers who just recently speed Angular. Thanks for merging it.

PR #11150, opened 2015-02-23
docs($animate): describe your change...

GregM:
Adding a clarifying sentence to $animate method to explain that a CSS Style in `from` will be applied immediately if there is no corresponding CSS Style in `to`.


GregM:
I signed the CLA

PR #11152, opened 2015-02-23
feat($interpolate): MessageFormat extensions

chirayuk:
See [**plunker example**](http://plnkr.co/edit/QBVRQ70dvKZDWmHW9RyR?p=preview)

Extend interpolation with MessageFormat like syntax.

Ref: https://docs.google.com/a/google.com/document/d/1pbtW2yvtmFBikfRrJd8VAsabiFkKezmYZ_PbgdjQOVU/edit

Example:

``` html

{{recipients.length, plural, offset:1
    =0 {You gave no gifts}
    =1 { {{ recipients[0].gender, select,
              male {You gave him a gift.}
              female {You gave her a gift.}
              other {You gave them a gift.}
          }}
       }
    one { {{ recipients[0].gender, select,
              male {You gave him and one other person a gift.}
              female {You gave her and one other person a gift.}
              other {You gave them and one other person a gift.}
          }}
       }
    other {You gave {{recipients[0].gender}} and # other people gifts. }
}}
```

This is a SEPARATE module so you MUST include
angular-messageformat.min.js.  In addition, your application module
should depend on the "ngMessageFormat".
(e.g. angular.module('myApp', ['ngMessageFormat']);)

$interpolate automatically gets the new behavior.

Quick note on syntax differences from MessageFormat:
- MessageFormat directives are always inside {{ }} instead of
  single { }.  This ensures a consistent interpolation syntax (else you
  could interpolate in more than one way and have to pick one based on
  feature availability for that syntax.)
- The first word inside such syntax can be an arbitrary Angular
  expression instead of a single identifier.
- You can nest them as deep as you want.  As mentioned earlier, you
  would use {{ }} to start the nested interpolation that may optionally
  include select/plural extensions.
- Only "select" and "plural" keywords are currently recognized.
- Quoting support is coming in a future commit.
- Positional arguments/placeholders are not supported. They don't make
  sense in Angular templates anyway (they are only helpful when using
  API calls from a programming language.)
- Redefining of the startSymbol and endSymbol used for interpolation is
  not currently supported yet.

### DIFF
--- a/docs/content/error/$interpolate/dupvalue.ngdoc
+++ b/docs/content/error/$interpolate/dupvalue.ngdoc
@@ -1,0 +1,11 @@
+@ngdoc error
+@name $interpolate:dupvalue
+@fullName Duplicate choice in plural/select
+@description
+
+You have repeated a match selection for your plural or select MessageFormat
+extension in your interpolation expression.  The different choices have to be unique.
+
+For more information about the MessageFormat syntax in interpolation
+expressions, please refer to MessageFormat extensions section at
+{@link guide/i18n#MessageFormat Angular i18n MessageFormat}
--- a/docs/content/error/$interpolate/logicbug.ngdoc
+++ b/docs/content/error/$interpolate/logicbug.ngdoc
@@ -1,0 +1,12 @@
+@ngdoc error
+@name $interpolate:logicbug
+@fullName Bug in ngMessageFormat module
+@description
+
+You've just hit a bug in the ngMessageFormat module provided by angular-messageFormat.min.js.
+Please file a github issue for this and provide the interpolation text that caused you to hit this
+bug mentioning the exact version of AngularJS used and we will fix it!
+
+For more information about the MessageFormat syntax in interpolation
+expressions, please refer to MessageFormat extensions section at
+{@link guide/i18n#MessageFormat Angular i18n MessageFormat}
--- a/docs/content/error/$interpolate/nochgmustache.ngdoc
+++ b/docs/content/error/$interpolate/nochgmustache.ngdoc
@@ -1,0 +1,17 @@
+@ngdoc error
+@name $interpolate:nochgmustache
+@fullName Redefinition of start/endSymbol incompatible with MessageFormat extensions
+@description
+
+You have redefined `$interpolate.startSymbol`/`$interpolate.endSymbol` and also
+loaded the `ngMessageFormat` module (provided by angular-messageFormat.min.js)
+while creating your injector.
+
+`ngMessageFormat` currently does not support redefinition of the
+startSymbol/endSymbol used by `$interpolate`.  If this is affecting you, please
+file an issue and mention @chirayuk on it.  This is intended to be fixed in a
+future commit and the github issue will help gauge urgency.
+
+For more information about the MessageFormat syntax in interpolation
+expressions, please refer to MessageFormat extensions section at
+{@link guide/i18n#MessageFormat Angular i18n MessageFormat}
--- a/docs/content/error/$interpolate/reqother.ngdoc
+++ b/docs/content/error/$interpolate/reqother.ngdoc
@@ -1,0 +1,13 @@
+@ngdoc error
+@name $interpolate:reqother
+@fullName Required choice "other" for select/plural in MessageFormat
+@description
+
+Your interpolation expression with a MessageFormat extension for either
+"plural" or "select" (typically used for gender selection) does not contain a
+message for the choice "other".  Using either select or plural MessageFormat
+extensions require that you provide a message for the selection "other".
+
+For more information about the MessageFormat syntax in interpolation
+expressions, please refer to MessageFormat extensions section at
+{@link guide/i18n#MessageFormat Angular i18n MessageFormat}
--- a/docs/content/guide/i18n.ngdoc
+++ b/docs/content/guide/i18n.ngdoc
@@ -137,3 +137,115 @@ The Angular datetime filter uses the time zone settings of the browser. The same
 application will show different time information depending on the time zone settings of the
 computer that the application is running on. Neither JavaScript nor Angular currently supports
 displaying the date with a timezone specified by the developer.
+
+
+<a name="MessageFormat"></a>
+## MessageFormat extensions 
+
+AngularJS interpolations via `$interpolate` and in templates
+support an extended syntax based on a subset of the ICU
+MessageFormat that covers plurals and gender selections.
+
+Please refer to our [design doc](https://docs.google.com/a/google.com/document/d/1pbtW2yvtmFBikfRrJd8VAsabiFkKezmYZ_PbgdjQOVU/edit)
+for a lot more details.  You may find it helpful to play with our [Plnkr Example](http://plnkr.co/edit/QBVRQ70dvKZDWmHW9RyR?p=preview).
+
+You can read more about the ICU MessageFormat syntax at
+[Formatting Messages | ICU User Guide](http://userguide.icu-project.org/formatparse/messages#TOC-MessageFormat).
+
+This extended syntax is provided by way of the
+`ngMessageFormat` module that your application can depend
+upon (shipped separately as `angular-messageFormat.min.js` and
+`angular-messageFormat.js`.)  A current limitation of the
+`ngMessageFormat` module, is that it does not support
+redefining the `$interpolate` start and end symbols. Only the
+default `{{` and `}}` are allowed.
+
+This syntax extension, while based on MessageFormat, has
+been designed to be backwards compatible with existing
+AngularJS interpolation expressions.  The key rule is simply
+this: **All interpolations are done inside double curlies.**
+The top level comma operator after an expression inside the
+double curlies causes MessageFormat extensions to be
+recognized.  Such a top level comma is otherwise illegal in
+an Angular expression and is used by MessageFormat to
+specify the function (such as plural/select) and it's
+related syntax.
+
+To understand the extension, take a look at the ICU
+MessageFormat syntax as specified by the ICU documentation.
+Anywhere in that MessageFormat that you have regular message
+text and you want to substitute an expression, just put it
+in double curlies instead of single curlies that
+MessageFormat dictates.  This has a huge advantage.  **You
+are no longer limited to simple identifiers for
+substitutions**.  Because you are using double curlies, you
+can stick in any arbitrary interpolation syntax there,
+including nesting more MessageFormat expressions!  Some
+examples will make this clear.  In the following example, I
+will only be showing you the AngularJS syntax.
+
+
+### Simple plural example
+
+```
+{{numMessages, plural,
+      =0 { You have no new messages }
+      =1 { You have one new message }
+   other { You have # new messages }
+}}
+```
+
+While I won't be teaching you MessageFormat here, you will
+note that the `#` symbol works as expected.  You could have
+also written it as:
+
+```
+{{numMessages, plural,
+      =0 { You have no new messages }
+      =1 { You have one new message }
+   other { You have {{numMessages}} new messages }
+}}
+```
+
+where you explicitly typed in `numMessages` for "other"
+instead of using `#`.  They are nearly the same except if
+you're using "offset".  Refer to the ICU MessageFormat
+documentation to learn about offset.
+
+Please note that **other** is a **required** category (for
+both the plural syntax and the select syntax that is shown
+later.)
+
+
+### Simple select (for gender) example
+
+```
+{{friendGender, select,
+       male { Invite him }
+     female { Invite her }
+      other { Invite them }
+}}
+```
+
+### More complex example that combines some of these
+
+This is taken from the [plunker example](http://plnkr.co/edit/QBVRQ70dvKZDWmHW9RyR?p=preview) linked to earlier.
+
+```
+{{recipients.length, plural, offset:1
+    =0 {You ({{sender.name}}) gave no gifts}
+    =1 { {{ recipients[0].gender, select,
+              male {You ({{sender.name}}) gave him ({{recipients[0].name}}) a gift.}
+              female {You ({{sender.name}}) gave her ({{recipients[0].name}}) a gift.}
+              other {You ({{sender.name}}) gave them ({{recipients[0].name}}) a gift.}
+          }}
+       }
+    one { {{ recipients[0].gender, select,
+              male {You ({{sender.name}}) gave him ({{recipients[0].name}}) and one other person a gift.}
+              female {You ({{sender.name}}) gave her ({{recipients[0].name}}) and one other person a gift.}
+              other {You ({{sender.name}}) gave them ({{recipients[0].name}}) and one other person a gift.}
+          }}
+       }
+    other {You ({{sender.name}}) gave {{recipients.length}} people gifts. }
+}}
+```
--- a/src/ng/interpolate.js
+++ b/src/ng/interpolate.js
@@ -1,6 +1,16 @@
 'use strict';
 
-var $interpolateMinErr = minErr('$interpolate');
+var $interpolateMinErr = angular.$interpolateMinErr = minErr('$interpolate');
+$interpolateMinErr.throwNoconcat = function(text) {
+  throw $interpolateMinErr('noconcat',
+      "Error while interpolating: {0}\nStrict Contextual Escaping disallows " +
+      "interpolations that concatenate multiple expressions when a trusted value is " +
+      "required.  See http://docs.angularjs.org/api/ng.$sce", text);
+};
+
+$interpolateMinErr.interr = function(text, err) {
+  return $interpolateMinErr('interr', "Can't interpolate: {0}\n{1}", text, err.toString());
+};
 
 /**
  * @ngdoc provider
@@ -244,10 +254,7 @@ function $InterpolateProvider() {
       // make it obvious that you bound the value to some user controlled value.  This helps reduce
       // the load when auditing for XSS issues.
       if (trustedContext && concat.length > 1) {
-          throw $interpolateMinErr('noconcat',
-              "Error while interpolating: {0}\nStrict Contextual Escaping disallows " +
-              "interpolations that concatenate multiple expressions when a trusted value is " +
-              "required.  See http://docs.angularjs.org/api/ng.$sce", text);
+          $interpolateMinErr.throwNoconcat(text);
       }
 
       if (!mustHaveExpression || expressions.length) {
@@ -277,9 +284,7 @@ function $InterpolateProvider() {
 
               return compute(values);
             } catch (err) {
-              var newErr = $interpolateMinErr('interr', "Can't interpolate: {0}\n{1}", text,
-                  err.toString());
-              $exceptionHandler(newErr);
+              $exceptionHandler($interpolateMinErr.interr(text, err));
             }
 
           }, {
@@ -304,9 +309,7 @@ function $InterpolateProvider() {
           value = getValue(value);
           return allOrNothing && !isDefined(value) ? value : stringify(value);
         } catch (err) {
-          var newErr = $interpolateMinErr('interr', "Can't interpolate: {0}\n{1}", text,
-            err.toString());
-          $exceptionHandler(newErr);
+          $exceptionHandler($interpolateMinErr.interr(text, err));
         }
       }
     }
@@ -346,4 +349,5 @@ function $InterpolateProvider() {
     return $interpolate;
   }];
 }
+
 
--- a/src/ngMessageFormat/messageFormatCommon.js
+++ b/src/ngMessageFormat/messageFormatCommon.js
@@ -1,0 +1,77 @@
+'use strict';
+
+// NOTE: ADVANCED_OPTIMIZATIONS mode.
+//
+// This file is compiled with Closure compiler's ADVANCED_OPTIMIZATIONS flag! Be wary of using
+// constructs incompatible with that mode.
+
+var $interpolateMinErr = angular['$interpolateMinErr'];
+
+var noop = angular['noop'],
+    isFunction = angular['isFunction'],
+    toJson = angular['toJson'];
+
+function stringify(value) {
+  if (value == null /* null/undefined */) { return ''; }
+  switch (typeof value) {
+    case 'string':     return value;
+    case 'number':     return '' + value;
+    default:           return toJson(value);
+  }
+}
+
+// Convert an index into the string into line/column for use in error messages
+// As such, this doesn't have to be efficient.
+function indexToLineAndColumn(text, index) {
+  var lines = text.split(/\n/g);
+  for (var i=0; i < lines.length; i++) {
+    var line=lines[i];
+    if (index >= line.length) {
+      index -= line.length;
+    } else {
+      return { line: i + 1, column: index + 1 };
+    }
+  }
+}
+var PARSE_CACHE_FOR_TEXT_LITERALS = Object.create(null);
+
+function parseTextLiteral(text) {
+  var cachedFn = PARSE_CACHE_FOR_TEXT_LITERALS[text];
+  if (cachedFn != null) {
+    return cachedFn;
+  }
+  function parsedFn(context) { return text; }
+  parsedFn['$$watchDelegate'] = function watchDelegate(scope, listener, objectEquality) {
+    var unwatch = scope['$watch'](noop,
+        function textLiteralWatcher() {
+          if (isFunction(listener)) { listener.call(null, text, text, scope); }
+          unwatch();
+        },
+        objectEquality);
+    return unwatch;
+  };
+  PARSE_CACHE_FOR_TEXT_LITERALS[text] = parsedFn;
+  parsedFn.exp = text; // Needed to pretend to be $interpolate for tests copied from interpolateSpec.js
+  parsedFn.expressions = []; // Require this to call $compile.$$addBindingInfo() which allows Protractor to find elements by binding.
+  return parsedFn;
+}
+
+function subtractOffset(expressionFn, offset) {
+  if (offset === 0) {
+    return expressionFn;
+  }
+  function minusOffset(value) {
+    return (value == void 0) ? value : value - offset;
+  }
+  function parsedFn(context) { return minusOffset(expressionFn(context)); }
+  var unwatch;
+  parsedFn['$$watchDelegate'] = function watchDelegate(scope, listener, objectEquality) {
+    unwatch = scope['$watch'](expressionFn,
+        function pluralExpressionWatchListener(newValue, oldValue) {
+          if (isFunction(listener)) { listener.call(null, minusOffset(newValue), minusOffset(oldValue), scope); }
+        },
+        objectEquality);
+    return unwatch;
+  };
+  return parsedFn;
+}
--- a/test/ngMessageFormat/messageFormatSpec.js
+++ b/test/ngMessageFormat/messageFormatSpec.js
@@ -1,0 +1,699 @@
+'use strict';
+
+/* TODO: Add tests for:
+ • Whitespace preservation in messages.
+ • Whitespace ignored around syntax except for offset:N.
+ • Escaping for curlies and the # symbol.
+ • # symbol value.
+ • # symbol value when gender is nested inside plural.
+ • Error with nested # symbol.
+ • parser error messages.
+ • caching.
+ • watched expressions.
+ • test parsing angular expressions
+ • test the different regexes
+ • test the different starting rules
+*/
+
+describe('$$ngMessageFormat', function() {
+  describe('core', function() {
+    var $$messageFormat, $parse, $interpolate, $locale, $rootScope;
+
+    function Person(name, gender) {
+      this.name = name;
+      this.gender = gender;
+    }
+
+    var alice   = new Person("Alice", "female"),
+        bob     = new Person("Bob", "male"),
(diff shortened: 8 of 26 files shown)


PascalPrecht:
@chirayuk thank you for implementing this! Where can I start helping out for the docs?

chirayuk:
Changes since the last time:

-   Separate self contained module.  Use `angular-messageformat.js` and depend on `ngMessageFormat`.
-   NO changes to existing `$interpolate` or `$parse`.  I did a major refactor and the core no longer knows or cares about this.  However, this is a private service and relies on core implementation.
-   Full $interpolate replacement
    -   Supports mustHaveExpression, trustedContext and allOrNothing.
    -   Escaping compatible with $interpolate.
    -   Includes the *entire contents* of interpolateSpec.js to test the replacement behavior.  See <https://github.com/angular/angular.js/pull/11152/files#diff-3746da2fc13376182459459ead442464R229>
-   Support for `#` symbol to mean (number-offset) inside plural messages.  (For compatibility with messageformat.)
-   Escaping support inside MessageFormat expressions.
-   Accurate recognition of angular expressions—understands subexpressions, strings/string escapes, etc.
-   Compilation with Closure's `ADVANCED_OPTIMIZATIONS` mode.
-   And other stuff I can't remember. :)

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+ *
+ * @description
+ * Angular internal service to recognize MessageFormat extensions in interpolation expressions.
+ * For more information, see:
+ * https://docs.google.com/a/google.com/document/d/1pbtW2yvtmFBikfRrJd8VAsabiFkKezmYZ_PbgdjQOVU/edit
+ */
+
+var noop = angular['noop'],
+    isFunction = angular['isFunction'],
+    toJson = angular['toJson'];
```
petebacondarwin: Why use lookup (`angular['noop']`) rather than property (`angular.noop`) syntax here. Is it because of the Google Closure compilation optimizations?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+
+function stringify(value) {
+  if (value == null) { return ''; }
```
petebacondarwin: `==` is to catch `undefined` too?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+  switch (typeof value) {
+    case 'string':     return value;
+    case 'number':     return '' + value;
+    default:           return toJson(value);
+  }
+}
+
+var PARSE_CACHE_FOR_TEXT_LITERALS = new Object(null);
+var PARSE_CACHE_FOR_INTERPOLATIONS = new Object(null);
```
petebacondarwin: Do you mean to use `Object.create(null)` here, which creates objects with no prototype, rather than `new Object(null)`, which is really just an empty object: `{}`?

petebacondarwin:
There are a bunch of jshint and jscs errors. Can you ensure that `grunt ci-checks` passes?

petebacondarwin:
I think that the filenames should be `messageFormat.js` and `messageFormatSpec.js` for consistency with other angular files.

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+
+function parseTextLiteral(text) {
+  var cachedFn = PARSE_CACHE_FOR_TEXT_LITERALS[text];
+  if (cachedFn != null) {
+    return cachedFn;
+  }
+  function parsedFn(context) { return text; };
+  var unwatch;
```
petebacondarwin: Why is this `unwatch` variable declared outside the `watchDelegate` function?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+        function () {
+          if (isFunction(listener)) { listener.call(null, text, text, scope); }
+          unwatch();
+        },
+        objectEquality);
+    return unwatch;
+  };
+  PARSE_CACHE_FOR_TEXT_LITERALS[text] = parsedFn;
+  if (goog.DEBUG) {
+    // Only needed in order to pretend to be $interpolate for tests copied from interpolateSpec.js
```
petebacondarwin: These properties are also used by `$compile.$$addBindingInfo()` to allow Protractor to find elements by binding. We need to have these here at least for the unminified build. Is it true that `goog.DEBUG` will be truthy except in the minified build or is it only truthy when running the tests?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+    parsedFn.exp = text;
+    parsedFn.expressions = [];
+  }
+  return parsedFn;
+}
+
+function subtractOffset(expressionFn, offset) {
+  if (offset == 0) {
```
petebacondarwin: `===`?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+function MessageSelectorWatchers(msgSelector, scope, listener, objectEquality) {
+  var self = this;
+  this.scope = scope;
+  this.msgSelector = msgSelector;
+  this.listener = listener;
+  this.objectEquality = objectEquality;
+  this.lastMessage = void 0;
+  this.messageFnWatcher = noop;
+  var expressionFnListener = function(newValue, oldValue) { return self.expressionFnListener(newValue, oldValue); };
+  this.expressionFnWatcher = scope['$watch'](msgSelector.expressionFn, expressionFnListener, objectEquality);
```
petebacondarwin: Could we not simplify some of these uses of instance methods as callbacks using `Function.bind()`...

```js
this.expressionFnWatcher = scope['$watch'](msgSelector.expressionFn, this.expressionFnListener.bind(this), objectEquality);
```

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+  var argType = match[1];
+  switch (argType) {
+    case "plural": this.rule = this.rulePluralStyle; break;
+    case "select": this.rule = this.ruleSelectStyle; break;
+    default: this.errorInParseLogic();
+  }
+};
+
+MessageFormatParser.prototype.rulePluralStyle = function rulePluralStyle() {
+  this.choices = {};
```
petebacondarwin: `Object.create(null)`?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+    this.choiceKey = parseInt(match[1], 10);
+  } else {
+    this.choiceKey = match[2];
+  }
+  if (this.choices.hasOwnProperty(this.choiceKey)) {
+    var position = indexToLineAndColumn(this.text, this.match.index);
+    throw $$messageFormatMinErr('dupvalue',
+        'The choice "{0}" is specified more than once. Duplicate key is at line {1}, column {2} in text "{3}"',
+        this.choiceKey, position.line, position.column, this.text);
+  }
```
petebacondarwin: This check for duplicate `choiceKey` is repeated code - it could be factored out to DRY this up.

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+       *}
+       *return parsedFn;
+       */
+    }
+
+    return {
+      'interpolate': function cachedInterpolate(text, mustHaveExpression, trustedContext, allOrNothing) {
+        return cacheCall(interpolate, text, mustHaveExpression, trustedContext, allOrNothing);
+      },
+      'parseMustache': function cachedParseMustache(text, mustHaveExpression, trustedContext, allOrNothing) {
```
petebacondarwin: The `parseMustache` code paths never seem to be used. Is this for some unimplemented feature? It seems like we could get rid of a bunch of code if we removed this call and the `parser.ruleMustache` stuff.

petebacondarwin:
I think it would make it easier to grok this module if it was broken up into a few smaller src files. For instance you have message selector stuff, parsing rules, decorating $interpolate.

I can see that the code is generally structured as classes. In the theme of writing stuff that will run the same code base in V2 as V1, perhaps this code could be written as ES6 classes and then transpiled down for this V1 `ngMessageFormat` module?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+};
+
+function getEndOperator(opBegin) {
+  switch (opBegin) {
+    case "{": return "}";
+    case "[": return "]";
+    case "(": return ")";
+    default: return null;
+  }
+}
```
petebacondarwin: Would this (and `getBeginOperator`) benefit from being written as:

```js
var END_OPERATORS { '{': '}', '[': ']', '(': ')' };
function getEndOperator(opBegin) {
 return END_OPERATORS[opBegin];
}
```

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+var INTERP_END_RE = /\s*}}/g;
+MessageFormatParser.prototype.ruleEndMustache = function ruleEndMustache() {
+  var match = this.matchRe(INTERP_END_RE);
+  if (match == null) {
+    var position = indexToLineAndColumn(this.text, this.index);
+    throw $$messageFormatMinErr('reqendinterp',
+        'Expecting end of interpolation symbol, "{0}", at line {1}, column {2} in text "{3}"',
+        '}}', position.line, position.column, this.text);
+  }
+  if (this.parsedFn == null) {
```
petebacondarwin: In what situation would the `parsedFn` already be defined here?
Or alternatively, why would the `parsedFn` not have been computed here already?

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+  var match = re.exec(this.text);
+  if (match != null && match.index == this.index) {
+    this.index = re.lastIndex;
+    return true;
+  } else {
+    return false;
+  }
+};
+
+// Run through our grammar avoiding deeply nested function call chains.
```
petebacondarwin: Good idea

src/ngMessageFormat/messageformat.js
```diff
@@ -0,0 +1,938 @@
...
+  }
+  return null;
+};
+
+MessageFormatParser.prototype.searchRe = function searchRe(re) {
+  return this.matchRe(re, true);
+};
+
+
+MessageFormatParser.prototype.consumeRe = function consumeRe(re) {
```
petebacondarwin: Can this method not just use `matchRe`?

```js
return !!this.matchRe(re, false);
```

petebacondarwin:
A few nitpicks and questions but generally looking good. It would be nice to see a state diagram of the parser somewhere for good measure.

src/ngMessageFormat/messageformat.js
chirayuk: Yes.  I'm using the `ADVANCED_OPTIMIZATIONS` mode of the Closure compiler which will rewrite `angular.noop` into something like `angular.x`.  This module has limited interaction with outside code (and injection has existing array syntax to guard against minification) so there aren't too many places that do this kind of explicit lookup.

src/ngMessageFormat/messageformat.js
chirayuk: Yup.  Added a comment to make it clear.

src/ngMessageFormat/messageformat.js
chirayuk: Fixed.  Thanks for catching that!

src/ngMessageFormat/messageformat.js
chirayuk: Removed.  It was cruft from a refactor.

src/ngMessageFormat/messageformat.js
chirayuk: Ah.  I knew there was something I was missing but hadn't found any helpful comments to guide me when I had grepped our code.

Yes, `goog.DEBUG` is `false` *only* in the minified version and if `true` otherwise which should allow Protractor to work correctly.  I went ahead and renamed the symbol to `goog.MINIFIED` and switched the bool value to make this clear.  In the minified build, because I'm using `ADVANCED_OPTIMIZATIONS`, the entire `if` block is optimized away.

src/ngMessageFormat/messageformat.js
chirayuk: Fixed.

src/ngMessageFormat/messageformat.js
chirayuk: I decided against it as a matter of balancing style and performance.  From the perspective of style, I think the entire file should stick to just one approach—either using `Function.bind` or an explicit wrapper—instead of having to pick at each call site.  It turns out that `Function.bind` is typically about 100x slower than the method used here.  I especially do not want to someday realize that such calls were chained in some flow paying 100x times 100x for the call.  The current approach also has the advantage that a .toString() on the wrapper function helps with debugging while Function.bind is useless for it.  So I've made a style choice to use the simple wrapper functions everywhere in this module.  (Ref: <http://stackoverflow.com/questions/18895305/will-function-prototype-bind-always-be-slow>)

src/ngMessageFormat/messageformat.js
chirayuk: Right!  Fixed.

src/ngMessageFormat/messageformat.js
chirayuk: Ran it through the dryer.  It's now nicely DRY! :)

src/ngMessageFormat/messageformat.js
chirayuk: Yes, it was for a bunch of intermediate commits before I would replace $interpolate completely.  I decided to replace $interpolate fully sooner than later so it's no longer required.

As a heads up, in a future commit, I plan to `$decorate` our original `$parse` to support these extensions to allow things like `ng-bind` to work well (as wells as Angular 2 as well.)  Luckily, the parser is architected such that one can parse a varied set of things just by picking the correct starting state and "enclosing" expressions such push onto the rule stack before dispatching to them so this shouldn't be big change to the parser.

src/ngMessageFormat/messageformat.js
chirayuk: A jsperf I wrote a little while back strongly favors the switch statement.  (<https://jsperf.com/map-vs-switch-vs-if>)  I don't particularly think one reads any clearer than the other so I went with the switch statement.

src/ngMessageFormat/messageformat.js
chirayuk: Added comment elaborating on this.

Here it is:
If we parsed a MessageFormat extension, (e.g. select/plural today, maybe more some other day), then the result *has* to be a string and those rules would have already set this.parsedFn.  If there was no MessageFormat extension, then there is no requirement to stringify the result and `parsedFn` isn't set.  We set it here.  While we could have set it unconditionally when exiting the Angular expression, I intend for us to not just replace `$interpolate`, but also to replace `$parse` in a future version (so ng-bind can work), and in such a case we do not want to unnecessarily stringify something if it's not going to be used in a string context.

src/ngMessageFormat/messageformat.js
chirayuk: Thanks. :)

src/ngMessageFormat/messageformat.js
chirayuk: Yes, it can and now it does, given that I can't use the sticky flag.  I'm leaving the comment in there.

chirayuk:
-   [X] ensure that grunt ci-checks passes
-   [X] filenames should be messageFormat.js and messageFormatSpec.js
-   [X] split into multiple files
-   [X] add error docs
-   [X] expand guide/i18n documentation

chirayuk:
@petebacondarwin @IgorMinar ready for another review.

docs/content/error/$interpolate/dupvalue.ngdoc
```diff
@@ -0,0 +1,12 @@
+@ngdoc error
+@name $interpolate:dupvalue
+@fullName Duplicate choice in plural/select
+@description
+
+You have repeated a match selection for your plural or select MessageFormat
+extension in your interpolation interpolation expression.  The different
```
petebacondarwin: imterpolation interpolation?

docs/content/error/$interpolate/logicbug.ngdoc
```diff
@@ -0,0 +1,11 @@
+@ngdoc error
+@name $interpolate:logicbug
+@fullName Bug in ngMessageFormat module
+@description
+
+You've just hit a bug in the ngMessageFormat module provided by provided by
```
petebacondarwin: provided by provided by?

docs/content/error/$interpolate/nochgmustache.ngdoc
```diff
@@ -0,0 +1,17 @@
...
+@description
+
+You have redefined `$interpolate.startSymbol`/`$interpolate.endSymbol` and also
+loaded the `ngMessageFormat` module (provided by angular-messageFormat.min.js)
+while creating your injector.
+
+`ngMessageFormat` currently does not support redefinition of the
+startSymbol/endSymbol used by `$interpolate`.  If this is affecting you, please
+file an issue and mention @chirayuk on it.  This is intended to be fixed in a
+future commit and the github issue will help gauage urgency.
```
petebacondarwin: guage

docs/content/error/$interpolate/reqother.ngdoc
```diff
@@ -0,0 +1,13 @@
+@ngdoc error
+@name $interpolate:reqother
+@fullName Required choice "other" for select/plural in MessageFormat
+@description
+
+Your interpolation expression with a MessageFormat extension for either
+"plural" or "select" (typically used for gender selection) does not contain a
+message for the choice "other".  Using either select or plural MessageFormat
+extensions require that you provide a message for the selection "other".
```
petebacondarwin: So for gender selectors you would need to provide: `male`, `female` **and** `other`? Or would you just put `male` and `other`?

docs/content/guide/i18n.ngdoc
```diff
@@ -137,3 +137,106 @@ The Angular datetime filter uses the time zone settings of the browser. The same
...
+
+You can read more about the ICU MessageFormat syntax at
+[Formatting Messages | ICU User Guide](http://userguide.icu-project.org/formatparse/messages#TOC-MessageFormat).
+
+This extended syntax is provided by way of the
+`ngMessageFormat` module that your application can depend
+upon (shipped separately as `angular-messageFormat.min.js` and
+`angular-messageFormat.js`.)  A current limitation of this
+module is that it is not compatible with `$interpolate` used
+with redefined start and end symbols.
```
petebacondarwin: I would rephrase this slightly to:

> A current limitation of the `ngMessageFormat` module, is that it does not support redefining the `$interpolate` start and end symbols. Only the default `{{` and `}}` are allowed.

docs/content/guide/i18n.ngdoc
```diff
@@ -137,3 +137,106 @@ The Angular datetime filter uses the time zone settings of the browser. The same
...
+
+This syntax extension, while based on MessageFormat, has
+been designed to be backwards compatible with existing
+AngularJS interpolation expressions.  The key rule is simply
+this: **All interpolations are done inside double curlies.**
+Consider valid MessageFormat syntax.  Anywhere in that
```
petebacondarwin: The sentence

> Consider valid MessageFomat syntax.

Doesn't seem to fit here?

docs/content/guide/i18n.ngdoc
```diff
@@ -137,3 +137,106 @@ The Angular datetime filter uses the time zone settings of the browser. The same
...
+MessageFormat that you have regular message text and you
+want to substitute an expression, just put it in double
+curlies instead of single curlies that MessageFormat
+dictates.  This has a huge advantage.  **You are no longer
+limited to simple identifiers for substitutions**.  Because
+you are using double curlies, you can stick in any arbitrary
+interpolation syntax there, including nesting more
+MessageFormat!  Some examples will make this clear.  In the
```
petebacondarwin: including nesting more MessageFormat expressions!

docs/content/guide/i18n.ngdoc
```diff
@@ -137,3 +137,106 @@ The Angular datetime filter uses the time zone settings of the browser. The same
...
+```
+{{numMessages, plural,
+      =0 { You have no new messages }
+      =1 { You have one new message }
+   other { You have # new messages }
+}}
+```
+
+While I won't be teaching you MessageFormat here, you will
+not that the `#` symbol works as expected.  You could have
```
petebacondarwin: note

docs/content/guide/i18n.ngdoc
```diff
@@ -137,3 +137,106 @@ The Angular datetime filter uses the time zone settings of the browser. The same
...
+`ngMessageFormat` module that your application can depend
+upon (shipped separately as `angular-messageFormat.min.js` and
+`angular-messageFormat.js`.)  A current limitation of this
+module is that it is not compatible with `$interpolate` used
+with redefined start and end symbols.
+
+This syntax extension, while based on MessageFormat, has
+been designed to be backwards compatible with existing
+AngularJS interpolation expressions.  The key rule is simply
+this: **All interpolations are done inside double curlies.**
```
petebacondarwin: I think it is worth, at some point in this guide, calling out the fact that it is the comma character that triggers the expression to be considered as a messageFormat expression.

src/ng/interpolate.js
```diff
@@ -1,6 +1,16 @@
...
+$interpolateMinErr.noconcat = function(text) {
+  return $interpolateMinErr('noconcat',
+      "Error while interpolating: {0}\nStrict Contextual Escaping disallows " +
+      "interpolations that concatenate multiple expressions when a trusted value is " +
+      "required.  See http://docs.angularjs.org/api/ng.$sce", text);
+};
+
+$interpolateMinErr.interr = function(text, err) {
+  return $interpolateMinErr('interr', "Can't interpolate: {0}\n{1}", text, err.toString());
+};
```
petebacondarwin: The downside of DRYing up these errors is that they cause yet more warnings when minifying the files...

```bash
>> build/angular-sanitize.js minified into build/angular-sanitize.min.js
build/angular-messageFormat.js:290: WARNING - Throw expression is not a minErr instance.
    throw $interpolateMinErr['noconcat'](originalText);
    ^

0 error(s), 1 warning(s)
>> build/angular-messageFormat.js minified into build/angular-messageFormat.min.js
>> build/angular-animate.js minified into build/angular-animate.min.js
build/angular.js:4297: WARNING - Throw expression is not a minErr instance.
          throw err;
          ^

build/angular.js:9967: WARNING - Throw expression is not a minErr instance.
            throw e;
            ^

build/angular.js:10290: WARNING - Throw expression is not a minErr instance.
          throw $interpolateMinErr.noconcat(text);
          ^

build/angular.js:11488: WARNING - Throw expression is not a minErr instance.
        throw e;
        ^

build/angular.js:15322: WARNING - Throw expression is not a minErr instance.
            throw e;
            ^

0 error(s), 5 warning(s)
```

test/ngMessageFormat/messageFormatSpec.js
```diff
@@ -0,0 +1,699 @@
+'use strict';
+
+/* TODO: Add tests for:
+ • Whitespace preservation in messages.
+ • Whitespace ignored around syntax except for offset:N.
+ • Escaping for curlies and the # symbol.
+ • # symbol value.
+ • # symbol value when gender is nested inside plural.
```
petebacondarwin: This seems like quite an important use case. I guess the actual code does already work. We should prioritize this test after we get this merged.

src/ngMessageFormat/messageFormatCommon.js
```diff
@@ -0,0 +1,79 @@
...
+          if (isFunction(listener)) { listener.call(null, text, text, scope); }
+          unwatch();
+        },
+        objectEquality);
+    return unwatch;
+  };
+  PARSE_CACHE_FOR_TEXT_LITERALS[text] = parsedFn;
+  if (!goog.MINIFIED) {
+    parsedFn.exp = text; // Needed to pretend to be $interpolate for tests copied from interpolateSpec.js
+    parsedFn.expressions = []; // Unminified builds require this to call $compile.$$addBindingInfo() which allows Protractor to find elements by binding.
```
petebacondarwin: I am pretty sure that we need this even for the minified build too. People do use Protractor on the minified builds. In fact, this is a good practice in general for Protractor tests since we can then catch certain errors that can be introduced by minification.

So I think we need this information even in minified builds.

Instead, we could check `$compile.$$addBindingInfo` to find whether we need to generate this information.

petebacondarwin:
Nice one @chirayuk - this looks even better!

The only serious concern I have is with the `bindingInfo` bits.

Other than that it looks good to merge, once the few docs fixes are in place.

docs/content/error/$interpolate/dupvalue.ngdoc
chirayuk: Fixed.  Thanks!

docs/content/error/$interpolate/logicbug.ngdoc
chirayuk: Fixed.  Thanks!

docs/content/error/$interpolate/nochgmustache.ngdoc
chirayuk: Fixed.  Thanks!

docs/content/error/$interpolate/reqother.ngdoc
chirayuk: Yup.  You provide all the choices that you expect to match and "other" if there is no match.  For plurals, for instance, you can omit categories that you think can't occur, but you must still include "other".  For gender, you should include all the choices which might require a different message.  Since male, female and other have different messages, typically, would include all three.  However, for a message like, "You have no posts", you might include *just* other as the "You" in English is gender neutral but might not be in a different language.

docs/content/guide/i18n.ngdoc
chirayuk: Done.

docs/content/guide/i18n.ngdoc
chirayuk: Rewrote it to make a little more sense.

docs/content/guide/i18n.ngdoc
chirayuk: Done.

docs/content/guide/i18n.ngdoc
chirayuk: Fixed!

docs/content/guide/i18n.ngdoc
chirayuk: Good point!  Done.

src/ng/interpolate.js
chirayuk: Fixed it.  Not super happy with the fix.  It adds a stack frame for `noconcat` and also breaks the symmetry a bit.  I think we should really make the `minErr` pass of the closure compiler smarter but that's a lot more work at this point in time.  I think that the current choice is an acceptable compromise for the present time.

test/ngMessageFormat/messageFormatSpec.js
chirayuk: Poring over what little information is there on the ICU site does not shed information on whether this should be valid.  The 3rd party JS lib supports "#" even in select without a plural.  That might be a bug.  I don't like the "#" in general because it has a bunch of limitations.  So this won't work today and if it break someone, they can file a bug and I'll fix it.

Looking at the actual ICU MessageFormat source test file, `test/intltest/tmsgfmt.cpp`, I don't see any `select` message using the `#` symbol, only `plural`s and `selectordinal` (which we don't support.)  e.g. line 1863

```cpp
void TestMessageFormat::TestSelectOrdinal() {
    IcuTestErrorCode errorCode(*this, "TestSelectOrdinal");
    // Test plural & ordinal together,
    // to make sure that we get the correct cached PluralSelector for each.
    MessageFormat m(
        "{0,plural,one{1 file}other{# files}}, "
        "{0,selectordinal,one{#st file}two{#nd file}few{#rd file}other{#th file}}",
        Locale::getEnglish(), errorCode);
    if (errorCode.logDataIfFailureAndReset("Unable to instantiate MessageFormat")) {
        return;
    }
```

Let's leave this "unspecified" for now until I can dig into the gory details of the actual implementation to see if it's "really really" supposed to work in select messages.

src/ngMessageFormat/messageFormatCommon.js
chirayuk: Ok.  Point taken.  I'm willing to add it.  This should have been trivial to do, except that `$compile` does not expose `debugInfoEnabled` on `$compile` (just on the provider.)  Which is fixable on `$compile`.  I can also just add it unconditionally like `$interpolate` used to do.  The latter is super easy to do and I would actually just be deleting and simplifying code.

For AngularDart, I think we had it so that we didn't add debug/test stuff unless you ran with debug/test mode enabled (so the testability service wouldn't work without test mode enabled either.)  The equivalent here is debugInfoEnabled.  So the right thing to do would be to actually use that one.  Sounds good?

chirayuk:
Other than the `bindingInfo` bits—that are easy to fix—the other comments have been addressed.  I'll address the bindingInfo bit this after after I get back from the Japanese Consulate.  Thanks for all the comments!

PR #11154, opened 2015-02-23
chore(grunt): use rimraf to remove node_modules on Windows

Narretz:
This works very well.
The way it's called however is quite obtuse at the moment, so I'm happy with any suggestions.


petebacondarwin:
Obtuse, indeed. Is there any need for `install-dependencies.sh` to be a bash script? It seems to me that we would get better interoperability in general if we made that a node.js script; and then we could also keep the decision whether to use `rimraf` or `rm -rf` in one file.

PR #11155, opened 2015-02-24
 GET file://query.yahooapis.com/...    net::ERR_FILE_NOT_FOUND

gonghaima:
This is an error due to index.html request from the local hosting server. Faninace.js reference the URL address  start from //query.yahooapi.com/.... 
Solution: Add http:  in front of the //query.yahooapis.com/....

### DIFF
--- a/docs/content/guide/concepts.ngdoc
+++ b/docs/content/guide/concepts.ngdoc
@@ -321,7 +321,7 @@ The following example shows how this is done with Angular:
     angular.module('finance3', [])
       .factory('currencyConverter', ['$http', function($http) {
         var YAHOO_FINANCE_URL_PATTERN =
-              '//query.yahooapis.com/v1/public/yql?q=select * from '+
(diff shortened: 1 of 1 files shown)


gonghaima:
GET file://query.yahooapis.com/v1/public/yql?q=select%20*%20from%20yahoo.financ…env=store://datatables.org/alltableswithkeys&callback=angular.callbacks._0 net::ERR_FILE_NOT_FOUND

//query...  should be http://query...

pkozlowski-opensource:
@gonghaima this is done on purpose so people can use this example correctly both using http and https, see http://www.paulirish.com/2010/the-protocol-relative-url/

If anything we should change it to https. Going to close this one but if you care to send us to send another PR switching to https and testing that it doesn't cause problems we would merge it.

Please follow our [commit message guidelines](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit) and sign the CLA if you plan to open a new PR. Thnx!

PR #11156, opened 2015-02-24
step_05.ngdoc Bug Fixes

martellaj:
The code snippets demonstrating dependency injection don't show the dependencies being injected, but only as arguments. These changes are untested, but should be correct.


martellaj:
I submitted a CLA.

martellaj:
Disregard... I'm new. Finished reading the page about minification and understand why the documentation is as it is.

PR #11159, opened 2015-02-24
feat($qRaf): introduce requestAnimationFrame-based promise service

matsko:


lib/promises-aplus/promises-aplus-test-adapter.js
```diff
@@ -11,7 +11,8 @@ var minErr = function minErr (module, constructor) {
   };
 };
 
-var $q = qFactory(process.nextTick, function noopExceptionHandler() {});
+var noop = function() {};
+var $q = qFactory(noop, process.nextTick, noop);
```
IgorMinar: this doesn't look right. qFactory signature is `function qFactory(nextTick, exceptionHandler) {` no?

IgorMinar: oh. you changed it :)

src/ng/q.js
```diff
@@ -223,22 +222,62 @@ function $QProvider() {
...
+ *   promise, the second parameter is a function which rejects the promise.
+ *
+ * @returns {Promise} The newly created promise.
+ */
+function $QRafProvider() {
+  this.$get = ['$$rAF', '$exceptionHandler', function($$rAF, $exceptionHandler) {
+    return qFactory(
+      function(data) {
+        $$rAF(function() {
+          data.$$rafCleared = true;
```
IgorMinar: what's `data` and why are we monkeypatching it?

src/ng/q.js
```diff
@@ -223,22 +222,62 @@ function $QProvider() {
...
+        });
+      },
+      function(callback, data) {
+        data.$$rafCleared ? callback() : $$rAF(callback);
```
IgorMinar: so you are calling the callback synchronously in some cases? that breaks all the guarantees that promises give you.

src/ng/q.js
```diff
@@ -325,14 +364,15 @@ function qFactory(nextTick, exceptionHandler) {
     }
   }
 
-  function scheduleProcessQueue(state) {
+  function scheduleProcessQueue(promise, state) {
```
IgorMinar: I find it weird that you make the new argument first and not last.

src/ng/q.js
```diff
@@ -223,22 +222,62 @@ function $QProvider() {
...
+ * @name $qRaf
+ *
+ * @description
+ * `$qRaf` is a promise library that coordinates itself based on calls to `requestAnimationFrame`. The
+ * purpose of `$qRaf` is to allow animation-based code to be sequenced together without the need
+ * to hook into the digest cycle (which is how `$q` works).
+ *
+ * `$qRaf` is gauranteed to be delivered once one or more animation frames have passed. This means that
+ * even if the promise is resolved immediately then it will not be delivered until the next frame has passed.
+ * However if an animation does occur in between promise creation and resolution then the promise
```
IgorMinar: `animation` -> `animation frame` or something more specific.

IgorMinar:
I'm not a big fan of proliferation of promise like apis in Angular especially since they are not compatible and this incompatibility is not even properly documented. The docs don't clearly tell me when to use this and when not to use this api.

I also have strong doubts about the synchronous execution of the callbacks.

lib/promises-aplus/promises-aplus-test-adapter.js
petebacondarwin: What is the point of this change?

petebacondarwin: Oh, I see. You have added a new parameter at the beginning.

src/ng/q.js
```diff
@@ -223,22 +222,62 @@ function $QProvider() {
...
 /**
+ * @ngdoc service
+ * @name $qRaf
+ *
+ * @description
+ * `$qRaf` is a promise library that coordinates itself based on calls to `requestAnimationFrame`. The
+ * purpose of `$qRaf` is to allow animation-based code to be sequenced together without the need
+ * to hook into the digest cycle (which is how `$q` works).
+ *
+ * `$qRaf` is gauranteed to be delivered once one or more animation frames have passed. This means that
```
petebacondarwin: guaranteed

src/ng/q.js
```diff
@@ -223,22 +222,62 @@ function $QProvider() {
...
-    return qFactory(function(callback) {
+    return qFactory(noop, function(callback, data) {
       $browser.defer(callback);
     }, $exceptionHandler);
   }];
 }
 
 /**
+ * @ngdoc service
+ * @name $qRaf
```
petebacondarwin: Is this a "public" API? Do application developers need to know about this service or use it in their apps?

Would it be safer to keep it as undocumented semi-private for now?

test/ng/qSpec.js
```diff
@@ -210,7 +209,7 @@ describe('q', function() {
       });
     };
 
-    afterEach(function() {
+    beforeEach(function() {
```
petebacondarwin: Doesn't this mean that we don't end up cleaning up after the tests?
If we need these values to be `null` instead of `undefined`, we could leave this line as `afterEach` and then just initialize the variables where they are declared above.

petebacondarwin:
I have the same concerns as @IgorMinar about the modifications to the QFactory API. If the new parameters are appending the and had reasonable defaults then the other calls to QFactory would not need to be changed.

src/ng/q.js
petebacondarwin: I believe that, here, `callback()` is only being called synchronously if an animation frame has already occurred between the creation of the deferred and the resolution of it.  Is that right? But I agree that this might not be enough asynchronicity.

The other versions of Q ($q and $$q) use `$evalAsync()` and `$browser.defer()` to ensure that the `callback` is not called until the calling function's stack has been unwound. In the case of `$$q`, it is always in a new browser loop; in the case of `$q` it may be a new browser loop or after the current digest completes.

@matsko - is the concern that waiting for the next animation frame is too late in some cases to prevent jank? My understanding of `requestAnimationFrame` is that you use it to ensure your callback occurs before the next repaint. If this is the case, then we should not need to call the `callback()` synchronously, right?

Removing this synchronous call would also allow us to ditch all the other API changes too, right?

@matsko - obviously you have added this for a reason but it would be good if you could clarify the benefits.

src/ng/q.js
```diff
@@ -276,7 +315,7 @@ function qFactory(nextTick, exceptionHandler) {
 
       this.$$state.pending = this.$$state.pending || [];
       this.$$state.pending.push([result, onFulfilled, onRejected, progressBack]);
-      if (this.$$state.status > 0) scheduleProcessQueue(this.$$state);
+      if (this.$$state.status > 0) scheduleProcessQueue(result.promise, this.$$state);
```
lgalfaso: There is something very weird with this change... why would you want the generated promise and only if the current promise is fulfill (and in this case, only some of the promises will be *marked* and some will not...

matsko: Can you break down what you're saying please. I don't understand.

src/ng/q.js
```diff
@@ -325,14 +364,15 @@ function qFactory(nextTick, exceptionHandler) {
     }
   }
 
-  function scheduleProcessQueue(state) {
+  function scheduleProcessQueue(promise, state) {
     if (state.processScheduled || !state.pending) return;
```
lgalfaso: Does this implementation understands that we schedule the resolution of promises only when we do not have it already scheduled?

matsko: Please explain what you mean in a different way. I don't understand this either.

src/ng/q.js
lgalfaso: `$$state.pending` is an array that has all the promises that should be fulfill once this promise is fulfill. The check `if (this.$$state.status > 0)` checks if this promise is fulfill. At `scheduleProcessQueue` we check if the given `$$state` was scheduled before, and if that is the case, then do nothing.

This is what the comment https://github.com/angular/angular.js/pull/11159/files#r26261523 is talking about.

Eg

```js
var parent = $q.when('XXX'); // parent is fulfill
var child1 = parent.then(...); // given that parent is fulfill, parent.$$state is scheduled (but not called)
var child2 = parent.then(...); // parent.$$state is already scheduled, then it is not scheduled again
```

So, this implies that the current cases will have some discrepancies
```js
// Case 1, there is a previously schedule promise resolution
var parent = $qRaf.when('XXX'); // parent is fulfill
var child1 = parent.then(...); // parent.$$state is scheduled (but not called)
$$rAF(function() {
  var child2 = parent.then(...); // parent.$$state is already scheduled, so this is going to be resolve with `$$rAF(callback)`
});
```


```js
// Case 2, there is no previously schedule promise resolution
var parent = $qRaf.when('XXX'); // parent is fulfill
$$rAF(function() {
  var child2 = parent.then(...); // parent.$$state is not scheduled, so this is going to be resolve synchronously
});
```


In the case I ma missing something, please let me know

matsko: I understand what you mean, and I appreciate the effort you put into your comment reply, but the line you're pointing to doesn't change any of that. The only reason the promise is passed into the function is so that a particular property can be read later on by it.

I still might be missing something. I think the only way this is going to be clear is to have a test that proves that `case 2` fails.

lgalfaso: Here are two scenarios that I would expect to have the same result

```js
// Case 1
var signature = '';
var parent = $qRaf.defer();
var child1 = parent.promise.then(function() {
  signature += 'A';
});
$$rAF(function() {
  var child2 = parent.promise.then(function() {
    signature += 'B';
  });
  parent.resolve('X');
});
$$rAF.flush();
expect(signature).toBe(???); // In this case, signature is 'A'
```

```js
// Case 2
var signature = '';
var parent = $qRaf.defer();
var child1 = parent.promise.then(function() {
  signature += 'A';
});
$$rAF(function() {
  parent.resolve('X');
  var child2 = parent.promise.then(function() {
    signature += 'B';
  });
});
$$rAF.flush();
expect(signature).toBe(???); // In this case, signature is 'AB'
```



To make it fail, I had to chance the implementation of mock.$rAF.flush to only move one frame at a time
```js
  rafFn.flush = function() {
    if (queue.length === 0) {
      throw new Error('No rAF callbacks present');
    }

    var length = queue.length;
    var toFlush = queue;
    queue = [];
    for (var i = 0; i < length; i++) {
      toFlush[i]();
    }
  };
```

petebacondarwin:
@matsko - you were going to knock up an example of animation jank when we resolve the promises asynchronously, which would provide the motivation for this synchronous code change.

matsko:
@petebacondarwin 

Looks like ONE raf is OK (no flicker). Haven't tested on IE yet.

** No RAF (instant delete) ** 
https://s3.amazonaws.com/angularjs-dev/raf-testing/index.html

** One RAF **
https://s3.amazonaws.com/angularjs-dev/raf-testing/index-raf.html

** Two Nested RAFs **
https://s3.amazonaws.com/angularjs-dev/raf-testing/index-raf-two.html

matsko:
@petebacondarwin  I'll do some more testing and then I'll make the changes and keep it as one tick.

petebacondarwin:
I can see from your examples that having animation frames between the `fade` class being removed and the element itself being removed causes flicker. On Safari I even see this happening in the second example: https://s3.amazonaws.com/angularjs-dev/raf-testing/index-raf.html

But I am still uncomfortable with putting this decision making in to the promise provider.

Moreover, my understanding of Animation Frames is that we have even less control over when they occur that with Timeouts.  In other words the browser is free to delay triggering an animation frame for any length of time if it is busy working on other stuff. This is great for saving on CPU usage, but doesn't really help us in terms of ticks in promise resolution...

What you are trying to achieve here, correct me if I'm wrong, is ensure that child promises are resolved before the next animation frame, if the parent promise was resolved in a previous animation frame.

I am not sure if the aim of the forthcoming rewrite is going to do this, but it would make more sense to me to move this decision making into an animation "scheduling engine" that  will ensure that visual updates, such as adding/removing classes/elements happen in the "same animation frame", rather than modifying promises to sort of provide this for us.

petebacondarwin:
Closing this in favour of the new `animateRunner` idea: #11369

PR #11168, opened 2015-02-24
chore(ci): update Karma to 0.12.32-beta.0

juliemr:
This will hopefully make the CI more stable because of its updated
version of socket.io.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "jasmine-node": "~1.14.5",
     "jasmine-reporters": "~1.0.1",
     "jshint-stylish": "~1.0.0",
-    "karma": "~0.12.0",
+    "karma": "0.12.32",
     "karma-browserstack-launcher": "0.1.2",
     "karma-chrome-launcher": "0.1.5",
     "karma-firefox-launcher": "0.1.3",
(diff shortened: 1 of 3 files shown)


juliemr:
NOTE - this is a work in progress pointing to github instead of npm. I will update to point to 0.12.32-beta.0 once that is released, if this passes.

package.json
```diff
@@ -44,7 +44,7 @@
     "jasmine-node": "~1.14.5",
     "jasmine-reporters": "~1.0.1",
     "jshint-stylish": "~1.0.0",
-    "karma": "~0.12.0",
+    "karma": "^0.12.32-beta.0",
```
IgorMinar: do we want to use `^`? that will update us automatically to 0.13/0.14 when that is available and somebody does `npm update`. Sounds too risky to me.

IgorMinar:
otherwise lgtm

package.json
juliemr: Done

juliemr:
I had to rename the karma release to 0.12.32 because otherwise peer dependencies would fail and npm wouldn't install.

juliemr:
Merged to master as 581ee9d0b6aacd76d9ede56ab6c735b2691f9193

juliemr:
Merged to v1.3.x as 01a725a7697181dbcc92853c18f486430c26e625

PR #11171, opened 2015-02-24
fix(select): Select option with ngOptions when label equals $viewValue

### LINKED ISSUE #11170: Select don't show the selected option with ngOptions label equals the $viewValue
Go to this JSFiddle:
http://jsfiddle.net/felixcoto/v450hf7o/
1. Push in the button "Load Options"
2. The first select doesn't show the correct option

This issue is tested in Angular 1.3.13

felixcoto:
Fix selectController when using ngOptions and the label equals the $viewValue of the ngModel.
Otherwise, when the options are loaded, selectController set the value of the select to a 
non existing value

Fixes #11170

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -205,7 +205,8 @@ var selectDirective = ['$compile', '$parse', function($compile,   $parse) {
           optionsMap = {},
           ngModelCtrl = nullModelCtrl,
           nullOption,
-          unknownOption;
+          unknownOption,
+          optionsExp = $attrs.ngOptions;
 
 
       self.databound = $attrs.ngModel;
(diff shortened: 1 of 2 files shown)


petebacondarwin:
This is not really fixing the problem, which is that the `optionsMap` is storing the label rather than the value of the option and we are trying to compare the `$viewValue` with the label rather than the "value" of the option, which is sometimes also different to the value of the `<option>` element. Trouble is that the element value (in Angular 1.3.x) is well related to the value of the option (i.e. it is basically the index value of the element).

This is fixed in Angular 1.4, which is about to be released. Is it possible to update to that version? It would probably be quicker than trying to fix this version.

PR #11172, opened 2015-02-24
docs(ngModel.NgModelController): Remove "newer browsers only" clarification on contenteditable

jkjustjoshing:
contenteditable is supported in many more browsers than Angular itself is.

http://caniuse.com/#feat=contenteditable


jkjustjoshing:
Signed!

PR #11173, opened 2015-02-25
perf(rootScope): Fix a memory leak

### LINKED ISSUE #11169: AngularJS: Memory Leak with ng-repeat using custom objects (w/simple PLUNKR)
(SIMPLE PLUNKR HERE:  http://plnkr.co/edit/1Adh2P65ZS8YHYJZhLMk)
(DETAILED STACKOVERFLOW EXPLANATION HERE: http://stackoverflow.com/questions/28683992/angularjs-memory-leak-with-ng-repeat-using-custom-objects-w-simple-plunkr)

SUMMARY:

There is a leak using ng-repeat after the 2nd wave iterating over an array of custom objects like this :

    <div ng-repeat="d_sampleObject in mySampleObjects">
        {{d_sampleObject.description}}
    </div>

NOTE: 'mySampleObjects' is an array of the following object instances:

        ml.MySampleObject = function (id) {

            this.id = id;
            this.description = 'this is object #:' + ' '+id;
        }

Memory profile reveals an extra 'd_sampleObject' left over and not de-referenced. More details (via a controller and an injected service) in the Stackoverflow link above. A simple demonstration also in the provided plunkr link. Any thoughts and help greatly appreciated in advance!

realityking:
Closes #11169

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -83,6 +83,10 @@ function $RootScopeProvider() {
   this.$get = ['$injector', '$exceptionHandler', '$parse', '$browser',
       function($injector, $exceptionHandler, $parse, $browser) {
 
+    function destroyChildScope($event) {
+        $event.currentScope.$$destroyed = true;
+    }
+
     /**
      * @ngdoc type
      * @name $rootScope.Scope
@@ -233,13 +237,9 @@ function $RootScopeProvider() {
         // prototypically. In all other cases, this property needs to be set
         // when the parent scope is destroyed.
         // The listener needs to be added after the parent is set
-        if (isolate || parent != this) child.$on('$destroy', destroyChild);
+        if (isolate || parent != this) child.$on('$destroy', destroyChildScope);
 
         return child;
-
-        function destroyChild() {
-          child.$$destroyed = true;
-        }
       },
 
       /**
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -80,6 +80,20 @@ function $RootScopeProvider() {
     return TTL;
   };
 
+  function createChildScopeClass(parent) {
+    function ChildScope() {
+      this.$$watchers = this.$$nextSibling =
+          this.$$childHead = this.$$childTail = null;
+      this.$$listeners = {};
+      this.$$listenerCount = {};
+      this.$$watchersCount = 0;
+      this.$id = nextUid();
+      this.$$ChildScope = null;
+    }
+    ChildScope.prototype = parent;
+    return ChildScope;
+  }
+
   this.$get = ['$injector', '$exceptionHandler', '$parse', '$browser',
       function($injector, $exceptionHandler, $parse, $browser) {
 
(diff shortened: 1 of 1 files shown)


Narretz:
Hi, thanks for the PR. Can you explain why that fixes the memory leak?

realityking:
The issue occurs in `Scope.prototype.$new`. The code currently looks like this (I've removed some parts to aid readability):

```javascript
function(isolate, parent) {
    var child;

    if (isolate) {
      child = new Scope();
      child.$root = this.$root;
    } else {
      // Only create a child scope class if somebody asks for one,
      // but cache it to allow the VM to optimize lookups.
      if (!this.$$ChildScope) {
        this.$$ChildScope = function ChildScope() {
          this.$$watchers = this.$$nextSibling =
              this.$$childHead = this.$$childTail = null;
          this.$$listeners = {};
          this.$$listenerCount = {};
          this.$$watchersCount = 0;
          this.$id = nextUid();
          this.$$ChildScope = null;
        };
        this.$$ChildScope.prototype = this;
      }
      child = new this.$$ChildScope();
    }
    child.$parent = parent;
    child.$$prevSibling = parent.$$childTail;
    if (parent.$$childHead) {
      parent.$$childTail.$$nextSibling = child;
      parent.$$childTail = child;
    } else {
      parent.$$childHead = parent.$$childTail = child;
    }

    if (isolate || parent != this) child.$on('$destroy', destroyChild);

    return child;

    function destroyChild() {
      child.$$destroyed = true;
    }
 }
```

There are two closures used in this function `ChildScope` and `destroyChild`. In V8 closures created in the same lexical scope share one context, which contains all variables used in *any* of the closures.

Now if we replace all child scopes created in an ngRepeat, `ChildScope` is cached. Since it shares a context with `destroyChild` so is the variable `child`, which is the first child scope created for this scope. The memory would be freed when the parent scope is destroyed.

By not using `child` in  `destroyChild` we keep it out of the context and thus the memory can be freed earlier. I moved the function outside of `$new` so it's not recreated for every child scope.

jbedard:
Thanks for the nice explanation @realityking! I didn't know V8 did that. That could lead to some very interesting memory issues...

Narretz:
LGTM. For good measure, @petebacondarwin do you think we need a test for this?

petebacondarwin:
Would an alternative be to move the creation of the `ChildScope` "class" outside of this function. Something like:

```js

function createChildScopeClass(parent) {
  function ChildScope() {
    this.$$watchers = this.$$nextSibling =
        this.$$childHead = this.$$childTail = null;
    this.$$listeners = {};
    this.$$listenerCount = {};
    this.$$watchersCount = 0;
    this.$id = nextUid();
    this.$$ChildScope = null;
  };
  ChildScope.prototype = parent;
  return ChildScope; 
}

function(isolate, parent) {
    var child;

    if (isolate) {
      child = new Scope();
      child.$root = this.$root;
    } else {
      // Only create a child scope class if somebody asks for one,
      // but cache it to allow the VM to optimize lookups.
      if (!this.$$ChildScope) {
        this.$$ChildScope = createChildScopeClass(this);
      }
      child = new this.$$ChildScope();
    }
    child.$parent = parent;
    child.$$prevSibling = parent.$$childTail;
    if (parent.$$childHead) {
      parent.$$childTail.$$nextSibling = child;
      parent.$$childTail = child;
    } else {
      parent.$$childHead = parent.$$childTail = child;
    }

    if (isolate || parent != this) child.$on('$destroy', function destroyChild() { child.$$destroyed = true; });

    return child;
 }
```

shahata:
Nice catch! I've mentioned this leak in https://github.com/angular/angular.js/pull/10895#issuecomment-72467810 but I didn't know why it leaked :)

realityking:
@Narretz I don't know how you'd want to test this. GC behavior is not observable from JS code.

@petebacondarwin That would work too - and IMO is easier to read. Mine has the (probably very small) advantage, that it doesn't create a new `destroyChild` function for every child scope.

Maybe do both?

petebacondarwin:
@realityking  - Both works for me :-) Let's do that. I also agree that we are unlikely to be able to create a unit test for this...

realityking:
Alright, I'll update the patch tonight.

realityking:
I've added a commit with @petebacondarwin's changes

PR #11180, opened 2015-02-25
docs(guide/Directives): mentioned shorthand for "return function()"

zbycz:
The shorthand is little confusing, added clarification.

### DIFF
--- a/docs/content/guide/directive.ngdoc
+++ b/docs/content/guide/directive.ngdoc
@@ -867,6 +867,7 @@ element?
   </file>
 </example>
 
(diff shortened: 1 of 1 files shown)


Narretz:
Hi, thanks for your PR! However, before we can merge we need you to sign the CLA.

Narretz:
Imo, we should change the example to use the verbose way of defining a directive. That makes it clearer.

lgalfaso:
Agree with @Narretz

zbycz:
Quite contrary, i'd like to learn that the shorthand exists - it can be seen in other projects as well. Just the clarification was missing. 

I signed the cla, should i do something more? (On https://cla.developers.google.com/clas i see my individual CLA with correct email and github name..)

PR #11182, opened 2015-02-25
Fix ngjq

Narretz:
@mboudreau This is a wip for the e2e tests. You can play around with it. (use `grunt test:protractor`) You can als comment out the other test paths in `protractor.conf.js`, so only the e2e folders get run:

```
// 'build/docs/ptore2e/**/*.js',
// 'docs/app/e2e/**/*.scenario.js'
```

The problem is currently that we don't have a reliable method of detecting what is loaded, because `jq` is private. We might have to expose it. Currently we can only test that it doesn't break.

### DIFF
--- a/test/e2e/fixtures/ngJq/script.js
+++ b/test/e2e/fixtures/ngJq/script.js
@@ -1,0 +1,6 @@
+'use strict';
+
+angular.module('test', [])
+  .run(function($rootScope) {
+    $rootScope.jqueryVersion = window.angular.element().jquery || 'jqLite';
+  });
--- a/test/e2e/fixtures/ngJqJquery/index.html
+++ b/test/e2e/fixtures/ngJqJquery/index.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html ng-app="test" ng-jq="jQuery_2_1_0">
+  <body>
+      {{jqueryVersion}}
+
+      <script src="../../../../bower_components/jquery/dist/jquery.js"></script>
+      <script src="http://ajax.googleapis.com/ajax/libs/jquery/2.1.0/jquery.min.js"></script>
+      <script>
+        var jQuery_2_1_0 = jQuery.noConflict();
+      </script>
+      <script src="angular.js"></script>
+      <script type="text/javascript" src="script.js"></script>
+    </body>
+</html>
--- a/test/e2e/tests/ngJqSpec.js
+++ b/test/e2e/tests/ngJqSpec.js
@@ -1,0 +1,12 @@
+describe('Customizing the jqlite / jquery version', function() {
+
+  it('should be able to force jqlite', function() {
+    loadFixture("ngJq").andWaitForAngular();
+    expect(element(by.binding('jqueryVersion')).getText()).toBe('jqLite');
+  });
+
+  it('should be able to use a specific version jQuery', function() {
+    loadFixture("ngJqJquery").andWaitForAngular();
+    expect(element(by.binding('jqueryVersion')).getText()).toBe('2.1.0');
+  });
+});
--- a/test/e2e/tools/util.js
+++ b/test/e2e/tools/util.js
@@ -24,6 +24,10 @@ function testExists(testname) {
 }
 
 function rewriteTestFile(testname, testfile) {
+  if (testfile.indexOf('http') === 0) {
+    return testfile;
+  }
+
   var i = 0;
   while (testfile[i] === '/') ++i;
   testfile = testfile.slice(i);
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -943,16 +943,13 @@ var csp = function() {
 var jq = function() {
   if (isDefined(jq.name_)) return jq.name_;
   var el;
-  var i, ii = ngAttrPrefixes.length;
+  var i, ii = ngAttrPrefixes.length, prefix, name;
   for (i = 0; i < ii; ++i) {
-    if (el = document.querySelector('[' + ngAttrPrefixes[i].replace(':', '\\:') + 'jq]')) {
+    prefix = ngAttrPrefixes[i];
+    if (el = document.querySelector('[' + prefix.replace(':', '\\:') + 'jq]')) {
+      name = el.getAttribute(prefix + 'jq');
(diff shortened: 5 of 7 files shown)


petebacondarwin:
Looks good so far. We ought to be able to work out what jquery-like library is being used from inside the fixture app. For instance, in the fixture app, call `angular.element()` and then look for something that only appears in JQuery or jqLite.

gkalpak:
We can just check that `angular.element !== window.jQuery` (either from inside the fixture, or using `executeScript`).
It would also be nice to test that one can load a different version of jQuery than the rest of the app (testing against `window.jQuery.fn.jquery`).

Narretz:
thx @gkalpak that works very well. 

I'm not quite sure how we should test for another version of jquery without including it in the repo.

petebacondarwin:
Since it is just an e2e test, perhaps we could hit a CDN with well defined versions of jQuery for the test?

Narretz:
I updated the PR with expects & a test for ensuring forcing a specific jQuery version works via google cdn. The http request should be negligible as the e2e take a long time anyway.

test/e2e/tests/ngJqSpec.js
```diff
@@ -0,0 +1,12 @@
+describe('Customizing the jqlite / jquery version', function() {
+
+  it('should be able to force jqlite', function() {
+    loadFixture("ngJq").andWaitForAngular();
+    expect(browser.executeScript('return window.angular.element !== window.jQuery')).toBe(true);
+  });
```
gkalpak: It might be a good idea to also test that `window.jQuery` is defined.

E.g. imagine 'jquery.js' isn't properly loaded in the future for whatever reason (e.g. directory is moved). I would be good to catch this in the test (and fix).

test/e2e/tests/ngJqSpec.js
```diff
@@ -0,0 +1,12 @@
...
+
+  it('should be able to use a specific version jQuery', function() {
+    loadFixture("ngJqJquery").andWaitForAngular();
+    expect(browser.executeScript('return window.angular.element !== window.jQuery_2_0_0')).toBe(true);
```
petebacondarwin: Should this be `===`?

Narretz: Yeah, definitely. That probably means that something is not working yet, since the test passes.

petebacondarwin: The problem we have here is that e2e tests do not run in the context of the browser. In fact they do not run on the browser at all. Protractor converts JS in the tests into commands that are sent via Selenium to the browsers. We have to consider the browser a black box.

(OK, so it is possible to send a JavaScript block to be run on the browser but this is not the best approach).

Instead what we want is the fixture applications to expose the information about JQuery/jqLite in HTML that we can then inspect using Protractor

petebacondarwin: OK, so I see that you are indeed sending a block of script to be executed but I think it is still better not to try to run scripts from e2e tests

Narretz: I'm switching to using a directive that exposes a scope property. There still seems to be something wrong, though.

Narretz:
I refactored the tests to use a directive that exposes a binding on the scope.

The test with forcing a specific version of jQuery fails when the file is on the google cdn. It works when I have the file locally. It also works with the cdn when I try it in a non-protractor environment, so I think it might be something with protractor, that it is not waiting for the file to download.

petebacondarwin:
I think the e2e middleware is rewriting the path to the CDN for the jquery at this point: https://github.com/angular/angular.js/blob/master/test/e2e/tools/util.js#L34

petebacondarwin:
In my branch here: https://github.com/angular/angular.js/compare/angular:master...petebacondarwin:ngJq 

then run `grunt webserver`

and go to  http://localhost:8000/e2e/fixtures/ngJqJquery

You can see in the source that the CDN src is changed to `false`

Narretz:
Thanks @petebacondarwin I was so focused on protractor being the problem that I didn't consider the rewrite.

Narretz:
I updated the PR. script src starting with `http` are now exempt from rewrite. I also included the simplified code from @petebacondarwin

petebacondarwin:
And now it works, yes?

test/e2e/fixtures/ngJq/script.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
```
petebacondarwin: This line breaks jshint rules. It should either be made into the functional form, or perhaps easier just remove it.

test/e2e/fixtures/ngJqJquery/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html ng-app="test" ng-jq="jQuery_2_1_0">
+  <body>
+      {{jqueryVersion}}
+
+      <script src="../../../../bower_components/jquery/dist/jquery.js"></script>
+      <script src="http://ajax.googleapis.com/ajax/libs/jquery/2.1.0/jquery.min.js"></script>
```
petebacondarwin: We should switch the order of these two jquery script lines because the last loaded jquery wins. If we dropped the `ng-jq` attribute in this app, it would still pass the e2e test.

test/e2e/tools/util.js
```diff
@@ -24,6 +24,10 @@ function testExists(testname) {
 }
 
 function rewriteTestFile(testname, testfile) {
+  if (testfile.indexOf('http') === 0) {
+    return testfile;
+  }
+
```
petebacondarwin: Looks good

petebacondarwin: Although this would match... `script="httpInterceptor.js"` Perhaps we should make it a little more complex: `/https?:\/\/`

petebacondarwin:
Merging...

test/e2e/tools/util.js
Narretz: Yes, that is much better. Honestly I just wanted to get the tests to work. ;)

Narretz:
Thanks @petebacondarwin !

PR #11189, opened 2015-02-26
docs(ngMessages): ng-messages-include defined on ng-messages element

fredrikekelund:
I was having issues implementing `ngMessages` together with `ngMessagesInclude` where the template I was trying to include simply wouldn't show up. This was until I realized that, contrary to how the docs describe it, Angular expected ng-messages-include to be defined on the same element as ng-messages (which I later discovered was how eg. [this tutorial](http://www.yearofmoo.com/2014/05/how-to-use-ngmessages-in-angularjs.html) described the implementation of ngMessages).

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -80,9 +80,7 @@ var jqLite = angular.element;
  *   <div ng-message="minlength">This field is too short</div>
  * </script>
  *
- * <div ng-messages="myForm.myField.$error">
- *   <div ng-messages-include="error-messages"></div>
- * </div>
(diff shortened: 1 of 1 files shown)


fredrikekelund:
Signed

gkalpak:
You are obviously seeing the docs for the latest build of Angular (master), while using an older (stable) version. `ngMessages` has been recently revamped (landed in v1.4.0-beta.5), so the latest docs are reflecting the latest changes.

For more info, take a look at this [breaking change notice][1] in the changelog.

[1]: https://github.com/angular/angular.js/blob/master/CHANGELOG.md#breaking-changes

fredrikekelund:
Ah, thanks. My bad!

PR #11194, opened 2015-02-26
docs(tutorial/0 - Bootstrapping): Clearing browser cache guarantees that your app will retrieve new files.

BorlandP:
Clearing browser cache guarantees that your app will retrieve new files.


Narretz:
Normally, the grunt server should not let the browser cache the files beyond a normal page reload. Did you encounter this problem yourself?

Mishurin:
Hi, I tried to reproduce this step (cloned the repo again and checked out to the first step). This issue wasn't reproduced this time. Maybe the thing was in running node server in separate window, which I had to stop. Let's reject this PR.

Narretz:
Okay I'll close it, but thanks for the contribution anyway!

PR #11199, opened 2015-02-26
fix(ngMessages): ensure that multi-level transclusion works with ngMessagesInclude

### LINKED ISSUE #11196: bug(ngMessages): ngMessagesInclude crashes inside ngRepeat
Bug introduced in 1.4.0-beta.5: using ngMessagesInclude inside an ngRepeat (with ngForm - might be relevant) crashes with error:

```
Error: [$compile:ctreq] Controller 'ngMessages', required by directive 'ngMessage', can't be found!
```

See demo: http://plnkr.co/edit/9tEwZzf1q6qeUrL0rBxn?p=preview
Code is working fine in 1.3.14: http://plnkr.co/edit/yAuxsFGYymCj9fAAqTVw?p=preview

 Crash does not occur if `msgs.html` is empty, it must have (any) content., i.e.:

```html
<span ng-message="required">Field is required</span>
<span ng-message="number">Fill in a number</span>
```

I'll paste the demo code here, too:

```html
<html>

<head>
  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.0-beta.5/angular.js"></script>
  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.0-beta.5/angular-messages.js"></script>
  <script src="script.js"></script>
</head>

<body>
  <div ng-app="myApp">
    <form name="myForm">
      <ul>


        <li ng-repeat="i in [1, 2, 3, 4]" ng-form="theForm">{{ i }}:
          <input type="number" ng-model="number" name="number">
          <div ng-messages="theForm.number.$error">
            <ng-messages-include src="msgs.html"></ng-messages-include>
          </div>
        </li>


      </ul>
    </form>
  </div>
</body>
</html>
```

```javascript
angular.module('myApp', ['ngMessages'])
  .controller('MyCtrl', ['$scope', function($scope) {
  }]);
```

@matsko this is related to https://github.com/angular/angular.js/pull/10676 and https://github.com/matsko/angular.js/commit/d97733f1dc70c35d75d4e0c3ce9085db3e7defdd

matsko:
ngRepeat and any other directives that alter the DOM structure using
transclusion may cause ngMessagesInclude to behave in an unpredictable
manner. This fix ensures that the element containing the ngMessagesInclude
directive will stay in the DOM to avoid these issues.

Closes #11196

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -500,27 +500,21 @@ angular.module('ngMessages', [])
 
      return {
        restrict: 'AE',
-       require: '^^ngMessages',
-       compile: function(element, attrs) {
-         var comment = jqLite($document[0].createComment(' ngMessagesInclude: '));
-         element.after(comment);
+       require: '^^ngMessages', // we only require this for validation sake
+       link: function($scope, element, attrs) {
+         var src = attrs.ngMessagesInclude || attrs.src;
+         $templateRequest(src).then(function(html) {
+           $compile(html)($scope, function(contents) {
+             element.after(contents);
 
-         return function($scope, $element, attrs, ngMessagesCtrl) {
-           // we're removing this since we only need access to the newly
-           // created comment node as an anchor.
-           element.remove();
+             // the anchor is placed for debugging purposes
+             var anchor = jqLite($document[0].createComment(' ngMessagesInclude: ' + src + ' '));
+             element.after(anchor);
 
-           $templateRequest(attrs.ngMessagesInclude || attrs.src).then(function(html) {
-             var elements = jqLite('<div></div>').html(html).contents();
-             var cursor = comment;
-             forEach(elements, function(elm) {
-               elm = jqLite(elm);
-               cursor.after(elm);
-               cursor = elm;
-             });
-             $compile(elements)($scope);
+             // we don't want to pollute the DOM anymore by keeping an empty directive element
+             element.remove();
            });
-         };
+         });
        }
      };
    }])
--- a/test/ngMessages/messagesSpec.js
+++ b/test/ngMessages/messagesSpec.js
@@ -4,6 +4,10 @@ describe('ngMessages', function() {
   beforeEach(inject.strictDi());
   beforeEach(module('ngMessages'));
 
+  function messageChildren(element) {
+    return (element.length ? element[0] : element).querySelectorAll('[ng-message], [ng-message-exp]');
+  }
+
   function they(msg, vals, spec, focus) {
     forEach(vals, function(val, key) {
       var m = msg.replace('$prop', key);
@@ -236,7 +240,7 @@ describe('ngMessages', function() {
         $rootScope.col = {};
       });
 
-      expect(element.children().length).toBe(0);
+      expect(messageChildren(element).length).toBe(0);
       expect(trim(element.text())).toEqual('');
 
       $rootScope.$apply(function() {
@@ -246,7 +250,7 @@ describe('ngMessages', function() {
         };
       });
 
-      expect(element.children().length).toBe(1);
+      expect(messageChildren(element).length).toBe(1);
       expect(trim(element.text())).toEqual('This message is blue');
 
       $rootScope.$apply(function() {
@@ -255,13 +259,13 @@ describe('ngMessages', function() {
         };
       });
 
-      expect(element.children().length).toBe(1);
+      expect(messageChildren(element).length).toBe(1);
       expect(trim(element.text())).toEqual('This message is red');
 
       $rootScope.$apply(function() {
         $rootScope.col = null;
       });
-      expect(element.children().length).toBe(0);
+      expect(messageChildren(element).length).toBe(0);
       expect(trim(element.text())).toEqual('');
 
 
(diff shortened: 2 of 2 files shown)


src/ngMessages/messages.js
```diff
@@ -496,31 +496,16 @@ angular.module('ngMessages', [])
...
-               cursor = elm;
-             });
-             $compile(elements)($scope);
-           });
-         };
+       require: '^^ngMessages', // we only require this for validation sake
+       link: function($scope, element, attrs) {
+         $templateRequest(attrs.ngMessagesInclude || attrs.src).then(function(html) {
+           element.html(html);
+           $compile(element.contents())($scope);
```
caitp: I'm not sure this is the best approach. Why not just make use of the clone attach feature of the link function?

```
$templateRequest(attrs.ngMessagesInclude || attrs.src).then(function(html) {
  $compile(html)($scope, function(dom) {
    // attach to DOM using your favourite method --- finding the parent ngMessages directive will no longer fail.
  });
```

src/ngMessages/messages.js
```diff
@@ -496,31 +496,16 @@ angular.module('ngMessages', [])
...
-
-         return function($scope, $element, attrs, ngMessagesCtrl) {
-           // we're removing this since we only need access to the newly
-           // created comment node as an anchor.
-           element.remove();
-
-           $templateRequest(attrs.ngMessagesInclude || attrs.src).then(function(html) {
-             var elements = jqLite('<div></div>').html(html).contents();
-             var cursor = comment;
-             forEach(elements, function(elm) {
```
caitp: +1 for removing all this though, it's more complicated than is really needed

src/ngMessages/messages.js
```diff
@@ -496,31 +496,16 @@ angular.module('ngMessages', [])
...
    .directive('ngMessagesInclude',
-     ['$templateRequest', '$document', '$compile', function($templateRequest, $document, $compile) {
+     ['$templateRequest', '$compile', function($templateRequest, $compile) {
 
      return {
        restrict: 'AE',
-       require: '^^ngMessages',
-       compile: function(element, attrs) {
-         var comment = jqLite($document[0].createComment(' ngMessagesInclude: '));
-         element.after(comment);
```
caitp: making this directive an element transclusion directive makes it unnecessary to add this manually

src/ngMessages/messages.js
matsko: What's the benefit to having this code be inserted in a different way? There's no animation stuff here since that's managed in the `<ng-message>` directive level which transcludes on its own anyway.

src/ngMessages/messages.js
matsko: If we set terminal to true then will this run last if the priority is left unset?

src/ngMessages/messages.js
caitp: the benefit is not adding a breaking change --- keeping the DOM looking the way it did before

src/ngMessages/messages.js
caitp: directives are evaluated in the deterministic order, but they'll stop when they reach the terminal one. similarly, child nodes of an element with a terminal directive won't be compiled (but there are no children, except the ones you load from the server and compile on your own.)

matsko: Yes I know how this works, but please answer my question if it will be handled last if a priority is not assigned in the directive definition.

caitp: yes, it will. the default priority is 0

src/ngMessages/messages.js
matsko: That works amazing! Thanks! You just saved a very inconvenient breaking change.

src/ngMessages/messages.js
matsko: The PR has been updated with a new commit. I don't think we should set this to terminal since it is already working.

caitp:
LGTM, but I would suggest adding some comments explaining the flow and assertions in that large unit test.

src/ngMessages/messages.js
```diff
@@ -500,27 +500,13 @@ angular.module('ngMessages', [])
...
-           $templateRequest(attrs.ngMessagesInclude || attrs.src).then(function(html) {
-             var elements = jqLite('<div></div>').html(html).contents();
-             var cursor = comment;
-             forEach(elements, function(elm) {
-               elm = jqLite(elm);
-               cursor.after(elm);
-               cursor = elm;
-             });
-             $compile(elements)($scope);
+       require: '^^ngMessages', // we only require this for validation sake
```
caitp: actually, hang on. Previously, you were basically replacing the element with a comment placeholder. I think we should keep doing this, there's no good reason to keep it in the DOM (is there?)

matsko: All fixed up now. And I added some comments to the unit test.

caitp:
thanks, lgtm.

matsko:
Merged as https://github.com/angular/angular.js/commit/d7ec5f392e1550658ddf271a30627b1749eccb69

matsko:
Thanks @caitp

Issue #11198, opened 2015-02-26
require in directive DDO only causes an error if the DDO also includes a link function


bborowin:
Repro: make a simple app w/this custom directive:

    foo.directive('bar', function() {
      return {
        scope: false,
        require: 'your mom',
        template: '<span>foobar</span>',
        //link: function() {}
      };
    });

Expected: compile [pukes on require](https://docs.angularjs.org/error/$compile/ctreq) not matched
Actual: everything compiles fine

Related SO question is [here](http://stackoverflow.com/questions/28730346/require-ddo-option-of-angular-directive-does-not-throw-an-error-when-it-should)

As far as I can tell, it's because the [getController](https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L1841) function is only called from pre and post, which doesn't happen when a link function is not specified in the DDO.

Narretz:
Yeah, you are explicitly requiring the controller into your link function, not the presence of the parent to your directive. I wonder if it would would break many apps, did we tighten require to throw even without a link function?

Sidenote: I can see how this is confusing. I remember, when I first started with angular, I thought require was also a way to make sure a certain directive can only be nested inside another directive, even if the required directive didn't actually define a controller.

bborowin:
Thanks! Any chance of updating the docs / tutorials / official recipes to explicitly state that require only works when link function is specified?

bborowin:
Thanks! Added a pull request to update the docs: https://github.com/angular/angular.js/pull/11206

PR #11206, opened 2015-02-27
Updated the require section to reflect issue 11198

bborowin:
To make things less confusing, explicitly state that require WILL NOT cause a compile error if a link function is not specified.

https://github.com/angular/angular.js/issues/11198
stackoverflow.com/questions/28730346/require-ddo-option-of-angular-directive-does-not-throw-an-error-when-it-should


bborowin:
CLA signed, thanks botty mc bot

pkozlowski-opensource:
Woops, it looks like it is not ready to be merged, as it makes TravisCI upset. @bborowin could you please fix the jscs style errors and update the commit message to follow our guidelines? Thnx!

Issue #6432, opened 2014-02-24
ngTouch: iOs form elements loose focus when tapped to enter text (textareas,inputs)


doner-astogdill:
While developing using ngTouch in an angular project I found that when tapping on a form field to enter text the keyboard would open and immediately close.  I looked into ngTouch and found that line 308:

```Javascript
    // Blur focused form elements
    event.target && event.target.blur();
```
Was causing this.  Once removed everything worked fine with no issues in my project.  May want to look into a work around that would still allow function with iPad text entry.

PR #2989, opened 2013-06-17
prevents unwanted opening of the soft keyboard

JanLaussmann:
Ghost clicks are busted but the corresponding form elements are still focused. This means that for example on smartphones the soft keyboard will be opened. This pull request prevents the unwanted opening of the soft keyboard.

PR #11212, opened 2015-02-28
fix(ngTouch): don't blur form elements

### LINKED ISSUE #4030: ngTouch: ng-click directive is broking default behaviour of child elements

Here is an example:
http://plnkr.co/edit/BFhC6n

mikec:
Form element focus should not be prevented by ng-click.
The issue was introduced by #2989, and happens on mobile browsers that support touch events.
Plunker to demonstrate the issue: http://embed.plnkr.co/CxeuWocoPE70vzrevrIi

Closes #4030, #6432

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11212)

### DIFF
--- a/src/ngTouch/directive/ngClick.js
+++ b/src/ngTouch/directive/ngClick.js
@@ -158,9 +158,6 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
     // If we didn't find an allowable region, bust the click.
     event.stopPropagation();
     event.preventDefault();
-
-    // Blur focused form elements
-    event.target && event.target.blur && event.target.blur();
   }
 
 
--- a/test/ngTouch/directive/ngClickSpec.js
+++ b/test/ngTouch/directive/ngClickSpec.js
@@ -314,18 +314,6 @@ describe('ngClick (touch)', function() {
 
       expect($rootScope.count1).toBe(1);
 
-      time = 90;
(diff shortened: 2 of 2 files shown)


OlenDavis:
Might this be included in the 1.4.1 milestone rather than the amorphous 1.4.x milestone? Pretty please?

christianv:
@mikec looks like there are some merge conflicts - would you able to rebase?

mikec:
@christianv: rebased

christianv:
@petebacondarwin would you mind doing a review? Currently we have to load a modified version of the directive with every release until it's fixed.

bruteforce:
The bug was with NgTouch in my case.
Got rid of the NgTouch and replaced it with fastclick directive  for 300ms delay (https://github.com/burnedikt/angular-fastclick) and things are working as expected now.

petebacondarwin:
@christianv - I don't think we can land this PR as it is since you are basically removing something, that even has a test, and was fixing another bug. This means that this PR would cause that bug to reappear.

Instead how about we consider the more drastic action of removing ngClick from ngTouch altogether?
https://github.com/angular/angular.js/pull/2989#issuecomment-22402807

christianv:
@petebacondarwin created a follow up at #12317 - would you be able to take a look?

petebacondarwin:
Closing in favour of #12317

Issue #11099, opened 2015-02-18
angular.copy Very poor performance (large objects)


Tomino2112:
Our application loads resources from API and renders them in a table. API request takes ~200ms, rendering table ~140ms but when you actually looking at the page it takes about 6s to display it.

I have drilled through the code and realized this was caused by using angular.copy . I need to keep track of the old data so I copied the whole object to backup object via angular.copy() function. The copying itself took over 2s (~2500ms). 

I have searched for better ways, to copy objects and stumbled upon this code: 

```
cloneObject: function(o){
            const gdcc = "__getDeepCircularCopy__";
            if (o !== Object(o)) {
                return o; // primitive value
            }

            var set = gdcc in o,
                cache = o[gdcc],
                result;
            if (set && typeof cache == "function") {
                return cache();
            }
            // else
            o[gdcc] = function() { return result; }; // overwrite
            if (o instanceof Array) {
                result = [];
                for (var i=0; i<o.length; i++) {
                    result[i] = this.cloneObject(o[i]);
                }
            } else {
                result = {};
                for (var prop in o)
                    if (prop != gdcc)
                        result[prop] = this.cloneObject(o[prop]);
                    else if (set)
                        result[prop] = this.cloneObject(cache);
            }
            if (set) {
                o[gdcc] = cache; // reset
            } else {
                delete o[gdcc]; // unset again
            }
            return result;
        },
```
This is my implementation of this: http://stackoverflow.com/a/10729242/820942

Replacing all angular.copy() in my code with this **hugely** improved the webapp overall. The same example above takes ~41ms instead of ~2s. 

I should mention that the data I am mentioning is pretty large (renders into 10 000 cells table)

After very brief look on the angular.copy code I am aware of that there are some added angular goodies ($$hashkey) but still the difference in performance is just too big.

pkozlowski-opensource:
@Tomino2112 did you try to put this code and see if it passes all the existing tests? As you've noticed, `angular.copy` is more of an internal utility and we kind of regret exposing it as a public API... but now the damage is done... So if your version works better for your particular case - by all means use it!

It is kind of hard to say more without an isolated reproduction scenario which we could profile to actually see what is going on. We are missing this as well as other essential info (ex.: browser and its version used for tests).

In short: please provide an isolated reproduce scenario, otherwise it is not very actionable.

VitalyTomilov:
[...]

PR #11215, opened 2015-03-02
fix(copy): do not copy the same object twice

jbedard:
#11099 made me curious if there are any easy performance improvements in `copy`. Instead I found some weird edge cases and inconsistencies.  Performance also improved in all the cases I tried (~25% less time & memory in the latest dataset I've been testing, but this will vary greatly depending on data).

Changes:
- Move the stack check before creating new objects instead of before recursing them. This avoids creating objects/arrays just to throw them out in the recursive call. This fixes non-recursed objects being duplicated when there are multiple references.
- Move the stack push to after creating new objects instead of after the recursive call to avoid object/arrays being pushed twice. The extra `return` on the recursive calls is also needed for this.
- Added various tests that were useful while testing different solutions. The 'should handle date/regex objects with multiple references' is the only one that previously failed.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -795,9 +795,18 @@ function copy(source, destination, stackSource, stackDest) {
 
   if (!destination) {
     destination = source;
-    if (source) {
+    if (isObject(source)) {
+      var index;
+      if (stackSource && (index = stackSource.indexOf(source)) !== -1) {
+        return stackDest[index];
+      }
+
+      // TypedArray, Date and RegExp have specific copy functionality and must be
+      // pushed onto the stack before returning.
+      // Array and other objects create the base object and recurse to copy child
+      // objects. The array/object will be pushed onto the stack when recursed.
       if (isArray(source)) {
-        destination = copy(source, [], stackSource, stackDest);
+        return copy(source, [], stackSource, stackDest);
       } else if (isTypedArray(source)) {
         destination = new source.constructor(source);
       } else if (isDate(source)) {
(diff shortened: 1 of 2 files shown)


jbedard:
Rebased. Anyone have a chance to look at this? I'd really like to do a few more things here but was hoping to get this in first to avoid an even more complicated commit...

lgalfaso:
Overall, I like the changes. Just a few comments:

* I do not expect a 25% performance improvements in the general case, just when there is a high self-referential or the same `Date` or `RegEx` objects used multiple times. Is this the case?
* The reason why in the case without a `destination` we sometimes `return` and sometimes do not, is not clear. A comment would be highly welcome.
* All the checks on `stackSource` make me think if the main recursive function should be an inner function and these parameters removed. Ie

```js
function copy(source, destination) {
  var stackSource = [];
  var stackDest = [];

  copy_(source, destination);

  function copy_(source, destination) {
    // Recursive logic goes here, but without the checks on stackSource.
  }
}
```

??

jbedard:
I've added a comment above the confusing sometimes-returning block.

Almost all cases seem to be ~25% faster. I think for a few reasons...
* Previously nested object/arrays were pushed onto the stack twice (=> `indexOf` is more expensive?) [A](https://github.com/angular/angular.js/blob/f7b999703f4f3bdaea035ce692f1a656b0c1a933/src/Angular.js#L824) + [B](https://github.com/angular/angular.js/blob/f7b999703f4f3bdaea035ce692f1a656b0c1a933/src/Angular.js#L877).
* Previously `isObject(result)` and the removed function call were done [per recursive call](https://github.com/angular/angular.js/blob/f7b999703f4f3bdaea035ce692f1a656b0c1a933/src/Angular.js#L876). Now that the `stack.push` has moved this isn't necessary.
* Previously `isArray`/`isRegExp`/`isTypedArray`/`isObject` were called on [every copied value](https://github.com/angular/angular.js/blob/f7b999703f4f3bdaea035ce692f1a656b0c1a933/src/Angular.js#L798). Now these are only called `if (isObject(source))`. This probably makes the biggest difference.

lgalfaso:
landed as 0e622f7b5bc3d5d0ab0fbc1a1bc69404bd7216d5

PR #11222, opened 2015-03-02
ng-cookies squashed

shahata:
@petebacondarwin - the commits are not precisely as you requested since some of the commits you listed didn't seem right to me but I hope it is close enough :)


petebacondarwin:
Apart from a couple of docs tweaks and removing the use of `extend` in `$$cookieWriter` this LGTM.
@shahata - let me know when those are done and we can land this

petebacondarwin:
LGTM - just checking the doc-gen and merging

PR #11233, opened 2015-03-03
docs($q): describe your change...

eddiejhong:
Included an example of how $q.all() is used

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -506,7 +506,7 @@ function qFactory(nextTick, exceptionHandler) {
    *
    * ```js
    *   promiseC = $q.all([promiseA, promiseB]).then(function(result) {
-   *     success: do something and resolve promiseC
+   *     // success: do something and resolve promiseC
    *     return result;
    *   }, function(reason) {
    *     // error: handle the error if possible and
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -504,6 +504,21 @@ function qFactory(nextTick, exceptionHandler) {
    * Combines multiple promises into a single promise that is resolved when all of the input
    * promises are resolved.
    *
+   * ```js
+   *   promiseC = $q.all([promiseA, promiseB]).then(function(result) {
+   *     success: do something and resolve promiseC
+   *     return result;
+   *   }, function(reason) {
+   *     // error: handle the error if possible and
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@eddiejhong I'm afraid that the code example in the current form doesn't add much, as it doesn't illustrate 2 key points about `$q.all`:
* how to get access to results from the individual promises
* what happens if one of the promises fail

If you can think of illustrating those points with a more meaningful example I would be happy to merge it but we will pass on this PR in the current form. Thnx for the suggestion, though!

src/ng/q.js
```diff
@@ -504,6 +504,21 @@ function qFactory(nextTick, exceptionHandler) {
    * Combines multiple promises into a single promise that is resolved when all of the input
    * promises are resolved.
    *
+   * ```js
+   *   promiseC = $q.all([promiseA, promiseB]).then(function(result) {
+   *     // success: do something and resolve promiseC
```
gkalpak: Since you are at it, it might be useful to somehow indicate what `result` looks like (i.e. an array with the resolved values for `promiseA`, `promiseB`).

src/ng/q.js
```diff
@@ -504,6 +504,21 @@ function qFactory(nextTick, exceptionHandler) {
...
+   *     return result;
+   *   }, function(reason) {
+   *     // error: handle the error if possible and
+   *     //        resolve promiseC with newPromiseOrValue,
+   *     //        otherwise forward the rejection to promiseC
+   *     if (canHandle(reason)) {
+   *      // handle the error and recover
+   *      return newPromiseOrValue;
```
gkalpak: The indentation is a little of.

src/ng/q.js
```diff
@@ -504,6 +504,21 @@ function qFactory(nextTick, exceptionHandler) {
...
+   *     }
+   *   });
```
gkalpak: I think you should also show how to properly handle the case where `canHandle()` returns false (i.e. propagate the rejection).

gkalpak:
Ooops ! I commented before seeing @pkozlowski-opensource's comment.
I agree that in its current form, this example doesn't add much (in the sense that it is as likely to confuse someone as it is likely to help them).

(BTW, it would be nice to also showcase the use of an object (instead of Array of promises).)

PR #11236, opened 2015-03-04
Conditional to avoid error message

k8adev:
In some moment, <code>parent</code> returns undefined on load view


gkalpak:
Have we determined under what circumstances this may occur ? It would be preferrable to address the root cause (instead of working around the symptom).

Narretz:
We need more information to consider merging this, such as the scenario where this happens and a test to show it. Feel free to reopen this issue in that case.

PR #11239, opened 2015-03-04
feat(ngMock): allow mock $controller service to set up controller bindings

### LINKED ISSUE #9425: Make controllers used with a bindToController directive easier to test
There doesn't seem to be a clear mechanism of unit-testing a controller that's used in a directive with `bindToController: true`. These controllers assume certain properties being set before the execution of the constructor.
I could build a mechanism myself using a temporary constructor etc. but it would be nice to have something in Angular itself.

I discovered the private boolean arg `later` in the `$controller` service that serves the purpose:

    ctrlFn = $controller(MyCtrl, {
        $scope: scope,
    }, true):
    ctrlFn.instance.foo = "bar";
    ctrl = ctrlFn();

Would you consider making that flag public and thus officially supported?
Or maybe add a helper to angular mock?

See also this [Stack Overflow question](http://stackoverflow.com/questions/25837774/bindtocontroller-in-unit-tests).

caitp:
Adds a new mock for the $controller service, in order to simplify testing using the
bindToController feature.

``` js
var dictionaryOfControllerBindings = {
  data: [
    { id: 0, phone: '...', name: '...' },
    { id: 1, phone: '...', name: '...' },
  ]
};

// When the MyCtrl constructor is called, `this.data ~= dictionaryOfControllerBindings.data`
$controller(MyCtrl, myLocals, dictionaryOfControllerBindings);
```

Closes #9425

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1781,6 +1781,44 @@ angular.mock.$RootElementProvider = function() {
 };
 
 /**
+ * @ngdoc service
+ * @name $controller
+ * @description
+ * Mock $controller service, used to simplify testing controllers when using the
+ * bindToController feature.
+ *
+* @param {Function|string} constructor If called with a function then it's considered to be the
+*    controller constructor function. Otherwise it's considered to be a string which is used
+*    to retrieve the controller constructor using the following steps:
+*
+*    * check if a controller with given name is registered via `$controllerProvider`
+*    * check if evaluating the string on the current scope returns a constructor
+*    * if $controllerProvider#allowGlobals, check `window[constructor]` on the global
+*      `window` object (not recommended)
+*
+*    The string can use the `controller as property` syntax, where the controller instance is published
+*    as the specified property on the `scope`; the `scope` must be injected into `locals` param for this
+*    to work correctly.
+*
+* @param {Object} locals Injection locals for Controller.
+* @param {Object=} bindings Properties to add to the controller before invoking
+*    the constructor. This is used to simulate the `bindToController` feature
+*    and simplify certain kinds oftests.
+* @return {Object} Instance of given controller.
+ */
+angular.mock.$ControllerDecorator = ['$delegate', function($delegate) {
+  return function(expression, locals, later, ident) {
+    if (later && typeof later === 'object') {
+      var create = $delegate(expression, locals, true, ident);
+      angular.extend(create.instance, later);
+      return create();
+    }
+    return $delegate(expression, locals, later, ident);
+  };
+}];
+
+
+/**
  * @ngdoc module
  * @name ngMock
  * @packageName angular-mocks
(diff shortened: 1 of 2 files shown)


caitp:
So:

Should it be a shallow copy, as extend() currently does? This works better if the dictionary contains native values.

src/ngMock/angular-mocks.js
```diff
@@ -1781,6 +1781,44 @@ angular.mock.$RootElementProvider = function() {
...
+*      `window` object (not recommended)
+*
+*    The string can use the `controller as property` syntax, where the controller instance is published
+*    as the specified property on the `scope`; the `scope` must be injected into `locals` param for this
+*    to work correctly.
+*
+* @param {Object} locals Injection locals for Controller.
+* @param {Object=} bindings Properties to add to the controller before invoking
+*    the constructor. This is used to simulate the `bindToController` feature
+*    and simplify certain kinds oftests.
```
gkalpak: oftests  -->  of tests

caitp:
@SonofNun15 / @cesarandreu would an API like this work for you guys? or do you prefer having the actual bindings set up

SonofNun15:
That looks like exactly what I was hoping for. Bindings are not required from my perspective. Thanks so much!

cesarandreu:
@caitp this works for me :D. I don't need bindings.

I've basically gotten rid of big link functions in my directives. In my link function I'll setup watchers and call controller actions, and handle any kind of DOM transformation in as much isolation as possible. The controller is left responsible for tracking state and providing glue to services.

caitp:
okay --- @petebacondarwin if you aren't busy watching the conference, does this look good to you with the typo fixed?

test/ngMock/angular-mocksSpec.js
```diff
@@ -1751,6 +1751,28 @@ describe('ngMock', function() {
...
+      var called = false;
+      var data = [
+        { name: 'derp1', id: 0 },
+        { name: 'testname', id: 1 },
+        { name: 'flurp', id: 2 }
+      ];
+      module(function($controllerProvider) {
+        $controllerProvider.register('testCtrl', function() {
+          called = true;
+          expect(this.data).toEqual(data);
```
petebacondarwin: Do we want to be more strict and say `toBe` here? I know it doesn't really matter but since you asked the question about shallow copying I guess we could make it clear what the result of that was.

petebacondarwin:
LGTM, merging with tweaked docs. Do we want to change the unit test WRT https://github.com/angular/angular.js/pull/11239#discussion_r25936353 ?

caitp:
> LGTM, merging with tweaked docs. Do we want to change the unit test WRT #11239 (comment) ?

sounds good to me

thelgevold:
This is great!
I have added a quick write up with an example of how to use it:
http://www.syntaxsuccess.com/viewarticle/5520197d61d7e9d80a9f52db
Thanks for adding this!

petebacondarwin:
Thanks @thelgevold

PR #9905, opened 2014-11-04
fix($rootScope): remove history event handler when app is torn down

randombk:
Remember the popstate and hashchange handler registered with window
when the application bootstraps, and remove it when the application
is torn down

Closes #9897

PR #11241, opened 2015-03-04
fix($rootScope): allow destroying a root scope

petebacondarwin:
Builds on #10895


randombk:
There are also a few global event handlers that needs to be removed (see PR #9905)

PR #11242, opened 2015-03-04
feat(isNaN): isNumberNaN

PatrickJS:
window.isNaN(‘lol’); //=> true
Number.isNaN(‘lol’); //=> false

isNaN converts it’s arguments into a Number before checking if it’s NaN. Number.isNaN fixes this but we can also polyfill it

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -399,6 +399,10 @@ function toInt(str) {
   return parseInt(str, 10);
 }
 
+var isNumberNaN = Number.isNaN || function isNumberNaN(num) {
+  return num !== num;
+};
+
 
 function inherit(parent, extra) {
   return extend(Object.create(parent), extra);
@@ -1113,7 +1117,7 @@ function fromJson(json) {
 
 function timezoneToOffset(timezone, fallback) {
   var requestedTimezoneOffset = Date.parse('Jan 01, 1970 00:00:00 ' + timezone) / 60000;
-  return isNaN(requestedTimezoneOffset) ? fallback : requestedTimezoneOffset;
+  return isNumberNaN(requestedTimezoneOffset) ? fallback : requestedTimezoneOffset;
 }
 
 
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1259,7 +1259,7 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
       if (isDefined(val) && !isNumber(val)) {
         val = parseFloat(val, 10);
       }
-      minVal = isNumber(val) && !isNaN(val) ? val : undefined;
+      minVal = !isNumberNaN(val) ? val : undefined;
       // TODO(matsko): implement validateLater to reduce number of validations
       ctrl.$validate();
     });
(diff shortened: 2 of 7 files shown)


pkozlowski-opensource:
@gdi2290 is there any reason to polyfill this particular [ES6 method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) inside the core? Shouldn't we rather advice people to include a full / partial polyfill if they need it?

gdi2290:
well I didn't keep this global on the angular object so it's for internal use since it's inside of the closure. Throughout the code base the team is checking if it is a number before running isNaN. Adding this would save a few lines and from this js gotcha 

```javascript
if (isNumber(ctrl.$modelValue) && isNaN(ctrl.$modelValue)) {
// would be
if (isNaN(ctrl.$modelValue) {
//
minVal = isNumber(val) && !isNaN(val) ? val : undefined;
// would be
minVal = !isNaN(val) ? val : undefined;
```

pkozlowski-opensource:
@gdi2290 so, if you believe that the changed isNaN version can be used to improve existing code, could you please include those changes in this PR?

gdi2290:
I think would be better if I name it `isNumberNaN` and replace all `isNaN` to `isNumberNaN` I'll update the PR

jbedard:
and instead of defining a function that wraps both the `Number.isNaN` and pollyfill it should only provide a pollyfill in place of `Number.isNaN`. Something such as...
```javascript
var isNumberNaN = Number.isNaN || function(n) {
   ...
};
```

gdi2290:
good idea
```javascript
var _isNaN = window.isNaN;
var isNumberNaN = Number.isNaN || function isNumberNaN(num) {
  return isNumber(num) ? _isNaN(num) : false;
};
```

realityking:
Nice idea.

`Number.isNaN` can be more easily (and a bit faster) polyfilled like this

```javascript
function isNumberNaN(value) {
  return value !== value;
}
```

src/ng/filter/limitTo.js
```diff
@@ -96,12 +96,12 @@ function limitToFilter() {
...
       limit = toInt(limit);
     }
-    if (isNaN(limit)) return input;
+    if (isNumberNaN(limit)) return input;
 
     if (isNumber(input)) input = input.toString();
     if (!isArray(input) && !isString(input)) return input;
 
-    begin = (!begin || isNaN(begin)) ? 0 : toInt(begin);
+    begin = (!begin || isNumberNaN(begin)) ? 0 : toInt(begin);
```
jbedard: I don't think this one should change. This one specifically wanted `"123"` to pass...

src/Angular.js
```diff
@@ -399,6 +399,11 @@ function toInt(str) {
   return parseInt(str, 10);
 }
 
+var _isNaN = window.isNaN;
+var isNumberNaN = Number.isNaN || function isNumberNaN(num) {
+  return isNumber(num) ? _isNaN(num) : false;
```
jbedard: This could be just `isNumber(num) && _isNaN(num)`. Or what @realityking said probably even better...

src/ng/directive/input.js
```diff
@@ -1259,7 +1259,7 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
       if (isDefined(val) && !isNumber(val)) {
         val = parseFloat(val, 10);
       }
-      minVal = isNumber(val) && !isNaN(val) ? val : undefined;
+      minVal = !isNumberNaN(val) ? val : undefined;
```
caitp: this changes semantic behaviour of the code

PR #11243, opened 2015-03-04
fix(input[type=email]) validate email array created by ng-list

christopher-ramirez:
Fix validation of mails addresses in an array generated by ng-list.


christopher-ramirez:
Bot, I'm the whole author of the commit.

Narretz:
Hi @christopher-ramirez thanks for your PR. However, we cannot merge this right now as we want to tackle the problem differently in the future (1.5ish). The disadvantage of your solution is that you'd need to build the array condition into every validator that should support multiple text inputs. We hope to handle this better in the future, without this overhead. But thanks for contributing!

PR #11244, opened 2015-03-04
aria-checked="{{checked}}" - to reflect check state 

neilyoung:
IMHO this needs to be changed from 
<some-checkbox role="checkbox"  ng-model="checked" ng-class="{active: checked}"
to
<some-checkbox role="checkbox" aria-checked="{{checked}}" ng-model="checked" ng-class="{active: checked}"

in order to be proper read by a screen-reader


Narretz:
This is actually a bug in the current version. If an element has `role="checkbox"` it should set the attribute automatically.

PR #11246, opened 2015-03-05
test(IE8): Remove workarounds for IE8

realityking:
For some reason shrink-wrap just blew on my system, so this is untested.


lgalfaso:
landed as 9b2a4c6851b783bdf15507201c213ac5dc91cae1

PR #9365, opened 2014-10-01
perf($compile): only use document fragments and copy data when necessary (replaceWith)

jbedard:
This does a few things, all within the internal `replaceWith`...
- only creates the document fragment if needed and clarifies why it's there
- only copies data if the node `hasData` to avoid creating empty data to copy (a jqLite version of `hasData` was added)
- uses the static data methods to avoid creating jqLite/jQuery objects
- cleans the data of all replaced elements the same way (this will no longer trigger $destroy on nodes 1+, like node 0, which seems correct to me)
- uses jqLite/jQuery/Array.splice to do the array modifications

PR #11248, opened 2015-03-05
refactor($compile): use indexof/splice instead of manually manipulating jquery object

jbedard:
jQuery/jqLite objects have a `splice` method on them, and `indexof.call` works on them, so those 2 methods can be used instead of the double loop.

The standard case (replacing one element) just does an assignment while other cases use `splice`. This could be changed to always use splice if wanted, but I thought the 99% case was worth it.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -56,6 +56,7 @@
   isElement,
   makeMap,
   includes,
+  indexOf
   arrayRemove,
   copy,
   equals,
@@ -731,8 +732,10 @@ function nodeName_(element) {
   return lowercase(element.nodeName || (element[0] && element[0].nodeName));
 }
 
+var indexOf = Array.prototype.indexOf;
+
 function includes(array, obj) {
-  return Array.prototype.indexOf.call(array, obj) !== -1;
+  return indexOf.call(array, obj) !== -1;
 }
 
 function arrayRemove(array, value) {
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -11,6 +11,8 @@
  *     Or gives undesired access to variables likes document or window?    *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
+/* global indexOf: true */
+
 /* ! VARIABLE/FUNCTION NAMING CONVENTIONS THAT APPLY TO THIS FILE!
  *
  * DOM-related variables:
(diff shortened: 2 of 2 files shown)


lgalfaso:
This looks interesting, would it be possible to get some numbers to know if there are any performance implications?

jbedard:
Seems to make no difference when running `$compile("<div ng-if='true'></div>")` in a loop. Although if I take out the `if (removeCount === 1)` case it is a couple percent slower.

There are other things in this method that can be changed if we want to speed it up though (see #9365).

lgalfaso:
@jbedard can you also try the performance on the case where multiple elements would need to be replaced?

What remains at #9365, I think it would be best if they can be split into more PRs (it you think that only one PR would be needed, then just rebase it once this lands)

jbedard:
Multiple elements seems to have the same result as removing the `if (removeCount === 1)` case for single elements - it is a couple percent slower and a fraction of a percent more memory (because splice returns a new array, I assume).

If you profile only the code changed on its own (`indexOf`+`splice` vs the loop) the loop one is 1-2x faster. But compared to the document fragment and data calls in this method this is insignificant.

Think it's worth removing that loop?

lgalfaso:
I would leave it as it... maybe just
```diff
diff --git a/src/ng/compile.js b/src/ng/compile.js
index 3de0d53..f582206 100644
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2424,12 +2424,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             $rootElement[i++] = newNode;
             for (var j = i, j2 = j + removeCount - 1,
                      jj = $rootElement.length;
-                 j < jj; j++, j2++) {
-              if (j2 < jj) {
-                $rootElement[j] = $rootElement[j2];
-              } else {
-                delete $rootElement[j];
-              }
+                 j2 < jj; j++, j2++) {
+              $rootElement[j] = $rootElement[j2];
             }
             $rootElement.length -= removeCount - 1;
```

jbedard:
I still think this is worth removing that ugly loop, but feel free to close it if you disagree.  Note that your snippet above won't work (I think...) because these are (always? often?) `jqLite` objects and not arrays so the `delete` is needed in addition to the `length` change...

mgol:
I'm going through old jqLite-related issues/PRs. What's the state of this one? Is it still worth landing, @jbedard?

jbedard:
I still think this makes things prettier, but others seemed to disagree. I'm fine closing it...

jbedard:
Looks like this won't happen, closing.

PR #11250, opened 2015-03-05
possible

naterexw:
the expected should be false right? test


gkalpak:
Why do you think it should be `false` ?

Narretz:
I don't see why it should be `false`. Once the request went true, the promise should be resolved. Or did I miss something?

PR #11269, opened 2015-03-07
docs(tutorial/0 - Bootstrapping): Add a "the" before "imperative / manua...

b0ri5:
...l way"

I'm assuming "imperative / manual" is modifying "way" in which case I think "the" is needed.

I don't really know grammar, but as a native speaker it sounds odd without "the".


b0ri5:
I've just signed a CLA.

PR #11270, opened 2015-03-07
docs(tutorial/0 - Bootstrapping): Clarifying where the callback is defined

kjarnet:
Saying that the script-tag registers a callback was for me very confusing, and I think specifying that it is the code in the script that registers the callback makes this a bit clearer.


kjarnet:
Signed the CLA.

Narretz:
Hi, thanks for the PR! I committed another version, which I found a bit simpler. Thanks for the contribution though!

PR #8756, opened 2014-08-24
fix($http): propagate status -1 for aborted requests

pkozlowski-opensource:
Fixes #4491

Issue #4491, opened 2013-10-17
fix($http) Indicate whether a request was aborted via status.


jgrund:
Currently, there does not appear to be a way to determine whether a request was aborted or failed for another reason, as status is normalized to 0 in $http:

```
 /**
  * Resolves the raw $http promise.
  */
  function resolvePromise(response, status, headers) {
    // normalize internal statuses to 0
    status = Math.max(status, 0);

    (isSuccess(status) ? deferred.resolve : deferred.reject)({
      data: response,
      status: status,
      headers: headersGetter(headers),
      config: config
    });
  }
```

Internally a -1 is used to flag that the request either timed out or it was aborted via promise resolution.

What if we exposed the -1 as a valid status to give a better idea to rejection handlers why the request failed?

If this is the case I can create a PR to adress this issue.

PR #11272, opened 2015-03-07
feat($http): flag timed out requests

gabrielmaldi:
This allows differentiating a network error from a timeout.

If you guys thinks this is OK I'll write some tests and update the docs.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1085,7 +1085,7 @@ function $HttpProvider() {
        *  - resolves the raw $http promise
        *  - calls $apply
        */
-      function done(status, response, headersString, statusText) {
+      function done(status, response, headersString, statusText, timedOut) {
         if (cache) {
           if (isSuccess(status)) {
             cache.put(url, [status, response, parseHeaders(headersString), statusText]);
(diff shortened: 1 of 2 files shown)


src/ng/http.js
```diff
@@ -1111,17 +1111,25 @@ function $HttpProvider() {
...
         // normalize internal statuses to 0
         status = Math.max(status, 0);
 
-        (isSuccess(status) ? deferred.resolve : deferred.reject)({
+        var result = {
           data: response,
           status: status,
           headers: headersGetter(headers),
           config: config,
           statusText: statusText
```
lgalfaso: why not just add `timeOut: timeOut` here?

lgalfaso:
this looks interesting, but it needs some tests and documentation.

src/ng/http.js
gabrielmaldi: I didn't want to alter the current success result, since it logically can never be a timeout (hence `timeOut: false` would be redundant; I think of `timeOut` as a possible outcome of an error result). But maybe it'd better to add `timeOut: [true | false]` every time.

Another question: would you go with `timedOut`, `timeOut`, or `timeout`?

pkozlowski-opensource:
We've got an issue for this already (https://github.com/angular/angular.js/issues/4491) with a associated PR (#8756). Would be good to discuss pros and cons of the respective approaches. But yeh, in any case we need tests, doc updates etc.

gabrielmaldi:
I gave it a bit of thought and think that the status approach is better. So I'll wait for #8756 :+1:

PR #11273, opened 2015-03-08
fix(ngAnimate): Change to falsy value in addClass and removeClass

### LINKED ISSUE #11268: animate.js removeClass seems to call setClass with a wrong type
I was getting tracebacks in my angular application.

I tracked it down to what I believe is removeClass calling setClass wrong:

https://github.com/angular/angular.js/blob/master/src/ngAnimate/animate.js#L1114

shows

return this.setClass(element, [], className, options);

while the second argument to setClass is supposed to be a string according to:

* @param {string} add the CSS classes which will be added to the element

but instead a list ([]) is passed in; which throws an error in a foreach:

    forEach(cssClasses.split(' '), function(cssClass) {

where this fails because you can't call split on an empty list.

wesleycho:
- Change addClass & removeClass to pass in string to setClass so
  $$setClassImmediately does not trigger due to arrays always being
  truthy

This fixes #11268 .

### DIFF
--- a/src/ngAnimate/animate.js
+++ b/src/ngAnimate/animate.js
@@ -1077,7 +1077,7 @@ angular.module('ngAnimate', ['ng'])
          * @return {Promise} the animation callback promise
         */
         addClass: function(element, className, options) {
-          return this.setClass(element, className, [], options);
+          return this.setClass(element, className, '', options);
         },
 
         /**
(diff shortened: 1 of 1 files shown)


wesleycho:
Closing, this does not look quite right - need to think some more.

PR #11274, opened 2015-03-09
docs(tutorial/3 - Filtering Repeaters): trim title

wuyiw:
trims the title when query is empty.


Narretz:
I think that's too much for a tutorial. The examples should focus on the exact stuff that is implemented, not the fine-tuning (which might be quite confusing to beginners).

wuyiw:
OK,for a tutorial that does be too much

PR #11275, opened 2015-03-09
fix(jqLite.attr): ignore comment, text and attribute nodes when setting/...

### LINKED ISSUE #11038: JQLite issue setting attributes on comments ¯\_(ツ)_/¯
Here's [a demo](http://jsbin.com/rojexo/2/edit) of the issue (open the console) and you'll see something like this (asside from all the jsbin logs):

screen shot 2015-02-11 at 5 31 22 pm

This is referencing [this line](https://github.com/angular/angular.js/blob/d5b1d6edde194da75f235349783a0712aaab5f7e/src/jqLite.js#L607) which chokes if the element is a comment (evidently). jQuery proper handles this (somehow). If you edit the jsbin, simply uncomment the script tag to include jQuery and you'll see that the error is now gone.

I apologize if the demo is a little contrived, it was how I discovered the bug.

RobiFerentz:
...getting attributes

follow jQuery handling of the attr function

Fixes #11038

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -587,6 +587,10 @@ forEach({
   },
 
   attr: function(element, name, value) {
+    var nodeType = element.nodeType;
+    if (nodeType === NODE_TYPE_TEXT || nodeType === NODE_TYPE_ATTRIBUTE || nodeType === NODE_TYPE_COMMENT) {
+      return;
+    }
     var lowercasedName = lowercase(name);
     if (BOOLEAN_ATTR[lowercasedName]) {
       if (isDefined(value)) {
--- a/test/jqLiteSpec.js
+++ b/test/jqLiteSpec.js
@@ -612,6 +612,30 @@ describe('jqLite', function() {
       expect(elm.attr('readOnly')).toBeUndefined();
       expect(elm.attr('disabled')).toBeUndefined();
     });
+
+    it('should do nothing when setting or getting on attribute nodes', function() {
+      var attrNode = jqLite(document.createAttribute('myattr'));
+      expect(attrNode).toBeDefined();
+      expect(attrNode[0].nodeType).toEqual(2);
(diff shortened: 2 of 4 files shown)


src/jqLite.js
```diff
@@ -587,6 +587,9 @@ forEach({
   },
 
   attr: function(element, name, value) {
+    if (element.nodeType === 2 || element.nodeType === 3 || element.nodeType === 8) {
```
lgalfaso: * please use `NODE_TYPE_TEXT` and `NODE_TYPE_COMMENT`, for nodeType === 2, please create `NODE_TYPE_ATTRIBUTE` to src/Angular.js and use it
* add `var nodeType = element.nodeType` and do the comparison to it

test/jqLiteSpec.js
```diff
@@ -612,6 +612,14 @@ describe('jqLite', function() {
       expect(elm.attr('readOnly')).toBeUndefined();
       expect(elm.attr('disabled')).toBeUndefined();
     });
+
+    it('should ignore nodes that cannot have attributes', function() {
```
lgalfaso: should do nothing when trying to set an attribute to a comment text node or attribute node

and please add the tests for all of them

lgalfaso:
otherwise, LGTM

src/jqLite.js
RobiFerentz: I've added the NODE_TYPE_ATTRIBUTE = 2 to src/Angular.js (also in the jshint comment at the top, to expose it as global). However, running grunt test still gives me an error that NODE_TYPE_ATTRIBUTE is not defined. I must be missing something.

RobiFerentz:
Done. :satisfied:

lgalfaso:
landed as bb5bf7f8162d11610a53428e630b47030bdc38e5

Issue #10007, opened 2014-11-11
Feature Request: angular.component helper


johnlindquist:
Please see this demo: https://jsbin.com/yixote/11/edit?html,js,output

Thanks to "bindToController" in angular 1.3, we're able to create a much simpler api for building components. 

Just as "angular.service" helps us to easily build an "angular.provider", I strongly believe a simplified "angular.component" could help us easily build components with "angular.directive".

I'm sure everyone will have plenty of feedback :)

a-lucas:
@demisx 

I totally agree on the component naming. widget is much more appropriate than component.

demisx:
@a-lucas :+1:  FYI, here is how we define and break down app into components: [AngularAtom component based organization](http://demisx.github.io/angularjs/atom/component-feature-based-organization/2014/12/02/angular-1-component-organization-1.html)) post. A component is so much more, than just a directive.

a-lucas:
@demisx 
I still don't understand the difference between a module and a component. This folder structure described in the link you provided matches almost exactly  the one used in the MEAN.js project. 

I usually organize my code the same way, because it makes it easier to find and visualize files in large projects.  ( The project I actually work on has around a hundred controllers) 

For re-usability purpose, it is better to have every folder being an independent module.  But this is not that easy. Taking the example of a profile module (or profile component). One of the mandatory condition for re usability  of the profile 'component' is that both project uses a very similar data structure.

But again, this can be tweaked into sub-module :  I do have for example my profile module organized this way : 

- profile module
   - personal informations sub-module
       /partials & controllers & routes
   - address sub-module
       /partials & controllers & routes
   - documents sub-modules
       /partials & controllers & routes
   - status sub-modules
       /partials & controllers & routes
and the list goes on.


I just don't get  what advantages components would bring to this existing code organisation.

demisx:
@a-lucas We currently don't use the concept of a widget at all. Entire app is broken down into components. A component can be a widget too. I've just suggested the term `widget` if Angular team were still going to implement this directive-based approach. So it would be `angular.widget` instead of `angular.component`, since directives deal with UI and components deal with so much more, than UI. We see a `component` as a grouping of various files and other sub-components. Components are similar to modules in other languages, but we chose the term component so it's not confused with the Angular modules themselves.

thorn0:
Agree on this. Widget is a better word.

pkozlowski-opensource:
[...]

PR #11278, opened 2015-03-09
An example implementation of .directive `bind`

kentcdodds:
Example implementation to implement #10007

Based on [this comment](https://github.com/angular/angular.js/issues/10007#issuecomment-77763231) by @gautelo.

**Note** This PR was created using the github editor and is expected to be more instructional than anything. Not expected to be merged.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -806,16 +806,22 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             forEach(hasDirectives[name], function(directiveFactory, index) {
               try {
                 var directive = $injector.invoke(directiveFactory);
+                var defaults = {
+                  priority: 0,
+                  name: name,
+                  require: directive.controller && directive.name,
+                  restrict: 'EA'
+                }
+                if (directive.bind) {
+                  addBindDefaults(defaults, name, directive);
+                }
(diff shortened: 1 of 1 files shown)


kentcdodds:
By the way, do we not have a `reverseExtend` function already? Basically, that function is just like `angular.extend` but will not overwrite properties. If there's something already there then awesome.

kentcdodds:
Note, I closed the PR because this is not expected to be merged. Just wanted to provide an example. This PR would amount to an api like so:

```javascript
ngModule.directive('myFoo', function() {
  return {
    bind: {
      bar: '='
    },
    template: '<pre ng-bind="myFoo.bar | json"></pre>'
  };
});
```

Pros:

 - People don't have to learn a new method - simplifies learning curve... maybe
 - Very easily override default options (such as `controllerAs`)

Cons: 

 - People don't have to learn a new method - Overloaded DDO is one of the most confusing parts of angular.

Very interested to hear what others have to say about this.

mzgol:
I'd prefer to see a new method. Having bind set some other properties by default seems confusing. Also, I'd like no new API to set restrict to EA; they should be separated: E for components, A for decorators.

gautelo:
Nice! :)

PR #11280, opened 2015-03-09
Changed aria-live attribute value from polite to assertive

RaphStein:
For ngMessages directive ngAria generates aria-live with value assertive and not polite.

### DIFF
--- a/docs/content/guide/accessibility.ngdoc
+++ b/docs/content/guide/accessibility.ngdoc
@@ -235,7 +235,7 @@ Becomes:
 
 The new ngMessages module makes it easy to display form validation or other messages with priority
 sequencing and animation. To expose these visual messages to screen readers,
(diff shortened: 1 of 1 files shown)


Narretz:
Hi, thanks for the PR! Can you please sign the CLA?

RaphStein:
Hi, i have signed the CLA.

marcysutton:
"Assertive" was added for a reason: so that feedback would be given immediately as users type in text inputs. We talked about making this a configuration option, was that not considered?

Narretz:
I'm not sure if I follow; is this now incorrect? It was only a change in the docs to match the behavior in code, so I assumed it was safe.

marcysutton:
Ah ha, I totally read that wrong! Disregard. :)

Issue #11109, opened 2015-02-20
module.directive('FirstLetterCapitalized', ...) does not work and does not log an error.


jamestalmage:
```javascript
//app.js
angular.module('myModule', []).directive('MyDirective', function() {
  console.log('Hello World!');
  /* ....  make a directive .... */
});
```
```html
<!--- index.html -->
<my-directive></my-directive>
```
This produces absolutely zero output on the console. Changing the directive name to start with a lower case letter fixes the problem. IMO, it should either log an error that I capitalized the first letter, or just work.

pkozlowski-opensource:
Yeh, kind of corner case, but you are right...

PR #11281, opened 2015-03-09
feat($compile): Throw error on incorrect casing

wesleycho:
- Adds error message for when the first letter of the directive name is
  not capitalized in the definition

This addresses #11109 .

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -779,6 +779,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     return bindings;
   }
 
+  function assertValidDirectiveName(name) {
+    var letter = name.charAt(0);
+    if (!letter || letter !== lowercase(letter)) {
+      throw ngMinErr('badname', "Directive name '{0}' is invalid. The first letter of a directive must be a lowercase letter");
+    }
+    return name;
+  }
+
   /**
    * @ngdoc method
    * @name $compileProvider#directive
@@ -797,6 +805,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    this.directive = function registerDirective(name, directiveFactory) {
     assertNotHasOwnProperty(name, 'directive');
     if (isString(name)) {
+      assertValidDirectiveName(name);
       assertArg(directiveFactory, 'directiveFactory');
       if (!hasDirectives.hasOwnProperty(name)) {
         hasDirectives[name] = [];
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -147,6 +147,15 @@ describe('$compile', function() {
 
 
   describe('configuration', function() {
+
+    it('should fail to register a directive that does not start with a lowercase letter', function() {
+      module(function($compileProvider) {
+        expect($compileProvider.directive('BadDirectiveName', function() {
+          return {};
+        })).toThrowMinErr('badname', "Directive name 'BadDirectiveName' is invalid. The first letter of a directive must be a lowercase letter");
+      });
+    });
+
     it('should register a directive', function() {
       module(function() {
         directive('div', function(log) {


src/ng/compile.js
```diff
@@ -779,6 +779,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     return bindings;
   }
 
+  function assertFirstLetterIsLowerCased(name) {
+    var letter = name.charAt(0);
```
lgalfaso: please use the `lowercase` function as it handles some locale issues

src/ng/compile.js
```diff
@@ -779,6 +779,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     return bindings;
   }
 
+  function assertFirstLetterIsLowerCased(name) {
```
lgalfaso: rename to `assertValidDirectiveName`. The function should also check that the directive name is not the empty string

test/ng/compileSpec.js
```diff
@@ -7430,3 +7431,19 @@ describe('$compile', function() {
       }));
   });
 });
+
+describe('$compile', function() {
```
lgalfaso: why does this adds a top level `describe` and not just use the current one?

test/ng/compileSpec.js
```diff
@@ -7430,3 +7431,19 @@ describe('$compile', function() {
...
+  describe('directive', function() {
+    it('should fail to register a directive', function() {
```
lgalfaso: the test description does not describe what it is actually testing

test/ng/compileSpec.js
```diff
@@ -7430,3 +7431,19 @@ describe('$compile', function() {
...
+      try {
+        module(provideLog, function($provide, $compileProvider) {
+          $compileProvider.directive('BadDirectiveName', function(log) {
```
lgalfaso: `log` is not used

test/ng/compileSpec.js
```diff
@@ -7430,3 +7431,19 @@ describe('$compile', function() {
       }));
   });
 });
+
+describe('$compile', function() {
+  describe('directive', function() {
+    it('should fail to register a directive', function() {
+      try {
```
lgalfaso: use `expect(...).toThrowMinErr(...)`

src/ng/compile.js
```diff
@@ -779,6 +779,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+    var letter = name.charAt(0);
+    if (letter !== letter.toLowerCase()) {
+      throw ngMinErr('badname', "Directive '{0}' is not a valid name. Make sure the first letter is lowercase");
```
lgalfaso: Rewrite the error to "The directive name '{0}' is invalid. The first letter of a directive must be a low case letter"

test/ng/compileSpec.js
```diff
@@ -7430,3 +7431,19 @@ describe('$compile', function() {
...
+        module(provideLog, function($provide, $compileProvider) {
```
lgalfaso: `$provide` is not used

lgalfaso: why do you need `provideLog` for this test?

rexyolo13: Need a lil help plz

test/ng/compileSpec.js
wesleycho: This is because it is defining the directive in the provide step - it needs to catch the error when the directive is defined.  It cannot lie in the existing top level describe because in the beforeEach, the module is defined then.  It needs to be defined in the spec itself - is it possible to just define a new module and run the spec, so it happens regardless of the beforeEach?

src/ng/compile.js
wesleycho: `lower` would be more correct than `low` in this case.

test/ng/compileSpec.js
wesleycho: I don't need `provideLog` in particular - is there a default noop module I can instantiate here, or should I create a new one?

src/ng/compile.js
Narretz: I think the second sentence is okay.

lgalfaso: "low case" or "lower case" both work for me, but please make the error describe what the error is and do not make a reference to the person nor what should he/she do

test/ng/compileSpec.js
lgalfaso: what is wrong with just using the function as the first and only argument?

wesleycho: I realized that this morning - just a complete brain fart on my part.

src/ng/compile.js
```diff
@@ -779,6 +779,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     return bindings;
   }
 
+  function assertValidDirectiveName(name) {
+    var letter = name.charAt(0);
+    if (letter && (letter !== lowercase(letter))) {
```
lgalfaso: I think this should be `if (!letter || letter !== lowercase(letter)) {`

test/ng/compileSpec.js
```diff
@@ -147,6 +147,15 @@ describe('$compile', function() {
 
 
   describe('configuration', function() {
+
+    it('should fail to register a directive that does not start with a lowercase letter', function() {
+      module(function($provide, $compileProvider) {
```
lgalfaso: `$provide` is not used

src/ng/compile.js
```diff
@@ -779,6 +779,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+      throw ngMinErr('badname', "Directive '{0}' is invalid. The first letter of a directive must be a lowercase letter");
```
Narretz: Not the directive is invalid, the directive's name is invalid.

petebacondarwin:
We can't use ng:badname as this refers to using `hasOwnProperty` where we shouldn't, see https://docs.angularjs.org/error/ng/badname
I'll knock up a new error and merge this. Thanks @wesleycho

petebacondarwin:
Moreover, the test was not actually running, since you have to also request that the injector instantiates the `$compile` service. You could see that the test was not failing by removing the "fix".

It is better to follow the example of the `hasOwnProperty` test: https://github.com/angular/angular.js/blob/master/test/ng/compileSpec.js#L196-L203

I will fix this up too.

PR #11284, opened 2015-03-10
docs(angular.extend): final param documentation does not match functionality

davej:
The `deep` property is hard-coded to `false` and is not configurable as the docs suggest. I'm not sure if the documentation is incorrect or the implementation is incorrect.


davej:
CLA signed

PR #11300, opened 2015-03-11
Add provider.$decorator to angular.Module

RodneyEbanks:
Allow for decorators to be added prior to the angular.config process.

Use case: I am working on a project using ui.router and would like to modify the $stateProvide.state service and would like to ensure that my modification will always be applied prior to any angular.config blocks which could contain $stateProvide.state requests.

This would also allow any decorators to be clearly delineated in your code and not lost in a config block.


RodneyEbanks:
CLA Signed - Individual email: github@rodneyebanks.com username: RodneyEbanks.

lgalfaso:
Hi, thanks for the PR.
* Please fix the style issues
* A few tests would not hurt :)

PR #11305, opened 2015-03-12
feat(ng, decorator): Add decorator Method (provider.$decorator) to angular.Module

RodneyEbanks:
Allow for decorators to be added prior to the angular.config process.

Use case: I am working on a project using ui.router and would like to modify the $stateProvide.state service and would like to ensure that my modification will always be applied prior to any angular.config blocks which could contain $stateProvide.state requests.

This would also allow any decorators to be clearly delineated in your code and not lost in a config block.

Replaces Pull Request #11300

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -193,6 +193,18 @@ function setupModuleLoader(window) {
            */
           constant: invokeLater('$provide', 'constant', 'unshift'),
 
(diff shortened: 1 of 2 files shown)


RodneyEbanks:
CLA email updated.

PR #11307, opened 2015-03-12
docs(guide/Conceptual Overview): adding a hyphen for clarity

meldaravaniel:
Minor change, but the heading for "View independent business logic..." would be more clear if it had a hyphen, "View-independent".  Perhaps it's because I'm new to javascript and its terminology, but I read "View" as a verb rather than a noun on the first pass and had to read on a bit to understand that it was, instead, referring to The View.   If you go just by grammar rules, making "view independent" into the compound adjective, "view-independent", makes it clear that it is modifying "business logic".  (http://www.grammarbook.com/punctuation/hyphens.asp - see Rule 1).

### DIFF
--- a/docs/content/guide/concepts.ngdoc
+++ b/docs/content/guide/concepts.ngdoc
@@ -179,10 +179,10 @@ The following graphic shows how everything works together after we introduced th
 
 <img style="padding-left: 3em; padding-bottom: 1em;" src="img/guide/concepts-databinding2.png">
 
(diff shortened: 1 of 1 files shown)


meldaravaniel:
Signed a CLA!  Just me by self, no corporation.

Narretz:
Landed as 68146cc09205c3f6eec91626edcd5724abb34b56, thanks!

PR #11318, opened 2015-03-13
docs(guide/Data Binding): add heading to page

meldaravaniel:
the other pages have one, this makes them look more unified.


lgalfaso:
landed, thanks!

PR #11319, opened 2015-03-13
docs(guide/Forms): format headers to match other docs

meldaravaniel:
Primary header with page title, secondary headers for each description section.


lgalfaso:
landed, thanks!

PR #11321, opened 2015-03-13
fix(ngAria): handle custom elements with role="checkbox"

Narretz:
In https://github.com/angular/angular.js/commit/d6eba21733c6e67e90e3a4763d8d41ad89a73a0c#diff-91e9129201746ba107348c9a0a7735edR218, we fixed some issues with checkboxes with string and integer models using ngTrueValue / ngFalseValue, but I did not take into consideration custom `role="checkbox"` elements (as in the ngAria docs). They are currently broken, because the new way of checking "checkedness" is by using ngModel.$isEmpty, which has a special implementation for checkboxes. Custom elements obviously lack this.
My idea is to provide the checkbox $isEmpety implementation as a default to all elements that come with role="checkbox" in the preLink fn. It can then be overwritten by the custom implementation. (since ngAria is simply assuming that a custom checkbox will have a boolean checkstate, which might not be the case).

@marcysutton can you have a look at this?

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -211,88 +211,101 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
     restrict: 'A',
     require: '?ngModel',
     priority: 200, //Make sure watches are fired after any other directives that affect the ngModel value
-    link: function(scope, elem, attr, ngModel) {
+    compile: function(elem, attr) {
       var shape = getShape(attr, elem);
-      var needsTabIndex = shouldAttachAttr('tabindex', 'tabindex', elem);
 
-      function ngAriaWatchModelValue() {
-        return ngModel.$modelValue;
-      }
+      return {
+        pre: function(scope, elem, attr, ngModel) {
+          if (shape === 'checkbox' && attr.type !== 'checkbox') {
+            //Use the input[checkbox] $isEmpty implementation for elements with checkbox roles
+            ngModel.$isEmpty = function(value) {
+              return value === false;
+            };
+          }
+        },
+        post: function(scope, elem, attr, ngModel) {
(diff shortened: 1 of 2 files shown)


src/ngAria/aria.js
```diff
@@ -211,88 +211,100 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
-        case 'checkbox':
-          if (shouldAttachRole(shape, elem)) {
-            elem.attr('role', shape);
-          }
-          if (shouldAttachAttr('aria-checked', 'ariaChecked', elem)) {
-            scope.$watch(ngAriaWatchModelValue, shape === 'radio' ?
-                getRadioReaction() : ngAriaCheckboxReaction);
+      return {
+        pre: function(scope, elem, attr, ngModel) {
+          if (shape === 'checkbox' && attr.role === 'checkbox') {
```
marcysutton: Hasn't `getShape` already checked for `role="checkbox"`? Also, will shimming `ngModel.$isEmpty` for custom checkboxes affect native ones at all?

test/ngAria/ariaSpec.js
```diff
@@ -111,6 +111,15 @@ describe('$aria', function() {
       expect(element.eq(0).attr('aria-checked')).toBe('false');
     });
 
+    it('should handle custom elements with role="checkbox"', function() {
```
marcysutton: :+1:

marcysutton:
@Narretz this looks great. thanks for doing the work!

src/ngAria/aria.js
Narretz: That's true; I was looking for a way to set $isEmpty only for elements that are not inputs. There's probably a better way to do it.

PR #11322, opened 2015-03-13
wip-chore: refactor npm install

Narretz:
This is an attempt to nodejs-ify the install-dependencies.sh script. 

`Gruntfile-test.js` is almost empty, so you can test the install script without having to do a complete `npm install`. Basically, you just need to `npm install rimraf grunt shelljs q` if the removing / install does not go through completely.

The problem is now that `grunt npm-install --gruntfile=Gruntfile-test.js` does not actually call 'rimraf'. It works when I call the file directly via `node` and append `installDependencies()` at the end of the .js file.

### DIFF
--- a/Gruntfile-test.js
+++ b/Gruntfile-test.js
@@ -1,0 +1,10 @@
+'use strict';
+
+var npmInstall = require('./scripts/npm/install-dependencies');
+
+module.exports = function(grunt) {
+
+  grunt.registerTask('npm-install', function() {
+    npmInstall.installDependencies();
+  });
+}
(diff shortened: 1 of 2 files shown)


petebacondarwin:
How about just using the `rimraf.sync` version or the shelljs `rm`, which is also sync. It is not really worth stressing about the async benefits since we want to wait for it to complete anyway. I'll give these a try...

petebacondarwin:
Actually, you are using the `shell.rm` already even on linux. This is already Win32 compatible.

PR #11324, opened 2015-03-13
Install dependencies

petebacondarwin:
Simplifies #11322


petebacondarwin:
How about this @Narretz ?

Narretz:
Unfortunately `shell.rm` is causing the same "Path too long" errors on Windows as regulat `rm`. :/

petebacondarwin:
OK, so `rimraf.sync` it is then, eh?

Narretz:
Looks like it! I'll test it later today, let's hope it's fast (although that shouldn't really matter because npm install is so slow)

Narretz:
I tested it twice and it looks good!

Gruntfile-test.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+var npmInstall = require('./scripts/npm/install-dependencies');
+
+module.exports = function(grunt) {
+
+  grunt.registerTask('npm-install', function() {
+    npmInstall.installDependencies();
+  });
+};
```
petebacondarwin: We need to remove this once we are finished testing

PR #11326, opened 2015-03-14
respect explicit return values from controller functions (fixes #11147)

### LINKED ISSUE #11147: Controller from require is empty when using an explicit return
<strong>AngularJS v1.3.13</strong>

When a directive uses ``require: '^outer'``, the injected controller is ``{}`` in case an explicit return is used. (``controller: function () { return ...; }``)

<strong>This issue does not occur in angular version 1.2.28</strong>

Example: http://jsfiddle.net/s6oyL2xk/
```
var app = angular.module('myApp',[]);

app.directive('outer', function () {
    return {
        restrict: 'E',
        template: '<div>Outer directive</div><div ng-transclude></div>',
        transclude: true,
        controller: function OuterController() {   
            return {
                data: 'outerController data'
            }
        }
    }
});

app.directive('inner', function () {
    return {
        restrict: 'E',
        template: '<div>Inner Directive</div><div ng-bind="test"></div>',
        require: '^outer',
        link: function (scope, element, attrs, outerController) {
            _.isEmpty(outerController) === true; // This should not happen
            scope.test = outerController.data;
        }
    }
});
```

jamestalmage:
When controller functions return an explicit value that value should be what is passed to the linking functions, and to any child/sibling controllers that `require` it. It should also be bound to the data store on the DOM element.

Closes #11147

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1967,7 +1967,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           for (i in elementControllers) {
             controller = elementControllers[i];
             var controllerResult = controller();
-            if (controllerResult !== controller.instance && (isObject(controllerResult) || isFunction(controllerResult))) {
+            if (controllerResult !== controller.instance) {
               controller.instance = controllerResult;
               $element.data('$' + directive.name + 'Controller', controllerResult);
               if (controller === controllerForBindings) {
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1967,13 +1967,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           for (i in elementControllers) {
             controller = elementControllers[i];
             var controllerResult = controller();
-            if (controllerResult !== controller.instance &&
-                controller === controllerForBindings) {
-              // Remove and re-install bindToController bindings
-              thisLinkFn.$$destroyBindings();
-              thisLinkFn.$$destroyBindings =
+            if (controllerResult !== controller.instance && (isObject(controllerResult) || isFunction(controllerResult))) {
+              controller.instance = controllerResult;
+              $element.data('$' + directive.name + 'Controller', controllerResult);
+              if (controller === controllerForBindings) {
+                // Remove and re-install bindToController bindings
+                thisLinkFn.$$destroyBindings();
+                thisLinkFn.$$destroyBindings =
                   initializeDirectiveBindings(scope, attrs, controllerResult,
-                                              bindings, scopeDirective);
+                    bindings, scopeDirective);
+              }
             }
           }
         }
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1974,8 +1974,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
                 // Remove and re-install bindToController bindings
                 thisLinkFn.$$destroyBindings();
                 thisLinkFn.$$destroyBindings =
-                  initializeDirectiveBindings(scope, attrs, controllerResult,
-                    bindings, scopeDirective);
+                  initializeDirectiveBindings(scope, attrs, controllerResult, bindings, scopeDirective);
               }
             }
           }
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -4197,6 +4197,27 @@ describe('$compile', function() {
     });
 
 
+    it('should respect controller return value', function() {
+      module(function() {
+        directive('logControllerProp', function(log) {
+          return {
+            controller: function($scope) {
+              this.foo = 'baz'; // value should not be used.
+              return {foo: 'bar'};
+            },
+            link: function(scope, element, attrs, controller) {
+              log(controller.foo);
+            }
+          };
+        });
+      });
+      inject(function(log, $compile, $rootScope) {
+        element = $compile('<log-controller-prop></log-controller-prop>')($rootScope);
+        expect(log).toEqual('bar');
+      });
+    });
+
+
     it('should get required parent controller', function() {
       module(function() {
         directive('nested', function(log) {
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -4197,7 +4197,7 @@ describe('$compile', function() {
     });
 
 
-    it('should respect controller return value', function() {
+    it('should respect explicit return value from controller', function() {
       module(function() {
         directive('logControllerProp', function(log) {
           return {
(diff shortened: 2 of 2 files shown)


jamestalmage:
CLA updated

petebacondarwin:
@caitp you have done some work on this part of the code. Can you take a look at this one?

caitp:
the only thing that's needed here is updating the element data

src/ng/compile.js
```diff
@@ -1967,13 +1967,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           for (i in elementControllers) {
             controller = elementControllers[i];
             var controllerResult = controller();
-            if (controllerResult !== controller.instance &&
-                controller === controllerForBindings) {
-              // Remove and re-install bindToController bindings
-              thisLinkFn.$$destroyBindings();
-              thisLinkFn.$$destroyBindings =
+            if (controllerResult !== controller.instance && (isObject(controllerResult) || isFunction(controllerResult))) {
```
caitp: in particular, don't worry about the `isObject() || isFunction()` stuff, since that's handled in controller.js anyways

src/ng/compile.js
```diff
@@ -1967,13 +1967,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+              controller.instance = controllerResult;
+              $element.data('$' + directive.name + 'Controller', controllerResult);
```
caitp: there are a few different circumstances that this happens, so we might need to be a bit more careful about this --- be sure to include test coverage that covers controllers for transclude directives too

src/ng/compile.js
```diff
@@ -1967,13 +1967,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+              if (controller === controllerForBindings) {
+                // Remove and re-install bindToController bindings
+                thisLinkFn.$$destroyBindings();
+                thisLinkFn.$$destroyBindings =
                   initializeDirectiveBindings(scope, attrs, controllerResult,
-                                              bindings, scopeDirective);
+                    bindings, scopeDirective);
```
caitp: don't change indentation

caitp:
LGTM with comments addressed

src/ng/compile.js
jamestalmage: I'm not sure what you are asking for here. We should mimic the behavior of native javascript: If they return an object or function use that as the result of the instantiation, otherwise return "as normal":

```javascript
function returnsStr() {
  this.foo = "bar";
  return "baz";
}

new returnsStr() === {foo:"bar"} // "baz" return value is ignored
```

jamestalmage:
unclear what this means

> the only thing that's needed here is updating the element data

Are you asking me to delete the first of [these two lines](https://github.com/jamestalmage/angular.js/commit/57fe7014c7187cb796af4d483e913c3a0dd29d38#diff-a732922b631efed1b9f33a24082ae0dbR1971)?:
```javascript
$element.data('$' + directive.name + 'Controller', controllerResult);
```
Both lines are required for my test to pass.

src/ng/compile.js
caitp: this code is duplicating code which already runs in the `controller()` function itself, it's not needed here

caitp: particularly [here](https://github.com/angular/angular.js/blob/62d514b06937cc7dd86e973ea11165c88343b42d/src/ng/controller.js#L139-L146)

31812015: this diff

caitp:
no, you need that line as mentioned above --- I would like you to add additional tests, though.

1. add another test where a controller which returns a custom object is require'd by another directive, assert that the directive gets the right controller (this is sort of what you're testing already, but it tests what we actually want, rather than the mechanics of it)
2. add similar test cases which verify the transclude cases

src/ng/compile.js
jamestalmage: Nevermind,
I see what you are saying, my above concern is already handled by the call to the `controller()` callback. I am adding a test case that verifies it.

jamestalmage: argh!
Come on Github! Real time updates for line comments!

jamestalmage:
> `2. add similar test cases which verify the transclude cases

your use of plural case(**s**) is sending me down a rabbit hole in `compileSpec`. There are many tests related to transclusion and controllers. 

I've added this one so far:
```javascript
    it('should support controllerAs', function() {
      module(function() {
        directive('main', function() {
          return {
            templateUrl: 'main.html',
            transclude: true,
            scope: {},
            controller: function() {
              this.name = 'lucas';
              return {name: 'george'};
            },
            controllerAs: 'mainCtrl'
          };
        });
      });
      inject(function($templateCache, $compile, $rootScope) {
        $templateCache.put('main.html', '<span>template:{{mainCtrl.name}} <div ng-transclude></div></span>');
        element = $compile('<div main>transclude:{{mainCtrl.name}}</div>')($rootScope);
        $rootScope.$apply();
        expect(element.text()).toBe('template:george transclude:');
      });
    });
```

But there are plenty of other cases. You want me to go nuts and add dozens of test cases, or are there specific scenarios you want tested.

caitp:
you don't need "dozens" of tests, but add ones that deal with a directive whose controller gets the transclude function

jamestalmage:
@caitp  I believe I have addressed all of your concerns, but I am not familiar enough with the inner workings of transclude to be sure I have covered all cases. Please let me know if there is a scenario I have missed (pointers to existing tests would be helpful).

test/ng/compileSpec.js
```diff
@@ -5525,6 +5615,41 @@ describe('$compile', function() {
...
+      it('should copy the explicit return value to all clones', function() {
+        module(function() {
+          directive('makeThree', valueFn({
+            transclude: 'content',
+            controller: function($transclude) {
+              this.foo = 'baz';
+              return {transclude:$transclude, foo: 'bar'};
+            },
+            link: function(scope, el, attr, ctrl) {
+              ctrl.transclude(cloneAttach);
```
caitp: we don't need to actually run the transclude fn, we just need to assert that it's there

caitp: It would be good to make these test cases a little closer, so that the only difference between them is the presence of the transclude property of the DDO. almost there =)

jamestalmage: so proving each clone gets the correct object reference to the parent controller has no value? (serious question)

caitp: no, this is tested elsewhere --- we only really care that the nested directive gets the right controller

jamestalmage: in my mind
`"each clone gets the correct object reference to the parent controller" === "the nested directive gets the right controller"`.

I need to call `$transclude` at least once to see what the nested controller gets, right?

caitp: well to link the nested directive yes, but you don't need to call it 3 times --- an `expect(controller).toBe(the right thing)` is the only real assertion that is needed

jamestalmage: understood.
on it.

jamestalmage:
@caitp, made most of the changes requested, except this one:

> It would be good to make these test cases a little closer, so that the only difference between them is the presence of the transclude property of the DDO

Can you clarify which tests you want to look closer?

caitp:
merged, thanks for the work

jamestalmage:
awesome. Thank you!

Issue #11157, opened 2015-02-24
AngularJS version 1.3.13 throws an error Error: [ngModel:numfmt] Expected `0.00` to be a number


NaomiN:
I am getting the following error:

```
 Error: [ngModel:numfmt] Expected `0.00` to be a number
http://errors.angularjs.org/1.3.13/ngModel/numfmt?p0=0.00
    at REGEX_STRING_REGEXP (angular.js:63)
    at Array.<anonymous> (angular.js:19884)
    at Object.ngModelWatch (angular.js:23289)
    at Scope.$get.Scope.$digest (angular.js:14235)
    at Scope.scopePrototype.$digest (hint.js:1468)
    at Scope.$get.Scope.$apply (angular.js:14506)
    at Scope.scopePrototype.$apply (hint.js:1478)
    at done (angular.js:9659)
    at completeRequest (angular.js:9849)
    at XMLHttpRequest.requestLoaded (angular.js:9790)
```

-----------------------
However, 0.00 is a number, so why would I see such an error?

The object generating this error is this:
```html
  <input type="number" name="creditLimit"
                           id="creditLimit"
                           ng-model="currentAccount.crLimit"
                           class="form-control"
                           min="0"
                           placeholder="1500.00"
                           data-sm:number data-sm:number-format data-accuracy="2" />
```

NaomiN:
If I remove the last line for the custom directive smNumber, I don't get that error.

matsko:
Could you please provide a working example of your code using plunkr? http://plnkr.co/

NaomiN:
It's a bit tricky as this is part of a big complex project, but I'll try later on. I've traced the code and the value was coming as "0.00" (e.g. string) and the first check for not empty was returning false and then the second check for isNumber was returning false as well (perhaps because the value was passed as a string).

NaomiN:
Hi Matias,

The failing code is this:

ctrl.$formatters.push(function(value) {
    if (!ctrl.$isEmpty(value)) {
      if (!isNumber(value)) {
        throw $ngModelMinErr('numfmt', 'Expected `{0}` to be a number',
value);
      }
      value = value.toString();
    }
    return value;
  });

The isNumber function checks the typeof value and not if that value is a
number or not.

In our case we use the following code in HTML:

 <div class="form-group">
                            <label class="control-label col-md-3
col-lg-3">@Labels.percentage:</label>
                            <div class="col-md-6 col-lg-6">
                                <div class="input-group">
[...]

PR #11334, opened 2015-03-16
possible solution for #11157

jamjamg:
- https://github.com/angular/angular.js/issues/11157


petebacondarwin:
I updated the docs (both the input[number] directive and numfmt error) with better information and a runnable example about to deal with this situation.

PR #11340, opened 2015-03-17
fix(ngAria): change accessibility keypress event to use event.which if it is provided

andrewaustin:
fix(ngAria): change accessibility keypress event to use event.which if it is provided

In Firefox, keyboard events for printable characters (e.g. space) do not use event.keyCode. Use event.which if it is provided before falling back to event.keyCode.

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -334,7 +334,8 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
 
         if ($aria.config('bindKeypress') && !attr.ngKeypress && !isNodeOneOf(elem, nodeBlackList)) {
           elem.on('keypress', function(event) {
-            if (event.keyCode === 32 || event.keyCode === 13) {
+            var keyCode = event.which || event.keyCode;
+            if (keyCode === 32 || keyCode === 13) {
               scope.$apply(callback);
             }
 
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -586,7 +586,7 @@ describe('$aria', function() {
 
     var clickFn;
 
-    it('should a trigger click from the keyboard', function() {
+    it('should trigger click from the keyboard', function() {
       scope.someAction = function() {};
 
       var elements = $compile('<section>' +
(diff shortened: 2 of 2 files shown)


andrewaustin:
Oops, this computer had my git email set to another one of my emails. I've signed a CLA using that other email as well.

andrewaustin:
I authored the commit, and both emails are mine.

marcysutton:
:+1: nice work!

caitp:
MDN says:

> Warning: This attribute is deprecated; you should use key instead, if available.

Why do we want to use this?

caitp:
Okay, so DOM 3 keyboard events say you want to use `key` and check for values `Enter` or `' '` --- but the various DOM level specs are so crazy and not consistently followed, that it's really hard to come up with anything concrete as a suggestion.

I think it might be wiser to use `key` if it's available, though.

andrewaustin:
@caitp According to MDN, the ```keypress``` event does not populate ```key``` and there are a number of related bugs in FF. (Chrome and Safari don't see to populate key at all, according to MDN compatibility tables).

See: 

https://developer.mozilla.org/en-US/docs/Web/Events/keypress

and also the related bugs:

https://bugzilla.mozilla.org/show_bug.cgi?id=680830
https://bugzilla.mozilla.org/show_bug.cgi?id=900372

I'm more than happy to update the PR as needed.

caitp:
> The API is also slightly different, which uses key values (string names) instead of keyCodes, but that is easy enough to fix.

I'm aware of that --- but I think we should go with the most consistent implementations of the most recent DOM events spec. Anyway I stopped caring, lgtm

caitp:
FWIW, D3E KeyboardEvent is a work in progress in Blink (https://crbug.com/227231), there seem to be some issues with different vendors shuffling around different versions of the draft, which is making things difficult and slow in the classic W3 style. Blink and WebKit currently both implement a `keyIdentifier` (and apparently have for a number of years) in KeyboardEvent, which has been removed from D3E for some time (last seen in a 2009 version of the draft, http://www.w3.org/TR/2009/WD-DOM-Level-3-Events-20090908/)

It's kind of amazing that WK/Blink are so far behind in this area

andrewaustin:
@caitp What do you think is the best way forward is for this PR? Would you like to see me try and incorporate DOM Level 3? Or do you think this PR still looks okay?

My current understanding is as follows:

| Browser |  ```KeyboardEvent.key``` | ```keyIdentifier``` | event.which |
|-------------| ----------------------------|-----------------------------| -----------------|
| FF (<37) | | | x |
| FF (>37) | x | | |
| Chrome | | x | |
| Safari | | x | |
| IE9 | x | |
| IE10 | x | |
| IE11 | x | |

Source for FF: https://bugzilla.mozilla.org/show_bug.cgi?id=900372
Source for IE: https://msdn.microsoft.com/en-us/library/ie/ff974892(v=vs.85).aspx
Source for Chrome: https://crbug.com/227231

caitp:
What you have is fine for now, I had said LGTM above

Issue #11343, opened 2015-03-17
bindToController does not bind correctly for multiple directives


jtorbicki:
The problem occurs when defining multiple directives on the same element. All of the directives use bindToController as an object and scope: true. Only the first directive's controller gets scope bound.

Here is a fiddle showing the bug: https://jsfiddle.net/rds5muzk/

Current console output:
```
first = 2
second = undefined
```

Expected console output:
```
first = 2
second = 4
```

JWGmeligMeyling:
Can confirm this issue.

majello:
can confirm this issue too. this seems only to refer to bindToController, no dependency on other factors as far as i can tell.

PR #11344, opened 2015-03-17
fix for #11343 - bindToController for multiple directives

jtorbicki:
Fix for issue https://github.com/angular/angular.js/issues/11343

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1968,7 +1968,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             if (scopeDirective && elementControllers[scopeDirective.name]) {
               bindings = scopeDirective.$$bindings.bindToController;
               controller = elementControllers[scopeDirective.name];
-
               if (controller && controller.identifier && bindings) {
                 controllerForBindings = controller;
                 thisLinkFn.$$destroyBindings =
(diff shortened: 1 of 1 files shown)


src/ng/compile.js
```diff
@@ -1960,19 +1960,21 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-          if (scopeDirective && elementControllers[scopeDirective.name]) {
-            bindings = scopeDirective.$$bindings.bindToController;
-            controller = elementControllers[scopeDirective.name];
-
-            if (controller && controller.identifier && bindings) {
-              controllerForBindings = controller;
-              thisLinkFn.$$destroyBindings =
+          for (i in elementControllers) {
+            // Initialize bindToController bindings for new/isolate scopes
+            var scopeDirective = newIsolateScopeDirective || controllerDirectives[i];
```
wesleycho: This does not look right - if `newIsolateScopeDirective` exists, it will just use that N times.

jtorbicki: If you try to combine scope: {} with scope: true on the same element you'll get an error earlier: 
```
Multiple directives [first, second] asking for new/isolated scope on: <div first="{{1+1}}" second="{{2+2}}">
```
So it's either one iteration for isolated scope, or multiple for scope: true

PR #11345, opened 2015-03-17
fix for #11343 - bindToController for multiple directives

jtorbicki:
Second attempt due to merge conflicts :) Fix for issue https://github.com/angular/angular.js/issues/11343

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1970,26 +1970,25 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         }
         if (elementControllers) {
           // Initialize bindToController bindings for new/isolate scopes
-          var scopeDirective = newIsolateScopeDirective || newScopeDirective;
-          var bindings;
-          var controllerForBindings;
-          if (scopeDirective && elementControllers[scopeDirective.name]) {
-            bindings = scopeDirective.$$bindings.bindToController;
-            controller = elementControllers[scopeDirective.name];
-
-            if (controller && controller.identifier && bindings) {
-              controllerForBindings = controller;
-              thisLinkFn.$$destroyBindings =
-                  initializeDirectiveBindings(scope, attrs, controller.instance,
-                                              bindings, scopeDirective);
-            }
-          }
           for (i in elementControllers) {
+            var scopeDirective = newIsolateScopeDirective || controllerDirectives[i];
+            var bindings;
+            var controllerForBindings;
+            if (scopeDirective && elementControllers[scopeDirective.name]) {
+              bindings = scopeDirective.$$bindings.bindToController;
+              controller = elementControllers[scopeDirective.name];
+              if (controller && controller.identifier && bindings) {
+                controllerForBindings = controller;
+                thisLinkFn.$$destroyBindings =
+                    initializeDirectiveBindings(scope, attrs, controller.instance,
+                                                bindings, scopeDirective);
+              }
+            }
             controller = elementControllers[i];
             var controllerResult = controller();
             if (controllerResult !== controller.instance) {
               controller.instance = controllerResult;
-              $element.data('$' + directive.name + 'Controller', controllerResult);
+              $element.data('$' + controllerDirectives[i].name + 'Controller', controllerResult);
               if (controller === controllerForBindings) {
                 // Remove and re-install bindToController bindings
                 thisLinkFn.$$destroyBindings();
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -4033,6 +4033,68 @@ describe('$compile', function() {
     });
 
 
+    it('should bind to multiple directives controllers via object notation (new scope)', function() {
+      var controllerCalled = false;
+      var secondControllerCalled = false;
+      module(function($compileProvider, $controllerProvider) {
+        $controllerProvider.register('myCtrl', function() {
+          expect(this.data).toEqualData({
+            'foo': 'bar',
+            'baz': 'biz'
+          });
+          expect(this.str).toBe('Hello, world!');
+          expect(this.fn()).toBe('called!');
+          controllerCalled = true;
+        });
+        $controllerProvider.register('secondCtrl', function() {
+          expect(this.data).toEqualData({
+            'foo2': 'bar2',
+            'baz2': 'biz2'
+          });
+          expect(this.str).toBe('Hello, second world!');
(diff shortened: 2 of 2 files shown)


pkozlowski-opensource:
@jtorbicki could you please add test(s) for this as well? Thnx!

jtorbicki:
Test added

test/ng/compileSpec.js
```diff
@@ -4033,6 +4033,68 @@ describe('$compile', function() {
...
+          scope: true,
+          controller: 'myCtrl as myCtrl'
+        }));
+        $compileProvider.directive('barDir', valueFn({
+          bindToController: {
+            'data': '=barData',
+            'str': '@barStr',
+            'fn': '&barFn'
+          },
+          scope: true,
```
caitp: technically, you're not supposed to have multiple "new scope" directives on the same element, so this doesn't really work in practice (if it does work, it's a bug!).

I'm a bit uncomfortable with this because, given the above, it could become unclear what's happening if you have directives which implicitly depend on there being a new scope directive on the same element to work correctly.

jtorbicki: @caitp This is not what the issue is about. The "scope: true" functionality creating the common scope that's being shared among the directives is already in the Angular. Those directives can interact with one another by reading and writing to the scope directly without the need of bindToController. Whether this is good or not is probably up for another discussion. This issue is about bindToController binding the data for only single directive leaving all others in the undefined state. That's obviously not correct. It should either bind for all, or bindToController should be disabled for "scope: true". 
On a side note: In my view there is nothing wrong with multiple directives sharing the same common scope, especially with the bindToController option that binds the data for different controllers in all directives, so there is no danger of polluting the scope.

caitp: Believe it or not, I understand what this issue is about --- The problem is the "scope: true" behaviour being shared by multiple directives on the same element. This was never meant to be supported, there is a WONTFIX'd bug about this, because it's explicitly __not__ supported, it's just that we fail to throw the error sometimes (depending on types of scopes used, and directive priorities). This is never something that should be depended on

sbialy:
@caitp I believe @jtorbicki is right - it seems perfectly fine for multiple directives to share a scope using ```scope: true```. https://github.com/angular/angular.js/issues/4402 is probably the one you're referring to - and it seems to me that this behaviour **is** intended and supported.
Using ```bindToController``` makes it inconsistent so this PR is valid as long as multiple ```scope: true``` directives on the same element are allowed.

caitp:
@sbialy no, that's not the issue I'm referring to --- It is not supported for there to be multiple new scope directives, whether they are new isolate scopes, or new "proper" scopes, on the same element.

jtorbicki:
@caitp Can you provide any reference that multiple "scope: true" are not supported? All I could find in the docs was this: https://docs.angularjs.org/error/$compile/multidir and it doesn't confirm what you'r saying. 
Another thing is a comment from compile.js: 
<code>
This directive is trying to add a child scope. Check that there is no isolated scope already
</code>
Again, only isolated scope is mentioned.

sbialy:
@caitp I know what the issue is about - but the fact that it's allowed to create multiple ```scope: true``` and in the same time it's **not** allowed to have an isolated scope and any other scope says something. If it was not supported it would be very easy to throw an exception in that situation as well, and yet it's not the case.

caitp:
unfortunately it's not that easy @sbialy, the compiler is a pretty complicated piece of machinery, there are a lot of factors at play, and they aren't organized in a way to make this trivial. We do __attempt__ to throw, but it doesn't always work out

sbialy:
I think the reason is more obvious @caitp - it's just legal: https://docs.angularjs.org/error/$compile/multidir specifies invalid configuration and ```scope: true``` is not among them.

JWGmeligMeyling:
The [docs](https://docs.angularjs.org/api/ng/service/$compile) say:
> child scope + child scope => Both directives will share one single child scope
Doing so currently breaks `bindToController`, and I think it unexpectedly does so.
I will test this PR and see if it works for me :smile:

test/ng/compileSpec.js
petebacondarwin: We do support multiple directives asking for new normal scope (i.e. not isolate) as can be seen in the docs (https://docs.angularjs.org/api/ng/service/$compile#-scope-):

> The scope property can be true, an object or a falsy value:
>
>**true**: A new child scope that prototypically inherits from its parent will be created for the directive's element. If multiple directives on the same element request a new scope, only one new scope is created. The new scope rule does not apply for the root of the template since the root of the template always gets a new scope.

So this is a valid bug and the fix looks pretty good to me too. I will give it a review and land this week.

src/ng/compile.js
```diff
@@ -1970,26 +1970,25 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-
-            if (controller && controller.identifier && bindings) {
-              controllerForBindings = controller;
-              thisLinkFn.$$destroyBindings =
-                  initializeDirectiveBindings(scope, attrs, controller.instance,
-                                              bindings, scopeDirective);
-            }
-          }
           for (i in elementControllers) {
+            var scopeDirective = newIsolateScopeDirective || controllerDirectives[i];
```
petebacondarwin: Why have you changed this from checking `newScopeDirective` to checking `controllerDirectives[i]`?

src/ng/compile.js
```diff
@@ -1970,26 +1970,25 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+            var bindings;
+            var controllerForBindings;
+            if (scopeDirective && elementControllers[scopeDirective.name]) {
```
petebacondarwin: I think is probably not quite right any more, since `elementControllers[scopeDirective.name]` is not related to the current iteration of this for loop

src/ng/compile.js
petebacondarwin: Oh hold on, you are actually calculating the controller directive here - probably should use a different name for the var.

src/ng/compile.js
petebacondarwin: I see now that the `scopeDirective` is actually the current controller directive

petebacondarwin:
@jtorbicki,  @sbialy and @JWGmeligMeyling - I agree that this is a valid bug and the fix is approximately correct. There was some churn in this part of the compiler not long ago, so I am rebasing this and merging. Thanks

PR #11350, opened 2015-03-17
chore(ci): make wait_for_browser_provider time out after 2 minutes

juliemr:
Before, if something went wrong, wait_for_browser_provider.sh would
wait indefinitely, and logs would never get printed. Now, we'll bail
early, and get some actual logs on what the problem was.


petebacondarwin:
LGTM - merging

petebacondarwin:
and cherry-picked into 1.3.x

PR #11351, opened 2015-03-17
docs(error/inprog): describe your change...typos?

Cayan:


petebacondarwin:
@Cayan: Thanks for your pull request but this change does not improve the documentation here. The point of the error is that some bugs are hard to detect if they do not throw an error. Instead you get weird and inexplicable behaviour that is hard to track down. If you would like to reword the error doc along those lines that would be fine.

Issue #11143, opened 2015-02-22
grunt with new npm:install has problems on Windows


Narretz:
The new npm install / caching behavior in grunt is causing some issues. Basically, when the new npm-install scripts tries to remove the contents of the node_modules folder I am getting a whole lot of cannot remove <really long path>: File or Path name too long. 
This is the stupid Windows restriction on very long file paths, I think. 
Because of this, the following npm install + shrinkwrap copy is never called, so I am ending up with a half empty node_modules folder.
I can empty the node_modules folder manually and call `scripts/npm/install-dependencies.sh` directly and this works, but I'd be running into this problem every time the shrinkwrap changes. 
Is it possible to delete the contents of node_modules differently?

cc @IgorMinar

IgorMinar:
if rm fails, we don't want to try to install because npm will almost certainly not install all the files. so you'll see random errors at runtime.

can you figure out what needs to be done on windows to make the rm command just work?

I don't have easy access to windows nor the required expertise 😞

Narretz:
It seems we can't use the rm command because Windows has a stupid 260 char restriction on file paths. It's a wontfix on joyents's side: https://github.com/joyent/node/issues/6960
There exists a module that emulates rm -rf specifically for node: https://github.com/isaacs/rimraf but I am not sure what would happen if it removes itself.
Or could we use npm uninstall x for all registered modules? I just checked, this works perfectly
That is, we'd need to get a list of modules in the package.json and then uninstall each of them manually

IgorMinar:
try `rimraf`. measure if it adds significant overhead. if it does use it only on windows.

`npm uninstall` sounds very unefficient, but I don't know if it actually is.

Bit-Rot:
rimraf doesn't seem to like deleting itself.  I get build errors attempting this.

I've found success in replacing the rm -rf call in install-dependencies.sh with:

```
npm ls --depth=0 --parseable=true | xargs npm remove
```

I'm not confident that this is solution in all environments, however.

Narretz:
@Bit-Rot have you tried the code in this PR, and if not can you test if it works? https://github.com/angular/angular.js/pull/11353

Bit-Rot:
@Narretz Just verified that the code in #11353 works for me.  Thanks :)

xie-qianyue:
@Narretz I also tested the commit in #11353, it works for me. But why nobody solve the merge problem to make the pull request success?

AmazingJaze:
[...]

PR #11353, opened 2015-03-17
refactor npm install (again)

Narretz:


petebacondarwin:
@IgorMinar  - can you take a quick look at this when you get a chance?

lib/grunt/utils.js
```diff
@@ -6,11 +6,39 @@ var shell = require('shelljs');
...
+
+function shrinkwrapChanged() {
+  return shell.exec('diff ' + SHRINKWRAP_FILE + ' ' + SHRINKWRAP_CACHED_FILE, {silent: true}).output;
+}
+
+function cleanNodeModules() {
+  if (process.platform == "win32") {
+    rimraf.sync('node_modules');
+    console.log('cleaned node_modules using rimraf');
+  } else {
```
mzgol: Why this `else`? `rimraf` is cross-platform.

Narretz: I think rm -rf might still be more robust, so we should use it if possible. @IgorMinar should probably have a word here.

petebacondarwin: @IgorMinar wanted to use `rm -rf` where possible for performance reasons

Narretz:
This has been confirmed to work on Windows in https://github.com/angular/angular.js/issues/11143 so I'm gonna merge it at the end of the week unless somebody has any objections.

Narretz:
Although I'd like @IgorMinar to take a quick look. :)

petebacondarwin:
Let's go with this @Narretz

Narretz:
Aah, travis was actually using the install-dependencies.sh script. We need to fix this first.

Narretz:
@petebacondarwin do you have an idea how we could handle this? I fear we have to duplicate the functionality, because we can't use the js file when travis is installing.

petebacondarwin:
Hmm. That is a pain. And of course this doesn't work if `npm install` has not already been run as `rimraf` would not be available anyway.
Perhaps we just need an option not to use this at all on Windows?

mzgol:
> I fear we have to duplicate the functionality, because we can't use the js file when travis is installing.

Why can't we? Node is already available, instead of:
```yaml
  - scripts/npm/install-dependencies.sh
```
we'd just use:
```yaml
  - node scripts/npm/install-dependencies.js
```
What am I not seeing here?

petebacondarwin:
@mzgol the problem is that the new version of the `install-dependencies.xxx`, which happens to be a Grunt task now, relies upon things like:

```js
var rimraf = require('rimraf');
```

which of course will fail if we have not run `npm install` at least once already as that module will not yet exist locally.

Or am I missing something?

mzgol:
@petebacondarwin Right... Could we just include `rimraf` manually in the repository and require it from this specific folder instead of `node_modules`? It's just 412 KB.

Or we could copy over the [rimraf.js file](https://github.com/isaacs/rimraf/blob/master/rimraf.js) and edit out `glob` which adds most of these 412 KB since we don't use glob patterns in our command (it would make it harder to upgrade the package later, though).

petebacondarwin:
On non-Windows machines we can just use `rm -rf`. It is only on Windows machines that we need an alternative solution. So I think that we should try to do the switch at a higher level.

petebacondarwin:
Closing in favour of #12792

PR #11354, opened 2015-03-17
chore(ci): force travis to print logs after driver provider timeout

juliemr:
Travis does not do the after_script step if before_script fails,
so wait_for_browser_provider.sh was not printing out logs.
Force it to print them manually.


scripts/travis/wait_for_browser_provider.sh
```diff
@@ -9,6 +9,7 @@ while [ ! -f $BROWSER_PROVIDER_READY_FILE ]; do
   let "counter++"
   if [ $counter -gt 240 ]; then
     echo "Timed out after 2 minutes waiting for browser provider ready file"
+    ./print_logs.sh
```
IgorMinar: can you please add a comment about the before_script / after_script issue?

otherwise lgtm

juliemr: Done.

juliemr:
Merged to master and 1.3.x

PR #11355, opened 2015-03-18
feat($http): CORS requests support for IE9

e-oz:
IE9 use it's own way to do CORS requests, so for this case (only) we should use XDomainRequest. More about it: http://www.html5rocks.com/en/tutorials/cors/

### DIFF
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -4,6 +4,10 @@ function createXhr() {
     return new window.XMLHttpRequest();
 }
 
+function createXdr() {
+    return new window.XDomainRequest();
+}
+
 /**
  * @ngdoc service
  * @name $httpBackend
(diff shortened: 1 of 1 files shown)


jamm:
IE9 still supported by AngularJS 1.3 and 1.4, so I decided to submit code I use for CORS requests in IE9.  I'm not sure we need MockXdr function to mock cross-origin requests and responses (because nobody use it except this rare case, which is temporary), so I just tested if it works in existing project in IE9 and it works. Also all tests by "grunt test" are passing and Travis showing green labels for most test jobs (I think `"ScriptTimeoutError: Timed out awaiting response to command "switchToFrame"` is some bug, not related to CORS).

pkozlowski-opensource:
@jamm this was [previously rejected](https://github.com/angular/angular.js/pull/1047) after a [long discussion](https://github.com/angular/angular.js/issues/934) due to the fact that XDomainRequest has several limitations and doesn't behave the same way as XHR in other browsers.

Has anything changed from the last time we were looking into this?

jamm:
@pkozlowski-opensource this commit has nothing common with one you mentioned. My code doesn't require any changes in config, works fine with $resource (all tests are green) and actually it's contribution from "production" project where this code works fine. Only difference in AngularJS with this code and without this code - CORS requests don't work in IE8 and IE9, and work with this code. 
Yes, XDomainRequest have limitations (such as "GET or POST" only, no returned headers), but from my experience, if project requires support of IE9, then in this project programmers will deal with these limitations. As AngularJS current, 1.4, 1.5 are going to support IE9, I thought it will be useful (it takes hours to implement workaround for CORS, there are some non-documented points). It was strange for to discover that AngularJS doesn't have this workaround as IE9 support promised.

pkozlowski-opensource:
So, I think that reasoning was that XDomainRequest request has so many limitations that it would be unfair to bring it to the framework as users of IE9 would expect $http to work the same across all the browsers and it wouldn't be the case with XDomainRequest. We just simply can't make IE9 behave as other browsers when it comes to CORS...

jamm:
There's a huge difference between "doesn't work at all" and "can't use some features". Following your rule, we should declare lack of CORS requests support, because currently they don't work similarly in all supported browsers. Kidding, of course. But if you think AngularJS will be better without it (or this code will make AngularJS worse), feel free to close this PR.

pkozlowski-opensource:
Yeh, I think that we are better of by not pretending that we support sth that we can't really support fully due to IE9 limitations :-(

PR #11359, opened 2015-03-18
fix(ngTouch): prevent touchmove from canceling a click

pomerantsev:
Usually browsers fire touchstart and touchend, and ng-click works.
But some browsers (I only experienced it on iOS) fire touchmove
between touchstart and touchend, and so before this fix ng-click
would not work in this case.
Also, the test for touchmove was incorrect: it did not test
the touchmove event itself, but rather the coordinates change,
which is handled elsewhere. If the coordinates are the same and
the interval is not too long, ng-click should fire, no matter if
there was a touchmove or not.


petebacondarwin:
Hi @pomerantsev - this is a duplicate of #10985

PR #11365, opened 2015-03-18
Ensure boolean values for aria-* attributes

fredsa:
aria-\* (e.g. aria-hidden) values should mirror the truthiness of their ng-*
counterparts (e.g. ng-show, ng-hide) instead of their actual value

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -115,9 +115,8 @@ function $AriaProvider() {
       var ariaCamelName = attr.$normalize(ariaAttr);
       if (config[ariaCamelName] && !attr[ariaCamelName]) {
         scope.$watch(attr[attrName], function(boolVal) {
-          if (negate) {
-            boolVal = !boolVal;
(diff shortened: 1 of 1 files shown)


fredsa:
Example of a problem:

<input type="text" model="foo">
<div ng-show="foo">…</div>

If a user types "bar" into the text field, foo will have a string value of "bar" and ng-show will correctly make the div visible. However, with ngAria module in use, aria-hidden attribute will (without this pull request) also have the value "bar", which is invalid. Instead, we want aria-hidden to mirror the truthiness of ng-show, and just have a value of true or false.

Narretz:
Hi, thanks for the PR. Would you be able to write a test for this?

Narretz:
Just to make sure, @marcysutton do you think this a valid issue?

marcysutton:
That does sound valid to me–`aria-hidden` needs a string `true` or `false` value to properly hide.

PR #11336, opened 2015-03-16
ngAnimate 1.4 Refactor

matsko:
All of ngAnimate has been rewritten to make the internals of the
animation code more flexible, reuseable and performant.

BREAKING CHANGE: JavaSript and CSS animations can no longer be run in
parallel. With earlier versions of ngAnimate, both CSS and JS animations
would be run together when multiple animations were detected. This
feature has now been removed, however, the same effect, with even more
possibilities, can be achieved by injecting `$animateCss` into a
JavaScript-defined animation and creating custom CSS-based animations
from there. Read the ngAnimate docs for more info.

BREAKING CHANGE: The function params for `$animate.enabled()` when an
element is used are now flipped. This fix allows the function to act as
a getter when a single element param is provided.

``` js
// < 1.4
$animate.enabled(false, element);

// 1.4+
$animate.enabled(element, false);
```

BREAKING CHANGE: In addition to disabling the children of the element,
`$animate.enabled(element, false)` will now also disable animations on
the element itself.

BREAKING CHANGE: Animation-related callbacks are now fired on
`$animate.on` instead of directly being on the element.

``` js
// < 1.4
element.on('$animate:before', function(e, data) {
  if (data.event === 'enter') { ... }
});
element.off('$animate:before', fn);

// 1.4+
$animate.on(element, 'enter', function(data) {
  //...
});
$animate.off(element, 'enter', fn);
```

BREAKING CHANGE: There is no need to call `$scope.$apply` or
`$scope.$digest` inside of a animation promise callback anymore
since the promise is resolved within a digest automatically (but a
digest is not run unless the promise is chained).

``` js
// < 1.4
$animate.enter(element).then(function() {
  $scope.$apply(function() {
    $scope.explode = true;
  });
});

// 1.4+
$animate.enter(element).then(function() {
  $scope.explode = true;
});
```

BREAKING CHANGE: When an enter, leave or move animation is triggered then it
will always end any pending or active parent class based animations
(animations triggered via ngClass) in order to ensure that any CSS
styles are resolved in time.

PR #11369, opened 2015-03-19
feat($animateRunner): prototype code for $animateRunner

matsko:
DO NOT MERGE. JUST A PR FOR REVIEW PURPOSES.

### DIFF
--- a/src/ngAnimate/animateRunner.js
+++ b/src/ngAnimate/animateRunner.js
@@ -1,0 +1,132 @@
+'use strict';
+
+var $$rAFMutexFactory = ['$$rAF', function($$rAF) {
+  return function() {
+    var passed = false;
+    $$rAF(function() {
+      passed = true;
+    });
+    return function(fn) {
+      passed ? fn() : $$rAF(fn);
+    };
+  }
+}];
+
+var $AnimateRunnerFactory = ['$q', '$$rAFMutex', function($q, $$rAFMutex) {
+  var DONE_PENDING_STATE = 1;
+  var DONE_COMPLETE_STATE = 2;
+
+  AnimateRunner.chain = function(chain, callback) {
+    var status = true;
+    var index = 0;
+
+    next();
+    function next() {
+      if (index === chain.length) {
+        callback(status);
+        return;
+      }
+
+      chain[index](function(response) {
+        status = status && response;
+        index++;
+        next();
+      });
+    }
+  };
+
+  AnimateRunner.all = function(runners, callback) {
+    var count = 0;
+    var status = true;
+    forEach(runners, function(runner) {
+      runner.done(onProgress);
+    });
+
+    function onProgress(response) {
+      status = status && response;
+      if (++count === runners.length) {
+        callback(status);
+      }
+    }
+  };
+
+  function AnimateRunner(host) {
+    this.setHost(host);
+
+    this._doneCallbacks = [];
+    this._atleastOneRAFPassedSinceStart = $$rAFMutex();
+    this._state = 0;
+  };
+
+  AnimateRunner.prototype = {
+    setHost: function(host) {
+      this.host = host || {};
+    },
+
+    done: function(fn) {
+      if (this.state === DONE_COMPLETE_STATE) {
+        fn();
+      } else {
+        this._doneCallbacks.push(fn);
+      }
+    },
+
+    progress: noop,
+
+    getPromise : function() {
+      if (!this.promise) {
+        var self = this;
+        this.promise = $q(function(resolve, reject) {
+          self.done(function(status) {
+            status === false ? reject() : resolve();
+          });
+        });
+      }
+      return this.promise;
+    },
+
+    then: function(resolveHandler, rejectHandler) {
+      return this.getPromise().then(resolveHandler, rejectHandler);
+    },
+
+    pause: function() {
+      this.host.pause && this.host.pause();
+    },
+
+    resume: function() {
+      this.host.resume && this.host.resume();
+    },
+
+    end: function() {
+      this.host.end && this.host.end();
+      this._resolve(true);
+    },
+
+    cancel: function() {
+      this.host.cancel && this.host.cancel();
+      this._resolve(false);
+    },
+
+    complete: function(response) {
+      var self = this;
+      if (self._state === 0) {
+        self._state = DONE_PENDING_STATE;
+        self._atleastOneRAFPassedSinceStart(function() {
+          self._resolve(response);
+        });
+      }
+    },
+
+    _resolve: function(response) {
+      if (this._state !== DONE_COMPLETE_STATE) {
+        angular.forEach(this._doneCallbacks, function(fn) {
+          fn(response);
+        });
+        this._doneCallbacks.length = 0;
+        this._state = DONE_COMPLETE_STATE;
+      }
+    }
+  };
+
+  return AnimateRunner;
+}];
--- a/test/ngAnimate/animateRunnerSpec.js
+++ b/test/ngAnimate/animateRunnerSpec.js
@@ -1,0 +1,251 @@
+'use strict';
+
+describe("$$rAFMutex", function() {
+  beforeEach(module('ngAnimate'));
+
+  it('should fire the callback only when one or more RAFs have passed',
+    inject(function($$rAF, $$rAFMutex) {
+
+    var trigger = $$rAFMutex();
+    var called = false;
+    trigger(function() {
+      called = true;
+    });
+
+    expect(called).toBe(false);
+    $$rAF.flush();
+    expect(called).toBe(true);
+  }));
+
+  it('should immediately fire the callback if a RAF has passed since construction',
+    inject(function($$rAF, $$rAFMutex) {
+
+    var trigger = $$rAFMutex();
+    $$rAF.flush();
+
+    var called = false;
+    trigger(function() {
+      called = true;
+    });
+    expect(called).toBe(true);
+  }));
+});
+
+describe("$animateRunner", function() {
+
+  beforeEach(module('ngAnimate'));
+
+  they("should trigger the host $prop function",
+    ['end', 'cancel', 'pause', 'resume'], function(method) {
+
+    inject(function($animateRunner) {
+      var host = {};
+      var spy = host[method] = jasmine.createSpy();
+      var runner = new $animateRunner(host);
+      runner[method]();
+      expect(spy).toHaveBeenCalled();
+    });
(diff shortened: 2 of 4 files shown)


src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
+'use strict';
+
+var $$rAFMutexFactory = ['$$rAF', function($$rAF) {
```
petebacondarwin: Not sure about calling this a Mutex...

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+      passed ? fn() : $$rAF(fn);
+    };
+  }
+}];
+
+var $AnimateRunnerFactory = ['$q', '$$rAFMutex', function($q, $$rAFMutex) {
+  var DONE_PENDING_STATE = 1;
+  var DONE_COMPLETE_STATE = 2;
+
+  AnimateRunner.chain = function(chain, callback) {
```
petebacondarwin: This is a function to run a sequence of functions in series right?

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+
+      chain[index](function(response) {
+        status = status && response;
+        index++;
+        next();
+      });
+    }
+  };
+
+  AnimateRunner.all = function(runners, callback) {
```
petebacondarwin: Looks good

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+
+    this._doneCallbacks = [];
+    this._atleastOneRAFPassedSinceStart = $$rAFMutex();
+    this._state = 0;
+  };
+
+  AnimateRunner.prototype = {
+    setHost: function(host) {
+      this.host = host || {};
+    },
```
petebacondarwin: So this is where the host gets added at some point after the runner was created, right?
Looks good

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+        callback(status);
+      }
+    }
+  };
+
+  function AnimateRunner(host) {
+    this.setHost(host);
+
+    this._doneCallbacks = [];
+    this._atleastOneRAFPassedSinceStart = $$rAFMutex();
```
petebacondarwin: I would just call this something like `batchCall` or `runInAnimationFrame`. The fact that it uses rAF does not need to be exposed.

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+    },
+
+    cancel: function() {
+      this.host.cancel && this.host.cancel();
+      this._resolve(false);
+    },
+
+    complete: function(response) {
+      var self = this;
+      if (self._state === 0) {
```
petebacondarwin: Use a constant for `0` here, perhaps `INITIAL_STATE`

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+      }
+    }
+  };
+
+  function AnimateRunner(host) {
+    this.setHost(host);
+
+    this._doneCallbacks = [];
+    this._atleastOneRAFPassedSinceStart = $$rAFMutex();
+    this._state = 0;
```
petebacondarwin: Use a constant here, maybe `INITIAL_STATE`

petebacondarwin:
I haven't had time to look at the tests but the main code looks very nice.

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+        self._state = DONE_PENDING_STATE;
+        self._atleastOneRAFPassedSinceStart(function() {
+          self._resolve(response);
+        });
+      }
+    },
+
+    _resolve: function(response) {
+      if (this._state !== DONE_COMPLETE_STATE) {
+        angular.forEach(this._doneCallbacks, function(fn) {
```
gkalpak: `angular.` can be ommitted (for consistency).

src/ngAnimate/animateRunner.js
```diff
@@ -0,0 +1,132 @@
...
+      }
+      return this.promise;
+    },
+
+    then: function(resolveHandler, rejectHandler) {
+      return this.getPromise().then(resolveHandler, rejectHandler);
+    },
+
+    pause: function() {
+      this.host.pause && this.host.pause();
```
gkalpak: I think this type statements are discouraged (if not "frown upon") inside the Angular codebase.
Just saying... :)

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
+'use strict';
+
+describe("$$rAFMutex", function() {
```
gkalpak: No, no, no, no !
No !
No, no annnd no !

Consistent use of `'`, please :stuck_out_tongue:

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
...
+    var capturedValue;
+    runner.done(function(val) {
+      capturedValue = val;
+    });
+    runner.complete('special value');
+    $$rAF.flush();
+    expect(capturedValue).toBe('special value');
+  }));
+
+  they("should immediately resolve each combined runner in depth-first order when $prop is called",
```
gkalpak: I don't think `depth-first` is an accurate description. You probably meant something like `bottom-up` or `fifo` or something.

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
...
+      runner2.then(
+        function() { status2 = 'pass'; },
+        function() { status2 = 'fail'; });
+
+      runner1[method]();
+
+      var expectedStatus = method === 'end' ? 'pass' : 'fail';
+
+      expect(status1).toBeFalsy();
+      expect(status2).toBeFalsy();
```
gkalpak: Aren't they expected to be undefined ? If so, using `toBeUndefined()` would be easier to follow.
(`toBeFalsy()` had me wondering if I had missed something.)

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
...
+  they("should resolve/reject using a newly created promise when .then() is used upon $prop",
+    ['end', 'cancel'], function(method) {
+
+    inject(function($animateRunner, $rootScope) {
+      var runner1 = new $animateRunner();
+      var runner2 = new $animateRunner();
+      runner1.setHost(runner2);
+
+      var status1;
+      runner2.then(
```
gkalpak: Was this supposed to be `runner1` ?

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
...
+      expect(status1).toBe(expectedStatus);
+      expect(status2).toBe(expectedStatus);
+    });
+  });
+
+  it("should expose/create the contained promise when getPromise() is called",
+    inject(function($animateRunner, $rootScope) {
+
+    var runner = new $animateRunner();
+    expect(isPromiseLike(runner.getPromise()));
```
gkalpak: I don't think you are testing anything (`.toBeTruthy()` missing ?).
<sub>(At least Jasmine 2.x won't verify anything just by calling `expect()`. Too lazy to try with 1.x :D)</sub>

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
...
+      runner3.complete(true);
+
+      expect(status).toBeUndefined();
+
+      $$rAF.flush();
+
+      expect(status).toBe(true);
+    }));
+
+    they("should immediately resolve if and when all runners have resolved been $prop",
```
gkalpak: _have been $prop_ ?

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
...
+
+        runner1[method]();
+        runner2[method]();
+        runner3[method]();
+
+        expect(status).toBe(expectedStatus);
+      });
+    });
+
+    it("should return a status of `false` if more than one runner was cancelled",
```
gkalpak: You probably mean _if one or more_.
(_"more than one"_ means _at least 2_.)

test/ngAnimate/animateRunnerSpec.js
```diff
@@ -0,0 +1,251 @@
...
+      items.push(function(fn) { log.push(3); runner3.done(fn); });
+
+      var status;
+      $animateRunner.chain(items, function(response) {
+        status = response;
+      });
+
+      runner1.complete(true);
+      runner2.complete(true);
+      runner3.complete(true);
```
gkalpak: This does not make it clear (nor tests) what happens if the runners are complete in a different order than the order they appear in `items`.

Would it be better/more explicit to `complete()` them in a different order and verify that the log order still matches their order in `items` ?

gkalpak:
I really like where the whole "animations in Angular" story is going !
I am pretty sure @matsko will put this `$animateRunner` in good use for us :)

I have left some comments inline. (Nothing major, apart from the `'`/`"` inconsistency :P)

--
Some final thoughts/questions:

1. Angular is supposed to:
    * _**prefer simple objects**, instead of complex inheritance hierarchies_
    * _use prototypal inheritance only when absolutely necessary_
    * _**love functions and closures**_

  I know there isn't any _complex inheritance hierarchy_ here, but still wondering if there would be any benefit in using closures instead (probably not).

2. I understand this is not merge-ready yet, but (before merging) it would be nice to have more docs/comments (not necessarily public) on what each method does and what purpose it serves. It would be much easier to maintain/enhance. 
<sub>`ngAnimate` is pretty much a "one-matsko-show" right now, but this might (have to) change soon, considering the increasing complexity of animations and the rapidly expading "surface" they need to cover (Material Design/ngMaterial, Angular2, need for complex animation choreographies etc).</sub>

src/ngAnimate/animateRunner.js
matsko: I would prefer (fn || noop) but it doesn't work for some reason.

matsko:
@gkalpak thanks a bunch for the great work with the review. I've made the changes, with the exception of the last one. I think I will redo that test.

matsko:
Closed in favour of #11336.

PR #11381, opened 2015-03-20
fix(ngInput): change URL_REGEXP to better match RFC3987

### LINKED ISSUE #11341: URL type input does not support file:/// protocol.
Very easy to reproduce. Any form input of type 'url' will return undefined for a file:/// protocol entry. 

To reproduce go to the following Plunker and enter 'file://some/file' to the input. The HTML5 validation succeeds, but the file protocol is not supported.

http://plnkr.co/edit/ywKtnQQWzWZqdG9gKccS

andrewaustin:
The URL_REGEXP in use to perform validation in ngInput is too restrictive and fails to
follow RFC3987. In particular, it only accepts ftp, http, and https scheme components and
rejects perfectly valid schemes such as "file", "mailto", "chrome-extension",
etc. The regex also requires the scheme to be followed by two "/" but the RFC says
0 to n are acceptable. This change fixes both of these issues to better align to
the standard.

Closes #11341

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -11,7 +11,8 @@
 
 // Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
 var ISO_DATE_REGEXP = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/;
-var URL_REGEXP = /^(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-\/]))?$/;
+// See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
+var URL_REGEXP = /^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/]*)?$/;
 var EMAIL_REGEXP = /^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+@[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i;
 var NUMBER_REGEXP = /^\s*(\-|\+)?(\d+|(\d*(\.\d*)))\s*$/;
 var DATE_REGEXP = /^(\d{4})-(\d{2})-(\d{2})$/;
--- a/test/ng/directive/inputSpec.js
+++ b/test/ng/directive/inputSpec.js
@@ -2387,8 +2387,18 @@ describe('input', function() {
     describe('URL_REGEXP', function() {
       /* global URL_REGEXP: false */
       it('should validate url', function() {
+        // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
         expect(URL_REGEXP.test('http://server:123/path')).toBe(true);
+        expect(URL_REGEXP.test('https://server:123/path')).toBe(true);
+        expect(URL_REGEXP.test('file:///home/user')).toBe(true);
+        expect(URL_REGEXP.test('mailto:user@example.com?subject=Foo')).toBe(true);
+        expect(URL_REGEXP.test('r2-d2.c3-p0://localhost/foo')).toBe(true);
+        expect(URL_REGEXP.test('abc:/foo')).toBe(true);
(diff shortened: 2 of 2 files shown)


andrewaustin:
Does anything else need to be done for this?

thatmarvin:
:+1: I need to validate `rtsp://` urls, so this would be useful for me too.

plestik:
Anything holding this back?

Narretz:
Is the regey taken from somewhere or is it a custom one? I'd like to follow an established regex if possible. What does Chrome's regex look like these days?

andrewaustin:
It's custom, if you can tell me where to find Google's I'll update it to match.

src/ng/directive/input.js
```diff
@@ -11,7 +11,8 @@
 
 // Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
 var ISO_DATE_REGEXP = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/;
-var URL_REGEXP = /^(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-\/]))?$/;
+// See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
+var URL_REGEXP = /^([0-9]+|\w+|\.|\+|\-)+:\/?\/?(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-\/]))?$/;
```
gkalpak: Why `\w+` (in `/^([0-9]+|\w+...` instead of `[A-Za-z]+` ?
It seems like `\w+` will (incorrectly) match characters like `_`.

If fact, I would expect more something like:

`/^([A-Za-z][A-Za-z0-9.+-]+):...`

gkalpak: Or are non-latin letters also allowed ?

gkalpak: According to KURL's implementation (used by WebKit (used by Chromium)), only `A-Z`, `a-z` and `0-9` are allowed: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/platform/weborigin/KURL.cpp&q=kurl.&sq=package:chromium&type=cs&l=116

gkalpak:
A little research:

* (As expected), Chrome (<= Chromium <= WebKit) does not use a RegExp, they [parse the URL][1] instead.
* They use [KURL][2] for that.

--
I don't think it's worth implementing a URL parser into core.
Let's work o that RegExp :)


[1]: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/core/html/forms/URLInputType.cpp&q=typeMismatch&sq=package:chromium&type=cs&l=56
[2]: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/platform/weborigin/KURL.cpp&q=kurl.&sq=package:chromium&type=cs&l=1

gkalpak:
Based on a quick walk-through the code used in Chromium (and specifically [url_parse.cc][1]) and also considering the fact that we don't want to parse/break-up the url (only validate it), I would propose the following RegExp:

`/^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/]*)?$/`

This is not strictly following neither the RFC nor KURL's implementation, but I think it works fine for our needs (considering the previous RegExp wasn't following RFC/Chrome either).

--
I made some minor improvements to the original one (removed duplicate duplicate characters, removed unnecessary escapes, stopped tracking groups etc) and fixed a bug:

By trying to match the host with `\S+` it would match everything including the host, port, path, query and fragment (since `+` matches eagerly). The intention was clearly to match the host only, so I replaced it with `[^\s/]+` (which more closely matches KURL's implementation).

Besides the cases fixed by @andrewaustin's version, there are a couple more cases that are recognized incorrectly by his RegExp (which are also fixed with the above):

* `0scheme://example.com`: Should be invalid (because `scheme` shouldn't start with a digit or `.+-`).
* ```http://example.com:9999/~~`` ```:	Should be invalid (because of `~`s ans `` ` ``s in the path).

[1]: https://code.google.com/p/chromium/codesearch#chromium/src/url/third_party/mozilla/url_parse.cc

andrewaustin:
@gkalpak I updated the PR with your suggested REGEX. I also added two more tests for the two urls you pointed out that should be invalid.

gkalpak:
It LGTM (obviously :P).
Considering how the previous RegExp didn't work perfectly either, I would be fine with this getting merged, if it looks good to some other team-member as well :)

PR #11386, opened 2015-03-20
[WIP] feat($httpUrlBuilderFactory): introduce new service abstracting url building

jmendiara:
Here you have my proposal for having configurable strategies for `$http` url building.
This is a WIP, and tries to solve same problems than #11238 addresses, but with a different perspective, covering more use cases and adding more freedom.

**Key points**: 
- New service `$httpUrlBuilderFactory`, that serves as **foundation** for making totally custom urls.
- Default overridable implementation in core
- Users can customize _per request_ how a url is built, or set a new default to the whole app.

**Rationale**:
URL Building and parameter serialization are complementary problems.
Serialization can be seen as converting objects to a set of `{key, value}` Strings, and URL Building is aggregating this set to create an encoded URL (`?key=value`) With this separation, users can rehuse some of the parts: how to serialize (`var arr = [1,2] => {arr[], 1}, {arr[],2}`) and how to build (encode, concat params with `;` ...) 

Having in mind keeping core small, and give full control to user for easy customization, only the current building URL is shipped, but the users can make whatever they want. `jquery` and other param serialization would be coded easily in userland.

**Examples**

``` js
// Example 1: Making pirate URLs
angular.module('App')
.factory('pirateBuildUrl', function($httpUrlBuilderFactory) {
  function pirateSerializer(params, addKeyValue) {
    //params is the $http request `params` config property
    angular.forEach(params, function(value, key) {
      addKeyValue('ARR' + key, value);
    });
  }
  return $httpUrlBuilderFactory(pirateSerializer);
})
.config(function($httpProvider) {
  //Specify default urlBuilder for the whole app 
  //Compatibility with defaults.cache usage: you can set also a function here
  $httpProvider.defaults.buildUrl = 'pirateBuildUrl';
})
.run(function() {
  $http.get('http://myapi.com/things', {
    params: {
      id: 5
    },
    // or specify per request urlBuild (or a function, like the `cache` property)
    buildUrl: 'pirateBuildUrl'
}); // GET http://myapi.com/things?ARRid=5
```

``` js
// Example 2: freedom in urlBuilding. Crazy things are possible. Nice for testing
angular.module('App')
.factory('testBuildUrl', function() {
  return function(url, params) {
    return url + '?' + JSON.stringify(params); 
  };
})
.run(function() {
  $http.get('http://myapi.com/things', {
    params:  { id: 5 },
    buildUrl: 'testBuildUrl'
}); // GET http://myapi.com/things?{"id":5}
```

I'd this **full** core customization available in userland, and forget about crazy tests and interceptors to communicate with complex backends. WDYT?


jmendiara:
Closed by decision taken in #11238

PR #11388, opened 2015-03-21
fix(cookieReader): safely access $document so it can be mocked

### LINKED ISSUE #11373: lastCookieString.split error with httpBackend.flush()
Using v1.4.0-build.3918+sha.4588e62 my unit test are now having the following error.

```
    TypeError: 'undefined' is not an object (evaluating 'lastCookieString.split'
)
        at C:/TR Forks/fef/bower_components/angular/angular.js:17354
        at sendReq (C:/TR Forks/fef/bower_components/angular/angular.js:9824)
        at C:/TR Forks/fef/bower_components/angular/angular.js:9542
        at processQueue (C:/TR Forks/fef/bower_components/angular/angular.js:140
60)
        at C:/TR Forks/fef/bower_components/angular/angular.js:14076
        at C:/TR Forks/fef/bower_components/angular/angular.js:15283
        at C:/TR Forks/fef/bower_components/angular/angular.js:15098
        at C:/TR Forks/fef/bower_components/angular-mocks/angular-mocks.js:1508
        at C:/TR Forks/fef/test/spec/services/FefLoagger.js:263
```

The unit test is this

```js
it('test mult call is correct format.', function() {
        var success = true;
        httpBackend.expectPOST('/demo/logs')
            .respond(
                function(method, url, data, headers) {
                    var validData = {
                        message: ['test'],
                        level: 'log',
                        name: 'fef:log1'
                    };
                    data = JSON.parse(data);
                    if (data.message[0] === validData.message[0] &&
                        data.level === validData.level &&
                        data.name === validData.name) {
                        return [200, {
                            token: '12345'
                        }];
                    } else {
                        success = false;
                        return [400, {
                            failure: 'bad'
                        }];
                    }
                });
        var log1 = FefLogger.getLogger('log1');
        log1.log('test');
        httpBackend.flush();
        expect(success).toBe(true);

    });
```

the line 263 coralates with the httpBackend.flush();
this unit test did work with 1.4.0-beta.0 and 1.3.15

shahata:
Closes #11373

### DIFF
--- a/src/ng/cookieReader.js
+++ b/src/ng/cookieReader.js
@@ -10,7 +10,7 @@
  * @return {Object} a key/value map of the current cookies
  */
 function $$CookieReader($document) {
-  var rawDocument = $document[0];
+  var rawDocument = $document[0] || {};
   var lastCookies = {};
   var lastCookieString = '';
 
@@ -24,9 +24,10 @@ function $$CookieReader($document) {
 
   return function() {
     var cookieArray, cookie, i, index, name;
+    var currentCookieString = rawDocument.cookie || '';
 
-    if (rawDocument.cookie !== lastCookieString) {
-      lastCookieString = rawDocument.cookie;
+    if (currentCookieString !== lastCookieString) {
+      lastCookieString = currentCookieString;
       cookieArray = lastCookieString.split('; ');
       lastCookies = {};
 
--- a/test/ng/documentSpec.js
+++ b/test/ng/documentSpec.js
@@ -6,4 +6,24 @@ describe('$document', function() {
   it("should inject $document", inject(function($document) {
     expect($document).toEqual(jqLite(document));
   }));
+
+
+  it('should be able to mock $document object', function() {
+    module({$document: {}});
+    inject(function($httpBackend, $http) {
+      $httpBackend.expectGET('/dummy').respond('dummy');
+      $http.get('/dummy');
+      $httpBackend.flush();
+    });
+  });
+
+
+  it('should be able to mock $document array', function() {
+    module({$document: [{}]});
+    inject(function($httpBackend, $http) {
+      $httpBackend.expectGET('/dummy').respond('dummy');
+      $http.get('/dummy');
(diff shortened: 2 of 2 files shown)


shahata:
I'm really not sure that this is the right way to solve this. I'm thinking it might be a good idea to have some mock implementation of `cookieReader`/`cookieWriter` in `angular-mocks` which can be configured during tests similarly to `$httpBackend`. We used to have something similar in `$browser`, but it wasn't documented and it got refactored out during the `ngCookies` re-write (which is why this problem only occurred now)

Narretz:
@petebacondarwin probably comes up with a good idea how to solve this, he always does :)

src/ng/cookieReader.js
```diff
@@ -22,6 +22,13 @@ function $$CookieReader($document) {
     }
   }
 
+  //null object implementation in case document was mocked out
+  if (!rawDocument || typeof rawDocument.cookie !== 'string') {
+    return function() {
+      return lastCookies;
+    };
+  }
+
```
petebacondarwin: I think this is probably overkill. If the developer mocks out window and doesn't include a document then lots of stuff is going to break. Instead, how about we simply do

```js
lastCookieString = rawDocument.cookie || '';
```

In the function below.

shahata: actually this is the only place that breaks, you can mock out `$window` and not include `$document`. See https://github.com/angular/angular.js/blob/master/test/ng/windowSpec.js

shahata: I can also do something like this, if you prefer:
```js
rawDocument = $document[0] || {};
lastCookieString = rawDocument.cookie || '';
```

petebacondarwin: Ugh! Mocking window without a document is surely crazy juice!
OK let's go with this version.

shahata: okay, amended.

petebacondarwin:
LGTM - feel free to merge

PR #11405, opened 2015-03-23
Fix ng style lazy one-time binding regression

Narretz:
There's no easy way to make the one-time binding work woth $watchCollection, so the easiest way is probably to revert the commit, as suggested by @lgalfaso


lgalfaso:
Landed

PR #11411, opened 2015-03-24
fix(ngModel): allow setting model to NaN when asyncValidator is present

### LINKED ISSUE #11315: Setting model to NaN causes infdig if $asyncValidators is used
An infinite digest error is thrown when a model is set to `NaN` and there is an asynchronous validator. 

Please check plunker at http://plnkr.co/edit/u9Sq12?p=preview.
When we set `model` to a string or to null - everything works as expected.
When we set `model` to `NaN` there is an infdig error:

```
Watchers fired in the last 5 iterations: []
http://errors.angularjs.org/1.3.14/$rootScope/infdig?p0=10&p1=%5B%5D
    at angular.js:63
    at Scope.$digest (angular.js:14281)
    at Scope.$apply (angular.js:14506)
    at HTMLButtonElement.<anonymous> (angular.js:21443)
    at HTMLButtonElement.eventHandler (angular.js:3014)
angular.js:63 Uncaught Error: [$rootScope:infdig] 10 $digest() iterations reached. Aborting!
Watchers fired in the last 5 iterations: []
http://errors.angularjs.org/1.3.14/$rootScope/infdig?p0=10&p1=%5B%5D
```

This happens because [`ngModelWatch`](https://github.com/angular/angular.js/blob/59205d7809630fb061c849171aad9c6b2a22f09a/src/ng/directive/ngModel.js#L813) checks that `modelValue !== ctrl.$modelValue`, which is true for `NaN`.

Thus, when we set the model to `NaN`, `$$runValidators` is called in every digest cycle. If we have an async validator, it causes a push to `asyncQueue` and so another cycle is needed.

`ngModelWatch` returns `modelValue`, which is `NaN` so the watch is not considered dirty because [`$digest()`](https://github.com/angular/angular.js/blob/59205d7809630fb061c849171aad9c6b2a22f09a/src/ng/rootScope.js#L766) treats `NaN` as equal to `NaN`.

I think a good solution would be to treat `NaN` as equal to `NaN` in `ngModelWatch`
Edit: 
Check http://plnkr.co/edit/aZZmyP?p=preview
`ngModelWatch` is changed to
```
if (modelValue !== ctrl.$modelValue &&
      !(typeof modelValue === 'number' && typeof ctrl.$modelValue === 'number' && isNaN(modelValue) && isNaN(ctrl.$modelValue))) {
      ctrl.$modelValue = ctrl.$$rawModelValue = modelValue;
```
and this indeed solved the infdig

Narretz:
Fixes #11315

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -810,7 +810,10 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
 
     // if scope model value and ngModel value are out of sync
     // TODO(perf): why not move this to the action fn?
-    if (modelValue !== ctrl.$modelValue) {
+    // Second check is needed to allow setting the model to NaN when there's an asyncValidator
+    if (modelValue !== ctrl.$modelValue &&
+       (ctrl.$modelValue === ctrl.$modelValue || modelValue === modelValue)
+    ) {
       ctrl.$modelValue = ctrl.$$rawModelValue = modelValue;
       parserValid = undefined;
 
--- a/test/ng/directive/ngModelSpec.js
+++ b/test/ng/directive/ngModelSpec.js
@@ -578,6 +578,30 @@ describe('ngModel', function() {
 
         dealoc(form);
       }));
+
+
+      it('should set NaN as the $modelValue when an asyncValidator is present',
+        inject(function($q) {
+
+        ctrl.$asyncValidators.test = function() {
+          return $q(function(resolve, reject) {
+            resolve();
+          });
+        };
+
+        scope.$apply('value = 10');
+        expect(ctrl.$modelValue).toBe(10);
+
+        expect(function() {
+          scope.$apply(function() {
+            scope.value = NaN;
+          });
+        }).not.toThrowMinErr('$rootScope', 'infdig', '10 $digest() iterations reached. Aborting!\n' +
+          'Watchers fired in the last 5 iterations: []');
+
+        expect(ctrl.$modelValue).toBeNaN();
+
+      }));
     });
 
 


test/ng/directive/ngModelSpec.js
```diff
@@ -578,6 +578,30 @@ describe('ngModel', function() {
...
+        };
+
+        scope.$apply('value = 10');
+        expect(ctrl.$modelValue).toBe(10);
+
+        expect(function() {
+          scope.$apply(function() {
+            scope.value = NaN;
+          });
+        }).not.toThrowMinErr('$rootScope', 'infdig', '10 $digest() iterations reached. Aborting!\n' +
```
gkalpak: I would prefer a plain `not.toThrow()`, because this test might become useless if the minErr changes.

src/ng/directive/ngModel.js
```diff
@@ -810,7 +810,10 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
 
     // if scope model value and ngModel value are out of sync
     // TODO(perf): why not move this to the action fn?
-    if (modelValue !== ctrl.$modelValue) {
+    // Second check is needed to allow setting the model to NaN when there's an asyncValidator
+    if (modelValue !== ctrl.$modelValue &&
+       (ctrl.$modelValue === ctrl.$modelValue || modelValue === modelValue)
```
petebacondarwin: Can we use `!isNaN(...)` here?

Narretz: Quick tests shows that `isNaN` does not work, but `Number,isNaN` (ECMA6) does. :(

petebacondarwin: `isNaN(undefined) === true`!

gkalpak: Don't take it personally, but we all are `NaN`: `isNaN('@petebacondarwin') === true` :smiley:

PR #11412, opened 2015-03-24
fix(select): don't call $render twice if $viewValue ref changes

### LINKED ISSUE #11329: select multi: $render is called twice when $viewValue reference changes
In the select directive there is a deep [watch](https://github.com/angular/angular.js/blob/master/src/ng/directive/select.js#L201) that calls `$render` if `$viewValue` deep-changes. This happens either if it changes inside or if it's reference changed.

It should only call $render in the first cases.

If the reference of `$viewValue` changes than this call is redundant because the shallow [watch](https://github.com/angular/angular.js/blob/master/src/ng/directive/ngModel.js#L826) in `ngModel` will call `$render`.

I suggest something like:
```
    var lastView, lastViewRef = NaN;
    scope.$watch(fucntion() {
      if (lastViewRef === ngModelCtrl.$viewValue and !equals(lastView, ngModelCtrl.$viewValue)) {
        lastView = shallowCopy(modelCtrl.$viewValue);
        modelCtrl.$render();
      }
      lastViewRef = ngModelCtrl.$viewValue;
    })
```

Edit: Here is a plunker that show the issue
http://plnkr.co/edit/dWpTKH?p=preview

Narretz:
Credits to @tepez for the fix

Closes #11329

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -200,11 +200,13 @@ var selectDirective = function() {
 
         // we have to do it on each watch since ngModel watches reference, but
         // we need to work of an array, so we need to see if anything was inserted/removed
+        var lastView, lastViewRef = NaN;
         scope.$watch(function selectMultipleWatch() {
-          if (!equals(lastView, ngModelCtrl.$viewValue)) {
+          if (lastViewRef === ngModelCtrl.$viewValue && !equals(lastView, ngModelCtrl.$viewValue)) {
             lastView = shallowCopy(ngModelCtrl.$viewValue);
             ngModelCtrl.$render();
           }
+          lastViewRef = ngModelCtrl.$viewValue;
         });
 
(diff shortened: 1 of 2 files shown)


src/ng/directive/select.js
```diff
@@ -200,11 +200,13 @@ var selectDirective = function() {
 
         // we have to do it on each watch since ngModel watches reference, but
         // we need to work of an array, so we need to see if anything was inserted/removed
+        var lastView, lastViewRef = NaN;
```
petebacondarwin: `lastView` is already declared up top isn't it?

Narretz: True - now that I see it, isn't it a bit weird that it is defined outside the link fn? So every selectDirective uses the same variable?

petebacondarwin: I agree so I moved it inside.

PR #11413, opened 2015-03-24
docs($routeProvider): Document resolved values

fredsa:
Document that resolved values are available in `$route.current.locals`

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -110,6 +110,7 @@ function $RouteProvider() {
    *        resolved before its value is injected into the controller. Be aware that
    *        `ngRoute.$routeParams` will still refer to the previous route within these resolve
    *        functions.  Use `$route.current.params` to access the new route parameters, instead.
+   *        Finally, use `$route.current.locals` to access resolved dependencies.
    *
    *    - `redirectTo` – {(string|function())=} – value to update
    *      {@link ng.$location $location} path with and trigger route redirection.


petebacondarwin:
In what situation are they available? If you want them in the controller then they are available via injection.

fredsa:
This is about the resolved promises provided in the 'resolve' map of the route.

pkozlowski-opensource:
@fredsa even if those are available, why not using DI for them (all the locals are injectable to controllers defined on a route level)? Is it about a "big" number of locals to be injected? If so maybe we should say in the docs when one way of doing things "wins" over another?

fredsa:
@pkozlowski-opensource AFAICT the resolved values are only available via `$route.current.locals` and not injectable.

Here's the code I'm using. Note the assignment of `$scope.products` from `$route.current.locals.products` in `$routeChangeSuccess` event handler:

```
.config(['$routeProvider', function($routeProvider) {
    $routeProvider
      .when('/products', {
        templateUrl: 'partial/products.html',
        resolve: {
          products: 'products_promise'
        },
        reloadOnSearch: false,
      })
})

.factory('products_promise', ['$http', function($http) {
  //fetch JSON from server
  return $http.get('/api/products')
  .then(function(result) {
    return result.data;
  });
})

.controller('ProductsController', ['$scope', '$route', function($scope, $route) {
  $scope.$on('$routeChangeSuccess', function(evt, current, previous) {
    $scope.products = $route.current.locals.products;
  });
}])
```


Note further:
```
// This doesn't work because there is no `products` service/factory:
.controller('ProductsController', ['$scope', '$route', 'products', function($scope, $route, products) {
  $scope.$on('$routeChangeSuccess', function(evt, current, previous) {
    $scope.products = products;
  });
}])
```

and:
```
// This doesn't work because `products_promise` is a promise, rather than the desired resolved value:
.controller('ProductsController', ['$scope', '$route', 'products_promise', function($scope, $route, products_promise) {
  $scope.$on('$routeChangeSuccess', function(evt, current, previous) {
    // Can't really use products_promise here because I want the resolved value
  });
}])
```

gkalpak:
@petebacondarwin, @pkozlowski-opensource, are referring to controllers defined for the route (in which the resolved values **can** be injected).

@fredsa, is referring to injecting resolved dependencies to other controllers (e.g. a parent controller), in which case I think the route's `locals` is the only way.

**[Example of both][1]**

--
I supposed (although undocumented) `locals` is not supposed to be private or something, since there's no `$$` prefix. 

[1]: http://codepen.io/ExpertSystem/pen/pvYeOW?editors=001

gkalpak:
BTW, if we decide to document the ability to access resolved dependencies via `locals`, I think it shouldn't be related to `resolved` property, but a "stand-alone" note and an explanation is needed for when it makes sense to do so (e.g. in controllers not related to the route).

petebacondarwin:
I don't think that accessing resolves from an arbitrary controller is a good pattern to follow:

* If you are not inside a route's template then it makes no sense to try to access these resolves since you can't be sure that the promises have been resolved.

* If you are inside a template then it makes more sense to use a route controller, inject the resolves into this and then expose the values to the rest of the template for other controllers to access.

That being said, if you are trying to set up a global route change success handler, from inside a service, then I guess you might want to access the resolves. In this case you would indeed use `$route.current.locals` to access these values but this is a special case.

petebacondarwin:
So, agreeing with @gkalpak, I would rather document this in the `$routeChangeSuccess` event as this is the only place it is really relevant.

PR #11443, opened 2015-03-27
fix(ngSanitize): use the smallest match on searching end tags

### LINKED ISSUE #11442: ngSanitize removes HTML strings between script elements
ngSanitize removes HTML strings between scripts elements. If I have the following HTML strings, the expected result is 'ABC' but the actual is 'AC'. The B is gone.

```
A
<script>evil</script>
B
<script>evil</script>
C
```
Plunker
http://plnkr.co/lQcfIh

After finding the begin of the script tag, ngSanitize searches the string of the end tag, like ```</script>``` with the longest match, so it removes all of the HTML strings between script elements. I think we should use the smallest possible match.

I will send the PR for it later. Thanks in advance!

----
Angular Version(s) : 1.4.0-beta
Browsers and Operating System : Chrome 41 / Mac OS 10.8.5

memolog:
ngSanitize searches the string of the end tag with the longest match.
So it removes all of the HTML strings between script elements.
We should use the smallest possible match.

Closes #11442

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -350,7 +350,7 @@ function htmlParser(html, handler) {
 
     } else {
       // IE versions 9 and 10 do not understand the regex '[^]', so using a workaround with [\W\w].
-      html = html.replace(new RegExp("([\\W\\w]*)<\\s*\\/\\s*" + stack.last() + "[^>]*>", 'i'),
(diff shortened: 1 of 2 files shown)


petebacondarwin:
This should be fixed by #11790

lgalfaso:
Closing this in favor of #11790

PR #11445, opened 2015-03-27
fix(filterFilter): fix matching against `null`/`undefined`

### LINKED ISSUE #11432: filter Filter will convert null and undefined to strings
filterFilter will convert null and undefined to strings and use that to filter
This makes the default comparator unusable with arrays that have fields with null or undefined values.
This seems to be a regression - it works as expected in <=1.3.5

Here is plnkr
http://plnkr.co/edit/P67JlZv83ea8J5ZFw9Ce?p=preview

gkalpak:
Included fixes:
- Do not convert `null`/`undefined` to strings for substring matching in
  non-strict comparison mode. Prevents `null`/`undefined` from being
  matched against e.g. 'u'.
- Let `null` (as a top-level filter expression) match "deeply" (as do
  booleans, numbers and strings).
  E.g. let `filterFilter(arr, null)` match an item like `{someProp: null}`.

Closes #11432
## 

<sub>1st and 3rd commits are refactorings (not important for the fix).
The 2nd commit contains the actual fix(es).
I kept them separate, for easier reviewing or eclectic merging.I can squash them if you think it is better.
</sub>

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -135,7 +135,7 @@ function filterFilter() {
       }
     }
 
-    var expressionType = (expression !== null) ? typeof expression : 'null';
+    var expressionType = getTypeForFilter(expression);
     var predicateFn;
     var matchAgainstAnyProp;
 
@@ -204,8 +204,8 @@ function createPredicateFn(expression, comparator, matchAgainstAnyProp) {
 }
 
 function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatchWholeObject) {
-  var actualType = (actual !== null) ? typeof actual : 'null';
-  var expectedType = (expected !== null) ? typeof expected : 'null';
+  var actualType = getTypeForFilter(actual);
+  var expectedType = getTypeForFilter(expected);
 
   if ((expectedType === 'string') && (expected.charAt(0) === '!')) {
     return !deepCompare(actual, expected.substring(1), comparator, matchAgainstAnyProp);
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@gkalpak it looks good to me after taking a quick look, but one of the commits upsets TravisCI (jsce). Could you please have a look?

gkalpak:
@pkozlowski-opensource: I fixed the error and tests are green now.
Shall I merge then (hopefully doing it the right way this time :smile:) ?

pkozlowski-opensource:
@gkalpak yeh, LGTM

Narretz:
One question, are these regression fixes or is this something that never worked correctly before?

gkalpak:
@Narretz, they are regression fixes:

> Do not convert `null`/`undefined` to strings for substring matching in non-strict comparison mode. 
   Prevents `null`/`undefined` from being matched against e.g. 'u'.

Broken since v1.3.6 / v1.4.0-beta.1.

> Let `null` (as a top-level filter expression) match "deeply" (as do booleans, numbers and strings).
   E.g. let `filterFilter(arr, null)` match an item like `{someProp: null}`.

Broken since v1.3.15 / v1.4.0-beta.6.

gkalpak:
@pkozlowski-opensource / @petebacondarwin, can you please take a look to make sure I got this right this time :smile:

pkozlowski-opensource:
@gkalpak commits look fine and TravisCI is green so things look in order :-)

Issue #11372, opened 2015-03-19
Error using ngOptions with objects containing circular references


dubejf:
The new `ngOptions` in Angular `1.4-beta` doesn't handle objects containing circular references well.

Steps to reproduce:
1. Define a `ng-options` with `ng-model`. At least one of the items in the source array must be a recursive object (such as a Breeze entity).
2. Select the recursive object.
3. *Uncaught RangeError: Maximum call stack size exceeded* is printed on the console.

http://jsfiddle.net/dubejf/tn102eh8/    (using `1.4.0-beta6`)

The watcher of `ngModel`, set by the `ngOptionsDirective`, compares the selected item with the last value by deep equality (using `angular.equals`). When a recursive object is selected, this causes the stack to blow up (as described in #7724).
 
Using `track by` doesn't alleviate the problem. 

This error is not reproducible in Angular 1.3.

mgbee8:
I can confirm and reproduce this error with Breeze entities, it was working correctly in beta5 but beta6 displays this behavior

dubejf:
You're right, it was broken by this commit: 6a03ca274314352052c3082163367a146bb11c2d

booleanbetrayal:
#11448 offers a way to work around this issue, provided you aren't trying to use `track by`

petebacondarwin:
Having landed #11448 I can see that this bug appears to be fixed: http://jsfiddle.net/L9vqc9ew/

dubejf:
Yes, this bug is fixed in `1.4.0-rc.0`, as long as you are not using `track by`. `track by` with cyclic objects still gives an error, but that regression seems less important.

In any case, with the recent performance optimizations to `ngOptions` (identity comparison, no copies), is it fair to say to `track by` should be *avoided* when using objects with `ngOptions`? As opposed to `ngRepeat`, where the guideline appears to be to use `track by` whenever possible.

pranaydutta89:
:+1:

petebacondarwin:
I think this is a better fix #11743

booleanbetrayal:
@petebacondarwin - I think the select documentation also needs to be updated as well - https://github.com/angular/angular.js/blob/abf59c285c3ff6af20dbf4236eba5204ae735abb/src/ng/directive/select.js#L131-L133

petebacondarwin:
Yes that block is not valid in that directive anyway, since it references `ngOptions`

petebacondarwin:
[...]

PR #11448, opened 2015-03-27
perf(ngOptions): only perform deep equality check on ngModel if using track by

### LINKED ISSUE #11447: ngSelect + ngOptions performs deep equality check unnecessarily in certain scenarios
Looks to be a result of #10893 which was first reported in #10869. Basically, the deep equality check should be avoided if operating in a context where object references are maintained.

booleanbetrayal:
Closes #11447

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -32,8 +32,9 @@ var ngOptionsMinErr = minErr('ngOptions');
  * option. See example below for demonstration.
  *
  * <div class="alert alert-warning">
- * **Note:** `ngModel` compares by reference, not value. This is important when binding to an
- * array of objects. See an example [in this jsfiddle](http://jsfiddle.net/qWzTb/).
+ * **Note:** By default, `ngModel` compares by reference, not value. This is important when binding to an
+ * array of objects. See an example [in this jsfiddle](http://jsfiddle.net/qWzTb/). When using `track by`
+ * in an `ngOptions` expression, however, deep equality checks will be performed.
  * </div>
  *
  * ## `select` **`as`**
@@ -275,6 +276,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
     }
 
     return {
+      trackBy: trackBy,
       getWatchables: $parse(valuesFn, function(values) {
         // Create a collection of things that we would like to watch (watchedArray)
         // so that they can all be watched using a single $watchCollection
@@ -500,8 +502,9 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
 
       // We also need to watch to see if the internals of the model changes, since
       // ngModel only watches for object identity change
-      scope.$watch(attr.ngModel, function() { ngModelCtrl.$render(); }, true);
-
+      if (ngOptions.trackBy) {
+        scope.$watch(attr.ngModel, function() { ngModelCtrl.$render(); }, true);
+      }
       // ------------------------------------------------------------------ //
 
 
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -135,8 +135,9 @@ var SelectController =
  * option. See example below for demonstration.
  *
  * <div class="alert alert-warning">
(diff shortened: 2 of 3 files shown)


test/ng/directive/ngOptionsSpec.js
```diff
@@ -1657,7 +1701,7 @@ describe('ngOptions', function() {
         scope.values.pop();
       });
 
-      expect(element.val()).toEqual('');
+      expect(element.val()).toEqualUnknownValue();
```
booleanbetrayal: Wish I had a better understanding of `unknownValue` usage in select + options. Had to revert the spec back to its form in ef894c87eaead76d90169113ab6acc9287654ea3.

petebacondarwin:
I'll take a better look over the weekend but I think this is good.

booleanbetrayal:
:+1:

src/ng/directive/ngOptions.js
```diff
@@ -32,8 +32,9 @@ var ngOptionsMinErr = minErr('ngOptions');
  * option. See example below for demonstration.
  *
  * <div class="alert alert-warning">
- * **Note:** `ngModel` compares by reference, not value. This is important when binding to an
- * array of objects. See an example [in this jsfiddle](http://jsfiddle.net/qWzTb/).
+ * **Note:** By default, `ngModel` compares by reference, not value. This is important when binding to an
+ * array of objects. See an example [in this jsfiddle](http://jsfiddle.net/qWzTb/). When using `track by`
+ * in an `ngOptions` expression, however, equality checks will be performed.
```
gkalpak: I might add _deep_ (equality) or something, to make it more clear.
(Technically, even in the default case, it is checking for equality (by reference).)

src/ng/directive/select.js
```diff
@@ -135,8 +135,9 @@ var SelectController =
  * option. See example below for demonstration.
  *
  * <div class="alert alert-warning">
- * **Note:** `ngModel` compares by reference, not value. This is important when binding to an
- * array of objects. See an example [in this jsfiddle](http://jsfiddle.net/qWzTb/).
+ * **Note:** By default, `ngModel` compares by reference, not value. This is important when binding to an
+ * array of objects. See an example [in this jsfiddle](http://jsfiddle.net/qWzTb/). When using `track by`
+ * in an `ngOptions` expression, however, equality checks will be performed.
```
gkalpak: Same as above.

Narretz:
There's a also regression with circular references in models which was introduced by https://github.com/angular/angular.js/commit/6a03ca274314352052c3082163367a146bb11c2d here: https://github.com/angular/angular.js/issues/11372

src/ng/directive/ngOptions.js
booleanbetrayal: good point! updated in rebased commit below.

src/ng/directive/select.js
booleanbetrayal: Same as above.  =]

barillax:
:+1:

nbrustein:
:+1:

test/ng/directive/ngOptionsSpec.js
petebacondarwin: OK, so the reason for this is that, if the model is `null` then `element.val()` is empty but if the model contains a value that doesn't match any option then `element.val()` is `'?'`.

When we remove the option the value becomes invalid so we get `'?'` but then the model is cleared out and becomes `null`. If we have a watch on the model then this will cause the `ngModel` to run `$render`, which update the element value to `''`.

I think that what we need is to trigger a new call to `$render` when the selected option is removed...

petebacondarwin: Actually what we need to do is fix the bit of code inside `updateOptions` (https://github.com/angular/angular.js/pull/11448/files#diff-0a9e86a58b2663074f51454f7a18a0efR646) that checks if the model has changed

petebacondarwin:
Thanks @booleanbetrayal for this

booleanbetrayal:
np ... thanks @petebacondarwin !

PR #11455, opened 2015-03-29
fix($parse): fix parse errors on older Android WebViews which choke with reserved keywords

gabrielmaldi:
Older Android WebViews (2.3) can't handle calling methods named after reserved keywords such as `.if()`, `.return()`, `.default()`, `.catch()`, `.finally()`.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -784,7 +784,7 @@ ASTCompiler.prototype = {
       self.state.computing = fnKey;
       var intoId = self.nextId();
       self.recurse(watch, intoId);
-      self.return(intoId);
+      self['return'](intoId);
       self.state.inputs.push(fnKey);
       watch.watchId = key;
     });
@@ -871,7 +871,7 @@ ASTCompiler.prototype = {
     recursionFn = recursionFn || noop;
     if (!skipWatchIdCheck && isDefined(ast.watchId)) {
       intoId = intoId || this.nextId();
-      this.if('i',
+      this['if']('i',
         this.lazyAssign(intoId, this.computedMember('i', ast.watchId)),
         this.lazyRecurse(ast, intoId, nameId, recursionFn, create, true)
       );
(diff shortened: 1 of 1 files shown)


lgalfaso:
Hi, are you talking about WebView 2.3 (this would be a version I am not aware of), or Android 2.3?

gabrielmaldi:
> Hi, are you talking about WebView 2.3 (this would be a version I am not aware of), or Android 2.3?

I mean WebViews on Android 2.3, which are the same as the Android Stock Browser.

This is quite common behavior on legacy Android and very similar issues have already been fixed, e.g. #11051

lgalfaso:
@petebacondarwin do we officially support Android 2.3?

Note: This does not mean that this will/will not be merged, and it is just to know what the expectations will be

petebacondarwin:
@lgalfaso at the moment we do not automatically test Angular 1 on any mobile devices. The most recent release of Android 2.3 was in 2011. I am not sure what our support is on this. Let's discuss it at the next team meeting. It would be helpful if anyone could point us in the direction of some usage statistics for Android versions...

gabrielmaldi:
According to [official Google Play data](https://developer.android.com/about/dashboards/index.html#Platform) Android 2.3 still holds `~7%`. And by my own experience, that number dramatically increases in emerging markets (I have apps published with over `20%` installs on 2.3 devices).

Believe me, I'd **love** to drop support for Android 2.3 in all my apps, it's a huge pain because it's really crappy and forces you to write workarounds for almost everything (e.g. doesn't support `overflow: scroll;`!). But I don't see myself being able to do this anytime soon because of market share.

My two cents: *supporting* surely means a lot more for the Angular team than just "having Angular be parsed correctly" :laughing: so perhaps you come to the conclusion to not officially support it like with IE8, but some small changes like the one on this PR can go a long way and not be a hassle to maintain.

Thanks guys!

lgalfaso:
Given that all of these are private methods, I just renamed them to `return_` and `if_`, and landed this as c55a494433e619aad0c7ef9fddadc0b3fdf53915

gabrielmaldi:
Great, thanks!

For posterity, I think you copied the wrong SHA: 10ae33b

PR #11459, opened 2015-03-30
docs(tutorial/Tutorial): added npm init step

mircare:
I lost some minutes to understand why there were errors launching npm install


Mirk01:
done

Mirko

pkozlowski-opensource:
@Mirk01 thnx for this pull request but the proposed change doesn't sound right so we will pass on this one.

PR #11460, opened 2015-03-30
Made for loops more consistent

James-Dunn:
Some for loops were storing the obj.length in a variable, some weren't. I converted them to all cache the length.


HichemBenChaaben:
i up vote on this if the line 678 does have the var ii declared first it won't break the build ;)

PR #11463, opened 2015-03-30
docs(error/infdig): Boolean directives taking toggle values by mistake can generate this error

mansabbloch:
One common mistake is passing a toggle value to a directive which takes boolian.


gkalpak:
_boolian_ --> _boolean_

Besides that, I don't think it is that common an error...
I am not sure this notice helps more than it confuses (e.g. what are directives that _take a boolean_).

BTW, the example works fine. Did you mean `showMyDialog=!showMyDialog` ?

Narretz:
I agree with @gkalpak, this is not that common. Stackoverflow questions with lots of upvotes or similar might make me change my mind, though :)

PR #11471, opened 2015-03-31
fix(ngTouch): register touches properly when jQuery is used

### LINKED ISSUE #4001: angular-touch with jQuery fails to "fastclick"
If jQuery is loaded before AngularJS it uses it and fails to get proper touches/changedTouches properties from TouchEvent in 

angular-touch.js:1.2-RC2
353:element.on('touchstart', function(event) {
380:element.on('touchend', function(event) {

adding
event = event.originalEvent;
solves the issue

mgol:
If jQuery was used with Angular the touch logic was looking for touches
under the original event object. However, jQuery wraps all events, keeping
the original one under the originalEvent property and copies/normalizes some
of event properties. Not all properties are copied, e.g. touches which caused
them to not be recognized properly.

Thanks to @mcmar & @pomerantsev for original patch ideas.

Fixes #4001
Closes #8584
Closes #10797

### DIFF
--- a/src/ngScenario/browserTrigger.js
+++ b/src/ngScenario/browserTrigger.js
@@ -77,6 +77,8 @@
           evnt.initAnimationEvent(eventType, null, null, null, eventData.elapsedTime || 0);
         }
       }
+    } else if (/touch/.test(eventType) && supportsTouchEvents()) {
+      evnt = createTouchEvent(element, eventType, x, y);
     } else {
       evnt = document.createEvent('MouseEvents');
       x = x || 0;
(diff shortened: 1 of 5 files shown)


mzgol:
The problem with tests for this PR is that without those changes they still pass in currently tested desktop browsers. If you run them in e.g. mobile Safari, they'd fail (with my PR they pass). I didn't find a good way to make them fail on desktop browsers as well, can we make the test run on at least one mobile browser? At lest unit tests?

mzgol:
cc @petebacondarwin @lgalfaso

src/ngScenario/browserTrigger.js
```diff
@@ -112,4 +114,37 @@
...
+    supportsTouchEvents._cached = true;
+    return true;
+  }
+
+  function createTouchEvent(element, eventType, x, y) {
+    var evnt = document.createEvent('TouchEvent');
+    x = x || 0;
+    y = y || 0;
+
+    var touch = document.createTouch(window, element, new Date().getTime(), x, y, x, y);
```
LeonardoBraga: I would change to Date.now(), as I've seen in other places of Angular's codebase.

pomerantsev:
This is great.

mzgol:
I'll need to merge #11479 before this one to be able to run the tests on mobile Safari.

petebacondarwin:
Landed as 06a9f0a95f0e72fa2e9879fe8a49e9bf69986a5f

Issue #10533, opened 2014-12-19
using ngResource when property name starts with @


ronaldjeremy:
I have objects that have properties that begin with @.  Example "@itemID" (this from the BadgerFish convention of turning an XML document into JSON, so i need to stick with this namaing convention)

I cannot figure out how to get a configuration object that ngResource will accept.  For example i have tried:
$scope.menuResource = $resource(url + "/restapi/items/:itemID", {itemID: "@@itemID"});
and Angular will throw an error:badmember exception.

How can I escape the @ in the property name to get it to work with ngResource?

gkalpak:
You can submit a [Pull Request][1] (including the fix and appropriate test-coverage).
For more details, take a look at [CONTRIBUTING.md#submitting-a-pull-request][2].

[1]: https://help.github.com/articles/creating-a-pull-request/
[2]: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#submitting-a-pull-request

ronaldjeremy:
@gkalpak I am following your instructions on Pull Requests and it says i need to create a new branch -- but it does not appear i am allowed to create a new branch at https://github.com/angular/bower-angular-resource

It it something you have permission to do?

gkalpak:
You need to fork the repo (the main `angular.js`, not `bower-angular-xyz`), create a branch on your fork, make the changes and submit a pull request against the original (forked) repo.

The process is discribed in more detail [here][1].

[1]: https://docs.angularjs.org/misc/contribute

mustela:
Just pushed a [PR](https://github.com/angular/angular.js/pull/11473) for this one.

PR #11473, opened 2015-03-31
fix($resource) add @ support for properties names

mustela:
Add support for properties that starts with @. This is useful when working with BadgerFish convention.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -5,7 +5,7 @@ var $resourceMinErr = angular.$$minErr('$resource');
 // Helper functions and regex to lookup a dotted path on an object
 // stopping at undefined/null.  The path must be composed of ASCII
 // identifiers (just like $parse)
-var MEMBER_NAME_REGEX = /^(\.[a-zA-Z_$][0-9a-zA-Z_$]*)+$/;
+var MEMBER_NAME_REGEX = /^(\.[a-zA-Z_$@][0-9a-zA-Z_$@]*)+$/;
 
 function isValidDottedPath(path) {
   return (path != null && path !== '' && path !== 'hasOwnProperty' &&
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -44,6 +44,7 @@ describe("resource", function() {
       expect(isValidDottedPath('1abc')).toBe(false);
       expect(isValidDottedPath('.')).toBe(false);
       expect(isValidDottedPath('$')).toBe(true);
+      expect(isValidDottedPath('@')).toBe(true);
       expect(isValidDottedPath('a')).toBe(true);
       expect(isValidDottedPath('A')).toBe(true);
       expect(isValidDottedPath('a1')).toBe(true);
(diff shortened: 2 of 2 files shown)


gkalpak:
@mustela, thx for the PR. It generally looks good to me (with a couple of concenrs; see below).
It would be nice to change the commit message though, adding `Closes #10533` at the end, in order to automatically close that related issue when/if this is merged.

I wonder:

- Is there any particular reason why the original `MEMBER_NAME_REGEX` only allowed `a-zA-Z_$` ?
  (I.e. did someone knew something we don't ?)

- Since we are allowing a pahth segment to start with `@` why not allow it to contain `@` ?
  I think it is better to allow `@` anywhere in the path (if it doesn't hurt at the beginning, I doubt it will hurt anywhere else).

mustela:
@gkalpak those are valid comments, and maybe the patter come from the emac standard (?), take a look to [this section](http://www.ecma-international.org/ecma-262/5.1/#sec-7.6). 

```
This standard specifies specific character additions: The dollar sign ($) and the underscore (_) are permitted anywhere in an IdentifierName.
```

Wondering if this decision really come from that and how strict is angular about it. If we *must* follow the standard, then this is no longer a bug :)

Also about the commit message. I didn't see nothing related to use the "Closes #xxx" sintaxis to automatically close issues. So should it be placed in the `header` or the `body`? 

Thanks

gkalpak:
The section you link to (7.6) is about _Identifiers_. But this is not the same as object _PropertyNames_ (e.g. take a look at [section 11.1.5][1]). A _PropertyName_ can be an _IdentifierName_ or a **_StringLiteral_** or a _NumericLiteral_.

What this means is:

```js
var _test;   // OK, because Identifiers are allowed to start with `_`
var @test;   // NOT OK, because Identifiers are not allowed to start with `@`

var obj = {_test: null};   // OK, because PropertyName can be an Identifier and
                           // Identifiers are allowed to start with `_`
var obj = {@test: null};   // NOT OK, because PropertyName can be an Identifier, but
                           // Identifiers are NOT allowed to start with `@`

var obj = {'@test': null};   // OK, because PropertyName can be a StringLiteral and
                             // '@test' is a valid StringLiteral
```

So, from that perspective I think it is pretty safe to allow `@` (both at the beginning and anywhere in the property name).

--
Regarding the `Closes #xxx`, is should go in the footer (at the very bottom of the commit message).
For more info you can take a look at [CONTRIBUTING.md#footer][2] and [this doc][3].

[1]: http://www.ecma-international.org/ecma-262/5.1/#sec-11.1.5
[2]: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#footer
[3]: https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#heading=h.j8e4paqkfz0q

mustela:
@gkalpak you are right, sorry to point the wrong link.

I update the branch with the changes. Let me know if you want me to do anything else. 

Thanks!

gkalpak:
Thx @mustela !

It generally LGTM...buuuut (since you ask for it), I would feel safer having a test that actually uses a `@prop`...

mustela:
@gkalpak your wish is my command! Let me know if thats ok. 

Thanks!

gkalpak:
:+1: @geni...ehrr...@mustela. This looks great !
<sub>
Now we just sit back and wait for someone from the team to give the green light.
(Actually, you sit back; I go pull some strings :P)
</sub>

mustela:
Thanks @gkalpak!

PR #11485, opened 2015-04-02
Update Croatian first day of the week

toncid:
It's strange to see Sunday as the first day of the week. This patch changes it to Monday.


toncid:
@googlebot I have just acquired a CLA.

petebacondarwin:
@toncid thanks for this PR but we actually generate these files automatically from the Google Closure library files. See https://github.com/angular/angular.js/tree/master/i18n

It may be that we need to update these libraries. But in the meantime we are not landing manual changes to these files as they will just be overwritten the next time we update.

PR #11497, opened 2015-04-02
docs(guide/Unit Testing): describe your change...

monir:
The test says "descending" order but it actually expects "ascending" order ( jeff -> jack -> igor).


gkalpak:
But `jeff -> jack -> igor` **IS** descending order.

Narretz:
@gkalpak I agree

PR #11502, opened 2015-04-03
fix(ngAnimate): use negative delay values to drive animation blocking

### LINKED ISSUE #11481: ngAnimate blocking causes issues with interrupted CSS transitions
This is a placeholder for a fixed issue in 1.4.

matsko:
DO NOT MERGE YET. IT STILL NEEDS MORE LIVE TESTING USING ANYTHING < 1.4.

Applies to 1.2 and 1.3.

when CSS transitions are used via ngAnimate a pre-emptive "block" is
applied to the element (via `transition: 0s none`) in order to gaurantee
that the `.ng-enter`, `.ng-leave` and/or `.ng-move` CSS classes are
applied instantly. By setting `0s none` we gaurantee that the styles
present within these classes actually appear on the element. While this
works it causes a major issue with CSS classes since they disupt an ongoing
transition since the `0s none` value is destructive.

This fix makes use of negative delay values in transitions to handle
this "blocking" mechanism yet it doesn't trigger the problems that setting

Closes #11481

### DIFF
--- a/src/ngAnimate/animate.js
+++ b/src/ngAnimate/animate.js
@@ -177,49 +177,6 @@
  * **version 1.3**, this workaround has been removed with ngAnimate and all non-ngAnimate CSS
  * class transitions are compatible with ngAnimate.
  *
- * There is, however, one special case when dealing with class-based transitions in ngAnimate.
- * When rendering class-based transitions that make use of the setup and active CSS classes
- * (e.g. `.fade-add` and `.fade-add-active` for when `.fade` is added) be sure to define
- * the transition value **on the active CSS class** and not the setup class.
- *
- * ```css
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@matsko what do we want to do about this PR?

petebacondarwin:
I doubt that we are going to get round to fixing this, especially in 1.3.x, since it is not too difficult to migrate to 1.4.x to get the fix.

Does anyone have a production requirement for this fix?

Narretz:
I'm going out on a limb here and say that this is not going to land in 1.2 or 1.3 Please reopen if I'm wrong!

matsko:
Sounds good to me.

PR #11507, opened 2015-04-03
fix @param tag for $provide.decorator

KevinBrogan:
The $provide.decorator function, as per the documentation, "is called using the auto.injector.invoke method and is therefore fully injectable."

The current @param contradicts this by stating that only Functions may be used as an argument.


KevinBrogan:
Signed the CLA

PR #11508, opened 2015-04-03
fix(jqLite): Check "length" in obj in isArrayLike for iOS JIT bug

snapwich:
to prevent iOS8 JIT bug (https://bugs.webkit.org/show_bug.cgi?id=142792) from surfacing

This is the fix implemented in [underscore.js](https://github.com/jridgewell/underscore/commit/086a56d04c0214054219d83eff25a4cbba524fc2) and most likely the fix that will be implemented in [jQuery](https://github.com/jquery/jquery/issues/2145#issuecomment-88650887).

It's not a very easy to recreate bug, but I stumbled upon it myself in our production app and was able to narrow it down to isArrayLike after several hours of banging my head.  This plunkr was the closest I got to consistently surfacing the bug in Angular: http://plnkr.co/edit/UQ5NhZnRuafNxWiJuuyN

That ng-repeat will fail, but only on iOS8 (and i've personally only had it fail on iPad) about 50% of the time after a few digests.  If it doesn't fail quick reloading the plunkr a few times will usually get it to fail.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -200,7 +200,9 @@ function isArrayLike(obj) {
     return false;
   }
 
-  var length = obj.length;
+  // Support: iOS 8.2 (not reproducible in simulator)
+  // "length" in obj used to prevent JIT error (gh-11508)
+  var length = "length" in Object(obj) && obj.length;
 
   if (obj.nodeType === NODE_TYPE_ELEMENT && length) {
     return true;


snapwich:
Updated pull request to follow updated isArrayLike solution in jQuery https://github.com/jquery/jquery/pull/2185

Unfortunately no tests since this is iOS8 only and doesn't surface in iOS simulator.

realityking:
@snapwich Could you add a comment explaining why it's needed? Otherwise someone will probably try to remove it in a couple of months or years ;)

snapwich:
@realityking sure.  I just added a comment similar to the jQuery commit.

snapwich:
Any update on merging this?  This fix has made its way into jQuery now and would be a quick fix for Angular.

caitp:
LGTM

jdalton:
jQuery found that based on their usage they didn't need the `Object(...)` use. Might be worth checking out here too.

caitp:
@jdalton that works in jQuery's case because the caller knows that it's a non-null "object" already, I'm not sure the same assumptions fly here. It would be good to make those assumptions work but I won't consider that a blocker

snapwich:
@jdalton @caitp yeah jQuery and Angular make some different assumptions I think.

Removing the Object(...) causes these unit tests to fail that otherwise don't:

    angular forEach should handle string values like arrays FAILED
    angular forEach ES spec api compliance should follow the ES spec when called with string FAILED

petebacondarwin:
For reference https://github.com/jquery/jquery/issues/2145

petebacondarwin:
@mzgol - you worked on this bug in JQuery - what is your recommendation here?

mzgol:
We should work around it; relevant jQuery commit: https://github.com/jquery/jquery/commit/154166458284bcce7d6a86328b7fd13483232a1a

mzgol:
This looks good but in jQuery it turned out some people relied on undocumented behavior that jQuery.map worked on strings and not just arrays (this broke a popular DataTables jQuery plugin). Not sure if this is an issue here.

petebacondarwin:
OK, LGTM. Let's merge this.

PR #11509, opened 2015-04-04
docs(ngSwitch): Replace tt tag with code tag

rparrapy:
Use of tt is discouraged, see:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt
http://www.w3.org/wiki/HTML/Elements/tt


rparrapy:
I signed it!

Issue #11470, opened 2015-03-31
Dynamic select directive doesn't support empty value


slavede:
Branch 1.2.X issue.

In case I have select directive which get options generated from server, it throws following error:

        TypeError: 'undefined' is not an object (evaluating 'emptyOption.prop')
            at /angularjs/angular-1.2.28.js:21717
            at ngModelWatch (/angularjs/angular-1.2.28.js:17899)
            at /angularjs/angular-1.2.28.js:12642
            at /angularjs/angular-1.2.28.js:12915

 in case I'm trying to update model in my test (it works perfectly fine when it comes to browser, but it can't update itself based on model change) and here's why. This piece of code executes when linking select initially:

      for (var i = 0, children = element.children(), ii = children.length; i < ii; i++) {
        if (children[i].value === '') {
          emptyOption = nullOption = children.eq(i);
          break;
        }
      }

and in this moment of course, my select doesn't have emptyOption. Later on, I populate options and try to update model in my test. Then this piece of code is triggered:

          var viewValue = ngModelCtrl.$viewValue;

          if (selectCtrl.hasOption(viewValue)) {
            if (unknownOption.parent()) unknownOption.remove();
            selectElement.val(viewValue);
            if (viewValue === '') emptyOption.prop('selected', true); // to make IE9 happy
          } else {
            if (isUndefined(viewValue) && emptyOption) {
              selectElement.val('');
            } else {
              selectCtrl.renderUnknownOption(viewValue);
            }
          }

Now, in this moment emptyOption is undefined, but I really do have it, only thing, it was generated later (after link function).

Can we(I can do it if you let me) have additional element.find when it comes to assigning to emptyOption?

petebacondarwin:
So your issues is that the `select` directive does not allow you to specify the `empty` option asynchronously. Correct?

slavede:
Yes, that's right...

petebacondarwin:
So the design of this directive is that the empty option should be hard-coded - even if the rest of the options are loaded from the server. Could you try doing that as a workaround?

slavede:
Unfortunately no, because even display for this empty value should has to come from server. I have simple ng-repeat that iterates options I get from server and display can be different depending on place this directive is used.

slavede:
If you agree, we can do in this line additional check if emptyOption is not defined, try to search for it:

    if (viewValue === '') emptyOption.prop('selected', true); // to make IE9 happy

I would search it in the way initial search is done:

[...]

PR #11512, opened 2015-04-04
Dynamic option can hold empty value now so you don't have to initially p...

slavede:
...ut it inside select (issue #11470)

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -86,7 +86,7 @@ var SelectController =
   self.addOption = function(value, element) {
     assertNotHasOwnProperty(value, '"option value"');
     if (value === '') {
-      self.emptyOption = element
+      self.emptyOption = element;
     }
     var count = optionsMap.get(value) || 0;
     optionsMap.put(value, count + 1);
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -83,8 +83,11 @@ var SelectController =
 
 
   // Tell the select control that an option, with the given value, has been added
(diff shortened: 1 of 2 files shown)


slavede:
I signed it!

petebacondarwin:
@slavede - thanks for this PR. It looks good for the scenario of adding in a dynamic empty element. Isn't it great when a fix can be so simple.

I have two concerns that we need to resolve:

* what happens if there was already a static empty option when a dynamic empty option appears?
* what happens when a dynamic empty element is removed?
  * in the case that there was already a static empty option
  * in the case that there was no empty option already

slavede:
Sorry, I was/am kinda busy this week. Let me think about this over the weekend and come up with some suggestion

petebacondarwin:
Great! Here is the link you need https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit

slavede:
1) that really shouldn't matter because self.emptyOptions is only used in checking if ngmodel is set to unknown value so we can set it to empty:

    if (isUndefined(value) && self.emptyOption) {
        self.removeUnknownOption();
        $element.val('');
      } else {
        self.renderUnknownOption(value);
      }
2) tbh, currently self.emptyOption is not even checked in removeOption and if you now remove empty option, self.emptyOption would remain set so that's I believe another bug we could open/fix

slavede:
@petebacondarwin what do you think?

petebacondarwin:
@slavede thanks for this. I tweaked the tests, added the bit about removing the empty option and then squashed the commits.

PR #11520, opened 2015-04-06
fix($compile): do not use jQuery class API

dtritus:
jQuery class API causes issues with SVG elements

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -964,9 +964,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
   this.$get = [
             '$injector', '$interpolate', '$exceptionHandler', '$templateRequest', '$parse',
-            '$controller', '$rootScope', '$document', '$sce', '$animate', '$$sanitizeUri',
+            '$controller', '$rootScope', '$document', '$sce', '$animate', '$$sanitizeUri', '$$jqLite',
     function($injector,   $interpolate,   $exceptionHandler,   $templateRequest,   $parse,
-             $controller,   $rootScope,   $document,   $sce,   $animate,   $$sanitizeUri) {
+             $controller,   $rootScope,   $document,   $sce,   $animate,   $$sanitizeUri, $$jqLite) {
 
     var Attributes = function(element, attributesToCopy) {
       if (attributesToCopy) {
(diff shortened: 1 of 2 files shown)


Narretz:
Hi, thanks for the PR. We need additinal info though, before this can be considered for merge:
1. what kinds of issues are these?
2. there need to be tests for these issues

dtritus:
1) [Example](http://plnkr.co/edit/QlcVv5eNSuWiAzvyvsCS?p=preview)
In latest Chrome and Firefox you will see blue rectangle, in Safari 7.0.6 - red (works in Safari 7.1).

This happens because `safeAddClass` uses jQuery to add `ng-isolate-scope` class to SVG element.
Internally jQuery uses `element.className` to read class attribute, but in Safari `element.className` returns `SVGAnimatedString` object instead of string. This object converted to `[object SVGAnimatedString]` string and we lose existing classes on element.

2) Not sure how test these issues

dtritus:
I've rebased my changes on top of current master. Issue is still actual for Angular 1.4.x

#### Few notes about this issue
* It happens only on SVG elements
* It happens only with jQuery included
* It happens only in Safari 6.1/7.0 (in Safari 7.1+ works fine)

Example: http://plnkr.co/edit/9FCmyOXQcXq0hTQ5jDTp?p=preview
Screenshot: http://i.imgur.com/JH4pL1o.png
Rectangles and text should be blue.

#### Why it happens
jqLite implementation of `addClass`/`removeClass` uses `getAttribute`/`setAttribute` functions to add/remove `class` attribute. This function works correct for SVG elements.

But jQuery uses `className` property of element. Safari 6.1/7.0 returns `SVGAnimatedString` as value of `className` property for SVG elements. jQuery casts `SVGAnimatedString` to string and as result we see `[object SVGAnimatedString]`.

So, this issue isn't Angular issue itself, but it breaks existing code when user adds jQuery.

See #9292 for discussion about the same issue in ngAnimate.

lgalfaso:
@dtritus would it be possible for you to add a test?

dtritus:
@lgalfaso I've added a test

loueradun:
Here is a simple shim to get this working if someone stumbles across this later:
SVGAnimatedString.prototype.toString = function () { return this.baseVal; }

Narretz:
Since the Safari bug was fixed in 7.1, I don't think we need to add this. The version was released 2 years ago, and we usually only support the latest browser versions anyway.

markTreveil:
The SVGAnimatedString shim breaks the click handler in angular.js:
`if (isObject(absHref) && absHref.toString() === '[object SVGAnimatedString]') {`

PR #11524, opened 2015-04-07
fix(ngMock): Add responseType function parameter in mock httpbackend for passthrough

### LINKED ISSUE #5415: Mock httpBackend does not support blobs for passThrough.
I have tried to use the Mock httpbackend with passThrough for some audio files retrieved from a server.

It works fine when not using the Mock httpBackend, but when I switch to the Mock backend the data returned are not blobs, but strings. The data is there is seems because chrome freezes when debugging and watching the variable - like the data contained in the string is just to big for it to handle.

The way I use $http is like this:
```Javascript
$http({method: 'GET', url: url, responseType: 'blob'}).success(...);
```

prateek14:
The ngMock $httpBackend allows http request routes that have not been defined
for mocking to be passed through to the real http backend service. The
responseType parameter was missing in the ngMock httpbackend parameter
list.

closes #5415

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1129,7 +1129,7 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
   }
 
   // TODO(vojta): change params to: method, url, data, headers, callback
-  function $httpBackend(method, url, data, callback, headers, timeout, withCredentials) {
+  function $httpBackend(method, url, data, callback, headers, timeout, withCredentials, responseType) {
     var xhr = new MockXhr(),
         expectation = expectations[0],
         wasExpected = false;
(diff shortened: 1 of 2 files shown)


brphelps:
+1

RobertjanTuit:
+1

petebacondarwin:
This is not what ngMock $httpBackend was designed for - it is supposed to be for unit tests that should never touch a slow external service. Perhaps you wanted to use ngE2EMock instead?

felixfbecker:
When I use ngMockE2E and specify to passThrough() a request for some binary file (lets say, a jpeg image) and then request it somewhere in my code with $http responseType: 'blob' the response.data property is always a string, not a blob. When I use it without ngMockE2E it works as expected. It seems like ngMockE2E is swallowing the responseType when passing through a request.

PR #11534, opened 2015-04-09
docs(filter): Note about exact match in object

jpribesh:
Added note that when filtering for specific matches in javascript you MUST wrap the object key in quotes. Also provided example.

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -28,6 +28,8 @@
  *     the string with `!`.
  *     For example `{name: "!M"}` predicate will return an array of items which have property `name`
  *     not containing "M".
+ *     Note that if filtering for an exact match in javascript, you MUST wrap the object key in quotes.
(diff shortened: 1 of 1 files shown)


jpribesh:
I signed it!

gkalpak:
I don't think you have to ([proof][1]).
Have you seen this issue with a particular version ?

[1]: http://codepen.io/ExpertSystem/pen/BygNQN?editors=101

jpribesh:
I am seeing this issue with v1.2.28

jpribesh:
I just tested this on 1.3.15, and looks like the quotes aren't needed there. Perhaps there is some other issue causing this side effect for me.

jpribesh:
I was also referring to using the filter in the js file. Either way it seems to work without the quotes [Proof](http://codepen.io/anon/pen/JoQEKg) Ill go ahead and close this pr, and open a new one if I figure out what is causing it in my code.

PR #11544, opened 2015-04-10
docs($resource): Make a correction in the example. 

yilinglu:
Parameter object property name should match the extraction string


gkalpak:
This isn't right. The first argument to `get` is the `parameters` object (not a `data` object) and the keys must match the parameters names (in this case `userId`).

Thx for taking the time to look into it and submit a PR. Keep them coming :)

PR #11546, opened 2015-04-10
docs(tutorial/11 - REST and Custom Services): describe your change...

jpgrace:
For someone that is totally unfamiliar with AngularJS (or any code segment for that matter), it is not ever easier to understand what is happening when an abstraction is used.

I propose dropping this sentence entirely. Regardless, the phrase "controllers is doing" is grammatically incorrect. It should be "controllers are doing".


jpgrace:
I signed it!

Narretz:
I don't see how this is confusing. Controllers are an essential part of an angular app, and are introduced by the tutorial in previous steps. The grammar is also okay, imo (I'm not a native speaker though)

Issue #1934, opened 2013-02-01
Ability to react to progress events of $http XHR


ryankshaw:
I would like to be able to bind to the progress event of the xhr upload.

for example (in plain javascript):

var xhr = new XMLHttpRequest();
xhr.upload.addEventListener("progress", uploadProgress, false)

(so I can show a progress meter for uploaded files)

I want to be able to do the same thing by doing $http.post...

for this and any other case edge case where you need to interact with the raw XHR object itself it might be nice to provide something that lets you just get at the original xhr object.

ajwhite:
It would be nice if we could just use the promise's progress callback

From [kriskowal/q](https://github.com/kriskowal/q/wiki/API-Reference#promisethenonfulfilled-onrejected-onprogress) docs

#### promise.then(onFulfilled, onRejected, onProgress)
The `then` method from the [Promises/A+ specification](http://promises-aplus.github.com/promises-spec/) with an additional progress handler.

I see no reason why we can't use `.then` here and expect `onProgress` to be reported to

```js
$http({
  method: 'POST',
  data: jsonData
}).then(function () {
  // on succes
}, function () {
  // on error
}, function () {
  // on progress
});
```

Or better yet, using [promise.progress](https://github.com/kriskowal/q/wiki/API-Reference#promiseprogressonprogress) 
```js
$http.post(url, data).success(function () {
  // success
}).error(function () {
  // error
}).progress(function () {
  // progress
});
```

benjamingr:
@ajwhite I humbly disagree, it would be horrible to use the progress event, not only is it being removed from Kris Kowal's Q (check the v2 branch) and the Q authors are themselves against it - it will not be included in native promises (ever) and is a generally poor idea.

Please refer to the above discussion starting here:

https://github.com/angular/angular.js/issues/1934#issuecomment-41469790

ajwhite:
@benjamine please refresh, I noticed the depreciation notice in the docs & redacted

ajwhite:
@benjamine also caught up on your conversation. With the points being made towards the deprecation of all progress events (even in `.then(null, null, progress)`), it makes less sense to build support for it this way. With Bluebird & Q moving away from that support, a configuration property seems to make a bit more sense, and leave promises to their responsibility

benjamingr:
Glad we agree promise progress events are not the way forward :)

caitp:
[...]

PR #11547, opened 2015-04-10
feat($http): add XHR events configuration option

chrisirhc:
Allows for arbitrary event handling on requests.

There are no tests here, it's just a proposal for discussion of #1934 .
I'll add tests if a decision is made to go ahead with it.
This may also simplify testing by just firing events on the `mockXhr` object.

### DIFF
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -28,7 +28,7 @@ function $HttpBackendProvider() {
 
 function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDocument) {
   // TODO(vojta): fix the signature
-  return function(method, url, post, callback, headers, timeout, withCredentials, responseType) {
+  return function(method, url, post, callback, headers, timeout, withCredentials, responseType, eventHandlers) {
     $browser.$$incOutstandingRequestCount();
     url = url || $browser.url();
 
@@ -87,6 +87,20 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
 
       xhr.onerror = requestError;
       xhr.onabort = requestError;
+
+      if (eventHandlers) {
+        forEach(eventHandlers, function(value, key) {
+          if (key !== 'upload') {
+            xhr.addEventListener(key, value);
+          }
+        });
+
+        if (eventHandlers.upload) {
+          forEach(eventHandlers.upload, function(value, key) {
+            xhr.upload.addEventListener(key, value);
+          });
+        }
+      }
 
       if (withCredentials) {
         xhr.withCredentials = true;
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1683,6 +1683,17 @@ function MockXhr() {
   };
 
   this.abort = angular.noop;
+
+  this.$$events = {};
+  this.addEventListener = function(name, listener) {
+    if (angular.isUndefined(this.$$events[name])) this.$$events[name] = [];
+    this.$$events[name].push(listener);
+  };
+
+  this.upload = {
+    $$events: {},
+    addEventListener: this.addEventListener
+  };
 }
 
 
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1155,7 +1155,7 @@ function $HttpProvider() {
         }
 
         $httpBackend(config.method, url, reqData, done, reqHeaders, config.timeout,
-            config.withCredentials, config.responseType);
+            config.withCredentials, config.responseType, config.events);
       }
 
       return promise;
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -694,6 +694,8 @@ function $HttpProvider() {
      *    - **headers** – `{Object}` – Map of strings or functions which return strings representing
      *      HTTP headers to send to the server. If the return value of a function is null, the
      *      header will not be sent. Functions accept a config object as an argument.
+     *    - **events** - `{Object}` - Event listeners to be bound to the XMLHttpRequest object.
+     *      To bind events to the XMLHttpRequest upload object, nest it under the upload key.
      *    - **xsrfHeaderName** – `{string}` – Name of HTTP header to populate with the XSRF token.
      *    - **xsrfCookieName** – `{string}` – Name of cookie containing the XSRF token.
      *    - **transformRequest** –
--- a/test/ng/httpBackendSpec.js
+++ b/test/ng/httpBackendSpec.js
@@ -214,6 +214,19 @@ describe('$httpBackend', function() {
     expect(MockXhr.$$lastInstance.withCredentials).toBe(true);
   });
 
+  it('should set up event listeners', function() {
+    var progressFn = function() {};
+    var uploadProgressFn = function() {};
+    $backend('GET', '/url', null, callback, {}, null, null, null, {
+      progress: progressFn,
+      upload: {
+        progress: uploadProgressFn
+      }
+    });
+    xhr = MockXhr.$$lastInstance;
+    expect(xhr.$$events.progress[0]).toBe(progressFn);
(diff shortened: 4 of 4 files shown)


src/ng/httpBackend.js
```diff
@@ -88,6 +88,20 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
...
+
+        if (eventHandlers.upload) {
+          forEach(eventHandlers.upload, function (value, key) {
+            xhr.addEventListener(key, value);
+          });
+        }
+      }
+
       if (withCredentials) {
         xhr.withCredentials = true;
```
lovemyboo: H

pkozlowski-opensource:
@chrisirhc I don't think is right - be doing so we would allow people to override already added (by Angular itself) handlers, no? 

I know that this change would make things super-flexible but this doesn't seem to be in-line with the current API. 

Do you have any other use-cases (besides download / upload events) on your mind for this one?

chrisirhc:
@pkozlowski-opensource This doesn't override existing handlers, addEventListener always adds on existing listeners. In fact, this is safer than allowing users to do anything to the `onreadystatechange` attribute.

Use Case: I'm streaming a response and want to read the incomplete response as it is being streamed. Think of a timeseries chart that renders live as the data is coming in.

chrisirhc:
@pkozlowski-opensource Ping. I saw that perhaps you put this in Purgatory because you thought that it would allow people to override the handlers, but it won't. As for the being in line with the API, it would be interesting to design an API that's in line that adds such functionality.

petebacondarwin:
@pkozlowski-opensource:
I think that something like this might be quite nice, although I wonder if we would need to include `$apply` in there somewhere.

What are the downsides to this?

src/ng/httpBackend.js
```diff
@@ -88,6 +88,20 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
...
       xhr.onabort = requestError;
 
+      if (eventHandlers) {
+        forEach(eventHandlers, function(value, key) {
+          if (key !== 'upload') {
+            xhr.addEventListener(key, value);
+          }
+        });
+
+        if (eventHandlers.upload) {
```
gkalpak: Why put listeners inside `upload` if it's not going to make any difference ?

petebacondarwin: ?

The `eventHandlers` object has the following structure:

```
{
  someEvent: ...,
  someOtherEvent: ...,
  upload: {
    someUploadEvent: ...,
    someOtherUploadEvent: ...
  }
}
```

gkalpak: Yes, but we end up doing `xhr.addEventListener(key, value)` for all of them.
I.e. I could write you example as:

```js
{
  someUploadEvent: ...,
  someOtherUploadEvent: ...
  upload: {
    someEvent: ...,
    someOtherEvent: ...,
  }
}

// or

{
  someEvent: ...,
  someOtherEvent: ...,
  someUploadEvent: ...,
  someOtherUploadEvent: ...
}
```

and the result would be the same.

So, I am wondering if there is any benefit in separating the two. Why not keep it flat ?

chrisirhc: Sorry, this is a typo on my part. This is meant to attach events on `xhr.upload`. The line 100 should read `xhr.upload.addEventListener(key, value);`.

gkalpak: Makes sense now; I didn't know about `xhr.upload` :confused:

chrisirhc:
Updated PR to correct typo (`xhr.upload`).

petebacondarwin:
OK, so all we need now are some unit tests and docs then this is good to go.

chrisirhc:
- Added a unit test
- Added brief docs in $http

petebacondarwin:
LGTM - @pkozlowski-opensource ?

src/ngMock/angular-mocks.js
```diff
@@ -1683,6 +1683,17 @@ function MockXhr() {
...
+  this.$$events = {};
+  this.addEventListener = function(name, listener) {
+    if (angular.isUndefined(this.$$events[name])) this.$$events[name] = [];
+    this.$$events[name].push(listener);
+  };
+
+  this.upload = {
+    $$events: {},
+    addEventListener: this.addEventListener
+  };
```
gkalpak: Since we are mocking the addition of event listeners, would it make sense to also mock emitting some event (so that people can test their code once it relies on a callback to be called on event _X_) ?
Would it be possible in a clean and reliable way without a breaking change ?

chrisirhc: That makes sense and could be implemented like [MockWindow's fire method](https://github.com/angular/angular.js/blob/master/test/ng/browserSpecs.js#L41-L45) .
However, I don't want to put too many changes in this PR at the moment as it'll affect reviewing, and the chances of it getting merged.

mchambaud:
This is not in 1.5.0-RC.0?

petebacondarwin:
Sorry, @mchambaud, it didn't make the cut for 1.5 - but worry not, 1.6 will be coming in not too far behind 1.5, which should be released early after the New Year.

mchambaud:
@petebacondarwin Perhaps the milestone should be updated!

petebacondarwin:
Thanks @mchambaud - I have updated the milestone :-)

mchambaud:
:+1:

marcelrouw:
When will 1.6 be released?
Or can this item still be added to 1.5.1 release?

petebacondarwin:
All non-breaking changes will be included in 1.5.x

marcelrouw:
May be a stupid question. But is this a breaking change?

petebacondarwin:
I don't think it is a BC

petebacondarwin:
Let's get this into 1.5.x (probably 2 or 3)

buraktamturk:
This isn't merged yet... And I don't think it is gonna be. I've been waiting a such feature on Angular for years, yet it is 2016 and we still can't track upload with $http.

I found a **pretty good** workaround somewhere in the web, before this code, I were using XHR directly, using XHR directly has a few disadvantages: 

* I weren't able to refresh the token if 401 returns from the server. All of the token refreshing mechanism were in custom service $api wrapped around $http.
* the service $api was including token in the request, with using XHR directly, I had manually get the token and call setRequestHeader etc. There would two authentication mechanism and it is bad for a project structure. 
* it is just **ugly**.

It was good idea to keep $http / $api logic in my project. 

**here is an example**

```javascript
// XHR decorators were recently added to angular.

/* app */.decorator("$xhrFactory", function($delegate, $rootScope) {
    'ngInject';

    return function(method, url) {
        var xhr = $delegate(method, url);

        xhr.setRequestHeader = (function(sup) {
            return function(header, value) {
                if ((header === "__XHR__") && angular.isFunction(value))
                    value(this);
                else
                    sup.apply(this, arguments);
            };
        })(xhr.setRequestHeader);

        return xhr;
    };
});

// somewhere in the project

$http({
  method: 'PUT',
  url: `files/user/private/${file.name}`,
  data: 'the file object',
  headers: {
      'Content-Type': 'application/blablabla',

      __XHR__: function() {
          return function(xhr) {
              // here you can access the XHR
              xhr.upload.addEventListener("progress", function(event) {
                  upload_callback(event.loaded, event.total);
              });
          };
      }
  }
});
```
**It would be much better if we could track the upload without hooking setRequestHeader.** But i think it is the best solution to use something like until this PR get merged.

chrisirhc:
Is this blocked on me rebasing this PR / making it merge-able? If it is, I can look into it.

petebacondarwin:
I think we are OK, it just got lost in the huge pile of papers in the in-tray

buraktamturk:
@petebacondarwin Are there any update on this?

petebacondarwin:
I have a pull request here that should land soon https://github.com/angular/angular.js/pull/14367

PR #11561, opened 2015-04-13
fix($http): stop $httpBackend conversion of falsy values to null before xh...

samjacobclift:
...r.send()

stop $httpBackend from potentially overriding false values to null when sending a falsy value to a web service via POST or PUT et al


royka:
Will redo with my none work email

PR #11564, opened 2015-04-13
refactor(directives.js): restrict default AC -> AE

PatrickJS:
let's be honest, no one uses Class directives in Angular1 and this will be more inline with Angular2. I can also change it to 'AEC' for the defaults to be backwards compatible.

### DIFF
--- a/src/ng/directive/directives.js
+++ b/src/ng/directive/directives.js
@@ -6,6 +6,6 @@ function ngDirective(directive) {
       link: directive
     };
   }
-  directive.restrict = directive.restrict || 'AC';
+  directive.restrict = directive.restrict || 'AE';
   return valueFn(directive);
 }


lgalfaso:
Hi @gdi2290, thanks for the PR. This is a breaking change that can cause too many headaches, so I think it is best not to change this

Narretz:
The default has been changed to `AE` already some time in the past (that's also what the $compile docs say). I assume compile will overwrite this restriction defintion somwhere. Maybe we should clean this up, then.

gkalpak:
This is not about the global default. It is the default of the internal `ngDirective` function which is used for creating directives such as `ngCloak`, `ngInit`, `ngNonBindable`, `ngStyle`, `ngSwitchDefault/When` etc. 

For most of them `E` wouldn't make sense, so I think it is fine (although it would be fine for some of them).

PR #11567, opened 2015-04-13
stop coercing falsy values to null before xhr.send()

### LINKED ISSUE #11552: $httpBackend coersion of falsy values to null before xhr.send()
Can anyone explain why this line in $httpBackend should coerce falsy values to null?

    xhr.send(post || null);
https://github.com/angular/angular.js/blob/master/src/ng/httpBackend.js#L112

This means that, if I need to POST, PUT, etc. an actual boolean false to a web service, I have to work around AngularJS.

I'd like the line to read something like this:

    xhr.send(post);

or if it just can't be undefined or something:

    xhr.send(isDefined(post) ? post : null);

johnhoffman:
This is intended to resolve https://github.com/angular/angular.js/issues/11552.

### DIFF
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -109,7 +109,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
         }
       }
 
-      xhr.send(post || null);
+      xhr.send(post);
     }
 
     if (timeout > 0) {
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -109,7 +109,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
         }
       }
 
-      xhr.send(post);
+      xhr.send(isDefined(post) ? post : null);
     }
 
     if (timeout > 0) {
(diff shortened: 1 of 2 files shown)


johnhoffman:
I signed it!

johnhoffman:
I authored the commits.

royka:
I've had problems with the CLA too :D I've had a little play with this issue. I've added a test I think is valid to check the behaviour.

see: https://github.com/royka/angular.js/commit/89ec5af6425c624f6c2f641ac5718fa69ec3ee98

pkozlowski-opensource:
@johnhoffman your commit is missing tests so we can't merge it as-is. There also seems to be an issue with the CLA - I suspect you've used a different e-mail / GitHub user name to author the commit and different one to sign the CLA.

@royka would you mind opening a PR with your change to see if the existing test pass? For example - do we need to convert `undefined` to `null`?

royka:
@pkozlowski-opensource see here https://github.com/angular/angular.js/pull/11579 

unit tests pass

johnhoffman:
I fixed the CLA and updated the PR. Still unable to build angular and run tests on my machine (https://docs.angularjs.org/misc/contribute) so I guess we resort to Travis CI?

PR #11568, opened 2015-04-14
feat($animate): provide support for animations on elements outside of $rootElement

matsko:
By default, ngAnimate only considers animations for elements that live within `$rootElement`. This is to prevent the accidental animation of elements that fall outside of an application. This feature, however, is limited since it prevents animations from running on elements that are situated by the `body` tag while being outside of `ng-app`. Drop down menus for example use absolute positioning and are therefore situated directly inside of `document.body`. This fix allows for this to work by using `$animate.pin()` to flag elements that live outside of `$rootElement`.

--- commit ---

Beforehand it was impossible to issue an animation via $animate on an
element that is outside the realm of an Angular app. Take for example a
dropdown menu where the menu is positioned with absolute positioning...
The element will most likely need to be placed by the `<body>` tag, but
if the angular application is bootstrapped elsewhere then it cannot be
animated.

This fix provides support for `$animate.pin()` which allows for an
external element to be virtually placed in the DOM structure of a host
parent element within the DOM of an angular app.

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -72,6 +72,7 @@ var $$CoreAnimateQueueProvider = function() {
       enabled: noop,
       on: noop,
       off: noop,
+      pin: noop,
 
       push: function(element, event, options, domOperation) {
         domOperation        && domOperation();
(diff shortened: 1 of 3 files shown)


matsko:
LGTM (from Igor in person).

matsko:
Merged as https://github.com/angular/angular.js/commit/e41faaa2a155a42bcc66952497a6f33866878508

PR #11569, opened 2015-04-14
fix(ngAnimate): ensure that a filtered-out leave animation always runs its DOM operation

### LINKED ISSUE #11555: $animateProvider.classNameFilter() broken with angular 1.4.0 rc0
I'm using the following code to enable ng-animate only for certain elements:
```
.config(function($animateProvider) {
   $animateProvider.classNameFilter(/ng-animation/);
})
```

It was working fine with all the previous angular versions, but with the new rc0 it breaks up everything: all the hidden elements (with ng-hide="true" or ng-show="false") are ALWAYS showing up

matsko:
This patch fixes the issue where filtered-out leave animations were not
properly run the DOM operation when closed.

Closes #11555

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -106,7 +106,7 @@ describe("animations", function() {
       module(function($animateProvider) {
         $animateProvider.classNameFilter(/only-allow-this-animation/);
       });
-      inject(function($animate, $rootScope, $document, $rootElement) {
+      inject(function($animate, $rootScope) {
         expect(element).not.toHaveClass('only-allow-this-animation');
 
         $animate.enter(element, parent);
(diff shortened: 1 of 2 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -121,6 +121,22 @@ describe("animations", function() {
       });
     });
 
+    it('should complete the leave DOM operation incase the classNameFilter fails', function() {
```
gkalpak: _incase_  -->  _in case_ (?)

test/ngAnimate/animateSpec.js
```diff
@@ -121,6 +121,22 @@ describe("animations", function() {
...
+      module(function($animateProvider) {
+        $animateProvider.classNameFilter(/memorable-animation/);
+      });
+      inject(function($animate, $rootScope, $document, $rootElement) {
```
gkalpak: `$document` and `$rootElement` seem to be unnecessary here (in the previous test too).

matsko:
LGTM (from Igor in person).

matsko:
All fixed up @gkalpak. Thanks for the review.

matsko:
Merged as https://github.com/angular/angular.js/commit/6dd64ab5f34fa19db8f90e6eabc810843089ba14

PR #11575, opened 2015-04-14
fix(ngAnimate): ensure that animations work when the app is bootstrapped on the document node

### LINKED ISSUE #11574: 1.4 animations do not work when the application is bootstrapped on `document`.

matsko:
Closes #11574

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -491,6 +491,12 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       var animateChildren;
 
       while (parent && parent.length) {
+        if (!rootElementDetected) {
+          // angular doesn't want to attempt to animate elements outside of the application
+          // therefore we need to ensure that the rootElement is an ancestor of the current element
+          rootElementDetected = isMatchingElement(parent, $rootElement);
+        }
+
         var parentNode = parent[0];
         if (parentNode.nodeType !== ELEMENT_NODE) {
           // no point in inspecting the #document element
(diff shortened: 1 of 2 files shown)


src/ngAnimate/animateQueue.js
```diff
@@ -491,6 +491,12 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       var animateChildren;
 
       while (parent && parent.length) {
+        if (!rootElementDetected) {
+          // angular doesn't want to attempt to animate elements outside of the application
+          // therefore we need to ensure that the rootElement is an ancestor of the current element
+          rootElementDetected = isMatchingElement(parent, $rootElement);
+        }
```
gkalpak: I wonder why moving this part here solves the issue...
Any insights ?

matsko: The `nodeType` if statement still breaks the loop, but the rootElement is detected beforehand now so that sets `rootElementDetected` to true.

ozzeD: y>3x²

gkalpak: Right ! Thx @matsko and @ozzeD for the explanation (actually @ozzeD's explanation was more to the point).

gkalpak:
Totally unrelated to this PR, but I noticed that the `details`' lookup on [line 506][1] could be moved inside the `if` block (I think).

[1]: https://github.com/angular/angular.js/pull/11575/files#diff-c15d87c15661cba4a721501ba4fda5b2R506

matsko:
LGTM (from Igor in person).

matsko:
Merged in as https://github.com/angular/angular.js/commit/bee14ed1e7b77ea7dc62326611380da36dec297e

PR #11576, opened 2015-04-14
docs(i18n): expand the MessageFormat syntax documentation

chirayuk:


chirayuk:
@petebacondarwin—I also have a bugfix for MessageFormat that needs to go in this release.  I also missed the bower publish/unpublish for this module that I'll be adding (also a separate PR.)

docs/content/guide/i18n.ngdoc
```diff
@@ -18,8 +18,10 @@ application means providing translations and localized formats for the abstracte
 Angular supports i18n/l10n for {@link ng.filter:date date}, {@link ng.filter:number number} and
 {@link ng.filter:currency currency} filters.
 
-Additionally, Angular supports localizable pluralization support through the {@link
-ng.directive:ngPluralize `ngPluralize` directive}.
+Localizable pluralization is supported via the {@link ng.directive:ngPluralize `ngPluralize`
```
petebacondarwin: @chirayuk do you think we should deprecate ngPluralize in 1.5 and put it in its own separate module?

docs/content/guide/i18n.ngdoc
```diff
@@ -142,84 +144,179 @@ displaying the date with a timezone specified by the developer.
...
+The syntax extension is based on a subset of the ICU MessageFormat syntax that covers plurals and
+gender selections.  Please refer to the links in the “Further Reading” section at the bottom of this
+section.
+
+You may find it helpful to play with our [Plnkr Example](http://plnkr.co/edit/QBVRQ70dvKZDWmHW9RyR?p=preview)
+as you read the examples below.
+
+### Plural Syntax
+
+The syntax for plural based message selection looks like this following:
```
petebacondarwin: ~~this~~ the following

docs/content/guide/i18n.ngdoc
```diff
@@ -142,84 +144,179 @@ displaying the date with a timezone specified by the developer.
...
+one, that message is used.  If there was no match, then the offset value is subtracted from the
+value of the expression and locale specific pluralization rules are applied to this new value to
+obtain its plural category (such as “one”, “few”, “many”, etc.) and a match is attempted against the
+keyword selectors and the matching message is used.  If there was no match, then the “other”
+category (required) is used.  The value of the `#` character inside a message is the value of
+original expression reduced by the offset value that was specified.
+
+### Escaping / Quoting
+
+You might need to quote the curly braces or the `#` character inside the message texts so that they
```
petebacondarwin: You might need to ~~quote~~ _use_ curly braces or the `#` character

docs/content/guide/i18n.ngdoc
```diff
@@ -142,84 +144,179 @@ displaying the date with a timezone specified by the developer.
...
+
+Here, `EXPRESSION` is an Angular expression that evaluates to the gender of the person that
+is used to select the message that should be displayed.
+
+The Angular expression is followed by `, select,` where the spaces are optional.
+
+This is followed by a list of selection keyword and corresponding message pairs.  The "other"
+keyword and corresponding message are **required** but you may have as few or as many of the other
+gender values as you need (i.e. it isn't restricted to male/female.)  Note however, that the
+matching is **case-sensitive**.
```
petebacondarwin: What happens if you wanted to actually match the string `"other"` but also have the catch-all `other` selector?

petebacondarwin:
This looks great @chirayuk!
Apart from a few minor tweaks above it looks good. Feel free to fix up and merge.

docs/content/guide/i18n.ngdoc
```diff
@@ -142,84 +144,179 @@ displaying the date with a timezone specified by the developer.
...
+Following the Angular expression, you would denote the plural extension syntax by the `, plural,`
+syntax element.  The spaces there are optional.
+
+This is followed by a list of selection keyword and corresponding message pairs.  The "other"
+keyword and corresponding message are **required** but you may have as few or as many of the other
+categories as you need.
+
+#### Selection Keywords
+
+The selection keywords can be either exact matches or language dependant [plural
```
gkalpak: _dependant_  -->  _dependent_ ?

docs/content/guide/i18n.ngdoc
```diff
@@ -142,84 +144,179 @@ displaying the date with a timezone specified by the developer.
...
 
-```
+```text
 {{numMessages, plural,
-      =0 { You have no new messages }
-      =1 { You have one new message }
-   other { You have {{numMessages}} new messages }
+      =0 {You have no new messages}
+      =1 {You have one new message}
+   other {You have {{numMessages}} new messages}
```
gkalpak: You could even right it as:

```
    =1 {You have one new message}
other {You have {{numMessage||'no'}} new messages}
```

:P

docs/content/guide/i18n.ngdoc
```diff
@@ -142,84 +144,179 @@ displaying the date with a timezone specified by the developer.
...
-Please note that **other** is a **required** category (for
-both the plural syntax and the select syntax that is shown
-later.)
+### Plural syntax with optional `offset`
+
+The plural syntax supports an optional `offset` syntax that is used in matching.  It's simpler to
+explain this with an example.
+
+```text
+{{recipients.length, plural, offset:1
```
gkalpak: I don't know if it is intended, but having 0 recipients (assuming the offset of `1`), results in a `-1` count and displays the:
_"You gave  and -1 other people a gift"_

But there doesn't seem to be a way to specify a messages for `-1` (i.e. 0 recipients - 1 offset).
Is there another solution to this ? If so it should be documented (if not it should be probably implemented :))

docs/content/guide/i18n.ngdoc
gkalpak: Erh...I think this comment is not valid. (Unless you mean the whole sentence should be changed/re-phrased.)

docs/content/guide/i18n.ngdoc
```diff
@@ -227,11 +324,20 @@ later.)
...
 
-### More complex example that combines some of these
+### Nesting
+
+As mentioned in the syntax for plural and select, the embedded messages can contain Angular
+interpolation syntax.  Since you can use MessageFormat extensions in Angular interpolation, this
+allows you to nest plural and gender expressions in any order.
+
+Please note that if these are intended to reach a translator and be translated, it is recommended
+that the messages appear as a while and not be split up.
```
gkalpak: _while_  -->  _whole_ ?

docs/content/guide/i18n.ngdoc
```diff
@@ -249,3 +355,27 @@ This is taken from the [plunker example](http://plnkr.co/edit/QBVRQ70dvKZDWmHW9R
...
+documentation.  Anywhere in that MessageFormat that you have regular message text and you want to
+substitute an expression, just put it in double curlies instead of single curlies that MessageFormat
+dictates.  This has a huge advantage.  **You are no longer limited to simple identifiers for
+substitutions**.  Because you are using double curlies, you can stick in any arbitrary interpolation
+syntax there, including nesting more MessageFormat expressions!  Some examples will make this clear.
+In the following example, I will only be showing you the AngularJS syntax.
+
+### Further Reading
+For more details, please refer to our [design doc](https://docs.google.com/a/google.com/document/d/1pbtW2yvtmFBikfRrJd8VAsabiFkKezmYZ_PbgdjQOVU/edit)
+for a lot more details.  You can read more about the ICU MessageFormat syntax at
```
gkalpak: Too many `details` in this sentence ?

docs/content/guide/i18n.ngdoc
```diff
@@ -249,3 +355,27 @@ This is taken from the [plunker example](http://plnkr.co/edit/QBVRQ70dvKZDWmHW9R
...
+is otherwise illegal in an Angular expression and is used by MessageFormat to specify the function
+(such as plural/select) and it's related syntax.
+
+To understand the extension, take a look at the ICU MessageFormat syntax as specified by the ICU
+documentation.  Anywhere in that MessageFormat that you have regular message text and you want to
+substitute an expression, just put it in double curlies instead of single curlies that MessageFormat
+dictates.  This has a huge advantage.  **You are no longer limited to simple identifiers for
+substitutions**.  Because you are using double curlies, you can stick in any arbitrary interpolation
+syntax there, including nesting more MessageFormat expressions!  Some examples will make this clear.
+In the following example, I will only be showing you the AngularJS syntax.
```
gkalpak: Ehm...where's the example ?

gkalpak:
I like this (I originally thought MessageFormat would be complex to use, but it turns out it is pretty much straight-forward). :+1: @chirayuk 

Apart from the minor inline comments, I have a couple (not so minor ?) not-inline comments/questions/remarks:

1. The minified version of `angular-messageFormat` of the current snapshot seems to have issues ([example][1]). Using the non-minified version works fine.
  Without digging too deep, I think it is because the IIFE is invoked with `window.Ga` instead of `window.angular`. (Something related to the config of Closure compiler perhaps ?)

2. [Elsewhere][2], the module file is referred to as `angular-messageformat[.min].js` (all lowercase), but it seems to be `angular-messageFormat` (capital `F`) now. I would suggest changing it to something lowercase (either `angular-messageformat` or `angular-message-format`) to avoid issues with OSes treating file names and casing in different/weird ways.
  In any case, the name should appear consistently (and correctly) across all docs.

<sub>@chirayuk, let me know if you would like me to open separate issues.</sub>

[1]: http://codepen.io/ExpertSystem/pen/pvMNzj?editors=101
[2]: https://docs.angularjs.org/api/ngMessageFormat

docs/content/guide/i18n.ngdoc
petebacondarwin: Yes I was being lazy. I think the whole sentence should read:

> You will need to escape curly braces or the `#` character inside message texts if you want them to be treated literally with no special meaning.

docs/content/guide/i18n.ngdoc
chirayuk: I hadn't done that because I had mistakenly assumed that ngPluralize allowed its embedded messages to contain HTML.  Reading the docs and testing on plnkr.co shows that that is not the case.

So yes, we should definitely deprecate this for 1.5!  Should I say something about it in the docs for 1.4 right now?

docs/content/guide/i18n.ngdoc
chirayuk: Sadly, that's not an option. :)  That just seems to be how MessageFormat has been defined.  We could deviate from the spec and use something else instead of "other" for the default case, but I feel like that is a bit too much for a rare case.  (Our other deviations are more understandable.)

docs/content/guide/i18n.ngdoc
chirayuk: I should actually add a section on best practices but that would touch on many things so I have not done so.

The example you have written is actually wrong.  I mean—it's correct for English, but it's highly unlikely to work if you use regular translation tools.  You're basically embedding a fragment of English text inside a subexpression and expecting that it will work.  The Angular expression typically would become an opaque placeholder to the translator that they can't change (and can't see that it contains "no" inside it.)  Also, in a different language, the position of "no" might need to move around.  This scheme will definitely work if one is only supporting a single language.

The following is also bad practice as it relies on the same type of "concatenation" assumption.  (The best practice is to let the translator see the messages in their entirety and not piecemeal.)

```text
You have {{numMessages, plural, =0 {no} =1 {one} other {#} }} new messages
```

docs/content/guide/i18n.ngdoc
chirayuk: For pluralization, negative numbers don't make sense—it's a logic bug.  If you're using an offset, your messages are automatically written to use this information (e.g. "… and # other people…").  You can simply add exact matches for the cases that would cause you to have negative "#" evaluations.

docs/content/guide/i18n.ngdoc
chirayuk: Fixed.

docs/content/guide/i18n.ngdoc
chirayuk: Oh yes, good call!  Fixed.

docs/content/guide/i18n.ngdoc
chirayuk: Good catch.  Fixed.

docs/content/guide/i18n.ngdoc
chirayuk: Fixed.

docs/content/guide/i18n.ngdoc
chirayuk: Yes.  Detail level reduced.

docs/content/guide/i18n.ngdoc
chirayuk: Removed the sentence.  This paragraph occurred a lot earlier in the previous version and it doesn't make sense anymore.

petebacondarwin:
@gkalpak - I think your first issue has just been fixed with #11592

petebacondarwin:
LGTM

petebacondarwin:
@gkalpak for your second point, I wonder if the capitalization of the module name is being changed by dgeni when generating that page...

chirayuk:
@gkalpak—regarding your comments:

The minified version bug has been fixed in a PR with an E2E test for it.  See PR 11592.

The casing—I fixed the only place it was inconsistently cased and made it part of this PR.  However, that does not fix the problem because the stuff generated in the `build/` tree contains lowercased references.

Would you file two issues for this—(1) casing should be changed to lowercase everywhere for consistency, (2) for me to create the bower repository and update `scripts/bower/(un)publish.sh` to use it so that this is published to bower as well.  I would like to tackle those as separate issues.  Thanks.

chirayuk:
On bower repo/module file name casing:  I'm going to go with bower-angular-message-format and angular-message-format.min.js. based on chatting with @IgorMinar and @btford.  This avoid the issue of casing that might affect case insensitive filesystem and has better consistency with names in URLs (i.e. using the hyphen as the separator character.)  Let me know if you have other thoughts.

docs/content/guide/i18n.ngdoc
gkalpak: You are right !
I am not in a 100% i18n mindset yet :)

docs/content/guide/i18n.ngdoc
gkalpak: I still don't get how am I supposed to handle a case with 0 recipients and `offset="1"`.

gkalpak:
@chirayuk, I see you have already opened the issues.
I too think that `angular-message-format` is the most consistent naming.

chirayuk:
@gkalpak @petebacondarwin the rename to `angular-message-format` is complete.  It's in sha 0d64f08005d235ad76208000e34517591b6192e5 (was PR #11597).  I'm going to merge this one in now.

PR #11577, opened 2015-04-14
Update ngClick.js - find label inside parent element to correctly prevent second click from being busted

mobilabgit:
In our code we have an input element inside a label. The click on the label generates two clicks, but the second click was not recognised by angular-touch as a label click, because it only looks at the current element, instead of looking for a label in parent elements. Therefore, the second click gets busted, and clicking on the label doesn't do anything for 2500ms.

I added the additional lookup for a containing label inside one of the conditions inside the onClick method.


mobilabgit:
I signed it

petebacondarwin:
I have tidied up this PR but it needs a unit test: https://github.com/angular/angular.js/compare/master...petebacondarwin:pr-11577

PR #11579, opened 2015-04-14
fix($http): stop $httpBackend coersion of falsy values to null before xh...

samjacobclift:
...r.send()

stop $httpBackend from potentially overriding false values to null when sending a falsy value to a web service via POST or PUT et al

### DIFF
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -109,7 +109,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
         }
       }
 
-      xhr.send(post || null);
+      xhr.send(post);
     }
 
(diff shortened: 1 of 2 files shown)


johnhoffman:
This passed: https://github.com/angular/angular.js/pull/11593. Apparently, the E2E tests can fail for extraneous reasons.

gkalpak:
Just to be clear, it is (almost) always possible to say if the test failures are flakes or _real_ failures by looking at the Travis job logs.

For the record, the test failures for this commit are _real_, but unrelated to this fix.
(Someone broke master with 2a156c2 :stuck_out_tongue:, but it's all good now.)

Closing this in favor of #11593 (which has better test coverage).

PR #11581, opened 2015-04-14
fix($location): prevent infinite $digest from no trailing slash in IE9

### LINKED ISSUE #11439: IE9 infinite $digest bug if loading a base URL without a trailing slash

```javascript
if (!sameBase) {
    reloadLocation = url;
}
```

to:

```javascript
if (replace || !sameBase) {
    reloadLocation = url;
}
```

hamfastgamgee:
fix($location): prevent infinite $digest from no trailing slash in IE9

This fix prevents IE9 from throwing an infinite $digest error when the user accesses the base
URL of the site without a trailing slash. Suppose you owned http://www.mysite.com/app
and had an Angular app hosted in a subdirectory "app". If an IE9 user accessed
http://www.mysite.com/app infinite $digest errors would be thrown on the console, but the app
itself would eventually resolve properly and work fine. Now the infinite $digest errors will
not be thrown.

Closes #11439

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -883,7 +883,7 @@ function $LocationProvider() {
       $browser.url($location.absUrl(), true);
     }
 
-    var initializing = true;
+    var initializing = true, skipLocationWatch = false;
 
     // update $location when $browser url changes
     $browser.onUrlChange(function(newUrl, newState) {
@@ -916,6 +916,11 @@ function $LocationProvider() {
 
     // update browser
     $rootScope.$watch(function $locationWatch() {
+      if (skipLocationWatch && !$location.$$replace) {
+        //prevent infinite $digest errors in hashbang mode
+        skipLocationWatch = false;
+        return;
+      }
       var oldUrl = trimEmptyHash($browser.url());
       var newUrl = trimEmptyHash($location.absUrl());
       var oldState = $browser.state();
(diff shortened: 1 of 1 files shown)


hamfastgamgee:
I signed it!

hamfastgamgee:
I fully expect this to get some comments as to its correctness.  I'm just not sure what the consequences of changing reloadLocation in $browser.url() or trimEmptyHash are, which are where the real problems are found.

hamfastgamgee:
The general gist is: In the initial pass through $locationWatch(), the afterLocationChange call will either add the trailing slash or add the "otherwise" route (say, /Home) to the URL after the hash.  If the latter case, $locationWatch will fire again with $location.$$replace=true, and we need to let that go through to actually change the browser location and get us out of replace mode.  But then $locationWatch will fire again (or for the second time if there isn't an "otherwise" route), and this time it will see trimEmptyHash($browser.url()) and trimEmptyHash($location.absUrl()) as unequal and try to change the URL continuously.  $browser.url() never ends up updating because sameBase always ends up as true, so we don't actually update reloadLocation which is feeding the $browser.url() getter.  So we end up in an infinite recursion of trying to update the URL and failing.

I'm totally open to something that fixes the real problem rather than basically whacking off the infinite recursion. :)  I'm pretty sure the reason this is all fine with browsers with the History API is that the sameBase check returns false in one of those iterations and so reloadLocation gets updated properly.

hamfastgamgee:
Blast.  Seems this doesn't fully work anyway.  It works on the simplified testcase that's linked off #11439, but it doesn't work in my actual fully fledged app (probably because I have more digest cycles fire).  So maybe we have to change $browser.url()'s setter to actually set reloadLocation?  I'm not sure why it's doing the sameBase check to begin with, so I'm not sure what the consequences are of changing that.

hamfastgamgee:
As this one doesn't work, closing pull request.

Issue #11562, opened 2015-04-13
Migration from 1.3 to 1.4 docs?


pocesar:
When the compilation of breaking changes and the guide for migrating from 1.3 to 1.4 is coming to the angularjs.org website docs?

petebacondarwin:
They will be in place before we go to full 1.4.0

pocesar:
alright, just asking because I couldn't seen any dates or milestones for it

matsko:
This PR is the basis of the migration log: https://github.com/angular/angular.js/pull/11584

petebacondarwin:
You can now see these migration notes on the angularjs.org site: https://docs.angularjs.org/guide/migration

pocesar:
:+1:

PR #11584, opened 2015-04-14
docs(MIGRATION): add notes for the migration from 1.3 to 1.4

matsko:


petebacondarwin:
Great stuff and very efficient @matsko. I haven't had a chance to read through the whole thing but my only suggestion would be to have a kind of introduction that lists the high level tasks that might need to be considered if I were about to embark on a migration of my 1.3.x app.

docs/content/guide/migration.ngdoc
```diff
@@ -13,6 +13,286 @@ which drives many of these changes.
...
+Previously, '&' expressions would always set up a function in the isolate scope. Now, if the binding
+is marked as optional and the attribute is not specified, no function will be added to the isolate scope.
+
+
+
+## Cookies (`ngCookies`)
+
+Due to [38fbe3ee](https://github.com/angular/angular.js/commit/38fbe3ee8370fc449b82d80df07b5c2ed2cd5fbe),
+`$cookies` will no longer expose properties that represent the current browser cookie
+values. Now you must explicitly the methods described above to access the cookie
```
gkalpak: _Now you must explicitly ... the methods_ (call ?)

_described above_ (Where ? :))

docs/content/guide/migration.ngdoc
```diff
@@ -13,6 +13,286 @@ which drives many of these changes.
...
+```
+
+
+
+
+## Filters (`filter`, `limitTo`)
+
+### `filter` filter
+Due to [cea8e751](https://github.com/angular/angular.js/commit/cea8e75144e6910b806b63a6ec2a6d118316fddd),
+Previously, with the standard `filter:` filter, the filter process was not applied and was silently ignore
```
gkalpak: _`filter:`_  -->  _`filter`_

_ignore_  -->  _ignored_

gkalpak:
BTW, this (to some extend ?) addresses #11562.

docs/content/guide/migration.ngdoc
```diff
@@ -13,6 +13,301 @@ which drives many of these changes.
...
+```
+
+Depending on where the `ngMessagesInclude` directive is placed it will be prioritized inline with the other messages
+before and after it.
+
+### ngOptions
+
+Due to [7fda214c](https://github.com/angular/angular.js/commit/7fda214c4f65a6a06b25cf5d5aff013a364e9cef),
+when `ngOptions` renders the option values within the DOM, the resulting HTML code is different. Normally this
+should effect your application at all, however, if your code relies on inspecing the value property of `<option>`
```
petebacondarwin: should **not affect** your application

docs/content/guide/migration.ngdoc
```diff
@@ -13,6 +13,301 @@ which drives many of these changes.
...
+https://github.com/petebacondarwin/angular-toArrayFilter
+or some other mechanism, and then sort them manually in the order you need.
+
+
+### $compile
+
+Due to [6a38dbfd](https://github.com/angular/angular.js/commit/6a38dbfd3c34c8f9efff503d17eb3cbeb666d422),
+previously, '&' expressions would always set up a function in the isolate scope. Now, if the binding
+is marked as optional and the attribute is not specified, no function will be added to the isolate scope.
+
```
petebacondarwin: I think there might be a breaking change here, where now if you don't mark a `&` attribute as optional then fail to specify it, you will get an error instead of a `noop`.

petebacondarwin:
Just a couple of notes. Otherwise LGTM

docs/content/guide/migration.ngdoc
```diff
@@ -13,6 +13,301 @@ which drives many of these changes.
...
+when an enter, leave or move animation is triggered then it will always end any pending or active parent
+class based animations (animations triggered via ngClass) in order to ensure that any CSS styles are resolved in time.
+
+
+
+
+## Forms (`ngMessages`, `ngOptions`)
+
+### ngMessages
+The ngMessages module has also been subject to an internal refactor to allow the featire to be more flexible
```
pocesar: allow the featire => allow the feature

docs/content/guide/migration.ngdoc
matsko: Would you mind making another commit on top of this one? I don't have enough knowledge of how '&' works and the fix that was made for it.

matsko:
Merged in as https://github.com/angular/angular.js/commit/042558a380de1fe17c20254149461123e7656693. Thank you guys.

docs/content/guide/migration.ngdoc
petebacondarwin: I can.

petebacondarwin: Actually looking back through the PR and issue, I see that we decided not to make the second breaking change that I indicated.

PR #11592, opened 2015-04-14
fix(ngMessageFormat): minified symbol and nested required expression

### LINKED ISSUE #11414: angular-messageFormat select statement bug
I have the following in my template
```
<span>
   <!--i18n: Gender fun-->
   {{model.gender, select,
         male   {He gave me a cactus}
         female {She gave me a cactus}
         other  {They gave me a cactus}
   }}
   <!--/i18n-->
</span>
```
And I am getting an error that says 'other is a required option' with stacktrace:

```
at http://localhost:1080/static/lib/angular/angular.min.js:6:434
    at new MessageSelectorBase (http://localhost:1080/static/lib/angular/angular-messageFormat.js:102:11)
    at MessageSelectorBase.SelectMessage (http://localhost:1080/static/lib/angular/angular-messageFormat.js:165:23)
    at MessageFormatParser.ruleSelectKeyword (http://localhost:1080/static/lib/angular/angular-messageFormat.js:577:21)
    at MessageFormatParser.run (http://localhost:1080/static/lib/angular/angular-messageFormat.js:451:12)
    at interpolate (http://localhost:1080/static/lib/angular/angular-messageFormat.js:891:14)
```

I tried copy-pasting the example from the docs for the messageformat pr and got the same result.

chirayuk:
Add an E2E test that works against the minified module to test that the
minified build works correctly.

Fix a bug where mustHaveExpression was passed through to submessages
unchanged. Use of the messageFormat syntax automatically means that you
are using an expression.  Therefore, submessages should not be required
to also have messages.  This closes #11414

### DIFF
--- a/src/ngMessageFormat/messageFormatCommon.js
+++ b/src/ngMessageFormat/messageFormatCommon.js
@@ -5,11 +5,11 @@
 // This file is compiled with Closure compiler's ADVANCED_OPTIMIZATIONS flag! Be wary of using
 // constructs incompatible with that mode.
 
-var $interpolateMinErr = angular['$interpolateMinErr'];
+var $interpolateMinErr = window['angular']['$interpolateMinErr'];
 
-var noop = angular['noop'],
-    isFunction = angular['isFunction'],
-    toJson = angular['toJson'];
+var noop = window['angular']['noop'],
+    isFunction = window['angular']['isFunction'],
+    toJson = window['angular']['toJson'];
 
 function stringify(value) {
   if (value == null /* null/undefined */) { return ''; }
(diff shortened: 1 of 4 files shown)


src/ngMessageFormat/messageFormatCommon.js
```diff
@@ -5,11 +5,11 @@
 // This file is compiled with Closure compiler's ADVANCED_OPTIMIZATIONS flag! Be wary of using
 // constructs incompatible with that mode.
 
-var $interpolateMinErr = angular['$interpolateMinErr'];
+var $interpolateMinErr = window['angular']['$interpolateMinErr'];
```
gkalpak: This shouldn't be necessary, since _normally_ the code would be wrapped into an IIFE and passed `window.angular` as an argument:

```js
(function (window, angular, undefined) { ... })(window, window.angular);
```

Any idea why it doesn't work ?

PR #11593, opened 2015-04-14
stop coercing falsy values to null before xhr.send()

### LINKED ISSUE #11552: $httpBackend coersion of falsy values to null before xhr.send()

johnhoffman:
This is intended to resolve https://github.com/angular/angular.js/issues/11552.

### DIFF
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -109,7 +109,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
         }
       }
 
-      xhr.send(post || null);
+      xhr.send(post);
     }
 
     if (timeout > 0) {
--- a/test/ng/httpBackendSpec.js
+++ b/test/ng/httpBackendSpec.js
@@ -50,6 +50,23 @@ describe('$httpBackend', function() {
     expect(xhr.$$data).toBe(null);
   });
 
+  it('should pass the correct falsy value to send if falsy body is set (excluding NaN)', function() {
+    var values = [false, 0, "", null, undefined];
+    angular.forEach(values, function(value) {
+      $backend('GET', '/some-url', value, noop);
+      xhr = MockXhr.$$lastInstance;
+
+      expect(xhr.$$data).toBe(value);
+    });
+  });
+
+  it('should pass NaN to send if NaN body is set', function() {
+    $backend('GET', '/some-url', NaN, noop);
+    xhr = MockXhr.$$lastInstance;
+
+    expect(isNaN(xhr.$$data)).toEqual(true);
+  });
+
   it('should call completion function with xhr.statusText if present', function() {
     callback.andCallFake(function(status, response, headers, statusText) {
       expect(statusText).toBe('OK');


src/ng/httpBackend.js
```diff
@@ -109,7 +109,12 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
         }
       }
 
-      xhr.send(post || null);
+      if (post === false) {
```
gkalpak: Why special-case `false` only and not all falsy values ?

johnhoffman: That would be ideal. See https://github.com/angular/angular.js/issues/11552. Unfortunately Travis CI E2E tests failed in that case. If this fails, I'll dig deeper since it implies strongly that something is wrong with the tests. This is my first attempt at contributing to angular, so perhaps I don't understand the intended workflow well yet. I've been unable to build/test on my local machine per these instructions: https://docs.angularjs.org/misc/contribute.

gkalpak: @johnhoffman, this build has failed because of a style error: `else` should be on the same line as the closing `}`:

```js
// WRONG
}
else {

// RIGHT
} else {
```

---
BTW, not all Travis failures means there is a problem with the code (you have to see the details; sometimes it's an unrelated timeout or something similar).
Additionally, you shouldn't create new PRs just to rerun the tests on Travis. Amending the latest commit and force-pushing to your repo, will update the PR and restart the Travis job.

Just so you know for future PRs :)
In any case, this is a legitimate failure, but I still wonder why only allow `false` ?

---
> I've been unable to build/test on my local machine per these instructions

What was the problem ? You should be able to run the tests and it is always better to have them pass locally first, in order to avoid losing time (Travis takes much longer) and wasting resources.

johnhoffman: Thanks for your guidance, @gkalpak. I was able to fix my npm installation using https://github.com/brock/node-reinstall. I updated the code and new tests to reflect what I believe we should expect. Unit tests passed and E2E may be failing for extraneous reasons as you suggested. I'll update the PR per your instructions.

gkalpak:
@johnhoffman, could you, please, squash all commits into one and add fix the commit message header and add `Closes #11552` at the bottom of the commit message ? (For more info on commit message conventions take a look [here][1]).

Other than that it LGTM.

johnhoffman:
Thanks, @gkalpak. Commits are squashed. Does the commit message look good?

gkalpak:
It LGTM. Thx @johnhoffman !

caitp:
LGTM

caitp:
I'm of the opinion that we should also land this into 1.3 --- but lets let it bake for a bit first

PR #11597, opened 2015-04-15
chore(ngMessageFormat): rename angular-messageFormat to angular-message-format

### LINKED ISSUE #11595: MessageFormat module name casing
`angular-messageFormat` should be named `angular-message-format` (and corresponding files to `angular-message-format.min.js`, etc.)  This keeps the name cleaner and avoids possible issues with case-insensitive systems (where the index.html might reference the script with a lowercased name and it accidentally works until moving to a case sensitive file system.)

chirayuk:
Closes #11595

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -204,7 +204,7 @@ module.exports = function(grunt) {
         src: util.wrap(files['angularModules']['ngResource'], 'module')
       },
       messageformat: {
-        dest: 'build/angular-messageFormat.js',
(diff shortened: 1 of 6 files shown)


chirayuk:
Note that this will conflict with PR 11576 that improves the i18n docs.  However, the conflict is just two lines related to angular-messageFormat.js so it should be an easy resolution.

I think it is more correct to land this PR first followed by PR 11576.

Issue #11361, opened 2015-03-18
ng-required="true" does not allow to empty field when using ng-model-options="{ getterSetter: true }"


mmazo:
When using both ng-required="true" and ng-model-options="{ getterSetter: true }",
there is no possibility to empty the field. You always get the last character 'standing' in the input field.
This happens because 'newName' value in getter-setter function for name is undefined when value is empty. It happens only when using ng-required="true". If the field is not required, we get just empty string and everyrhing works fine.

Here is the code example:

```html
<input type="text" name="userName"
           ng-required="true"
           ng-model="user.name"
           ng-model-options="{ getterSetter: true }" />
```
```javascript
(function(angular) {
  'use strict';
angular.module('getterSetterExample', [])
  .controller('ExampleController', ['$scope', function($scope) {
    var _name = '';
    $scope.user = {
      name: function(newName) {
        console.log(newName);
        if (angular.isDefined(newName)) {
          _name = newName;
        }
        return _name;
      }
    };
  }]);
})(window.angular);
```

Narretz:
Can you please put the code as a runnable example on a plnkr.co or similar?

mmazo:
Hi Narretz,

here is the example: http://plnkr.co/edit/P5NaLBwWYB2v8O9U0W0A?p=preview

andrewaustin:
@mmazo You need to turn off validation by passing ```allowInvalid: true``` to your ng-model-options. You can see a working plnkr here: http://plnkr.co/edit/6vuK2vAb9m24eRL5uQ0Q?p=preview

You can see a more details from ngConf here: https://www.youtube.com/watch?v=k3t3ov6xHDw&t=211

gkalpak:
Using `allowInvalid` will work around the issue, but it is not solving the actual problem.
The actual problem lies in the implementation of `name()`, which treats passing an argument with value `undefined` as not passing an argument (although these are obviously two very different things).

You should use a more "accurate" way of defining whether the function is called as a getter or as a setter (i.e. whether `newName` has been passed (even if `undefined`) or not).
E.g.:

```js
name: function(newName) {
  if (arguments.length) {
    _name = newName || '';
  }
  return _name;
}
```

[Demo][1]

--
I think everything works as expected, so there's nothing actionable here.

[1]: http://plnkr.co/edit/CF4O6rzfGWHSjHTQamln?p=preview

Narretz:
Thanks @gkalpak, I agree, this is not a bug.

[...]

PR #11604, opened 2015-04-15
docs(ngModel): getterSetter and `arguments.length` vs. `angular.isDefined`

yankee42:
When using `ng-model-options="{ getterSetter: true }"` one should be aware that the getter/setter method could be called with an argument set to undefined even if the method is used as setter. This e.g. happens when the model is invalid:
https://github.com/angular/angular.js/blob/dd8b157299520e8167b918439c2d3534e8cba550/src/ng/directive/ngModel.js#L535

Using `arguments.length` reliably detects whether the getter/setter should act as getter or setter.
Credits for the concept:
https://github.com/angular/angular.js/issues/11361#issuecomment-84022489

I believe that this should be documented as failing to realise this can cause weird behaviour and large time spendings on hunting bugs (at least that happened to me).

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -1205,7 +1205,11 @@ var DEFAULT_REGEXP = /(\s+|^)default(\s+|$)/;
           var _name = 'Brian';
           $scope.user = {
             name: function(newName) {
-              return angular.isDefined(newName) ? (_name = newName) : _name;
+              // Note that newName can be undefined for two reasons:
+              // 1. Because it is called as a getter and thus called with no arguments
+              // 2. Because the property should actually be set to undefined. This happens e.g. if the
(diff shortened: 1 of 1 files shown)


yankee42:
I signed and also added my github username. I suppose I do not need to set the username AND the emailaddress?!

icorne:
I just ran into this issue, Should this not be on master as well?

gkalpak:
Not sure what you mean. This is a docs change and it has been merged into master as well.

icorne:
I'm not sure either, yesterday I still saw it with the angular.isDefined , today, it's changed to this version. Maybe my mind is playing tricks on me. Sorry for the disturbance!

gkalpak:
Np :smiley: You were probably viewing the docs of an older version the other day (there is version picker in the top-left).

Issue #11598, opened 2015-04-15
ngSanitize syntax error in Chrome/IE11


luanshixia:
[Message] Uncaught SyntaxError: Invalid regular expression: missing /
[File] angular-sanitize.js, Line: 619, Column: 98
[Version] 1.3.15

This error is not observed in 1.2.x.

Narretz:
Can you please provide a runnable code example that shows this error? (plnkr.co etc.) This is not enough information to act on this issue.

luanshixia:
Interestingly, this error is only observable with the unminified angular-sanitize.js. When using angular-sanitize.min.js it is OK. 

```html
<!DOCTYPE html>
<html>
<head>
	<title>Angular Bug</title>
</head>
<body>

	<script type="text/javascript" src="angular.js"></script>
	<script type="text/javascript" src="angular-sanitize.js"></script>
</body>
</html>
```

Open this html in Chrome, you can find error in the console.

pkozlowski-opensource:
I can't reproduce: http://plnkr.co/edit/PaSOmNTTeXG8yq8aZfl3?p=preview

Maybe your file got messed up during the download process? Edited by accident? Or maybe you've got a particular Chrome extension installed?

In any case I can't reproduce it with a simple plunker.

luanshixia:
In the uncompressed file, there is a regex:

    /((ftp|https?):\/\/|(www\.)|(mailto:)?[A-Za-z0-9._%+-]+@)\S*[^\s.;,(){}<>"”’]/

But in the min.js file, this regex is:

    /((ftp|https?):\/\/|(www\.)|(mailto:)?[A-Za-z0-9._%+-]+@)\S*[^\s.;,(){}<>"\u201d\u2019]/

I guess there must be something wrong.

pkozlowski-opensource:
@luanshixia yes, minification process seems to change some quote literals to their \u equivalents. Bot both regexps are still valid (at least syntactically). 

> I guess there must be something wrong.

Might be, but we still need a reproduce scenario. I've linked a simple plunker that represents what you've been describing (or at least my understanding of it!) and doesn't throw any errors?

Could you please modify this plunk to expose the error?

luanshixia:
@pkozlowski-opensource just save the plunker sample to local hard drive and run. You will see that the filter does not work. At least on my machine.

pkozlowski-opensource:
[...]

PR #11609, opened 2015-04-16
fix(ngSanitize): escape the wide char quote marks in a regex in linky.js

luanshixia:
[#11598](https://github.com/angular/angular.js/issues/11598)

It is all about the wide quote mark chars introduced in 1.3.x. I have escaped these chars.

Actually the escaped chars are what is in the compressed sanitize.min.js. If not escaped, there is a chance to get messed up on some machines, causing the script not working.


luanshixia:
Signed the CLA.

PR #11612, opened 2015-04-16
docs(misc/Downloading): upgrade CDN path 1.3.15

johannes-weber:
In order to reference to the latest stable angular version.
Background: a lot of people, including me, copy this snipped and use them in their project without checking for the latest stable version.


johannes-weber:
I signed it!

PR #11614, opened 2015-04-16
docs(angular.element): css() api incompatibility.

adam77:
"When a number is passed as the value, jQuery will convert it to a string and add px to the end of that string."
http://api.jquery.com/css/#css2

jqLite does not appear to do this. 

Partial fix for #11626 .
(can submit full fix if desired)

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -55,7 +55,7 @@
  * - [`children()`](http://api.jquery.com/children/) - Does not support selectors
  * - [`clone()`](http://api.jquery.com/clone/)
  * - [`contents()`](http://api.jquery.com/contents/)
- * - [`css()`](http://api.jquery.com/css/) - Only retrieves inline-styles, does not call `getComputedStyle()`
+ * - [`css()`](http://api.jquery.com/css/) - Only retrieves inline-styles, does not call `getComputedStyle()`. On write, does not convert numbers to strings and append 'px'.
  * - [`data()`](http://api.jquery.com/data/)
  * - [`detach()`](http://api.jquery.com/detach/)
  * - [`empty()`](http://api.jquery.com/empty/)


adam77:
Not sure why test is failing, the change is trivial.

gkalpak:
It's OK now :)

src/jqLite.js
```diff
@@ -55,7 +55,7 @@
  * - [`children()`](http://api.jquery.com/children/) - Does not support selectors
  * - [`clone()`](http://api.jquery.com/clone/)
  * - [`contents()`](http://api.jquery.com/contents/)
- * - [`css()`](http://api.jquery.com/css/) - Only retrieves inline-styles, does not call `getComputedStyle()`
+ * - [`css()`](http://api.jquery.com/css/) - Only retrieves inline-styles, does not call `getComputedStyle()`. On write, does not convert numbers to strings and append 'px'.
```
mzgol: Shouldn't it be:
```
On write, does not convert numbers to strings or append 'px'.
```
? Native speaker required. ;-)

mzgol: Also, this line is too long, it should be broken.

Narretz: I fixed the wording but not the line length. Next time!

----- Ursprüngliche Nachricht -----
Von: "Michał Gołębiowski" <notifications@github.com>
Gesendet: ‎27.‎04.‎2015 22:35
An: "angular/angular.js" <angular.js@noreply.github.com>
Cc: "Martin Staffa" <mjstaffa@googlemail.com>
Betreff: Re: [angular.js] docs(angular.element): css() api incompatibility.(#11614)

In src/jqLite.js:
> @@ -55,7 +55,7 @@
>   * - [`children()`](http://api.jquery.com/children/) - Does not support selectors
>   * - [`clone()`](http://api.jquery.com/clone/)
>   * - [`contents()`](http://api.jquery.com/contents/)
> - * - [`css()`](http://api.jquery.com/css/) - Only retrieves inline-styles, does not call `getComputedStyle()`
> + * - [`css()`](http://api.jquery.com/css/) - Only retrieves inline-styles, does not call `getComputedStyle()`. On write, does not convert numbers to strings and append 'px'.
Also, this line is too long, it should be broken.

gkalpak: Not a native speaker either, but I think it should be "nor" :)
<sub>(Just noting for next time.)</sub>

PR #11620, opened 2015-04-16
fix(ngAnimate): close parent animations only when there are classes to resolve

matsko:
Previously if a parent animation was cancelled then it would not resolve
the runner when that happens. This is now fixed in this patch. Another
fix in this patch ensures that a parent animation is only cancelled if
the animation contains any classes to resolve. This prevents inline
animations from being cancelled.


matsko:
Merged in as https://github.com/angular/angular.js/commit/1459be170dabfca40501dcf219dfced5ba513169

PR #11623, opened 2015-04-16
sec(http): JSONP should require trusted resource URLs

### LINKED ISSUE #11328: Inconsistent handling of $sce trustedUrl with $http
```
// Works:
$http.get($sce.trustAsUrl('/')).success(function(d){$log.log(d);});

// Fails:
$http.get($sce.trustAsUrl('/'), {params: {timestamp: 0}}).success(function(d){$log.log(d);});
```

In the first case, when the trusted URL reaches the XmlHttpRequest, its toString method is used to get the wrapped URL. In the second case, the buildUrl function attempts to add parameters, but the given URL is not a string as expected.

I feel that the buildUrl function should reject a non-string argument for url in all cases.

### LINKED ISSUE #11352: $http's JSONP should require trusted resource URLs
$http does not perform any security checks on the URLs that it receives.  It should require URLs that pass checks for `$sce.RESOURCE_URL`.

This should generally not have a huge impact in terms of breaking backwards compatibility.  The cases that break, are where folks are performing JSONP requests to 3rd party servers.  In such cases, it would be a good idea for the apps that do break to configure the $sce whitelist to include that specific trusted 3rd party server to unbreak themselves.

It might be a good idea to backport this to earlier versions if we can feel confident that we won't actually be breaking most applications.

chirayuk:
WORK-IN-PROGRESS.  Do **NOT** merge.  More work needs to be done and the tests are currently broken.
- JSONP should require trusted resource URLs.  This would be a breaking
  change but maybe not too onerous since same origin URLs are trusted in
  the default config and you can easily whitelist any 3rd party URLs you
  trust in one single place (your app/module config.)
- fix a bug where $http can't handle $sce wrapper URLs.

Closes #11352
Closes #11328

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1110,7 +1110,11 @@ function $HttpProvider() {
           cache,
           cachedResp,
           reqHeaders = config.headers,
-          url = buildUrl(config.url, config.paramSerializer(config.params));
+          // origUrl could be a $sce trusted URL which used a wrapper class
+          origUrl = config.url,
+          url = $sce.valueOf(origUrl);
+
+      url = buildUrl(url, config.paramSerializer(config.params));
 
       $http.pendingRequests.push(config);
       promise.then(removePendingReq, removePendingReq);
@@ -1154,7 +1158,7 @@ function $HttpProvider() {
           reqHeaders[(config.xsrfHeaderName || defaults.xsrfHeaderName)] = xsrfValue;
         }
 
-        $httpBackend(config.method, url, reqData, done, reqHeaders, config.timeout,
+        $httpBackend(config.method, (url === $sce.valueOf(origUrl) ? origUrl : url), reqData, done, reqHeaders, config.timeout,
             config.withCredentials, config.responseType);
       }
 
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -21,14 +21,20 @@ function createXhr() {
  * $httpBackend} which can be trained with responses.
  */
 function $HttpBackendProvider() {
-  this.$get = ['$browser', '$window', '$document', function($browser, $window, $document) {
-    return createHttpBackend($browser, createXhr, $browser.defer, $window.angular.callbacks, $document[0]);
+  this.$get = ['$sce', '$browser', '$window', '$document', function($sce, $browser, $window, $document) {
+    return createHttpBackend($sce, $browser, createXhr, $browser.defer, $window.angular.callbacks, $document[0]);
   }];
 }
 
-function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDocument) {
+function createHttpBackend($sce, $browser, createXhr, $browserDefer, callbacks, rawDocument) {
   // TODO(vojta): fix the signature
   return function(method, url, post, callback, headers, timeout, withCredentials, responseType) {
+    if (lowercase(method) == 'jsonp') {
+      // This is a pretty sensitive operation where we're allowing a script to have full access to
+      // our DOM and JS space.  So we require that the URL satisfies SCE.RESOURCE_URL.
+      url = $sce.getTrustedResourceUrl(url);
+    }
+
     $browser.$$incOutstandingRequestCount();
     url = url || $browser.url();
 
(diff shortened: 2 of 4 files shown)


test/ng/httpBackendSpec.js
```diff
@@ -255,7 +256,22 @@ describe('$httpBackend', function() {
   });
 
 
-  describe('JSONP', function() {
+  [true, false].forEach(function(trustAsResourceUrl) {
```
gkalpak: It seems like a good place for the `they()` helper :)

test/ng/httpBackendSpec.js
```diff
@@ -255,7 +256,22 @@ describe('$httpBackend', function() {
...
+    describe('JSONP: trustAsResourceUrl='+trustAsResourceUrl, function() {
+
+
+    function $backendCall() {
+      var args = Array.prototype.slice.call(arguments);
```
gkalpak: There is also the `sliceArgs()` function. (Just saying...)

test/ng/httpBackendSpec.js
gkalpak: Or maybe not (since it's on the `describe` level, not the `it` level)...

src/ng/httpBackend.js
```diff
@@ -142,8 +148,8 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
     // - adds and immediately removes script elements from the document
     var script = rawDocument.createElement('script'), callback = null;
     script.type = "text/javascript";
-    script.src = url;
     script.async = true;
+    script.src = url;
```
caitp: phantom edit?

src/ng/http.js
```diff
@@ -1154,7 +1158,7 @@ function $HttpProvider() {
           reqHeaders[(config.xsrfHeaderName || defaults.xsrfHeaderName)] = xsrfValue;
         }
 
-        $httpBackend(config.method, url, reqData, done, reqHeaders, config.timeout,
+        $httpBackend(config.method, (url === $sce.valueOf(origUrl) ? origUrl : url), reqData, done, reqHeaders, config.timeout,
```
caitp: is this necessary? why not just `url = $sce.valueOf(origUrl);`

caitp:
are the ng-mock edits enough to not break application tests?

caitp:
@chirayuk it generally looks good ok, a few nits --- I'll leave it up to you whether you think this can land today or not.

petebacondarwin:
@chirayuk - the description says DO NOT MERGE - is this still accurate or are you happy that this is ready to go in now?

IgorMinar:
can we move this to 1.5?

Also, I don't see an easy way to define the whitelist, do we piggy-back on the existing means to define whitelist or is there no current way to whitelist urls that are ok for jsonp?

src/ngMock/angular-mocks.js
```diff
@@ -1095,7 +1095,7 @@ angular.mock.dump = function(object) {
    ```
  */
 angular.mock.$HttpBackendProvider = function() {
-  this.$get = ['$rootScope', '$timeout', createHttpBackendMock];
+  this.$get = ['$sce', '$rootScope', '$timeout', createHttpBackendMock];
```
petebacondarwin: I don't see why the changes in this file are needed...

test/ng/httpBackendSpec.js
```diff
@@ -255,7 +256,22 @@ describe('$httpBackend', function() {
...
+      var url = args[1];
+      if (trustAsResourceUrl || url == null) {
+        args[1] = $sce.trustAsResourceUrl(url);
+        return $backend.apply(null, args);
+      } else {
+        var badUrlPrefix = url ? url.replace('JSON_CALLBACK', '') : '';
+        expect(function() { $backend.apply(null, args); }).toThrowMinErr(
+          '$sce', 'insecurl', 'Blocked loading resource from url not allowed by $sceDelegate policy.  URL: ' + badUrlPrefix);
+      }
+    }
```
petebacondarwin: There is too much work going on here for a test. We should just duplicate what is needed so that we can easily read the tests.

src/ng/http.js
petebacondarwin: This is related to https://github.com/angular/angular.js/issues/11328

petebacondarwin:
Closing in favour of #15143 which is a rebased and tidied up version of this.

PR #11628, opened 2015-04-17
fix($compile): exception when using "watch" as isolated scope binding variable in Firefox

### LINKED ISSUE #11627: Exception when using "watch" as isolated scope binding variable in Firefox
Firefox throws "Error: can't convert undefined to object" exception when using "watch" as a isolated scope binding.
Reproduced in Firefox 35 and Angular 1.4.0-rc.0

See: http://plnkr.co/edit/Hu8t5THUimqXJBZ2BusM?p=preview

randing89:
Firefox throws "Error: can't convert undefined to object" exception when using "watch" as a isolated scope binding
It is because in Firefox Object.prototype.watch is defined and the related code is not using hasOwnProperty to check property availability.

Closes #11627

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2545,9 +2545,19 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         lastValue,
         parentGet, parentSet, compare;
 
+        if (!hasOwnProperty.call(attrs, attrName)) {
+          // In the case of user defined a binding with the same name as a method in Object.prototype but didn't set
+          // the corresponding attribute. We need to make sure subsequent code won't access to the prototype function
+          attrs[attrName] = undefined;
+        }
+
         switch (mode) {
 
           case '@':
+            if (!attrs[attrName] && !optional) {
+              destination[scopeName] = undefined;
+            }
+
             attrs.$observe(attrName, function(value) {
               destination[scopeName] = value;
             });
@@ -2564,6 +2574,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               return;
             }
             parentGet = $parse(attrs[attrName]);
+
             if (parentGet.literal) {
               compare = equals;
             } else {
@@ -2602,9 +2613,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             break;
 
           case '&':
-            // Don't assign Object.prototype method to scope
-            if (!attrs.hasOwnProperty(attrName) && optional) break;
-
             parentGet = $parse(attrs[attrName]);
 
             // Don't assign noop to destination if expression is not valid
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -2231,6 +2231,45 @@ describe('$compile', function() {
               }}
             };
           });
+          directive('prototypeMethodNameAsScopeVarA', function() {
+            return {
+              scope: {
+                'constructor': '=?',
+                'valueOf': '='
+              },
+              restrict: 'AE',
+              template: '<span></span>'
+            };
+          });
+          directive('prototypeMethodNameAsScopeVarB', function() {
+            return {
+              scope: {
+                'constructor': '@?',
+                'valueOf': '@'
+              },
+              restrict: 'AE',
+              template: '<span></span>'
+            };
+          });
+          directive('prototypeMethodNameAsScopeVarC', function() {
+            return {
+              scope: {
+                'constructor': '&?',
+                'valueOf': '&'
+              },
+              restrict: 'AE',
+              template: '<span></span>'
+            };
+          });
+          directive('watchAsScopeVar', function() {
+            return {
+              scope: {
+                'watch': '='
+              },
+              restrict: 'AE',
+              template: '<span></span>'
+            };
+          });
         }));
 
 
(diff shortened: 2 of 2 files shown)


lgalfaso:
#11627 looks like a real issue and the diagnose also looks accurate, but this fix does not look right:

* This patch only fixes the case that the property `watch` is optional, there should be a more comprehensive solution that also fixes this for non-optional bindings
* It needs new tests that fail before the patch and pass with the patch

randing89:
@lgalfaso Thanks for your comment. I have revised my pull request and added few test cases.

randing89:
@lgalfaso 
Hi. Can you please have a review when you got time?
Thanks.

src/ng/compile.js
```diff
@@ -2540,9 +2545,23 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+          // Override property to make sure subsequent code do not have access to prototype method
+          attrs[attrName] = undefined;
+        }
+
         switch (mode) {
 
           case '@':
+            if (!attrs[attrName]) {
+              if (optional) {
+                break;
```
lgalfaso: so if the value is optional and it does not have an initial value, then we are not observing the attribute? This implies that
`<div attr-abc="{{abc}}" some-directive-that-binds-to-the-attribute-abc></div>`
will stop working. This is a breaking change

src/ng/compile.js
```diff
@@ -2540,9 +2545,23 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+        if ((attrName in attrs) && !attrs.hasOwnProperty(attrName)) {
+          // User defined a binding with same as a method in Object.prototype and also the attribute is not present
+          // Override property to make sure subsequent code do not have access to prototype method
+          attrs[attrName] = undefined;
+        }
+
         switch (mode) {
 
           case '@':
+            if (!attrs[attrName]) {
```
lgalfaso: This check should be with `hasOwnProperty` not to run into the original issue

src/ng/compile.js
```diff
@@ -737,6 +737,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
     var bindings = {};
 
+    if (scope.hasOwnProperty !== Object.prototype.hasOwnProperty) {
```
lgalfaso: there is a function named `hasOwnProperty` that lives in the angular closure that should be used here

src/ng/compile.js
```diff
@@ -737,6 +737,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+      // Throw exception
+      assertNotHasOwnProperty('hasOwnProperty', 'binding');
```
lgalfaso: There is no need for this

src/ng/compile.js
```diff
@@ -2555,10 +2574,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             break;
 
           case '=':
-            if (optional && !attrs[attrName]) {
-              return;
-            }
             parentGet = $parse(attrs[attrName]);
+
+            if (parentGet === noop && optional) break;
```
lgalfaso: this is wrong as in the example `watch` is a function and `$parse(fn) === fn` for al functions `fn`

lgalfaso:
On top, a test that breaks in Firefox using the property `watch` is missing

lgalfaso:
One more thing, please rebase and squash everything into one commit

Thanks!

src/ng/compile.js
randing89: As from line 2548
```javascript
if ((attrName in attrs) && !attrs.hasOwnProperty(attrName)) {
     // User defined a binding with same as a method in Object.prototype and also the attribute is not present
    // Override property to make sure subsequent code do not have access to prototype method
    attrs[attrName] = undefined;
}
```
has set `attrs[attrName]` to `undefined` in the case `attrName` is a method name but not actually defined in attributes. It is unnecessary to check with `hasOwnProperty `.

src/ng/compile.js
randing89: In the case of `attrs['watch']` is a function, it would been set to `undefined` in line 2551. So $parse(undefined) would be not valid.

caitp:
I think we should just createMap the Attributes class' prototype so that we don't have to deal with non-standard Object.prototype properties. Unfortunately, doing that is a breaking change

lgalfaso:
> I think we should just createMap the Attributes class' prototype so that we don't have to deal with non-standard Object.prototype properties. Unfortunately, doing that is a breaking change

This would make the code simpler, but given that it is a breaking change, I would like other opinions

randing89:
I have revised the patch. I set the `attrs[attrName]` to `undefined` in the case user defined a scope binding with same name as prototype method & attribute not presenting. So that the following code will never access to the actual method function.

src/ng/compile.js
```diff
@@ -2545,9 +2545,19 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         lastValue,
         parentGet, parentSet, compare;
 
+        if ((attrName in attrs) && !attrs.hasOwnProperty(attrName)) {
```
lgalfaso: Please change this to `if (hasOwnProperty(attrs, attrName)) {`

src/ng/compile.js
lgalfaso: Does this behaves exactly as the previous code? If this is the case, please keep the code as it was before

lgalfaso:
Otherwise, LGTM

src/ng/compile.js
gkalpak: You mean `hasOwnProperty.call(attrs, attrName)`, right ?

lgalfaso: @gkalpak yep :)

randing89:
Updated. Thanks for the review

src/ng/compile.js
```diff
@@ -2545,9 +2545,19 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         lastValue,
         parentGet, parentSet, compare;
 
+        if ((attrName in attrs) && !hasOwnProperty.call(attrs, attrName)) {
```
lgalfaso: is `(attrName in attrs) && ` needed?

randing89: Yes. In the case `attrs === {}` and `attrName === 'watch'` we want to make sure code in the switch below accessing attrs['watch'] instead of attrs.prototype.watch. So this is use to find out if `attrName` is in the prototype chain but not actually defined in `attrs` object.

lgalfaso: I mean, is `if (!hasOwnProperty.call(attrs, attrName)) {` good?

randing89: Em... Tried to run the test and seems all good. I've updated the commit. Thanks.

lgalfaso:
LGTM, will merge right after we cut the next release.

lgalfaso:
landed as a6339d30d1379689da5eec9647a953f64821f8b0

PR #11635, opened 2015-04-17
fix($httpParamSerializerJQLike): Follow jQuery parameter serialization l...

### LINKED ISSUE #11551: $httpParamSerializerJQLike serializing differently than jQuery.param
There's a difference when serializing nested objects:

```js
jQuery.param({ myObject: { prop1: 5, prop2: { propA: 10 } } })
```
produces
```
myObject%5Bprop1%5D=5&myObject%5Bprop2%5D%5BpropA%5D=10
```

While `$httpParamSerializerJQLike` produces
```
myObject%5Bprop1%5D=5&myObject%5Bprop2%5D=%7B%22propA%22:10%7D
```

`ASP.NET Web API` understands `jQuery.param` but not `$httpParamSerializerJQLike`.

BTW, thanks for adding this feature, @pkozlowski-opensource! :smiley:

lgalfaso:
...ogic

Follow jQuery parameter serialization logic for nested objects.

Closes #11551

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -11,6 +11,8 @@ var JSON_PROTECTION_PREFIX = /^\)\]\}',?\n/;
 
 function paramSerializerFactory(jQueryMode) {
 
+  return jQueryMode ? jQueryLikeParamSerializer : paramSerializer;
+
   function serializeValue(v) {
     if (isObject(v)) {
       return isDate(v) ? v.toISOString() : toJson(v);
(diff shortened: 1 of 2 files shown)


test/ng/httpSpec.js
```diff
@@ -2014,6 +2014,12 @@ describe('$http param serializers', function() {
       expect(jqrSer({a: 'b', foo: {'bar': 'barv', 'baz': 'bazv'}})).toEqual('a=b&foo%5Bbar%5D=barv&foo%5Bbaz%5D=bazv');
                                                                            //a=b&foo[bar]=barv&foo[baz]=bazv
     });
+
+    it('should serialize nested objects by repeating param name with [kay] suffix', function() {
```
gkalpak: _kay_  -->  _KEY_

lgalfaso: Done

src/ng/http.js
```diff
@@ -34,7 +35,32 @@ function paramSerializerFactory(jQueryMode) {
...
 
     return parts.length > 0 ? parts.join('&') : '';
-  };
+  }
+
+  function jQueryLikeParamSerializer(params) {
+    if (!params) return '';
+    var parts = [];
+    serialize(params, '', true);
+    return parts.length > 0 ? parts.join('&') : '';
```
gkalpak: Minor, but I think this can be just `return parts.join('&');`
In `paramSerializer()` too.

lgalfaso: Done

lgalfaso:
The travis failure looks like a flake

gabrielmaldi:
I think we need to make a special case for date parameters:

```js
jQueryLikeParamSerializer({ dateFrom: new Date() }) === ""
```

src/ng/http.js
```diff
@@ -18,23 +20,47 @@ function paramSerializerFactory(jQueryMode) {
...
+    serialize(params, '', true);
+    return parts.join('&');
+
+    function serialize(toSerialize, prefix, topLevel) {
+      if (toSerialize === null || isUndefined(toSerialize)) return;
+      if (isArray(toSerialize)) {
+        forEach(toSerialize, function(value) {
+          serialize(value, prefix + '[]');
+        });
+      } else if (isObject(toSerialize)) {
```
gabrielmaldi: Maybe `} else if (isObject(toSerialize) && !isDate(toSerialize)) {`?

lgalfaso:
@gabrielmaldi you are right, fixed this case

gabrielmaldi:
I thought this would make it into `v1.4.0-rc.1` :cry:

src/ng/http.js
```diff
@@ -11,6 +11,8 @@ var JSON_PROTECTION_PREFIX = /^\)\]\}',?\n/;
 
 function paramSerializerFactory(jQueryMode) {
 
+  return jQueryMode ? jQueryLikeParamSerializer : paramSerializer;
```
petebacondarwin: If we are going to do this then we might as wel ditch the `paramSerializerFactory` altogether and just reference these two serializers directly in their respective providers.

petebacondarwin:
One comment but otherwise LGTM

src/ng/http.js
lgalfaso: They only share `serializeValue`, maybe move a little more logic to this function but keep it as is?

gabrielmaldi: It's already refactored in 2420a0a77e27b530dbb8c41319b2995eccf76791

PR #11640, opened 2015-04-17
fix($animate): make sure the JS animation lookup is an object lookup

### LINKED ISSUE #11619: ngAnimate: lookUpAnimations function breaks if CSS class is an Array.prototype function name.
Hello, I'm having a problem with the lookUpAnimations function in [src/ngAnimate/animateJs.js](https://github.com/angular/angular.js/blob/master/src/ngAnimate/animateJs.js#L236). It iterates over the CSS classes of the element and gets the [registered animation provider](https://github.com/angular/angular.js/blob/master/src/ngAnimate/animateJs.js#L241):

```javascript
// Extract from src/ngAnimate/animateJs.js#L241
var klass = classes[i],
    animationFactory = $animateProvider.$$registeredAnimations[klass];
if (animationFactory && !flagMap[klass]) {
  matches.push($injector.get(animationFactory));
  flagMap[klass] = true;
}
```
My element has a CSS class named "filter" so, when doing:
`animationFactory = $animateProvider.$$registeredAnimations[klass];`
the `animationFactory` variable ends up being a reference to `Array.prototype.filter` (because `$$registeredAnimations` [is of type Array](https://github.com/angular/angular.js/blob/master/src/ng/animate.js#L166)). After that, I got an unknown provider exception because it calls `$injector.get()` with the filter function.

I already changed the CSS class, but shouldn't $$registeredAnimations be just a plain object (like Object.create(null)) instead of an Array if it's going to be used that way?

Thoughts?

matsko:
The lookup type was an array before, but it should be an empty object.

Closes #11619

### DIFF
--- a/test/ng/animateSpec.js
+++ b/test/ng/animateSpec.js
@@ -115,6 +115,21 @@ describe("$animate", function() {
       inject();
     });
 
+    it("should register the animation and be available for lookup", function() {
+      var provider;
+      module(function($animateProvider) {
+        provider = $animateProvider;
+      });
+      inject(function() {
+        // by using hasOwnProperty we know for sure that the lookup object is an empty object
+        // instead of inhertiting properties from its original prototype.
+        expect(provider.$$registeredAnimations.hasOwnProperty).toBeFalsy();
+
+        provider.register('.filter', noop);
+        expect(provider.$$registeredAnimations['filter']).toBe('.filter-animation');
+      });
+    });
+
     it("should apply and retain inline styles on the element that is animated", inject(function($animate, $rootScope) {
       var element = jqLite('<div></div>');
       var parent = jqLite('<div></div>');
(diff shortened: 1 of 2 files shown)


test/ng/animateSpec.js
```diff
@@ -115,6 +115,18 @@ describe("$animate", function() {
       inject();
     });
 
+    it("should register the animation and be available for lookup", function() {
+      var provider;
+      module(function($animateProvider) {
+        provider = $animateProvider;
+      });
+      inject(function() {
+        expect(provider.$$registeredAnimations['filter']).toBeFalsy();
```
IgorMinar: check for hasOwnProperty instead, it's more generic and also add a comment that we are testing with a well known property name that is usually on objects or arrays, we must ensure that we don't create collisions with them

matsko:
Merged as https://github.com/angular/angular.js/commit/103a39ca8dad0300bead15c358aad846510b2229

PR #11644, opened 2015-04-18
fix(ngMessageFormat): ensure bindings are valid for Protractor

petebacondarwin:
Yes, yes, I know. This PR requires unit tests, etc, etc.
Just pushing here for @chirayuk to take care of. Thanks @chirayuk !

### DIFF
--- a/src/ngMessageFormat/messageFormatSelector.js
+++ b/src/ngMessageFormat/messageFormatSelector.js
@@ -24,6 +24,7 @@ function MessageSelectorBase(expressionFn, choices) {
   this.parsedFn['$$watchDelegate'] = function $$watchDelegate(scope, listener, objectEquality) {
     return self.watchDelegate(scope, listener, objectEquality);
   };
+  this.parsedFn.exp = expressionFn.exp;
 }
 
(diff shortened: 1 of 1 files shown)


petebacondarwin:
To recreate the issue we just need to have an ICU plural message interpolation, the run a protractor spec that tries to find an element by a binding....

```html
      {{ app.filteredRestaurants.length, plural,
        =0 {No restaurants found!}
        =1 {Only one restaurant found!}
        other {# restaurants found.}
      }}
```

```js
    var userNameDisplay = element(by.binding('app.user.name'));
```

The call to `by.binding('...')` triggers a call in angular.js of `testability.findBindings` which tries to find the expression in the binding name:

```
if (bindingName.indexOf(expression) != -1) {
```

But without this fix we have a binding that consists of an array with on `undefined` item.

chirayuk:
Thanks. I'll work on this PR in the next hour or so!

chirayuk:
I'm waiting for Travis status.  However, I think this one is good to merge once it's confirmed to pass on Travis.  See #11649 which has the changes to the test files and some symbol renaming protection.  (Please merge that one if I go to bed before I'm able to merge this.)

PR #11646, opened 2015-04-19
docs(ngJq): fix directive usage

cexbrayat:
Should fix the usage example, see https://docs.angularjs.org/api/ng/directive/ngJq


Narretz:
Landed: https://github.com/angular/angular.js/commit/03d4bbc16f620d13ac509ab93cabbc77bbfc3f59

PR #11648, opened 2015-04-19
Note about required quotes for dashed class names

aschriner:
This is a nasty hidden gotcha.  http://stackoverflow.com/a/15582433/710394
Any time someone says, "After hours of hacking around, it turns out..." it's probably a good candidate for inclusion in the docs.


aschriner:
I signed it!

lgalfaso:
This works just like JSON object literal syntax. Ie `{icon-white: someBooleanValue}` is an invalid JSON definition...

Narretz:
The docs were changed to contain a hyphenated class some time ago; explictly expainling this is overkill, it's not an angular specific thing.

PR #11673, opened 2015-04-21
feat(compile): add 'directives' property to ddo to preload templates

shahata:
this can be useful when you have many nested directives and you want to download all of the required templates in parallel instead of downloading the next template serially only after each downloaded template is compiled.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1662,6 +1662,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         var attrStart = directive.$$start;
         var attrEnd = directive.$$end;
 
+        downloadTemplateDependencies(directive);
+
         // collect multiblock sections
         if (attrStart) {
           $compileNode = groupScan(compileNode, attrStart, attrEnd);
@@ -2167,6 +2169,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           dstAttr[key] = srcAttr[key];
         }
       });
+    }
+
+
+    function downloadTemplateDependencies(directive) {
+      if (directive.directives) {
+        directive.directives.forEach(function(name) {
+          $injector.get(name + Suffix).forEach(function(dependency) {
+            downloadTemplateDependencies(dependency);
+            if (isString(dependency.templateUrl)) {
+              $templateRequest($sce.getTrustedResourceUrl(dependency.templateUrl));
+            }
+          });
+        });
+      }
     }
 
 
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -1170,6 +1170,79 @@ describe('$compile', function() {
         }));
       });
 
+      describe('directives', function() {
+        beforeEach(module(function() {
+          directive('grandParentFn', valueFn({
+            restrict: 'E',
+            directives: ['parentFn'],
+            templateUrl: 'grand-parent.html'
+          }));
+          directive('parentFn', valueFn({
+            restrict: 'E',
+            directives: ['child'],
+            templateUrl: function() {
+              return 'parent.html';
+            }
+          }));
+          directive('grandParent', valueFn({
+            restrict: 'E',
+            directives: ['parent'],
+            templateUrl: 'grand-parent.html'
+          }));
+          directive('parent', valueFn({
+            restrict: 'E',
+            directives: ['child'],
+            templateUrl: 'parent.html'
+          }));
+          directive('child', valueFn({
+            restrict: 'E',
+            templateUrl: 'child.html'
+          }));
+        }));
+
+
+        it('should download templateUrl of nested directives immediately', function() {
+          inject(function($compile, $httpBackend, $rootScope) {
+            $httpBackend.expect('GET', 'child.html').respond('World');
+            $httpBackend.expect('GET', 'parent.html').respond('Hello <child/>');
+
+            element = $compile('<parent/>')($rootScope);
+            $httpBackend.flush();
+            expect(sortedHtml(element)).toEqual('<parent>Hello <child>World</child></parent>');
+          });
+        });
+
+
+        it('should download templateUrl of multi-nested directives immediately', function() {
+          inject(function($compile, $httpBackend, $rootScope) {
+            $httpBackend.expect('GET', 'child.html').respond('World');
+            $httpBackend.expect('GET', 'parent.html').respond('Hello <child/>');
+            $httpBackend.expect('GET', 'grand-parent.html').respond('Hello <parent/>');
+
+            element = $compile('<grand-parent/>')($rootScope);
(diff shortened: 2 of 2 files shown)


shahata:
This is just a POC in order to see what people think about it. In general the idea is that if directive a contains directive b which contains directive c, we want to download a.html, b.html and c.html in parallel and not one after the other. In order to achieve this, each directive declares in its DDO what directives it depends on and then we know to download their templates before the directive's template is downloaded.

test/ng/compileSpec.js
```diff
@@ -1170,6 +1170,76 @@ describe('$compile', function() {
...
+
+            element = $compile('<grand-parent/>')($rootScope);
+            $httpBackend.flush();
+            expect(sortedHtml(element)).toEqual(
+              '<grand-parent>Hello <parent>Hello <child>World</child></parent></grand-parent>');
+          });
+        });
+
+
+        it('should not download nested templateUrl functions', function() {
```
gkalpak: How are we testing this ? `parent.HTML` seems to still be fetched :confused:

shahata: It is fetched immediately after `grand-parent.html` is compiled. I've modified the test to make this more clear.

gkalpak: Since the templates can be fetched either by `downloadTemplateDependencies()` or via the "normal" $compiling process, how are the tests above testing `downloadTemplateDependencies()` ?
AFAICT, they should work even without the change, no ?

shahata: It is not possible that `child.html` will be requested before `parent.html` via normal `$compile` process.

shahata: I've considered making the test more clear by verifying that `$httpBackend.verifyNoOutstandingRequest();` throws within the response handlers, but that seemed to much. I'm more interested in feedback on the actual feature :)

gkalpak: Sorry, I didn't realize it was the order of the requests that's being verified here. Feel free to revert to the former (less explicit) state.

gkalpak:
I am still not sure what the objective of this feature is. Just initiating a request sooner doesn't mean we are going to get a response sooner. So, I guess the point is to start downloading deep nested templates sooner rather than later, but is it really worth it ?

I am not sure it will make that much of a difference (I might be wrong though).
Besides, if someone is _really_ concerned about template download speed, wouldn't they inline the templates instead (either manually or as part of their build-process) ?

pkozlowski-opensource:
Yeh, I'm also not sold on this one... As @gkalpak commented, it won't necessarily speed up things (well, it might, since templates might be downloaded eagerly in parallel) but will reduce flickering. Having said this I don't think that additional complexity is worth it - I would argue that one needs directives templates into `$templateCache` to avoid delays / flickering. This seems to me a much better and simpler mechanism.

@shahata do you have a use-case where preloading things into `$templateCache` doesn't work for you?

shahata:
I'll try to describe the objective in more detail:

Usually you should inline templates only if they are immediately visible to the user when navigating to your application, thus saving a lot of precious page load time until your application is visible. You shouldn't, however, inline ALL of your templates since many of the screen in your application are not immediately visible to the user, thus loading them on page load is a waste of precious time.

Now, let's say your has loaded and now the user clicks some button to display some modal. This modal contains 10 nested directives (which were not inlined since they are not part of the initial page view, they are inside some modal). This means you need to wait for 10 templates download and since they are nested, they are downloaded serially, so it is going to be slow. With this mechanism, you are going to download the templates in parallel, which at best case means it might take 10% of the time to render.

pkozlowski-opensource:
I see. However don't think I can agree with this statement:
> Usually you should inline templates only if they are immediately visible to the user

I would argue that one should pre-load all the "small" templates for "commonly" used directives. For example, I don't want users to wait for download of a, say, datepicker template while opening a datepicker for the very first time. The negative user experience here is worse than having some memory / download time consumed initially. 

Now,  "small" and "commonly" might mean different things for different applications, so if we move to the model where everything is a directive (like in ng2) the approach of pre-loading everything becomes impractical. 

In short: I see the use case, but I'm still not sold on the solution. But at the same time can't think of anything better atm :-(

shahata:
Yeah, this is not an exact science, there are obviously more reasons to preload templates. I've came to this proposal after seeing many apps buying into the "everything is a component" approach and becoming worried about their response time when navigating inside the app. In ng2 you are required to declare the directives you depend on, so tis is how I got the idea to declare them in 1.x as well and use this information to optimize template load time.

src/ng/compile.js
```diff
@@ -2170,6 +2172,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     }
 
 
+    function downloadTemplateDependencies(directive) {
```
lgalfaso: It looks like this will break if you have two directives and each depend on the other.

src/ng/compile.js
```diff
@@ -2170,6 +2172,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+      if (directive.directives) {
+        directive.directives.forEach(function(name) {
```
lgalfaso: the name `directives` sounds too generic... but this can be changed if this is moving forward

src/ng/compile.js
```diff
@@ -2170,6 +2172,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+          $injector.get(name + Suffix).forEach(function(dependency) {
+            downloadTemplateDependencies(dependency);
```
lgalfaso: traversing the graph without knowing what was already traversed, can get quite expensive fast

shahata: just a POC :)

src/ng/compile.js
shahata: yeah, I borrowed the name from ng2 though, so that's why I like it.

src/ng/compile.js
shahata: you mean it would end up in an infinite loop, right? you are right, but let's discuss a bit if we want this feature before I put more effort into this.

lgalfaso:
If we are talking about *just* the optimization to load in advance some templates, then I lean towards not landing this. On the other hand, if we are talking about something else (eg: limit the directive template to the directives specified), then that would be something worth exploring. That said, it is clear that the later would be a lot more work, so only someone with a good amount of time in their hands should look into what is needed for that to happen.

Narretz:
In 1.6.7, it'll be possible to officially add modules to the injector after initial app bootstrap: https://github.com/angular/angular.js/commit/34237f929927295392bbb1a600e78cbde581839a. One use case of this is to break up the app into separate parts (based on routing for example), which means the main use case of this PR - to defer the loading of templates until needed while at the same time loading them parallel - can be handled at the module level with templateUrl to $templateCache solutions or already inlined templates.

PR #11688, opened 2015-04-23
fix($compile): Better error message on empty template

SomeKittens:
If a user ends up trying to call $compile with an empty string, they will get an error from jqLite
This commit introduces a much more accurate error to aid in debugging.

Plunker: http://plnkr.co/edit/YNxDVQNMTsiNV3WhmjcT?p=preview

### DIFF
--- a/docs/content/error/$compile/tplrt.ngdoc
+++ b/docs/content/error/$compile/tplrt.ngdoc
@@ -51,3 +51,8 @@ well.  Consider the following template:
 
 The `<!-- container -->` comment is interpreted as a second root element and causes the template to
 be invalid.
+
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Hi @SomeKittens 
Thanks for this PR. While it is true that passing an empty string to `$compile` is an error, it is not that templates cannot be empty strings.  As you can see in this plunker: http://plnkr.co/edit/m4AL8BofE5sisMsFzKq7?p=preview

For us to merge this PR, we would need a different error and also a unit test to demonstrate that it is working correctly.

Narretz:
Closing because of inactivity.

PR #11689, opened 2015-04-23
Materialize

eros2123:
vxbn


jkurz:
Is this something the team is still trying to achieve? Or is it not a priority anymore? If it is, then this PR should probably be closed and the branch should be given some more love.

lgalfaso:
In this state, it looks impossible to understand what the fix is. Please rebase this or just open a new PR with the fix

Issue #11694, opened 2015-04-23
Unwatch a particular $swipe event


matthieusieben:
I am using the $swipe service to implement a drag and drop functionality. But in the case there is nothing under the mouse when the start event is triggered, I am not interested in subsequent move & end events. This is no big deal since a can keep a variable to determine if an element is currently being dragged.

However, when I try to select some text, the $swipe listeners are triggered, which means that the event gets [defaultPrevented](https://github.com/angular/angular.js/blob/master/src/ngTouch/swipe.js#L153), preventing me to select my text.

It would be nice if we could either:
- cancel the calls to the listeners (up until next 'start') if the start listeners returns false.
- not call automatically event.preventDefault(); and let the callback do it instead.

Narretz:
I'm closing this issue since the ngTouch module will be deprecated in AngularJS 1.7.0, and we will not do any more work on it. This is because touch handling is a complex topic, and there are stand-alone libraries that are better equipped to handle all the nuances and edge cases. For example, https://hammerjs.github.io/, which is also used by Angular.

matthieusieben:
- [x] Wait long enough for the feature to be deprecated
 - [x] Mark issue as resovled

PR #11695, opened 2015-04-23
$swipe: Make a swipe event inactive from the start callback.

matthieusieben:
Implementation proposition for issue #11694


matthieusieben:
I signed it!

PR #11698, opened 2015-04-23
$swipe: Make a swipe event inactive from the start callback.

matthieusieben:
Implementation proposition for issue #11694


solispauwels:
+1

Narretz:
I'm closing this PR since the ngTouch module will be deprecated in AngularJS 1.7.0, and we will not do any more work on it. This is because touch handling is a complex topic, and there are stand-alone libraries that are better equipped to handle all the nuances and edge cases. For example, https://hammerjs.github.io/, which is also used by Angular.

PR #11699, opened 2015-04-23
docs($httpBackend): compare $httpBackend API to $http API

ciwchris:
Documentation states $httpBackend has the same API as $httpBackend, it compares the API to itself. This change compares the API to $http instead.


pkozlowski-opensource:
I don't think it is correct. What is compared here is the API of the _mock_ `$httpBackend` vs. `$httpBackend` which should be identical. So we are talking about 2 services with the same name, but from the different modules.

ciwchris:
Okay, I think I understand. This can be closed.

PR #11701, opened 2015-04-23
docs(error/nonassign): add optional binding example

kentcdodds:


kentcdodds:
Thought this would be useful information :-)

PR #11702, opened 2015-04-23
feat(directives.js): default replace true

PatrickJS:
better for animations/css/seo


Narretz:
This is a huge breaking change and replace is actually deprecated. This can't be done.

PR #11704, opened 2015-04-23
Fixes for ngAnimate to get it work for RC1

matsko:


src/ngAnimate/module.js
```diff
@@ -392,6 +392,93 @@
...
+ * ## Animation Anchoring (via `ng-animate-ref`)
+ *
+ * ngAnimate in AngularJS 1.4 comes packed with the ability to cross-animate elements between
+ * structural areas of an application (like views) by pairing up elements using an attribute
+ * called `ng-animate-ref`.
+ *
+ * Let's say for example we have two views that are managed by `ng-view` and we want to identify
+ * that there is a relationship between two components situated in different views. By using the
+ * `ng-animate-ref` attribute we can identify that the two components are paired together and we
+ * can then attach an animation between then which is triggered when the view changes.
```
jkurz: ',between them,' or maybe just 'an animation, which is'

src/ngAnimate/module.js
```diff
@@ -392,6 +392,93 @@
...
+ *
+ * ### How is the morphing handled?
+ *
+ * CSS Anchoring relies on transitions and keyframes and the internal code is intelligent enough to figure out
+ * what CSS classes differ between the starting element and the destination element. These different CSS classes
+ * will be added/removed on the anchor element and a transition will be applied (the transition that is provided
+ * in the anchor class). Long story short, ngAnimate will figure out what classes to add and remove which will
+ * make the transition of the element as smooth and automatic as possible. Be sure to use simple CSS classes that
+ * do not rely on DOM nesting structure so that the anchor element appears the same as the starting element (since
+ * the cloned element is placed inside of body element).
```
gkalpak: In the previous section, it says the element is placed in the root element of the application. Which one is it ?

src/ngAnimate/module.js
```diff
@@ -392,6 +392,93 @@
...
+ *
+ * ### How is the element transported?
+ *
+ * When an anchor animation occurs, ngAnimate will clone the starting element and position it exactly where the starting
+ * element is located on screen via absolute positioning. The cloned element will be placed inside of the root element
+ * of the application (where ng-app was defined) and all of the CSS classes of the starting element will be applied. The
+ * element will then animate into the `out` and `in` animations and will eventually reach the coordinates and match
+ * the dimensions of the destination element. During the entire animation a CSS class of `.ng-animate-shim` will be applied
+ * to both the starting and destination elements in order to hide them from being visible (the CSS styling for the class
+ * is: `visibility:none`). Once the anchor reaches its destination then it will be removed and the destination element
```
gkalpak: `visibility: hidden`

src/ngAnimate/module.js
petebacondarwin: Or "from one to the other"?

src/ngAnimate/module.js
```diff
@@ -392,6 +392,93 @@
  *
  * To learn more about what's possible be sure to visit the {@link ngAnimate.$animateCss $animateCss service}.
  *
+ * ## Animation Anchoring (via `ng-animate-ref`)
+ *
+ * ngAnimate in AngularJS 1.4 comes packed with the ability to cross-animate elements between
+ * structural areas of an application (like views) by pairing up elements using an attribute
+ * called `ng-animate-ref`.
+ *
+ * Let's say for example we have two views that are managed by `ng-view` and we want to identify
```
petebacondarwin: identify -> show

src/ngAnimate/module.js
```diff
@@ -392,6 +392,93 @@
...
+ * <img src="./banner.jpg" ng-animate-ref="banner" class="banner-animation">
+ * ```
+ *
+ * Now, when the view changes (once the link is clicked), ngAnimate will examine the
+ * HTML contents to see if there is a match reference between any components in the view
+ * that is leaving and the view that is entering. It will then attempt to trigger a CSS
+ * animation on the `.banner-animation-anchor` CSS class (notice how `.banner-animation` is
+ * a shared CSS class on both elements containing the `ng-animate-ref` pairing).
+ *
+ * So the two images match since they share the same ref value. They also share a similar CSS
```
petebacondarwin: "So the two images match since they share the same ref value." is not a valid sentence but also does not make sense to me.  Should we just drop the "So"?

matsko:
Merged as https://github.com/angular/angular.js/commit/d97b6587973aac4249b43d9d76fb6d61fcccdf19

matsko:
Thanks everyone for the reviewing.

PR #11720, opened 2015-04-24
fix(merge): treat dates as atomic values instead of objects.

gabrielmaldi:
Makes `angular.merge` copy dates correctly.

Before and after this change:

image

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -352,8 +352,12 @@ function baseExtend(dst, objs, deep) {
       var src = obj[key];
 
       if (deep && isObject(src)) {
-        if (!isObject(dst[key])) dst[key] = isArray(src) ? [] : {};
-        baseExtend(dst[key], [src], true);
+        if (isDate(src)) {
+          dst[key] = new Date(src.valueOf());
+        } else {
+          if (!isObject(dst[key])) dst[key] = isArray(src) ? [] : {};
+          baseExtend(dst[key], [src], true);
+        }
       } else {
         dst[key] = src;
       }
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -481,6 +481,16 @@ describe('angular', function() {
       // make sure we retain the old key
       expect(hashKey(dst)).toEqual(h);
     });
+
+
+    it('should copy dates by reference', function() {
+      var src = { date: new Date() };
+      var dst = {};
+
+      extend(dst, src);
+
+      expect(dst.date).toBe(src.date);
+    });
   });
 
 
@@ -547,6 +557,18 @@ describe('angular', function() {
         foo: [1,2,3]
       });
       expect(dst.foo).not.toBe(src.foo);
+    });
+
+
+    it('should copy dates by value', function() {
+      var src = { date: new Date() };
+      var dst = {};
+
+      merge(dst, src);
+
+      expect(dst.date).not.toBe(src.date);
+      expect(isDate(dst.date)).toBeTruthy();
+      expect(dst.date.valueOf()).toEqual(src.date.valueOf());
     });
   });
 


gkalpak:
On the docs it sys that `merge()` _"recursively descends into object properties of source objects, performing a deep copy"_. That is so not true :(
It performs a "deep extend".

In that sense, this PR seems to be doing the right thing.

gabrielmaldi:
Perhaps we should also include something like this (line 341):
```js
dst[key] = isDate(src) ? new Date(src.valueOf()) : src;
```
So that dates get cloned instead of assigning a reference to the same object.

src/Angular.js
```diff
@@ -334,11 +334,11 @@ function baseExtend(dst, objs, deep) {
       var key = keys[j];
       var src = obj[key];
 
-      if (deep && isObject(src)) {
+      if (deep && isObject(src) && !isDate(src)) {
         if (!isObject(dst[key])) dst[key] = isArray(src) ? [] : {};
         baseExtend(dst[key], [src], true);
       } else {
-        dst[key] = src;
+        dst[key] = isDate(src) ? new Date(src.valueOf()) : src;
```
gkalpak: I don't think this is the right think to do. It will also apply to non-deep copying, where I believe it would be better to just assign the same Date object.

It's OK fr deep extending I guess, so maybe something like:

```js
       if (deep && isObject(src)) {
         if (isDate(src)) {
           dst[key] = new Date(src.valueOf());
         } else {
           if (!isObject(dst[key])) dst[key] = isArray(src) ? [] : {};
           baseExtend(dst[key], [src], true);
         }
       } else {
         dst[key] = src;
       }
```

gabrielmaldi: I agree! Done and thanks for the snippet.

gkalpak:
The implementation lgtm.  But we should have some tests in place, in order to ensure there will be no accidental regressions in the future.

@gabrielmaldi, could add a couple of tests (i.e. ones that would fail **without** this fix and pass **with** it) ?

gabrielmaldi:
> @gabrielmaldi, could add a test (i.e. ones that would fail without this fix and pass with it) ?

Sure, I'll work on it later today.

gabrielmaldi:
I wrote a couple of tests but didn't push them because one is failing and I believe it's an issue with the version of `Jasmine` which `Angular` is using.

Here's a [gist](https://gist.github.com/gabrielmaldi/cb8af7f2b1e5c2c79fbf#file-angularspec-js-L40) with the tests; it fails on the last line with this error:

screen shot 2015-05-02 at 3 54 45 pm

I wrote this [pen](http://codepen.io/anon/pen/QbbZRo) which uses `Jasmine 2.0.0` and `Angular 1.4.0-rc.1` and the test passes just fine.

I also tried using:
```js
var dst2 = {};
brokenMerge(dst2, src);
expect(dst2).not.toEqual({
  foo: date
});
```
but it fails with the same error:

screen shot 2015-05-02 at 4 07 20 pm

Thanks for your help.

gkalpak:
Angular is using jamsine 1.3. In this [pen][1], I have a couple of tests that are currently failing.
I would expect them to pass with your fix. (If that is indeed the case, you can just use those two (maybe update ther descriptions if necessary).)
<sub>
In case it wasn't clear, you don't have a `brokenMerge` in your testcase.
It suffices to have tests that **would** have failed, were they run prior to the fix and pass **with** the fix.
</sub> 

[1]: http://codepen.io/ExpertSystem/pen/qddQrX

gabrielmaldi:
Yes, your tests pass with the fix. I guess `Jasmine 1.3` behaves differently and using `valueOf` to compare dates is a valid workaround. Thanks for your you guidance.

PR #11722, opened 2015-04-25
fix(ngAnimate): Fix text node animation

### LINKED ISSUE #11703: (bug) ngAnimate element.parent is not a function
I'm using version 1.4.0-rc.0 of ngAnimate and there's an issue when `$animate.leave` is called with a `TextNode` as an argument. Here's the stack trace:

```
TypeError: element.parent is not a function
    at queueAnimation (https://code.angularjs.org/1.4.0-rc.0/angular-animate.js:1991:28)
    at Object.push (https://code.angularjs.org/1.4.0-rc.0/angular-animate.js:1943:16)
    at Object.leave (https://code.angularjs.org/1.4.0-rc.0/angular.js:5017:31)
    at update (https://rawgit.com/rubenv/angular-gettext/master/dist/angular-gettext.js:198:34)
    at post (https://rawgit.com/rubenv/angular-gettext/master/dist/angular-gettext.js:207:21)
    at invokeLinkFn (https://code.angularjs.org/1.4.0-rc.0/angular.js:8416:9)
    at nodeLinkFn (https://code.angularjs.org/1.4.0-rc.0/angular.js:7925:11)
    at compositeLinkFn (https://code.angularjs.org/1.4.0-rc.0/angular.js:7326:13)
    at nodeLinkFn (https://code.angularjs.org/1.4.0-rc.0/angular.js:7920:24)
    at compositeLinkFn (https://code.angularjs.org/1.4.0-rc.0/angular.js:7326:13)
```

I've managed to isolate the issue in a [Plunker](http://plnkr.co/edit/wSTszGVuuwxZHXL0xsPA?p=preview) and it appears to be caused by [this line](https://github.com/angular/angular.js/blob/master/src/ngAnimate/shared.js#L84) which returns a boolean (usually `false`) to the `queueAnimation` function. It then tries to call the `parent` method and throws the error.

In the Plunker posted above, the H1's text gets passed to the `leave` method by the `translate` directive. If the text is wrapped in another element, the error is gone. For example:

```html
<h1 translate>hi</h1> <!-- breaks -->
<h1 translate> <!-- works -->
  <span>hi</span>
</h1> 
```

EDIT: Here's a better isolated example of the same issue: https://jsfiddle.net/pLvmf9qg/

circlingthesun:
Change to comment stripping code to return text nodes in addition to element nodes

Closes #11703

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -2,6 +2,7 @@
 
 var $animateMinErr = minErr('$animate');
 var ELEMENT_NODE = 1;
+var TEXT_NODE = 3;
 
 function mergeClasses(a,b) {
   if (!a && !b) return '';
@@ -12,10 +13,10 @@ function mergeClasses(a,b) {
   return a + ' ' + b;
 }
 
-function extractElementNode(element) {
+function extractNonCommentNode(element) {
   for (var i = 0; i < element.length; i++) {
     var elm = element[i];
-    if (elm.nodeType === ELEMENT_NODE) {
+    if (elm.nodeType === ELEMENT_NODE || elm.nodeType === TEXT_NODE) {
       return elm;
     }
   }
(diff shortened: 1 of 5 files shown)


circlingthesun:
This PR seems a bit buggy still. Please don't merge it.

circlingthesun:
It appears that the unresolved issues with this PR was a result of the library that I was testing it against. It resulted in some early remove animation not taking affect. I have submitted a PR to fix said library at https://github.com/rubenv/angular-gettext/pull/187. This PR should be fine to merge.

circlingthesun:
Fixed by #11803

PR #11726, opened 2015-04-25
docs(tutorial/7 - Routing): Remove an extra "/"

b0ri5:


pkozlowski-opensource:
@b0ri5 why? Is it causing any problems for you?

b0ri5:
Sorry for the late reply, but yes, the documentation as is does not work:
  app/index.html/#/phones and
  app/#/phones both work
but
  app/index.html/#/phones
as the documentation has it, does not.

This should be easy to repo. In fact, the test case does not have the extra "/" after "index.html".

gkalpak:
Closed with c2033d7.

PR #11734, opened 2015-04-27
feat($injector): add strictDi property to $injector instance

### LINKED ISSUE #11728: Add isStrictDi to $injector
I'm writing a utility library where I have a use case to know when strict DI is enabled on a given `$injector`. I could really use some reasonable way of knowing without having to do something crazy like:

```javascript
function isStrictDi($injector) {
  try {
    $injector.invoke(function(a) {});
    return false;
  } catch(e) {
    return true;
  }
}
```

I'm thinking that we could just add a line that says `instanceInjector.strictDi = strictDi` right [here](https://github.com/angular/angular.js/blob/0baa17a3b7ad2b242df2b277b81cebdf75b04287/src/auto/injector.js#L649) before the `instanceInjector` is returned.

I would happily open a PR with tests which would do this. Just want to make sure this is the correct approach.

kentcdodds:
Add a strictDi property which is true or false when creating an instance of an injector.

Closes #11728

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -646,6 +646,8 @@ function createInjector(modulesToLoad, strictDi) {
 
   forEach(loadModules(modulesToLoad), function(fn) { instanceInjector.invoke(fn || noop); });
 
+  instanceInjector.strictDi = strictDi;
+
   return instanceInjector;
 
   ////////////////////////////////////
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -116,6 +116,10 @@ describe('injector', function() {
     expect($injector).not.toBe(providerInjector);
   }));
 
+  it('should have an false strictDi property', inject(function($injector) {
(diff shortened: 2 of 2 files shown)


kentcdodds:
Just doing some clean-up. This will obviously never get merged (because it's not really needed). Thanks!

lgalfaso:
@kentcdodds if you look at #12446 then that posted a very reasonable reason to merge this. Anyhow, I would like to consider this for 1.5. Can you put the PR back?

kentcdodds:
Oh, cool. Didn't notice that. Sure thing :-) I've made several PRs to angular. But if this one gets merged, it'll be the first time I've contributed code to angular that actually runs in people's code :smile:

lgalfaso:
In a few days we should branch for 1.5, will merge this just after that

kentcdodds:
Yay! Thanks! https://twitter.com/kentcdodds/status/645610842814181376

PR #11735, opened 2015-04-27
$animate callbacks using DOM queries

matsko:
This patch fixes ngView to inject `$element` and also patches `$animate.on` and `$animate.off` to provide support for triggering callbacks based on a DOM-query.

All of this is to make using animation callbacks within controllers easier.

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -278,20 +278,29 @@ var $AnimateProvider = ['$provide', function($provide) {
        * @name $animate#on
        * @kind function
        * @description Sets up an event listener to fire whenever the animation event (enter, leave, move, etc...)
-       *    has fired on the given element or among any of its children. Once the listener is fired, the provided callback
-       *    is fired with the following params:
+       *    has fired on the given element or among any of its children or within a matched query selector.
+       *    Once the listener is fired, the provided callback is fired. The example code below shows that in action:
        *
        * ```js
-       * $animate.on('enter', container,
+       * // using a DOM container
+       * $animate.on('enter', containerNode,
        *    function callback(element, phase) {
        *      // cool we detected an enter animation within the container
+       *    }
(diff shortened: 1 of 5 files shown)


matsko:
@petebacondarwin @IgorMinar this code works, but there is a potential leak where if a controller is destructed then the associated callback function is never removed unless element.off is called.

Perhaps we should document that you always MUST remember to deregister the callback?

gkalpak:
> All of this is to make using animation callbacks within controllers easier.

I am frightened by this line :fearful: 
I hopw you mean directive controllers, right ? Right ?

petebacondarwin:
@gkalpak I think we are talking about route controllers here, right?

@matsko could you not attach an event handler to the jqLite '$destroy' event on each DOM element that has an animation callback attached? This handler would remove all events on that element. This would prevent the leak since the element would be destroyed alongside the controller.

gkalpak:
Hm...I thought accessing the DOM (even indirectly) from within a controller was an anti-pattern (unless it was a directive, which again is debatable if the DOM should be accessed from the controller or the linking functions).
Of course with the new router/ng2 each route/view will be a components (aka directive), but currently I am skeptical about suggesting having animations in route controllers.

--
Regarding the auto-$destroy suggestion, I don't think it's a fool-proof solution, since the controller might be animating elements out-living it. But it's definitely a step in the right direction.

Narretz:
I am also not really sure we need animate callbacks in route controllers. @matsko is this something that was requested by someone?

petebacondarwin:
I tend to agree with @Narretz here. What is the use case for handling animation callbacks inside a route controller? Surely this could be done by applying a new directive either inside the template or on the ng-view element?

petebacondarwin:
Moving to post 1.4.0

PR #11743, opened 2015-04-27
fix(ngOptions): use watchCollection not deep watch of ngModel

petebacondarwin:


lgalfaso:
Tests comments, code comments and documentation need to be updated to reflect the change. Otherwise, LGTM

petebacondarwin:
@lgalfaso I added tests and docs. Can you review?

gkalpak:
After a superficial look it lgtm, but I wonder if this qualifies as a breaking change.
I find it unlikely that someone was relying on having the view reflect a change in a deeper level of the selected item, but it a breaking change notice might be inplace (just in case).

--
I don't know why I am mentoning it; I probably shouldn't, but I just have to get it out (please ignore :smile:):

```js
// A few (insignificant) lines of (test) code can be saved and 
// readability can be marginally imrpoved (imo), by replacing occurrences of:
scope.$apply(function() {
  scope.x = y;
});

// with:
scope.$apply('x = y');
```

petebacondarwin:
@gkalpak - yeah I know what you mean about the BC message. I was trying to get away without one. But I guess it doesn't do any harm to highlight this.

@gkalpak - yeah I know what you are saying about $apply expressions and I tend to fluctuate back and forth between the two styles. On one hand I want to be consistent and on the other concise. The former is tricky when there are occasions when you cannot express what you want to change on the scope in an Angular expression.

petebacondarwin:
In 1.3.15 it totally doesn't work if you update deepter properties on the model so this is not a breaking change, IMO

Compare 1.3.15:  http://plnkr.co/edit/zsgnhflQ3M1ClUSrsne0?p=preview
to snapshot: http://plnkr.co/edit/hI48vBc0GscyYTYucJ0U?p=preview

gkalpak:
I was referring to a breaking change since 1.4.0-rc.1 (for people upgrading from that version to 1.4.0-rc.2 or later). But it's a minor one anyway.

test/ng/directive/ngOptionsSpec.js
```diff
@@ -901,8 +949,7 @@ describe('ngOptions', function() {
 
       // Update the properties on the object in the selected array, rather than replacing the whole object
```
lgalfaso: I think this comment should be updated.

lgalfaso:
Just two small comments. Otherwise, LGTM

test/ng/directive/ngOptionsSpec.js
petebacondarwin: @lgalfaso - looking at this a little more I am now thinking that in the specific situation of **multiple** + **track by** we should handle the value of the **track by** changing within each of the items in the collection of selected values.

PR #11745, opened 2015-04-28
docs($http): Adds detail about param sorting.

wesalvaro:
Parameters are sorted when added to the URL:
https://github.com/angular/angular.js/blob/c8700f04fb6fb5dc21ac24de8665c0476d6db5ef/src/ng/http.js#L24
This is a very important detail, especially for testing.


Narretz:
Actually, this behavior isn't even tested. We should probably test this, too.

pkozlowski-opensource:
And with the custom param serializers we should probably move this doc to a serializer itself...

PR #11755, opened 2015-04-29
fix(ngAnimate): close follow-up class-based animations when the same class is added/removed when removed/added

### LINKED ISSUE #11717: ngShow gives inconsistent results with ngAnimate
`ng-show` gives inconsistent results when the watched valued is changed multiple time in the same digest cycle and `ngAnimate` is included.

Here is a simple reproduction with `1.4.0-rc.0`: http://jsfiddle.net/dubejf/xz6ppj9v/2/

- The controller has a flag, `vm.showIt`.
- This flag controls a red div using a `ng-show` directive, with `ng-show="vm.showIt"`.
- Clicking on the button toggles the flag, twice. First to set it to `true`, then to set it to `false`. The value are changed using a `$timeout`, in the hope that the `ng-show` watch will update twice in the same digest.

Expected behavior: Click on 'Click me'. At the end of the digest, `vm.showIt` should always be `false`. The red div should always be invisible.

Actual behavior: `vm.showIt` is always `false` at the end of the cycle (good). The red div is sometime visible, sometime invisible (bad).

The behavior is easier to observe in Firefox r37, where the fiddle gives the wrong results half the time. The bad behavior can also be observed in Chrome r42, with less consistency - I haven't been able to figure the pattern, you may need a few dozen push of the button to make it fail.

When the test fails (red div visible), the class 'ng-hide' is missing on the div. The class should be set, because the condition for `ng-show` is `false`.

The root of the problem appears to be the handling of classes inside the `$animate` service, specifically when successive calls to `removeClass` \ `addClass` are made.

The problem only occurs when the `ngAnimate` module is included. Tested with `1.4.0-rc.0` and `1.4.0-build.3988`.

matsko:
This patch ensures that if the same CSS class is added/removed within a
follow-up digest then the previous class-based animation is cancelled
beforehand.

Closes #11717

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -876,6 +876,51 @@ describe("animations", function() {
         expect(enterComplete).toBe(true);
       }));
 
+      it('should cancel the previously running removeClass animation if a follow-up addClass animation is using the same class value',
+        inject(function($animate, $rootScope, $$rAF) {
+
+        parent.append(element);
+        var runner = $animate.addClass(element, 'active-class');
+        $rootScope.$digest();
+
+        var closed = false;
+        runner.done(function(status) {
+          closed = true;
+        });
+
+        $$rAF.flush();
+
+        expect(closed).toBe(false);
+
+        $animate.removeClass(element, 'active-class');
+        $rootScope.$digest();
+
+        expect(closed).toBe(true);
+      }));
+
+      it('should cancel the previously running addClass animation if a follow-up removeClass animation is using the same class value',
+        inject(function($animate, $rootScope, $$rAF) {
+
+        element.addClass('active-class');
+        parent.append(element);
+        var runner = $animate.removeClass(element, 'active-class');
+        $rootScope.$digest();
+
+        var closed = false;
+        runner.done(function(status) {
+          closed = true;
+        });
+
+        $$rAF.flush();
+
+        expect(closed).toBe(false);
+
+        $animate.addClass(element, 'active-class');
+        $rootScope.$digest();
+
+        expect(closed).toBe(true);
+      }));
+
       it('should skip the class-based animation entirely if there is an active structural animation',
         inject(function($animate, $rootScope) {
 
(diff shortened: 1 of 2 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -876,6 +876,51 @@ describe("animations", function() {
...
+
+        expect(closed).toBe(false);
+
+        $animate.removeClass(element, 'active-class');
+        $rootScope.$digest();
+
+        expect(closed).toBe(true);
+      }));
+
+      it('should cancel the previously running addClass animation if a follow-up removeClass animation is using the same class value',
```
mzgol: aren't descriptions of those 2 tests swapped?

petebacondarwin: I think you are right @mzgol.

test/ngAnimate/animateSpec.js
```diff
@@ -876,6 +876,51 @@ describe("animations", function() {
...
+        inject(function($animate, $rootScope, $$rAF) {
+
+        parent.append(element);
+        var runner = $animate.addClass(element, 'active-class');
+        $rootScope.$digest();
+
+        var closed = false;
+        runner.done(function(status) {
+          closed = true;
+        });
```
petebacondarwin: Is there a reason why you are not using spies for this kind of thing? I find it much more readable:

```js
var doneHandler = jasmine.createSpy('addClass done');
runner.done(doneHandler);

$$rAF.flush();

expect(doneHandler).not.toHaveBeenCalled();

$animate.removeClass(element, 'active-class');
$rootScope.$digest();

expect(doneHandler).toHaveBeenCalled();
```

petebacondarwin:
Other than comments about descriptions and spies, this LGTM.

petebacondarwin:
I rebased this to merge but it causes another spec to fail at this line: https://github.com/angular/angular.js/blob/213c2a703293ee0af8229dde2b608687cd77ccfa/test/ngAnimate/animateSpec.js#L1053

My suggested fix is to clear the animation state if a newly merged animation is invalid: https://github.com/angular/angular.js/blob/213c2a703293ee0af8229dde2b608687cd77ccfa/src/ngAnimate/animateQueue.js#L361

So this bit of code would look like:

```js
      if (!isValidAnimation) {
        close();
        clearElementAnimationState(element);
        return runner;
      }
```

petebacondarwin:
Closing in favour of #11927

matsko:
Landed as https://github.com/angular/angular.js/commit/db246eb701529b41049fc118908e528920f13b24

PR #11760, opened 2015-04-29
fix(ngAnimate): ensure that all jqLite elements are deconstructed properly

### LINKED ISSUE #11658: ng-animate breaks with ngMaterial
AN exception in angular-animate.js is allowing the creation of more than one ng-view instance in teh pnge.:

```
"Error: animationDetails.runner is undefined
closeChildAnimations/<@http://192.168.1.24:8087/angular-animate.js:2202:13
n@http://192.168.1.24:8087/angular.min.js:7:300
closeChildAnimations@http://192.168.1.24:8087/angular-animate.js:2197:7
queueAnimation@http://192.168.1.24:8087/angular-animate.js:2042:9
$$AnimateQueueProvider</this.$get</<.push@http://192.168.1.24:8087/angular-animate.js:1937:16
Ne</this.$get</<.leave@http://192.168.1.24:8087/angular.min.js:186:468
z@http://192.168.1.24:8087/angular-route.min.js:6:210
x/n<@http://192.168.1.24:8087/angular-route.min.js:6:456
T/<@http://192.168.1.24:8087/angular.min.js:52:320
S/<@http://192.168.1.24:8087/angular.min.js:54:282
m@http://192.168.1.24:8087/angular.min.js:58:484
x@http://192.168.1.24:8087/angular-route.min.js:6:364
df/this.$get</m.prototype.$broadcast@http://192.168.1.24:8087/angular.min.js:135:43
n/<@http://192.168.1.24:8087/angular-route.min.js:11:426
f/<@http://192.168.1.24:8087/angular.min.js:118:182
df/this.$get</m.prototype.$eval@http://192.168.1.24:8087/angular.min.js:132:217
df/this.$get</m.prototype.$digest@http://192.168.1.24:8087/angular.min.js:129:225
df/this.$get</m.prototype.$apply@http://192.168.1.24:8087/angular.min.js:132:497
@http://192.168.1.24:8087/angular-touch.min.js:12:106
n.event.dispatch@http://192.168.1.24:8087/jquery-2.1.1.min.js:3:6352
n.event.add/r.handle@http://192.168.1.24:8087/jquery-2.1.1.min.js:3:3162
"
```

matsko:
Prior to this fix if a form DOM element was fed into parts of the
ngAnimate queuing code it would attempt to detect if it is a jqLite
object in an unstable way which would allow a form element to return an
inner input element by index. This patch ensures that jqLite instances
are properly detected using a helper method.

Closes #11658

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -117,7 +117,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     }
 
     function findCallbacks(element, event) {
-      var targetNode = element[0];
+      var targetNode = getDomNode(element);
 
       var matches = [];
       var entries = callbackRegistry[event];
@@ -198,7 +198,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
             // (bool) - Global setter
             bool = animationsEnabled = !!element;
           } else {
-            var node = element.length ? element[0] : element;
+            var node = getDomNode(element);
             var recordExists = disabledElementsLookup.get(node);
 
             if (argCount === 1) {
@@ -224,7 +224,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       var node, parent;
       element = stripCommentsFromElement(element);
       if (element) {
-        node = element[0];
+        node = getDomNode(element);
         parent = element.parent();
       }
 
@@ -411,7 +411,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
           close(!status);
           var animationDetails = activeAnimationsLookup.get(node);
           if (animationDetails && animationDetails.counter === counter) {
-            clearElementAnimationState(element);
+            clearElementAnimationState(getDomNode(element));
           }
           notifyProgress(runner, event, 'close', {});
         });
@@ -438,7 +438,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     }
 
     function closeChildAnimations(element) {
-      var node = element[0];
+      var node = getDomNode(element);
       var children = node.querySelectorAll('[' + NG_ANIMATE_ATTR_NAME + ']');
       forEach(children, function(child) {
         var state = parseInt(child.getAttribute(NG_ANIMATE_ATTR_NAME));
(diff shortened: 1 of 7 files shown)


chirayuk:
I have two comments—one about the naming of `getDomElement` and another about a naming convention for jQuery objects vs DOM nodes.

###### On naming `getDomElement`
As an example, when you replaced (in animateQueue.js:411) `clearElementAnimationState(element)` with `clearElementAnimationState(getFirstNode(element))`, to the reader, it looks like you already had an element and then you got the first node from it.  The reader will typically wonder if element is a jQuery object or some DOM node collection or if getFirstNode returns some kind of child node.  If instead, one writes, `clearElementAnimationState(asDomNode(element))` / `clearElementAnimationState(toDomNode(element))` / `clearElementAnimationState(ensureDomNode(element))` or some such call, then it becomes very clear that while `element` itself might be a DOM node or a jQuery element, the result of the call is not a jQuery element but an actual plain DOM node.

###### Naming conventions for jQuery instances vs DOM nodes
There are parts of the code that require DOM nodes, some that require jQuery (or angular.element) instances, and some that can work with either.  To the reader (and all future maintainers), it can become pretty hard to keep track of this given the lack of a type safe language.  Perhaps we should use a naming convention—`element` **always** means it's a jQuery instance, `node` **always** means it's a DOM node, and perhaps `elementOrNode` or some such name is used for the few places where it could be either.  That would greatly simplify the cognitive load for the reader, allow one to get rid of gratuitous wrapping/unwrapping calls and help catch bug more easily.  Thoughts?

src/ngAnimate/animateQueue.js
```diff
@@ -454,9 +454,9 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
...
     function clearElementAnimationState(element) {
-      element = element.length ? element[0] : element;
-      element.removeAttribute(NG_ANIMATE_ATTR_NAME);
-      activeAnimationsLookup.remove(element);
+      var node = getFirstNode(element);
+      node.removeAttribute(NG_ANIMATE_ATTR_NAME);
+      activeAnimationsLookup.remove(node);
     }
 
     function isMatchingElement(a,b) {
```
chirayuk: What are `a` and `b` here—jQuery instances of DOM nodes?  Do they need to be wrapped/unwrapped before calling isMatchingElement at all call sites?

chirayuk:
In `areAnimationsAllowed(element, parent, event)`, I take it that `element` and `parent` are always jQuery instances.  What functions/methods require DOM nodes instead of jQuery elements?  If it's the rare function that requires a raw DOM node, maybe just those parameters/variables can use a naming convention to indicate that (e.g. with a prefix like domNode, domParent, etc.)?

gkalpak:
:+1: for having some naming conventions to indicate whether the value is expected to be a DOM node, JQLite object or any.

I am not sure that `element vs node` will cut it though, since there are cases where we need to reference a `parent` or `container`.

I some projects I have tried prefixing JQLite variables with `jq`. E.g.:
DOM elements: `element`, `node`, `body`, `parent`
JQLite elements: `jqElement`, `jqNode`, `jqBody`, `jqParent`

(But somehow (and without being sure why) I wasn't too happy about it.)

petebacondarwin:
I thought that the convention in the angular code base was that "wrapped" DOM nodes (i.e. jqLite elements always start with a `$`? See https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L20

gkalpak:
I don't think this is consistently the case in all files (maybe just `compaile.js`).

In practice, I don't like that convention (although it would be indeed very convenient), because everything `$`-prefixed is supposed to be Angular stuff (i.e. usually injectable stuff). So having to write `function someFunc($element)` while `$element` is not injectable makes me very unhappy.

--
<sub>I particularly hate DDOs with `link: function ($scope, $element) {...}` even if the arguments are not injectable. E.g. compare it to `controller: function ($scope, $element) {...}` where the arguments are injectable (and order does not make a difference).</sub>

But it's a matter of personal preferrence I guess and it is very common (even in Angular core :cry:).

matsko:
Merged as https://github.com/angular/angular.js/commit/64d05180a667e586328fbdbd328889d3b003571d

matsko:
We'll have another PR in RC2+ which renames things properly.

PR #11761, opened 2015-04-29
Clarify's test case for ng-repeat stability test.

nbrookie:
Hey Angular devs,

I was going through the `ng-repeat` directive test cases, and noticed a typo (or at least, I think its a typo) in the setup for the stability test case.

Having `{{key}}:{{val}}` in the repeated element did not seem to make sense to me, especially when the expression is `ng-repeat="item in items"`. 

Anyways, hope this helps!


petebacondarwin:
Assuming Travis now passes (it was flaking previously) then this LGTM

PR #11764, opened 2015-04-30
fix($http): do not modify the config object passed into $http short methods

maxthyen:
Update $http's `createShortMethods` and `createShortMethodsWithData` to extend an empty object instead of the passed-in config.

Previously, since $http was extending the passed-in config, the changes to the config object persisted even after the call to $http's get/post/etc. returned:

```
var config = { some_key: 1 };
$http.get('/testUrl', config);
// config now contains { some_key: 1, url: '/testUrl', method: 'GET' }
```

This causes unexpected behavior if that config object is reused in subsequent calls to $http. The existing test in httpSpec was not properly testing this situation.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1076,7 +1076,7 @@ function $HttpProvider() {
     function createShortMethods(names) {
       forEach(arguments, function(name) {
         $http[name] = function(url, config) {
-          return $http(extend(config || {}, {
+          return $http(extend({}, config || {}, {
             method: name,
             url: url
           }));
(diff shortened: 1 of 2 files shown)


johnculviner:
:+1: this caused a pretty lengthy head scratching session for @maxthyen and I until the behavior was realized. Can't think of any reasons why this would be desirable?

matthewdordal:
:+1: being able to reuse a confit object without it being changed under the hood is ideal.

lgalfaso:
I like this PR, just wondering if anybody is relying on this odd behavior to think that this is a breaking change.
@petebacondarwin WDYT?

petebacondarwin:
I agree. It looks good. We should merge it now and label it as a small breaking change - just in case.

petebacondarwin:
Do you want to take that @lgalfaso ?

lgalfaso:
landed as f7a4b48121ed2b04af89bd2b754f500d1872360d

PR #11765, opened 2015-04-30
fix(ngOptions): correctly watch displayed value for changes

dremonkey:
ngOptions does not update the displayed value properly if the property being used as the label changes when not using the "select as" or "track by" syntax, i.e `<select ng-options="value.label for value in array"></select>` will not update if `array[0].label` is changed but `<select ng-options="value.label as value.label for value in array></select>` will correctly update.

The bug is demonstrated here: [https://jsfiddle.net/76qrsctv/17/](https://jsfiddle.net/76qrsctv/18/)

The proposed change fixes this issue.


dremonkey:
I signed it!

petebacondarwin:
Thanks for highlighting this issue. We needed to provide a unit test too.

PR #11771, opened 2015-04-30
fix(ngOptions): iterate over the options collection in the same way as `...

### LINKED ISSUE #11733: ng-options include array properties
If an array contains additional properties (notably, as when fetched with Restangular, which adds several functions to the array), these extra properties are included in the iteration for ng-options, leading to unexpected "junk" entries in the select dropdown.

This differs from the iteration when using ng-repeat over the same array, which does not include such properties.

(This is new behavior in v1.4 rc0 and rc1. v1.3 does not show this behavior)

petebacondarwin:
...ngRepeat`

In `ngRepeat` if the object to be iterated over is "array-like" then it only iterates
over the numerical indexes rather than every key on the object. This prevents "helper"
methods from being included in the rendered collection.

This commit changes `ngOptions` to iterate in the same way.

BREAKING CHANGE:

Although it is unlikely that anyone is using it in this way, this change does change the
behaviour of `ngOptions` in the following case:
- you are iterating over an array-like object, using the array form of the `ngOptions` syntax
  (`item.label for item in items`) and that object contains non-numeric property keys.

In this case these properties with non-numeric keys will be ignored.

*\* Here array-like is defined by the result of a call to this internal function:
https://github.com/angular/angular.js/blob/v1.4.0-rc.1/src/Angular.js#L198-L211 **

To get the desired behaviour you need to iterate using the object form of the `ngOptions` syntax
(`value.label` for (key, value) in items)`).

Closes #11733

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -82,7 +82,7 @@ function $HttpParamSerializerJQLikeProvider() {
           parts.push(encodeUriQuery(prefix) + '=' + encodeUriQuery(serializeValue(toSerialize)));
         }
       }
-    }
+    };
   };
 }
 


src/ng/directive/ngOptions.js
```diff
@@ -325,12 +325,28 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+        }
+
+        var optionValuesLength = optionValuesKeys.length;
+
+        for (var index = 0; index < optionValuesLength; index++) {
+          var key = (optionValues === optionValuesKeys) ? index : optionValuesKeys[index];
 
           // Ignore "angular" properties that start with $ or $$
-          if (key.charAt(0) === '$') return;
+          if (key.charAt && key.charAt(0) === '$') return;
```
petebacondarwin: Actually this line is no longer needed as it is dealt with in the key extraction logic above

petebacondarwin: And it is actually totally **wrong** now that it is not using `forEach`

src/ng/directive/ngOptions.js
```diff
@@ -325,12 +325,28 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         // The option values were already computed in the `getWatchables` fn,
         // which must have been called to trigger `getOptions`
         var optionValues = valuesFn(scope) || [];
+        var optionValuesKeys;
 
-        var keys = Object.keys(optionValues);
-        keys.forEach(function getOption(key) {
+
+        if (!keyName && isArrayLike(optionValues)) {
```
caitp: I'm not sure I understand why `!keyName` is needed --- don't we want this to enable the use of other expressions? Like:  `ng-options="foo.value as foo.key for foo in foos"` or something

petebacondarwin: This check ensures that we don't treat array-like objects as array-like if we are using the object form of the ngOptions syntax.  Without this the second test would fail.

caitp: Oh you mean this prevents `(key, value) in foos` from working? I don't think there's much point in that either

petebacondarwin: Just to be clear `keyName` is the value of the bit of the `ngOptions` expression in capitals below:

```
foo.value as foo.key for (KEYNAME, foo) in foos
```

petebacondarwin: It doesn't prevent it from working, it is saying that if you explicitly are iterating over an object's keys (via `(key,value) in obj`) then you don't want `ngOptions` to treat it like an array.

caitp: see my above comment --- so, if fixing that's a breaking change, then I guess we can't, but it's a really dumb API

src/ng/directive/ngOptions.js
```diff
@@ -325,13 +325,25 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+            if (optionValues.hasOwnProperty(itemKey) && itemKey.charAt(0) !== '$') {
+              optionValuesKeys.push(itemKey);
+            }
+          }
+        }
 
+        var optionValuesLength = optionValuesKeys.length;
+
+        for (var index = 0; index < optionValuesLength; index++) {
+          var key = (optionValues === optionValuesKeys) ? index : optionValuesKeys[index];
```
petebacondarwin: I guess I should probably move the declaration of `index` and `key` up to the top of the function body.

src/ng/directive/ngOptions.js
petebacondarwin: I don't understand? If I have an object that *happens* to look like an array but I am explicitly iterating over it as an object, then I would want all the properties to be included, whether they are numeric or not, right?

Or are you saying that iterating over an object's keys in general is a bad idea?

src/ng/directive/ngOptions.js
petebacondarwin: Actually I don't think I will move these as then we would have to move **all** the other declarations in this block, which would not make the code any clearer.

src/ng/directive/ngOptions.js
caitp: It's a bad idea for a few reasons

1. enumeration (for/in) order isn't totally hammered down even in es6
2. the api is inconsistent

petebacondarwin: OK, I hear that you are saying that the idea of iterating over an object's keys is a bad idea.
Unfortunately this would be a momentous breaking change for many apps and so we will not be changing this in the near future - if at all in 1.x.

Other than that are you happy with the PR?

src/ng/directive/ngOptions.js
```diff
@@ -325,13 +325,25 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
-        keys.forEach(function getOption(key) {
 
-          // Ignore "angular" properties that start with $ or $$
-          if (key.charAt(0) === '$') return;
+        if (!keyName && isArrayLike(optionValues)) {
+          optionValuesKeys = optionValues;
+        } else {
+          // if object, extract keys, in enumeration order, unsorted
+          optionValuesKeys = [];
+          for (var itemKey in optionValues) {
```
caitp: Lets use `Object.keys()`, then we can avoid the hasOwnProperty check, and use the nice optimized implementation in v8.

caitp: you could also just skip the test for `'$'` until iterating over the keys later on, so that you don't have to do complicated array filtering

caitp: for that matter, maybe split the key iterating into 2 functions, and just pick the one to call depending on whether `optionValues === optionValuesKeys` or not

petebacondarwin: Yes I thought of all those options and I do agree. The reason I went with this is that it is much closer (almost line for line) with the `ngRepeat` implementation and I thought it would be best to keep the two as similar as possible.

I believe that inside `ngRepeat` the argument is to try to minimize function calls to keep things fast. I have no idea if that is a valid performance issue in this case. I am happy to take your advice on that.

caitp: Function calls aren't that big of a deal --- what you want to avoid is method lookup in megamorphic code (which is really just as bad as property lookup). IC cache misses are the heavy hits which angular2 does gymnastics to try and avoid

petebacondarwin: I guess the best bet is to benchmark the difference :-) I will knock something up tonight.

caitp: it's not even worth it, a microbenchmark is going to be a complete waste of energy --- the key is really just to write it as simply as possible

petebacondarwin:
I refactored the code to make it cleaner, rather than follow the `ngRepeat` logic and ran the ng-options-bp benchmark on both strategies. You can see the refactored code here https://github.com/angular/angular.js/commit/864bee5645c4bf21ca81d5f3914c4a6219bed4ec

| Code Strategy  | add-options    | set-model-1    | set-model-2    | remove-options | set-view-1     | set-view-2     | 
|----------------|----------------|----------------|----------------|----------------|----------------|----------------|
| ngRepeat style | 7.45           | 7.39           | 13.90          | 0.03           | 14.32          | 12.72          |
| Clean style    | 13.83          | 10.77          | 24.80          | 0.02           | 21.32          | 23.02          |

It seems clear that it is best to leave the logic as it is.  @caitp if you want to look into why the refactored version was so much slower that would be great and we can do a refact commit later.

But in the meantime I am going to go with the `ngRepeat` style.

caitp:
using `forEach` will include the HasProperty check, which is super slow. We don't care about holes in arrays, so we don't want to eat that performance hit.

petebacondarwin:
Right, I was using `forEach` because if the collection is only "array-like" then it wouldn't necessarily have the `Array.prototype.forEach` method available.

caitp:
Well, ignoring "holes" in arrays has been asked for, and forEach() would do it --- a way around that is to just write a quick fast version of forEach that only deals with array-like types, and always ignores holes

petebacondarwin:
I am going to land this and leave refactoring for post 1.4.0

PR #11772, opened 2015-04-30
fix($compile): validate directive name for whitespaces

### LINKED ISSUE #11397: directive name should be trimmed
Directive name should probably be either trimmed or angular should throw an exception, so a typo is easier to spot. Leading or following space is clearly an error.

In the following example the name is 'directiveName ' instead of 'directiveName':

https://jsfiddle.net/c2ztgdjr/

Maybe same rule should be applied to controllers, factories etc, although it's still possible to inject those with an extra space.

lugovsky:
Throw an exception if directive name contains leading or trailing whitespaces

Closes #11397

### DIFF
--- a/docs/content/error/$compile/baddir.ngdoc
+++ b/docs/content/error/$compile/baddir.ngdoc
@@ -5,4 +5,4 @@
 
 This error occurs when the name of a directive is not valid.
 
-Directives must start with a lowercase character.
+Directives must start with a lowercase character and must not contain leading or trailing whitespaces.
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -802,6 +802,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     if (!letter || letter !== lowercase(letter)) {
       throw $compileMinErr('baddir', "Directive name '{0}' is invalid. The first character must be a lowercase letter", name);
     }
+    if (name !== name.trim()) {
+      throw $compileMinErr('baddir',
+            "Directive name '{0}' is invalid. The name should not contain leading or trailing whitespaces",
+            name);
+    }
   }
 
   /**
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -211,6 +211,21 @@ describe('$compile', function() {
       });
       inject(function($compile) {});
     });
+    it('should throw an exception if a directive name has leading or trailing whitespace', function() {
+      module(function() {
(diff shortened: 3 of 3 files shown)


gkalpak:
I think you would also need a new `baddirwspc.ngdoc` file under `docs/content/error/$compile/`.
E.g. take a look at [docs/content/error/$compile/baddir.ngdoc][1].

[1]: https://github.com/angular/angular.js/blob/master/docs/content/error/%24compile/baddir.ngdoc

pkozlowski-opensource:
LGTM, provided that an existing error page is used and its content is updated accordingly

lugovsky:
@pkozlowski-opensource Is the updated error page good enough for you, or probably you would like to structure conditions in a list?

pkozlowski-opensource:
Should be fine as a starting point. Could you please squash all the 3 commits into one so it is easier to merge?

lugovsky:
@pkozlowski-opensource Done

lugovsky:
@pkozlowski-opensource Sorry for interrupting you again. I can't understand why CI build failed. Don't see any errors related to my changes. Could you advice on that? :)

gkalpak:
It's all well now :)

petebacondarwin:
@pkozlowski-opensource  and @gkalpak can you assign this to a milestone and/or deal with it?
Thanks

gkalpak:
It LGTM as is, but if we wanted to go the extra mile, we could verify there is no whitespace in the name whatsoever.

petebacondarwin:
Is this a breaking change or did the compiler just fail to match directives with leading/trailing whitespace before?

Should we just be more compliant and simply trim directive names?

gkalpak:
Previously, the compiler just failed to match the directives, so this is not a breaking change.
I am not sure what is the best approach: throwing an error or trimming names (I'd be happy with either).

If we decide to also check for internal whitespace, then I guess throwing makes more sense.

pkozlowski-opensource:
> If we decide to also check for internal whitespace, then I guess throwing makes more sense

Personally I would only add additional checks if users keep complaining about common mistakes. Each check cost us frw size and CPU cycles, so checks should be used with moderation, IMO.

pkozlowski-opensource:
Ah, and I'm for throwing - I really don't like systems that try to work-around common mistakes, Fail fast is the way to go IMO.

petebacondarwin:
OK, let's merge this then.

PR #11774, opened 2015-04-30
fix(ngAria): update `aria-valuemin/max` when `min/max` change

### LINKED ISSUE #11770: bug(ngAria): update `aria-valuemin/max` when `min`/`max` change
Currently, `ngAria` checks `attr.min/max` in a 200-priority directive's post-linking function and sets `aria-valuemin`/`aria-valuemax` accordingly.

There are 2 issues with this approach:

1. `ngMin`/`ngMax` will also result in `attr.min/max` being updated, but only **after** `ngAria` has set `aria-valuemin/max`. As a result, `ngAria` is effectively **not** taking into account `ngMin`/`ngMax` (while I think it should).

2. If `min`/`max` change later on, `aria-valuemin/max` are not updated, resulting in misleading info (accessibility loses :cry:).

--
Luckily, solving (2) will automatically fix (1) as well.

So, I believe that `ngAria` should `attr.$observe` certain attributes and update aria values accordingly.
This is probably true for other attributes besides `min`/`max` (@marcy should know better).

gkalpak:
As a result of thi fix, `ngMin/Max` also set `aria-valuemin/max` on
"range"-shaped elements.

Fixes #11770

### DIFF
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -510,6 +510,34 @@ describe('$aria', function() {
       expectAriaAttrOnEachElement(element, 'aria-valuemin', 'userSetValue2');
       expectAriaAttrOnEachElement(element, 'aria-valuemax', 'userSetValue3');
     });
+
+    it('should update `aria-valuemin/max` when `min/max` changes dynamically', function() {
+      scope.$apply('min = 25; max = 75');
+      compileElement('<input type="range" ng-model="val" min="{{min}}" max="{{max}}" />');
+
+      expect(element.attr('aria-valuemin')).toBe('25');
+      expect(element.attr('aria-valuemax')).toBe('75');
+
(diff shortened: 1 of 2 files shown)


test/ngAria/ariaSpec.js
```diff
@@ -510,6 +510,34 @@ describe('$aria', function() {
...
+      expect(element.attr('aria-valuemax')).toBe('75');
+
+      scope.$apply('min = 0');
+      expect(element.attr('aria-valuemin')).toBe('0');
+
+      scope.$apply('max = 100');
+      expect(element.attr('aria-valuemax')).toBe('100');
+    });
+
+    it('should update `aria-valuemin/max` when using `ng-min/max`', function() {
```
Narretz: This should be worded similar to the test above. Or both wrapped in a describe.

PR #11778, opened 2015-04-30
docs(changelog): wrap jqLite example containing html with code block

Narretz:
This prevents the markdown parser from garbling the input and putting
out broken html.


gkalpak:
Since you are at it, there is one more similar break starting at `0.9.10`'s _Bug Fixes_.

Narretz:
Fixed! :laughing:

petebacondarwin:
@Narretz thanks for sorting this. Please merge.

PR #11780, opened 2015-04-30
docs(ngJq): update to indicate common pitfall

### LINKED ISSUE #11779: ng-jq only works on html tag
I tried the new tag 'ng-jq' to disable trying to use my very old jQuery lib

<div  ng-jq ng-app="myApp"> (not working)

<html ng-jq ...> (working)

In my case I tried to improve a small module in a very old Joomla page so changing the layout template is not really a nice way to do it - would be better if I could use ng-jq on the app tag too)

AT least the description stated it not madatory to do it in the <html> tag

tried with: 1.4.0-rc.1

kentcdodds:
change docs for ngJq so it mentions that the placement of the directive is important with regards to the angular script.

Closes #11779

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -964,10 +964,10 @@ var csp = function() {
  * used to force either jqLite by leaving ng-jq blank or setting the name of
  * the jquery variable under window (eg. jQuery).
  *
- * Since this directive is global for the angular library, it is recommended
- * that it's added to the same element as ng-app or the HTML element, but it is not mandatory.
- * It needs to be noted that only the first instance of `ng-jq` will be used and all others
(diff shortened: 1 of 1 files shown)


caitp:
It looks good to me. I think fixing the caveat would cause more problems than it's worth, but @petebacondarwin can make a call on that

petebacondarwin:
This LGTM. Feel free to merge.

caitp:
maybe the wording could be improved a bit, I mean we're starting a sentence with a preposition here =) but it's good enough for now, followup wording fixups welcome

petebacondarwin:
Thanks @caitp

PR #11784, opened 2015-05-01
fix(ngOptions): ensure that tracked properties are always watched

petebacondarwin:
Commit 47f9fc3e70bc361e8c11fe68dc3ec4489238efb3 failed to account for changes to
the tracked value of model items in a collection where the select was `multiple`.

See https://github.com/angular/angular.js/pull/11743#discussion_r29424578

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -261,9 +261,13 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
     // Get the value by which we are going to track the option
     // if we have a trackFn then use that (passing scope and locals)
     // otherwise just hash the given viewValue
-    var getTrackByValue = trackBy ?
-                              function(viewValue, locals) { return trackByFn(scope, locals); } :
-                              function getHashOfValue(viewValue) { return hashKey(viewValue); };
+    var getTrackByValueFn = trackBy ?
+                              function(value, locals) { return trackByFn(scope, locals); } :
+                              function getHashOfValue(value) { return hashKey(value); };
+    var getTrackByValue = function(value, key) {
+      return getTrackByValueFn(value, getLocals(value, key));
+    };
+
     var displayFn = $parse(match[2] || match[1]);
     var groupByFn = $parse(match[3] || '');
     var disableWhenFn = $parse(match[4] || '');
@@ -290,6 +294,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
 
     return {
       trackBy: trackBy,
+      getTrackByValue: getTrackByValue,
       getWatchables: $parse(valuesFn, function(values) {
         // Create a collection of things that we would like to watch (watchedArray)
         // so that they can all be watched using a single $watchCollection
@@ -299,7 +304,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
 
         Object.keys(values).forEach(function getWatchable(key) {
           var locals = getLocals(values[key], key);
-          var selectValue = getTrackByValue(values[key], locals);
+          var selectValue = getTrackByValueFn(values[key], locals);
           watchedArray.push(selectValue);
 
           // Only need to watch the displayFn if there is a specific label expression
@@ -347,7 +352,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
           var value = optionValues[key];
           var locals = getLocals(value, key);
           var viewValue = viewValueFn(scope, locals);
-          var selectValue = getTrackByValue(viewValue, locals);
+          var selectValue = getTrackByValueFn(viewValue, locals);
           var label = displayFn(scope, locals);
           var group = groupByFn(scope, locals);
           var disabled = disableWhenFn(scope, locals);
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngOptions.js
```diff
@@ -439,44 +444,54 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+          return null;
+        };
 
+        // If we are using `track by` then we must watch the tracked value on the model
+        // since ngModel only watches for object identity change
+        if (ngOptions.trackBy) {
+          scope.$watch(
+            function() { return ngOptions.getTrackByValue(ngModelCtrl.$viewValue); },
+            function() { ngModelCtrl.$render(); }
+          );
```
petebacondarwin: So now, when the select is not `multiple`, there is only a simple watch on the tracked value.

src/ng/directive/ngOptions.js
```diff
@@ -508,6 +523,22 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+
+          scope.$watchCollection(function() {
+            if (isArray(ngModelCtrl.$viewValue)) {
+              return ngModelCtrl.$viewValue.map(function(value) {
+                return ngOptions.getTrackByValue(value);
+              });
+            }
+          }, function() {
+            ngModelCtrl.$render();
+          });
```
petebacondarwin: Now, when the select is `multiple` there is a $watchCollection on the array of tracked values.

src/ng/directive/ngOptions.js
```diff
@@ -534,11 +565,6 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
       // We will re-render the option elements if the option values or labels change
       scope.$watchCollection(ngOptions.getWatchables, updateOptions);
 
-      // We also need to watch to see if the internals of the model changes, since
-      // ngModel only watches for object identity change
-      if (ngOptions.trackBy) {
-        scope.$watchCollection(attr.ngModel, function() { ngModelCtrl.$render(); });
-      }
```
petebacondarwin: Using $watchCollection on the model was misguided.
This would:
* unnecessarily watch all the properties on a non-multiple select's model
* not react to changes if the tracked value was based on a deeper property in the model on a non-multiple select
* not react to changes in a multple select if only the tracked property value was changed on an item in the model collection.

src/ng/directive/ngOptions.js
```diff
@@ -261,9 +261,13 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
-    var getTrackByValue = trackBy ?
-                              function(viewValue, locals) { return trackByFn(scope, locals); } :
-                              function getHashOfValue(viewValue) { return hashKey(viewValue); };
+    var getTrackByValueFn = trackBy ?
+                              function(value, locals) { return trackByFn(scope, locals); } :
+                              function getHashOfValue(value) { return hashKey(value); };
+    var getTrackByValue = function(value, key) {
+      return getTrackByValueFn(value, getLocals(value, key));
+    };
+
```
petebacondarwin: `getTrackByValueFn` and `getTrackByValue` are both useful functions.
The first allows us to get performance benefits by not recomputing the locals unnecessarily
The second gives us a clean function that we can expose outside this function without having to expose `getLocals`

test/ng/directive/ngOptionsSpec.js
```diff
@@ -988,9 +988,9 @@ describe('ngOptions', function() {
 
       expect(element.val()).toEqual(['10']);
 
-      // Update the properties on the object in the selected array, rather than replacing the whole object
+      // Update the tracked property on the object in the selected array, rather than replacing the whole object
       scope.$apply(function() {
-        scope.selected[0] = {id: 20, label: 'new twenty'};
+        scope.selected[0].id = 20;
```
petebacondarwin: The earlier commit that changed this line was wrong.

test/ng/directive/ngOptionsSpec.js
```diff
@@ -1138,13 +1138,13 @@ describe('ngOptions', function() {
...
       scope.$apply(function() {
-        scope.selected = scope.arr[0];
+        scope.selected = {id: 10, label: 'ten'};
       });
 
       spyOn(element.controller('ngModel'), '$render');
 
       scope.$apply(function() {
-        scope.selected.label = 'changed';
+        scope.arr[0].id = 20;
```
petebacondarwin: Because we are using `track by` here, the model and the option are different objects.

So we are not interested in the `label` of the selected item, as this doesn't affect the visual display of the options, which are only interested in the `label` property on the options themselves.

test/ng/directive/ngOptionsSpec.js
```diff
@@ -1138,13 +1138,13 @@ describe('ngOptions', function() {
       });
 
       scope.$apply(function() {
-        scope.selected = scope.arr[0];
+        scope.selected = {id: 10, label: 'ten'};
```
petebacondarwin: We need to use a totally different object here (which is matched by the `id` property to the option) so that we can then change the value of the `id` property of the model without changing the `id` property on the option.

test/ng/directive/ngOptionsSpec.js
```diff
@@ -1130,7 +1130,7 @@ describe('ngOptions', function() {
       }).not.toThrow();
     });
 
-    it('should re-render if a propery of the model is changed when using trackBy', function() {
+    it('should re-render if the tracked property of the model is changed when using trackBy', function() {
```
petebacondarwin: We are not actually interested in watching any old property on the model. The important thing is the `track by` expression.

src/ng/directive/ngOptions.js
Narretz: Maybe I don't get it, but it looks like if the $viewValue is not an array, nothing is actually watched, right? So is this only valid for select[multiple]? And if so, could we add this $watch only if it is multiple?

petebacondarwin: This is only added if multiple. I reorganized the if statement above. This is in the else clause now.

PR #11786, opened 2015-05-01
fix($rootScope): stop memory leak in IE9

### LINKED ISSUE #10706: Memory Leak in IE9 with AngularJS 1.2.28
The following gist demonstrate the problem. It is a very basic example that switches between two views. With IE9 (Version 9.0.15) on Windows 7 the memory consumption for process "iexplore" continues to grow until it crashes.

https://gist.github.com/m-unterberger/e90ec1552df6b43bafc9

petebacondarwin:
DO NOT MERGE
This is a work in progress for discussion - it is not ideal is lacking tests.

Closes #10706

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -101,6 +101,29 @@ function $RootScopeProvider() {
         $event.currentScope.$$destroyed = true;
     }
 
+    function cleanUpScope($scope) {
+
+      if (msie === 9) {
+        // There is a memory leak in IE9 if all child scopes are not disconnected
+        // completely when a scope is destroyed. So this code will recurse up through
+        // all this scopes children
+        //
+        // See issue https://github.com/angular/angular.js/issues/10706
+        $scope.$$childHead && cleanUpScope($scope.$$childHead);
+        $scope.$$nextSibling && cleanUpScope($scope.$$nextSibling);
+      }
+
+      // The code below works around IE9 and V8's memory leaks
+      //
+      // See:
+      // - https://code.google.com/p/v8/issues/detail?id=2073#c26
+      // - https://github.com/angular/angular.js/issues/6794#issuecomment-38648909
+      // - https://github.com/angular/angular.js/issues/1313#issuecomment-10378451
+
+      $scope.$parent = $scope.$$nextSibling = $scope.$$prevSibling = $scope.$$childHead =
+          $scope.$$childTail = $scope.$root = $scope.$$watchers = null;
+    }
+
     /**
      * @ngdoc type
      * @name $rootScope.Scope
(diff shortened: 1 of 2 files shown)


src/ng/rootScope.js
```diff
@@ -101,6 +101,15 @@ function $RootScopeProvider() {
         $event.currentScope.$$destroyed = true;
     }
 
+    function cleanUpScope($scope) {
+      // Recurse bottom up
+      $scope.$$childHead && cleanUpScope($scope.$$childHead);
+      $scope.$$nextSibling && cleanUpScope($scope.$$nextSibling);
```
IgorMinar: this is not correct for the first call to `cleanUpScope`. we don't want to traverse into siblings of the scope being destroyed.

src/ng/rootScope.js
```diff
@@ -900,6 +909,13 @@ function $RootScopeProvider() {
         this.$on = this.$watch = this.$watchGroup = function() { return noop; };
         this.$$listeners = {};
 
+        var that = this;
```
IgorMinar: can you use more descriptive name? e.g. `currentScope`

IgorMinar: also please add a comment linking back to the bug.

IgorMinar:
lgtm. needs cleanup, comments and tests.

src/ng/rootScope.js
```diff
@@ -900,6 +909,13 @@ function $RootScopeProvider() {
...
+        if (msie === 9) {
+          this.$evalAsync(function() {
+            cleanUpScope(that);
```
petebacondarwin: So we can just set `that.$$nextSibling = null` before calling `cleanUpScope(that)` to prevent `cleanUpScope` from trying to kill sibling nodes.

petebacondarwin:
I did a full clean up.
I disconnect the `$$nextSibling` so that `cleanUpScope` doesn't try to disconnect all the real siblings of the destroyed scope.
I have moved the disconnection stuff (that fixes a Chrome memory leak) into `cleanUpScope` to prevent repeating ourselves. Therefore we don't need to use `$evalAsync` anymore.

src/ng/rootScope.js
```diff
@@ -101,6 +101,29 @@ function $RootScopeProvider() {
...
+    function cleanUpScope($scope) {
+
+      if (msie === 9) {
+        // There is a memory leak in IE9 if all child scopes are not disconnected
+        // completely when a scope is destroyed. So this code will recurse up through
+        // all this scopes children
+        //
+        // See issue https://github.com/angular/angular.js/issues/10706
+        $scope.$$childHead && cleanUpScope($scope.$$childHead);
+        $scope.$$nextSibling && cleanUpScope($scope.$$nextSibling);
```
gkalpak: What is the purpose of this line ? We already have _nulled out_ `$scope.$$nextSibling` before calling `cleanUpScope()`, haven't we ?

PR #11789, opened 2015-05-01
docs(protractor-conf): fix syntax error

ghost:
Fix comma in line 12 as there is only one key in config.capabilities object.


Narretz:
This is valid syntax starting from ES5. The file is also only used in a node context, so it should never cause an error. Did you encounter any problems with this?

PR #11790, opened 2015-05-01
chore(sanitize): new implementation of the html sanitized parser

mhevery:
This implementation is based on inert document

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11790)


test/ngSanitize/sanitizeSpec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-describe('HTML', function() {
+ddescribe('HTML', function() {
```
gkalpak: _ddescribe_

src/ngSanitize/sanitize.js
```diff
@@ -262,11 +236,34 @@ function makeMap(str, lowercaseKeys) {
   return obj;
 }
 
+var baseNode;
+(function(window) {
+  var doc;
+  if (window.DOMDocument) {
+    doc = new window.DOMDocument();
```
lgalfaso: Is a `DOMDocument` *inert*? I was not able to find anything that claims this

test/ngSanitize/sanitizeSpec.js
```diff
@@ -88,15 +67,15 @@ describe('HTML', function() {
     });
 
     it('should throw badparse if text content contains "<" followed by "/" without matching ">"', function() {
```
lgalfaso: given that this is not throwing, the test should be renamed

test/ngSanitize/sanitizeSpec.js
```diff
@@ -88,15 +67,15 @@ describe('HTML', function() {
...
-      expect(function() {
-        htmlParser('foo </ bar', handler);
-      }).toThrowMinErr('$sanitize', 'badparse', 'The sanitizer was unable to parse the following block of html: </ bar');
+      htmlParser('foo </ bar', handler);
+      expect(start).toEqual(undefined);
+      expect(text).toEqual('foo ');
     });
 
     it('should throw badparse if text content contains "<" followed by an ASCII letter without matching ">"', function() {
```
lgalfaso: same

src/ngSanitize/sanitize.js
```diff
@@ -262,11 +236,34 @@ function makeMap(str, lowercaseKeys) {
...
+  var bodys = docElement.getElementsByTagName('body');
+  if (bodys.length === 1) {
+    baseNode = bodys[0];
+  } else {
+    var html = doc.createElement('html');
+    var body = baseNode = doc.createElement('body');
+    html.appendChild(body);
+    doc.appendChild(html);
+  }+
+})(window);
```
gkalpak: What would be the drawback of using `$window` instead (after moving this inside `$get()`) ?

src/ngSanitize/sanitize.js
```diff
@@ -262,11 +236,34 @@ function makeMap(str, lowercaseKeys) {
...
+  var docElement = doc.documentElement || doc.getDocumentElement();
+  var bodys = docElement.getElementsByTagName('body');
+  if (bodys.length === 1) {
+    baseNode = bodys[0];
+  } else {
+    var html = doc.createElement('html');
+    var body = baseNode = doc.createElement('body');
+    html.appendChild(body);
+    doc.appendChild(html);
+  }+
```
campersau: Is the `+` behind `}` intended?

src/ngSanitize/sanitize.js
```diff
@@ -262,11 +236,34 @@ function makeMap(str, lowercaseKeys) {
...
+    doc = new window.DOMDocument();
+  } else if (window.document && window.document.implementation) {
+    doc = window.document.implementation.createHTMLDocument("inert");
+  } else if (window.ActiveXObject) {
+    doc = new ActiveXObject("Msxml.DOMDocument");
+  } else {
+    throw "DOMDocument not supported!";
+  }
+  var docElement = doc.documentElement || doc.getDocumentElement();
+  var bodys = docElement.getElementsByTagName('body');
```
IgorMinar: bodys -> bodyElements

src/ngSanitize/sanitize.js
IgorMinar: +1

src/ngSanitize/sanitize.js
```diff
@@ -262,11 +236,34 @@ function makeMap(str, lowercaseKeys) {
...
+  if (bodys.length === 1) {
+    baseNode = bodys[0];
+  } else {
+    var html = doc.createElement('html');
+    var body = baseNode = doc.createElement('body');
+    html.appendChild(body);
+    doc.appendChild(html);
```
IgorMinar: so we just append it as is since we assume that it's inert?

src/ngSanitize/sanitize.js
IgorMinar: there is no drawback, but I don't think there is any benefit either. we need to test this against the real window.

src/ngSanitize/sanitize.js
mhevery: That is my understanding, and also I was able to verify this imperatively. If you try to create ``  in regular document you can see either the alert or that it was canceled in the network tab. Creating the same thing on the new document does nothing.

src/ngSanitize/sanitize.js
mhevery: So some platforms create `html` > `body` but others do not. This is just to normalize it.

src/ngSanitize/sanitize.js
mhevery: Nope a mess up, cleaned up.

src/ngSanitize/sanitize.js
mhevery: I thought that `$window` is only in the main module, but maybe I am mistaken.

test/ngSanitize/sanitizeSpec.js
mhevery: removed

mhevery:
@IgorMinar I don't know how to fix the build. Can you have a look?

src/ngSanitize/sanitize.js
gkalpak: I meant get it through DI. But as @igor mentioned there isn't any real benefit either, so...

src/ngSanitize/sanitize.js
lgalfaso: I do not have access to any IE, but if you already tested this with IE, then it works for me

IgorMinar:
you didn't update docs and used the wrong object to throw the error.

pull the commit from this branch: https://github.com/angular/angular.js/compare/master...IgorMinar:sanitize-fix?expand=1

review the docs for correctness..

mhevery:
@IgorMinar merged your changes here, you can close your branch.

lgalfaso:
The sanitization process looks fine, but this PR makes some unexpected html behavior changes that change how html is rendered. eg

```js
it('should not duplicate <br> tags', function() {
  expectHTML('<br />').not.toEqual('<br></br>');
  expectHTML('<br>').not.toEqual('<br></br>');
});
```

IgorMinar:
This PR has been superseded by #12524

Issue #11775, opened 2015-04-30
Error message on directive namespace conflict


sod:
I got this error message today:

```
Error: [$compile:multidir] Multiple directives [pagination, pagination] asking for new/isolated scope on: <pagination>
http://errors.angularjs.org/1.4.0-rc.1/$compile/multidir?p0=pagination&p1=pagination&p2=new%2Fisolated%20scope&p3=%3Cpagination%3E
```

Took me a while to debug whats wrong with my `pagination` directive. Until I discovered another directive with the exact same name `pagination` from ui.bootstrap. It wasn't obvious as I didn't include `ui.boostrap` myself.

If the error message would include the module name the directive came from, this had helped a lot. Something like:

```
Error: [$compile:multidir] Multiple directives [pagination (module: foundation), pagination (module: ui.bootstrap)] asking for new/isolated scope on: <pagination>
```

Example code: http://plnkr.co/edit/Mgxutpvfr4ngqRD6fV1F?p=preview
Version: angular.js/1.4.0-rc.1

petebacondarwin:
Feel free to provide a PR for this. Otherwise we will try to get to it in 1.5. At the moment there is no link back from injectable components to their originating modules.

kentcdodds:
Trying my hand at a PR #11793. Would love some feedback to make sure I'm taking this the right direction. A little iffy on the idea of passing the module to all the invoke queue. But I can't think of any other way to accomplish this.

lugovsky:
@kentcdodds Probably there could be other solution to do this. I'm just not 100% sure if this is possible, cuz I'm fairly new to angular.js sources. 

I assume, that each directive is an object. If so, there could probably be a way to find module in which directive is defined by using object comparison. Even if this solution is not good in terms of performance, I think it could be quite okay during exception throwing.

And again, it's only an assumption, so don't judge me strictly please :)

As well I can provide pull request, if this solution is implementable and somebody will tell me which parts of this solution are not implemented at all.

petebacondarwin:
It would be better find a way to search the loaded modules for directives with a given name.

kentcdodds:
That could be reasonably done. I've [done this before](http://kent.doddsfamily.us/kcd-angular/#/) (the list on the left is auto-generated based on the contents of the `_invokeQueue`). That would also touch much less of the entire library.

lugovsky:
Yeah, I see that. But in your example you have root module hardcoded:
```javascript
var allTheThings = [];
[...]

PR #11793, opened 2015-05-02
PROOF OF CONCEPT #11775 (show module name for conflicting directives), DO NOT MERGE

kentcdodds:


src/auto/injector.js
```diff
@@ -722,13 +722,14 @@ function createInjector(modulesToLoad, strictDi) {
...
 
-      function runInvokeQueue(queue) {
+      function runInvokeQueue(queue, moduleFn) {
         var i, ii;
         for (i = 0, ii = queue.length; i < ii; i++) {
           var invokeArgs = queue[i],
               provider = providerInjector.get(invokeArgs[0]);
-
-          provider[invokeArgs[1]].apply(provider, invokeArgs[2]);
+          var args = Array.prototype.slice.call(invokeArgs[2]);
```
kentcdodds: Not so sure I like adding another arg... But can't think of another way to do it.

src/ng/compile.js
```diff
@@ -2299,7 +2300,15 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     function assertNoDuplicate(what, previousDirective, directive, element) {
       if (previousDirective) {
         throw $compileMinErr('multidir', 'Multiple directives [{0}, {1}] asking for {2} on: {3}',
-            previousDirective.name, directive.name, what, startingTag(element));
+            getDirectiveLabel(previousDirective), getDirectiveLabel(directive), what, startingTag(element));
+      }
+      
+      function getDirectiveLabel(dir) {
+        if (dir.ngModule) {
```
kentcdodds: Forced to do this check because sometimes the directive function can be called without passing an `ngModule`.

test/ng/compileSpec.js
```diff
@@ -2113,6 +2113,7 @@ describe('$compile', function() {
         var iscope;
 
         beforeEach(module(function() {
+          var fakeModule = {name: 'fakeModule'};
```
kentcdodds: Faking out a module to pass to the directive function so the test will show the result of passing a module.

PR #11794, opened 2015-05-03
chore(jscs): active disallowMultipleLineBreaks rule

sebholstein:


caitp:
not a fan of this, the vertical whitespace between blocks makes a lot of sense, and is used in other goog projects like closure-library and v8

SebastianM:
@caitp I'm not a fan of this too. But as this rule is listed in .jscs.json.todo I thought "it's an open TODO - I'll try to fix it".

caitp:
It makes perfect sense for linebreaks that aren't between function definitions/their surrounding comments --- but I'd prefer we enforce that there are two blank lines between function definitions. Still, since that will cause a lot of churn, it's probably not worth it

petebacondarwin:
I think we meant to remove that TODO list now. I think we agreed not to implement any more of those rules.

hzoo:
Yeah I think I forgot to remove the file a while ago - was from https://github.com/angular/angular.js/pull/10772#issuecomment-71631999.

SebastianM:
Ok guys, closing this PR

PR #11795, opened 2015-05-03
fix(ngOptions): fix frozen ui in ie with more than one select[multiple]

### LINKED ISSUE #11314: Bug. Select ng-options with multiple doesn't work IE10/11 when next to each other.
There seems to be an issue in IE10/11 using select with multiple attribute. The issue only seems to be happening when you have more than one select in a page. The unexpected behaviour renders all of the selects unresponsive minus the last one. It would appear to be a rendering issue, since if you toggle the visibility of the element it works as intended.

Here is a plnkr to reproduce the issue: http://plnkr.co/edit/E1asWesEj90OqAf5RIt2?p=preview

Narretz:
If there are more than one select, and the first select is wrappend
in an element with display: inline or display: inline-block, all but
the last select are completely unresponsive to any user input.

This cannot be tested in a unit-test, as the events must come directly
from the ui.

Close #11314 

This is pretty tricky, as we probably need an e2e test for this, but they currently don't run on IE. I'm also not sure if we can simply leave the label setting out.

@petebacondarwin should we wait with this until we have IE e2e tests?

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -576,11 +576,16 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
       function updateOptionElement(option, element) {
         option.element = element;
         element.disabled = option.disabled;
-        if (option.value !== element.value) element.value = option.selectValue;
+        // NOTE: The label must be set before the value, otherwise IE10/11/EDGE create unresponsive
+        // selects in certain circumstances when multiple selects are next to each other and display
+        // the option list in listbox style, i.e. the select is [multiple], or specifies a [size].
+        // See https://github.com/angular/angular.js/issues/11314 for more info.
+        // This is unfortunately untestable with unit / e2e tests
         if (option.label !== element.label) {
           element.label = option.label;
           element.textContent = option.label;
         }
+        if (option.value !== element.value) element.value = option.selectValue;
       }
 
       function addOrReuseElement(parent, current, type, templateElement) {


petebacondarwin:
If you can see that this fixes the bug on IE and that we are not breaking other browsers by removing then we should merge this. I am trying to remember why both assignments are there in the first place. I have a feeling it was getting around another IE bug.

Narretz:
Looks like you are right:
```
IE 9.0.0 (Windows Vista) ngOptions should update the label property FAILED
	Expected '' to equal 'A'.
	Expected '' to equal 'B'.
	Expected '' to equal 'C'.
IE 9.0.0 (Windows Vista) ngOptions should update the label if only the property has changed FAILED
	Expected '' to equal 'A'.
	Expected '' to equal 'B'.
	Expected '' to equal 'C'.
	Expected '' to equal 'X'.
```

petebacondarwin:
So we need to find a different solution...

Narretz:
I haven't looked into a different solution, but I created an e2e test - which doesn't work. It's seems like this is essentially untestable. I have this spec to click on an option inside a select:
```  
it('should be able to select the first select', function() {
    loadFixture("ngOptions").andWaitForAngular();
    element(by.cssContainingText('option', 'two')).click();
    expect(element(by.binding('selected1')).getText()).toBe('two');
  });
```
and this works! But the bug is that actual clicks from the ui don't work at all on the first select. So we can't even show the bug with an e2e test.

PR #11796, opened 2015-05-03
fix($compile): workaround for IE11 MutationObserver

### LINKED ISSUE #11781: Interpolations containing HTML entities fail in IE11 when MutationObservers are active
Text nodes that contain unicode / HTML entities are not properly interpolated in IE11 when there is a `MutationObserver` active.  Consider the following example.

```html
<!doctype html>
<html>
  <head>
    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.0-rc.1/angular.min.js"></script>
    <script>
      window.addMutationObserver = function(){
        if(!window.MutationObserver){
          return;
        }
        new window.MutationObserver(function(){}).observe(document.body, {
          childList: true,
          subtree: true
        });
      };

      angular.module('example', []);
    </script>
  </head>
  <body onload="addMutationObserver()" ng-app="example">
    <span>{{ 'This works.' }}</span>
    <span>&mdash; {{ "This doesn't." }}</span>
  </body>
</html>
```

The output is:

```This works. — {{ "This doesn't." }}```

If you remove the `MutationObserver` the code works as expected. If you don't set `subtree: true` the code works as expected. If you remove the entity it also works as expected.  If instead you replace the entity with the actual character (—) it remains broken.  It's also broken in 1.3.

We recently started seeing this bug because Wordpress has some Emoji polyfills that use `MutationObserver`.  This issue breaks interpolation even when the Angular application is running in an iframe and the code that is attaching the `MutationObserver` is in the *parent* frame, which seems pretty crazy. Because of this we seek a workaround on the Angular side, as we can't change every site that might embed Angular in an iframe -- perhaps this is a bug in IE.

### PR DESCRIPTION
**Pull request by lgalfaso**
IE11 MutationObserver breaks consecutive text nodes into several text nodes.
This patch merges consecutive text nodes into a single node before
looking for interpolations.

Closes #11781

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -2762,6 +2762,23 @@ describe('$compile', function() {
     }));
 
 
+    it('should handle consecutive text elements as a single text element', inject(function($rootScope, $compile) {
+      window.addMutationObserver = function() {
+        if (!window.MutationObserver) {
+          return;
+        }
+        new window.MutationObserver(function() {}).observe(document.body, {
+          childList: true,
+          subtree: true
+        });
+      };
+      var base = jqLite('<div>&mdash; {{ "This doesn\'t." }}</div>');
+      element = $compile(base)($rootScope);
+      $rootScope.$digest();
+      expect(element.text()).toBe("— This doesn't.");
+    }));
+
+
     it('should support custom start/end interpolation symbols in template and directive template',
         function() {
       module(function($interpolateProvider, $compileProvider) {
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1540,6 +1540,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
           break;
         case NODE_TYPE_TEXT: /* Text Node */
+          if (msie === 11) {
+            // Workaround for #11781
+            while (node.parentNode && node.nextSibling && node.nextSibling.nodeType === NODE_TYPE_TEXT) {
+              node.nodeValue = node.nodeValue + node.nextSibling.nodeValue;
+              node.parentNode.removeChild(node.nextSibling);
+            }
+          }
           addTextInterpolateDirective(directives, node.nodeValue);
           break;
         case NODE_TYPE_COMMENT: /* Comment */

### CODE REVIEW

**Comment by petebacondarwin** [comment]
Does this only happen on IE11 or other versions of IE?
Can you wrap this code in a conditional: `if(msie===11) {...}`?

**Comment by lgalfaso** [comment]
@petebacondarwin AFAICT, this is IE11 only and when there is a MutationObserver

**Comment by lgalfaso** [comment]
BTW, the test as is fails on all browsers. The main issue is that only IE11 generates multiple consecutive text elements/

File: test/ng/compileSpec.js [hunk 1]
```diff
@@ -2762,6 +2762,16 @@ describe('$compile', function() {
     }));
 
 
+    it('should handle consecutive text elements as a single text element', inject(function($rootScope, $compile) {
+      var base = jqLite('<div></div>');
+      base[0].appendChild(document.createTextNode('Lucas {{1+'));
+      base[0].appendChild(document.createTextNode('2}} Lucas'));
```
petebacondarwin: I think we should test the specific situation that causes the bug, i.e. that there is a mutation observer and that the text contains some special characters that trigger IE11 to split it into multiple text nodes.

lgalfaso:
@petebacondarwin can you take a second look at this?

test/ng/compileSpec.js
```diff
@@ -2762,6 +2762,23 @@ describe('$compile', function() {
     }));
 
 
+    it('should handle consecutive text elements as a single text element', inject(function($rootScope, $compile) {
+      window.addMutationObserver = function() {
```
petebacondarwin: Should we really be modifying the `window` object here?

petebacondarwin:
Other than my comment about `window` this looks good to me.

test/ng/compileSpec.js
gkalpak: Does this have any effect anyway ? Shouldn't we call the function as well ?
(And indeed there doesn't seem to be a reason to modify `window`.)

lgalfaso: The test, without fiddling with `window` passes without the fix on IE 11.
IE11 parses text node differently when there is a mutation observer, this is why the test add it.

gkalpak: But if the function isn't called, then there is no MutationObserver added right ?
How does the function help if we don't call it ?
And we does the function need to be assigned on window ?

lgalfaso: @gkalpak I think I am missing something. This is a test. IE11 parses html different when there is any mutation observer registered. The test reproduces this scenario.

gkalpak: If I recall correctly, the issue appeared when there was a childList+subtree MutationObserver **registered** on the page. I would expect a test that calls `new MutationObserver(...).observe(...)`.

Instead I see:

```js
window.addMutationObserver = function () {
  // A function that _would_ register a MutationObserver on body
  // IF it where executed.
};

// Then I see the actual test code being executed without ever calling `window.addMutationObserver()`
// which would have registered the MutationObserver.
```

So, afaict, there is **no** MutationObserver during this test, so it would pass even without the fix (I am just speculating - I haven't actually tried it).

--
I would expect something along these lines:

```js
it ('...', inject(function($compile, $rootScope) {
  // No point it running thetest, if there is no MutationObserver
  if (!window.MutationObserver) return;

  // Create and register the MutationObserver
  var observer = new window.MutationObserver(noop);
  observer.observe(document.body, {childList: true, sublist: true});

  // Run the actual test
  var base = jqLite('<div>&mdash; {{ "This doesn\'t." }}</div>');
  element = $compile(base)($rootScope);
  $rootScope.$digest();
  expect(element.text()).toBe("— This doesn't.");

  // Unregister the MutationObserver (and hopeit doesn't mess up with subsequent tests)
  observer.disconnect();
}));
```

lgalfaso: Updated the test

lgalfaso:
landed as f3b1d0b723298a5f8ea21d0704405649cce1b5fc

test/ng/compileSpec.js
gkalpak: @lgalfaso, I noticed that the added test would pass on IE11 even without the fix.
Looking into it, it turned out I (unwittingly) tricked you into changing `subtree: true` to `sublist: true`, which effectively makes the test useless :disappointed: 

Submitted a fix in #12061.

PR #11800, opened 2015-05-04
fix(core): ensure that multiple requests to requestAnimationFrame are buffered

### LINKED ISSUE #11791: ngAnimate makes IE11 grind to a halt
This jsfiddle (http://jsfiddle.net/oga2z7bb/3/) uses ng-repeat to render 10000 numbers. This jsfiddle (http://jsfiddle.net/oga2z7bb/4/) does the same thing but includes ngAnimate.

Both examples work fine in Chrome & Firefox. The first example also works in IE11, however if you open http://jsfiddle.net/oga2z7bb/4/ in IE11 your browser should stop responding. I've profiled (with 1 000 numbers instead of 10 000) this and it seems that 99% of the time is spent in ```window.requestAnimationFrame()```.

I use ng-repeat to render a calendar in my app. With IE11 it is effectively unusable unless I do some browser sniffing and not load ngAnimate when IE is detected.

matsko:
IE11 (and maybe some other browsers) do not optimize multiple calls to
rAF. This code makes that happen internally within the $$rAF service
before the next frame kicks in.

Closes #11791

### DIFF
--- a/src/ng/raf.js
+++ b/src/ng/raf.js
@@ -10,7 +10,7 @@ function $$RAFProvider() { //rAF
                                $window.webkitCancelRequestAnimationFrame;
 
     var rafSupported = !!requestAnimationFrame;
-    var raf = rafSupported
+    var rafFn = rafSupported
       ? function(fn) {
           var id = requestAnimationFrame(fn);
           return function() {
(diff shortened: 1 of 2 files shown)


src/ng/raf.js
```diff
@@ -24,8 +24,42 @@ function $$RAFProvider() { //rAF
...
+    var taskCount = 0;
+    var taskQueue = [];
+    return queueFn;
+
+    function flush() {
+      for (var i = 0; i < taskQueue.length; i++) {
+        var task = taskQueue[i];
+        if (task) {
+          taskQueue[i] = null;
+          taskCount--;
```
gkalpak: This seems redundant.

src/ng/raf.js
```diff
@@ -24,8 +24,42 @@ function $$RAFProvider() { //rAF
...
+      var index = taskQueue.length;
+
+      taskCount++;
+      taskQueue.push(asyncFn);
+
+      if (index === 0) {
+        cancelLastRAF = rafFn(flush);
+      }
+
+      return function() {
```
gkalpak: This should be named (for debugging purposes etc).

src/ng/raf.js
```diff
@@ -24,8 +24,42 @@ function $$RAFProvider() { //rAF
...
+        taskQueue[index] = null;
+        if (--taskCount === 0 && cancelLastRAF) {
+          cancelLastRAF();
+          cancelLastRAF = null;
+        }
+      };
+    }
```
gkalpak: This function seems broken altogether.

What if it is called **after** the next raf ?
Won't it cancel another unrelated task ?

Also, cancelling will not properly empty the queue (it will leave it full of `null`s).

src/ng/raf.js
matsko: The `taskCount` variable is essential to know when to cancel the underlying RAF request when there are no more queued rAF functions.

gkalpak: But here you are running a loop that decreases `taskCount` and right after that you srt it to `0`.
So, decrementing it is redundant.

matsko: Ah yes. Good call.

matsko:
LGTM (from Igor in person).

matsko:
Merged in as https://github.com/angular/angular.js/commit/db20b830fc6074a00dc11d3f47d665c55e8bb515

gkalpak:
What does Igor in person think about `cancelQueueFn` being called at a later time and cancelling a titally unrelated animation (just because the `index` is the same) ?

Did I miss something ?

PR #11807, opened 2015-05-05
Enforce a best practice on the `ng-animate` CSS class

### LINKED ISSUE #11431: $animateProvider.classNameFilter removing class name for leave event that follows an enter event
I'm building a notification system and i set

```
$animateProvider.classNameFilter(/ng-animate/);
```

to prevent conflict with other components in website. When the animation comes in everything work fine, and ng-enter class provide the animation, but it seems that filter class ng-animate is removed from the div and hence the ng-leave class is not been applied!! With you duplicate the ng-animate class in the component you can work around the problem!! Check out this plunker to see what i'm talking about: http://plnkr.co/edit/7IEgccUvteHFCGiML6Fy?p=preview

The bounce in animation works fine when enter, but when leaving it does not trigger the bounce out animation!

matsko:
Since ngAnimate uses the `ng-animate` CSS class internally to track
state it is better to keep this as a reserved CSS class to avoid
accidentally adding / removing the CSS class when an animation is
started and closed.

BREAKING CHANGE: partially or fully using a regex value containing
`ng-animate` as a token is not allowed anymore. Doing so will trigger a
minErr exception to be thrown.

So don't do this:

``` js
// only animate elements that contain the `ng-animate` CSS class
$animateProvider.classNameFilter(/ng-animate/);

// or partially contain it
$animateProvider.classNameFilter(/some-class ng-animate another-class/);

// but this is OK
$animateProvider.classNameFilter(/ng-animate-special/);
```

Closes #11431

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -2,6 +2,7 @@
 
 var $animateMinErr = minErr('$animate');
 var ELEMENT_NODE = 1;
+var NG_ANIMATE_CLASSNAME = 'ng-animate';
 
 function mergeClasses(a,b) {
   if (!a && !b) return '';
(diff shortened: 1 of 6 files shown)


mzgol:
:+1:

Issue #11808, opened 2015-05-05
Add "any" method on $q service


candreoliveira:
$q is a service inspired by [Kris Kowal's Q](https://github.com/kriskowal/q). 

It's usual in frontend applications make requests to some APIs and based on (at least) an answer continue processing the business logic. 

For example: Imagine an e-commerce application. It almost always needs to receive coupons, vouchers and/or freight discounts. A good architecture will separate responsabilities into distinct micro-services. However, the "perfect" ui could be a single input text. So, we'd use a single input to dispatch 3 requests (3 end-points: coupon, voucher and freight discount) and based on any answer apply the user discount. Hence, we need the ability to manage all requests into a single promise.

The Kris Kowal's Q has the method "any" and I think It's very helpful. What do you think about it @IgorMinar ?

SanderElias:
@candreoliveira, I think this is a bad idea. There is the [.race](https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) in the ES standard, it's an better idea to implement that. 

I like the functionality, but if there is a documented standart, it's better to comply to that right?

jrencz:
@candreoliveira I can confirm `$q.all` fulfils your needs: In the example you've given you definitely want to wait until all services respond: All in all you need to know whether user has 1, 2 or 3 sources of discount and you want to sum up the discount. 

@SanderElias I see no cons of adding both `.any` and `.race` - they have different semantics: `.race` will get rejected after the first one of the promises given to it gets rejected. `.all` will wait for all of them to get rejected.

There's #12929 for race

jrencz:
@SanderElias I was wrong about the semantics of `.all` and rejection: `.all` gets rejected with a reason the first of rejected promise got rejected. Still `.all`, `.race` and `.any` have different semantics clearly.

`.any` with semantics as described by @candreoliveira in [his PR](https://github.com/candreoliveira/angular.js/commit/7675d2e710c8d3f1e7dc69cb4a3d941f47e75390#diff-0cfb390955e411a4a5437fa337d52481R1824) seems to be a `.race` of `.race` and a promise constructor call (using ES6 syntax for brevity):

```javascript
// This code is not meant for production!
// It assumes 
angular.module('myModule').config(function ($provide) {
  $provide.decorator('$q', function ($delegate) {
    $delegate.race = () => {}; // Let's assume it's implemented according to the Promise spec

    $delegate.any = promises => $delegate.race([
[...]

PR #11809, opened 2015-05-05
feat(q): Resolve any promise of array/hash input (q.any)

candreoliveira:
Issue reference: https://github.com/angular/angular.js/issues/11808

$q is a service inspired by Kris Kowal's Q.

It's usual in frontend applications to make requests to some APIs and based on (at least) a response continue processing the business logic.

For example: An e-commerce almost always needs to receive coupons, vouchers and/or freight discounts. A good architecture will separate responsabilities into distinct micro-services. However, the "perfect" ui could be a single input text. So, we'd use a single input to dispatch 3 requests (3 end-points: coupon, voucher and freight discount) and based on any answer apply the user discount. Hence, we need the ability to manage all requests into a single promise.

The Kris Kowal's Q has the method "any" and I think It's very helpful.

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -515,23 +515,61 @@ function qFactory(nextTick, exceptionHandler) {
 
   function all(promises) {
     var deferred = new Deferred(),
-        counter = 0,
-        results = isArray(promises) ? [] : {};
+        counter  = 0,
+        results  = isArray(promises) ? [] : {};
 
     forEach(promises, function(promise, key) {
       counter++;
       when(promise).then(function(value) {
-        if (results.hasOwnProperty(key)) return;
         results[key] = value;
         if (!(--counter)) deferred.resolve(results);
       }, function(reason) {
-        if (results.hasOwnProperty(key)) return;
         deferred.reject(reason);
       });
     });
 
+    // Empty promises array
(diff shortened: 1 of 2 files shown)


realityking:
Since Angular is aligning it's promises with ES6, I think it would be better to name it after [`Promise.race`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race).

motin:
+1 on including this, albeit called $q.race() as per @realityking's suggestion

twilczak:
I think this feature as implemented is fundamentally different than Promise.race, and probably shouldn't use that name. It don't see a native feature for this in ES6.

ryanray:
+1

jrencz:
Note that `.any` as implemented here has different semantics than `.race` as in the Spec.

`.race` will get rejected after the first one of the promises given to it gets rejected. `.all` will wait for all of them to get rejected.

This implementation cannot be named `.race`.

There's #12929 for race

petebacondarwin:
We are going to implement `$q.race()` (because it is part of the ES6 Promise spec) and only document how one could implement one's one version of `any()` (because it is not part of the spec).

PR #11814, opened 2015-05-05
fix(ngClass): add/remove classes which are properties of Object.prototype

### LINKED ISSUE #11813: Firefox ignores 'watch' class name in ng-class directive.
I'm not sure if this is an Angular or Firefox bug (or maybe I'm just missing something else) because no other browsers are having this problem...

But given a code snippet like:

```
<div ng-class="{className: true, watch: true, sympathy: true}">
```

Firefox will generate:
```
<div class="className sympathy"><div>
```

You can see this using the following Pen (Angular 1.3.2):
http://codepen.io/joshmatz/pen/Xbmame

caitp:
- fast paths for various internal functions when createMap() is used
- Make createMap() safe for internal functions like copy/equals/forEach
- Use createMap() in more places to avoid needing hasOwnProperty()

Closes #11813

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -36,6 +36,7 @@
   isUndefined: true,
   isDefined: true,
   isObject: true,
+  isBlankObject: true,
   isString: true,
   isNumber: true,
   isDate: true,
@@ -175,6 +176,7 @@ var
     splice            = [].splice,
     push              = [].push,
     toString          = Object.prototype.toString,
+    getPrototypeOf    = Object.getPrototypeOf,
     ngMinErr          = minErr('ng'),
 
     /** @name angular */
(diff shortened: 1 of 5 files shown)


matsko:
Mind rewriting the commit message to clearly explain what the problem is/was and what the patch is fixing. There's a lot of domain-specific stuff in here that is hard to understand.

caitp:
@matsko PTAL, I've made the commit message more thorough and refactored it a bit.

src/Angular.js
```diff
@@ -267,12 +269,25 @@ function forEach(obj, iterator, context) {
       }
     } else if (obj.forEach && obj.forEach !== forEach) {
         obj.forEach(iterator, context, obj);
-    } else {
+    } else if (isBlankObject(obj)) {
+      // createMap() fast path --- Safe to avoid hasOwnProperty check because prototype chain is empty
+      for (key in obj) {
+        iterator.call(context, obj[key], key, obj);
+      }
+    } else if (typeof obj.hasOwnProperty === 'function') {
```
gkalpak: We probably don't care, but shadowing `hasOwnProperty` with a function will not be handled correctly.
<SUB>Just saying...</SUB>

caitp: Depends on how it's shadowed --- I don't think we care

caitp:
huh --- I guess this patch discovered a new IE quirk :<

caitp:
ping

caitp:
pong

matsko:
LGTM. Excellent work.

caitp:
thanks

PR #11822, opened 2015-05-06
feat(ngAnimate): generate shared anchor classes between anchor elements as well

matsko:
Prior to this fix when anchored CSS animations were put to use, only the
shared CSS classes between **container** elements were suffixed and
applied to the anchor element. This patch now ensures that in addition
to the container classes, the shared classes between the starting anchor
element and the destination anchor element are prefixed and applied as
well.

### DIFF
--- a/src/ngAnimate/animateCssDriver.js
+++ b/src/ngAnimate/animateCssDriver.js
@@ -35,17 +35,14 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
       return classes.replace(/\bng-\S+\b/g, '');
     }
 
-    function getUniqueValues(a, b) {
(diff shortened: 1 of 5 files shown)


src/ngAnimate/animateCssDriver.js
```diff
@@ -142,7 +139,7 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
 
       function prepareInAnimation() {
         var endingClasses = filterCssClasses(inAnchor.attr('class'));
```
gkalpak: `getClassVal()` here too ?

matsko:
This has been closed in favour of: https://github.com/angular/angular.js/pull/11825

PR #11826, opened 2015-05-07
fix($animate): silently ignore the options param if an object is not passed in

### LINKED ISSUE #11713: (bug) ngAnimate options.domOperation is not a function
Few days ago at production error occured:

TypeError: options.domOperation is not a function
options.domOperation();
angular...mate.js (line 2747, col 8)

https://github.com/angular/angular.js/blob/master/src/ngAnimate/animateQueue.js#L432

First I don't have this error at local machine, but after update Bower dependencies(bower update) I got it.

Here is my bower.json
...
"angular": "~1.4.0",
"angular-animate": "~1.4.0",
...

matsko:
Earlier versions of Angular expected a function to be passed into the
same param as the options value. This causes a nasty issue since the
internal animation code expects the options value to be an object
instead of a function. This patch adds the code to convert a function
value into an empty object when that occurs.

Closes #11713

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -486,6 +486,54 @@ describe("animations", function() {
       });
     });
 
+    they('$prop() shouldn\'t blow up when a function is passed as the options param',
+      ['enter', 'move', 'leave', 'addClass', 'removeClass', 'setClass', 'animate'], function(event) {
+
+      inject(function($animate, $rootScope, $document) {
+        var element = $document[0].createElement('div');
+        element.setAttribute('id', 'crazy-man');
+        if (event !== 'enter' && event !== 'move') {
+          parent.append(element);
+        }
+
+        var fakeOption = function() { };
+
+        switch (event) {
+          case 'enter':
+          case 'move':
+            $animate[event](element, parent, parent2, fakeOption);
+            break;
+
+          case 'addClass':
+            $animate.addClass(element, 'klass', fakeOption);
+            break;
+
+          case 'removeClass':
+            element.className = 'klass';
+            $animate.removeClass(element, 'klass', fakeOption);
+            break;
+
+          case 'setClass':
+            element.className = 'two';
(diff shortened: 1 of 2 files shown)


Narretz:
Does it make sense to remove this first thing in 1.5 and add it as a breaking change?

Narretz:
We can also add a deprecation notice for this.

test/ngAnimate/animateSpec.js
```diff
@@ -486,6 +486,54 @@ describe("animations", function() {
       });
     });
 
+    they('$prop() shouldn\'t blow up when a function is passed as the options param',
+      ['enter', 'move', 'leave', 'addClass', 'removeClass', 'setClass', 'animate'], function(event) {
+
+      inject(function($animate, $rootScope, $document) {
+        var element = $document[0].createElement('div');
+        element.setAttribute('id', 'crazy-man');
```
Narretz: Is that left over from the previous test? The id is not used.

petebacondarwin:
It looks like this callback was removed in https://github.com/matsko/angular.js/commit/bf0f5502b1bbfddc5cdd2f138efd9188b8c652a9

This is the breaking change for 1.4.0, and it is documented as such in the commit, since if you were relying on the callback to be called this no longer happens.

The question is whether we should let it slip through by merging this code. I feel bad doing so because, now instead of an error your callback just never gets called.

How about we add an assertion on that parameter that says that it cannot be a `function` then document the breaking change in the error docs for it? This would encourage people to upgrade to using promises and also highlight to them that their code is not going to work as expected.

matsko:
Works for me. @petebacondarwin would be able to take this one on for me?

petebacondarwin:
@matsko: will do

matsko:
Note that we had to revert this and add the 1.3 behaviour of ignoring non-object properties since UI-router, AngularStrap and other libraries still rely on it.

This commit makes the change:
https://github.com/angular/angular.js/commit/72edd4dff931c644eecb8f0d1c878dc839c76947

PR #11829, opened 2015-05-07
feat($compile): show module name during multidir error

### LINKED ISSUE #11775: Error message on directive namespace conflict

lugovsky:
Show module name if possible when multidir error happens.

Closes #11775

P.S. Hey guys, please check out this solution. Let me know if the way I implemented it looks correct at least from afar.
Honestly, I've spent a lot of time trying to understand the way, that affects core functionality as less as possible and I think that's it. But if you see some another solution, I will be glad to hear that. 
In case you think that these changes are critical and this feature doesn't worth it, I'm ready to this as well :).

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -146,7 +146,7 @@ function setupModuleLoader(window) {
            * @description
            * See {@link auto.$provide#provider $provide.provider()}.
            */
-          provider: invokeLater('$provide', 'provider'),
+          provider: invokeLaterAndSetModuleName('$provide', 'provider'),
 
           /**
            * @ngdoc method
@@ -157,7 +157,7 @@ function setupModuleLoader(window) {
            * @description
            * See {@link auto.$provide#factory $provide.factory()}.
            */
-          factory: invokeLater('$provide', 'factory'),
+          factory: invokeLaterAndSetModuleName('$provide', 'factory'),
 
           /**
            * @ngdoc method
@@ -168,7 +168,7 @@ function setupModuleLoader(window) {
            * @description
            * See {@link auto.$provide#service $provide.service()}.
            */
-          service: invokeLater('$provide', 'service'),
+          service: invokeLaterAndSetModuleName('$provide', 'service'),
 
           /**
            * @ngdoc method
@@ -203,7 +203,7 @@ function setupModuleLoader(window) {
            * @description
            * See {@link auto.$provide#decorator $provide.decorator()}.
            */
-          decorator: invokeLater('$provide', 'decorator'),
+          decorator: invokeLaterAndSetModuleName('$provide', 'decorator'),
 
           /**
            * @ngdoc method
@@ -237,7 +237,7 @@ function setupModuleLoader(window) {
            * See {@link ng.$animateProvider#register $animateProvider.register()} and
            * {@link ngAnimate ngAnimate module} for more information.
            */
-          animation: invokeLater('$animateProvider', 'register'),
+          animation: invokeLaterAndSetModuleName('$animateProvider', 'register'),
 
           /**
            * @ngdoc method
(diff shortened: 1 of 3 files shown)


kentcdodds:
Looks good to me :-)

lgalfaso:
I see this as a real issue, but there are a few things that needs some cleanup.

- It exposes angular specific properties with a `_` prefix and not with `$$`
- Exposing the factory looks like a bad idea, why not expose the module name itself?
- A test when the module name is missing, is missing :)
- (nice to have) A test with a directive that has an interceptor


All this a side, I think this solves a real struggle

lugovsky:
Sorry guys for the delay in response. Unfortunately I've caught some illness which didn't let me do anything.
@lgalfaso I changed code to achieve your points 1 and 3. I didn't actually get about what interceptor you told in point 4.
There is one problem implementing point 2. Unfortunately directive factory doesn't have any back reference to module in which it was defined. As you can see, in my original solution, to understand in which module the directive was defined I'm doing a search through invokeQueue's of all loaded modules. The idea was, that the code was executing only when failure was detected, so it shouldn't affect performance. But if I will use that search every time directive object is creating, that would definitely affect performance. So that's what I don't want to do :). And I don't see any good solution here.
One the other hand, there is some other solution. In that solution small modifications to loader should be done: special case of invokeLater should be created for directive. I've made some sample here https://github.com/akveo/angular.js/commit/6eb899e9a2eb85d7b6fac622a51d70778d094095 . That as well will put most of code I created in this PR to the trash bin, but as well will reduce the complexity of this task.

lgalfaso:
@lugovsky I like the path that https://github.com/akveo/angular.js/commit/6eb899e9a2eb85d7b6fac622a51d70778d094095 is going. In fact if this would be added strait to `invokeLater` (with the exception of `config`, `constant` and `value`). I also think that this would make this PR a lot cleaner

lugovsky:
@lgalfaso Did I got you correctly and you wish to add that code straight to invokeLater, so every factory/service/provider/decorator/animation/filter/controller (except config/constant/value) has string representing module in which it was defined as well? Do you want to provide additional extensibility for those recipies in the future?

lgalfaso:
@lugovsky yes, and yes. The later is to show in the error the module when a dependency is not fulfill.

Just to be sure, I would like to keep the current scope of this PR.

In the future (PR welcome) showing the module when a dependency is not fulfill would be nice.

lugovsky:
@lgalfaso Made updates according to our discussion. I like it much more this way actually :). Please make another review and suggest new changes if any.

If you want me to handle showing the module when a dependency is not fulfilled, please create or find the issue with the appropriate description :). Honestly, I haven't got the idea. But anyway I will be glad to help you with this problem!

lgalfaso:
LGTM, will merge after the cut

lugovsky:
@lgalfaso Awesome, thanks!

PS: If you need help with that problem when dependency is not fulfilled, just create the issue after the merge and reference me. I will be happy to help :)

lgalfaso:
landed as 351fe4b79c50a45a11af2fcd2aa7b6fd3b70058d

PR #11833, opened 2015-05-07
docs(ngAnimate): Change event to ng-enter

stovenator:
The event needs to map to the css class, or else the css class won't fire

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11833)

### DIFF
--- a/src/ngAnimate/module.js
+++ b/src/ngAnimate/module.js
@@ -364,7 +364,7 @@
  *   return {
  *     enter: function(element, doneFn) {
  *       var animation = $animateCss(element, {
- *         event: 'enter'
+ *         event: 'ng-enter'
  *       });
  *
  *       if (animation) {
@@ -390,7 +390,7 @@
  *   return {
  *     enter: function(element, doneFn) {
  *       var animation = $animateCss(element, {
- *         event: 'enter',
+ *         event: 'ng-enter',
  *         addClass: 'maroon-setting',
  *         from: { height:0 },
  *         to: { height: 200 }


stovenator:
I signed it!

petebacondarwin:
@matsko - can you check this is correct and merge?

gkalpak:
This doesn't seem correct (unless the documentation is wrong).
But there is definitely something that needs fixing.

petebacondarwin:
It seem to me that the original wording is correct. See https://github.com/angular/angular.js/blob/master/src/ngAnimate/animateCss.js#L185
In other words, the `event` is mapped to the "DOM event", and that the class is then generated from this.

gkalpak:
The problem is that in practice, it doesn't work as described in the docs (but I think there is another issue about this).

petebacondarwin:
Can you provide a demo of this not working?
This line of code here seems to follow what the docs say: 
https://github.com/angular/angular.js/blob/master/src/ngAnimate/animateCss.js#L528

gkalpak:
I am pretty sure there was another issue (where I had commented) that showed the issue with `enter` and `enter-active` being added (without the `ng-` prefix). But I can't find it.

@matsko has pushed some fixes that may have taken care of the issue. I'll check and create a demo if it hasn't been fixed on master.

gkalpak:
I found it: the issue is #12202 (which is not directily related, but lead me to noticing the unexpected(?) behaviour).

Here is the [demo][1], I created for that issue.
(I would expect `ng-enter`/`ng-enter-active` to be added, but `enter`/`enter-active` are added instead. The same happens with the current snapshot.)

[1]: http://codepen.io/ExpertSystem/pen/KpybXO?editors=011

PR #11841, opened 2015-05-08
docs(tutorial/3 - Filtering Repeaters): Missing a variable declaration

tdyck12345:
Need this variable to make the unit test work


tdyck12345:
I signed it!

gkalpak:
But the variable is already defined on [line 160][1].

[1]: https://github.com/tdyck12345/angular.js/blob/patch-1/docs/content/tutorial/step_03.ngdoc#L160

Narretz:
I think so, too.

PR #11846, opened 2015-05-10
Fix bug in JSON vulnerability regex

damoncasale:
The old regex fails on Windows or Mac carriage returns/linefeeds.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -7,7 +7,7 @@ var JSON_ENDS = {
   '[': /]$/,
   '{': /}$/
 };
-var JSON_PROTECTION_PREFIX = /^\)\]\}',?\n/;
+var JSON_PROTECTION_PREFIX = /^\)\]\}',?\r?\n/;
 
 function serializeValue(v) {
   if (isObject(v)) {


damoncasale:
I signed the CLA.

lgalfaso:
Would it be possible for you to add a test that shows the vulnerability? Thanks!

gkalpak:
@lgalfaso, I don't think there is a vulnerability. It's a fix on the regex for removing the _vulnerability protection prefix_, so without the fix the prefix would fail to be removed and the data would fail to be parsed as JSON (if the newline is `\r\n` instead of `\n`).

If we decide we want the fix, I agree there should be a test failing (showcasing the failure to parse the data without the fix). (Just pointing out this is not a security fix.)

lgalfaso:
I think I am missing something, the current regex is
`JSON_PROTECTION_PREFIX = /^\)\]\}',?\n/;`
this is, it will understand the strings `"^)]}',\n"` and the string `"^)]}'\n"` as valid prefixes (check that the optional is on `","` not on `"\n"` (this was added at fe633dd0cf3d52f84ce73f486bcbd4e1d3058857). This change adds `"\r"` as optional and keeps the `"\n"` as mandatory. If the idea is that this las part can be any of *CR*, *LF* or *CR LF*, then the PR looks wrong as it is doing something else.

BTW, the test is really important here.

gkalpak:
I think the purpose of the PR is (should be) to add support for _CRLF_ (Window-style), in addition to the already supported _LF_.
I am not sure about Macs, but ([from what I read][1]) OS X seems to be using _LF_ as well (or not).

Not sure why the OP mentions `Mac`, but this PR seems valid for "fixing" Windows-style endings.

--
I wonder if this is necessary though, since this prefix is programmatically (and explicitly) added by the server (it's not like anyone will type the character-sequence in a program that will then add OS-specific line-breaks).
AFAICT, this fix shouldn't be necessary.

[1]: http://superuser.com/questions/439440/did-mac-os-lion-switch-to-using-line-feeds-lf-n-for-line-breaks-instead-of#answer-439443

lgalfaso:
I have to agree with @gkalpak this prefix is added programmatically and with this protection in mind, and also agree that this fix does not look as needed.

lgalfaso:
Without some context on why this is needed, then I think this is not an issue. If you are able to provide a context were this is needed, feel free to reopen this issue

PR #11852, opened 2015-05-11
fix($animateCss): ensure that custom durations do not confuse the gcs cache

### LINKED ISSUE #11723: Weird behavior of $animateCss with ng-if and quick toggling of elements in 1.4.0rc1
Angular version 1.4.0rc1

Old

I wrote a custom javascript animation and applied it to `ng-if`:
```js
mod.animation('.slide-margin', function($animateCss) {
    var animation = {
      enter : function(element, done) {
        var height = element[0].offsetHeight;
        var animator = $animateCss(element, {
          from: {
            marginTop: '-' + height + 'px'
          },
          to: {
            marginTop: '0px'
          },
          duration: 0.5
        });
        animator.start().finally(done);
      },
      leave : function(element, done) {
        var height = element[0].offsetHeight;
        var animator = $animateCss(element, {
          from: {
          },
          to: {
            marginTop: '-' + height + 'px'
          },
          duration: 0.3
        });
        animator.start().finally(done);
      },
      move : function(element, done) {
      },
      addClass : function(element, className, done) {
      },
      removeClass : function(element, className, done) {
      }
    };
    return animation;
  });
```

```jade
a(ng-click='show=!show') CLICK
br
br
div.slide-margin(ng-if='show')
  p A
  p B
  p C
```

Simple behavior is correct when you toggle an item, it slides down and if you click again in the middle of the animation it slides back from the same position. But if you click yet again for it mid-animation to slide down ng-if creates a new DOM content and weirdly enough old DOM content starts animating down along with the new content. Is this expected behavior? Do you know of a workaround?

matsko:
Closes #11723

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -621,7 +621,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
 
       var timings = computeTimings(node, fullClassName, cacheKey);
       var relativeDelay = timings.maxDelay;
-      maxDelay = Math.max(relativeDelay, 0);
+      maxDelay = Math.max(relativeDelay, 0) || 0;
       maxDuration = timings.maxDuration;
 
       var flags = {};
@@ -660,7 +660,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       // we need to recalculate the delay value since we used a pre-emptive negative
       // delay value and the delay value is required for the final event checking. This
       // property will ensure that this will happen after the RAF phase has passed.
-      if (timings.transitionDuration > 0) {
+      if (options.duration == null && timings.transitionDuration > 0) {
         flags.recalculateTimingStyles = flags.recalculateTimingStyles || isFirst;
       }
 
@@ -849,7 +849,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
 
             timings = computeTimings(node, fullClassName, cacheKey);
             relativeDelay = timings.maxDelay;
-            maxDelay = Math.max(relativeDelay, 0);
+            maxDelay = Math.max(relativeDelay, 0) || 0;
             maxDuration = timings.maxDuration;
 
             if (maxDuration === 0) {
@@ -866,7 +866,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
                   ? parseFloat(options.delay)
                   : relativeDelay;
 
-            maxDelay = Math.max(relativeDelay, 0);
+            maxDelay = Math.max(relativeDelay, 0) || 0;
 
             var delayStyle;
             if (flags.applyTransitionDelay) {
--- a/test/ngAnimate/animateCssSpec.js
+++ b/test/ngAnimate/animateCssSpec.js
@@ -1203,6 +1203,43 @@ describe("ngAnimate $animateCss", function() {
       });
     });
 
+    it('should avoid applying the same cache to an element a follow-up animation is run on the same element',
+      inject(function($animateCss, $rootElement, $document) {
+
+      function endTransition(element, elapsedTime) {
+        browserTrigger(element, 'transitionend',
+          { timeStamp: Date.now(), elapsedTime: elapsedTime });
+      }
+
+      function startAnimation(element, duration, color) {
+        $animateCss(element, {
+          duration: duration,
+          to: { background: color }
+        }).start();
+        triggerAnimationStartFrame();
+      }
+
+      var element = jqLite('<div></div>');
+      $rootElement.append(element);
+      jqLite($document[0].body).append($rootElement);
+
+      startAnimation(element, 0.5, 'red');
+      expect(element.attr('style')).toContain('transition');
+
+      endTransition(element, 0.5);
+      expect(element.attr('style')).not.toContain('transition');
+
+      startAnimation(element, 0.8, 'blue');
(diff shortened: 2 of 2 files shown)


src/ngAnimate/animateCss.js
```diff
@@ -742,6 +742,9 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
         applyAnimationClasses(element, options);
         applyAnimationStyles(element, options);
 
+        // we increment the counter so that if the same element is animated
+        // then the existing cache will not apply to it
+
```
petebacondarwin: I don't understand the context of this comment. Where is the counter that is being incremented?

test/ngAnimate/animateCssSpec.js
```diff
@@ -1203,6 +1203,47 @@ describe("ngAnimate $animateCss", function() {
...
+        { timeStamp: Date.now(), elapsedTime: 0.5 });
+
+      expect(element.attr('style')).toContain('transition');
+
+      browserTrigger(element, 'transitionend',
+        { timeStamp: Date.now(), elapsedTime: 0.8 });
+
+      expect(element.attr('style')).not.toContain('transition');
+    }));
+
```
petebacondarwin: This test is hard to follow. How about tidying it up as follows:

```
    it('should avoid applying the same cache to an element a follow-up animation is run on the same element',
      inject(function($animateCss, $rootElement, $document) {

      function endTransition(element, elapsedTime) {
        browserTrigger(element, 'transitionend',
          { timeStamp: Date.now(), elapsedTime: elapsedTime });
      }

      function startAnimation(element, duration, color) {
        $animateCss(element, {
          duration: duration,
          to: { background: color }
        }).start();
        triggerAnimationStartFrame();
      }

      var element = jqLite('<div class=""></div>');
      $rootElement.append(element);
      jqLite($document[0].body).append($rootElement);

      startAnimation(element, 0.5, 'red');
      expect(element.attr('style')).toContain('transition');

      endTransition(element, 0.5);
      expect(element.attr('style')).not.toContain('transition');

      startAnimation(element, 0.8, 'blue');
      expect(element.attr('style')).toContain('transition');

      // Trigger an extra transitionend event that matches the original transition
      endTransition(element, 0.2);
      expect(element.attr('style')).toContain('transition');

      endTransition(element, 0.8);
      expect(element.attr('style')).not.toContain('transition');
    }));
```

test/ngAnimate/animateCssSpec.js
```diff
@@ -1203,6 +1203,47 @@ describe("ngAnimate $animateCss", function() {
       });
     });
 
+    it('should avoid applying the same cache to an element a follow-up animation is run on the same element',
+      inject(function($animateCss, $rootElement, $document) {
+
+      var element = jqLite('<div></div>');
+      element.attr('class', '');
```
petebacondarwin: No need for this line. You can just do:
```js
var element = jqLite('<div class=""></div>');
```

But what I don't understand is why the test passes without the fix when this class attribute is missing...

test/ngAnimate/animateCssSpec.js
```diff
@@ -1203,6 +1203,47 @@ describe("ngAnimate $animateCss", function() {
...
+      runner = $animateCss(element, {
+        duration: 0.8,
+        to: { background: 'blue' }
+      }).start();
+
+      triggerAnimationStartFrame();
+      expect(element.attr('style')).toContain('transition');
+
+      browserTrigger(element, 'transitionend',
+        { timeStamp: Date.now(), elapsedTime: 0.5 });
```
petebacondarwin: Why does the test pass if `elapsedTime` here is less that 0.5 secs? E.g.

```js
browserTrigger(element, 'transitionend',
       { timeStamp: Date.now(), elapsedTime: 0.2 });
```

src/ngAnimate/animateCss.js
```diff
@@ -665,7 +665,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       // we need to recalculate the delay value since we used a pre-emptive negative
       // delay value and the delay value is required for the final event checking. This
       // property will ensure that this will happen after the RAF phase has passed.
-      if (timings.transitionDuration > 0) {
+      if (options.duration == null && timings.transitionDuration > 0) {
```
petebacondarwin: This line seems to be sufficient to cause the test to pass. Are the other lines (i.e. `maxDelay = Math.max(relativeDelay, 0) || 0;`) unnecessary?

src/ngAnimate/animateCss.js
matsko: sorry that is dead code from a previous fix I had in mind.

test/ngAnimate/animateCssSpec.js
matsko: This was also dead code. It's fine without it.

test/ngAnimate/animateCssSpec.js
matsko: It passes because the transition is still active. Only when `800ms` kicks in will the animation be over. Is this what you mean or did I miss something?

src/ngAnimate/animateCss.js
matsko: The reason why we have `|| 0` is because getComputedStyle is mocked out in a lot of parts of the tests and it returns NaN when the computeTimings code is run. Normally (outside of the testing world) getComputedStyle will always return a value of `0s` for transition-delay or transition-duration or animation-delay or animation-duration so there is no need to check for null values. This is really just here to make the mocking in tests work.

PR #11853, opened 2015-05-11
fix(ngAnimate): class-based animations must set the addClass/removeClass class on the element

### LINKED ISSUE #11810: ngAnimate: the addClass/removeClass class is added during the animation
See http://plnkr.co/edit/iAgdX7hon6aEqMvSko9A?p=preview and click the green box while looking at its styles. The `addClass`/`removeClass`/`addClass-active`/`removeClass-active` class is being added during the animation. Is this intended? Note that HTML is case-insensitive so this looks weird.

Angular 1.4.0-rc.1.

screen shot 2015-05-05 at 18 43 28

cc @matsko

matsko:
With the abstraction system that ngAnimate uses, $animateCss internally
sets the provided `event` as a CSS class on the element. In this
situation the `addClass` and `removeClass` events on the element as a
CSS class. This should not happen with class-based animations and this
feature is unnecessary and has now been removed.

Closes #11810

### DIFF
--- a/test/ngAnimate/animateCssDriverSpec.js
+++ b/test/ngAnimate/animateCssDriverSpec.js
@@ -104,6 +104,14 @@ describe("ngAnimate $$animateCssDriver", function() {
         driver({ element: element });
         expect(capturedAnimation[1].applyClassesEarly).toBeFalsy();
       }));
+
+      it("should not provide a `method` as an option value if the animation is not structural", inject(function() {
+        driver({ element: element, structural: true, event: 'superman' });
+        expect(capturedAnimation[1].event).toBe('superman');
+
+        driver({ element: element, event: 'batman' });
+        expect(capturedAnimation[1].event).toBeFalsy();
+      }));
     });
 
     describe("anchored animations", function() {
@@ -214,7 +222,9 @@ describe("ngAnimate $$animateCssDriver", function() {
             'out': jqLite('<div></div>')
           };
 
+          fromAnimation.structural = true;
           fromAnimation.element.append(anchorAnimation['out']);
+          toAnimation.structural = true;
           toAnimation.element.append(anchorAnimation['in']);
 
           var animator = driver({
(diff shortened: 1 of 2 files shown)


test/ngAnimate/animateCssDriverSpec.js
```diff
@@ -103,6 +103,14 @@ describe("ngAnimate $$animateCssDriver", function() {
         driver({ element: element });
         expect(capturedAnimation[1].applyClassesEarly).toBeFalsy();
       }));
+
+      it("should not provide a method into the $animateCss animation if the animation is not structural", inject(function() {
```
Narretz: "provide ... into" sounds not right somehow.

matsko: it should not provide a `method` as an option value if the animation is not structural

test/ngAnimate/animateCssDriverSpec.js
```diff
@@ -104,6 +104,14 @@ describe("ngAnimate $$animateCssDriver", function() {
         driver({ element: element });
         expect(capturedAnimation[1].applyClassesEarly).toBeFalsy();
       }));
+
+      it("should not provide a `method` as an option value if the animation is not structural", inject(function() {
```
Narretz: I think we could simplify this by changing it "should only provide ... if the animation is structural". What still confuses me, is that "method"  doesn't really reflect what is set. It's really the event (name) that is set, right?

petebacondarwin:
I think the commit message ought to be 

> fix(ngAnimate): class-based animations must **not** set addClass/removeClass CSS classes on the element

test/ngAnimate/animateCssDriverSpec.js
petebacondarwin: I agree with @Narretz here.  Let's change this to:

```js
it("should only set the event name if the animation is structural", inject(function() {
```

Also the previous it clause should probably change:

```js
      it("should signal $animateCss to apply the classes early when animation is structural", inject(function() {
```

petebacondarwin:
Apart from tweaks to test descriptions this LGTM

matsko:
Fixed up and merged as https://github.com/angular/angular.js/commit/3a3db690a16e888aa7371e3b02e2954b9ec2d558

PR #11854, opened 2015-05-11
fix(ngAnimate): throw an error if a callback is passed to animate met…

### LINKED ISSUE #11713: (bug) ngAnimate options.domOperation is not a function

petebacondarwin:
…hods

As of bf0f5502b1bbfddc5cdd2f138efd9188b8c652a9 (released in 1.3.0) it is no longer
valid to pass a callback to the following functions: `enter`, `move`, `leave`, `addClass`,
`removeClass`, `setClass` and `animate`.

To prevent confusing error messages, this change asserts that this parameter is
not a function.

Closes #11826
Closes #11713

### DIFF
--- a/docs/content/error/$animate/nocb.ngdoc
+++ b/docs/content/error/$animate/nocb.ngdoc
@@ -1,0 +1,12 @@
+@ngdoc error
+@name $animate:nocb
+@fullName Do not pass a callback to animate methods
+@description
+
+Since Angular 1.3, the methods of {@link ng.$animate} do not accept a callback as the last parameter.
+Instead, they return a promise to which you can attach `then` handlers to be run when the animation completes.
+
+If you are getting this error then you need to update your code to use the promise-based API.
+
+See https://github.com/angular/angular.js/commit/bf0f5502b1bbfddc5cdd2f138efd9188b8c652a9 for information about
+the change to the animation API and the changes you need to make.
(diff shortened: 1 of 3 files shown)


Narretz:
LGTM
Although for the record I prefer to not do anything here at all. It's a documented BC.

petebacondarwin:
Trouble is that not doing anything leads to a very confusing error, which is hard to track down to having left this callback in place.

One thing missing here is actually the error doc. I will add it.

docs/content/error/$animate/nocb.ngdoc
```diff
@@ -0,0 +1,12 @@
+@ngdoc error
+@name $animate:nocb
+@fullName Do not pass a callback to animate methods
+@description
+
+Since Angular 1.3, the methods of {@link ng.$animate} do not accept a callback as the last parameter.
```
Narretz: Should we mention that the last parameter now takes the option object?

matsko:
@petebacondarwin make sure to retrofit both the assertion code and the docs so that it can be merged into the 1.3 branch as well.

petebacondarwin:
@matsko - by retrofit you mean backport?

matsko:
Yessir

matsko:
Landed as https://github.com/angular/angular.js/commit/98048678dc7d1b5ea549e2e43e6e27870bb60a31 and https://github.com/angular/angular.js/commit/05156143c8552e9d615802c37103d18d9db31a12

PR #11868, opened 2015-05-13
fix($templateRequest): Set `Accept` header on template requests

### LINKED ISSUE #6860: angular directive templateUrl should allow "text/html" Accept and content-type headers
templateUrl is generally going to be expecting to receive HTML back from the server. To inform the server of that (according to RESTful best practices) it would be useful to be able to specify the headers for the $http request that fetches the actual template.

The default being used by angularJS is "application / json". It would be useful to be able to specify "text/html" instead for templateUrl. In addition I submit that should be the default for any request that expects HTML back from the server.

luckycadow:
Set the `Accept` header on HTTP requests for templates to `text/html, text/ng-template`.
Previously the default request headers sent by `$http` were used.

Closes #6860

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11868)

### DIFF
--- a/src/ng/templateRequest.js
+++ b/src/ng/templateRequest.js
@@ -36,7 +36,8 @@ function $TemplateRequestProvider() {
 
       var httpOptions = {
         cache: $templateCache,
-        transformResponse: transformResponse
+        transformResponse: transformResponse,
+        headers: { 'Accept': 'text/html, text/ng-template' }
       };
 
       return $http.get(tpl, httpOptions)


src/ng/templateRequest.js
```diff
@@ -36,7 +36,8 @@ function $TemplateRequestProvider() {
 
       var httpOptions = {
         cache: $templateCache,
-        transformResponse: transformResponse
+        transformResponse: transformResponse,
+        headers: { 'Accept': 'text/html, text/ng-template' }
```
petebacondarwin: I have a feeling it would be safest to move this to the provider so that it could always be modified on an app by app basis.

pkozlowski-opensource: Agreed. The proposal here sounds like a good default value.

BTW, this needs tests.

petebacondarwin:
We have a new PR that overrides this one, as it is configurable and has tests.

PR #11878, opened 2015-05-15
docs(angular.element): describe your change...

wmyers:
Adding mention of jqLite support of jQuery FX animation engine - this is mentioned in the jqLite page (https://code.google.com/p/jqlite/wiki/UsingJQLite), but not mentioned in Angular documentation which is confusing

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -83,6 +83,9 @@
  * - [`val()`](http://api.jquery.com/val/)
  * - [`wrap()`](http://api.jquery.com/wrap/)
  *
+ * ## jQuery/jqLite Animation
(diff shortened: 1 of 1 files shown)


wmyers:
I signed it!

Narretz:
Did you find any connection between angular's jqlite and this library? As far as I know, they only share the name, and as you can see, the googlecode project has last been updated in 2010.

wmyers:
It was my silly mistake. I was looking for a reason why I could use the jQuery `animate` method in Angular when I thought I only had jqLite. I didn't notice (or remember) that angular-fullstack generator had included jQuery (full version) in my project...ho hum.

Narretz:
I don't think we need this, then. :)

PR #11881, opened 2015-05-15
fix(ngAnimate): ensure that "element" caintains html node

interstella-5555:
Ensure that "element" jqlite wrapper contains html node. It results in an error if it is defined but empty.


Narretz:
Thanks for the PR. Can you please:
- show the bug / scenario / issue that is fixed by this change
- write a test that shows the problem

kwypchlo:
Sure @Narretz, I will try to reproduce the problem in an isolated case, though I will need some days for that as it happened in my project at work.

matsko:
+1 :)

Narretz:
Closing because we have no demo of the problem :(

PR #11888, opened 2015-05-17
docs(guide/Controllers): I think that it is better to say that a controller is defined using a JavaScript constructor function as we define a class through a constructor.

niteshthakur:
A controller is a instantiated object created out of the constructor function. So, I feel we should not call a Controller a constructor function.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11888)

### DIFF
--- a/docs/content/guide/controller.ngdoc
+++ b/docs/content/guide/controller.ngdoc
@@ -5,7 +5,7 @@
 
 # Understanding Controllers
 
-In Angular, a Controller is a JavaScript **constructor function** that is used to augment the
+In Angular, a Controller is defined using a JavaScript **constructor function** that is used to augment the
 {@link scope Angular Scope}.
 
 When a Controller is attached to the DOM via the {@link ng.directive:ngController ng-controller}


niteshthakur:
The cla has been signed.


Thanks

petebacondarwin:
I am not convinced that this makes any difference. Constructor functions in JavaScript are not specific to classes. They are simply functions that create and initialise objects, which is exactly what a controller's constructor function is used for.

niteshthakur:
I will try to make my point more clear. The doc says "In Angular, a Controller is a JavaScript constructor function that is used to augment the Angular Scope."

However, I believe that a Controller is an instance or Javascript object that has been created/configured using the constructor and thus is not a JavaScript constructor function itself. Also I believe that the object is created using the 'new' keyword which is referred by 'this' in the constructor function. So strictly speaking, the object is just initialized/configured by the constructor function.

So, in my opinion, it would be better to say that "In Angular, a Controller is defined using a JavaScript constructor function that is used to augment the Angular Scope."

petebacondarwin:
Let's say you have a function that is going to "define" a controller: `function MyController($scope, ...)`.

The it is true that we do not exactly call `new MyController($scope, ...)` to "construct" the controller object. This is because we must apply the injectables to the parameters when creating the object. Inside `$controller` we have the following line of code:

```
instance = $injector.instantiate(expression, locals, constructor);
```

where `constructor` is exactly the function that we are discussing (e.g. `MyController`).

Then inside the `$injector` the instantiate method looks like this:

```js
    function instantiate(Type, locals, serviceName) {
      // Check if Type is annotated and use just the given function at n-1 as parameter
      // e.g. someModule.factory('greeter', ['$window', function(renamed$window) {}]);
      // Object creation: http://jsperf.com/create-constructor/2
      var instance = Object.create((isArray(Type) ? Type[Type.length - 1] : Type).prototype || null);
      var returnedValue = invoke(Type, instance, locals, serviceName);

      return isObject(returnedValue) || isFunction(returnedValue) ? returnedValue : instance;
    }
```

You can see that we are basically simulating `new Constructor(...)` using a combination of `Object.create` and `$injector.invoke`.

Now are you saying that because we are only simulating `new Constructor(...)` that the function that we provide when registering a controller is not really a constructor? Or something else?

In any case, I am still of the opinion, unless you can convince me otherwise, that there is no semantic or practical value in distinguishing in the way that you suggest.

niteshthakur:
I will try to be even more elaborate. The doc says "In Angular, a Controller is a JavaScript constructor ".  After reading the above line in quotes, a question comes to my mind. Is the controller really a constructor and the answer to this is 'NO'(at least for me). To me, controller is something which controls some part of our logic and hence it is not a constructor function according to me but a outcome(an instance) of running of constructor function. Thus I still think it is better to say that  "a controller is defined using a java script constructor function". Hope this will convey my thought better.

petebacondarwin:
Oh right! I get it. OK. Sounds reasonable.

Issue #11797, opened 2015-05-04
Allow global configuration of the default value for ngMessages' multiple


realityking:
In a large application with hundereds of form fields it's rather tedious to always remember to add `ng-messages-multiple` to get a consistent expierence for users.

It would be great if a default could be set globally. This might entail changing `ng-messages-multiple` to accept an expression (or a string) to allow disabling it when the default is to show multiple messages.

Narretz:
Sounds reasonable.

kentcdodds:
I think this feature would definitely be handy.

I want to mention angular-formly. Disclosure, it's my library, so feel free to take this with a grain of salt. But making this happen with angular-formly would be trivial using the `wrappers` feature. I threw together a quick [example](http://angular-formly.com/#/example/other/multiple-errors). Hopefully this is more helpful than spammy.

lili21:
ngMessagesProvider???

PR #11891, opened 2015-05-18
feat(ngMessagesConfig): global configuration of the default value for ngMessages' multiple

lili21:
#11797


lili21:
I signed it!

Narretz:
Hi, thanks for the PR. There a few things however that need to be changed before this can be added:
- [ ] app-wide configs should use providers instead of constants
- [ ] the behavior needs tests (e.g. if multiple is set when the config is set)

gkalpak:

PR #11894, opened 2015-05-18
fix($animate): accept unwrapped DOM elements as inputs for enter + move

### LINKED ISSUE #11848: $animate.enter with third parameter in 1.4.x
I have been doing `$animate.enter()` for a while in Angular `1.3.x` in the following way:

```javascript
$animate.enter(element, parent, parent[0].lastChild);
```

So instead of adding a new `element` at the beginning of `parent`, I add them after the last item on `parent`.

That worked good, but after I updated to Angular `1.4.x`, it says that what I pass as the third parameter `parent[0].lastChild` doesn't have the `after` method. That suggested to me that I needed to wrap that element into `angular.element`:

```javascript
$animate.enter(element, parent, angular.element(parent[0].lastChild));
```

And that did work. The thing is, I tried to replicate this issue on a plunker and it didn't work either in 1.3.x if I don't wrap the third parameter.

So my question is: Having to wrap the third parameter with `angular.element` was always needed? Or maybe with the whole refactor it needs it now?

It surely fixed it to work with `1.3.x` as well but it feels weird now.

There is a [plunker](http://plnkr.co/edit/Mkyq3ViX74rRpGpw5j7P?p=info)

Maybe I am doing it wrong, but it worked for me in the past when it was just `after.after`.

matsko:
Prior to this fix the $animate.enter() and $animate.move() events caused
an error when a parent or after element was provided that was not
already wrapped as a jqLite element. This patch ensures that both
wrapped and unwrapped DOM nodes are allowed.

Closes #11848

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -424,6 +424,8 @@ var $AnimateProvider = ['$provide', function($provide) {
        */
       enter: function(element, parent, after, options) {
         assertNoCallback(options);
+        parent = parent && jqLite(parent);
+        after = after && jqLite(after);
         parent = parent || after.parent();
         domInsert(element, parent, after);
         return $$animateQueue.push(element, 'enter', options);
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@matsko does this need to go in before we can release 1.4.0? Or can it go into 1.4.1?

petebacondarwin:
Scrap that - it looks good to me. @matsko can you land this one please?

matsko:
Landed as https://github.com/angular/angular.js/commit/f26fc26f6ea283b2fc5ddb18627b13850de2663e

PR #11895, opened 2015-05-18
fix(ngAnimate): ensure that repeated structural calls during pre-digest function

### LINKED ISSUE #11867: ng-view duplication in the 1.4.RC
Hi there,

I've upgraded my application to the 1.4-RC2 and found that after changing of the route e.g. from 'route1' to 'route2' there is several dom elements with `ng-view` attribute.
Issue is reproducible in my application with versions RC0, RC1 and RC2 and only on the route that will be handled in `otherwise`. E.g. '/Book'

Image of bug

~~I haven't able to reproduce this issue on jsbin, but it exists in my application.~~
http://jsbin.com/buzuwe/2/edit?html,js,output

Simplified version of my config looks like this:

```js
app.config(function($routeProvider, $compileProvider) {
  $routeProvider
   .when('/Book/:bookId', {
    templateUrl: 'book.html',
    controller: 'BookController',
    resolve: {
      delay: function($q, $timeout) {
        var delay = $q.defer();
        $timeout(delay.resolve, 1000);
        return delay.promise;
      }
    }
  })
  .when('/Book/:bookId/ch/:chapterId', {
    templateUrl: 'chapter.html',
    controller: 'ChapterController'
  })
  .otherwise({
    redirectTo: '/',
    templateUrl: 'noRoute.html'
  });

$compileProvider.aHrefSanitizationWhitelist(...);
```

Could anyone help to find, what's the problem here? Is it a bug or just wrong structure of my application?

Thanks in advance

matsko:
Prior to this fix if `$animate.enter()` or `$animate.leave()` was called
before a digest was issued then the element may not be cancelled early
enough. This fix ensures that the previous structural animation is
cancelled immediately when a follow-up animation is kicked off.

Closes #11867

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -1003,6 +1003,20 @@ describe("animations", function() {
         expect(enterComplete).toBe(true);
       }));
 
+      it('should cancel the previous structural animation if a follow-up structural animation takes over before the postDigest',
+        inject(function($animate, $$rAF) {
+
+        var enterComplete = false;
+        $animate.enter(element, parent).done(function() {
+          enterComplete = true;
+        });
+
+        expect(enterComplete).toBe(false);
+        $animate.leave(element);
+        $$rAF.flush();
+        expect(enterComplete).toBe(true);
+      }));
+
       it('should skip the class-based animation entirely if there is an active structural animation',
         inject(function($animate, $rootScope) {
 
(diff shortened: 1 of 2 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -1003,6 +1003,20 @@ describe("animations", function() {
...
+        $animate.enter(element, parent).done(function() {
+          enterComplete = true;
+        });
+
+        expect(enterComplete).toBe(false);
+        $animate.leave(element);
+        $$rAF.flush();
+        expect(enterComplete).toBe(true);
+      }));
+
```
petebacondarwin: Use spies:

```js
      it('should cancel the previous structural animation if a follow-up structural animation takes over before the postDigest',
        inject(function($animate, $$rAF) {

        var enterDone = jasmine.createSpy('enter animation done');
        $animate.enter(element, parent).done(enterDone);
        expect(enterDone).not.toHaveBeenCalled();

        $animate.leave(element);
        $$rAF.flush();
        expect(enterDone).toHaveBeenCalled();
      }));
```

petebacondarwin:
Except for not using spies, LGTM

matsko:
Landed as https://github.com/angular/angular.js/commit/2327f5a0a7e018a9b03aefabe1fbd0c9330e2eeb.

PR #11899, opened 2015-05-18
fix(ngAnimate): ensure that cancelled class-based animations are properly cleaned up

### LINKED ISSUE #11652: Problem with <input type="email" /> and form validation
After migration to 1.4 I discovered that form element does is not changing and does not get ng-dirty class is one of the form elements is input with type="email". Also with this input and forms are marked with attribute data-ng-animate="1"

matsko:
Prior to this fix if the same CSS class was added and removed quickly
then the element being animated would be left with a stale cache of the
cancelled out animation. This would then result in follow-up animations
being added to the previous animation which would then never run. A
stale cache was to blame for that. This patch takes care of this issue.

Closes #11652

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -394,6 +394,13 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
             runner.end();
           }
 
+          // in the event that the element animation was not cancelled or a follow-up animation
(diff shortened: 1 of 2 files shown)


gkalpak:
LGTM :wink:

Narretz:
LGTM2!

matsko:
Merged as https://github.com/angular/angular.js/commit/718ff84405558ac64402e1fca5caefd7d307ea1e

PR #11905, opened 2015-05-19
refactor(ngCsp): Make use of document.head

realityking:
document.head is available since IE9 (https://developer.mozilla.org/en-US/docs/Web/API/Document/head)

Saves us a bit of DOM querying during startup and a couple of bytes.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11905)


petebacondarwin:
LGTM - although you might like to chime in on #11933 as this could be a moot point.

PR #11908, opened 2015-05-20
docs(guide/Directives): correct the sentence describing compiling a HTML

cockroachzl:
> The reason we use the term "compile" is that the recursive process of attaching <directives>
> mirrors the process of compiling source code in [compiled programming languages]

substitute "directives" with "event listeners" in the above sentence.


cockroachzl:
I signed it!

Narretz:
"directives" is actually more correct. Could you change the reference to "event listeners" to "directives"?

PR #11910, opened 2015-05-20
docs($resourceProvider): provide info and example for configuring $re…

yilinglu:
…sourceProvider

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11910)

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -57,6 +57,24 @@ function shallowClearAndCopy(src, dst) {
  * <div doc-module-components="ngResource"></div>
  *
  * See {@link ngResource.$resource `$resource`} for usage.
+ *
+ * See {@link ngResource.$resourceProvider `$resourceProvider`} for usage.
+ */
+
+/**
+ * @ngdoc provider
+ * @name $resourceProvider
+ *
+ * @description
+ *
+ * Use for configuring {@link ngResource.$resource `$resource`} in module configuration phase.
+ *
+ * ## Example
+ * See {@link ngResource.$resourceProvider#defaults `Properties`} for configuring `ngResource`.
+ *
+ * ## Dependencies
+ * Requires {@link ngResource `ngResource`} module to be installed.
+ *
  */
 
 /**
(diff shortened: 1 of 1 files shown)


src/ngResource/resource.js
```diff
@@ -57,6 +57,24 @@ function shallowClearAndCopy(src, dst) {
  * <div doc-module-components="ngResource"></div>
  *
  * See {@link ngResource.$resource `$resource`} for usage.
+ *
+ * See {@link ngResource.$resourceProvider `$resourceProvider`} for usage.
```
gkalpak: Merge these two lines into:

```
See {@link ngResource.$resource `$resource`} and {@link ngResource.$resourceProvider `$resourceProvider`} for usage.
```

src/ngResource/resource.js
```diff
@@ -57,6 +57,24 @@ function shallowClearAndCopy(src, dst) {
...
+ */
+
+/**
+ * @ngdoc provider
+ * @name $resourceProvider
+ *
+ * @description
+ *
+ * Use for configuring {@link ngResource.$resource `$resource`} in module configuration phase.
```
gkalpak: I would rather change it to something like:

_Use `$resourceProvider` to change the default behavior of the {@link ngResource.$resource `$resource`} service._

--
<sub>(You don't have to mention the configuration phase. It's not specific to this provider and we don't want to mention such stuff in every provider's documentation.)</sub>

src/ngResource/resource.js
```diff
@@ -57,6 +57,24 @@ function shallowClearAndCopy(src, dst) {
...
+ *
+ * ## Example
+ * See {@link ngResource.$resourceProvider#defaults `Properties`} for configuring `ngResource`.
```
gkalpak: This _Example_ section is redundant imo.

src/ngResource/resource.js
```diff
@@ -363,6 +427,22 @@ angular.module('ngResource', ['ng']).
...
+     * This shows the fact that {@link ngResource.$resource $resource}
+     * is a high level service built by leveraging {@link ng.$http $http} service.
+     *
+     * @param {Object} $http An instance of {@link ng.$http $http} service.
+     *
+     * @param {Object} $q An instance {@link ng.$q $q}, Angular's implementation of deferred API.
+     *
+     *
+     * @returns {Function} A factory function, when invoked, returns a $resource object.
+     */
```
gkalpak: `$get` methods should not be documented. They are idiomatic and kind of private to Angular.
As you mention, they should only be called by the framework itself.

src/ngResource/resource.js
```diff
@@ -349,6 +367,52 @@ angular.module('ngResource', ['ng']).
...
+     *        delete:   {method: 'DELETE'}
+     *    };
+     *
+     *    // Don't strip trailing slashes from calculated URLs.
+     *    // Consult your application server's configuration to work in concert with this setting.
+     *    $resourceProvider.defaults.stripTrailingSlashes = false;
+     * }]);
+     * ```
+     *
+     */
```
gkalpak: Could you bring this more inline with how [`$cookiesProvider.defaults`][1] are formatted ?

[1]: https://github.com/angular/angular.js/blob/v1.4.0-rc.2/src/ngCookies/cookies.js#L28-L49

gkalpak:
@yilinglu: Thx for the PR. I left a couple of comments.
(Feel free to ping me again once you have gone through them.)

petebacondarwin:
@gkalpak - we should put things like this into a milestone once you have triaged it. In this case, since we are awaiting response it can go into Purgatory.

PR #11915, opened 2015-05-20
docs(guide/Filters): describe your change...

swick7:
I had a really hard time understanding that if I use a custom filter from a controller or service, I needed to add the string "Filter" to the name of the filter when listing it as a dependency of my controller or service.  I hope, by adding this example of the custom filter, "name", it becomes apparent.  

It's weird that a developer could refer to the custom filter by it's name in a view, like `<div ng-repeat="user in array | name:'a'">{{user.name}}</div>` but they have to refer to it by "nameFilter" from a controller or service instead of just "name".

In this example, the error message from angular is really unhelpful when the controller fails to refer to the "name" filter by the string "nameFilter".  The error is, "Error: [$injector:unpr] Unknown provider: nameProvider <- name <- FilterController".

### DIFF
--- a/docs/content/guide/filter.ngdoc
+++ b/docs/content/guide/filter.ngdoc
@@ -64,7 +64,18 @@ or the filter expression is changed).
 
   <file name="script.js">
     angular.module('FilterInControllerModule', []).
-      controller('FilterController', ['filterFilter', function(filterFilter) {
+      filter('name', function () {
+        return function (names, value) {
+          var matches = [];
+          for(var i=0; i<names.length; i++) {
+            if (names[i].match(value)) {
+              matches.push(names[i]);
+            }
+          }
+          return matches;
+        };
+      }).
+      controller('FilterController', ['nameFilter', function(nameFilter) {
         this.array = [
           {name: 'Tobias'},
           {name: 'Jeff'},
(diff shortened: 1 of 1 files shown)


swick7:
I signed it!

gkalpak:
FWIW, this explained (clearly imo) in the [Filters][1] section of the Developer Guide.

[1]: https://docs.angularjs.org/guide/filter#using-filters-in-controllers-services-and-directives

Narretz:
@gkalpak But we are in the filters guide

gkalpak:
@Narretz, indeed :) I guess I meant in the **"Using filters in controllers, services, and directives"** section of the Filter guide, where it says:

> You can also use filters in controllers, services, and directives. For this, inject a dependency with the name `<filterName>Filter` to your controller/service/directive. E.g. using the dependency `numberFilter` will inject the `number` filter.

bradhenslee:
After suddenly struggling with this same issue (despite having been using custom filters within templates for a long time), I'd like to urge you to merge this PR. In my opinion, the 'nameFilter' makes a clearer example of what's happening than the 'filterFilter' example.

Narretz:
I've made a change for this, but  did not include this new example. It doesn't really add anything new and might be confusing actually. It's explained in the first paragraph how to require filters - I've made the style more explicit. I've also added an example of using the filter in the controller in the custom filter example.

PR #11919, opened 2015-05-21
docs(input[datetime-local]): change type of ngModel in docs

mikaelharsjo:
In the docs for `input[datetime-local]` it says in the text that model should always be a `Date` object but under Usage and Arguments it says `string`. This fixes so both says `Date`.

Docs here: https://docs.angularjs.org/api/ng/input/input%5Bdatetime-local%5D


mikaelharsjo:
I did us a different email. I've added that to my google account now and it displays in the CLA.

mikaelharsjo:
As I said. I have added the git `user.email` to my account. Does it have to be my primary email?

mikaelharsjo:
Closing and reopening to see if it makes a difference.

PR #11920, opened 2015-05-21
docs(input[datetime-local]): change type of ngModel in docs

mikaelharsjo:
In the docs for `input[datetime-local]` it says in the text that model should always be a `Date` object but under Usage and Arguments it says `string`. This fixes so both says `Date`.

Docs here: https://docs.angularjs.org/api/ng/input/input%5Bdatetime-local%5D

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -226,7 +226,7 @@ var inputType = {
     * The timezone to be used to read/write the `Date` instance in the model can be defined using
     * {@link ng.directive:ngModelOptions ngModelOptions}. By default, this is the timezone of the browser.
     *
-    * @param {string} ngModel Assignable angular expression to data-bind to.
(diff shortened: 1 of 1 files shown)


mikaelharsjo:
I did author them. My email is listed in the CLA so I'm not sure what's wrong.

realityking:
I'm not with Google, but they need to have a CLA on file for the email address with which the commit was made. In this case that's mikael.harsjo@cryptzone.com

mikaelharsjo:
I have signed the CLA. And the e-mail address is listed there. weird.

Narretz:
Hm, since we are talking about element attributes, everything is a string. Sure, on the scope it must be a model, but on the element it can't be anything else but a string.

mikaelharsjo:
That's a valid point. I'm just saying that the way the docs are written now it fooled me to think my model should be a `string`.

PR #11923, opened 2015-05-21
docs(ngModel): ng-model scope inheritance update

tylermcginnis:
ng-model will look to its current scope, then look to all its parent scopes before it implicitly creates a property on the current scope. The current docs don't talk about the parent lookup. http://jsfiddle.net/tylermcginnis/ma9a20kq/


tylermcginnis:
I signed it!

PR #11927, opened 2015-05-22
fix(ngAnimate): ensure invalid merged animations are cleared from the element

petebacondarwin:
@matsko can you just check that this fix makes sense?


matsko:
Landed as https://github.com/angular/angular.js/commit/db246eb701529b41049fc118908e528920f13b24

PR #11928, opened 2015-05-22
docs(ngPattern): add note about using the `g` flag

### LINKED ISSUE #11917: Issue with ng-pattern
I have an issue with ng-pattern. It works perfectly when adding characters to the input, it vaildates and shows error when invalid but acts weirdly while deleting characters from the input even if its is a valid string. If you delete a character from the input it show invalid and if you delete one more character it shows it as a valid. this repeats for every character you delete.

I observed this issue in 1.2.18, 1.3.15 and 1.4.0.rc2
I have added an example in the plunker. 

http://plnkr.co/edit/A4kHHthnEYzaxyVzu7w9?p=preview

gkalpak:
Add a note to point out that using the `g` flag on the validation RegExp, will cause each search to start at the index of the last search's match, thus not taking the whole input value into account.

Closes #11917
Closes #11928

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -46,9 +46,13 @@ var inputType = {
    *    as in the ngPattern directive.
    * @param {string=} ngPattern Sets `pattern` validation error key if the ngModel value does not match
    *    a RegExp found by evaluating the Angular expression given in the attribute value.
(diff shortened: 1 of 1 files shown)


petebacondarwin:
LGTM - please merge!

PR #11933, opened 2015-05-22
Refine csp checks for `unsafeEval`.

jdalton:
This PR refines the csp checks for function compilation because the default policy in WWA (Windows Web Apps) allows `unsafe-eval`. This would give a [~30% boost](https://docs.angularjs.org/api/ng/directive/ngCsp) when evaluating expressions.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/11933)

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -978,22 +978,27 @@ function equals(o1, o2) {
   return false;
 }
 
+var unsafeEval = function() { /* jshint ignore:line */
+  if (isDefined(unsafeEval.isActive_)) return unsafeEval.isActive_;
+
+  var active = true;
+  try {
+    /* jshint -W031, -W054 */
+    new Function('');
+    /* jshint +W031, +W054 */
+  } catch (e) {
+    active = false;
+  }
+  return (unsafeEval.isActive_ = active);
+};
+
 var csp = function() {
   if (isDefined(csp.isActive_)) return csp.isActive_;
 
   var active = !!(document.querySelector('[ng-csp]') ||
                   document.querySelector('[data-ng-csp]'));
 
-  if (!active) {
-    try {
-      /* jshint -W031, -W054 */
-      new Function('');
-      /* jshint +W031, +W054 */
-    } catch (e) {
-      active = true;
-    }
-  }
-
+  active = active || !unsafeEval();
   return (csp.isActive_ = active);
 };
 
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1698,12 +1698,14 @@ function $ParseProvider() {
   var cacheExpensive = createMap();
 
   this.$get = ['$filter', '$sniffer', function($filter, $sniffer) {
+    var csp = $sniffer.csp && !$sniffer.unsafeEval;
+
     var $parseOptions = {
-          csp: $sniffer.csp,
+          csp: csp,
           expensiveChecks: false
         },
         $parseOptionsExpensive = {
-          csp: $sniffer.csp,
+          csp: csp,
           expensiveChecks: true
         };
 
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -65,6 +65,7 @@
     "shallowCopy": false,
     "equals": false,
     "csp": false,
+    "unsafeEval": false,
     "concat": false,
     "sliceArgs": false,
     "bind": false,
--- a/src/ng/sniffer.js
+++ b/src/ng/sniffer.js
@@ -79,6 +79,7 @@ function $SnifferProvider() {
         return eventSupport[event];
       },
       csp: csp(),
+      unsafeEval: unsafeEval(),
       vendorPrefix: vendorPrefix,
       transitions: transitions,
       animations: animations,
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -61,6 +61,7 @@
     "shallowCopy": false,
     "equals": false,
     "csp": false,
+    "unsafeEval": false,
     "jq": false,
     "concat": false,
     "sliceArgs": false,
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -711,6 +711,7 @@ describe('angular', function() {
     afterEach(function() {
       window.Function = originalFunction;
       delete csp.isActive_;
+      delete unsafeEval.isActive_;
     });
 
 


src/Angular.js
```diff
@@ -978,6 +978,20 @@ function equals(o1, o2) {
...
+  var active = true;
+  try {
+    /* jshint -W031, -W054 */
+    new Function('');
+    /* jshint +W031, +W054 */
+  } catch (e) {
+    active = false;
+  }
+  return (unsafeEval.isActive_ = active);
+};
```
jdalton: Having `unsafeEval` split out of `csp` may make it chatty in the console. I can add guards to it if you all would like.

jdalton:
Corporate CLA: Microsoft.

src/Angular.js
```diff
@@ -985,15 +999,8 @@ var csp = function() {
...
 
   if (!active) {
-    try {
-      /* jshint -W031, -W054 */
-      new Function('');
-      /* jshint +W031, +W054 */
-    } catch (e) {
-      active = true;
-    }
+    active = !unsafeEval();
```
petebacondarwin: this whole bit could be changed to

```js
active = active || !unsafeEval();
```

src/ng/parse.js
```diff
@@ -1698,12 +1698,14 @@ function $ParseProvider() {
   var cacheExpensive = createMap();
 
   this.$get = ['$filter', '$sniffer', function($filter, $sniffer) {
+    var csp = $sniffer.csp && !$sniffer.unsafeEval;
```
petebacondarwin: Does `$sniffer.csp` not include `!sniffer.unsafeEval`?

caitp: the point is that CSP headers could turn off things like unsafe-inline, but leave unsafe-eval enabled

caitp: Oh I see what you're saying, we don't actually have a check for unsafe-inline anymore, I guess that disappeared. So yeah, `csp` is just an alias for this at the moment.

src/Angular.js
jdalton: Nice!

src/ng/parse.js
jdalton: Depending `csp` doesn't necessarily hit the `!$niffer.unsafeEval` check so this ensures that your `$sniffer.csp` is `true` and that `unsafeEval` is not allowed.

jdalton:
Any other concerns or issues?

IgorMinar:
@jdalton can you please explain how this improves anything on Windows?

It looks like you are trying to force non-csp mode even when developer opts in to it, which changes the semantics of the opt-in.

What's the real world scenario you are trying to fix?

jdalton:
@IgorMinar Currently the csp check in Angular assumes that csp means unsafe eval is not allowed, however in Windows 10 WWA the [default csp policy allows unsafe eval](http://blogs.windows.com/buildingapps/2015/06/08/winjs-4-0-released/).

snippet

The benefit is to give a perf boost out of the box for those using Angular in WWAs.

IgorMinar:
The only side effect of opting into the CSP mode is that we don't use function constructor. So just by not opt-ing into the CSP mode you get the perf benefit you are looking for.

As far as I can tell the only impact of your change is that if someone explicitly opts into the CSP mode via ng-csp directive, we will ignore it and still use function constructor in `$parse`.

IgorMinar:
There is one more place where csp mode plays any role and that's when including the default stylesheet. I assume that on WWA this would throw, in which case you'd want to prevent that via ng-csp, which then forces you down the slow path in $parse. Is that the case?

jdalton:
>  if someone explicitly opts into the CSP mode via ng-csp directive, we will ignore it and still use function constructor in `$parse`.

It's not ignoring CSP, it's using `unsafeEval` if it's allowed by the users CSP policy which is a better check than blanket CSP "on" or "off".

> There is one more place where csp mode plays any role and that's when including the default stylesheet. I assume that on WWA this would throw, 

How's it loading the default stylesheet? Inline scripts are still not allowed by the Win 10 default policy.

IgorMinar:
We append an inline stylesheet to the head while booting, unless ng-csp is found in the doc.

jdalton:
unsafe-inline is allowed by the Win 10 default CSP policy too.

IgorMinar:
In that case then just don't use `ng-csp` in your main template and you are all set. no?

petebacondarwin:
I understand now where you are coming from.

Angular does two things that are naughty from a CSP perspective:

* Injecting inline styles into the DOM (see the bottom line of angular.js)
* Evaluating strings as code (in the parser)

This PR changes things so that Angular will now always test to see if we are allowed to eval strings, even if CSP has been set.

I would like to suggest a more explicit API, which is to allow `ngCsp` to be able to configure the rules that we care about:

<html ng-csp="

petebacondarwin:
Sorry, slipped on the wrong button there

IgorMinar:
if for whatever reason we need $sniffer.csp to distinguish between unsafe-inline and unsafe-eval, it would be preferable to remove the need to inline the stylesheet (and instead use https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule to add the default styles to the document).

afterwards ng-csp would only control if we attempt to autodetect unsafe-eval or not.

the primary reason for ng-csp was to avoid autodetection which relies on throwing exceptions which is not nice and interferes with debugging.

but from what you described, you actually don't need to change angular in order to get your performance benefit.

jdalton:
> but from what you described, you actually don't need to change angular in order to get your performance benefit.

Outside of default CSP policies users can provide their own CSP policies in apps or sites that while they may restrict inline stylesheets can allow `unsafeEval`. This would allow devs who still indicate CSP to be able to get a boost if their policies allow `unsafeEval`.

petebacondarwin:
I think we can fix this in a cleaner way.

Either we remove the inline stylesheet from Angular (using `insertRule`) so that `ng-csp` only refers to `unsafeEval` or make `ng-csp` more configurable (by providing a value to the directive).

```html
<html ng-csp-inline-style ng-csp-unsafe-eval>
```

or

```html
<html ng-csp="inline-style;unsafe-eval">
```

The first option is preferable to us as it means that we don't have to worry about people tweaking their CSP config in subtle ways that we don't support.

petebacondarwin:
Closing for now. @jdalton if you would like to progress this along the lines of one of the above solutions, then please open a new PR. I would suggest trying to fix it by removing our dependency on inline styles first.

jdalton:
@petebacondarwin I can dig it. Could you point me in the right direction?

jdalton:
nm found it :mag_right:

jdalton:
It looks like `insertRule` won't work with `style-src 'self'`. Is there something else you had in mind?

jdalton:
Ping @petebacondarwin. If the style issue can't be resolved then I think this PR still has merit.

jdalton:
Ping again. I'd like to continue this discussion since there doesn't seem like there's a way to avoid the css csp restriction.

petebacondarwin:
@jdalton sorry about that - got a bit distracted by AngularU. Can we continue this next week? I am flying back to the UK tomorrow.

jdalton:
> Can we continue this next week?

Sure thing :octocat:

petebacondarwin:
OK, so the problem with `style-src 'self'` is that we can add a style element dynamically to a page, right.

So for example the following fails:

```html
<!DOCTYPE html>
<html>
<head>
  <title>Content Security Policy Test</title>
  <meta http-equiv="Content-Security-Policy" content="style-src 'self'">
</head>
<body>
      var styleEl = document.createElement('style'),
      document.head.appendChild(styleEl);
  </script>
</body>
</html>
```

petebacondarwin:
I wonder if we can use a SHA of the stylesheet to get around this somehow?

```html
<!DOCTYPE html>
<html>
<head>
  <title>Content Security Policy Test</title>
  <meta http-equiv="Content-Security-Policy" content="style-src 'self' 'sha256-SYgyy9Wm0ECiqsk_A24OSP3RCfkJ0lt3HGKlCXLCjqc='">
  <style>
  @charset "UTF-8";

  [ng\:cloak], [ng-cloak], [data-ng-cloak], [x-ng-cloak],
  .ng-cloak, .x-ng-cloak,
  .ng-hide:not(.ng-hide-animate) {
    display: none !important;
  }

  ng\:form {
    display: block;
  }

  .ng-animate-shim {
    visibility:hidden;
  }

  .ng-anchor {
    position:absolute;
  }
  </style>
</head>
<body>
</body>
</html>
```

petebacondarwin:
So this works:

```html
<!DOCTYPE html>
<html>
<head>
  <title>Content Security Policy Test</title>
  <meta http-equiv="Content-Security-Policy" content="style-src 'self' 'sha256-1PxuDsPyGK6n-LZsMv0gG4lMX3i3XigG6h0CzPIjwrE='">
  <script src="angular.js"></script>
</head>
<body>
</body>
</html>
```

Here the SHA of the style block that is loaded by angular.js is provided as a meta tag.

petebacondarwin:
Right! Now we are rocking. We can replace the bit that gets added to the end of angular.js, i.e. this bit:

```js
!window.angular.$$csp() && window.angular.element(document).find('head').prepend('<style type="text/css">@charset "UTF-8";[ng\\:cloak],[ng-cloak],[data-ng-cloak],[x-ng-cloak],.ng-cloak,.x-ng-cloak,.ng-hide:not(.ng-hide-animate){display:none !important;}ng\\:form{display:block;}.ng-animate-shim{visibility:hidden;}.ng-anchor{position:absolute;}</style>');
```

with this bit of code instead:

```js
var head = window.angular.element(document).find('head')
head.prepend('<meta http-equiv="Content-Security-Policy" content="style-src \'self\' \'sha256-1PxuDsPyGK6n-LZsMv0gG4lMX3i3XigG6h0CzPIjwrE=\'">')
head.prepend('<style type="text/css">@charset "UTF-8";[ng\\:cloak],[ng-cloak],[data-ng-cloak],[x-ng-cloak],.ng-cloak,.x-ng-cloak,.ng-hide:not(.ng-hide-animate){display:none !important;}ng\\:form{display:block;}.ng-animate-shim{visibility:hidden;}.ng-anchor{position:absolute;}</style>');
```

I haven't tested it with server provided CSP headers. I am not sure if the `meta` tag would be ignored if the server provides a `style-src` rule.

petebacondarwin:
So whether or not this would work, I think we could state that if you want to have `style-src` rules in place then you must include our SHA as one of the options.

This means that we could dispense with checking for csp when adding the styles to the document. The overhead for the application developer is not much more than having to provide an external stylesheet when using such a CSP style rule.

petebacondarwin:
@jdalton and @IgorMinar what do you think?

jdalton:
I'll test this in the morning. So your proposing it attempt to write the styles twice (one via the sha and one via style element)?

Might be good to add a code comment with how you sha encoded the css.

petebacondarwin:
No, not twice. The SHA is just there to tell CSP that the inline styles are whitelisted.

petebacondarwin:
I SHA encoded the string by simply causing the error in Chrome - it dumps out a valid SHA that you can then use. :-)

petebacondarwin:
Just to be clear - my suggestion is that we simply include the inline style block in the angular.js codebase, without any CSP check, and publish the SHA of the block in our documentation.

If someone wants to turn on `style-src 'self'``, for instance, then they would have to add our SHA to the CSP rules that are provided to the browser.

jdalton:
Nooice!

lgalfaso:
> Just to be clear - my suggestion is that we simply include the inline
> style block in the angular.js codebase, without any CSP check, and publish
> the SHA of the block in our documentation.
>
Would this mean that any change to the css is a breaking change?
The fact that most developers do not care about the css that we add, does
not make things any simpler

>

petebacondarwin:
@lgalfaso - If we made a change to the inline CSS styles then when you upgrade to that version you will need to change the SHA that is provided in your CSP directives. This means that if you use bower or npm to install your angular files and it chose to automatically update angular from say 1.5.0 to 1.5.1 then your CSP controlled application would immediately fail.

But the failure would be immediate and obvious, and updating to the new SHA should not be particularly difficult. In fact if we provided the SHA in a file along with the distribution of angular then one could have arrange for your web server to use the SHA directly from that file when serving a CSP controlled application. In that case there would be nothing to do when upgrading.

lgalfaso:
@petebacondarwin ok. One more thing, why would we need to add the `meta` tag as in your proposal?
```js
var head = window.angular.element(document).find('head')
head.prepend('<meta http-equiv="Content-Security-Policy" content="style-src \'self\' \'sha256-1PxuDsPyGK6n-LZsMv0gG4lMX3i3XigG6h0CzPIjwrE=\'">')
head.prepend('<style type="text/css">@charset "UTF-8";[ng\\:cloak],[ng-cloak],[data-ng-cloak],[x-ng-cloak],.ng-cloak,.x-ng-cloak,.ng-hide:not(.ng-hide-animate){display:none !important;}ng\\:form{display:block;}.ng-animate-shim{visibility:hidden;}.ng-anchor{position:absolute;}</style>');
```

This is, what would not work with just
```js
var head = window.angular.element(document).find('head')
head.prepend('<style type="text/css">@charset "UTF-8";[ng\\:cloak],[ng-cloak],[data-ng-cloak],[x-ng-cloak],.ng-cloak,.x-ng-cloak,.ng-hide:not(.ng-hide-animate){display:none !important;}ng\\:form{display:block;}.ng-animate-shim{visibility:hidden;}.ng-anchor{position:absolute;}</style>');
```

petebacondarwin:
@lgalfaso  absolutely. My final proposal did not include the meta tags

lgalfaso:
@petebacondarwin then your solution SGTM

realityking:
I don't think relying on the SHA1 is a good idea.

All modern browser [support CSP](http://caniuse.com/#feat=contentsecuritypolicy) (ignoring IE 10-11 which only support sandbox policy which isn't of interest here)

While the hash usage appears to be supported by [Chrome 40](https://www.chromestatus.com/feature/4957003285790720) and [Firefox 29](https://bugzilla.mozilla.org/show_bug.cgi?id=883975), Microsoft only [claims](https://msdn.microsoft.com/en-us/library/dn904195(v=vs.85).aspx) to support CSP 1.0 (which does not include hash support) and god knows what Safari currently supports.

jdalton:
> Microsoft only claims to support CSP 1.0

Microsoft's default CSP restrictions for MS Edge allow inline styles sheets.

realityking:
I'm not talking about default CSP restrictions.

A while ago (#8459) @petebacondarwin figured out a, albeit hacky, way how to run Angular while disabling inline styles but still having the speed advantage of eval. I'd be rather unfortunate to loose that.

jdalton:
So for this PR is there anything that would need to tweak or change to make room for more style sheet discussions?

realityking:
To be honest, I haven't really understood the aim of this PR.

From what I read here, WWAs have a default CSP policy which includes both `unsafe-eval` for javascript and `unsafe-inline` for Stylesheets. Since these are the only two directives which affect AngularJS, I'm missing what needs to change here. If people manually set `ng-csp` they're gonna sufffer some performance degradation, but do we really need to hold hands?

Slightly off topic: The best solution would be if browsers exposed the current security policy. This was once [specified](https://w3c.github.io/webappsec/specs/content-security-policy/#strawman-dom-api) but for some reason it's absent from newer drafts.

jdalton:
@realityking See https://github.com/angular/angular.js/pull/11933#issuecomment-110505004.

> If people manually set ng-csp they're gonna sufffer some performance degradation, but do we really need to hold hands?

Ya. Having CSP restrictions is fine or getting rid or the directives, if doable, is great. The issue is the assumption that CSP is an auto-out for compilation. In this case the assumption is incorrect and needlessly costing 30%. The PR attempts to break down the checks to be a bit more granular. I think the change is a positive one. The discussion on styles would be a good follow up PR.

petebacondarwin:
So the deal is that Angular does two things that can be locked down by CSP:

* unsafe eval
* unsafe inline

Currently if you don't specify the `ng-csp` directive then Angular just tests for whether eval-ing a string throws an error. This causes an unwanted error message in the console.
To prevent this the developer can specify the `ng-csp` directive which tells Angular not to do either of the two things above.

@jdalton would like to refine this so that we can lock down only one or the other of these two things.

The SHA idea would mean that we don't need to lock down the inline styles issue, which means that Angular only need to lock down the unsafe eval, which could be done with `ng-csp`.

If this doesn't work then we should make `ng-csp` configurable to specify which of the two rules we need to lock down.

I would like to test the SHA approach on the browsers that we support before going with the second option.

petebacondarwin:
A final option that I would consider is to simply have the option of a custom build that would exclude restricted aspects. I believe that @IgorMinar has generally been against this idea.

realityking:
Yeah the custom build for the CSS was something I submitted way back: #8459

From the other options, I think I prefer this one:
```
<html ng-csp-inline-style ng-csp-unsafe-eval>
```

petebacondarwin:
I think it would be more intuitive if it was:

```html
<html ng-csp-no-inline-style ng-csp-no-unsafe-eval>
```

petebacondarwin:
Here is another idea:

How about we use the fact that CSP rules can be defined as meta tags in the HTML. Angular will parse the meta tags for CSP rules and work out whether it is allowed to do inline-style and/or unsafe-eval.

This may require rules to be duplicated, both on the server and in the HTML but is not requiring specialised `ng-csp` style directives.

petebacondarwin:
:-( I did some testing and indeed the SHA CSP header for styles doesn't work on

Safari 8.0.7
Firefox 34.0.5

So this is not a viable strategy. Although I think it was a nice idea.

jdalton:
Ok so for the unsafe eval side of things is there anything else to do/tweak?

petebacondarwin:
We need to have a configurable `csp` sniffer value. I see that there are four options:

* different builds of angular for different CSP configurations
* configuration via the value of the `ng-csp="..."` attribute
* configuration via new directives, e.g. `ng-csp-no-unsafe-eval` etc
* configuration via standard CSP meta tags, e.g. `<meta http-equiv="Content-Security-Policy" content="style-src \'self\'>`

I would like to use the last one because, a) it doesn't require multiple build files, b) it is a standard way of specifying CSP directives anyway, and so would make sense to someone reading the HTML who doesn't know about angular.

What do people think?

petebacondarwin:
After a discussion with @btford I am going to knock up a PR with the 4th idea

petebacondarwin:
Actually, I began to implement that and found that there are so many potential combinations of directive rules that we would need to be able to parse it added far too much code.

Instead, I am going to go with the following:

You can choose from the following declarations in your application:

No declaration means that Angular will assume that you can do inline styles, but it will do a runtime check for unsafe-eval. [This is the same as we have currently]
```
<body>
```

A simple `ng-csp` (or `data-ng-csp`) attribute will tell Angular not to do either the inline styles or the unsafe eval. [This is the same as we have currently]
```
<body ng-csp>
```

Providing a value of `allow-unsafe-eval` will tell Angular that it should be able to do unsafe eval; there will be no runtime check. [This is new]
```
<body ng-csp="allow-unsafe-eval">
```

Providing a value of `allow-unsafe-inline` will tell Angular that it should be able to add the inline style; there will be no runtime check. [This is new]
```
<body ng-csp="allow-unsafe-inline">
```

lgalfaso:
@petebacondarwin what would happen if tomorrow there is another CSP parameter we would like it to be configurable? Are we going to keep on adding things to the `ng-csp` attribute? What would happen with the proposal if someone does `ng-csp="allow-unsafe-inline;allow-unsafe-eval"`?

petebacondarwin:
@lgalfaso:

We only care about CSP rules that we actively break inside core, which are doing eval and injecting inline styles. If we started doing something else in core, which would break a CSP rule then we would need to add an option to deactivate that code if the application developer wanted to have that rule turned on.

In the scenario described above `ng-csp="allow-unsafe-inline;allow-unsafe-eval"` would say that we expect there to be no CSP rules that block eval or inline styles, so Angular will go ahead and use them. This is pretty much the same as not using the `ng-csp` attribute at all, except, in this case, we wouldn't even do the runtime check as we have explicitly said that we support unsafe-eval.

That being said I am wondering whether the rules should be the other way around, i.e. the two options would be:

* `no-inline-style`: this would stop Angular from injecting the styles inline (and would require the app developer to link to an external CSS file.
* `no-unsafe-eval`: this would stop Angular from optimising `$parse` with unsafe eval of strings.

Then you would have the following options for `ng-csp`:

No declaration means that Angular will assume that you can do inline styles, but it will do a runtime check for unsafe-eval. [This is the same as we have currently]
```
<body>
```

A simple ng-csp (or data-ng-csp) attribute will tell Angular not to do either the inline styles or the unsafe eval. [This is the same as we have currently]
```
<body ng-csp>
```


This tells Angular that we must not use eval, but that we can inject inline styles.
```
<body ng-csp="no-unsafe-eval">
```

This tells Angular that we must not inject styles, but that we can run eval.
```
<body ng-csp="no-inline-style">
```

This tells Angular that we must not inject styles nor use eval, which is the same as an empty: `ng-csp`
```
<body ng-csp="no-inline-style;no-unsafe-eval">
```

If we add a new feature to Angular that potentially breaks a CSP rule then we would add a switch to turn it off.

jdalton:
@petebacondarwin This sounds reasonable to me. Can you post back a link to your PR once you have it up?

I'll close this one in favor of your PR.

petebacondarwin:
Thanks. I will have it ready in a few hours - just going through the tests and docs

petebacondarwin:
Here you go @jdalton: #12346

jdalton:
Thanks, will check it out :shipit:

PR #11934, opened 2015-05-22
docs(ngRepeat): document that track by must be the last expression

bcherny:
this is a point of confusion that's not documented. see #5520


bcherny:
I signed it!

PR #11935, opened 2015-05-23
fix($location): Infinite digest in $location replace in IE9

petebacondarwin:
@hamfastgamgee - I think this is the right unit test. Can you take a look. The nice thing is that it even fails on Chrome as well (since we are simulating no-history support) without the fix.

### DIFF
--- a/test/ng/locationSpec.js
+++ b/test/ng/locationSpec.js
@@ -996,6 +996,49 @@ describe('$location', function() {
         }
       );
     });
+
+
+    function mockUpBrowser(options) {
+      module(function($windowProvider, $browserProvider) {
+        $windowProvider.$get = function() {
+          var win = {};
+          angular.extend(win, window);
+          win.addEventListener = angular.noop;
+          win.removeEventListener = angular.noop;
+          win.location = {
+            href: options.initialUrl,
+            replace: function(val) {
+              win.location.href = val;
+            }
+          };
+          return win;
+        };
+        $browserProvider.$get = function($document, $window, $log, $sniffer) {
+          /* global Browser: false */
+          var b = new Browser($window, $document, $log, $sniffer);
+          b.baseHref = function() {
+            return options.baseHref;
+          };
+          return b;
+        };
+      });
+    }
+
+
+    it('should not get caught in infinite digest when replacing empty path with slash', function() {
+      initService({html5Mode:true,supportHistory:false});
+      mockUpBrowser({initialUrl:'http://server/base', baseHref:'/base/'});
+      inject(
+        function($browser, $location, $rootScope, $window) {
+          $rootScope.$on('$locationChangeSuccess', function() {
+            if ($location.path() !== '/') {
+                $location.path('/').replace();
+            }
+          });
+          $rootScope.$digest();
+        }
+      );
+    });
   });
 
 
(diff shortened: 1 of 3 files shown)


hamfastgamgee:
Well, this is certainly a much more concise version of what I was directly fixing.  I'm going to update my patch in the other pull request with something I was knocking around this morning that I'd like you to ponder.  I believe the problem is somewhat more widespread than previously thought (impacting more cases) because the mock browser didn't expose the bug.

petebacondarwin:
Perhaps we do need to update ngMock browser across the board as you say?

hamfastgamgee:
Seems at the least like it would be a more honest accounting of what's really going wrong in the app, which should lead to more bullet-proof fixes. :)

hamfastgamgee:
Is there an easy way to run the karma tests on another browser locally?  (I'm relatively new to grunt.)

petebacondarwin:
Easiest way to run tests locally is to use `grunt autotest`. This will setup karma to watch the source and test files and automatically run the tests on any change. Once this is running, you can then attach additional browsers by navigating them to http://localhost:9876

hamfastgamgee:
Thanks!  Please see new version over on #11675.

hamfastgamgee:
So is this fixed, or...?

petebacondarwin:
Moving focus to #11675. I will try to land this in the next couple of days :-)

hamfastgamgee:
Gotcha. Thanks again for all your help! 
On Jun 1, 2015 7:04 AM, Pete Bacon Darwin <notifications@github.com> wrote:Moving focus to #11675. I will try to land this in the next couple of days :-)

test/ng/locationSpec.js
```diff
@@ -996,6 +996,49 @@ describe('$location', function() {
...
+
+    it('should not get caught in infinite digest when replacing empty path with slash', function() {
+      initService({html5Mode:true,supportHistory:false});
+      mockUpBrowser({initialUrl:'http://server/base', baseHref:'/base/'});
+      inject(
+        function($browser, $location, $rootScope, $window) {
+          $rootScope.$on('$locationChangeSuccess', function() {
+            if ($location.path() !== '/') {
+                $location.path('/').replace();
+            }
```
IgorMinar: add an assertion here

test/ng/locationSpec.js
```diff
@@ -996,6 +996,49 @@ describe('$location', function() {
...
+    }
+
+
+    it('should not get caught in infinite digest when replacing empty path with slash', function() {
+      initService({html5Mode:true,supportHistory:false});
+      mockUpBrowser({initialUrl:'http://server/base', baseHref:'/base/'});
+      inject(
+        function($browser, $location, $rootScope, $window) {
+          $rootScope.$on('$locationChangeSuccess', function() {
+            if ($location.path() !== '/') {
```
petebacondarwin: need to assert that this handler was actually called

test/ng/locationSpec.js
```diff
@@ -996,6 +996,49 @@ describe('$location', function() {
...
+                $location.path('/').replace();
+            }
+          });
+          $rootScope.$digest();
+        }
```
petebacondarwin: Need to check that the $location actually got updated after the digest

IgorMinar: add an assertion that we executed the locationChangeSuccess callback

PR #11937, opened 2015-05-24
docs(ngDisabled): describe your change...

Nitish4561:
"disabled" should be written as "ng-disabled".


gkalpak:
This is the _not working_ example, so it is deliberately broken.

Narretz:
We need to do something about these "wrong" examples.

PR #11938, opened 2015-05-24
docs(ngDisabled): correct directive.

Rohit5890:
In example of "ng-disabled" previously disabled was used in place of "ng-disabled" directive.
untitled


gkalpak:
@Rohit5890, thx for the PR, but an identical PR has already been submitted (just a few hours before) :smiley:
Closing as duplicate of #11937.

PR #11946, opened 2015-05-26
fix($animate): ignore invalid option parameter values

matsko:
Prior to this fix there was another patch that threw an exception if the
provided options value was not of an object type. While this is correct
in terms of logic, it caused issues with plugins and tools that are
designed to work with multiple version of Angular. This fix ensures that
these plugins work since an invalid options value is ignored by
`$animate`.

Closes #11826

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -40,6 +40,19 @@ function splitClasses(classes) {
   return obj;
 }
 
+// if any other type of options value besides an Object value is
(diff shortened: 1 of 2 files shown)


matsko:
LGTM (from Igor in person)

matsko:
Landed as https://github.com/angular/angular.js/commit/72edd4dff931c644eecb8f0d1c878dc839c76947

PR #9795, opened 2014-10-26
feat(ngRepeat): add $prev and $next properties to ngRepeat

interstella-5555:
- Adds 2 new properties to ngRepeat item scopes: $prev and $next
- Adds tests to corresponding changes
- Makes simple refactoring on comparison used couple of times for better readability

Closes #9740

PR #11948, opened 2015-05-27
feat(ngRepeat): add $prev and $next

nswbmw:
In some cases, i need to get previous element before this `$index` or get next element after this `$index`. Array works well, but Object not. Maybe others need this feature too.

### DIFF
--- a/src/ng/directive/ngRepeat.js
+++ b/src/ng/directive/ngRepeat.js
@@ -19,6 +19,8 @@
  * | `$last`   | {@type boolean} | true if the repeated element is last in the iterator.                       |
  * | `$even`   | {@type boolean} | true if the iterator position `$index` is even (otherwise false).           |
  * | `$odd`    | {@type boolean} | true if the iterator position `$index` is odd (otherwise false).            |
+ * | `$prev`   | {@type number} {@type string} {@type null} | previous index/key of the iterator position `$index` (null if `$index` is `$first`).  |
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngRepeat.js
```diff
@@ -302,6 +304,8 @@ var ngRepeatDirective = ['$parse', '$animate', function($parse, $animate) {
     // jshint bitwise: false
     scope.$odd = !(scope.$even = (index&1) === 0);
     // jshint bitwise: true
+    scope.$prev = prev;
+    scope.$next = next;
```
caitp: So, I can understand why you'd want this feature, but we do see a real performance impact from adding properties to each scope, it's not ideal. I think we need to be very careful about adding new ones

gkalpak:
Just adding a little context:

There's been a couple of similar PRs already (e.g. #9795, #10362).

--
IIRC, it was discussed during the [planning meeting for 1.4][1].
IIRC again, people (@IgorMinar ?) weren't too keen on adding new properties on all `ngRepeat` generated scopes, but there were some interesting alternatives discussed (e.g. adding them _on-demand_, having a getter property that caclulates them only if they are requested).

I am not sure what the final verdict was (some memory, huh ?) :smiley: 

[1]: https://www.youtube.com/watch?v=Uae9_8aFo-o

Narretz:
I think the consensus was to add the extra properties on-demand. But we didn't really work on it.

PR #11952, opened 2015-05-27
docs: New note in docs for Angular.copy

luckylooke:
Note about keeping same constructor as source for new copied objects.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -739,6 +739,9 @@ function arrayRemove(array, value) {
  *   are deleted and then all elements/properties from the source are copied to it.
  * * If `source` is not an object or array (inc. `null` and `undefined`), `source` is returned.
  * * If `source` is identical to 'destination' an exception will be thrown.
+ * 
+ * **Note:** If you desire to keep same constructor for new copy, you can do following:
+ * `angular.copy(source, new source.constructor());`
  *
  * @param {*} source The source that will be used to make a copy.
  *                   Can be any type, including primitives, `null`, and `undefined`.


luckylooke:
I signed it!

luckylooke:
I dont understand why those checks failed :/

petebacondarwin:
I am not sure this is the correct use of `angular.copy`. It doesn't inherit the prototype from the source.
See http://plnkr.co/edit/CiBJv0v6l4HsFq9oDbuH?p=preview

luckylooke:
If you change method definition to:
```
Obj.prototype.method= function() {
    console.log('in method');
  };
```
it works:
http://plnkr.co/edit/Wgcohybf2yqFh8Wvpxfk

But dont ask me why.. :) Maybe you tell me..

petebacondarwin:
The reason for this is that the original `Obj.prototype` has a reference to the `Obj` function on it as `Obj.prototype.constructor`. In other words, when you create a function, it assigns itself to its prototype property's constructor property.

In my plunker, replacing the prototype with one created from an object literal meant that there was no constructor property specified and the call to `new z.constructor()` falls back on the `Object.constructor` function.

I am now not sure what the aim of your copy is. The `angular.copy` method will delete all local properties from the destination object, so there is little point in running the constructor function to initialize the destination properties.

Are you saying that you want to have an object that inherits the prototypical properties and methods of the source object? If so then I would prefer to use something like `angular.copy(source, Object.create(source.prototype))`.

petebacondarwin:
Actually, my last suggestion is exactly what `angular.copy` does if you don't provide a destination. See https://github.com/angular/angular.js/blob/master/src/Angular.js#L809

luckylooke:
In my use case I needed to copy to be instance of same constructor function as source, and to have same methods in prototype. When I didnt provide destination, then copy was instance of Object not MyObject. You can see it in plunker that y is not Obj but Object. But I agree that the note is no interesting to be in docs anymore. Because I didnt notice that prototype is copied by default ;)

petebacondarwin:
Oh, right, so you were doing testing the "type" of an object by checking its constructor?

luckylooke:
Not exactly, I just have been confused by getting arrays with mixed objects in console. Because objects created via angular.copy were displayed like Objects not MyObject, and I thought that it means that they didn't inherit prototypes (I didn't check that, sorry). :)

Issue #11626, opened 2015-04-17
angular.element.css() method has undocumented inconsistentcy with jQuery api


adam77:
"When a number is passed as the value, jQuery will convert it to a string and add px to the end of that string."
http://api.jquery.com/css/#css2

jqLite does not appear to do this.

Can result in some nasty subtle bugs which only show up when jqLite in use.

petebacondarwin:
@mzgol can you take a look at this?

mzgol:
I wish jQuery didn't do it. :) This basically means any new CSS property that might accept plain numbers as values needs to be added to a whitelist that we maintain; we've recently had to add flexbox-related properties to the list.

A current list can be viewed at http://api.jquery.com/jQuery.cssNumber/, we've made this object public recently.

jqLite diverges from jQuery in many places but in this case. That said, we haven't removed this mechanism exactly because it might broke a lot of code and it won't throw errors but just won't apply some CSS rules so it might, indeed, be hard to debug. This would be pretty trivial to add to jqLite, maybe we should do it for compat with jQuery.

This would be a breaking change for jqLite, though so we shouldn't touch it in 1.3 IMO.

petebacondarwin:
Let's add it as a candidate for 1.5 and just merge the doc update (#11614) for now.

mzgol:
Makes sense, this API requires one to react to new CSS properties so it adds responsibility for jqLite maintenance if we do want to match this jQuery design mistake. ;)

adam77:
jQuery reference:
* [convert parameters of type number] (https://github.com/jquery/jquery/blob/master/src/css.js#L287)
* [list of props not to add 'px' to ('possibly unitless')] (https://github.com/jquery/jquery/blob/master/src/css.js#L230)

mzgol:
@adam77 No need to link to source, it's officially documented under http://api.jquery.com/jQuery.cssNumber/ with more info than in the code as I mentioned in my comment https://github.com/angular/angular.js/issues/11626#issuecomment-96643955

mgol:
We'll be removing this logic in jQuery 4.0, just with a closed whitelist of properties to which we'll still append `px` just for back compat reasons. Because of that, adding it in jqLite doesn't make sense. Closing.

PR #11953, opened 2015-05-27
feat($animateCss): automatically append "px" for positional numeric values

matsko:
Plugins/frameworks like jQuery and GreenSock automatically append a px
value when a number value is provided a style value for an animation.
This patch ensures that the same thing happens on `$animateCss` for
the `from` and `to` styles. This patch also ensures that common
properties such as `z-index` and string-based values are not appended
with a `px` suffix.

### DIFF
--- a/src/ngAnimate/shared.js
+++ b/src/ngAnimate/shared.js
@@ -12,6 +12,7 @@ var isUndefined = angular.isUndefined;
 var isDefined   = angular.isDefined;
 var isFunction  = angular.isFunction;
 var isElement   = angular.isElement;
+var isNumber    = angular.isNumber;
 
 var ELEMENT_NODE = 1;
 var COMMENT_NODE = 8;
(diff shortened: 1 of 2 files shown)


petebacondarwin:
This popped up before. It is actually a discrepancy between jqLite and jQuery. Perhaps we should fix it there if we are bothered about it?

matsko:
Do you have more info? Yes the style application code does run through jqLite/jQuery, but by the time it runs through it then it already has `px` appended to it. Right now with `$animateCss` we require upfront that the user append `px` on their own. So by the time that both scenarios reach jqLite/jQuery they already have `px` appended as a suffix.

matsko:
@petebacondarwin or did you mean that we bring this feature to a lower level and just patch jqLite to do this for us?

petebacondarwin:
This is what I was thinking of : https://github.com/angular/angular.js/issues/11626

src/ngAnimate/shared.js
```diff
@@ -144,10 +145,40 @@ function prepareAnimationOptions(options) {
...
   }
   return options;
 }
 
+function pixelifyNumericStyles(styles) {
+  for (var prop in styles) {
+    // this should satisfy `z-index` and `zIndex` since
+    // both values are always numeric, but don't require a
+    // `px` prefix.
+    if (prop.charAt(0) === 'z') continue;
```
lgalfaso: would it be possible to have a list of the properties that do not need the additional `'px'`?
BTW, jQuery list is
```js
cssNumber: {
		"columnCount": true,
		"fillOpacity": true,
		"flexGrow": true,
		"flexShrink": true,
		"fontWeight": true,
		"lineHeight": true,
		"opacity": true,
		"order": true,
		"orphans": true,
		"widows": true,
		"zIndex": true,
		"zoom": true
	}
```

lgalfaso:
other than this, LGTM

petebacondarwin:
I think we should move this logic into JQLite - see #11626

petebacondarwin:
Closing in favour of #11626 - @matsko do you want to make a PR?

Issue #10527, opened 2014-12-19
ngAnimate doesn't cleanup correctly when on root Element


Narretz:
When you have something like

```<body ng-app="myApp"></body>```

and do
```$animate.addClass(body, 'abc');```

the `ng-animate` class is not removed after the animation finishes. This is because cleanup has a special path if it's the root element: https://github.com/angular/angular.js/blob/master/src/ngAnimate/animate.js#L1512 
(Simple) animations work anyway, so it looks like a minor thing - although the default path does a little more stuff.

matsko:
@Narretz this is fixed in 1.4 as of now. We should fix this in 1.3 however.

Narretz:
1.3 is not worked on anymore.

PR #11956, opened 2015-05-27
fix(ngAnimate): allow animations on body and root elements

OlenDavis:
##### Overview of the Issue:

Animations on `<body>` and the root element don't work.
##### Motivation for or Use Case:

I have a class-based animation on the `<body>` element (though I noticed this would also be an issue for animations on the root element).
##### Angular Version(s):

v1.4.0-rc.2
##### Browsers and Operating System:

Chrome Version 45.0.2414.0 canary (64-bit), Mac OS X Yosemite Version 10.10.3 (14D136)
##### Reproduce the Error:

See [this Plunker](http://plnkr.co/AZEHn1).
##### Related Issues:

The only open issue related to animate and body or root I found was #10527.
##### Suggest a Fix:

All that needed to be fixed was initializing `(body|root)ElementDetected` to whether the current element matches the body or root element rather than false; otherwise, the animated element is never checked as being within the body and root element and can therefore never be animated.

I believe all that needs to be added to this fix is testing if you agree this is a valid and relevant issue as fixed.

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -524,8 +524,8 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     }
 
     function areAnimationsAllowed(element, parentElement, event) {
-      var bodyElementDetected = false;
-      var rootElementDetected = false;
+      var bodyElementDetected = isMatchingElement(element, bodyElement);
+      var rootElementDetected = isMatchingElement(element, $rootElement);
       var parentAnimationDetected = false;
       var animateChildren;
 


petebacondarwin:
@matsko could you take a look at this?
@OlenDavis this change will require a unit test before it can be merged.

matsko:
@petebacondarwin @OlenDavis here is the completed commit with your tests:

https://github.com/matsko/angular.js/commit/2d0913efe9e19f87fe4cc3f0280481c508ad1ebc

OlenDavis:
Do you need me to add those tests to my commit? I'm sorry; it's just that I can't actually run the tests or even complete the normal grunt build locally because I get a `Warning: stderr maxBuffer exceeded. Use --force to continue.` failure every time.

matsko:
No that's fine. The commit I posted can be merged in so long as you're happy with the tests.

OlenDavis:
I'm unfamiliar with $$AnimateRunner and $$animate, but the tests look great; thank you! And, I'm very glad you went the extra inch and added the check for the HTML/document.documentElement element as well as body.

petebacondarwin:
@matsko - do you want to merge this then?

matsko:
This is the final PR that will go in: https://github.com/angular/angular.js/pull/12245

PR #11963, opened 2015-05-28
docs(filter/number): Change description of returned value

calamarico:
I think this is not correct because since 1.3.0-RC.0 sonic-boltification this change: https://github.com/angular/angular.js/commit/2ae10f67fcde3e172f695956301ef796b68a50c2 and null and undefined are passing through


calamarico:
I signed it!

2015-05-28 14:35 GMT+02:00 googlebot <notifications@github.com>:

Issue #11930, opened 2015-05-22
ngOptions incorrectly evaluates label for $$hashKey when watching labels


ryanhart2:
**Overview of the Issue**
Within a directive, there is a variable on the scope which is an array of arrays. The directive template uses ng-repeat to repeat a `<select>` for each of the "child" arrays in this array variable. Each `<select>` has an ng-options attribute with the format *"item as translate(item) for item in array"* where array is provided by the ng-repeat and translate is a function on the scope that translates the item value for display.
At this point the array provided by ng-repeat has a $$hashkey property. This property is correctly omitted from the options on the `<select>`, however, the translate function is invoked for this property in addition to the other items in the array, which is not expected.

**Motivation for or Use Case**
The motivation is to have a directive that can be provided with an array of selections that the user is required to make, with each selection variable being an array of the available options. The directive will then display a select element for each selection containing the relevant options. The directive accepts an optional translate-options attribute which, when set to true, will result in the options being translated. The translation should only affect the display of the options and not the model value.

**Angular Version(s)**
This issue does *not* occur with version 1.3.15
This issue *does* occur with versions 1.4.0-beta.0 through to at least 1.4.0-rc.2

**Browsers and Operating System**
Chrome  43.0.2357.65 m, Windows 8.1

**Reproduce the Error**
http://plnkr.co/edit/nyOOCbqcz3QlPvjddW3R?p=preview

**Related Issues**
None

**Suggest a Fix**
When `Object.keys(values).forEach` is executed below, `values` has a $$hashKey property which is then passed in to the getWatchable function. This results in displayFn being invoked with the locals parameter being an object with a value similar to {value: "object:50"}. A potential fix is for the getWatchable function to ignore keys that begin with '$$'.
```
getWatchables: $parse(valuesFn, function(values) {
  // Create a collection of things that we would like to watch (watchedArray)
  // so that they can all be watched using a single $watchCollection
  // that only runs the handler once if anything changes
  var watchedArray = [];
  values = values || [];

  Object.keys(values).forEach(function getWatchable(key) {
    var locals = getLocals(values[key], key);
    var selectValue = getTrackByValueFn(values[key], locals);
    watchedArray.push(selectValue);

    // Only need to watch the displayFn if there is a specific label expression
    if (match[2] || match[1]) {
      var label = displayFn(scope, locals);
      watchedArray.push(label);
    }

    // Only need to watch the disableWhenFn if there is a specific disable expression
    if (match[4]) {
[...]

PR #11968, opened 2015-05-28
let ngOptions ignore array values starting with `$$`

Narretz:
@gkalpak @petebacondarwin 
I wanted to get the ball rolling on this, but the solution looks pretty ugly tbh. Also, should we equally ignore elements starting with `$`?

edit: actually, it's not array elements, but specifically ngRepeat that sets the `$$hashKey` property on array which are used for ngOptions. So the fix doesn't actually apply.

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -303,6 +303,13 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         values = values || [];
 
         Object.keys(values).forEach(function getWatchable(key) {
+          if (isString(values[key]) &&
+              values[key].charAt(0) == '$' &&
+              values[key].charAt(1) == '$'
+          ) {
+            return;
+          }
+
           var locals = getLocals(values[key], key);
           var selectValue = getTrackByValueFn(values[key], locals);
           watchedArray.push(selectValue);
@@ -348,6 +355,12 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         var optionValuesLength = optionValuesKeys.length;
 
         for (var index = 0; index < optionValuesLength; index++) {
+          if (isString(optionValuesKeys[index]) &&
+              optionValuesKeys[index].charAt(0) === '$' &&
+              optionValuesKeys[index].charAt(1) === '$'
+          ) {
+            continue;
+          }
           var key = (optionValues === optionValuesKeys) ? index : optionValuesKeys[index];
           var value = optionValues[key];
           var locals = getLocals(value, key);
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -922,6 +922,25 @@ describe('ngOptions', function() {
       expect(scope.selected).toEqual([20]);
       expect(element).toEqualSelectValue([20], true);
     });
+
+
+    it('should ignore array values beginning with $$', function() {
+      scope.arr = ['one', 'two', '$$hashKey'];
+
+      scope.modify = function(item) {
+        return item;
+      };
+
+      spyOn(scope, 'modify');
+
+      createSelect({
+        'ng-model': 'selected',
+        'multiple': false,
+        'ng-options': 'item as modify(item) for item in arr'
+      });
+
+      expect(scope.modify).not.toHaveBeenCalledWith('$$hashKey');
+    });
   });
 
 


caitp:
It would be helpful to only include the last commit in this PR, could you fix that and force-push?

Narretz:
I am on it.

Narretz:
Done!

src/ng/directive/ngOptions.js
```diff
@@ -303,6 +303,13 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         values = values || [];
 
         Object.keys(values).forEach(function getWatchable(key) {
+          if (isString(values[key]) &&
```
caitp: If it hasn't been monkey-patched, Object.keys() only produces a list of strings strings (even numeric indexes are strings).

On this point, we should probably make adjustments so that we always use the original un-monkey-patched Object.keys function

src/ng/directive/ngOptions.js
```diff
@@ -348,6 +355,12 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         var optionValuesLength = optionValuesKeys.length;
 
         for (var index = 0; index < optionValuesLength; index++) {
+          if (isString(optionValuesKeys[index]) &&
+              optionValuesKeys[index].charAt(0) === '$' &&
```
caitp: we should cache `optionValuesKeys[index]` so that we only load the field once in unoptimized code

src/ng/directive/ngOptions.js
caitp: Wait, I misread this, `values[key]` might be a string =)

test/ng/directive/ngOptionsSpec.js
```diff
@@ -922,6 +922,25 @@ describe('ngOptions', function() {
       expect(scope.selected).toEqual([20]);
       expect(element).toEqualSelectValue([20], true);
     });
+
+
+    it('should ignore array values beginning with $$', function() {
+      scope.arr = ['one', 'two', '$$hashKey'];
```
caitp: I'm not sure why it makes sense to ignore array elements which are strings beginning with "$$", that doesn't communicate "private" to me

caitp: It makes sense to ignore properties of an array beginning with "$$", but for elements I'm not really convinced

src/ng/directive/ngOptions.js
Narretz: It's not checking the keys, but the values - they can be anything.

caitp: See the comment

caitp: Also as I mention further down, it makes no sense to care about array values, period

test/ng/directive/ngOptionsSpec.js
Narretz: The problem is that this can happen when your ngOptions values are also repeated over by ngRepeat - see this plnkr: http://plnkr.co/edit/eXyWQp9jH2AyflxSCTqL?p=preview This definitely worked differently in 1.3.x

Narretz: But I just saw that the `$$hashKey` is not an array element in this case, so it's probably wrong to check for it like I did.

caitp: I don't think there's ever a good reason to ignore array indexes, unless we decide we want to ignore holes --- but we've decided against that due to the performance impact. This change probably poses a similar performance impact

gkalpak: Performance consideration aside, I don't think this solves the actual issue anyway.
It's indeed the `keys` we need to look for `$$` not the values.

petebacondarwin:
I don't think this is really a problem with Angular. If you pass an array to ngOptions then the $$hashKey will not turn up (unless you iterate over the array as though it was an object - which I don't think you want to do).

petebacondarwin:
@Narretz - let me know if this is not the right decision

ryanhart2:
@petebacondarwin You are correct that the $$hashKey does not turn up in the options. However, a label for $$hashKey is still evaluated (even though the label will never be used). Refer to https://github.com/angular/angular.js/issues/11930.

This is a problem if the function generating the option labels throws an error when it's unexpectedly called with the value of $$hashKey, which is in the format {value: "object:50"}

petebacondarwin:
Thanks @ryanhart2 it helps to have the original issue to refer to.

PR #11969, opened 2015-05-28
Ensure aria-hidden is always a "boolean" (true/false string)

Narretz:
@marcysutton can you take a look?


marcysutton:
@Narretz looks good to me! Nice test coverage. :+1:

gkalpak:
Since this also affects `ngDisabled`/`aria-disabled`, it might be a good idea to have a test for that too.

Narretz:
I added the tests and landed it. Thanks!

PR #11973, opened 2015-05-29
fix($http): continue promise chain with HttpPromise

### LINKED ISSUE #11972: Inconsistent promise chaining with HttpPromise
The `$http` special promise methods `success` and `error` do not always chain as expected.  These should work with the promise chain as expected and also be fully interoperable with the `$q` methods `.then` and `.catch`.

http://plnkr.co/edit/csyUaNojCaUvN6k47V6C?p=preview

        $http.get("https://cors-test.appspot.com/test").success(function (data) {
          return $q(function (_, reject) { reject() });
        }).error(function () {
          console.log("first error should be logged, isn't");
        });
        
        $http.get("https://cors-test.appspot.com/test").success(function (data) {
          return $http.get("http://doesnt.exist")
        }).success(function () {
          console.log("second error is incorrectly considered a success");
        }).catch(function () {
          console.log("second error should be logged, isn't");
        });
        
        $http.get("https://cors-test.appspot.com/test").then(function (data) {
          return $q(function (_, reject) { reject() });
        }).catch(function () {
          console.log("third error should be logged, is");
        });
        
        $http.get("https://cors-test.appspot.com/test").then(function (data) {
          return $http.get("http://doesnt.exist")
        }).catch(function () {
          console.log("fourth error should be logged, is");
        });

This will emit

> second error is incorrectly considered a success
> third error should be logged, is
> fourth error should be logged, is

The issue seems to be that if a promise is returned from the `.success` and `.error` callbacks, it is treated as a truthy value rather than continuing the promise chain.  Instead, they should continue the promise chain.

ajcrites:
`.success` and `.error` methods will now continue the promise chain.
Returning a promise in either of these callbacks will be appropriately
handled by the chain.

Closes #11972

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -911,7 +911,7 @@ function $HttpProvider() {
         assertArgFn(fn, 'fn');
 
         promise.then(function(response) {
-          fn(response.data, response.status, response.headers, config);
+          return fn(response.data, response.status, response.headers, config);
         });
         return promise;
       };
@@ -920,7 +920,7 @@ function $HttpProvider() {
         assertArgFn(fn, 'fn');
 
         promise.then(null, function(response) {
-          fn(response.data, response.status, response.headers, config);
+          return fn(response.data, response.status, response.headers, config);
         });
         return promise;
       };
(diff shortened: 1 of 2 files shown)


gkalpak:
Duplicate of #11097.

PR #11979, opened 2015-05-29
add a endpoint in Readme.closure.md

jb7959:
add a endpoint.


gkalpak:
The sentence continues on the next line (so, there should be no period at the end of the first line).
The `JavaScript` part (at the beginning of the second line) is confusing though...

PR #11982, opened 2015-05-29
docs(guide/Expressions): one-time binding of object literals

pholly:
One time binding of object literals are treated differently than simple expressions. Added a link under the Value stabilization algorithm section to Ben Nadel's article describing how object literals's keys are checked for undefined.


pholly:
Signed CLA

PR #11986, opened 2015-05-30
V1.2.26 is the last working version compatible with Famo.us/Angular

abyss-outpost:
"Error: $animate.enter(...) is undefined" is encountered when using Famo.us/Angular with 
https://ajax.googleapis.com/ajax/libs/angularjs/1.4.0/angular*.js and earlier versions up to 1.2.26.

Could it be readded back as a deprecation for compatibility's sake since another API depends on it? @matsko is who i was referred to by @caitp about this issue.


caitp:
whoa --- this is a bit much =) But @matsko, it might be a good idea to write a compatibility service for the crazy new ngAnimate stuff to minimize the breaking changes (why do we keep adding these pointless breaking changes that make the API harder to use?)

PR #11989, opened 2015-05-31
fix($parse): default expressions of the form `a.b` to `undefined`

### LINKED ISSUE #11959: A problem with boolean expression evaluation in 1.4
Greetings,

After upgrading to angular 1.4 we're having a problem with expressions like:

true && aMissingObject.aMissingProperty

They evaluate to true, instead of false (incidentally, true && !!aMissingObject.aMissingProperty evaluates to false).

Here is an example on plunkr:

http://plnkr.co/edit/4wVwHOO3lsc10JaSrIXI

Could you please comment if this is expected behaviour?

Many thanks

lgalfaso:
When there is an expression of the form `a.b` and there is a recursion that specifies where the value should be placed, then when `a == null` then set the value of the returned identifier to `undefined`

Closes #11959

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -983,6 +983,8 @@ ASTCompiler.prototype = {
             }
           }
           recursionFn(intoId);
+        }, function() {
+          self.assign(intoId, 'undefined');
         });
       }, !!create);
       break;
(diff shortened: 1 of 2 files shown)


lgalfaso:
@petebacondarwin can you take a look at this?

petebacondarwin:
LGTM.

The commit message is a bit confusing though. It sounds to focused on the implementation rather than the issue that is being fixed.

How about:

```
fix($parse): default null references to undefined

When there is an expression of the form `a.b` and where `a == null`, then set
the value of `a.b` to `undefined`.

Closes #11959
```

Or is that too general?

Other than that feel free to merge.

petebacondarwin:
Well done on providing such a quick fix!

lgalfaso:
Fixed and landed as d19504a179355d7801d59a8db0285a1322e04601

PR #11996, opened 2015-06-01
fix(ngOptions): use reference check only when not using trackBy

### LINKED ISSUE #11936: ngOptions with "track by" triggers ngChange when ngModel did not change
**Overview of the Issue**

```html
<select
  ng-options="option.id for option in vm.options track by option.id"
  ng-model="vm.model"
  ng-change="vm.update()">
</select>

```

*vm.options* is an array of objects with a unique *id* property. When the ```<select>``` is first created, ngChange is triggered if ngModel is an object that does not have the same reference as any of the objects in vm.options even though it may be identical in value to one of them. As there has not been a change to the ngModel value, it is not expected that ngChange is triggered.

There are two ```<select>``` elements for the user to select the make and model of a car. The options available for ```<select>```*model* are dependent on the selected value in ```<select>```*make*. Therefore, when the selected value in ```<select>```*make* changes, the selected value in ```<select>```*model* should be cleared. This works with no issues. However, when navigating away from and back to the ```<select>``` elements, the ```<select>```*model* is cleared even though the value of ```<select>```*make* did not change. This is not expected.

**Angular Version(s)**
This issue *does not* occur with version 1.3.15 through to 1.4.0-beta.6
This issue *does* occur with versions 1.4.0-rc.0 through to at least 1.4.0-rc.2

Chrome 43.0.2357.65 m, Firefox 38.0.1, IE 11.0.9600.17801
Windows 8.1 Pro

**Reproduce the Error**
http://plnkr.co/edit/xe4HPLZVEHqO7XYW1kGX?p=preview

**Related Issues**
https://github.com/angular/angular.js/pull/11448
https://github.com/angular/angular.js/issues/11447

**Suggest a Fix**
The issue seems to result from this commit: https://github.com/angular/angular.js/commit/171b9f7f2339ef9047b8526b2c3f36bb58d14feb

In the code below, when ```previousValue !== nextValue``` is true, ```ngModelCtrl.$setViewValue(nextValue)``` is invoked, even though ```ngOptions.trackBy``` is true and ```!equals(previousValue, nextValue)``` is false.

```js
// Check to see if the value has changed due to the update to the options
if (!ngModelCtrl.$isEmpty(previousValue)) {
  var nextValue = selectCtrl.readValue();
  if (ngOptions.trackBy && !equals(previousValue, nextValue) ||
        previousValue !== nextValue) {
    ngModelCtrl.$setViewValue(nextValue);
    ngModelCtrl.$render();
  }
}
```

The code to compare *previousValue* and *nextValue* could be changed to:
```js
if (ngOptions.trackBy ? !equals(previousValue, nextValue) : previousValue !== nextValue) {
```

This would result in a deep comparison when *track by* is used and a reference comparison when *track by* is not used.

seaster:
Change the check on changed value for ngOptions to ensure that a reference check is only done
when trackBy is not used.  Previously, if trackBy was being used but the values of the objects were
equal a reference check was then used which would cause the check to be true if the objects had
different memory references but the values were equal.  This can cause issues with forms being
marked dirty when the value of the model as not actually changed.

Closes #11936

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -706,8 +706,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         // Check to see if the value has changed due to the update to the options
         if (!ngModelCtrl.$isEmpty(previousValue)) {
           var nextValue = selectCtrl.readValue();
-          if (ngOptions.trackBy && !equals(previousValue, nextValue) ||
-                previousValue !== nextValue) {
+          if (ngOptions.trackBy ? !equals(previousValue, nextValue) : previousValue !== nextValue) {
             ngModelCtrl.$setViewValue(nextValue);
             ngModelCtrl.$render();
           }
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -1176,6 +1176,26 @@ describe('ngOptions', function() {
 
     });
 
+    it('should not set view value again if the tracked property of the model has not changed when using trackBy', function() {
+
+      createSelect({
+        'ng-model': 'selected',
+        'ng-options': 'item for item in arr track by item.id'
+      });
+
+      scope.$apply(function() {
+        scope.selected = {id: 10, label: 'ten'};
+      });
+
+      spyOn(element.controller('ngModel'), '$setViewValue');
+
+      scope.$apply(function() {
+        scope.arr[0] = {id: 10, label: 'ten'};
+      });
+
+      expect(element.controller('ngModel').$setViewValue).not.toHaveBeenCalled();
+    });
+
     it('should not re-render if a property of the model is changed when not using trackBy', function() {
 
       createSelect({


test/ng/directive/ngOptionsSpec.js
```diff
@@ -1176,6 +1176,28 @@ describe('ngOptions', function() {
...
+        scope.selected = {id: 10, label: 'ten'};
+      });
+
+      spyOn(element.controller('ngModel'), '$setViewValue');
+
+      scope.$apply(function() {
+        scope.arr[0] = {id: 10, label: 'ten'};
+      });
+
+      // update render due to equality watch
```
Narretz: This line can be removed

src/ng/directive/ngOptions.js
```diff
@@ -707,7 +707,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         if (!ngModelCtrl.$isEmpty(previousValue)) {
           var nextValue = selectCtrl.readValue();
           if (ngOptions.trackBy && !equals(previousValue, nextValue) ||
-                previousValue !== nextValue) {
+              !ngOptions.trackBy && previousValue !== nextValue) {
```
Narretz: I'd prefer `if (ngOptions.trackBy ? !equals(previousValue, nextValue) : previousValue !== nextValue)`

Narretz:
LGTM

petebacondarwin:
LGTM - feel free to merge

PR #11997, opened 2015-06-01
fix(ngAnimate): do not assume $document[0] exists

### LINKED ISSUE #11964: $window mocking not working with ngAnimate since 1.4.0
I have tests for a directive that uses `$window` and `ngAnimate`. Since 1.4.0 (or 1.4.0-rc.0) I cannot mock `$window` without having `document` property set in `$window`.

This issue is due to `var bodyElement = jqLite($document[0].body);` in `ngAnimate` `animateQueue.js` https://github.com/angular/angular.js/blob/master/src/ngAnimate/animateQueue.js#L108.

I have a JSFiddle demonstrating this problem:
- [Both tests are passing in 1.3.15](http://jsfiddle.net/tjokimie/g7wndnb5/1/)
- [Test without `document` is failing in 1.4.0](http://jsfiddle.net/tjokimie/s9jLkhhj/1/)

tjokimie:
Use document.body instead of $document[0].body. Enable mocking of $window
without having to add an empty document object into it.

Closes #11964

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -66,10 +66,10 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     return (nO.addClass && nO.addClass === cO.removeClass) || (nO.removeClass && nO.removeClass === cO.addClass);
   });
 
-  this.$get = ['$$rAF', '$rootScope', '$rootElement', '$document', '$$HashMap',
-               '$$animation', '$$AnimateRunner', '$templateRequest', '$$jqLite',
-       function($$rAF,   $rootScope,   $rootElement,   $document,   $$HashMap,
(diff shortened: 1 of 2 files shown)


Narretz:
@matsko can you take a look at this? I'm not sure how strict we should be with these injections.

lgalfaso:
I think this is not an issue. If you want to mock `$window` then you should provide the document. Someone else might want to mock `$document` and if we use `document` here, then it would be impossible

tjokimie:
@lgalfaso Previously this has been an issue in `$sce` and it was fixed https://github.com/angular/angular.js/pull/9663.

matsko:
The reason why we use `$document` in the first place is to allow for mocking. This commit further limits things since we have to change the code afterwards to pull document from $window. Unfortunately you'll need to inject both. Sorry :{

PR #12001, opened 2015-06-01
Allow interpolation inside ng-message (fix for #11616)

### LINKED ISSUE #11616: ng-messages no longer interprets its attribute value properly
In order to minimize the amount of html needed for validation, at our office, we created a directive on top of ng-messages. Sadly, moving from 1.4.0-beta.4/angular-messages.js to 1.4.0-beta.5/angular-messages.js (and later version). Our code no longer works.

The error is as follows:
```
Syntax Error: Token '{' invalid key at column 2 of the expression [{{::messages.fullFieldName}}.$error] starting at [{::messages.fullFieldName}}.$error].
```

The template for this directive is as follows:

```html
<div class="validation-message" ng-messages="{{::messages.fullFieldName}}.$error" ng-show="{{::messages.fullFieldName}}.$invalid && !{{::messages.fullFieldName}}.$pending && ({{::messages.formName}}.$submitted || {{::messages.fullFieldName}}.$touched)">
  <div ng-transclude></div>
</div>
```

The directive is defined as follows:
```javascript
function indMessages() {
    var controller = function ($element) {
      var vm = this;
      var form = $element.inheritedData('$formController');

      // Unfortunately required controllers are not injectable into our own controller, only into our link function.
      // Here's a solution, feels a bit hacky but nicer than creating a link function just for this; based on: http://stackoverflow.com/a/21252383/198797; also used in core angular code: https://github.com/angular/angular.js/blob/adcc5a00bf582d2b291c18e99093bb0854f7217c/src/ng/directive/input.js#L1614
      vm.formName = form.$name;
      vm.fullFieldName = vm.formName + '.' + vm.fieldName;
    }

    controller.$inject = ['$element'];

    return {
      restrict: 'E',
      require: '^^form',
      templateUrl: '/views/templates/ind-messages.html',
      bindToController: {
        fieldName: "@for"
      },
      controllerAs: 'messages',
      transclude: true,
      replace: true,
      scope: true,

      controller: controller
    };
  }
```

Is the above due to a programming error on our part? Or was there a change with unexpected side-effects inside the angular-messages.js code?

sravenhorst:
https://github.com/angular/angular.js/issues/11616

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -320,8 +320,13 @@ angular.module('ngMessages', [])
      var INACTIVE_CLASS = 'ng-inactive';
 
      return {
-       require: 'ngMessages',
        restrict: 'AE',
+       bindToController: {
+         ngMessages: '@',
+         forAttr: '@for'
+         },
+       scope: true,
+       controllerAs: 'ngMessagesController',
        controller: ['$element', '$scope', '$attrs', function($element, $scope, $attrs) {
          var ctrl = this;
          var latestKey = 0;
@@ -384,7 +389,7 @@ angular.module('ngMessages', [])
               : $animate.setClass($element, INACTIVE_CLASS, ACTIVE_CLASS);
          };
 
-         $scope.$watchCollection($attrs.ngMessages || $attrs['for'], ctrl.render);
+         $scope.$watchCollection(ctrl.ngMessages || ctrl.forAttr, ctrl.render);
 
          this.reRender = function() {
            if (!renderLater) {


sravenhorst:
I have just signed the "first contribution" report.

Narretz:
Thanks for the PR. As you wrote in the issue, couldn't you also simply move the watch logic back to the link function?

Narretz:
Also, we need a test for this.

src/ngMessages/messages.js
```diff
@@ -320,8 +320,13 @@ angular.module('ngMessages', [])
...
 
      return {
-       require: 'ngMessages',
        restrict: 'AE',
+       bindToController: {
+         ngMessages: '@',
+         forAttr: '@for'
+         },
+       scope: true,
+       controllerAs: 'ngMessagesController',
```
Narretz: This and `scope: true` are changing the semantics of the directive. Not sure if we want this.

lgalfaso: Agree with @Narretz we do not want to add this new scope

PR #12010, opened 2015-06-02
fix(ngOptions): do not watch properties starting with $

### LINKED ISSUE #11930: ngOptions incorrectly evaluates label for $$hashKey when watching labels

petebacondarwin:
Expressions that compute labels and track by values for ngOptions were
being called for properties, that start with $ even though those properties
were being ignored as options.

Closes #11930

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -303,6 +303,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         values = values || [];
 
         Object.keys(values).forEach(function getWatchable(key) {
+          if (key.charAt(0) === '$') return;
           var locals = getLocals(values[key], key);
           var selectValue = getTrackByValueFn(values[key], locals);
           watchedArray.push(selectValue);
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -448,6 +448,41 @@ describe('ngOptions', function() {
   });
 
 
+  it('should not watch array properties that start with $ or $$', function() {
+    createSelect({
+      'ng-options': 'value as createLabel(value) for value in array',
+      'ng-model': 'selected'
+    });
+    scope.createLabel = jasmine.createSpy('createLabel').andCallFake(function(value) { return value; });
+    scope.array = ['a', 'b', 'c'];
+    scope.array.$$private = 'do not watch';
+    scope.array.$property = 'do not watch';
+    scope.selected = 'b';
+    scope.$digest();
+
+    expect(scope.createLabel).toHaveBeenCalledWith('a');
+    expect(scope.createLabel).toHaveBeenCalledWith('b');
+    expect(scope.createLabel).toHaveBeenCalledWith('c');
+    expect(scope.createLabel).not.toHaveBeenCalledWith('do not watch');
+  });
+
+
+  it('should not watch object properties that start with $ or $$', function() {
+    createSelect({
+      'ng-options': 'key as createLabel(key) for (key, value) in object',
+      'ng-model': 'selected'
+    });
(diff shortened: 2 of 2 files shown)


src/ng/directive/ngOptions.js
```diff
@@ -303,6 +303,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         values = values || [];
 
         Object.keys(values).forEach(function getWatchable(key) {
+          if (key.charAt(0) === '$') return;
```
gkalpak: Do we really want to ignore `$` (and not just `$$`) properties ?
`$$` are (by convention) private, but `$` are (by convention again) Angular-specific, but not private.

petebacondarwin: @gkalpak - good question. Right now we are ignoring such properties as options, so there is a precedent: https://github.com/angular/angular.js/blob/master/test/ng/directive/ngOptionsSpec.js#L434-L448

Also `angular.equals` and `ngRepeat` ignore single `$` too.

gkalpak: Good point. For some reason, [`ToJsonReplacer()`][1] was modifed some time ago to only ignore `$$` properties (but everything else (e.g. `ngRepeat`, `ngOptions`, `equals`) kept ignoring `$` as well. Too much breaking change I guess :)

So, it's fine as it is :smiley:

[1]: https://github.com/angular/angular.js/blob/d7dc14dc0cdeb9c187d227e19acc8aca7df9d740/src/Angular.js#L1102-L1116

Narretz:
LGTM

PR #12012, opened 2015-06-02
fix:$http:Silently fail when isJsonLike wrongly detects JSON

fungiboletus:
isJsonLike can have false positives.  Instead of crashing, the defaultHttpHandler should return the string.

It may be a problem for developers trying to load a broken JSON document without using the correct Content-Type. A log message could be added to the try-catch.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -93,8 +93,12 @@ function defaultHttpResponseTransform(data, headers) {
 
     if (tempData) {
       var contentType = headers('Content-Type');
-      if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData)) {
+      if (contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) {
         data = fromJson(tempData);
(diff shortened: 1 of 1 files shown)


gkalpak:
@yellowiscool, thx for the PR. We should have a test verifying the new behaviour (i.e. a test that would fail prior to this change and passes with the change).

petebacondarwin:
Is this PR still relevant to 1.5 and 1.6?

petebacondarwin:
Looks like it was not changed yet. Adding to 1.6.x

Narretz:
We made a different [change](https://github.com/ellimist/angular.js/commit/efdd5a11ba98e648a8c0b57950924e57718ef0ee) to this behavior - http will now throw a specfic error when the data cannot be parsed as json.
It's not ideal that $http tries to parse everything as JSON based on isJSONLike, but changing this now would be a significant breaking change (we don't know how many people send JSON without proper headers). If you expect pure text or broken JSON, then changing the default transforms is the better option, see https://github.com/angular/angular.js/issues/15897

PR #12014, opened 2015-06-02
docs(filter): Documented third argument of predicate function

ThomasLandauer:


ThomasLandauer:
I signed it!

src/ng/filter/filter.js
```diff
@@ -34,8 +34,8 @@
  *     `{name: {first: 'John', last: 'Doe'}}` will **not** be matched by `{name: 'John'}`, but
  *     **will** be matched by `{$: 'John'}`.
  *
- *   - `function(value, index)`: A predicate function can be used to write arbitrary filters. The
- *     function is called for each element of `array`. The final result is an array of those
+ *   - `function(value, index, array)`: A predicate function can be used to write arbitrary filters. The
+ *     function is called for each element of `array`, with 3 arguments beeing passed in: the current element's value; its index in the array; the entire array. The final result is an array of those
```
gkalpak: This line is too long. According to the [coding rules][1], lines should be wrapped at 100 characters.

Nitpicking, I find it too verbose (and the phrase `the current element's value` is confusing imo; just `element` sounds clearer). I would probably prefer something like:

> The function is called for each element of `array`, with the element's index and the array itself as extra arguments.

[1]: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-coding-rules

Narretz:
@ThomasLandauer can you please squash these two commits into one and update the branch (git push -f)?

ThomasLandauer:
Sorry, I just use GitHub through the Website and don't know how to do this :-( You can just go ahead and submit it with your name - or whatever. It's just a minor change.......

Narretz:
Landed as 46b7cf7464ef7f37b3d2919847627e58ff7fdca8

PR #12016, opened 2015-06-03
docs(ngClass): Put single quotes around classes

qbzzt:
For the classes in the map notation ({'strike': deleted, 'bold': important, 'red': error}). I don't know why, but on Chrome sometimes it doesn't work without those quotes.

### DIFF
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -162,7 +162,7 @@ function classDirective(name, selector) {
  * @example Example that demonstrates basic bindings via ngClass directive.
    <example>
      <file name="index.html">
-       <p ng-class="{strike: deleted, bold: important, red: error}">Map Syntax Example</p>
+       <p ng-class="{'strike': deleted, 'bold': important, 'red': error}">Map Syntax Example</p>
        <label>
           <input type="checkbox" ng-model="deleted">
           deleted (apply "strike" class)


qbzzt:
I signed it!

gkalpak:
Do you mean the docs example doesn't work without the quotes ir something in your app ?
I have never come across this issue before (and I've used my fair share of `ngClass` :smiley:).

--
Note that the quotes are necessary for classes contsining hypens (or other special characters).

E.g. `{bordered: hasBorder}` is fine, but `{'with-border': hasBorder}` needs the quotes.

qbzzt:
You're right, I was using bootstrap with btn-lg. I was just too tired to
remember that it needs quotes in that case.

However, I would still request that you keep my edit. Those quotes don't
hurt, and they might save somebody else who copies without thinking.

    Thanks,
         Ori

gkalpak:
I don't know what others think, but those quotes do hurt my eyes a little bit (especially when I have to escape yhem inside a directive template string) :smiley:

I would rather not merge this (after all we can't really help people that _copy without thinking_ :wink:).
I would be happy to merge a PR which changes on of the classes to include a hypen (and puts quotes around that one only).

How does this sound to you, @qbzzt ?

qbzzt:
Great. I'll write that later today.

PR #12032, opened 2015-06-05
docs(CONTRIBUTING): revert should be a valid type

stevemao:
EG: https://github.com/angular/angular.js/commit/462f444b06ae5cad3ccb761b1dba7131df01a655

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12032)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -238,6 +238,7 @@ The body should include the motivation for the change and contrast this with pre
 The footer should contain any information about **Breaking Changes** and is also the place to
 reference GitHub issues that this commit **Closes**.
 
+**Breaking Changes** should start with the word `BREAKING CHANGE:` with a space or two newlines. The rest of the commit message is then used for this.
 
 A detailed explanation can be found in this [document][commit-message-format].
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -199,6 +199,8 @@ format that includes a **type**, a **scope** and a **subject**:
 <footer>
 ```
 
+The **header** is mandatory and the **scope** of the header is optional.
+
 Any line of the commit message cannot be longer 100 characters! This allows the message to be easier
 to read on github as well as in various git tools.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -202,6 +202,9 @@ format that includes a **type**, a **scope** and a **subject**:
 Any line of the commit message cannot be longer 100 characters! This allows the message to be easier
 to read on github as well as in various git tools.
 
+### Revert
+If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
+
 ### Type
 Must be one of the following:
 


petebacondarwin:
Well, strictly, `revert` is a modifier that you stick on the front of another commit's message.

petebacondarwin:
Can you describe it like that instead?

stevemao:
Since there is only one "modifier", should I just say `revert`?
I'm not 100% sure how you wanna write it but I'll push up something.

stevemao:
Also I'm maintaining the module [conventional-changelog](https://github.com/ajoslin/conventional-changelog) used by you guys.
what do you expect conventional-changelog to do to this? I would simply parse it as a `type` so it will look the same as other types. 

```md
## Bug Fixes
...

## Reverts
...
```

petebacondarwin:
Well at the moment, we are not actually using your package, Instead we use an script inside the angular.js project, `changelog.js`. And in this script you can see that we ignore all commits that do not start with `fix`, `feat`, `perf` or contain the word `BREAKING`. See https://github.com/angular/angular.js/blob/master/changelog.js#L194

The general idea, though, as far I am concerned, is that reverted commits should be removed from the changelog. So assuming that the original commit is also in the current release then we should simply ignore both commits. If it is reverting from a previous release, then I agree that we should have some kind of `## Reverts` section.

stevemao:
Yup, that script is actually the very first version of conventional-changelog. You guys should definitely use the package once I release `v0.1.x`. It's much better :)
I agree with you. So far it's not smart enough to remove both commits within one release. Thanks for the feedback @petebacondarwin 

Let me know what you think of the docs.

CONTRIBUTING.md
```diff
@@ -192,7 +192,7 @@ Each commit message consists of a **header**, a **body** and a **footer**.  The
 format that includes a **type**, a **scope** and a **subject**:
 
 ```
-<type>(<scope>): <subject>
+(revert: )<type>(<scope>): <subject>
```
petebacondarwin: Hmm, the parentheses here on `(revert: )` are because it is optional, right? But the parentheses on `(<scope>)` are because they are needed in the string. Perhaps we should just not include the `revert` bit in this and just explain about reverts as you do very clearly below.

CONTRIBUTING.md
```diff
@@ -202,6 +202,10 @@ format that includes a **type**, a **scope** and a **subject**:
 Any line of the commit message cannot be longer 100 characters! This allows the message to be easier
 to read on github as well as in various git tools.
 
+### Revert
+
+If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`
+
```
petebacondarwin: (where the hash is the SHA of the commit being reverted) :-)

CONTRIBUTING.md
stevemao: Oh right. I was thinking the scope is actually optional. Maybe we should note what's optional what's not?

stevemao:
One thing I found about the commit message guide is that it's actually not very strict:
1. It doesn't say what's optional what's not.
2. The exact format of a part.
For example, it doesn't say how **Breaking Changes** should be written. The sentense should start with `Breaking Change: ` and there are optional two newlines between `Breaking Change: ` and the content of it.

stevemao:
This applies to the google doc you link to in the file. It gives more examples but still not very strict.

stevemao:
I added two commits just to support my points (Maybe they should be in different PR but they kinda related and I need your opinions). Let me know what you think. Thanks.

stevemao:
ping @petebacondarwin

gkalpak:
Fun fact: According to the `validate-commit-msg.js` file, `revert` is considered a valid type since Nov 2012 (and there are tests that verify it - which probably no one runs :smiley:).

stevemao:
hm... lol. `conventional-changelog` treats `revert` as a type too (if it reverts a commit that's not in the current release) and it works well. There is no need to make an exceptional rule for it.

for `validate-commit-msg.js` I think it should verify the reverted commit actually exists.

petebacondarwin:
LGTM!

stevemao:
Thanks @petebacondarwin , @gkalpak anything you would like to add here?

stevemao:
Would be great if you guys could update the docs on google docs
https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#

Or, even better, create a git repo just for the conventions?

stevemao:
Thanks @petebacondarwin

petebacondarwin:
@stevemao I have also updated the google doc.

stevemao:
Nice, will check it out

PR #12035, opened 2015-06-05
Decreases indentation to fix markdown

josherick:
Two bullet points were indented causing the entire lines to be formatted as code by markdown. Decreased indentation level by one, as it seems this was not the intention.


josherick:
I signed it!

PR #12036, opened 2015-06-05
fix(compile): assign controller return value to correct controller

### LINKED ISSUE #12029: ngModel generates errors when in a form element with a ngController directive
The following HTML snippet generates errors under 1.4.0, but works with 1.3.15.

ngModelDirective generates errors in ngModelPreLink because it inadvertently references myController rather than the expected FormController.

```html
<html>
<body ng-app="myApp">
	<form ng-controller="myController">
		<input ng-model="myModel" />
	</form>
	<script src="http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.0/angular.js"></script>
	<script>
		angular.module("myApp", []).controller("myController", function () { 
                  return {}; 
                })
	</script>
</body>
</html>
```

Narretz:
Fixes #12029

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1997,9 +1997,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           for (i in elementControllers) {
             controller = elementControllers[i];
             var controllerResult = controller();
+
             if (controllerResult !== controller.instance) {
+              // If the controller constructor has a return value, overwrite the instance
+              // from setupControllers and update the element data
               controller.instance = controllerResult;
-              $element.data('$' + directive.name + 'Controller', controllerResult);
+              $element.data('$' + i + 'Controller', controllerResult);
               if (controller === controllerForBindings) {
                 // Remove and re-install bindToController bindings
                 thisLinkFn.$$destroyBindings();
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -4573,6 +4573,41 @@ describe('$compile', function() {
     });
 
 
+    it('should correctly assign controller return values for multiple directives', function() {
+      var directiveController, otherDirectiveController;
+      module(function() {
+
+        directive('myDirective', function(log) {
+          return {
+            scope: true,
+            controller: function($scope) {
+              return directiveController = {
+                foo: 'bar'
+              };
+            }
+          };
+        });
+
+        directive('myOtherDirective', function(log) {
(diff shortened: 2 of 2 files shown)


src/ng/compile.js
```diff
@@ -1983,9 +1983,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           var scopeDirective = newIsolateScopeDirective || newScopeDirective;
           var bindings;
           var controllerForBindings;
+          var scopeController;
           if (scopeDirective && elementControllers[scopeDirective.name]) {
             bindings = scopeDirective.$$bindings.bindToController;
-            controller = elementControllers[scopeDirective.name];
+            controller = scopeController = elementControllers[scopeDirective.name];
```
Narretz: I don't know what else scopeDirective can be, so scopeController might not be the best name.

src/ng/compile.js
```diff
@@ -1997,9 +1998,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           for (i in elementControllers) {
             controller = elementControllers[i];
             var controllerResult = controller();
+
             if (controllerResult !== controller.instance) {
               controller.instance = controllerResult;
-              $element.data('$' + directive.name + 'Controller', controllerResult);
+              var name = controller === scopeController ? scopeDirective.name : directive.name;
```
caitp: if `controller === scopeController`, isn't `directive === scopeDirective` too?

caitp: maybe not.

Narretz:
@caitp Can you check this again? The problem is actually much more general. Basically, the return values for each controller was always updated for the same directive.

src/ng/compile.js
```diff
@@ -1997,9 +1997,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           for (i in elementControllers) {
             controller = elementControllers[i];
             var controllerResult = controller();
+
             if (controllerResult !== controller.instance) {
+              // Assign controller return value to the instance
               controller.instance = controllerResult;
-              $element.data('$' + directive.name + 'Controller', controllerResult);
+              $element.data('$' + i + 'Controller', controllerResult);
```
caitp: Looks good to me, maybe the comment could be a bit more clear --- we're overwriting stuff that was done in `setupControllers()` because the return value was different from the pre-allocated instance.

caitp:
lgtm

PR #12041, opened 2015-06-06
perf($compile): only use document fragments when necessary

jbedard:
This mainly just explains the TODO(perf). I'm sure the actual impact is negligible, but will avoid creating fragments in the majority of cases (where only one node is being replaced).

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2494,9 +2494,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         parent.replaceChild(newNode, firstElementToRemove);
       }
 
-      // TODO(perf): what's this document fragment for? is it needed? can we at least reuse it?
-      var fragment = document.createDocumentFragment();
-      fragment.appendChild(firstElementToRemove);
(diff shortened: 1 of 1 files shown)


lgalfaso:
I am not sure about this change. I see that there might be some performance improvements, but it will also break calling `parent.children` when there is only one element. I see that this is a very edge case, but people might be using this

jbedard:
Really it's just for [groupScan](https://github.com/angular/angular.js/blob/9efb0d5ee961b57c8fc144a3138a15955e4010e2/src/ng/compile.js#L1601). Once a node has been replaced I don't think anything is supported on that node? Doesn't mean people don't use it though. I could change this to just update the comment instead?

jbedard:
Note that if #12094 merges as is then this PR won't be possible anymore (since #12094 always makes use of the document fragment).

lgalfaso:
I think it is better to land #12094. Later, if there is a way to remove the document fragment, then I am sure that @jbedard will find a way :)

petebacondarwin:
Closing in favour of #12094

PR #12042, opened 2015-06-06
perf($compile): avoid jquery data calls when there is no data

jbedard:
This avoids copying jq data for the majority of cases where there is no data. Previously when there was no data `.data()` was called which would create an empty data object ([using `defineProperties` on the node](https://github.com/jquery/jquery/blob/2.1-stable/src/data/Data.js#L36-51) which is pretty slow, just to be cleaned right after).


lgalfaso:
landed as 9efb0d5ee961b57c8fc144a3138a15955e4010e2

PR #12043, opened 2015-06-06
perf($compile): use static jquery data method to avoid creating new instances

jbedard:
Really minor, but the static `jqLite.data` does the same thing when there is only one node...

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2505,7 +2505,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         // Copy over user data (that includes Angular's $scope etc.). Don't copy private
         // data here because there's no public interface in jQuery to do that and copying over
         // event listeners (which is the main use of private data) wouldn't work anyway.
-        jqLite(newNode).data(jqLite(firstElementToRemove).data());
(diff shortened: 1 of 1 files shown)


lgalfaso:
Mmm... not sure about this change. Creating a new instance should not be that expensive for us to have to optimize this...

jbedard:
It's not expensive at all in a method like this. But it's also simpler (less code) imo so it seemed like a positive change to me...

lgalfaso:
using `data` (or any other method) this way is documented nor supported. I see no reason why we should start using them in this way

jbedard:
The jquery version is [documented](https://api.jquery.com/jquery.data/) and the jqlite version is equivalent (although not documented).  They are already used in a few places where it does actually make a difference (linking).

jbedard:
Also note that the `hasData` call right above this line is also the static version (and there is no instance version), so I think it makes sense for this line to do the same.

lgalfaso:
Mmm... ok.

LGTM

petebacondarwin:
Landed as 9b90c32f31fd56e348539674128acec6536cd846

PR #12044, opened 2015-06-07
fix(angularCopy): out putValue function from copy

behzad888:
Out putValue function from copy


behzad888:
I signed it!

lgalfaso:
The function itself it going out once #11215 lands, so there is no need for this. Thanks for the PR anyhow!

PR #12046, opened 2015-06-07
docs(orderBy): correctly is not the right word

dwanderton:
The use of correctly implies that Angular is doing something incorrect, however it is that we expect a number passed as a string to be sorted as a number. Angular does not do what we are expecting, although it responds correctly to what we have actually asked - sorting based on the string representation of a number.


dwanderton:
I signed it!

PR #12049, opened 2015-06-08
Update README.md

hardikvasa:
Minor modifications to the readme.md for clear understanding

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@ AngularJS [![Build Status](https://travis-ci.org/angular/angular.js.svg?branch=m
 =========
 
 AngularJS lets you write client-side web applications as if you had a smarter browser.  It lets you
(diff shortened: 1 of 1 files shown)


hardikvasa:
I signed it!

Narretz:
Thanks for the PR. However, I am not 100% sold that these are clearer. What's wrong with "syntax" instead of "functionality"? And two-way data-binding is a pretty established concept.

pkozlowski-opensource:
Agree with @Narretz - IMO the existing wording is better. -1 for this change from me.

hardikvasa:
I just wanted to make it a little clearer. The word 'Syntax' is little specific to its usage pattern. The sentense denotes that one can extend its features (and not modify HTML statements)...and so a little generalized term 'functionality'. Hope it helps! Sorry if it is not found appropriate.

Narretz:
While the readme text isn't that great, this is not a huge improvement either.

PR #12050, opened 2015-06-08
fix(angularCopy): out putValue function from copy 

behzad888:


lgalfaso:
This is a duplicate of #12044. Please do not open multiple PRs with the
same content.

behzad888:
Ok

PR #12054, opened 2015-06-09
perf(copy): avoid regex in isTypedArray

jbedard:
Simplifying and avoiding the regex makes all the copy scenarios I tried 5-15% faster...

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -659,9 +659,9 @@ function isPromiseLike(obj) {
 }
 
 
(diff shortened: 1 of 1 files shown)


lgalfaso:
Would it be possible for you to share the setup that you are using to check for these performance gains?

jbedard:
Basically something such as f7bf118ecaaab16ed8b3d16febb0ee6bb1d6749c. Random test snippets within a benchpress app...

lgalfaso:
LGTM

PR #12061, opened 2015-06-09
test($compile): fix test

gkalpak:
The fixed test is supposed to test a fix for an IE11 bug/peculiarity that arises when using a specifically configured MutationObserver on the page (see #11781 for more info).

The configuration contained a typo (`sublist` instead of `subtree`), which effectively failed to set up the MutationObserver in a way that would make the IE11 bug appear.


lgalfaso:
LGTM

PR #12065, opened 2015-06-09
perf(copy): use different arrays for different types

lgalfaso:
Use different arrays for different types.
Remove forEach to clean the destination object.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -783,98 +783,121 @@ function arrayRemove(array, value) {
  </file>
  </example>
  */
-function copy(source, destination, stackSource, stackDest) {
-  if (isWindow(source) || isScope(source)) {
-    throw ngMinErr('cpws',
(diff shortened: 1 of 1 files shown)


lgalfaso:
using the same test at f7bf118ecaaab16ed8b3d16febb0ee6bb1d6749c, there is a 10% improvement

jbedard:
I haven't created a PR yet but see ec8d48835a74b8ccb468193c1a8e28ed50d9e415. It reduces the copy time by 30-60% while also splitting the method up similar to this.

I didn't think of separate arrays though. But I'm not sure if it's worth the extra code and duplication imo. I'd still vote for using `Map`, but that had its own issues...

lgalfaso:
The use of `Map` would be ideal, but is not present in all the supported browsers and adding a polyfill would add a lot of code.

PR #12068, opened 2015-06-10
perf(copy): only validate/clear user specified destination

jbedard:
This (combined with #12054) makes all the copy situations I tried 30-60% faster (~1.5-3x).

This mainly just moves/splits:
- input validation/clearing
- different object type copying
- different object type recursing

Into 3 separate methods, allowing:
- only the input destination (if specified) is validated
- only the input destination (if specified) has the content cleared
- only the public method has the `if (destination)` confusion
- removing all the checks if the stack vars have been initialized yet

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -794,100 +794,106 @@ function arrayRemove(array, value) {
  </file>
  </example>
  */
-function copy(source, destination, stackSource, stackDest) {
+function copy(source, destination) {
+  if (destination) {
+    if (isTypedArray(destination)) {
+      throw ngMinErr('cpta', "Can't copy! TypedArray destination cannot be mutated.");
+    }
+    if (source === destination) {
+      throw ngMinErr('cpi', "Can't copy! Source and destination are identical.");
+    }
+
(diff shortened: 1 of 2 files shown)


jbedard:
Anyone ever have a chance to look at this? I found it much clearer then the previous implementation and is much faster...

jbedard:
Rebased...

lgalfaso:
did some of the cleanup talked offline and merged as 33c67ce785cf8be7f0c294b3942ca4a337c5759d

jbedard:
Thanks! I'll try to get a PR that uses map now...

petebacondarwin:
@lgalfaso did this get back ported to 1.4?

lgalfaso:
@petebacondarwin it landed only in 1.5, do you think it should be backported?

petebacondarwin:
I have backported this and @jbedard's other commit: 

* d1293540e13573eb9ea5f90730bb9c9710c345db
* 19fab4a1d79d2445795273f1622344353cf4d104

PR #4282, opened 2013-10-04
feat(orderBy): speed optimization

agentcooper:
Closes #4177.

Issue #11312, opened 2015-03-13
orderBy filter has odd results when nulls are compared to Date objects


hamfastgamgee:
Granted, "odd results" are better than bombing completely, which I think is what used to happen in some cases for orderBy.  When you have nulls in the property that is being compared in an orderBy filter and that property is generally a Date (or anything else where .valueOf evaluates to a number but typeof <value> === "object"), the results are effectively unpredictable.

In compare, typeof null evaluates to "object", and typeof <the Date> also evaluates to "object".  Thus, we fall into the t1 === t2 && t1 === "object" block, which calls objectToString() on both values.  The null ends up getting converted into a string ("null"), and then the Date ends up getting converted into a number by .valueOf.  Back up in compare(), the code goes into the next t1 === t2 block.  The values are not equal, so it tries the less than check.  "null" < any number always evaluates to false, and so the sort order ends up inconsistent (since the null can be either v1 or v2 depending on the call to compare()).  The nulls don't go to either the top or the bottom of the list consistently, regardless of sorting "regular" or "reverse".

I can throw together a plunkr later (too late tonight), but I suspect this is at least enough to get someone thinking. :)

Angular version 1.3.13

hamfastgamgee:
I strongly suspect the same would be true for, say, Number, but I care about the Date situation, and that's less easy to work around without negative consequences. :)

hamfastgamgee:
Also, mind you, I'm not sure if there's a general solution to how nulls *should* sort against all object types or even that whatever gets picked for the default behavior will work for me.  I might have to write a custom comparer anyway.  But at least the default behavior should probably be well-defined.

pkozlowski-opensource:
@hamfastgamgee could you please share a plunker with a reproduce scenario? I think that it is already fixed in 1.3.14 or master but a live reproduce scenario would allow us to confirm this.

hamfastgamgee:
http://embed.plnkr.co/dU2LCUMlqIQ44QgjIdWw/preview

You can see that the nulls sort to the middle of the list in default ascending order, but if you click the column header to sort descending, they end up second through fourth in the list.  As I said, "odd". :)

pkozlowski-opensource:
@hamfastgamgee oh, I see, thnx for the reproduce scenario! Yeh, I agree it is odd, let's do something about it!

pkozlowski-opensource:
[...]

PR #12072, opened 2015-06-10
fix(orderBy): ensure correct ordering with arrays of objects and no predicate

### LINKED ISSUE #11866: orderBy doesn't reverse order if no predicate specified.
According to the documentation: https://docs.angularjs.org/api/ng/filter/orderBy.
> If no property is provided, (e.g. '+') then the array element itself is used to compare where sorting.

This seems to be working fine in 1.3.5 but it seems to have been broken in 1.3.6+.

Probably caused by: https://github.com/angular/angular.js/commit/8bfeddb5d671017f4a21b8b46334ac816710b143

Plnkr with the issue: http://plnkr.co/edit/vaOJzuDdkJUookDVSuDK?p=preview

petebacondarwin:
This fix also considerably refactors how the `orderBy` filter works in internally.

The new algorithm precomputes the predicate values for the array being ordered. This makes the algorith easier to follow but also ensures that this computation is done a maximum of n times, whereas in the previous algorithm it could be greater. The downside is that we must temporarily store this intermediate array.

There could be performance implications, but they are as likely to be beneficial as detrimental. It could probably do with benchmarking.

Closes #11866

### DIFF
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -176,88 +176,114 @@
 orderByFilter.$inject = ['$parse'];
 function orderByFilter($parse) {
   return function(array, sortPredicate, reverseOrder) {
+
     if (!(isArrayLike(array))) return array;
-    sortPredicate = isArray(sortPredicate) ? sortPredicate : [sortPredicate];
+
+    if (!isArray(sortPredicate)) { sortPredicate = [sortPredicate]; }
     if (sortPredicate.length === 0) { sortPredicate = ['+']; }
-    sortPredicate = sortPredicate.map(function(predicate) {
-      var descending = false, get = predicate || identity;
-      if (isString(predicate)) {
-        if ((predicate.charAt(0) == '+' || predicate.charAt(0) == '-')) {
-          descending = predicate.charAt(0) == '-';
-          predicate = predicate.substring(1);
-        }
-        if (predicate === '') {
-          // Effectively no predicate was passed so we compare identity
-          return reverseComparator(compare, descending);
-        }
-        get = $parse(predicate);
-        if (get.constant) {
-          var key = get();
-          return reverseComparator(function(a, b) {
-            return compare(a[key], b[key]);
-          }, descending);
-        }
-      }
-      return reverseComparator(function(a, b) {
-        return compare(get(a),get(b));
-      }, descending);
-    });
-    return slice.call(array).sort(reverseComparator(comparator, reverseOrder));
 
-    function comparator(o1, o2) {
-      for (var i = 0; i < sortPredicate.length; i++) {
-        var comp = sortPredicate[i](o1, o2);
-        if (comp !== 0) return comp;
-      }
-      return 0;
-    }
-    function reverseComparator(comp, descending) {
-      return descending
-          ? function(a, b) {return comp(b,a);}
-          : comp;
+    var predicates = processPredicates(sortPredicate, reverseOrder);
+
+    // The next three lines are a version of a Swartzian Transform idiom from Perl
+    // (sometimes called the Decorate-Sort-Undecorate idiom)
+    // See https://en.wikipedia.org/wiki/Schwartzian_transform
+    var compareValues = Array.prototype.map.call(array, getComparisonObject);
+    compareValues.sort(doComparison);
+    array = compareValues.map(function(item) { return item.value; });
+
+    return array;
+
+    function getComparisonObject(value, index) {
+      return {
+        value: value,
+        predicateValues: predicates.map(function(predicate) {
+          return getPredicateValue(predicate.get(value), index);
+        })
+      };
     }
 
-    function isPrimitive(value) {
-      switch (typeof value) {
-        case 'number': /* falls through */
-        case 'boolean': /* falls through */
-        case 'string':
-          return true;
-        default:
-          return false;
+    function doComparison(v1, v2) {
+      var result = 0;
+      for (var index=0, length = predicates.length; index < length; ++index) {
+        result = compare(v1.predicateValues[index], v2.predicateValues[index]) * predicates[index].descending;
+        if (result) break;
       }
-    }
-
-    function objectToString(value) {
-      if (value === null) return 'null';
-      if (typeof value.valueOf === 'function') {
-        value = value.valueOf();
-        if (isPrimitive(value)) return value;
-      }
-      if (typeof value.toString === 'function') {
-        value = value.toString();
-        if (isPrimitive(value)) return value;
-      }
-      return '';
-    }
-
-    function compare(v1, v2) {
-      var t1 = typeof v1;
-      var t2 = typeof v2;
-      if (t1 === t2 && t1 === "object") {
-        v1 = objectToString(v1);
-        v2 = objectToString(v2);
-      }
-      if (t1 === t2) {
-        if (t1 === "string") {
-           v1 = v1.toLowerCase();
-           v2 = v2.toLowerCase();
-        }
-        if (v1 === v2) return 0;
-        return v1 < v2 ? -1 : 1;
-      } else {
-        return t1 < t2 ? -1 : 1;
-      }
+      return result;
     }
   };
+
+  function processPredicates(sortPredicate, reverseOrder) {
+    reverseOrder = reverseOrder ? -1 : 1;
+    return sortPredicate.map(function(predicate) {
+      var descending = 1, get = identity;
+
+      if (isFunction(predicate)) {
+        get = predicate;
+      } else if (isString(predicate)) {
+        if ((predicate.charAt(0) == '+' || predicate.charAt(0) == '-')) {
+          descending = predicate.charAt(0) == '-' ? -1 : 1;
+          predicate = predicate.substring(1);
+        }
+        if (predicate !== '') {
+          get = $parse(predicate);
+          if (get.constant) {
+            var key = get();
+            get = function(value) { return value[key]; };
+          }
+        }
+      }
+      return { get: get, descending: descending * reverseOrder };
+    });
+  }
+
+  function isPrimitive(value) {
+    switch (typeof value) {
+      case 'number': /* falls through */
+      case 'boolean': /* falls through */
+      case 'string':
+        return true;
+      default:
+        return false;
+    }
+  }
+
+  function objectValue(value, index) {
+    // If `valueOf` is a valid function use that
+    if (typeof value.valueOf === 'function') {
+      value = value.valueOf();
+      if (isPrimitive(value)) return value;
+    }
+    // If `toString` is a valid function and not the one from `Object.prototype` use that
+    if (hasCustomToString(value)) {
+      value = value.toString();
+      if (isPrimitive(value)) return value;
+    }
+    // We have a basic object so we use the position of the object in the collection
+    return index;
+  }
+
+  function getPredicateValue(value, index) {
+    var type = typeof value;
+    if (value === null) {
+      type = 'string';
+      value = 'null';
+    } else if (type === 'string') {
+      value = value.toLowerCase();
+    } else if (type === 'object') {
+      value = objectValue(value, index);
+    }
+    return { value: value, type: type };
+  }
+
+  function compare(v1, v2) {
+    var result = 0;
+    if (v1.type === v2.type) {
+      if (v1.value !== v2.value) {
+        result = v1.value < v2.value ? -1 : 1;
+      }
+    } else {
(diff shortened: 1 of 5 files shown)


petebacondarwin:
@Narretz and wawyed could you take a look?

petebacondarwin:
I guess this should be backported to 1.3.x - but not to 1.2.x since it uses `Array.map` and `Array.reduce`, which are not available in IE8.

petebacondarwin:
Here is a plunker with this build of AngularJS demonstrating the original bug is fixed: http://plnkr.co/edit/WxAW671zu7BePcN4wgBn?p=preview

test/ng/filter/orderBySpec.js
```diff
@@ -11,15 +11,15 @@ describe('Filter: orderBy', function() {
     it('should return sorted array if predicate is not provided', function() {
       expect(orderBy([2, 1, 3])).toEqual([1, 2, 3]);
 
-      expect(orderBy([2, 1, 3], '')).toEqual([1, 2, 3]);
-      expect(orderBy([2, 1, 3], [])).toEqual([1, 2, 3]);
-      expect(orderBy([2, 1, 3], [''])).toEqual([1, 2, 3]);
+      // expect(orderBy([2, 1, 3], '')).toEqual([1, 2, 3]);
+      // expect(orderBy([2, 1, 3], [])).toEqual([1, 2, 3]);
+      // expect(orderBy([2, 1, 3], [''])).toEqual([1, 2, 3]);
```
Narretz: Is this intentionally commented out?

wawyed:
Seems to be working fine when setting reverse true, although it doesn't seem to be working when using the '-' notation instead of reverse true for array with objects.

Pulnker with the issue:

http://plnkr.co/edit/Z3cdF3wMJ7Uzubw2Gma9?p=preview

test/ng/filter/orderBySpec.js
petebacondarwin: Oops. No

petebacondarwin:
Agh! I knew it wasn't going to be so easy.
Now here is the problem. If you are reversing with a predicate then you need a way of telling the algorithm that objects should be ordered by index, rather than identity - but only if they don't have a `toString` or `toValue` or are being mapped by specific predicate.

petebacondarwin:
OK, so I fixed it...

Actually, the reason for the refactoring was that it made this scenario much easier to handle.

But I am not 100% happy that we have to guess at whether the object is a POJO or not, when we decide to use the index rather than the identity for its ordering value.

Also, we now treat `null` as lower order than an object - hence the change to two of the tests. I don't think this should really affect anyone but perhaps it should be marked as a potential breaking change?

petebacondarwin:
Agh! So safari wasn't happy comparing nulls, so I have now fixed that too

petebacondarwin:
http://plnkr.co/edit/BSkUBBUfg0c9TJ9tTK5F?p=preview

wawyed:
For example in this plunkr is this behaviour desired for the third ng-repeat?:

http://plnkr.co/edit/gyw8B10H9tpGdu77g1e7?p=preview

According to documentation if reverse is true it should reverse the order of the array, no matter what type the contents are. It's not that clear what it should do with '-'.

src/ng/filter/orderBy.js
```diff
@@ -176,88 +176,110 @@
 orderByFilter.$inject = ['$parse'];
 function orderByFilter($parse) {
   return function(array, sortPredicate, reverseOrder) {
+
     if (!(isArrayLike(array))) return array;
-    sortPredicate = isArray(sortPredicate) ? sortPredicate : [sortPredicate];
+
+    if (!isArray(sortPredicate)) { sortPredicate = [sortPredicate]; }
     if (sortPredicate.length === 0) { sortPredicate = ['+']; }
```
lgalfaso: Should this be
```js
sortPredicate.push('+');
```

?

src/ng/filter/orderBy.js
```diff
@@ -176,88 +176,110 @@
...
-    sortPredicate = sortPredicate.map(function(predicate) {
-      var descending = false, get = predicate || identity;
-      if (isString(predicate)) {
+
+    var predicates = processPredicates(sortPredicate);
+    var compareValues = Array.prototype.map.call(array, getComparisonObject);
```
lgalfaso: This is usually called *schwartzian transform*, a comment would help people understand what is going on

lgalfaso: and please add Close #4282

src/ng/filter/orderBy.js
```diff
@@ -176,88 +176,110 @@
...
+
+    compareValues.sort(doComparison);
+    array = compareValues.map(function(item) { return item.value; });
+    if (reverseOrder) { array.reverse(); }
```
lgalfaso: Can this be made part of the main comparison?

src/ng/filter/orderBy.js
lgalfaso: Mmm... thinking a little more about this, I think my proposal and the current code are wrong as we are modifying the original `sortPredicate` and we should not do that

src/ng/filter/orderBy.js
```diff
@@ -176,88 +176,110 @@
...
+      return {
+        value: value,
+        predicateValues: predicates.map(function(predicate) {
+          return getPredicateValue(predicate.get(value), index);
+        })
+      };
+    }
+
+    function doComparison(v1, v2) {
+      return predicates.reduce(function(result, predicate, index) {
```
lgalfaso: this may be a little too fancy just to avoid a for-loop

lgalfaso:
> Also, we now treat null as lower order than an object - hence the change to two of the tests. I don't think this should really affect anyone but perhaps it should be marked as a potential breaking change?

Do you think it would make the code any harder to keep the original behavior? I do not mind this going one way or another, but do think this is a small breaking change

hamfastgamgee:
I'm guessing that this "small breaking change" might be accidentally fixing (or at least changing behavior of) #11312?

src/ng/filter/orderBy.js
petebacondarwin: I agree. We should not modify the passed in predicate

src/ng/filter/orderBy.js
petebacondarwin: I am not sure what is fancy about this. It is reducing a collection of comparison predicates down to a final comparison result.

petebacondarwin: The for loop would look like:

```js
function doComparison(v1, v2) {
  var result = 0;
  for(var index=0, length = predicates.length; index<length; ++index) {
    result = result || compare(v1.predicateValues[index], v2.predicateValues[index]) * predicates[index].descending;
  }
  return result;
}
```

caitp: honestly, for people who aren't familiar with `reduce`, this isn't significantly more readable, and is guaranteed to be slower. That and, comparison doesn't really map very well with the "reduce" operation

petebacondarwin: @caitp The `reduce` is guaranteed to be slower?

caitp: yes --- the spec mandates `HasProperty` checks for each item which are slow almost everywhere, and can't be optimized away because the callback can manipulate the array

caitp: Basically for something that runs each digest on a potentially large collection, I'd prefer if it wasn't a reduce operation, both for readability and perf

src/ng/filter/orderBy.js
gkalpak: How is the current implementation modifying the passed in predicate ?

src/ng/filter/orderBy.js
petebacondarwin: OK, I'll change this

petebacondarwin:
I think we can probably move `null` in the sort order. I need to work out exactly where that is though.

src/ng/filter/orderBy.js
petebacondarwin: It could but if you look at the previous implementation it meant complicating the processing of the predicate functions, which makes it significantly more messy.
What is the cost of `Array.reverse()`? It would seem to me that it could be done rather quickly inside the JavaScript engine. Any thoughts @caitp ?

petebacondarwin: Actually, it is simple to move this into the predicates.

src/ng/filter/orderBy.js
petebacondarwin: It's not. But I agree that we should start doing that either :-)

petebacondarwin:
No breaking change no more: the algorithm sorts nulls in the same way as `array.sort()`.
Objects are now dealt with in the following manner:

* has a valid `toValue()` method: sort on the return value of this method
* has a valid custom `toString()` method (i.e. not `Object.prototype.toString()`): sort on the return value of this method
* otherwise sort on index of the object in the array

This means that dates are sorted as expected, but also that arrays are sorted by their `toString` representation, which is to convert each item to a string and then join them together, separated by commas. This gives the same result as calling `array.sort()` on such an array of arrays.

I have noted that the idiom used is a Schwartzian Transform and that it closes #4282.

src/ng/filter/orderBy.js
```diff
@@ -176,88 +176,110 @@
...
+        value: value,
+        predicateValues: predicates.map(function(predicate) {
+          return getPredicateValue(predicate.get(value), index);
+        })
+      };
+    }
+
+    function doComparison(v1, v2) {
+      var result = 0;
+      for (var index=0, length = predicates.length; index < length; ++index) {
```
lgalfaso: for (...; !result && index < length; ...) {

?

src/ng/filter/orderBy.js
```diff
@@ -176,88 +176,110 @@
...
+  function objectValue(value, index) {
+    if (typeof value.valueOf === 'function') {
+      value = value.valueOf();
+      if (isPrimitive(value)) return value;
+    }
+    if (hasCustomToString(value)) {
+      value = value.toString();
+      if (isPrimitive(value)) return value;
     }
+    return index;
```
lgalfaso: trying to think when we end up returning `index` and what does it mean

src/ng/filter/orderBy.js
petebacondarwin: I like

src/ng/filter/orderBy.js
petebacondarwin: I'll add a comment. Is that enough or are you saying that the code might be wrong?

src/ng/filter/orderBy.js
petebacondarwin: But it is a bit obscure. I would prefer to make it more explicit:

```js
    function doComparison(v1, v2) {
      var result = 0;
      for (var index=0, length = predicates.length; index < length; ++index) {
        result = compare(v1.predicateValues[index], v2.predicateValues[index]) * predicates[index].descending;
        if (result) break;
      }
      return result;
    }
```

petebacondarwin:
@lgalfaso I updated the code.

src/ng/filter/orderBy.js
lgalfaso: > In src/ng/filter/orderBy.js
> <https://github.com/angular/angular.js/pull/12072#discussion_r32614045>:
>
> >      }
> > +    return index;
>
> I'll add a comment. Is that enough or are you saying that the code might
> be wrong?
>

Before, we had that if the property was an object (that does not meet a
criteria) then the value was the blank string. Now it returns the position
in the index of the object.

This is, if you sort [{foo: 2}, {foo: 3}, {foo: 4}, {foo: {}}] by "foo",
over and over, then sorting it once will be different than sorting it twice
(assuming sort is stable)

petebacondarwin: Well.... it is true that the sort will change on the second pass but I think it is reasonable to say that sorting mixed types like that is a rather strange case, no?

petebacondarwin: Actually, this is not a problem. First of all the `orderBy` filter does not mutate the original array. Second if the types of things are different then they are ordered by their type first rather than their value.

src/ng/filter/orderBy.js
lgalfaso: Works for me

src/ng/filter/orderBy.js
lgalfaso: I think this is very minor (famous last words), so there is no needs to keep this going on. Lets land this

lgalfaso:
LGTM

PR #12074, opened 2015-06-10
Modified not to be case sensitive Fixes #12073

### LINKED ISSUE #12073: Linky does not work with case-insensitive schemes
According to RFC 3986 (http://tools.ietf.org/html/rfc3986#section-3.1) scheme such as http or mailto are case-insensitive. So links such as http://server/ and HTTP://server/ are valid and equivalent.

However, the linky filter recognizes only links with lowercase schemes as links. While http://server/ will be recognized as a link, HTTP://server will be not.

NOTE: This is a duplicate of #3210

BiohZn:


BiohZn:
I signed it!

petebacondarwin:
Closes #12073

petebacondarwin:
LGTM. Squashing and merging. Thanks @BiohZn

petebacondarwin:
Landed as 8dc09e6dabb84c2c611cdc9e40adfac989648200

PR #12075, opened 2015-06-10
fix(ngModel): form validation and Object.prototype enumerable values

### LINKED ISSUE #12066: Having custom Object.prototype method prevents form validation
I'm running into this issue that if I set a custom Object.prototype method, it stops form validation from updating:

    <!DOCTYPE html>
    <html lang="en">
    	<head>
    		<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
    		<title>Test</title> 
    	</head>
    	<body ng-app>
    		<form name="testForm" novalidate>
    			
    			<input type="text" name="test" ng-model="test" required> Field Valid? {{testForm.test.$valid | json}}
    		</form>
    		Form valid? {{testForm.$valid | json}}
    
    		<script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.4.0/angular.min.js"></script>
    		<script>
    			// remove this line and validation works again
    			Object.prototype.someThing = function() {}
    		</script>
    	</body>
    </html>

lgalfaso:
When adding an Object.prototype enumerable property, this should not be confused
as a form error

Closes #12066

### DIFF
--- a/test/ng/directive/ngModelSpec.js
+++ b/test/ng/directive/ngModelSpec.js
@@ -1091,6 +1091,31 @@ describe('ngModel', function() {
       }));
 
 
+      it('should be possible to extend Object prototype and still be able to do form validation',
+        inject(function($compile, $rootScope) {
+        Object.prototype.someThing = function() {};
+        var element = $compile('<form name="myForm">' +
+                                 '<input type="text" name="username" ng-model="username" minlength="10" required />' +
+                               '</form>')($rootScope);
+        var inputElm = element.find('input');
+
+        var formCtrl = $rootScope.myForm;
+        var usernameCtrl = formCtrl.username;
+
+        $rootScope.$digest();
+        expect(usernameCtrl.$invalid).toBe(true);
+        expect(formCtrl.$invalid).toBe(true);
+
+        usernameCtrl.$setViewValue('valid-username');
+        $rootScope.$digest();
+
+        expect(usernameCtrl.$invalid).toBe(false);
+        expect(formCtrl.$invalid).toBe(false);
+        delete Object.prototype.someThing;
+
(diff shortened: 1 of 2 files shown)


test/ng/directive/ngModelSpec.js
```diff
@@ -1091,6 +1091,31 @@ describe('ngModel', function() {
...
+        expect(formCtrl.$invalid).toBe(true);
+
+        usernameCtrl.$setViewValue('valid-username');
+        $rootScope.$digest();
+
+        expect(usernameCtrl.$invalid).toBe(false);
+        expect(formCtrl.$invalid).toBe(false);
+        delete Object.prototype.someThing;
+
+        dealoc(element);
```
gkalpak: Theoretically [not needed][1].

[1]: https://github.com/lgalfaso/angular.js/blob/issue-12066/test/ng/directive/ngModelSpec.js#L35

lgalfaso: and in practice, generates the error "Error: Found jqCache references that were not deallocated! count: 1"

Keep in mind that this is a new `element` (not the global one)

gkalpak: Right !
Why not reuse the _global_ `element` then ? :)

lgalfaso: The global `element` is already compiled with a content that is not good for this test. This same thing is done for other tests in the same file.

gkalpak: I should have taken a more detailed look :)
<sub>(Sorry for the noise...)</sub>

lgalfaso:
landed as 0934b76b72cec86093414834ac4cb7f0946b651d

PR #12078, opened 2015-06-10
perf($compile): Lazily compile the `transclude` function

### LINKED ISSUE #6072: Directive: transclude in conjuction with terminal
I have tried to understand what's the difference between this:
````
{ 
    translcude: 'element',
    terminal: true 
}
```
And this:
```
{
    translcude: 'element',
    terminal: false
}
```

I made a plunker: http://plnkr.co/edit/yryUHelOMW0gITotKiC7?p=preview
It seems that the contents are pre-compiled on both cases so I though `terminal: true` just get ignored when `transclude` is used.

Then I saw that these directives: **ngIf , ngInclude , ngRepeat**  all use:
`{ trasclude:'element', terminal:true }`.
while other directives like  **ngSwitch , ngSwitchWhen**  only use:
`{ translcude:'element }'` without _a terminal option_.

Now I want a clarification to what are the use cases of these two options in conjunction.

Let say I have a piece of DOM:
```
<div ng-repeat="item in items">
   {{ item }}
</div>
```
Even if `$scope.items = []` the contents get compiled before any clones are linked.

I currently develop a module which provides directives that show/hide contents based on media queries ( inspired by _zurb foundation interchange_), If the media query is not matched then the contents stay out of the DOM ( like `ngIf`). here is the repo: https://github.com/IlanFrumer/angular-match-media

What I suggest is that the behavior of `transclude` and `terminal` together would be to defer compilation and lazy compile it on the first time $transclude is called inside the linking function.
If no clones are needed the contents will never get compiled.

Thanks!

dcherman:
For transcluded directives, the transclude function can be lazily compiled
most of the time since the contents will not be needed until the
`transclude` function was actually invoked.  For example, the `transclude`
function that is passed to `ng-if` or `ng-switch-when` does not need to be
invoked until the condition that it's bound to has been matched.  For
complex trees or switch statements, this can represent significant
performance gains since compilation of branches is deferred, and that
compilation may never actually happen if it isn't needed.

In a large application where I work that recently had some very complex `ngSwitch` cases added, this change significantly reduce the time spent in the compilation stage of the $digest cycle and further reduced the used JS heap size from **123MB** to **56.8MB**.

There are two instances where compilation will not be lazy; when we scan
ahead in the array of directives to be processed and find at least two of
the following:
- A directive that is transcluded and does not allow multiple transclusion
- A directive that has templateUrl and replace: true

In both of those cases, we will need to continue eager compilation in
order to generate the multiple transclusion exception at the correct time.  If anyone has a better idea on how to support this while maintaining backwards compatibility with existing unit tests, I'm all ears. 

Fixes #6072

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1626,6 +1626,37 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     }
 
     /**
+     * A function generator that is used to support both eager and lazy compilation
+     * linking function.
+     * @param eager
+     * @param $compileNodes
+     * @param transcludeFn
+     * @param maxPriority
+     * @param ignoreDirective
+     * @param previousCompileContext
+     * @returns {Function}
+     */
+    function compilationGenerator(eager, $compileNodes, transcludeFn, maxPriority, ignoreDirective, previousCompileContext) {
+        if (eager) {
+            return compile($compileNodes, transcludeFn, maxPriority, ignoreDirective, previousCompileContext);
+        }
+
+        var compiled;
+
+        return function() {
+            if (!compiled) {
+                compiled = compile($compileNodes, transcludeFn, maxPriority, ignoreDirective, previousCompileContext);
+
+                // Null out all of these references in order to make them eligible for garbage collection
+                // since this is a potentially long lived closure
+                $compileNodes = transcludeFn = previousCompileContext = null;
+            }
+
+            return compiled.apply(this, arguments);
+        };
+    }
+
+    /**
      * Once the directives have been collected, their compile functions are executed. This method
      * is responsible for inlining directive templates as well as terminating the application
      * of the directives if the terminal directive has been reached.
@@ -1669,6 +1700,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           replaceDirective = originalReplaceDirective,
           childTranscludeFn = transcludeFn,
           linkFn,
+          didScanForMultipleTransclusion = false,
+          mightHaveMultipleTransclusionError = false,
           directiveValue;
 
       // executes all directives on the current element
@@ -1711,6 +1744,27 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
         directiveName = directive.name;
 
+        // If we encounter a condition that can result in transclusion on the directive,
+        // then scan ahead in the remaining directives for others that may cause a multiple
+        // transclusion error to be thrown during the compilation process.  If a matching directive
+        // is found, then we know that when we encounter a transcluded directive, we need to eagerly
+        // compile the `transclude` function rather than doing it lazily in order to throw
+        // exceptions at the correct time
+        if (!didScanForMultipleTransclusion && ((directive.replace && (directive.templateUrl || directive.template))
+            || (directive.transclude && !directive.$$tlb))) {
+                var candidateDirective;
+
+                for (var scanningIndex = i + 1; candidateDirective = directives[scanningIndex++];) {
+                    if ((candidateDirective.transclude && !candidateDirective.$$tlb)
+                        || (candidateDirective.replace && (candidateDirective.templateUrl || candidateDirective.template ))) {
+                        mightHaveMultipleTransclusionError = true;
+                        break;
+                    }
+                }
+
+                didScanForMultipleTransclusion = true;
+        }
+
         if (!directive.templateUrl && directive.controller) {
           directiveValue = directive.controller;
           controllerDirectives = controllerDirectives || createMap();
@@ -1740,7 +1794,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             compileNode = $compileNode[0];
             replaceWith(jqCollection, sliceArgs($template), compileNode);
 
-            childTranscludeFn = compile($template, transcludeFn, terminalPriority,
+            childTranscludeFn = compilationGenerator(mightHaveMultipleTransclusionError, $template, transcludeFn, terminalPriority,
                                         replaceDirective && replaceDirective.name, {
                                           // Don't pass in:
                                           // - controllerDirectives - otherwise we'll create duplicates controllers
@@ -1754,7 +1808,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           } else {
             $template = jqLite(jqLiteClone(compileNode)).contents();
             $compileNode.empty(); // clear contents
-            childTranscludeFn = compile($template, transcludeFn);
+            childTranscludeFn = compilationGenerator(mightHaveMultipleTransclusionError, $template, transcludeFn);
           }
         }
 
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -6596,6 +6596,27 @@ describe('$compile', function() {
         });
       });
 
+      it('should only allow one element transclusion per element when replace directive is in the mix', function() {
+        module(function() {
+          directive('template', valueFn({
+            template: '<p second></p>',
+            replace: true
+          }));
+          directive('first', valueFn({
+            transclude: 'element',
+            priority: 100
+          }));
+          directive('second', valueFn({
+            transclude: 'element'
+          }));
+        });
+        inject(function($compile) {
+          expect(function() {
+            $compile('<div template first></div>');
+          }).toThrowMinErr('$compile', 'multidir', /Multiple directives \[first, second\] asking for transclusion on: <p .+/);
+        });
+      });
+
 
       it('should support transcluded element on root content', function() {
         var comment;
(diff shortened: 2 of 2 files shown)


lgalfaso:
This PR proposes a tradeoff

Pro
* The initial compilation is faster :)
* Uses less memory

Con
* If there is a directive with `templateUrl` that causes the compilation to continue asynchronously, then there is an extra wait when this needs to get compiled
* Compilation errors on parts that are not displayed will not be detected by just loading the page


The PR still needs a few cleanups here and there and more than a few tests, but that aside, I like the idea proposed.

@petebacondarwin WDYT for 1.5?

src/ng/compile.js
```diff
@@ -1711,6 +1744,27 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
         directiveName = directive.name;
 
+        // If we encounter a condition that can result in transclusion on the directive,
+        // then scan ahead in the remaining directives for others that may cause a multiple
+        // transclusion error to be thrown during the compilation process.  If a matching directive
+        // is found, then we know that when we encounter a transcluded directive, we need to eagerly
+        // compile the `transclude` function rather than doing it lazily in order to throw
+        // exceptions at the correct time
+        if (!didScanForMultipleTransclusion && (directive.templateUrl && directive.replace)
```
jbedard: Should this be `!didScanForMultipleTransclusion && (a || b)`? Right now it is `!didScanForMultipleTransclusion && a || b`...

dcherman: Absolutely, logic fail on my part.

jbedard:
Just to confirm... if you have an `ng-if` that never evaluates to true, it will never compile? Or a switch with 10 cases but only 1 case is ever used, it will never compile the other 9?

dcherman:
So the deferred loading of child templateUrl/async directives is something that I thought about quite a bit and figured would come up in this discussion.  I think I agree that this will probably have to wait for another significant version bump.

I think that is mitigated by two factors:

1.  In many applications ( including my own ), a build step pre-warms the $tempateCache with the content of these templates, thereby avoiding the delayed HTTP request.

2.  In very large applications ( again including some of my own ) where it's not reasonable to inline everything due to file size concerns, adding proper cache headers should help mitigate repeated performance issues since only a single request would be made for the template.

I avoided doing it in this pull request since I am not nearly familiar enough with the compile flow yet, but since the generated link functions are often long lived, I wonder there are references that can be nulled out to make them GC eligible once compilation for the directive is complete.

dcherman:
@jbedard Precisely.  I encountered this exact situation in an application that I support where one of the developers had written a giant ngSwitch statement ( like 20 cases ), and each of the cases had different templates to compile.  By implementing this change, I brought the app startup and interactivity time from 20-30 seconds down to 1-2.

At the moment, I'm having them work around the performance issue by manually using $compile with a real switch statement in another directive, but I feel like this concept would positively impact nearly all applications that use transclusion that isn't immediately called whether it's through ngIf, ngSwitch, custom directives, etc...

fenduru:
@lgalfaso your point regarding templateUrl's is absolutely correct; however, even in this situation this is a balancing act because while you might save some async processing later, you're causing a flood of HTTP requests at load-time for all of the nested templates that may never even be needed.

jbedard:
Maybe it's time for the giant directive loop in `applyDirectivesToNode` to be split up a bit? Could loop over the directives once before the main loop to look for the multiple transclude / template errors?

That would be a change in how error handling works though since it would throw before compiling anything instead of half way through. But that actually sounds better to me anyway...

dcherman:
The only problem with that is that when we encounter the templateUrl + replace case, we can't synchronously know whether or not we will actually encounter the multiple transclusion error.

I can very easily move the loop outside of the for loop, however I don't think I'd be able to maintain the exact same error messages as I did with this PR since `$compileNode` can be modified by directives based on their definition, and that is used to generate the error message ( looks like it pulls the starting tag name ).

The file could probably benefit from some maintenance though for sure.

dlongley:
This is just a note that this PR is related to #9413. Lazy compilation of transcluded content was the original goal that led to discovery of that bug.

There's a test here in the core suite that exercises it: https://github.com/angular/angular.js/blob/master/test/ng/compileSpec.js#L6064-L6077

And a directive here that has been used to improve performance as mentioned above: https://github.com/digitalbazaar/bedrock-angular-lazy-compile/blob/master/lazy-compile-directive.js

A very significant performance improvement was seen in start up time (and memory usage) for applications that include modals (or "sheets") that only need to be shown once activated. Note that the above directive also includes an experimental feature that attempts to cache the same transcluded content only once, further saving memory in certain situations.

Having this sort of behavior on by default in angular core would be great.

lgalfaso:
@dcherman we would like to get this into master and for it to make it into 1.5, but for this to happen it needs some tests and documentation changes. Would it be possible for you to add these to this PR?

dcherman:
@lgalfaso Happy to do so.  Other than the confirming that the new lazy behavior is working, were there any other tests that wanted added related to this?

lgalfaso:
@dcherman 
* Test that check that the lazy behavior works when expected
* That it does not in the cases outlined in the code
In all cases, using `templateUrl` and `template`

dcherman:
Working on this now.  @lgalfaso out of curiosity, what's the timeline for 1.5?  An alternative solution if you feels it's worthwhile could be to add `$compileProvider.lazyCompilation( boolean )` which would default to false ( the current behavior ).  If users wanted the optimization, they could opt into it without having to wait for 1.5.  Once 1.5 hits, you could choose to swap the default to `true` if you wanted.

lgalfaso:
@dcherman I would start by making this into 1.5, and then talk about back-porting this as something optional. The compiler core is very complex, and adding configurable behavior changes in this specific part of the code is not something I am very fond of. Anyhow, if there is enough interest and traction, then it is something that can be discussed.

dcherman:
I've added some tests and am thinking about how to document some things.  You'll notice that some tests for the eager cases weren't added; those code paths are already covered by earlier tests that check for multiple transclusion errors ( the reason eager loading was added in the first place ).

I can add tests if you want, but they'll literally be copied and pasted from those earlier tests with the name of the test changed.

src/ng/compile.js
```diff
@@ -1711,6 +1744,27 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+
+                for (var scanningIndex = i + 1; candidateDirective = directives[scanningIndex++];) {
+                    if ((candidateDirective.transclude && !candidateDirective.$$tlb)
+                        || (candidateDirective.templateUrl && candidateDirective.replace)) {
+                        mightHaveMultipleTransclusionError = true;
+                        break;
+                    }
+                }
+
+                didScanForMultipleTransclusion = true;
```
lgalfaso: I wonder if there is some corner case and we should add at line 1859
`didScanForMultipleTransclusion = false`?

This is, after a directive with a template and `replace: true` is found, then we might need to rescan once again

src/ng/compile.js
```diff
@@ -1711,6 +1744,27 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
         directiveName = directive.name;
 
+        // If we encounter a condition that can result in transclusion on the directive,
+        // then scan ahead in the remaining directives for others that may cause a multiple
+        // transclusion error to be thrown during the compilation process.  If a matching directive
+        // is found, then we know that when we encounter a transcluded directive, we need to eagerly
+        // compile the `transclude` function rather than doing it lazily in order to throw
+        // exceptions at the correct time
+        if (!didScanForMultipleTransclusion && ((directive.templateUrl && directive.replace)
```
lgalfaso: why is `templateUrl` different than `template`?

lgalfaso:
Overall, I like how this looks like. We should take a look at some of the most weird test cases, but overall, it looks fine.

src/ng/compile.js
dcherman: It's not, good catch!  That was actually indicative of a missing unit test - I've added one that tests that the template/replace:true combination throws a multiple transclusion error exactly as the templateUrl case does.  As expected, that test failed with my code, so I've checked that case too.

src/ng/compile.js
dcherman: Rather than doing that, template and replace: true should just be another case that can cause an eager compile if a second one is found.  I'm not overly concerned about that path compiling eagerly since `replace: true` is deprecated

lgalfaso: Ok, sounds good

src/ng/compile.js
```diff
@@ -1626,6 +1626,37 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+     * linking function.
+     * @param eager
+     * @param $compileNodes
+     * @param transcludeFn
+     * @param maxPriority
+     * @param ignoreDirective
+     * @param previousCompileContext
+     * @returns {Function}
+     */
+    function compilationGenerator(eager, $compileNodes, transcludeFn, maxPriority, ignoreDirective, previousCompileContext) {
```
lgalfaso: [Not a blocker to land this]
I am somehow torn on having all the parameters here or if we should use `arguments`. i.e.

```js
function compilationGenerator(eager) {
  var compileArguments = Array.slice(arguments);
  compileArguments.shift();
  [...]
    return compile.apply(null, compileArguments);
  [...]
}
```

It is somehow more code, but it should be more resilient to changes. WDYT?

lgalfaso:
Other that that, LGTM

src/ng/compile.js
dcherman: If someone makes a change that makes the arguments passed invalid, we should have sufficient unit tests in order to catch any resulting bugs.  Since the compile stage is generally considered a hot path, before making this change I'd be curious to see what ( if any ) perf implications there might be from V8 being unable to optimize this function due to the use of `arguments` like that.

We can also consider copying the arguments manually with a for loop which avoids the deopt if you think it really would make it more resilient.

lgalfaso: SGTM

lgalfaso:
Will wait until we branch 1.4.x to land this.

Thanks @dcherman !

lgalfaso:
landed as 652b83eb226131d131a44453520a569202aa4aac

Narretz:
Nice! Let's see what it breaks ;)
Am 19.09.2015 22:35 schrieb "Lucas Mirelmann" <notifications@github.com>:

PR #12083, opened 2015-06-11
fix($location): do not get caught in infinite digest in IE9

### LINKED ISSUE #11439: IE9 infinite $digest bug if loading a base URL without a trailing slash

petebacondarwin:
The problem appears to occur when we have triggered a reload (via replace()) that updates the `reloadLocation` but then before the reload happens we trigger another change to the URL that would not have triggered a reload if we had already reloaded. In this case, the `$browser` is ignoring the `reloadLocation` because it thinks that we will no longer reload.

So to be more explicit:

Navigate to http://my.server.com/Home
Because we are on IE9 this get rewritten (with replace()) to http://my.server.com/#/Home, which would cause a reload on IE9
The SuccessChangeHandler kicks in and then changes the $location.path to /, i.e. the URL gets changed to http://my.server.com/#/`.
Since http://my.server.com/#/Home -> http://my.server.com/#/ would not normally trigger a reload we don't update the reloadLocation, which gets stuck at http://my.server.com/#/Home
We get stuck in an infinite digest as the $location keeps trying to update the URL to http://my.server.com/#/ but failing because the reloadLocation is stuck at http://my.server.com/#/Home
This is why @hamfastgamgee's fix appears to work. When we see that we are in an infinite loop and there is a reloadLocation then we force it to change to the proper URL.

A better solution is simply to realize that we are updating a URL in $browser that is expected to cause a reload by changing the line at https://github.com/angular/angular.js/blob/v1.4.0/src/ng/browser.js#L150:

```
    cacheState();
  } else {
    if (!sameBase) {
      reloadLocation = url;
    }
    if (replace) {
      location.replace(url);
    } else if (!sameBase) {
      location.href = url;
    } else {
      location.hash = getHash(url);
    }
  }
```

to

```
    cacheState();
  } else {
    if (!sameBase || reloadLocation) {
      reloadLocation = url;
    }
    if (replace) {
      location.replace(url);
    } else if (!sameBase) {
      location.href = url;
    } else {
      location.hash = getHash(url);
    }
  }
```

In other words we check not only whether the base has changed but also whether we are in the middle of a reload any way.

Closes #11439
Closes #11935
Closes #11675

### DIFF
--- a/test/ng/locationSpec.js
+++ b/test/ng/locationSpec.js
@@ -669,12 +669,17 @@ describe('$location', function() {
       mockUpBrowser({initialUrl:'http://server/base/home', baseHref:'/base/'});
       inject(
         function($browser, $location, $rootScope, $window) {
+          var handlerCalled = false;
           $rootScope.$on('$locationChangeSuccess', function() {
+            handlerCalled = true;
             if ($location.path() !== '/') {
                 $location.path('/').replace();
             }
           });
+          expect($browser.url()).toEqual('http://server/base/#/home');
           $rootScope.$digest();
+          expect(handlerCalled).toEqual(true);
+          expect($browser.url()).toEqual('http://server/base/#/');
         }
       );
     });
@@ -768,10 +773,10 @@ describe('$location', function() {
 
 
     describe('location watch for HTML5 browsers', function() {
-      beforeEach(initService({html5Mode: true, supportHistory: true}));
-      beforeEach(inject(initBrowser({basePath: '/app/'})));
 
       it('should not infinite $digest when going to base URL without trailing slash when $locationChangeSuccess watcher changes path to /Home', function() {
+        initService({html5Mode: true, supportHistory: true});
+        mockUpBrowser({initialUrl:'http://server/app/', baseHref:'/app/'});
         inject(function($rootScope, $injector, $browser) {
           var $browserUrl = spyOnlyCallsWithArgs($browser, 'url').andCallThrough();
 
@@ -790,6 +795,8 @@ describe('$location', function() {
       });
 
       it('should not infinite $digest when going to base URL without trailing slash when $locationChangeSuccess watcher changes path to /', function() {
+        initService({html5Mode: true, supportHistory: true});
+        mockUpBrowser({initialUrl:'http://server/app/', baseHref:'/app/'});
         inject(function($rootScope, $injector, $browser) {
           var $browserUrl = spyOnlyCallsWithArgs($browser, 'url').andCallThrough();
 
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@hamfastgamgee can you take a look at this alternative?

test/ng/locationSpec.js
```diff
@@ -663,6 +663,22 @@ describe('$location', function() {
...
+      initService({html5Mode:true,supportHistory:false});
+      mockUpBrowser({initialUrl:'http://server/base/home', baseHref:'/base/'});
+      inject(
+        function($browser, $location, $rootScope, $window) {
+          $rootScope.$on('$locationChangeSuccess', function() {
+            if ($location.path() !== '/') {
+                $location.path('/').replace();
+            }
+          });
+          $rootScope.$digest();
```
petebacondarwin: need to check that the change handler actually gets called and that the path is updated correctly

test/ng/locationSpec.js
```diff
@@ -672,6 +688,162 @@ describe('$location', function() {
...
+          expect($location.path()).toEqual('/');
+          expect($browserUrl.calls.length).toEqual(1);
+        });
+      });
+    });
+
+
+    describe('location watch for HTML5 browsers', function() {
+      beforeEach(initService({html5Mode: true, supportHistory: true}));
+      beforeEach(inject(initBrowser({basePath: '/app/'})));
```
petebacondarwin: need to update these tests to use `mockupBrowser()`

test/ng/locationSpec.js
```diff
@@ -1040,6 +1212,49 @@ describe('$location', function() {
...
+        function($browser, $location, $rootScope, $window) {
+          $rootScope.$on('$locationChangeSuccess', function() {
+            if ($location.path() !== '/') {
+                $location.path('/').replace();
+            }
+          });
+          $rootScope.$digest();
+        }
+      );
+    });
```
petebacondarwin: Need to remove this test as it duplicates the above

hamfastgamgee:
Well, leaving aside the failed tests, I can verify that it does in fact fix my case.  I had tried a similar fix previously (never exactly this; it was mostly removing the sameBase check) and had at least three failed tests, but it appears that your fix at least satisfies most browsers on travis. :)

petebacondarwin:
OK, I'll fix up these tests and then if you can give it a once-over, I'll be able to merge it tomorrow

hamfastgamgee:
Sure thing.  Can this please be backported into the 1.3.x branch as well?
My app might be able to go up to 1.4.x, but at present we're not allocating
any time to do the level of regression testing that would make our business
users comfortable.  (I realize it should be a relatively easy upgrade when
we have that time.)

petebacondarwin:
I think this should be able to be back-ported

petebacondarwin:
Updated.

hamfastgamgee:
LGTM.  Still works in my app. :)

petebacondarwin:
If Travis is green by the morning I will merge.

IgorMinar:
lgtm especially with all the tests!

petebacondarwin:
And it's in! Backporting to 1.3...
Then to check whether all those other infinite digest issues are fixed too...

PR #12086, opened 2015-06-11
test($location): ensure mock window can be wrapped by jqLite

caitp:
CC @petebacondarwin


caitp:
The fact that we have this bug in jqLite makes me suspicious that we might do the wrong thing with the real `window` object when there are e.g. iframes on the page

PR #12091, opened 2015-06-11
feat(ngTrim): allow ng-trim to work for elements input[type=radio]

lgalfaso:
Allow elements of type input[type=radio] to specify ng-trim.
Do not trim attributes values by default.

BREAKING CHANGE:
Attribute values are not trimmed

Migration: when depending on attribute values to be trimmed, then manually trim them.

DISCLAIMER: I am not too fond on not trimming attribute values by default, but put the PR here to start the discussion

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1498,7 +1498,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
             attr = nAttrs[j];
             name = attr.name;
-            value = trim(attr.value);
+            value = attr.value;
 
             // support ngAttr attribute binding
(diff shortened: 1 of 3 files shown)


realityking:
Wouldn't it be better to default to true, like it is input[type=text]?

lgalfaso:
Wouldn't it be better to default to true, like it is input[type=text]?

It defaults to true

realityking:
Sorry, I misread your comment.

Narretz:
See also https://github.com/angular/angular.js/pull/5597

The title is a bit misleading, because it's a bigger change

PR #8695, opened 2014-08-20
test(jQuery): test not firing $destroy on jQuery.cleanData with jQuery UI

mgol:
So far it wasn't tested that Angular's logic for skipping it triggering
the $destroy event on jQuery.cleanData in the replaceWith internal function
works correctly when Angular is not the last one to patch the cleanData method
(e.g. if jQuery UI does the patching later). This commits adds the relevant
test.

Ref #8486

cc @IgorMinar

PR #12094, opened 2015-06-12
refactor($compile): remove skipDestroyOnNextJQueryCleanData, fix data leak on some replaced nodes

jbedard:
This removes the `skipDestroyOnNextJQueryCleanData` global while also fixing an edge case where some elements removed by `replaceWith` did not have child elements cleaned and could leak jq data/handlers. It's a bit of an edge case but would be reproduced if jq data ever got onto the child element of a `ng-if` or `ng-repeat` before the transclude `replaceWith` call (or similar situations with `replace:true`). This is captured in the added tests (3 previously failed). Also added a test ensuring the `firstElementToRemove` does not trigger the element $destroy.

`skipDestroyOnNextJQueryCleanData` is removed by doing `jqLite(firstElementToRemove).off('$destroy')` before doing `cleanData(fragment.qSA(*))`.

Doing `cleanData(fragment.qSA(*))` avoids `jqLite(elements[1...n]).remove()` for the multi element case while also fixing the bug mentioned above by also fetching child elements. This makes use of the fragment meaning #12041 that avoided creating it would no longer work.

`JQLite.cleanData` is added so this code doesn't need JQLite vs jQuery specific logic.

For the single node case this adds the `.qSA(*)` and may add `.off('$destroy')` but all the benchmarks I tried showed no effect. The `replaceWith` is just too small/infrequent when doing compile/link/digest.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2634,9 +2634,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         parent.replaceChild(newNode, firstElementToRemove);
       }
 
-      // TODO(perf): what's this document fragment for? is it needed? can we at least reuse it?
+      // Append all the `elementsToRemove` to a fragment. This will...
+      // - remove them from the DOM
+      // - allow them to still be traversed with .nextSibling
+      // - allow a single fragment.qSA to fetch all elements being removed
       var fragment = document.createDocumentFragment();
-      fragment.appendChild(firstElementToRemove);
+      for (i = 0; i < removeCount; i++) {
+        fragment.appendChild(elementsToRemove[i]);
+      }
 
       if (jqLite.hasData(firstElementToRemove)) {
         // Copy over user data (that includes Angular's $scope etc.). Don't copy private
@@ -2644,31 +2649,18 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         // event listeners (which is the main use of private data) wouldn't work anyway.
         jqLite.data(newNode, jqLite.data(firstElementToRemove));
 
-        // Remove data of the replaced element. We cannot just call .remove()
-        // on the element it since that would deallocate scope that is needed
-        // for the new node. Instead, remove the data "manually".
-        if (!jQuery) {
-          delete jqLite.cache[firstElementToRemove[jqLite.expando]];
-        } else {
-          // jQuery 2.x doesn't expose the data storage. Use jQuery.cleanData to clean up after
-          // the replaced element. The cleanData version monkey-patched by Angular would cause
-          // the scope to be trashed and we do need the very same scope to work with the new
-          // element. However, we cannot just cache the non-patched version and use it here as
-          // that would break if another library patches the method after Angular does (one
-          // example is jQuery UI). Instead, set a flag indicating scope destroying should be
-          // skipped this one time.
-          skipDestroyOnNextJQueryCleanData = true;
-          jQuery.cleanData([firstElementToRemove]);
-        }
+        // Remove $destroy event listeners from `firstElementToRemove`
+        jqLite(firstElementToRemove).off('$destroy');
       }
 
-      for (var k = 1, kk = elementsToRemove.length; k < kk; k++) {
-        var element = elementsToRemove[k];
-        jqLite(element).remove(); // must do this way to clean up expando
-        fragment.appendChild(element);
-        delete elementsToRemove[k];
-      }
+      // Cleanup any data/listeners on the elements and children.
+      // This includes invoking the $destroy event on any elements with listeners.
+      jqLite.cleanData(fragment.querySelectorAll('*'));
 
+      // Update the jqLite collection to only contain the `newNode`
+      for (i = 1; i < removeCount; i++) {
+        delete elementsToRemove[i];
+      }
       elementsToRemove[0] = newNode;
       elementsToRemove.length = 1;
     }
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -5844,7 +5844,9 @@ describe('$compile', function() {
 
             testCleanup();
 
-            expect(cleanedCount).toBe(xs.length);
+            // The ng-repeat template is removed/cleaned (the +1)
+            // and each clone of the ng-repeat template is also removed (xs.length)
+            expect(cleanedCount).toBe(xs.length + 1);
 
             // Restore the previous jQuery.cleanData.
             jQuery.cleanData = currentCleanData;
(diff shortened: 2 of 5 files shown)


lgalfaso:
with the patch, is the issue outlined at the comment about *jQuery UI* an issue?

jbedard:
I believe that's covered by [this test](https://github.com/jbedard/angular.js/blob/compile-replacewith-jqremove/test/ng/compileSpec.js#L5609) added in b9389b26ba2cf6aa70372fa32a7b28c62d174bf5. As long as we always reference the current `jqLite.cleanData` it should be fine.

test/ng/compileSpec.js
```diff
@@ -7815,4 +7817,125 @@ describe('$compile', function() {
         expect(element.hasClass('fire')).toBe(true);
       }));
   });
+
+  describe('element replacement', function() {
+    it('should broadcast $destroy only on removed elements, not replaced', function() {
```
jbedard: Does anyone know if this is actually intentional? Why is this done? It seems like the replaced node was never `.remove()`ed so this was always the case. But when multi-element directives were added (e46100f7097d9a8f174bdb9e15d4c6098395c3f2) they used `.remove()` which does fire `$destroy`.

Should `$destroy` actually be fired on the replaced node as well?

jbedard:
@mzgol, this is changing what you originally did for jQuery2. What do you think?

mzgol:
@jbedard I'll take a look on Thursday, I'm on vacation until then.

src/ng/compile.js
```diff
@@ -2507,31 +2512,19 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-          // element. However, we cannot just cache the non-patched version and use it here as
-          // that would break if another library patches the method after Angular does (one
-          // example is jQuery UI). Instead, set a flag indicating scope destroying should be
-          // skipped this one time.
-          skipDestroyOnNextJQueryCleanData = true;
-          jQuery.cleanData([firstElementToRemove]);
-        }
+        // Remove $destroy event listeners from `firstElementToRemove` since it is being
+        // replaced, not destroyed.
+        jqLite(firstElementToRemove).off('$destroy');
```
mzgol: How did I not think about it... :)

mzgol: I wouldn't delete the whole above comment. The following:
```
        // Remove data of the replaced element. We cannot just call .remove()
        // on the element it since that would deallocate scope that is needed
        // for the new node. Instead, remove the data "manually".
```
seems still relevant to the `.off('$destroy')` code.

test/ng/compileSpec.js
mzgol: Did you try to remove the `jqLite(firstElementToRemove).off('$destroy');` line and run unit tests? From what I remember the whole `skipDestroyOnNextJQueryCleanData` was added because the scope is reused on the replaced elements so removing this logic made existing unit tests fail.

test/ng/compileSpec.js
```diff
@@ -7815,4 +7817,125 @@ describe('$compile', function() {
...
+          return {
+            multiElement: true,
+            replace: true,
+            templateUrl: 'template123'
+          };
+        });
+
+        $compileProvider.directive('foo', function() {
+          return {
+            priority: 1, //before the replace directive
```
mzgol: a space after `//`, please. :)

test/ng/compileSpec.js
```diff
@@ -7815,4 +7817,125 @@ describe('$compile', function() {
...
+        expect(linkCalls).toEqual(['2', '3']);
+        expect(destroyCalls).toEqual([]);
+        $rootScope.$apply();
+        expect(linkCalls).toEqual(['2', '3', '1']);
+        expect(destroyCalls).toEqual(['2', '3']);
+      });
+    });
+
+    function getAll($root) {
+      return [$root[0]].concat($root[0].querySelectorAll ? sliceArgs($root[0].querySelectorAll('*')) : []);
```
mzgol: Where can `$root[0].querySelectorAll` be falsy?

test/ng/compileSpec.js
```diff
@@ -7815,4 +7817,125 @@ describe('$compile', function() {
...
+        forEach(preCompiledChildren, function(element, i) {
+          jqLite.data(element, 'foo', 'template#' + i);
+        });
+
+        var linkedElements = $compile(toCompile)($rootScope);
+        $rootScope.$apply();
+        linkedElements.remove();
+
+        forEach(preCompiledChildren, function(element, i) {
+          expect(jqLite.hasData(element)).toBe(false, "template#" + i);
```
mzgol: What does `toBe` with two parameters do? I haven't seen this signature.

test/ng/compileSpec.js
```diff
@@ -7815,4 +7817,125 @@ describe('$compile', function() {
...
+      });
+    }
+    it('should clean data of element-transcluded link-cloned elements', function() {
+      testCompileLinkDataCleanup('<div><div ng-repeat-start="i in [1,2]"><span></span></div><div ng-repeat-end></div></div>');
+    });
+    it('should clean data of element-transcluded elements', function() {
+      testCompileLinkDataCleanup('<div ng-if-start="false"><span><span/></div><span></span><div ng-if-end><span></span></div>');
+    });
+
+    function testReplaceElementCleanup(replace, transclude, asyncTemplate) {
```
mzgol: Could you change this to an object with properties? The following:
```js
testReplaceElementCleanup({replace: false, transclude: false, asyncTemplate: false});
```
would read *much* better than:
```js
testReplaceElementCleanup(false, false, false);
```

mzgol:
A few questions/remarks, overall I like the direction of this PR. :+1:

src/ng/compile.js
jbedard: I'm not sure if that comment is 100% correct. `.remove()` doesn't "deallocate [the] scope" it just fires a DOM $destroy event (not the scope $destroy event).  It doesn't actually touch the scope...

test/ng/compileSpec.js
jbedard: The second param becomes part of the error message if the `toBe` fails.

test/ng/compileSpec.js
jbedard: Removing `jqLite(firstElementToRemove).off('$destroy');` doesn't cause anything to fail. Even this test designed to fail in that case isn't failing!  :worried:

I thought of another case though which I think might show that not invoking the DOM $destory is (sometimes) wrong...

```html
<div ng-form="myForm" lower-priority-replace-directive>
```

In this case the `ng-form` directive runs first which adds itself to the parent form then does `formElement.on('$destroy', function() { parentFormCtrl.$removeControl(controller); ... })`. Then the other lower priority directive replaces that element with something brand new, that replaced element is not removed from the ng-form because no $destroy was fired. The replaced ng-form controller left behind a) is leaked and b) might cause form validation issues?

However if that low priority directive was doing transclude instead of replace then maybe not invoking the DOM $destory is valid? Meaning `replaceWith` would need an `invokeDestroy` argument...

test/ng/compileSpec.js
jbedard: For comment nodes in the transclude case...

test/ng/compileSpec.js
mzgol: Interesting; just removing the `skipDestroyOnNextJQueryCleanData` logic from my commit definitely did fail unit tests.

This might be related, I totally forgot about this WIP PR: #8695

jbedard: Was that PR ready or was there more to look into still?

jbedard: I'm tempted to just drop this test for now. The cases when `$destroy` should/should-not fire are too unclear and I think the current implementation is wrong (or at least super weird) so I'm not sure if I want to enforce that with a test. I think all the other tests are good though (ensuring all elements get passed to `cleanData`).

petebacondarwin:
Moving to 1.5 - @mzgol & @jbedard can we try to combine this with #8695?

mzgol:
@petebacondarwin Sure, I'll have a look. It seems all my comments have been addressed so I'll take the last look and it'll most likely be ready to land.

PR #12101, opened 2015-06-12
fix($parse): set null reference properties to `undefined`

### LINKED ISSUE #12099: Logical operators return incorrect results when used on forgiving expressions.
**Version:** 1.4.0-rc.2

Given the following statements:

- `scope.$eval("true && does.not.exist");`
- `scope.$eval("true && !does.not.exist");`

Both return `true`, as long as `does.not` is not an object on the current scope.

However `scope.$eval("true && doesnotexist");` and `scope.$eval("true && undefined");` both return `false` as expected as long as `doesnotexist` is not a truthy value on the current scope.

This makes me think it has something to do with using `scope.$parse` as it only occurs when attempting to access nested properties of undefined values on the scope which `scope.$parse` suppresses errors for. 

For this reason any directive such as `ng-if` and `ng-switch`. For examples see: http://jsbin.com/pesatiloku/1/edit?html,js,output

lgalfaso:
When there is an expression of the form
- true && a.b.c
- true && a()
- true && a()()
- false || a.b.c
- false || a()
- false || a()()

where `a == null`

Closes #12099

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -982,10 +982,10 @@ ASTCompiler.prototype = {
               nameId.name = ast.property.name;
             }
           }
-          recursionFn(intoId);
         }, function() {
           self.assign(intoId, 'undefined');
         });
+        recursionFn(intoId);
       }, !!create);
       break;
     case AST.CallExpression:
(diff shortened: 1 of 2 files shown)


lgalfaso:
landed as c6b48e13db1272031c10da1f91bf97a19ba923fe

PR #12106, opened 2015-06-13
fix(jqLite): Firefox validation trigger

### LINKED ISSUE #12102: Bug: Firefox: Input fields with 'required' get red shadow/border before submitting
Firefox 38.0.5
AngularJS 1.4.0

`input` fields get the red invalid border/shadow on pageload (i.e. before the form is submitted), if all three conditions are met:
* `ng-app` defined
* `ng-model` set to any name
* `required` attribute present

Here's a fiddle: https://jsfiddle.net/5yh58orm/11/

In Chrome 43.0 and Internet Explorer 11 there is (as expected) no border.

lgalfaso:
Do not trigger Firefox validation on form initialization.
- Do not set a value to an <input> field if the field already has the same value

Closes #12102

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1125,7 +1125,11 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
   element.on('change', listener);
 
   ctrl.$render = function() {
-    element.val(ctrl.$isEmpty(ctrl.$viewValue) ? '' : ctrl.$viewValue);
+    // Workaround for Firefox validation #12102.
+    var value = ctrl.$isEmpty(ctrl.$viewValue) ? '' : ctrl.$viewValue;
+    if (element.val() !== value) {
+      element.val(value);
+    }
   };
 }
 


src/jqLite.js
```diff
@@ -671,7 +671,10 @@ forEach({
       }
       return element.value;
     }
-    element.value = value;
+    // Workaround for Firefox validation #12102.
+    if (element.value != value) {
```
gkalpak: Is the `!=` intentional ? It can lead to bugs in corner cases.
E.g. when `value === ''` and one calls `elem.val(0)`, they would expect `elem.val() === '0'`, but since `0 == ''`, nothing will happen (I think).

gkalpak:
If we decide that we want this change, would it be possible to make the change in a "higher" level (i.e. somewhere in `ngModel`), so that it will work the same as jQuery (or would it be too messy) ?

src/jqLite.js
lgalfaso: `element.value` is always a string. If the value we are going to set it to will be transformed to the same string, then do not copy the value

gkalpak: Yes,  but `value` can be anything the user passes (e.g. `0`).
If `element.value === ''` and `value === 0`, I would expect the element's value to be set to `'0'`, but because `'' != 0` evaluates to false, the element's value will remain `''`, right ?

gkalpak: The rules for transforming to string are not the same as the rules for `!=` comparison.
Maybe you should compare `element.value` to `String(value)` or something ?

gkalpak: Hm...although that is not accurate for `value === null`, because `String(null) === 'null'`, but `element.value = null ==> element.value === ''`, so that need to be taken into account as well.

lgalfaso:
It is possible to do this here https://github.com/angular/angular.js/blob/master/src/ng/directive/input.js#L1128

src/jqLite.js
lgalfaso: Maybe we should just be on the safe side and do `!==`

gkalpak: But, does it solve the initial Firefox issue :) ?

lgalfaso: It solves it as long as the initial value of `ngModel` is not the same as the `input[value]` (in most cases we expect that `input[value]` to be an empty string, so yes, just doing `!==` should be fine (famous last words)

gkalpak: Indeed, since https://github.com/angular/angular.js/blob/master/src/ng/directive/input.js#L1128 takes care of the _empty_ values (null, undefined), it should be fine :)
So, using `!==` sounds like the simplest (aka best) way out...

lgalfaso:
landed as e7423168fbf439a8798fdbbffb57955c272c2d74

PR #12112, opened 2015-06-14
feat(http): Add 'useLegacyMethods' to $httpProvider

realityking:
For now it defaults to true. This will open up a path to remove the short hand methods in the long run.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -8,6 +8,7 @@ var JSON_ENDS = {
   '{': /}$/
 };
 var JSON_PROTECTION_PREFIX = /^\)\]\}',?\n/;
+var $httpMinErr = minErr('$http');
 
 function serializeValue(v) {
   if (isObject(v)) {
@@ -330,6 +331,29 @@ function $HttpProvider() {
     return useApplyAsync;
   };
 
+  var useHttpPromise = true;
+  /**
+   * @ngdoc method
+   * @name $httpProvider#useLegacyPromiseExtensions
+   *
+   * Configure $http service to return promises without the shorthand methods `success` and `error`. It should
+   * be used to make sure that applications work without these methods.
+   *
+   * Defaults to false. If no value is specified, returns the current configured value.
+   *
+   * @param {boolean=} value If true, $http will return a promise without the `success` and `error methods.
+   *
+   * @returns {boolean|Object} If a value is specified, returns the $httpProvider for chaining.
+   *    otherwise, returns the current configured value.
+   **/
+  this.useLegacyPromiseExtensions = function(value) {
+    if (isDefined(value)) {
+      useHttpPromise = !!value;
+      return this;
+    }
+    return useHttpPromise;
+  };
+
   /**
    * @ngdoc property
    * @name $httpProvider#interceptors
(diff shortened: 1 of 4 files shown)


petebacondarwin:
I like the idea behind this. `success()` and `error()` have always worried me...

petebacondarwin:
A few things

* The documentation for it needs to be beefed up
* a deprecation notice needs to go in the commit message
* if the legacy methods have been turned off then they should be changed to throw a helpful error rather than simply missing.

realityking:
* I'll get to the documentation. Just wanted to get some feedback first ;)
 * Is there a special format for deprecation notices?

petebacondarwin:
Perhaps something like this would work: https://github.com/angular/angular.js/commit/38fbe3ee8370fc449b82d80df07b5c2ed2cd5fbe

petebacondarwin:
This would close #10508

realityking:
Sorry, it took me way too long to get back to this.

I've changed the documentation to to mainly cover `then` and added a deprecation notice to both the commit and the documentation. I've also added the error messaged.

Small aside: Once this is merged, it would be great if this could be uplifted to 1.4, as this would allow switching the default to off in 1.5.

src/ng/http.js
```diff
@@ -954,23 +975,34 @@ function $HttpProvider() {
...
-      };
+          promise.then(null, function(response) {
+            fn(response.data, response.status, response.headers, config);
+          });
+          return promise;
+        };
+      }
+      else {
+        promise.success = function() {
+          throw namespaceMinErr('nosuccess', 'The method `success` on the $http result has been disabled.');
```
petebacondarwin: These should be `$httpMinErr` not `namespaceMinErr`.
Also we must create error docs for these errors in https://github.com/angular/angular.js/tree/master/docs/content/error/%24http

test/ng/httpSpec.js
```diff
@@ -1975,6 +1975,25 @@ describe('$http with $applyAsync', function() {
...
+
+  it('should not have the success and error methods if useLegacyMethods is false', function() {
+    $httpBackend.expect('GET', '/url').respond('');
+    var promise = $http({url: '/url'});
+    expect('success' in promise).toBe(false);
+    expect('error' in promise).toBe(false);
+  });
+});
+
 describe('$http param serializers', function() {
```
petebacondarwin: We also need a test of throwing the error if the legacy methods are disabled.

petebacondarwin:
@realityking - thanks for improving this. Just a couple of issues with the minerr stuff and then we are good to go.

petebacondarwin:
If we default to false then this is a breaking change and needs to go into 1.5

realityking:
I've fixed minErr and added the tests. Is there a script to generate the scaffolding for the error docs or is that all by hand?

As for 1.5, my I idea was to ship this with default true in 1.4.x and change the default to false in 1.5. That would allow people to fix their libraries and apps ahead of the release. Also, we'd get rid of the legacy methods being front and center in the docs.

petebacondarwin:
That is better.

For the error doc, you just make a copy of a file like https://github.com/angular/angular.js/blob/master/docs/content/error/%24http/badreq.ngdoc and doctor it to your needs. The rest is magic.

One last comment about the provider method, I feel that `useLegacyMethods` is a bit vague. Perhaps `useLegacyPromiseExtensions` or something?

realityking:
I think I've addressed everything.

petebacondarwin:
I will land this week.

PR #12113, opened 2015-06-14
Check for for strict equality.

mitogh:


mitogh:
I signed it!

petebacondarwin:
This doesn't actually make any difference since `typeof` always returns a string and we are comparing against a static string, so there will be no type coercion attempted.

PR #12114, opened 2015-06-15
refactor(animate): add the new operator to injected $q

pocesar:
When replacing $q with Bluebird, it complains `this is undefined`. Missing the new keyword

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -54,7 +54,7 @@ function prepareAnimateOptions(options) {
 }
 
 var $$CoreAnimateRunnerProvider = function() {
-  this.$get = ['$q', '$$rAF', function($q, $$rAF) {
+  this.$get = ['$q', '$$rAF', function(Q, $$rAF) {
     function AnimateRunner() {}
     AnimateRunner.all = noop;
     AnimateRunner.chain = noop;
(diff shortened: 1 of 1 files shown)


pocesar:
I signed it!

src/ng/animate.js
```diff
@@ -65,7 +65,7 @@ var $$CoreAnimateRunnerProvider = function() {
       pause: noop,
       complete: noop,
       then: function(pass, fail) {
-        return $q(function(resolve) {
+        return new Q(function(resolve) {
```
Narretz: It's not `Q`, it's `$q`.

pocesar: jshint complains "A constructor name should start with an uppercase letter"

lgalfaso:
I think this is a bad idea. If someone wants to replace `$q` with Bluebird, then when the code that intercepts and replaces `$q` should take care of this

pocesar:
Bluebird doesn't allow `Bluebird(function(resolve, reject) {})`, needs to use new on it; returning from the constructor is bad practice anyway

lgalfaso:
Let me reiterate what I just wrote

> If someone wants to replace $q with Bluebird, then **the code that intercepts and replaces $q should take care of this**

Not Angular, not Bluebird, but the code that replaces `$q`

lgalfaso:
No feedback in a month, closing

Issue #11115, opened 2015-02-20
ngVisible


VitalyTomilov:
Guys, please add ngVisible to 1.3. This is probably the most generic directive still not implemented, even though it is needed in just about every project. And it is 100% core functionality.

Something along these lines, I believe:
````
.directive('ngVisible', function () {
    return function (scope, element, attr) {
        scope.$watch(attr.ngVisible, function (visible) {
            element.css('visibility', visible ? 'visible' : 'hidden');
        });
    };
})
````

Narretz:
Compelling arguments, although personaly, I hardly ever use`visible: hidden`.

gkalpak:
I can't recall ever using `visibility: <anything>`...

VitalyTomilov:
CSS visibility is the key property for maintaining reserved space for invisible objects. There is no substitute in Angular for that, or at least one that would work as reliably as the basic CSS.

From my code example:
````
<p ng-visible="initialized">
       grid summary clause
</p>
<div>
    fixed-size grid
</div>
````
If I do not use ng-visible, then my data grid will jump after initialization, which looks bad. I want to reserve space in the simplest possible way, which is how I do it in my example, coherent with the standard CSS approach.

I do not want to do an Angular work-around for such a basic CSS feature, so I always end up extending my code with ng-visible as shown in my first post.

meanJim:
Front end developers use `visibility` very frequently. I totally agree with @VitalyTomilov here, and would add that designers and interaction designers **don't** want to see their interfaces jump around all over the place. Visibility helps claim visual space for a DOM element to prevent that jumping.

Have a look here at a mockup of **1** of **~40** modals we built with varying heights:

![screen shot 2015-02-20 at 9 51 22 am](https://cloud.githubusercontent.com/assets/310223/6306175/0dbf4450-b8e6-11e4-8ba8-cfff2debe91a.png)

* **The Goal:** Have text appear after some delay inside of this modal without using a fixed height. A fixed height is unacceptable.
* Using `display:none` -> `display:block` forces the modal to adjust to the contents of the modal container element. Will cause a jump. Can make the jump look nice if I want.
* Using `visibility: hidden` -> `visibility: visible` we avoid this problem all together.

Also some CSS transitions with `opacity` declared as `transition: opacity 300ms ease;` do not work with the use of `display: none;` and `display: block;` anyways. I am aware of `ngAnimate` and other work arounds, but having something we always extend built in would be nice.

[...]

PR #12138, opened 2015-06-16
feat(ng): add directive for visible and invisible elements

jacobscarter:
when you need to preserve a hidden elements dom space ngVisible and ngInvisible will allow you to do so


jacobscarter:
The `ngInvisibleDirective` undefined variable I will figure out, in regards to the error for `ngNonBindableDirective` being undefined, I did not create that directive or modify it, I don't know why that is throwing an error but ill look into it.

petebacondarwin:
@jacobscarter - thanks for putting this PR together. I am afraid that we have had long debates about this directive in the past in #11115. We have chosen not to include it in the core. See https://github.com/angular/angular.js/issues/11115#issuecomment-76154055

caitp:
You could totally publish it as a third party module, though, or maybe get it upstream into one of angular-ui's "utility" packages (I can't recall the names OTOH)

jacobscarter:
@caitp @petebacondarwin thanks for the replies, it was a cool learning experience and thanks for the advice.

PR #12155, opened 2015-06-17
fix(ngMock): support `before` test framework hook.

### LINKED ISSUE #12154: ngMock doesn’t support `inject` from `before` hooks.
Both Jasmine 2.1+ and Mocha support `before` hooks that run setup code once before all `beforeEach` hooks and test cases are run. `angular-mocks.js` does not use these hooks to configure `currentSpec` (the way, for example, it uses `beforeEach` to do so). When `inject` is called from a `before` hook, it returns without invoking `workFn` and therefore does not invoke the specified function block, configure the injector, etc.

We would like to use `before` hooks with ngMock to avoid repeating common, state-agnostic, possibly expensive test fixture setup work.

mikol:
Both Jasmine 2.1+ and Mocha support `before` hooks that run setup
code once before all `beforeEach` hooks and test cases are run.
Initialize `annotatedFunctions` and `currentSpec` for `before`,
`beforeAll`, and `suiteSetup` test framework synonyms just as we do
for `beforeEach` and `setup` so that `before` hooks can be used with
ngMock.

Closes #12154

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2197,11 +2197,26 @@ if (window.jasmine || window.mocha) {
     return angular.mock.$$annotate.apply(this, arguments);
   };
 
-
-  (window.beforeEach || window.setup)(function() {
+  var doSetUp = function doSetUp() {
     annotatedFunctions = [];
     currentSpec = this;
-  });
+  };
+
+  // Hook `before` to provide the same module and injector configuration as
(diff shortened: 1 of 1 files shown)


mikol:
This won’t work and my local test case was too limited to uncover that. Because it is defined on the outermost test scope, `before` will only be called once, period. No matter how many test suites or spec files are in play. `beforeEach` works for pretty much the same reason – it’s set up once in the outermost scope so gets applied before each and every test case.

PR #12156, opened 2015-06-18
feat($rootScope): Add $once method

noam3127:
This listens on events just like $on, but will deregister
the listener automatically upon the first call.

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -1109,19 +1109,7 @@ function $RootScopeProvider() {
        * @returns {function()} Returns a deregistration function for this listener.
        */
       $on: function(name, listener) {
-        var namedListeners = this.$$listeners[name];
-        if (!namedListeners) {
-          this.$$listeners[name] = namedListeners = [];
-        }
-        namedListeners.push(listener);
-
-        var current = this;
-        do {
-          if (!current.$$listenerCount[name]) {
-            current.$$listenerCount[name] = 0;
-          }
-          current.$$listenerCount[name]++;
-        } while ((current = current.$parent));
+        registerListener(this, name, listener);
 
         var self = this;
         return function() {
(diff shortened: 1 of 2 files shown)


noam3127:
I signed it!

src/ng/rootScope.js
```diff
@@ -1125,14 +1125,58 @@ function $RootScopeProvider() {
...
+       *   - `stopPropagation` - `{function=}`: calling `stopPropagation` function will cancel
+       *     further event propagation (available only for events that were `$emit`-ed).
+       *   - `preventDefault` - `{function}`: calling `preventDefault` sets `defaultPrevented` flag
+       *     to true.
+       *   - `defaultPrevented` - `{boolean}`: true if `preventDefault` was called.
+       *
+       * @param {string} name Event name to listen on.
+       * @param {function(event, ...args)} listener Function to call when the event is emitted.
+       */
+      $once: function(name, listener) {
```
Narretz: To keep it dry, `$once` should share most of its code with `$on`.

Narretz:
It's essentially sugar, so I am not a huge fan of adding it. Let's see what others say.

lgalfaso:
I think this can be added as a third party module. With this in mind, I agree with @Narretz that it should not be added

src/ng/rootScope.js
noam3127: Updated the code to make it dry, @Narretz

src/ng/rootScope.js
```diff
@@ -1109,30 +1109,50 @@ function $RootScopeProvider() {
...
+        var self = this;
+        var modifiedListener = function() {
+          deregisterListener(self, name, modifiedListener);
+          listener.apply(self, arguments);
```
lgalfaso: why use `self` when calling the listener?

src/ng/rootScope.js
lgalfaso: I think that what @Narretz is trying to say is that `$once` should be implemented as

```js
// Untested code
$once: function(name listener) {
  var delegate = this.$on(name, function() {
    delegate();
    listener.apply(null, arguments);
  });
},
```

and leave `$on` unmodified.

lgalfaso:
I still do not like the idea of adding the `$once` method. It should be possible to add this method thru a third party module by modifying the prototype.

Narretz:
Ok, let's close this. This is good for a 3rd party module.

PR #9319, opened 2014-09-28
feat ($http): Add custom createXhr factory to config object

sjurba:
To be able to construct non-standard or customised XHR objects a factory an optional factory function can be added to the config object. This is required for use of the mozAnon and mozSystem options for FirefoxOS.
See https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest#XMLHttpRequest() for details.

This is an alternative solution to the one suggested in PR  #7903. Closes #2318

PR #12159, opened 2015-06-18
feat($http): add $xhrFactory service to enable creation of custom xhr objects

### LINKED ISSUE #2318: mozSystem parameter for window.XMLHttpRequest on FirefoxOS
To be allowed to do Cross-Domain requests in FirefoxOS you need the permission, and window.XMLHttpRequest has to be passed a constructor parameter like this:

    new window.XMLHttpRequest({mozSystem: true}); 

See https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest#XMLHttpRequest()

I currently patch Angular in the creatHttpBackend function to get this working, would be nice to not have to do that.

sjurba:
Extract the createXhr function into a service called $xhrFactory which can easily be decorated to be able to construct non-standard or customised XMLHttpRequest objects. This is required for use of the mozAnon and mozSystem options for FirefoxOS.
See https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest#XMLHttpRequest() for details.

Alternate solution to PR #9319
Closes #2318

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12159)


src/ng/httpBackend.js
```diff
@@ -1,7 +1,33 @@
...
-function createXhr() {
-    return new window.XMLHttpRequest();
+/**
+ * @ngdoc service
+ * @name $xhrFactory
+ *
+ * @description
+ * Factory function used to create XMLHttpRequest objects.
+ *
+ * Decorate this service to create your own custom XMLHttpRequest objects.
```
pkozlowski-opensource: I would say: "Replace or decorate ..."

src/ng/httpBackend.js
```diff
@@ -1,7 +1,33 @@
...
+ *
+ * ```
+ * .config(function($provide) {
+ *   $provide.decorator('$xhrFactory', function() {
```
pkozlowski-opensource: This example is not ideal as here we are replacing the entire implementation really. I would simply show how to register a factory with the same (`$xhrFactory`, that is) name.

pkozlowski-opensource:
I've left a couple of comments regarding decorators usage, but otherwise it LGTM after taking a quick look.

sjurba:
It's funny, my first thought was to do exactly like you suggested. Then for some reason I decided to only go for the decorator version. Anyways, I've updated the text and the example.

src/ng/httpBackend.js
```diff
@@ -46,7 +72,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
       });
     } else {
 
-      var xhr = createXhr();
+      var xhr = createXhr(method, url);
```
gkalpak: Just curious: Why are we passing the `method` and `URL` ?

pkozlowski-opensource: We could pass more (essentially everything we've got here :-)). We can't really predict what people might need... The other approach would be to expose nothing by default and expand the API as soon as we've got more use cases.

sjurba: Not for us, but for the benefit of end users who will override the service. Let's say you want to use a custom object for certain urls only..

sjurba: I think method and url is a good middle ground to start with.

Narretz:
Wouldn't it make more sense for this to use the same mechanism as the httpInterceptors and the httpParamSerializers?
With this PR, users are limited to having one custom xhrObject per app, which is used for all requests. If I remember correctly, we have at least two use cases:

- FirefoxOS needs an object passed to the constructor
- upload / download progress listeners

Having only one implementation per app might not be the best idea.

sjurba:
@Narretz you will have one xhrObject pr request. The xhrFactory will be called for each request and should return a unique xhrObject every time. It will also be called with the method and the url as parameters so you could customise it to only add certain features for certain urls.

But still, you can't have a separate implementation of the factory for each request, so this implementation only solves the Firefox issue, it doesn't seem like a good solution to the progress listener case.

pkozlowski-opensource:
IMO we shouldn't try to fix progress events here, it is a separate issue. And IMO we shouldn't try to have a factory per-request - this is what we've discussed during a meeting: adding per-request handling will add one more param to the $http's object and this is the very issue we wanted to avoid

sjurba:
Any update on the status of this PR? Anything else I need to do?

pkozlowski-opensource:
This one LGTM, I'm +1 for merging this as-is. @petebacondarwin @caitp @Narretz - any objections?

src/ng/httpBackend.js
```diff
@@ -1,7 +1,32 @@
...
-function createXhr() {
-    return new window.XMLHttpRequest();
+/**
+ * @ngdoc service
+ * @name $xhrFactory
+ *
+ * @description
+ * Factory function used to create XMLHttpRequest objects.
+ *
+ * Replace or decorate this service to create your own custom XMLHttpRequest objects.
```
Narretz: This affects $http and $resource requests alike, right? (since resource uses http). We should probably note this here.

Narretz:
Still not a big fan of this solution, since it's at odds with the way we do other things for $http. @petebacondarwin to the rescue!

petebacondarwin:
@Narretz do you have a better suggestion for how we could deal with this?

Narretz:
@petebacondarwin I was thinking more about how the paramSerializers are implemented. This would bea) more consistent and b) more flexible. But since I don't know if this per-request flexibility is actually needed, let's just merge it once with switch to 1.5 and then see if somebody needs the flexibility.

PR #12162, opened 2015-06-18
added simple example of how to use $timeout

rncrtr:
[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12162)


rncrtr:
I signed it!

Narretz:
I don't think something like $timeout that also exists in vanilla JS needs an example, and if so a runnable example would be better. Closing.

PR #12166, opened 2015-06-19
WIP angular.component implementation [ci skip]

caitp:
Needs tests and experimentation, it would be nice to make this work well
with TypeScript decorators for ease of use.

The whole "controller instantiated after templateUrl is loaded" thing seems
not-ideal, it would be cool if a templateUrl function could be a method of
the component controller.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12166)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -814,6 +814,52 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
   /**
    * @ngdoc method
+   * @name $compileProvider#component
+   * @kind function
+   *
+   * @description
+   * Register a new component with the compiler.
+   *
+   * A component is a shorthand for a directive with a controller, template, and a
+   * new scope.
+   *
+   * @returns {ng.$compileProvider} Self for chaining.
+   */
+  this.component = registerComponent;
+  function registerComponent(name, bindings, controller) {
+    if (isString(name)) {
+      assertValidDirectiveName(name);
+      if (arguments.length < 3) {
+        controller = bindings;
+        bindings = controller.bindings || null;
+      }
+      return registerDirective(name, valueFn({
+        controller: controller,
+        controllerAs: controller.controllerAs || '$' + name,
+        scope: controller.isolate ? {} : true,
+        bindToController: bindings,
+        template: controller.template || false,
+        templateUrl: controller.templateUrl || false
+      }, true));
+    } else {
+      forEach(name, function(value, name) {
+        var controller = null;
+        var bindings = null;
+        if (isArray(value) || isFunction(value)) {
+          controller = value;
+          bindings = value.bindings || null;
+        } else if (isObject(value)) {
+          controller = value.controller || null;
+          bindings = value.bindings || null;
+        }
+        return registerComponent(name, bindings, controller);
+      });
+    }
+    return this;
+  }
+
+  /**
+   * @ngdoc method
    * @name $compileProvider#directive
    * @kind function
    *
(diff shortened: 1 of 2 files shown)


src/ng/compile.js
```diff
@@ -814,6 +814,52 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+   *
+   * A component is a shorthand for a directive with a controller, template, and a
+   * new scope.
+   *
+   * @returns {ng.$compileProvider} Self for chaining.
+   */
+  this.component = registerComponent;
+  function registerComponent(name, bindings, controller) {
+    if (isString(name)) {
+      assertValidDirectiveName(name);
```
gkalpak: Isn't this going to be called by `registerDirective()` anyway ?

src/ng/compile.js
```diff
@@ -814,6 +814,52 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+      if (arguments.length < 3) {
+        controller = bindings;
+        bindings = controller.bindings || null;
+      }
+      return registerDirective(name, valueFn({
+        controller: controller,
+        controllerAs: controller.controllerAs || '$' + name,
```
gkalpak: `$` was an unexpected default for me (fwiw).

gkalpak:
By having the controller by some sort of configuration object, we are restricting the use of a string in certain cases (which might not be a good practice).

I would probably prefer an optional config object as a 4th parameter.

src/ng/compile.js
ewinslow: IMO this should be `$ctrl` or something so that it's easy to do a global find/replace when switching to Angular 2. As-is, you'd have to find/replace each directive individually.

DavidSouther: Agreed - when I created a similar approach, we'd just done a bunch of work with React. We called it `state`.

src/ng/compile.js
DavidSouther: I'd rather see the error thrown in the registration method I call, rather than the underlying implementation.

src/ng/compile.js
```diff
@@ -827,7 +873,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    *    {@link guide/directive} for more info.
    * @returns {ng.$compileProvider} Self for chaining.
    */
-   this.directive = function registerDirective(name, directiveFactory) {
+  this.directive = function callRegisterDirective(name, directiveFactory) {
+    return registerDirective(name, directiveFactory, false);
+  };
+   function registerDirective(name, directiveFactory, isComponent) {
```
DavidSouther: Extra space?

src/ng/compile.js
egor-smirnov: Maybe `ctrl` or `vm`?

ewinslow:
Would it be possible to get docs on what usage of this API is going to look like? It's hard to visualize just looking at the code...

caitp:
I envision it as using some kind of decorator library (we used to have a.js but that seems to have vanished somewhere), something like:

```js
// TypeScript decorators:
@Component({
  templateUrl: 'meow.html',
  controllerAs: 'MEOWCTRL'
})
class MeowComponent {
  constructor() {
    // Do important controller-initialization-stuff!
  }
}

// ES5 JS with some kind of a.js-like library
Decorate(function MeowComponent() {
  // Do important controller-initialization stuff!
}).With(Component, {
  templateUrl: 'meow.html',
  controllerAs: 'MEOWCTRL'
});
```

But you know, we are kind of just making it up as we go. No docs yet, I'll have to put more work into this before it gets that far

src/ng/compile.js
aciccarello: We use `vm` which I like because it is short. I'd only be concerned that it may be confusing for someone who isn't familiar with the "view model" terminology. That term would have to be explained well in the docs.

DavidSouther: Horrible terrible no good but very flexible idea: let users specify the default `controllerAs` in a provider, because no one is every going to like the default.

caitp: yeah we need to really figure out what we want to do here

aciccarello: see ngUpgraders/ng-forward#6 for discussion on configuration in that decorators implementation

shahata:
Closing in favor of #12933

PR #12167, opened 2015-06-19
docs($rootScope.Scope): reinsert lost example

louisik1:
This doc example was lost long ago here https://github.com/angular/angular.js/commit/2845dd159087fc59eb29845a578f32c7c462e8e6

As it is now, it says 'here is a simple scope snippet...' but has a non-linking file link to the test doc.  Doesn't make sense.  This change simply puts back the example as it originally was.

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -128,7 +128,7 @@ function $RootScopeProvider() {
          // still old value, since watches have not been called yet
          expect(scope.greeting).toEqual(undefined);
 
-         scope.$digest(); // fire all  the watches
+         scope.$digest(); // fire all the watches
          expect(scope.greeting).toEqual('Hello Misko!');
      * ```
      *
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -112,8 +112,24 @@ function $RootScopeProvider() {
      * compiled HTML template is executed.)
      *
      * Here is a simple scope snippet to show how you can interact with the scope.
-     * ```html
-     * <file src="./test/ng/rootScopeSpec.js" tag="docs1" />
+     * ```js
+         var scope = $rootScope.$new();
+         scope.salutation = 'Hello';
+         scope.name = 'World';
+
+         expect(scope.greeting).toEqual(undefined);
+
+         scope.$watch('name', function() {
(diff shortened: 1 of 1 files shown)


src/ng/rootScope.js
```diff
@@ -112,8 +112,24 @@ function $RootScopeProvider() {
...
+         scope.$watch('name', function() {
+           scope.greeting = scope.salutation + ' ' + scope.name + '!';
+         }); // initialize the watch
+
+         expect(scope.greeting).toEqual(undefined);
+         scope.name = 'Misko';
+         // still old value, since watches have not been called yet
+         expect(scope.greeting).toEqual(undefined);
+
+         scope.$digest(); // fire all  the watches
```
gkalpak: Since you are at it, why not fix the double space after "all" :D

louisik1: Good catch. Done.

src/ng/rootScope.js
```diff
@@ -112,8 +112,24 @@ function $RootScopeProvider() {
...
+         scope.$watch('name', function() {
+           scope.greeting = scope.salutation + ' ' + scope.name + '!';
+         }); // initialize the watch
+
+         expect(scope.greeting).toEqual(undefined);
+         scope.name = 'Misko';
+         // still old value, since watches have not been called yet
+         expect(scope.greeting).toEqual(undefined);
+
+         scope.$digest(); // fire all the watches
```
Narretz: We should probably change this to `scope.$apply()`, since the use of $digest is basically not recommened anywhere in the docs.

louisik1: I don't disagree regarding $digest not being recommended, but changing the example is not quite the scope of this request.  This example was accidentally removed so I am just putting it back. This sample comes directly from the unit tests. All the examples on this page use $digest; none of them use $apply.  The notes do indicate that $digest is not recommended, but similar notes regarding 'better practices' exist throughout the documentation.  If you would like to file an issue to have the examples all changed to $apply, I think that would be great and hopefully someone can pick it up.  Thanks!

Narretz:
I decided to not add the example, because it introduces concepts like $watch and $digest without actually explaining them. The examples for the methods do a better job at this. I also added a link to the guide.

PR #12174, opened 2015-06-19
fix($rootScope): Internal state consistency

lgalfaso:
During a recursive $apply, clear the state only if the state was set in that call.

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -1040,14 +1040,17 @@ function $RootScopeProvider() {
       $apply: function(expr) {
         try {
           beginPhase('$apply');
-          return this.$eval(expr);
-        } catch (e) {
+          try {
+            return this.$eval(expr);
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Although this is logically correct, I find it unnecessarily complicated to follow. Instead, we should simply wrap the `$eval` in its own try-finally block:

```js
        try {
          beginPhase('$apply');
          try {
            return this.$eval(expr);
          } finally {
            clearPhase();
          }
        } catch(e) {
          $exceptionHandler(e);
        } finally {
          try {
            $rootScope.$digest();
          } catch(e) {
            $exceptionHandler(e);
            throw e;
          }
        }
```

lgalfaso:
@petebacondarwin I wanted to avoid the second `try`, but I think that your version of the fix makes it easier to read. Updated the patch with the recommended change

PR #12180, opened 2015-06-21
Add hasBody to ngResource action configuration

yathos:
The use of hasBody would allow developers of REST interfaces to decide if a request body is necessary or not on a per method basis.

The way ngResource currently handles request bodies in custom requests make the definition of custom requests pretty useless. It’s not possible to transfere a request body through a  other requests than POST, PUT, PATCH. Or am I wrong?

It may be true that some browser implementations are also brocken when it comes to methods like DELETE but this broken behavior should not be enforced by angular.

Even as i read https://github.com/angular/angular.js/issues/3207 after creating the patch its somehow picking up pmariduenas comment on creating a enableDeleteRequestBody. Its just the generic version for all custom requests.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -161,6 +161,8 @@ function shallowClearAndCopy(src, dst) {
  *   - **`interceptor`** - `{Object=}` - The interceptor object has two optional methods -
  *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
  *     with `http response` object. See {@link ng.$http $http interceptors}.
+ *   - **`hasBody`** - `{boolean}` - allows to specify if a request body is to be used (not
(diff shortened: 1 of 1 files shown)


kaufholdr:
I confirm that I am fine with the contribution of this patch to Google.

yathos:
Can someone please tell me what to do to get the cia:yes tag?

caitp:
@yathos the CLA bot doesn't like it when someone elses commits are in your PR --- Perhaps @kaufholdr  could submit the PR themselves instead?

yathos:
Thank you. Will try it with the correct approach. :-)

yathos:
The corrected pull request is: https://github.com/angular/angular.js/pull/12181

PR #12181, opened 2015-06-21
Add hasBody to ngResource action configuration

kaufholdr:
The use of hasBody would allow developers of REST interfaces to decide if a request body is necessary or not on a per method basis.

The way ngResource currently handles request bodies in custom requests make the definition of custom requests pretty useless. It’s not possible to transfere a request body through a  other requests than POST, PUT, PATCH. Or am I wrong?

It may be true that some browser implementations are also brocken when it comes to methods like DELETE but this broken behavior should not be enforced by angular.

Even as i read the https://github.com/angular/angular.js/issues/3207 after creating the patch its somehow picking up pmariduenas comment on creating a enableDeleteRequestBody. Its just the generic version for all custom requests.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -194,7 +194,7 @@ function shallowClearAndCopy(src, dst) {
  *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
  *     with `http response` object. See {@link ng.$http $http interceptors}.
  *   - **`hasBody`** - `{boolean}` - allows to specify if a request body should be included or not.
- *     If not specified only POST, PUT and PATCH will have a request.
+ *     If not specified only POST, PUT and PATCH requests will have a body.
  *
  * @param {Object} options Hash with custom settings that should extend the
  *   default `$resourceProvider` behavior.  The supported options are:
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -193,6 +193,8 @@ function shallowClearAndCopy(src, dst) {
  *   - **`interceptor`** - `{Object=}` - The interceptor object has two optional methods -
  *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
  *     with `http response` object. See {@link ng.$http $http interceptors}.
+ *   - **`hasBody`** - `{boolean}` - allows to specify if a request body is to be used (not
+ *     required for POST,PUT,PATCH and can't disable body inclusion on this methods).
  *
  * @param {Object} options Hash with custom settings that should extend the
  *   default `$resourceProvider` behavior.  The supported options are:
@@ -640,7 +642,7 @@ angular.module('ngResource', ['ng']).
         };
 
         forEach(actions, function(action, name) {
-          var hasBody = /^(POST|PUT|PATCH)$/i.test(action.method);
+          var hasBody = /^(POST|PUT|PATCH)$/i.test(action.method) || action.hasBody === true;
           var numericTimeout = action.timeout;
           var cancellable = isDefined(action.cancellable) ?
               action.cancellable : route.defaults.cancellable;
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -194,7 +194,7 @@ function shallowClearAndCopy(src, dst) {
  *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
  *     with `http response` object. See {@link ng.$http $http interceptors}.
  *   - **`hasBody`** - `{boolean}` - allows to specify if a request body should be included or not.
- *     If not specified POST,PUT and PATCH will have a request body.
+ *     If not specified only POST, PUT and PATCH will have a request.
  *
  * @param {Object} options Hash with custom settings that should extend the
  *   default `$resourceProvider` behavior.  The supported options are:
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -193,8 +193,8 @@ function shallowClearAndCopy(src, dst) {
  *   - **`interceptor`** - `{Object=}` - The interceptor object has two optional methods -
  *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
  *     with `http response` object. See {@link ng.$http $http interceptors}.
- *   - **`hasBody`** - `{boolean}` - allows to specify if a request body is to be used (not
- *     required for POST,PUT,PATCH and can't disable body inclusion on this methods).
+ *   - **`hasBody`** - `{boolean}` - allows to specify if a request body should be included or not.
+ *     If not specified POST,PUT and PATCH will have a request body.
  *
  * @param {Object} options Hash with custom settings that should extend the
  *   default `$resourceProvider` behavior.  The supported options are:
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -642,7 +642,7 @@ angular.module('ngResource', ['ng']).
         };
 
         forEach(actions, function(action, name) {
-          var hasBody = /^(POST|PUT|PATCH)$/i.test(action.method) || action.hasBody === true;
+          var hasBody = action.hasBody === true || (action.hasBody !== false && /^(POST|PUT|PATCH)$/i.test(action.method));
           var numericTimeout = action.timeout;
           var cancellable = isDefined(action.cancellable) ?
               action.cancellable : route.defaults.cancellable;
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -102,6 +102,10 @@ describe('basic usage', function() {
     var condition = {at: '2038-01-19 03:14:08'};
 
     $httpBackend.expect('CREATE', '/fooresource', instant).respond({fid: 42});
+    $httpBackend.expectPOST('/fooresource').respond(function(method, url, data) {
+      expect(data).not.toBeUndefined();
+      return [200, {id: 42}];
+    });
     $httpBackend.expect('DELETE', '/fooresource', condition).respond({});
 
     var r = $resource('/fooresource', {}, {
@@ -110,16 +114,21 @@ describe('basic usage', function() {
     });
 
     var creationResponse = r.create(instant);
+    var saveResponse = r.save(null, {id: -1});
     var deleteResponse = r.delete(condition);
 
     $httpBackend.flush();
 
     expect(creationResponse.fid).toBe(42);
+    expect(saveResponse.id).toBe(42);
     expect(deleteResponse.$resolved).toBe(true);
   });
 
   it('should not include a request body if hasBody is false on POST, PUT and PATCH', function() {
-    $httpBackend.expect('POST', '/foo').respond({});
+    $httpBackend.expectPOST('/foo').respond(function(method, url, data) {
+      expect(data).toBeUndefined();
+      return [200, {id: 42}];
+    });
     $httpBackend.expect('PUT', '/foo').respond({});
     $httpBackend.expect('PATCH', '/foo').respond({});
 
@@ -135,7 +144,7 @@ describe('basic usage', function() {
 
     $httpBackend.flush();
 
-    expect(postResponse.$resolved).toBe(true);
+    expect(postResponse.id).toBe(42);
     expect(putResponse.$resolved).toBe(true);
     expect(patchResponse.$resolved).toBe(true);
   });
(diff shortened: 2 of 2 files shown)


caitp:
I don't think custom http methods ate necessarily recommended for REST services. This does need some consideration

kaufholdr:
I thought that the decision of enabling custom methods in ngResource has already been made as it is possible right now.

The current behavior of silently dropping the content body for custom requests is just weired (it took me several hours of debugging to locate the problem in angularjs). For me it feels like a bug and not a feature.

Do you think that the inclusion of this patch could be harmfull to the functionality of ngResource?

caitp:
Custom methods and custom http methods are different things, are there any other methods you want a body for? I know people have asked for bodies on DELETE requests

kaufholdr:
I know that there is a difference between custom http methods and custom methods but they are currently both possible in ngResource.

I would not include a body in the DELETE request as I currently have no use for it but i would like to include a body in a LOGIN request which carries the user information in a JSON (not only username an pasword). Some other custom http methods where i would like to include a body are: CLOCKIN, CLOCKOUT, STORNO, ASSIGN, NOTIFY, DISABLE but they are all specific to the project I'm working on. That's why I would like to have a generic attribute on the action specification.

yathos:
I created a modified bower-angular-resource version that includes the requested fix in version 1.4.1  https://github.com/yathos/bower-angular-resource If someone would like to test it.

kaufholdr:
Please tell me if there is anything that I could do to get this pull request included.

Narretz:
Honestly, I don't think custom HTTP methods are a big enough use case to warrant inclusion into the angular core. Especially since $resource is pretty much built for "normal" REST interfaces. If you are doing something special, the recommendation has always been to built your own wrapper around $http.

ElayneB:
+1 for this feature. We have built a client on top of $resource and we consider DELETE with bodies to be part of "normal" rest interfaces.

The use case is bulk relationship delete.
E.G.
`DELETE model/:id/relatedModel` `body: [relatedModelId, relatedModelId, ..]`

kaufholdr:
I'm still interested in getting this pull request integrated.

gkalpak:
Let's consider this after 1.6.0 is released and decide what direction to go (either merge or close).
It is a minimal addition and doesn't seem to do any harm; plus it could take care of the recurring debate about whether `delete` requests should have a body or not.

src/ngResource/resource.js
```diff
@@ -640,7 +642,7 @@ angular.module('ngResource', ['ng']).
         };
 
         forEach(actions, function(action, name) {
-          var hasBody = /^(POST|PUT|PATCH)$/i.test(action.method);
+          var hasBody = /^(POST|PUT|PATCH)$/i.test(action.method) || action.hasBody === true;
```
gkalpak: I would find something like the following more intuitive:

```js
var hasBody = action.hasBody === true || (action.hasBody !== false && /^(POST|PUT|PATCH)$/i.test(action.method));
```

Am I missing any reason why we shouldn't allow people to overwrite `hasBody` for POST, PUT, PATCH? (Not that they should.)

kaufholdr: I'm fine with both versions of the statement. Should I modify the patch?

kaufholdr: I had to like twice to see the error but the correct statement would be:
`var hasBody = action.hasBody === true || (action.hasBody === false && /^(POST|PUT|PATCH)$/i.test(action.method));`

gkalpak: I actually meant it the way I wrote it above. Your version is the same as:

```js
var hasBody = action.hasBody === true || /^(POST|PUT|PATCH)$/i.test(action.method);
```

...which is not what we want.

kaufholdr:
sry for taking so long it should be the correct version now.

gkalpak:
@kaufholdr, we won't get to this before next week, so take your time :smiley:

kaufholdr:
Is there anything more that I can do to get this feature included?

src/ngResource/resource.js
```diff
@@ -193,6 +193,8 @@ function shallowClearAndCopy(src, dst) {
  *   - **`interceptor`** - `{Object=}` - The interceptor object has two optional methods -
  *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
  *     with `http response` object. See {@link ng.$http $http interceptors}.
+ *   - **`hasBody`** - `{boolean}` - allows to specify if a request body should be included or not.
+ *     If not specified only POST,PUT and PATCH will have a request.
```
gkalpak: Nit: Missing whitespace after `,`.

gkalpak (changes requested):
We also need tests for this.

gkalpak:
@kaufholdr, feel free to ping me once you've made the requested changes, so I can take another look.

kaufholdr:
Sure. I have added the required test cases. Are they sufficient?

src/ngResource/resource.js
```diff
@@ -193,6 +193,8 @@ function shallowClearAndCopy(src, dst) {
  *   - **`interceptor`** - `{Object=}` - The interceptor object has two optional methods -
  *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
  *     with `http response` object. See {@link ng.$http $http interceptors}.
+ *   - **`hasBody`** - `{boolean}` - allows to specify if a request body should be included or not.
+ *     If not specified only POST, PUT and PATCH will have a request.
```
gkalpak: _only POST, PUT and PATCH will have a request_ --> _only POST, PUT and PATCH requests will have a body_

test/ngResource/resourceSpec.js
```diff
@@ -97,6 +97,43 @@ describe('basic usage', function() {
     $httpBackend.flush();
   });
 
+  it('should include a request body when calling custom delete with hasBody is true', function() {
```
gkalpak: _custom delete_ --> _custom method_

Can you also test with another method (in the same test), just to avoid giving the impression this is just about `DELETE`.

gkalpak (changes requested):
We also need a test for POST/PUT/PATCH methods with `hasBody: false`.

kaufholdr:
I have added the requested changes.

Should I squash down my changes to one commit on completion or would you like to take them as several commits if it's ready?

test/ngResource/resourceSpec.js
```diff
@@ -97,21 +97,50 @@ describe('basic usage', function() {
...
+
+    var R = $resource('/foo', {}, {
+      post: {method: 'POST', hasBody: false},
+      put: {method: 'PUT', hasBody: false},
+      patch: {method: 'PATCH', hasBody: false}
+    });
+
+    var postResponse = R.post();
+    var putResponse = R.put();
+    var patchResponse = R.patch();
```
gkalpak: Would these 3 calls fail if `hasBody` was not set? I don't think so (but I haven't tried it either). Can you check?

test/ngResource/resourceSpec.js
```diff
@@ -97,21 +97,50 @@ describe('basic usage', function() {
     $httpBackend.flush();
   });
 
-  it('should include a request body when calling custom delete with hasBody is true', function() {
+  it('should include a request body when calling custom method with hasBody is true', function() {
+    var instant = {name: 'info.txt', value: 'V2hlbiB0aGUgdGltZSBlbmRzLg=='};
+    var fid = 42;
+    var created = {fid: fid, filname: 'fooresource', value: 'V2hlbiB0aGUgdGltZSBlbmRzLg=='};
```
gkalpak: Nit: You don't really need these 2 extra variables (`fid`, `created`). You could inline them (and only keep the necessary props) to avoid visual noise.

gkalpak:
You don't need to squash them, btw. We can take care of it while merging.

test/ngResource/resourceSpec.js
kaufholdr: I have cleaned up those variables.

test/ngResource/resourceSpec.js
kaufholdr: No they don't fail if hasBody is not set. This is a bug or a feature of ngResource not enforcing the described api.

kaufholdr: I haven't found a way to distinguish between request parameters and the request body in httpBackend.

Did I miss something?

kaufholdr:
Is there anything that I should fix in this Pull request right now?

test/ngResource/resourceSpec.js
gkalpak: If you pass two params (e.g. `R.post(null, {foo: 'bar'})`), then the first one is interpreted as `params` and the second one as `data`.

For example, th following test currently fails, because `{id: -1}` is not undefined. If something similar passes with your PR, then we are good:

```js
    $httpBackend.expectPOST('/foo').respond(function(method, url, data) {
      expect(data).not.toBeUndefined();
      return [200, {id: 42}];
    });

    var R = $resource('/foo');
    var response = R.save(null, {id: -1});

    $httpBackend.flush();
    expect(response.id).toBe(42);
```

gkalpak:
@kaufholdr, [that][1] should be the last bit :smiley:

[1]: https://github.com/angular/angular.js/pull/12181/files#r105556810

test/ngResource/resourceSpec.js
kaufholdr: sry for the delayed response on my side. I have merged your proposed test case into the test method with hasBody true. The only problem for me was that I couldn't reproduce the failing of the test case. (Even when I used your version.)

Can you see my error or give me the error that you get?

test/ngResource/resourceSpec.js
```diff
@@ -102,6 +102,10 @@ describe('basic usage', function() {
     var condition = {at: '2038-01-19 03:14:08'};
 
     $httpBackend.expect('CREATE', '/fooresource', instant).respond({fid: 42});
+    $httpBackend.expectPOST('/fooresource').respond(function(method, url, data) {
```
gkalpak: This is not needed here. `POST` is not a custom method (which what this test is about).

test/ngResource/resourceSpec.js
```diff
@@ -110,16 +114,21 @@ describe('basic usage', function() {
...
     $httpBackend.flush();
 
     expect(creationResponse.fid).toBe(42);
+    expect(saveResponse.id).toBe(42);
     expect(deleteResponse.$resolved).toBe(true);
   });
 
   it('should not include a request body if hasBody is false on POST, PUT and PATCH', function() {
-    $httpBackend.expect('POST', '/foo').respond({});
+    $httpBackend.expectPOST('/foo').respond(function(method, url, data) {
```
gkalpak: Make this change to all three methods. E.g. something like:

```js
function verifyRequest(method, url , data) {
  expect(data).toBeUndefined();
  return [200, {id: 42}];
}

$httpBackend.expect('POST', '/foo').respond(verifyRequest);
$httpBackend.expect('PUT', '/foo').respond(verifyRequest);
$httpBackend.expect('PATCH', '/foo').respond(verifyRequest);

...
```

gkalpak (changes requested):

test/ngResource/resourceSpec.js
gkalpak: I was probably not clear. What I meant is that my test fails with the code currently in master (i.e. without your PR), which is expected, because we don't support the `hasBody` yet. If the test passes with your PR, then it verifies that the feature you added (`hasBody`) does indeed work for `PUT/POST/PATCH` requests too, so it is good to be merged.

kaufholdr:
Fell free to tell me if there is still something missing. :-)

test/ngResource/resourceSpec.js
```diff
@@ -97,6 +97,76 @@ describe('basic usage', function() {
...
+
+    var postResponse = R.post();
+    var putResponse = R.put();
+    var patchResponse = R.patch();
+
+    $httpBackend.flush();
+
+    expect(postResponse.id).toBe(42);
+    expect(putResponse.$resolved).toBe(true);
+    expect(patchResponse.$resolved).toBe(true);
```
gkalpak: Please, change these two lines to `....id).toBe(42)`.

gkalpak (changes requested):

gkalpak (approved):

kaufholdr:
Thank. You. :-)

gkalpak:
I tweaked the docs a little bit and merged. Thank you for working on this, @kaufholdr 👍

PR #12189, opened 2015-06-22
fix(select): select doesn't work when label and ngModel value match.

### LINKED ISSUE #11835: Work-around for major 1.3x ng-options issue?
`ng-options` has major issues in the 1.3x branch (see #9714 and Breaking Changes under '1.4.0-beta.0' in the CHANGELOG: https://github.com/angular/angular.js/blob/master/CHANGELOG.md ). 

My Angular 1.3.3 production app just hit a wall with this issue: http://plnkr.co/edit/JCM2NOxxO4OFoWybmQhp?p=preview

``` javascript 
angular.module('MyApp', [])

.controller('App', function App($scope, $timeout) {
  $scope.options = null;

  $scope.selected = {
    //try toggling the below IDs

    // works
    //id: 400 

    // DOES NOT work in 1.3x... works in 1.4.0-rc.1+
    id: 1001 // 1002 doesn't work either ... notice 1001 is also a value for 'name'
  };

  // let's pretend ajax call
  $timeout(function() {
    $scope.options = [{
      "id": 400,
      "name": "1001"
    }, {
      "id": 401,
      "name": "1002"
    }, {
      "id": 1001,
      "name": "1003"
    }, {
      "id": 1002,
      "name": "1004"
    }]
  }, 1000);

});
```

This issue is addressed in 1.4.0-rc.0+. However, while 1.4 continues to bake (and hopefully comes out soon), what would be a possible work around?

rgaskill:
The selectCtrl `addOption` compares the label value with the
ngModel value when determining when to set the value of the select
element.  This causes a problem if the label matches model value but
not the selectAs value of the option.

fixes #11835

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -715,13 +715,15 @@ var selectDirective = ['$compile', '$parse', function($compile,   $parse) {
             }
             optionGroup[0].element.remove();
           }
-          forEach(labelMap, function(count, label) {
-            if (count > 0) {
-              selectCtrl.addOption(label);
-            } else if (count < 0) {
-              selectCtrl.removeOption(label);
-            }
-          });
+          if (!anySelected) {
+            forEach(labelMap, function(count, label) {
+              if (count > 0) {
+                selectCtrl.addOption(label);
+              } else if (count < 0) {
+                selectCtrl.removeOption(label);
+              }
+            });
+          }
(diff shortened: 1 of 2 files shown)


Narretz:
We've decided not to fix this issue in 1.3.x. See #11835 (comment) for more info

PR #12196, opened 2015-06-23
feat(linky): support tel scheme

gabrielmaldi:


moonklhh:
docs/app/src/errors.js
src/ngSanitize/filter/linky.js

PR #12201, opened 2015-06-24
fix(ngResource): fix query url params encoding

alediaferia:
Encoding URL Query Parameters should be "more aggressive" than the URL Segment encoding.
You can see what I mean in the added test case.

I'm currently using AngularJS 1.3.16 but I found this bug is still there in master.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12201)

### DIFF
--- a/scripts/npm/install-dependencies.sh
+++ b/scripts/npm/install-dependencies.sh
@@ -10,7 +10,7 @@ if diff -q $SHRINKWRAP_FILE $SHRINKWRAP_CACHED_FILE; then
 else
   echo 'Blowing away node_modules and reinstalling npm dependencies...'
   rm -rf node_modules
-  npm install
+  npm install --registry http://registry.npmjs.org/
   cp $SHRINKWRAP_FILE $SHRINKWRAP_CACHED_FILE
   echo 'npm install successful!'
 fi
--- a/scripts/npm/install-dependencies.sh
+++ b/scripts/npm/install-dependencies.sh
@@ -10,7 +10,7 @@ if diff -q $SHRINKWRAP_FILE $SHRINKWRAP_CACHED_FILE; then
 else
   echo 'Blowing away node_modules and reinstalling npm dependencies...'
   rm -rf node_modules
-  npm install --registry http://registry.npmjs.org/
+  npm install
   cp $SHRINKWRAP_FILE $SHRINKWRAP_CACHED_FILE
   echo 'npm install successful!'
 fi
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -433,7 +433,7 @@ angular.module('ngResource', ['ng']).
             }
             if (!(new RegExp("^\\d+$").test(param)) && param &&
               (new RegExp("(^|[^\\\\]):" + param + "(\\W|$)").test(url))) {
-              urlParams[param] = true;
+              urlParams[param] = { isQueryParam: (new RegExp("\\?(?:.*):" + param + "(?:\\W|$)")).test(url) };
             }
           });
           url = url.replace(/\\:/g, ':');
@@ -443,10 +443,14 @@ angular.module('ngResource', ['ng']).
           });
 
           params = params || {};
-          forEach(self.urlParams, function(_, urlParam) {
+          forEach(self.urlParams, function(paramInfo, urlParam) {
             val = params.hasOwnProperty(urlParam) ? params[urlParam] : self.defaults[urlParam];
             if (angular.isDefined(val) && val !== null) {
-              encodedVal = encodeUriSegment(val);
+              if (paramInfo.isQueryParam === true) {
+                encodedVal = encodeUriQuery(val, true);
+              } else {
+                encodedVal = encodeUriSegment(val);
+              }
               url = url.replace(new RegExp(":" + urlParam + "(\\W|$)", "g"), function(match, p1) {
                 return encodedVal + p1;
               });
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -433,7 +433,7 @@ angular.module('ngResource', ['ng']).
             }
             if (!(new RegExp("^\\d+$").test(param)) && param &&
               (new RegExp("(^|[^\\\\]):" + param + "(\\W|$)").test(url))) {
-              urlParams[param] = { isQueryParamValue: (new RegExp("\\?(?:.*)=:" + param + "(?:\\W|$)")).test(url) };
+              urlParams[param] = { isQueryParamValue: (new RegExp("\\?.*=:" + param + "(?:\\W|$)")).test(url) };
             }
           });
           url = url.replace(/\\:/g, ':');
@@ -446,7 +446,7 @@ angular.module('ngResource', ['ng']).
           forEach(self.urlParams, function(paramInfo, urlParam) {
             val = params.hasOwnProperty(urlParam) ? params[urlParam] : self.defaults[urlParam];
             if (angular.isDefined(val) && val !== null) {
-              if (paramInfo.isQueryParamiValue === true) {
+              if (paramInfo.isQueryParamValue === true) {
                 encodedVal = encodeUriQuery(val, true);
               } else {
                 encodedVal = encodeUriSegment(val);
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -433,7 +433,7 @@ angular.module('ngResource', ['ng']).
             }
             if (!(new RegExp("^\\d+$").test(param)) && param &&
               (new RegExp("(^|[^\\\\]):" + param + "(\\W|$)").test(url))) {
-              urlParams[param] = { isQueryParam: (new RegExp("\\?(?:.*):" + param + "(?:\\W|$)")).test(url) };
+              urlParams[param] = { isQueryParamValue: (new RegExp("\\?(?:.*)=:" + param + "(?:\\W|$)")).test(url) };
             }
           });
           url = url.replace(/\\:/g, ':');
@@ -446,7 +446,7 @@ angular.module('ngResource', ['ng']).
           forEach(self.urlParams, function(paramInfo, urlParam) {
             val = params.hasOwnProperty(urlParam) ? params[urlParam] : self.defaults[urlParam];
             if (angular.isDefined(val) && val !== null) {
-              if (paramInfo.isQueryParam === true) {
+              if (paramInfo.isQueryParamiValue === true) {
                 encodedVal = encodeUriQuery(val, true);
               } else {
                 encodedVal = encodeUriSegment(val);
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -338,7 +338,7 @@ describe("resource", function() {
   });
 
 
-  it('should handle url encoding in mapped params', function() {
+  it('should encode & in query params unless in query param value', function() {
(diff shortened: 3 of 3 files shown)


alediaferia:
I signed it!

nacmacfeegle:
+1 to this fix, thanks @alediaferia

src/ngResource/resource.js
```diff
@@ -431,16 +431,20 @@ angular.module('ngResource', ['ng']).
             }
             if (!(new RegExp("^\\d+$").test(param)) && param &&
               (new RegExp("(^|[^\\\\]):" + param + "(\\W|$)").test(url))) {
-              urlParams[param] = true;
+              urlParams[param] = { isQueryParam: (new RegExp("\\?(.*)\\:" + param)).test(url) };
```
gkalpak: You need the `'(\\W|$)'` part or else you might get false positive. (Better yet `'(?:\\W|$)'`.)
E.g. `param: 'foo'` and `url: '/path/foo?key=:foobar'` will lead to a false positive.

alediaferia: I agree my regex is too broad but I'm not sure I completely get the one proposed by you.
Would something like the following be more accurate?
`new RegExp("\\?((\\w+)\\=\\w+&)*\\w+=:" + param + "(\\W|$)").test(url)`

gkalpak: No need to make it too strict either :)

IMO, `new RegExp('\\?(.*):' + param + '(\\W|$)')` (or `new RegExp('\\?(?:.*):' + param + '(?:\\W|$)')`) is fine.

alediaferia: So what does `?:` match exactly? This is the only bit I don't understand. :)

alediaferia: Ooh, it's the non-capturing group! :)

alediaferia: I've updated the code accordingly and split my commit messages to better comply with contribution guidelines.

alediaferia:
I've updated the code accordingly and split my commit messages to better comply with contribution guidelines.

src/ngResource/resource.js
```diff
@@ -431,16 +431,20 @@ angular.module('ngResource', ['ng']).
             }
             if (!(new RegExp("^\\d+$").test(param)) && param &&
               (new RegExp("(^|[^\\\\]):" + param + "(\\W|$)").test(url))) {
-              urlParams[param] = true;
+              urlParams[param] = { isQueryParam: (new RegExp("\\?(?:.*)\\:" + param + "(?:\\W|$)")).test(url) };
```
gkalpak: Nitpick: You don't need `\\` before `:` (it's not a special character in this context).

gkalpak:
LGTM

davidesantangelo:
LGMT

alediaferia:
Is there a plan for when this will go in the main line?

alediaferia:
Bump!

alediaferia:
Is there something I can do to ease the merge for this patch?

gkalpak:
@alediaferia, sorry it wasn't brought up earlier (I didn't spot it originally - thx @petebacondarwin for pointing it out), but this seems to be a breaking change.
E.g., one might want to specify the whole query string as param, like this:

```js
$resource('/api/myapp/resource?:query').get({query: 'foo&bar'});
```

--
So, in order to avoid the breaking change and still benefit from the fix, I propose that the `&` is only encoded, if the param is a recognized as query string value part (of a key=value pair).

This can be achieved, with the following changes:

1. Change this line:
        `urlParams[param] = { isQueryParam: (new RegExp("\\?(?:.*):" + param + "(?:\\W|$)")).test(url) };`
    to that:
        `urlParams[param] = { isQueryParamValue: (new RegExp("\\?(?:.*)=:" + param + "(?:\\W|$)")).test(url) };`

2. Change `paramInfo.isQueryParam` to `paramInfo.isQueryParamValue` (or whatever you name it).

3. Modify the test, in order to check for both cases:

        it('should encode & in query params unless in query param value', function() {
          var R1 = $resource('/api/myapp/resource?:query');
          $httpBackend.expect('GET', '/api/myapp/resource?foo&bar').respond({});
          R1.get({query: 'foo&bar'});
          
          var R2 = $resource('/api/myapp/resource?from=:from');
          $httpBackend.expect('GET', '/api/myapp/resource?from=bar%20%26%20blanks').respond({});
          R2.get({from: 'bar & blanks'});
        });

--
@alediaferia, would you be up to making those changes and see what people think ?

alediaferia:
@gkalpak will change according to your suggestions ASAP. Thank you for getting back :)

petebacondarwin:
@alediaferia nice one. Keep your ki cool. Thanks for the help.

src/ngResource/resource.js
```diff
@@ -433,7 +433,7 @@ angular.module('ngResource', ['ng']).
             }
             if (!(new RegExp("^\\d+$").test(param)) && param &&
               (new RegExp("(^|[^\\\\]):" + param + "(\\W|$)").test(url))) {
-              urlParams[param] = true;
+              urlParams[param] = { isQueryParamValue: (new RegExp("\\?(?:.*)=:" + param + "(?:\\W|$)")).test(url) };
```
gkalpak: Hm...looking at it again, why not change `(?:.*)` to `.*` ?

test/ngResource/resourceSpec.js
```diff
@@ -338,6 +338,17 @@ describe("resource", function() {
   });
 
 
+  it('should handle url encoding in mapped params', function() {
```
gkalpak: I'd prefer a more descriptive testcase name (e.g. `should encode & in query params unless in query param value` or something).

src/ngResource/resource.js
alediaferia: Makes sense

src/ngResource/resource.js
```diff
@@ -443,10 +443,14 @@ angular.module('ngResource', ['ng']).
           });
 
           params = params || {};
-          forEach(self.urlParams, function(_, urlParam) {
+          forEach(self.urlParams, function(paramInfo, urlParam) {
             val = params.hasOwnProperty(urlParam) ? params[urlParam] : self.defaults[urlParam];
             if (angular.isDefined(val) && val !== null) {
-              encodedVal = encodeUriSegment(val);
+              if (paramInfo.isQueryParamiValue === true) {
```
gkalpak: _isQueryParamiValue_  -->  _isQueryParamValue_

(It's a good idea to locally run the tests before pushing, to ensure nothing is (obviously) breaking :smiley:)

gkalpak:
@alediaferia, thx for the update. I left a couple of comments (mostly minor).
Feel free to ping me once you've pushed changes, so I can take another look.

src/ngResource/resource.js
alediaferia: I'm sorry for that. Actually I thought I pushed just to my `origin` and not to `upstream` as well. Still didn't manage to run the tests. Will do ASAP

gkalpak: NP :)

BTW, when you have submitted a PR, you don't need to push upstream or anything; pushing to your origin/<branch> will auto-update the PR (GitHub feature), so that's probably what happened.

Just ping me when you're done, so I take a final look. Thx for all the work so far !

alediaferia:
@gkalpak BTW do you have any hint on why now my `grunt package` fails with `stderr maxBuffer exceeded` ?

gkalpak:
@alediaferia, no idea :) Does it happen with the code currently on the PR ?
(If not, you can push the code somewhere so I can take a look.)

petebacondarwin:
Sounds like something to do with the closure compiler?

alediaferia:
@gkalpak @petebacondarwin sorry for not getting back earlier. It's happening with the current PR but it's happening with updated master as well. It may be something wrong with my npm/node configuration, but not sure. I'm not much into JavaScript :)

gkalpak:
@alediaferia, (un)fortunately, I'm not getting any issues with `grunt package` :disappointed:

test/ngResource/resourceSpec.js
```diff
@@ -338,6 +338,17 @@ describe("resource", function() {
   });
 
 
+  it('should encode & in query params unless in query param value', function() {
```
alediaferia: I'm wondering if part of this test is a duplicate of the one preceding this test.

alediaferia:
@gkalpak ping

test/ngResource/resourceSpec.js
gkalpak: I would say close but not quite.
I would be happy to have them merged into one `it` block (keeping all 3 cases).

Feel free to merge them.

scripts/npm/install-dependencies.sh
```diff
@@ -10,7 +10,7 @@ if diff -q $SHRINKWRAP_FILE $SHRINKWRAP_CACHED_FILE; then
 else
   echo 'Blowing away node_modules and reinstalling npm dependencies...'
   rm -rf node_modules
-  npm install
+  npm install --registry http://registry.npmjs.org/
```
gkalpak: I guess this has to do with your `grunt package` issues, but it definitely does not belong in this PR :)

alediaferia: Woops! sorry :)

gkalpak:
LGTM, thx @alediaferia !

/cc @petebacondarwin

alediaferia:
If this is ok, shall I clean up the git history squashing the latest commits and following your commit message conventions?

petebacondarwin:
OK, LGTM. @alediaferia thanks for your work on this. If you want to squash and tidy up the commit message that would be great. Otherwise, perhaps @gkalpak could do it when merging?

gkalpak:
Sure, I'll take care of squashing and merging !

alediaferia:
Thank you @gkalpak @petebacondarwin

gkalpak:
@petebacondarwin, should this be backported to 1.4.x ?

petebacondarwin:
If you are confident it is not a breaking change.

gkalpak:
It's one of those times that I find it hard to distinguish between a fix and a BC.

The only change introduced is when a param (`:xyz`) is used as the _value part_ of a URL query key-value pair. In any other case, there is no difference. But it could still break an app (theoretically).

E.g., assuming `$resource('/path?key=:val').get({val: 'foo&bar'})`, the request is sent to this URL:

1.4.x: `/path?key=foo&bar`
1.5.x: `/path?key=foo%26bar`

alediaferia:
Although I feel the "new" behaviour is the right one I unfortunately think there may be a lot of apps out there relying on a workaround for this "issue". 

If you look for e.g. `"angularjs ngresource query parameters"` on Google there are some issues open on SO with a few workarounds that usually involve using `$http` directly.

Just my 2 cents.

Happy to have contributed to such a big project anyway :)

petebacondarwin:
OK, so no backport.

gkalpak:
@alediaferia, I couldn't find any SO questions/answers that would be broken by this "fix". Relying on a workaround isn't a problem (it won't get broken).
Do you have any specific SO question/answer in mind ?

The problem is if someone has as a URL template such as `/path?key=:val` and expects to set other query param via `:val` (e.g. `/path?key=1&key2=2&key3=3`). It is a weird usecase imo (but theoretically possible nontheless).

alediaferia:
@gkalpak I cannot manage to find the question I had in mind, strange. Anyway, as you point out, relying on a workaround won't make things break. That's true. I don't have further objections.

PR #12202, opened 2015-06-24
docs(ngAnimate): change .ng-enter to .ng-animate

yichaowang:
.ng-enter class is not added by animateCss service. .enter and .ng-animate is added. Maybe a fix of adding .ng-enter rather than .enter is more proper.

### DIFF
--- a/src/ngAnimate/module.js
+++ b/src/ngAnimate/module.js
@@ -402,7 +402,7 @@
  *
  * ```css
  * /&#42; the transition tells ngAnimate to make the animation happen &#42;/
- * .slide.ng-enter { transition:0.5s linear all; }
+ * .slide.ng-animate { transition:0.5s linear all; }
  *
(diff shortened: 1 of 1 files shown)


gkalpak:
Indeed it seems that the `enter` and `enter-active` classes are added, but according to the docs they should `ng-enter` and `ng-enter-active`. @matsko, is this a bug ?

Also, the examples on https://docs.angularjs.org/api/ngAnimate seem rather outdated (and have several types (e.g. missing closing `)`, `height: XYZ` instead of `height: 'XYZpx'` etc).

matsko:
No the CSS code is right, but we forgot to add the property that instructs `$animateCss` to apply the `ng-` prefix.

matsko:
This fix has taken care of that: https://github.com/angular/angular.js/commit/344dffbc54df221d03eaa6dbbb677b2cefd20a23

matsko:
Thank you for bringing this up.

PR #12203, opened 2015-06-24
docs(ngAnimate): if statement typo. use element css selector.

yichaowang:
Replace ? to :. The directive is a element in example not an attribute.


gkalpak:
LGTM

PR #12213, opened 2015-06-25
feat(ngMessages): add support for default message

### LINKED ISSUE #12008: ngMessageDefault - Add support for showing values not mapped with ng-message
Many times it is necessary that we show a generic error message for unexpected values.
Would be nice to have something like:

``` HTML
<div ng-messages="controller.errors">
    <div ng-message="error1">Option 1</div>
    <div ng-message="error2">Option 2</div>
    <div ng-message-default>Option 3</div>
</div>
```

This way, whenever there is an "error3" in "controller.errors", as it is not mapped, the value under the ng-message-default would be shown.

gdangelo:
Added support for showing default message when no values are mapped with ng-message.

Closes #12008

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12213)

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -21,7 +21,7 @@ var jqLite = angular.element;
  * sequencing based on the order of how the messages are defined in the template.
  *
  * Currently, the ngMessages module only contains the code for the `ngMessages`, `ngMessagesInclude`
- * `ngMessage` and `ngMessageExp` directives.
+ * `ngMessage`, `ngMessageExp` and `ngMessageDefault` directives.
  *
  * # Usage
  * The `ngMessages` directive listens on a key/value collection which is set on the ngMessages attribute.
@@ -239,6 +239,19 @@ var jqLite = angular.element;
  * .some-message.ng-leave.ng-leave-active {}
  * ```
  *
+ * ## Displaying a default message
+ * If the ngMessages renders no inner ngMessage directive (that is to say when the key values does not
+ * match the attribute value present on each ngMessage directive), then it will render a default message
+ * using the ngMessageDefault directive.
+ *
+ * ```html
+ * <div ng-messages="myForm.myField.$error" role="alert">
+ *   <div ng-message="required">This field is required</div>
+ *   <div ng-message="minlength">This field is too short</div>
+ *   <div ng-message-default>This is a default message</div>
+ * </div>
+ * ```
+ *
  * {@link ngAnimate Click here} to learn how to use JavaScript animations or to learn more about ngAnimate.
  */
 angular.module('ngMessages', [])
@@ -260,6 +273,9 @@ angular.module('ngMessages', [])
     * at a time and this depends on the prioritization of the messages within the template. (This can
     * be changed by using the `ng-messages-multiple` or `multiple` attribute on the directive container.)
     *
+    * A default message can also be displayed when no `ngMessage` directive is inserted, using the
+    * `ngMessageDefault` directive.
+    *
     * A remote template can also be used to promote message reusability and messages can also be
     * overridden.
     *
@@ -272,6 +288,7 @@ angular.module('ngMessages', [])
     *   <ANY ng-message="stringValue">...</ANY>
     *   <ANY ng-message="stringValue1, stringValue2, ...">...</ANY>
     *   <ANY ng-message-exp="expressionValue">...</ANY>
+    *   <ANY ng-message-default>...</ANY>
     * </ANY>
     *
     * <!-- or by using element directives -->
@@ -279,6 +296,7 @@ angular.module('ngMessages', [])
     *   <ng-message when="stringValue">...</ng-message>
     *   <ng-message when="stringValue1, stringValue2, ...">...</ng-message>
     *   <ng-message when-exp="expressionValue">...</ng-message>
+    *   <ng-message-default>...</ng-message-default>
     * </ng-messages>
     * ```
     *
@@ -307,6 +325,7 @@ angular.module('ngMessages', [])
     *         <div ng-message="required">You did not enter a field</div>
     *         <div ng-message="minlength">Your field is too short</div>
     *         <div ng-message="maxlength">Your field is too long</div>
+    *         <div ng-message-default>This is a default message</div>
     *       </div>
     *     </form>
     *   </file>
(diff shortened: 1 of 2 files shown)


lgalfaso:
it looks like this PR adds quite a few of duplicated code. Before doing a proper review, would it be possible to try to keep the core a little bit more DRY?

gdangelo:
I will make the modifications soon.

petebacondarwin:
This looks reasonable let's get it into 1.6 early as soon as 1.5.0 is released.

mikeycgto:
Any updates on this making it into 1.6?

src/ngMessages/messages.js
```diff
@@ -239,6 +239,19 @@ var jqLite = angular.element;
...
+ * ## Displaying a default message
+ * If the ngMessages renders no inner ngMessage directive (that is to say when the key values does not
+ * match the attribute value present on each ngMessage directive), then it will render a default message
+ * using the ngMessageDefault directive.
+ *
+ * ```html
+ * <div ng-messages="myForm.myField.$error" role="alert">
+ *   <div ng-message="required">This field is required</div>
+ *   <div ng-message="minlength">This field is too short</div>
+ *   <div ng-message-default>This is a default message</div>
```
gkalpak: I would use a more meaningful message here to give a better impression of how to use the directive.

src/ngMessages/messages.js
```diff
@@ -379,9 +398,17 @@ angular.module('ngMessages', [])
              messageCtrl.detach();
            });
 
-           unmatchedMessages.length !== totalMessages
-              ? $animate.setClass($element, ACTIVE_CLASS, INACTIVE_CLASS)
-              : $animate.setClass($element, INACTIVE_CLASS, ACTIVE_CLASS);
+           if (unmatchedMessages.length !== totalMessages) {
+             // Unset default message if setted
```
gkalpak: _setted_  -->  _set_

src/ngMessages/messages.js
```diff
@@ -397,23 +424,32 @@ angular.module('ngMessages', [])
...
-           var nextKey = latestKey.toString();
-           messages[nextKey] = {
-             message: messageCtrl
-           };
-           insertMessageNode($element[0], comment, nextKey);
-           comment.$$ngMessageNode = nextKey;
-           latestKey++;
+         this.register = function(comment, messageCtrl, isDefault) {
+           if (isDefault) {
+            ctrl.default = messageCtrl;
```
gkalpak: Wrong indentation (here and below).

src/ngMessages/messages.js
```diff
@@ -397,23 +424,32 @@ angular.module('ngMessages', [])
...
          };
 
-         this.deregister = function(comment) {
-           var key = comment.$$ngMessageNode;
-           delete comment.$$ngMessageNode;
-           removeMessageNode($element[0], comment, key);
-           delete messages[key];
+         this.deregister = function(comment, isDefault) {
+           if (isDefault) {
+            delete ctrl.default;
```
gkalpak: I think you also need to detach `ctrl.default` (if defined), because if it is visible, it won't get removed by `ctrl.reRender()`.

src/ngMessages/messages.js
```diff
@@ -594,35 +630,74 @@ angular.module('ngMessages', [])
...
+    *
+    * <!-- or by using element directives -->
+    * <ng-messages for="expression" role="alert">
+    *   <ng-message when="stringValue">...</ng-message>
+    *   <ng-message when="stringValue1, stringValue2, ...">...</ng-message>
+    *   <ng-message-default>...</ng-message-default>
+    * </ng-messages>
+    * ```
+    *
+    * @param {expression} ngMessageDefault|when no ngMessage matches.
```
gkalpak: It doesn't take any params.

test/ngMessages/messagesSpec.js
```diff
@@ -401,6 +401,24 @@ describe('ngMessages', function() {
...
   });
 
+  it('should render a default message when no one matched', inject(function($rootScope, $compile) {
+    element = $compile('<div ng-messages="col">' +
+                       '  <div ng-message="val">Message is set</div>' +
+                       '  <div ng-message-default>Default message is set</div>' +
+                       '</div>')($rootScope);
+    $rootScope.$digest();
+
+    //expect(element.text()).not.toContain('Message is set');
```
gkalpak: Remove comment.

test/ngMessages/messagesSpec.js
```diff
@@ -401,6 +401,24 @@ describe('ngMessages', function() {
...
+    expect(element.text()).toContain('Default message is set');
```
gkalpak: Why not use `.toBe()`?

test/ngMessages/messagesSpec.js
```diff
@@ -401,6 +401,24 @@ describe('ngMessages', function() {
...
+
+    $rootScope.$apply(function() {
+      $rootScope.col = { val: true };
+    });
```
gkalpak: `$rootScope.$apply('col = {val: true}')` is more concise :smiley:

test/ngMessages/messagesSpec.js
```diff
@@ -401,6 +401,24 @@ describe('ngMessages', function() {
...
+
+    //expect(element.text()).toContain('Message is set');
+    expect(element.text()).not.toContain('Default message is set');
```
gkalpak: Again, why not use `.toBe()` and verify the exact contents?

gkalpak:
A few more tests would be (e.g. using `ng-messages-multiple` or using `ngIf` on `ngMessageDefault` etc). But overall looks good.

src/ngMessages/messages.js
mikeycgto: Do you mean to call `removeMessageNode` here for the default?

src/ngMessages/messages.js
mikeycgto: Not sure I follow...?

src/ngMessages/messages.js
mikeycgto: Anything in specific that should be mentioned here?

gkalpak: Anything that you would use in an actual app. In this case, since these seem to be validation messages, the default message would be something like 'Valid' or a checkmark or 'This field is valid` etc.

src/ngMessages/messages.js
gkalpak: `X  -->  Y` is my way of telling that there is a typo in `X`, it should be changed to `Y` :stuck_out_tongue: 
(Basically, change `setted` to `set`.)

src/ngMessages/messages.js
gkalpak: I don't think `removeMessageNode()` works for `ctrl.default`. I was thinking `ctrl.default.detach()`.
<sub>And there should also be a test about dynamically adding/removing an `<ng-message-default>` element (e.g. using `ngIf`).</sub>

gdangelo:
Hi @gkalpak, @mikeycgto, thanks to follow this feature request. Do we need anything else to include the feature in the 1.6 release?

petebacondarwin:
@gdangelo - yes, @gkalpak provided a number of comments that need addressing, including some more tests.

PR #12217, opened 2015-06-26
fix($compile): throw error when requesting new and isolate scopes (async)

### LINKED ISSUE #12215: Isolate scope weird behavior based on directive's name.
I have 2 directives on the same element. One has isolate scope (scope: {}), the other one inherits scope from parent (scope: true). Angular version 1.3.16 (and even 1.4).

This should throw an exception, but it doesn't if my directive's name starts with letter "o" or any following letter in the alphabet.

See for your self in the following plunk by simply changing the directive's name first letter:
http://plnkr.co/edit/oL1WLw

gkalpak:
While directives are not allowed to request both a new (normal) and an isolate scope on the same element, the relevant check was not performed correctly when the directive requesting the isolate scope had a lower priority and specified a `templateUrl`.
In that case, the check was deferred until the template was fetched, but the info about other directives requesting a new (normal) scope was not properly preserved (in `previousCompileContext`).

This commit fixes this, so now an error is thrown (as expected).

Fixes #12215

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12217)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1654,7 +1654,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       previousCompileContext = previousCompileContext || {};
 
       var terminalPriority = -Number.MAX_VALUE,
-          newScopeDirective,
+          newScopeDirective = previousCompileContext.newScopeDirective,
           controllerDirectives = previousCompileContext.controllerDirectives,
           newIsolateScopeDirective = previousCompileContext.newIsolateScopeDirective,
           templateDirective = previousCompileContext.templateDirective,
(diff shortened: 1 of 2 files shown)


Narretz:
We should really fix this. This might be a breaking change for some apps, although it's technically a fix. If you didn't know you were not supposed to use it this way, you'll simply experience your app breaking.

petebacondarwin:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/angular/angular.js/12217)

petebacondarwin:
Reviewed 1 of 2 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/angular/angular.js/12217)

petebacondarwin:
LGTM

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/angular/angular.js/12217)

petebacondarwin:
I don't think it would be a breaking  change, since it is invalid to have new scope and isolate scope together.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/angular/angular.js/12217)

Issue #9485, opened 2014-10-08
Cannot read property 'ng339' of undefined error inside ng-switch-when and directive on 1.2.25


p3rsist:
I have a directive inside an ng-switch-when directive. When I do some changes to the scope models inside that directive (saving data to the server and updating the models) and I switch view (going to another ng-switch view) and go back again to that same ng-switch-when I get this ng339 error.

seaside98:
I am getting this same error in 1.4.1.

lgalfaso:
@seaside98 would it be possible for you to create a plunker with the error?

seaside98:
@lgalfaso It seems to be a bit too complex for me to replicate outside of my app, and I can't just paste the app because it is a private project. I will keep working on the plunker to see if I can replicate it. 

If it helps at all, here is the full error log:
```
[Error] Error: undefined is not an object (evaluating 'element.ng339')
jqLiteRemoveData@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:2807:26
jqLiteDealoc@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:2764:41
jqLiteRemove@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:2960:30
https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:3503:19
https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:5218:25
push@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:4832:44
leave@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:5217:35
ngSwitchWatchAction@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:27696:68
$digest@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:15667:31
$apply@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:15935:31
https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:17691:42
completeOutstandingRequest@https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:5387:15
https://ajax.googleapis.com/ajax/libs/angularjs/1.4.1/angular.js:5659:33
	(anonymous function) (angular.js, line 12314)
[...]

PR #12226, opened 2015-06-29
fix(jqLite): return if element is undefined

seaside98:
ngAnimate has problems removing elements, causing errors. A detailed explanation of this can be found at #8796.

See #9485

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12226)

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -315,6 +315,9 @@ function jqLiteOff(element, type, fn, unsupported) {
 }
 
 function jqLiteRemoveData(element, name) {
+  if(element == undefined) {
+    return;
+  }
   var expandoId = element.ng339;
   var expandoStore = expandoId && jqCache[expandoId];
 
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -315,7 +315,7 @@ function jqLiteOff(element, type, fn, unsupported) {
 }
 
 function jqLiteRemoveData(element, name) {
-  if(element == undefined) {
+  if (element === null) {
     return;
   }
   var expandoId = element.ng339;


seaside98:
I signed it!

lgalfaso:
Without a way to know how the original issue happen, I am inclined not to merge this PR.

@matsko do you have enough information to know if this is still an issue with 1.4.x?

seaside98:
How am I the only one dealing with this? I just started a new, completely unrelated project, and I'm getting ng339 errors. The only similarity I can find is that they both have long (300+ items) ```ng-repeat```s.

seaside98:
@lgalfaso @matsko This bug is breaking yet another one of my apps. I have uploaded some code to jsbin for demonstration purposes. Basically, I am just ng-repeating a large list and then rapidly changing the values of the list. Obviously this isn't how I am encountering the bug in my apps, but it is the easiest way to demonstrate the problem. This bug is quite annoying to replicate, sometimes the app breaks on my first try, other times it will go for several minutes with no problems at all. 

http://jsbin.com/vayade/edit?html,js,output You will probably have to rerun the script a number of times to get the bug to show up (```b.ng339 does not exist```). 

Also, I am using **Safari 8.0.8**. I have not been able to replicate this issue in Chrome, though I don't spend nearly as much time with Chrome.

lgalfaso:
@seaside98 run the script at the jsbin ~50 times and not even once the error show up. Is there some specific step that needs to happen for this error to trigger?

seaside98:
Luck, really. I've found that lag helps.

In the meantime, I will try to build something that will break more consistently. I don't you if you have already read this, but if you haven't, the following is a detailed technical explanation of the bug. https://github.com/angular/angular.js/issues/8796

td-edge:
@seaside98 You're not the only one dealing with this issue. A co-worker and I just stumbled across it with similar specifications. We are using angular 1.4.4 with ng-repeat of a long list and get the error randomly. The main development is being completed and checked in safari. I can post our stack trace if you want.

seaside98:
@td-edge After updating to OS X El Capitan, I haven't seen the problem. The only pattern I can seem to find is that somehow the safari dev console must be causing enough lag to make this bug show up. Glad to hear I'm not the only one.

Narretz:
Closing, see https://github.com/angular/angular.js/issues/9485#issuecomment-170402817

Issue #11580, opened 2015-04-14
ngAria : role="button" added with ng-click directive cannot be disabled and is dangerous


wayofspark:
Automatically adding unwanted role="button" on elements with ng-click directives can be very dangerous with some screen reader's behaviour. This ngAria feature cannot be disabled in ngAria config, which force developper to add a role attribute to element with ng-click when they dont want ngAria to add the button role to their clickable elements.

Narretz:
Can you explain why this is dangerous?

marcysutton:
Some specific examples would help. There is a node blacklist preventing it from being added to certain elements, which can be expanded. I'd argue it's more dangerous to put ng-click on elements that aren't interactive, and it isn't difficult to override the role.

wayofspark:
Sure, I can explain some specific example 

We had the issue with a a list of radio button :
Each radio is composed of a div container, with an ng-click handler containing the radio button itself. (The ng-click handler was here for mobile device purpose, to increase the click zone to select the radio buttons - ng-click='set-model-to-value-of-radio, basically).

With this configuration ( <div role=button> <label>blalbla</label> <input type=radio/> <div> ) VoiceOver (and maybe NVDA too, but not 100% sure) simply dont understand there is nested elements, stoping at the button role, with make the radio list description broken on the screen reader.

Sure, adding role="presentation" to the div do fix the issue. But honestly, I could find 100 more case where I can use a click handler on an element with simply doesnt match the "button" role, and forcing the developer to verbosely add a dummy role simply to say "hey, plz, do not mess around with inappropriate button role" seems unecessary.

Allowing the feature to be disabled ( it's the ONLY one ngAria doesnt allow to disable ) seems... legitimate imho.

marcysutton:
Cool, thanks for the information. We can totally make that configurable.

OlsonDev:
I just ran into this myself. Until those commits are available on bower, I threw a `role="presentation" tabindex="-1"` on my container. Is there a declarative way to change the focus to my `<md-switch>` inside the container like a native `<label for="some-id">` would do? Or do I have to write some JavaScript to do that in my `ng-click` handler?

In case my question isn't clear, this is a snippet I'm messing around with:
```html
<div layout ng-click="fullscreen = !fullscreen;" role="presentation" tabindex="-1">
    <div flex layout layout-align="start center"><md-icon class="md-primary">fullscreen</md-icon> Fullscreen</div>
[...]

Issue #11174, opened 2015-02-25
ngAria module and $ariaProvider configurations are global


sriprabha:
Currently, when I inject ngAria dependency for one view in the application, it defaults to injecting it across the application and adds all the defaults like tabindex to the HTML semantics. 
When I configure ngAria further using $ariaProvider to NOT add the tabindex (for test purpose), the configuration gets applied at other places as well where the dependency on ngAria was not injected. Making ngAria or $ariaProvider more private scoped may provide more flexibility to a developer to provide accessibility support.

marcysutton:
I wonder if a view-specific override system would work. On Angular Material, we're going to start checking the `$ariaProvider` to make sure we don't override it. We'd have to take things like that into account when building a more locally-scoped config. https://github.com/angular/material/issues/600

PR #12234, opened 2015-06-30
feat(ngAria): add option to disable role=button

marcysutton:
I am separating a larger ngAria refactor into smaller pieces and getting them submitted! Starting with a config option for `role="button"`, which is added by ngClick onto divs or custom elements like `<radio-group>`. If `bindRoleForClick` is set to false in the config, ngAria will not add the button role to custom elements using `ng-click`. I'm totally open to renaming the config variable.

Closes https://github.com/angular/angular.js/issues/11580

Note config options are app-wide–an awesome improvement would be to scope them by controllers or directives. Related to #11174.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12234)

### DIFF
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -750,6 +750,18 @@ describe('$aria', function() {
     });
   });
 
+  describe('actions when bindRoleForClick is set to false', function() {
+    beforeEach(configAriaProvider({
+      bindRoleForClick: false
+    }));
+    beforeEach(injectScopeAndCompiler);
+
+    it('should not add a button role', function() {
+      compileElement('<radio-group ng-click="something"></radio-group>');
+      expect(element.attr('role')).toBeUndefined();
+    });
+  });
+
   describe('actions when bindKeypress is set to false', function() {
     beforeEach(configAriaProvider({
       bindKeypress: false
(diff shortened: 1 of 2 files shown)


test/ngAria/ariaSpec.js
```diff
@@ -750,6 +750,18 @@ describe('$aria', function() {
...
+      beforeEach(configAriaProvider({
+          bindRoleForClick: false
+      }));
+      beforeEach(injectScopeAndCompiler);
+
+      it('should not add a button role', function() {
+        compileElement('<radio-group ng-click="something"></radio-group>');
+        expect(element.attr('role')).toBeUndefined();
+      });
+  });
```
gkalpak: Indentation :)

marcysutton: It matches the rest of the file. I didn't want to introduce whitespace changes, especially to make the whole file consistent. But if it's important, I can....

gkalpak: Inside the function you are using 4 spaces (most of the time). It should be 2 spaces (according to Angular's "styleguide".
(I don't think it's a big deal, but I have a thing about whitespace inconsistency :P)

gkalpak:
LGTM (other than the whitespace nitpick) as long as Travis is happy.

marcysutton:
@gkalpak I fixed the whitespace, looks like Travis quit whining about whatever it was failing on. It looked very unrelated to these changes when I dug into them.

Issue #12197, opened 2015-06-23
Support the (event) / on-event syntax in AngularJS


mhevery:
# Goal

In order to support the [ng2migration](https://docs.google.com/document/d/19NpH_Fzr-omI_Qm6lUt4mZW0L_f6Ksil4pTTp47AhYk/edit#) effort the AngularJS needs to learn how to process the Angular 2 event binding syntax

# Syntax Variants

* `(event)` / `(^event)`
* `on-event` / `onbubble-event`

# Semantics

Angular should process this markup directly in the compiler by attaching the appropriate event to the DOM without any need for a directive.

IgorMinar:
per discussion in https://docs.google.com/document/d/1xvBZoFuNq9hsgRhPPZOJC-Z48AHEbIBPlOCBTSD8m0Y/edit# I'm going to close this issue for now so that we don't give wrong impression as to what is actually going to happen. The main problem with adding this syntax into Angular 1 core is that it fragments the api/docs/etc and exposes large api surface that we'd need to maintain - which is especially challenging because of the semantical differences between Angular 1 and 2.

We can reopen this issue in the future if needed.

PR #12236, opened 2015-06-30
feat(compile): add support for Angular 2 event bindings

wesleycho:
- Adds support for event bindings to match Angular 2, i.e. `(event)`, `(^event)`, `on-event`, and `onbubble-event`

This implements #12197.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12236)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -728,7 +728,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       COMMENT_DIRECTIVE_REGEXP = /^\s*directive\:\s*([\w\-]+)\s+(.*)$/,
       CLASS_DIRECTIVE_REGEXP = /(([\w\-]+)(?:\:([^;]+))?;?)/,
       ALL_OR_NOTHING_ATTRS = makeMap('ngSrc,ngSrcset,src,srcset'),
-      REQUIRE_PREFIX_REGEXP = /^(?:(\^\^?)?(\?)?(\^\^?)?)?/;
+      REQUIRE_PREFIX_REGEXP = /^(?:(\^\^?)?(\?)?(\^\^?)?)?/,
+      EVENT_DIRECTIVE_REGEXP = /\(\^?([\w]+)\)$/,
+      ON_EVENT_DIRECTIVE_REGEXP = /^on(?:bubble)?-([\w]+)$/;
 
   // Ref: http://developers.whatwg.org/webappapis.html#event-handler-idl-attributes
   // The assumption is that future DOM event attribute names will begin with
@@ -1491,7 +1493,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               directiveNormalize(nodeName_(node)), 'E', maxPriority, ignoreDirective);
 
           // iterate over the attributes
-          for (var attr, name, nName, ngAttrName, value, isNgAttr, nAttrs = node.attributes,
+          for (var attr, name, nName, ngAttrName, value, isNgAttr, isEventAttr, nAttrs = node.attributes,
                    j = 0, jj = nAttrs && nAttrs.length; j < jj; j++) {
             var attrStartName = false;
             var attrEndName = false;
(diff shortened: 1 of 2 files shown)


test/ng/compileSpec.js
```diff
@@ -7524,6 +7516,35 @@ describe('$compile', function() {
...
+
+      $rootScope.clicked = false;
+      element = $compile('<div on-click="onClick()"></div>')($rootScope);
+      browserTrigger(element, 'click');
+      expect($rootScope.clicked).toBe(true);
+
+      $rootScope.clicked = false;
+      element = $compile('<div onbubble-click="onClick()"></div>')($rootScope);
+      browserTrigger(element, 'click');
+      expect($rootScope.clicked).toBe(true);
```
LeonardoBraga: In order to make it DRY, you could refactor the tests to:

```js
var testEventBinding = function(template) {
  $rootScope.clicked = false;
  element = $compile(template)($rootScope);
  browserTrigger(element, 'click');
  expect($rootScope.clicked).toBe(true);
};

testEventBinding('template 1');
testEventBinding('template 2');
...
testEventBinding('template N');
```

test/ng/compileSpec.js
```diff
@@ -7184,14 +7184,6 @@ describe('$compile', function() {
           "Please use the ng- versions (such as ng-click instead of onclick) instead.");
     }));
 
-    it('should pass through arbitrary values on onXYZ event attributes that contain a hyphen', inject(function($compile, $rootScope) {
```
gkalpak: What was the purpose of this test ?

wesleycho: If I had to guess, that `on-` were valid attributes, although this test in itself is a bit weird since I don't think it should have had the opinionation of it testing interpolation.

The test conflicts with the new binding syntax though, since the expected behavior is to accept an expression as opposed to an interpolation.

src/ng/compile.js
```diff
@@ -1509,6 +1511,17 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
                 });
             }
 
+            // support (event), (^event), on-event, and onbubble-event bindings
```
caitp: `on-event` potentially breaks existing apps. Do we want to do that?

caitp: Maybe a workaround for this is to ignore attributes which are present in the bindings object

src/ng/compile.js
```diff
@@ -728,7 +728,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       COMMENT_DIRECTIVE_REGEXP = /^\s*directive\:\s*([\w\-]+)\s+(.*)$/,
       CLASS_DIRECTIVE_REGEXP = /(([\w\-]+)(?:\:([^;]+))?;?)/,
       ALL_OR_NOTHING_ATTRS = makeMap('ngSrc,ngSrcset,src,srcset'),
-      REQUIRE_PREFIX_REGEXP = /^(?:(\^\^?)?(\?)?(\^\^?)?)?/;
+      REQUIRE_PREFIX_REGEXP = /^(?:(\^\^?)?(\?)?(\^\^?)?)?/,
+      EVENT_DIRECTIVE_REGEXP = /\(\^?([\w]+)\)$/,
```
caitp: `\w` will break many "custom" dom events, like jQuery namespaced events, or special ones that people define for themselves

caitp: I guess `-` isn't really a problem since these are normalized already? but `.` seems like a big one for the jquery people

src/ng/compile.js
```diff
@@ -2455,6 +2469,24 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     }
 
 
+    function addEventBindingDirective(directives, eventName, isEventDirective) {
+      if (!isEventDirective) return;
+
+      directives.push({
+        compile: function(tElement, tAttr) {
```
caitp: doing this in compile seems dangerous for directives which use `replace`? Or element transclusion?

wesleycho: The main logic is in the linking function - should I be caching the parsed expression in the linking function then?

caitp: oh you're right, didn't notice you were returning a function. Regardless, I was speaking with Igor about this a moment ago and I think we're going to avoid bringing this into angular 1.x, to minimize confusion, and in my view avoid the edge cases which is a big problem

wesleycho: Gotcha - feel free to close this then if it is decided it is out of scope.

caitp: yeah I think we're going to close this, although I believe Igor was going to do it. Well don't delete the branch just in case minds are changed later on =)

IgorMinar:
I closed the original issue because we'd like to implement an alternative approach instead. Rather than making Angular 1 look like Angular 2, we'd like to make it possible to run Angular 1 in Angular 2 and vice versa. See: https://docs.google.com/document/d/1xvBZoFuNq9hsgRhPPZOJC-Z48AHEbIBPlOCBTSD8m0Y/edit#

Issue #12220, opened 2015-06-26
Allow cached requests to bypass sce validation in $templateRequest


IgorMinar:
Once #12219 is implemented, we will be able to bypass sce validation for requests that are already in the template cache.

This is desirable because it allows developers to disable the whitelist, especially remove the 'self' rule for allowing requests to the same domain origin to make it through. The same origin requests are currently allowed by default, which is problematic for applications that contain redirection services (e.g. for tracking clicks to 3rd party sites).

Doing it in this way, would allow us to be backwards compatible for application that populate the template cache as part of the build, while allowing us to ban the whitelist in these applications.

We need to backport this change to the 1.3.x branch as well.

Issue #12219, opened 2015-06-26
Move sce validation from $compile/$route/ngInclude into $templateRequest


IgorMinar:
The objectives are:
- move the responsibility to the more correct place
- reduce the duplication of code
- make it possible to fetch templates from cache before even validating $sce

Currently the sce testing logic is all over the place.

One complication I see is how the ngInclude is watching the expression and validating the result. If I remember correctly, this was done for performance reason, so we should be careful in that area.

PR #12240, opened 2015-06-30
refactor($templateRequest): move $sce checks and trust the cache

rjamet:
Move all the calls to $sce.getTrustedUrl inside $templateRequest, and
also trust the contents of the cache. This allows prefetching templates
and to bypass the checks on templates directly populated in the cache, 
like for instance templates specified in script tags.

Fixes both #12219 and #12220.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12240)

### DIFF
--- a/src/ng/templateRequest.js
+++ b/src/ng/templateRequest.js
@@ -7,10 +7,12 @@ var $compileMinErr = minErr('$compile');
  * @name $templateRequest
  *
  * @description
- * The `$templateRequest` service downloads the provided template using `$http` and, upon success,
- * stores the contents inside of `$templateCache`. If the HTTP request fails or the response data
- * of the HTTP request is empty, a `$compile` error will be thrown (the exception can be thwarted
- * by setting the 2nd parameter of the function to true).
+ * The `$templateRequest` service runs security checks then downloads the provided template using
+ * `$http` and, upon success, stores the contents inside of `$templateCache`. If the HTTP request
+ * fails or the response data of the HTTP request is empty, a `$compile` error will be thrown (the
+ * exception can be thwarted by setting the 2nd parameter of the function to true). Note that the
+ * contents of `$templateCache` are trusted, so the call to `$sce.getTrustedUrl(tpl)` is omitted
+ * when `tpl` is of type string and `$templateCache` has the matching entry.
  *
  * @param {string|TrustedResourceUrl} tpl The HTTP request template URL
  * @param {boolean=} ignoreRequestError Whether or not to ignore the exception when the request fails or the template is empty
--- a/src/ng/templateRequest.js
+++ b/src/ng/templateRequest.js
@@ -12,7 +12,7 @@ var $compileMinErr = minErr('$compile');
  * of the HTTP request is empty, a `$compile` error will be thrown (the exception can be thwarted
  * by setting the 2nd parameter of the function to true).
  *
- * @param {string} tpl The HTTP request template URL
+ * @param {string|TrustedResourceUrl} tpl The HTTP request template URL
  * @param {boolean=} ignoreRequestError Whether or not to ignore the exception when the request fails or the template is empty
  *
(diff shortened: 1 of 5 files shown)


src/ng/templateRequest.js
```diff
@@ -12,7 +12,7 @@ var $compileMinErr = minErr('$compile');
  * of the HTTP request is empty, a `$compile` error will be thrown (the exception can be thwarted
  * by setting the 2nd parameter of the function to true).
  *
- * @param {string} tpl The HTTP request template URL
+ * @param {string|TrustedResourceUrl} tpl The HTTP request template URL
```
lgalfaso: nit: an update on the description would be nice

src/ngRoute/route.js
```diff
@@ -591,9 +591,8 @@ function $RouteProvider() {
                 if (angular.isFunction(templateUrl)) {
                   templateUrl = templateUrl(nextRoute.params);
                 }
-                templateUrl = $sce.getTrustedResourceUrl(templateUrl);
```
lgalfaso: Is this a behavior change?

lgalfaso:
Overall, looks good. If possible, would like to see some numbers to know if this PR has any performance implications

src/ngRoute/route.js
rjamet: - If the templateUrl is a string in the cache, then it is a behavior change, as it will not call getTrustedResourceUrl, but that's the point. 
- If the templateUrl is not a string or not in the cache, the call to valueOf below unwraps trusted types without side-effects, and we're just moving the call to getTrustedResourceUrl inside the call to templateRequest two lines below. You could maybe build a templateUrl equal to $sce.trustAs(SCE_CONTEXTS.RESOURCE_URL, undefined) which will get in the if block after the patch but not before, but the default $sce.trustAs implementation does not allow that, and it would only fail to load anyways.

rjamet:
I updated the description as you asked. Regarding performance, the only impact I see is that we will do more cache lookups and less white/blacklisting, but I'm not sure of the overall effect or how to measure it.

test/ng/compileSpec.js
```diff
@@ -1296,14 +1296,24 @@ describe('$compile', function() {
...
+          function($compile, $httpBackend, $rootScope, $sce) {
             expect(function() {
-              $templateCache.put('http://example.com/should-not-load.html', 'Should not load even if in cache.');
               $compile('<div class="crossDomainTemplate"></div>')($rootScope);
             }).toThrowMinErr('$sce', 'insecurl', 'Blocked loading resource from url not allowed by $sceDelegate policy.  URL: http://example.com/should-not-load.html');
           }
         ));
 
+        it('should trust what is already in the template cache', inject(
+          function($compile, $httpBackend, $rootScope, $templateCache, $sce) {
```
gkalpak: Is `$sce` needed here ?

test/ng/compileSpec.js
```diff
@@ -1296,14 +1296,24 @@ describe('$compile', function() {
         ));
 
         it('should not load cross domain templates by default', inject(
-          function($compile, $rootScope, $templateCache, $sce) {
+          function($compile, $httpBackend, $rootScope, $sce) {
```
gkalpak: Are `$httpBackend` and `$sce` needed here ?

rjamet: They are both useless, you are right. Removed.

test/ng/compileSpec.js
rjamet: It's not, I removed it.

wesleycho:
@lgalfaso this likely has performance implications on `ng-include`, as the `$watch` is now watching an object as opposed to a string - not sure if it's considered a huge issue, but is worth noting.

IgorMinar:
@wesleycho, @lgalfaso I don't expect any significant perf impact. the watch in ngInclude should not cause problems because ngInclude doesn't do any deep watching.

PR #12245, opened 2015-06-30
Fixes to body element animation

matsko:
[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12245)

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -25,12 +25,12 @@ describe("animations", function() {
       });
     });
 
-    inject(function($animate, $rootScope, $document) {
+    inject(function($animate, $rootScope, $$body) {
       $animate.enabled(true);
 
       element = jqLite('<div></div>');
 
-      $animate.enter(element, jqLite($document[0].body));
+      $animate.enter(element, $$body);
       $rootScope.$digest();
 
       expect(capturedAnimation).toBeTruthy();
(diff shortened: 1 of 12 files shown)


lgalfaso:
I am slightly inclined into making the new `$body` service an undocumented service. That a side, LGTM

test/ngAnimate/animateSpec.js
```diff
@@ -1209,6 +1209,34 @@ describe("animations", function() {
...
+        return function(element, method, options) {
+          capturedAnimation = arguments;
+          return new $$AnimateRunner();
+        };
+      });
+    });
+    inject(function($animate, $rootScope, $document, $rootElement) {
+      $animate.enabled(true);
+
+      var body = jqLite($document[0].body);
```
Narretz: Does it make sense to use the new body service here instead?

Narretz:
I also think $body should be $$body.

matsko:
@petebacondarwin please review so we can merge this in tomorrow.

petebacondarwin:
It looks fine to me. The only comment I have is that since `$$body` is only used once in the core `ng` module, we should move the definition of `$$body` over to the `ngAnimate` module so we don't need to make `angular.js` file any bigger.

matsko:
Merged in as https://github.com/angular/angular.js/commit/44ce9c8288fc6c12043567027271a09bd0594d74. Excellent work @OlenDavis.

PR #12247, opened 2015-06-30
docs(guide/Services): forgot "ng-app"

christianreed:
Compiler won't run on this code if you type it out yourself.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12247)

### DIFF
--- a/docs/content/guide/services.ngdoc
+++ b/docs/content/guide/services.ngdoc
@@ -32,7 +32,7 @@ subsystem takes care of the rest.
 
 <example module="myServiceModule">
   <file name="index.html">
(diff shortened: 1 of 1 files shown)


christianreed:
I signed it!

Narretz:
Hi, thanks for the PR. This is a little problematic, because when you open the examples in the plnkr, there will be two ng-app directives. It doesn't hurt, but it's also not very pretty. This is because the example generator will add ng-app automatically to the example. Since you have to add html boilerplate to the examples anyway when you copy / paste them, I think we should not add ng-app explicitly to these examples. But thanks for your effort!

PR #12248, opened 2015-06-30
Synchronous Classes in ngAnimate

### LINKED ISSUE #11975: Animations should not cancel parent animations
1.4 has broken some important animations in my application. Apparently this is because ng-enter animations have been modified to cancel running ng-class animations on their parents, for some insane reason?

I have a container element with a router ui-view nested inside it. The container has an ng-class that sets 'active' on the container when a particular route becomes active, which is then used to slide the container into view. The ui-view element itself has an ng-enter animation that transitions its opacity.

In 1.3 this all worked perfectly. In 1.4 it is broken, because the framework apparently thinks that, because I'm animating the opacity of one element, I mustn't want the position animation—that I have *explicitly specified*—on its parent to run. In what universe is this desirable behaviour?

### LINKED ISSUE #12276: ngAnimate issue on ng-show with a nested ng-repeat
Hi, i was actually helping on stackoverflow for some issues and i ran into this one : 
http://stackoverflow.com/questions/31245780/parent-child-animation-unexpected-different-angular-1-3-to-1-4/

After some try i was able to figure out where was the trouble : 

On "ng-show/hide" with a nested "ng-repeat" there is no class "ng-hide-animate" added on the element during the animation.

This doesn't work on 1.4.0 - 1.4.1 - 1.4.2: http://codepen.io/anon/pen/EjEora
This works on 1.3.5 : http://codepen.io/anon/pen/VLXygP

Not having this class cause this to be applied during the animation.

.ng-hide:not(.ng-hide-animate) {
  display: none !important;
}

Which result into not hiding the element before the animation end.

I'm also a bit confused about the documentation : 

**1.4.X documentation**
> CSS Class-based Animations
> Class-based animations (animations that are triggered via ngClass, ngShow, ngHide and some other > directives) have a slightly different naming convention. Class-based animations are basic enough that a standard transition or keyframe can be referenced on the class being added and removed.

> For example if we wanted to do a CSS animation for ngHide then we place an animation on the .ng-> hide CSS class:

The 1.4.X documentation specify that we shoudn't have this "ng-hide-animate". So i guess that the ng-hide:not(.ng-hide-animate) in the CSS shoudn't exist.

What do you think about this ? Is that a bug or just a miss-use ?

matsko:
This PR ensures that classes added via $animate are always applied after the first digest instead of after the first requestAnimationFrame.

Closes #11975
Closes #12276


IgorMinar:
reviewed with Matsko. once all tests and demos are passing then this LGTM

matsko:
@petebacondarwin could you please review?

matsko:
Landed as https://github.com/angular/angular.js/commit/32d3cbb3aadf71492102f9318fcac570fb60bef8

petebacondarwin:
Great!

PR #12258, opened 2015-07-02
revert: "fix($compile): do not write @-bound properties if attribute …

jeffbcross:
…is not present"

This reverts commit 8a1eb1625c080445ce1e519762e1f2d4fd842b72.

This commit broke the tabs component on the material project,
which caused internal breakages. Will open a separate issue to
look into the issue.


gkalpak:
FYI, I think the reverted commit fixed other issues in `ngMaterial` caused by a previous commit.
Maybe this needs closer investigation.

Issue #11374, opened 2015-03-19
[ngAria] aria-required only applied when ng-model has $error.required


dpogue:
Based on a Twitter conversation with @marcysutton: https://twitter.com/dpogue/status/569999089967849472

If I have `<input type="text" ng-model="username" name="username" required>` and I'm using ngAria then it should have the `aria-required` attribute added to it.

This currently only happens when the field doesn't have content (i.e., when `username.$error.required` is true):
https://github.com/angular/angular.js/blob/master/src/ngAria/aria.js#L292-L298

marcysutton:
A native input doesn't need `aria-required`. Custom elements do, though. If it was `<md-checkbox role="checkbox" required>`, `aria-required` would need to be added. That said, I'm actually seeing the attribute already on a custom element with `required`: http://codepen.io/marcysutton/pen/wBwVqM

Narretz:
So this works as expected @marcysutton?

marcysutton:
Unless I'm missing something, it looked like it worked fine to me. @dpogue can you elaborate? Is the issue happening when `ng-model` is set up in a specific way? Can you provide a Codepen or Plunker that shows the issue?

marcysutton:
@dpogue can you reply to this issue? Otherwise we'll close it as not a bug.

dpogue:
My concern was just that the `aria-required` attribute was being added and removed based on the error state of the ngModel, instead of being added once and maintained. If that's not an issue, then this can be closed.

Narretz:
According to this mdn example, it looks like `aria-required` is always present: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute

So, for non-native inputs (controls) with ng-model that have a required attribute should this mean?
- if they have only the `required` attribute, ngAria should add `aria-required` once.
- if they have ngRequired, they should watch that and toggle aria-required appropriately.

marcysutton:
That sounds right. Upon looking closer, I found that `aria-required` in ngAria is super buggy for expected use cases (custom elements with roles). Depending on whether the custom control has `type="checkbox"` (or other form type) or `role="checkbox"` (which it needs for assistive tech), it can get out of sync with `required`. For example:
```
<some-checkbox type="checkbox" ng-required="true" ng-model="data">
 Checkbox
</some-checkbox>
```
[...]

PR #12262, opened 2015-07-02
fix(ngAria): clean up tabindex usage

### LINKED ISSUE #11500: ngAria: tabIndex unnecessarily added to buttons & other interactive elements
As it is currently written, ngAria is a bit too generous about adding `tabindex="0"` to elements. It adds the attribute to any element with `ngModel` or `ngClick` without checking whether it should actually do that. This includes buttons, inputs and other elements that might not need it. Adding unnecessary `tabindex` contributes to code bloat and is generally a poor practice.

marcysutton:
- Do not put `tabindex` on natively interactive controls using `ng-model` or `ng-click`
- Use a single `nodeBlacklist` to limit which elements receive support (not user-configurable)

Closes #11500

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -53,6 +53,16 @@ var ngAriaModule = angular.module('ngAria', ['ng']).
                         provider('$aria', $AriaProvider);
 
 /**
+* Internal Utilities 
+*/
+var nodeBlackList = ['BUTTON', 'A', 'INPUT', 'TEXTAREA', 'SELECT', 'DETAILS', 'SUMMARY'];
+
+var isNodeOneOf = function (elem, nodeTypeArray) {
+  if (nodeTypeArray.indexOf(elem[0].nodeName) !== -1) {
+    return true;
+  }
+}
+/**
  * @ngdoc provider
  * @name $ariaProvider
  *
@@ -113,10 +123,10 @@ function $AriaProvider() {
     config = angular.extend(config, newConfig);
   };
 
-  function watchExpr(attrName, ariaAttr, negate) {
+  function watchExpr(attrName, ariaAttr, nodeBlackList, negate) {
     return function(scope, elem, attr) {
       var ariaCamelName = attr.$normalize(ariaAttr);
-      if (config[ariaCamelName] && !attr[ariaCamelName]) {
+      if (config[ariaCamelName] && !isNodeOneOf(elem, nodeBlackList) && !attr[ariaCamelName]) {
         scope.$watch(attr[attrName], function(boolVal) {
           // ensure boolean value
           boolVal = negate ? !boolVal : !!boolVal;
@@ -125,7 +135,6 @@ function $AriaProvider() {
       }
     };
   }
-
   /**
    * @ngdoc service
    * @name $aria
@@ -184,10 +193,10 @@ function $AriaProvider() {
 
 
 ngAriaModule.directive('ngShow', ['$aria', function($aria) {
-  return $aria.$$watchExpr('ngShow', 'aria-hidden', true);
+  return $aria.$$watchExpr('ngShow', 'aria-hidden', [], true);
 }])
 .directive('ngHide', ['$aria', function($aria) {
-  return $aria.$$watchExpr('ngHide', 'aria-hidden', false);
+  return $aria.$$watchExpr('ngHide', 'aria-hidden', [], false);
 }])
 .directive('ngModel', ['$aria', function($aria) {
 
@@ -261,6 +270,9 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
                 scope.$watch(ngAriaWatchModelValue, shape === 'radio' ?
                     getRadioReaction() : ngAriaCheckboxReaction);
               }
+              if (needsTabIndex) {
+                elem.attr('tabindex', 0);
+              }
               break;
             case 'range':
               if (shouldAttachRole(shape, elem)) {
@@ -289,16 +301,15 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
                   });
                 }
               }
+              if (needsTabIndex) {
+                elem.attr('tabindex', 0);
+              }
               break;
             case 'multiline':
               if (shouldAttachAttr('aria-multiline', 'ariaMultiline', elem)) {
                 elem.attr('aria-multiline', true);
               }
               break;
-          }
-
-          if (needsTabIndex) {
-            elem.attr('tabindex', 0);
           }
 
           if (ngModel.$validators.required && shouldAttachAttr('aria-required', 'ariaRequired', elem)) {
@@ -322,7 +333,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
   };
 }])
 .directive('ngDisabled', ['$aria', function($aria) {
-  return $aria.$$watchExpr('ngDisabled', 'aria-disabled');
+  return $aria.$$watchExpr('ngDisabled', 'aria-disabled', []);
 }])
 .directive('ngMessages', function() {
   return {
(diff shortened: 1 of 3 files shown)


src/ngAria/aria.js
```diff
@@ -126,6 +128,11 @@ function $AriaProvider() {
     };
   }
 
+  function isNodeOneOf(elem, nodeTypeArray) {
```
marcysutton: This function used to live inside of the `ngClick` override directive below. This exposes it to the entire module.

src/ngAria/aria.js
```diff
@@ -322,7 +333,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
   };
 }])
 .directive('ngDisabled', ['$aria', function($aria) {
-  return $aria.$$watchExpr('ngDisabled', 'aria-disabled');
+  return $aria.$$watchExpr('ngDisabled', 'aria-disabled', []);
```
marcysutton: `[]` here is a placeholder for forthcoming `ng-disabled` improvements. This PR is purposefully limited.

test/ngAria/ariaSpec.js
```diff
@@ -640,26 +665,6 @@ describe('$aria', function() {
       compileElement('<div ng-dblclick="someAction()" tabindex="userSetValue"></div>');
       expect(element.attr('tabindex')).toBe('userSetValue');
     });
-
-    it('should set proper tabindex values for radiogroup', function() {
```
marcysutton: I decided Angular shouldn't be in the business of fixing custom radiogroups...it's too big of a can of worms when it comes to expected keyboard behavior.

test/ngAria/ariaSpec.js
```diff
@@ -616,10 +616,35 @@ describe('$aria', function() {
...
+        $compile("<a ng-href='#/something'>")(scope),
+        $compile("<input ng-model='val'>")(scope),
+        $compile("<textarea ng-model='val'></textarea>")(scope),
+        $compile("<select ng-model='val'></select>")(scope),
+        $compile("<details ng-model='val'></details>")(scope)
+      ];
+      expectAriaAttrOnEachElement(element, 'tabindex', undefined);
+    });
+
+    it('should not attach to rando ng-model elements', function() {
```
gkalpak: 1. `rando`  -->  `random` (?)

2. Why not ? If they have `ngModel` then they are supposed to have two-way binding (aka expect user input), so why not get a tabindex ?

marcysutton: 1. "rando" is what I would say verbally...it was intentional, but could be changed as it is colloquial.
2. `tabindex` is only useful if it has interaction events as well, for which ngModel alone is not responsible. That's more of a job for ngClick (or similar), which does add tabindex. The whole problem with ngAria was that it was way too liberal with adding tabindex. This PR fixes that.

gkalpak:
> Use a single nodeBlacklist to limit which elements receive support (not user-configurable)

I don't know why/if it should not be user-configurable, but putting it onto the service makes it user-configurable.

--
Not sure if it's for better or worse, but you could use the `nodeBlacklist` and the `isNodeOneOf()` directly (not through the service). Just mentioning it...

marcysutton:
Good point...I swapped that around at the last minute and didn't think it through. Two things: 
 
1. I actually think it makes sense to expose the element config, but I've been warned about this before by @caitp. Once it's in there, it's really hard to remove. see https://github.com/angular/angular.js/pull/10766#discussion_r23463437.
1. This is a dumb question, but how can I expose those functions inside of ngAria without exposing them in the service or polluting the global scope?

patrickfox:
`tabindex="0"` is used to make something user focusable/tabbable, and only elements that are intended to be actionable should be able to receive this. Does it make sense to allow a random DIV to have `tabindex="0"` without it having a specified interactive `role`? Would it make sense to allow `tabindex="0"` on non-blacklisted elements only when they have a `role="{interactive role}"`or at least throw out a console.warn?

Interactive roles are:
- radio
- checkbox
- button
- link
- tab
- menuitem
- textbox
- menuitemcheckbox
- menuitemradio
- spinbox
- textbox
- option

gkalpak:
> I actually think it makes sense to expose the element config, but I've been warned about this before by @caitp. Once it's in there, it's really hard to remove. see #10766 (comment).

It is indeed difficult/undesirable to refactor something, once it's been made public. I would refrain from it for now and re-consider if/when we have several users asking for it. Especially, since `ngAria` seems to be about to get some refactoring and internal things might be more volatile right now.
In any case, if we decide to make it public, we should do it as a configuration option on the provider (not on the service) to avoid inconsistencies.

--
> [...] how can I expose those functions inside of ngAria without exposing them in the service or polluting the global scope?

All modules are wrapped in an IIFE, so you will not be polluting the global scope.

marcysutton:
@patrickfox so glad to hear your take on it! I definitely don't think `ngModel` should add `tabindex` by itself, so I restricted it to use cases with roles...perhaps those use cases could be expanded upon. I like your idea about adding tabindex for interactive roles, and that's pretty much how it works already: if `ngModel` matches the scenarios we want to backfill with support, `tabindex` is added if not already present. Interaction events are still required to take accessibility the rest of the way–perhaps missing events would be a good opportunity to log a warning, like in https://github.com/rackt/react-a11y or https://github.com/angular/material. I can tell you already that people really like disabling warnings. :)

For `ngClick` on a non-blacklisted element without a role, it adds `tabindex` and `role=button` (if config allows). I will admit it's pretty heavy-handed, but it addresses one of the biggest abuses in Angular (ngClick on DIV elements). ngAria should handle common use cases before tackling intricate ARIA patterns like tabs, and that one is pretty straight-forward.

If we take a step back and look at what people expect from ngAria, it's to get accessibility for free: e.g. "I include this module, it makes my app accessible". Just including the module isn't enough, however–most of the time you also have to add a `role` to tell ngAria to "fix" something. This restricts it to certain use cases, but makes it less "automagical", since you still have to know which role to choose. At that point, how much do we gain by adding `tabindex` when you're already on the hook to manage your own roles? Should we even encourage people to use interactive roles like `link` when the native `a` tag does the same job, but better? 

This last question I think about a lot when adding features to ngAria–I don't think an Angular module should encourage ARIA patterns like `role=radiogroup` when native radio inputs do the job so well, and you can style them. At some point I decided "fixing" those widgets with `tabindex` had diminishing returns. There is often too much interaction complexity for ngAria to anticipate everything, but we can consider common use cases, starting the list of interactive roles–many of which are already addressed.

marcysutton:
I should add, I have another PR forthcoming to limit ARIA usage on native controls, since right now they introduce redundancy. For example, native inputs currently get `aria-checked` from ngModel, when they shouldn't. That work will come after this PR but I'm definitely keeping it in mind as we go through this discussion.

marcysutton:
Okay, I finally got your feedback incorporated @gkalpak. I considered @patrickfox's points about interactive roles, but decided to keep the logic as-is. ngAria already checks for some of those roles under the hood for ngModel before adding tabindex (checkbox, radio, etc.)–it compares those roles against a `nodeBlacklist` including buttons, links, and form controls. Many custom ARIA patterns are simply too complex for ngAria to support, such as `radiogroup` and `menuitem`...so I left those out. Keep in mind the `aria-checked` refactor will be in a different PR, this is only concerning `tabindex`.

src/ngAria/aria.js
```diff
@@ -226,7 +235,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
           }
         },
         post: function(scope, elem, attr, ngModel) {
-          var needsTabIndex = shouldAttachAttr('tabindex', 'tabindex', elem);
+          var needsTabIndex = shouldAttachAttr('tabindex', 'tabindex', elem, nodeBlackList);
```
gkalpak: `shouldAttachAttr()` does not take `nodeBlackList` into account. I think ti should...
(Either take it into account or remove this 4th argument - to avoid confusion.)

test/ngAria/ariaSpec.js
```diff
@@ -616,10 +616,35 @@ describe('$aria', function() {
...
+        $compile("<a ng-href='#/something'>")(scope),
+        $compile("<input ng-model='val'>")(scope),
+        $compile("<textarea ng-model='val'></textarea>")(scope),
+        $compile("<select ng-model='val'></select>")(scope),
+        $compile("<details ng-model='val'></details>")(scope)
+      ];
+      expectAriaAttrOnEachElement(element, 'tabindex', undefined);
+    });
+
+    it('should not attach to random ng-model elements', function() {
```
gkalpak: :disappointed:

gkalpak:
It LGTM (with the exception of little comment regarding `shouldAttachAttr()`).

I am wondering if there should be a BREAKING CHANGE notice about not trying to `tabindex` radiogroup.

marcysutton:
@gkalpak I removed `nodeBlackList` from that `shouldAttachAttr()` call, since the directive is already checking for valid use cases before applying attributes.

The radiogroup reference was only in a test, there is no change to the code. So I don't think it's a breaking change.

gkalpak:
@marcysutton, I don't think the `ngModel` directive checks for valid usecases before adding a tabindex.
AFAIKT, it only checks the "shape", which does not account for elements in the `nodeBlackList`.
WDYT ?

--
> The radiogroup reference was only in a test, there is no change to the code.

Good point :smile: No breaking change then !

marcysutton:
The check that it does is a bit weird, to be honest...the fact that it checks for `type` OR `role` is an artifact of the incorrect functionality this refactor is trying to fix. ngAria added ARIA attributes to native elements that didn't need them, i.e. `<input>`, and did so by checking for types. I'm thinking that `type` should just go away, since we are factoring the native input stuff out completely, but that is a pretty major breaking change. 

It reminds me of the bugs I uncovered with `aria-required` in https://github.com/angular/angular.js/issues/11374#issuecomment-128749894: ngAria behaves differently when elements use `type` vs. `role`, the latter of which is what's actually valid and needed for custom controls. 

Do you think we can kill the `type` checks and streamline ngAria to hinge off of `role` attributes and tagNames instead?

gkalpak:
I am not sure how much of a breaking change it would be. Since this is on the 1.4.x milestone, we shouldn't introduce significantly different behaviour.

If you think it will not break stuff (much), I'd say go for it. Else, let's improve what can be "backwards compatibly" improved and defer bigger changes to 1.5.x.

marcysutton:
Ok, I think it's worth keeping it as-is for now and swapping the `type` check for the `nodeBlackList` later, in 1.5x. Keeping in mind there is another PR coming for `aria-checked` that could warrant that change earlier....might as well let this work go forward.

gkalpak:
@marcysutton, is this PR ready for final review/merging ?

marcysutton:
@gkalpak yup! I think it should stay as-is until the next PR, which I can start as soon as this is merged.

gkalpak:
LGTM

petebacondarwin:
LGTM

PR #12264, opened 2015-07-03
docs($cookiesProvider): fixed a typo

ashishdasnurkar:
"This is _import_ so that cookies will be visible" should be "This is _important_ so that cookies will be visible"


ashishdasnurkar:
I signed it

Issue #12249, opened 2015-07-01
ng-form  cannot show message & remove item from arry


kimx:
Hello 
when I upgrade version from 1.3.16 to 1.4.1 ,I occur some errors.
* ng-message cannot show， But if I downgrade to angular-messages .1.3.16 it can be work.
* When some input has validation in ng-form ,It cannot be remove from an array.I found the problem cause by when inject module of  ng-animate e.g
` angular.module("App", ["ngMessages",  "ngAnimate"])`
see [Plunker](http://plnkr.co/edit/mUi24IA2yyww0TyoDeRD)

please help me ,thanks a lot

Narretz:
Can you please provide a plnkr or similar for these issues, especially the first? The second one sounds like a known ngAnimate bug.

kimx:
For first Issue ng-message cannot show ,[this plunker can show] (http://plnkr.co/edit/lML0DrYXj6XkT9fASgsn) ,for this sample just clear any value of input,that will show validation message,but if I change 1.3.16 to 1.4.1 ,the message cannot show.

gkalpak:
Regarding the 1st issue, there have been some changes in `ngMessages` on 1.4.x, so the `ng-messages` expression is not interpolated.
To make it work on both versions, change `rowForm['item.name{{$index}}'].$error` to `rowForm['item.name'+$index].$error` ([updated plnkr][1]).

[1]: http://plnkr.co/edit/gmIthPYnztmO7l6jRPxb?p=preview

kimx:
@gkalpak thanks 1st issue can work, @Narretz  about second issue will be fix in next version? thanks a lot .

sreeramu:
#12271 should fix this issue.

Narretz:
@kimx Sorry, we don't have a fix for that yet. It's been tracked in issue https://github.com/angular/angular.js/issues/12161, so I'll close this one.

Issue #12161, opened 2015-06-18
digest issue with ng-form with validators on a ng-repeat when using ng-animate on 1.4.1


jlowcs:
When there is a ng-form directive on a ng-repeat directive, and there are validations in fields inside that form, the deletion of an element in the list is not reflected in the ng-repeat.

http://plnkr.co/edit/Er0thSHyCg0s2lvGO8SP

```html
    <div ng-controller="MyCtrl as myCtrl">
      list length: {{myCtrl.list.length}}
      <div ng-form ng-repeat="elt in myCtrl.list track by elt.id">
        <input required ng-model="elt.text">
        <button ng-click="myCtrl.deleteElt($index)">delete</button>
      </div>
      <button ng-click="myCtrl.addElt()">add</button>
    </div>
```

(ng-form and ng-repeat are on the same element)

```javascript
var app = angular.module('exampleApp', ['ngAnimate']);

var cpt = 0;

app.controller('MyCtrl', function ($scope, $element) {
  this.list = [];
  for (var i = 0; i < 6; ++i) {
    this.list.push({
      text: 'Elt ' + i,
      id: ++cpt
    });
  }
  
  var _this = this;
  
  this.deleteElt = function (index) {
    _this.list.splice(index, 1);
  }
  this.addElt = function (index) {
    _this.list.push({
      text: '',
      id: ++cpt
    });
  }
});
```

nomack84:
I have exactly the same issue. I have this code working on 1.3.16 but when I update to 1.4 it stop working. The ng-repeat is not updating the data.

nomack84:
ping @Narretz @petebacondarwin I think this is a **bug**

gkalpak:
Wow ! Impressive combination of prerequisites:

1. `ngAnimate`
2. `ngForm` and `ngRepeat` on the same element
3. Inputs with some validation directive

I suspect it has something to do with some class-based animation on the `[ng-form]` element (e.g. `ng-valid`/`ng-invalid` etc) that cancels `ngRepeat`'s structural animation (removal).

--
FWIW, it broke in `v1.4.0-rc.0` (it works OK up to `v1.4.0-beta.6`).

nomack84:
@gkalpak well, if you think about it, is not that hard to get that combination in practice. 
This currently is stopping me from upgrading to Angular 1.4.1, as I have many cases where this combination is applied. @matsko any chance to have this solved ASAP?
Thanks a lot

jlowcs:
[...]

PR #12271, opened 2015-07-05
Leave Animation callback overridden by next animation

sreeramu:
This patch will fix the overriding or loosing the leave callback object of the animation.

Example:

```
var app = angular.module("app", ['ngMessages', 'ngAnimate']);

app.directive('iholder', ['$animate','$timeout', function($animate,$timeout) {
    return {
        restrict: 'EA',
        scope: {},
        template: '<div class="level1 level2" ng-click="clickHandler()" style="width:100%;height:10em;background-color:red;"></div>',
        replace: true,
        link: function($scope, element, attrs) {
            $scope.clickHandler = function() {
                $animate.leave(element);
                $animate.removeClass(element,'level2');
            }
        }
    }
}]);
```

In above example removeClass is called as soon as the leave is called so in queueAnimation the the leave animation option (callback) is overridden with the removeClass option so the element is not removed.

This patch will fix the #12249 issue , in this issue leave is called by ngRepeatDirective for removing the element and on destroy ngForm called removeClass so the leave callback is overridden in mergeAnimationOptions so the element is not removed.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/12271)

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -185,6 +185,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       push: function(element, event, options, domOperation) {
         options = options || {};
         options.domOperation = domOperation;
+        options.hasDomOpreation = !isUndefined(domOperation);
         return queueAnimation(element, event, options);
       },
 
--- a/src/ngAnimate/shared.js
+++ b/src/ngAnimate/shared.js
@@ -171,8 +171,15 @@ function mergeAnimationOptions(element, target, newOptions) {
   var toAdd = (target.addClass || '') + ' ' + (newOptions.addClass || '');
   var toRemove = (target.removeClass || '') + ' ' + (newOptions.removeClass || '');
   var classes = resolveElementClasses(element.attr('class'), toAdd, toRemove);
+  var targetDomOperation = target.hasDomOpreation ? target.domOperation : null;
 
   extend(target, newOptions);
+
+  // TODO : proper fix is to maintain all animation callback in array and call at last,but now only leave has the callback so no issue with this.
+  if (targetDomOperation) {
+    target.hasDomOpreation = true;
+    target.domOperation = targetDomOperation;
+  }
 
   if (classes.addClass) {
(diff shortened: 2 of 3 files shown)


sreeramu:
I signed it!

sreeramu:
due to this fix below test cases are failing.

'''
it('class-based animations, however it should also cancel former structural animations in the process',
        inject(function($animate, $rootScope) {

        element.addClass('green lime');

        $animate.enter(element, parent);
        $animate.addClass(element, 'red');
        $animate.removeClass(element, 'green');

        $animate.leave(element);
        $animate.addClass(element, 'pink');
        $animate.removeClass(element, 'lime');

        expect(element).toHaveClass('red');
        expect(element).not.toHaveClass('green');
        expect(element).not.toHaveClass('pink');
        expect(element).toHaveClass('lime');

        $rootScope.$digest();

        expect(capturedAnimation[0]).toBe(element);
        expect(capturedAnimation[1]).toBe('leave');

        // $$hashKey causes comparison issues
        expect(element.parent()[0]).toEqual(parent[0]);

        options = capturedAnimation[2];
        expect(options.addClass).toEqual('pink');
        expect(options.removeClass).toEqual('lime');
      }));
'''
according to this patch animate.leave is going on so we don't merge the animation object for subsequent animation on the element we just close the event, but in close there is applyAnimationClasses(element, options); which will apply the classed so pink is added and lime is removed after leave also , is it ok that i update test case also.?

Narretz:
Hi, thanks for the PR. Would it be possible for you to add a test for this?

sreeramu:
@Narretz Test script are updated for this bug.

Narretz:
@matsko can you take a look at this?

hparfr:
Thanks sreeramu, this PR fix an issue I had (similar to #12161)

matsko:
@sreeramu I've taken your PR and removed some extra code: https://github.com/angular/angular.js/pull/12567

I'm waiting for Travis to be green and then it's in.

Excellent work on this! Thank you for putting in the time.

sreeramu:
Thanks @matsko

PR #12278, opened 2015-07-06
Fix for #12096 : Fixed leave start & close callback

### LINKED ISSUE #12096: ngAnimate on leave event miss "close" phase
Basically the problem is I don't get "close" phase of "leave" event. Works fine for "enter" event.

Here's the plunkr (see the output in console when you click checkbox): http://plnkr.co/edit/V7lgKLzPbwkBXoerUTAG?p=preview

sreeramu:
This PR fixes #12096 issue.

As leave event removes the node from the document `entry.node.contains(targetNode)` will not match for leave animation so no callbacks are called for leave.

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -124,14 +124,17 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       return mergeAnimationOptions(element, options, {});
     }
 
-    function findCallbacks(element, event) {
+    function findCallbacks(parent, element, event) {
       var targetNode = getDomNode(element);
+      var targetParentNode = getDomNode(parent);
 
       var matches = [];
       var entries = callbackRegistry[event];
       if (entries) {
         forEach(entries, function(entry) {
           if (entry.node.contains(targetNode)) {
+            matches.push(entry.callback);
+          } else if (event === 'leave' && entry.node.contains(targetParentNode)) {
             matches.push(entry.callback);
           }
         });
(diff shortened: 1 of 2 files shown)


Narretz:
Hi, thanks for the PR. Would it be possible for you to add a test for this?

src/ngAnimate/animateQueue.js
```diff
@@ -131,7 +131,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       var entries = callbackRegistry[event];
       if (entries) {
         forEach(entries, function(entry) {
-          if (entry.node.contains(targetNode)) {
+          if (event == 'leave' || entry.node.contains(targetNode)) {
```
wesleycho: This should probably be `===`

sreeramu:
Thanks @Narretz & @wesleycho , i had update this PR with update and the test scripts are also added.

Narretz:
@sreeramu  It looks like one of your commits has a different email than entered in the CLA. Can you please squash the commits into one that has the correct email? That should do the trick.

frankweb:
Hi @Narretz, there is another issue that could be solved with this fix https://github.com/angular/angular.js/issues/13054, is there a chance to merge the changes? Thank you.

Narretz:
I'd like for @matsko to review this, but I don't think he will have time before angular connect.

petebacondarwin:
I rebased, tweaked and landed this. Thanks @sreeramu and @frankweb

Issue #12281, opened 2015-07-07
ngAnimate 1.4.1 causes rootScope digest/apply per element animated


startswithaj:
ngAnimate inconjunction with ng-repeat causes a rootScope digest/apply per element animated.

This is the actual angular ng-repeat animation plunkr modified to demostrate the issue. You need to open inspector and view console log. Then type 'j' in the search field. When you filter down the elements and they animate you get a digest per animated row. Without the animation you get just the single digest or double digest for dirty checking.

No Animation - 2 digests happen on the rootscope level (for dirty checking)
http://plnkr.co/edit/lf70ifdNEh0wftQCqmEy

Animation class on ng-repeat 8 digests happen - 1 rootscope digest per element animated
http://plnkr.co/edit/l1vXyxqBYVLrOHnBc1vK

This is the easiest way to illustrate it. Is this desired behaviour? If its going to digest shouldnt it just digest the scope its in rather than do an $apply and bubble digests all the way up/down from the rootscope?

startswithaj:
https://github.com/angular/angular.js/blob/071be609275558f177cdb28d8f8891bd1805af1c/src/ngAnimate/animateCss.js#L915

```
$timeout(onAnimationExpired, maxDelayTime + CLOSING_TIME_BUFFER * maxDurationTime);
```

This line is missing the third invokeApply parameter false.

```
$timeout(onAnimationExpired, maxDelayTime + CLOSING_TIME_BUFFER * maxDurationTime, false);
```

The stagger code @ https://github.com/angular/angular.js/blob/071be609275558f177cdb28d8f8891bd1805af1c/src/ngAnimate/animateCss.js#L915

Does apply the this third parameter and does not cause multiple $applys.

startswithaj:
#12282

Den-dp:
+1

PR #12282, opened 2015-07-07
fix(ngAnimate): $timeout without invokeApply

startswithaj:
Currently when using ngAnimate in-conjunction
with ng-repeat a $rootScope.$digest happens
per element animated.
This change calls $timeout with the invokeApply
parameter set to false which stops ngAnimate
from invoking its changes inside an $apply block.

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -912,7 +912,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
 
           startTime = Date.now();
           element.on(events.join(' '), onAnimationProgress);
-          $timeout(onAnimationExpired, maxDelayTime + CLOSING_TIME_BUFFER * maxDurationTime);
+          $timeout(onAnimationExpired, maxDelayTime + CLOSING_TIME_BUFFER * maxDurationTime, false);
 
           applyAnimationToStyles(element, options);
         }


startswithaj:
resolves #12281

I have signed the cla

wesleycho:
This looks like a good change to me, but probably needs another pair of eyes to be sure.

captainclam:
+1 thanks

startswithaj:
There are plunkrs that demo this bug on #12281

margh:
solid change +1

startswithaj:
@wesleycho the $timeout definitely shouldn't be forcing an apply but the code doesn't make total sense to me. So I agree it needs another set of eyes.

The $timeout queues up onAnimationExpired() to run after a given time period but then says in the code comments the below.

```
        function onAnimationExpired() {
          // although an expired animation is a failed animation, getting to
          // this outcome is very easy if the CSS code screws up. Therefore we
          // should still continue normally as if the animation completed correctly.
          close();
        }
```

A reference to the $timeout isn't stored and cancelled if the animation is successful in onAnimationProgress(). So the $timedout function onAnimationExpired() runs for every element regardless of where its failed or not. So it goes 
onAnimationExpired -> 
 close() -> 
   if (animationClosed || (animationCompleted && animationPaused)) return;

And then because of no invokeApply false passed to the timeout it does a full apply/digest cycle per element.

matsko:
This is great. Thank you for finding this.

matsko:
Merged in now to master. Thanks again for finding this.

PR #12285, opened 2015-07-07
fix(injector): check that modulesToLoad isArray.

mprobst:
When users accidentally just pass a single string, e.g.
`angular.injector('myModule')`, this change give them a better error message.

Currently Angular just reports that the module with the name 'm' is missing,
as it's iterating through all characters in the string, instead of all strings
in the module.


lgalfaso:
Thanks for the PR, can you please add a test?

mprobst:
@lgalfaso test added, please take another look.

PR #12286, opened 2015-07-07
fix(merge): clone elements instead of threating them like objects

luckylooke:
Similar fix to #11720. Without this fix there is only empty object created in destination object.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -354,6 +354,8 @@ function baseExtend(dst, objs, deep) {
       if (deep && isObject(src)) {
         if (isDate(src)) {
           dst[key] = new Date(src.valueOf());
+        } else if (isElement(src)) {
+          dst[key] = src[0] ? jqLite(src).clone()[0] : jqLite(src).clone();
         } else {
           if (!isObject(dst[key])) dst[key] = isArray(src) ? [] : {};
           baseExtend(dst[key], [src], true);
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -491,6 +491,17 @@ describe('angular', function() {
 
       expect(dst.date).toBe(src.date);
     });
+
+    it('should copy elements by reference', function() {
+      var src = { element: document.createElement('div'),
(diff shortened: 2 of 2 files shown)


luckylooke:
I was not sure whether referencing the object would not be better, but according to similar issue with dates #11720 I made it to be cloned. Because docs says merge is deep copy.

test/AngularSpec.js
```diff
@@ -570,6 +581,21 @@ describe('angular', function() {
...
+
+
+    it('should copy(clone) elements', function() {
+      var src = { element: document.createElement('div'),
+        jqObject: jqLite("<p><span>s1</span><span>s2</span></p>").find("span") };
+      var dst = {};
+
+      merge(dst, src);
+
+      expect(dst.element).not.toBe(src.element);
```
petebacondarwin: should we be checking that `dst.element` is actually an Element here?

luckylooke: @petebacondarwin I think we should (it is checked on line 596).

petebacondarwin: Right!

petebacondarwin:
LGTM

src/Angular.js
```diff
@@ -354,6 +354,8 @@ function baseExtend(dst, objs, deep) {
       if (deep && isObject(src)) {
         if (isDate(src)) {
           dst[key] = new Date(src.valueOf());
+        } else if (isElement(src)) {
+          dst[key] = src[0] ? jqLite(src).clone()[0] : jqLite(src).clone();
```
gkalpak: A little late to the party, but shouldn't it be the other way around?
Namely: `!src[0] ? jqLite(src).clone()[0] : jqLite(src).clone();`

(In fact, I would rather expect: `!src[0] ? jqLite(src).clone()[0] : src.clone();`)

petebacondarwin: Gadzooks you are right!

petebacondarwin: Fixing now

petebacondarwin: @gkalpak - thanks for that. See 75292a6cb5e17d618902f7996e80eb3118eff7b0

PR #12290, opened 2015-07-07
fix($compile): ignore optional =-bound properties with empty value

### LINKED ISSUE #12144: Isolated scope property for directive is being re-digested in version 1.4.1

### LINKED ISSUE #12259: Troubleshoot Angular Material breakage with "do not write @-bound properties" commit
Troubleshoot Angular Material breakage with https://github.com/angular/angular.js/commit/8a1eb1625c080445ce1e519762e1f2d4fd842b72, which was reverted.

This spec is failing: https://github.com/angular/material/blob/786b0ed3652b7460c2c802efa1aa79972bd96f5d/src/components/chips/contact-chips.spec.js#L46

With this error:

```
Error: [$compile:nonassign] Expression '' used with directive 'mdAutocomplete' is non-assignable!
	http://errors.angularjs.org/1.4.2-build.4088+sha.4da1cc3/$compile/nonassign?p0=&p1=mdAutocomplete
	minErr/<@/material/temp/angular.js:68:12
	initializeDirectiveBindings/</parentSet<@/material/temp/angular.js:8705:1
	parentValueWatch@/material/temp/angular.js:8718:19
	regularInterceptedExpression@/material/temp/angular.js:14227:16
	constantWatch@/material/temp/angular.js:14208:16
	$RootScopeProvider/this.$get</Scope.prototype.$digest@/material/temp/angular.js:15674:34
	$RootScopeProvider/this.$get</Scope.prototype.$apply@/material/temp/angular.js:15950:13
	buildChips/<@/material/src/components/chips/contact-chips.spec.js:98:7
	invoke@/material/temp/angular.js:4452:14
	workFn@/material/temp/angular-mocks.js:2420:11
	angular.mock.inject@/material/temp/angular-mocks.js:2392:30
	buildChips@/material/src/components/chips/contact-chips.spec.js:96:5
	@/material/src/components/chips/contact-chips.spec.js:51:23
	invoke@/material/temp/angular.js:4452:14
	workFn@/material/temp/angular-mocks.js:2420:11

	angular.mock.inject@/material/temp/angular-mocks.js:2391:25
	buildChips@/material/src/components/chips/contact-chips.spec.js:96:5
	@/material/src/components/chips/contact-chips.spec.js:51:23
	invoke@/material/temp/angular.js:4452:14
	workFn@/material/temp/angular-mocks.js:2420:11

	angular.mock.inject@/material/temp/angular-mocks.js:2391:25
	@/material/src/components/chips/contact-chips.spec.js:46:27
	jasmineInterface.describe@/material/node_modules/karma-jasmine/lib/boot.js:59:14
	@/material/src/components/chips/contact-chips.spec.js:45:5
	jasmineInterface.describe@/material/node_modules/karma-jasmine/lib/boot.js:59:14
	@/material/src/components/chips/contact-chips.spec.js:36:3
	jasmineInterface.describe@/material/node_modules/karma-jasmine/lib/boot.js:59:14
	@/material/src/components/chips/contact-chips.spec.js:1:1
```

CC @ThomasBurleson @caitp

caitp:
Previously, optional empty '='-bound expressions were ignored ---
erroneously they stopped being ignored, and no tests were caused to fail
for this reason. This change restores the original ignoring behaviour while
still preventing other errors fixed by 8a1eb16

Closes #12144
Closes #12259

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2591,6 +2591,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               if (optional) break;
               attrs[attrName] = void 0;
             }
+            if (optional && !attrs[attrName]) break;
 
             parentGet = $parse(attrs[attrName]);
             if (parentGet.literal) {
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2567,24 +2567,19 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         lastValue,
         parentGet, parentSet, compare;
 
-        if (!hasOwnProperty.call(attrs, attrName)) {
-          // In the case of user defined a binding with the same name as a method in Object.prototype but didn't set
-          // the corresponding attribute. We need to make sure subsequent code won't access to the prototype function
-          attrs[attrName] = undefined;
-        }
-
         switch (mode) {
 
           case '@':
-            if (!attrs[attrName] && !optional) {
-              destination[scopeName] = undefined;
+            if (!optional && !hasOwnProperty.call(attrs, attrName)) {
+              destination[scopeName] = attrs[attrName] = void 0;
             }
-
             attrs.$observe(attrName, function(value) {
-              destination[scopeName] = value;
+              if (isString(value)) {
+                destination[scopeName] = value;
+              }
             });
             attrs.$$observers[attrName].$$scope = scope;
-            if (attrs[attrName]) {
+            if (isString(attrs[attrName])) {
               // If the attribute has been provided then we trigger an interpolation to ensure
               // the value is there for use in the link fn
               destination[scopeName] = $interpolate(attrs[attrName])(scope);
@@ -2592,11 +2587,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             break;
 
           case '=':
-            if (optional && !attrs[attrName]) {
-              return;
+            if (!hasOwnProperty.call(attrs, attrName)) {
+              if (optional) break;
+              attrs[attrName] = void 0;
             }
-            parentGet = $parse(attrs[attrName]);
 
+            parentGet = $parse(attrs[attrName]);
             if (parentGet.literal) {
               compare = equals;
             } else {
@@ -2635,7 +2631,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             break;
 
           case '&':
-            parentGet = $parse(attrs[attrName]);
+            // Don't assign Object.prototype method to scope
+            parentGet = attrs.hasOwnProperty(attrName) ? $parse(attrs[attrName]) : noop;
 
             // Don't assign noop to destination if expression is not valid
             if (parentGet === noop && optional) break;
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -3618,6 +3618,33 @@ describe('$compile', function() {
     });
 
 
+    it('should ignore optional "="-bound property if value is the emptry string', function() {
+      module(function($compileProvider) {
+        $compileProvider.directive('testDir', valueFn({
+          scope: {prop: '=?'},
+          controller: function($scope) {
+            $scope.prop = $scope.prop || 'default';
+            this.getProp = function() {
+              return $scope.prop;
+            };
+          },
+          controllerAs: 'ctrl',
+          template: '<p></p>'
+        }));
+      });
+      inject(function($compile, $rootScope) {
+        element = $compile('<div test-dir></div>')($rootScope);
+        var scope = element.isolateScope();
+        expect(scope.ctrl.getProp()).toBe('default');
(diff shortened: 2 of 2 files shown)


caitp:
@petebacondarwin / @jeffbcross PTAL --- I've tested this against the angular-material tests and it seems to be working okay, but I'm not sure what the issue was with the tab control (so it would be good to get another look just in case there's another sneaky bug)

Narretz:
Thanks @caitp ! It looks like @ThomasBurleson is involved with this, too.

caitp:
is travis really being this slow today? wow

src/ng/compile.js
```diff
@@ -2592,11 +2587,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             break;
 
           case '=':
-            if (optional && !attrs[attrName]) {
-              return;
+            if (!hasOwnProperty.call(attrs, attrName)) {
```
gkalpak: Why not just use `attrs.hasOwnProperty` (instead of `hasOwnProperty.call(attrs`) here and above ?
(It is being used below anyway ;))

caitp: attrs shouldn't share [[Prototype]] with Object, there are a lot of bugs which have been fixed because of this. the code below just came from a revert of a revert, so it's part of the pieces that haven't been fixed yet

petebacondarwin:
Without running the tests and stuff myself, this LGTM

caitp:
I'm a bit worried that it could break things in ways that aren't being tested in Material. @jeffbcross did you know of any demo apps in material which were affected? The ones I tried all looked fine

jeffbcross:
I don't know any demo apps that were affected, but I have limited familiarity with the material project. Could you give @ThomasBurleson some guidance on where this change could potentially cause problems?

src/ng/compile.js
ThomasBurleson: @caitp - in https://github.com/angular/angular.js/issues/12144#issuecomment-119284219 you mentioned some additional code required for Angular Material. 
*  Is that still true?
*  If yes, should we modify the Angular Material directives with optional 2-way binding?

ThomasBurleson:
@jeffbcross, @petebacondarwin - see comment https://github.com/angular/angular.js/pull/12290/files#r34271903
I want to make sure I understand the issue, the fix, and any changes needed in Angular Material. Thx.

src/ng/compile.js
caitp: The code in that snippet is contained in the second commit here, just organized slightly differently --- I decided to avoid the `isString()` check and just test if it's falsy (values in `attrs` should always be a string, unless it gets shadowed above, in which case it's undefined --- so also falsy) --- so the simpler version in this PR should be fine

caitp: Oh, the example snippet in that issue also handles values which are exclusively whitespace, but I think these are pretty rare

petebacondarwin: @ThomasBurleson - see @caitp's reply to your question, that the code snippet that is mentioned in the previously linked comment is included in this PR. In other words you shouldn't need to make changes to Material for all your (non-disabled) tests to pass.

I think the only thing we can do now is see if this PR works for the latest on Material's master branch and if so then merge this PR.

ThomasBurleson: @petebacondarwin - Re: ... see if this PR works for the latest on Material's master branch.  Is this something I should be doing now; or is someone else assigned this task ?

matthiasg:
is this being actively investigated ? we are blocked by this for our next release and would have to change a number of sources to workaround. we would like to avoid that if this is being expected to be merged soon

src/ng/compile.js
Narretz: @ThomasBurleson would be nice if you could test that!

ThomasBurleson: @Narretz - I will test tonight and respond tomorrow. Will that work ?

petebacondarwin:
This PR still causes observers to be called with `undefined`. We should probably prevent this.

caitp:
it only does in certain condition, when the listeners are not optional. We don't have a great other way to do this

Narretz:
I say, let's merge it in time for 1.4.4. We can always look into this observers fired with undefined thing later.

matthiasg:
just merge this in .. this is causing a number of issues already and is making some directives quite cumbersome to write

petebacondarwin:
A project for tomorrow for me. OK

petebacondarwin:
I have merged this and I have a PR for #12383 at #12464

caitp:
Thanks Pete

PR #12291, opened 2015-07-07
docs(jqLite): added missing support for unbind/off with event object as parameter

gersomvg:


gkalpak:
Not sure what you mean by _"missing support for unbind(event) and off(event)"_.
They seem supported to me.

grsmvg:
Have you tested? For me they didn't work with jQlite, until I included the full jQuery lib.

gkalpak:
@grsmvg, yes [I have tested][1].

<sub>Good job, @gkalpak, for valiantly resisting the unbeatable urge to be sarcastic about _"have you tested"_.
I'm seeing some serious progress here ! :smile:</sub>

[1]: http://codepen.io/ExpertSystem/pen/mJLrdJ?editors=101

grsmvg:
In your example you are not using the exact syntax I'm talking about ;) And glad you're not too sarcastic because we're all here to help an awesome open source project!

The exact syntax that does not seem to be working is like this:

```
$(someElm).on('click', function(event){
  $(someElm).off(event);
});
```

And I'm not saying it needs to be fixed, but it should at least be stated correctly in the docs imho.

gkalpak:
Ah, I see what you mean now :)
(TBH I've never used such a thing.)

I think it would be better phrazed as _"event object as parameter"_.
But it also seems trivial to implement...

I would be fine with both :smile:

grsmvg:
The reason someone would use this is because this looked like the only way to unbind one specific event binding without using namespaces (because namespaces are also not supported in jqlite).

I've already ended up using full jQuery lib, but for future readers it's good to change this in docs. See my last commit for a better description.

grsmvg:
Responding to Google bot at top^^ : 'I signed it!'

Pull request is ready for merge.

PR #12294, opened 2015-07-08
docs(guide/Scopes): added ng-app to the examples

christianreed:
ng-app was left off of all the examples, so if you type them out and try to run them they don't work.

### DIFF
--- a/docs/content/guide/scope.ngdoc
+++ b/docs/content/guide/scope.ngdoc
@@ -55,7 +55,7 @@ the applications.
       }]);
   </file>
   <file name="index.html">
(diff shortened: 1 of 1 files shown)


Narretz:
Hi, thanks for the PR. This is a little problematic, because when you open the examples in the plnkr, there will be two ng-app directives. It doesn't hurt, but it's also not very pretty. This is because the example generator will add ng-app automatically to the example. Since you have to add html boilerplate to the examples anyway when you copy / paste them, I think we should not add ng-app explicitly to these examples. But thanks for your effort!

Issue #12287, opened 2015-07-07
angular.js 1.4 isFunction is not defined in loader.js


jerryorta-dev:
Loading the seed-app with index-async.html, loader.js throws an error ```Uncaught ReferenceError: isFunction is not defined```.

I will have a pull request to fix loader.js.

lgalfaso:
I think that `isFunction` should be added to *src/loader.prefix*, and that should make it.

jerryorta-dev:
@lgalfaso thank you. Didn't see this at first. Adjusting pull request.

jerryorta-dev:
ok... this should be ready to merge per instructions.

jerryorta-dev:
I see the pr was declined. Any suggestions how this may be fixed?

pkozlowski-opensource:
@jerryorta-dev sorry, I've closed your PR by accident / lack of knowledge. I've re-opened it.

lgalfaso:
landed as 9ea52d818bcd2fb3ea8ccc85bf47f9fd5af68843

PR #12298, opened 2015-07-08
Fix loader 12287

jerryorta:
Sorry for all the commits, this is my first time doing this process.

Fix for issue #12287, define isFunction in loader


Narretz:
Can you please clean up this PR / create a new PR which only includes your changes? Otherwise we can't work with this.

PR #12299, opened 2015-07-08
fix(loader): define isFunction

jerryorta:
Issue #12287: loader throws error using index-async.html in seep-app.

### DIFF
--- a/src/loader.prefix
+++ b/src/loader.prefix
@@ -5,3 +5,4 @@
  */
 'use strict';
 (function() {
+    function isFunction(value) {return typeof value === 'function';};


pkozlowski-opensource:
It is defined [here](https://github.com/angular/angular.js/blob/d518a64d93b4f63e9c55678f5a094cc439e02ec0/src/Angular.js#L614) already. Why would you add it to the loader file as well?

lgalfaso:
landed as 9ea52d818bcd2fb3ea8ccc85bf47f9fd5af68843

PR #12304, opened 2015-07-08
fix(ngLocale): Fixed Chilean locales

hugomenesesp:
- Chilean translation to September is "Septiembre", "Setiembre" is a slang translation.
- Short days and months should not have an ending dot, it's dependent on the usage.
- Weekends are Sunday and Saturday.
- First day of week is Monday.

### DIFF
--- a/src/ngLocale/angular-locale_es-cl.js
+++ b/src/ngLocale/angular-locale_es-cl.js
@@ -24,7 +24,7 @@ $provide.value("$locale", {
       "a. C.",
       "d. C."
     ],
-    "FIRSTDAYOFWEEK": 0,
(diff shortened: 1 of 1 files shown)


lgalfaso:
This should be split into several parts.

1/ "Setiembre" vs "Septiembre" should be a simple fix with just regenerating
2/ The extra dot at the end of the short month, for that check [Localization Issue](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-found-an-issue)
3/ Day positions follow ISO-8601 where Monday is day 0

Closing this, in a few will post a PR that regenerates the locales, this should solve only 1/, for 2/ please follow the guidelines and 3/ is not an issue.

PR #12305, opened 2015-07-08
Fixed Chilean locales

hugomenesesp:
- Chilean translation to September is "Septiembre", "Setiembre" is a slang translation.
- Short days and months should not have an ending dot, it's dependent on the usage.
- Weekends are Sunday and Saturday.
- First day of week is Monday.


lgalfaso:
dup of #12304

PR #12317, opened 2015-07-10
fix(ngTouch) Remove ngClick override from ngTouch

### LINKED ISSUE #4030: ngTouch: ng-click directive is broking default behaviour of child elements

christianvuerings:
#2989 introduces a bug (when adding the `ng-click` directive) which prevents the default behavior of child elements like `a` and `input` elements when you click them.

Instead we should recommend people to use https://github.com/ftlabs/fastclick to eliminate the 300ms delay.
This is a follow up from #11212 http://git.io/vqyF6 where it was suggested to remove the ngClick override

BREAKING CHANGE: Remove ngClick override from ngTouch, use fastclick inststead
Closes #4030, #6432


christianv:
@petebacondarwin - could you have a look at this PR? Let me know if you would like me to make any changes.

petebacondarwin:
I'll take a look in the next week

christianv:
@petebacondarwin thanks!

petebacondarwin:
This is a big breaking change so will have to go into 1.5.

We also need to have a clear message for what people need to do if they rely on this `ngClick` implementation. We need to update the docs for `ngTouch` and the migration guide doc. In these docs we should talk about how to use fastclick.js and why we are proposing that rather than what we had before.

Also I think the breaking change message should be a bit more involved - is it really simply a case of loading fastclick.js or do we need to do some configuration?

Finally do we have any examples or e2e tests that rely on this behaviour?

PR #12318, opened 2015-07-10
ngTemplator - Build Website form Json

joelviseu:
The idea is create a website from a json schema, and cross with the templates base HTML


joelviseu:
I signed it!

pkozlowski-opensource:
@joelviseu this sounds like a potential 3rd party module that you should maintain and submit to http://ngmodules.org/ (or similar repo), but it doesn't belong to the core. Thnx for the proposal, though.

PR #12307, opened 2015-07-08
chore(ngLocale): regenerate locale files

lgalfaso:

PR #12323, opened 2015-07-10
Month translation bug and currency update

AlexanderZon:
Is not setiembre is septiembre, and the actually currency prefix (sufix really) is Bsf.


AlexanderZon:
I signed it!

Narretz:
We are using https://github.com/google/closure-library/tree/master/closure/goog/i18n for our locale files. So any bugs must either be fixed in the library, or we must import the library to update the code. Would you be able to check if this is still bugged in closure?

lgalfaso:
This should get fixed once #12307 lands

PR #12326, opened 2015-07-11
docs(guide/Forms): The CSS example is missing binding results

maurya777:
It will be good to have the binding results of CSS example, similar to the Simple Form example.


maurya777:
I signed it!

PR #12328, opened 2015-07-11
Update angular-locale_hr-hr.js

TommyZG:
Changed "MONTH" values to be more consistent with Croatian language. The genitiv forms are rarely used, only in very formal occasions, it is much more useful to have nominativ forms
.
Changed "FIRSTDAYOFWEEK" from 0 to 1 because in Croatia, as in most European countries, Monday is the first day of the week.

### DIFF
--- a/src/ngLocale/angular-locale_hr-hr.js
+++ b/src/ngLocale/angular-locale_hr-hr.js
@@ -42,20 +42,20 @@ $provide.value("$locale", {
       "pr. Kr.",
       "p. Kr."
     ],
-    "FIRSTDAYOFWEEK": 0,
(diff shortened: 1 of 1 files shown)


TommyZG:
I signed it!

Narretz:
Thanks for the PR. However, ngLocale basically imports the i18n part of https://github.com/google/closure-library, so all fixes must be made there before they can land in angular.

Narretz:
ACtually, they must be made further down the rabbit hole: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#localization-issue

PR #12329, opened 2015-07-12
fix(cacheFactory): check key exists before decreasing size

### LINKED ISSUE #12321: $cacheFactory.remove() does always just size-- but it doesn't have really delete something
see line:

https://github.com/angular/angular.js/blob/master/src/ng/cacheFactory.js#L227

there size-- is done but just before it we do:

 delete data[key];

but there is no test if it is really in the data, so if remove is called with something thats not inside the data nothing is deleted but the size is decreased.

the put function does have a check:

if (!(key in data)) size++;
data[key] = value;

shouldn't remove not have a similar one?

tealtail:
When $cacheFactory.remove() is called, if the key does not exist, return
before modifying size. There is currently no check for the key's
existence.

Closes #12321

### DIFF
--- a/src/ng/cacheFactory.js
+++ b/src/ng/cacheFactory.js
@@ -93,9 +93,9 @@ function $CacheFactoryProvider() {
 
       var size = 0,
           stats = extend({}, options, {id: cacheId}),
-          data = {},
+          data = createMap(),
           capacity = (options && options.capacity) || Number.MAX_VALUE,
-          lruHash = {},
+          lruHash = createMap(),
           freshEnd = null,
           staleEnd = null;
 
@@ -223,6 +223,8 @@ function $CacheFactoryProvider() {
             delete lruHash[key];
           }
 
+          if (!(key in data)) return;
+
           delete data[key];
           size--;
         },
@@ -237,9 +239,9 @@ function $CacheFactoryProvider() {
          * Clears the cache object of any entries.
          */
         removeAll: function() {
-          data = {};
+          data = createMap();
           size = 0;
-          lruHash = {};
+          lruHash = createMap();
           freshEnd = staleEnd = null;
         },
 
@@ -399,4 +401,3 @@ function $TemplateCacheProvider() {
     return $cacheFactory('templates');
   }];
 }
-
--- a/test/ng/cacheFactorySpec.js
+++ b/test/ng/cacheFactorySpec.js
@@ -133,6 +133,19 @@ describe('$cacheFactory', function() {
         expect(cache.info().size).toBe(0);
       }));
 
+      it('should only decrement size when an element is actually removed via remove', inject(function($cacheFactory) {
+        cache.put('foo', 'bar');
+        expect(cache.info().size).toBe(1);
+
+        cache.remove('undefined');
+        expect(cache.info().size).toBe(1);
+
+        cache.remove('hasOwnProperty');
+        expect(cache.info().size).toBe(1);
+
+        cache.remove('foo');
+        expect(cache.info().size).toBe(0);
+      }));
 
       it('should return cache id', inject(function($cacheFactory) {
         expect(cache.info().id).toBe('test');


src/ng/cacheFactory.js
```diff
@@ -223,6 +223,8 @@ function $CacheFactoryProvider() {
             delete lruHash[key];
           }
 
+          if (!(key in data)) return;
```
gkalpak: This will still has unexpected results if `key` is prototypally inherited by `data`.
You could use `!data.hasOwnProperty(key)`. It might also be possible to use `data = createMap()` and not have to deal with the prototype chain (but I haven't checked if more modifications are needed in other places).

lgalfaso: I think that the simplest and best of the two options would be to do `data = createMap()`

tealtail: Is that literally what this line should be? Sorry, this is my first time contributing here and I'm having trouble finding anything about a `createMap()` function. Is this just pseudo-code, or something internal and already available?

caitp: `createMap` is defined in src/Angular.js. It just performs `Object.create(null);` to build an object with a null prototype

tealtail: Could I get a bit more guidance as to how I should make use of `createMap` in this case, though? If I'm setting `data = createMap()`, wouldn't that effectively be clearing out everything that was being held inside of `data`?

lgalfaso: Change https://github.com/angular/angular.js/blob/master/src/ng/cacheFactory.js#L96 and https://github.com/angular/angular.js/blob/master/src/ng/cacheFactory.js#L240 to `data = createMap();`

tealtail:
Thanks for pointing me in the right direction. I've updated my PR.

Narretz:
@tealtail Thanks! The final thing we need now is a test that ensures, we only decrement the counter if an element has actually been removed from the cache.

lgalfaso:
Just to complement what @Narretz just said. A test that without this chance would fail.

tealtail:
Alright, I've added a new test case to `cacheFactorySpec.js`.

src/ng/cacheFactory.js
```diff
@@ -93,7 +93,7 @@ function $CacheFactoryProvider() {
 
       var size = 0,
           stats = extend({}, options, {id: cacheId}),
-          data = {},
+          data = createMap(),
           capacity = (options && options.capacity) || Number.MAX_VALUE,
           lruHash = {},
```
lgalfaso: should `lruHash` also be created with `createMap()`?

test/ng/cacheFactorySpec.js
```diff
@@ -133,6 +133,16 @@ describe('$cacheFactory', function() {
         expect(cache.info().size).toBe(0);
       }));
 
+      it('should only decrement size when an element is actually removed via remove', inject(function($cacheFactory) {
+        cache.put('foo', 'bar');
+        expect(cache.info().size).toBe(1);
+
+        cache.remove('undefined');
```
lgalfaso: would it be possible to add a case that the key is `hasOwnProperty` (or some other property that is standard to `Object.prototype`?

lgalfaso:
besides the comment, and that a test on the LRU case would be great. LGTM

test/ng/cacheFactorySpec.js
tealtail: Do you mean adding in something like:
```js
cache.remove(hasOwnProperty);
expect(cache.info().size).toBe(1);
```

tealtail:
Now using `createMap` when creating `lruHash`, and I've added the additional test using `hasOwnProperty` as a key. You mentioned an LRU test as well, but I'm not sure what the test case should be for that.

test/ng/cacheFactorySpec.js
```diff
@@ -133,6 +133,19 @@ describe('$cacheFactory', function() {
...
+        expect(cache.info().size).toBe(1);
+
+        cache.remove(hasOwnProperty);
```
gkalpak: What @lgalfaso meant was `cache.remove('hasOwnProperty')`.

gkalpak:
Regarding the `lruHash` test, it is not easy to test it directly without its being exposed, but an indirect test is described in this [CodePen][1].

---
Not directly related to this PR, but the `destroy` method needs some massaging I think.
(At the very least, unsetting `freshEnd` and `staleEnd`, to allow them to be GC'ed. You can make a different PR if you fancy :smiley:)

[1]: http://codepen.io/ExpertSystem/pen/VLdoGZ?editors=001

PR #12338, opened 2015-07-13
fix(ngAnimate): ensure that orphaned elements do not throw errors when animated

### LINKED ISSUE #11975: Animations should not cancel parent animations

matsko:
This fix ensures that for both `$animateCss` and `$animate` capture the error
when an animation takes place where the element is removed from the
parent element sometime before or during the preparation stages of the
animation.

Closes #11975

### DIFF
--- a/test/ngAnimate/animateCssSpec.js
+++ b/test/ngAnimate/animateCssSpec.js
@@ -51,6 +51,43 @@ describe("ngAnimate $animateCss", function() {
   describe('when active', function() {
     if (!browserSupportsCssAnimations()) return;
 
+    it("should silently quit the animation and not throw when an element is not attached to a parent during preparation",
+      inject(function($animateCss, $$rAF, $rootScope, $document, $rootElement) {
+
+      var element = jqLite('<div></div>');
+      expect(function() {
+        $animateCss(element, {
+          duration: 1000,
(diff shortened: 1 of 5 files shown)


test/ngAnimate/animateCssSpec.js
```diff
@@ -51,6 +51,43 @@ describe("ngAnimate $animateCss", function() {
   describe('when active', function() {
     if (!browserSupportsCssAnimations()) return;
 
+    it("should silently quit the animation and not throw when an element is not attached to a parent during preparation",
```
petebacondarwin: `when an element has no parent` removes the potential confusion of a double negative.

PR #12340, opened 2015-07-14
fix($animateCss): ensure animations execute if only a keyframeStyle is provided

### LINKED ISSUE #12124: ngAnimate - Providing only keyframe/keyframeStyles to $animateCss does nothing
After realising that `keyframe` should actually be `keyframeStyles` for $animateCss I see that passing `keyframeStyles` alone will decide that there's nothing to do and return: `closeAndReturnNoopAnimator()`.

To get around this I have to do something like:
```js
$animateCss(ele, { 
  keyframeStyles: '1s my-anim ease',
  addClass: 'meh'
});
```

matsko:
`$animateCss` is a fan of transition animations, but it turns out that
if only a keyframeStyle is provided into the animation upon constrution
then it will quit because it assumes that nothing will be animated
(since no classes or styles are being applied). This patch ensures that
a keyframe style can solely be applied to an animation triggered with
`$animateCss`.

``` js
// this will now work as expected
$animateCss(element, { keyframeStyle: '1s rotate' }).start();
```

Closes #12124

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -556,10 +556,11 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       var fullClassName = classes + ' ' + setupClasses;
       var activeClasses = pendClasses(setupClasses, '-active');
       var hasToStyles = styles.to && Object.keys(styles.to).length > 0;
+      var containsKeyframeAnimation = (options.keyframeStyle || '').length > 0;
 
       // there is no way we can trigger an animation since no styles and
       // no classes are being applied which would then trigger a transition
-      if (!hasToStyles && !setupClasses) {
+      if (!containsKeyframeAnimation && !hasToStyles && !setupClasses) {
         return closeAndReturnNoopAnimator();
       }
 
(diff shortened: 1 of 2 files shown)


src/ngAnimate/animateCss.js
```diff
@@ -556,10 +556,14 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       var fullClassName = classes + ' ' + setupClasses;
       var activeClasses = pendClasses(setupClasses, '-active');
       var hasToStyles = styles.to && Object.keys(styles.to).length > 0;
+      var containsKeyframeAnimation = (options.keyframeStyle || '').length > 0;
 
       // there is no way we can trigger an animation since no styles and
       // no classes are being applied which would then trigger a transition
-      if (!hasToStyles && !setupClasses) {
+      // is there a raw keyframe value that is applied to the element
```
petebacondarwin: This comment seems a bit weird in the context of the previous two lines...

test/ngAnimate/animateCssSpec.js
```diff
@@ -2092,6 +2092,29 @@ describe("ngAnimate $animateCss", function() {
           expect(element.css('transition-property')).toMatch('color');
           expect(style).toContain('ease-in');
         }));
+
+        it("should only execute the animation if any CSS to styles are mixed into together",
```
petebacondarwin: So this should probably read:

```
it("should execute the animation only if a CSS `to` style is also mixed in"
```

right?

matsko: should execute the animation only if there is any CSS styling to go with the transition

PR #12346, opened 2015-07-14
fix(ngCsp): allow CSP to be configurable

petebacondarwin:
There are two different features in Angular that can break CSP rules:
use of `eval` to execute a string as JavaScript and dynamic injection of
CSS style rules into the DOM.

This change allows us to configure which of these features should be turned
off to allow a more fine grained set of CSP rules to be supported.

Closes #11933
Closes #8459

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -984,22 +984,39 @@ function equals(o1, o2) {
 }
 
 var csp = function() {
-  if (isDefined(csp.isActive_)) return csp.isActive_;
+  if (!isDefined(csp.rules)) {
 
-  var active = !!(document.querySelector('[ng-csp]') ||
-                  document.querySelector('[data-ng-csp]'));
 
-  if (!active) {
+    var ngCspElement = (document.querySelector('[ng-csp]') ||
+                    document.querySelector('[data-ng-csp]'));
+
+    if (ngCspElement) {
+      var ngCspAttribute = ngCspElement.getAttribute('ng-csp') ||
+                    ngCspElement.getAttribute('data-ng-csp');
+      csp.rules = {
+        noUnsafeEval: !ngCspAttribute || (ngCspAttribute.indexOf('no-unsafe-eval') !== -1),
+        noInlineStyle: !ngCspAttribute || (ngCspAttribute.indexOf('no-inline-style') !== -1)
+      };
+    } else {
+      csp.rules = {
+        noUnsafeEval: noUnsafeEval(),
+        noInlineStyle: false
+      };
+    }
+  }
+
+  return csp.rules;
+
+  function noUnsafeEval() {
     try {
       /* jshint -W031, -W054 */
       new Function('');
       /* jshint +W031, +W054 */
+      return false;
     } catch (e) {
-      active = true;
+      return true;
     }
   }
-
-  return (csp.isActive_ = active);
 };
 
 /**
--- a/src/ng/directive/ngCsp.js
+++ b/src/ng/directive/ngCsp.js
@@ -6,27 +6,29 @@
  *
  * @element html
  * @description
- * Enables [CSP (Content Security Policy)](https://developer.mozilla.org/en/Security/CSP) support.
+ *
+ * Angular has some features that can break certain
+ * [CSP (Content Security Policy)](https://developer.mozilla.org/en/Security/CSP) rules.
+ *
+ * If you intend to implement these rules then you must tell Angular not to use these features.
  *
  * This is necessary when developing things like Google Chrome Extensions or Universal Windows Apps.
  *
- * CSP forbids apps to use `eval` or `Function(string)` generated functions (among other things).
- * For Angular to be CSP compatible there are only two things that we need to do differently:
  *
- * - don't use `Function` constructor to generate optimized value getters
- * - don't inject custom stylesheet into the document
+ * The following rules affect Angular:
  *
- * AngularJS uses `Function(string)` generated functions as a speed optimization. Applying the `ngCsp`
- * directive will cause Angular to use CSP compatibility mode. When this mode is on AngularJS will
- * evaluate all expressions up to 30% slower than in non-CSP mode, but no security violations will
- * be raised.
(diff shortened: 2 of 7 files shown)


petebacondarwin:
@jdalton, @realityking, @lgalfaso and @btford - can you take a look?

src/Angular.js
```diff
@@ -984,22 +984,39 @@ function equals(o1, o2) {
...
+    var ngCspElement = (document.querySelector('[ng-csp]') ||
+                    document.querySelector('[data-ng-csp]'));
+
+    if (ngCspElement) {
+      var ngCspAttribute = ngCspElement.getAttribute('ng-csp') ||
+                    ngCspElement.getAttribute('data-ng-csp');
+      csp.rules = {
+        noUnsafeEval: !ngCspAttribute || (ngCspAttribute.indexOf('no-unsafe-eval') !== -1),
+        noInlineStyle: !ngCspAttribute || (ngCspAttribute.indexOf('no-inline-style') !== -1)
+      };
```
jdalton: Shouldn't the `ngCspAttribute.indexOf` checks be looking for `> -1` as if they're declared then they're `true`?

src/ng/directive/ngCsp.js
```diff
@@ -35,11 +37,37 @@
...
+ * styles and unsafe eval. E.g. `<body ng-csp>`.
+ *
+ * * Specifying only `no-unsafe-eval` tells Angular that we must not use eval, but that we can inject
+ * inline styles. E.g. `<body ng-csp="no-unsafe-eval">`.
+ *
+ * * Specifying only `no-inline-style` tells Angular that we must not inject styles, but that we can
+ * run eval - no automcatic check for unsafe eval will occur. E.g. `<body ng-csp="no-inline-style">`
+ *
+ * * Specifying both `no-unsafe-eval` and `no-inline-style` tells Angular that we must not inject
+ * styles nor use eval, which is the same as an empty: ng-csp, except that no runtime check for
```
jdalton: the comment
> except that no runtime check for unsafe eval will occur.

can be removed as the runtime check won't occur for `ng-csp` either at the moment.

src/ng/parse.js
```diff
@@ -1703,11 +1703,11 @@ function $ParseProvider() {
 
   this.$get = ['$filter', '$sniffer', function($filter, $sniffer) {
     var $parseOptions = {
-          csp: $sniffer.csp,
+          csp: $sniffer.csp && $sniffer.csp.noUnsafeEval,
```
realityking: Looking at [$sniffer](https://github.com/petebacondarwin/angular.js/blob/ae2241522611cebc8d0833f0bed58bb726c6fdb1/src/ng/sniffer.js#L81), I think the check if  `$sniffer.csp ` exists is unnecessary.

src/Angular.js
realityking: Checking for `!== -1` has the same effect, so this is fine.

realityking:
@petebacondarwin I like it (a lot)!

src/Angular.js
jdalton: > Checking for `!== -1` has the same effect, so this is fine.

Doh! I blame it on all the double negatives :smile_cat:

src/ng/directive/ngCsp.js
petebacondarwin: Oh yes you are right.

src/ng/parse.js
petebacondarwin: It would be except that there are loads of tests that mock out sniffer and don't provide a `csp` object. I debated back and forth between fixing the tests or fixing this line and in the end I felt that this was simpler as it meant all these other tests didn't need to know about the csp object.

You can see this in action if you remove this check and then run `grunt autotest`

realityking: Ah good point. Probably an indication that `$sniffer`is doing too much, but that ship has sailed.

test/AngularSpec.js
```diff
@@ -793,36 +804,51 @@ describe('angular', function() {
...
 
-    it('should return the true when CSP is enabled manually via [data-ng-csp]', function() {
-      spyOn(document, 'querySelector').andCallFake(function(selector) {
-        if (selector == '[data-ng-csp]') return {};
-      });
-      expect(csp()).toBe(true);
+    it('should return true when CSP is enabled manually via [data-ng-csp]', function() {
+      var spy = mockCspElement('data-ng-csp');
+      expect(csp()).toEqual({ noUnsafeEval: true, noInlineStyle: true });
       expect(document.querySelector).toHaveBeenCalledWith('[data-ng-csp]');
```
gkalpak: `expect(spy)` for consistency ? :smile:

gkalpak: `expect(spy)` for consistency ? :smile:

test/AngularSpec.js
```diff
@@ -793,36 +804,51 @@ describe('angular', function() {
...
     });
+
+
+    it('should return true for noUnsafeEval if it is specified in the `ng-csp` attribute value', function() {
+      var spy = mockCspElement('ng-csp', 'no-unsafe-eval');
+      expect(csp()).toEqual({ noUnsafeEval: true, noInlineStyle: false });
+    });
```
gkalpak: It might be also nice to test that no check is performed (i.e. spying on `window.Function` to verify it hasn't been called or something).

test/AngularSpec.js
petebacondarwin: Right!

test/AngularSpec.js
petebacondarwin: Right, too

src/ng/parse.js
petebacondarwin: Actually, perhaps we don't need to use `$sniffer` here at all? Since we are inside the angular.js closure we could just use the `csp()` function. I don't think we are mocking it anywhere are we?

petebacondarwin:
I have added a commit with changes thanks to the feedback.

gkalpak:
LGTM :+1:

btford:
:+1:

PR #12347, opened 2015-07-14
fix(ngAnimate): buffer repeated calls to RAF to improve performance for animations

### LINKED ISSUE #12280: $$rAF performance issue in Angular Material
When scrolling *sticky* md-subheader, recent changes in $$rAF seem to create excessive CPU usages.
See https://github.com/angular/material/issues/3531 for details and CodePen demos.

matsko:
Prior to this patch ngAnimate buffered all repeated calls to
requestAnimationFrame into a task queue that is flushed once the next
animation frame has passed. While this works it causes a random lag to
jump out when any CPU intensive rAF requests are processed. This patch
sets a configurable task range which then offsets followup rAF requests
into the next available frame.

Closes #12280

### DIFF
--- a/src/ng/raf.js
+++ b/src/ng/raf.js
@@ -1,6 +1,10 @@
 'use strict';
 
 function $$RAFProvider() { //rAF
+
+  var provider = this;
+  provider.$$bufferLimit = 10;
+
   this.$get = ['$window', '$timeout', function($window, $timeout) {
     var requestAnimationFrame = $window.requestAnimationFrame ||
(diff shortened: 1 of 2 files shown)


gkalpak:
LGTM

<sub>The implementation seems unnecessarily complex imo. I don't have proof though and since I can't spot any functional issue/bug (e.g. as was the case with the previous implementation), I'll leave it to the LGTM :smile:</sub>

matsko:
Closed in favour of: https://github.com/angular/angular.js/pull/12619

PR #12358, opened 2015-07-16
docs($http): specifies that url is the cache key

dod38fr:
When using a custom cache, it may be handy to clear some cached data. For instance, when you have a sequence of GET, POST and then GET on the same URL, the second GET must not use a cached data. Using a custom cache object from `$cacheFactory`, cached data can be cleared with `myCache.remove(url)`.

This works but uses an undocumented feature: the fact that `$http` use the required url as cache key. This change specifies that point.

Hope this helps

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -582,8 +582,10 @@ function $HttpProvider() {
      * cache) or to a custom cache object (built with {@link ng.$cacheFactory `$cacheFactory`}).
      * When the cache is enabled, `$http` stores the response from the server in the specified
      * cache. The next time the same request is made, the response is served from the cache without
-     * sending a request to the server.
-     *
+     * sending a request to the server. `$http` store the response using the required url as key in the cache.
+     * When using a custom cache object, you can then clear cached data with something like
+     * `myCache.remove(requestedUrl)`
+     * 
      * Note that even if the response is served from cache, delivery of the data is asynchronous in
      * the same way that real requests are.
      *


dod38fr:
Mon company (HP) has a corporate agreement. My work address is dominique.dumont at hp.com

jnizet:
As far as I understand the code, the url passed to $http() or $http.get() is not the actual key of the cache. The key is the url of the config followed by the query string built from the parameters:

    var url = buildUrl(config.url, config.paramSerializer(config.params));

So I think the suggested addition would make things more confusing: readers might fear that parameters are not taken into account, or might expect a cached response to be evicted if they don't take parameters into account.

dod38fr:
So, if I understand correctly, there's no simple way to clean up the cache passed to `$http.get` except by calling `removeAll()` (which clean *all* cache entries instead of just one).

Is this assessment correct ?

gkalpak:
@dod38fr, there is indeed no simple way to know the key if you are using query parameters.
It's not super complex or impossible to build the key yourself (you have all necessary pieces (`url`, `params`, `paramSerializer`) and the implementation of `buildUrl()` is really simple), but I definitely wouldn't recommend it.

Depending on your usecase, you might be able to create a very simple wrapper around a cache that does what you want.

gkalpak:
BTW, I agree with @jnizet - the suggested docs addition is inaccurate and could prove misleading.

dod38fr:
@gkalpak , the use case is: I have a remote storage that I mostly read from and sometimes write to. When I write to this remote storage, the next read should contain the data that was just written.

This is currently not possible using the simple cache facility provided by `$http`.

BTW, I agree with both of you. The doc modif I suggest is not good currently. I'll amend it or close this PR.

I just wish there was a simple way to know the cache key used by `$http`. Would it be possible to modify `$http` to provide this cache key in the `then` callback ? With this information, managing the cache content would then be possible.

gkalpak:
There are several ways to achieve the same I guess, but off the top of my head, I would try something [that][1].
(I.e. augment the cache used by `$http` to store the key somewhere.)

[1]: http://codepen.io/ExpertSystem/pen/NqBwWE?editors=101

dod38fr:
There's also a more simple (even if more resource consuming) way: use one cache object per API endpoint and use `removeAll` on this object when the cache needs to be flushed.

This avoid adding more complexity in my project for a problem that should be handled by AngularJS framework. 

I agree that the augment trick may not be much by itself, but it adds up with other workarounds and make on-boarding of new developers much more difficult.

Narretz:
Ok, it looks like the docs changes are not really correct. I am gonna close this then. Still, @dod38fr, thanks for contribution - you can open a new issue that addresses the problem with getting the cache key and / or another PR that correctly explains what the cache key is for $http is.

PR #12361, opened 2015-07-16
fix($location): don't crash if navigating outside the app base

### LINKED ISSUE #11667: Back button broken in Chrome when setting location.href in HTML5 mode
When clicking a link from ```http://example.com/``` to ```http://example.com/app``` where ```/app``` runs angular with HTML5 mode and the baseURL set to ```/app```, the following error is raised when you press the back button:

```Error: [$location:ipthprfx] Invalid url "http://example.com/", missing path prefix "http://example.com/app/"```

The offending line:
```var pathUrl = beginsWith(appBaseNoFile, url);```

Can be found here:
https://github.com/angular/angular.js/blob/a509e9aa149d0f88cc39f703d539f7ffd4cd6103/src/ng/location.js#L107

```pathUrl``` returns undefined which results in an error being thrown and the page not navigating back to ```http://example.com/``` as expected.

### PR DESCRIPTION
**Pull request by petebacondarwin**
This should fix #11667

### DIFF
--- a/test/ng/locationSpec.js
+++ b/test/ng/locationSpec.js
@@ -860,7 +860,6 @@ describe('$location', function() {
       });
     });
 
-
     // location.href = '...' fires hashchange event synchronously, so it might happen inside $apply
     it('should not $apply when browser url changed inside $apply', function() {
       initService({html5Mode:false,hashPrefix: '!',supportHistory: true});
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/ng/locationSpec.js [hunk 1]
```diff
@@ -601,35 +601,35 @@ describe('$location', function() {
 
 
       it('should return decoded characters for search specified in URL', function() {
-        var locationUrl = new LocationHtml5Url('http://host.com/');
+        var locationUrl = new LocationHtml5Url('http://host.com/', 'http://host.com/', 'http://host.com/');
```
gkalpak: Is the `basePrefix` really needed here ?

petebacondarwin: Oops. No that was a bit of over zealous cutting and pasting :-)

Narretz:
Is this resolved by https://github.com/angular/angular.js/commit/9e492c358c19549696577c86c2c61b93f50ab356 and https://github.com/angular/angular.js/commit/92c7ce5becb9b90216c49f50e0f57f5dbeda8eba?

petebacondarwin:
Yes.

PR #12362, opened 2015-07-16
I18n changes

### LINKED ISSUE #10158: Negative currency amount is displayed as $-xxx,xxx.xx for en-us culture
Shouldn't a negative currency amount be displayed -$xxx,xxx.xx for en-us ?

With that [commit](https://github.com/angular/angular.js/commit/0eb2c2af82d09f8494e44952a0a385ce7584fb4c) it will show as $-xxx,xxx.xx which is kind of weird...

### LINKED ISSUE #12134: Inconsistent in en_US localization
Hi, during development on localizations I've noticed that in core angular code localization for ```en_US``` looks different than in a file with localizations exposed as bower dep.

value in angular core
https://github.com/angular/angular.js/blob/291d7c467fba51a9cb89cbeee62202d51fe64b09/src/ng/locale.js#L36

value in bower dep
https://github.com/angular/bower-angular-i18n/blob/master/angular-locale_en-us.js#L118

I'm not sure which one is correct so don't want to create PR but would be great if we could have the same value in both cases.

petebacondarwin:
Closes #10158
Closes #8174 
Closes #12134


lgalfaso:
Please add Closes #12307. Otherwise, LGTM

PR #12372, opened 2015-07-17
docs($q): Missing docs for when / resolve callback arguments

ArchmageInc:


gkalpak:
Is it even possible to call the error or progress callback ?
Are they there to be compatiblt with `Q` or something ?

PR #12373, opened 2015-07-17
Bug: Address issue where headers modified in transformers are ignored

snayagar:
This bug looks like it was introduced in Angular 1.4.0 as I found it when upgrading from angular 1.2.26.

The issue is reproduced when you define a request transformer that modifies the headers. In angular 1.3.x and lower you are able to change the request headers in the transformer and inject additional / modify request parameters.

In Angular 1.4.0 and newer this no longer works any more.


snayagar:
Looks like this is by design...I don't know why.

PR #12377, opened 2015-07-18
Adding 'null' checks

RolfDeVries:
Under some race conditions $window will be null or undefined

### DIFF
--- a/src/ng/anchorScroll.js
+++ b/src/ng/anchorScroll.js
@@ -231,7 +231,7 @@ function $AnchorScrollProvider() {
           var elemTop = elem.getBoundingClientRect().top;
           $window.scrollBy(0, elemTop - offset);
         }
-      } else {
+      } else if ($window) {
         $window.scrollTo(0, 0);
       }
     }


RolfDeVries:
I signed it!

lgalfaso:
Would it be possible to see a plunker and a test that triggers this issue?

RolfDeVries:
I'm sorry the issue was filed a while ago in our app, and we don't have a test not a plunker.
As I agree the issue is somewhat strange, because $window is initialized and should not become null.
It is found only on a WP8.1 which makes it even harder to debug.
But as you can see the check is quite simple and harmless, so should cause no issues.
Can you please pull in the request so we don't have to change the code after each new release, thanks!

Here is some more detailed information for our logs:

  "device": "{\"ua\":\"Mozilla/5.0 (Mobile; Windows Phone 8.1; Android 4.0; ARM; Trident/7.0; Touch; WebView/2.0; rv:11.0; IEMobile/11.0; NOKIA; Lumia 820) like iPhone OS 7_0_3 Mac OS X AppleWebKit/537 (KHTML, like Gecko) Mobile Safari/537\",\"browser\":{\"name\":\"IEMobile\",\"version\":\"11.0\",\"major\":\"11\"},\"engine\":{\"name\":\"Trident\",\"version\":\"7.0\"},\"os\":{\"name\":\"Windows Phone\",\"version\":\"8.1\"},\"device\":{\"model\":\"Lumia 820\",\"vendor\":\"Nokia\",\"type\":\"mobile\"},\"cpu\":{}}",
      "timestamp": "2015-06-23T08:11:39.171Z",
      "category": "ERROR",
      "message": "Unexpected error occoured: Unable to get property 'scrollTo' of undefined or null reference\r\nTypeError: Unable to get property 'scrollTo' of undefined or null reference\n   at Anonymous function

lgalfaso:
@RolfDeVries I understand that this change is somehow innocuous, but without a test nor a plunker that shows the error, then I lean towards not merging this.

RolfDeVries:
I understand your point, but unfortunally I cannot deliver, thanks anyway.

PR #12378, opened 2015-07-18
docs(filter.js): documents second parameter to .register

matthewjh:
Is there any reason the second parameter of $FilterProvider#register wasn't documented? If not, here.


lgalfaso:
I do not expect a lot of people to actually use this API, but `module.filter`, anyhow, thanks for the PR.

matthewjh:
Agreed Lucus. I used it for the first time with typescript and the definition was off (it spoke nothing of a second parameter), presumably because of this doc.





On Sun, Jul 19, 2015 at 5:52 AM -0700, "Lucas Mirelmann" <notifications@github.com> wrote:
I do not expect a lot of people to actually use this API, but `module.filter`, anyhow, thanks for the PR.

---

PR #12381, opened 2015-07-19
Fix($animateCss): should clear cache if element has no animation.

### LINKED ISSUE #12214: $animate.enter/leave does not animate properly when animation property is defined by css's descandant selector
I have a case that need to apply different animation style to a dynamically generated content that enter/leave the scene. The animation is defined in CSS like:

    .slide-left [data-content] ng-enter { animation: ... }
    .slide-left [data-content] ng-leave { animation: ... }

The html would be

    <div class="container" ng-class="transitionStyle">
        
    </div>

When add/remove content, the following is called

    $scope.transitionStyle = 'slide-left';
    ...
    $animate.leave(oldContent);
    $animate.enter(newContent, $('.container'));

The problem is that if once the transition style has no animation (i.e. `$scope.transitionStyle = ''`), it will be remembered and the following change to transition style (e.g. `$scope.transitionStyle = 'slide-left'`) cannot be applied.

Please check this [plunker](http://plnkr.co/edit/iDCj7f1ivVjqWLEhDiVs?p=preview), if you click 'slide left' or 'slide right' first, the animation is fine. After you have clicked 'no style', the will be no animation even if you click 'slide left/right'.

Normally, I should apply the style to the dynamic content itself. But in this case the the dynamic content is generated and `enter/leave` by 3rd party library (ui-router) which I have no control over.

sreeramu:
We should remove animation cache if the element don't have animation so follow-up animation on the same element will not be from cache.

This PR will close #12214 #12518

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -777,6 +777,8 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
           cancel: cancelFn
         });
 
+        // should flush the cache animation
+        waitUntilQuiet(noop);
         close();
 
         return {
--- a/test/ngAnimate/animateCssSpec.js
+++ b/test/ngAnimate/animateCssSpec.js
@@ -1277,6 +1277,25 @@ describe("ngAnimate $animateCss", function() {
       expect(element.attr('style')).not.toContain('transition');
     }));
 
+    it("should clear cache if no animation so follow-up animation on the same element will not be from cache",
+      inject(function($animateCss, $rootElement, $$body, $$rAF) {
+        var element = jqLite('<div class="rclass"></div>');
+        var options = {
+          event: 'enter',
+          structural: true
+        };
+        $rootElement.append(element);
+        $$body.append($rootElement);
+        var animator = $animateCss(element, options);
+        expect(animator.$$willAnimate).toBeFalsy();
+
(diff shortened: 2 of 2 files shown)


sreeramu:
@matsko can you check this PR, is the fix is OK.?

matsko:
Hey @sreeramu. Thanks again for putting in work into fixing ngAnimate.

The cache needs to be cleared in the next RAF. The reason for this is if a big list of repeated items attempts to detect an animation then it will only perform the getComputedStyle detection code once. Once things are quiet then it clears the cache. The final code should look like this:

```js
      function closeAndReturnNoopAnimator() {
        runner = new $$AnimateRunner({
          end: endFn,
          cancel: cancelFn
        });

        // this will clear the cache internally
        waitUntilQuiet(noop);
        close();

        return {
          $$willAnimate: false,
          start: function() {
            return runner;
          },
          end: endFn
        };
      }
```

That one line is all that needs to be changed. Then inside of your test code just add a call to `$$rAF.flush()` before running the next animation.

sreeramu:
Thanks @matsko for explaining the issue clearly, i had updated the PR according to your comments.

matsko:
Landed as https://github.com/angular/angular.js/commit/0a75a3db6ef265389c8c955981c2fe67bb4f7769. Thank you @sreeramu.

PR #12398, opened 2015-07-21
fix(httpParamSerializerJQLike): Follow jQuery for arrays of objects

### LINKED ISSUE #12393: $httpParamSerializerJQLike doesn't encode array indices (as opposed to $.param)
Why does pressing Enter while in the title posts the issue?

Anyway, found an important encoding difference between `$.param` and `$httpParamSerializerJQLike`.
When encoding arrays, sometimes it's fine to leave the index out like that: `a[]=1&a[]=2`, which comes out as `a = [1, 2]` on the other side.

Sometimes this is not the case, mostly when you have array of objects: `a[][x]=1&a[][y]=2`, which comes out as `a = [{ x: 1, y: 2 }]`, which is sometimes true, but what if you meant `a = [{ x: 1 }, { y: 2 }]`? In this case you need to encode the array index like so: `a[0][x]=1&a[1][y]=2`. JQuery does that with `$.param`, and Rails for example knows how to handle it.

If indices are not given, the server only appends a new object to the array when there are duplicate keys, like so: `a[][x]=1&a[][y]=2&a[][x]=3`, which should come out as `a = [{ x: 1, y: 2 }, { x: 3 }]`.

lgalfaso:
Follow jQuery when serializing arrays that contain objects

Close #12393

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -108,8 +108,8 @@ function $HttpParamSerializerJQLikeProvider() {
       function serialize(toSerialize, prefix, topLevel) {
         if (toSerialize === null || isUndefined(toSerialize)) return;
         if (isArray(toSerialize)) {
-          forEach(toSerialize, function(value) {
-            serialize(value, prefix + '[]');
+          forEach(toSerialize, function(value, index) {
+            serialize(value, prefix + '[' + (isObject(value) ? index : '') + ']');
           });
         } else if (isObject(toSerialize) && !isDate(toSerialize)) {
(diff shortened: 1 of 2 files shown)


odedniv:
Sorry, I feel stupid not creating this PR myself. Next time beer is on me.

Narretz:
LGTM

lgalfaso:
landed as 18a2e4fbfc44216c31bbcdf7705ca87c53e6f1fa

Issue #12405, opened 2015-07-22
orderBy $index descending with track by $index not working


OlsonDev:
Hi. I came across #8334 but it's slightly different than my use case. I have an array of `0`'s and `1`'s I need to list in reverse order. I put together a [plunk](http://plnkr.co/edit/cnNaRr7Wo68lSEFjQugr?p=preview) to demonstrate the issue. In the plunk, I made it an array of `0`, `1`, `2`, & `3` to demonstrate the issue more clearly. Because my array items are not objects I need to use `track by $index` but I need the array reversed visually so I need an `orderBy` filter. I'd rather not duplicate the array and reverse it because then that duplication would happen every digest cycle or I'd have to track changes manually. Many others seem content with [this solution](http://stackoverflow.com/a/15267754/221867) though. However, even back in 2013, people seemed to think [this solution should work](http://stackoverflow.com/a/18947039/221867). They just didn't file an issue... so I am.

PR #12408, opened 2015-07-22
feat(orderBy): Stable sort the input array

lgalfaso:
Stable sort the input array

### DIFF
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -183,6 +183,10 @@ function orderByFilter($parse) {
     if (sortPredicate.length === 0) { sortPredicate = ['+']; }
 
     var predicates = processPredicates(sortPredicate, reverseOrder);
+    // Add a predicate at the end that evaluates to the element index. This makes the
+    // sort stable as it works as a tie-breaker when all the input predicates cannot
+    // distinguish between two elements.
+    predicates.push({ get: function() { return {}; }, descending: reverseOrder ? -1 : 1});
 
     // The next three lines are a version of a Swartzian Transform idiom from Perl
     // (sometimes called the Decorate-Sort-Undecorate idiom)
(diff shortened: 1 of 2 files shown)


lgalfaso:
This indirectly fixes #12405

Narretz:
Does this have a performance implications?

lgalfaso:
> Does this have a performance implications?
>
 This only adds a natural comparator with the original element position, so
this adds some cycles but much smaller than even one $parse evaluation

src/ng/filter/orderBy.js
```diff
@@ -183,6 +183,7 @@ function orderByFilter($parse) {
     if (sortPredicate.length === 0) { sortPredicate = ['+']; }
 
     var predicates = processPredicates(sortPredicate, reverseOrder);
+    predicates.push({ get: function() { return {}; }, descending: reverseOrder ? -1 : 1});
```
caitp: I think we should leave a comment explaining this, it won't be very nice to have to look up this issue in the blame for this function to figure out what this means once it stops being clear.

Just a quick header comment with an overview explaining how the predicate works, imho

caitp:
> This only adds a natural comparator with the original element position, so
this adds some cycles but much smaller than even one $parse evaluation

It might still be a good idea to look at the impact this has on the bigtable benchmarks, and put some numbers in the commit message and issue --- so that there's a bit more confidence there. Can we do that?

lgalfaso:
Run order_by benchpress several times with and without the patch

With the patch
mean: 152.74ms ± 22% 
(stddev 33.61)

Without the patch
mean: 156.36ms ± 32% 
(stddev 53.76) 


This PR does not make the code any faster (as the numbers look like they claim), but do show that the variations between runs is much bigger than whatever performance penalty this patch adds

Narretz:
Can you add to the commit message that it fixes #12405? Otherwise LGTM

PR #12415, opened 2015-07-23
feat(ngSrcAlt): add ng-src-alt directive for image error handling

Nano1237:
I've added a simple directive, that replaces the `src` attribute of an `` Tag with the content of the `ngSrcAlt` Attribute if the file linking to doesn't exist.

That means this:

``` html
<img src="someNonExisting.png" ng-src-alt="{{'someExisting.png'}}" >
```

Would then resolve to this:

``` html
<img src="someExisting.png" ng-src-alt="{{'someExisting.png'}}">
```

if the image `someNonExisting.png` does not exist.

### DIFF
--- a/angularFiles.js
+++ b/angularFiles.js
@@ -69,6 +69,7 @@ var angularFiles = {
     'src/ng/directive/ngPluralize.js',
     'src/ng/directive/ngRepeat.js',
     'src/ng/directive/ngShowHide.js',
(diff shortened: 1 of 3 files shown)


Nano1237:
I signed it!

Narretz:
Hi @Nano1237, thanks for your contribution, but I think this is something that should be a 3rd party module / directive, as it's not a super-common use case.

PR #12416, opened 2015-07-23
refactor(ngModel): use minErr directly for correct doc creation

Narretz:
The minerr_errors extractor of the closure compiler cannot detect minErrs that are not
thrown in the form of 'throw minErr(...'

This causes the error page code blocks which display the error template to stay empty.

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -16,8 +16,7 @@ var VALID_CLASS = 'ng-valid',
     TOUCHED_CLASS = 'ng-touched',
     PENDING_CLASS = 'ng-pending';
 
-
-var $ngModelMinErr = new minErr('ngModel');
+var ngModelMinErr = minErr('ngModel');
 
 /**
  * @ngdoc type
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngModel.js
```diff
@@ -599,7 +596,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
       forEach(ctrl.$asyncValidators, function(validator, name) {
         var promise = validator(modelValue, viewValue);
         if (!isPromiseLike(promise)) {
-          throw $ngModelMinErr("$asyncValidators",
+          throw minErr('ngModel')("$asyncValidators",
```
Narretz: This is actually a separate problem, too: there's no error page for "$asyncValidators" (and the err code doesn't conform to our standards)

Narretz:
@lgalfaso can you give this a quick review?

lgalfaso:
LGTM

Narretz:
I realized that the closure compiler can detect the minErr if is assigned to another variable, we just need to remove the `new`, as it is done in the Angular.js file.

PR #12418, opened 2015-07-23
fix(ngModel): remove reference to parentForm from removed control

### LINKED ISSUE #12263: Input validity reflected on form after `$removeControl`
After removing a field from a form using `$removeControl`, if you change the field's value [to be invalid] it will affect the validity of the form it was removed from.
Not too sure if this is working as expected or not, but seems to be undesired.
Affected versions [at least] 1.3.x, 1.4.x

Steps to reproduce (given the following plnkr)
http://plnkr.co/edit/NZJn0zwWdKxvrIlQxUmn?p=preview

1. Add an invalid value into the field (it accepts numbers).
2. Note the validity of the form and field. (Form: invalid; Field: invalid)
3. Click the button to remove the control via `$removeControl(...)`
4. Note the validity of the form and field. (Form: valid; Field: [not on formCtrl])
5. Edit the field to be invalid.
6. Note the validity of the form and field. (Form: invalid; Field: [not on formCtrl])

For step 6 I would expect the form to no longer be affected by the field, especially since step 4 set the form to valid when the invalid field was removed.

Narretz:
Otherwise, once the removed control's validity changes,
it will continue to be reflected on its former parent form.

Fixes #12263

@gkalpak can you take a look?

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -64,8 +64,6 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
   var form = this,
       controls = [];
 
-  var parentForm = form.$$parentForm = element.parent().controller('form') || nullFormCtrl;
-
   // init state
   form.$error = {};
   form.$$success = {};
@@ -76,8 +74,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
   form.$valid = true;
   form.$invalid = false;
   form.$submitted = false;
-
-  parentForm.$addControl(form);
+  form.$$parentForm = nullFormCtrl;
 
   /**
    * @ngdoc method
@@ -131,6 +128,8 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
     if (control.$name) {
       form[control.$name] = control;
     }
+
+    control.$$parentForm = form;
   };
 
   // Private API: rename a form control
@@ -204,7 +203,6 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
         delete object[property];
       }
     },
-    parentForm: parentForm,
     $animate: $animate
   });
 
@@ -223,7 +221,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
     $animate.addClass(element, DIRTY_CLASS);
     form.$dirty = true;
     form.$pristine = false;
-    parentForm.$setDirty();
+    form.$$parentForm.$setDirty();
   };
 
   /**
@@ -279,7 +277,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
   form.$setSubmitted = function() {
     $animate.addClass(element, SUBMITTED_CLASS);
     form.$submitted = true;
-    parentForm.$setSubmitted();
+    form.$$parentForm.$setSubmitted();
   };
 }
 
@@ -453,6 +451,7 @@ var formDirectiveFactory = function(isNgForm) {
     var formDirective = {
       name: 'form',
       restrict: isNgForm ? 'EAC' : 'E',
+      require: ['form', '^^?form'], //first is the form's own ctrl, second is an optional parent form
       controller: FormController,
       compile: function ngFormCompile(formElement, attr) {
         // Setup initial state of the control
@@ -461,7 +460,9 @@ var formDirectiveFactory = function(isNgForm) {
         var nameAttr = attr.name ? 'name' : (isNgForm && attr.ngForm ? 'ngForm' : false);
 
         return {
-          pre: function ngFormPreLink(scope, formElement, attr, controller) {
+          pre: function ngFormPreLink(scope, formElement, attr, ctrls) {
+            var controller = ctrls[0];
+
             // if `action` attr is not present on the form, prevent the default action (submission)
             if (!('action' in attr)) {
               // we can't use jq events because if a form is destroyed during submission the default
@@ -490,7 +491,9 @@ var formDirectiveFactory = function(isNgForm) {
               });
             }
 
-            var parentFormCtrl = controller.$$parentForm;
+            var parentFormCtrl = ctrls[1] || controller.$$parentForm;
+            parentFormCtrl.$addControl(controller);
+
             var setter = nameAttr ? getSetter(controller.$name) : noop;
 
             if (nameAttr) {
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -167,6 +167,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
     });
 
     arrayRemove(controls, control);
+    control.$$parentForm = nullFormCtrl;
   };
 
 
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -113,11 +113,23 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
   /**
    * @ngdoc method
    * @name form.FormController#$addControl
+   * @param {object} control control object, either a {@link form.FormController} or an
+   * {@link ngModel.NgModelController}
    *
    * @description
-   * Register a control with the form.
+   * Register a control with the form. Input elements using ngModelController do this automatically
+   * when they are linked.
    *
-   * Input elements using ngModelController do this automatically when they are linked.
+   * Note that the current state of the control will not be reflected on the new parent form. This
+   * is not an issue with normal use, as freshly compiled and linked controls are in a `$pristine`
+   * state.
+   *
+   * However, if the method is used programmatically, for example by adding dynamically created controls,
+   * or controls that have been previously removed without destroying their corresponding DOM element,
+   * it's the developers responsiblity to make sure the current state propagates to the parent form.
+   *
+   * For example, if an input control is added that is already `$dirty` and has `$error` properties,
+   * calling `$setDirty()` and `$validate()` afterwards will propagate the state to the parent form.
    */
   form.$addControl = function(control) {
     // Breaking change - before, inputs whose name was "hasOwnProperty" were quietly ignored
@@ -146,11 +158,18 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
   /**
    * @ngdoc method
    * @name form.FormController#$removeControl
+   * @param {object} control control object, either a {@link form.FormController} or an
+   * {@link ngModel.NgModelController}
    *
    * @description
    * Deregister a control from the form.
    *
    * Input elements using ngModelController do this automatically when they are destroyed.
+   *
+   * Note that only the removed control's validation state (`$errors`etc.) will be removed from the
+   * form. `$dirty`, `$submitted` states will not be changed, because the expected behavior can be
+   * different from case to case. For example, removing the only `$dirty` control from a form may or
+   * may not mean that the form is still `$dirty`.
    */
   form.$removeControl = function(control) {
     if (control.$name && form[control.$name] === control) {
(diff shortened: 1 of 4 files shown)


src/ng/directive/form.js
```diff
@@ -168,6 +168,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
     });
 
     arrayRemove(controls, control);
+    control.$$parentForm = nullFormCtrl;
```
gkalpak: I think a similar change is need in [`$addControl()`][1]: `control.$$parentForm = form`

[1]: https://github.com/angular/angular.js/pull/12418/files#diff-1838ed9bd28307b6d42de98549826ae5R134

src/ng/directive/ngModel.js
```diff
@@ -1311,7 +1309,7 @@ function addSetValidityMethod(context) {
     }
 
     toggleValidationCss(validationErrorKey, combinedState);
-    parentForm.$setValidity(validationErrorKey, combinedState, ctrl);
+    ctrl.$$parentForm.$setValidity(validationErrorKey, combinedState, ctrl);
```
gkalpak: I think a similar change is needed in [`$setDirty()`][1]: `ctrl.$$parentForm.$setDirty()`

[1]: https://github.com/Narretz/angular.js/blob/fix-parent-form/src/ng/directive/form.js#L222

gkalpak:
Since nested `ngForm`s are also reated as controls to parent-forms (i.e. can be added/removed via `$add/removeControl()`), similar changes need to be made in `formDirective` (i.e. referencing `form.$$parentForm` instead of the _private_ `parentForm` variable.

Note, that this issue does not only affect validity states, but other properties or forms/controls, such as `$dirty`, `$submitted` etc.

It would also be nice to have more thorough testing (testing those other properties - not only `$error`).

--
But other than the comments above, it LGTM :smile:

gkalpak:
(And of course, there is the fail on Travis, but it's an easy fix :stuck_out_tongue:)

Narretz:
@gkalpak Thanks for the review! Do you remember what failure Travis recorded? It's such a pain to look through the logs.

src/ng/directive/form.js
caitp: seems right

caitp:
> @gkalpak Thanks for the review! Do you remember what failure Travis recorded? It's such a pain to look through the logs.

Running "jshint:ng" (jshint) task
   src/ng/directive/ngModel.js
    241 |  var parentForm = this.$$parentForm = $element.inheritedData('$formController') || nullFormCtrl
                                                                                                         ^ Missing semicolon.
>> 1 error in 70 files
Warning: Task "jshint:ng" failed. Use --force to continue.
Aborted due to warnings.
The command "./scripts/travis/build.sh" exited with 3.
cache.2
store build cache
1.74snothing changed, not updating cache
$ ./scripts/travis/print_logs.sh

test/ng/directive/ngModelSpec.js
```diff
@@ -19,7 +19,6 @@ describe('ngModel', function() {
       };
 
       element = jqLite('<form><input></form>');
-      element.data('$formController', parentFormCtrl);
```
Narretz: Since the parentFormCtrl is now extracted via regular require, this mechanism doesn't work anymore. This might break people's tests.

Narretz:
Thanks @caitp 
This turned out to be a bit more complex, since the adding of controls is handled differently in forms / ngModel. I refactored this now, so that the initial adding of controls happens in the preLink function for both form and ngModel (this actually saves one instance of accessing the parent form controller in ngModel). The logic of assigning the parentForm to the control is moved to the $addControl.
When a control is added, its validity, dirty and submit states will be reflected automatically on the parent form. 
I'll have to test if these changes impact performance.

Narretz:
@gkalpak / @caitp can you please take another look?

src/ng/directive/form.js
```diff
@@ -131,6 +128,28 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
     if (control.$name) {
       form[control.$name] = control;
     }
+
+    control.$$parentForm = form;
+
+    forEach(control.$pending, function(value, name) {
```
caitp: this may not be a big deal in most cases, but 3x `forEach()` calls seems like a perf hit, which would be most noticeable in forms with visibility controlled by ng-if or ng-switch. I'm not saying it's a huge problem, but yeah

caitp:
apart from 1 comment, 3rd commit looks ok to me

src/ng/directive/form.js
```diff
@@ -131,6 +128,28 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
...
+    forEach(control.$$success, function(value, name) {
+      form.$setValidity(name, true, control);
+    });
+
+    if (control.$dirty && form.$pristine) {
+      form.$setDirty();
+    }
+
+    if (control.$submitted && !form.$submitted) {
+      form.$setSubmitted();
```
gkalpak: Do we really want this ?
It is really a corner case, but if we already have unsubmitted form (with more fields/child-forms), then adding a submitted child-form shouldn't necessarily mark the form as submitted.
(The same holds for a submitted form and unsubmitted child-form being added.)

Neither way seems intuitive (so maybe it's better to leave the property untouched).

test/ng/directive/formSpec.js
```diff
@@ -58,6 +58,78 @@ describe('form', function() {
...
+    var form = scope.myForm;
+
+    var input = doc.find('input').eq(0);
+    var inputController = input.controller('ngModel');
+
+    // remove control so we can add it again
+    form.$removeControl(form.control);
+    expect(form.control).toBeUndefined();
+    expect(form.$error.maxlength).toBeFalsy();
+    expect(form.$dirty).toBe(false);
```
gkalpak: I would move these 3 expectations at line 119 (after the input has been edited, but before it has been added to the form), in order to verify that the form has been affected **only** after the control has been added back).
(It's not a big deal, since the previous test already verifies that a removed control does not affect the form, but it still makes me feel better :smiley:)

test/ng/directive/formSpec.js
```diff
@@ -577,6 +649,83 @@ describe('form', function() {
...
+
+      changeInputValue(input, 'abc');
+      scope.$apply();
+
+      expect(form.$error.maxlength).toBeFalsy();
+      expect(form.$dirty).toBe(false);
+    });
+
+
+    it('should react ti changes in manually added child forms', function() {
```
gkalpak: _ti_  -->  _to_

test/ng/directive/formSpec.js
```diff
@@ -577,6 +649,83 @@ describe('form', function() {
...
+    it('should ignore changes in manually removed child forms', function() {
+      doc = $compile(
+        '<form name="myForm">' +
+          '<ng-form name="childform">' +
+            '<input name="childformcontrol" ng-maxlength="1" ng-model="value"/>' +
+          '</ng-form>' +
+        '</form>')(scope);
+
+      var form = scope.myForm;
+      var childformController = doc.find('ng-form').eq(0).controller('form');
```
gkalpak: You know you could simply write `form.childform`, right ? :stuck_out_tongue: 
<sub>(It applies in all tests.)</sub>

src/ng/directive/form.js
gkalpak: I guess this is in accordance to what happens in `$setSubmitted()` (where the parent form is also marked as submitted), but it feels unintuitive.

Anyway, since this is how things are, let's leave it as is.

src/ng/directive/form.js
```diff
@@ -490,7 +513,9 @@ var formDirectiveFactory = function(isNgForm) {
               });
             }
 
-            var parentFormCtrl = controller.$$parentForm;
+            var parentFormCtrl = ctrls[1] || ctrls[0].$$parentForm;
+            parentFormCtrl.$addControl(controller);
```
gkalpak: Hard-coding `parentFormCtrl`suffers from the same issues in [`attr.$observe(...)`][1] and [`formElement.on('$destroy', ...)`][2]; i.e. if the form has been since removed and re-added to another parent form, it won't be properly renamed/removed.

You need to use `ctrls[0].$$parentForm`.

(It's OK to use `parentFormCtrl` on line 517 though.)

[1]: https://github.com/angular/angular.js/pull/12418/files#diff-1838ed9bd28307b6d42de98549826ae5R523
[2]: https://github.com/angular/angular.js/pull/12418/files#diff-1838ed9bd28307b6d42de98549826ae5R531

src/ng/directive/form.js
```diff
@@ -204,7 +224,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
         delete object[property];
       }
     },
-    parentForm: parentForm,
+    parentForm: form.$$parentForm,
```
gkalpak: This property is not used, since you are accessing the parent form via `ctrl.$$parentForm` inside `addSetValidity()`.

src/ng/directive/ngModel.js
```diff
@@ -1022,7 +1020,7 @@ var ngModelDirective = ['$rootScope', function($rootScope) {
       return {
         pre: function ngModelPreLink(scope, element, attr, ctrls) {
           var modelCtrl = ctrls[0],
-              formCtrl = ctrls[1] || nullFormCtrl;
+              formCtrl = ctrls[1] || ctrls[0].$$parentForm;
```
gkalpak: Similar problem as https://github.com/angular/angular.js/pull/12418#discussion_r38092011.

test/ng/directive/formSpec.js
Narretz: I think I was trying to be safe in case the form gets removed etc. But I guess it does work with these tests

Narretz:
Thanks for the review @gkalpak, I have fixed the cases where the wrong parent was referenced.
I've also reverted the behavior where calling addControl would try to set the current state (error, success, pending, dirty, submitted) on the parentForm. This way, we match the previous behavior, and we avoid possible performance penalties. I've added documentation about this.

test/ng/directive/formSpec.js
```diff
@@ -58,6 +58,114 @@ describe('form', function() {
...
+    expect(form.control).toBeUndefined();
+    expect(form.$error.maxlength).toBeFalsy();
+    expect(form.$dirty).toBe(false);
+
+    //re-add the control
+    form.$addControl(inputController);
+    changeInputValue(input, 'a');
+
+    expect(form.$error.maxlength).toBeFalsy();
+  });
```
gkalpak: In this test you are not actually testing that the form has reacted to changes in the added control.
After you add the control you only check `form.$error.maxlength` (which hasn't changed since the last time you checked).

test/ng/directive/formSpec.js
```diff
@@ -58,6 +58,114 @@ describe('form', function() {
...
+
+    expect(form.childControlMoved).toBeUndefined();
+    expect(otherForm.childControl).toBeUndefined();
+    expect(otherForm.childControlMoved).toBe(childControl);
+
+    scope.hasChildControl = false;
+    scope.$digest();
+
+    expect(form.childControlMoved).toBeUndefined();
+    expect(otherForm.childControlMoved).toBeUndefined();
```
gkalpak: lol

test/ng/directive/formSpec.js
```diff
@@ -547,35 +655,147 @@ describe('form', function() {
...
+      changeInputValue(input, 'ab');
+
+      expect(form.childform).toBeUndefined();
+      expect(form.$dirty).toBe(false);
+      expect(form.$error.maxlength).toBeFalsy();
+
+      form.$addControl(childformController);
+      changeInputValue(input, 'a');
+
+      expect(form.$error.maxlength).toBeFalsy();
```
gkalpak: Same as https://github.com/angular/angular.js/pull/12418/files#r38523628.

gkalpak:
Apart for a couple of nitpicks in test, it LGTM :+1:

test/ng/directive/formSpec.js
Narretz: Hmm I don't understand!

gkalpak: Sorry - I was just thinking loud (laughing actually).

I find it funny (in a good way), that you can have a control be one form's child (in DOM), but belong to another form (in Angular terms).
I thought it was the wrong expectation at first site, but it is correct and find it nice that it works.

Sorry for the noise and confusion :disappointed_relieved:

Narretz:
Thx again for the reviews! Landed as f8a07dd^..aa11dfc

PR #12421, opened 2015-07-23
docs($http): update docs to include statusText

arkarkark:
also wondering if response.statusText should be included around
https://github.com/angular/angular.js/blob/master/src/ng/http.js#L961
https://github.com/angular/angular.js/blob/master/src/ng/http.js#L970

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -402,11 +402,11 @@ function $HttpProvider() {
      * ```js
      *   // Simple GET request example :
      *   $http.get('/someUrl').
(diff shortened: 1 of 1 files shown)


arkarkark:
I signed it!

caitp:
If I recall correctly, `statusText` is not available to the success/error methods, only `then()` (and conferring with @petebacondarwin it sounds like I do), I don't think we want to merge this one. The success/error api are in the process of being deprecated and aren't going to have new functionality added to them.

PR #12425, opened 2015-07-23
fix($injector): Allows ES6 function syntax

### LINKED ISSUE #12424: ES2015 syntax results in TypeError: Cannot read property '1' of null
I put together [this plunk](http://plnkr.co/edit/09uiBvV5CUV82Tk3fH8v?p=preview) to demonstrate the issue.

The issue itself (I'm reasonably sure) is the dependency annotator doesn't understand [ES2015 syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_6) for shorthand method names even though the browser does (Chrome v43.0.2357.134 m, Windows 10).

The weird thing is it only matters for the `controller` key, best I can tell. If I use the same syntax for the `link` key, the `TypeError` doesn't happen and `$scope` is successfully injected.

Here's the full stack trace:
```
TypeError: Cannot read property '1' of null
    at Function.annotate (angular.js:3725)
    at Object.invoke (angular.js:4428)
    at $get.extend.instance (angular.js:8999)
    at nodeLinkFn (angular.js:8109)
    at compositeLinkFn (angular.js:7541)
    at compositeLinkFn (angular.js:7545)
    at compositeLinkFn (angular.js:7545)
    at publicLinkFn (angular.js:7416)
    at angular.js:1635
    at Scope.$get.Scope.$eval (angular.js:15846)
```

lgalfaso:
Closes #12424

### DIFF
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -237,6 +237,15 @@ describe('injector', function() {
     });
 
 
+    // Only Chrome and Firefox support this syntax.
+    if (/chrome|firefox/i.test(navigator.userAgent)) {
+      it('should be possible to annotate functions that are declared using ES6 syntax', function() {
+        // The function is generated using `eval` as just having the ES6 syntax can break some browsers.  
+        expect(annotate(eval('({ fn(x) { return; } })').fn)).toEqual(['x']);
+      });
+    }
+
+
     it('should publish annotate API', function() {
       expect(angular.mock.$$annotate).toBe(annotate);
       spyOn(angular.mock, '$$annotate').andCallThrough();
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -62,7 +62,7 @@
  * Implicit module which gets automatically added to each {@link auto.$injector $injector}.
  */
 
-var FN_ARGS = /^function\s*[^\(]*\(\s*([^\)]*)\)/m;
+var FN_ARGS = /^[^\(]*\(\s*([^\)]*)\)/m;
 var FN_ARG_SPLIT = /,/;
 var FN_ARG = /^\s*(_?)(\S+?)\1\s*$/;
 var STRIP_COMMENTS = /((\/\/.*$)|(\/\*[\s\S]*?\*\/))/mg;


lgalfaso:
BTW, the travis failures are because the Chrome and Firefox versions running on Travis do not support this syntax

test/auto/injectorSpec.js
```diff
@@ -237,6 +237,15 @@ describe('injector', function() {
     });
 
 
+    // Only Chrome and Firefox support this syntax.
+    if (/chrome|firefox/i.test(navigator.userAgent)) {
+      it('should be possible to annotate functions that are declared using ES6 syntax', function() {
+        // The function is generated using `eval` as just having the ES6 syntax can break some browsers.  
+        expect(annotate(eval('({ fn(x) { return; } })').fn)).toEqual(['x']);
```
caitp: Can you explain this better? How does it break?

lgalfaso: If I change the test to just
```js
expect(annotate(({ fn(x) { return; } }).fn)).toEqual(['x']);
```

then the JS parser complains about the syntax. It claims there is a missing `:` after `fn` and stops

caitp: Which JS parser? What browser versions are you talking about? Are you talking about the test file failing to parse in older browsers? Or what?

lgalfaso: Without the `eval` trick, Safari 8.0.7 (10600.7.12)  [This is the latest Safari version]
throws this error when parsing the file to run the tests
```
SyntaxError: Unexpected token '('. Expected a ':' following the property name 'fn'.
```
After throwing this error, no tests are executed.

caitp: Alright, I'd rather we jUst put the test in a different sOurce file which is only loaded in known supported browsers, because I really don't want to rely on eval. Take your pick and document clearly the approach you've taken. Otherwise LGTM

test/auto/injectorSpec.js
```diff
@@ -237,6 +237,15 @@ describe('injector', function() {
     });
 
 
+    // Only Chrome and Firefox support this syntax.
+    if (/chrome|firefox/i.test(navigator.userAgent)) {
```
gkalpak: This regex will also match MS Edge's navigator (but it's OK, since it understands the syntax as well).
(But the comment is misleading :P)

lgalfaso: I do not have access to MS Edge and it is currently not supported by saurcelabs, so this will have to make it.

On a side note, I find it somehow strange for MS Edge user agent to have the word "Chrome" in it

gkalpak: I know.
BTW, it also has `Safari` and (of course) `Edge` (for interoperability reasons - or so I've read :P).

PR #12428, opened 2015-07-24
feat(jqLite): change camelCase to convert first letter to lowercase

Strikeskids:
There currently is no way to match class directives of the form
`"MyDirective"` or `"My_Directive"`, because directives must start
with a lowercase letter and the `camelCase` function does not change
the case of the first letter. This change allows matching directives
that start with an uppercase letter.

Before, `camelCase('FooBar') === 'FooBar'` and
`camelCase('Foo_Bar') === 'FooBar'`. Now, the first letter will be
lowercase: `camelCase('FooBar') === 'fooBar'`.

### DIFF
--- a/test/jqLiteSpec.js
+++ b/test/jqLiteSpec.js
@@ -2013,8 +2013,15 @@ describe('jqLite', function() {
 
     it('should covert browser specific css properties', function() {
       expect(camelCase('-moz-foo-bar')).toBe('MozFooBar');
+      expect(camelCase('MozFooBar')).toBe('MozFooBar');
       expect(camelCase('-webkit-foo-bar')).toBe('webkitFooBar');
       expect(camelCase('-webkit-foo-bar')).toBe('webkitFooBar');
+    });
+
+    it('should convert strings that start with an upper-case letter', function() {
+      expect(camelCase('Foo_Bar')).toBe('fooBar');
+      expect(camelCase('Foo-Bar')).toBe('fooBar');
+      expect(camelCase('FooBar')).toBe('fooBar');
     });
   });
 
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -133,7 +133,7 @@ JQLite._data = function(node) {
 function jqNextId() { return ++jqId; }
 
 
-var SPECIAL_CHARS_REGEXP = /([\:\-\_]+(.))/g;
+var SPECIAL_CHARS_REGEXP = /((?:^|[\:\-\_])[\:\-\_]*(.))/g;
 var MOZ_HACK_REGEXP = /^moz([A-Z])/;
 var MOUSE_EVENT_MAP= { mouseleave: "mouseout", mouseenter: "mouseover"};
 var jqLiteMinErr = minErr('jqLite');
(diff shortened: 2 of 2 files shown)


caitp:
I think this change breaks Moz vendor prefixes, which we probably don't want to do despite wanting to discourage the use of vendor prefixes

Strikeskids:
Moz vendor prefixes are already special-cased within the camelCase [function](https://github.com/angular/angular.js/blob/a5221f320a8c1644354003c0e78201add44f11e6/src/jqLite.js#L151) and [spec](https://github.com/angular/angular.js/blob/a5221f320a8c1644354003c0e78201add44f11e6/test/jqLiteSpec.js#L2015). I did not change that behavior.

caitp:
I know it's already handled specially, but I don't see any new tests verifying that it still works properly including newly supported inputs which should result in the Moz prefix

Strikeskids:
I don't understand what you think is lacking. Can you explain a test case that would fail here? I'm reasonably confident that the test cases used cover all of the cases.

caitp:
I'm asking you to add a test case which asserts the vendor prefix hack still works, even on the newly supported inputs in your patch.

Strikeskids:
Given the ordering of my patch and the the vendor prefix hack, I assumed that the single `camelCase('-moz-foo-bar')` would be enough to verify that the hack still works. However, I now understand the desire for a direct test, and have updated accordingly.

caitp:
Thank you =)

test/jqLiteSpec.js
```diff
@@ -2013,9 +2013,16 @@ describe('jqLite', function() {
 
     it('should covert browser specific css properties', function() {
       expect(camelCase('-moz-foo-bar')).toBe('MozFooBar');
+      expect(camelCase('MozFooBar')).toBe('MozFooBar');
```
petebacondarwin: Surely this is not correct. If there is no `-moz-` at the start then we should be lower casing the first character according to the requirements of this patch. In other words I expect the behaviour to be as described in these two expectations...

```js
expect(camelCase('-moz-foo-bar')).toBe('MozFooBar');
expect(camelCase('MozFooBar')).toBe('mozFooBar');
```

Strikeskids:
@petebacondarwin are you suggesting that `camelCase(camelCase(x)) === x` is not true for all x?

What you are suggesting is 

    camelCase('-moz-foo-bar') //=> MozFooBar
    camelCase('MozFooBar') //=> mozFooBar
    camelCase(camelCase('-moz-foo-bar')) //=> mozFooBar

I think that behavior is far weirder than the current behavior.

mgol:
I don't like the idea of generally special-casing strings that start with `moz`. This is non-intuitive, breaks logic encapsulation & mixes the CSS context with a general utility. If we want to do this, I'd prefer to change the general `camelCase` method to not care about `moz` at all and use a separate function inside jqLite for the purposes of CSS prefixing.

mgol:
I've done the camelCasing refactor in https://github.com/angular/angular.js/commit/73050cdda04675bfa6705dc841ddbbb6919eb048.

mgol:
@Strikeskids Would you like to try to update the patch? Note that the `Moz`-related tests are now incorrect and should be removed; we only handle vendor prefixes in jqLite, `$compile` not uses [its own `directiveNormalize` function](https://github.com/angular/angular.js/blob/f1e6778/src/ng/compile.js#L3602).

mgol:
@Strikeskids Since after my changes in jqLite this has to be implemented differently (not in jqLite but in $compile) I'm going to close this PR. Please open a new one if you want to fix that the new way.

PR #12429, opened 2015-07-24
docs($rootScope.Scope): improve clarity describing $watch with no listener

Strikeskids:
The previous explanation in parentheses created a bit of confusion because the documentation stated to leave off the `listener`, but then said "be prepared for multiple calls to your listener". The new explanation clarifies that it is indeed the `watchExpression` that will be executed multiple times.


caitp:
lgtm

PR #12440, opened 2015-07-27
test(expression): add timeout to alert protractor spec to prevent failure

petebacondarwin:
In Chrome, if two alert boxes pop up, without enough time between them,
Protractor (or possibly ChromeDriver) sometimes fails to recognize the
second alert.


lgalfaso:
If this makes it work. LGTM

petebacondarwin:
I merged a better fix thanks to Hank: addb1ae37d775937688ae6f09e6f0ebd79225849

PR #12443, opened 2015-07-27
docs($compile): fix typo

patyatka:
I'm not native speaker, but "like like" construction is weird for me...


patyatka:
I signed it!

caitp:
yes, this looks like an improvement. will land

PR #12444, opened 2015-07-27
chore(angularFiles): add default en-us locale to angularFiles

jeffbcross:
The lack of a default locale was breaking some internal builds 
that depend on some provider of an ngLocale module to be present.

Note: I'm not sure this is the best solution. I'm still talking with @petebacondarwin 
about this. Please don't merge


petebacondarwin:
I don't this works this way because it will put the angular module `ngLocale` "inside" the closure where we define angular itself. These broken projects just need to include an `ngLocale` module in their distribution.

PR #12449, opened 2015-07-28
Add new type of switch item, that accept Variables

deadmann:


deadmann:
sry i forgot to edit other required file, that hold directive

Issue #12442, opened 2015-07-27
Switch


deadmann:
it's kinda good that switch should take direct value, but there are also cases that it is in fault...

Ex.
i have a Scope Variable:

    $scope.WindowType={
        'Users' : 0,
        'Chat'   : 1,
        ...
    }
    $scope.MyModel ={
                ...
                Sections       : {
                    ...
                    SwitchWindow : $scope.WindowType.Chat,
                }
    }

In here, there are three issue,...
1. Readability is low : <div data-ng-switch-when="0">
2. You can't remember the conditions for life time: 
3. Changing Model Cause System Failure...

So i'm wonder if you provide ng-switch with another bind-able directive, or attribute.

deadmann:
I don't know which version of angular i have, but i hope see it in next nuGet package i fetch from server...

i fix it by creating this directive:

    var ngSwitchWhenBindDirective = ngDirective({
        transclude: 'element',
        priority: 1200,
        require: '^ngSwitch',
        multiElement: true,
        link: function (scope, element, attrs, ctrl, $transclude) {
            ctrl.cases['!' + scope.$eval(attrs.ngSwitchWhenBind)] = (ctrl.cases['!' + scope.$eval(attrs.ngSwitchWhenBind)] || []);
            ctrl.cases['!' + scope.$eval(attrs.ngSwitchWhenBind)].push({ transclude: $transclude, element: element });
        }
    });

So older project can continue working...

lgalfaso:
This was discussed and this is not something for the core.

PR #12451, opened 2015-07-28
Add new type of switch item, that accept Variables

deadmann:
There are cases that you may want to define some types, and these case are issued...
read more in here:
https://github.com/angular/angular.js/issues/12442


deadmann:
I signed it!

deadmann:
What is this ? :O

https://travis-ci.org/angular/angular.js/jobs/72984296

 continuous-integration/travis-ci/pr — The Travis CI build failed: ...
Running "jscs:src" (jscs) task
Illegal trailing whitespace at src/ng/directive/ngSwitch.js :
     8 | * @description
     9 | * The `ngSwitch` directive is used to conditionally swap DOM structure on your template based on a scope expression.
    10 | * Elements within `ngSwitch` but without `ngSwitchWhen` or `ngSwitchWhenBind` or `ngSwitchDefault` directives will be 
-------------------------------------------------------------------------------------------------------------------------------^
    11 | * preserved at the location as specified in the template.
    12 | *
Illegal trailing whitespace at src/ng/directive/ngSwitch.js :
    41 | * </ANY>
    42 | * ```
    43 | * 
-----------^
    44 | *  * ```
    45 | * <ANY ng-switch="expression">
Illegal trailing whitespace at src/ng/directive/ngSwitch.js :
    53 | * @scope
    54 | * @priority 1200
    55 | * @param {*} ngSwitch|on expression to match againsts <code>ng-switch-when</code> or 
----------------------------------------------------------------------------------------------^
    56 | * <code>ng-switch-when-bind</code>.
    57 | * On child elements add:
Illegal trailing whitespace at src/ng/directive/ngSwitch.js :
    60 | *   case will be displayed. If the same match appears multiple times, all the
    61 | *   elements will be displayed.
    62 | * * * `ngSwitchWhenBind`: the bindable case statement to match against. If match 
------------------------------------------------------------------------------------------^
    63 | *   then this case will be displayed. If the same match appears multiple times, 
    64 | *   all the elements will be displayed.
Illegal trailing whitespace at src/ng/directive/ngSwitch.js :
    61 | *   elements will be displayed.
    62 | * * * `ngSwitchWhenBind`: the bindable case statement to match against. If match 
    63 | *   then this case will be displayed. If the same match appears multiple times, 
-----------------------------------------------------------------------------------------^
    64 | *   all the elements will be displayed.
    65 | * * `ngSwitchDefault`: the default case when no other case match. If there
Illegal space before opening round brace at src/ng/directive/ngSwitch.js :
   212 |    require: '^ngSwitch',
   213 |    multiElement: true,
   214 |    link: function (scope, element, attrs, ctrl, $transclude) {
--------------------------^
   215 |        ctrl.cases['!' + scope.$eval(attrs.ngSwitchWhenBind)] = (ctrl.cases['!' + scope.$eval(attrs.ngSwitchWhenBind)] || []);
   216 |        ctrl.cases['!' + scope.$eval(attrs.ngSwitchWhenBind)].push({ transclude: $transclude, element: element });
>> 6 code style errors found!
Warning: Task "jscs:src" failed. Use --force to continue.
Aborted due to warnings.
The command "./scripts/travis/build.sh" exited with 3.

DikshaDeo:
Please check the files and let me know.....

deadmann:
@DikshaDeo 
the previous post, shows built error, says it's due to style of writing code, and the white space exist within file, as you can see those errors...

the error happens on Build Job: "JOB=unit BROWSER_PROVIDER=sauselab"

i have no idea what cause this, or what to do

lgalfaso:
Hi, thanks for the PR, but similar issues were discusses several times, and in all cases the result is that we do not want to have expressions in `ng-switch-when`. The main reason being what should be the behavior when the expression value changes? There is simply no good answer that does not also involve adding watchers and killing any performance benefit of `ngSwitch` and that does not make developers surprised about the result.

Sorry, but it looks like we are going to pass on this

deadmann:
We doesn't talk about a normal condition, we talk about times we want define types, and any language allows that, javascript sure doesn't have any type, but has ways to define it, and that's where it come to scene, also this way, it won't effect a normal switch case you always do, or kill any performance, it just do when you need it.

lgalfaso:
@deadmann This was discusses many times, allowing any form of expressions in `ngSwitch` has many side-effects and the option of just evaluation things on link time has its limitations (eg, what would happen if the value is set asynchronously).

deadmann:
angular though us to manage our code from controller, and to have static things over view, it's kinda, you know... and no one change Enum type in middle of application, sure js has no exact enum type, and they are changable, but then should i stop using for example substring function because someone may override it? no, we normally teach them, and here i have ngSwitchWhenBind, which it self says, hey i may change, and the user should be notified that, hey this should be only used when you have statically typed variable...

but, as you recall, i'm not aware of performance side effect, or the limitation of calling $eval function...

but still user can go on and use ngSwitchWhen...

my word are done at this moment, i'm not aware of the system it self it's true, so it's depend on guys such as your self to decide.

BTW, thank you for the time,
Hassan Faghihi

lgalfaso:
@deadmann nothing prevents you from adding this directive to your project and use it in the way that best suits your needs. The question is if the functionality fulfills the expectations to a broader audience to be part of the core

deadmann:
i though they gonna need it, maybe they don't as much as i do

PR #12459, opened 2015-07-28
fix(ngAnimate): ensure that only string-based addClass/removeClass values are applied

### LINKED ISSUE #12458: ngAnimate throws jqLite/jQuery errors if a non-string value is passed as an addClass/removeClass option
This issue came up because of Material, however it is very hard to debug and ngAnimate should just check for this on its own.

matsko:
Closes #12458

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -106,31 +106,31 @@ var $$CoreAnimateQueueProvider = function() {
     };
 
     function addRemoveClassesPostDigest(element, add, remove) {
-      var data = postDigestQueue.get(element);
-      var classVal;
+      var classVal, data = postDigestQueue.get(element);
 
       if (!data) {
         postDigestQueue.put(element, data = {});
         postDigestElements.push(element);
       }
 
-      if (add) {
-        forEach(add.split(' '), function(className) {
-          if (className) {
-            data[className] = true;
-          }
-        });
-      }
+      var updateData = function(classes, value) {
+        var changed = false;
+        if (classes) {
+          classes = isString(classes) ? classes.split(' ') :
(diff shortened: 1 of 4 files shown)


petebacondarwin:
@matsko - looks good but I think the implementation could be DRYer with a helper:

```js
function addRemoveClassesPostDigest(element, add, remove) {
  var data = postDigestQueue.get(element);
  var dataSet = false;
  var classVal;

  var updateData = function(classes, value) {
    var changed = false;
    if (classes) {
      add = isString(classes)
             ? classes.split(' ')
             : isArray(classes) ? classes : [];
      forEach(classes, function(className) {
        if (className) {
          changed = true;
          data[className] = value;
        }
      });
    }
    return changed;
  };

  if (!data) {
    postDigestQueue.put(element, data = {});
    postDigestElements.push(element);
  }

  dataSet = updateData(add, true);
  dataSet = dataSet || updateData(remove, false);


  if (!dataSet || postDigestElements.length > 1) return;

  ...

```

ThomasBurleson:
@matsko, @petebacondarwin - Love the DRY refactor! Question: no support for `functions that return strings`?

matsko:
@petebacondarwin the code has been updated. Please double check and merge in.

petebacondarwin:
LGTM

petebacondarwin:
Landed. Thanks @matsko.
Although I did tweak the test because I felt that it was clearer to have two `it` clauses with no conditional logic than the `they` clause. Hope you don't mind!

petebacondarwin:
Doh! I only changes one of the test. Sorry.

matsko:
Thank you sir :)

PR #12462, opened 2015-07-29
refactor($locale): use en-us as generic built-in locale

### LINKED ISSUE #12134: Inconsistent in en_US localization

lgalfaso:
Previously there was a custom built en-us locale that was included with
angular.js. This made likely that it would get out of sync with the real
en-us locale that is generated from the closure library.

This change removes that custom one and uses the generated one instead.
This also has the benefit of preventing the unwanted caught error on trying
to load `ngLocale` during angular bootstrap.

Closes #12134
Closes #8174


caitp:
> This also has the benefit of preventing the unwanted caught error on trying
to load ngLocale during angular bootstrap.

Nice =) The build code looks good to me, please assign a milestone and handle merging this one though.

PR #12465, opened 2015-07-29
feat(angular.info): optionally store and access additional info on mo…

petebacondarwin:
…dules

This feature allows the developer to store additional meta data about a
module when it is defined by adding an additional parameter containing
an object to the `angular.module(moduleName, deps, info, configFn)` call.

Developers can then access this information by calling `angular.info(moduleName)`

See this https://github.com/angular/material/issues/3842 for more background.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -79,6 +79,9 @@ function setupModuleLoader(window) {
      * @returns {module} new module with the {@link angular.Module} api.
      */
     return function module(name, requires, configFn) {
+
+      var info = {};
+
       var assertNotHasOwnProperty = function(name, context) {
         if (name === 'hasOwnProperty') {
           throw ngMinErr('badname', 'hasOwnProperty is not a valid {0} name', context);
@@ -113,6 +116,37 @@ function setupModuleLoader(window) {
           _invokeQueue: invokeQueue,
           _configBlocks: configBlocks,
           _runBlocks: runBlocks,
+
+          /**
+           * @ngdoc method
+           * @name angular.Module#info
+           * @module ng
+           *
+           * @param {Object=} info Information about the module
+           * @returns {Object|Module} The current info object for this module if called as a getter,
+           *                          or `this` if called as a setter.
+           *
+           * @description
+           * Additional info about this module
+           * For example you could put the version of the module in here.
+           *
+           * ```js
+           * angular.module('myModule', []).info({ version: '1.0.0' });
+           * ```
+           *
+           * The global method `angular.info()` can be used to retrieve this info:
+           *
+           * ```js
+           * var version = angular.info('myModule').version;
+           * ```
+           */
+          info: function(value) {
+            if (isDefined(value)) {
+              info = value;
+              return this;
+            }
+            return info;
+          },
 
           /**
            * @ngdoc property
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -1614,6 +1614,21 @@ describe('angular', function() {
     });
   });
 
+
+  describe('info', function() {
+    it('should return the additional info for the named module', function() {
+      angular.module('a', []).info({some: 'thing'});
+      angular.module('b', ['dep'], function configFn() {}).info({other: 'thang'});
+
+      expect(info('a')).toEqual({some: 'thing'});
+      expect(info('b')).toEqual({other: 'thang'});
+    });
+
+    it('should return anh empty object if there is no such module', function() {
+      expect(info('no-such-module')).toEqual({});
+    });
+  });
+
   describe('bootstrap', function() {
     it('should bootstrap app', function() {
       var element = jqLite('<div>{{1+2}}</div>');
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -81,6 +81,7 @@
     "encodeUriSegment": false,
     "encodeUriQuery": false,
     "angularInit": false,
+    "info": false,
     "bootstrap": false,
     "getTestability": false,
     "snake_case": false,
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1504,6 +1504,25 @@ function angularInit(element, bootstrap) {
   }
 }
 
+
+/**
+ * @ngdoc function
+ * @name angular.info
+ * @module ng
+ * @returns { Object }
+ *    An object containing the info about a module or an empty object if
+ *    the module has not been defined.
+ * @description
+ * Get the info about a given module.
+ */
+function info(moduleName) {
+  try {
+    return angularModule(moduleName).info();
+  } catch (e) {
+    return {};
+  }
+}
+
 /**
  * @ngdoc function
  * @name angular.bootstrap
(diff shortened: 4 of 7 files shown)


lgalfaso:
Even when this approach may solve the issue of getting some meta-data out of a module (if a new bootstrap is made). On the other hand, given that it is possible to have modules redefinition, there would be no mechanism to see the metadata of the instance of a module that is used in any given `$injector`. Is the new functionality going to be used to know before bootstrapping the information of a given module or is the information going to be used after bootstrapping to know the metadata of a module loaded in the current `$injector`?

petebacondarwin:
The primary use case cam from Angular Material, where they wanted to publish the version of the module so that when you are debugging some app, you can quickly drop into the browser console and see what versions are loaded.

petebacondarwin:
@lgalfaso is it common to load a module, create an injector, override the module with a new module and then create another injector in a single application?

test/AngularSpec.js
```diff
@@ -1602,6 +1602,23 @@ describe('angular', function() {
     });
   });
 
+
+  describe('info', function() {
+    it('should return the additional info for the named module', function() {
+      angular.module('a', [], {some: 'thing'});
+      angular.module('b', ['dep'], {other: 'thang'}, function configFn() {});
+
+      expect(info('a')).toEqual({some: 'thing'});
```
ThomasBurleson: Where is the `function info()` defined ?

ThomasBurleson:
@petebacondarwin - The function signature 
```js
return function module(name, requires, configFnOrInfo, configFn) {
};
```
looks like it will present issues with Angular version <1.4.4; we would have to check use `angular.module.length` right ? 

It appears that an alternative signature would not require an argument count precheck.
```js
return function module(name, requires, configFn, info) {
};
```

> It is late, perhaps I am not thinking clearly here.

petebacondarwin:
The `configFn` is optional and often left out so we would need to do some wrangling either way.

test/AngularSpec.js
petebacondarwin: It is defined in the closure and on `angular` so we could make this test more clear.

ThomasBurleson: kk. The definition is in `src/Angular.js`.

lgalfaso:
> The `configFn` is optional and often left out so we would need to do some wrangling either way.

I think that the intention of the comment is that if someone wants to make a module that is 1.4.3 and 1.4.4 compatible, then, to be safe, would need to put the `info` parameter in the forth position anyhow. So why just not make it the forth position in the first place.

lgalfaso:
> @lgalfaso is it common to load a module, create an injector, override the module with a new module and then create another injector in a single application?

For apps where there are multiple non-trivial angular apps running at the same time and browserify (or any of the similar libraries are used), then the chances of having a module redefinition is quite high (even when in most cases the redefinition is just to a slightly different version of the same module (eg ui-router).


I think that this falls into the less than 5% of the cases. That said, I do not think it would be a more code to make it work as `$injector.info(moduleName)`

petebacondarwin:
> > The configFn is optional and often left out so we would need to do some wrangling either way.

> I think that the intention of the comment is that if someone wants to make a module that is 1.4.3 and 1.4.4 compatible, then, to be safe, would need to put the info parameter in the forth position anyhow. So why just not make it the forth position in the first place.

@lgalfaso but both the old (`configFn`) and new (`info`) parameters are optional so it is backward compatible: you can still do:

```js
angular.module(moduleName, depencencyArray, configFn);
```

I know that it is the norm to add new parameters to the end of the argument list but when both are optional then it doesn't make much difference. In this case, I believe that it is more pleasant, if you are going to provide both `info` and a `configFn` for the `configFn` to come at the end:

```js
angular.module('someModule`, ['someDependency'], {version: '1.0.2'}, function() {
  ...
});
```

lgalfaso:
@petebacondarwin I am just saying that, if people want to be safe, then they will need to put the `info` parameter in the forth position. The reason being that if they put it in the third position and this module is used with angular 1.4.3, then it would fail.

Note: If people have complete control of the environment their module will run, then this is not an issue, but this is an issue for material design as they should work fine with angular 1.4.3

petebacondarwin:
Oh right! The concern is that people using this feature will cause their modules to be incompatible with previous versions of angular.

This would still be the case if we add it as a fourth  parameter but the user chooses not to provide a configFn.

So I think we should move this out of the module function and have it as a method.

ThomasBurleson:
@petebacondarwin :+1:

src/loader.js
```diff
@@ -115,6 +118,37 @@ function setupModuleLoader(window) {
...
+           *
+           * The global method `angular.info()` can be used to retrieve this info:
+           *
+           * ```js
+           * var version = angular.info('myModule').version;
+           * ```
+           */
+          info: function(value) {
+            if (isDefined(value)) {
+              info = value;
```
ThomasBurleson: I think we should do `angular.extend(info, angular.isObject(value) ? value : {})`

petebacondarwin: That would mean one could add to the info but never remove anything, right?

IgorMinar:
What if angular.info() returned a map with all available module names as keys and an object with module specific data (or an empty object if no data is provided).

that way it's easy to see what's available without many angular.info(moduleName) calls.

ThomasBurleson:
I think both would be great:

`angular.info(<moduleName>)` :  for specific module info object
`angular.info( )` : for hashmap of all modules info objects keyed by moduleName

IgorMinar:
`angular.info().moduleName` is the same as `angular.info(moduleName)`, so why provide both?

ThomasBurleson:
@IgorMinar - Agreed: superfluous.

src/loader.js
petebacondarwin: @ThomasBurleson - I would rather not use `extend` here. Can you give a concrete use case?

ThomasBurleson: @petebacondarwin 
```js
angular
   .module('ngMaterial,['ngAnimate'])
   .info({ 
      version : "1.0",
      name : "Angular Material"
   });


// Later during runtime
var info = angular.info().ngMaterial;
info.license = "MIT";
```

The goal of the 

```js
info = angular.extend(info, angular.isObject(value) ? value : {});
```

was to require that a hashamp **object** was registered not just a simply value.

petebacondarwin: we could ensure that `value` is an object with a check for `isObject`, rather than using `extend()`.

src/loader.js
```diff
@@ -79,6 +79,9 @@ function setupModuleLoader(window) {
      * @returns {module} new module with the {@link angular.Module} api.
      */
     return function module(name, requires, configFn) {
+
+      var info = {};
```
petebacondarwin: probably better to use a `createMap()` call here

test/AngularSpec.js
```diff
@@ -1614,6 +1614,21 @@ describe('angular', function() {
...
+  describe('info', function() {
+    it('should return the additional info for the named module', function() {
+      angular.module('a', []).info({some: 'thing'});
+      angular.module('b', ['dep'], function configFn() {}).info({other: 'thang'});
+
+      expect(info('a')).toEqual({some: 'thing'});
+      expect(info('b')).toEqual({other: 'thang'});
+    });
+
+    it('should return anh empty object if there is no such module', function() {
```
petebacondarwin: typo

petebacondarwin:
Closing in favour of #15225

PR #12469, opened 2015-07-30
Drag and Drop directives

eislasq:
This change will help to handle drag and drop events.
Example at http://plnkr.co/edit/d4OgfpeyI8dYE0mty5lP?p=preview

### DIFF
--- a/src/ng/directive/ngEventDirs.js
+++ b/src/ng/directive/ngEventDirs.js
@@ -78,6 +78,56 @@ forEach(
 
 /**
  * @ngdoc directive
+ * @name ng-drop
+ *
+ * @description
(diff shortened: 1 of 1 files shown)


eislasq:
I signed it!

caitp:
This might be a bit niche to live in core, have you considered releasing this as a standalone library?

eislasq:
I think this is not a bit nichem, rather it is very common to use drag and drop. But anyway I will publish it as standalone library.
Thanks

caitp:
of course it's very common, but the requirements for different applications D&D behaviours can differ dramatically, as can the browser support and performance --- these things make it very hard to justify living in core. Thank you for publishing this externally =)

PR #12471, opened 2015-07-31
Update step_11.ngdoc

ColinFletch:
changed pronoun to we


ColinFletch:
I signed it!

ColinFletch:
I signed it!

Narretz:
Hm, all the steps begin with "You will ..." and then use You / we kind of arbitrarily. I don't think it makes a lot of sense to change this now.

PR #12473, opened 2015-07-31
fix(ngSrc) don't update the src property for iframe elements in IE:

### LINKED ISSUE #9843: ngSrc duplicate request for iframe in IE
Hi.

I have the issue with iframe and ngSrc directive in IE 10 in Standard mode. 
When I set the ngSrc value it duplicate request for iframe resources.
One of them is with status 'aborted' and another is '200 OK'.

But sometimes on the server side we get 2 requests instead of 1.
It's critical for us because one of params(guid) should be unique, but when two same requests are sent the guids are the same too.

The jsfiddle exampe is here : http://jsfiddle.net/6au10sk4/4/

In angular  source code i find :

Line:15704

          attr.$set(name, value);
          // on IE, if "ng:src" directive declaration is used and "src" attribute doesn't exist
          // then calling element.setAttribute('src', 'foo') doesn't do anything, so we need
          // to set the property as well to achieve the desired effect.
          // we use attr[attrName] value since $set can sanitize the url.
          if (msie && propName) element.prop(propName, attr[name]);
        });

I think here is the problem :

          if (msie && propName) element.prop(propName, attr[name]);

backlit-studio:
no longer set the "src" property on iframe elements in Internet Explorer to avoid sending a duplicate request for iframe content

Closes #9843

### DIFF
--- a/src/ng/directive/attrs.js
+++ b/src/ng/directive/attrs.js
@@ -439,7 +439,7 @@ forEach(['src', 'srcset', 'href'], function(attrName) {
           // then calling element.setAttribute('src', 'foo') doesn't do anything, so we need
           // to set the property as well to achieve the desired effect.
           // we use attr[attrName] value since $set can sanitize the url.
-          if (msie && propName) element.prop(propName, attr[name]);
+          if (msie && propName && element[0].tagName !== 'IFRAME') element.prop(propName, attr[name]);
         });
       }
     };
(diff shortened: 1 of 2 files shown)


ramirezd42:
I signed it!

ramirezd42:
Closing this. I re-opened a new pull request with the correct email address attached to the commits. #12479

PR #12477, opened 2015-07-31
fix(sanitize): Do not duplicate self closing elements

lgalfaso:
This is a refinement on top of #11790 that handles self closing elements


Narretz:
We should probably wait until 1.5 for this, right?

lgalfaso:
> We should probably wait until 1.5 for this, right?
>

Yep

lgalfaso:
Close in favor of #12524

PR #12479, opened 2015-08-01
fix(ngSrc) don't update the src property for iframe elements in IE:

### LINKED ISSUE #9843: ngSrc duplicate request for iframe in IE

backlit-studio:
no longer set the "src" property on iframe elements in Internet Explorer to avoid sending a duplicate request for iframe content

Closes #9843

### DIFF
--- a/src/ng/directive/attrs.js
+++ b/src/ng/directive/attrs.js
@@ -439,7 +439,7 @@ forEach(['src', 'srcset', 'href'], function(attrName) {
           // then calling element.setAttribute('src', 'foo') doesn't do anything, so we need
           // to set the property as well to achieve the desired effect.
           // we use attr[attrName] value since $set can sanitize the url.
-          if (msie && propName) element.prop(propName, attr[name]);
+          if (msie && propName && element[0].tagName !== 'IFRAME') element.prop(propName, attr[name]);
         });
       }
     };
--- a/test/ng/directive/booleanAttrsSpec.js
+++ b/test/ng/directive/booleanAttrsSpec.js
@@ -209,6 +209,16 @@ describe('ngSrc', function() {
 
           dealoc(element);
         }));
+
+    it('should NOT update the element property for iframe elements', inject(
+        function($compile, $rootScope, $sce) {
+          var element = angular.element('<iframe ng-src="http://somewhere"></iframe>');
+          $compile(element);
+          spyOn(element, "prop");
+          $rootScope.$digest();
+          expect(element.prop).not.toHaveBeenCalled();
+          dealoc(element);
+        }));
   }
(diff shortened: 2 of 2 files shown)


lgalfaso:
Would it be possible to know for what specific IE versions this fix is aimed at? Also, please open an issue with Microsoft and post the link to the issue here

fasfsfgs:
As mentioned in the issue, this fixed the bug for at least IE 8, 10 and 11.

fasfsfgs:
Any news on this merge? Is it going to happen?

petebacondarwin:
As an alternative to "hacking" this for IE, could we consider whether we need to actually set the attribute on the element at all? Instead only ever set the property, rather than the attribute. Perhaps that will be enough to trigger the `href` or `src` to become active.

Some initial playing got me to this:

```js
forEach(['src', 'srcset', 'href'], function(attrName) {
  var normalized = directiveNormalize('ng-' + attrName);
  ngAttributeAliasDirectives[normalized] = function() {
    return {
      priority: 99, // it needs to run after the attributes are interpolated
      link: function(scope, element, attr) {
        var propName = attrName,
            name = attrName;

        if (attrName === 'href' &&
            toString.call(element.prop('href')) === '[object SVGAnimatedString]') {
          name = 'xlinkHref';
          attr.$attr[name] = 'xlink:href';
          propName = null;
        }

        attr.$observe(normalized, function(value) {
          console.log(attr, element.href)
          if (!value) value = null;
          attr.$set(name, value, !propName);
          if (propName) {
            element.prop(propName, attr[name]);
          }
          console.log(attr, element.href)
        });
      }
    };
  };
});
```

It fails a dozen unit tests but these could be because they are expecting the attributes rather than the properties to have changed.

PR #12483, opened 2015-08-02
refactor(form): remove the use of the private setter function

lgalfaso:
Remove the private `setter` function from $parse
Replace the use of `setter` from the `form` directive to use $parse

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -449,7 +449,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
  *                       related scope, under this name.
  */
 var formDirectiveFactory = function(isNgForm) {
-  return ['$timeout', function($timeout) {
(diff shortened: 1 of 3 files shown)


src/ng/directive/form.js
```diff
@@ -514,6 +514,13 @@ var formDirectiveFactory = function(isNgForm) {
     };
 
     return formDirective;
+
+    function getSetter(expression) {
+      if (expression === '') {
+        return $parse('this[""]').assign;
```
Narretz: What exactly does this do?

lgalfaso: 410f7c682633c681be641cd2a321f9e51671d474 added some code changes and a test so when the name is the empty string, then the form object is stored in the scope under the `''` name. This code workarounds this case as the empty expression is not assignable

Narretz: Thanks!

PR #12484, opened 2015-08-02
Revlin/1.4.x svg

dancingfrog:
When AngularJS runs within an SVG file there's an issue with handling URL parts... I tried to address this, but please check out the code


dancingfrog:
I signed it!

PR #12498, opened 2015-08-04
Update ngModel.js documentation

akhlesh:
ngModel documentation says that $setViewValue doesn't trigger a '$digest' but its does. it triggers $apply.


akhlesh:
I signed it!

Narretz:
Hey @akhlesh thanks for the PR. The change is not completely correct, though. First, an $apply triggers a $digest, so this is basically correct. 
The second part is a bit more complex. $setViewValue will only trigger a $digest if you haven't specified specific updateOn triggers, like `blur`. If you have specified a `debounce` it'll be triggered only after the debounce. 
Do you think you could incorporate this information into the PR?

akhlesh:
@Narretz so $setviewValue always calls $digest at some point of time.
 sure i'll make this change in my PR.

Issue #12185, opened 2015-06-22
add filename to comment header


tomByrer:
eg https://code.angularjs.org/1.4.1/angular-route.min.js
```js
/* angular-route.min.js
 AngularJS v1.4.1
 (c) 2010-2015 Google, Inc. http://angularjs.org
 License: MIT
*/
```

Handy when scripts are collated (locally or via a CDN like [jsDelivr](https://github.com/jsdelivr/jsdelivr#load-multiple-files-with-single-http-request)) or when the filename is obfuscated, to find all source files.
Thanks for having a header with version either way!

philipooo:
I tried my best on #12499. Maybe you find it useful, otherwise feel free to discard.

tomByrer:
Thanks for your work @philipooo !
I haven't tested, but I trust you know how to extract a filename ;)
Adding the sha code is interesting; I can see it could help identify a mis-tag or new PRs after version tagging.

I'm not sure how Google Closure works, but I believe that Uglify2 is often configured to find `@license` anywhere in the comment field.  So perhaps:
```
* "NG_FILENAME"
* AngularJS v"NG_VERSION_FULL"
* (c) 2010-2015 Google, Inc. http://angularjs.org
* @license MIT
```

philipooo:
Thank you for your feedback!

I reused a regular expression from https://github.com/philipooo/angular.js/blob/add-filename-to-header/lib/grunt/utils.js#L193 and extraced it to the function to get the filename. So this should work :)

The sha code is added automatically when building locally, I have not added something there.

First I tried to place the filename right above the `@license` annotation

```
/** 
 * "NG_FILENAME"
 * @license AngularJS v"NG_VERSION_FULL"
 * (c) 2010-2015 Google, Inc. http://angularjs.org
 * License: MIT
 */
```

what results in: (Things before `@license` seems to get stripped out by the Closure Compiler)

```
/*
 AngularJS v1.4.4-local+sha.b191085
 (c) 2010-2015 Google, Inc. http://angularjs.org
 License: MIT
*/
```

Then I tried using the `@preserve` annotation at the line of the filename. But this kept the `@license` annotation.

```
/*
 angular-route.min.js
 @license AngularJS v1.4.4-local+sha.b191085
 (c) 2010-2015 Google, Inc. http://angularjs.org
 License: MIT
*/
```

tomByrer:
[...]

PR #12499, opened 2015-08-04
chore(grunt): add filename to comment header

philipooo:
Add the filename of each .js and .min.js file to it's own comment header.

### DIFF
--- a/src/angular.prefix
+++ b/src/angular.prefix
@@ -1,5 +1,6 @@
-/**
- * @license AngularJS v"NG_VERSION_FULL"
+/** 
+ * @license "NG_FILENAME"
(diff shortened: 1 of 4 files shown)


src/angular.prefix
```diff
@@ -1,5 +1,6 @@
-/**
- * @license AngularJS v"NG_VERSION_FULL"
+/** 
+ * @license "NG_FILENAME"
```
Narretz: Is there any specific reason why the filename is now the first item in the header?

philipooo: It's just how it was requested in issue #12185

philipooo:
Sorry, accidentally deleted the branch. Restored it.

PR #12500, opened 2015-08-05
docs(guide/Directives): add info about scope types

mightyhorst:
Isolate scope is a fundamental concept that must be explained in order to master directives. Accordingly  the different settings for the 'scope' option should be explained before applying them to the example.


Narretz:
I assume this has been copy & pasted from the $compile docs? If so, wouldn't it make more sense to simply link to that section? Possibly together with a brief introduction to the different scope types. Currently, this huge block of text feels like an anomaly in the guide.

PR #12502, opened 2015-08-05
fix($animate) : will not skip the animation of ng-hide remove.

### LINKED ISSUE #12453: ngAnimate with ngHide fails to add hook 'ng-hide-animate' class in case of binding
I have this case, where I want to apply animation to a DOM object with ngHide directive:
http://plnkr.co/edit/r4i7WbCepl2w9Svv0YHn?p=preview

In this example, I have this JSON objects array:

    $scope.items = [
        {"key": 1, "values": []},
        {"key": 2, "values": [21, 22, 23]}, 
        {"key": 3, "values": [31, 32, 33, 34]}
    ];

This list is bound to a `<ul>` element which will show the keys of this list. When selecting an item of them, values should be displayed in another `<div>` with id detailsList. which will be shown using `ng-show="selectedItem"` statement.

Animation is added using CSS: `#detailsList.ng-hide`, according to Angular documentation of ngAnimate.

When selecting/deselecting the first item with no values, animation works well, but not with the items with values.

When inspecting detailsList element, I found that hook class: `ng-hide-animate` is not added during animation as described by documentation.

Is it a bug or I missed sth??

sreeramu:
According to current ngAnimate code both adding/removing of animate classes and real change classes are done in single requestAnimationFrame , so when ng-hide is removed from the element it will skip the transition.

This is one way of fixing or other way can be just do only adding of classes at same frame with animate for removing the classes wait for one  requestAnimationFrame after applying of animate classes, or we want to go back to the previous behavior that ngAnimate having to wait for one
requestAnimationFrame before CSS classes were added/removed.

closes #12453

### DIFF
--- a/css/angular.css
+++ b/css/angular.css
@@ -2,7 +2,7 @@
 
 [ng\:cloak], [ng-cloak], [data-ng-cloak], [x-ng-cloak],
 .ng-cloak, .x-ng-cloak,
-.ng-hide:not(.ng-hide-animate) {
+.ng-hide:not(.ng-hide-animate):not(.ng-hide-remove) {
   display: none !important;
 }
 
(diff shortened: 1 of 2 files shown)


matsko:
Closing this issue since the root cause was something else that has already been fixed: https://github.com/angular/angular.js/issues/12453#issuecomment-130470653

matsko:
Otherwise thank you @sreeramu for making the PR.

PR #12504, opened 2015-08-05
Make NgModelController $setValidity a noop if the scope is destroyed

rwlogel:
Prevent setValidity from doing anything if the scope it is associated to has been destroyed.  This can happen if an asynchronous event in the directive used the controller after the scope had been destroyed.

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -182,6 +182,7 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
    */
   addSetValidityMethod({
     ctrl: this,
+    $scope: $scope,
     $element: element,
     set: function(object, property, controller) {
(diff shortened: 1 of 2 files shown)


Narretz:
Can you please provide a test / scenario for this case? I think this is very rare, and in the scenario you describe the async event should probably test for the scope.$$destroyed itself.

rwlogel:
I will try and create a test case.  I did fix the problem in my code by preventing the async event from triggering the event.  It just seemed like something the NgModelController should protect against.

Also I was under the impressions that variables names with $$ like $$destroyed should be considered private and there was no guarantee that variable name wouldn't change in a future release.

Narretz:
Sorry, without a test case we cannot merge this. Please reopen when you have a test case.

PR #12515, opened 2015-08-06
fix typo useLegacyPromise

isyun:


realityking:
My bad, but at least I was consistent ;)

PR #12524, opened 2015-08-07
Sanitize innert document

IgorMinar:


IgorMinar:
hmm.. tests are failing because different browsers enumerate over the list of tags in a different order which results in non-deterministic. I'll either have to sort the attribute list or change the matcher to account for this.

lgalfaso:
@IgorMinar even after sorting the attributes, there are two test failures with IE9

```
IE 9.0.0 (Windows 7) HTML htmlParser should throw badparse if text content contains "<" followed by an ASCII letter without matching ">" FAILED
	Expected { tag : 'a', attrs : { body : '', bar< : '' } } to equal undefined.
IE 9.0.0 (Windows 7) HTML htmlParser should parse empty value attribute of node FAILED
	Expected null to equal { tag : 'option', attrs : { selected : '', value : '' } }.
```

I think it should be possible to fix the second issue with something like https://github.com/angular/angular.js/blob/master/src/jqLite.js#L159
The first issue, I am not sure on how to work around it

src/ngSanitize/sanitize.js
```diff
@@ -153,6 +154,42 @@ function $SanitizeProvider() {
...
+  /**
+   * @ngdoc method
+   * @name $sanitizeProvider#enableSvg
+   * @kind function
+   *
+   * @description
+   * Enables a subset of svg to be supported by the sanitizer.
+   *
+   * **Warning**: By enabling this setting without taking other precautions, you might expose your
+   * application to click-hijacking attacks. In these attacks, a sanitize svg could be positioned
```
gkalpak: _sanitize svg_ ??

src/ngSanitize/sanitize.js
```diff
@@ -153,6 +154,42 @@ function $SanitizeProvider() {
...
+   *
+   * ```
+   * .rootOfTheIncludedContent svg {
+   *   overflow: hidden !important;
+   * }
+   * ```
+   *
+   * @param {boolean=} regexp New regexp to whitelist urls with.
+   * @returns {boolean|ng.$sanitizeProvider} Returns the currently configured value if called
+   *    without a argument or self for chaining otherwise.
```
gkalpak: _a_  -->  _an_

hackvertor:
Found a bug with closing br elements. Closing </br> get converted to opening br's
</br> -> <br>

src/ngSanitize/sanitize.js
IgorMinar: fixed. thanks.

src/ngSanitize/sanitize.js
IgorMinar: fixed.

.travis.yml
```diff
@@ -19,10 +19,10 @@ env:
...
     - JOB=e2e TEST_TARGET=jqlite BROWSER_PROVIDER=saucelabs
     - JOB=e2e TEST_TARGET=jquery BROWSER_PROVIDER=saucelabs
-    - JOB=unit BROWSER_PROVIDER=browserstack
-    - JOB=docs-e2e BROWSER_PROVIDER=browserstack
-    - JOB=e2e TEST_TARGET=jqlite BROWSER_PROVIDER=browserstack
-    - JOB=e2e TEST_TARGET=jquery BROWSER_PROVIDER=browserstack
+#    - JOB=unit BROWSER_PROVIDER=browserstack
+#    - JOB=docs-e2e BROWSER_PROVIDER=browserstack
+#    - JOB=e2e TEST_TARGET=jqlite BROWSER_PROVIDER=browserstack
+#    - JOB=e2e TEST_TARGET=jquery BROWSER_PROVIDER=browserstack
```
petebacondarwin: Is this disabling just temporary or should we give up with browserstack altogether?

src/ngSanitize/sanitize.js
```diff
@@ -153,6 +154,42 @@ function $SanitizeProvider() {
...
+
+  /**
+   * @ngdoc method
+   * @name $sanitizeProvider#enableSvg
+   * @kind function
+   *
+   * @description
+   * Enables a subset of svg to be supported by the sanitizer.
+   *
+   * **Warning**: By enabling this setting without taking other precautions, you might expose your
```
petebacondarwin: I would put this warning inside a `<div class="alert alert-warning">` block to make it stand out more.

src/ngSanitize/sanitize.js
```diff
@@ -262,11 +289,30 @@ function makeMap(str, lowercaseKeys) {
...
+(function(window) {
+  var doc;
+  if (window.document && window.document.implementation) {
+    doc = window.document.implementation.createHTMLDocument("inert");
+  } else {
+    throw $sanitizeMinErr('noinert', "Can't create an inert html document");
+  }
+  var docElement = doc.documentElement || doc.getDocumentElement();
+  var bodyElements = docElement.getElementsByTagName('body');
+  if (bodyElements.length === 1) {
```
petebacondarwin: Is it possible that `bodyElements` can have `length > 1`? If so, can we not just take the first one?

src/ngSanitize/sanitize.js
```diff
@@ -276,169 +322,76 @@ function makeMap(str, lowercaseKeys) {
...
+    switch (node.nodeType) {
+      case 1: // ELEMENT_NODE
+        handler.start(node.nodeName.toLowerCase(), attrToMap(node.attributes));
+        break;
+      case 3: // TEXT NODE
+        handler.chars(node.textContent);
+        break;
+    }
+    var nextNode;
+    if (!(nextNode = node.firstChild)) {
```
petebacondarwin: I think this block can be simplified:

```js
    var nextNode;
    if (!(nextNode = node.firstChild)) {
      if (node.nodeType == 1) {
        handler.end(node.nodeName.toLowerCase());
      }
      nextNode = node.nextSibling;
      if (!nextNode) {
        while (nextNode == null) {
          node = node.parentNode;
          if (node === inertBodyElement) break;
          nextNode = node.nextSibling;
          if (node.nodeType == 1) {
            handler.end(node.nodeName.toLowerCase());
          }
        }
      }
    }
    node = nextNode;
```

petebacondarwin:
A few small comments but otherwise LGTM

petebacondarwin:
How far back can this be backported?

IgorMinar:
@petebacondarwin since there are breaking changes included in here, I don't think that we should backport this to older versions. It's safer to keep it in master only.

.travis.yml
IgorMinar: I think it's just temporary

src/ngSanitize/sanitize.js
IgorMinar: usually there should be only one body, but in IE there is none, so we need to create it. I'll add a comment to that code

src/ngSanitize/sanitize.js
IgorMinar: nice catch

h3rald:
@IgorMinar, @petebacondarwin 

Hi, what do you mean with "it's safer to keep it in master only"? Is there any plan to include this security fix in a 1.4.x release?

petebacondarwin:
> safer to keep it in master only

This means that since there are breaking changes, trying to backport it to older versions could cause havoc with live applications. The message is that if you are concerned about these security features then you should upgrade to 1.5 as soon as it comes out.

h3rald:
Hi Pete, thanks a lot for the clarification.

I was wondering if you or someone else could elaborate a bit more on the vulnerability fixed by this pull request: in particular, is it browser-dependent? Does the vulnerability apply just to some/older browsers? Which?

Again thanks a lot for your time!

petebacondarwin:
I'll have a further chat with @IgorMinar about this.

IgorMinar:
aside from the option to disable svg support (which is not a security issue if you set `overflow: hidden` on the element containing the sanitized output), I don't believe that there is a reason for considering the previous implementation insecure. The new implementation is simpler and easier to maintain going forward.

The reason why we didn't backport this to 1.4 was that the new implementation could parse html slightly differently which could result in hard-to-debug issues in the rendered output  (but not security vulnerabilities).

h3rald:
Hi Igor,

Thanks a lot for your comment: it really clarifies things a lot!

My original concern about backporting was due to the fact that this particular pull request was notified to me as a "fix" for possible vulnerabilities affecting Angular sanitization. Based on your comments (and also after doing a bit of testing with both implementations at this point) I think that perhaps the security notification I received was perhaps a bit misleading or simply tagged incorrectly. I will investigate further within my company on why this happened. 

Thanks a lot for your support!

moshere:
> Hi Pete, thanks a lot for the clarification.
> 
> I read through the commits and the only breaking change seems to be the fact that SVG sanitization is not enabled by default (and if enabled you recommend extra care), right?
> 
> The fact that you are now relying on the browser itself to parse HTML code by creating an inert HTML document should not be too much of a problem because AFAIK this functionality is available in all major browsers (including IE9+...
> 
> I was notified of this fix by a security notification issued throughout my company which recommended that we apply the patch manually as a temporary workaround until the fix is officially distributed.
> If the fix is not applied, apparently it may be possible to perform a XSS attack in certain conditions -- is it right?
> 
> Thanks again for your time!
> 
> P.S.: I actually tried to apply the fixes to v1.4.3 and it doesn't seem to break much (I still have to test it with directives using SVG though).

@h3rald - Can I get the relevant fix for 1.4.x version? thank you!

PR #12528, opened 2015-08-09
refactor($compile): remove unnecessary and invalid $destroyed listener

jbedard:
I noticed when bindToController is used the scope `$destroyed` event is listened for, but it should be `$destroy`.  But looking further into it I don't think we need to listen for any destroy event in this case so this PR just deletes it.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1437,11 +1437,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             if (nodeLinkFn.scope) {
               childScope = scope.$new();
               compile.$$addScopeInfo(jqLite(node), childScope);
-              var destroyBindings = nodeLinkFn.$$destroyBindings;
-              if (destroyBindings) {
-                nodeLinkFn.$$destroyBindings = null;
-                childScope.$on('$destroyed', destroyBindings);
-              }
             } else {
               childScope = scope;
             }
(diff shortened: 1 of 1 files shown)


src/ng/compile.js
```diff
@@ -2644,16 +2639,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
       });
-      var destroyBindings = onNewScopeDestroyed ? function destroyBindings() {
+
+      return onNewScopeDestroyed && function destroyBindings() {
         for (var i = 0, ii = onNewScopeDestroyed.length; i < ii; ++i) {
           onNewScopeDestroyed[i]();
         }
-      } : noop;
-      if (newScope && destroyBindings !== noop) {
-        newScope.$on('$destroy', destroyBindings);
```
jbedard: I moved this out to the 1 of 3 `initializeDirectiveBindings` invocations that actually need it. I figured now that `initializeDirectiveBindings` binds to any `destination` and not specifically a scope this logic no longer belonged here...

lgalfaso:
@petebacondarwin WDYT?

src/ng/compile.js
```diff
@@ -1985,24 +1978,28 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           compile.$$addScopeClass($element, true);
           isolateScope.$$isolateBindings =
               newIsolateScopeDirective.$$isolateBindings;
-          initializeDirectiveBindings(scope, attrs, isolateScope,
-                                      isolateScope.$$isolateBindings,
-                                      newIsolateScopeDirective, isolateScope);
+          var parentWatchDestoryer = initializeDirectiveBindings(scope, attrs, isolateScope,
```
Narretz: -> *Destroyer* (also on line 1984 and 1985)

Narretz:
I think the commit message should be more explicit: why exactly do we not need to listen to $destroy in this case?

jbedard:
Fixed the var name typo and rebased. I'll have to look into it again to remember why the listener isn't actually needed...

petebacondarwin:
I tweaked, embellished and merged. Thanks @jbedard - do you just spend your evenings scouring the compiler??

PR #12529, opened 2015-08-09
perf($compile): use createMap() for directive bindings to allow fast forEach

jbedard:
Very minor and I haven't actually measured it. But since `forEach` now has a [fast path for blank objects](https://github.com/angular/angular.js/blob/ed3a33a063f09d7ca356d15c278d95ad82e680a0/src/Angular.js#L273) I thought it would be worth it.


jbedard:
Still seems worth doing to me... :sweat_smile:

gkalpak:
LGTM

PR #12531, opened 2015-08-09
Update http.js

mazenharake:
The isSuccess() function now returns true for values between the range 200 and 300 as, I guess, intended.

I didn't test this or anything, I just stumbled across this as I was reading through the code.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -239,7 +239,7 @@ function transformData(data, headers, status, fns) {
 
 
 function isSuccess(status) {
-  return 200 <= status && status < 300;
+  return 200 >= status && status < 300;
 }
 
 


mazenharake:
Haven't signed anything. I don't have time to read through all that crap just to contribute some code.

You are welcome.

lgalfaso:
@mazenharake without signing the CLA, this cannot be merged. Closing

Narretz:
The change wouldn't have been correct anyway. The current behavior makes sure all codes from 200 to theoretically 299 are considered successes, the change would not catch anything above 200 as a success

mazenharake:
You are absolutely right! :) I was thrown off by the convention, stupid mistake. What you are writing was what I thought I fixed :) Reading code while you have two kid running around playing in the room isn't the best time :) Sorry about wasting your time.

PR #12532, opened 2015-08-09
fix(ngResources): support IPv6 URLs

### LINKED ISSUE #12512: angular-resource breaks IPv6 URLs
There is a regular expression in angular-resource.js that is breaking IPv6 resource URLs:

    url = url.replace(new RegExp("(\/?):" + urlParam + "(\\W|$)", "g"), function(match,
        leadingSlashes, tail) {
      if (tail.charAt(0) == '/') {
        return tail;
      } else {
        return leadingSlashes + tail;
      }
    });

Example:
http://jsfiddle.net/c2franz/pszLs0kg/4/

Expected behavior: a GET request is performed to http://[2620:0:861:ed1a::1]
Observed behavior: a GET request is performed to http://[2620:0:861::1]

lgalfaso:
Do not confuse IPv6 URLs domains and resource parameters.

Closes #12512

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -348,6 +348,7 @@ function shallowClearAndCopy(src, dst) {
  */
 angular.module('ngResource', ['ng']).
   provider('$resource', function() {
+    var PROTOCOL_AND_DOMAIN_REGEX = /^http(?:s)?:\/\/[^\/]*/;
     var provider = this;
 
     this.defaults = {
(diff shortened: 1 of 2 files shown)


src/ngResource/resource.js
```diff
@@ -348,6 +348,7 @@ function shallowClearAndCopy(src, dst) {
  */
 angular.module('ngResource', ['ng']).
   provider('$resource', function() {
+    var PROTOCOL_AND_DOMAIN_REGEX = /^http(?:s)?:\/\/[^\/]*/;
```
gkalpak: Why not `(?:s)?` --> `s?` ?

gkalpak:
What about the `https?` part of the RegExp ? Couldn't people be using other protocols ?
(I can't think of any good reason, but people do all sorts of stuff without a good reason :))

Should we add a breaking change notice (or avoid it altogether) ?

---
Other than that, it LGTM :+1:

PR #12535, opened 2015-08-10
perf($q): small $q performance optimization

lgalfaso:
Only generate a new promise when `then` receives some non-undefined parameter

### DIFF
--- a/lib/promises-aplus/promises-aplus-test-adapter.js
+++ b/lib/promises-aplus/promises-aplus-test-adapter.js
@@ -4,6 +4,8 @@
 var isFunction = function isFunction(value){return typeof value == 'function';};
 var isPromiseLike = function isPromiseLike(obj) {return obj && isFunction(obj.then);};
 var isObject = function isObject(value){return value != null && typeof value === 'object';};
+var isUndefined = function isUndefined(value) {return typeof value === 'undefined';};
+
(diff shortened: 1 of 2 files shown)


lgalfaso:
This saves a few cycles when calling `when` without any parameters as will resolve to a value in one step instead of two

IgorMinar:
lgtm since the a+ specs are ok with this

akoshelev:
It seems to be a breaking change for projects that intensively use angular.mocks.$timeout service and unit testing. After this change had applied, many of our tests began crashing with
> Error: No deferred tasks to be flushed

before Angular 1.4.4 we had to use ```$timeout.flush(); ``` in our unit tests when working with ```$q.when```. For now, there are no deferred tasks and ```$timeout.flush()``` crashes. It is not a big deal and can be fixed easily but imho it is worth mentioning that in the change log.

lgalfaso:
Thanks for the information. Would it be possible for you to create a
plunker so the release note would have a concrete example and the steps to
follow to make things work?

akoshelev:
For sure, please find it here http://plnkr.co/edit/1ajOE3JEiSZvauvjct3D?p=preview
Let me know if you need further information.

lgalfaso:
@akoshelev #12705 should take care of this issue

PR #12541, opened 2015-08-10
fix(ngOptions): allow empty option selection with multiple attribute

### LINKED ISSUE #12511: ngOptions.js throws type error when selecting empty value and using 'multiple' on select
When selecting the empty value `<option value=""></option>` on a `<select multiple="multiple" ng-options="v.label for(k,v) in options">` a type error is thrown.

OS/Browser: All
Version: Angular 1.4.3
Fiddle: https://jsfiddle.net/HB7LU/16138/
Code: https://github.com/angular/angular.js/blob/14638f4a60053b085565e597fc74bd31cf0d372b/src/ng/directive/ngOptions.js#L529

Stack Trace:
```
TypeError: Cannot read property 'disabled' of undefined
    at angular.min.js:272
    at m (angular.min.js:7)
    at u.readValue (angular.min.js:272)
    at angular.min.js:285
    at n.$eval (angular.min.js:134)
    at n.$apply (angular.min.js:135)
    at HTMLSelectElement.<anonymous> (angular.min.js:285)
    at HTMLSelectElement.Gf.c (angular.min.js:35)
```

Fix:

Replace

`if (!option.disabled) selections.push(options.getViewValueFromOption(option));`

with

`if (option && !option.disabled) selections.push(options.getViewValueFromOption(option));`

same as in [Line 517](https://github.com/angular/angular.js/blob/14638f4a60053b085565e597fc74bd31cf0d372b/src/ng/directive/ngOptions.js#L517).

arm1n:
fixes #12511

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -526,7 +526,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
 
           forEach(selectedValues, function(value) {
             var option = options.selectValueMap[value];
-            if (!option.disabled) selections.push(options.getViewValueFromOption(option));
+            if (option && !option.disabled) selections.push(options.getViewValueFromOption(option));
           });
 
           return selections;
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -2058,6 +2058,18 @@ describe('ngOptions', function() {
       // ensure the option has not changed following the digest
       expect(element[0].selectedIndex).toEqual(0);
     });
+
+    // bug fix #12511
+    it('should be selectable if multiple attribute is applied on select element',function() {
+      createMultiSelect(true);
+
+      // select the empty option
+      setSelectValue(element,0);
+
+      // ensure selection and correct binding
+      expect(element[0].selectedIndex).toEqual(0);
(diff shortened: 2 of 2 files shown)


Narretz:
Hi @arm1n thanks for the PR! It looks good. I just wondered: did you upgrade your app to 1.4.x? I noticed that the behavior is different in 1.3.x and before, because the empty option is never displayed for multi selects (because you can deselect until the selection is empty). 
I just wanted to get your opinion - if you consider the current behavior "correct".

arm1n:
Hi @Narretz! I admit that it's quite rare to use empty options for `<select>` with `multiple` attribute. Nevertheless, I think it should be up to the developer to make use of it or not, this seems to be a matter of preference. To be consistent and meet assumed user expectations, it should behave the same way as if `multiple` is absent, but I don't think it should be selected by default, which is already reflected by current behavior, but it simply doesn't reset `ng-model` caused by type error. This would be fixed by this pull request.

As I started building my app from `1.4.1` and recently updated to `1.4.3` I cannot tell you, if this would irritate an updating user. But if nobody ever complained, it's a really rare use case and maybe nobody has any expectations about such behavior, what do you think?

Narretz:
@arm1n I agree, it makes the most sense to have the option for an empty option whether the select is multiple or not. Thanks for the PR!

PR #12543, opened 2015-08-10
docs(angular.module): simplify wording

frankadrian:
A suggestion on the wording of retrieving a moudle

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -38,9 +38,9 @@ function setupModuleLoader(window) {
      * All modules (angular core or 3rd party) that should be available to an application must be
      * registered using this mechanism.
      *
-     * When passed two or more arguments, a new module is created.  If passed only one argument, an
-     * existing module of that name is retrieved.
-     *
+     * Passing one argument retrieves an existing {@link angular.Module}, 
+     * whereas passing more than one argument creates a new {@link angular.Module}
+     * 
      *
      * # Module
      *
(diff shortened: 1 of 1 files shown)


frankadrian:
signed

src/loader.js
```diff
@@ -39,7 +39,7 @@ function setupModuleLoader(window) {
      * registered using this mechanism.
      *
      * When passed two or more arguments, a new module is created.  If passed only one argument, an
-     * existing module (the name passed as the first argument to `module`) is retrieved.
+     * existing module of that name is retrieved.
```
caitp: I like that this is concise --- but how about we make it clear that we're talking about a specific parameter, and maybe just reword the paragraph entirely, because it wasn't particularly well written to begin with. It sounds sort of Yoda-ish, you know? (Referring especially to the preceding "When passed two or more arguments, a new module created is." bit), and just doesn't come across as particularly natural.

Something more like "Creates a new {@link angular.Module} when passed more than one argument, or retrieves an existing {@link angular.Module} when only passed a single argument.", maybe? It's not perfect, but sounds slightly less weird to me.

If we're going to mention "name" at all, lets wrap it in codeblocks (surround with backticks) to hint that it refers to a parameter. And, if we aren't going to mention the position in the list, maybe at least refer to it as "the `name` argument" or something.

frankadrian: Yes i like that! How about:
"Passing one argument retrieves an existing  {@link angular.Module} while passing more than one argument creates a new {@link angular.Module}"

caitp: that sounds good to me! --- Minor nit, comma before "while" --- and you might consider replacing "while" with "whereas", but I'll leave that up to you

caitp:
lgtm

PR #12545, opened 2015-08-10
fix(ngTouch): allow use of href and ng-click with ng-touch

### LINKED ISSUE #5307: Touch event on a link issue
It's in AngularJS 1.2.3.
This is only happening on touch devices or in Google Chrome with touch event emulation enabled.
This link works (redirects to the url):
```<a href="#/search">search</a>```
And this link doesn't (no redirect occurs on tap):
```<a href="#/search" ng-click="v = 1">search</a>```
Click handler invoked and after that something strange happens inside of 
```element.on('touchend', function(event)``` (angular-touch.js:385). 
Somehow it decides to prevent event by ```calling preventGhostClick(x, y);```
It's really sad because it blocked our big mobile app to use AngularJS 1.2.3 in production. We're still on 1.0.8.

### PR DESCRIPTION
**Pull request by ca136**
changes browser url on elements with ng-click if element has an href attribute

fixes #5307

### DIFF
--- a/src/ngTouch/directive/ngClick.js
+++ b/src/ngTouch/directive/ngClick.js
@@ -48,8 +48,8 @@ ngTouch.config(['$provide', function($provide) {
   }]);
 }]);
 
-ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
-    function($parse, $timeout, $rootElement) {
+ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement', '$location',
+    function($parse, $timeout, $rootElement, $location) {
   var TAP_DURATION = 750; // Shorter than 750ms is a tap, longer is a taphold or drag.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jonricaurte** [comment]
Is it possible to add this to 1.4.4? We just ran into this issue and the current workaround is not ideal. We have a mobile navigation slide in on the left side of the screen, and when clicking any of the a tags with hrefs and ng-clicks within that mobile nav, we want the user to be directed to that link and also close the mobile nav through the ng-click.

Thanks.

**Comment by martinmcwhorter** [comment]
I can't wait to try it out. The ngTouch bugs have been persistent.

**Comment by zhongsp** [comment]
@ca136 thanks for your job. I've met the same issue. But I want to share some info with your solution.
I'm using ngTouch and ui-router. Given the current url  */app/#/display/water*. And the html
```html
<a ng-href="#/home" ng-click="doSomething()">Home</a>
```
By your fixing, the final url would be */app/#/display/water/#%2Fhome*

The proper url should be */app/#/home*

PR #12546, opened 2015-08-11
docs(select): explain how to set default value

meliaxford:
Setting the default value in a select is a real trap for beginners, questions about how to do this on StackOverflow have been view more than 100,000 times in the last year.  This change updates the documentation to make it clearer.


meliaxford:
I signed it!

PR #12552, opened 2015-08-11
minor fix of the $interval doc

chrootsu:
The promise can be undefined 
https://github.com/angular/angular.js/blob/v1.4.3/src/ng/interval.js#L180-L188
https://github.com/angular/angular.js/blob/v1.4.3/test/ng/intervalSpec.js#L309-L312

### DIFF
--- a/src/ng/interval.js
+++ b/src/ng/interval.js
@@ -174,7 +174,7 @@ function $IntervalProvider() {
       * @description
       * Cancels a task associated with the `promise`.
       *
-      * @param {promise} promise returned by the `$interval` function.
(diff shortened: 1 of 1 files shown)


chrootsu:
I signed it!

caitp:
I suppose this matches the docs for `$timeout.cancel()`, so lgtm --- the PR text is not super clear, though. The doc isn't saying anything about the type being undefined, it's saying the parameter is optional, and changing the type name. But it's fine.

chrootsu:
@caitp Thanks for the clarification )

PR #12554, opened 2015-08-12
fix(ngAnimate): always apply a preparation reflow for CSS-based animations

### LINKED ISSUE #12267: Animations not working with 1.4.x releases, ok with 1.3
I'm building a single page application with angular material and for some reason the animations are not functioning.

It looks like it is not just angular material as I can't get the basic example on the https://docs.angularjs.org/guide/animations page to work.

It appears that the classes that angular material relies on to be added for the css animations to hook on to are not being applied. Inspecting the elements in chrome I just see the hide and show classes but none of the transition classes.

Relevant parts of my bower file:

```
    "angular": "~1.4.1",
    "angular-animate": "~1.4.1",
    "angular-sanitize": "~1.4.1",
    "angular-messages": "~1.4.1",
    "angular-i18n": "~1.4.1",
    "angular-ui-router": "~0.2.15",
    "angular-material": "~0.10.0",
```

My site (in development) is at http://www.zakhenry.com and the animation test page is at http://www.zakhenry.com/#/blog
When the viewport is small (ie mobile width) I _should_ be able to see the sidepane animate out like on the https://material.angularjs.org/latest/ site; instead it just appears and disappears.

Source code is at https://github.com/xiphiaz/zakhenry.com

*update*
 I can actually access the nganimate api eg if I run `$animate.leave();` in a controller I get an error thrown from angular-animate, indicating the module is loaded successfully. Also adding breakpoints in the ng animate vendor works.

If I call `console.log($animate.enabled())` in the `run()` phase I get `false` in the console. I tried setting it to true with `$animate.enabled(true)`, to no effect.

### LINKED ISSUE #12553: ngShow is broken in master
No version has been released just yet, but ngShow and ngHide do not get caught with the new class based reflow fix.

Broken:
http://plnkr.co/edit/qWpK7YWCwMjSKb5eTtXk?p=preview

Fixed:
http://plnkr.co/edit/zqhheQydJH7uta6QYgHG?p=preview

matsko:
It's unpredictable sometimes to ensure that a browser triggers a reflow
for an animation. Prior to this patch, reflows would be applied
carefully in between parent/child DOM structure, but that doesn't seem
to be enough for animations that contain complex CSS styling rules.

Closes #12553
Closes #12554
Closes #12267

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -67,9 +67,9 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
   });
 
   this.$get = ['$$rAF', '$rootScope', '$rootElement', '$document', '$$body', '$$HashMap',
-               '$$animation', '$$AnimateRunner', '$templateRequest', '$$jqLite',
+               '$$animation', '$$AnimateRunner', '$templateRequest', '$$jqLite', '$$forceReflow',
        function($$rAF,   $rootScope,   $rootElement,   $document,   $$body,   $$HashMap,
-                $$animation,   $$AnimateRunner,   $templateRequest,   $$jqLite) {
+                $$animation,   $$AnimateRunner,   $templateRequest,   $$jqLite,   $$forceReflow) {
 
     var activeAnimationsLookup = new $$HashMap();
     var disabledElementsLookup = new $$HashMap();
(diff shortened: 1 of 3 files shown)


lgalfaso:
I am slightly worried that this can cause some performance issues, but the problems are just too big not to merge this

lgalfaso:
landed as d33cedda1624114d7e97a97b79705685c6cc40a2

PR #12556, opened 2015-08-12
Added invokeApply parameter to $http to skip apply

Utsav2:
There are a bunch of times when we don't want a $rootScope.$apply after an http request succeeds. [This](http://stackoverflow.com/questions/21935215/make-angularjs-skip-running-a-digest-loop-if-http-get-resulted-in-no-new-data) stackoverflow question gives a good use case: polling.
We can also use this to do scope specific digests, which really helps performance and snappiness on mobile devices.

It also gives us a little more control over the API, and also makes it similar to $timeout.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -920,7 +920,7 @@ function $HttpProvider() {
 </file>
 </example>
      */
-    function $http(requestConfig) {
+    function $http(requestConfig, invokeApply) {
 
       if (!angular.isObject(requestConfig)) {
         throw minErr('$http')('badreq', 'Http request configuration must be an object.  Received: {0}', requestConfig);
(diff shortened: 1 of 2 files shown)


Utsav2:
I signed it!

src/ng/http.js
```diff
@@ -1276,11 +1284,15 @@ function $HttpProvider() {
...
         }
 
-        if (useApplyAsync) {
-          $rootScope.$applyAsync(resolveHttpPromise);
-        } else {
-          resolveHttpPromise();
-          if (!$rootScope.$$phase) $rootScope.$apply();
+        var skipApply = (isDefined(invokeApply) && !invokeApply);
+
+        if (!skipApply) {
```
lgalfaso: so if `skipApply` is true, then `resolveHttpPromise` will not get called?

Utsav2: Yeah, that's a mistake.

Utsav2:
Could you have a look at this #12557? I've fixed the problem.

PR #12558, opened 2015-08-12
Add custom attributes to linky

stianjensen:
Optional extra attributes may be defined either as:
- a map of attributes and values
- a function that takes the url as a parameter and returns a map

### DIFF
--- a/src/ngSanitize/filter/linky.js
+++ b/src/ngSanitize/filter/linky.js
@@ -15,6 +15,16 @@
  *
  * @param {string} text Input text.
  * @param {string} target Window (_blank|_self|_parent|_top) or named frame to open links in.
+ * @param {object|function(url)} [attributes] Add custom attributes to the link element.
+ *
+ *    Can be one of:
+ *
+ *    - `object`: A map of attributes
+ *    - `function`: Takes the url as a parameter and returns a map of attributes
+ *
+ *    If the map of attributes contains a value for `target`, it overrides the value of
+ *    the target parameter.
+ *
  * @returns {string} Html-linkified text.
  *
  * @usage
@@ -32,7 +42,7 @@
                'mailto:us@somewhere.org,\n'+
                'another@somewhere.org,\n'+
                'and one more: ftp://127.0.0.1/.';
-             $scope.snippetWithTarget = 'http://angularjs.org/';
+             $scope.snippetWithSingleURL = 'http://angularjs.org/';
            }]);
        </script>
        <div ng-controller="ExampleController">
@@ -55,10 +65,19 @@
          <tr id="linky-target">
           <td>linky target</td>
           <td>
-            <pre>&lt;div ng-bind-html="snippetWithTarget | linky:'_blank'"&gt;<br>&lt;/div&gt;</pre>
+            <pre>&lt;div ng-bind-html="snippetWithSingleURL | linky:'_blank'"&gt;<br>&lt;/div&gt;</pre>
           </td>
           <td>
-            <div ng-bind-html="snippetWithTarget | linky:'_blank'"></div>
+            <div ng-bind-html="snippetWithSingleURL | linky:'_blank'"></div>
+          </td>
+         </tr>
+         <tr id="linky-custom-attributes">
+          <td>linky custom attributes</td>
+          <td>
+            <pre>&lt;div ng-bind-html="snippetWithSingleURL | linky:'_self':{rel:'nofollow'}"&gt;<br>&lt;/div&gt;</pre>
+          </td>
+          <td>
+            <div ng-bind-html="snippetWithSingleURL | linky:'_self':{rel:'nofollow'}"></div>
           </td>
          </tr>
          <tr id="escaped-html">
@@ -95,9 +114,16 @@
 
        it('should work with the target property', function() {
         expect(element(by.id('linky-target')).
-            element(by.binding("snippetWithTarget | linky:'_blank'")).getText()).
+            element(by.binding("snippetWithSingleURL | linky:'_blank'")).getText()).
             toBe('http://angularjs.org/');
         expect(element(by.css('#linky-target a')).getAttribute('target')).toEqual('_blank');
+       });
+
+       it('should optionally add custom attributes', function() {
+        expect(element(by.id('linky-custom-attributes')).
+            element(by.binding("snippetWithSingleURL | linky:'_self':{rel: 'nofollow'}")).getText()).
+            toBe('http://angularjs.org/');
+        expect(element(by.css('#linky-custom-attributes a')).getAttribute('rel')).toEqual('nofollow');
        });
      </file>
    </example>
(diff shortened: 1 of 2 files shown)


stianjensen:
I signed it!

lgalfaso:
A test would be nice

Narretz:
Would it make sense to make the third parameter more powerful - i.e. allowing it to add arbitrary attributes by passing a map or a function that returns a map?

stianjensen:
I guess that would be even better. I'm sure many would like the option of adding classes to links for instance.

I'll take a stab at changing the parameter.
I guess you still want to keep the target parameter as a dedicated one, just for the sake of backwards compatibility?

Narretz:
@stianjensen Great! Yes, the target parameter should stay. I imagine the API looking something like this (I haven't checked with the current code though):
- the 3rd parameter is either a map of attributes, or a function that receives the url as a parameter and returns a map.
- if the 2nd parameter is set, and the map contains the target attribute, it overrides it

stianjensen:
> if the 2nd parameter is set, and the map contains the target attribute, it overrides it

The 2nd parameter overrides the target attribute from the map, or the other way around?

Also, am I correct that there is no way you could provide a value for the 3rd parameter without also providing a value for the 2nd parameter?

stianjensen:
I updated the pull request with support for custom attributes now :smile:

src/ngSanitize/filter/linky.js
```diff
@@ -137,8 +163,22 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
     }
 
     function addLink(url, text) {
+      var key;
+      var computedAttributes = {};
       html.push('<a ');
-      if (angular.isDefined(target)) {
+      if (typeof attributes === 'object') {
```
Narretz: We can use `angular.isObject`

src/ngSanitize/filter/linky.js
```diff
@@ -137,8 +163,22 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
...
+        computedAttributes = attributes;
+        for (key in computedAttributes) {
+          html.push(key + '="' + computedAttributes[key] + '" ');
+        }
+      }
+      if (typeof attributes === 'function') {
```
Narretz: We can use `angular.isFunction`

test/ngSanitize/filter/linkySpec.js
```diff
@@ -54,4 +54,19 @@ describe('linky', function() {
     expect(linky("http://example.com", "someNamedIFrame")).
       toEqual('<a target="someNamedIFrame" href="http://example.com">http://example.com</a>');
   });
+
+  it('should optionally add custom attributes', function() {
+    expect(linky("http://example.com", "_self", {rel: "nofollow"})).
+      toEqual('<a rel="nofollow" target="_self" href="http://example.com">http://example.com</a>');
+  });
+
+  it('should optionally add custom attributes from function', function() {
```
Narretz: There should also be an exra test that ensures the url is actually passed to the attr function

src/ngSanitize/filter/linky.js
```diff
@@ -137,8 +163,22 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
...
+      var key;
+      var computedAttributes = {};
       html.push('<a ');
-      if (angular.isDefined(target)) {
+      if (typeof attributes === 'object') {
+        computedAttributes = attributes;
+        for (key in computedAttributes) {
+          html.push(key + '="' + computedAttributes[key] + '" ');
+        }
+      }
```
Narretz: Make this an `else if`?

src/ngSanitize/filter/linky.js
stianjensen: Nice!

src/ngSanitize/filter/linky.js
stianjensen: Sure

stianjensen:
Updated

src/ngSanitize/filter/linky.js
```diff
@@ -137,8 +163,21 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
     }
 
     function addLink(url, text) {
+      var key;
+      var computedAttributes = {};
       html.push('<a ');
-      if (angular.isDefined(target)) {
+      if (angular.isObject(attributes)) {
```
campersau: Can this code duplication be prevented using something like this?

```js
var key;
var computedAttributes = {};
if (angular.isFunction(attributes)) {
  attributes = attributes(url);
}
if (angular.isObject(attributes)) {
  computedAttributes = attributes;
  for (key in computedAttributes) {
    html.push(key + '="' + attributes[key] + '" ');
  }
}
```

This would also check if the return value from the `attributes` function is an object.

stianjensen: I guess that's a good idea. At first I didn't want to reassign the `attributes` variable, but the extra check for the return value of attributes as a function, is a compelling case for doing it like this.

src/ngSanitize/filter/linky.js
```diff
@@ -137,8 +163,19 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
     }
 
     function addLink(url, text) {
+      var key;
+      var computedAttributes = {};
```
petebacondarwin: Do we need to use a new variable `computedAttributes` here?
Can't we just assign `attributes = {}` in an else block starting at [line 177](https://github.com/angular/angular.js/pull/12558/files#diff-c0c7c3cd309480c6baa2b90491d610f3R176)?

src/ngSanitize/filter/linky.js
```diff
@@ -137,8 +163,19 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
...
       html.push('<a ');
-      if (angular.isDefined(target)) {
+      if (angular.isFunction(attributes)) {
+        attributes = attributes(url);
+      }
+      if (angular.isObject(attributes)) {
+        computedAttributes = attributes;
+        for (key in computedAttributes) {
+          html.push(key + '="' + computedAttributes[key] + '" ');
```
petebacondarwin: Do we need to HTML encode the key and value here or does the $sanitize service deal with that?

petebacondarwin:
Other than those two questions, it LGTM

src/ngSanitize/filter/linky.js
stianjensen: Good idea, I'll do that!

src/ngSanitize/filter/linky.js
stianjensen: My understanding was that this is dealt with by $sanitize, but anyone correct me if I'm wrong.

stianjensen:
ping

Narretz:
The generated HTML is sanitized before it is inserted. I'll add a test for that.

Narretz:
Looks like IE and FF insert the attributes in a different order! (see Travis failures). Looks like we'll have to change these expects to be more robust.

Narretz:
Landed as 06f002b161f61079933d482668440d8649fd84fc
thx @stianjensen for working on this!

PR #12566, opened 2015-08-12
chore: rename angular.bind to angular.bind.js

matsko:
Some internal tests were failing since `.bind` is a JS file which
is expected to have a `.js` file name suffix.


IgorMinar:
lgtm

PR #12570, opened 2015-08-12
feat(ngAnimate): expose a core version of `$animateCss`

### LINKED ISSUE #12509: Make $animateCss available in Angular
For library maintainers that allow users to opt-in to animations, there is a gap in support - currently `$animateCss` is only available in ngAnimate. It would be useful to have a matching API in Angular itself with a `$animateCss` service, similar to how `$animate` is handled. The only way to detect if the user has $animateCss is to do a try-catch with fetching via `$injector` and swallow the exception, and fall back to $animate usage in that case.

matsko:
A core version of `$animateCss` can now be injected when
ngAnimate is not present. This core version doesn't trigger any
animations in any way. All that it does is apply the provided from
and/or to styles as well as the addClass and removeClass values.

The motivation for this feature is to allow for directives to activate
animations automatically when ngAnimate is included without the need to
use `$animate`.

Closes #12509

### DIFF
--- a/angularFiles.js
+++ b/angularFiles.js
@@ -14,6 +14,7 @@ var angularFiles = {
 
     'src/ng/anchorScroll.js',
     'src/ng/animate.js',
+    'src/ng/animateCss.js',
     'src/ng/browser.js',
     'src/ng/cacheFactory.js',
     'src/ng/compile.js',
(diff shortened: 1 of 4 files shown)


matsko:
LGTM (from Igor in person).

PR #12576, opened 2015-08-13
ng-animate-ref leave clone element fix

lee-tunnicliffe:
Call $animate.leave on the cloned element rather than just remove to provide some more animation hooks.


lee-tunnicliffe:
Closing this as I had the wrong credentials configured in git.

Issue #12539, opened 2015-08-10
Anchoring Animations (ng-animate-ref) no way to transition between cloned element and new element


lee-tunnicliffe:
I am trying to achieve a morphing animation between views using ng-animate-ref. The animation should follow the steps below:

1) When the user clicks the element to go forwards or the back link to go backwards, Fade out the contents of the element by transitioning the opacity of the contents.

2) Morph the element into it's new size and position using ng-animate-ref.

3) Either fade out the cloned element to reveal the actual element underneath or fade in the actual element when the cloned element is removed from the dom.

It is the 3rd step that I cannot achieve. Here is a plunkr with what I have so far.

http://plnkr.co/edit/H41tfHjpmK6o5jm2XhoE?p=preview

The problem I have is that ng-anchor-out-remove happens too early (during the morph) and then there is no ng-animate-shim-remove on the real element that I could use. I can't think what else to do.

Another smaller issue that I have is that on ie11 I get a flash of the content showing just before the morphing starts after I have faded out the content.

gjr-qm:
You could introduce a delay in your enter animation so it starts when the leave animation ends.

Link to working plnkr fork: http://plnkr.co/edit/yxy655o8caCf4svZOLAQ?p=preview

What did i change?

Line 51 of your style.css changed from
    transition: 2s linear opacity;
into
    transition: 2s 2s linear opacity;

The second 2s is the delay

lee-tunnicliffe:
Thanks for your suggestion but the issue with this is I don't see the element morph anymore.

gjr-qm:
Somehow the anchor was removed... Try again or just change your own plnkr...

Still not the most robust solution, but if the animation is fast enough it will work i think.

matsko:
@lee-tunnicliffe would this feature provide the functionality that you're looking for? https://github.com/angular/angular.js/pull/12577

Once the anchoring animation completes you could optionally fade out the leave element to reveal the one below.

lee-tunnicliffe:
Yes, that was my PR I referenced this issue in it. Didn't I link it correctly?

matsko:
My mistake. I just scanned over this issue super quick. Let me think this over to see what we can do. But it sounds like a good idea.

wawyed:
Is there any ETA on this feature?

PR #12577, opened 2015-08-13
ng-animate-ref leave clone element fix - Call $animate.leave on the cloned element rather than just remove to provide some more animation hooks.

lee-tunnicliffe:
ng-animate-ref leave clone element fix - Call $animate.leave on the cloned element rather than just remove to provide some more animation hooks.
https://github.com/angular/angular.js/issues/12539

### DIFF
--- a/src/ngAnimate/animateCssDriver.js
+++ b/src/ngAnimate/animateCssDriver.js
@@ -9,8 +9,8 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
   var NG_OUT_ANCHOR_CLASS_NAME = 'ng-anchor-out';
   var NG_IN_ANCHOR_CLASS_NAME = 'ng-anchor-in';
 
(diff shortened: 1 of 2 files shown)


matsko:
This is a good idea, but I think it's better to use:

```js
prepareRegularAnimation({
   element: clone,
   event: 'leave',
   structural: true
}, noop);
```

On the element so it doesn't have to cycle through the list of operations inside of `$animate.leave` since the animation has already been detected as CSS.

lee-tunnicliffe:
Thanks for your reply. Could I submit another pull request with this change or does self-assigned mean you're going to take it from here?

matsko:
I can take care of it from here if you want.

lee-tunnicliffe:
Thanks very much :)

NickBolles:
Any update on this? @matsko

petebacondarwin:
This doesn't appear to have been done. Moving to Backlog for now.

wawyed:
Is this something that could be looked at?

petebacondarwin:
@matsko is overloaded with the new Angular animation stuff, so we don't have any resource to look at this feature request right now. Well implemented PRs are welcome.

PR #12580, opened 2015-08-13
test(select): ensure that options with interpolated text are added / …

Narretz:
…updated

### DIFF
--- a/test/ng/directive/selectSpec.js
+++ b/test/ng/directive/selectSpec.js
@@ -964,21 +964,83 @@ describe('select', function() {
 
   describe('option', function() {
 
-    it('should populate value attribute on OPTION', function() {
+    it('should populate a missing value attribute with the option text', function() {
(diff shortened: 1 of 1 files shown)


test/ng/directive/selectSpec.js
```diff
@@ -964,22 +964,84 @@ describe('select', function() {
...
+        browserTrigger(element.find('option').eq(0));
+        expect(scope.selected).toBe('Option 1');
+
+        scope.selected = 'Option 2';
+        scope.$digest();
+        expect(element.find('option').eq(1).prop('selected')).toBe(true);
+        expect(element.find('option').eq(1).text()).toBe('Option 2');
+
+        //Make sure the bindings are destroyed
+        dealoc(element);
```
Narretz: I am not completely sure if this is the best way to make sure the text bindings are destroyed. (without this line, it complains about leaking bindings)

PR #12582, opened 2015-08-13
fix(select): update option if interpolated value attribute changes

### LINKED ISSUE #12005: ng-repeat with <option> binding issues
I have an odd problem where I cannot get an `ng-model` value bound to a `<select ng-model>` to work against a list bound with `ng-repeat`. I can't duplicate the scenario in a [simple plunkr](http://plnkr.co/edit/YD8OXNHKk1M9C11X6Kae?p=preview). 

This plunkr works, but the same code in an app with more complex layout fails with the first list displaying the wrong item in the first list. In my actual application, I see the model values displayed correctly on the bottom but the actual select list always selects DOG. IOW, the model is correct, but the binding in the select list is not.

The only way I've been able to work around this is to use $timeout() to delay assigning the value which works but is ugly:

```javascript
        // special handling for 'remembering' selection
        var type = animalService.animalFilter.type;
        vm.filter.type = '';
        $timeout(function () {            
            vm.filter.type = type; 
        },30);
```

It appears that ng-model on the list is binding before the list is bound and depending on how quickly the page renders this causes a binding issue. The second list which uses static values bind correctly.

I have run into this a few times now, but it throws me off every single time until I remember that I need to delay binding.  This really shouldn't be necessary - is there a better way to accomplish this or possibly a binding bug with a timing issue?

Narretz:
Previously, the option would only update if the fallback
(the interpolated option text) was used. Now the value attribute will
be observed if it contains an interpolation.

Closes #12005

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -270,9 +270,12 @@ var optionDirective = ['$interpolate', function($interpolate) {
     priority: 100,
     compile: function(element, attr) {
 
-      // If the value attribute is not defined then we fall back to the
-      // text content of the option element, which may be interpolated
-      if (isUndefined(attr.value)) {
+      if (isDefined(attr.value)) {
+        // If the value attribute is defined, check if it contains an interpolation
+        var valueInterpolated = $interpolate(attr.value, true);
+      } else {
+        // If the value attribute is not defined then we fall back to the
+        // text content of the option element, which may be interpolated
         var interpolateFn = $interpolate(element.text(), true);
         if (!interpolateFn) {
           attr.$set('value', element.text());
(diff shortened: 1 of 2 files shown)


src/ng/directive/select.js
```diff
@@ -288,24 +291,38 @@ var optionDirective = ['$interpolate', function($interpolate) {
...
-              selectCtrl.addOption(newVal, element);
-              selectCtrl.ngModelCtrl.$render();
-              chromeHack(element);
+              addOption(newVal);
             });
           } else {
-            selectCtrl.addOption(attr.value, element);
-            selectCtrl.ngModelCtrl.$render();
-            chromeHack(element);
+            // Either the value attribute or the text content is static
```
gkalpak: I think the comment would be more accurate as _"Both the value attribute and the text content are static"_, right ?

src/ng/directive/select.js
```diff
@@ -288,24 +291,38 @@ var optionDirective = ['$interpolate', function($interpolate) {
...
         // Only update trigger option updates if this is an option within a `select`
         // that also has `ngModel` attached
         if (selectCtrl && selectCtrl.ngModelCtrl) {
 
-          if (interpolateFn) {
+          if (valueInterpolated) {
+            // The value attribute is interpolated
+            var oldVal;
+            attr.$observe('value', function valueAttributeObserveAction(newVal) {
+              if (oldVal && oldVal !== newVal) {
```
gkalpak: Is it really possible to have a truthy `oldVal` that is equal to the `newVal` ?
If so, should we also refrain from adding a new option ?

Also, what happens if `oldVal` is the empty string ? Wouldn't we still want to remove the option ?

src/ng/directive/select.js
Narretz: Hmm, it's a bit obscure: If the attr is defined, we create an interpolation fn for it (to check if we should observe). We then skip everything related to the option text, because the value takes precedence over it. Which means, while the option value attr might not be interpolated, the option text could still be. Does that make sense?

if the value attribute is undefined, we create an interpolationFn for the text. If there's nothing to interpolate, we set the text as the value attr. That's why at this point it could be

src/ng/directive/select.js
Narretz: Yeah, since empty strings are allowed, this should probably work. So simply change to `oldVal || oldVal === ''` or `isDefined(oldVal)`?

gkalpak: `isDefined(oldVal)` sgtm (I think :smiley:)

src/ng/directive/select.js
gkalpak: You are right; then both wordings are inaccurate. It should be more like:
`Either the value attribute exists and is static or it doesn't exists and the text content is static`
(which although more accurate sounds messy anyway - so I leave it up to you :)

Narretz: I think just noting that the attr value is static should be enough then. The code isn't that big, it should be relatively easy to find out what exactly happens.

PR #12585, opened 2015-08-14
fix($animateCss): the transitions options delay value should be applied before class application

### LINKED ISSUE #12584: $animateCss options.delay doesn't work with transitions

matsko:
Applying a CSS transition delay after a reflow has hit occured where a
CSS class is added/removed doesn't work. This feature does however work
with keyframe animations. This patch ensures that the provided
`options.delay` value is applied beforehand.

Closes #12584

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -598,6 +598,18 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
         return closeAndReturnNoopAnimator();
       }
 
+      if (options.delay != null) {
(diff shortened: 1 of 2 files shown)


Narretz:
LGTM. But can you rewrite this sentence in the commit message:
> Applying a CSS transition delay after a reflow has hit occured where a
CSS class is added/removed doesn't work.

It doesn't really make sense to me

matsko:
Closed as https://github.com/angular/angular.js/commit/0c81e9fd25285dd757db98d458919776a1fb62fc.

The commit message has been updated.

Thank you @Narretz.

PR #12590, opened 2015-08-14
Fix test issue

elviocb:
Unnecessary split. The url returns a string without the hash, resulting in an undefined value and making the test fails.


elviocb:
I signed it!

caitp:
in the actual phonecat codebase we're using `toEqual()` for some reason (???) but this (otherwise) matches up with it, so looks good to me

PR #12593, opened 2015-08-15
Please let us use shims for old IE

S-YOU:


S-YOU:
I signed it!

Narretz:
which ie9 versions does this affect? IE8 is unsupported since 1.3

S-YOU:
You don't have to support IE8. but, that angular works well with shims.
But with above code, IE throw error for the whole file.

lgalfaso:
I have to disagree with @S-YOU, IE8 is not supported starting with 1.3. This is, IE8 is strictly not supported and this is with or without shims.
If you would like to have this change, then there are other projects that do code transformation to make some browsers happy (eg https://github.com/StoryCloud/webkit-assign) you are free to clone their projects and make Angular 1.4.x work for IE8, but I see no reason for merging this.

S-YOU:
Sigh! the whole point of using shims is to be able to use official angularjs, not someone else build.
And StoryCloud/webkit-assign is irrelevant, its about webkit, not related to IE.

realityking:
@S-YOU 

What you could do use is use babel's [propertyLiterals](http://babeljs.io/docs/advanced/transformers/es3/property-literals/) transform in your build process to do this automatically.

S-YOU:
Thanks for info @realityking. Good to know that. I indeed built my own.

Narretz:
Turns out we need to do this anyway to support Android 2.3: https://github.com/angular/angular.js/commit/1cc9c9ca9d9698356ea541517b3d06ce6556c01d

S-YOU:
IE haters, hahaha. Now, I have my own fork of what I wanted. No more negative noises!

gkalpak:
@S-YOU, fwiw, this particular one is already fixed as @Narretz mentioned in https://github.com/angular/angular.js/pull/12593#issuecomment-133112958.

PR #12598, opened 2015-08-17
feat($injector): support instantiating classes.

### LINKED ISSUE #12597: $injector cannot instantiate ES6 classes
This code (paraphrasing) fails with `TypeError: Class constructors cannot be invoked without 'new'`:

```js
class Foo { constructor() { } }
var foo = $injector.instantiate(Foo);
```

The reason is that ES6 disallows using `fn.apply` on class constructors.

mprobst:
ES6's `class Foo {}` constructors cannot be instantiated using
`fn.apply`. This change extracts injection argument collection and then
uses `new (Function.bind.apply(ctor, args))` to instantiate the service
instance.

Fixes #12597.

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -801,31 +801,33 @@ function createInjector(modulesToLoad, strictDi) {
       }
     }
 
+
+    function injectionArgs(fn, locals, serviceName) {
+      var args = [],
+          $inject = createInjector.$$annotate(fn, strictDi, serviceName);
+
+      for (var i = 0, length = $inject.length; i < length; i++) {
+        var key = $inject[i];
+        if (typeof key !== 'string') {
+          throw $injectorMinErr('itkn',
+                  'Incorrect injection token! Expected service name as string, got {0}', key);
+        }
+        args.push(locals && locals.hasOwnProperty(key) ? locals[key] :
+                                                         getService(key, serviceName));
+      }
+      return args;
+    }
+
+
     function invoke(fn, self, locals, serviceName) {
       if (typeof locals === 'string') {
         serviceName = locals;
         locals = null;
       }
 
-      var args = [],
-          $inject = createInjector.$$annotate(fn, strictDi, serviceName),
-          length, i,
-          key;
-
-      for (i = 0, length = $inject.length; i < length; i++) {
-        key = $inject[i];
-        if (typeof key !== 'string') {
-          throw $injectorMinErr('itkn',
-                  'Incorrect injection token! Expected service name as string, got {0}', key);
-        }
-        args.push(
-          locals && locals.hasOwnProperty(key)
-          ? locals[key]
-          : getService(key, serviceName)
-        );
-      }
+      var args = injectionArgs(fn, locals, serviceName);
       if (isArray(fn)) {
-        fn = fn[length];
+        fn = fn[fn.length - 1];
       }
 
       // http://jsperf.com/angularjs-invoke-apply-vs-switch
@@ -833,15 +835,19 @@ function createInjector(modulesToLoad, strictDi) {
       return fn.apply(self, args);
     }
 
+
     function instantiate(Type, locals, serviceName) {
       // Check if Type is annotated and use just the given function at n-1 as parameter
       // e.g. someModule.factory('greeter', ['$window', function(renamed$window) {}]);
-      // Object creation: http://jsperf.com/create-constructor/2
-      var instance = Object.create((isArray(Type) ? Type[Type.length - 1] : Type).prototype || null);
-      var returnedValue = invoke(Type, instance, locals, serviceName);
-
-      return isObject(returnedValue) || isFunction(returnedValue) ? returnedValue : instance;
+      var ctor = (isArray(Type) ? Type[Type.length - 1] : Type);
+      var args = injectionArgs(Type, locals, serviceName);
+      // Empty object at position 0 is ignored for invocation with `new`, but required.
+      args.unshift({});
+      /*jshint -W058 */ // Applying a constructor without immediate parentheses is the point here.
+      return new (Function.prototype.bind.apply(ctor, args));
+      /*jshint +W058 */
     }
+
 
     return {
       invoke: invoke,
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -270,6 +270,16 @@ describe('injector', function() {
         it('should take args before first arrow', function() {
           expect(annotate(eval('a => b => b'))).toEqual(['a']);
         });
+
+        it('should be possible to instantiate ES6 classes', function() {
+          // Only Chrome (not even the FF we use) supports ES6 classes.
+          if (!/chrome/i.test(navigator.userAgent)) return;
+          providers('a', function() { return 'a-value'; });
+          var clazz = eval('(class { constructor(a) { this.a = a; } aVal() { return this.a; } })');
+          var instance = injector.instantiate(clazz);
+          expect(instance).toEqual({a: 'a-value'});
+          expect(instance.aVal()).toEqual('a-value');
+        });
         /*jshint +W061 */
       });
     }


mprobst:
Note: this change requires using `Function.bind`, which is quite a bit slower in instantiating objects in a synthetic benchmark ([x20](http://jsperf.com/object-create-vs-function-bind)). However I wouldn't expect that the `instantiate` code flow is very critical - it's usually only used for services and controllers, both of which should be instantiated relatively rarely in a typical application's lifecycle.

IgorMinar:
tests are failing

mprobst:
@IgorMinar should be fixed (FF doesn't enable `class` syntax by default).

test/auto/injectorSpec.js
```diff
@@ -245,12 +245,25 @@ describe('injector', function() {
...
+           function() { expect(annotate(eval('({ fn(x) { return; } })').fn)).toEqual(['x']); });
+
+        it('should be possible to instantiate ES6 classes', function() {
+          // Only Chrome (not even the FF we use) supports ES6 classes.
+          if (!/chrome/i.test(navigator.userAgent)) return;
+          providers('a', function() { return 'a-value'; });
+          var clazz = eval('(class { constructor(a) { this.a = a; } aVal() { return this.a; } })');
+          var instance = injector.instantiate(clazz);
+          expect(instance).toEqual({a: 'a-value'});
+          // Check if prototype is set up correctly.s
```
IgorMinar: you are not checking the prototype + there is a typo at the end of the comment. I'd just remove this.

src/auto/injector.js
```diff
@@ -827,16 +829,18 @@ function createInjector(modulesToLoad, strictDi) {
...
-      // Object creation: http://jsperf.com/create-constructor/2
-      var instance = Object.create((isArray(Type) ? Type[Type.length - 1] : Type).prototype || null);
-      var returnedValue = invoke(Type, instance, locals, serviceName);
-
-      return isObject(returnedValue) || isFunction(returnedValue) ? returnedValue : instance;
+      var ctor = (isArray(Type) ? Type[Type.length - 1] : Type);
+      var args = injectionArgs(Type, locals, serviceName);
+      // Empty object at position 0 is ignored for invocation with `new`, but required.
+      args.unshift({});
+      return new (Function.bind.apply(ctor, args));
```
gkalpak: Wouldn't this cause issues with CSP (no 'unsafe-eval') ?

test/auto/injectorSpec.js
mprobst: Well, the call to `aVal()` would fail if the prototype was incorrect. But point taken, dropped the comment.

src/auto/injector.js
mprobst: I don't think so - this is not `eval`ing code, it's just rebinding the constructor function.

mprobst:
PTAL.

Narretz:
@mprobst While you are at it, could you check this related issue: https://github.com/angular/angular.js/issues/12614? If $controller uses $injector, shouldn't this PR make it work?

mprobst:
#12614 is not fixed by this – `$controller` uses `$injector`, but `bindToController` has its own logic to instantiate controllers on directives. I'm not sure it'll be possible to fix `bindToController`, it's way of instantiating a class on top of an existing object, then `apply`ing the ctor to it might be inherently incompatible with ES6.

caitp:
It generally is, but this might not be too much of a problem for much longer. It's likely that the "new-only" restriction will be lifted, or otherwise a way of specializing a "call" behaviour/factory method for classes.

In the mean time, you could add some other compat mechanism, like falling back on a static "init()" method if calling the constructor fails. There's no reason this couldn't be solved using conventions rather than language features

mprobst:
@caitp: indeed, but what about doing that in another PR?

caitp:
Yeah, doesn't need to be this one, although that feature probably shouldn't be specific to controllers if we do that

stefan--:
when can we expect this to work?

mprobst:
@IgorMinar PTAL, I believe this is good to merge. It does not solve the issue for `controllerAs` yet, but that's more tricky (and may or may not be impossible).

src/auto/injector.js
lgalfaso: Should this be `Function.prototype.bind.apply`?

mprobst: With `Function` being a `Function`, thus `bind` resolving on the `prototype` of `Function` in any case, I don't think it makes a difference. But done nonetheless for consistency.

IgorMinar:
lgtm. @petebacondarwin should we wait with a merge until master is open for 1.5 beta features?

IgorMinar:
@petebacondarwin, one more thing - this feature is not documented at all. where should the documentation live?

src/auto/injector.js
```diff
@@ -795,31 +795,33 @@ function createInjector(modulesToLoad, strictDi) {
...
       var args = [],
-          $inject = createInjector.$$annotate(fn, strictDi, serviceName),
-          length, i,
-          key;
+          $inject = createInjector.$$annotate(fn, strictDi, serviceName);
 
-      for (i = 0, length = $inject.length; i < length; i++) {
-        key = $inject[i];
+      for (var i = 0, length = $inject.length; i < length; i++) {
+        var key = $inject[i];
```
petebacondarwin: Why move the declarations of `i` and `key` here? Being `var` declarations they will still be hoisted to the top of the function block.

petebacondarwin: Also we seem to have lost the declaration of the `length` variable here.

petebacondarwin:
> However I wouldn't expect that the instantiate code flow is very critical - it's usually only used for services and controllers, both of which should be instantiated relatively rarely in a typical application's lifecycle.

I am not sure about this statement. Controllers can be instantiated very often. Every instance of a directive which defines a "directive controller" would need to instantiate a directive. Also imagine if you use `ng-controller` inside a frequently changing `ng-repeat` block.

src/auto/injector.js
caitp: ` length` is delayed as part of the var declaration

caitp: Declared, darned Swype

petebacondarwin:
I guess the documentation for this should be in the `$injector` class: https://docs.angularjs.org/api/auto/service/$injector

src/auto/injector.js
mprobst: @petebacondarwin I was moving the code around anyway, and I personally find it easier to declare things as close as possible to their respective use site, hoisting or not. Makes it easier to see what's used where and why, and also makes it easier to refactor code without forgetting to remove declarations.

mprobst:
@petebacondarwin absent a good benchmark, this is all guesswork. However I would imagine the cost of this particular call being dwarfed by the cost of both `ng-repeat` and the DOM manipulations for directive controllers or `ng-controller` in a loop. I'd expect you can construct a synthetic benchmark that does slow down, but I wouldn't expect actual apps experiencing that.

mprobst:
Friendly ping?

IgorMinar:
@petebacondarwin is this too high risk for 1.5? several people at Google are asking for this change.

IgorMinar:
+ @lgalfaso - what do you think?

lgalfaso:
@IgorMinar it mostly moves code around, and it adds a little *magic* but it is reasonably documented. From my point of view, it looks safe for 1.5. If @petebacondarwin does not oppose, then I will merge it EOD Friday.

petebacondarwin:
Go for it @lgalfaso

bisubus:
@petebacondarwin It looks like the commit does not affect [*later* argument](https://github.com/angular/angular.js/blob/62d514b06937cc7dd86e973ea11165c88343b42d/src/ng/controller.js#L117) in `$controller` which still uses `$injector.invoke`.

IgorMinar:
sgtm

petebacondarwin:
@bisubus - thanks for pointing that out.
Can we create a new issue for that case so that we can track it and work on fixing it?

petebacondarwin:
I have created the issue at https://github.com/angular/angular.js/issues/13510

PR #12603, opened 2015-08-17
Fixes for ngAnimate in 1.3

### LINKED ISSUE #10205: ngAnimate cause errors after updating to 1.3 (Cannot read property 'parentNode' of undefined)
I think the guy who filed #9953 gave up too easily. :)  I'm seeing the same error message in Angular 1.3.  I'm using a modified version of the angular-ui bootstrap datepicker that uses ng-if to only render the control if it is open, and so I'm reasonably certain the analysis by @caitp in that issue is exactly the problem for me.  If I add the following at line 1201 in angular-animate.js after the call to extractElementNode(), the error message goes away:

```javascript
if (!elementNode) {
    done();
    return;
}
```

There's also a comment by @matsko there about refactoring some code into animationsDisabled() after #8092 is fixed, which it now is, so I'm not sure if there's some more work to be done in that area.

I have not noticed any consequences to this, since presumably we can't do much to animate a DOM element that isn't there...

### LINKED ISSUE #12604: class-based animations in ngAnimate for 1.3 do not clear cache when the animation fails

matsko:
Closes #10205
Closes #12604
Closes #12603

### DIFF
--- a/src/ngAnimate/animate.js
+++ b/src/ngAnimate/animate.js
@@ -1194,9 +1194,12 @@ angular.module('ngAnimate', ['ng'])
           }
 
           return cache.promise = runAnimationPostDigest(function(done) {
-            var parentElement = element.parent();
-            var elementNode = extractElementNode(element);
-            var parentNode = elementNode.parentNode;
+            var parentNode, parentElement, elementNode = extractElementNode(element);
+            if (elementNode) {
+              parentElement = element.parent();
+              parentNode = elementNode.parentNode;
+            }
+
             // TODO(matsko): move this code into the animationsDisabled() function once #8092 is fixed
             if (!parentNode || parentNode['$$NG_REMOVED'] || elementNode['$$NG_REMOVED']) {
               done();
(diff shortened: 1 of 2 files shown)


src/ngAnimate/animate.js
```diff
@@ -1194,18 +1194,22 @@ angular.module('ngAnimate', ['ng'])
           }
 
           return cache.promise = runAnimationPostDigest(function(done) {
-            var parentElement = element.parent();
-            var elementNode = extractElementNode(element);
-            var parentNode = elementNode.parentNode;
+            var cache, parentNode, parentElement, elementNode = extractElementNode(element);
+            var DOM_FRAGMENT = 11;
```
campersau: It looks like `DOM_FRAGMENT` is never used.

matsko: This get removed when it was merged to master: https://github.com/angular/angular.js/commit/2c03a3574336ed814d020cf7ba36cee5b87e65b5

matsko:
Landed as https://github.com/angular/angular.js/commit/2c03a3574336ed814d020cf7ba36cee5b87e65b5

PR #12613, opened 2015-08-18
fix($compile): backport to 1.2 of workaround for IE11 MutationObserver

### LINKED ISSUE #11781: Interpolations containing HTML entities fail in IE11 when MutationObservers are active

smdvdsn:
Backport #11796 to 1.2 branch.
IE11 MutationObserver breaks consecutive text nodes into several text nodes.
This patch merges consecutive text nodes into a single node before looking for interpolations.

Closes #11781

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3705,7 +3705,7 @@
                   "version": "0.2.1"
                 },
                 "i": {
-                  "version": "0.3.2"
+                  "version": "0.3.3"
                 },
                 "mkdirp": {
                   "version": "0.4.0"
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1088,6 +1088,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
           break;
         case 3: /* Text Node */
+          if (msie === 11) {
+            // Workaround for #11781
+            while (node.parentNode && node.nextSibling && node.nextSibling.nodeType === 3 /* Text Node */) {
+              node.nodeValue = node.nodeValue + node.nextSibling.nodeValue;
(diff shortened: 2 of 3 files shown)


Narretz:
Landed as https://github.com/angular/angular.js/commit/fccce96d444f442ba5ecfb67201505a445d0c209

PR #12619, opened 2015-08-18
fix(ngAnimate): only buffer rAF requests within the animation runners

### LINKED ISSUE #12280: $$rAF performance issue in Angular Material

matsko:
Closes #12280

Here are demos to show that this works (test in IE):

Multiple RAF Requests in Angular 1.4 ([#11791](https://github.com/angular/angular.js/issues/11791))
- Pre 1.4.0 functionality (really slow on **IE 11** ... this was the reason why we decided to bundle the calls into a single function): https://jsfiddle.net/oga2z7bb/3/
- Current functionality (fixed for **IE 11**): https://angular-development.appspot.com/raf-perf/fast/index.html
- Patched functionality (fixed for **IE 11**): https://angular-development.appspot.com/raf-perf/slow/index.html

Material Design $$rAF perf issue ([#12280](https://github.com/angular/angular.js/issues/12280))
- Current functionality (**slow**): https://angular-development.appspot.com/slow-raf/slow/index.html
- Patched functionality (**fast**): https://angular-development.appspot.com/slow-raf/fast/index.html

### DIFF
--- a/src/ngAnimate/animateRunner.js
+++ b/src/ngAnimate/animateRunner.js
@@ -1,18 +1,33 @@
 'use strict';
 
-var $$rAFMutexFactory = ['$$rAF', function($$rAF) {
+var $$AnimateAsyncRunFactory = ['$$rAF', function($$rAF) {
+  var waitQueue = [];
+
+  function waitForTick(fn) {
+    waitQueue.push(fn);
+    if (waitQueue.length > 1) return;
+    $$rAF(function() {
+      for (var i = 0; i < waitQueue.length; i++) {
+        waitQueue[i]();
+      }
+      waitQueue = [];
+    });
+  }
+
   return function() {
     var passed = false;
-    $$rAF(function() {
+    waitForTick(function() {
       passed = true;
     });
-    return function(fn) {
-      passed ? fn() : $$rAF(fn);
+    return function(callback) {
+      passed ? callback() : waitForTick(callback);
     };
   };
 }];
 
-var $$AnimateRunnerFactory = ['$q', '$$rAFMutex', function($q, $$rAFMutex) {
+var $$AnimateRunnerFactory = ['$q', '$sniffer', '$$animateAsyncRun',
+                      function($q,   $sniffer,   $$animateAsyncRun) {
+
   var INITIAL_STATE = 0;
   var DONE_PENDING_STATE = 1;
   var DONE_COMPLETE_STATE = 2;
@@ -57,7 +72,7 @@ var $$AnimateRunnerFactory = ['$q', '$$rAFMutex', function($q, $$rAFMutex) {
     this.setHost(host);
 
     this._doneCallbacks = [];
-    this._runInAnimationFrame = $$rAFMutex();
+    this._runInAnimationFrame = $$animateAsyncRun();
     this._state = 0;
   }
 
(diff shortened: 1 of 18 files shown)


matsko:
@ThomasBurleson @AdriVanHoudt @MichhDiego this should fix the issues you guys are having.

matsko:
@IgorMinar please review.

test/ngRoute/directive/ngViewSpec.js
```diff
@@ -913,9 +912,11 @@ describe('ngView animations', function() {
 
       $location.path('/foo');
       $rootScope.$digest();
-      expect($animate.queue.shift().event).toBe('enter');
-      $animate.triggerCallbacks();
 
+      $animate.flush();
+      $rootScope.$digest();
```
IgorMinar: why do we need the extra digest after flush? it wasn't necessary before and we digest just before flushing already.

src/ngAnimate/animateRunner.js
```diff
@@ -1,18 +1,33 @@
 'use strict';
 
-var $$rAFMutexFactory = ['$$rAF', function($$rAF) {
+var $$AnimateMutexFactory = ['$$rAF', function($$rAF) {
```
IgorMinar: Why do you call this a mutex? it's not quite the same as regular mutex. How about $$AnimateAsyncRun or $$AnimateAsyncExecutor?

src/ngMock/angular-mocks.js
```diff
@@ -763,25 +763,50 @@ angular.mock.animate = angular.module('ngAnimateMock', ['ng'])
...
-        triggerCallbackEvents: function() {
-          $$rAF.flush();
-        },
-        triggerCallbackPromise: function() {
-          $timeout.flush(0);
-        },
-        triggerCallbacks: function() {
-          this.triggerCallbackEvents();
-          this.triggerCallbackPromise();
+        flush: function() {
```
IgorMinar: didn't you want to document the flush method as public api?

IgorMinar:
the rest looks good. thanks!

AdriVanHoudt:
thank you so much :clap:

src/ngMock/angular-mocks.js
matsko: Let's do this one in a follow up commit later this week. That way we can have a page in the guide that explains animation testing.

src/ngAnimate/animateRunner.js
matsko: Changed to `$$AnimateAsyncRun`

test/ngRoute/directive/ngViewSpec.js
matsko: So beforehand what happened was:

1. `triggerCallbacks()` issued a `$$rAF.flush()` internally when called.
2. `triggerCallbacks()` also calls `$timeout.flush(0)` which calls a digest somewhere down the line.

But with the changes to use `$animate.flush()` we end up having to do this:

1. `$animate.flush()` clears out the RAF for the runner.
2. `$rootScope.$digest()` allows for the resolved promise to flush (which can only be done after the RAF has flushed).

So we need to call RAF + digest always.

matsko:
Closed with https://github.com/angular/angular.js/commit/dc48aadd26bbf1797c1c408f63ffde99d67414a9

PR #12626, opened 2015-08-19
fix(ngAria): simplify aria-required attribute detection to support angular v1.2.x

KrekkieD:
Remove the ngModel.$validators check to support angular 1.2.x. The check
serves no purpose but breaks aria on older angular versions.

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -301,7 +301,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             elem.attr('tabindex', 0);
           }
 
-          if (ngModel.$validators.required && shouldAttachAttr('aria-required', 'ariaRequired', elem)) {
+          if (shouldAttachAttr('aria-required', 'ariaRequired', elem)) {
             scope.$watch(function ngAriaRequiredWatch() {
               return ngModel.$error.required;
             }, function ngAriaRequiredReaction(newVal) {


gkalpak:
`ngAria` is not supported in 1.2.x (there is no 1.2.x-compatible version of `ngAria`).
I don't think this change makes sense.

KrekkieD:
This makes it compatible. The only thing that breaks it is the missing '$validators' reference. I was quite surprised myself.

gkalpak:
It doesn't make it compatible; it just prevents throwing an error. In fact, this change breaks `ngAria` for both 1.2.x and future releases.

The point is not if it is throwing errors and not if it kind-of-works right now, but if it works correctly (which it doesn't) and if it won't break in the future (and it probably will).

--
Basically, in order to ensure correct behaviour, you should always use the same version of external modules (such as `ngAria`) as your core angular. E.g. with angular v1.4.3 you should use `ngAria` v1.4.3, with angular v1.4.4 `ngAria` v1.4.4 etc.

Right now there is not version of `ngAria` for v1.2.x. If one wanted to backport it, they should create (and maintain) a different version of `ngAria` which would be compatible with angular v1.2.x. This version would be tested against angular version 1.2.x (so they need to create and maintain appropriate test-cases) etc.

Since the 1.2.x is the legacy version and only accepts critical bug-fixes atm, I don't think it makes sense to commit resources and cycles to such an undertaking.

KrekkieD:
The error breaks the full app with loss of all functionality. If nothing else, the condition could be changed to `ngModel.$validators && ngModel.$validators.required` which will lose the `aria-required` implementation on v1.2.x and nothing more.

I understand it's far from ideal, and I understand not wanting to waste resources. But a small change like this can help many users by getting the aria attributes, I would say it's a quick win. I'm not asking you to officially support 1.2.x, just asking for a minor patch that makes it work for 1.2.x.

What I do notice is that the change I proposed sets `aria-required` on all `ng-model` elements, watcher included. This may not be entirely ideal, also because this influences behaviour in angular ^1.3.0.

It could also be patched to this to keep the same behaviour, but it is a little more verbose with no direct benefit for ^1.3.0;

          var needsAriaRequired = (ngModel.$validators ? ngModel.$validators.required : true) &&
              shouldAttachAttr('aria-required', 'ariaRequired', elem);

          if (needsAriaRequired) {
            scope.$watch(function ngAriaRequiredWatch() {
              return ngModel.$error.required;
            }, function ngAriaRequiredReaction(newVal) {
              elem.attr('aria-required', !!newVal);
            });
          }


In terms of it working for 1.2.x, you said it doesn't work correctly. Can you elaborate on this? I tried all features in the aria lib and did not see anything weird?

Narretz:
@KrekkieD You said it yourself, it breaks ngAria by attaching aria-required to all models, regardless if they are actually required. There's a test missing for that. I agree with @gkalpak here:
```
Right now there is not version of ngAria for v1.2.x. If one wanted to backport it, they should create (and maintain) a different version of ngAria which would be compatible with angular v1.2.x. This version would be tested against angular version 1.2.x (so they need to create and maintain appropriate test-cases) etc.
```

KrekkieD:
That's too much overhead for me, I just wanted to contribute with a minor patch that would make it backwards compatible.

A better solution specifically for 1.2.x with identical behaviour to ^1.3.0 would be this (also needs injection of `$timeout` in the `ngModel` directive):

```javascript
    if (shouldAttachAttr('aria-required', 'ariaRequired', elem)) {
        // needs an additional $digest for ngModel.$error to populate
        $timeout(function () {
            if (ngModel.$error.hasOwnProperty('required')) {
                scope.$watch(function ngAriaRequiredWatch() {
                    return ngModel.$error.required;
                }, function ngAriaRequiredReaction(newVal) {
                    elem.attr('aria-required', !!newVal);
                });
            }
        });
    }
```

I think copying the `ngAria` tests from ^1.3.0 would run fine on 1.2.x.

Could include `ngMessages` too, as that also seems to be working fine for 1.2.x, only seems to need `Array.indexOf` to be patched for IE8 (at least that's what makes `ng-messages` and `ng-message` work for my needs, have not tried `ng-message-exp`). Simply replacing `Array.indexOf` with a custom `indexOf` function (like done with `Array.isArray`) would patch this right up without polyfilling (which is what I do in my implementation of ngMessages@1.4.4 for 1.2.x).

If there are active developers for 1.2.x I hope they'll be willing to integrate it. I think these two libraries would make a huge contribution for 1.2.x, and I expect they are very easy to backport. The fact that they're very small makes this easier, and the code fixes I provided might cover all issues.

From a corporate perspective, with the business requirement to support older browsers (hello IE8) and thus angular 1.2.x, we're unfortunately left in the dark because of these very minor incompatibilities, which hurts our Chrome/Firefox users too.

If you want to help me along the way, pretty please, tag a developer of angular 1.2.x, who has the CI and everything set up for 1.2.x and could give this a try in minutes (copy tests from 1.3.x to 1.2.x, patch code in `ngAria` and `ngMessages` as described, and run tests). If it would work out of the box in just a few minutes time, wouldn't that be a beautiful thing for everyone involved?

Narretz:
See my comment https://github.com/angular/angular.js/pull/12646#issuecomment-142562845

Issue #12618, opened 2015-08-18
ng-pattern only work in angular 1.2.1, but not on 1.3.17, 1.4.3 and 1.4.4


pokaxperia:
Hello, I have a problem, I trying to validate an input number, setting 00.00. what I need?, only two digits after and before decimal dot, no matter if the two digits after decimal dot are 00.

Between many patterns, I found this that it serves me:
**/^\d{0,1}(\d\.\d)?\d?$/**

I created four examples on plunker with different versions of angularjs, that explain my problem and no matter if set **pattern** instead **ng-pattern** attribute, still not working.

Angularjs 1.4.4 http://plnkr.co/edit/K00ffy?p=preview
Angularjs 1.4.3 http://plnkr.co/edit/jCYj97?p=preview
Angularjs 1.3.17 http://plnkr.co/edit/93absW?p=preview
**Angularjs 1.2.1 http://plnkr.co/edit/ooFneS?p=preview**

The last one works, but my question or the question is, what it's wrong?, me or the versions of AngularJS?.

Thank you.

kacecode:
It has to do with the way angular is handling the input -- After your 1.2.1 example the input fed into the pattern validator is a number type instead of a string. Thus, when it's `00000.000000000` it's passed into the validator as `0` then cast to the string "0".

There was a pretty heavy refactor in 1.3 that appears to have added the $parser to the inputDirective which is what is causing your woes.

Narretz:
Thanks for opening the issue. It's being tracked here already: #12344

pokaxperia:
Thank you for the answer, but..., the solution is????, or there is no solution ?, so it stays?. Greetings. :)

kacecode:
@pokaxperia the fix is the above patch, which you will have to wait on the release of. In the interim, you can get regex matching from a text field

PR #12633, opened 2015-08-20
fix(ngPattern): Changed pattern to validate against input field value

### LINKED ISSUE #12344: ng-pattern binding is not working in v1.4
I understand that v1.3 introduced some changes in the recommended usage of ng-pattern when the validation pattern / RegExp is exposed as a model property. This change is documented in the [migration notes](https://docs.angularjs.org/guide/migration) and works correctly in v1.3, as demonstrated in [this JSFiddle](http://jsfiddle.net/Lzgts/500/).

However, using the same implementation in v1.4 does not work, as [this JSFiddle](http://jsfiddle.net/Lzgts/502/) shows. Instead, pattern validation always fails.

kacecode:
Set patternDirective validator to test against viewValue instead of
modelValue.

Closes #12344, #12618

### DIFF
--- a/src/ng/directive/validators.js
+++ b/src/ng/directive/validators.js
@@ -43,8 +43,8 @@ var patternDirective = function() {
         ctrl.$validate();
       });
 
-      ctrl.$validators.pattern = function(value) {
-        return ctrl.$isEmpty(value) || isUndefined(regexp) || regexp.test(value);
+      ctrl.$validators.pattern = function(modelValue, viewValue) {
+        return ctrl.$isEmpty(viewValue) || isUndefined(regexp) || regexp.test(viewValue);
       };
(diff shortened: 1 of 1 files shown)


gkalpak:
I think we need some tests for this fix, in order to catch potential regressions in the future.
@kacecode, could you push a second commit with some tests (i.e. testcases that would fail **without** the fix, but pass **with** the fix) ?

Narretz:
I actually created a commit with the fix, but didn't get around to post it. Here it is: https://github.com/angular/angular.js/pull/12640

Narretz:
I'll have a test for this, so we'll go with #12640. But thanks for the PR, @kacecode. It's great to see contributions.

Issue #12635, opened 2015-08-20
$timeout / $browser.defer broken on iOS 9 Cordova UIWebView


jd-carroll:
On iOS 9 runing Cordova with UIWebView, $timeout will invoke multiple times for a single request.

Here's a simple example:

```
<!DOCTYPE html>
<html>
    <head>
        <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.3.16/angular.js"></script>
        <script>
            angular.module('timeout', [])
            .config(function($timeout) {
              var i = 0;
              $timeout(function() {
                i++;
                console.log('$timeout invoked ' + i + ' times');
              }, 0);
        </script>
    </head>
    <body ng-app="timeout">
        <div ng-view></div>
    </body>
</html>
```

It does not matter how long the delay is.

jd-carroll:
I should note that the issue is with the `window.setTimeout` which will affect everything that uses it

Narretz:
If the problem is with setTimeout then Apple needs to fix it. no? Have you reported this to Apple?

jd-carroll:
Open Radar: https://openradar.appspot.com/22360634

Narretz:
Thanks. Let's hope Apple fixes this before the release, it's a pretty big bug. If you see that it is resolved, can you close this issue?

jd-carroll:
Absolutely, I'll keep the rdar updated as best as possible with any updates from apple (if it doesn't happen automatically)

realityfilter:
I tried the code snippet from the apple bug report and got only one debug line. Is used an iPhone 5 with iOS 9 beta5 and the simulator for an iPad Air. 

I also decorated $browser.defer the following way:

```javascript
    app.config(['$provide', ($provide) => {
        $provide.decorator('$browser', ['$delegate', ($delegate) => {
            var origDefer = $delegate.defer;

            var saveInvoke = function(fn) {
                var invoked = 0;
                return function () {
                    if (invoked > 0) return console.log("Called multiple times: " + invoked);

                    try {
                        fn.apply(this, arguments);
                    } finally {
                        ++invoked;
                        fn = angular.noop;
                    }
                }
            };

            function newDefer(fn, timeout) {
                return origDefer(saveInvoke(fn), timeout);
            }

            $delegate.defer = newDefer;
            $delegate.defer.cancel = origDefer.cancel;

            return $delegate;
        }]);
    }]);
```

And received no multiple invocations. I tested this on an iPad Air, too.

[...]

PR #12636, opened 2015-08-20
Workaround for multiple invocations on iOS 9 Cordova UIWebView

jd-carroll:


jd-carroll:
This is a proposed fix/workaround for #12635

jd-carroll:
Closing as @realityfilter offered a better workaround in the issue.

Issue #12631, opened 2015-08-20
ngHide/ng-show directive don't  animate in ng 1.4.4


e-cloud:
when i update to angular 1.4.4, there is a strange thing -- some animation i define for the element with ng-hide directive doesn't seem to work. Then i check the api docs. Ok, [the 1.4.4 version docs of ng-show] (https://code.angularjs.org/1.4.4/docs/api/ng/directive/ngShow)'s [example plunkr](https://code.angularjs.org/1.4.4/docs/api/ng/directive/ngShow#example) seems to have no animation too.

**Does Angular have enough tests before the new version is about to release?**

Narretz:
Can you please provide a demo of the problem? Otherwise, we cannot do anything to fix the problem.

A11oW:
Just look how work [https://code.angularjs.org/1.4.4/docs/api/ng/directive/ngShow#example](example ) or [http://plnkr.co/edit/y2JZH3xTB9MI6VXZ0o4b?p=preview](plunker)
when check labels angular don't add intermediates classes:
```
.ng-hide-add
.ng-hide-add-active,
.ng-hide-remove
.ng-hide-remove-active
```

A11oW:
I think that proble related with #12612, #12583

Narretz:
Ok, I see now. @matsko can you take a look at this? It seems pretty bad if even the "official" ng-hide animations aren't working.

matsko:
Turns out that the API docs example code is incorrect. Transitions always need to be applied to the non active CSS class:
http://plnkr.co/edit/JzI0NnxtPdySR2sMcGeW?p=preview

Or on the class that is being added or removed:
http://plnkr.co/edit/EdhLQkurZRRYjhmSUrrX?p=preview

Or on a shared CSS class that is always there:
http://plnkr.co/edit/HX9BQwtwJ5WvylbWVOa3?p=preview

Placing it on the active CSS classes may have worked before, and I will look into making the fix properly for 1.4.5. Here is an issue to keep note of that:
https://github.com/angular/angular.js/issues/12638

In the meantime here is a commit that will clean up the example in the docs a bit so nobody else comes across this issue:
https://github.com/angular/angular.js/pull/12637

matsko:
OK 1.4.5 will fix this entirely: http://plnkr.co/edit/y2JZH3xTB9MI6VXZ0o4b?p=preview

matsko:
[...]

Issue #12594, opened 2015-08-16
Animation staggering issue in 1.4.4


serjrd:
Here's a plunker snippet that illustrates the issue: http://plnkr.co/edit/PSuu2IOG03fv2ifNTTwQ?p=preview

With the version 1.4.4 there's no staggering.
But If you change the included angular version to 1.4.3 - staggering works fine.

This is under Ubuntu / Chrome Version 44.0.2403.125 (64-bit).

datnordstrom:
1+

I've experienced this as well.

ordinaryAmsel:
+1

Narretz:
@matsko can you take a look?

matsko:
This isn't entirely broken. You can still get the stagger to work by adding `transition-duration: 0s` into the stagger CSS selector.

http://plnkr.co/edit/PSuu2IOG03fv2ifNTTwQ?p=preview
By the way there is no need to define the -webkit- properties for transitions anymore.

This was mentioned in the stagger docs:
https://docs.angularjs.org/api/ngAnimate#css-staggering-animations

Unfortunately with 1.4.4 and onwards this is **always** the case. You must always define the duration value as 0s since we changed the way the ng-EVENT classes are applied. I completely missed making note of this in the CHANGELOG as a breaking change. I will make sure it shows up in the 1.4.5 release notes.

matsko:
This PR contains a fix to the docs: https://github.com/angular/angular.js/pull/12637

serjrd:
Thanks, that helped for the transition animation.

And what about the keyframes animation? I tried it with the 'animation-duration: 0s' property, but it didn't help.
Same code works fine under the 1.4.3 version.
See the example: http://plnkr.co/edit/uAFFOhZ32mDYQVp4Fcco?p=preview

abruzzihraig:
I met the issue like @serjrd, the 1.4.4 version still has a bug in keyframes when use stagger.

And my code like below:

abruzzihraig:
I met the issue like @serjrd too, the 1.4.4 version still has a bug in keyframes when use stagger.

And my code like below:
http://codepen.io/abruzzi/pen/VLodOK

when you choose 1.4.3 version, the code works fine.

Another thing I have noticed that when you write 'transition-duration: 0' without 's', the stagger not performed like we expected, but in 1.4.3, it works.

matsko:
[...]

PR #12637, opened 2015-08-20
simplify the animation docs

matsko:
The animation example contains unnecessarily complex CSS animation
code in it and the conventions are off.

Related #12631
Related #12594


matsko:
Thanks @Narretz

PR #12640, opened 2015-08-20
fix(ngModel): validate pattern against the viewValue

### LINKED ISSUE #12344: ng-pattern binding is not working in v1.4

Narretz:
Since the HTML5 pattern validation constraint validates the input value,
we should also validate against the viewValue. This allows e.g. input[date]
to validate both the input format of the date and the date object itself.

Fixes #12344 

(Also needs to be backported to 1.3)

### DIFF
--- a/test/ng/directive/validatorsSpec.js
+++ b/test/ng/directive/validatorsSpec.js
@@ -204,6 +204,21 @@ describe('validators', function() {
       expect($rootScope.form.test.$error.pattern).toBe(true);
       expect(inputElm).not.toBeValid();
     });
+
+
+    it('should validate the viewValue and not the modelValue', function() {
+      var inputElm = helper.compileInput('<input type="text" name="test" ng-model="value" pattern="\\d{4}">');
+      var ctrl = inputElm.controller('ngModel');
+
+      ctrl.$parsers.push(function(value) {
+        return (value * 10) + '';
+      });
+
+      helper.changeInputValueTo('1234');
+      expect($rootScope.form.test.$error.pattern).not.toBe(true);
+      expect($rootScope.form.test.$modelValue).toBe('12340');
+      expect(inputElm).toBeValid();
+    });
   });
 
 
--- a/src/ng/directive/validators.js
+++ b/src/ng/directive/validators.js
@@ -43,8 +43,9 @@ var patternDirective = function() {
         ctrl.$validate();
       });
 
-      ctrl.$validators.pattern = function(value) {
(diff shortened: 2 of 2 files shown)


lgalfaso:
Would it be possible to know when this broke? Would like to know this to know to what extend this is a regression vs a breaking change.

Narretz:
It's not completely clear-cut. It worked last in 1.2.x, see https://github.com/angular/angular.js/issues/12618. In 1.3.x we changed two things: validation is done in the $validators, not the $parsers, and input[number/date] parse the input to Number/Date, so validating the pattern against the modelValue doesn't work anymore. That's the regression. The potential bc is that pattern now explicitly validates the viewValue, not the modelValue.

I actually had a bc note with a workaround for the latter case, but I removed it because the actual use case being broken is a regression. We can add it as a potential breaking change:

test/ng/directive/validatorsSpec.js
```diff
@@ -204,6 +204,21 @@ describe('validators', function() {
       expect($rootScope.form.test.$error.pattern).toBe(true);
       expect(inputElm).not.toBeValid();
     });
+
+
+    it('should validate the viewValue and not the modelValue', function() {
+      var inputElm = helper.compileInput('<input type="text" name="test" ng-model="value" pattern="\\d{4}$">');
```
gkalpak: I suspect this is a copy-paste from the previous test, but the trailing `$` doesn't seem to be necessary (neither here nor in other tests).

Narretz: Ah, you got me. Guilty of c&p! ;)

lgalfaso:
I am ok for this landing into the 1.4 and 1.3 branches with a BC notice. If you think this is not needed, then ping @petebacondarwin

Narretz:
Updated - please take a look

petebacondarwin:
LGTM!!

lgalfaso:
LGTM!

gkalpak:
Merged as 0e00108.

PR #12643, opened 2015-08-21
docs$rootScope.Scope#$watchGroup: clarify listener's oldValues parameter

rehmsen:
Mention explicitly that oldValues may contain changes that happened before the last call to listener, as this seems non-obvious, see
- https://github.com/angular/angular.js/issues/8671
- https://github.com/angular/angular.js/issues/12452
- http://plnkr.co/edit/Ir1X88bRL95mvm7l6f4N?p=preview

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -418,8 +418,9 @@ function $RootScopeProvider() {
        *    expression in `watchExpressions` changes
        *    The `newValues` array contains the current values of the `watchExpressions`, with the indexes matching
        *    those of `watchExpression`
-       *    and the `oldValues` array contains the previous values of the `watchExpressions`, with the indexes matching
-       *    those of `watchExpression`
+       *    and the `oldValues` array contains the value of each `watchExpression` before the last change (it 
+       *    will contain an older value if this `watchExpression` has not changed since `listener` was last
+       *    called). The indexes match those of `watchExpression`. 
        *    The `scope` refers to the current scope.
        * @returns {function()} Returns a de-registration function for all listeners.
        */


smashercosmo:
The most valuable PR ever. And I'm only half-joking)) This was a great source of confusion for me.

src/ng/rootScope.js
```diff
@@ -418,8 +418,9 @@ function $RootScopeProvider() {
        *    expression in `watchExpressions` changes
        *    The `newValues` array contains the current values of the `watchExpressions`, with the indexes matching
        *    those of `watchExpression`
-       *    and the `oldValues` array contains the previous values of the `watchExpressions`, with the indexes matching
-       *    those of `watchExpression`
+       *    and the `oldValues` array contains the value of each `watchExpression` before the last change (it
```
Narretz: I don't think this is clear enough. My head actually hurts a bit reading it. I will also see if I can come up with a better wording.

Narretz:
Thanks for the PR, this is important info indeed. I think we should also add an example that shows how the watchGroup executes the listener, and with what values.

src/ng/rootScope.js
rehmsen: I agree it is not perfect. It is pretty tricky to describe this behavior concisely and clearly. New attempt:

and the `oldValues` array contains the value of each `watchExpression` before it was last changed (even if that change did not happen in this digest cycle).

Happy for any suggestion with regards to wording - as long as we mention this for me unexpected behavior explicitly. If you can tell which part exactly find hard to understand, I will try to improve that specifically.

rehmsen:
I signed it!

Narretz:
Docs have been updated in https://github.com/angular/angular.js/commit/06baf1869bb6bd6a7255a47e52ac83de23ce0341

PR #12654, opened 2015-08-22
Fix #1678 by allowing headers.Authorization to be a 2 item username/password array

SchizoDuckie:
Just to kick things off after my $provide.decorator fix in #1678 

todo:
- [ ] docs
- [ ] tests (how?)


SchizoDuckie:
CLA signed.

PR #12525, opened 2015-08-08
Issue with canceling XHR request using promises - fixed

netman92:
Fixes #10890 and #9332.

Issue #9332, opened 2014-09-29
Can no longer cancel $resource request with a promise


simonjm:
I've tested this on Chrome 37.0.2062.124 m, Firefox 32.0.3, and IE11 on Windows 8.1

Here is some code to reproduce the issue.

```html
<body ng-app="testApp">
    <div ng-controller="TestCtrl">
        <button type="button" ng-click="clear()">Clear</button>
        <ul>
            <li ng-repeat="val in dataset">
                {{val | json}}
            </li>
        </ul>
        <button type="button" ng-click="start()">Start</button>
        <button type="button" ng-click="stop()">Stop</button>
    </div>
    <script type="text/javascript" src="https://code.angularjs.org/1.3.0-rc.3/angular.js"></script>
    <script type="text/javascript" src="https://code.angularjs.org/1.3.0-rc.3/angular-resource.js"></script>
    <script type="text/javascript">
        angular.module('testApp', ['ngResource']).controller('TestCtrl', function ($scope, $resource, $q) {                                   
            var promises = [];
            function getResource() {
                var defered = $q.defer();
                promises.push(defered);
                return $resource('/data', {}, {
                    query: {
                        method: 'GET',
                        isArray: true,
                        timeout: defered.promise
                    }
                });
            }

            $scope.start = function () {
                getResource().query(function (res) {
                    $scope.dataset = res;
                });
            };

            $scope.stop = function () {
                promises.forEach(function (p) { p.resolve(); });
                promises = [];
            };

            $scope.clear = function() {
                $scope.dataset = [];
            }
        });
    </script>
</body>
```
https://github.com/angular/angular.js/commit/23bc92b17df882a907fb326320f0622717fefe7b appears to be the culprit. Before this the above code works as expected.

tsonev:
This is a very valuable article, around which I have created my own duplicate request aborter service. 

https://developer.rackspace.com/blog/cancelling-ajax-requests-in-angularjs-applications/

Does it mean that from 1.3 on this will not be possible anymore?

Orcrist78:
try to change this line of ngResource source code:

httpConfig[key] = copy(value);

in

httpConfig[key] = key !== 'timeout' ? copy(value) : value;

the problem is the copied promise

valente:
I tracked down this issue to [resource.js, 563](https://github.com/angular/angular.js/blob/master/src/ngResource/resource.js#L563-567):

<pre lang="javascript"><code>
            forEach(action, function(value, key) {
[...]

PR #12657, opened 2015-08-23
Issue with canceling XHR request using promises - fixed #12525

### LINKED ISSUE #10890: $resource does not abort the requests through the promise
https://github.com/angular/angular.js/blob/master/src/ngResource/resource.js#L564

netman92:
Fixes #10890 and #9332.

@jtheoof no conflicts, is it ok now?

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -568,8 +568,10 @@ angular.module('ngResource', ['ng']).
               undefined;
 
             forEach(action, function(value, key) {
-              if (key != 'params' && key != 'isArray' && key != 'interceptor') {
+              if (key != 'params' && key != 'isArray' && key != 'interceptor' && key != 'timeout') {
                 httpConfig[key] = copy(value);
+              } else if (key == 'timeout') {
+                httpConfig[key] = value;
               }
             });
 
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -568,10 +568,17 @@ angular.module('ngResource', ['ng']).
               undefined;
 
             forEach(action, function(value, key) {
-              if (key != 'params' && key != 'isArray' && key != 'interceptor' && key != 'timeout') {
-                httpConfig[key] = copy(value);
-              } else if (key == 'timeout') {
-                httpConfig[key] = value;
+              switch (key) {
+                default:
+                  httpConfig[key] = copy(value);
+                  break;
+                case 'params':
+                case 'isArray':
+                case 'interceptor':
+                  break;
+                case 'timeout':
+                  httpConfig[key] = value;
+                  break;
               }
             });
 
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -1357,7 +1357,7 @@ describe('resource', function() {
       );
   });
 
-  it('If timeout promise is resolved, cancel the request', function() {
+  it('should cancel the request if timeout promise is resolved', function() {
     var canceler = $q.defer();
 
     $httpBackend.when('GET', '/CreditCard').respond({data: '123'});
@@ -1372,7 +1372,7 @@ describe('resource', function() {
     CreditCard.query();
 
     canceler.resolve();
-    expect(function() { $httpBackend.flush();}).toThrow(new Error("No pending request to flush !"));
+    expect($httpBackend.flush).toThrow(new Error("No pending request to flush !"));
 
     canceler = $q.defer();
     CreditCard = $resource('/CreditCard', {}, {
@@ -1383,9 +1383,7 @@ describe('resource', function() {
     });
 
     CreditCard.query();
-    expect(function() { $httpBackend.flush();}).not.toThrow(new Error("No pending request to flush !"));
-
-
+    expect($httpBackend.flush).not.toThrow();
   });
 
 
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -1308,7 +1308,7 @@ describe("resource", function() {
 });
 
 describe('resource', function() {
-  var $httpBackend, $resource;
+  var $httpBackend, $resource, $q;
 
   beforeEach(module(function($exceptionHandlerProvider) {
     $exceptionHandlerProvider.mode('log');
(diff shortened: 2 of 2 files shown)


jtheoof:
:+1: Although the commit message does not seem to follow the angular commit message conventions. I would preprend the commit the message with: `fix(ngResource): ...`

netman92:
@jtheoof fixed

src/ngResource/resource.js
```diff
@@ -568,8 +568,10 @@ angular.module('ngResource', ['ng']).
               undefined;
 
             forEach(action, function(value, key) {
-              if (key != 'params' && key != 'isArray' && key != 'interceptor') {
+              if (key != 'params' && key != 'isArray' && key != 'interceptor' && key != 'timeout') {
```
caitp: What if we just made this a switch statement for readability?

netman92: @caitp Done

jtheoof:
:+1:

test/ngResource/resourceSpec.js
```diff
@@ -1356,5 +1357,36 @@ describe('resource', function() {
       );
   });
 
+  it('If timeout promise is resolved, cancel the request', function() {
```
caitp: nit: can you change the test description to match the grammar of other ones in the file?

caitp:
Other than the nit it looks great to me. Would be good if @gkalpak or @Narretz would take a look as well before merging

test/ngResource/resourceSpec.js
netman92: @caitp done

test/ngResource/resourceSpec.js
```diff
@@ -1356,5 +1357,36 @@ describe('resource', function() {
...
+      query: {
+        method: 'GET',
+        timeout: canceler.promise
+      }
+    });
+
+    CreditCard.query();
+
+    canceler.resolve();
+    expect(function() { $httpBackend.flush();}).toThrow(new Error("No pending request to flush !"));
```
gkalpak: Nit: `expect($httpBackend.flush).toThrow(...);` is more concise and readable (here and below).
(Else you need a space after `flush();`.)

caitp: If it (flush) uses `this` at all, it's not an option anyways

test/ngResource/resourceSpec.js
```diff
@@ -1356,5 +1357,36 @@ describe('resource', function() {
...
+    canceler = $q.defer();
+    CreditCard = $resource('/CreditCard', {}, {
+      query: {
+        method: 'GET',
+        timeout: canceler.promise
+      }
+    });
+
+    CreditCard.query();
+    expect(function() { $httpBackend.flush();}).not.toThrow(new Error("No pending request to flush !"));
```
gkalpak: I would leave it at `.not.toThrow()` (to account for the unlikely situation where the error message changes and the behaviour breaks (simultaneously)).
It's also more concise and readable.

test/ngResource/resourceSpec.js
netman92: @gkalpak fixed

test/ngResource/resourceSpec.js
```diff
@@ -1356,5 +1357,36 @@ describe('resource', function() {
...
+    canceler = $q.defer();
+    CreditCard = $resource('/CreditCard', {}, {
+      query: {
+        method: 'GET',
+        timeout: canceler.promise
+      }
+    });
+
+    CreditCard.query();
+    expect($httpBackend.flush).not.toThrow(new Error("No pending request to flush !"));
```
gkalpak: I would leave it at `not.toThrow()` (accounting for the unlikely situation where the error message changes and the behaviour breaks (simultaneously)).
It's also more concise and readable.

Also, remove the 2 empty lines below.

netman92: @gkalpak Fixed

gkalpak:
LGTM :+1: 
<sub>
TBH, I don't find it very useful, since having to recreate the resource class everytime a timeout promise is resolved doesn't seem very practical. But if people are using it, it's fine by me :smiley:
</sub>

netman92:
@caitp Can you merge it?

Narretz:
What @gkalpak said seems like the behavior is not well defined. Do we need to update the docs? Also, the commits must be squashed and the commit message doesn't follow our guidelines.

caitp:
the behaviour seems pretty well defined to me @Narretz. It would be cool if you could override it on a per-request basis, but hey.

Some time ago, I had proposed that we allow letting you specify a factory function to produce the timeout promise, I can't remember if that's been done yet or not. That would solve the usability concerns.

@netman92 can you update the docs to indicate that the timeout parameter can be specified during initial resource class construction? Then I'll squash and merge

netman92:
@caitp https://docs.angularjs.org/api/ngResource/service/$resource
and param timeout

```
timeout – {number|Promise} – timeout in milliseconds, or promise that should abort the request when resolved.
```
 I think, it is already in docs.

caitp:
Not really, it's not an accurate depiction of what happens

netman92:
@caitp Sorry, but I do not know, what to add to docs.

jtheoof:
I agree with @netman92. The documentation seems correct, the behaviour was incorrect in a sense that the `Promise` was copied, therefore not cancellable. I think it's good to merge.

jvorcak:
The documentation seems correct to me as well. Do you think it will be merged eventually @caitp? I'd really need this for one of my apps.

gkalpak:
You know what they say: It's never too late, right ?)
Well, I just realized there is an issue with this change:

If the user fails to re-instantiate the resource after having cancelled a request (i.e. after having resolved the `timeout` promise), **all** subsequent requests will fail (and this behaviour isn't even documented).
You can see it in action [here][1] (with the current master).

IMO, we should come up with a better approach to request cancelling for `$resource` (let's discuss in #13058).
At the very least, a warning should be added in the docs and a BC notice in the changelog (but I would prefer fixing the behaviour instead). 

[1]: http://codepen.io/ExpertSystem/pen/meLoqE?editors=001

petebacondarwin:
I agree - let's expedite #13058

gkalpak:
#13058 has been superseeded by #13210.

PR #12658, opened 2015-08-23
fix(ngModel): let aliased validator directives work on any element

### LINKED ISSUE #12158: minlength validator binded to ng-model not input
I'm using custom directive to build my form.
Like `<my-input ng-model="vm.value" required="true" minlength="6">`

This example shows that in validation required is bind to ng-model and works great. But `minlength` or `ng-minlength` is different - I suppose it works only with `input` tag.

Can You bing `ng-minlength` with `ng-model`?

Should I make my own validator?

Narretz:
`ng-(pattern|minlength|maxlength)` directives will now validate the
`ngModel` when on an element that is not an `input` or
a `textarea`.
The commit also adds a test for `ng-required`, although that validator
worked on all elements before this fix.

Fixes #12158

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -556,9 +556,8 @@ function getBooleanAttrName(element, name) {
   return booleanAttr && BOOLEAN_ELEMENTS[nodeName_(element)] && booleanAttr;
 }
 
-function getAliasedAttrName(element, name) {
-  var nodeName = element.nodeName;
-  return (nodeName === 'INPUT' || nodeName === 'TEXTAREA') && ALIASED_ATTR[name];
+function getAliasedAttrName(name) {
+  return ALIASED_ATTR[name];
 }
 
 forEach({
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1077,7 +1077,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
         var node = this.$$element[0],
             booleanKey = getBooleanAttrName(node, key),
-            aliasedKey = getAliasedAttrName(node, key),
+            aliasedKey = getAliasedAttrName(key),
             observer = key,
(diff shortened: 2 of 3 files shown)


gkalpak:
Doesn't it also affect `ngMin/Max` ? If so, it would be nice to have tests for them as well.
LGTM other than that.

Narretz:
Hmm ... min / ngMin and max / ngMax validators are only added for `input|textarea[number|date]`. So they cannot exist on a custom element.
The whole problem stems from the fact that we made min/maxlength, required, and pattern into separate directives, which are expected to work with ngModel on all elements. 

But before that, we made it so that ngPattern, ngMaxlength, ngMinlength, ngMax and ngMin will set their corresponding non-ng attributes, so that the validator directives only have to observe the non-ng attributes (instead of having another set of directives that watch ngXYZ), see https://github.com/angular/angular.js/commit/d9b90d7c10a8e1bacbee0aeb7e86093cca9e8ed2 and https://github.com/angular/angular.js/commit/25541c1f876a16c892d71faae11727bec7bba98c

It's a bit confusing, but there I can't think of any drawbacks atm. There's just the case that e.g. ngMax on a non-input element will set max, too and make it observable, without actually adding a validator.

gkalpak:
Hm...you are right...weird (that all alliased attributes are directives except for `min`/`max`).
I guess ideally `min`/`max` should be their own directives as well (unless there is some caveat I am missing), so we could make them also work on any element.

But this is not an ideal world, so this LGTM as is :+1:

Narretz:
It seems we have basically split our validators into two groups: generic one's that validate the input / viewValue (which is always a string) and specific one's, such as date / number. min / max really only make sense for date / number, so it'd be hard to make them generic.
And thanks for the review!

PR #12665, opened 2015-08-24
docs($injector): fix inaccuracy in $provide.service docs

### LINKED ISSUE #12664: Inacurracy in $provide.service documentations
From [Documentations](https://docs.angularjs.org/api/auto/service/$provide#service):
> Register a **service constructor**, which will be invoked with `new` to create the service instance. This is short for registering a service where its provider's `$get` property is the service constructor function that will be used to instantiate the service instance.

I think this is technically wrong, because as stated in [Docs](https://docs.angularjs.org/api/auto/service/$provide), "An Angular **service** is a singleton object created by a **service factory**". So `service` method is a shorthand for registering a service where its provider's `$get` property is **a function that returns an instance created by the service constructor**, not the constructor itself.
This is exactly inline with the code for `$provide.service`:
```javascript
function service(name, constructor) {
    return factory(name, ['$injector', function($injector) {
      return $injector.instantiate(constructor);
    }]);
  }
```

alirezamirian:
closes #12664

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -487,8 +487,8 @@ function annotate(fn, strictDi, name) {
  *
  * Register a **service constructor**, which will be invoked with `new` to create the service
  * instance.
- * This is short for registering a service where its provider's `$get` property is the service
- * constructor function that will be used to instantiate the service instance.
+ * This is short for registering a service where its provider's `$get` property is
+ * a function that returns an instance created by the service constructor.
  *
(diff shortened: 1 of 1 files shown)


petebacondarwin:
Although the original text is inaccurate, I don't think that this description is much more helpful. Instead I suggest that we provide some pseudo code to explain the situation.

petebacondarwin:
Actually I left your change and tweaked it with a code example

Issue #12656, opened 2015-08-22
ngAnimate assumes transition is linear instead of using the value in CSS


wesleycho:
There are two scenarios which are not being caught - the easing function in CSS when the transition is specified in the `transition` CSS property, and the easing function if it is specified in the `transition-timing-function` property.

[Here](http://plnkr.co/edit/D4s7PEvBprheJwwAQDSz?p=preview) is a stripped down example of the `transition-timing-function` situation.

[Here is a plnkr](http://plnkr.co/edit/5arAwP5JsCu3igrlwj1l?p=preview), a stripped down example of the `transition` situation.

Note that $animateCss (& $animate as well it turns out) assumes that the easing function is `linear` instead of `ease` here.

This is a regression from 1.3.x

Narretz:
Thanks for reporting @wesleycho! Contrary to many other ngAnimate bug, this looks like it's not too difficult to fix. Do you think you could give a PR a shot?

wesleycho:
Sure, I'll try to put together a PR today

trajano:
Here's a plnkr that you can use http://plnkr.co/edit/rHRFzotMG5pZGeBKF8Lw?p=preview

Narretz:
We had some issues on Travis, so the commit was reverted

PR #12666, opened 2015-08-24
fix($animateCss): use timing functions when present

wesleycho:
- Uses transition timing functions when present in styles instead of defaulting to `linear`

This is for #12656.

I need some guidance here @matsko - the style information is needed from the element on call to $animateCss when doing class manipulation, but I am not sure the best way to cache/retrieve the information here. Would you advise using the existing `computeCachedCssStyles` method here?

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -220,6 +220,7 @@ var DETECT_CSS_PROPERTIES = {
   transitionDuration:      TRANSITION_DURATION_PROP,
   transitionDelay:         TRANSITION_DELAY_PROP,
   transitionProperty:      TRANSITION_PROP + PROPERTY_KEY,
+  transitionTimingFunction: TRANSITION_PROP + TIMING_KEY,
   animationDuration:       ANIMATION_DURATION_PROP,
   animationDelay:          ANIMATION_DELAY_PROP,
   animationIterationCount: ANIMATION_PROP + ANIMATION_ITERATION_COUNT_KEY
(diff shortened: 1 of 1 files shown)


Narretz:
@matsko Could you give this a look?

src/ngAnimate/animateCss.js
```diff
@@ -232,6 +233,8 @@ var DETECT_STAGGER_CSS_PROPERTIES = {
   animationDelay:          ANIMATION_DELAY_PROP
 };
 
+var TRANSITION_EASING_VALUE = /ease|ease-in|ease-out|ease-in-out|linear|step-start|step-end|steps\(.+\)|cubic-bezier\(.+\)/;
```
Narretz: the regex fails when I have something like `transition: 1s ease-in-out color' (It will take the first match, which is `ease`).

src/ngAnimate/animateCss.js
```diff
@@ -437,6 +443,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
 
       var temporaryStyles = [];
       var classes = element.attr('class');
+      var computedStyles = $window.getComputedStyle(element[0]);
```
Narretz: We should probably not call this twice. It's called anyway in the "init" fn.

src/ngAnimate/animateCss.js
```diff
@@ -286,13 +289,16 @@ function truthyTimingValue(val) {
...
+function getCssTransitionStyle(computedStyles, duration, applyOnlyDuration) {
   var style = TRANSITION_PROP;
   var value = duration + 's';
   if (applyOnlyDuration) {
     style += DURATION_KEY;
   } else {
-    value += ' linear all';
+    var transitionStyle = computedStyles[TRANSITION_PROP];
+    var match = TRANSITION_EASING_VALUE.test(transitionStyle);
+    transitionStyle = match ? transitionStyle.match(TRANSITION_EASING_VALUE)[0] : 'linear';
```
gkalpak: Why exec the RegExp twice ? Is there a (noticeable) perf gain ?

Narretz:
Here's another PR to solve this problem: https://github.com/angular/angular.js/pull/13333
It's a bit more work than I expected. Basically, $animateCss would ignore all transition properties already on the element (except duration). I'll check with @matsko tomorrow if I got this right.

wesleycho:
Closing this due to #13333 superceding this

PR #12669, opened 2015-08-25
fix(ngAnimate): use requestAnimationFrame to space out child animations

### LINKED ISSUE #12187: Animation issue on Android 4.1.2.
Hi, 

I'm facing issue regarding angular animations in my Cordova app running on Android 4.1.2. It happen time to time, its hard to reproduce and I don't know what it could be caused by. I have tried both versions of Angular (1.3.16 and 1.4)

Here is stacktrace:

```
Details:Cannot set property 'WebkitTransitionProperty' of undefined: TypeError: Cannot set property 'WebkitTransitionProperty' of undefined
    at blockTransitions (file:///android_asset/www/js/lib/angular-animate.js:1982:70)
    at animateSetup (file:///android_asset/www/js/lib/angular-animate.js:1813:11)
    at animateBefore (file:///android_asset/www/js/lib/angular-animate.js:1990:13)
    at Object.beforeAddClass [as fn] (file:///android_asset/www/js/lib/angular-animate.js:2086:36)
    at file:///android_asset/www/js/lib/angular-animate.js:741:46
    at forEach (file:///android_asset/www/js/lib/angular.js:326:20)
    at run (file:///android_asset/www/js/lib/angular-animate.js:729:11)
    at Object.before (file:///android_asset/www/js/lib/angular-animate.js:770:13)
    at performAnimation (file:///android_asset/www/js/lib/angular-animate.js:1413:16)
    at file:///android_asset/www/js/lib/angular-animate.js:1219:17
```

Dont you have and idea what could be wrong with it? Thanks for help.

### LINKED ISSUE #12594: Animation staggering issue in 1.4.4

### LINKED ISSUE #12612: Enter/leave animation not working in 1.4.4
Hello, I have a library that animates the pop up of notifications and it is not working anymore in 1.4.4

It animates the first notification but not the rest of them (Foxandxss/angular-toastr#121).

I have been able to create a minimal plunker with the idea. The plunker is on 1.4.3 and except the first popup (probably a mistake on my part), all the toasts are being animated.

If you bump angular and angular-animate to 1.4.4 on the plunker, no animation (the ng-enter classes and stuff are not applied).

Plunker doesn't show the .leave animations, but it is the same behavior.

http://plnkr.co/edit/7JlKsjf4XG5aHGaFnuSG?p=info

Whats going on? :)

### LINKED ISSUE #12631: ngHide/ng-show directive don't  animate in ng 1.4.4

### LINKED ISSUE #12655: View transition animations break with single ng-animate-ref 1.4.4 and snapshot
When using a view transition animation and ng-animate-ref, if you have only a single ng-animate-ref it breaks the view transition animation.

Take this lightly modified version of the ng-animate-ref example from the Angular documentation. I have modified it to just show a link to a single record rather than a list.  Notice the page sliding is no longer working properly:
http://plnkr.co/edit/n5WPBEyF04WxAEb2bnFi?p=preview

By adding an empty span with a nonsense ng-animate-ref, the issue goes away:
http://plnkr.co/edit/ua4e00iUajeh0dcu7UF9?p=preview

The issue also resolves by simply reverting to 1.4.3:
http://plnkr.co/edit/Lq9AoWo6Hz3qpCbNie92?p=preview

Updating to snapshot version does not resolve the issue.

matsko:
This reverts the previous behaviour of using foreced reflows to deal
with preparation classes in favour of a system that uses
requestAnimationFrame (RAF).

Closes #12669
Closes #12594
Closes #12655
Closes #12631
Closes #12612
Closes #12187


Narretz:
Does this mean we are going back to the system that was implemented before? Or is this a new approach?

matsko:
This is a new approach that uses a tree data structure to fix the parent class animation blocking.

test/ngAnimate/rafSchedulerSpec.js
```diff
@@ -0,0 +1,142 @@
...
+
+    expect(spies.b).toHaveBeenCalled();
+    expect(spies.y).toHaveBeenCalled();
+
+    $$rAF.flush();
+
+    expect(spies.c).toHaveBeenCalled();
+    expect(spies.z).toHaveBeenCalled();
+
+    function spy(c) {
```
IgorMinar: what's the purpose of the param? :)

test/ngAnimate/rafSchedulerSpec.js
```diff
@@ -0,0 +1,142 @@
...
+      y: spy('y'),
+      z: spy('z')
+    };
+
+    var items = [[spies.a, spies.x],
+                 [spies.b, spies.y],
+                 [spies.c, spies.z]];
+
+    $$rAFScheduler(items);
+    expect(spies.a).toHaveBeenCalled();
```
IgorMinar: can you verify that spies weren't called before you schedule and flush?

test/ngAnimate/rafSchedulerSpec.js
```diff
@@ -0,0 +1,142 @@
...
+
+      expect(log).toEqual(['task1', 'quiet', 'task2', 'task3']);
+
+      $$rAFScheduler.waitUntilQuiet(quietFn);
+
+      $$rAF.flush();
+
+      expect(log).toEqual(['task1', 'quiet', 'task2', 'task3', 'quiet', 'task4']);
+
+      function logFactory(token) {
```
IgorMinar: you can add `module(provideLog)` and then `inject(log)` to get logging

see: [provideLog](https://github.com/angular/angular.js/blob/2f6b6fb7a1dee0ff97c5d2959b927347eeda6e8b/test/helpers/testabilityPatch.js#L289-L326), [usage](https://github.com/angular/angular.js/blob/291d7c467fba51a9cb89cbeee62202d51fe64b09/test/ngRoute/routeSpec.js#L82)

IgorMinar:
the rest looks good

PR #12670, opened 2015-08-25
docs: cleanup of directive animation classes 

brettdewoody:


brettdewoody:
I signed it!

brettdewoody:
Being my first PR on this repo I'm not sure how to proceed. Do I need to fix my PR?

Narretz:
I don't think the changes are quite correct. `add`, `enter` etc. don't mean the classes that are added, these are the actual events that are fired by the animation code.

brettdewoody:
Do you think it would be helpful to add an examples of the classes added using for each directive?

Narretz:
We had a great overview over the phases until the 1.3.x docs: https://code.angularjs.org/1.3.18/docs/api/ngAnimate/service/$animate We should get those back in.some way.

PR #12672, opened 2015-08-25
feat($rootScope): implement event.stopPropagation() for $broadcast-ed events

just-boris:
It would more consistent if both ways to trigger an event will have `stopPropagation` feature.

Implemented as described in https://github.com/angular/angular.js/pull/914

Also there is a related question on [stackoverlow](http://stackoverflow.com/questions/15420765/how-to-stop-broadcast-events-in-angularjs/15422749#15422749).

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -1099,7 +1099,7 @@ function $RootScopeProvider() {
        *     event propagates through the scope hierarchy, this property is set to null.
        *   - `name` - `{string}`: name of the event.
        *   - `stopPropagation` - `{function=}`: calling `stopPropagation` function will cancel
-       *     further event propagation (available only for events that were `$emit`-ed).
+       *     further event propagation.
        *   - `preventDefault` - `{function}`: calling `preventDefault` sets `defaultPrevented` flag
        *     to true.
        *   - `defaultPrevented` - `{boolean}`: true if `preventDefault` was called.
@@ -1232,9 +1232,11 @@ function $RootScopeProvider() {
         var target = this,
             current = target,
             next = target,
+            stopPropagation = false,
             event = {
               name: name,
               targetScope: target,
+              stopPropagation: function() {stopPropagation = true;},
               preventDefault: function() {
                 event.defaultPrevented = true;
               },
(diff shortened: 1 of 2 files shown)


just-boris:
I signed it as just-boris@hotmail.com

just-boris:
I signed it!

src/ng/rootScope.js
```diff
@@ -1270,8 +1272,8 @@ function $RootScopeProvider() {
           // yes, this code is a bit crazy, but it works and we have tests to prove it!
           // this piece should be kept in sync with the traversal in $digest
           // (though it differs due to having the extra check for $$listenerCount)
-          if (!(next = ((current.$$listenerCount[name] && current.$$childHead) ||
-              (current !== target && current.$$nextSibling)))) {
+          if (!(next = ((current.$$listenerCount[name] && !stopPropagation && current.$$childHead) ||
+              (current !== target && ((stopPropagation = false) || current.$$nextSibling))))) {
```
just-boris: Maybe it should split that complicated expression into more `if`s?
But then it will be completely different than traversal in `$digest`

Narretz:
I am not sure what the reasons were for not having stopPropagation on broadcast. However, what's the use case for using stopPropagation this way? Normally, you use events pretty sparingly in an angular app.

test/ng/rootScopeSpec.js
```diff
@@ -2007,6 +2007,20 @@ describe('Scope', function() {
...
 
 
+        it('should allow stopping event propagation', inject(function($rootScope) {
+          child2.$on('myEvent', function(event) { event.stopPropagation(); });
+          $rootScope.$broadcast('myEvent');
+          expect(log).toEqual('0>1>11>2>3>');
+        }));
+
+
+        iit('should stop propagation only for one scope chain', inject(function($rootScope) {
```
Narretz: iit -> it

just-boris:
@Narretz fixed. Now all tests are passing fine.

The reasons are the same that stoping `$emit`-ed events. If some component on top level already handled this event (showed popover, for example), then children don't need to handle this event again. Now as a workaround I have to use some flags and check that on children and code looks ugly.

I guess that wasn't implemented earlier, because it is more rare case than `$emit` and harder to implement (you need to stop only one branch in the tree and go on by others).

PR #12682, opened 2015-08-26
Add support for multiple ignored directives

skaldo:
Sometimes it is useful to have a possibility to ignore more than one directive.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1372,7 +1372,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             : compileNodes(childNodes,
                  nodeLinkFn ? (
                   (nodeLinkFn.transcludeOnThisElement || !nodeLinkFn.templateOnThisElement)
-                     && nodeLinkFn.transclude) : transcludeFn);
+                     && nodeLinkFn.transclude) : transcludeFn, undefined, undefined, ignoreDirective);
 
         if (nodeLinkFn || childLinkFn) {
           linkFns.push(i, nodeLinkFn, childLinkFn);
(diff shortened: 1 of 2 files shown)


Narretz:
Hi, I have a few questions. What exactly does this change and why do we need it? Also, tests are failing and the change itself must be tested.

skaldo:
Hi, this change allows us to pass an array containing the names of the directives, we don't want to compile for some reason, to the $compile function.
Regarding tests, I've run them on my fork and the last two have passed, the 1st and 2nd failed because of timeouts..

Narretz:
Travis says there are some failures in the e2e tests. They don't look like they are timeouts / flakes. Also, we still need a test for your change.

src/ng/compile.js
```diff
@@ -2104,6 +2104,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     function addDirective(tDirectives, name, location, maxPriority, ignoreDirective, startAttrName,
                           endAttrName) {
       if (name === ignoreDirective) return null;
+      if (angular.isArray(ignoreDirective)) {
+        for (var i = 0; i < ignoreDirective.length; i++) {
+          if (name === ignoreDirective[i]) return null;
+        }
+      }
```
gkalpak: This whole block can be reduced to:

```js
if (isArray(ignoreDirective) && includes(ignoreDirective, name)) return null;
```

gkalpak:
I am a bit sceptical, because this is undocumented (private) API, but this change will not be serving the needs of Angular core.

skaldo:
Well, in our application this change is really beneficial. On the other hand, if it's going to be useless for the core and if it's big job to make it working, let's close this request.

Anyway, what's failing the tests is the recursive call of compileNodes function with the ignoreDirective parameter.

caitp:
you know, this probably could be useful for internal purposes, like helping people build recursive tree structured directives easily using the directive API. It would need a bit more than that, but I can see how you could build that kind of API on top of it.

gkalpak:
As pointed out by @Narretz, this is essentially the same functionality as #14635. Although this also has test coverage, #14635's approach is more powerful/flexible, so I'm going to close this in favor of #14635.
<sub>(I doubt we are likely to implement this feature, but if we would #14635 would be the way to go.)</sub>

PR #12683, opened 2015-08-26
test($parse): fix csp setup

lgalfaso:
Run the parse interpreter for csp enabled tests

### DIFF
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -1681,8 +1681,9 @@ describe('parser', function() {
 
       beforeEach(module(function($provide) {
         $provide.decorator('$sniffer', function($delegate) {
-          $delegate.csp = cspEnabled;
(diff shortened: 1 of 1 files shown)


jbedard:
Is there a way to actually test this? I think this has happened before...

lgalfaso:
I can add an internal flag that express that the function is compiled or
interpreted and add a test. Slightly an overkill, but should work
>
>

jbedard:
I'm not sure if it's worth adding a flag. What about just asserting the flag we are changing is already a boolean?

lgalfaso:
@jbedard mmm... ok updated the PR

jbedard:
LGTM then, hopefully that stops it from happening again...

lgalfaso:
landed as ea2518fceac22f2ee2a9ce688e9608f776c207f1

PR #12687, opened 2015-08-26
docs(guide/Forms): replace `form` with `user`

resalisbury:
Referring to the `user` as `form` in the previews is confusing, since it makes it seem as though the data being displayed is attached to the `form` object, when the `form` object is separate, albeit controlling whether data is synced to the `user` object.


resalisbury:
I signed it!

PR #12689, opened 2015-08-27
fix($angular): use strict equality comparison

JungMinu:
There is no type-conversion to be done because both parameters are already the same type. 
Therefore, the === operator should be used for better performance.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -198,7 +198,7 @@ msie = document.documentMode;
  *                   String ...)
  */
 function isArrayLike(obj) {
-  if (obj == null || isWindow(obj)) {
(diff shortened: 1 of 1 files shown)


JungMinu:
I signed it!

gkalpak:
`obj == NULL` is equivalent to `(obj === undefined || obj === null)`, so this change breaks the semantics of the statement.

We should either expand it to the more verbose form or leave as is.

Condidering the current form is more concise and is used throughout the codebade, I'd vote for leaving it as is (unless we decide to change it everywhere).

lgalfaso:
Thanks for the PR, but @gkalpak is spot on and also think that leaving it as-is is best

PR #12696, opened 2015-08-27
fix($animate): `$animate.enabled(false)` should disable animations on $animateCss as well

### LINKED ISSUE #12685: $animateCss can't disable animations.
It would appear that when I call $animate.enabled(false) the code in $animateCss doesn't seem to be affected by it, meaning that it will add the animation classes regardless of the status of animationsEnabled.

Is this intended? If so, how do I disable animations completely?

matsko:
Closes #12685


IgorMinar:
lgtm

PR #12701, opened 2015-08-28
fix issue 12699: Context hierachy is not destroyed within a destroyed…

yoann-antoviaque:
… context, making memory leaks worse

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -865,13 +865,6 @@ function $RootScopeProvider() {
       $destroy: function() {
         // we can't destroy the root scope or a scope that has been already destroyed
         if (this.$$destroyed) return;
-        var parent = this.$parent;
-
-        // let's help garbage collecting
-        while (this.$$childHead) {
-        	this.$$childHead.$destroy();
-        }
-		
         this.$broadcast('$destroy');
         this.$$destroyed = true;
         if (this === $rootScope) return;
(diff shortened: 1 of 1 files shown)


yoann-antoviaque:
I signed it!

lgalfaso:
Would it be possible to see something that shows that this makes the GC any better?

yoann-antoviaque:
I refactored the code to prevent changing the order '$destroyed' is broadcasted and clarify what is done: just remove links between each scope of the destroyed tree, preventing leaking the whole tree when someone keeps a reference on a child.

this helps in the application i am working on which has some memory leaks:
- each leak retains less memory
- it is easier to find the leak origin in a memory dump as there is no more references between destroyed contexts and therefore less retainer objects to go through

jbedard:
Note that this could hurt performance, traversing an entire scope tree just in case an app has a memory leak...

yoann-antoviaque:
Well, if i'm not mistaken we already traverse the whole tree when broadcasting $destroyed without performance issue.
It does not add complexity (while it's true that we add a small overhead on destroy).

jbedard:
Event broadcasting only traverses branches which have a listener for that event: https://github.com/angular/angular.js/blob/681affef591273615e4505811eec663d0c0d1bb4/src/ng/rootScope.js#L1273

yoann-antoviaque:
The pointed line is difficult to understand (thanks for the "insanity warning" ;)) and I will trust you on this.
There is added complexity but I have not found it problematic in my relatively complex case

dcherman:
Copying a comment from the issue:

There should be no logical changes here. In both cases, you're severing the path to the GC roots, so the scope and all descendants should immediately become eligible for GC ( assuming nothing leaked in descendant scopes ).

IMO, it seems like the solution here is to not have memory leaks in your code, or fix them if they're present.

Narretz:
Based on the lack of a good test for this, and the skeptical comments here and the comments in the issue, I'm going to close this. A change that touches such hot code should be really bulletproof, and this it is not.

PR #12705, opened 2015-08-30
docs(CHANGELOG): Add breaking change notice for issue 12506

lgalfaso:


CHANGELOG.md
```diff
@@ -257,6 +257,38 @@ and/or `beforeRemoveClass` then the CSS classes will not be applied
 in time for the children (and the parent class-based animation will not
 be cancelled by any child animations).
 
+- **$q** due to [6838c979](https://github.com/angular/angular.js/commit/6838c979451c109d959a15035177ccee715ccf19),
+  When writting tests, there is no need to call `$timeout.flush()` to resolve a call to `$q.when` with a value.
+
+The previous behavior involve creating an extra promise that needed to be resolved. This is no longer needed when
```
gkalpak: _involve_  -->  _involved_

CHANGELOG.md
```diff
@@ -257,6 +257,38 @@ and/or `beforeRemoveClass` then the CSS classes will not be applied
 in time for the children (and the parent class-based animation will not
 be cancelled by any child animations).
 
+- **$q** due to [6838c979](https://github.com/angular/angular.js/commit/6838c979451c109d959a15035177ccee715ccf19),
+  When writting tests, there is no need to call `$timeout.flush()` to resolve a call to `$q.when` with a value.
```
gkalpak: _writting_  -->  _writing_

CHANGELOG.md
lgalfaso: done

CHANGELOG.md
lgalfaso: done

IgorMinar:
hmm... we didn't realize that this would break tests when we merged the original PR.

I think it's too late to revert the breaking commit, so this PR lgtm.

@petebacondarwin do you agree?

lgalfaso:
> hmm... we didn't realize that this would break tests when we merged the
> original PR.
>
Keep in mind that this is only for test that create a promise using `when`
using a value and then doing nothing (not calling `then` nor using it to
resolve other promise)

IgorMinar:
yes, I understand. still we should avoid any breaking changes in stable
releases in the future.

PR #12708, opened 2015-08-30
fix($parse): `assign` returns the new value

### LINKED ISSUE #12675: Angular 1.4.4, $parse, assign function does not return the value set
Hi guys,

I was using angular 1.3.15 and I had this code:
```javascript
function getDataRoot(data, expression){
      var getter = $parse(expression);
      
      return getter(data) || getter.assign(data, {});
}

var data = {details: {orderId: ''}};
console.log(getDataRoot(data, 'details')); // should return: {orderId: ''}
 console.log(getDataRoot(data, 'address')); // should return a new object {}
```
The function above always returned a value either an existing or a new one object, because getter.assign always returned the value set, now I have upgraded to angular 1.4.4 and getter.assign no longer return anything.

Here is an [example](http://plnkr.co/edit/OQq0FLonOhr3ZvzOVBEH?p=preview), just change the versions and you will see the issue.   

Cheers,
Fabián

lgalfaso:
The `.assign` function returns the new value.
The version with csp enabled already has this behavior.

Closes #12675

### DIFF
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -2797,6 +2797,14 @@ describe('parser', function() {
           expect(scope).toEqual({a:123});
         }));
 
+        it('should return the assigned value', inject(function($parse) {
+          var fn = $parse('a');
+          var scope = {};
+          expect(fn.assign(scope, 123)).toBe(123);
+          var someObject = {};
+          expect(fn.assign(scope, someObject)).toBe(someObject);
+        }));
+
         it('should expose working assignment function for expressions ending with brackets', inject(function($parse) {
           var fn = $parse('a.b["c"]');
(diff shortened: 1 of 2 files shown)


test/ng/parseSpec.js
```diff
@@ -2793,7 +2793,7 @@ describe('parser', function() {
           var fn = $parse('a');
           expect(fn.assign).toBeTruthy();
           var scope = {};
-          fn.assign(scope, 123);
+          expect(fn.assign(scope, 123)).toBe(123);
```
Narretz: If we are super nitpicky, this is out of scope for this test (i.e. it should be its own it). But it's no dealbreaker

Narretz:
LGTM, expect for that nitpick

PR #12709, opened 2015-08-30
Fixing problems with large numbers and formatNumber

### LINKED ISSUE #8674: number formatter outputs NaN.0 for large numbers with decimals
http://plnkr.co/edit/bB1m5UOpLJJUIlHOYKAO?p=preview

For very large numbers when converting to a string the browser might use the "#e#" format which breaks the use of appending `"e"+fractionSize` and produces NaN.

The MDN page that `formatNumber` references appears to handle this by doing a `.split('e')` before generating the "#e#" string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

### LINKED ISSUE #12707: formatNumber produces NaN for numbers near 999999999999999934463
At the moment, the `formatNumber` method does some [hand built rounding](https://github.com/angular/angular.js/blob/master/src/ng/filter/filters.js#L180) for decimal numbers, in order to not lose precision.

It uses the `e` (scientific) notation (e.g. 1000e2) to shift the number and remove unnecessary precision, before rounding it and then doing the inverse `100000e-2` to return the number to it's original state.

If the number is already in scientific notation then it is [diverted away beforehand](https://github.com/angular/angular.js/blob/master/src/ng/filter/filters.js#L169). However, if the number is not in scientific notation, but scaling it up with `'e' + fractionSize` will take it above the browser's limit (999999999999999934463) then the string ends up with two `e`s. 

For example, take 444444444400000000000 and lets say we want to round it to 2 points of precision.

```
+(444444444400000000000 + 'e' + 2)
// becomes scientific notation
4.444444444e+22

// then we round it
Math.round(4.444444444e+22)
4.444444444e+22

// then we scale it back
4.444444444e+22 + 'e' + -2
"4.444444444e+22e-2"

// this version has two exponents and can't be coerced with +
+("4.444444444e+22e-2")
NaN
```

I think this issue may have already been opened as #8674, but the conditions for reproducing are a lot less clear in that issue.

danprince:
Use a simple check to prevent the decimal rounding code doesn't try to round integers. This prevents formatNumber from returning NaN for numbers that fit the criteria described in #12707.

Tests show it working with numbers lower and upper limit numbers as well as the example from the above issue.

Fixes #12707
Fixes #8674

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -174,10 +174,13 @@ function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
       fractionSize = Math.min(Math.max(pattern.minFrac, fractionLen), pattern.maxFrac);
     }
 
-    // safely round numbers in JS without hitting imprecisions of floating-point arithmetics
-    // inspired by:
-    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round
-    number = +(Math.round(+(number.toString() + 'e' + fractionSize)).toString() + 'e' + -fractionSize);
+    // round the number only if it's decimal
+    if(number.toString().indexOf('.') >= 0) {
+      // safely round numbers in JS without hitting imprecisions of floating-point arithmetics
+      // inspired by:
+      // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round
+      number = +(Math.round(+(number.toString() + 'e' + fractionSize)).toString() + 'e' + -fractionSize);
+    }
 
     var fraction = ('' + number).split(DECIMAL_SEP);
     var whole = fraction[0];
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -175,7 +175,7 @@ function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
     }
 
     // round the number only if it's decimal
-    if(number.toString().indexOf('.') >= 0) {
+    if (number.toString().indexOf('.') >= 0) {
       // safely round numbers in JS without hitting imprecisions of floating-point arithmetics
       // inspired by:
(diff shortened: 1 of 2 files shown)


lgalfaso:
It looks like this does not solve the problem entirely. eg.
```js
expect(formatNumber(999999999999999.9, pattern, ',', '.', 7))
    .toBe('999,999,999,999,999.9000000');
    // Returns 'NaN.0000000'
```

lgalfaso:
Wondering if changing
```js
if (number.toString().indexOf('.') >= 0) {
```

for
```js
if (number.toString().indexOf('.') >= 0 && number.toString().split('.')[1].length > fractionSize) {
```

should take care of all the issues

danprince:
Wouldn't this end up as a tradeoff between fixing those NaNs but also directing some decimal numbers away from the safe decimal rounding function?

petebacondarwin:
It looks like #8705 had a different way of solving this, which also tests corner cases from different VMs. @danprince can you take a look at the tests there - and the implementation of the fix.

petebacondarwin:
@lgalfaso can we give stuff like this a milestone if you are commenting on it (along with its related issue), so that it doesn't get triaged multiple times. Thanks

danprince:
I would suggest that this is a solution to a much smaller part of the problem that #8705 is try to address. However this is definitely a short term solution and it makes the edge cases more complex to understand. 

If the ideal outcome would be to solve the problem to the depth of the idiosyncrasies of specific VM implementations, then you only have to look at the size of [decimal.js](https://github.com/MikeMcl/decimal.js/blob/master/decimal.js) to see that this is a non-trivial thing to hack into the format number function.

Maybe a more elegant short term solution is to add a disclaimer to the docs for currency (and anything else that relies on formatNumber) that explains these edge cases? Or at the very least reminds users of the restrictions regarding precision with 64-bit floating point numbers?

Narretz:
We could also make it configurable via service etc. which function is used to format / process the number. Then people could include decimal.js or whatever if they need it. But it's probably difficult to match the APIs.

PR #12715, opened 2015-08-31
Troubleshooting instructions for step 7

tyrasd:
This shows troubleshooting instructions as in [step 11](https://github.com/angular/angular.js/blob/master/docs/content/tutorial/step_11.ngdoc#L49) of the tutorial.


tyrasd:
I signed it!

PR #12716, opened 2015-08-31
docs($animate): Consistent naming

evermax:
In the description of the example, you use `element` to refer to the container parameter and `listenerFn` to refer to the callback parameter.


evermax:
I signed it!

PR #12721, opened 2015-08-31
fix(ngAnimate): allow animations if the body node is changed

### LINKED ISSUE #12366: ngIf on body stop animation
When i use a condition (in my case `ngIf`) in the `body`, all animation using `ngAnimate` dos not work anymore.

reproduced:
http://codepen.io/anon/pen/MwByOL

I tried using a old version (1.3.15) and everything back to normal.

matsko:
If an ngIf is placed on the body then resulting node is different from
what was there originally and that causes issues with ngAnimate. This
patch fixes this.

Closes #12366


petebacondarwin:
The fix LGTM but really it should have some kind of test...

matsko:
This is closed. We got rid of the `$$body` service.

PR #12722, opened 2015-08-31
chore(scripts/publish): get dist-tag from package.json

petebacondarwin:
This change would need to be cherry-picked to the previous version branches (with suitable change to package.json) to prevent releases from them breaking npm dist-tags.

Where does this need to be documented? New big version release tasks? Contributor documentation?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "license": "MIT",
   "branchVersion": "^1.4.0-beta.0",
   "branchPattern": "1.4.*",
+  "distTag": "latest",
   "repository": {
     "type": "git",
     "url": "https://github.com/angular/angular.js.git"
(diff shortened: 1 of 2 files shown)


petebacondarwin:
The branch to dist-tag mapping will be as follows (once we fork and start working on 1.5.0):

* 1.2.x -> ie8_compat
* 1.3.x -> previous_1_3
* 1.4.x -> latest
* master -> beta

Narretz:
I think it's okay, although it's a bit cumbersome to update the branches.
Do we have access to our git in this step? Then we could check if there's a larger version number, and not set it in this case.

petebacondarwin:
Going forward it will be quite straightforward. You only ever have to change the branch that is going from `latest` to `previous...` and then master swaps between `latest` and `beta`. We need to specify some dist-tag when publishing or it defaults to `latest`

IgorMinar:
nice! lgtm

PR #12725, opened 2015-09-01
fix($animate): invalid CSS class names should not break subsequent elements

### LINKED ISSUE #12674: ng-show is not working when ng-class is having array expression
ng-show is not working (angularjs version 1.4.4) when ng-class is having array expression

Here is the plnkr to demonstrate the issue. 

http://plnkr.co/edit/qNTsC8sWS7b1u76zvfbO?p=preview

petebacondarwin:
The postDigest handler was not being added if the first element in
to modify the CSS classes contained invalid CSS class names. This meant
that subsequent valid CSS class changes were not being handled since we
were not then adding the handler for those correct cases.

Closes #12674

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -105,6 +105,51 @@ var $$CoreAnimateQueueProvider = function() {
       }
     };
 
+
+    function updateData(data, classes, value) {
+      var changed = false;
+      if (classes) {
+        classes = isString(classes) ? classes.split(' ') :
+                  isArray(classes) ? classes : [];
+        forEach(classes, function(className) {
+          if (className) {
+            changed = true;
+            data[className] = value;
+          }
+        });
+      }
+      return changed;
+    }
+
+    function handleCSSClassChanges() {
+      forEach(postDigestElements, function(element) {
+        var data = postDigestQueue.get(element);
+        if (data) {
+          var existing = splitClasses(element.attr('class'));
+          var toAdd = '';
+          var toRemove = '';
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Here is the Plunker showing the issue with the fixed version of Angular: http://plnkr.co/edit/Zux7ICpKvsh6oa25vlzr?p=preview

petebacondarwin:
The diff looks more complicated that it is. I pulled two functions, `updateData` and `handleClassChanges`, from out of the `addRemoveClassesPostDigest` function. This means that we are not having to recreate them on every call and also makes the `addRemoveClassesPostDigest` easier to grok.

Narretz:
Without the refactor, it would be easier to see what actually changed semantically, though ;)

matsko:
LGTM. Excellent work.

petebacondarwin:
@Narrate you're right. I should have made two commits

petebacondarwin:
Here is the plunker with the most recent fix: http://plnkr.co/edit/ppS7j623Wz4DifRRBjtu?p=preview

petebacondarwin:
@Narretz - there I have broken the PR into two commits.

Narretz:
@petebacondarwin I almost down dare to say it, but ... it's actually `refactor`, not `refact`.

petebacondarwin:
Doh!

petebacondarwin:
@Narretz shouldn't you be asleep?

PR #12727, opened 2015-09-01
Adjusted example

MrPeanuts:
example was using disabled which always disable the button, correct should be ng-disabled


Narretz:
No, this is actually correct - this is a wrong example, to show you how not to do it.

PR #12729, opened 2015-09-01
<feat>($location): add post parse app url hook

### LINKED ISSUE #8905: AngularJS Routing Fails when running within a Juniper SSL VPN
I write corporate web apps that need to be accessible via our SSL VPN. In particular we have a Juniper appliance, and it handles secure proxying of the app. The issue is that it adds a bunch of extra stuff to the URL and this causes Angular routes to fail since the extra stuff is random in nature (session ids, etc). I've found the easiest way to deal with this is to patch the `parseAppUrl` function in angular.js. 

This doesn't make sense to include in the core Angular library since it's Juniper SSL VPN specific, but I wanted to post it for anyone struggling with similar URL re-write issues from external appliances. Though perhaps a better fix would be to provide a hook into the `parseAppUrl` method, to allow a user to configure some external processing of the `locationObj` prior to `parseAppUrl` returning.

``` javascript
function parseAppUrl(relativeUrl, locationObj, appBase) {
  var prefixed = (relativeUrl.charAt(0) !== '/');
  if (prefixed) {
    relativeUrl = '/' + relativeUrl;
  }
  var match = urlResolve(relativeUrl, appBase);
  locationObj.$$path = decodeURIComponent(prefixed && match.pathname.charAt(0) === '/' ?
      match.pathname.substring(1) : match.pathname);
  locationObj.$$search = parseKeyValue(match.search);
  locationObj.$$hash = decodeURIComponent(match.hash);

  //Detect Juniper re-write functions and handle the $$path issue
  if(locationObj.$$path === "[object Object]" && typeof(DanaOrigUrl) === 'function') {
    var __strH = 'href';
    var __tmpHack = match[__strH];
    var __nn = ("" + __tmpHack).match(/^(https?:\/\/[^\/]+)?([^?|#]*)/);
    locationObj.$$path = __nn[2];
  }
  // make sure path starts with '/';
  if (locationObj.$$path && locationObj.$$path.charAt(0) != '/') {
    locationObj.$$path = '/' + locationObj.$$path;
  }
}
```

mikepugh:
Add a hook into the end of the $location service's parse app url method. The hook allows the locationObj to be repaired in the event it gets
damaged by some external source (such as an SSL VPN appliance).

Closes #8905

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -1,8 +1,9 @@
 'use strict';
 
 var PATH_MATCH = /^([^\?#]*)(\?([^#]*))?(#(.*))?$/,
-    DEFAULT_PORTS = {'http': 80, 'https': 443, 'ftp': 21};
+  DEFAULT_PORTS = {'http': 80, 'https': 443, 'ftp': 21};
 var $locationMinErr = minErr('$location');
+
 
 
 /**
@@ -13,7 +14,7 @@ var $locationMinErr = minErr('$location');
  */
 function encodePath(path) {
   var segments = path.split('/'),
-      i = segments.length;
+    i = segments.length;
 
   while (i--) {
     segments[i] = encodeUriSegment(segments[i]);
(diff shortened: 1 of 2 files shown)


mikepugh:
Looks like my WebStorm IDE auto fixed a bunch of spacing discrepancies based upon the .editorconfig. I'm going to pull that out so that this commit is more focused on what I actually changed, as well as fix the <feat> in the commit message.

PR #12738, opened 2015-09-02
A new directive for the focusout event.

drakir:
I could not find a built in directive for the focusout event, so here it is.

The problem I was facing was to close a "layer" (typically a small popup) when clicking outside of the directive's markup. The idea I had and initially solved it with, was to use ngBlur and check the $event.relatedTarget to see if it was known to the scope of the directive's html elements or not.
However when using Internet Explorer the relatedTarget does not seem to be set. So the closest thing to the blur event is a focusout event, which solved this particular problem for me.

Please let me know if you have any other solution to this problem, otherwise I believe a focusout directive in AngularJs wouldn't hurt :-)

### DIFF
--- a/src/ng/directive/ngEventDirs.js
+++ b/src/ng/directive/ngEventDirs.js
@@ -43,10 +43,11 @@ var ngEventDirectives = {};
 // so that they are not executed in an inconsistent state.
 var forceAsyncEvents = {
   'blur': true,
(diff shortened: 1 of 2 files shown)


Narretz:
Closing. Two issues:
1) no CLA
2) The event is still not supported in Firefox

PR #12745, opened 2015-09-03
fix($httpBackend): send `null` when post-data is undefined

### LINKED ISSUE #12141: $http.delete causes non-zero Content-Length header in IE10 due to passing undefined to xhr.send
Since 1.4.0-rc1 and confirmed in 1.4.1, calling `$http.delete('url')` in IE10 causes a non-zero Content-Length header to be added to the request. I noticed this when the webserver was returning HTTP error code 415 on all `$http.delete` calls.

I tracked this down to 1.4.0-rc1 where `xhr.send(post || null)` in createHttpBackend was changed to `xhr.send(post)`. When post is undefined, this problem occurs in IE10. The code below reproduces this bug in IE10 and is essentially what `$http.delete('url')` is doing.

    var xhr= new window.XMLHttpRequest();
    var post;
    xhr.open('DELETE', 'url', true);
    xhr.send(post);
Changing `xhr.send(post)` to `xhr.send(post || null)` fixes the problem.

### LINKED ISSUE #12739: Bad content-length for $resource remove in IE 11, angularjs 1.4.5
See plunk at http://plnkr.co/edit/1Pf5wTbjdMqmVQCkJ8il?p=preview.

The angular version is initially 1.4.3 (same problem exists in 1.4.5).  In IE 11, when you click the "Test Delete" button and examine the request, notice that the Content-Length is 9 (which corresponds to the text 'undefined').

If you change the version to angular 1.3.18 and repeat, Content-Length is 0, which is correct.

The content-length of 9 with the value of 'undefined' causes the DELETE request to fail at the REST server.

This issue only exists in IE, I'm using version 11.

gkalpak:
IE11 (and maybe others) converts an `undefined` argument to `xhr.send()` to string (`'undefined'`) for certain request methods (e.g. DELETE). This causes the request to appear having a body, when it shouldn't.

Fixes #12141
Fixes #12739

### DIFF
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -109,7 +109,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
         }
       }
 
-      xhr.send(post);
+      xhr.send(isUndefined(post) ? null : post);
     }
 
     if (timeout > 0) {
--- a/test/ng/httpBackendSpec.js
+++ b/test/ng/httpBackendSpec.js
@@ -44,21 +44,23 @@ describe('$httpBackend', function() {
   });
 
   it('should pass null to send if no body is set', function() {
-    $backend('GET', '/some-url', null, noop);
+    $backend('GET', '/some-url', undefined, noop);
     xhr = MockXhr.$$lastInstance;
 
     expect(xhr.$$data).toBe(null);
   });
 
-  it('should pass the correct falsy value to send if falsy body is set (excluding NaN)', function() {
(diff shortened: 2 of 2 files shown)


pkozlowski-opensource:
This should also fix https://github.com/angular/angular.js/issues/12141

gkalpak:
Thx @pkozlowski-opensource - added that to the commit message.

pkozlowski-opensource:
LGTM

gkalpak:
Merged as 6f39f10.

PR #12746, opened 2015-09-03
fix(ngJq): properly detect when `ng-jq` is empty

### LINKED ISSUE #12741: ngJq causes warning in Firefox
If ngJq is used without a value passed to it Firefox warns "Empty string passed to getElementById()."
Tested in Firefox 40.0.3.

http://jsbin.com/kimatejiqa/1/edit?html,js,output

gkalpak:
Previously, even when `ng-jq` was empty (which should force the use of jqLite), Angular tried to find jQuery on `window['']`. If it didn't find anything there, it would fall back to jqLite (as expected).

Nonetheless, trying to access `window['']` calls `getElementById('')`, which issues a warning in Firefox (maybe others).

This fix properly detects when `ng-jq` is empty and avoids trying to access `window['']`.

Fixes #12741

### DIFF
--- a/test/e2e/fixtures/ngJq/index.html
+++ b/test/e2e/fixtures/ngJq/index.html
@@ -4,6 +4,10 @@
     {{jqueryVersion}}
 
     <script src="../../../../bower_components/jquery/dist/jquery.js"></script>
+    <script type="text/javascript">
+      // Trying to trick `bindJQuery()`
+      window[''] = window.jQuery;
+    </script>
     <script src="angular.js"></script>
     <script type="text/javascript" src="script.js"></script>
   </body>
--- a/test/e2e/fixtures/ngJq/index.html
+++ b/test/e2e/fixtures/ngJq/index.html
@@ -5,7 +5,8 @@
 
     <script src="../../../../bower_components/jquery/dist/jquery.js"></script>
     <script type="text/javascript">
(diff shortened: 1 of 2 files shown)


test/e2e/fixtures/ngJq/index.html
```diff
@@ -4,6 +4,10 @@
     {{jqueryVersion}}
 
     <script src="../../../../bower_components/jquery/dist/jquery.js"></script>
+    <script type="text/javascript">
+      // Trying to trick `bindJQuery()`
```
Narretz: I think this should say more explicit what we are trying to do.

gkalpak: @Narrezt, I didn't want to be too explicit, since this is basically fixing a bug and it wouldn't be obvious what is does by looking at the corrected code.

Do you think I should add a reference to the issue or the PR (that explains the issue) ?
Or just add a brief description of what the issue was (and how this tests it's fixed) inline ?
<sub>I'm fine with both.</sub>

Narretz: Link to issue is fine!

gkalpak: Done.

gkalpak:
Merged as 19ecdb5.

PR #12762, opened 2015-09-04
update angular-mocks.js url matching

deodeveloper:
url matching was incorrect and updated. if all query parameters in url are in different order, it should still match

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1628,7 +1628,7 @@ function MockHttpExpectation(method, url, data, headers) {
     if (!url) return true;
     if (angular.isFunction(url.test)) return url.test(u);
     if (angular.isFunction(url)) return url(u);
-    return url == u;
+    return (url == u || this.compareUrl(u));
   };
 
   this.matchHeaders = function(h) {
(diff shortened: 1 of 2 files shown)


deodeveloper:
I signed it!

petebacondarwin:
Thanks for this @deodeveloper - before we can land this change it would need to have a unit test to show that it is doing what you expect.

deodeveloper:
For merging with master, updated a test case for the changes related to this the pull request as commented by @petebacondarwin

gkalpak:
There are some stying issues (e.g. see https://travis-ci.org/angular/angular.js/jobs/106143760).
You can also run the ci-checks locally (`grunt ci-checks`) - I believe there will be some more (e.g. whitespace between function parameters) once you fix the semicolon issues :wink:

src/ngMock/angular-mocks.js
```diff
@@ -1646,6 +1646,15 @@ function MockHttpExpectation(method, url, data, headers) {
     }
     return data == d;
   };
+  
+  this.getUrlParams = function (u){
```
gkalpak: Do these two methods (`getUrlParams` and `compareUrl`) need to be public ?

gkalpak:
This doesn't properly support URLs with a hash fragment, but I guess we don't really care, because it doesn't make sense to make requests against URLs with a hash fragment (since hash fragments have no meaning for the server).
(In any case, this PR doesn't make them any less supported than they already were, I just mention it as a possible (probably unnecessary) enhancement.)

petebacondarwin:
@deodeveloper can you address the issues that @gkalpak has highlighted?

PR #12765, opened 2015-09-05
fix($Angular): fix length to define constants

JungMinu:
length var is constant.
Therefore, this commit replaces length var statements with const statements


gkalpak:
`const` is not supported by all the currently supported browsers.

lgalfaso:
Agree with https://github.com/angular/angular.js/pull/12765#issuecomment-138054182, unfortunately this cannot be merged

PR #12766, opened 2015-09-05
fix($interpolate): use const to define constants

JungMinu:
startSymbolLength,endSymbolLength are constants.
Therefore, this commit replaces length var statements with const statements


lgalfaso:
Const expressions are not supported by all the currently supported browsers

PR #12777, opened 2015-09-07
fix(ngMock/$httpBackend): fail if a url is provided but is `undefined`

### LINKED ISSUE #8442: $httpBackend expectations with undefined URL
Using : Angular-1.3.0-beta.17

$httpBackend.expect methods return true when URL is undefined or missing, even though the documentation states that the URL is not optional.

This is because of the following code in angular-mocks.js (class MockHttpExpectation):
```
this.matchUrl = function(u) {
    if (!url) return true;
```

A Plunkr to illustrate the point: http://plnkr.co/edit/k1bntn?p=preview

Is this an undocumented feature or a bug?

### LINKED ISSUE #10934: $httpBackend.expectGET doesn't fail with undefined url parameter
$httpBackend.expectGET() accepts all GET requests when it's url parameter is undefined. IMHO it should fail. 

I have a test where the URL parameter for expectGET() is retrieved from an array. After I renamed the array variable the tests continued to pass, when I was expecting them to fail.

    $httpBackend.expextGET(void 0).respond(response);
    issue GET
    $httpBackend.flush();

does not fail.

Angular 1.3.11, PhantomJS

petebacondarwin:
While the `url` parameter is optional for `$httpBackend.when`,
`$httpBackend.expect` and related shortcut methods, it should not have the
value of `undefined` if it has been provided.

This change ensures that an error is thrown in those cases.

Closes #8442
Closes #8462
Closes #10934

BREAKING CHANGE

It is no longer valid to explicitly pass `undefined` as the `url` argument
to any of the `$httpBackend.when...()` and `$httpBackend.expect...()`
methods.

While this argument is optional, it must have a defined value if it is
provided.

Previously passing an explicit `undefined` value was ignored but this
lead to invalid tests passing unexpectedly.

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1423,7 +1423,7 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
    * Creates a new backend definition.
    *
    * @param {string} method HTTP method.
-   * @param {string|RegExp|function(string)} url HTTP url or function that receives a url
+   * @param {string|RegExp|function(string)=} url HTTP url or function that receives a url
    *   and returns true if the url matches the current definition.
    * @param {(string|RegExp|function(string))=} data HTTP request body or function that receives
    *   data string and returns true if the data is as expected.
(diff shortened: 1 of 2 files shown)


Narretz:
This needs a rebase but otherwise looks good.

PR #12781, opened 2015-09-08
fix(limitTo): start at 0 if `begin` is negative and exceeds input length

### LINKED ISSUE #12775: Bug in limitTo filter when using negative begin parameter
There is an error in the limitTo filter begin/end position calculations.
I want to show a maximum 5 characters from a string no matter which length it is.

Therefore I am using this snippet: ```{{string | limitTo : 5 : -5  }}```
I expect the filter to show max. 5 chars beginning from the -5 index. This works as long as the string is at least 5 characters long. If it is shorter the filter won't display anything!

Example here:
http://plnkr.co/edit/tSdeeQ1Ue8rxyapC11vV?p=preview

Bug seems to be in any 1.4.x version of angular.

gkalpak:
Previously, specifying a negative `begin` whose abs value exceeds the input's length, would behave unexpectedly (depending on the value of `limit` relative to `begin`). E.g.:

```
limitToFilter('12345', 3, -7) === '1'
// but
limitToFilter('12345', 10, -7) === '123'
```

This commit fixes the unexpected behaviour, by setting `begin` to 0 in the aforementioned cases. Thus, the previous examples become:

```
limitToFilter('12345', 3, -7) === limitToFilter('12345', 3, 0) === '123'
// and
limitToFilter('12345', 10, -7) === limitToFilter('12345', 10, 0) === '12345'
```

Fixes #12775

### DIFF
--- a/src/ng/filter/limitTo.js
+++ b/src/ng/filter/limitTo.js
@@ -111,7 +111,7 @@ function limitToFilter() {
     if (!isArray(input) && !isString(input)) return input;
 
     begin = (!begin || isNaN(begin)) ? 0 : toInt(begin);
-    begin = (begin < 0 && begin >= -input.length) ? input.length + begin : begin;
+    begin = (begin < 0) ? Math.max(0, input.length + begin) : begin;
 
     if (limit >= 0) {
       return input.slice(begin, begin + limit);
--- a/test/ng/filter/limitToSpec.js
+++ b/test/ng/filter/limitToSpec.js
@@ -140,16 +140,19 @@ describe('Filter: limitTo', function() {
 
   it('should return an empty array if Y exceeds input length', function() {
     expect(limitTo(items, '3', 12)).toEqual([]);
-    expect(limitTo(items, 4, '-12')).toEqual([]);
     expect(limitTo(items, -3, '12')).toEqual([]);
-    expect(limitTo(items, '-4', -12)).toEqual([]);
   });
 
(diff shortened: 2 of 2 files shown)


frankweb:
Thank you for your fix, I've reviewed your solution and I think it's right. I don't have write access but I'd like someone merge this pull request.

frankweb:
This solution works fine for me, I wonder if someone could merge this pull request. I'd like to do other contribution based on this one.

petebacondarwin:
LGTM - merging

src/ng/filter/limitTo.js
```diff
@@ -111,7 +111,7 @@ function limitToFilter() {
     if (!isArray(input) && !isString(input)) return input;
 
     begin = (!begin || isNaN(begin)) ? 0 : toInt(begin);
-    begin = (begin < 0 && begin >= -input.length) ? input.length + begin : begin;
+    begin = (begin < 0) ? Math.max(0, input.length + begin) : begin;
```
lgalfaso: is this just a fancy way to do
```js
begin = Math.max(0, begin, input.length + begin);
```

?

gkalpak:
@petebacondarwin, thx for merging :smiley:
Should this also be backported to `v1.4.x` ?

petebacondarwin:
Yes. Can you do that @gkalpak ?

gkalpak:
Backported to `v1.4.x` as 4fc40bc.

petebacondarwin:
Thanks!!

PR #12782, opened 2015-09-08
docs(CONTRIBUTING.md): mention gitter community

gkalpak:


Narretz:
Since all the channels are unofficial anyway (maybe except the IRC), I think we can add it.

PR #12785, opened 2015-09-08
fix(input): ignore min/max if they are empty on all input types

### LINKED ISSUE #12363: ngAttr added attributes don't trigger directives
This may be by design, but I thought I'd bring it up.

Since `ngAttr`-prefixed attributes are interpolated with allOrNothing, an attribute may or may not be present initially. If the expression is defined on the first evaluation, it looks like the resulting attribute _will_ trigger directives. However, if the expression is initially undefined, when it later becomes defined the attribute that is added will not cause any directives to be applied to the element.

My use case is this: I'm using `ng-attr-min` and `ng-attr-max` on an `input[type=date]` with something like `ng-attr-min="{{minDate | date:'yyyy-MM-dd'}}"`. If `minDate` is undefined, I don't want there to be any min validation. When `minDate` becomes defined, the min attribute is properly set but the `min` directive is not applied retroactively, therefore I don't get my angular min validation.

The alternatives are
* Use `ngMin`, but this doesn't set the `min` attribute and on Chrome the datepicker won't be properly restricted.
* Use interpolation on `min` directly (e.g. `min="{{minDate | date:'yyyy-MM-dd'}}"`), but when `minDate` is undefined, this leaves a dangling `min` attribute with no value which incorrectly makes _any_ value invalid (this might be a bug).

gkalpak:
When the min/max attributes are empty (i.e. `attrs.min/max === ''`), there should be no min/max validation applied (i.e. all values should be valid wrt min/max). This works correctly for `input[number]`, but not for date input family (`input[date/datetime-local/time/week/month]`).

In the case on `input[number]`, an empty string for `attrs.min/max` is translated to `undefined` for `minVal/maxVal` and a check for `isUndefined(minVal/maxVal)` ensures that no min/max validation takes place.
For the data input family, an empty string for `attrs.min/max` is translated to `NaN` (by `parseDate()`), so an additional check (for `isNaN(minVal/maxVal)`) is required.

Fixes #12363

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1301,7 +1301,7 @@ function createDateInputType(type, regexp, parseDate, format) {
     if (isDefined(attr.min) || attr.ngMin) {
       var minVal;
       ctrl.$validators.min = function(value) {
-        return !isValidDate(value) || isUndefined(minVal) || isNaN(minVal) || parseDate(value) >= minVal;
+        return !isValidDate(value) || isUndefined(minVal) || parseDate(value) >= minVal;
       };
       attr.$observe('min', function(val) {
         minVal = parseObservedDateValue(val);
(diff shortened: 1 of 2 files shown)


Narretz:
Couldn't we check the emptiness inside `parseObservedDateValue` and return undefined in that case, too? It'd  be one check less for the actual validator to make.

gkalpak:
Good point, @Narretz. Updated.

Narretz:
LGTM after squashing the commits!

gkalpak:
Thx @Narretz ! <sub>(I just added the extra commits for "reviewability".)</sub>

Squashing and merging...

PR #12786, opened 2015-09-08
fix(angular.copy): support copying XML nodes

### LINKED ISSUE #5429: 1.2.5 angular-mocks copy change breaks tests using XML datatype
Apparently the XML datatype in Chrome/Firefox (and others?) contains pointers to itself.  The change here (https://github.com/angular/angular.js/commit/f69dc16241c8b631123ad0b09674f0a5e0ff32fe) copies the HTTP response.

If you are mocking an HTTP service with XML datatype as the return type, copy() never terminates.

petebacondarwin:
Closes #5429

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -825,6 +825,8 @@ function copy(source, destination, stackSource, stackDest) {
       } else if (isRegExp(source)) {
         destination = new RegExp(source.source, source.toString().match(/[^\/]*$/)[0]);
         destination.lastIndex = source.lastIndex;
+      } else if (isFunction(source.cloneNode)) {
+          destination = source.cloneNode(true);
       } else {
         var emptyObject = Object.create(getPrototypeOf(source));
         return copy(source, emptyObject, stackSource, stackDest);
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -386,6 +386,18 @@ describe('angular', function() {
       expect(aCopy).toBe(aCopy.self);
     });
 
+    it('should deeply copy XML nodes', function() {
+      var anElement = document.createElement('foo');
+      anElement.appendChild(document.createElement('bar'));
+      var theCopy = anElement.cloneNode(true);
+      expect(copy(anElement).outerHTML).toEqual(theCopy.outerHTML);
+      expect(copy(anElement)).not.toBe(anElement);
+    });
+
+    it('should not try to call a non-function called `cloneNode`', function() {
+      expect(copy.bind(null, { cloneNode: 100 })).not.toThrow();
+    });
+
     it('should handle objects with multiple references', function() {
       var b = {};
       var a = [b, -1, b];


petebacondarwin:
Good old Internet Explorer:

```
IE 11.0.0 (Windows 8.1) angular copy should support XML nodes FAILED
	Expected '<?XML:NAMESPACE PREFIX = "PUBLIC" NS = "URN:COMPONENT" /><foo></foo>' to equal '<foo></foo>'.
```

gkalpak:
Did you force-push a fix for the IE issue ?

test/AngularSpec.js
```diff
@@ -386,6 +386,13 @@ describe('angular', function() {
       expect(aCopy).toBe(aCopy.self);
     });
 
+    it("should support XML nodes", function() {
```
gkalpak: <sub>`"`  -->  `'` ? :blush: </sub>

src/Angular.js
```diff
@@ -825,6 +825,8 @@ function copy(source, destination, stackSource, stackDest) {
       } else if (isRegExp(source)) {
         destination = new RegExp(source.source, source.toString().match(/[^\/]*$/)[0]);
         destination.lastIndex = source.lastIndex;
+      } else if (source.cloneNode) {
```
lgalfaso: should this be `} else if (isFunction(source.cloneNode)) {` ?

src/Angular.js
```diff
@@ -825,6 +825,8 @@ function copy(source, destination, stackSource, stackDest) {
...
+          destination = source.cloneNode();
```
jbedard: Should this be a deep clone?

petebacondarwin: I guess you are right. I will add a unit test.

src/Angular.js
petebacondarwin: I guess that is tighter

test/AngularSpec.js
petebacondarwin: :-) cut and paste from the original commit

petebacondarwin:
@gkalpak - yes I hid my shame with a forced push. It was not really IE's fault (well yes it probably was, but it was mine too). Can you spot the mistake in my test:

```js
    it("should support XML nodes", function() {
      var anElement = document.createElement("foo");
      var theCopy = anElement.cloneNode();
      expect(copy(anElement.outerHTML)).toEqual(theCopy.outerHTML);
      expect(copy(anElement)).not.toBe(anElement);
    });
```

petebacondarwin:
@gkalpak and @jbedard I added a commit with your suggested changes. Thanks

gkalpak:
@petebacondarwin 
> Can you spot the mistake in my test

You mean besides the `"..."` ? :smile:
Oh...and the misplaced parenthesis :)

LGTM

petebacondarwin:
Thanks. I'll merge

PR #12790, opened 2015-09-08
Fix documentation formatting for angular.version

akshayagarwal:
Fixes summary description of angular.version on https://docs.angularjs.org/api/ng#object
The property description is best kept for the detailed page.
screen shot 2015-09-09 at 12 50 01 am


akshayagarwal:
I signed it!

akshayagarwal:
The failing test seems probably due to a socket connection error

Issue #12649, opened 2015-08-21
Can't build Cygwin and Windows 10 without --force


jrgleason:
I am trying to build and getting....

    $ grunt package
    Running "shell:npm-install" (shell) task
    Warning: EINVAL, invalid argument Use --force to continue.

    Aborted due to warnings.

Going to try and troubleshoot

    grunt-cli v0.1.13
    grunt v0.4.5

jrgleason:
In addition when I do use --force I get the following...

     >> build/angular-message-format.js minified into build/angular-message-format.min.js
    build\angular.js:4437: WARNING - Throw expression is not a minErr instance.
          throw err;
          ^

    build\angular.js:10656: WARNING - Throw expression is not a minErr instance.
            throw e;
            ^

    build\angular.js:12125: WARNING - Throw expression is not a minErr instance.
        throw e;
        ^

    build\angular.js:15989: WARNING - Throw expression is not a minErr instance.
            throw e;
            ^

    0 error(s), 4 warning(s)
    >> build/angular.js minified into build/angular.min.js


But it does in fact finish

jrgleason:
Could it be related to this line in the gruntfile...

    "npm-install": {
        command: path.normalize('scripts/npm/install-dependencies.sh')
      },

And the fact that I am on windows w/ cygwin (aka paths getting hosed)?

lgalfaso:
I do not have access to this environment, but if you can figure this out
and create a clean PR, I see no problem in merging it

Narretz:
I think this is is only half the solution. You'll probably run into https://github.com/angular/angular.js/issues/11143 after this is fixed

Narretz:
Actually, the path.normalize should handle different platform file paths, so it's probably a different problem. Can you run the command --verbose?

gkalpak:
`shell:npm-install` has been re-implemented in #12792.
@jrgleason, can you test if your issues are still there with the code on master ?

gkalpak:
Closing since the part that was causing issues has been re-implemented and there hasn't been any response for over a month. Feel free to re-open, if you are still facing this problem.

PR #12792, opened 2015-09-09
chore(check-node-modules): make check/reinstall node_modules work across platforms

### LINKED ISSUE #11143: grunt with new npm:install has problems on Windows

gkalpak:
The previous implementations (based on shell scripts) threw errors on Windows, because it was not able to `rm -rf` 'node_modules' (due to the 255 character limit in file-paths).

This implementation works consistently across platforms and is heavily based on 'https://github.com/angular/angular/blob/3b9c08676a4c921bbfa847802e08566fb601ba7a/tools/npm/check-node-modules.js'.

Fixes #11143
Closes #11353


gkalpak:
This might also be related to #12649.

TODO:

- [ ] See what Travis thinks.
- [ ] Test more thoroughly (espacially on platforms other than Windows).
- [ ] Decide if it is better to show the output of `npm install` or hide it and just show a progress indication (printing dots every few seconds).
    (Currently, the output is **not** shown.)

scripts/npm/check-node-modules.js
```diff
@@ -0,0 +1,175 @@
...
+  function deleteDirOrFileSync(subpath) {
+    var curPath = path + '/' + subpath;
+
+    if (fs.lstatSync(curPath).isDirectory()) {
+      deleteDirSync(curPath);
+    } else {
+      fs.unlinkSync(curPath);
+    }
+  }
+}
```
gkalpak: Would we care to use `rm -rf` on *NIX based systems for performance reasons ?
I didn't bother, because this will be running only rarely and then again `npm install` is the main bottleneck.

Narretz: In the initial discussion, we preferred this. But imo we can see if this works, and then change it back afterwards if it's much slower.

Narretz:
I noticed that this changes the behavior - previously, if there was a mismatch between the current and the cached lock file, it would automatically install, now it only warns It should work with the --reinstall option, but can you actually use this from the grunt level? I tried `grunt npm:install --reinstall` but it didn't work.

petebacondarwin:
The fact that it doesn't automatically reinstall is a good thing. It gives you a chance to finish what you are doing before wiping out all your node_modules. We should make the force reinstallation process reliable though

Narretz:
Hmm I'm getting quite a few of these (on Windows): 
`Error: ENOTEMPTY, directory not empty 'c:\Users\Martin\static\angular.js\node_modules\angular-benchpress\node_modules\browserify\node_modules\module-deps\node_modules\resolve\test\node_path\y'`

petebacondarwin:
Is that the old path too long problem?

petebacondarwin:
With this PR you force an install by running `node scripts/npm/check-node-modules.js --reinstall`

scripts/npm/check-node-modules.js
```diff
@@ -0,0 +1,175 @@
...
+
+function checkNodeModules(reinstallIfStale) {
+  var nodeModulesOk = compareMarkerFiles();
+
+  if (nodeModulesOk) {
+    console.log(':-) npm dependencies are looking good!');
+  } else {
+    console.warn(':-( npm dependencies are stale or in an unknown state!');
```
petebacondarwin: We ought to make this BIGGER and add a message about how to fix it. Something like:

```
console.warn('=====================================================================');
console.log(':-( npm dependencies are stale or in an unknown state!');
console.log('You can rebuild the dependencies by running `node scripts/npm/check-node-modules.js --reinstall`');
console.warn('=====================================================================');
```

petebacondarwin:
It is working well on my Mac

Narretz:
Strange, Now it finally uninstalled everything and installed. I got this:     
```
Running `npm install` (this may take a while)............................................................................................................................................................................................................................................
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
Operation completed with errors:
[Error: stderr maxBuffer exceeded.]
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
But I still get modules are stale error.
I assume it's because I'm not really running in Windows cmd, but in git bash inside ConEmu. Maybe the problem lies there somewhere.

petebacondarwin:
I think this is because, if `npm install` failed then the npm-shrinkwrap.json file never gets copied into the node_modules folder, which is what is used to check for out of date dependencies.

scripts/npm/check-node-modules.js
Narretz: Very good idea!

petebacondarwin:
Trying on windows 10

petebacondarwin:
Still waiting ...

petebacondarwin:
I just got the same maxBuffer exception. Will look into this tomorrow.

Narretz:
I've also got the buffer thing again. This time I've tried with admin rights, and that fixed the purge errors. Btw, they are actually different than path too long errors, it's probably some process holding on to the file. But if it works as admin, it's fine by me.

petebacondarwin:
I have been looking at the Angular 2 version. They don't bother trying to run `npm install` from the script.
Instead they have a preinstall hook that purges the old node_modules and a postinstall hook that copies the npm-shrinkwrap file. I quite like this since to ensure you have exactly the right modules you just run `npm install` and you are done.

gkalpak:
I tried it a few times on Windows 10 and never got that maxBuffer error. (But it is indeed too slow; it's `npm install` though that takes much time, so I don't think we can do better here.)

BTW (as I'm sure you've already figured out :)), if you uncomment [line 125][1] and [line 126][2] you get to see the stdout and stderr of `npm install`, which might provide useful contxt for errors (if they happen during `npm install`).

Could that also help with the `stderr maxBuffer exceeded` error ?

--
> They don't bother trying to run npm install from the script.
Instead they have a preinstall hook that purges the old node_modules and a postinstall hook that copies the npm-shrinkwrap file.

@petebacondarwin, you mean you like it as a convenience if you want to update the dependencies ?
(I.e. you can run `npm install` instead of `node scripts/npm/check-node-modules.js --reinstall`)
Or is there any other benefit as well ?

[1]: https://github.com/angular/angular.js/pull/12792/files#diff-69a37fa022e145ada77ff070e0c9a9ffR125
[2]: https://github.com/angular/angular.js/pull/12792/files#diff-69a37fa022e145ada77ff070e0c9a9ffR126

petebacondarwin:
There are two benefits.

1) It is intuitive to simply run `npm install` if you want to refresh the dependencies; rather than having to "know" about the script
2) It is less magic to have the `npm install` script run directly rather than using `exec()` to do it.

gkalpak:
Agreed :)

So, I will make the following changes:

1. `grunt shell:npm-install`: Will only check if the modules are up-to-date (via 'npm-shrinkwrap.json') and inform the user (with a BIGGER message in case of outdated deps, instructing them to run `npm install` to update the deps).

2. npm's `preinstall`: Will check if the modules are up-to-date and purge 'node_modules/` if they aren't.

3. npm's `postinstall`: Will copy 'npm-shrinkwrap.json' to 'node_modules/npm-shrinkwrap.cached.json'.

4. In `.travis.yml`: Will run `npm install` (instead of calling our script).

petebacondarwin:
Perfick!

gkalpak:
Added a 2nd (squashable upon merging) commit with the aforementioned changes.

Unfortunately, I am getting errors when purging from `npm install` (but it works fine when running `node scripts/npn/check-node-modules.js --purge` directly). It seems like the same `ENOTEMPTY` error you reported earlier :(

`npm install` output:

```bash
> angularjs@ preinstall C:\...\angular.js
> node scripts/npm/check-node-modules.js --purge

:-( npm dependencies are stale or in an unknown state!
    Purging 'node_modules'...
C:\...\angular.js\scripts\npm\check-node-modules.js:0
(function (exports, require, module, __filename, __dirname) { // Implementatio
^
Error: ENOTEMPTY, directory not empty 'C:\...\angular.js\node_modules\karma\node_modules\chokidar\node_modules'
    at Error (native)
    at Object.fs.rmdirSync (fs.js:711:18)
    at deleteDirSync (C:\...\angular.js\scripts\npm\check-node-modules.js:61:8)
    at deleteDirOrFileSync (C:\...\angular.js\scripts\npm\check-node-modules.js:69:7)
    at Array.forEach (native)
    at deleteDirSync (C:\...\angular.js\scripts\npm\check-node-modules.js:60:26)
    at deleteDirOrFileSync (C:\...\angular.js\scripts\npm\check-node-modules.js:69:7)
    at Array.forEach (native)
    at deleteDirSync (C:\...\angular.js\scripts\npm\check-node-modules.js:60:26)
    at deleteDirOrFileSync (C:\...\angular.js\scripts\npm\check-node-modules.js:69:7)

npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files\\nodejs\\\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install"
npm ERR! node v0.12.7
npm ERR! npm  v2.11.3
npm ERR! code ELIFECYCLE
npm ERR! angularjs@ preinstall: `node scripts/npm/check-node-modules.js --purge`
npm ERR! Exit status 1
...
```

It fails consistently, but on a different directory each time.
I'll look into it tomorrow (unless anyone has any insights in the meantime :smile:).

Narretz:
I had this error yesterday. Have you tdied running the cmd prompt / shell as admin

gkalpak:
Well, I tried it again today (with user rights - no admin) and it worked like a charm !?
I can't reproduce the ENOTEMPTY error (or any other error).
I also tried on a Linux (Debian) VM a couple of times and didn't run into any issues.

// @Narretz , @petebacondarwin

Narretz:
I think we should merge it and see if it causes any issues / makes it better on Windows.

petebacondarwin:
I agree. Let's merge this.

petebacondarwin:
Woohoo!

petebacondarwin:
Can we cherry pick it to 1.3 and 1.4 too @gkalpak ?

gkalpak:
I guess we could (since it only touches "chore" stuff).
Should I go for it right away, or give it a little time to see how it behaves ?

petebacondarwin:
Go for it. I have only seen it fail on windows. And since even that is intermittent, it is better than we have right now.

gkalpak:
Backported to:

* **v1.4.x** (9c1f8ea7)
* **v1.3.x** (bce1d8e)

petebacondarwin:
Smashing!

PR #12814, opened 2015-09-10
fix($browser/$location): add a workaround of iOS9 regression in UIWeb…

### LINKED ISSUE #12241: Infinite digest on location change on iOS 9 w/ UIWebView (not in Safari/ WKWebView)
The following simple HTML demonstrates the issue:
``` html
<!DOCTYPE html>
<html>
	<head>
		<script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.3.16/angular.js"></script>
		<script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.3.16/angular-route.js"></script>
		<script>
			angular.module('fail', ['ngRoute'])
			.config(function($routeProvider) {
				$routeProvider
				.when('/a', {
					template: '<a ng-href="#/b">a</a>'
				})
				.when('/b', {
					template: '<a ng-href="#/a">b</a>'
				})
				.otherwise({
					redirectTo: '/a'
				});
			});
		</script>
	</head>
	<body ng-app="fail">
		<div ng-view></div>
	</body>
</html>
```

This runs as expected on most devices, but it throws an infinite digest exception on iOS 9.
I'm able to reproduce on both iPad Air 2 and iPad 4th generation with iOS 9 beta 2.
I realize it's probably an issue in iOS, but it might still be worth investigating.

IgorMinar:
…View

This works around: https://openradar.appspot.com/22186109

Closes #12241
Related to driftyco/ionic#4082

### DIFF
--- a/src/ng/browser.js
+++ b/src/ng/browser.js
@@ -373,22 +373,18 @@ function applyIOS9Shim(browser) {
   patchedBrowser.url = function() {
     if (arguments.length) {
       pendingLocationUrl = arguments[0];
-
-      // reset the current pendingLocationUrl in the next task
-      (function() {
-        var pendingLocationUrlToReset = pendingLocationUrl;
-        setTimeout(function() {
-          if (pendingLocationUrlToReset === pendingLocationUrl) {
-            pendingLocationUrl = null;
-          }
-        },0);
-      }());
-
       return browser.url.apply(patchedBrowser, arguments);
     }
 
     return pendingLocationUrl || browser.url();
   };
+
+  jqLite(window).on('popstate', clearPendingLocationUrl);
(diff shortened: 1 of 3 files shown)


IgorMinar:
I've converted this change into an externalized [drop in patch](https://gist.github.com/IgorMinar/863acd413e3925bf282c) that can be added to Angular 1.2-1.4.5. @petebacondarwin is working on an alternative patch that will land in master (see https://github.com/angular/angular.js/compare/master...petebacondarwin:ios-9-fix-4)

petebacondarwin:
Actually it is in a PR now: https://github.com/angular/angular.js/pull/12819

PR #12817, opened 2015-09-10
feat(ngAnimate): introduce `$animate.flush` for unit testing

matsko:


test/ngAnimate/animateSpec.js
```diff
@@ -5059,8 +5042,8 @@ describe("ngAnimate", function() {
       });
       $rootScope.$digest();
 
-      $animate.triggerReflow();
-      $animate.triggerCallbackPromise();
+      $animate.flush();
+      $animate.flush();
```
IgorMinar: is this necessary? it should be sufficient to call flush just once.

test/ngAnimate/animateSpec.js
```diff
@@ -5089,12 +5072,12 @@ describe("ngAnimate", function() {
...
```
IgorMinar: same here

src/ngMock/angular-mocks.js
```diff
@@ -803,6 +803,29 @@ angular.mock.animate = angular.module('ngAnimateMock', ['ng'])
...
+            $$rAF.flush();
+          }
+          if ($$asyncCallback.queue.length) {
+            somethingFlushed = true;
+            this.triggerCallbackEvents();
+          }
+          if ($browser.deferredFns.length) {
+            somethingFlushed = true;
+            this.triggerCallbackPromise();
+          }
```
IgorMinar: we should put this into a loop and keep on flushing stuff until everything settles

IgorMinar:
the rest looks ok, but we need to test this at google to see if there will be issue with flushing browser.deferreds in this way

matsko:
Merged as https://github.com/angular/angular.js/commit/f98e038418f7367b2373adcf4887f64a8e8bdcb0

PR #12818, opened 2015-09-10
feat(ngModel): public method to force run model -> view binding pipeline

### LINKED ISSUE #12815: ngModel: force running formatters for custom things
Hello,

I am working on the application, where I have input field with ng-model,  inside which I need to show number in different units ata time (like px or %),
From the application point of view it is convenient to keep single unit in the model for calculations (i.e model contains only px values), and show different units in input field with help of parsers and formatters. I also have a toggle button for changing the unit.

The problem is , when I try to change the unit, I need formatters to run and store different viewValue inside input field (like instead of px value, show % value in input, though model still has px value).
I did not find any good way of doing this.
I end up incrementing my px value, and in $timeout decrement again, which runs second digest, and causes unwanted UI behaviour (like if model value is binded to something else, it will increment and decrement in few milliseconds, but you can still notice it).

Can we have an api which will trigger formatters for ngModelController and update view value?

aleksanyan:
Added ngModel.NgModelController#$forceModelToViewBound.
It triggers $formatters pipeline, and in case of changed $viewValue,
runs validators and update the UI without changing actual model.
Basically it simulates fake model value change.

Closes #12815

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -296,6 +296,21 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
 
   /**
    * @ngdoc method
+   * @name ngModel.NgModelController#$forceModelToViewBound
+   *
+   * @description
+   * This might be called to force update model -> view binding without changing actual model
+   *
+   * It will trigger $formatters pipeline, and in case of changed $viewValue, will run validators and
+   * update the UI.
+   *
+   * For instance you might use this to change the input's value, but keep the original scope.value the same.
+   */
+  this.$forceModelToViewBound = ngModelWatch.bind(this, true);
+
+
+  /**
+   * @ngdoc method
    * @name ngModel.NgModelController#$isEmpty
    *
    * @description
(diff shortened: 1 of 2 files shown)


Narretz:
Thanks for the PR. It's something that's come up a few times, we'll consider it for 1.5
In the example in your issue, where would you call this function? In a $parser?

aleksanyan:
Welcome. I would call it from ngModelController (as in PR). So it could be called publically in the end application.
I realize the PR's code is not clean enough, so if you like the idea, I can clean up and push.

Also would be good to have similar function for $parsers pipeline.
http://stackoverflow.com/questions/27565742/trigger-ngmodelcontroller-parsers-pipeline-from-a-watch/32149600#32149600

Narretz:
I meant from  where would you call it? As in from which part of your
application? Inside an ngModel parser, inside a directive when something
chnages or an event happens, or from ngChange ...
Am 11.09.2015 13:21 schrieb "Artur Aleksanyan" <notifications@github.com>:

aleksanyan:
Inside a directive, on click event for some button.

Narretz:
Implemented in https://github.com/angular/angular.js/commit/05fdf918d97a379634c8b0e15da8615edfc6fb8d

PR #12819, opened 2015-09-10
fix($browser): handle async updates to location

### LINKED ISSUE #12241: Infinite digest on location change on iOS 9 w/ UIWebView (not in Safari/ WKWebView)

petebacondarwin:
This is hopefully the fix that will go into master.
We will publish a patch for older versions separately.

Closes #12241

### DIFF
--- a/src/ng/browser.js
+++ b/src/ng/browser.js
@@ -87,7 +87,7 @@ function Browser(window, document, $log, $sniffer) {
   var cachedState, lastHistoryState,
       lastBrowserUrl = location.href,
       baseElement = document.find('base'),
-      reloadLocation = null;
+      pendingLocation = null;
 
   cacheState();
   lastHistoryState = cachedState;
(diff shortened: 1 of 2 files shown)


IgorMinar:
lgtm

PR #12829, opened 2015-09-11
fix(ngAnimateMock): $animate.flush should work for looping animations

matsko:


matsko:
Merged as https://github.com/angular/angular.js/commit/472d076cca2ffb99bd87d3c026ef69afc713268d

PR #12830, opened 2015-09-11
feat(ngList): add support for preserving blank tokens

mkalmanson:
Added feature to allow preserving blank tokens when 2 delimiters are back to back.  Previously, the blank tokens would be dropped.


Narretz:
What's your use case for preserving empty tokens?

mkalmanson:
We are using ngList in an administration screen for controlling the text on
a multi-line display that accepts an array of strings as input.  Our users
would like to be able put blank lines on the display, but when they do, the
blank lines get stripped.  We are currently telling our users to put a
space on the blank line as a workaround, but that feels like a hack.

PR #12831, opened 2015-09-12
refactor: use isDefined and isUndefined consistently

### LINKED ISSUE #4365: isUndefined and isDefined should be used more consistently
I noticed the functions ```isUndefined``` and ```isDefined``` in src/Angular.js, which seems like a good idea to standardise the checking of defined/undefined variables. However, I noticed that still checks for undefined/defined are done without these functions in quite some places throughout the code. Have a look at the ```equals``` function defined in src/Angular.js around line 752, as an example. There the statement ```o2[key] !== undefined```can be found. It seems to me that this should instead be ```isDefined(o2[key])``` to keep the consistency and obey the defined/undefined check standardisation.

tl;dr; Only ```isDefined```and ``isUndefined``` should be used the check for undefined/defined variables.

If this is desired, I would happily create a pull request fixing this.

lucastetreault:
Fix any place that x == undefined or x != undefined
to use isUndefined and isDefined instead

Closes #4365

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -977,7 +977,7 @@ function equals(o1, o2) {
         for (key in o2) {
           if (!(key in keySet) &&
               key.charAt(0) !== '$' &&
-              o2[key] !== undefined &&
+              isDefined(o2[key]) &&
               !isFunction(o2[key])) return false;
         }
         return true;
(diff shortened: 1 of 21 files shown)


lgalfaso:
Hi, thanks for the PR. Can you please
* Squash everything into a single commit
* Fix the commit comment as this is not changing `x == undefined` and `x != undefined` as it is changing `x === undefined` and `x !== undefined` (that have a slightly different meaning)

petebacondarwin:
@lgalfaso thanks for triaging this one. I think that if it looks good then we can to the squash and commit message fix up when merging to save on cycles.

Thanks @lucastetreault for fixing this very long running issue!

Issue #12787, opened 2015-09-08
ngResource is not suitable for single primitive value(such as string or number) response


kuitos:
the web api(restful most of time) may return a single primitive value as response,such as a string or a number, like this

```js
GET /users/count  ----> 10
GET /users/1/name ------> "Jack"
```
single primitive value(string or number or true&false and so on) is conform to the specifications of JSON,but now ngResource just support Array or Object response,and I know the reason when I saw the ngResource source code:

```js
........
if (action.isArray) {
  value.length = 0;
  forEach(data, function(item) {
    if (typeof item === "object") {
      value.push(new Resource(item));
    } else {
      // Valid JSON values may be string literals, and these should not be converted
      // into objects. These items will not have access to the Resource prototype
      // methods, but unfortunately there
      value.push(item);
    }
  });
} else {
  shallowClearAndCopy(data, value);
  value.$promise = promise;
}
..........
/**
 * Create a shallow copy of an object and clear other fields from the destination
 */
function shallowClearAndCopy(src, dst) {
  dst = dst || {};

  angular.forEach(dst, function(value, key) {
    delete dst[key];
  });

  for (var key in src) {
    if (src.hasOwnProperty(key) && !(key.charAt(0) === '$' && key.charAt(1) === '$')) {
      dst[key] = src[key];
    }
  }

  return dst;
}
```
Obviously it can not resolve single string or number response.
What should I do when the response is single string with ngResource?

kuitos:
Now I have to overwrite the shallowClearAndCopy method to support my business

```js
function shallowClearAndCopy(src, dst) {
  dst = dst || {};

  angular.forEach(dst, function(value, key) {
    delete dst[key];
  });

  if (angular.isString(src) || angular.isNumber(src)) {
      dst._data = src;
    } else {
      for (var key in src) {
        if (src.hasOwnProperty(key) && !(key.charAt(0) === '$' && key.charAt(1) === '$')) {
          dst[key] = src[key];
        }
      }
    }

  return dst;
}
```
what it do is to wrap the single value response to {_data:xxx},and then I can get the value by resource._data.
and then override the $resource service use $provide.decorator.
But it is so ugly and I do not have any other ideas

Narretz:
Are you saying that `"string"` and `4434` are valid JSOn responses? I don't think that is the case.

Narretz:
[...]

PR #12832, opened 2015-09-13
fix(ngResource): compatible with the primitive single value in standard json

kuitos:
wrap response into {_value:response} format when the response is a single value of json

see #12787


kuitos:
I signed it!

PR #12833, opened 2015-09-13
fix($parse): throw error when accessing a restricted property indirectly

lgalfaso:
When accessing an instance thru a computed member and the property is an array, then also check the string value of the array.


lgalfaso:
@petebacondarwin please review this when you have a chance. I would like this to make it into the next 1.4.x release

petebacondarwin:
Can you describe the real life scenario?

petebacondarwin:
Merging. Thanks @lgalfaso

PR #12834, opened 2015-09-13
Typo/mismatch fix

ghost:
There is an mismatch for status in controller and test. 
In controller $scope.status = 'ERROR!' and in test we expect($rootScope.status).toBe('Failed...') so the test will fail;


toastman:
I signed it!

toastman:
@Narretz You don't agree with my suggestion ? Please look at $htmlBackend docs https://docs.angularjs.org/api/ngMock/service/$httpBackend

gkalpak:
@toastman, he must totally agree with your suggestion, because he (@Narretz) merged it into master (159efdd) :smile:

toastman:
@gkalpak Thank you, I'm just misunderstood ("This pull request is closed, but the toastman:patch-1 branch has unmerged commits.")

PR #12839, opened 2015-09-14
fix(dateFilter): follow the CLDR on pattern escape sequences

lgalfaso:
When there are two single quotes "''" (quotes for clarification) that are not part of an escape sequence, then this sequence should be handled as one single quote. See http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns second and forth examples


petebacondarwin:
LGTM - please merge.

PR #12840, opened 2015-09-14
fix(ngOptions): throw if ngModel is not present

Narretz:
BREAKING CHANGE:
`ngOptions` will now throw if `ngModel` is not present on the `select`
element. Previously, having no `ngModel` let `ngOptions` silently
fail, which could lead to hard to debug errors. The change should
therefore not affect any applications, as it simply makes the
requirement more strict and alerts the developer explicitly.


petebacondarwin:
LGTM. Please merge.

PR #12843, opened 2015-09-14
feat($parse): allow forcing expensive checks at the provider level

lgalfaso:
Allow forcing expensive checks at the $parse provider level

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1687,6 +1687,25 @@ function getValueOf(value) {
 function $ParseProvider() {
   var cacheDefault = createMap();
   var cacheExpensive = createMap();
+  var forceExpensiveChecks = false;
+
+  /**
+   * @ngdoc method
+   * @name $parseProvider#expensiveChecks
+   * @description
+   * Force expensive checks on all parsed expressions. Defaults to `false`.
+   *
+   * @param {boolean=} value new value to set the force expensive checks to.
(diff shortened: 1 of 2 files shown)


Narretz:
Can you please add labels / milestone to the issue? And should the description be more explicitly what these expensive checks are and when you should enable them?

src/ng/parse.js
```diff
@@ -1687,6 +1687,25 @@ function getValueOf(value) {
 function $ParseProvider() {
   var cacheDefault = createMap();
   var cacheExpensive = createMap();
+  var forceExpensiveChecks = false;
+
+  /**
+   * @ngdoc method
+   * @name $parseProvider#expensiveChecks
+   * @description
+   * Force expensive checks on all parsed expressions. Defaults to `false`.
```
gkalpak: I think a little explanation on what `false` means is in place (i.e. that will still apply expensive checks in some places or even what "expensive checks" include).

src/ng/parse.js
```diff
@@ -1687,6 +1687,25 @@ function getValueOf(value) {
...
+   *
+   * @param {boolean=} value new value to set the force expensive checks to.
```
gkalpak: _force expensive checks_  -->  _forceExpensiveChecks_ ?

lgalfaso:
Sorry for the noise, I thought that adding this option would help with another problem, but this is not the case. Closing

PR #12844, opened 2015-09-14
add standalone months to ngLocale

Narretz:
Cleaned up version of #3744. We could also add STANDALONESHORTMONTHS, but I don't think that's worth it.


lgalfaso:
this looks fine, but do you plan on adding the support for this at `dateFilter`?

Narretz:
I didn't think of that at all actually. I proposed it because a few people pointed out that it wasn't available. Can we add the support for dateFilter at a later time?

petebacondarwin:
LGTM - merging

petebacondarwin:
(For some reason the second commit added a full stop after `kr` currencies. I have reverted those changes when merging.)

PR #12848, opened 2015-09-14
feat(ngModel): provide ng-empty and ng-not-empty CSS classes

### LINKED ISSUE #10050: Proposal: .ng-prefilled class helper
This would be handy in the following use case:

You want to show validation messages either

* after the form is submitted

* on field touch, **provided the field has already a value**. In this case I'd like the user to notice the validity of the new value right after touch -- instead of waiting until form submission.

So it would work like this (relevant classes mentioned):

1. the field is in a *untouched state*, but has a value: `ng-untouched` `ng-pristine` `ng-prefilled`
2. user *changes* field value: `ng-untouched` `ng-dirty` `ng-prefilled`
3. user *blurs* field: `ng-touched` `ng-dirty` `ng-prefilled`

Does this makes sense to you, guys?

petebacondarwin:
I believe that this feature should be enough to fix #10050

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -14,7 +14,9 @@ var VALID_CLASS = 'ng-valid',
     DIRTY_CLASS = 'ng-dirty',
     UNTOUCHED_CLASS = 'ng-untouched',
     TOUCHED_CLASS = 'ng-touched',
-    PENDING_CLASS = 'ng-pending';
+    PENDING_CLASS = 'ng-pending',
+    EMPTY_CLASS = 'ng-empty',
+    NOT_EMPTY_CLASS = 'ng-not-empty';
 
 var ngModelMinErr = minErr('ngModel');
 
@@ -315,6 +317,17 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
   this.$isEmpty = function(value) {
     return isUndefined(value) || value === '' || value === null || value !== value;
   };
+
+  this.$$updateEmptyClasses = function(value) {
+    if (ctrl.$isEmpty(value)) {
+      $animate.removeClass($element, NOT_EMPTY_CLASS);
+      $animate.addClass($element, EMPTY_CLASS);
+    } else {
+      $animate.removeClass($element, EMPTY_CLASS);
+      $animate.addClass($element, NOT_EMPTY_CLASS);
+    }
+  };
+
 
   var currentValidationRunId = 0;
 
@@ -652,6 +665,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
     if (ctrl.$$lastCommittedViewValue === viewValue && (viewValue !== '' || !ctrl.$$hasNativeValidators)) {
       return;
     }
+    ctrl.$$updateEmptyClasses(viewValue);
     ctrl.$$lastCommittedViewValue = viewValue;
 
     // change to dirty
@@ -834,6 +848,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
         viewValue = formatters[idx](viewValue);
       }
       if (ctrl.$viewValue !== viewValue) {
+        ctrl.$$updateEmptyClasses(viewValue);
         ctrl.$viewValue = ctrl.$$lastCommittedViewValue = viewValue;
         ctrl.$render();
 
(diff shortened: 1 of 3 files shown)


test/ng/directive/ngModelSpec.js
```diff
@@ -1182,46 +1182,6 @@ describe('ngModel', function() {
...
-        $rootScope.$digest();
-
-        expect(input).toBeValid();
-        expect(input).not.toHaveClass('ng-invalid-maxlength');
-        expect(input).toHaveClass('ng-valid-maxlength');
-        expect(addClassCallCount).toBe(1);
-        expect(removeClassCallCount).toBe(0);
-
-        dealoc(element);
-      }));
```
gkalpak: Out of curiosity, why was this removed ?

gkalpak:
I wonder if people will ask for this to be propagated to a parent `FormController`...

petebacondarwin:
I was trying not to mention that :-P

Narretz:
For the original use case, you would then use`ng-pristine` and `ng-not-empty` together?

src/ng/directive/ngModel.js
```diff
@@ -316,6 +318,17 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
     return isUndefined(value) || value === '' || value === null || value !== value;
   };
 
+  this.$$updateEmptyClasses = function(value) {
+    if (this.$isEmpty(value)) {
```
Narretz: Should `this` be `ctrl`?

test/ng/directive/ngModelSpec.js
```diff
@@ -1368,6 +1328,26 @@ describe('ngModel', function() {
   describe('CSS classes', function() {
     var EMAIL_REGEXP = /^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+@[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i;
 
+    it('should set ng-empty or ng-not-empty the view value changes', inject(function($compile, $rootScope, $sniffer) {
```
Narretz: .. ng-not-empty (when) the view value changes ..

Narretz: Also, line lengths? I never know what the limit should be. 100?

gkalpak: Theoretically, yes, they should be 100 chars max :smiley:
See: [CONTRIBUTING.md#-coding-rules][1]

[1]: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-coding-rules

src/ng/directive/ngModel.js
petebacondarwin: It doesn't really matter because we have control over when it is called but I agree it is probably better for consistency.

test/ng/directive/ngModelSpec.js
petebacondarwin: Our jshint rules are now set to 200 for line length :-)

petebacondarwin:
Thanks for the reviews @Narretz and @gkalpak. I added a commit.

test/ng/directive/ngModelSpec.js
gkalpak: Pure evil :-P

petebacondarwin:
@Narretz  and @gkalpak can you take another look?

src/ng/directive/ngModel.js
```diff
@@ -316,6 +318,17 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
     return isUndefined(value) || value === '' || value === null || value !== value;
   };
 
+  this.$$updateEmptyClasses = function(value) {
```
Narretz: This could be simplified by always using ctrl.$viewValue instead of passing it as an arg. It should only be the viewValue anyway. You'll have to change the order on line https://github.com/angular/angular.js/pull/12848/files#diff-ae1963e24d6328c3aa019c0c1b9f4432R851 though, because otherwise it'll use the old viewValue

Narretz:
Regarding my question
> For the original use case, you would then use ng-pristine and ng-not-empty together?

because only ng-empty doesn't work if I understood the issue correctly.

petebacondarwin:
@Narretz: Yes, you would use a combination of the CSS classes for the original use case, but this seemed reasonable and intuitive to me.

src/ng/directive/ngModel.js
petebacondarwin: Actually it would have to use `$$lastCommittedViewValue` and we would have to change the order as you suggest. I am not that keen on doing this though as I don't feel that not passing in an argument makes it that much simpler.

petebacondarwin:
Thanks for the review @Narretz.
So assuming that we don't simplify the method and that the original use case is covered by a combination of CSS classes. Is this PR ready to merge?

Narretz:
Yep, feel free to merge!
Am 21.09.2015 22:45 schrieb "Pete Bacon Darwin" <notifications@github.com>:

PR #12850, opened 2015-09-14
fix(filters): ensure `formatNumber` observes i18n decimal separators

### LINKED ISSUE #10342: Number Decimal Separator Incorrect for exponentially small numbers
**Issue**

The number decimal separator are incorrect for the fr-fr locale in formatNumber for exponentially small numbers.

Expected

 - Decimal: ,

Actual

 - Decimal: .

In fact, it's an issue for all languages that do not have a dot as decimal separator.

I suggest the same use case than ["should filter exponentially small numbers"](https://github.com/angular/angular.js/blob/g3_v1_3/test/ng/filter/filtersSpec.js#L196) to test this issue:

```javascript
     it('should filter exponentially small numbers on non dot decimal separator languages', function() {
       $locale.NUMBER_FORMATS.DECIMAL_SEP = ',';
       expect(number(1e-50, 0)).toEqual('0');
       expect(number(1e-6, 6)).toEqual('0,000001');
       expect(number(1e-7, 6)).toEqual('0,000000');
 
       expect(number(-1e-50, 0)).toEqual('0');
       expect(number(-1e-6, 6)).toEqual('-0,000001');
       expect(number(-1e-7, 6)).toEqual('-0,000000');
     });
```

**Solve suggestion**

I figure out where the issue is localized: https://github.com/angular/angular.js/blob/g3_v1_3/src/ng/filter/filters.js#L206

```javascript
    if (fractionSize > 0 && number < 1) {
      formatedText = number.toFixed(fractionSize);
      number = parseFloat(formatedText);
    }
```

The [`.toFixed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) method is directly use to render the final number. And so, does NOT care about i18n at all.

To solve this issue, I suggest just to replace the dot by the expected decimal separator. As we deal this just number < 1, we don't have to manage thousand group separators.

```javascript
    if (fractionSize > 0 && number < 1) {
      formatedText = number.toFixed(fractionSize);
      number = parseFloat(formatedText);
      formatedText = formatedText.replace(DECIMAL_SEP, decimalSep);
    }
```

petebacondarwin:
Closes #10342

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -214,6 +214,7 @@ function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
     if (fractionSize > 0 && number < 1) {
       formatedText = number.toFixed(fractionSize);
       number = parseFloat(formatedText);
+      formatedText = formatedText.replace(DECIMAL_SEP, decimalSep);
     }
   }
 
--- a/test/ng/filter/filtersSpec.js
+++ b/test/ng/filter/filtersSpec.js
@@ -62,6 +62,8 @@ describe('filters', function() {
     it('should format according different separators', function() {
       var num = formatNumber(1234567.1, pattern, '.', ',', 2);
       expect(num).toBe('1.234.567,10');
+      num = formatNumber(1e-14, pattern, '.', ',', 14);
+      expect(num).toBe('0,00000000000001');
     });
 
     it('should format with or without fractionSize', function() {


src/ng/filter/filters.js
```diff
@@ -214,6 +214,7 @@ function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
     if (fractionSize > 0 && number < 1) {
```
gkalpak: It's not directly related to the fix, but what happens if `fractionSize` is 0 ?

gkalpak: Nevermind, I see what happens :)

lgalfaso:
LGTM

Narretz:
Tangential, but do you know if the Number input type takes i18n into consideration?

petebacondarwin:
It doesn't look like the HTML5 input[number] control handles i18n at all: The value (http://www.w3.org/TR/html-markup/input.number.html#input.number.attrs.value) of the input must be a string representing a floating point number (http://www.w3.org/TR/html-markup/datatypes.html#common.data.float)

PR #12852, opened 2015-09-14
fix(jqLite) - toggleClass & removeClass reflect documentation

### LINKED ISSUE #12793: element.toggleClass does not accept a function
Either update `toggleClass` to be compatible with jQuery 1.4 and [accept a function](http://api.jquery.com/toggleclass/#toggleClass-function-state) or note in the [angular.element docs](https://docs.angularjs.org/api/ng/function/angular.element) that it is not compatible with that method signature.

Running `toggleClass` with a function at the moment produces

> TypeError: selector.split is not a function

due to the assumption that the first argument, if present, is a string.

### LINKED ISSUE #12851: jqLite - removeClass cannot accept a function
Angular documentation redirects to [jQuery 1.4](http://api.jquery.com/removeClass/), which states that `removeClass` should be able to take a function as a parameter. Also they state that if no parameter is provided `removeClass` should remove all classes.

This is currently not the case. 

I am submitting a pull request to fix this issue.

JordanMajd:
Updates removeClass to accept function  in spec with 1.9 (current) jQuery docs.
Updates removeClass to accept no parameters in spec with 1.9 (current) jQuery docs.
Updates toggleClass to accept function in spec with 1.9 (current) jQuery docs.
Updates toggleClass to accept no parameters in spec with 1.9 (current) jQuery docs.
Provides tests for all changes.

fixes angular/angular.js#12793
fixes angular/angular.js#12851

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -381,15 +381,26 @@ function jqLiteHasClass(element, selector) {
       indexOf(" " + selector + " ") > -1);
 }
 
-function jqLiteRemoveClass(element, cssClasses) {
-  if (cssClasses && element.setAttribute) {
-    forEach(cssClasses.split(' '), function(cssClass) {
-      element.setAttribute('class', trim(
-          (" " + (element.getAttribute('class') || '') + " ")
-          .replace(/[\n\t]/g, " ")
-          .replace(" " + trim(cssClass) + " ", " "))
-      );
-    });
+function jqLiteRemoveClass(element, cssClassesOrFunc) {
+  if (element.setAttribute) {
+    if (cssClassesOrFunc) {
+      if (typeof cssClassesOrFunc === 'function') {
+        forEach(element.className.split(/\s+/), function(className, index) {
+          cssClassesOrFunc(index, className);
+        });
+      } else {
+        forEach(cssClassesOrFunc.split(' '), function(cssClass) {
+          element.setAttribute('class', trim(
+              (" " + (element.getAttribute('class') || '') + " ")
+              .replace(/[\n\t]/g, " ")
+              .replace(" " + trim(cssClass) + " ", " "))
+          );
+        });
+      }
+    } else {
+      // if no class names are specified in the parameter, all classes will be removed
+      element.setAttribute('class', '');
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/jqLite.js
```diff
@@ -381,15 +381,26 @@ function jqLiteHasClass(element, selector) {
...
-      element.setAttribute('class', trim(
-          (" " + (element.getAttribute('class') || '') + " ")
-          .replace(/[\n\t]/g, " ")
-          .replace(" " + trim(cssClass) + " ", " "))
-      );
-    });
+function jqLiteRemoveClass(element, cssClassesOrFunc) {
+  if (element.setAttribute) {
+    if (cssClassesOrFunc) {
+      if (typeof cssClassesOrFunc === 'function') {
```
Narretz: can use `isFunction`

src/jqLite.js
```diff
@@ -381,15 +381,26 @@ function jqLiteHasClass(element, selector) {
...
+        forEach(element.className.split(/\s+/), function(className, index) {
```
Narretz: I don't think this is doing the right thing. jquer docs say:
> A function returning one or more space-separated class names to be added to the existing class name(s). Receives the index position of the element in the set and the existing class name(s) as arguments. Within the function, this refers to the current element in the set.

src/jqLite.js
```diff
@@ -935,14 +946,35 @@ forEach({
...
-      forEach(selector.split(' '), function(className) {
-        var classCondition = condition;
-        if (isUndefined(classCondition)) {
-          classCondition = !jqLiteHasClass(element, className);
-        }
-        (classCondition ? jqLiteAddClass : jqLiteRemoveClass)(element, className);
-      });
+
+      // the class name to be toggled can be determined by passing in a function.
+      if (typeof selector === 'function') {
```
Narretz: -> `isFunction`

src/jqLite.js
```diff
@@ -935,14 +946,35 @@ forEach({
...
+
+        forEach(element.className.split(/\s+/), function(className, index) {
+          if (Array.isArray(condition) && condition.length >= index) {
```
Narretz: -> `isArray`

src/jqLite.js
JordanMajd: @Narretz 
I will look into this, thank you.

JordanMajd:
@Narretz , thanks for the reviews. I've fixed the items you had mentioned as well as a few other things I had noticed that could use improvement and am ready for another review.

gkalpak:
There is still the inconsistency of the passed index. (It's always 0.)

JordanMajd:
Thanks for the input. However, I think passing anything but 0 would be inconsistent. Please correct me if I'm mistaken, but it is my understanding that  `jqLite` and `jQuery` are intrinsically going to be different. This is because `jQuery` can work on multiple elements, while `jqLite` will only operate on a single element. Since `jqLite` is only operating on one element, the element will always be index 0.

gkalpak:
Off the top of my head, I think `jqLite` operates on multiple elements as well.

JordanMajd:
@gkalpak thanks, 

I just did some digging, you are correct! I will fix this.

Narretz:
Feel free to reopen this PR or open a new one if you have remedied the issues pointed out.

PR #12860, opened 2015-09-15
fix($parse): block assigning to fields of a constructor

lgalfaso:
Throw when assigning to a field of a constructor.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -102,6 +102,16 @@ function ensureSafeFunction(obj, fullExpression) {
   }
 }
 
+function ensureSafeAssignContext(obj, fullExpression) {
+  if (obj) {
+    if (obj === (0).constructor || obj === (false).constructor || obj === ''.constructor ||
+        obj === {}.constructor || obj === [].constructor || obj === Function.constructor) {
+      throw $parseMinErr('isecaf',
+        'Assigning to a constructor is disallowed! Expression: {0}', fullExpression);
+    }
+  }
+}
+
 var OPERATORS = createMap();
 forEach('+ - * / % === !== == != < > <= >= && || ! = |'.split(' '), function(operator) { OPERATORS[operator] = true; });
 var ESCAPE = {"n":"\n", "f":"\f", "r":"\r", "t":"\t", "v":"\v", "'":"'", '"':'"'};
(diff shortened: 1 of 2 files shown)


src/ng/parse.js
```diff
@@ -102,6 +102,16 @@ function ensureSafeFunction(obj, fullExpression) {
...
 }
 
+function ensureSafeAssignContext(obj, fullExpression) {
+  if (obj) {
+    if (obj === (0).constructor || obj === (false).constructor || obj === ''.constructor ||
+        obj === {}.constructor || obj === [].constructor || obj === Function.constructor) {
+      throw $parseMinErr('isecaf',
+        'Assigning to a constructor is disallowed! Expression: {0}', fullExpression);
+    }
+  }
```
petebacondarwin: So we only care about built-in constructors?

lgalfaso: we care the ones an expression can invoke and can be used to break the sandbox... the only other I can think about is `Scope`, but I really have no idea if it can be maliciously used. The alternative would be blocking to any property named `constructor`, but it might be just too much and there might be valid uses.

petebacondarwin: So I guess we can't protect against developers putting stuff on the scope via a controller, such as the `window` object, for instance?

test/ng/parseSpec.js
```diff
@@ -2703,6 +2703,32 @@ describe('parser', function() {
...
+          }).toThrow();
+          // foo.constructor is the object constructor.
+          expect(function() {
+            scope.$eval("foo.constructor[0] = ''", {foo: {}});
+          }).toThrow();
+          // foo.constructor is not a constructor.
+          expect(function() {
+            scope.$eval("foo.constructor[0] = ''", {foo: {constructor: ''}});
+          }).not.toThrow();
+        });
```
petebacondarwin: For good measure we could also test assigning to a variable that has been set to a constructor:

```js
expect(function() {
  scope.$eval("foo = {}.constructor; foo.join = ''");
}).toThrow();
```

petebacondarwin:
I merged this into master and 1.4.x.
It didn't go cleanly into 1.3.x. @lgalfaso is it backportable to 1.3 and 1.2?

src/ng/parse.js
lgalfaso: there are other protections in place that prevent any form of access to several objects. Eg. DOM nodes and `window`

test/ng/parseSpec.js
lgalfaso: done

lgalfaso:
@petebacondarwin sure, will do so later this week

petebacondarwin:
Magic!

Issue #12175, opened 2015-06-20
$locationChangeSuccess not fires with browser back and hash is empty


fspezi:
I am using **angular 1.3.16**, in Google Chrome.  
I found a wrong behavior of $locationChangeSuccess event.

To reproduce:
* open a page like "path/#"
* click to a link to go to a page like "path/#arg=123"
* click to a link to go to a page like "path/#arg=555"
* now, click to "Back" button of browser. You will go to "path/#arg=123"; $locationChangeSuccess event is fires.
* click another time to "Back" button of browser. You will go to "path/#"; sometimes, $locationChangeSuccess event is NOT fired ($locationChangeStart is fired).
* click to "foreward" button of browser. You will go to "path/#arg=123"; $locationChangeSuccess is fired.
* click to "back" button of browser. You will go to "path/#"; $locationChangeSuccess is NOT fired.

I found in "angular.js" source code that there is a comparison between $location.absUrl() and `newUrl` local variable. `$location.absUrl()` returns "path/", and `newUrl` returns "path/#".  
Row 11486: `if ($location.absUrl() !== newUrl) return;`  

This comparison is made after "$locationChangeStart" and before "$locationChangeSuccess". The comparison do exit by function.

The $succes

juniorjpf:
fspezi, i've tried to reproduce this bug, but no success. Can you make a example into Plunker http://plnkr.co/ or post here your files ?

fspezi:
I develop a test page. I discovered that the issue fires if the path ends with "/" like "www.acme.com/", or "www.acme.com/path/".  
You can [download code from here](http://plnkr.co/edit/hKv1Kd?p=preview), then install to local website, so you will open 'index.html' as default page of folder.  
Then, follow step, and watch fired events below.

(you can [open the code in my local server](http://angularjs.location.dev.xilium.it/#) )

Deklin:
This issue exists in 1.4.4 as well.

timmolendijk:
I think this demo illustrates the same issue. It uses version 1.4.6. Maybe it's helpful.

http://output.jsbin.com/zihoza/quiet/

Deklin:
Outstanding!

lbysun:
Hi, this bug still exists in 1.3x. Just wondering if there is any reason why the above [PR](https://github.com/angular/angular.js/pull/13251) is not merged into 1.3x. 

Thank you!

gkalpak:
[...]

PR #12862, opened 2015-09-15
fix($location): $locationchangesuccess not fires with browser back when path ends with '/#'. #12175

rrsivabalan:
In line #930, $location.absUrl() returns "path/", and newUrl returns "path/#". with browser back, so if condition is positive and returns. For fix I used trimEmptyHash function to trim # from newUrl, the same function is used in browser forward scenario @line#947.

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -922,12 +922,12 @@ function $LocationProvider() {
         $location.$$parse(newUrl);
         $location.$$state = newState;
 
-        defaultPrevented = $rootScope.$broadcast('$locationChangeStart', newUrl, oldUrl,
+        defaultPrevented = $rootScope.$broadcast('$locationChangeStart', trimEmptyHash(newUrl), oldUrl,
             newState, oldState).defaultPrevented;
 
         // if the location was changed by a `$locationChangeStart` handler then stop
         // processing this location change
-        if ($location.absUrl() !== newUrl) return;
+        if ($location.absUrl() !== trimEmptyHash(newUrl)) return;
 
         if (defaultPrevented) {
           $location.$$parse(oldUrl);


rrsivabalan:
This Pull Request fixes #12175

rrsivabalan:
@Narretz  can you please review  code changes( closes #12175 )

rrsivabalan:
@petebacondarwin Can you please review changes(closes #12175)

petebacondarwin:
@rrsivabalan - thanks for this PR.
First, it needs a unit test to demonstrate the bug you are fixing.
Also, it seems wasteful to trim the hash from the `newUrl` twice. Shouldn't we just do it once and store it in a variable?

gggritso:
This issue has been stale for about a month now, I'm hoping to get it pushed through; this is a somewhat important problem for me. I think this unit test shows the bug:

```javascript
it('should should detect location changes to bare hash',
  inject(function($location, $browser, $rootScope, $log) {

    $rootScope.$on('$locationChangeSuccess', function(event, newUrl, oldUrl) {
      $log.info('changed', newUrl, oldUrl);
    });

    $browser.url('http://server/#/');
    $browser.poll();

    expect($log.info.logs.shift()).
      toEqual(['changed', 'http://server/#/', 'http://server/']);

    $browser.url('http://server/#');
    $browser.poll();

    expect($log.info.logs.shift()).
      toEqual(['changed', 'http://server/#', 'http://server/#/']);

    expect($browser.url()).toEqual('http://server/#');
  })
);
```

Unfortunately, I don't think the code in this PR correctly covers this case. I would love to submit a PR myself, but right now I don't have enough Angular knowledge to propose a good solution. Is there anything else I can do to help out here?

petebacondarwin:
We are working our way through the 1.4.x bugs

gggritso:
That’s great to hear, thank you for your hard work!

rrsivabalan:
@petebacondarwin  sorry for delay, i have  fixed code review comments and added UT to demonstrate issue  
New PR: https://github.com/angular/angular.js/pull/13251
Please note that i got merge conflict  issue in existing PR, so raised new PR 
@gggrisot my UT is pretty similar to your UT, i think my PR covers your test case.

PR #12864, opened 2015-09-16
fix(ngAnimate): callback detection should only use RAF when necessary

matsko:
Callbacks are detected within the internals of ngAnimate whenever an
animation starts and ends. In order to allow the user to set callbacks
the callback detection needs to happen during the next tick. Prior to
this fix we used `$$rAF` to do the tick detection, however, with this
patch we intelligently use `$$postDigest` to do that for us and then
only issue a call to `$$rAF` if necessary.

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -890,7 +890,6 @@ describe("animations", function() {
         });
 
         $rootScope.$digest();
-        $animate.flush();
         expect(capturedAnimation).toBeTruthy();
         expect(runner1done).toBeFalsy();
 
@@ -910,7 +909,6 @@ describe("animations", function() {
         });
 
         $rootScope.$digest();
-        $animate.flush();
         expect(capturedAnimation).toBeTruthy();
         expect(runner2done).toBeFalsy();
 
@@ -990,8 +988,6 @@ describe("animations", function() {
         var doneHandler = jasmine.createSpy('addClass done');
         runner.done(doneHandler);
 
-        $animate.flush();
-
         expect(doneHandler).not.toHaveBeenCalled();
 
         $animate.removeClass(element, 'active-class');
@@ -1010,8 +1006,6 @@ describe("animations", function() {
 
         var doneHandler = jasmine.createSpy('addClass done');
         runner.done(doneHandler);
-
-        $animate.flush();
 
         expect(doneHandler).not.toHaveBeenCalled();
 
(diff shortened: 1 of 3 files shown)


matsko:
@petebacondarwin please review.

test/ngAnimate/animateSpec.js
```diff
@@ -1707,6 +1700,69 @@ describe("animations", function() {
...
+      // to fire when the animation ends since
+      // the callback detection happens again
+      expect(spy.callCount).toBe(3);
+
+      function registerCallback() {
+        $animate.on('enter', element, spy);
+      }
+    }));
+
+    it('should use use RAF if there are detected callbacks within the hierarcy of the element being animated',
```
gkalpak: _use use_  -->  _use_
_hierarcy_  -->  _hierachy_

test/ngAnimate/animateSpec.js
```diff
@@ -1707,6 +1700,69 @@ describe("animations", function() {
...
+      // was added after the first callback
+      // was fired
+      expect(spy.callCount).toBe(2);
+
+      spy.reset();
+      runner.end();
+
+      $animate.flush();
+
+      // no we expect all three callbacks
```
gkalpak: _no_  -->  _now_

test/ngAnimate/integrationSpec.js
```diff
@@ -319,6 +319,40 @@ describe('ngAnimate integration tests', function() {
...
+    it('should trigger callbacks at the start and end of an animation',
+      inject(function($rootScope, $rootElement, $animate, $compile) {
+
+      ss.addRule('.animate-me', 'transition:2s linear all;');
+
+      var parent = jqLite('<div></div>');
+      html(parent);
+
+      var element = jqLite('<div ng-if="exp" class="animate-me"></div>');
+      parent.append(element);
```
petebacondarwin: Given that you `find` this element again later on in the test, you might as well tidy up setup here to something like:

```
var element,
    parent = jqLite('<div><div ng-if="exp" class="animate-me"></div></div>');
html(parent);
```

petebacondarwin:
I don't really understand what you are trying to achieve here. Perhaps you could give me anothe explanation?

matsko:
@petebacondarwin this is to avoid having calls to RAF since it's more expensive. Beforehand even if there were no detected callbacks then a RAF would still be issued.

petebacondarwin:
OK. LGTM

matsko:
Landed as https://github.com/angular/angular.js/commit/8b27c3f064b34532ba99d709cadf09fc4c0cbeab

PR #12866, opened 2015-09-16
fix(ngClick): prevent soft keyboard from disappearing on input click

### LINKED ISSUE #11358: ngTouch doesn't work
When clicking on input and using ngTouch to remove the 300ms delay, it doesn't open up a keyboard. 

What happens:
- touchstart event gets triggered and there is no touchend/click event after that, because touchcancel and touchmove are called and they reset the state.  
- keyboard only opens up on click event

v1.3.15 - angularjs

steven10172:
Currently ngClick calls blur on all elements and this causes keyboards
on touch devices to bounce and sometimes not appear. This is a fix
for #11358 and should allow input fields to keep the keyboard visible
without bouncing in and out of view or never appearing.

Closes #11358

### DIFF
--- a/src/ngTouch/directive/ngClick.js
+++ b/src/ngTouch/directive/ngClick.js
@@ -125,6 +125,9 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
       return; // Too old.
     }
 
+    // Get the node name of the target
+    var nodeName = nodeName_(event.target);
+
     var touches = event.touches && event.touches.length ? event.touches : [event];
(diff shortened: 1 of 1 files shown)


steven10172:
Examples:

 - Using ngTouch: http://embed.plnkr.co/edBdpUbM4UWI6a5Pdr7d/preview
 - Without ngTouch: http://embed.plnkr.co/RHm7H2/preview
 - Using pull request: http://embed.plnkr.co/4e0owINFEsWFNOap1SSs/preview

PR #12871, opened 2015-09-16
WIP: transclude element and template elements

lgalfaso:
DO NOT MERGE
This PR is just to open the discussion to get some feedback if this would be an interesting feature.

When a directive of type transclude element is on a <template> element, then the transclude function contains the content of the <template> element and the <template> element is replaced with a comment node.


Narretz:
Would have been cool to explore this more, but alas ... it's too late ;)

PR #12875, opened 2015-09-17
Essential note when cloning into Angular repository

SibuStephen:
1) First change - Whenever cloning into directory there are chances of some users using "sudo" without knowing what it does exactly and what's the purpose of using sudo, leading them into a confusion and this lead to changes in git config and permission issues of which users may not be able to figure out at one spot.
2)  Second Change- There are chances of version unstability  issues in node which at times causes npm error and warnings while installing packages, I too had undergone some issues with node version conflicts at times, users who are not aware of this will be benefited with a Note prior.

### DIFF
--- a/docs/content/misc/contribute.ngdoc
+++ b/docs/content/misc/contribute.ngdoc
@@ -61,7 +61,8 @@ To build AngularJS, you clone the source code repository and use Grunt to genera
 minified AngularJS files:
 
 ```shell
-# Clone your Github repository:
+# Clone your Github repository(Note: Do not go with `sudo` user when cloning into this repository 
+if not known what it does essentially, please refer https://www.linux.com/learn/tutorials/306766:linux-101-introduction-to-sudo):
 git clone "git@github.com:<github username>/angular.js.git"
 
 # Go to the AngularJS directory:
@@ -70,7 +71,7 @@ cd angular.js
 # Add the main AngularJS repository as an upstream remote to your repository:
 git remote add upstream "https://github.com/angular/angular.js.git"
 
-# Install node.js dependencies:
+# Install node.js dependencies(Note: Please check for the stable version of node before installing the essential packages):
 npm install
 
 # Install bower components:
--- a/docs/content/misc/contribute.ngdoc
+++ b/docs/content/misc/contribute.ngdoc
@@ -61,9 +61,11 @@ To build AngularJS, you clone the source code repository and use Grunt to genera
 minified AngularJS files:
 
 ```shell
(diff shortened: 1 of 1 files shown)


SibuStephen:
I signed it

docs/content/misc/contribute.ngdoc
```diff
@@ -61,7 +61,8 @@ To build AngularJS, you clone the source code repository and use Grunt to genera
 minified AngularJS files:
 
 ```shell
-# Clone your Github repository:
+# Clone your Github repository(Note: Do not go with `sudo` user when cloning into this repository
```
Narretz: I don't see it as particularly helpful to tell the user not to use sudo even though this point doesn't mention sudo at all.

SibuStephen: Yes i totally agree with you @Narretz , but won't it be a before hand note to tell the users that using sudo might not be good always, if in case they use it to clone the repo and are not even aware of what exactly it does?

Narretz: We can't really provide a documentation for unix-specific commands. If somebody uses `sudo` to clone the repo, they should know what they are doing.

SibuStephen: @Narretz ok will do, you are right I will remove it, and is my second comment valid? please do provide feedback. Thank you

docs/content/misc/contribute.ngdoc
```diff
@@ -70,7 +71,7 @@ cd angular.js
 # Add the main AngularJS repository as an upstream remote to your repository:
 git remote add upstream "https://github.com/angular/angular.js.git"
 
-# Install node.js dependencies:
+# Install node.js dependencies(Note: Please check for the stable version of node before installing the essential packages):
```
Narretz: What is meant by "check for the stable version of node"? That you should make sure you are running the stable version of node?

docs/content/misc/contribute.ngdoc
Narretz: I think it can't hurt to add the link to the sudo man page to the first mention of sudo on the page on line 49.

SibuStephen: So should I include that and replace with the original Note (Do not go with `sudo` user when cloning into this repository ) ?

docs/content/misc/contribute.ngdoc
SibuStephen: Yes that's what i meant as such, sometimes there are version dependency errors while installing node.js .

Narretz:
I've checked the docs again and both a note for sudo / admin, and the required node version has been added in the meantime.

PR #12881, opened 2015-09-17
Animation Fixes

### LINKED ISSUE #12872: ngAnimate's ng-animate-ref does not work when manual bootstrapping App ( + Plunkr )
The Manual Initialization breaks the __ng-animate-ref animation__.
http://plnkr.co/edit/h6aIddcKKfdZKiDWf1ch?p=preview
Thank you, for your hard work by the way !

### LINKED ISSUE #12874: ngAnimate doesn't work when you use ngView on the body
http://plnkr.co/edit/PgfQaoDE5iCaxV2AX8OA

Example here. Basically it doesn't add the intermediate classes when showing and hiding.

The workaround is pretty simple, just don't use ngView on the body. In the example if you move ngView down onto a div it works fine.

There are a few scattered issues on stackoverflow where it does say you get some odd behaviour when adding ngView to the body but most of them aren't serious. If we shouldn't be doing this, I think there should be a warning in the docs about it.

matsko:
Closes #12874
Closes #12872

### DIFF
--- a/src/ngAnimate/animateCssDriver.js
+++ b/src/ngAnimate/animateCssDriver.js
@@ -9,13 +9,13 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
   var NG_OUT_ANCHOR_CLASS_NAME = 'ng-anchor-out';
   var NG_IN_ANCHOR_CLASS_NAME = 'ng-anchor-in';
 
-  this.$get = ['$animateCss', '$rootScope', '$$AnimateRunner', '$rootElement', '$$body', '$sniffer', '$$jqLite',
-       function($animateCss,   $rootScope,   $$AnimateRunner,   $rootElement,   $$body,   $sniffer,   $$jqLite) {
+  this.$get = ['$animateCss', '$rootScope', '$$AnimateRunner', '$rootElement', '$sniffer', '$$jqLite', '$document',
+       function($animateCss,   $rootScope,   $$AnimateRunner,   $rootElement,   $sniffer,   $$jqLite,   $document) {
 
     // only browsers that support these properties can render animations
     if (!$sniffer.animations && !$sniffer.transitions) return noop;
 
-    var bodyNode = getDomNode($$body);
+    var bodyNode = $document[0].body;
     var rootNode = getDomNode($rootElement);
 
     var rootBodyElement = jqLite(bodyNode.parentNode === rootNode ? bodyNode : rootNode);
--- a/src/ngAnimate/animateCssDriver.js
+++ b/src/ngAnimate/animateCssDriver.js
@@ -9,6 +9,10 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
   var NG_OUT_ANCHOR_CLASS_NAME = 'ng-anchor-out';
   var NG_IN_ANCHOR_CLASS_NAME = 'ng-anchor-in';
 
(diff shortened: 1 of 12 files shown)


src/ngAnimate/animateCssDriver.js
```diff
@@ -9,16 +9,25 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
...
 
-    var bodyNode = getDomNode($$body);
+    var bodyNode = $document[0].body;
     var rootNode = getDomNode($rootElement);
 
-    var rootBodyElement = jqLite(bodyNode.parentNode === rootNode ? bodyNode : rootNode);
+    var rootBodyElement = jqLite(
+      // this is to avoid using something that sites outside of the body
+      // we also special case the doc fragement case because our unit test code
+      // appens the $rootElement to the body after the app has been bootstrapped
```
petebacondarwin: `appens` -> `appends to`

src/ngAnimate/animateCssDriver.js
```diff
@@ -9,16 +9,25 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
...
     // only browsers that support these properties can render animations
     if (!$sniffer.animations && !$sniffer.transitions) return noop;
 
-    var bodyNode = getDomNode($$body);
+    var bodyNode = $document[0].body;
     var rootNode = getDomNode($rootElement);
 
-    var rootBodyElement = jqLite(bodyNode.parentNode === rootNode ? bodyNode : rootNode);
+    var rootBodyElement = jqLite(
+      // this is to avoid using something that sites outside of the body
```
petebacondarwin: `sites`? should this be `lives` or `exists`?

test/ngAnimate/animateCssDriverSpec.js
```diff
@@ -129,8 +129,14 @@ describe("ngAnimate $$animateCssDriver", function() {
...
-          // we need to do this so that style detection works
-          $$body.append($rootElement);
+          var doc = $document[0];
+
+          // there is one test in here that expects the rootElement
+          // to superceed the body node
+          if (!$rootElement[0].contains(doc.body)) {
+            // we need to do this so that style detection works
+            jqLite(doc.body).append($rootElement);
+          }
```
petebacondarwin: Doing this kind of conditional work in `beforeEach` makes me feel uncomfortable.
Actually almost all `beforeEach` blocks in general make me uncomfortable, especially when the spec file has over a thousand lines.
I often find myself looking at a spec that clearly has some magic happening somewhere in the file but there are potentially multiple nested decribe blocks who may or may not provide a `beforeEach`.
As a general rule I am moving more toward a style of helpers, except for very basic setup like getting hold of injectables.

But anyway, not to worry here on this occasion.

petebacondarwin:
Great! I am all for removing private undocumented services.
I see that this is supposed to fix animations not working when `ng-view` is placed on the body.
Can we have a test to show that this fixes that?

Other than that LGTM

gkalpak:
Am I the only one that finds it weird when people are trying to animate multiple `<body>` elements in and out ?

matsko:
Landed as https://github.com/angular/angular.js/commit/240d5896ecdfac2351f9bd6147b52de52c0b7608

I promised @petebacondarwin that I would create a test in a follow-up commit that tests against the `<body ng-view>` issue.

PR #12882, opened 2015-09-18
fix(ngForm): don't block submit when method=dialog

dpogue:
When a form's method is set to "dialog" it does not post to a server but instead closes the nearest parent <dialog> element. An action is not required.

https://html.spec.whatwg.org/multipage/forms.html#submit-dialog

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -364,7 +364,8 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
  * to handle the form submission in an application-specific way.
  *
  * For this reason, Angular prevents the default action (form submission to the server) unless the
- * `<form>` element has an `action` attribute specified.
+ * `<form>` element has a `method` attribute with the value `"dialog"` or has an `action` attribute
+ * specified.
  *
  * You can use one of the following two ways to specify what javascript method should be called when
  * a form is submitted:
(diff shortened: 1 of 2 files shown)


dpogue:
I signed it!

Narretz:
Shocking support for that element: http://caniuse.com/#feat=dialog
But I don't think it'll hurt, as someone who uses `method="dialog"` will know what they're doing. There's still some work to do for this though:
- Add a test for this
- Add a line to the docs that says submit will not be blocked when method is set to dialog.

src/ng/directive/form.js
```diff
@@ -485,7 +485,7 @@ var formDirectiveFactory = function(isNgForm) {
             var controller = ctrls[0];
 
             // if `action` attr is not present on the form, prevent the default action (submission)
-            if (!('action' in attr)) {
+            if (!('action' in attr) && attr['method'] !== 'dialog') {
```
Narretz: What happens when the action attr is set and method is set to dialog?

dpogue: According to the spec, if method is dialog then it closes the dialog and stops further processing. The action attribute should be ignored by the browser in that case.

Narretz: What I meant is: if the action attribute is set and the method is set to dialog, your code change would not prevent the submit. Isn't `!('action' in attr) || attr['method'] !== 'dialog' what we want?

dpogue:
I've added a test and updated the comments. I'm not sure that blocking submit when method="dialog" with an action is what we want? It should be up to the browser to handle that properly.

src/ng/directive/form.js
Narretz: Ah, I think I misunderstood the purpose. I though it should block submit when the method is dialog, doh. Having the test really helps understanding what a change is about.

Narretz:
Thanks for the test. Now that method="dialog" does not block submit anymore, I think we'd still want to have it call $commitViewValue and $setSubmitted on the form, do you agree? Previously, it would set these only if the form wasn't submitted, i.e. sent to the server. Now the form isn't send to the server, so I assume we still want that.

PR #12885, opened 2015-09-18
Reword for grammar

earthrid:
Reworded to correct grammar.


earthrid:
I've not signed it. I'll cancel this pull request in a while. Please feel free to incorporate the suggested changes if they are of use.

PR #12887, opened 2015-09-18
feat(ngMock): invoke nested calls to module immediately

shahata:
before this change calling a nested module was just ignored, but it makes sense that ppl might use module() inside some helper function in order to get providers from DI and those helper methods should work no matter if they are called before or during the injector creation.

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2282,16 +2282,21 @@ if (window.jasmine || window.mocha) {
       if (currentSpec.$injector) {
         throw new Error('Injector already created, can not register a module!');
       } else {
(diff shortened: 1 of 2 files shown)


shahata:
This actually worked in some weird way in angular 1.2 (nested modules were invoked last) since angular.forEach allowed adding items to array during iteration. I noticed it when migrating some app from 1.2 to 1.4, and it was actually really hard to create a workaround without making the code ugly: http://plnkr.co/edit/4XGPuQQGfXBGmMivgbio?p=preview

shahata:
@lgalfaso or @petebacondarwin can you plz take a look?

shahata:
Ping

petebacondarwin:
LGTM - merging

PR #12890, opened 2015-09-18
fix(injector): support arrow functions with no parenthesis

shahata:


shahata:
I initially upgraded jscs so it will parse es6 syntax correctly, but since eventually I used `eval` in order to not break old browsers, it is not really needed. I think we can upgrade jscs anyway, so I'm leaving this commit as is.

src/auto/injector.js
```diff
@@ -62,17 +62,23 @@
  * Implicit module which gets automatically added to each {@link auto.$injector $injector}.
  */
 
+var ARROW_ARG = /^([^\(]+)=>/;
```
gkalpak: This RegExp would incorrectly match `arg1 => { var arrow = '=>'; }` (it would match greedily until the last `=>` before the first `(`).
The `+` should be made non-greedy: `/^([^(]+?)=>/`

src/auto/injector.js
```diff
@@ -62,17 +62,23 @@
...
+}
+
 function anonFn(fn) {
   // For anonymous functions, showing at the very least the function signature can help in
   // debugging.
-  var fnText = fn.toString().replace(STRIP_COMMENTS, ''),
-      args = fnText.match(FN_ARGS);
+  var args = extractArgs(fn);
   if (args) {
     return 'function(' + (args[1] || '').replace(/[\s\r\n]+/, ' ') + ')';
```
gkalpak: I am not sure if it's worth fixing, but the signature won't be correct for fat-arrow functions.

gkalpak:
I left a couple of comments.
LGTM other than that.

src/auto/injector.js
shahata: :+1:

src/auto/injector.js
shahata: Yeah, I saw it too. I don't think it matters.

shahata:
Why isn't this running in travis? :(

gkalpak:
It did run initially I think (https://travis-ci.org/angular/angular.js/builds/81241841).
I've seen it too in other PRs. I might happen when you force-push an amended commit.

gkalpak:
Did someone restartd it or did it start on its own ?

shahata:
I created another pull request on a cloned branch in hopes of somehow triggering travis and it seems to have worked :)

shahata:
Okay, Travis finally approves. I'll merge this tomorrow morning unless anyone else has more comments

lgalfaso:
@shahata can you please squash c2e0ecfe093f99f25fc303dbc3f09ff4323401a9 427db62ec116c65d4ae45ea0618523a10f1048d4 into one commit. It is expected for some refactor to happen as part of a fix. The other commits are fine as is

PR #12892, opened 2015-09-19
fix($parse): do not convert to string computed properties multiple times

lgalfaso:
Do not convert to string properties multiple times.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -38,20 +38,30 @@ var $parseMinErr = minErr('$parse');
 
 
 function ensureSafeMemberName(name, fullExpression) {
+  if (name === "__defineGetter__" || name === "__defineSetter__"
+      || name === "__lookupGetter__" || name === "__lookupSetter__"
+      || name === "__proto__") {
+    throw $parseMinErr('isecfld',
+        'Attempting to access a disallowed field in Angular expressions! '
+        + 'Expression: {0}', fullExpression);
+  }
+  return name;
+}
+
+function getStringValue(name, fullExpression) {
   // From the JavaScript docs:
   // Property names must be strings. This means that non-string objects cannot be used
   // as keys in an object. Any non-string object, including a number, is typecasted
   // into a string via the toString method.
   //
   // So, to ensure that we are checking the same `name` that JavaScript would use,
-  // we cast it to a string, if possible
-  name =  (isObject(name) && name.toString) ? name.toString() : name;
-
-  if (name === "__defineGetter__" || name === "__defineSetter__"
-      || name === "__lookupGetter__" || name === "__lookupSetter__"
-      || name === "__proto__") {
-    throw $parseMinErr('isecfld',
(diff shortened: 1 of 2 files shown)


src/ng/parse.js
```diff
@@ -1187,6 +1200,10 @@ ASTCompiler.prototype = {
     return 'ensureSafeFunction(' + item + ',text)';
   },
 
+  getStringValue: function(item) {
+    this.assign(item, 'getStringValue(' + item + ',text)');
```
petebacondarwin: should this be returning the result of the assign?

src/ng/parse.js
```diff
@@ -38,20 +38,30 @@ var $parseMinErr = minErr('$parse');
...
 function ensureSafeMemberName(name, fullExpression) {
+  if (name === "__defineGetter__" || name === "__defineSetter__"
+      || name === "__lookupGetter__" || name === "__lookupSetter__"
+      || name === "__proto__") {
+    throw $parseMinErr('isecfld',
+        'Attempting to access a disallowed field in Angular expressions! '
+        + 'Expression: {0}', fullExpression);
+  }
+  return name;
+}
```
petebacondarwin: Have you checked that `ensureSafeMember` is never called with an unsafe name "thing"? I.E. That `getStringValue` is always called when needed?

src/ng/parse.js
lgalfaso: the value is replaced inline, this is, `item` will have the new value.

test/ng/parseSpec.js
```diff
@@ -2692,6 +2692,15 @@ describe('parser', function() {
           });
         });
 
+       it('should prevent the exploit', function() {
```
petebacondarwin: Is this message purposefully vague right now?

src/ng/parse.js
lgalfaso: ensureSafeMemberName is called in three cases.
* retrieving an identifier. Eg `foo`
* retrieving a non-computed member. Eg `foo.bar`
* retrieving a computed member. Eg `foo[bar]`

getStringValue is only used in the later. This is the expected behavior

test/ng/parseSpec.js
lgalfaso: If you have a better name, then I am ok with it. I do not know if the exploit can be explained in a one line.

petebacondarwin:
Marvellous. LGTM - please merge.

lgalfaso:
closed with 20cf7d5e3a0af766b1929e24794859c79439351c

lgalfaso:
@petebacondarwin merged this at master and v1.4.x branch. Do you think this should this be backported to prior versions?

petebacondarwin:
Yes.

lgalfaso:
backport to 1.2 with afb65c11e5e3425f8431ad5b82308ff6b8ecbc64
backport to 1.3 with 9d5ac2eb84f21159ff93dac6f3d5782988a5f515

petebacondarwin:
Good work!

PR #12895, opened 2015-09-20
fix(injector): support arrow functions with no parenthesis

shahata:
I initially upgraded jscs so it will parse es6 syntax correctly, but since eventually I used eval in order to not break old browsers, it is not really needed. I think we can upgrade jscs anyway, so I'm leaving this commit as is.


shahata:
closing in favor of #12890

PR #12896, opened 2015-09-20
Unicode identifiers at $parse

lgalfaso:
This PR has two commits:
- Adds the ability to use user-defined functions for the identifier start and identifier continue characters for an identifier in an Angular expression
- Adds a new optional module that extends $parse to allow all ES6 identifiers in an Angular expression


lgalfaso:
Updated the PR with the comments from @gkalpak

gkalpak:
Ooops, I just noticed I was leaving the comments on the commit (80c77d3), instead of the PR :disappointed: 
Taking another look...

src/ng/parse.js
```diff
@@ -185,12 +185,51 @@ Lexer.prototype = {
...
+        this.options.isIdentifierContinue(ch, this.codePointAt(ch)) :
+        this.isValidIdentifierContinue(ch);
+  },
+
+  isValidIdentifierContinue: function(ch) {
+    return this.isValidIdentifierStart(ch) || this.isNumber(ch);
+  },
+
+  codePointAt: function(ch) {
+    if (ch.length == 1) return ch.charCodeAt(0);
```
gkalpak: <sub>Nitpicking: `===`</sub>

src/ng/parse.js
```diff
@@ -1705,16 +1750,51 @@ function getValueOf(value) {
...
+  *
+  * Since this function will be called extensively, keep the implementation of these functions fast,
+  * as the performance of these functions have a direct impact on the expressions parsing speed.
+  *
+  * @param {function=} identifierStart The function that will decide whether the given character is
+  *   a valid identifier start character.
+  * @param {function=} identifierContinue The function that will decide whether the given character is
+  *   a valid identifier continue character.
+  */
+  this.setValidIdentifierCharacters = function(identifierStart, identifierContinue) {
```
gkalpak: I wonder if we could find a better name, because reading the current name, I expect to literally set the characters (e.g. an array of valid characters); but I'm not good with names anyway :)

src/ng/parse.js
```diff
@@ -1705,16 +1750,51 @@ function getValueOf(value) {
...
+  };
 
   this.$get = ['$filter', function($filter) {
     var noUnsafeEval = csp().noUnsafeEval;
     var $parseOptions = {
           csp: noUnsafeEval,
-          expensiveChecks: false
+          expensiveChecks: false,
+          isIdentifierStart: isFunction(identStart) && identStart,
+          isIdentifierContinue: isFunction(identContinue) && identContinue
```
gkalpak: Extremely minor, but the isFunction() checks could be done inside setValidIdentifierCharacter().

test/ng/parseSpec.js
```diff
@@ -104,6 +104,37 @@ describe('parser', function() {
       expect(spaces).toEqual(noSpaces);
     });
 
+    it('should use callback functions to know when an identifier is valid', function() {
+      function getText(t) { return t.text; }
+      var isIdentifierStart = jasmine.createSpy("start");
+      var isIdentifierContinue = jasmine.createSpy("continue");
```
gkalpak: <sub>Nitpicking: Single quotes for consistency (here and below).</sub>

gkalpak:
I left a couple of minor comments, but it basically LGTM.
 :+1: for `ngParseExt`.

One last thing:
`this.peekMultichar()` is called several times on the same index (e.g. it's called in `isIdentifierStart()` and then again in `readIdent()` to get it's length; same inside `readIdent()` on lines [283][1] and [286][2]). We could save a few calls, but is it worth the impact on readability ?

[1]: https://github.com/lgalfaso/angular.js/commit/2db70dace23b6900fce73e0d9eccbbcc9d027178#diff-780de070ede30180f6aedb6c5f7d57caR283
[2]: https://github.com/lgalfaso/angular.js/commit/2db70dace23b6900fce73e0d9eccbbcc9d027178#diff-780de070ede30180f6aedb6c5f7d57caR286

src/ng/parse.js
lgalfaso: the reason for doing the check here is that it should be possible to set the valid identifier functions to `undefined` to bring back the default behavior

lgalfaso:
fixed the remaining comments and partially resolved the issue of calling multiple times `peekMutichar` (now the calls only happen multiple times for the first character in the identifier).
The only comment that I did nothing is the one on `isFunction` inside `setValidIdentifierCharacter` as I think that the current behavior is best

src/ng/parse.js
lgalfaso: is `setIdentifierValidators` a better name?

petebacondarwin:
Note that creating a new module requires creation of a `bower-angular-...` project in GH and also publication to npm and bower. The project needs to contain a specific package.json, which can be "copied" from one of the other projects.

src/ng/parse.js
```diff
@@ -1705,16 +1749,51 @@ function getValueOf(value) {
...
   var cacheExpensive = createMap();
+  var identStart, identContinue;
+
+ /**
+  * @ngdoc method
+  * @name ng.$parseProvider#setValidIdentifierCharacters
+  * @methodOf ng.$parseProvider
+  * @description
+  *
+  * Allows defining the set of characters that allowed in Angular expressions. The function
```
petebacondarwin: that **are** allowed

src/ng/parse.js
petebacondarwin: I think we could simply call this method: `setIdentifierFns` or possibly `setIdentifierMatchers`; although the second one sounds like we should be providing RegExps. I agree that `setValidIdentifierCharacters` does not sound right.

petebacondarwin: To be honest we could also just have one for each:

```
setIdentifierStartFn(fn);
```

and

```
setIdentifierContinueFn(fn);
```

Although I accept that is may be rare to change only one of these.

src/ngParseExt/ucd.js
```diff
@@ -0,0 +1,1217 @@
+/******************************************************
+ *         Generated file, do not modify              *
```
petebacondarwin: You don't say!

src/ng/parse.js
petebacondarwin: I would go the other way and suggest that we move the conditional statement from out of the `isIdentifierStart` and `isIdentifierContinue`, around lines 188, and provide the built in one in the options here, if not provided. That would speed up those methods. No?

petebacondarwin:
A couple of minor comments but generally LGTM. I think it would be good for @IgorMinar to have a quick look too for good measure.

petebacondarwin:
Oh and we are missing some tests for `ngParseExt`

src/ng/parse.js
lgalfaso: I think it would be really odd to change only one, so will just rename it to `setIdentifierFns`

src/ng/parse.js
lgalfaso: I want to avoid calling `codePointAt` when using the build-in function, just making the function call would be slower than the if

src/ngParseExt/ucd.js
lgalfaso: after too many PRs on the locale files, we have to try something new :)

lgalfaso:
Sorry for taking this long, finally had the time to update this and to write the tests.

lgalfaso:
@petebacondarwin PTAL

petebacondarwin:
LGTM

- [ ] add some docs for the new module
- [ ] setup new GH repo and npm package for the new module

lgalfaso:
@petebacondarwin Created the docs for the new module, and fixed some doc issues. I do not have the permissions to create a new repo for `bower-parse-ext` nor the permissions to create the bower/npm repo in the name of `angularcore`. Can you please take care of this?

petebacondarwin:
Will do. Thanks

petebacondarwin:
Repository created at https://github.com/angular/bower-angular-parse-ext
The npm package should be created by the CI server when we release 1.5.4 (I think)

lgalfaso:
@petebacondarwin can I land this in master and v1.5.x ?

petebacondarwin:
Yep

petebacondarwin:
But @lgalfaso please add this repo to the https://github.com/angular/angular.js/blob/master/scripts/bower/repos.inc file too

lgalfaso:
Made the change to https://github.com/angular/angular.js/blob/master/scripts/bower/repos.inc in master and 1.5, and landed everything in both branches

petebacondarwin:
Doh! I gave the CI bot access to the new bower repo but only READ not WRITE!!

petebacondarwin:
I have fixed that now and so it should work on the next commit

PR #12899, opened 2015-09-21
Fix call angular.isUndefined in $cacheFactory doc

idoo:


idoo:
@googlebot I signed it!

Narretz:
These functions are usually available without `angular` while in the main angular closure. Did you experience them being undefined?

Narretz:
Oh I see now it's in the docs. Never mind my comment then.

gkalpak:
@Narretz, the code belongs to an example (in the docs), so it **is** needed.

idoo:
@gkalpak it's not working in current version
https://docs.angularjs.org/api/ng/service/$cacheFactory

Narretz:
Thx @gkalpak. Would also be nice to backport this to 1.4.

Narretz:
@idoo You have to give Travis about 20 minutes to complete the tests etc. Only then is the site updated.

gkalpak:
Backported to `v1.4.x` as 808f984.

PR #12903, opened 2015-09-21
fix(ngMessages): prevent race condition with ngAnimate

### LINKED ISSUE #12856: Race condition when using ngMessages with $asyncValidators
The issue is that $asyncValidators will set the validity to undefined while it fulfills its promise, which will cause the error to briefly go away. At this point, ngMessage's render method will process the $errors (which will be {}), and will detach the messageCtrl. But before that is complete, the async validator finishes and sets the error back to true, and ngMessage's render method is run. Then $destroy handler is invoked and ngMessage will deregister the message node, causing the message to go away and never come back.

See the error in action:

https://gist.github.com/anonymous/1bd3f9b6cb17498e73c0

petebacondarwin:
If `ngMessage` tried to add a message back in that was about to be removed
after an animation, the NgMessageController got confused and tried to detach
the newly added message, when the pending node was destroyed.

This change applies a unique `attachId` to the message object and its DOM
node when it is attached. This is then checked when a DOM node is being
destroyed to prevent unwanted calls to `detach`.

Closes #12856

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -325,6 +325,9 @@ angular.module('ngMessages', [])
        controller: ['$element', '$scope', '$attrs', function($element, $scope, $attrs) {
          var ctrl = this;
          var latestKey = 0;
+         var nextAttachId = 0;
+
+         this.getAttachId = function getAttachId() { return nextAttachId++; };
 
          var messages = this.messages = {};
          var renderLater, cachedCollection;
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@petebacondarwin - this fixes the issues of destroying/removing the appropriate node(s). Should we also think about **reusing** node elements that will be destroyed and re-added? Perhaps this is an optimization that is not worth the complexity?

petebacondarwin:
@ThomasBurleson:

So we are currently reusing the message controller objects but not the message nodes. This was the main cause for the problem since we would associate both the node that was being removed and the node that was being added with the same message controller object.

The trouble is that the old node has been already scheduled for removal ("leave") via `ngAnimate` by the time that we realise that it could be reused. So without some changes inside `ngAnimate` to allow such nodes to be "recovered", it would be rather tricky.

topherfangio:
@petebacondarwin Just tested and this does appear to fix the issue in material. Thanks so much!

ThomasBurleson:
@petebacondarwin - thx for the explanation.

petebacondarwin:
@matsko can you cast your eye over this fix before I merge, please?

petebacondarwin:
Cherry-picked to 1.4.x too

PR #12909, opened 2015-09-21
docs: fix link formatting on directive documentation

mlh758:
The link was showing up as [text](link) rather than actually displaying the formatted link. I removed the bold formatting so it display as a link again.


Narretz:
Hi, thanks for pointing this out. It looks like the problem was that the url was on a newline, so the markdown processor didn't pick it up. I changed that, and so we now have a bold link.

gkalpak:
:+1:

mlh758:
That works for me, thanks for fixing it either way!

PR #12914, opened 2015-09-22
Add missing mozRequestAnimationFrame fallback

hypery2k:
see #6535


Narretz:
We should not add this back. See https://github.com/angular/angular.js/issues/6535#issuecomment-143514836 
Why do you think we should add it?

hypery2k:
I see, but then we not also remove it from cancelAnimationFrame?

Narretz:
I don't see it in cancelAnimationFrame ..

PR #12918, opened 2015-09-22
WIP - feat(ngPattern): parse "stringified" regular expressions to `RegExp` Objects

### LINKED ISSUE #12913: Validation with passed pattern from variable doesn't work with modifiers
When you want to pass regular expression as validation pattern for input from variable it works well if you don't use slashes (delimeters) and modifiers, but when you try pass regexp with modifiers (as i, g ..) validation is broken. 

See [example fiddle](http://jsfiddle.net/champion/wmnegLdz/)

gkalpak:
Previously, `ngPattern`'s expression value was expected to be a `RegExp` Object. This made it difficult to use string representations of regular expressions (e.g. fetched from the backend) as values for `ngPattern` expressions.

This commit adds the ability to recognize and parse string values into `RegExp` Objects. If the value cannot be parsed into a `RegExp` Object, it behaves the same as before.

Fixes #12913

---

This is a POC to get some feedback/see what people think.

TODO:
- [ ] Update the docs.
- [ ] Add a BC notice to the commit message (if necessary).
## 

Possible BREAKING CHANGE:

Assuming `ng-pattern="pat"` and `$scope.pat = '/123/'`:

**Before:**
Would match a literal `/123/` as input value.

**After:**
Will be parsed into a `RegExp` and match any value containing `123`.
In order to match a literal `/123/` as input value, one can use `$scope = '\/123/'`.

### DIFF
--- a/src/ng/directive/attrs.js
+++ b/src/ng/directive/attrs.js
@@ -390,19 +390,43 @@ forEach(ALIASED_ATTR, function(htmlAttr, ngAttr) {
     return {
       priority: 100,
       link: function(scope, element, attr) {
-        //special case ngPattern when a literal regular expression value
-        //is used as the expression (this way we don't have to watch anything).
-        if (ngAttr === "ngPattern" && attr.ngPattern.charAt(0) == "/") {
-          var match = attr.ngPattern.match(REGEX_STRING_REGEXP);
-          if (match) {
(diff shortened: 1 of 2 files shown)


gkalpak:
This didn't gain any traction. Closing.

PR #12919, opened 2015-09-22
feat(ngMock): improve $httpBackend mock

### LINKED ISSUE #12399: $httpBackend mock should allow for testing that a particular request has not been made
$httpBackend mock right now allows for 

```javascript
$httpBackend.expect***
```
as a matcher to assert that something has been called.

However, a negative matcher is missing. We need 
```javascript
$httpBackend.expectNone***
```
 or a similar method that allows us to EXPLICITLY assert that a specific request has not been made. Specially when we want a few requests to be made and exclude some.

This might mean change in $http.flush() method which currently throws error when no requests have been made at all. If this is a valid use case, an alternate behaviour might be set a flag instead of throwing an error as it does now.

KostyaDanovsky:
In $httpBackend mock were no possibility to expect that specific function won't be called.
Added haveNotBeenCalled function to the $httpBackend mock expectation.

Closes #12399

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1136,6 +1136,7 @@ angular.mock.$HttpBackendProvider = function() {
 function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
   var definitions = [],
       expectations = [],
+      notCalledExpectations = [],
       responses = [],
       responsesPush = angular.bind(responses, responses.push),
       copy = angular.copy;
@@ -1185,6 +1186,10 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
           }
         }
       }
+    }
+
+    if (notCalledExpectations.some(function(exp) { return exp.match(method, url, data, headers); })) {
+      throw new Error('Request should not be called: ' + method + ' ' + url);
     }
 
     if (expectation && expectation.match(method, url)) {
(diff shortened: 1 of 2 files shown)


KostyaDanovsky:
Hey @lgalfaso, I did a fix. Does it need for you? ))

src/ngMock/angular-mocks.js
```diff
@@ -1394,6 +1399,14 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
           respond: function(status, data, headers, statusText) {
             expectation.response = createResponse(status, data, headers, statusText);
             return chain;
+          },
+          haveNotBeenCalled: function() {
```
frankweb: I'd like to use `hasNotBeenCalled` instead of `haveNotBeenCalled`.

src/ngMock/angular-mocks.js
```diff
@@ -1187,6 +1188,10 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
       }
     }
 
+    if (notCalledExpectations.some(function(exp) { return exp.match(method, url, data, headers); })) {
+      throw new Error('Request should not be called: ' + method + ' ' + url);
```
frankweb: I'm not sure whether this is necessary, but you could include data and headers in the error description.

KostyaDanovsky:
@frankweb thanks. I agree with your comments, I'll improve it.

petebacondarwin:
See https://github.com/angular/angular.js/issues/12399#issuecomment-174577924

PR #12923, opened 2015-09-23
docs(guide/Conceptual Overview): change "factory" to "constructor"

alirezamirian:
Controller functions are treated as **constructor** functions by angular, not **factory** functions and
controllers instances are created by instantiating their registered **constructor** function

### DIFF
--- a/docs/content/guide/concepts.ngdoc
+++ b/docs/content/guide/concepts.ngdoc
@@ -274,7 +274,7 @@ The code snippet `angular.module('invoice2', ['finance2'])`  specifies that the 
 `finance2` module. By this, Angular uses the `InvoiceController` as well as the `currencyConverter` service.
 
 Now that Angular knows of all the parts of the application, it needs to create them.
-In the previous section we saw that controllers are created using a factory function.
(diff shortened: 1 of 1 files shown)


Narretz:
Aren't thje docs saying the controllers are created by a controller factory? Sure, the actual controller is defined by a constructor, but it is created by a factory.

alirezamirian:
From [docs](https://docs.angularjs.org/guide/controller):
> In Angular, a Controller is defined by a JavaScript **constructor function**

I'm not sure if the module API `.controller` ( which is used for **registering** a controller inside some module, not **creating** it) is being called **factory**.
If this is the case, I still don't feel right about it.

Narretz:
Then maybe the sentence should be:
> In the previous section we saw that controllers are registered using a constructor function.

Regardless what the change is, the next sentence will feel unconnected if we change factory to constructor. So we should take that into account.

Narretz:
You were right here - this was chnaged here https://github.com/angular/angular.js/commit/88f3517db2a66b3b8a1a5888e07ccca7c6604c01

PR #12907, opened 2015-09-21
wip: module.component helper

btford:
This is taken right from https://github.com/angular/angular/blob/master/modules/angular1_router/test/integration/lifecycle_hook_spec.js#L451-L476

This only includes functionality for adding route configuration and lifecycle hooks to directive components, and does not include best practices or expose all of the options for directives.

I'm going to hand this off to @shahata to finish. It needs tests, docs, and improvements from the looooong thread in #10007.

Closes #10007

Issue #12007, opened 2015-06-02
ngAnimate disrupts correct validation of input type email


JPtenBerge:
Hi there,

With the newest version (1.4.0) it seems that email validation does not fully work correctly when the ngAnimate module gets included. Check out this Plnkr: http://plnkr.co/edit/mRvFhnpgZCKTI1YPWqkT?p=preview.

Try entering a value like '@@@', '|||' or 'qqq'. The model value only gets updated when a true valid value is entered, but the ng-valid/ng-invalid classes get applied wrongly. Removing the ngAnimate module as a dependency and the ng-valid/invalid classes work just fine.

I'm experiencing this on Windows 7 and 8, in IE as well as Chrome and Firefox.

JP

Narretz:
@matsko Wasn't there a similar bug that was already fixed? Or is this the same problem?

Narretz:
This seems to be the same issue: https://github.com/angular/angular.js/issues/11652

gkalpak:
It is quite similar, but I don't think it is the same issue, because #11652 has been fixed ([demo][1]), but this one is still [reproducible on master][2].

Also, the wrong behaviour is different: This one seems to incorrectly apply the valida classes every 3 keystrokes, even if the value is invalid. Weird and interesting :)

[1]: http://jsbin.com/tiwijowiza/3/edit
[2]: http://plnkr.co/edit/3YyAnVQWT5E7NImPdB2e?p=preview

matsko:
This is now fixed in master courtesy of @sreeramu.

matsko:
1.4.4 will land next week as well.

PR #9137, opened 2014-09-17
feat(compile): add human readable alternative to '@=&?'

shahata:
support using `interpolate:<name>`, `bind:<name>`, `eval:<name>` and `bind:optional:<name>` instead of `@<name>`, `=<name>`, `&<name>` and `=?<name>`

Closes #9125

PR #12933, opened 2015-09-24
feat(loader): add convenience method for creating components

### LINKED ISSUE #10007: Feature Request: angular.component helper

shahata:
The proposed syntax:

``` js
angular.module('myApp').component('myComponent', {
  template: '<div></div>', //string | Function, default: ''
  templateUrl: 'a.html', //string | Function, default: undefined
  controller: MyCtrl, //string | Function, default: function(){}
  controllerAs: 'vm', //string, default: component name
  transclude: false, //boolean | object, default: true
  isolate: false, //boolean, default: true (scope: {} .vs. scope: true)
  bindings: {abc: '@'}, //object, default: {} (passed to bindToController)
  $canActivate: MyCtrl.canActivate, //Function, default: undefined (passed to factory.$canActivate)
  $routeConfig: MyCtrl.routeConfig, //RouteConfig, default: undefined (passed to factory.$routeConfig)
});
```

This means that for the most common use-case of simple component ppl will only need to pass template, controller and bindings.

I've discussed this with @btford and we've decided to change a couple of things from #12907:
1. Controller lifecycle hooks (onActivate, onDeactivate, etc.) are not  going to be part of the component options since it makes more sense that the developer will just put those as methods of his controller (this way those methods have access to controller's state)
2. In case developer passes a function for template/templateUrl, this function is injectable with locals of $element and $attrs. This is important because today if you pass a function for template/templateUrl in DDO, you get element as attrs, but for DI you rely on the injectable DDO factory. In the component helper you don't have a factory, therefore we make all functions that can be passed in options injectable.

Docs are still missing obviously, but I would like to get some feedback before moving on.

Closes #10007

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
 
           /**
            * @ngdoc method
+           * @name angular.Module#component
+           * @module ng
+           * @param {string} name Name of the component in camel-case (i.e. myComp which will match as my-comp)
+           * @param {Object} options Component definition object, has the following properties (all optional):
+           *
+           *    - `controller` – `{(string|function()=}` – Controller fn that should be associated with
+           *      newly created scope or the name of a {@link angular.Module#controller registered
+           *      controller} if passed as a string.
+           *    - `controllerAs` – `{string=}` – An identifier name for a reference to the controller.
+           *      If present, the controller will be published to scope under the `controllerAs` name.
+           *      If not present, this will default to be the same as the component name.
+           *    - `template` – `{string=|function()=}` – html template as a string or a function that
+           *      returns an html template as a string which should be used as the contents of this component.
+           *
+           *      If `template` is a function, then it is {@link auto.$injector#invoke injected} with
+           *      the following locals:
+           *
+           *      - `$element` - Current element
+           *      - `$attrs` - Current attributes object for the element
+           *
+           *    - `templateUrl` – `{string=|function()=}` – path or function that returns a path to an html
+           *      template that should be used  as the contents of this component.
+           *
+           *      If `templateUrl` is a function, then it is {@link auto.$injector#invoke injected} with
+           *      the following locals:
+           *
+           *      - `$element` - Current element
+           *      - `$attrs` - Current attributes object for the element
+           *    - `transclude` – `{boolean=}` – whether {@link $compile#transclusion transclusion} is enabled.
+           *      enabled by default.
+           *    - `isolate` – `{boolean=}` – whether the new scope is isolated. Isolated by default.
+           *    - `bindings` – `{object=}` – define DOM attribute binding to component properties.
+           *      component properties are always bound to the component controller and not to the scope.
+           *    - `$canActivate` – `{function()=}` – TBD.
+           *    - `$routeConfig` – `{object=}` – TBD.
+           *
+           * @description
+           * Register a component definition with the compiler. This is short for registering a directive
+           * where its definition object is isolated, allows transclusion and bound to controller as the
+           * component name by default.
+           * See {@link ng.$compileProvider#directive $compileProvider.directive()}.
+           */
+          component: function(name, options) {
+            function factory($injector) {
+              function makeInjectable(fn) {
+                if (angular.isFunction(fn)) {
+                  return function(tElement, tAttrs) {
+                    return $injector.invoke(fn, this, {$element: tElement, $attrs: tAttrs});
+                  };
+                } else {
+                  return fn;
+                }
+              }
+
+              var template = (!options.template && !options.templateUrl ? '' : options.template);
+              return {
+                controller: options.controller || function() {},
+                controllerAs: identifierForController(options.controller) || options.controllerAs || name,
+                template: makeInjectable(template),
+                templateUrl: makeInjectable(options.templateUrl),
+                transclude: options.transclude === undefined ? true : options.transclude,
+                scope: options.isolate === false ? true : {},
+                bindToController: options.bindings || {}
+              };
+            }
+
+            if (options.$canActivate) {
+              factory.$canActivate = options.$canActivate;
+            }
+            if (options.$routeConfig) {
+              factory.$routeConfig = options.$routeConfig;
+            }
+            factory.$inject = ['$injector'];
+
+            return moduleInstance.directive(name, factory);
+          },
+
+          /**
+           * @ngdoc method
            * @name angular.Module#config
            * @module ng
            * @param {Function} configFn Execute this function on module load. Useful for service
--- a/src/loader.js
+++ b/src/loader.js
@@ -287,16 +287,19 @@ function setupModuleLoader(window) {
            * @name angular.Module#component
            * @module ng
            * @param {string} name Name of the component in camel-case (i.e. myComp which will match as my-comp)
-           * @param {Object} options Component definition object, has the following properties (all optional):
+           * @param {Object} options Component definition object (a simplified
+           *    {@link ng.$compile#directive-definition-object directive definition object}),
+           *    has the following properties (all optional):
            *
-           *    - `controller` – `{(string|function()=}` – Controller fn that should be associated with
-           *      newly created scope or the name of a {@link angular.Module#controller registered
-           *      controller} if passed as a string.
+           *    - `controller` – `{(string|function()=}` – Controller constructor function that should be
+           *      associated with newly created scope or the name of a {@link ng.$compile#-controller-
+           *      registered controller} if passed as a string. Empty function by default.
            *    - `controllerAs` – `{string=}` – An identifier name for a reference to the controller.
            *      If present, the controller will be published to scope under the `controllerAs` name.
            *      If not present, this will default to be the same as the component name.
            *    - `template` – `{string=|function()=}` – html template as a string or a function that
            *      returns an html template as a string which should be used as the contents of this component.
+           *      Empty string by default.
            *
            *      If `template` is a function, then it is {@link auto.$injector#invoke injected} with
            *      the following locals:
@@ -312,18 +315,47 @@ function setupModuleLoader(window) {
            *
            *      - `$element` - Current element
            *      - `$attrs` - Current attributes object for the element
-           *    - `transclude` – `{boolean=}` – whether {@link $compile#transclusion transclusion} is enabled.
-           *      enabled by default.
-           *    - `isolate` – `{boolean=}` – whether the new scope is isolated. Isolated by default.
-           *    - `bindings` – `{object=}` – define DOM attribute binding to component properties.
-           *      component properties are always bound to the component controller and not to the scope.
+           *    - `bindings` – `{object=}` – Define DOM attribute binding to component properties.
+           *      Component properties are always bound to the component controller and not to the scope.
+           *    - `transclude` – `{boolean=}` – Whether {@link $compile#transclusion transclusion} is enabled.
+           *      Enabled by default.
+           *    - `isolate` – `{boolean=}` – Whether the new scope is isolated. Isolated by default.
+           *    - `restrict` - `{string=}` - String of subset of {@link ng.$compile#-restrict- EACM} which
+           *      restricts the component to specific directive declaration style. If omitted, this defaults to 'E'.
            *    - `$canActivate` – `{function()=}` – TBD.
            *    - `$routeConfig` – `{object=}` – TBD.
            *
            * @description
-           * Register a component definition with the compiler. This is short for registering a directive
-           * where its definition object is isolated, allows transclusion and bound to controller as the
-           * component name by default.
+           * Register a component definition with the compiler. This is short for registering a specific
+           * subset of directives which represents actual UI components in your application. Component
+           * definitions are very simple and do not require the complexity behind defining directives.
+           * Component definitions usually consist only of the template and the controller backing it.
+           * In order to make the definition easier, components enforce best practices like controllerAs
+           * and default behaviors like scope isolation, restrict to elements and allow transclusion.
+           *
+           * Here are a few examples of how you would usually define components:
+           *
+           * ```js
+           *   angular.module(...).component.('myComp', {
+           *     template: '<div>My name is {{myComp.name}}</div>',
+           *     controller: function MyCtrl() {
+           *       this.name = 'shahar';
+           *     }
+           *   });
+           *
+           *   angular.module(...).component.('myComp', {
+           *     template: '<div>My name is {{myComp.name}}</div>',
+           *     bindings: {name: '@'}
+           *   });
+           *
+           *   angular.module(...).component.('myComp', {
+           *     templateUrl: 'views/my-comp.html',
+           *     controller: 'MyCtrl as ctrl',
+           *     bindings: {name: '@'}
+           *   });
+           *
+           * ```
+           *
            * See {@link ng.$compileProvider#directive $compileProvider.directive()}.
            */
           component: function(name, options) {
@@ -346,7 +378,8 @@ function setupModuleLoader(window) {
                 templateUrl: makeInjectable(options.templateUrl),
                 transclude: options.transclude === undefined ? true : options.transclude,
                 scope: options.isolate === false ? true : {},
-                bindToController: options.bindings || {}
+                bindToController: options.bindings || {},
+                restrict: options.restrict || 'E'
               };
             }
 
--- a/test/loaderSpec.js
+++ b/test/loaderSpec.js
@@ -123,7 +123,8 @@ describe('component', function() {
         templateUrl: undefined,
         transclude: true,
         scope: {},
-        bindToController: {}
+        bindToController: {},
+        restrict: 'E'
       }));
     });
   });
@@ -137,7 +138,8 @@ describe('component', function() {
       templateUrl: 'def.html',
       transclude: false,
       isolate: false,
-      bindings: {abc: '='}
+      bindings: {abc: '='},
+      restrict: 'EA'
     });
     module('my');
     inject(function(myComponentDirective) {
@@ -148,7 +150,8 @@ describe('component', function() {
         templateUrl: 'def.html',
         transclude: false,
         scope: true,
-        bindToController: {abc: '='}
+        bindToController: {abc: '='},
+        restrict: 'EA'
       }));
     });
   });
--- a/test/loaderSpec.js
+++ b/test/loaderSpec.js
@@ -87,3 +87,111 @@ describe('module loader', function() {
     expect(window.angular.$$minErr).toEqual(jasmine.any(Function));
   });
 });
+
+
+describe('component', function() {
+  it('should return the module', function() {
+    var myModule = window.angular.module('my', []);
+    expect(myModule.component('myComponent', {})).toBe(myModule);
+  });
+
+  it('should register a directive', function() {
+    var myModule = window.angular.module('my', []).component('myComponent', {});
+    expect(myModule._invokeQueue).toEqual(
+      [['$compileProvider', 'directive', ['myComponent', jasmine.any(Function)]]]);
+  });
+
+  it('should add router annotations to directive factory', function() {
+    var myModule = window.angular.module('my', []).component('myComponent', {
+      $canActivate: 'canActivate',
+      $routeConfig: 'routeConfig'
+    });
+    expect(myModule._invokeQueue.pop().pop()[1]).toEqual(jasmine.objectContaining({
+      $canActivate: 'canActivate',
+      $routeConfig: 'routeConfig'
+    }));
+  });
+
+  it('should return ddo with reasonable defaults', function() {
+    window.angular.module('my', []).component('myComponent', {});
+    module('my');
+    inject(function(myComponentDirective) {
+      expect(myComponentDirective[0]).toEqual(jasmine.objectContaining({
+        controller: jasmine.any(Function),
+        controllerAs: 'myComponent',
+        template: '',
+        templateUrl: undefined,
+        transclude: true,
+        scope: {},
+        bindToController: {}
+      }));
+    });
+  });
+
+  it('should return ddo with assigned options', function() {
+    function myCtrl() {}
+    window.angular.module('my', []).component('myComponent', {
+      controller: myCtrl,
+      controllerAs: 'ctrl',
+      template: 'abc',
+      templateUrl: 'def.html',
+      transclude: false,
+      isolate: false,
+      bindings: {abc: '='}
+    });
+    module('my');
+    inject(function(myComponentDirective) {
+      expect(myComponentDirective[0]).toEqual(jasmine.objectContaining({
+        controller: myCtrl,
+        controllerAs: 'ctrl',
+        template: 'abc',
+        templateUrl: 'def.html',
+        transclude: false,
+        scope: true,
+        bindToController: {abc: '='}
+      }));
+    });
+  });
+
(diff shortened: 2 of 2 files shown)


petebacondarwin:
For completeness we ought to mention that this PR is related to the discussion at #12007.

@shahata can you clarify how this version compares with the suggestion at https://github.com/angular/angular.js/issues/10007#issuecomment-77509179?

Also have we considered the idea of ditching the magic symbols: `@`, `=` and `&` and using more ng2 terminology: `properties`, `events` and ... `bindings`, such as described at https://github.com/angular/angular.js/issues/10007#issuecomment-83838553?

src/loader.js
```diff
@@ -284,6 +284,48 @@ function setupModuleLoader(window) {
...
           /**
            * @ngdoc method
+           * @name angular.Module#component
+           * @module ng
+           * @param {string} name Component name
+           * @param {Object} options Component definition object
+           * @description
+           * TODO...
+           */
+          component: function(name, options) {
```
petebacondarwin: should be using `invokeLaterAndSetModuleName`?

src/loader.js
```diff
@@ -284,6 +284,48 @@ function setupModuleLoader(window) {
...
+            function factory($injector) {
+              function makeInjectable(fn) {
+                if (angular.isFunction(fn)) {
+                  return function(tElement, tAttrs) {
+                    return $injector.invoke(fn, this, {$element: tElement, $attrs: tAttrs});
+                  };
```
petebacondarwin: I wonder if this additional invoke via the `$injector` is an unwanted performance hit, since it would be called often, compared to the original factory function. But I guess that we can't get away from that given this API.

One alternative, could be to have a `dependencies` property on the `options`, which specifies all the dependencies that should be available in some closure to all these methods... Then we can inject these dependencies only once for all instances of the component, right?

pkozlowski-opensource:
Once comment: IMO we should avoid the `bindings` name as it is really overloaded in ng2 and adding yet another meaning in ng1 won't help. Maybe "properties"?

src/loader.js
shahata: this is just a wrapper around module.directive. module.directive will do the invokeLaterAndSetModuleName

src/loader.js
shahata: I'm not sure how that would work, the methods already have a closure when they are placed on the options obj, I don't think we can update that closure. Best we could do is pass the dependencies as a map argument to the methods, but I think it really isn't worth it, it doesn't seem like a big performance hit for me.

Another option is to have module.component accept either a options object or a factory method and then if you want templateUrl/template function that use DI, you have to pass a factory instead of options. But again, I think having templateUrl/template functions injectable is much more elegant and not such a big perf hit. It is also consistent with how the controller and $canActivate will be invoked.

shahata:
@petebacondarwin this proposal is consistent with https://github.com/angular/angular.js/issues/10007#issuecomment-77509179 with the exceptions that module component accepts an options obj instead of factory method and that scope isolation is optional

I'll look into the `properties`/`events` proposal

shahata:
@petebacondarwin I thought a bit about ditching the magic symbols today and I think that I'm leaning against it. Main reasons are that terminology is not the same since ng1 doesn't really have one way binding (would you use `properties` for `@` or `=`? what will you call the other?) and that as we previously discussed in #9137, introducing new way to the same thing is something we should avoid. wdyt?

shahata:
@petebacondarwin @btford @pkozlowski-opensource @caitp 
Any thoughts regarding accepting `properties` and `events` instead of magic symbols in `module.component`? (see previous comment)

petebacondarwin:
@shahata - I kind of like ditching the magical `=`, `@` `&`, but I also appreciate that there is no direct mapping two properties - if we can't come up with a set of names that we are happy with then let's just stick with what we have. Also note that we now also have the additional "modifiers" on these magical characters: `?`-optional and `*`-collection.

aciccarello:
I don't think the functionality needs to line up 100% between ng1 and ng2 `properties`, `events`, etc.. But I do think that the mapping needs to be intuitive and well documented.

pkozlowski-opensource:
@shahata I'm with @petebacondarwin here - the pb is that both `@` and `=` mean "binding to a property" (using the ng2 terminology) - those just denote 2 different ways of binding. The same could be argued about `&` which can be also used to _read_ expression value from an attribute....

So no, we shouldn't map existing symbols to props / events.

shahata:
@petebacondarwin @pkozlowski-opensource if we decide to go forward with adding one way binding maybe we can use `properties` for one way binds, `events` for callbacks and optionally also `bindings` for the existing magic symbols syntax. wdyt?

pkozlowski-opensource:
@shahata just fyi `properties` were just renamed to `inputs` and `events` to `outputs` in ng2. This sparked quite a discussion which IMO steamed from the fact that different people got different mental model of those things. Add on top the fact that concepts don't match between ng1 and ng2 and you are looking for trouble :-)

I must say that I'm not huge fan of the whole "make ng1 look like ng2 even if concepts don't match"  story

shahata:
This is why I suggested to do this only if we decide to match the concept of one way binds. Regarding naming I'm not going to touch it, we'll just do whatever ng2 does. :)

petebacondarwin:
Let's stick with  `=`, `&` and `@` (and possibly `>` if it lands) for this PR.
That way we are not going to get caught in trying to keep parity with potentially changing ng2 API naming :-)
If ngForward wants to sugar this to look more like ng2 then that can sit on top of this easily enough.

pkozlowski-opensource:
@shahata what is the one way binding in ng1 for you? `@`? If so the problem is that semantics are not the same... The pb with `@` is that it can only accept strings (or at least it is going to convert bound value to string).

In ng2 you can bind any type so `[foo]="expr"` is more like `=` binding in ng1 if no one does assignment :-)

petebacondarwin:
@pkozlowski-opensource the one-way binding we are talking about is this PR - https://github.com/angular/angular.js/pull/12835 - but using a different symbol that `#`, such as `>`.


Here is an interesting idea, which @MikeRyan52 has:

* all components must have either a `template` or a `templateUrl`
* all components must transclude their contents

This seems reasonable. I can't think of a situation where you would want a "component" that doesn't provide a template. And if that is the case then we would always be clearing out any child nodes of the element where the component is instantiated, so we might as well stick it in a transclude function. Previously this would have incurred a performance penalty but now that we have lazy compilation of transcluded contents, it would be negligible.

---

Side note - if the multi-slot transclusion lands, then this proposal would change to:

* component must have `transclude: true` (the default) or `transclude: { ... }` to indicate multi slot transclusion

pkozlowski-opensource:
@petebacondarwin thnx for the info. I like the `>` symbol somehow, I must say!

And I completely agree that each component must have template / templateUrl. I would also add that each component need to use isolated scope (consequence of having a template and most probably state that goes with it)

> all components must transclude their contents

I don't think I understand reasoning behind it. There are many examples of components that don't need to transclude anything (ex.: pagination, rating etc.)

petebacondarwin:
Yes, I agree that components should also always have `scope: {}`

Regarding transclusion, it means that the configuration object that gets passed to `module.component` is even simpler in the majority of cases. I.E. that we would never need to put `transclude: true` in the options for this call.

If there was no content below the element then we could optmize so that nothing would be transcluded anyway. In any case, now that we have lazy compilation the cost will barely be noticed, and if it was a problem then we could allow people to explicitly override it with `transclude: false`. 

So I guess my suggestion is that for components the default would be `transclude: true`.

pkozlowski-opensource:
> So I guess my suggestion is that for components the default would be transclude: true.

@petebacondarwin makes sense if there are no side effect / perf hit

petebacondarwin:
So the API that I would be happy with is:

```js
angular.module('myApp').component('myComponent', {

  template: '<div></div>', //string | Function, default: undefined
  templateUrl: 'a.html', //string | Function, default: undefined

  controller: MyCtrl, //string | Function | Array, default: function(){}
  controllerAs: 'vm', //string, default: component name
  bindings: {abc: '@'}, //object, default: {} (passed to bindToController)

  $canActivate: MyCtrl.canActivate, //Function, default: undefined (passed to factory.$canActivate)
  $routeConfig: MyCtrl.routeConfig, //RouteConfig, default: undefined (passed to factory.$routeConfig)

  isolate: false, //boolean, default: true (scope: {} .vs. scope: true)
  transclude: false //boolean | object, default: true
});
```

where

* one of `template` or `templateUrl` must be provided
* controller can be a string (name of a registered controller), function/class (explicit controller) or an array (inline annotated explicit controller)
* isolate defaults to `true`
* tranclude defaults to `true`

pkozlowski-opensource:
> controller: MyCtrl, //string | Function | Array, default: function(){}
  controllerAs: 'vm', //string, default: component name

To be honest I've never was fan of 2 properties here. How about `MyCtrl as myName` with default being `vm` if no `as` is provided?

shahata:
np, I'll add `transclude: true` by default. regarding `controllerAs`, it is allowed even today (though purely documented) to pass `controller: 'MyCtrl as vm'`. The reason we still need `controllerAs` is in the case I pass a class/function in `controller`, which I think will be the more common case.

petebacondarwin:
I prefer the case as stated where it defaults to the name of the component

shahata:
+1 for @petebacondarwin

shahata:
But that made me think of a bug I might have when you do `controller: 'MyCtrl as vm'` and `module.component` passes by default `controllerAs: 'compName'` I'm not sure who will win (obviously `vm` *should* win). I'll add a test for this.

shahata:
Yes, confirmed bug :)

shahata:
@petebacondarwin Regarding throwing when you don't pass a template or templateUrl or when you don't pass a controller - do you think we should really protect against unreasonable usage in run time or just document it so?

petebacondarwin:
I can live without an explicit controller, since you might just pass through values to the template but what does it mean to have a component with no view (template/templateUrl)? Perhaps the template should default to the empty string?

shahata:
I've added first attempt of documenting this. Please review.

shahata:
Any comments regarding docs?

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
 
           /**
            * @ngdoc method
+           * @name angular.Module#component
+           * @module ng
+           * @param {string} name Name of the component in camel-case (i.e. myComp which will match as my-comp)
```
petebacondarwin: `@param {string} name name of the component in canonical camelCase form (i.e. ...`

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           * @param {Object} options Component definition object, has the following properties (all optional):
```
petebacondarwin: I think that the docs generally doesn't start these descriptions with an upper case character

petebacondarwin: Perhaps we should note that this is conceptually similar to the directive definition object, with a link?

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *
+           *    - `controller` – `{(string|function()=}` – Controller fn that should be associated with
+           *      newly created scope or the name of a {@link angular.Module#controller registered
```
petebacondarwin: Controller constructor function

petebacondarwin: Perhaps a link to the controller section in `$compile`: https://docs.angularjs.org/api/ng/service/$compile#-controller

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *      controller} if passed as a string.
```
petebacondarwin: We need to clarify what is the default of each of the if not passed

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *    - `bindings` – `{object=}` – define DOM attribute binding to component properties.
+           *      component properties are always bound to the component controller and not to the scope.
+           *    - `$canActivate` – `{function()=}` – TBD.
+           *    - `$routeConfig` – `{object=}` – TBD.
+           *
+           * @description
+           * Register a component definition with the compiler. This is short for registering a directive
+           * where its definition object is isolated, allows transclusion and bound to controller as the
+           * component name by default.
+           * See {@link ng.$compileProvider#directive $compileProvider.directive()}.
```
petebacondarwin: We should be more explicit about the reasons why one would use this helper.
We should provide some example usage, including common cases where we rely upon the defaults.

PascalPrecht:
So if this helper uses `controllerAs` behind the scenes, could we agree on *not* using `vm` as namespace? It's not a big deal but the thing is that ppl that are using `vm` for controllers as (also `var vm = this` inside a controller), think that the controller **is** the view model, which is **not** the case.

The view model is still the scope, it just happens that an instance of the controller is assigned to a property on that scope, with the name that is used in `controllerAs`. I'd rather vote for sth. like `ctrl` instead.

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *      template that should be used  as the contents of this component.
+           *
+           *      If `templateUrl` is a function, then it is {@link auto.$injector#invoke injected} with
+           *      the following locals:
+           *
+           *      - `$element` - Current element
+           *      - `$attrs` - Current attributes object for the element
+           *    - `transclude` – `{boolean=}` – whether {@link $compile#transclusion transclusion} is enabled.
+           *      enabled by default.
+           *    - `isolate` – `{boolean=}` – whether the new scope is isolated. Isolated by default.
```
PascalPrecht: Do we actually have a use case where we'd build a component that needs a child scope instead of an isolated scope?

petebacondarwin:
@PascalPrecht - the default `controllerAs` value is the component name. So it is not using `vm`

PascalPrecht:
Okay sounds good! :+1:

src/loader.js
petebacondarwin: I don't think so but I guess it doesn't do much harm to have it as an option.

src/loader.js
gkalpak: Or perhaps we should generally start these descriptions with an upper case character ! (Some of them are, some aren't .)
It makes it much more readable (in code) and I am happy when I see an uppercased description :)

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *    - `controller` – `{(string|function()=}` – Controller fn that should be associated with
+           *      newly created scope or the name of a {@link angular.Module#controller registered
+           *      controller} if passed as a string.
+           *    - `controllerAs` – `{string=}` – An identifier name for a reference to the controller.
+           *      If present, the controller will be published to scope under the `controllerAs` name.
+           *      If not present, this will default to be the same as the component name.
+           *    - `template` – `{string=|function()=}` – html template as a string or a function that
+           *      returns an html template as a string which should be used as the contents of this component.
+           *
+           *      If `template` is a function, then it is {@link auto.$injector#invoke injected} with
```
gkalpak: Why are we deviating from the "normal" DDO's `template`/`templateUrl` (which don't get injected) ?
There's probably a good reason; I just want to know it :)

Maybe we should stress it more, because more experienced users might get biten by this.

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *    - `templateUrl` – `{string=|function()=}` – path or function that returns a path to an html
+           *      template that should be used  as the contents of this component.
+           *
+           *      If `templateUrl` is a function, then it is {@link auto.$injector#invoke injected} with
+           *      the following locals:
+           *
+           *      - `$element` - Current element
+           *      - `$attrs` - Current attributes object for the element
+           *    - `transclude` – `{boolean=}` – whether {@link $compile#transclusion transclusion} is enabled.
+           *      enabled by default.
```
gkalpak: _enabled_  -->  _Enabled_ (?)

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *    - `isolate` – `{boolean=}` – whether the new scope is isolated. Isolated by default.
+           *    - `bindings` – `{object=}` – define DOM attribute binding to component properties.
+           *      component properties are always bound to the component controller and not to the scope.
+           *    - `$canActivate` – `{function()=}` – TBD.
+           *    - `$routeConfig` – `{object=}` – TBD.
+           *
+           * @description
+           * Register a component definition with the compiler. This is short for registering a directive
```
gkalpak: This sentence is not very accurate, since you are not forced to use isolated scope, transclusion or the component name for controller alias.

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+           *      If `templateUrl` is a function, then it is {@link auto.$injector#invoke injected} with
+           *      the following locals:
+           *
+           *      - `$element` - Current element
+           *      - `$attrs` - Current attributes object for the element
+           *    - `transclude` – `{boolean=}` – whether {@link $compile#transclusion transclusion} is enabled.
+           *      enabled by default.
+           *    - `isolate` – `{boolean=}` – whether the new scope is isolated. Isolated by default.
+           *    - `bindings` – `{object=}` – define DOM attribute binding to component properties.
+           *      component properties are always bound to the component controller and not to the scope.
```
gkalpak: _component_  -->  _Component_ (?)

src/loader.js
```diff
@@ -284,6 +284,85 @@ function setupModuleLoader(window) {
...
+
+              var template = (!options.template && !options.templateUrl ? '' : options.template);
+              return {
+                controller: options.controller || function() {},
+                controllerAs: identifierForController(options.controller) || options.controllerAs || name,
+                template: makeInjectable(template),
+                templateUrl: makeInjectable(options.templateUrl),
+                transclude: options.transclude === undefined ? true : options.transclude,
+                scope: options.isolate === false ? true : {},
+                bindToController: options.bindings || {}
```
gkalpak: Why set it to `{}` if `options.bindings` is falsy ?
Wouldn't `false`/`true` work just as well ?

test/loaderSpec.js
```diff
@@ -87,3 +87,111 @@ describe('module loader', function() {
...
+    }).value('myValue', 'blah');
+    module('my');
+    inject(function(myComponentDirective) {
+      myComponentDirective[0].template('a', 'b');
+      myComponentDirective[0].templateUrl('c', 'd');
+      expect(log).toEqual('template,a,b,blah\ntemplateUrl,c,d,blah\n');
+    });
+  });
+
+  it('should allow passing transclude as object', function() {
```
gkalpak: Why do we want it to pass `transclude` as object ?

test/loaderSpec.js
```diff
@@ -87,3 +87,111 @@ describe('module loader', function() {
...
+    inject(function(myComponentDirective) {
+      expect(myComponentDirective[0]).toEqual(jasmine.objectContaining({
+        transclude: {}
+      }));
+    });
+  });
+
+  it('should give ctrl as syntax priority over controllerAs', function() {
+    window.angular.module('my', []).component('myComponent', {
+      controller: 'MyCtrl as vm'
```
gkalpak: It would be even better to add `controllerAs` as well, to test that even that has lower priority.

gkalpak:
I left a coule of minor comments.

One thing that is missing imo, is the `restrict` property. Right now it falls back to the current default for directives (and is not configurable using `.component()`).
When using `.component()`, I would expect it to default to `'E'` and also be configurable.

--
It might also be useful to make all defaults globally configurable, but I am not sure it is a good idea to give too much power to the user :stuck_out_tongue:

src/loader.js
petebacondarwin: The reason is that in `module.component` there is no factory function. In classic DDO, the `template` and `templateUrl` properties are in the closure of the factory function and so have access to injectables from there.

test/loaderSpec.js
petebacondarwin: One (long) word: multi-slot-transclusion (see https://github.com/angular/angular.js/commit/a4ada8ba9c4358273575e16778e76446ad080054)

petebacondarwin:
The defaults need to be consistent globally, since people may be loading in 3rd party libraries that also register using `module.component`. If you could configure them per app then these 3rd party components might break.

gkalpak:
As a user (of the framework) I would prefer to be able to configure the defaults (espacially if I have a large number of components) and 3rd-party libraries better define explicit values to avoid being broken.

But I agree it might be too much for now - it's probably better to keep things as straightforward as possible.

mzgol:
IMO defaulting to `restrict: 'E'` for components would be clearer, that's what people use for components when they drop IE8 support. Is there any reason why it needs to stay `EA`?

test/loaderSpec.js
shahata: I don't think it is a realistic use case that the developer will pass string with ' as ' in controller and also controllerAs property

src/loader.js
shahata: I just feel more comfortable thinking about `bindToController`'s type as always object

src/loader.js
shahata: This is why it ends with "by default". :)
I'll try to clarify.

src/loader.js
shahata: I guess some people would feel comfortable with using this with components that work tightly together and the child component needs access to the parent component.

shahata:
Updated docs & didn't squash on purpose to make it easy to see diff. I'm staying with capital letters for now until we decide otherwise. Regarding `restrict: 'E'` I tend to agree that it should be the default, but I can go either way.

petebacondarwin:
I agree, let's default to `E` but allow it to be overridden in the options object.

shahata:
Cool. Done.

petebacondarwin:
Any more feedback before we try to get this merged?

gkalpak:
LGTM :smile_cat:

shairez:
We should probably change "bindings" to "providers" right?

petebacondarwin:
@shairez - nope. Bindings in this context are template bindings. Nothing to do with the DI system.

shairez:
@petebacondarwin sorry, skipped that part in the discussion, I see now it's like what used to be the scope object in the DDO. 
I agree with you that it might be confusing, but since the name changed in ng2 from "bindings" to "providers",  I hope it will not be confusing moving forward. 
Thanks for clarifying! I'll start and test it to give more feedback

gkalpak:
Let there be...:sparkles: component :sparkles:

Woohoo ! :tada:

petebacondarwin:
Well done everyone who had a hand in this, especially @shahata for the implementation

sarod:
Sorry for the late feedback but I find that using `transclude:true` by default is probably not very user friendly for beginners as it causes more problems for simple usecases.

For instance using a simple component in conjunction with other directives requiring transclude like ngSwichWhen result in an error

```
[$compile:multidir] Multiple directives [ngSwitchWhen, myComp (module: myApp)] asking for transclusion on: <my-comp ng-switch-when="1">
```

```
<div ng-switch="x">
<my-comp ng-switch-when="1"></my-comp>
<div ng-switch-when="2"></div>
</div>
```

PR #12934, opened 2015-09-24
feat($compile): multiple transclusion via named slots

petebacondarwin:
In this version of multiple transclusion you specify the transclusion slots in the `transclude` property of the DDO, where the keys are the node names of transclusion zones and the values are the names of the slots in the template:

``` js
directive('minionComponent', function() {
  return {
    restrict: 'E',
    scope: {},
    transclude: {
      minion: 'minionSlot',
      boss: 'bossSlot'
    },
    template:
      '<div class="boss" ng-transclude="bossSlot"></div>' +
      '<div class="minion" ng-transclude="minionSlot"></div>' +
      '<div class="other" ng-transclude></div>'
  };
});
```

So the following HTML

``` html
<minion-component>
  <minion>stuart</minion>
  <minion>bob</minion>
  <span>dorothy</span>
  <boss>gru</boss>
</minion-component>
```

will be compiled and rendered to 

``` html
<minion-component>
  <div class="boss" ng-transclude="bossSlot">
    <boss>gru</boss>
  </div>
  <div class="minion" ng-transclude="minionSlot">
    <minion>stuart</minion>
    <minion>bob</minion>
  </div>
  <div class="other" ng-transclude>
    <span>dorothy</span>
  </div>
</minion-component>
```

Replaces #12742 and #11736

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1486,6 +1486,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         });
       };
 
+      // We need  to attach the transclusion slots onto the `boundTranscludeFn`
+      // so that they are available inside the `controllersBoundTransclude` function
+      var boundSlots = boundTranscludeFn.$$slots = createMap();
+      for (var slotName in transcludeFn.$$slots) {
+        boundSlots[slotName] = createBoundTranscludeFn(scope, transcludeFn.$$slots[slotName], previousBoundTranscludeFn);
+      }
+
       return boundTranscludeFn;
     }
 
@@ -1827,9 +1834,56 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
                                           nonTlbTranscludeDirective: nonTlbTranscludeDirective
                                         });
           } else {
+
+            var slots = createMap();
             $template = jqLite(jqLiteClone(compileNode)).contents();
+
+            if (isObject(directiveValue)) {
+
+              // We have transclusion slots - collect them up and compile them and store their
+              // transclusion functions
+              $template = [];
+              var slotNames = createMap();
+              var filledSlots = createMap();
+
+              // Parse the slot names: if they start with a ? then they are optional
+              forEach(directiveValue, function(slotName, key) {
+                var optional = (slotName.charAt(0) === '?');
+                slotName = optional ? slotName.substring(1) : slotName;
+                slotNames[key] = slotName;
+                slots[slotName] = [];
+                // filledSlots contains `true` for all slots that are either optional or have been
+                // filled. This is used to check that we have not missed any required slots
+                filledSlots[slotName] = optional;
+              });
+
+              // Add the matching elements into their slot
+              forEach($compileNode.children(), function(node) {
+                var slotName = slotNames[directiveNormalize(nodeName_(node))];
+                var slot = $template;
+                if (slotName) {
+                  filledSlots[slotName] = true;
+                  slots[slotName].push(node);
+                } else {
+                  $template.push(node);
+                }
+              });
+
+              // Check for required slots that were not filled
+              forEach(filledSlots, function(filled, slotName) {
+                if (!filled) {
+                  throw $compileMinErr('reqslot', 'Required transclusion slot `{0}` was not filled.', slotName);
+                }
+              });
+
+              forEach(Object.keys(slots), function(slotName) {
+                slots[slotName] = compilationGenerator(mightHaveMultipleTransclusionError, slots[slotName], transcludeFn);
+              });
+            }
+
             $compileNode.empty(); // clear contents
             childTranscludeFn = compilationGenerator(mightHaveMultipleTransclusionError, $template, transcludeFn);
+            childTranscludeFn.$$slots = slots;
           }
         }
 
@@ -2134,11 +2188,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
         // This is the function that is injected as `$transclude`.
         // Note: all arguments are optional!
-        function controllersBoundTransclude(scope, cloneAttachFn, futureParentElement) {
+        function controllersBoundTransclude(scope, cloneAttachFn, futureParentElement, slotName) {
           var transcludeControllers;
-
           // No scope passed in:
           if (!isScope(scope)) {
+            slotName = futureParentElement;
             futureParentElement = cloneAttachFn;
             cloneAttachFn = scope;
             scope = undefined;
@@ -2149,6 +2203,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
           if (!futureParentElement) {
             futureParentElement = hasElementTranscludeDirective ? $element.parent() : $element;
+          }
+          if (slotName) {
+            var slotTranscludeFn = boundTranscludeFn.$$slots[slotName];
+            if (!slotTranscludeFn) {
+              throw $compileMinErr('noslot',
+               'No parent directive that requires a transclusion with slot name "{0}". ' +
+               'Element: {1}',
+               slotName, startingTag($element));
+            }
+            return slotTranscludeFn(scope, cloneAttachFn, transcludeControllers, futureParentElement, scopeToChild);
           }
           return boundTranscludeFn(scope, cloneAttachFn, transcludeControllers, futureParentElement, scopeToChild);
         }
--- a/src/ng/directive/ngTransclude.js
+++ b/src/ng/directive/ngTransclude.js
@@ -8,12 +8,20 @@
  * @description
  * Directive that marks the insertion point for the transcluded DOM of the nearest parent directive that uses transclusion.
  *
+ * You can specify that you want to insert a named transclusion slot, instead of the default slot, by providing the slot name
+ * as the value of the `ng-transclude` or `ng-transclude-slot` attribute.
+ *
  * Any existing content of the element that this directive is placed on will be removed before the transcluded content is inserted.
  *
  * @element ANY
  *
+ * @param {string} ngTransclude|ngTranscludeSlot the name of the slot to insert at this point. If this is not provided or empty then
+ *                                               the default slot is used.
+ *
  * @example
-   <example module="transcludeExample">
+ * ### Default transclusion
+ * This example demonstrates simple transclusion.
+   <example name="simpleTranscludeExample" module="transcludeExample">
      <file name="index.html">
        <script>
          angular.module('transcludeExample', [])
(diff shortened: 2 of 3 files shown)


petebacondarwin:
This is a WIP and, of course, needs optimizing and documenting.

petebacondarwin:
Added support for required and optional slots; where the compiler will error if a required slot is not filled.

petebacondarwin:
@kara, @IgorMinar & @fenduru what do you think of this approach?

Narretz:
This is more explicit than #12742, which is good. Are there any other major differences?

fenduru:
@petebacondarwin I like the way this is coming along. After playing with it a little bit, [here are a couple of breaking test cases](https://gist.github.com/fenduru/bd1585705a62ae42e907). The first is a bug, and the second is more of desired behavior.

Moving the elements around during the compile step means that we're relying on the DOM being available already. This prevents the child elements from being created via directives like `ngTransclude` and `ngInclude`, which are both useful when composing other directives. Perhaps instead of looking at the node's children to find matching elements it would somehow be possible to look at the node's parent to see if there is a matching slot instead (although this would require running that check on all elements, rather than just elements known to use transclusion)? This is the only real dealbreaker I see with this implementation.

Due to #8914, I strongly advocate never using the attribute version of `ngTransclude` to my coworkers. It would be nice to have a similar syntax for the element version, maybe something like: `<ng-transclude slot="slotName">`

Should the key in DDO's transclude object use `element-name` or the normalized `elementName` for consistency?

One use case I have for this kind of feature would be to ensure proper order of child elements (instead of relying on consumers providing them in the correct order). For instance, I have a directive like:

```js
directive('foo', function() {
  return {
    restrict: 'E',
    transclude: {
      'one': 'oneSlot',
      'two': 'twoSlot',
      'three': 'threeSlot'
    },
    template: '<div ng-transclude="oneSlot"></div>' +
              '<div ng-transclude="twoSlot"></div>' +
              '<div ng-transclude="threeSlot"></div>'
  };
})
```

So when a consumer writes:

```html
<foo>
  <two>...</two>
  <three>...</three>
  <one>...</one>
</foo>
```

they are rendered in the correct order. The problem with this is that this slots implementation requires an additional wrapper around each slot content (since you need an `ngTransclude`), which limits the ability to style using sibling or `:last-child` selectors. The way I'm avoiding this in my userspace implementation of slots is by having the slot element be `transclude: 'element'`. It would be really convenient if this was the case for `ngTransclude` too.

src/ng/compile.js
```diff
@@ -1826,6 +1833,47 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+                filledSlots[slotName] = true;
+                slot = slots[slotName] = (slots[slotName] || []);
+              }
+              slot.push(node);
+            });
+
+            // Check for required slots that were not filled
+            forEach(filledSlots, function(filled, slotName) {
+              if (!filled) {
+                throw $compileMinErr('reqslot', 'Required transclusion slot `{0}` was not filled.');
```
kara: I think we're missing the `slotName` here to replace `{0}`:

```javascript
throw $compileMinErr('reqslot', 'Required transclusion slot `{0}` was not filled.', slotName);
```

petebacondarwin: Spot on @kara! Thanks

kara:
I like this approach. Setting up a transclude map makes sense, and the `?` flag for optional slots is a pretty useful addition.

The only thing still missing for me is the ability to use selectors other than node names, like it would be possible to do with `ng-content select` in Angular 2.  Currently, it looks like you have to use something like `<boss>`, and can't select by attribute (`<div boss>`), due to `directiveValue[_nodeName(node)]`. Any way we can add this without causing performance issues?

petebacondarwin:
@fenduru - thank you for your detailed and helpful response. I took a look at the failing tests you provided. I have added a fix for the bug and I will look into how we can support the second.

By the way, you can combine the element `ng-transclude` with an attribute of the same name: `<ng-transclude ng-transclude="boss"></ng-transclude>`, which gives you what you need there.

petebacondarwin:
@kara I think it is reasonable to restrict the things that can be selected, both for performance reasons but also, IMO, to make the whole thing easier to grok when you come across a directive that makes use of this multi-transclusion feature.  Once this is in, I would be willing to have discussion about making the selector more general, since that would not be a breaking change to the currently proposed API.

petebacondarwin:
> Should the key in DDO's transclude object use element-name or the normalized elementName for consistency?

We could normalise the element name, and store match against the normalized form stored as the key in the `transclude` object. I will implement that.

petebacondarwin:
> The way I'm avoiding this in my userspace implementation of slots is by having the slot element be transclude: 'element'. It would be really convenient if this was the case for ngTransclude too.

Do you mean that we should have an `ng-transclude-element` directive that replaces itself with the transcluded content rather than appending it?

fenduru:
> By the way, you can combine the element ng-transclude with an attribute of the same name: <ng-transclude ng-transclude="boss"></ng-transclude>, which gives you what you need there.

@petebacondarwin are you sure that would work? I would expect the element's link function to run, transcluding all of the non-slot content, then the attribute directive to also run transcluding the "boss" slot as well.

Also, I'm saying it would be nice to have something like `ng-transclude-element` that replaces itself with the transcluded content rather than appending it.

petebacondarwin:
@fenduru yes it does work: https://github.com/angular/angular.js/pull/12934/files#diff-348c2f3781ed66a24894c2046a52c628R7365

fenduru:
@petebacondarwin That actually only happens to work by chance. Both instances of `ngTransclude` are being linked, but you are relying on the fact that the attribute directive gets linked second. More so, when the attribute directive links, it empties the element (removing the content transcluded by the element directive) and leaks a transclusion scope. So basically that example is transcluding everything, then removing the elements from the DOM (without destroying the scope), then transcluding just the slot content.

Using that would break certain directives I've written that register with a parent directive during `link`, and unregister during `scope.$on('$destroy', ...`.

fenduru:
@petebacondarwin I've alluded to a userspace version of slots that I've written. [Here is a gist with the approach](https://gist.github.com/fenduru/43dd77bbcf0644c5ad98) in case it provides any insight.

Couple of key points:
* I called them "outlets" instead of "slots" (I actually prefer slots I think)
* Doesn't require the content element to be a direct child (not sure if this is good or bad)
* Forces the child directive to know which parent/slot it belongs in
* Must pass the controller to the outlet, otherwise things get screwy when you start nesting directives. (This takes advantage of how transclusion scopes work to talk to the correct parent, rather than using `require` which would rely on DOM order)
* The parent needs a special controller (or needs to mix in that functionality)
* Only supports 1:1 child to slot mappings, rather than n:1 like in your implementation (n:1 would be very useful though)

petebacondarwin:
By jove @fenduru you are right! I never knew that. I always assumed that if the element matched then we wouldn't match the same on an attribute or class.

petebacondarwin:
I like your work @fenduru - but I would like to get the feature more baked into the core so that users don't need to use custom helpers and controllers. I will check it out and see if there are useful ideas we can use. I am wondering whether `transclude: {...}` should imply `$$tlb`.

fenduru:
@petebacondarwin in my implementation only the child elements have `$$tlb` since they are the ones that get moved, so not sure if `transclude: {...}` implying `$$tlb` buys us anything (although tbh I have no idea what `$$tlb` actually means/stands for)

But yeah, I would also love if this was baked in (hence my interest in this PR :P)

petebacondarwin:
I think (:embarassed:) that $$tlb stands for "transclude late bound".

src/ng/compile.js
```diff
@@ -1472,7 +1472,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
     function createBoundTranscludeFn(scope, transcludeFn, previousBoundTranscludeFn) {
 
-      var boundTranscludeFn = function(transcludedScope, cloneFn, controllers, futureParentElement, containingScope) {
+      var boundTranscludeFn = function boundTranscludeFn(transcludedScope, cloneFn, controllers, futureParentElement, containingScope) {
```
petebacondarwin: I just added this name to the function to aid with debugging. It will be stripped out during minification, I believe.

fenduru:
This is looking really good!

I'm sad to see the removal of the composition test case. Without it, it will be impossible to use this feature when constructing recursive directives. The only way I can think to solve this would be to allow things like `ngTransclude` and `ngInclude` to be special cased such that when they call `$transclude()` they can proxy to their parent's `$$slots`.

petebacondarwin:
@fenduru I'm afraid it is more difficult than that. Because the transclude slot parsing is done at compile time there is no chance for "dynamic" slots to be generated at link time (e.g. via `ng-transclude`) since it is too late by then.

I believe the only solution would be to go more down the lines of your library where all the transclusion slots are computed after compilation has finished.

I discussed this with @IgorMinar and we felt that for the time being this was the best way forward, since it gives us performance and also compile time checking that required slots have been filled.

fenduru:
@petebacondarwin I don't think requiring your slots to be present during compile time (i.e. part of your template) is unreasonable at all. On the other side though - the content that the user provided - we (directive authors) have no control over how the consumer puts the content there and would have to document that they cannot use ng-transclude/ng-include/similar to insert the child elements.

This limitation unfortunately means it won't be usable for my use case (reusable low level components) since it would prevent composition, which is kind of the point of using transclusion.

petebacondarwin:
I am going to merge this in as it is. The ability to specify content for zones dynamically would use the same API so it is feasible it could be added in the future without a BC.

pkozlowski-opensource:
Wooooohooooo!

fenduru:
@petebacondarwin you're right it wouldn't be a breaking change.

Do you know anything about transclusion in angular 2? Are the use cases I brought up being considered?

petebacondarwin:
@fenduru I'm not sure about this in Angular 2, yet. Perhaps @tbosch or similar would be able to shed some light on this?

gkalpak:
I just managed to take a proper look (:flushed:).
:clap: Good stuff ! :clap:

petebacondarwin:
Thanks @gkalpak

tbosch:
In Ng2 we are using `<ng-content select="...">`, i.e. a selector based approach. This works in Ng2 as we can execute the content projection in a build step so we don't get a performance penalty. For Ng1 the named approach is a lot easier and performant to implement...

petebacondarwin:
So from an upgrade point of view, the ng2 approach is a superset of this PR and so it will be straightforward to convert from ng1 named multislot approach to the ng2 content selector approach

ilyes-garifullin:
Hello! Really looking forward to when this feature will be in the release. I decided to try how it will work. And I noticed that it is not working properly, now. Content from both ng-tansclude slots dublicates instead of every slot to render its own content. 

It can can be seen in the multi-slot transclusion Plunker example, at official documentation.
https://docs.angularjs.org/api/ng/directive/ngTransclude

petebacondarwin:
@ilyes-garifullin - can you clarify what the problem is? The demos look like they are working for me.

ilyes-garifullin:
Here is a screen shot from documentation:
image

And here is from Plunker:
image

And here is a screen shot of rendered html markup from chrome console tool: 
image

pane-title and pane-body elements are renderen inside both ng-transclude="title" and ng-transclude="body" directives, but pane-title should be rendered inside ng-transclude="title" and pane-body inside ng-transclude="body". 

The same result I get when cloned the latest version of angular from github

lgalfaso:
The issue in the plunker is that it is pointing to 1.5.beta1 and the multi-transclude code is not there. If you change angular version in the plunker from beta1 to https://code.angularjs.org/snapshot/angular.js then it works as expected. This should get fixed once beta2 is out

ilyes-garifullin:
Oh, I see, thank you!

jtrancas:
I'm totally new to AngularJS, however, I believe I found an issue when jQuery is being included along with Angular, I could not reproduce the error when Angular is running without jQuery.
The situation is the following:
- jQuery is included.
- There's a directive with **all** the transclude slots marked as optional.
- There's a directive occurrence with no child slots defined.

You can check the following Plnkr: http://plnkr.co/edit/f9SN35IMPrmeyuPz0Pih?p=preview
The *pane* directive has no child nodes, all its transclude slots are optional, and you may notice the following error being raised in the console:
```
TypeError: Cannot read property 'replace' of undefined
    at directiveNormalize (angular.js:9033)
    at angular.js:8095
    at forEach (angular.js:354)
    at applyDirectivesToNode (angular.js:8094)
    at compileNodes (angular.js:7620)
    at compileNodes (angular.js:7632)
    at compileNodes (angular.js:7632)
    at compile (angular.js:7527)
    at angular.js:1677
    at Scope.$eval (angular.js:16127)
```
If you remove the jQuery inclusion it all goes well.

petebacondarwin:
@jtrancas - thanks. I have created an issue to track this bug https://github.com/angular/angular.js/issues/13169

josip:
From my quick tests this looks like a tiny regression introduced with this PR, in 1.4 this used to work:
 ```html
<div ng-transclude="ng-transclude"></div>
```

But in 1.5 it throws an error:
```
[$compile:noslot] No parent directive that requires a transclusion with slot name "ng-transclude"
```

Maybe `"ng-transclude"` could be added to the list of values that trigger the default behaviour (as empty string/value already do)?

---

Funny `ng-transclude="ng-transclude"` is generated by Jade (which doesn't like empty attributes):
```jade
div(ng-transclude)
```

petebacondarwin:
OK @josip, that sounds reasonable.

petebacondarwin:
@josip can you take a look at #13383?

gogromat:
I was also struggling to make multi ngTrasclude to work, even with beta 2.

The Plunkr already points to beta 2 (1.5.0-beta.2), and it produces:
Error: [$compile:reqslot] http://errors.angularjs.org/1.5.0-beta.2/$compile/reqslot?p0=pane-body

However if I put ngTransclude keys and rename html elements as in [this Plunkr](http://plnkr.co/edit/vveeFY3xqewVvtJOcBkk?p=preview) I made, it seems to work fine:
```js
  transclude: {
    'paneb': 'paneb',
    'panet': '?panet',
    'panef': '?panef'
  }
```
and:
```html
<pane>
    <panet><a ng-href="{{link}}">{{title}}</a></panet>
    <paneb><p>{{text}}</p></paneb> 
</pane>
```

petebacondarwin:
@gogromat - there has been a bit of churn this weekend with the API for multi-slot transclude.
The currently visible docs - https://docs.angularjs.org/api/ng/directive/ngTransclude - (as of 10am GMT 8 December 2015) is not correct.

We are releasing RC1 today, in which the API should be stabilized.

The transclude map should be:

```js
{
  slotName: 'camelCaseElementName`
}
```

PR #12935, opened 2015-09-24
Added some missing functions to $log

humancatfood:
I find myself using these console functions a lot, so I decided to add them to $log. This makes it easier to turn off or find/remove all my debug stuff before deployment and looks more consistent in the code.

### DIFF
--- a/src/ng/log.js
+++ b/src/ng/log.js
@@ -104,6 +104,171 @@ function $LogProvider() {
 
       /**
        * @ngdoc method
+       * @name $log#time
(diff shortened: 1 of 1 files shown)


humancatfood:
I signed it!

Narretz:
The methods $log provides are not meant to mirror every method console provides. $log only writes by default to console, but if you have another implementation, for example writing to a remote log service, methods like table etc. would not really make a lot of sense. On the other hand, log, info, warn, error and debug are common error levels. I suggest you use a decorator to add the missing methods if you use them often.

PR #12939, opened 2015-09-24
Update datetimeSymbolsExt.js

juanzero000:
There are some typos in locales es_PE and es_UY. In spanish the 9th month is "Septiembre" not "Setiembre".


juanzero000:
I signed it!

Narretz:
The i18n data is imported from the google closure i18n library: https://github.com/google/closure-library/tree/master/closure/goog/i18n These changes must be made there, otherwise they'll be lost when we next import.

Narretz:
More precisely, the changes must be done in the CLDR first: http://cldr.unicode.org/. File a bug: http://cldr.unicode.org/index/bug-reports

PR #12943, opened 2015-09-25
Update rootScope.js

xieranmaya:
create $scope.$assign method


lgalfaso:
I do not understand why there is a need for this call. A better context would make things simpler

lgalfaso:
BTW, things that this PR lacks:
* test
* `$assign` might not be present

Narretz:
We really need a use case for this feature. It looks like it's just sugar at this point. Please reopen when you can provide more info,

PR #12945, opened 2015-09-25
docs($sceProvider): XSS when turning of SCE

mprobst:
Document that turning off SCE is very, very dangerous and should normally not be used by applications.

### DIFF
--- a/src/ng/sce.js
+++ b/src/ng/sce.js
@@ -342,6 +342,11 @@ function $SceDelegateProvider() {
      * returns the originally supplied value if the queried context type is a supertype of the
      * created type.  If this condition isn't satisfied, throws an exception.
      *
+     * <div class="alert alert-danger">
+     * Disabling auto-escaping is extremely dangerous, it usually creates a Cross Site Scripting
+     * (XSS) vulnerability in your application.
+     * </div>
+     *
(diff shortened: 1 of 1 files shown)


src/ng/sce.js
```diff
@@ -341,6 +341,9 @@ function $SceDelegateProvider() {
      * Takes the result of a {@link ng.$sceDelegate#trustAs `$sceDelegate.trustAs`} call and
      * returns the originally supplied value if the queried context type is a supertype of the
      * created type.  If this condition isn't satisfied, throws an exception.
+     * 
+     * **WARNING:** Disabling auto-escaping is extremely dangerous, it usually creates a Cross
+     * Site Scripting (XSS) vulnerability in your application.
```
gkalpak: Should this be placed in an alert box to stand out more ?

mprobst: Sounds good, but how do you create an alert box?

gkalpak: You can wrap the content in a `<div class="alert alert-warning/danger">...</div>`.
Example: https://docs.angularjs.org/api/ng/directive/ngInit

mprobst: Done, thanks for the pointer.

gkalpak:
LGTM

PR #12946, opened 2015-09-25
docs(tutorial/6 - Templating Links

heidi-jungel:
Updating to include the alt attribute for the image.  Per WCAG 2.0 Level A 1.1.1, all images must have alternative text.  Since this is likely a tutorial developers will follow, better to include accessibility while they are learning angular and show how it can be done with angular.

### DIFF
--- a/docs/content/tutorial/step_06.ngdoc
+++ b/docs/content/tutorial/step_06.ngdoc
@@ -44,7 +44,7 @@ __`app/index.html`:__
 ...
         <ul class="phones">
           <li ng-repeat="phone in phones | filter:query | orderBy:orderProp" class="thumbnail">
-            <a href="#/phones/{{phone.id}}" class="thumb"><img ng-src="{{phone.imageUrl}}" alt="{{phone.alt}}></a>
+            <a href="#/phones/{{phone.id}}" class="thumb"><img ng-src="{{phone.imageUrl}}" alt="{{phone.alt}}"></a>
             <a href="#/phones/{{phone.id}}">{{phone.name}}</a>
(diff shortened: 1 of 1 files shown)


gilesedinburgh:
requesting to add alt attribute to tutorial file

gilesedinburgh:
I signed it!

docs/content/tutorial/step_06.ngdoc
```diff
@@ -43,7 +44,7 @@ __`app/index.html`:__
 ...
         <ul class="phones">
           <li ng-repeat="phone in phones | filter:query | orderBy:orderProp" class="thumbnail">
-            <a href="#/phones/{{phone.id}}" class="thumb"><img ng-src="{{phone.imageUrl}}"></a>
+            <a href="#/phones/{{phone.id}}" class="thumb"><img ng-src="{{phone.imageUrl}}" alt="{{phone.alt}}></a>
```
gkalpak: Missing closing quote.

gkalpak:
I tool a closer look. If we are adding `alt`, I think we should be consistent. This means that similar changes are needed in steps 5, 10 and 12 as well as in the `angular-phonecat` repo.

gilesedinburgh:
I totally agree!  We should have these in all steps for consistency. I can certainly make that request for each of these steps.

gkalpak:
@gilesedinburgh, awesome ! I think it would be best to make all changes in one PR (plus one more for the `angular-phonecat` repo).

gkalpak:
Closed with c2033d7.

PR #12948, opened 2015-09-25
docs(tutorial/10 - Event Handlers): describe your change...

heidi-jungel:
Added ngKeypress directive to support keyboard only users for WCAG 2.0 Level A, 2.1.1 Keyboard Accessibility. To support and encourage developers to think about accessibility these should be clickable to keyboard only users.

### DIFF
--- a/docs/content/tutorial/step_10.ngdoc
+++ b/docs/content/tutorial/step_10.ngdoc
@@ -32,13 +32,22 @@ phonecatControllers.controller('PhoneDetailCtrl', ['$scope', '$routeParams', '$h
     $scope.setImage = function(imageUrl) {
       $scope.mainImageUrl = imageUrl;
     };
+    
+    $scope.setImagesKeyed=function(e, imageUrl){
+    if ((e.which==13) || (e.which==31)){
+      $scope.mainImageUrl = imageUrl;
+      e.preventDefault();
+      }
+    };
+
   }]);
 ```
 
 In the `PhoneDetailCtrl` controller, we created the `mainImageUrl` model property and set its
 default value to the first phone image URL.
 
-We also created a `setImage` event handler function that will change the value of `mainImageUrl`.
+We also created a `setImage` and 'setImagesKeyed' event handler functions that will change the value of `mainImageUrl`.  Since this image does not normally support keyboard navigation, we have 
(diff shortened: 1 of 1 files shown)


gkalpak:
Even though I think promoting accessibility is important, I feel like this change is out of scope.
Note that the same effect could be achieved by using `ngAria`.

gilesedinburgh:
I work for a company that reviews accessible code.  We are seeing developers follow these exact tutorial steps, missing on accessibility.   If we can incorporate accessible code as users are learning angular, this becomes second nature.  

The way we have it, they  have to retrofit after they have coded.

gkalpak:
I case I wasn't clear:

I too believe that a11y is important and promiting good practices in example code and tutorials is paramount to building "good developer habbits" (wrt to a11y). In fact, we (read @marcysutton) have recently-ish gone though _all_ docs examples and improved a11y (see #11079).

Yet, the functionality added by this change is already built-into `ngAria`, so the user would get it for free just by adding `ngAria` as a dependency to their module. Considering this is a tutorial (so people are learning new concepts), I think it's best to keep it at the absolute minimum. This change adds clutter which may make it more difficult for inexperienced users to focus on what is being taught here.

While I am OK with simple/small changes that improve a11y without introducing "mental" overhead (e.g. #12946), I believe that the tutorial is not the place for more complex stuff. Even more so, if the same functionality (plus a ton more) can be easily achieved by using `ngAria` as a dependency.

I think it would be nice to more explicitly mention `ngAria` and what it offers in the tutorial (either in the introduction or in a dedicated step) and/or point to relevant docs.

WDYT, @gilesedinburgh ?

/cc @petebacondarwin

petebacondarwin:
I agree with @gkalpak. Perhaps we should indeed consider a dedicated step in the tutorial where we implement `ngAria`? That would be nice.

gilesedinburgh:
I “third" that agreement!

gkalpak:
@gilesedinburgh, would you be up to creating a PR ? (Even if not 100% complete, it could serve as a starting point and get some discussion going.)

gilesedinburgh:
Sure. I can certainly do that.  It might take me a couple of days, but I will get one started.

Thanks!

gkalpak:
That's awesome. Thx !

PR #12950, opened 2015-09-25
docs($http): link to the usage section when config is mentioned

### LINKED ISSUE #12949: Docs for $http interceptor - what is the config object?
The docs say that you have access to the config object, but don't say what the object looks like. It'd be helpful to know what the object looks like.

Narretz:
This makes it easier for beginners to find out what the config object looks like

Closes #12949


gkalpak:
:+1:

PR #12952, opened 2015-09-25
fix(ngOptions): skip comments when looking for option elements

### LINKED ISSUE #12190: ngOptions in Angular 1.4 broke <select> placeholder values using ng-if

```
      </select>
```

```
      </select>
```

Narretz:
Closes #12190


petebacondarwin:
LGTM

petebacondarwin:
Woohoo - look at your @Narretz "merging" PRs!

Narretz:
Yeah, it looks nice! Too bad you cannot merge / push multiple commits at once. And thanks for the review!

PR #12955, opened 2015-09-26
feat($parse): allow option to have one-time bind be default behavior

onumossn:
When one-time binds are high, it can be simpler to write it as default rather than the exception. Also, utilizing this option in an application can help developers get more into the mindset of adding watches only where necessary rather than removing where its not.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1727,6 +1727,30 @@ function getValueOf(value) {
 function $ParseProvider() {
   var cacheDefault = createMap();
   var cacheExpensive = createMap();
+  var defaultToOneTime = false;
+  /**
+   * @ngdoc method
+   * @name $parseProvider#useOneTimeAsDefault
+   * @description
+   *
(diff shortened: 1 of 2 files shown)


lgalfaso:
there are several things that look very dangerous with this PR:
* if a module turns this flag on, then it can change the behavior on the entire app
* modules will need to know if this flag is on or off and have different templates to handle expressions differently
* `$parse` is used extensively, and this change can have some disruptive effects on directives like `ngModel`

@onumossn did you actually try this patch on top of an existing mid-side app?

onumossn:
No, I hadn't had the opportunity to try it on an existing mid-sized app due to the first point. It would potentially require a great deal of changes to implement on an existing mid-sized app. I was thinking of it from the perspective of starting something new. I hadn't considered how it would interact with modules that rely on the current default behavior.

lgalfaso:
I am going to close this based on
* turning on this option can cause just too many problem
* if people feel like experimenting, then the same effect can be done with a third-party module

PR #12959, opened 2015-09-27
docs($animateCss): describe your change...

Shrulik:
There was an extra asterisk, the second stagger is just a reference by the documentation, not the definition of a new field


Shrulik:
I signed it!

PR #12964, opened 2015-09-27
Add $location.setBaseHref()

edrico37:
I'd like to add a method to $location service to allow apps to set the base href of the app to something different than what's been specified in the <base href="/app/base/"> tag on the page. This method is only supported in full HTML5 mode. This will let developers update the page URL via $location to something outside the current app base without causing the page to reload.

To give a more concrete example, I have an angular app running in a subdirectory, let's just pretend it's "http://www.domain.com/angular-app/list". The base href for the app is set to "/angular-app/". When a user clicks a button, I want to show a popup modal and update the current URL to "http://www.domain.com/other-app/123". Currently I can't do this via $location without causing the page to refresh.

With this change, I can call:

``` javascript
showModal();
$location.setBaseHref('/other-app/').path('123');
```

If the user closes the modal, I can go back to the previous URL by calling:

``` javascript
$location.setBaseHref('/angular-app/').path('list');
```

I realize this is a hack and could be dangerous/error-prone if not used correctly. But I haven't found any other clean way to solve the scenario I described above. I've been using this hack in my own copy of angular for a few months and it seems to work (at least for my purposes). I figured it might help someone else in a similar situation.

Thanks.

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -158,6 +158,15 @@ function LocationHtml5Url(appBase, appBaseNoFile, basePrefix) {
     }
     return !!rewrittenUrl;
   };
+
+  /**
+   * Sets the appBase and appBaseNoFile members to use a new base href.
+   * @private
+   */
+  this.$$setBaseHref = function(baseHref) {
+    appBase = serverBase(this.absUrl()) + (baseHref || '/');
+    appBaseNoFile = stripFile(appBase);
+  };
 }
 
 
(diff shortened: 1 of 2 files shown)


edrico37:
I just found a bug with this implementation. I didn't realize the appBase that's used by this code doesn't reference the same appBaseNoFile variable that gets set by the function I added.
```
if (isUndefined(beginsWith(appBaseNoFile, newUrl))) {
    // If we are navigating outside of the app then force a reload
    $window.location.href = newUrl;
    return;
}
```

So if the URL changes outside of the angular context (the back button) the new base won't be respected by that code. Closing this pull request for now until I get that fixed.

PR #12965, opened 2015-09-28
Use strict DI notation in executeWithElement()

Chris380:


Narretz:
Why do you think this needs to be  changed?

Chris380:
Without this change I ran into an error while e2e-testing an app with strict DI activated:
Error: [$injector:strictdi] function($browser) is not using explicit annotation and can not be invoked in strict mode
http://errors.angularjs.org/1.4.6/$injector/strictdi?p0=function(%24delegate)

Narretz:
Ah okay, I see. So you are using the scenario runner for e2e tests, and not protracotr? Because the scenario runner is deprecated.

Chris380:
Yes, we still use the scenario runner for one (old) project. A switch to Protractor is not planned yet.

Narretz:
If it's an old project, does it actually help you if this change lands in 1.4.x and above?

Chris380:
We do regular updates of Angular and are currently using 1.4.6. Maybe that's a special scenario, so I don't have problems if this pull request is rejected.

petebacondarwin:
The `ngScenario` module is deprecated and we are not accepting further code changes for it. Please look at using Protractor for your e2e testing or consider forking your own version of `ngScenario`.

Chris380:
Thank's for your comment, Pete. I've already switched to Protractor for e2e-testing - works fine!

Issue #12915, opened 2015-09-22
ngOptions does not populate options through selectCtrl


user291831:
Regression(?) of #8761 

Support for `selectCtrl.hasOption` in `ngOptions` was added in 1.3.x (09eeb48) but was lost as part of the ngOptions/select rewrite in 1.4.0(?) (408f89d)

JSBins - *see console whilst changing select values*.

**1.3.1:** http://jsbin.com/sifegicura/1/edit?html,js,output
**1.4.5:** http://jsbin.com/zufuzuweda/1/edit?html,js,output

Why? I'm trying to write a directive which I can apply across a variety select fields and requires access to the options values that are available to the user.

Narretz:
Hm, that's a pickle. Since ngOptions is now a standalone directive, there would need to be an interface between ngOptions and select that adds the ngOptions generated options to the selectCtrl. I'm not sure how easy this is, based on the fact that ngOptions now encodes the options differently. If you want to give a PR a shot, though, we'll be happy to look at it.

Narretz:
We can overwrite the `hasOption` method, same as we do with `writeValue` etc.
However, two issues:
- the function is undocumented, so we'd need to nail down that it should work a specific way
- case in point: Should hasOption work with the scope.value or with the $viewValue? if you have a parser on ngModel that modifies the currently selected value, you cannot use it with scope.value

PR #12968, opened 2015-09-28
fix(ngOptions): don't throw if options are unset inside writeValue

### LINKED ISSUE #11685: Hard coded <option> causes error "TypeError: Cannot read property 'getOptionFromViewValue' of undefined"
Hi, in documentation is stated that 'empty option' can be used when using ng-options. Here is direct quote

> Optionally a single hard-coded **option** element, with the value set to an empty string, can be nested into the **select** element. This element will then represent the null or "not selected" option

I used that in 1.3 very heavily  and it worked flawlessly, but in 1.4 i'm getting error
```
TypeError: Cannot read property 'getOptionFromViewValue' of undefined
```

Just to be clear, here is the template i'm using

```html
<select ng-options="option.id as option.name for option in selectable_options">
  <option value="">Choose an option</option>
</select>
```

Is this planed regression or a bug? If this is planed regression, is there a workaround for the thing i'm trying to accomplish?

Narretz:
This can happen in the following case:
- there's a blank option inside the select
- another directive on the select element compiles the contents of it before ngOptions is linked.

Now this happens:
- the option directive is compiled and adds an element $destroy listener that calls ngModel.$render
- when ngOptions processes the blank option, it removes the element, and
  triggers the $destroy listener
- ngModel.$render delegates to selectCtrl.writeValue, which accesses the options
- in that phase, the options aren't yet set

Fixes #11685

### DIFF
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -2,7 +2,7 @@
 
 describe('ngOptions', function() {
 
-  var scope, formElement, element, $compile;
+  var scope, formElement, element, $compile, linkLog;
 
   function compile(html) {
     formElement = jqLite('<form name="form">' + html + '</form>');
@@ -103,6 +103,53 @@ describe('ngOptions', function() {
       }
     });
   });
+
+  beforeEach(module(function($compileProvider, $provide) {
+    linkLog = [];
+
+    $compileProvider
+      .directive('customSelect', function() {
+        return {
+          restrict: "E",
+          replace: true,
+          scope: {
+            ngModel: '=',
+            options: '='
+          },
+          templateUrl: 'select_template.html',
+          link: function(scope, $element, attributes) {
+            scope.selectable_options = scope.options;
+          }
+        };
+      })
+
+      .directive('oCompileContents', function() {
+        return {
+          link: function(scope, element)  {
+            linkLog.push('linkCompileContents');
+            $compile(element.contents())(scope);
+          }
+        };
+    });
+
+    $provide.decorator('ngOptionsDirective', function($delegate) {
+
+      var origPreLink = $delegate[0].link.pre;
+      var origPostLink = $delegate[0].link.post;
+
+      $delegate[0].compile = function() {
+        return {
+          pre: origPreLink,
+          post: function() {
+            linkLog.push('linkNgOptions');
+            origPostLink.apply(this, arguments);
+          }
+        };
+      };
+
+      return $delegate;
+    });
+  }));
 
   beforeEach(inject(function($rootScope, _$compile_) {
     scope = $rootScope.$new(); //create a child scope because the root scope can't be $destroy-ed
@@ -2119,6 +2166,35 @@ describe('ngOptions', function() {
       option = element.find('option').eq(0);
       expect(option.text()).toBe('A');
     });
+
+
+    it('should not throw when a directive compiles the blank option before ngOptions is linked', function() {
+      expect(function() {
+        createSelect({
+          'o-compile-contents': '',
+          'name': 'select',
+          'ng-model': 'value',
+          'ng-options': 'item for item in items'
+        }, true);
+      }).not.toThrow();
+
+      expect(linkLog).toEqual(['linkCompileContents', 'linkNgOptions']);
+    });
+
+
+    it('should not throw with a directive that replaces', inject(function($templateCache, $httpBackend) {
+      $templateCache.put('select_template.html', '<select ng-options="option as option for option in selectable_options"> <option value="">This is a test</option> </select>');
+
+      scope.options = ['a', 'b', 'c', 'd'];
+
+      expect(function() {
+        element = $compile('<custom-select ng-model="value" options="options"></custom-select>')(scope);
+        scope.$digest();
+      }).not.toThrow();
+
+      dealoc(element);
+    }));
+
   });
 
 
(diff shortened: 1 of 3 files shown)


Narretz:
I think the check is reasonable, even if the underlying problem is a bit iffy. Basically, the `option` directive does its thing even if ngOptions is on the select parent. So we could also introduce a check inside the option directive to prevent this. However, this really just happens if somehow the contents of the select are compiled before ngOptions gets to it, because ngOptions removes the blank option before it compiles it (and it therefore doesn't have a select parent and does nothing at all).

test/ng/directive/ngOptionsSpec.js
```diff
@@ -2119,6 +2130,18 @@ describe('ngOptions', function() {
       option = element.find('option').eq(0);
       expect(option.text()).toBe('A');
     });
+
+
+    it('should not throw when a directive compiles the blank option before ngOptions is linked', function() {
+      expect(function() {
+          createSelect({
```
gkalpak: Too much indentation ?

petebacondarwin:
> another directive on the select element compiles the contents of it before ngOptions is linked.

The problem here is that `ngOptions` is `terminal: true`, which means it makes the assumption that the option elements below it are not compiled ... ever. If we manually compile these options then they will get hold of the `SelectController` and add themselves. I suggest that a better fix for this issue is to get `ngOptions` to also override `SelectController.addOption` and `SelectController.removeOption` with `noop`s so that even if these options are compiled the do not get registered with the `SelectController` at all.

What do you think? I am not sure, yet, whether this would fix the other part of this issue.

Narretz:
If both directives have priority 0, they should be both executed (at least according to the docs). However, I just noticed that my current test actually passes without the change, because `compileContent` is not executed, but `rompileContents` is, because `r` comes after ` n`. This sounds a bit fishy to be honest.

Overwriting `addOption` and `removeOption` is good for consistency, but it won't fix the bug, because the call in `option` that leads to the exception comes from `element.$on('$destroy')

Narretz:
Correction: rompileContents is linked before ngOptions, compileContents after it, so that's why it always works in the second case.

petebacondarwin:
OK, so the best trick is to set the `SelectController`'s `ngModelCtrl` property to `null` during the pre link function for `ngOptions`. This effectively "turns off" the select directive, from the point of view of the `option` directive and is really what `ngOptions` would like anyway.

Narretz:
That doesn't work, because the order is:
- ngOptions pre
- options post
- option post

So the ngModelCtrl prop is set after ngOptions pre runs.

petebacondarwin:
Take a look at this https://github.com/angular/angular.js/pull/12972

petebacondarwin:
It is not quite right as you say but I feel that this is a good direction to be looking for a fix.

petebacondarwin:
Actually I take it back. Setting `$modelCtrl` to `null` alone doesn't fix the problem...

Narretz:
Let's get a bit crazy:
- give select priority: 1, so it's preLink fn is run before ngOptions
- move (almost) all select ctrl logic to preLink fn (is it problematic to move element eventlisteners there?)
- override selectCtrl.ngModelCtrl inside ngOptions's preLink fn.
- Tests pass
- ???
- Userland stuff breaks?

Honestly, the directive order stuff is so confusing, especially when you throw terminal in the mix.

petebacondarwin:
The really tricky bit is when you have templateURL and replace because some of the directives run half of their code at weird times and arguably out of order.

Narretz:
The commit contains a test for the replace directive + templateUrl.

petebacondarwin:
I think I have the right fix - force pushed to #12972

Narretz:
If the compileContents directive looks like this, it still breaks:

      .directive('oCompileContents', function() {
        return {
          priority: 2,
          link: {
            pre: function(scope, element)  {
              dump('oCompileContents pre')
              linkLog.push('linkCompileContents');
              $compile(element.contents())(scope);
            }
          }
        };
      });

with 

        TypeError: Cannot read property '$render' of null
            at HTMLOptionElement.<anonymous> (c:/Users/Martin/static/angular.js/src/ng/directive/select.js:468:35)

petebacondarwin:
Arguably that is an invalid use of Angular

Narretz:
@petebacondarwin I updated the PR. Now it uses a function on the selectCtrl that sets a flag and additionally sets addOption and removeOption to noop.

petebacondarwin:
Good stuff. Can we call the method `disable ()`?

Narretz:
I didn't want to call it disable because `disabled` has meaning the html element context.

petebacondarwin:
Oh, right. I see that could be confusing... (Although we are disabling the controller and not the element).
Trouble is I found that `override` didn't mean anything to me, and even knowing what we are doing, I had to do a double take. Mainly because I would expect an `override` method to be passing a parameter that would be used to "instead" of the original thing, do you know what I mean?
The only other word I can come up with is `annul` but that is totally esoteric and not intuitive either.
How about `disableController` or `detachController`?

petebacondarwin:
I had another idea. I wonder if it is possible to actually remove the `select` controller from the DOM from inside the `ngOptions` directive...

gkalpak:
I like `detach` better than `disable`. Or maybe `suspend`.

Narretz:
@petebacondarwin I don't know if removing the selectCtrl completely is what we want. I've always understood that ngOptions replaces some functionality of the selectCtrl instead of replacing it completely. For example, there's this issue - https://github.com/angular/angular.js/issues/12915 The API is undocumented, but I can see how it makes sense to support hasOption for select with and without ngOptions.

PR #12972, opened 2015-09-29
fix(ngOptions): disable select directive when ngOptions is active

petebacondarwin:


petebacondarwin:
With this patch, both the Plunkr with custom directive that compiles the options after the fact and also the one where we use replace and templateUrl don't break.

Narretz:
Strange, I thought I tested that and it didn't work.

Narretz:
In the test you are not actually adding the directive, though.

petebacondarwin:
:-) yeah the test is dubious but I checked the build against the plunkers.

Narretz:
I actually thought about that, but didn't really like it too much, because it special cases ngOptions from the perspective of the option directive that doesn't really need to know about ngOptions at all. But it's definitely cleaner. If you think it's better, we can do it this way.
What did you think about my test setup that makes sure the directives are actually linked in the order we expect? It makes the test a bit more rigid.

src/ng/directive/ngOptions.js
```diff
@@ -395,6 +395,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
   return {
     restrict: 'A',
     terminal: true,
+    controller: function() { },
```
Narretz: `noop`?

test/ng/directive/ngOptionsSpec.js
```diff
@@ -2119,6 +2130,18 @@ describe('ngOptions', function() {
       option = element.find('option').eq(0);
       expect(option.text()).toBe('A');
     });
+
+
+    it('should not throw when a directive compiles the blank option before ngOptions is linked', function() {
+      expect(function() {
+          createSelect({
```
Narretz: too much indentation

test/ng/directive/ngOptionsSpec.js
```diff
@@ -104,6 +104,17 @@ describe('ngOptions', function() {
     });
   });
 
+  beforeEach(module(function($compileProvider) {
+    $compileProvider
+      .directive('compileContents', function($compile) {
+        return {
+          link: { pre: function(scope, element) {
```
Narretz: Ugly indentation! ;)

src/ng/directive/select.js
```diff
@@ -417,10 +417,15 @@ var optionDirective = ['$interpolate', function($interpolate) {
...
-              parent.parent().data(selectCtrlName); // in case we are in optgroup
+        var selectElement = element.parent();
+        if (nodeName_(selectElement) !== 'select') {
+          selectElement = selectElement.parent(); // in case we are in optgroup
+        }
+
+        var selectCtrl = selectElement.data('$selectController');
+
+        // If there is an ngOptions directive then bail out
+        if (selectElement.data('$ngOptionsController')) return;
```
Narretz: I personally prefer `selectElement.controller('select')` Is the other one faster?

petebacondarwin: Yes, `.controller(...)` will keep searching up the tree till it hits the root

petebacondarwin:
Closing in favour #13012

PR #12974, opened 2015-09-29
docs($httpProvider): Parameter Detail Fix

johnz133:
useLegacyPromiseExtensions's boolean parameter was the other way around.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -345,9 +345,9 @@ function $HttpProvider() {
    * Configure `$http` service to return promises without the shorthand methods `success` and `error`.
    * This should be used to make sure that applications work without these methods.
    *
-   * Defaults to false. If no value is specified, returns the current configured value.
+   * Defaults to true. If no value is specified, returns the current configured value.
    *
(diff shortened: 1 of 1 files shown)


src/ng/http.js
```diff
@@ -347,7 +347,7 @@ function $HttpProvider() {
    *
    * Defaults to false. If no value is specified, returns the current configured value.
```
gkalpak: Could you also correct this ? (It defaults to `true`.)

src/ng/http.js
```diff
@@ -347,7 +347,7 @@ function $HttpProvider() {
...
    *
-   * @param {boolean=} value If true, `$http` will return a normal promise without the `success` and `error` methods.
+   * @param {boolean=} value If true, `$http` will return a promise with the depreciated legacy `success` and `error` methods.
```
gkalpak: _depreciated_  -->  _deprecated_

johnz133:
@gkalpak fixed. Nice catch!

Issue #12936, opened 2015-09-24
ng-if as base element of replaced directive destroys isolate scope.


nosideeffects:
Example: https://jsfiddle.net/01mfdjwm/1/

mzdunek93:
You could use `{{$parent.dir.hello}}`, but this clearly seems to be a bug, since the same issue doesn't exist when using `templateUrl`. Turns out that the directives at the root of a replaced directive get moved "next to" the containing directive, so the scope for the transclusion is set to be outside the directive. I submitted a fix to change the scope's prototype in such case:  #12975

nosideeffects:
Super strange that it works with `templateUrl`. Thanks for submitting a fix!

mzdunek93:
My pleasure. Just to be clear, it's not in yet, we'll have to wait for somebody from the core team to approve it.

petebacondarwin:
Duplicate of #9837 ?

PR #12975, opened 2015-09-29
fix($compile): fix scoping for transclusion

mzdunek93:
Change the scope's prototype when a directive with transclude is at the root of a template of a replaced directive

Fix for the issue #12936

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1270,7 +1270,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
     //================================
 
-    function compile($compileNodes, transcludeFn, maxPriority, ignoreDirective,
+    function compile($compileNodes, transcludeFn, maxPriority, changeOuterScope, ignoreDirective,
                         previousCompileContext) {
       if (!($compileNodes instanceof jqLite)) {
         // jquery always rewraps, whereas we need to preserve the original selector so that we can
@@ -1291,6 +1291,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       var namespace = null;
       return function publicLinkFn(scope, cloneConnectFn, options) {
         assertArg(scope, 'scope');
+
+        if (changeOuterScope) {
+          scope = scope.$parent.$new(false);
+        }
 
         options = options || {};
         var parentBoundTranscludeFn = options.parentBoundTranscludeFn,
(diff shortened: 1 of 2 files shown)


mzdunek93:
Signed.

petebacondarwin:
This looks very promising! I will look this week.

src/ng/compile.js
```diff
@@ -2162,6 +2173,15 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       }
     }
 
+    function markDirectivesAsReplacing(directives) {
+      // mark transcluding directives as replacing, so that their outer scope is changed to the directive's scope
+      for (var j = 0, jj = directives.length; j < jj; j++) {
+        if (directives[j].transclude) {
```
lgalfaso: should this be `if (directive[j].transclude === 'element') {` ?

src/ng/compile.js
```diff
@@ -1657,16 +1661,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
      * @param previousCompileContext
      * @returns {Function}
      */
-    function compilationGenerator(eager, $compileNodes, transcludeFn, maxPriority, ignoreDirective, previousCompileContext) {
+    function compilationGenerator(eager, $compileNodes, transcludeFn, changeOuterScope, maxPriority, ignoreDirective, previousCompileContext) {
```
lgalfaso: it looks like the order of `maxPriority` and `changeOuterScope` are in the wrong order

src/ng/compile.js
```diff
@@ -1815,8 +1821,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             compileNode = $compileNode[0];
             replaceWith(jqCollection, sliceArgs($template), compileNode);
 
-            childTranscludeFn = compilationGenerator(mightHaveMultipleTransclusionError, $template, transcludeFn, terminalPriority,
-                                        replaceDirective && replaceDirective.name, {
+            childTranscludeFn = compilationGenerator(mightHaveMultipleTransclusionError, $template, transcludeFn, changeOuterScope,
```
lgalfaso: the order of `changeOuterScope` and `terminalPriority` is wrong

[I do see that the two order errors cancel each other, but it is better if the order is respected]

test/ng/compileSpec.js
```diff
@@ -5629,6 +5629,34 @@ describe('$compile', function() {
       });
 
 
+      //see issue https://github.com/angular/angular.js/issues/12936
+      it('should use the proper scope when being the root element of a replaced directive', function() {
```
lgalfaso: for completeness, would it be possible to have the same test with `templateUrl` ?

test/ng/compileSpec.js
```diff
@@ -5629,6 +5629,34 @@ describe('$compile', function() {
...
+          directive('parent', valueFn({
+            scope: {},
+            replace: true,
+            template: '<div trans>{{x}}</div>',
+            link: function(scope, element, attr, ctrl) {
+              scope.x = 'iso';
+            }
+          }));
+          directive('trans', valueFn({
+            transclude: 'content',
```
lgalfaso: `transclude: 'content'` is not a supported value (even when it fallbacks to `true`

mzdunek93: I copied it from the other tests.

test/ng/compileSpec.js
mzdunek93: I don't think it's necessary, this bug didn't appear for `templateUrl`, since it loaded and compiled the template separately.

src/ng/compile.js
mzdunek93: I did it so that I didn't have to explicitly pass `undefined` for maxPriority to `compilationGenerator` instead of just omitting the argument, but I can fix that.

test/ng/compileSpec.js
petebacondarwin: @lgalfaso : Originally the possible values for `transclude` were indeed `"content"` and `"element"` but it got shorthanded to `true` and `"element"`. So this is a valid value.

test/ng/compileSpec.js
petebacondarwin: I afree we probably don't need a test for `templateUrl` but what we do need a test for is child scope (i.e. not isolated scope), since this is a common failing case (e.g. `ngIf` or `ngRepeat` at the root of the template).

mzdunek93:
I've switched the order of parameters in `compilationGenerator` and added another test for a child scope. What do you think?

PR #12979, opened 2015-09-30
fix($http): apply `transformResponse` when `data` is empty (for non-HEAD requests)

### LINKED ISSUE #12976: 1.4.5 $http transformResponse property isn't fired when status code is 204 (No Content)
I keep wanting to use the transform response property of the config object with an http service wrapper. The docs claim that the transformResponse is fired when there is a successful status message. Instead, I get no transformResponse firing. 

I have tests that show OK, 200, works, but 204 with null or undefined data does not.

The reason why is I'd like to set up default data, but I cannot with this current behavior (I took out a few things, but this should get the gist:

```js
describe("When using the SubNavServiceUri", function() {
    var serviceUri;
    var requestUri = "/api/aliens";
    var collection = null;
    var httpBackend;

    beforeEach(function () {
        module("superSiteApp");

        inject(function (_$httpBackend_, _SubNavServiceUri_) {
            httpBackend = _$httpBackend_;

            serviceUri = _SubNavServiceUri_;
        });
    });

    afterEach(function () {
        httpBackend.verifyNoOutstandingExpectation();
        httpBackend.verifyNoOutstandingRequest();
    });

it("and NoContent it should return empty collection", function () {
        // set up correct end point
        httpBackend
            .expectGET(requestUri)
            .respond(HttpStatusCode.NoContent);

        var actual;

        serviceUri.get(requestUri).then(function (result) {
            actual = result.data;
        });

        httpBackend.flush();

        expect([]).toEqual(actual);
        expect(0).toEqual(actual.length);
    });


...
service("SubNavServiceUri", ["$http", function ($http) {
        var isNullOrUndefined = function(item) {
            return item === null || item === undefined;
        };

        var transformData = function(data) {
            var result = [];
            return result;
        };

        var get = function(uri) {
            var object = {
                cache: false,
                transformResponse: function (data, headersGetter, status) {
                    console.log("howdy ");
                    console.log("headersGetter " + headersGetter);
                    console.log("status " + status);
                    if (isNullOrUndefined(data)) {
                        return [];
                    } else {
                        return transformData(data);
                    }
                }
            };

            return $http.get(uri, object);
        };

        return {
            "get": get
        };
...
```

gkalpak:
Fixes #12976

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -999,11 +999,8 @@ function $HttpProvider() {
       function transformResponse(response) {
         // make a copy since the response must be cacheable
         var resp = extend({}, response);
-        if (!response.data) {
-          resp.data = response.data;
-        } else {
-          resp.data = transformData(response.data, response.headers, response.status, config.transformResponse);
-        }
+        resp.data = transformData(response.data, response.headers, response.status,
+                                  config.transformResponse);
         return (isSuccess(response.status))
           ? resp
           : $q.reject(resp);
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1391,6 +1391,25 @@ describe('$http', function() {
           expect(callback).toHaveBeenCalledOnce();
           expect(callback.mostRecentCall.args[0]).toBe('RESP-FIRST:V1');
         });
+
+
+        it('should apply `transformResponse` even if the response data is empty', function(data) {
+          var callback = jasmine.createSpy('transformResponse');
+          var config = {transformResponse: callback};
+
+          $httpBackend.expect('GET', '/url1').respond(200, undefined);
+          $httpBackend.expect('GET', '/url2').respond(200, null);
+          $httpBackend.expect('GET', '/url3').respond(200, '');
+          $http.get('/url1', config);
+          $http.get('/url2', config);
+          $http.get('/url3', config);
+          $httpBackend.flush();
+
+          expect(callback.callCount).toBe(3);
+          expect(callback.calls[0].args[0]).toBe(undefined);
+          expect(callback.calls[1].args[0]).toBe(null);
+          expect(callback.calls[2].args[0]).toBe('');
+        });
       });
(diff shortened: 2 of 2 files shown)


gkalpak:
Possible BREAKING CHANGE:

Previously, non-empty responses to `HEAD` requests were transformed' not any more.
I didn't consider this a BC though, because it was never explicitly supported and according to RFC2616: 
_"The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response."_

--
We could as well change it so that even `HEAD` responses are transformed, if we feel more comfortable :)

[1]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4

Narretz:
LGTM. Don't think it's a BC. @petebacondarwin wdyt?

petebacondarwin:
I agree. It is not a BC.
Moreover, I am inclined to agree that we should just transform everything. Although HEAD requests are not "supposed" to return any data, there are, no doubt, servers out there that do. What do we care if they return data or not? We should just transform everything but not crash if the data is empty.

gkalpak:
Updated the commit. Now we always trasform the data.
<sub>
It just occurred to me:
If someone were using their own tranform function(s) (not the default), and were operating under the assumption that the data wouldn't be empty, they will experience some "broken-ness" :)
FWIW, it isn't mentioned or implied anywhere that `data` passed to the transform callback won't be empty, but that's how it was.

I still don't think this should be documented as a BC; just pointing it out.
</sub>

--
@petebacondarwin, @Narretz: PTYL.

petebacondarwin:
To be on the safe side, let's ensure that the commit documents this change, highlighting the case where someone's code may break if they were relying on the "undocumented" behaviour.

gkalpak:
I updated the commit message.
<sub>
Just to make sure I interpreted your comment correctly, @petebacondarwin:
You didn't mean I should ad a BC notice, did you ?
</sub>

PR #12981, opened 2015-09-30
docs($resource): Corrected wrong character

jaydge:
I believe the correct description of the behavior when parameters before suffixes are empty is that `/.` will be collapsed down to a single `/` - not a single '.'.


jaydge:
I signed it!

jaydge:
Nevermind, I now understand that the original documentation was in fact correct; I had misunderstood it.

PR #12987, opened 2015-10-01
Added a comment on a misleading declaration on the $location guide

williamtroy:
When the guide is followed correctly, a seemingly innocuous $sniffer value is
 set on the module. For the new developer this is pretty normal.  But what is
completely hidden is that setting this value in that way, will silently prevent
all animations from ngAnimate. 
Unless the developer knows that the value is not intended to be modified
(it's an internal service after all, it usually never is) they might
end up breaking their animations unknowingly.


samhax:
I signed it!

PR #12991, opened 2015-10-01
docs(tutorial/11 - REST and Custom Services): edit tutorial copy to reflect source

spectre6000:
Code breaks if tutorial is followed without reset. bower.js exceprt copy does not match source. Change to reflect in text body.


spectre6000:
I signed it!

spectre6000:
Missed a spot.

PR #12994, opened 2015-10-01
doc(angular.bootstrap): document debugInfoEnabled

thorn0:
Or is it not documented for a reason?

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1549,6 +1549,8 @@ function angularInit(element, bootstrap) {
  * @param {Object=} config an object for defining configuration options for the application. The
  *     following keys are supported:
  *
+ * * `debugInfoEnabled` - overrides the `debugInfoEnabled` setting defined in the provided
+ *   `modules` (see {@link ng.$compileProvider#debugInfoEnabled} for more.).
  * * `strictDi` - disable automatic function annotation for the application. This is meant to
  *   assist in finding bugs which break minified code. Defaults to `false`.
  *


gkalpak:
I think it is a private option (aka not documented on purpose).
It is only useful for debugging, and for that there is already [angular.reloadWithDebugInfo][1].
[1]: https://docs.angularjs.org/api/ng/function/angular.reloadWithDebugInfo

thorn0:
`reloadWithDebugInfo` calls `location.reload()`. If someone resorts to using `angular.bootstrap`, it usually means that his/her app isn't bootstrapped automatically after the page is loaded.

gkalpak:
The `config.debugInfoEnabled` property in `bootstrap()` is only used internally for implementing `angular.reloadWithDebugInfo()`.

If the user wants to have debug info enabled:

* permanently: They should use `$compileProvider.debugInfoEnabled(true)`.
* temporarily (for debugging): They should use `angulat.reloadWithDebugInfo()`.

--
There is no need to use the `config.debugInfoEnabled` property in `bootstrap()`.
I think it should stay private (i.e. undocumented).

thorn0:
It's enabled by default. But I just reread the source of `angular.bootstrap` and found out that passing `debugInfoEnabled: false` to this function actually doesn't have any effect. Only `true` does.

thorn0:
So this option doesn't fully control `debugInfoEnabled` as it can only enable it, but can't disable. Thus it really doesn't look suitable for public API.

PR #13000, opened 2015-10-02
docs(ngOptions): explain using 'select as' and 'track by' together

ryanhart2:
- change warning to indicate that 'select as' and 'track by' can be used together with care
- provide an example that will work and an example that will not work
- provide a simple explanation of why the non-working example does not work

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -58,28 +58,27 @@ var ngOptionsMinErr = minErr('ngOptions');
  * ### `select` **`as`** and **`track by`**
  *
  * <div class="alert alert-warning">
- * Do not use `select` **`as`** and **`track by`** in the same expression. They are not designed to work together.
+ * Be careful when using `select` **`as`** and **`track by`** in the same expression.
  * </div>
  *
- * Consider the following example:
+ * This will work:
  *
  * ```html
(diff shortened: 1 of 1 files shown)


src/ng/directive/ngOptions.js
```diff
@@ -58,29 +58,28 @@ var ngOptionsMinErr = minErr('ngOptions');
...
- *   subItem: { name: 'bSubItem' }
- * }];
+ * This will not work:
  *
- * $scope.selected = { name: 'aSubItem' };
+ * ```html
+ * <select ng-options="item.value as item.label for item in items track by item.id" ng-model="selected"></select>
  * ```
  *
+ * In both examples the **`track by`** expression is applied sucessfully to each `item` in `items`. It is also
```
gkalpak: _sucessfully_  -->  _successfully_

gkalpak:
@ryanhart2, could you fix the typo and amend the commit message to add `#Fixes 11768` (so it get's closed automatically) ?

ryanhart2:
Thanks @gkalpak, I will do that. I also realized that I didn't remove all of the previous example that I was replacing; I need to remove rows 83 to 94.

ryanhart2:
Sorry @gkalpak, I messed that up. I deleted the origin branch and pushed again, instead of a force push. The updated pull request is #13007.

PR #13002, opened 2015-10-03
fix($parse): do evaluate several times expressions with interceptors

### LINKED ISSUE #12983: Registering watch seems to trigger it one too many times
http://plnkr.co/edit/brHNIEoTQxNQF9eZikgJ?p=preview

It contains a template:
`<p>Hello {{log('A')}}  {{log('B')}}!</p>`

Run the above plunker and open console. I would expect the console to print: `A, B` and then `A, B` as second digest runs. Instead I see `A, A, B, B` and then `A, B` when second digest runs.

Is this a bug or intended behavior?

lgalfaso:
For simple expressions without filters that need a stateless interceptor then handle the 2nd phase parse evaluation using `inputs`.

Closes #12983

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -947,6 +947,14 @@ describe('$compile', function() {
           expect(child).toHaveClass('log'); // merged from replace directive template
         }));
 
+        it('should interpolate the values once per digest',
+            inject(function($compile, $rootScope, log) {
+          element = $compile('<div>{{log("A")}} foo {{::log("B")}}</div>')($rootScope);
+          $rootScope.log = log;
+          $rootScope.$digest();
+          expect(log).toEqual('A; B; A; B');
+        }));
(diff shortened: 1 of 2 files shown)


Narretz:
Can you reword the commit message? I have a hard time understanding what it means.

lgalfaso:
@Narretz sure, done

test/ng/compileSpec.js
```diff
@@ -947,6 +947,14 @@ describe('$compile', function() {
           expect(child).toHaveClass('log'); // merged from replace directive template
         }));
 
+        it('should interpolate the values once per digest',
```
petebacondarwin: `it('should interpolate simple values once per digest', `?

petebacondarwin:
LGTM - I think the LONG STORY from the issue should be included in the commit message.

PR #13003, opened 2015-10-03
docs($http): expand caching section to cover FAQ

### LINKED ISSUE #11101: $http cache docs are not clear
According to the [docs](https://docs.angularjs.org/api/ng/service/$http),
it says 
> To enable caching, set the request configuration cache property to true (to use default cache) or to a custom cache object (built with $cacheFactory)
You can change the default cache to a new object (built with $cacheFactory) by updating the $http.defaults.cache property. All requests who set their cache property to true will now use this cache object.
If you set the default cache to false then only requests that specify their own custom cache object will be cached.

I'm confused by the default value. The doc doesn't specify this.
So I run `$http.defaults.cache` and gets `undefined`. So will it ignore all `cache-control` header? It seems that this is not true.

I have the following question:
1. If cache is `true`, it will follow the http cache-control header and store the cache in browser cache. Is this correct? If it's `true` but the server doesn't reply `cache-control` header, it won't cache.
2. If cache is `false` it won't cache even if the server reply `cache-control` header.
3. If it's a CacheFactory object, then the object will handle it. Also ignore the cache-control header.

ryanhart2:
- add note that only GET & JSONP requests are cached
- add note that cache keys are the url and search params (headers not considered)
- add note that cache-control header on response does not affect angular caching
- add note that $httpProvider.defaults.cache can be used to set cache default (in addition to $http.defaults.cache)
- expand note on how default cache property affects requests with cache set to true

Fixes #11101

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -257,10 +257,8 @@ function $HttpProvider() {
    *
    * Object containing default values for all {@link ng.$http $http} requests.
    *
-   * - **`defaults.cache`** - {Object} - an object built with {@link ng.$cacheFactory `$cacheFactory`}
-   * that will provide the cache for all requests who set their `cache` property to `true`.
-   * If you set the `defaults.cache = false` then only requests that specify their own custom
-   * cache object will be cached. See {@link $http#caching $http Caching} for more information.
+   * - **`defaults.cache`** - {boolean|Object} - A value to enable caching of HTTP responses
+   * by default. See {@link $http#caching $http Caching} for more information.
    *
    * - **`defaults.xsrfCookieName`** - {string} - Name of cookie containing the XSRF token.
    * Defaults value is `'XSRF-TOKEN'`.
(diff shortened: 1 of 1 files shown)


src/ng/http.js
```diff
@@ -608,26 +608,30 @@ function $HttpProvider() {
...
+     * default cache or to a custom cache object (built with {@link ng.$cacheFactory `$cacheFactory`}).
+     * When the cache is enabled, {@link ng.$http `$http`} stores the response from the server
+     * in the default or specified cache. The next time the same request is made, the response
+     * is returned from the cache without sending a request to the server.
+     *
+     * Things to note:
+     *
+     *   * Only GET and JSONP requests are cached.
+     *   * The cache key is the request url including search parameters; headers are not considered.
+     *   * Cached responses are return asynchronously, in the same way as responses from the server.
```
gkalpak: _return_  -->  _returned_

ryanhart2:
@gkalpak, I fixed the spelling error and also uppercased 'URL' to be consistent with the rest of the comments. I also added Fixes #11101 to the commit comment.

src/ng/http.js
```diff
@@ -608,26 +608,30 @@ function $HttpProvider() {
...
      * {@link ng.$cacheFactory `$cacheFactory`}) by updating the
-     * {@link ng.$http#defaults `$http.defaults.cache`} property. All requests who set
-     * their `cache` property to `true` will now use this cache object.
+     * {@link ng.$http#defaults `$http.defaults.cache`} property or the
+     * {@link $httpProvider#defaults `$httpProvider.defaults.cache`} property. Requests with the
+     * configuration `cache` property set to true will be cached using this cache object.
      *
-     * If you set the default cache to `false` then only requests that specify their own custom
-     * cache object will be cached.
+     * The default cache can also be set to `false`, which will mean that only requests that specify
```
gkalpak: This sentence seems incorrect (from taking a quick look at the code).

ryanhart2:
@gkalpak, you are correct: a default cache value of false can be overridden by a cache config value of true as well as a custom cache object, so I will update.

This means that rows 260-263 of the file (describing the defaults object) are also incorrect, so I will include a revision of these in my commit as well.

There is also no mention of the effect of setting the default cache to true (it results in the $http cache being used when the config cache is undefined), so I will expand my comments to include this.

I set up this [plunker] (http://plnkr.co/edit/QuHeEzUYLmyKNxuIjMUh?p=preview) to test the various combinations.

Narretz:
@ryanhart2 Did you ever update this PR?

ryanhart2:
@Narretz Apologies, I didn't and had forgotten about it. I will have a look this weekend and try to get it done.

Narretz:
Am 21.01.2016 um 15:04 schrieb Ryan Hart:
>
> @Narretz <https://github.com/Narretz> Apologies, I didn't and had 
> forgotten about it. I will have a look this weekend and try to get it 
> done.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/angular/angular.js/pull/13003#issuecomment-173577789>.
>
No problem. I just wouldn't want to see good work go to waste!

Narretz:
@ryanhart2 why did you close it?

ryanhart2:
@Narretz Just struggling to do the rebase... I think I am getting there!

ryanhart2:
@Narretz , @gkalpak I have updated this pull request to more accurately describe how http caching works. Appreciate if you have time to review and provide feedback. Thanks.

src/ng/http.js
```diff
@@ -805,10 +812,8 @@ function $HttpProvider() {
      *      by registering it as a {@link auto.$provide#service service}.
      *      The default serializer is the {@link $httpParamSerializer $httpParamSerializer};
      *      alternatively, you can use the {@link $httpParamSerializerJQLike $httpParamSerializerJQLike}
-     *    - **cache** – `{boolean|Cache}` – If true, a default $http cache will be used to cache the
-     *      GET request, otherwise if a cache instance built with
-     *      {@link ng.$cacheFactory $cacheFactory}, this cache will be used for
-     *      caching.
+     *    - **cache** – `{boolean|Object}` – A value to enable caching of the HTTP response. See
```
gkalpak: IMO it is still useful to mention that the value must be either boolean or a `Cache` object created with `$cacheFactory`.

src/ng/http.js
```diff
@@ -257,10 +257,8 @@ function $HttpProvider() {
    *
    * Object containing default values for all {@link ng.$http $http} requests.
    *
-   * - **`defaults.cache`** - {Object} - an object built with {@link ng.$cacheFactory `$cacheFactory`}
-   * that will provide the cache for all requests who set their `cache` property to `true`.
-   * If you set the `defaults.cache = false` then only requests that specify their own custom
-   * cache object will be cached. See {@link $http#caching $http Caching} for more information.
+   * - **`defaults.cache`** - {boolean|Object} - A value to enable caching of HTTP responses
```
gkalpak: IMO it is still useful to mention that the value must be either boolean or a `Cache` object created with `$cacheFactory`.

gkalpak:
I left a couple of comments (actually it's the same comment twice :stuck_out_tongue: ).
BTW, @ryanhart2, you don't need to rebase. (In fact it's preferrable to just push a new commit with the changes :smiley:)

ryanhart2:
Thanks @gkalpak. I agree, the revised version now reads better. Thanks for the hint on the rebase. I have pushed a new commit addressing your suggestions. Note that, on the same lines, I also updated the words "enable caching" to "enable or disable caching", given that the values could be `false`.

gkalpak:
Backported to `v1.5.x` (2a7d4c1) and `v1.4.x` (b9d3625).
Thx @ryanhart2 for working on this !

PR #13007, opened 2015-10-04
docs(ngOptions): explain using 'select as' and 'track by' together

### LINKED ISSUE #11768: Potential issue in the documentation
Hi,

I was looking for some documentation about ngOption and I found this potential issue.

Here is a caution that warns us to do not use 'as' and 'track by' in the same expression :
https://github.com/angular/angular.js/blob/master/src/ng/directive/ngOptions.js#L61

But the next example, uses 'as' and 'track by' in the same expression :
https://github.com/angular/angular.js/blob/master/src/ng/directive/ngOptions.js#L67

Is that normal?

Regards,

Guillaume

ryanhart2:
- change warning to indicate that 'select as' and 'track by' can be used together with care
- provide an example that will work and an example that will not work
- provide a simple explanation of why the non-working example does not work

Fixes #11768

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -58,17 +58,13 @@ var ngOptionsMinErr = minErr('ngOptions');
  * ### `select` **`as`** and **`track by`**
  *
  * <div class="alert alert-warning">
- * Do not use `select` **`as`** and **`track by`** in the same expression. They are not designed to work together.
+ * Be careful when using `select` **`as`** and **`track by`** in the same expression.
  * </div>
  *
- * Consider the following example:
- *
- * ```html
- * <select ng-options="item.subItem as item.label for item in values track by item.id" ng-model="selected"></select>
- * ```
+ * Given this array of items on the $scope:
  *
  * ```js
- * $scope.values = [{
+ * $scope.items = [{
  *   id: 1,
  *   label: 'aLabel',
  *   subItem: { name: 'aSubItem' }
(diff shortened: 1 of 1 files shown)


gkalpak:
I ([obviously][1]) like documenting that it **is** possible to use `select as` and `track by` together, but I also kind of liked the lengthier explanation. I'd prefer a more detailed explanation (not necessarily copy-pasting the old one, but maybe incorporating parts of it).

[1]: https://github.com/angular/angular.js/issues/11768#issuecomment-97717964

petebacondarwin:
Thanks @ryanhart2 and @gkalpak for working on this.
Once you are happy with it, @gkalpak, please merge.

Narretz:
I tried to incorporate the original example with the two new examples, but the results are pretty confusing: http://plnkr.co/edit/GBw4YXl3sDzkX8oqXR4s?p=preview

I feel that the example needs to be much clearer in which circumstances the examples will work and in which not. The new "wrong" example works if you only change the selection from th UI, but the problem with the track by expression only occurs when the model is preselected. So this part of the original explanation is completely lost.

I think it would be the best if we had an actual runnable example where you could manipulate the selects and set the model the different values.

ryanhart2:
@gkalpak, @Narretz:
I have updated to incorporate the original example to address your feedback. Do you think this revised version is clearer and correct?

src/ng/directive/ngOptions.js
```diff
@@ -77,20 +73,32 @@ var ngOptionsMinErr = minErr('ngOptions');
...
  *
- * 1. Apply **`track by`** to the elements in the array. In the example: `[1, 2]`
- * 2. Apply **`track by`** to the already selected value in `ngModel`.
- *    In the example: this is not possible as **`track by`** refers to `item.id`, but the selected
- *    value from `ngModel` is `{name: 'aSubItem'}`, so the **`track by`** expression is applied to
- *    a wrong object, the selected element can't be found, `<select>` is always reset to the "not
- *    selected" option.
+ * In both examples the **`track by`** expression is applied successfully to each `item` in the `items` array.
+ * Because the selected option has been set in code, the **`track by`** expression is also applied to the `ngModel`
+ * value. In the first example, the `ngModel` value is `item[x]` and the **`track by`** expression evaluates to
```
gkalpak: _item[x]_  -->  _items[x]_

src/ng/directive/ngOptions.js
```diff
@@ -77,20 +73,32 @@ var ngOptionsMinErr = minErr('ngOptions');
...
+ * `item[x].id` with no issue. In the second example, the `ngModel` value is `item.subItem` and the **`track by`**
```
gkalpak: _item[x]_  -->  _items[x]_
_item.subItem_  -->  _items[x].subItem_

src/ng/directive/ngOptions.js
```diff
@@ -77,20 +73,32 @@ var ngOptionsMinErr = minErr('ngOptions');
...
+ * expression evaluates to `item.subItem.id`. As a result, the selected element is never found and the `<select>`
```
gkalpak: _item.subItem.id_  -->  _items[x].subItem.id_

gkalpak:
@Narretz, wrt https://github.com/angular/angular.js/pull/13007#issuecomment-146523951, I don't get what you mean. What is the confusing part ?

I can see the problem happening when the `ngModel` is set outside of the `<select>` (either programmatically, or through the UI (e.g. `ngClick` handler on a button)). Preselected or not doesn't seem to play a role.

When the value is set via the `<select>` everything works as expected, because `track by` is used for "reverse mapping" the model value to an option (and not the other way around).

Do I miss something ?

gkalpak:
/ping @Narretz
(I'll be merging this soon, so speak up if you have any objections :smiley:)

petebacondarwin:
Go for it @gkalpak

Narretz:
Sorry for the delay, I'm checking this again, give me 10 minutes.

src/ng/directive/ngOptions.js
```diff
@@ -77,20 +73,32 @@ var ngOptionsMinErr = minErr('ngOptions');
...
+ * ```
  *
- * 1. Apply **`track by`** to the elements in the array. In the example: `[1, 2]`
- * 2. Apply **`track by`** to the already selected value in `ngModel`.
- *    In the example: this is not possible as **`track by`** refers to `item.id`, but the selected
- *    value from `ngModel` is `{name: 'aSubItem'}`, so the **`track by`** expression is applied to
- *    a wrong object, the selected element can't be found, `<select>` is always reset to the "not
- *    selected" option.
+ * In both examples the **`track by`** expression is applied successfully to each `item` in the `items` array.
+ * Because the selected option has been set in code, the **`track by`** expression is also applied to the `ngModel`
```
Narretz: "has been set in code" doesn't really tell me anything. How about "has been set from the scope"?

Narretz:
Apart from a minor nitpick, it's looking good now.

src/ng/directive/ngOptions.js
gkalpak: I'll change it to "has been set programmatically in the controller". How about that ?

gkalpak:
Tweaked a bit and merged. Also backported to `v1.4.x` as 980fb39.
<sub>
:+1: to @ryanhart2 for submitting (and to @Narretz and @petebacondarwin for reviewing) !
</sub>

Narretz:
:thumbsup: for merging. Technically I don't think there needs to be a
controller involved for setting it programmatically, but a scope does, but
yeah ... it's not terribly important.
Am 27.10.2015 21:08 schrieb "Georgios Kalpakas" <notifications@github.com>:

PR #13012, opened 2015-10-05
fix(ngOptions): override select option registration

### LINKED ISSUE #11685: Hard coded <option> causes error "TypeError: Cannot read property 'getOptionFromViewValue' of undefined"

petebacondarwin:
When ngOptions is present on a select, the option directive should not be able to
register options on the selectCtrl since this may cause errors during the
ngOptions lifecycle.

This can happen in the following cases:
- there is a blank option below the select element, an ngModel
  directive, an ngOptions directive and some other directive on the select
  element, which compiles the children of the select
  (i.e. the option elements) before ngOptions is has finished linking.
- there is a blank option below the select element, an ngModel
  directive, an ngOptions directive and another directive, which uses
  templateUrl and replace:true.

What happens is:
- the option directive is compiled and adds an element `$destroy` listener
  that will call `ngModel.$render` when the option element is removed.
- when `ngOptions` processes the option, it removes the element, and
  triggers the `$destroy` listener on the option.
- the registered `$destroy` listener calls `$render` on `ngModel`.
- $render calls `selectCtrl.writeValue()`, which accesses the `options`
  object in the `ngOptions` directive.
- Since `ngOptions` has not yet completed linking the `options` has not
  yet been defined and we get an error.

This fix moves the registration code for the `option` directive into the
`SelectController`, which can then be easily overridden by the `ngOptions`
directive as a `noop`.

Fixes #11685
Closes #12972
Closes #12968

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -2,6 +2,15 @@
 
 var noopNgModelController = { $setViewValue: noop, $render: noop };
 
+function chromeHack(optionElement) {
+  // Workaround for https://code.google.com/p/chromium/issues/detail?id=381459
+  // Adding an <option selected="selected"> element to a <select required="required"> should
+  // automatically select the new element
+  if (optionElement[0].hasAttribute('selected')) {
+    optionElement[0].selected = true;
+  }
+}
+
 /**
  * @ngdoc type
  * @name  select.SelectController
@@ -77,6 +86,8 @@ var SelectController =
     }
     var count = optionsMap.get(value) || 0;
     optionsMap.put(value, count + 1);
+    self.ngModelCtrl.$render();
+    chromeHack(element);
   };
 
   // Tell the select control that an option, with the given value, has been removed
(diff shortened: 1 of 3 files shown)


src/ng/directive/select.js
```diff
@@ -98,6 +109,39 @@ var SelectController =
   self.hasOption = function(value) {
     return !!optionsMap.get(value);
   };
+
+
+  self.register = function($scope, $optionElement, $optionAttrs, interpolateValueFn, interpolateTextFn) {
```
Narretz: can we call it `registerOption`?

Narretz: Also, we inject `$scope` into the selectController. It's possibly a bit confusing to have a parameter with the same name in this fn.

petebacondarwin: OK, so I'll change the function to :+1: 

```js

replaceOption($optionScope, $optionElement, $optionAttrs, interpolateValueFn, interpolateTextFn) { ... }
```

OK?

Narretz: Is replaceOption a typo? Otherwise sgtm

gkalpak: Why the `$`-prefix in the first 3 args ?

petebacondarwin: Yes, I meant `registerOption` :-)

So in the Angular code, elements that are wrapped in jqLite are prefixed with a dollar, and I always prefer to do the same with the attributes and scope, since they are "Angular" things, not from the user's application space. I guess that this is not that common so I don't mind dropping them if they would cause confusion.

godza:
Great news. Thanks!

petebacondarwin:
Backported to 1.4 too

PR #13013, opened 2015-10-05
fix(testabilityPatch): reset cache after memory leak detection

petebacondarwin:
When you have a test failure in the angular.js project, which also causes a memory leak, such as a failure to dealocate an element or a scope, you often get a false-negative failure for every subsequent test. This change ensures that after such an error, the cache is reset so that subsequent tests can run normally.

### DIFF
--- a/test/helpers/testabilityPatch.js
+++ b/test/helpers/testabilityPatch.js
@@ -1,4 +1,4 @@
-/* global jQuery: true, uid: true */
+/* global jQuery: true, uid: true, jqCache: true */
 'use strict';
 
 /**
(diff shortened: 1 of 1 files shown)


shahata:
I think you can now remove `jqLiteCacheSizeInit()` and change `jqLiteCacheSize()` to simply be:
```js
function jqLiteCacheSize() {
  return Object.keys(jqLite.cache).length;
}
```

shahata:
The whole purpose of those methods was to handle the fact that cache is not reset between tests :)

petebacondarwin:
Is that a LGTM for this change then @shahata?

petebacondarwin:
Although changing that means that we need to clear the cache even if there wasn't a leak.

shahata:
Yes, LGTM :)

PR #13014, opened 2015-10-05
docs(ngResource): fix error in example

alexeymetelev:
user must be a parameter of callback function
currently it is not defined

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -187,7 +187,7 @@ function shallowClearAndCopy(src, dst) {
  *   read, update, delete) on server-side data like this:
  *   ```js
  *   var User = $resource('/user/:userId', {userId:'@id'});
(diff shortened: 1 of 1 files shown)


gkalpak:
Duplicate of #9791 (among others).

tl;dr
This change is not needed. The example works just as well without it (see https://github.com/angular/angular.js/pull/9791#issuecomment-60495160 for a more detailed explanation).

--
Thx for taking the time to submit a PR anyway !
Keep'em coming :smiley:

alexeymetelev:
OK :), sorry for wasting your time

I would add callback to the fiddle:

        $scope.user = User.get({userId: userId}, function() {
                $scope.user.callbackWorks = 'fine';
            }

to show that it works exactly like it should

gkalpak:
@alexeymetelev, added a callback :)

PR #13018, opened 2015-10-05
chore(sanitize): new implementation of the html sanitized parser

### LINKED ISSUE #11442: ngSanitize removes HTML strings between script elements

tjgambs:


Narretz:
Sorry, what exactly are you doing? This PR and the two others seem strange.

tjgambs:
Sorry I am still learning how to do Pull Requests

PR #13024, opened 2015-10-06
docs(loader): fix pluralization of constant

nebulou5:


gkalpak:
Backported to `v1.4.x` as 216724b.

PR #13025, opened 2015-10-06
fix($compile): evaluate against the correct scope with bindToController on new scope

### LINKED ISSUE #13021: Behaviour bindToController in child scope versus isolated scope
I was playing around with the `bindToController` option for directives. I stumbled upon a seemingly strange difference between the behaviour using a child scope compared to an isolated scope. When I use an isolated scope, an new scope is created for the directive, but changes to the bound controller attributes are forwarded to the parent scope. Yet when I use a child scope instead, my example breaks. (Using `bindToController` using child scopes should be allowed according to http://blog.thoughtram.io/angularjs/2015/01/02/exploring-angular-1.3-bindToController.html#improvements-in-14 )

The code:
```js
    {
        restrict: 'E',
        scope: {},
        controller: 'FooDirCtrl',
        controllerAs: 'vm',
        bindToController: {
            name: '='
        },
        template: '<div><input ng-model="vm.name"></div>'
    };
```
Working demo https://jsfiddle.net/tthtznn2/

The version using a child scope:
```js
    {
        restrict: 'E',
        scope: true,
        controller: 'FooDirCtrl',
        controllerAs: 'vm',
        bindToController: {
            name: '='
        },
        template: '<div><input ng-model="vm.name"></div>'
    };
```

Demo: http://jsfiddle.net/ydLd1e00/

The changes to name are forwarded to the child scope, but not to the parent scope. This in contrast to binding to an isolated scope, where the changes are forwarded to the old scope. Why is this?

**As copied from http://stackoverflow.com/questions/32972354/behaviour-bindtocontroller-in-child-scope-versus-isolated-scope **

gkalpak:
Previously, the directive bindings were evaluated against the directive's new (non-isolate) scope, instead of the correct (parent) scope.
This went unnoticed most of the time, since a property would be eventually looked up in the parent scope due to prototypal inheritance. The incorrect behaviour was exhibited when a property on the child scope was shadowing that on the parent scope.

This commit fixes it.

Fixes #13021

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2027,8 +2027,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       function nodeLinkFn(childLinkFn, scope, linkNode, $rootElement, boundTranscludeFn,
                           thisLinkFn) {
-        var i, ii, linkFn, controller, isolateScope, elementControllers, transcludeFn, $element,
-            attrs;
+        var i, ii, linkFn, controller, isolateScope, parentScope, elementControllers, transcludeFn,
+            $element, attrs;
 
         if (compileNode === linkNode) {
           attrs = templateAttrs;
@@ -2041,6 +2041,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         if (newIsolateScopeDirective) {
           isolateScope = scope.$new(true);
         }
+
+        parentScope = (newScopeDirective && !newIsolateScopeDirective) ? scope.$parent : scope;
 
         if (boundTranscludeFn) {
           // track `boundTranscludeFn` so it can be unwrapped if `transcludeFn`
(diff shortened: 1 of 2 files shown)


gkalpak:
I suspect there is some refactoring that would improve readability/maintainability, but I don't trust myself too much with compiler stuff :smiley:

JWGmeligMeyling:
:+1+

lgalfaso:
@petebacondarwin if you are looking into #12975, can you please make sure that both solutions are compatible?

petebacondarwin:
will do

JWGmeligMeyling:
Maybe we can combine this PR with https://github.com/angular/angular.js/pull/11345 . That PR is another issue with `bindToController` using child scopes, but went out of sight because it missed a test. The tests are now added.

PR #13029, opened 2015-10-07
fix(ngOptions): provide robust identification of empty options with ngIf

### LINKED ISSUE #12190: ngOptions in Angular 1.4 broke <select> placeholder values using ng-if

Narretz:
When an empty option has ngIf, the compilation result is different from
the extracted emptyOption, so the default way of identifying the empty
option doesn't work anymore. In that case, we need to make sure that we
don't identify comment nodes and options whose value is '' as valid
option elements.

Related #12952
Closes #12190


petebacondarwin:
Another option would be to take a reference to the empty option before it gets compiled?

petebacondarwin:
See #13033

PR #13033, opened 2015-10-07
fix(ngOptions) skip authentic empty options when looking for options

### LINKED ISSUE #12190: ngOptions in Angular 1.4 broke <select> placeholder values using ng-if

petebacondarwin:
Related #12952
Closes #12190
Closes #13029

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -408,7 +408,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         }
       }
 
-      var providedEmptyOption = !!emptyOption;
+      var providedEmptyOption = emptyOption;
 
       var unknownOption = jqLite(optionTemplate.cloneNode(false));
       unknownOption.val('?');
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngOptions.js
```diff
@@ -618,10 +618,19 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         var emptyOption_ = emptyOption && emptyOption[0];
         var unknownOption_ = unknownOption && unknownOption[0];
 
+        // If the compiled empty option was replaced by a comment because
+        // it had an "element" transclusion directive on it (such as ngIf)
+        // then compare against the pre-compiled empty element instead
+        if (emptyOption && emptyOption_.nodeType === NODE_TYPE_COMMENT) {
```
Narretz: We can actually remove this whole block then, can't we? (the tests pass without it)

Narretz: Because of `providedEmptyOption = emptyOption`, providedEmptyOption is also the compiled result. But the added checks below do the trick.

petebacondarwin: I guess so. As long as there is not some crazy `replace` directive on the option :-)

Narretz: Even that works in its simplest form, I just tested it. :) This case is probably very rare anyway, because having an option that replaces itself with an option takes this whole issue to strange new places!

ShallI remove the block and commit?

gkalpak: If this block is removed, don't forget to also undo the change on [line 411][1].

[1]: https://github.com/angular/angular.js/pull/13033/files#diff-0a9e86a58b2663074f51454f7a18a0efR411

Narretz:
I'll clean this up like proposed and merge it, if that's okay for you guys.

petebacondarwin:
@Narretz sounds good

Issue #13031, opened 2015-10-07
reloadWithDebugInfo does not work when the host already uses window.name


klaascuvelier:
I am not sure how much of an edge case this is, but the `reloadWithDebugInfo` call relies on `window.name` not to be used outside of angular.

We have a script-tag in our head setting the `window.name` for our application, this causes the implementation for `reloadWithDebugInfo` not to work.

https://github.com/angular/angular.js/blob/19ecdb54bf85fc4e7bd3cde453aa6843f869a1ab/src/Angular.js#L1600-L1638

*Angular version:* 1.4 (but I assume this applies to 1.3 as well)
*Browser*: every browser
*Reproduce*: add a script tag with `window.name = 'some name'` in your page head, disable debug info in your config, confirm `angular.reloadWithDebugInfo()` does not work 
*Suggested fix*: maybe use a query string parameter instead? Seems to me like `NG_ENABLE_DEBUG_INFO!` won't be a widely used query string?

vdclouis:
+1

superole:
angular apparently uses `window.name` for setting various flags not just `NG_ENABLE_DEBUG_INFO!`
another example is `NG_DEFER_BOOTSTRAP!` .

IMO this is a serious, standard-breaking bug. -A real code smell.
window.name is used to identify a browsing context, and any multi-window application that uses targeted links or window.open(), would break if this value is modified by a debug-tool or frontend-framework.

PR #13036, opened 2015-10-07
fix(Angular.js): use cookie instead of window.name to load with debug…

klaascuvelier:
… info

Use cookie to store a check if the application needs to be loaded with debug info, instead of using window.name. As window.name can be used alread (and be overwritten on page load)
Fixes issue #13031

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1578,7 +1578,6 @@ function bootstrap(element, modules, config) {
   var defaultConfig = {
     strictDi: false
   };
-
   config = extend(defaultConfig, config);
   var doBootstrap = function() {
     element = jqLite(element);
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1578,6 +1578,7 @@ function bootstrap(element, modules, config) {
   var defaultConfig = {
     strictDi: false
   };
+
   config = extend(defaultConfig, config);
   var doBootstrap = function() {
     element = jqLite(element);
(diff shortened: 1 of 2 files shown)


klaascuvelier:
Apparently SauceLabs complains about page reload, which indeed happens. 
I was not able to spy on `window.location.reload` which indeed triggers a page reload when calling the `reloadWithDebugInfo()`. I am not sure how to fix this. 
- Don't test the calling of `reloadWithDebugInfo()`? (it was not tested before)
- Stub window somehow? (I've seen some article talking about doing it with sinon)
- Other options?

petebacondarwin:
@juliemr do you remember why we used window.name rather than a cookie?

juliemr:
If I recall correctly, we can't set a cookie on the site without navigating to a different site on its domain first. So, if we would like to be able to mock out modules before they are ever loaded, we need a different strategy.

klaascuvelier:
Oh, I missed the updates on this PR, my apologies.
Other options would be a query string parameter or localstorage. But those are not optimal either.
Can you let me know if you'd want me to do something here, @petebacondarwin @juliemr

petebacondarwin:
I don't think that localStorage would work either, since it is locked to a domain, right?

petebacondarwin:
Any other ideas @juliemr ?

klaascuvelier:
So what's left is a query string parameter, @petebacondarwin?
Or would there be any other solutions?

cc @juliemr

juliemr:
Sorry for the slow response here. I'll think about it more, but when we first designed this it was a toss-up between query string and window.name and we decided that window.name looked cleaner to the user and had less chance of interfering with something custom. Before changing to query params we would have to think carefully about what that could possibly break.

klaascuvelier:
It might look cleaner, but it can break some functionality (like it did for me).
Not sure if having it look cleaner is a real selling point here, it's something that is only visible when a developer reloads the page to do some debugging ...

Anyway looking forward to some possible solution. Happy to contribute if I can.

petebacondarwin:
@klaascuvelier - of course changing the query string could break apps too. It is a bit of a stone or a hard place. Perhaps the idea solution would be for it to be configurable in the projector config. That way, the default would continue to be changing the window name (no BC) but developers could explicitly choose a query parameter to use instead if they wished?

PR #13039, opened 2015-10-07
feat(NgRoute): allow resolve promises on routeUpdate event

ivomju:
When using the reloadOnSearch defined as false, if you have a resolve property for that path the promise will not be invoked. My pull request is to enable that functionality, in our product we have a scenario where we need this feature I hope that this could help somebody else

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -581,14 +581,29 @@ function $RouteProvider() {
       }
     }
 
+    function buildResolvePromises(resolve) {
+      var locals = angular.extend({}, resolve);
+      angular.forEach(locals, function(value, key) {
+        locals[key] = angular.isString(value) ?
+            $injector.get(value) : $injector.invoke(value, null, null, key);
+      });
+
+      return locals;
+    }
+
     function commitRoute() {
       var lastRoute = $route.current;
       var nextRoute = preparedRoute;
+      var resolvePromises;
 
       if (preparedRouteIsUpdateOnly) {
-        lastRoute.params = nextRoute.params;
-        angular.copy(lastRoute.params, $routeParams);
-        $rootScope.$broadcast('$routeUpdate', lastRoute);
+        resolvePromises = buildResolvePromises(nextRoute.$$route.resolve);
+
+        $q.all(resolvePromises).then(function() {
+          lastRoute.params = nextRoute.params;
+          angular.copy(lastRoute.params, $routeParams);
+          $rootScope.$broadcast('$routeUpdate', lastRoute);
+        });
       } else if (nextRoute || lastRoute) {
         forceReload = false;
         $route.current = nextRoute;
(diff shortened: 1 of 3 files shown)


d0g:
Can you check again for CLA? I've made a little mistake but if you check again it will be ok.

d0g:
@petebacondarwin can you give me some input about this?

d0g:
@IgorMinar can someone give me some input about this?

test/ngRoute/routeSpec.js
```diff
@@ -1063,7 +1063,8 @@ describe('$route', function() {
     });
 
 
-    it('should not reload a route when reloadOnSearch is disabled and only .search() changes', function() {
+    it('should not reload a route when reloadOnSearch is disabled and' +
+      'only .search() changes without resolve promises defined', function() {
```
petebacondarwin: Is this change to the test description correct?

src/ngRoute/route.js
```diff
@@ -551,14 +551,29 @@ function $RouteProvider() {
       }
     }
 
+    function buildResolvePromises(resolve) {
+      var locals = angular.extend({}, resolve);
```
petebacondarwin: I would call `locals` `resolvePromises` to keep it consistent with the name of the function.

src/ngRoute/route.js
```diff
@@ -577,14 +592,9 @@ function $RouteProvider() {
         $q.when(nextRoute).
           then(function() {
             if (nextRoute) {
-              var locals = angular.extend({}, nextRoute.resolve),
+              var locals = buildResolvePromises(nextRoute.resolve),
```
petebacondarwin: We could even rename `locals` here too

petebacondarwin:
I haven't tried running the tests but on the surface this looks right

test/ngRoute/routeSpec.js
d0g: Yes, here you test without any resolve promises

d0g:
@petebacondarwin I've updated what you commented. Thanks for the input about this

d0g:
@petebacondarwin Do you see this being accepted?

petebacondarwin:
Yes but not till after 1.5 is released now

d0g:
Ok, thanks mate ;)

ivomirra:
Hi guy's, can you estimate in witch version this will be released?

petebacondarwin:
@ivomirra - I will add it to our next meeting for discussion. I am slightly concerned about the change to async from sync code.

ivomirra:
Thanks @petebacondarwin. Can you give me feedback when you have a decision regarding this? Cheers

petebacondarwin:
will do

petebacondarwin:
Hi @ivomju we discussed this PR in our meeting today. The general feeling was that this didn't seem to fit with how the library currently works. What is the point of triggering the resolves again? We can't "reinject" them into the constructor, since we do not recreate the controller in this case.
Unless you can provide a really strong use case that is likely to impact a number of developers, we are not likely to accept this PR.

Narretz:
I'm going to close this issue because we haven't got any feedback. Leave a comment if you can provide new feedback.

PR #13042, opened 2015-10-07
docs(ngInclude): add warning to onload function cause breaks in IE11

### LINKED ISSUE #12493: ng-include's onload generates errors with SVG elements in some browsers
If use ng-include on an SVG element and include an onload handler errors are thrown in Internet Explorer 11 (and in our selenium test runner - chutzpah, which seems to use phantomjs). This is a problem when developers use IE and it breaks our automated unit tests during builds.

It appears that SVG elements also define an attribute called "onload", which conflicts with the angular version of onload. In IE it actually attempts to call my function itself (not via angular) which results in an error.
````
SCRIPT5009: 'myOnLoadFunction' is undefined
````

For example 
````
<svg>
<g ng-include="content.svg" onload="myload()"></g>
</svg>
````

will result in 2 calls to something call myload(). Once by IE, which results in an error in the console since window.myload() doesn't exist. Next angular will call scope.myload(), which is what I intended.

A workaround was to create a noop windows.myload() function. This gives something for IE to call, and prevents the error. There should be way to define an onload expression w/o interfering with the SVG onload eventing. Maybe an alternate name? (ng-onload? data-onload?)

Demo (IE 11 with console open):
http://plnkr.co/edit/OdgOanXOgLoCXDxpI5yN?p=info

kutyel:
Closes #12493

### DIFF
--- a/src/ng/directive/ngInclude.js
+++ b/src/ng/directive/ngInclude.js
@@ -34,10 +34,10 @@
  * @param {string} ngInclude|src angular expression evaluating to URL. If the source is a string constant,
  *                 make sure you wrap it in **single** quotes, e.g. `src="'myPartialTemplate.html'"`.
  * @param {string=} onload Expression to evaluate when a new partial is loaded.
- * 
+ *
  *                  <div class="alert alert-warning">
  *                  **Note:** Seems like the onload attribute is being handled differently by different browsers,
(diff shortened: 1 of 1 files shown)


Narretz:
Thanks! I changed the text a bit and committed it.

kutyel:
You are welcome :smile:

PR #13045, opened 2015-10-08
orderBy should not allow to sort on string

S-YOU:
Currently if you sort `'hello' | orderBy`, it return `["e","h","l","l","o"]`, result does not make sense.
Instead it should be `hello` without sorted, or `ehllo` sorted as string.

This PR let orderBy return `hello` without sorting for string.
I hope this is not a breaking change.


S-YOU:
this was just a waste of my time.

PR #13047, opened 2015-10-08
Update ngOptions.js

epplestun:
Check value type to avoid errors

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -503,7 +503,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
             option.element.selected = false;
           });
 
-          if (value) {
+          if (value && Array.isArray(value)) {
             value.forEach(function(item) {
               var option = options.getOptionFromViewValue(item);
               if (option && !option.disabled) option.element.selected = true;


epplestun:
I signed it!

src/ng/directive/ngOptions.js
```diff
@@ -503,7 +503,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
             option.element.selected = false;
           });
 
-          if (value) {
+          if (value && Array.isArray(value)) {
```
gkalpak: It's better to use Angular's `isArray()`.

epplestun:
Why is better?

gkalpak:
It is better if we use the same function for detecting arrays throughout the codebase, becaue that way it is easier to fix issues or add features.
E.g. right now, `isArray()` is just a wrapper for `Array.isArray()`, but it was not always the case (e.g. on older versions supporting IE8) and it might not alaways be the case.

By using the same function, we only have to make the change in one place and we get more consistent behaviour across different parts (e.g. imagine one directive recognizing something as array, while another won't).

epplestun:
Ok, I will change Array.isArray(value) by angular.isArray(value) :)

gkalpak:
BTW, you didn't have to close this one. If you pushed a new commit to your branch or amended the previous commit (and force-pushed), GitHub would automatically update the PR.

Issue #13046, opened 2015-10-08
ng-options don't check value type when is multiple


epplestun:
If selec tag has multiple attribute, the ng-options directive don't  checks the type of value, assuming that is an array, and then use .forEach method directly.

https://github.com/angular/angular.js/blob/master/src/ng/directive/ngOptions.js#L506

In my opinion it should be checked the type to avoid errors

epplestun:
A simple solution:

```js
if(value && Array.isArray(value)) {}
```

gkalpak:
Is there any (valid) usecase where `value` is not an array ?
If not, I think it's better to throw, than silently fail.

epplestun:
Imagine, a directive that inserts other directives dinamically, and one of this directives has a "<select>" with "ng-options" directive, when you remove a directive with ng-options from the DOM the method $render is fired and value can be other thing and not an array, this is my reason to check the value type.

gkalpak:
Could you share some code (or a live demo using CodePen, Plnkr etc) ?
It's not clear to me.

epplestun:
I can't share because is private product .... I will try to make a codepen demo to demostrate.

PR #13048, opened 2015-10-08
Check value type to avoid errors

epplestun:


src/ng/directive/ngOptions.js
```diff
@@ -503,7 +503,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
             option.element.selected = false;
           });
 
-          if (value) {
+          if (value && angular.isArray(value)) {
```
gkalpak: You can use `isArray()` directly from inside Angular (no need for `angular.isArray()`).

gkalpak:
It would be nice, if you could add `Fixes #13046` at the end of the commit message, so the issue is automatically closed when(/if) the PR gets merged.

You might also want to take a look at our [commit message conventions][1]. It's not a biggy (i.e. we can take care of it when merging), but it saves us time.

gkalpak:
Hey, it just occurred me:
We need a test, to guard this fix against future regressions :)

PR #13050, opened 2015-10-08
Do not suggest using a promise as timeout

darabos:
Promises don't work. (#9332) I wish they did, but if they don't at least do not explicitly document that they do.

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.js/13050)


darabos:
I signed it! We signed it as a corporation, Big Graph Kft.

But I'd much prefer #13058 as a solution to this problem!

PR #13058, opened 2015-10-09
WIP - feat($resource): add support for cancelling requests using promises

### LINKED ISSUE #9332: Can no longer cancel $resource request with a promise

gkalpak:
Previously, it was not possible to use a promise as value for the `timeout` property of an action's `config`, because that value would be copied before being passed to `$http`.

This commit introduces a special value for the `timeout`, namely `'promise'`.
Setting an action's `timeout` configuration property to `'promise'`, will create a new promise and pass it to `$http` for each request. The associated deferred, is attached to the resource instance, under `$timeoutDeferred`.

Example usage:

``` js
  query: {
    method: 'GET',
    isArray: true,
    timeout: 'promise'
  }
});

var posts = Post.query();   // GET /posts
...
// Later we decide to cancel the request
// in order to make a new one
posts.$timeoutDeferred.resolve();
posts.query({author: 'me'});   // GET /posts?author=me
...
```

Fixes #9332
Closes #13050

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -517,6 +517,7 @@ angular.module('ngResource', ['ng']).
 
         Resource.prototype.toJSON = function() {
           var data = extend({}, this);
+          delete data.$timeoutDeferred;
           delete data.$promise;
           delete data.$resolved;
           return data;
@@ -574,10 +575,14 @@ angular.module('ngResource', ['ng']).
               undefined;
 
             forEach(action, function(value, key) {
-              if (key != 'params' && key != 'isArray' && key != 'interceptor') {
+              if (key !== 'params' && key !== 'isArray' && key !== 'interceptor') {
                 httpConfig[key] = copy(value);
               }
             });
+            if (action.timeout === 'promise') {
+              var deferred = value.$timeoutDeferred = $q.defer();
+              httpConfig.timeout = deferred.promise;
+            }
 
             if (hasBody) httpConfig.data = data;
             route.setUrlParams(httpConfig,
@@ -585,8 +590,9 @@ angular.module('ngResource', ['ng']).
               action.url);
 
             var promise = $http(httpConfig).then(function(response) {
-              var data = response.data,
-                promise = value.$promise;
+              var data = response.data;
+              var promise = value.$promise;
+              var timeoutDeferred = value.$timeoutDeferred;
 
               if (data) {
                 // Need to convert action.isArray to boolean in case it is undefined
@@ -613,6 +619,7 @@ angular.module('ngResource', ['ng']).
                 } else {
                   shallowClearAndCopy(data, value);
                   value.$promise = promise;
+                  value.$timeoutDeferred = timeoutDeferred;
                 }
               }
 
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -25,8 +25,20 @@ describe("resource", function() {
         headers: {
           'If-None-Match': '*'
         }
+      },
+      cancellableQuery: {
+        method: 'GET',
+        isArray: true,
+        timeout: 'promise'
+      },
+      cancellableGet: {
+        method: 'GET',
+        timeout: 'promise'
+      },
+      cancellablePut: {
+        method: 'PUT',
+        timeout: 'promise'
       }
-
     });
     callback = jasmine.createSpy();
   }));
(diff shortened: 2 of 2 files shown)


gkalpak:
This is a POC to get some feedback. 

Still missing/pending (if we decide to move this forward):

* [ ] Documentation
* [ ] Decide whether to keep or remove the `$timeoutDeferred` property once receiving the response.
* [ ] Decide how to handle the fact that `$timeoutDeferred` might not be present.
* [ ] Decide if `$timeoutDeferred` is an appropriate name.

darabos:
I'm not qualified to review, but my opinion is that this is an excellent solution! Thanks a lot!

petebacondarwin:
@gkalpak I think that the fix in #12657 makes this unnecessary?

gkalpak:
Not quite (e.g. see https://github.com/angular/angular.js/pull/12657#issuecomment-134315124 and next couple of comments).
With #12657 you need to recreate the resource "class" every time you cancel a request or it won't work correctly. It's not very practical imo.

This is adding better support for cancelling resource requests.

petebacondarwin:
I see! The feature is a good idea. I am not sure of the API, though.
What if we allowed people to pass in a promise to the instance methods?

gkalpak:
I don't feel too strongly about the proposed API, but I would rather avoid the boiler plate of having to pass a promise every time (especially if we can easily avoid it :)).
Maybe we could just expose an extra method for cancelling the request and handle the promise creation/rejection internally (but then there's the overhead of creating a promise every time, just in case the user decides to use it).

I'm definitely open to better ideas though.

petebacondarwin:
Let's have a think about it.

gkalpak:
Here's another idea (thinking loud):

- Keep the current behaviour, when using `timeout: <NUMBER>` (in `$resource`).
- Deprecate using `timeout: <PROMISE>` (in `$resource`), since we can't make it work correctly.
- Introduce a new boolean property (in `$resource`'s config - either per method or per class): `cancellable`. Iff this is `true`, we will add a new `cancel()`/`abort()` method on the returned instance/array of each method (and handle the promise creation and resolving internally).

--
(I wonder what other libd

petebacondarwin:
@gkalpak I like the sound of that better

gkalpak:
From a quick look to other libs ([Restangular][1], [Angular Restmod][2], [ModelCore][3], [angular-restResource][4]), they don't seem to have any special handling in place.
So, I'll go ahead and implement a quick POC as described in https://github.com/angular/angular.js/pull/13058#issuecomment-152182903.

[1]: https://github.com/mgonto/restangular
[2]: https://github.com/platanus/angular-restmod
[3]: https://github.com/klederson/ModelCore/
[4]: https://github.com/lifeiscontent/angular-restResource

petebacondarwin:
Good stuff @gkalpak

Issue #12697, opened 2015-08-28
$animate.on('addClass') etc do not provide enough info


jamie-pate:
What class is being added? i don't know! The current implementation (1.4.3) only passes in an empty object `{}` as 'data'

https://github.com/angular/angular.js/blob/ea8016c4c8f55bc021549f342618ed869998e335/src/ngAnimate/animateQueue.js#L450

matsko:
That's a good point. We should add that info.

PR #13059, opened 2015-10-09
fix(ngAnimate): Provide information about what class is being added for addClass and removeClass events

fhernandezn:
Hi everyone!

This pull request is related to the issue https://github.com/angular/angular.js/issues/12697 about the lack of information for the events addClass and removeClass.

Please review the code and let me know is something is wrong,

Closes https://github.com/angular/angular.js/issues/12697

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -401,6 +401,16 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       markElementAnimationState(element, PRE_DIGEST_STATE, newAnimation);
 
       $rootScope.$$postDigest(function() {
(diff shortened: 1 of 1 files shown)


Ariel-Isaacm:
Hey! good job, your solution is pretty neat :) it looks pretty good to me

frankweb:
Hi @Narretz could you take a look to my PR?

Narretz:
Thanks for the PR. A few things should be done:
- [ ] Make sure the object passed to the notify fn is always structured the same (so it should always have the addClass and removeClass properties, and they should be `null` if nothing is set). 
- [ ] Should the other options also be passed to the notifier / events?
- [ ] The behavior needs to be tested.

PR #13061, opened 2015-10-09
Feat linky 12558

Narretz:


src/ngSanitize/filter/linky.js
```diff
@@ -23,26 +34,14 @@
...
-         <tr>
-           <td>Filter</td>
-           <td>Source</td>
-           <td>Rendered</td>
-         </tr>
+         <thead>
+           <th>Filter</th>
+           <th>Source</th>
+           <th>Rendered</th>
+         </thead>
```
S-YOU: `th` are encouraged to use under `thead` without `tr`? I understand all browsers work either way.

Narretz: Sorry, what do you mean? Should I use `tr` instead of `thead`?

S-YOU: browsers generated both, if you miss one of them. But I am not html spec expert.
Just wondering why it has to be changed.

Narretz: I think you are right that you shouldn't use `<thead>` without `<tr>`, even though browser generate the `<tr>`. What I really wanted was to have the `<th>` because that improves the formatting.

S-YOU: > What I really wanted was to have the <th> because that improves the formatting.

I see, you're right.

PR #13064, opened 2015-10-09
Improve the docs search results layout

Narretz:
Sometimes the docs search results look really bad. This is hopefully an improvement that works in all browsers. Tested in FF, Chrome, IE11.

Here's a comparison for Firefox, searching for http:
OLD:
http-old-firefox

NEW:
http-new-firefox

Firefox looks the best, mainly because it supports hyphenation. Chrome breaks the words rather ugly, but you can actually read them. IE11 also hyphenates, but since it doesn't support @supports it doesn't get multi columns, so it looks a bit weird when list items break. Overall, though I'd say it's an improvement.

### DIFF
--- a/docs/app/assets/css/docs.css
+++ b/docs/app/assets/css/docs.css
@@ -315,8 +315,13 @@ iframe.example {
   color:white;
 }
 
+.search-results-group .search-results {
+  padding: 0 5px 0;
+  list-style-type: none;
+}
+
 .search-results-frame > .search-results-group:first-child > .search-results {
-  border-right:1px solid #050505;
+  border-right:1px solid #222;
 }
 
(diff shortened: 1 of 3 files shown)


Narretz:
If someone could test on Safari and Edge, this would be cool!

gkalpak:
I tried on Edge and it looks nice (it also supports hyphenation).
newdocssearch-edge

Narretz:
@petebacondarwin are you okay with me merging this before 1.5?

petebacondarwin:
Absolutely

Narretz:
Landed as 7da22e66855b5794d12d0e99e44f0c6f52ce2d55 and 0efef2385fe4dd4b2d72a964b1cf855c71733b4f (only on 1.5)

PR #13074, opened 2015-10-12
fix(ngMessage): make ngMessage work with ngBind

### LINKED ISSUE #8089: ngMessage and ngBind do not work together
If you have a custom validation key (such as `server` in this case), and try bind the `ngMessage` element to some object you presumably have the error message in, it does not work.

```html
<div ng-if="myForm.test.$dirty" ng-messages="myForm.test.$error">
    <span ng-message="server" ng-bind="errors.test"></span>
</div>
```

The above does not work, the span remains empty but visible and `errors.test` is indeed populated (and visible in scope, by trying `angular.element($0).errors.test`).

If you use a child, like so:

```html
<span ng-message="server">
    <span ng-bind="errors.test"></span>
</span>
```

This works correctly. 

Is this expected functionality or a bug? I have no issue with having to use a child if it is expected to behave this way.

Example [here](http://jsfiddle.net/U3e2t/).

Narretz:
Fixes #8089 

(I'm not actually sure why this works tbh)

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -572,6 +572,7 @@ angular.module('ngMessages', [])
     * @ngdoc directive
     * @name ngMessageExp
     * @restrict AE
+    * @priority 1
     * @scope
     *
     * @description
@@ -604,6 +605,7 @@ function ngMessageDirectiveFactory() {
     return {
       restrict: 'AE',
       transclude: 'element',
+      priority: 1, // must run before ngBind, otherwise the text is set on the comment
       terminal: true,
       require: '^^ngMessages',
       link: function(scope, element, attrs, ngMessagesCtrl, $transclude) {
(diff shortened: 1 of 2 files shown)


gkalpak:
LGTM

> (I'm not actually sure why this works tbh)

LOL ! It works because `ngMessage` has `transclude: 'element'`, which means that the comment whihc replaces it will be (sort of) compiled with `maxPriority: directive.priority`.
Previously, both `ngMessage` and `ngBind` had the default priority (0), so `ngBind`'s `compile` function was applied to the comment node. 

With `priority: 1` on `ngMessage`, `ngBind` won't be applied to the comment node (as it has a lower priority), but to the transcluded element (when/if it gets inserted into the DOM).

--
I'm not sure if this is a little BC, since all directives with priority 0 (or undefined => default), will not be applied during the initial compilation phase.
Directives that should be applied, usually have (or should have) and higher priority anyway, but you never know.

Narretz:
Thanks @gkalpak ! I've merged it only into 1.5 and added a BC notice.

gkalpak:
:+1:

Issue #13076, opened 2015-10-12
Proposal for spread in promises


fernandomoraes:
Hello , someone else misses the spread functionality , that the Q have? I think it's much more intuitive than just retrieve the values ​​of an array. Something more like this:

```js
$q.all([promise1, promise2]).spread(function (value1, value2) {
  console.log(value1);
  console.log(value1);
});
```

over this:

```js
$q.all([promise1, promise2]).then(function (arr) {
  console.log(arr[0]);
  console.log(arr[1]);
});
```

For me, it gets a little more confusing when the return value is also an array:

```js
$q.all([promise1, promise2]).then(function (arr) {
  console.log(arr[0][1]);
});
```
I made a <a href="https://github.com/angular/angular.js/pull/13077">pull request</a> for this, and the increase of final file of angularjs is much low

mchlbrnd:
I like your approach, but I think would stick with an array or object. Where I would use an array if the number of promises is unknown beforehand. Otherwise I would use an object;

```javascript
$q.all({
    one: promise1, 
    two: promise2
})
.then(function (result) {
   // start using result.one and result.two
});
```
For you example to work, $q.all (with arbitrary arguments) also needs a change.

fernandomoraes:
Hello @mchlbrnd, I wrote the $q.all in wrong way, sorry for this, now is correct. Like you can see now, there is no changes to do with $q.all ...

fernandomoraes:
In my pull request code, is secure for who receive a promise as argument, because always can call spread. This work like a uniforme interface ...

```js
function foo (promise) {
    promise.spread(function (a) {
        console.log(a);
    });
}

foo($q.all([promise1]));
foo(promise1);
```

mchlbrnd:
Hi @fernandomoraes,

What I liked was:
```javascript
$q.all(x, y, z).spread(function (x, y, z) {})
```
Then $q.all needs to transform arguments before doing its magic. I think all and spread need to work similarly in order to improve readability and consistency (removing square brackets).

Additionally, for an object, your PR produces a second argument of undefined in spread's callback (see below). Another problem might be preserving order of object properties. I think this PR will lead to unexpected behaviour with the current implementation of $q.all.

```javascript
$q.all({
  one: promise1,
  two: promise2
})
.spread(function (one, two) {
  // two will be undefined
});
```

Narretz:
[...]

PR #13077, opened 2015-10-12
feat $q: spread

fernandomoraes:
add spread functionality to promises

Issue: <a href="https://github.com/angular/angular.js/issues/13076">13076</a>

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -296,6 +296,22 @@ function qFactory(nextTick, exceptionHandler) {
       }, function(error) {
         return handleCallback(error, false, callback);
       }, progressBack);
+    },
+
+    spread: function (onFulfilled, onRejected) {
+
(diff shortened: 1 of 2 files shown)


src/ng/q.js
```diff
@@ -296,6 +296,22 @@ function qFactory(nextTick, exceptionHandler) {
...
+    spread: function (onFulfilled, onRejected) {
+
+      function toArray (value) {
+        var arr = [];
+        arr.push(value);
+        return arr;
+      }
+
+      function spread (value) {
+        var arrData = isArray(value) ? value : toArray(value);
```
campersau: Instead of `toArray(value)` you can just use `[ value ]`.

gkalpak:
See https://github.com/angular/angular.js/issues/13076#issuecomment-226011726.

PR #13087, opened 2015-10-14
refactor(ngMessages): remove unused argument

gkalpak:


Narretz:
That makes me think. Can you actually use ngMessage as an element?

gkalpak:
[Yup][1]

[1]: http://codepen.io/ExpertSystem/pen/NGwPWx?editors=100

PR #13095, opened 2015-10-14
fix(ngAria): remove tabindex from radio buttons

### LINKED ISSUE #12492: ngAria disables tabbing to a radio set when no default value selected
- **Overview of the Issue** - When a radio set is configured with no default selected value (ie: the value starts out as undefined), all the radio buttons will have their `tabindex` set to `-1`, making it impossible to tab to the radio set.
- **Motivation for or Use Case** - This makes it difficult to use a normal radio set for focus when you don't want to preselect the value for the user.
- **Angular Version** - 1.4.1
- **Browsers and Operating System** - Tested on FF39 on Mac OS X 10.10.4
- **Reproduce the Error** - [View on Plunkr](http://embed.plnkr.co/oYvOGsHxcYPMZOuRLQg1/preview)
- **Related Issues** - A partial attempt to fix this was issue #10212
- **Suggest a Fix** - I'm not sure what the right solution here is.  Some ideas:
    - I don't know what the purpose of setting the `tabindex` to `-1` is, if it's not critical, it probably should be removed.
    - Another option would be to find the entire set of radios with a matching `name`, and if none are selected, then all should be active, otherwise only the selected on is active.

marcysutton:
There was some funky stuff going on when ngModel was set to null, which I fixed by adding a nodeBlackList check to the method setting tabindex. 

The logic added to `needsTabIndex` should eventually be moved into the `shouldAttachAttr` function but it will require refactoring of other directives, so I decided to wait on that. 

Closes #12492

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -235,7 +235,8 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
           }
         },
         post: function(scope, elem, attr, ngModel) {
-          var needsTabIndex = shouldAttachAttr('tabindex', 'tabindex', elem);
+          var needsTabIndex = shouldAttachAttr('tabindex', 'tabindex', elem)
+                                && !isNodeOneOf(elem, nodeBlackList);
 
           function ngAriaWatchModelValue() {
             return ngModel.$modelValue;
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -616,16 +616,18 @@ describe('$aria', function() {
   describe('tabindex', function() {
     beforeEach(injectScopeAndCompiler);
 
-    it('should not attach to native controls', function() {
-      var element = [
-        $compile("<button ng-click='something'></button>")(scope),
-        $compile("<a ng-href='#/something'>")(scope),
-        $compile("<input ng-model='val'>")(scope),
-        $compile("<textarea ng-model='val'></textarea>")(scope),
-        $compile("<select ng-model='val'></select>")(scope),
-        $compile("<details ng-model='val'></details>")(scope)
-      ];
-      expectAriaAttrOnEachElement(element, 'tabindex', undefined);
+    they('should not attach to native control $prop with ng-model', {
+      'button': "<button ng-click='something'></button>",
+      'a': "<a ng-href='#/something'>",
+      'input[text]': "<input type='text' ng-model='val'>",
+      'input[radio]': "<input type='radio' ng-model='val'>",
+      'input[checkbox]': "<input type='checkbox' ng-model='val'>",
+      'textarea': "<textarea ng-model='val'></textarea>",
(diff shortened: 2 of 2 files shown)


Narretz:
I think we should test that no element in the blacklist gets a tabindex, not only one. I think matsko's `they` helper might be good for this.

gkalpak:
I believe this has been brought up before (https://github.com/angular/angular.js/pull/12262#issuecomment-127953813). My understanding from that conversation was that it is a breaking change.

Is my understanding correct ? (If so a BC notice should added.)

--
On the minor side:

:+1: for checking all elements in `nodeBlackList`.

I would also change the commit message to something like `stop adding` (instead of `remove`), because it gives the impression that something is being removed from the element.

marcysutton:
@gkalpak I applied the nodeBlacklist to `needsTabIndex` as opposed to `shouldAttachAttr`, limiting it to only tabindex part of ngModel. Does this qualify as a breaking change?

@Narretz there was already a test ensuring tabindex didn't get added to those elements (https://github.com/marcysutton/angular.js/blob/ngaria-tabindex2/test/ngAria/ariaSpec.js#L619). What was missing was an ngModel test ensuring a `null` default value wouldn't trigger it. Should I expand on that, or does this cover it?

Narretz:
@marcysutton Oh, I see. Although it's not only `null`, it happens with any value expect `undefined` when it's a radio.

I think we don't need this new test, we can make the original test better. The problem is that this test block https://github.com/angular/angular.js/blob/master/test/ngAria/ariaSpec.js#L619 behaves differently than tests that use `compileElement`, as in it doesn't digest after compilation. I've found this out when I was adding inputs with type checkbox and type radio to the list. Without a digest, the test fails only for type checkbox, with digest, it fails for checkbox and for radio.
I think since you said in https://github.com/angular/angular.js/issues/12492#issuecomment-148143542 that no native inputs should get tabindex, it makes sense to test for these types explicitly.

What do you think about replacing the above mentioned test with this one, and leaving the new test out?

```
    tthey('should not attach to native control $prop with ng-model', {
      'button': "<button ng-click='something'></button>",
      'a': "<a ng-href='#/something'>",
      'input[text]': "<input type='text' ng-model='val'>",
      'input[radio]': "<input type='radio' ng-model='val'>",
      'input[checkbox]': "<input type='checkbox' ng-model='val'>",
      'textarea': "<textarea ng-model='val'></textarea>",
      'select': "<select ng-model='val'></select>",
      'details': "<details ng-model='val'></details>"
    }, function(html) {
        compileElement(html);
        expect(element.attr('tabindex')).toBeUndefined();
    });
```

The `they` has the advantage that the element are compiled in isolation (doesn't really matter here), and that the expects is in the actual test block.

radio and checkbox would fail without the patch.

marcysutton:
@Narretz I like it! I've updated the test with the one you provided, and it worked like a charm. Thank you.

PR #13099, opened 2015-10-15
Fixes the issue "ng-bind not working correctly with <textarea> & <input> elements.

nozelrosario:
This issue mostly occurs in chrome because the current ng-bind does not update 'value' property of element. [IE somehow updates the property automatically]
Eg. can be found at plunk : http://plnkr.co/edit/w89iACaEl5fgJIo0svkh?p=preview
When using the ng-bind directive on `<textarea>` , the results are strange , But when using the modified version of ng-bind directive => 'ng-text-area-bind' it works as expected.

### DIFF
--- a/src/ng/directive/ngBind.js
+++ b/src/ng/directive/ngBind.js
@@ -60,7 +60,7 @@ var ngBindDirective = ['$compile', function($compile) {
         $compile.$$addBindingInfo(element, attr.ngBind);
         element = element[0];
         scope.$watch(attr.ngBind, function ngBindWatchAction(value) {
-          element.textContent = isUndefined(value) ? '' : value;
+          element.textContent = element.value = isUndefined(value) ? '' : value;
         });
       };
     }


Narretz:
ng-bind was not designed to work with inputs and textareas. You would normally use ngModel for that. ng-bind is used for binding to text, like interpolation is.

What's your usecase? Can't you use ng-model? Because using ng-bind this way will not update you scope values. Is this intended?

nozelrosario:
Yes, my need is only one way binding  **Model => View** , and ng-bind seems a good fit. also ng-value would  have worked, but it just updates the attribute "value". And in case of `<textarea>` the value is specified as inner text. eg. `<textarea> some text value </textarea>`
As shown in plunk  : http://plnkr.co/edit/uoaJ7q9OpVqAD0IjkL43?p=preview
I need a explicit custom  **View => Model** binding, which is achieved with ng-keypress & default Model=> View binding that ng-bind would provide.
But since ng-bind does not affect the 'value' property of the `<textarea>` element, the updated value isnt reflected / rendered on browser. That gives a strange behaviour on chrome. You might see that updating the model does not update the values in `<textarea>`
Though IE handles this problem, you would notice correct behaviour in IE. [Try executing the plunk in IE]

So modifying the ng-bind directive to update vaue attribute solves the issue as shown in plunk : http://plnkr.co/edit/w89iACaEl5fgJIo0svkh?p=preview
[ng-bind directive modified & re-named as ng-textarea-bind in plunk. ] 

You might see that updating the model also updates the values in `<textarea>`

nozelrosario:
Also this fix breaks the working of ng-bind with ng-options & use of ng-bind with `<option>` element, But shouldn't the right way of using `<option>` be using ng-value in combination with interpolation ? eg. `<option ng-value="optionValue"> {{optionLabel }} </option>`

Have also created a plunk to check ng-value with `<textarea>`
http://plnkr.co/edit/qO2SSmIQWVm8oaqhkofX?p=preview

Narretz:
I still don't think, we should mix ngBind behavior with ngModel behavior; it would probably be a breaking change and I don't like this to happen in an important directive like ngBind.
You can actually use ngModel to bind from scope -> input, and then use ngModelOptions debounce / updateOn to tweak the update behavior to the scope.
You can also create an alternative to the ngBind directive and publish it on Github to see if there's demand for something like this.

PR #13103, opened 2015-10-15
feat($injector): Allow specifying a decorator on $injector

lgalfaso:
Allows the definition of a decorator on `$injector`

### DIFF
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -688,6 +688,23 @@ describe('injector', function() {
           expect(log.join('; ')).
             toBe('myDecoratedService:input,dependency1; myService:decInput; dec+origReturn');
         });
+
+
+        it('should allow for decorators to $inject', function() {
+          injector = createInjector(['ng', function($provide) {
+            $provide.decorator('$injector', function($delegate) {
+              return extend({}, $delegate, {get: function(val) {
+                if (val === 'key') {
+                  return 'value';
+                }
+                return $delegate.get(val);
+              }});
+            });
+          }]);
+
+          expect(injector.get('key')).toBe('value');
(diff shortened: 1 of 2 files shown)


gkalpak:
OOC, what's the usecase ?

test/auto/injectorSpec.js
```diff
@@ -688,6 +688,24 @@ describe('injector', function() {
           expect(log.join('; ')).
             toBe('myDecoratedService:input,dependency1; myService:decInput; dec+origReturn');
         });
+
+
+        it('should allow for decorators to $inject', function() {
+          injector = createInjector(['ng', function($provide) {
+            $provide.value('dep', 'dependency');
```
gkalpak: What's this used for ? Did yoy mean for the key to be `'key'` or something (to show it gets overwritten).

lgalfaso: sorry, this is a leftover

lgalfaso:
> OOC, what's the usecase ?

The use case is to allow something that is not defined within an angular module to be injected. This would be useful in projects like ng-upgrade that need to make two different worlds coexist

test/auto/injectorSpec.js
```diff
@@ -688,6 +688,23 @@ describe('injector', function() {
           expect(log.join('; ')).
             toBe('myDecoratedService:input,dependency1; myService:decInput; dec+origReturn');
         });
+
+
+        it('should allow for decorators to $inject', function() {
```
gkalpak: _$inject_  -->  _$injector_ (?)
<sub>I would probably change it to something like: _should allow decorating $injector_</sub>

gkalpak:
LGTM <sub>(other than a minor nitpick in a test description :))</sub>

PR #13109, opened 2015-10-16
feat($resource): pass isArray to http config

tautvilas:
This will allow to indentify request type in global interceptor.

I need to identify if request is array or object and put necessary array or object to httpCache accordingly if request does not need to be made.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -574,7 +574,7 @@ angular.module('ngResource', ['ng']).
               undefined;
 
             forEach(action, function(value, key) {
-              if (key != 'params' && key != 'isArray' && key != 'interceptor') {
+              if (key != 'params' && key != 'interceptor') {
                 httpConfig[key] = copy(value);
               }
             });


gkalpak:
I don't get the usecase (and I suspect this might not be something commonly useful to include in core).
Could you provide more details on why you need this/how you will use this ? (An example maybe ?)

tautvilas:
You are right, this is not a common usecase. But, why drop isArray param? I would expect all data that I enter in resource configuration to be found in http config object. I can find cache, method, header and transformer params, so why no isArray?

gkalpak:
Hm...good point. As a work-around, you can pass your custom param (but you must be in charge of all `$resource`s and it's less than ideal). E.g.:

```js
var Thing = $resource('/api/thing/:id', {id: '@id'}, {
  query: {method: 'GET', isArray: true, itsAnArray: true}
});
// ...then look for `config.itsAnArray` in your interceptor
```

--
Regarding the question "why not to include `isArray`":
I guess it was skipped in the first place, because it is a `$resource`-specific property and doesn't have any usefulness for `$http`.
On the other hand, it is (kind of) related to/characterizing the request.

I could definitely leave with passing `isArray` to `$http`.
So, it LGTM after all.

Let's see what other people have to say.

tautvilas:
Thanks, I will implement temporary workaround and will wait for what you decide

Issue #13108, opened 2015-10-16
Why is the call to $apply needed in $http?


thorn0:
I started to think about it looking at #12557. At first glance the answer is obvious: we need to update the views after we got a response, so we trigger a digest. But hey, it'll be triggered anyway because `$http` uses `$q` internally. And when a `$q` deferred is resolved, it schedules a digest. So can't we just remove that call to `$rootScope.$apply` (with all the `$applyAsync` thing altogether)? Am I missing something? If it's needed only for the tests, the `$apply` call should be moved to `$httpBackend.flush`.

gkalpak:
Hey, this sounds about right. I would try replacing `$q` with `$$q` inside `$http`and keep `$http`'s apply (sync or async) - probably disable-able with an extra argument.

(This has been discussed before (more or less). See #2049 and https://github.com/angular/angular.js/issues/2049#issuecomment-59020072.)

--
@thorn0, would fancy giving it a shot ?

thorn0:
> I would try replacing $q with $$q inside $http and keep $http's apply (sync or async)

It won't work this way. The `$apply` call in `$http` is useless (or seems so if I do miss something) because it happens just after the response comes, but before any callbacks are called, which means before any scopes are modified. After `git blame`'ing, it seems to me that this `$apply` call is merely a historical rudiment.

gkalpak:
True. Yet, by merely removing `$http`'s apply, we can't have the `applyAsync` perf improvement (which we don't have now either I think).

--
So, I would still try replacing `$$q` with `$q` and add another callback at the end of the promise chain to make the digest (sync or async). That way we can easily have normal apply (using `$rootScope.$degest()`), async apply (using `$rootScope.$applyAsync()`) or add an extra option to skip apply altogether (in which case we should not add anything at the end of the chain).

<sub>Again, this is off the top of my head, so take it with a grain of salt :smiley:</sub>

gkalpak:
Or do you mean removing the [`$rootScope.$apply()`][1], but keep the [`$rootScope.$applyAsync(...)`][2] ?
That would work as well, I guess, but we would still get an extra $digest in the async case (I think).

Plus, replacing `$q` with `$$q` would also enable skipping apply altogether (as an option).

[1]: https://github.com/angular/angular.js/blob/v1.5.0-beta.1/src/ng/http.js#L1275
[2]: https://github.com/angular/angular.js/blob/v1.5.0-beta.1/src/ng/http.js#L1272

[...]

PR #13111, opened 2015-10-16
Not a 'real' PR, but a demo for #13108

thorn0:
@gkalpak please check out this branch and run `grunt test:unit`. You'll see that removing the `$apply`/`$applyAsync` call has the same effect as using `$applyAsync`.


thorn0:
Basically, namesake, you don't even need to checkout anything locally. Travis has run the tests for us.

PR #13119, opened 2015-10-19
docs(select): add multiple attribute

Puigcerber:
In the documentation of the [select](https://docs.angularjs.org/api/ng/directive/select) directive there is an example of multiple selection, but the multiple attribute is not listed as an argument, so I think it can be a bit confusing.

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -187,6 +187,8 @@ var SelectController =
  *
  * @param {string} ngModel Assignable angular expression to data-bind to.
  * @param {string=} name Property name of the form under which the control is published.
+ * @param {string=} multiple Allows multiple options to be selected. The selected values will be bound to the model
+ * as an array.
  * @param {string=} required Sets `required` validation error key if the value is not entered.
  * @param {string=} ngRequired Adds required attribute and required validation constraint to
  * the element when the ngRequired expression evaluates to true. Use ngRequired instead of required


gkalpak:
I think this is because `multiple` doesn't have a special meaning in Angular (as far as the user is concerned at least). It is standard HTML for allowing multiple options to be selected. (Of course Angular does certain things under the hood to support multiple values, but that is not relevan here imo).

I am not sure if this needs to be documented.

Puigcerber:
I know. But in the same fashion both `name` and `required` are added to the documentation. Should we maybe write a note saying that all the HTML5 attributes are valid?

gkalpak:
`name` and `required` are documented because they carry extra semantics for Angular (e.g. `name` is the property under which the control is registered with its parent form and `required` sets the corresponding error key in the control's and its parent forms' error objects).
Also note that other HTML5 attributes (e.g. `read-only` or `disabled` etc) are not documented either.

Anyway, since `multiple` changes the bound value (into an array), we could document that.
Would you like to update the PR, explaining that adding `multiple` will result into an array of values as model-value ?

Puigcerber:
That makes perfect sense. I'll update the pull request. Thanks @gkalpak!

gkalpak:
Thx @Puigcerber ! Feel free to ping me when you're done, so I can take another look.

Puigcerber:
Hey @gkalpak! I just updated it.

gkalpak:
Merged. Thx @Puigcerber :+1:
(Backported to `v1.4.x` as 5758d73.)

PR #13124, opened 2015-10-19
fix(ngMockE2E): allow $httpBackend.passThrough() to work when ngMock is loaded

### LINKED ISSUE #1434: ngMockE2E $httpBackend "passThrough()" doesn't work
Hi, I'm trying to set up some test that actually do hit the backend, but I can't get this to work:

$httpBackend.whenPOST(/.\*/).passThrough()
$httpBackend.whenGET(/.\*/).passThrough()

But I get the error: "POST /some/url No more request expected"

I am still able to use .whenGET().respond(), and it's variants, to mock the requests. Am I doing something wrong with passThrough()?

schmod:
Allow $httpBackend.passThrough() to work normally when `ngMock` is loaded concurrently with `ngMockE2E`.  This also allows one to use `angular.mock.module()` to write tests that depend on `ngMockE2E`, as `module()` always adds a dependency on `ngMock`.

Fixes #1434

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2129,6 +2129,7 @@ angular.module('ngMock', ['ng']).provider({
  * the {@link ngMockE2E.$httpBackend e2e $httpBackend} mock.
  */
 angular.module('ngMockE2E', ['ng']).config(['$provide', function($provide) {
+  $provide.value('$httpBackend', angular.injector(['ng']).get('$httpBackend'));
   $provide.decorator('$httpBackend', angular.mock.e2e.$httpBackendDecorator);
 }]);
 
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -1846,12 +1846,12 @@ describe('ngMockE2E', function() {
     var hb, realHttpBackend, callback;
 
     beforeEach(function() {
-      module(function($provide) {
-        callback = jasmine.createSpy('callback');
+      callback = jasmine.createSpy('callback');
+      angular.module('ng').config(function($provide) {
         realHttpBackend = jasmine.createSpy('real $httpBackend');
         $provide.value('$httpBackend', realHttpBackend);
-        $provide.decorator('$httpBackend', angular.mock.e2e.$httpBackendDecorator);
       });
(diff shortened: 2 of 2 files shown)


schmod:
@googlebot I signed it!

schmod:
What does one need to do to get a PR merged around here?

Narretz:
Hi @schmod I see that this PR is a response to https://github.com/angular/angular.js/issues/1434

Can you give me a quick scenario where you need both ngMock and ngMockE2E in the same app?

schmod:
Hi @Narretz,

For me, the passthrough is useful when I have a test fixture that cannot be easily inlined into JavaScript (such as an image).  

I have an Angular directive and service that convert (user-provided) images into clickable thumbnails when they exceed a certain size.

The most straightforward (and realistic) way to test this functionality is to provide a real image as my test fixture, and actually let the browser make a real HTTP request to grab the image.

I suspect that there are also many other integration testing scenarios that could benefit from the semantics of a unit test, with occasional "real" HTTP requests thrown in.  This workflow is difficult to achieve in practice, because the `window.inject()` function provided by angular-mocks load `ngMock`, regardless of whether I ask for it.

Hope this helps!  I realize that my proposed solution isn't the most ideal, but it doesn't seem like it would make things *worse* for anybody.

petebacondarwin:
LGTM - merging

mike-bresnahan:
This change had a bad side effect on protractor. Protractor no longer waits for pass through requests.

https://github.com/angular/protractor/issues/3948

mike-bresnahan:
I'm not an expert on Angular internals, but this change smells bad because it causes 2 instances of $httpBackend and dependent services (e.g. $browser) to exist and that violates the implied "all services are singletons" policy of Angular. This is bound to cause people issues (e.g. protractor).

petebacondarwin:
@mike-bresnahan - can you create a new issue for this. I see that the problem is that calling `angular.injector(...)` creates a completely new injector and that not only creates a new backend but also an inaccessible copy of the `$browser` service.

mike-bresnahan:
Petebacondarwin, yes that is my understanding of the root cause as well and I have created a new issue for it.

https://github.com/angular/angular.js/issues/15593

PR #13140, opened 2015-10-21
 optimize angularInit function . do not foreach candidate if the element is retrived

omarProgrammer:
…ent is retrive


omarProgrammer:
If the  element Module is retrived then  do not  foreach for retrive a candidate children element

PR #13147, opened 2015-10-21
docs(modules):  example of testing run blocks

### LINKED ISSUE #12706: How do you test the run block?
http://stackoverflow.com/questions/32169240/how-should-the-run-block-be-dealt-with-in-angular-unit-tests

Would be useful to include a section on it in the [Unit Testing Guide](https://docs.angularjs.org/guide/unit-testing).

iamisti:
Add:
- example to the documentation of unit testing run blocks.

Closes #12706

### DIFF
--- a/docs/content/guide/module.ngdoc
+++ b/docs/content/guide/module.ngdoc
@@ -302,3 +302,25 @@ describe('myApp', function() {
   });
 });
 ```
+
+## Unit Testing Run blocks
+
+Sometimes you would like to test your run blocks. However it's not so obvious at the moment.
+
+```js
+it('should call something in a run block', function() {
+
+    // angular.module() will load your module which has the run block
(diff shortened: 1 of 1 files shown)


petebacondarwin:
My personal view is that you should not put code that needs unit testing in run blocks. Move it to a service, test it there, then simply call it from a run block. So I would rather not merge this as it implies that we would recommend such a practice.

iamisti:
@petebacondarwin I agree with you. Best practice is not having code in run blocks. But still even if you have your logic behind a service method, you might want to check if you call that service in your run block (making sure that nobody will just simply remove it with a refactor or whatever).

petebacondarwin:
Well you can check this by simply loading the module into the injector. By definition that will run the modules run blocks.

iamisti:
@petebacondarwin so true! :) Thanks for clarifying that.

PR #13151, opened 2015-10-22
feat(ngModel): pass the previous and current model to view change listeners

luaks:


Narretz:
Technically, when ngChange is called, the model should always have been changed. What's your use case for this?

luaks:
We are using this, to compare the old and new values of the form elements.

When autosaving we build a model from the valid values to be sent to the server, but want to keep the possibly invalid new value, for the user to still be editable, even if they leave the invalid field.

PR #13152, opened 2015-10-22
docs(tutorial/10 - Event Handlers): missing unit test

sandormajor:
In the doc there stands "You also have to refactor one of your unit tests because of the addition of the `mainImageUrl`
model property" but the spec was missing the test of `mainImageUrl`

### DIFF
--- a/docs/content/tutorial/step_10.ngdoc
+++ b/docs/content/tutorial/step_10.ngdoc
@@ -138,7 +138,9 @@ __`test/unit/controllersSpec.js`:__
       expect(scope.phone).toBeUndefined();
       $httpBackend.flush();
 
-      expect(scope.phone).toEqual(xyzPhoneData());
+      var xyzPhoneInstance = xyzPhoneData();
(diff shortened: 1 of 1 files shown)


Airborn22:
I signed it!

Airborn22:
There is a pull request for the phonecat app but it's failing now because of an issue on master: https://github.com/angular/angular-phonecat/pull/279

Narretz:
Thanks for the PR. The way I see it, the description is accurate: You need to update the test, because the `phone` on the scope is different now. We can add an additional expect/test for the mainImageUrl, but it is already tested in the e2e test.

Airborn22:
Thank you @Narretz, I created a pull request for the test but all the builds are failing there for some reason. I will create a new pull request later.

Narretz:
You don't have to open a new PR, you can push --force to your branch and it iwll update the PR.

gkalpak:
The tutorial has changed quite a bit (with #14416).
If you think this change is still relevant, please submit a new PR against the latest code.

Thx !

PR #13156, opened 2015-10-23
docs(tutorial/0 - Bootstrapping): describe your change...

349989153:
I think it's more clear to specify which element should be considered as the root element.


gkalpak:
I don't think this change is correct ("grammar-wise"). Any native speakers (e.g. @petebacondarwin) ?

petebacondarwin:
Sorry I agree. The change is not grammatically correct. The original sentence already indicates that we are defining the root element.

PR #13161, opened 2015-10-23
ignore: Will redo with a more correct implementation.

shelbymorris90:
Fixes an issue where the attribute 'disabled' was assigned to an element, but the value of the attribute wasn't enforced. This correctly assigns the 'disabled' attribute with the value 'disabled' (Fixes CSS selectors such as option[disabled] and option[disabled="disabled"], et al.) This is more consistent with other angular implementations (see: ngDisabled)


shelbymorris90:
I signed it.

Issue #13169, opened 2015-10-26
BUG: optional multi-slot transclusion with jquery


petebacondarwin:
@jtrancas commented 6 hours ago:

I'm totally new to AngularJS, however, I believe I found an issue when jQuery is being included along with Angular, I could not reproduce the error when Angular is running without jQuery.
The situation is the following:

* jQuery is included.
* There's a directive with all the transclude slots marked as optional.
* There's a directive occurrence with no child slots defined.

You can check the following Plnkr: http://plnkr.co/edit/f9SN35IMPrmeyuPz0Pih?p=preview
The pane directive has no child nodes, all its transclude slots are optional, and you may notice the following error being raised in the console:

```
TypeError: Cannot read property 'replace' of undefined
    at directiveNormalize (angular.js:9033)
    at angular.js:8095
    at forEach (angular.js:354)
    at applyDirectivesToNode (angular.js:8094)
    at compileNodes (angular.js:7620)
    at compileNodes (angular.js:7632)
    at compileNodes (angular.js:7632)
    at compile (angular.js:7527)
    at angular.js:1677
    at Scope.$eval (angular.js:16127)
```

If you remove the jQuery inclusion it all goes well.

lgalfaso:
it looks like this is a regression from 2c8d87e064dca99a49ed35d1db885b1f2e40dcf4
#13171 should take care of this

PR #13171, opened 2015-10-26
fix(forEach): handle jQuery objects of length 0

lgalfaso:
Closes: #13169

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -193,7 +193,8 @@ msie = document.documentMode;
 
 function isNodeList(obj) {
   return typeof obj.length == 'number' &&
-         typeof obj.item == 'function';
+         (typeof obj.item == 'function' ||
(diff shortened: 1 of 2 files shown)


src/Angular.js
```diff
@@ -193,7 +193,8 @@ msie = document.documentMode;
 
 function isNodeList(obj) {
   return typeof obj.length == 'number' &&
-         typeof obj.item == 'function';
+         (typeof obj.item == 'function' ||
+          isString(obj.jquery));
```
petebacondarwin: It is actually easier to identify jQuery and jqLite instances using `instanceof jqLite`.

lgalfaso:
thanks @petebacondarwin for the cleanup

petebacondarwin:
Thanks @lgalfaso for the original fix

PR #13175, opened 2015-10-26
fix(ngAnimate): do not alter the provided options data

### LINKED ISSUE #13040: animateCss shoul not alter the original options object
When using `$animateCss` if the options passed are stored in a variable after the animation the original options object is altered and using it in a new animation will fail (as you can see  on codepen http://codepen.io/anon/pen/avWgaX).

If a copy of the configuration object (made using `angular.copy`) is provided the configuration object is reusable.
But forcing the developer to copy the configuration object each time he wants to use it is quite annoying. And copying the whole object might be a performance bottleneck.

If altering some elements of the configuration object is needed, this should not affect the original provided object.

Tested on Angular 1.4.7

matsko:
Prior to this fix the provided options object would be
altered as the animation kicks off due to the underlying
mechanics of ngAnimate. This patch ensures that a
copy of the provided options is used instead. This patch
also works for when `$animateCss` is used by itself.

Closes #13040

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -130,6 +130,24 @@ describe("animations", function() {
       };
     }));
 
+    it("should not alter the provided options input in any way throughout the animation", inject(function($animate, $rootElement, $rootScope) {
+      var initialOptions = {
+        from: { height: '50px' },
+        to: { width: '50px' },
+        addClass: 'one',
+        removeClass: 'two'
+      };
+
+      var copiedOptions = copy(initialOptions);
+      expect(copiedOptions).toEqual(initialOptions);
+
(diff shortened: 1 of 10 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -130,6 +130,24 @@ describe("animations", function() {
       };
     }));
 
+    it("should not alter the provided options input in any way throughout the animation", inject(function($animate, $rootElement, $rootScope) {
```
gkalpak: `$rootElement` is not needed.

Narretz: If you remove the patch, this is the only test that fails. Shouldn't all tests fail if the patch is removed?

Narretz:
Sorry, I merged this by mistake! :(

PR #13180, opened 2015-10-27
docs: Fix broken link to HTML 5 spec in $anchorScroll.

srawlins:


gkalpak:
Backported to `v1.4.x` as 62ed26a.

PR #5420, opened 2013-12-16
fix($timeout.$flush): make $flush follow the timeline

lgalfaso:
Make the mock $timeout.$flush follow the timeline when the function
being calls $timeout

PR #13183, opened 2015-10-27
feat(ngMock) add atomic flushing strategy for $timeout

twlvl:
Hi dear Angular community,

During writing unit tests I found one interesting issue. I need to write tests for some code like this:

``` javascript
$timeout(function() {
 action1();
 $timeout(function() {
  action2();
 });
});
```

And I'm expecting that I need to execute `$timeout.flush()`  2 times. First time to reach `action1` function and the second time to execute `action2` function (let's call this kind of flush _"atomic"_ for example). But for now, for the first flush() it executes the both handlers. One after another.  Please check this example: http://jsbin.com/fuziri/edit?js,output .

On the other hand we have some cases in tests when this buggy, for the first look, behaviour can simplify our lifes. In such case for example:

``` javascript
 var callback = jasmine.createSpy();
 $timeout(function() {}, 0, false).then(callback);
 $timeout.flush();
 expect(callback).toHaveBeenCalled();
```

Here `$timeout` register new defered task to process callback using `$browser.defer` during flushing. So to execute callback (if you are using that _atomic_ flush strategy) you should call `$timeout.flush()` twise.

I think that the both strategies of flushing are pretty useful. And may be it will be a good idea to:
- implement that _atomic_ flushing
- to allow to switch between the strategies using some additional flag

What do you think, guys?

P.S. this is a duplicate of #13146, I opened a new one because can't solve the certificate issue for commit with wrong email address.

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -103,8 +103,11 @@ angular.mock.$Browser = function() {
    * Flushes all pending requests and executes the defer callbacks.
    *
    * @param {number=} number of milliseconds to flush. See {@link #defer.now}
+   * @param {boolean=} [atomic=false] If set to `true` — will handle only tasks
+   * available at the moment of execution, otherwise will process tasks that was
+   * added during execution as well.
    */
-  self.defer.flush = function(delay) {
+  self.defer.flush = function(delay, atomic) {
     if (angular.isDefined(delay)) {
       self.defer.now += delay;
     } else {
@@ -115,8 +118,25 @@ angular.mock.$Browser = function() {
       }
     }
 
-    while (self.deferredFns.length && self.deferredFns[0].time <= self.defer.now) {
-      self.deferredFns.shift().fn();
+    function shiftExecutableFns(handler) {
+      while (self.deferredFns.length && self.deferredFns[0].time <= self.defer.now) {
+        handler(self.deferredFns.shift());
+      }
+    }
+
+    if (atomic) {
+      var fnsToRun = [];
+      shiftExecutableFns(function(task) {
+        fnsToRun.push(task);
+      });
+
+      angular.forEach(fnsToRun, function(item) {
+        item.fn();
+      });
+    } else {
+      shiftExecutableFns(function(task) {
+        task.fn();
+      });
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/ngMock/angular-mocks.js
```diff
@@ -103,8 +103,11 @@ angular.mock.$Browser = function() {
    * Flushes all pending requests and executes the defer callbacks.
    *
    * @param {number=} number of milliseconds to flush. See {@link #defer.now}
+   * @param {boolean=} [atomic=false] If set to `true` — will handle only tasks
+   * available at the moment of execution, otherwise will process tasks that was
```
atogata: that were*

src/ngMock/angular-mocks.js
```diff
@@ -1948,9 +1968,11 @@ angular.mock.$TimeoutDecorator = ['$delegate', '$browser', function($delegate, $
    * Flushes the queue of pending tasks.
    *
    * @param {number=} delay maximum timeout amount to flush up until
+   * @param {boolean=} [atomic=false] If set to `true` — will not run the tasks
+   * added during execution of flash even if they added with `0` timeout.
```
atogata: flush*
they were*

atogata:
+1

elkeis:
@atogata, thanks. Spelling corrected.

petebacondarwin:
@elkeis - did you know that you can pass a `delay` to `$timeout.flush(delay)`, which will only flush timeouts that should have completed by the given delay. This is not quite what you want, perhaps, since if you have two timeouts with 0 delay then they would both always be flushed at the same time.
But this can help with testing sequencing issues.

petebacondarwin:
Oh I just looked at the code. I see that you want to ensure that new calls to timeout that are triggered as part of flushing the current timeouts are not triggered...

petebacondarwin:
Interesting idea. I think this would need some more API thought as the current `atomic` parameter is not that intuitive to me.

petebacondarwin:
This is very vaguely related to #5420. It would be interesting to solve both at once...

petebacondarwin:
@elkeis once #14686 lands then as long as your timeouts have a non-zero delay, then they will indeed become separable in tests.

Given your original example, if I add 1ms second delays:

```js
$timeout(function() {
 action1();
 $timeout(function() {
  action2();
 },1);
}, 1);
```

Then calling `$timeout.flush(1)` will only trigger `action1()`. Calling `$timeout.flush(1)` a second time will trigger `action2()`.

Is this enough for your testing?

petebacondarwin:
Closing as #14686 has landed and there has been activity on this PR for 10 months.

Issue #13127, opened 2015-10-19
Unit testing: Error thrown during promise resolution causes "$digest already in progress" message for the *next* call to $scope.$apply()


gregjacobs:
Angular 1.4.7

```javascript
it( 'thrown errors and $digest', inject( function( $rootScope, $q ) {
    $q.when().then( function() { 
        throw new Error( 'Test Error' ); 
    } );

    expect( function() {
        $rootScope.$apply();  // to finalize the promise
    } ).toThrowError( 'Test Error' );

    $rootScope.$apply();  // **digest already in progress** 
} ) );
```

Our codebase throws many errors to alert developers of problems, but we're having trouble testing them. This minimal test case showcases the problem.

Narretz:
Why are you adding the second $rootSCope.$apply()? Is something coming afterwards in this test? To me, it doesn't make sense to have essentially two $rootScope.$apply() after another.

gregjacobs:
It's just a minimalist example of the problem. 

An error thrown during a digest doesn't allow Angular to know that it's no longer in a digest.

Narretz:
Okay. I don't know maybe there's a way to manually clear the digest loop.
However, to keep the test focused shouldn't you simply test that the error
is thrown and nothing more?
Am 27.10.2015 21:18 schrieb "Gregory Jacobs" <notifications@github.com>:

Narretz:
I'm going to close this issue because we haven't got any feedback. Feel free to reopen this issue if you can provide new feedback.

PR #13184, opened 2015-10-27
fix($rootScope): reset internal phase on digest exception

lgalfaso:
Cleanup the internal phase when there is a non-handled exception.

Closes: #13127

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -730,98 +730,100 @@ function $RootScopeProvider() {
             logIdx, logMsg, asyncTask;
 
         beginPhase('$digest');
-        // Check for changes to browser url that happened in sync before the call to $digest
-        $browser.$$checkUrlChange();
+        try {
+          // Check for changes to browser url that happened in sync before the call to $digest
+          $browser.$$checkUrlChange();
 
-        if (this === $rootScope && applyAsyncId !== null) {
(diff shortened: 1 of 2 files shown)


lgalfaso:
Sorry, it looks like this broke an existing test, looking into it now

gkalpak:
I am not sure this is the right thing to do. There are already several try-catch blocks inside `$digest`.

--
AFAICT, the issue with phase not being cleared up only happens in tests (when using the mocked `$exceptionHandler`). The exception that causes this is already caught inside the `$digest()` and passed to the `$exceptionHandler`. Normally, the `$exceptionHandler` just logs the error and goes on.
The mocked handler used in tests (as part of `ngMock`) rethrows the error and that is why the phase doesn't get cleared.
Maybe, we could solve it more elegantly by taking care of it inside the mocked `$exceptionHandler`.

Still, this would not solve the issue in case someone overwrites the `$exceptionHandler` to actually throw errors, but anyone doing so would be asking for it anyway :stuck_out_tongue: 

--
I'm not saying this is wrong or that I know of a better way (have only looked into it superficially) - just _raising some concern_ :)

BTW, is there any perf-hit by the extra try-catch block ? This path is super-hot, isn't it ?

lgalfaso:
The more I think about this, then the more I think @gkalpak is right. The proper solution would need to come from *angular-mock* and `$exceptionHandler`

PR #13185, opened 2015-10-28
feat(ngAnimate): introduce ngSwap directive

matsko:


src/ngAnimate/ngSwap.js
```diff
@@ -0,0 +1,112 @@
...
+ *  .green { background:green; }
+ *  .blue { background:blue; }
+ *  .yellow { background:yellow; }
+ *  .orange { background:orange; }
+ *  </file>
+ * </example>
+ */
+var ngSwapDirective = ['$animate', '$rootScope', function($animate, $rootScope) {
+  return {
+    restrict: 'EA',
```
gkalpak: The `E` part is undocumented.

src/ngAnimate/ngSwap.js
```diff
@@ -0,0 +1,112 @@
...
+    transclude: 'element',
+    terminal: true,
+    priority: 600,
+    compile: function(elm, attrs) {
```
gkalpak: Why not just use `link` (since it doesn't do anything during compilation) ?

test/ngAnimate/ngSwapSpec.js
```diff
@@ -0,0 +1,175 @@
+'use strict';
+
+describe("ngSwap", function() {
+
+  beforeEach(module('ngAnimate'));
+  beforeEach(module('ngAnimateMock'));
+
+  var element;
+  afterEach(inject(function($$jqLite) {
```
gkalpak: Why the `inject` and `$$jqLite` ?

test/ngAnimate/ngSwapSpec.js
```diff
@@ -0,0 +1,175 @@
...
+    dealoc(element);
+  }));
+
+  it('should render a new container when the expression changes', inject(function($rootScope, $compile, $animate) {
```
gkalpak: Instead of injecting `$rootScope`, `$compile`, `$animate` in every test, why not inject them in a `beforeEach` block and make them available to all tests ?

test/ngAnimate/ngSwapSpec.js
```diff
@@ -0,0 +1,175 @@
...
+    $rootScope.$digest();
+
+    var first = element.find('div')[0];
+    expect(first).toBeFalsy();
+
+    $rootScope.exp = 'yes';
+    $rootScope.$digest();
+
+    var second = element.find('div')[0];
+    expect(second.innerText).toBe('yes');
```
gkalpak: I think `textContent` is preferred over `innerText`.
<sub>(Not that it makes a real difference in tests, but `innerText` used to lead to layout thrashing.)</sub>

test/ngAnimate/ngSwapSpec.js
```diff
@@ -0,0 +1,175 @@
...
+
+    var three = element.find('div')[0];
+    expect(three.innerText.trim()).toBe('2 2 3');
+
+    $rootScope.exp = { c: 4 };
+    $rootScope.$digest();
+
+    var four = element.find('div')[0];
+    expect(four.innerText.trim()).toBe('4');
+  }));
```
gkalpak: I don't think this test tests what it claims it tests.
Basically, what you want to test is that elements are swapped, not just that the `textContent` updates based on `exp`.
You should test that `one` !== `two` !== `three`.

test/ngAnimate/ngSwapSpec.js
```diff
@@ -0,0 +1,175 @@
...
+
+      $rootScope.value = value;
+      $rootScope.$digest();
+
+      var two = element.find('div')[0];
+      expect(two).toBeFalsy();
+    });
+  });
+
+  it('should consider "0" as a truthy value',
```
gkalpak: Technically `"0"` is truthy anyway. It's `0` that we're interested in :)

gkalpak:
This is nice, but I wonder if it belongs in ngAnimate-core (instead of being a 3rd-party module).
Is there a common usecase for this (or something we want to support) ?

matsko:
Thank you @gkalpak for the awesome code review. The reason why this code is in the ngAnimate module is because the directive doesn't do anything when animations are removed. If you can think of a way to integrate a lightweight pagination system into it then we could have something like `ng-page` which does the same stuff here but with some from of pagination.

We could also rename this to `ng-animate-swap`, but we do plan on having `*ng-swap` in Angular2.

gkalpak:
@matsko, my question is whydo we need to have this directive in core instead of having it as a 3rd-party module (or incrporating it is a UI framework such as `ngMaterial`) ?
It seems more like a widget than a feature of the framework (and in fact it is implemented with no access to internal stuff (neither core nor ngAnimate)).

The same question would apply to `*ng-swap` in Angular2 :smile: 

--
But, I guess if you decided to introduce it, there is probably some usecase that you want to support (and think is common/important enough), right ? So, what is it ? (Yes, I'm a curious person :stuck_out_tongue: )

petebacondarwin:
I am pretty cool with this going into 1.6 (we are about to lock down 1.5 so that we can get it released).

It is pretty small and so I don't feel too bad about including it in `angular-animate.js`. It seems rather overkill to create a whole new module in the `angular.js` project for it.

Narretz:
Landed as https://github.com/angular/angular.js/commit/78297d252de7c80f73ecf9e291ed71bd52578361

PR #13188, opened 2015-10-28
feat($templateRequest): support configuration of accept headers

### LINKED ISSUE #6860: angular directive templateUrl should allow "text/html" Accept and content-type headers

petebacondarwin:
It is now possible to configure the Accept header for template requests.
If a value is configured then this will override only the Accept property
of the headers that are passed.
If no value is configured then the request will use the default $http
Accept header.

Thanks to @luckycadow for help on this feature

Closes #11868
Closes #6860

### DIFF
--- a/src/ng/templateRequest.js
+++ b/src/ng/templateRequest.js
@@ -3,26 +3,55 @@
 var $compileMinErr = minErr('$compile');
 
 /**
- * @ngdoc service
- * @name $templateRequest
- *
+ * @ngdoc provider
+ * @name $templateRequestProvider
  * @description
- * The `$templateRequest` service runs security checks then downloads the provided template using
- * `$http` and, upon success, stores the contents inside of `$templateCache`. If the HTTP request
- * fails or the response data of the HTTP request is empty, a `$compile` error will be thrown (the
- * exception can be thwarted by setting the 2nd parameter of the function to true). Note that the
- * contents of `$templateCache` are trusted, so the call to `$sce.getTrustedUrl(tpl)` is omitted
- * when `tpl` is of type string and `$templateCache` has the matching entry.
- *
- * @param {string|TrustedResourceUrl} tpl The HTTP request template URL
- * @param {boolean=} ignoreRequestError Whether or not to ignore the exception when the request fails or the template is empty
- *
- * @return {Promise} a promise for the HTTP response data of the given URL.
- *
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@petebacondarwin the code looks good to me, but I wonder if the current design is not too restrictive. Maybe we should allow people to set any headers and not limit ourselves to `Accept` only?

If all the use-cases we know of are boiling down to setting `Accept` I'm fine with landing this as-is. But maybe we should make it more generic?

petebacondarwin:
@pkozlowski-opensource thanks for the review. I think that you are right. It might even bring the code size down a touch! I will update.

petebacondarwin:
@pkozlowski-opensource - any chance you can take another look?

src/ng/templateRequest.js
```diff
@@ -45,12 +79,10 @@ function $TemplateRequestProvider() {
...
-      var httpOptions = {
-        cache: $templateCache,
-        transformResponse: transformResponse
-      };
-
-      return $http.get(tpl, httpOptions)
+      return $http.get(tpl, extend({
+          cache: $templateCache,
+          transformResponse: transformResponse
+        }, httpOptions))
```
pkozlowski-opensource: Shouldn't it be other way around? I mean, we should always enforce `cache: $templateCache` and certain transform, no?

pkozlowski-opensource: Could you add a test for a case where someone tries to specify cache?

petebacondarwin: The point of modifying the `transformResponse` is that the default `$http` service adds in a JSON transformation, which we don't want in `$templateRequest`.

If we are going to allow people to configure the options then we should allow them to override everything, including this `transformResponse` and also to provide their own cache if we really wanted them to; although the same could be achieved by overriding the `$templateCache` service itself.

I can add a test for overriding the `cache` but it is really just the same code path as overriding the `transformResponse`.

petebacondarwin: Doh! I see what you mean. You cannot create a cache object in the provider...

PR #13198, opened 2015-10-29
docs(tutorial/Tutorial): add to PATH in Windows

bladefistx2:
user may need to add the directory to the PATH variable.

### DIFF
--- a/docs/content/tutorial/index.ngdoc
+++ b/docs/content/tutorial/index.ngdoc
@@ -105,6 +105,9 @@ Check the version of Node.js that you have installed by running the following co
 node --version
 ```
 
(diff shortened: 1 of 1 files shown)


bladefistx2:
Signed.

gkalpak:
@bladefistx2, the Node installer takes care of that automatically. Did you run into this issue yourself ? 

In any case, I think it is outside of the scope of the Angular tutorial to provide troubleshooting instructions for Node.js installation.
<sub>
BTW, I noticed that the download link mentioned a couple of lines above seems to have changed.
The url should be `https://nodejs.org/en/download/`. Feel free to submit a PR for that :smiley:
</sub>

gkalpak:
Closed with c2033d7.

PR #13199, opened 2015-10-29
docs/test($templateRequest): minor docs improvement and tests clean-up

gkalpak:


src/ng/templateRequest.js
```diff
@@ -6,7 +6,10 @@ var $compileMinErr = minErr('$compile');
  * @ngdoc provider
  * @name $templateRequestProvider
  * @description
- * Used to configure the Accept header that is sent to the server when requesting a template.
+ * Used to configure the options passed to the {@link $http} service when making a request.
```
Narretz: It's still *only* for template requests, I think that should be explicit.

gkalpak: Good point :+1:

gkalpak: Changed it. PTAL.

PR #13208, opened 2015-10-30
feat(checkbox): ngIndeterminateValue

calebcauthon:
adds support for an inderminate value in the same pattern as
ngTrueValue and ngFalseValue

Closes #1411

Here's a jsFiddle that demonstrates the feature: http://jsfiddle.net/tyk7kybe/


Narretz:
Thanks for the PR.
I see that due to the nature of "indeterminate", we cannot observe an attribute like we would normally do. But watching DOM attributes is actually a very bad idea, as it is much slower than watching primitives and objects.
We should probably have an attribute like `ng-indeterminate` that takes an expression, and sets `indeterminate` on the element.
Also, `indeterminate` is just visual style, and doesn't mean the checkbox can't be currently in a checked / unchecked state. So I think we should call $setViewValue() only if there's an actual `ng-indeterminate-value`.

calebcauthon:
Thanks for the info! I'll offer that up!

calebcauthon:
I opened a new PR with the suggested changes at #13216

PR #13210, opened 2015-10-30
feat($resource): add support for cancelling requests

### LINKED ISSUE #9332: Can no longer cancel $resource request with a promise

gkalpak:
Introduced changes:
- Deprecate passing a promise as `timeout` (for `$resource` actions).
  It never worked correctly anyway.
  Now a warning is logged (using `$log.debug()`) and the property is
  removed.
- Add support for a boolean `cancellable` property to actions'
  configuration, the `$resource` classes `options` of the
  `$resourceProvider`'s defaults.
  If true, the `$cancelRequest` method (added to all returned values for
  non-instance calls) will abort the request (if it's not already
  completed or aborted).
  If there is `timeout` specified on the action's configuration, the value
  of `cancellable` is ignored.

Example usage:

``` js
  get: {
    method: 'GET',
    cancellable: true
  }
});

var currentPost = Post.get({id: 1});
...
// A moment later the user selects another post, so
// we don't need the previous request any more
currentPost.$cancelRequest();
currentPost = Post.get({id: 2});
...
```

Fixes #9332
Closes #13050
Closes #13058

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -365,7 +365,7 @@ angular.module('ngResource', ['ng']).
       }
     };
 
-    this.$get = ['$http', '$q', function($http, $q) {
+    this.$get = ['$http', '$log', '$q', function($http, $log, $q) {
 
       var noop = angular.noop,
         forEach = angular.forEach,
@@ -525,6 +525,22 @@ angular.module('ngResource', ['ng']).
         forEach(actions, function(action, name) {
           var hasBody = /^(POST|PUT|PATCH)$/i.test(action.method);
 
+          var hasTimeout = action.hasOwnProperty('timeout');
+          if (hasTimeout && !angular.isNumber(action.timeout)) {
+            $log.debug('ngResource:\n' +
+                       '  Only numeric values are allowed as `timeout`.\n' +
+                       '  Promises are not supported in $resource, because the same value has to ' +
+                       'be re-used for multiple requests. If you are looking for a way to cancel ' +
+                       'requests, you should use the `cancellable` option.');
+            delete action.timeout;
+            hasTimeout = false;
+          }
+          action.cancellable = hasTimeout ?
+              false : action.hasOwnProperty('cancellable') ?
+              action.cancellable : (options && options.hasOwnProperty('cancellable')) ?
+              options.cancellable :
+              provider.defaults.cancellable;
+
           Resource[name] = function(a1, a2, a3, a4) {
             var params = {}, data, success, error;
 
@@ -581,6 +597,7 @@ angular.module('ngResource', ['ng']).
                 case 'params':
                 case 'isArray':
                 case 'interceptor':
+                case 'cancellable':
                   break;
                 case 'timeout':
                   httpConfig[key] = value;
@@ -588,14 +605,27 @@ angular.module('ngResource', ['ng']).
               }
             });
 
+            if (!isInstanceCall) {
+              if (!action.cancellable) {
+                value.$cancelRequest = angular.noop;
+              } else {
+                var deferred = $q.defer();
+                httpConfig.timeout = deferred.promise;
+                value.$cancelRequest = deferred.resolve.bind(deferred);
+              }
+            }
+
             if (hasBody) httpConfig.data = data;
             route.setUrlParams(httpConfig,
               extend({}, extractParams(data, action.params || {}), params),
               action.url);
 
-            var promise = $http(httpConfig).then(function(response) {
+            var promise = $http(httpConfig).finally(function() {
+              if (value.$cancelRequest) value.$cancelRequest = angular.noop;
+            }).then(function(response) {
               var data = response.data,
-                promise = value.$promise;
+                promise = value.$promise,
+                cancelRequest = value.$cancelRequest;
 
               if (data) {
                 // Need to convert action.isArray to boolean in case it is undefined
@@ -625,6 +655,7 @@ angular.module('ngResource', ['ng']).
                 }
               }
 
+              value.$cancelRequest = cancelRequest;
               value.$resolved = true;
 
               response.resource = value;
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -63,6 +63,8 @@ function shallowClearAndCopy(src, dst) {
  * @ngdoc service
  * @name $resource
  * @requires $http
+ * @requires ng.$log
+ * @requires $q
  *
  * @description
  * A factory which creates a resource object that lets you interact with
@@ -107,9 +109,9 @@ function shallowClearAndCopy(src, dst) {
  *   URL `/path/greet?salutation=Hello`.
  *
  *   If the parameter value is prefixed with `@` then the value for that parameter will be extracted
- *   from the corresponding property on the `data` object (provided when calling an action method).  For
- *   example, if the `defaultParam` object is `{someParam: '@someProp'}` then the value of `someParam`
- *   will be `data.someProp`.
+ *   from the corresponding property on the `data` object (provided when calling an action method).
+ *   For example, if the `defaultParam` object is `{someParam: '@someProp'}` then the value of
+ *   `someParam` will be `data.someProp`.
  *
  * @param {Object.<Object>=} actions Hash with declaration of custom actions that should extend
  *   the default set of resource actions. The declaration should be created in the format of {@link
@@ -143,15 +145,23 @@ function shallowClearAndCopy(src, dst) {
  *     `{function(data, headersGetter)|Array.<function(data, headersGetter)>}` –
  *     transform function or an array of such functions. The transform function takes the http
  *     response body and headers and returns its transformed (typically deserialized) version.
- *     By default, transformResponse will contain one function that checks if the response looks like
- *     a JSON string and deserializes it using `angular.fromJson`. To prevent this behavior, set
- *     `transformResponse` to an empty array: `transformResponse: []`
+ *     By default, transformResponse will contain one function that checks if the response looks
+ *     like a JSON string and deserializes it using `angular.fromJson`. To prevent this behavior,
+ *     set `transformResponse` to an empty array: `transformResponse: []`
  *   - **`cache`** – `{boolean|Cache}` – If true, a default $http cache will be used to cache the
  *     GET request, otherwise if a cache instance built with
  *     {@link ng.$cacheFactory $cacheFactory}, this cache will be used for
  *     caching.
- *   - **`timeout`** – `{number|Promise}` – timeout in milliseconds, or {@link ng.$q promise} that
- *     should abort the request when resolved.
+ *   - **`timeout`** – `{number}` – timeout in milliseconds.<br />
+ *     **Note:** In contrast to {@link ng.$http#usage $http.config}, {@link ng.$q promises} are
+ *     **not** supported in $resource, because the same value has to be re-used for multiple
+ *     requests. If you are looking for a way to cancel requests, you should use the `cancellable`
+ *     option.
+ *   - **`cancellable`** – `{boolean}` – if set to true, the request made by a "non-instance" call
+ *     will be cancelled (if not already completed) by calling `$cancelRequest()` on the call's
+ *     return value. Calling `$cancelRequest()` for a non-cancellable or an already
+ *     completed/cancelled request will have no effect.<br />
+ *     **Note:** If a timeout is specified in millisecondes, `cancellable` is ignored.
  *   - **`withCredentials`** - `{boolean}` - whether to set the `withCredentials` flag on the
  *     XHR object. See
  *     [requests with credentials](https://developer.mozilla.org/en/http_access_control#section_5)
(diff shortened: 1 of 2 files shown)


gkalpak:
This is an alternative approach to #13058 as discussed in https://github.com/angular/angular.js/pull/13058#issuecomment-152182903.
Still WIP - looking for feedback :smiley:

Pending:

- [ ] Add docs.
- [ ] Decide on names/message text etc.
- [ ] Decide if we need a BC notice ? (Probably...)
- [ ] Decide if we want something more "aggressive" whe using a promise as `timeout`.
      (E.g. $log.warn/error, throw error, etc.) (Probably not...)
- [ ] Decide if we want to support cancellation on instance calls as well (which usually are **not** idempotent) ? (Probably not...)
      If yes, how ?
      Implementation detail: The instance may still have a (noop'd out) `$cancelRequest()` method from when it was created. Could that trick the user into believing they can cancel the request ? Can we do anything about it ?
- [ ] Decide if we want to restrict this to some types of requests only (e.g. `GET`).
      It might be a good idea to only support cancellation on idempotent requests, but then again you don't really know what requests are idempotent and what aren't.
- [ ] ... ?
- [ ] Last but not least: Stop overthinking it and just get this merged :stuck_out_tongue: 

/cc @petebacondarwin

petebacondarwin:
The API looks good to me. I haven't looked at the implementation but I think it is worth pursuing this line.

iyel:
Good job! +1

petebacondarwin:
* Add docs. *Yes*
* Decide on/Improve names/warning message text etc. *Names look good to me*
* Decide if we need a BC notice ? (Probably...) *Yes*
* Decide if we want something more "aggressive" whe using a promise as timeout. (E.g. $log.warn/error, throw error, etc.) (Probably not...) *No*
* Decide if we want to support cancellation on instance calls as well (which usually are not idempotent) ? (Probably not...) If yes, how ? Implementation detail: The instance may still have a (noop'd out) $cancelRequest() method from when it was created. Could that trick the user into believing they can cancel the request ? Can we do anything about it ? *For simplicity we should start only with making static call requests cancellable. Can we delete the `$cancelRequest` method from the instance once the request completes (or is cancelled)?*
* Decide if we want to restrict this to some types of requests only (e.g. GET). It might be a good idea to only support cancellation on idempotent requests, but then again you don't really know what requests are idempotent and what aren't. *No*

petebacondarwin:
@gkalpak if you can get this up and running before next release it can go in 1.5.0 otherwise let's punt it to 1.6

gkalpak:
> Can we delete the $cancelRequest method from the instance once the request completes (or is cancelled)?

@petebacondarwin , the problem is that the user should then always check if the method is available before calling it (and that makes the API weird).

Consider the typical usecase, where the app loads some data based on a user's choice. As soon as the user changes their choice, the old request is irrelevant and a new one should be made. At this point, it is a good practice to cancel the previous request, but we don't really know if that request has already completed or is still pending.

petebacondarwin:
OK, an alternate API is that you pass the object back to a cancel request on the original resource object...

gkalpak:
Would be possible, but requires more "housekeeping", right ?
Could go either way, but slightly prefer this PR's.

What are the benefits of calling a static method (maybe I missed something) ?

petebacondarwin:
The benefit is that you don't need to modify the properties on the instance object.
Something like...

```js
var postsResource = $resource(...);
var posts = postsResource.get(...);
postsResource.cancelRequest(posts);
```

gkalpak:
Yes, but you need to keep an association between requested resources/instances and promises internally, so I feel there is more housekeeping to be done. It would be also more easy to introduce memory leaks (if at all avoidable without WeakMaps).

So, off the top of my head, implementation-wise, modifying the instance properties should be preferrable/simpler. It is also more consistent imo with `$promise` and `$resolved`, which are already attached to the instance (or array of instances).

--
But from a user's perspective, would there be any benefit ?

petebacondarwin:
Do the `$promise` and `$resolved` properties work for instance requests?

petebacondarwin:
If we provide a new promise in the `$promise` property for every request then how about adding a `cancel` method to that object?

gkalpak:
That would work as well, but: 

1. It would make that promise "special" (i.e. not a regular promise).
2. What is the benefit (from a user's perspective or implementation-wise) ?

petebacondarwin:
The `$resource` service makes me sad. OK, you have convinced me. My ideas do not improve the user experience.

gkalpak:
I added the docs (and the BC notice can be added upon squashing).
There is an issue with the `@requires $log` in the docs. It doesn't seem to be able to find `$log` as an available service, so it doesn't probably link to it.

Will try to get a look at it today.

gkalpak:
So, it seems this somehow related to how the [dgeni-packages/links#getDocFromAlias service][1] works
Namely, when there are several `docs` definitions, but none matches the `originatingDoc`s `module`, then an emty list is returned. IMO, it would make more sense to look for a "doc definition" under the `ng` module as a fallback.

In any case, the issue in this PR can be solved by changing `@requires $log` to `@requires ng.$log`.
/cc @petebacondarwin to decide if this needs to be fixed in `dgeni-packages` or works as expected.

[1]: https://github.com/angular/dgeni-packages/blob/94271a1a996b1cfa17f110513a0c784352172f2e/links/services/getDocFromAlias.js

petebacondarwin:
@gkalpak thanks for the investigation but that is indeed how dgeni-packages is supposed to work :-)

gkalpak:
I fixed the `@requires $log` issue and added BC notice in the first commit's message.
This is ready for final review afaict :)

@petebacondarwin, isn't it kind of unexpected that if there is only 1 "doc definition" then it is returned (even if doesn't match the `originatingDoc`'s `module`), but if there are >1 "doc definitions" (none of which matches `originatingDoc`'s `module`) then no doc is returned. I would expect at least on to be returned (e.g. the one on the `ng` module if available or all of them).

src/ngResource/resource.js
```diff
@@ -143,15 +145,23 @@ function shallowClearAndCopy(src, dst) {
  *     `{function(data, headersGetter)|Array.<function(data, headersGetter)>}` –
  *     transform function or an array of such functions. The transform function takes the http
  *     response body and headers and returns its transformed (typically deserialized) version.
- *     By default, transformResponse will contain one function that checks if the response looks like
- *     a JSON string and deserializes it using `angular.fromJson`. To prevent this behavior, set
- *     `transformResponse` to an empty array: `transformResponse: []`
+ *     By default, transformResponse will contain one function that checks if the response looks
+ *     like a JSON string and deserializes it using `angular.fromJson`. To prevent this behavior,
+ *     set `transformResponse` to an empty array: `transformResponse: []`
```
gkalpak: Here and above, I just truncated the lines to 100 chars max.
Sorry, I couldn't help myself :smiley:

petebacondarwin:
The way it works is that we match the text you provide against all the aliases of the docs in the system.
* If exactly one matches then we are happy and we use that one.
* If none match then we have an invalid link.
* If more than one match then we try (as a one off) looking in the current doc's module, since it makes sense that you might be referring to the service in the current module.
  * If there is exactly one match in the current module then we are happy
* Otherwise we have an ambiguous link that needs to be fixed by providing more context (e.g. the module).

gkalpak:
I see. This makes sense as well, I guess. Tweaking the "smartness" vs predictability ratio is a matter of preferrence after all :smiley:
Thx for clarifying !

petebacondarwin:
I guess we could also have a heuristic to favour `ng` if there are multiple but where does it end?

gkalpak:
It doesn't ! Now that I now the reasoning behind it, it makes sense, so it's fine by me :)

src/ngResource/resource.js
```diff
@@ -143,15 +145,23 @@ function shallowClearAndCopy(src, dst) {
...
+ *   - **`timeout`** – `{number}` – timeout in milliseconds.<br />
+ *     **Note:** In contrast to {@link ng.$http#usage $http.config}, {@link ng.$q promises} are
+ *     **not** supported in $resource, because the same value has to be re-used for multiple
+ *     requests. If you are looking for a way to cancel requests, you should use the `cancellable`
+ *     option.
+ *   - **`cancellable`** – `{boolean}` – if set to true, the request made by a "non-instance" call
+ *     will be cancelled (if not already completed) by calling `$cancelRequest()` on the call's
+ *     return value. Calling `$cancelRequest()` for a non-cancellable or an already
+ *     completed/cancelled request will have no effect.<br />
+ *     **Note:** If a timeout is specified in millisecondes, `cancellable` is ignored.
```
Narretz: Why is that so?

gkalpak: By design :smile:

Basically, both need to pass a value as timeout to `$http`'s config (either the number of milliseconds or a promise).
You can't use both, a numeric timeout and a promise, so I chose for `timeout` to take precedence.
(Note that `calcellable` can be configured as a global default or as _per resource_ default (using the `options` argument), while `timeout` can't.)

src/ngResource/resource.js
```diff
@@ -588,14 +657,23 @@ angular.module('ngResource', ['ng']).
               }
             });
 
+            if (!isInstanceCall && action.cancellable) {
+              var deferred = $q.defer();
+              httpConfig.timeout = deferred.promise;
+              value.$cancelRequest = deferred.resolve;
```
petebacondarwin: Do we need to worry about memory leaks if this promise gets attached to the http request?

petebacondarwin: and the returned `value`?

src/ngResource/resource.js
```diff
@@ -588,14 +657,23 @@ angular.module('ngResource', ['ng']).
...
+            }
+
             if (hasBody) httpConfig.data = data;
             route.setUrlParams(httpConfig,
               extend({}, extractParams(data, action.params || {}), params),
               action.url);
 
-            var promise = $http(httpConfig).then(function(response) {
+            var promise = $http(httpConfig).finally(function() {
+              if (value.$cancelRequest) value.$cancelRequest = angular.noop;
```
petebacondarwin: Do we even need to do this? Since resolving the timeout after the request completes would have no impact right?

src/ngResource/resource.js
```diff
@@ -625,6 +703,7 @@ angular.module('ngResource', ['ng']).
                 }
               }
 
+              value.$cancelRequest = cancelRequest;
```
petebacondarwin: we can just set this to `noop` here and then we don't need to hold on to the promise up above.

src/ngResource/resource.js
gkalpak: I haven't given it too much thought tbh, but here is what I think:

- `httpConfig` and `value` do not reference each other, so the question is whether each one is disposed of properly on its own.
- `httpConfig` should get gc'ed (but maybe it doesn't because it is referenced inside the [promise callback][1] - I am not sure). If it doesn't, we should fix it. In any case, this is not introduced by or in any way related to changes in this PR.
- `value.$cancelRequest` does live on, so we need to de-reference `deferred` s soon as that's possible, in order to enable it to be gc'ed. That is what's done in the `finally` block.

I might have missed something though, so take the above with a grain of salt.

[1]: https://github.com/gkalpak/angular.js/blob/feat-%24resource-abortable-requests/src/ngResource/resource.js#L685

src/ngResource/resource.js
gkalpak: See https://github.com/angular/angular.js/pull/13210#discussion_r45747318.
We need this in order to enable `deferred` to be gc'ed and we need `$cancelRequest()` to be a function (that does nothing).

src/ngResource/resource.js
gkalpak: Not holding to the promise is taken care of in the finally block.

--
At this point `cancelRequest` will either be `angular.noop` or undefined.
The idea was to add `noop` back if it was there or set the property to undefined if it wasn't defined.

On a second thought, it would make more sense to avoid adding a property that didn't exist in the first place:

```js
if (cancelRequest) value.$cancelRequest = cancelRequest;
```

src/ngResource/resource.js
gkalpak: Note that we can't take care of it inside the `then` success callback, because there might be an error (in which case we'll never make it to the success callback).

gkalpak:
I added a commit to replace calls to `.hasOwnProperty()` on `action` and `options` to `hasOwnProperty.call()` (as discussed off-GitHub).
I wonder: if we are trying to support custom objects (for `action` and `options`), would it make sense to omit the `hasOwnProperty` check altogether (e.g. use `action.cancellable` or `options.cancellable` etc) ?

gkalpak:
Note to self: This needs to be backported to `v1.4.x`, to solve the issue introduced with 7170f9d.

petebacondarwin:
Hmm - I feel uncomfortable about backporting to 1.4.8. I think it might be safer to just revert 7170f9d instead?

gkalpak:
Works for me as well.

gkalpak:
And maybe also log a warning on `v1.4.x` that promises are not supported as timeout in `$resource`.

gkalpak:
@petebacondarwin, thx for merging !

So, what about `v.1.4.x` ?

- [ ] Should we just revert 7170f9d ?
- [ ] Update the docs to explain that _promise as timeout_ is not supported ?
- [ ] Log a warning when someone does ?

PR #13213, opened 2015-10-31
fixes #12082 (ngTouch regression since 1.3)

### LINKED ISSUE #12082: Click not fired with ng-touch 1.3.16
I have removed all directives except ng-click from a non responsive button.

On touch devices I only see CSS change styles to :active, but the callback for ng-click does not run.

I am using Chrome 43.0 on Android but am seeing this with Safari on iPhone, and IE on windows phone too.

I remove ng-touch and the problem disappears. I downgrade to 1.3.15 and the problem is gone.

I'm fixing the version to 1.3.15 for now in my bower.json.

NorikDavtian:
Send original event to triggerHandler and bypass prevent ghost click if no coordinates.

If there is anchor with ng-click on it the click event was not triggering.

fixes #12082


Narretz:
Thanks for the PR!
It looks like a test failed because of this, though. We also need a test that shows the scenario that is being fixed.

PR #13216, opened 2015-10-31
feat(jqLite): add support for 'ngIndeterminate' as a boolean attribute

calebcauthon:
ng-indeterminate can be used in the same pattern as ng-readOnly, ng-checked, ng-multiple, etc.

Closes #1411

jsFiddle demo of this feature: http://jsfiddle.net/m6ruj542/

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -541,7 +541,7 @@ var JQLitePrototype = JQLite.prototype = {
 // value on get.
 //////////////////////////////////////////
 var BOOLEAN_ATTR = {};
-forEach('multiple,selected,checked,disabled,readOnly,required,open'.split(','), function(value) {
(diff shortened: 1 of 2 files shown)


lgalfaso:
this PR looks very reasonable, the only issue is that this is not the behavior that jQuery has :s 
@mzgol do you know why this is the case?

Narretz:
It should ve because indeterminate is a property, not an attribute, no?https://developer.mozilla.org/de/docs/Web/CSS/%3Aindeterminate

calebcauthon:
@Narretz Yes good point. Indeterminate doesn't behave like those other attributes, I've discovered after some playing around, hence these tests failing. Closing this.

iyel:
So what is the future of indeterminate?

PR #13223, opened 2015-11-02
docs(tutorial/11 - REST and Custom Services): bower install, not npm

acme:
If we change the bower.json then we should run "bower install", not "npm install"


acme:
I signed it!

acme:
Oops, I didn't notice the **Note:** below that. Closing.

PR #13230, opened 2015-11-02
fix(jqLite): deregister special `mouseenter` / `mouseleave` events correctly

### LINKED ISSUE #12795: jqLite handles on/off asymmetrically for mouseenter/mouseleave
After an alternating series of `on`/`off` calls for `mouseenter` or `mouseleave`, the event listener starts being invoked more often than expected.

Say you have a reference to a jqLite-wrapped element, `$element`, and you want to invoke some function when the mouse enters that element:

```javascript
$element.on('mouseenter', someFn);
```

Now, `someFn` is invoked once every time the mouse enters the DOM element wrapped by `$element`, as expected.

Now detach that listener:

```javascript
$element.off('mouseenter', someFn);
```

No, `someFn` is no longer invoked when the mouse enters that DOM element, again as expected.

Now, reattach the listener:

```javascript
$element.on('mouseenter', someFn);
```

Observe: Now `someFn` is invoked twice whenever the mouse enters the specified DOM element, which is _not_ expected.

More generally, after _n_ calls to `on` and _n - 1_ calls to `off`, the event listener starts getting invoked _n_ times per mouse entry, instead of once as expected.

This is true for `mouseenter` and `mouseleave` but does not appear to be the case for other events.

See [this gist](https://gist.github.com/VWoeltjen/c0d9b9bb182e96725891) illustrating the issue.

petebacondarwin:
An alternative solution...

Closes #12795
Closes #12799

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -313,17 +313,23 @@ function jqLiteOff(element, type, fn, unsupported) {
       delete events[type];
     }
   } else {
-    forEach(type.split(' '), function(type) {
-      if (isDefined(fn)) {
-        var listenerFns = events[type];
-        arrayRemove(listenerFns || [], fn);
-        if (listenerFns && listenerFns.length > 0) {
-          return;
-        }
-      }
 
-      removeEventListenerFn(element, type, handle);
-      delete events[type];
+    var removeHandler = function(type) {
+      var listenerFns = events[type];
+      if (isDefined(fn)) {
+        arrayRemove(listenerFns || [], fn);
+      }
+      if (!(isDefined(fn) && listenerFns && listenerFns.length > 0)) {
+        removeEventListenerFn(element, type, handle);
+        delete events[type];
+      }
+    };
+
+    forEach(type.split(' '), function(type) {
+      removeHandler(type);
+      if (MOUSE_EVENT_MAP[type]) {
+        removeHandler(MOUSE_EVENT_MAP[type]);
+      }
     });
   }
 }
@@ -779,6 +785,9 @@ function createEventHandler(element, events) {
       return event.immediatePropagationStopped === true;
     };
 
+    // Some events have special handlers that wrap the real handler
+    var handlerWrapper = eventFns.specialHandlerWrapper || defaultHandlerWrapper;
+
     // Copy event handlers in case event handlers array is modified during execution.
     if ((eventFnsLength > 1)) {
       eventFns = shallowCopy(eventFns);
@@ -786,7 +795,7 @@ function createEventHandler(element, events) {
 
     for (var i = 0; i < eventFnsLength; i++) {
       if (!event.isImmediatePropagationStopped()) {
-        eventFns[i].call(element, event);
+        handlerWrapper(element, event, eventFns[i]);
       }
     }
   };
@@ -795,6 +804,22 @@ function createEventHandler(element, events) {
   //       events on `element`
   eventHandler.elem = element;
   return eventHandler;
+}
+
+function defaultHandlerWrapper(element, event, handler) {
+  handler.call(element, event);
+}
+
+function specialMouseHandlerWrapper(target, event, handler) {
+  // Refer to jQuery's implementation of mouseenter & mouseleave
+  // Read about mouseenter and mouseleave:
+  // http://www.quirksmode.org/js/events_mouse.html#link8
+  var related = event.relatedTarget;
+  // For mousenter/leave call the handler if related is outside the target.
+  // NB: No relatedTarget if the mouse left/entered the browser window
+  if (!related || (related !== target && !jqLiteContains.call(target, related))) {
+    handler.call(target, event);
+  }
 }
 
 //////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


src/jqLite.js
```diff
@@ -313,17 +313,23 @@ function jqLiteOff(element, type, fn, unsupported) {
...
+    var removeHandler = function(type) {
+      var listenerFns = events[type];
       if (isDefined(fn)) {
-        var listenerFns = events[type];
         arrayRemove(listenerFns || [], fn);
-        if (listenerFns && listenerFns.length > 0) {
-          return;
-        }
       }
+      if (!(isDefined(fn) && listenerFns && listenerFns.length > 0)) {
```
gkalpak: Why only when `!isDefined(fn)` ? The previosu implementation would remove the native listener when the last handle listenerFn was unregistered. Why prevent that ?

gkalpak:
From a quick look, it LGTM.
Will take a more careful look and try some more tests tomorrow.

:+1:

src/jqLite.js
petebacondarwin: I think that this construct is correct, since we call `removeEventListenerFn` in the following cases:

* `fn` is not defined - i.e. we want to remove all handlers for this event
* `fn` is defined AND there are no more items in the `listenerFns` array

gkalpak: Right !

gkalpak:
Thinking about this, I realized there is the following issue (not introduced by this PR, but might be worth solving as part of it):

If the user registers the same callback for `mouseenter` and `mouseover`, then the callback is executed twice upon `mouseover`. (The same is true for `mouseleave`/`mouseout`.)

--
Here are a couple of tests I ran, if you want to add them to the suite:

```js
    // SUCCESS
    it('should deregister a mouseenter/leave listener that is also registered for mouseover/out',
      function() {
        var aElem = jqLite(a);
        var onMouseenter = jasmine.createSpy('onMouseenter');
        var onMouseleave = jasmine.createSpy('onMouseleave');

        aElem.on('mouseenter', onMouseenter);
        aElem.on('mouseleave', onMouseleave);
        aElem.off('mouseenter', onMouseenter);
        aElem.off('mouseleave', onMouseleave);
        aElem.on('mouseover', onMouseenter);
        aElem.on('mouseout', onMouseleave);

        browserTrigger(a, 'mouseover', {relatedTarget: b});
        expect(onMouseenter).toHaveBeenCalledOnce();

        browserTrigger(a, 'mouseout', {relatedTarget: b});
        expect(onMouseleave).toHaveBeenCalledOnce();
      }
    );


    // SUCCESS
    it('should deregister specific listener within the listener and call subsequent listeners', function() {
      var aElem = jqLite(a),
          clickSpy = jasmine.createSpy('click'),
          clickOnceSpy = jasmine.createSpy('clickOnce').andCallFake(function() {
            aElem.off('click', clickOnceSpy);
          });

      aElem.on('click', clickOnceSpy);
      aElem.on('click', clickSpy);

      browserTrigger(a, 'click');
      expect(clickOnceSpy).toHaveBeenCalledOnce();
      expect(clickSpy).toHaveBeenCalledOnce();

      browserTrigger(a, 'click');
      expect(clickOnceSpy).toHaveBeenCalledOnce();
      expect(clickSpy.callCount).toBe(2);
    });


    // FAILURE
    it('should call a mouseenter/leave listener that is also registered for mouseover/out once',
      function() {
        var aElem = jqLite(a);
        var onMouseenter = jasmine.createSpy('onMouseenter');
        var onMouseleave = jasmine.createSpy('onMouseleave');

        aElem.on('mouseenter', onMouseenter);
        aElem.on('mouseleave', onMouseleave);
        aElem.on('mouseover', onMouseenter);
        aElem.on('mouseout', onMouseleave);

        browserTrigger(a, 'mouseover', {relatedTarget: b});
        expect(onMouseenter).toHaveBeenCalledOnce();

        browserTrigger(a, 'mouseout', {relatedTarget: b});
        expect(onMouseleave).toHaveBeenCalledOnce();
      }
    );
```

gkalpak:
Other than the above comment, it LGTM.

petebacondarwin:
Isn't that effectively by design? The point is that we can't rely on the mouseenter event so we react to the mouseover event instead. If you add handlers to both the mouseenter and mouseover events then it is reasonable to expect to calls.

gkalpak:
> If you add handlers to both the mouseenter and mouseover events then it is reasonable to expect to calls.

Fair enough :)

--
New concern:
The previous implementation didn't actually add an event-listener for `mouseenter`/`mouseleave` events.

This new implementation does add event-listeners for both `mouseenter`/`mouseleave` and `mouseover`/`mouseout`. This will result in calling the same listener twice (although the user registered it once), when both `mouseenter` and `mouseover` are triggered.

gkalpak:
Basically, that's the test I have in mind:

```js
    it('should call a `mouseenter/leave` listener once when `mouseenter/leave` and `mouseover/out` '
       + 'are triggered simultaneously', function() {
      var aElem = jqLite(a);
      var onMouseenter = jasmine.createSpy('mouseenter');
      var onMouseleave = jasmine.createSpy('mouseleave');

      aElem.on('mouseenter', onMouseenter);
      aElem.on('mouseleave', onMouseleave);

      browserTrigger(a, 'mouseenter', {relatedTarget: b});
      browserTrigger(a, 'mouseover', {relatedTarget: b});
      expect(onMouseenter).toHaveBeenCalledOnce();

      browserTrigger(a, 'mouseleave', {relatedTarget: b});
      browserTrigger(a, 'mouseout', {relatedTarget: b});
      expect(onMouseleave).toHaveBeenCalledOnce();
    });
```

It does indeed pass with the previous implementation, but fails with the new one.

petebacondarwin:
@gkalpak I have updated the PR with this test and the fix. It passes both on jqlite and jquery so I think we are good to go.

src/jqLite.js
```diff
@@ -825,35 +850,27 @@ forEach({
...
-        eventFns = events[type];
       }
+
       eventFns.push(fn);
+    };
+
+    while (i--) {
+      type = types[i];
+      if (MOUSE_EVENT_MAP[type]) {
+        addHandler(MOUSE_EVENT_MAP[type], specialMouseHandlerWrapper);
```
gkalpak: I think we need an additional `(events[type] || (events[type] = [])).push(fn);` (or equivalent) to also support "manual" triggering.

gkalpak:
But now it fails the following test:

```js
    it('should call a `mouseenter/leave` listener when manually triggering the event', function() {
      var aElem = jqLite(a);
      var onMouseenter = jasmine.createSpy('mouseenter');
      var onMouseleave = jasmine.createSpy('mouseleave');

      aElem.on('mouseenter', onMouseenter);
      aElem.on('mouseleave', onMouseleave);

      a.triggerHandler('mouseenter');
      expect(onMouseenter).toHaveBeenCalledOnce();

      a.triggerHandler('mouseleave');
      expect(onMouseleave).toHaveBeenCalledOnce();
    });
```

Basically, it doesn't call the handler, when "manually" triggering `mouseenter`/`leave`.
At least I think so (I haven't tried it).

[1]: https://github.com/angular/angular.js/pull/13230/files#r44645939

PR #13233, opened 2015-11-03
fix(ngLocale): set FIRSTDAYOFWEEK to monday for german locales

ctaepper:
first day of week in germans locales should be monday


ctaepper:
I signed it!

ctaepper:
I signed it!

PR #13235, opened 2015-11-03
fix(ngLocale): set FIRSTDAYOFWEEK to monday for german locales

ctaepper:
first day of week should be monday for german locales


lgalfaso:
The property `FIRST_DAY_OF_THE_WEEK` follows ISO 8601 that is that Monday is the first day of the week. I know that there is a discrepancy between this and JavaScript, unfortunately this cannot be changed now :(

PR #13236, opened 2015-11-03
perf(merge): remove unnecessary wrapping of jqLite element

gkalpak:
Fixes https://github.com/angular/angular.js/commit/75292a6cb5e17d618902f7996e80eb3118eff7b0#commitcomment-14137538

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -363,7 +363,7 @@ function baseExtend(dst, objs, deep) {
         } else if (src.nodeName) {
           dst[key] = src.cloneNode(true);
         } else if (isElement(src)) {
-          dst[key] = jqLite(src).clone();
+          dst[key] = src.clone();
         } else {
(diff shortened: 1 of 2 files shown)


protractor-conf.js
```diff
@@ -9,7 +9,7 @@ config.specs = [
 ];
 
 config.capabilities = {
-  browserName: 'chrome'
+  browserName: 'firefox'
```
petebacondarwin: Why this change?

petebacondarwin:
One comment above. Otherwise LGTM please merge

protractor-conf.js
gkalpak: Ooops. I committed it by accident (I am using it locally, because of the problem I have with protractor+latest Chrome).

gkalpak:
Backported to `v1.4.x` as ce6a96b.

petebacondarwin:
Smashing, thanks

Issue #8659, opened 2014-08-18
Invalid argument in IE when using $compile


Matsemann:
A fiddle with a trivial example: http://jsfiddle.net/1t4crp6e/

````
<div ng-app="myApp">
    <div ng-controller="myctrl">
        <div test>the value is: {{testmodel}}</div>
        
        <br>
        <input type="text" ng-model="testmodel">
    </div>
</div>
....
link: function(scope, elm, attrs) {
            var standardText = elm.text();
            
            elm.html('something');
            
            $timeout(function() {
                console.log(standardText);
                elm.html('').append($compile('<span>' + standardText + '</span>')(scope.$parent));
            }, 2000);
        }
```


Wait 2 secs, see the content of directive change to a binding. In Firefox, everything works. In IE it *works*, but it throws a lot of errors in the console.
(Don't worry about what I'm doing here, I just made a small example showing the issue)

> Error: Invalid argument.
  at Anonymous function (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:62:80)
   at k.prototype.$digest (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:109:350)
   at k.prototype.$apply (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:112:343)
   at Anonymous function (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:18:158)
   at d (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:34:474)
   at c (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:18:66)
   at fc (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:18:270)
   at Xc (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:17:362)
   at Anonymous function (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.22/angular.min.js:214:78)
   at o (https://ajax.googleapis.com/ajax/libs/jquery/1.7.2/jquery.min.js:2:14725)

**However, most of my tries of actually writing this worked in all browsers *except* IE. For instance, if I remove the span in the compile, and instead of elm.text() use elm.contents().**
[...]

PR #13243, opened 2015-11-03
Fix ie 8659 (interpolation in extracted textNodes)

Narretz:
This is to test a fix for issue #8659. Locally, it caused another test to fail.

~~There are a few more commits in here that should be, I'll check this out later~~


lgalfaso:
I do not have access to IE, so there are many things that I cannot try. Anyhow, if https://github.com/angular/angular.js/issues/8659#issuecomment-67441529 is true, then IE is broken to an extend that there is simply no way to make IE happy.


BTW, the tests pass if the check is changed to `if (node[0].nodeValue == null) {`, anyhow, I doubt that this would be enough to fix other variations of this same problem under IE.

Narretz:
@lgalfaso `node[0].nodeValue == null` didn't fix the problem locally. Another option is to wrap it in a try-catch, but I'm not sure it has side effects.

Narretz:
It's unlikely that this makes it in, as the original problem is probably not a bug. See the issue.

PR #13261, opened 2015-11-05
refactor($interval): do not use `notify` to trigger the callback

### LINKED ISSUE #12533: Drop non-compliant Promises/A+  progress/notify on $q
Makes it really hard to replace $q when using a compliant library, since angular uses progress/notify handlers on core itself.

lgalfaso:
Do not use `$q.notify` to trigger the callback. This allows `$q` to be replaced
with another Promise/A+ compilant library

Note: this PR is not enough to close #12533, but should solve the main issue in trying to use Bluebird as a replacement of `$q`

### DIFF
--- a/src/ng/interval.js
+++ b/src/ng/interval.js
@@ -2,8 +2,8 @@
 
 
 function $IntervalProvider() {
-  this.$get = ['$rootScope', '$window', '$q', '$$q',
-       function($rootScope,   $window,   $q,   $$q) {
+  this.$get = ['$rootScope', '$window', '$q', '$$q', '$browser',
+       function($rootScope,   $window,   $q,   $$q,   $browser) {
     var intervals = {};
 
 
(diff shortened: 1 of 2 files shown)


Narretz:
If we are making this change to support a specific use case, shouldn't we add a test for that behavior? Otherwise someone else might refactor this back to what it was, and it would break again,

lgalfaso:
@Narretz done!

src/ng/interval.js
```diff
@@ -144,11 +144,14 @@ function $IntervalProvider() {
 
       count = isDefined(count) ? count : 0;
 
-      promise.then(null, null, (!hasParams) ? fn : function() {
-        fn.apply(null, args);
-      });
-
       promise.$$intervalId = setInterval(function tick() {
+        if (intervals[promise.$$intervalId]) {
```
gkalpak: Why is this check needed ?

src/ng/interval.js
```diff
@@ -164,6 +167,14 @@ function $IntervalProvider() {
       intervals[promise.$$intervalId] = deferred;
 
       return promise;
+
+       function callback() {
```
gkalpak: Too indented...

src/ng/interval.js
lgalfaso: you are right

src/ng/interval.js
lgalfaso: thanks!

PR #13265, opened 2015-11-06
refactor($compile): simplify multi element directive check

jbedard:
I think this is a bit simpler. It also avoids `directiveIsMultiElement()` (which does `$injector.get`) for every attribute of every element when compiling (which might have minor perf improvements when compiling).


gkalpak:
LGTM :+1:

Narretz:
LGTM
I assume the perf improvements are very minor, because directiveIsMultiElement() only calls the injector if an attribute indicates a directive, right?. Maybe for elements that have many directives.

Narretz:
Landed as b837fc3116e697aaf18977867a5defd9541f7f8c

PR #13267, opened 2015-11-06
refactor($compile): move check for interpolation of on-event attributes to compile time

jbedard:
BREAKING CHANGE:
Previously the nodomevents would be thrown at link time for any interpolation on\* event attributes. Now this will be throw at compile time similar to the selmulti error.

The breaking change should be rare, but is probably worth mentioning.  This makes the two interpolation errors consistent and avoids checking the same thing per link (which previously would log the same error per link).

The test changes are not necessary but do make them stricter and more like the selmulti error tests.


lgalfaso:
LGTM

Issue #11261, opened 2015-03-06
ngTouch check for blur() function before using it.


leandroz:
I am using ngTouch and i have an svg element on my mobile app. The problem is that svg elements don't have blur() function.
Solution:
````
if (typeof tapElement.blur === 'function') {
	tapElement.blur();
}
````

PR #13272, opened 2015-11-06
Fix TypeError: tapElement.blur is not a function

ghostd:
Hi,

I ran into the same issue than #11261 with the following HTML:

``` html
<span ng-click="scrollService.goToSection('my-section')">
  <svg><use xlink:href="#my-icon"></use></svg>
  Title
</span>
```

On Firefox mobile, when i tap onto the SVG icon, i get the following error which breaks the event handler:

```
TypeError: tapElement.blur is not a function
```

The tap element is a SVG node that does not have a 'blur' function.

This commit fixes the issue for me.

Regards


ghostd:
I signed it!

Issue #12095, opened 2015-06-12
Allow transformRequest to modify headers


remcohaszing:
In AngularJS 1.4.0-beta5 the possibility to modify request headers from `transformRequest` functions has been removed. This is stated in the [changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#breaking-changes-3). I found this a very useful feature to modify the `Content-Type` of the request being made.

For example:
```js
$http.post('', {
  foo: 'bar'
}, {
  transformRequest: function(data, headers) {
    headers()['Content-Type'] = 'application/x-www-form-urlencoded';
    var parsed = [];
    angular.forEach(data, function(value, key) {
      parsed.push(encodeURIComponent(key) + '=' + encodeURIComponent(value));
    });
    return parsed.join('&');
  }
});
```
I often use a similar approach to make `form-data` requests, which requires the `Content-Type` to be set to `undefined`.

In practice I have created an `xWwwFormUrlencoded` and `formData` factory, because I use them more often.

So the code would look like this:
```js
$http.post('', {
  foo: 'bar'
}, {
  transformRequest: xWwwFormUrlencoded
});
```

It feels like a huge drawback having to to explicitly specify the headers whenever making such a transformed request.

The following [plunkr](http://plnkr.co/edit/HnafBwgTFwi9Ete7SwjA?p=preview) demonstrates this issue. The test breaks when updating AngularJS to `1.4.0`.

Narretz:
Hmm. I can see how this in inconvenient. There are basically two ways
- setting headers explicitly
- using an interceptor (but you have to filter the requests to which they should apply)

However, if you have a factory for the transformRequest, couldn't you simply do:
```
$http.post('', {
  foo: 'bar'
}, {
  headers: xWwwFormUrlencoded()
});
```

pkozlowski-opensource:
Yeh, I can see how this is inconvenient and while interceptors work they add overhead of the async processing. having said this I really dislike APIs where input arguments are getting modified....

Maybe we should make it possible to return both transformed data and headers from the transform request function?

Narretz:
The headers property could also accept a function that gets passed the current config + default headers and is expected to return a headers object.

remcohaszing:
Actually using an interceptor sounds like a perfect solution in my case. I could just filter out requests which specify an `x-www-form-urlencoded` content type and modify the body accordingly.

[...]

Issue #13123, opened 2015-10-19
Cannot run `npm install` on a newly-cloned repository


schmod:
Running `npm install` on a freshly-cloned repository fails, and leaves no `node_modules` folder intact:

```
> angularjs@ preinstall /home/schmod/source/angular.js
> node scripts/npm/check-node-modules.js --purge

:-( npm dependencies are stale or in an unknown state!
    Purging 'node_modules'...

> angularjs@ postinstall /home/schmod/source/angular.js
> node scripts/npm/copy-npm-shrinkwrap.js

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
Failed to copy `npm-shrinkwrap.json` to `node_modules/npm-shrinkwrap.cached.json`:
{ [Error: ENOENT: no such file or directory, open 'node_modules/npm-shrinkwrap.cached.json']
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'node_modules/npm-shrinkwrap.cached.json' }
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

NPM v3.3.8; Node 4.2.1; Ubuntu 15.04

schmod:
It looks like the changes introduced in #12792 / d3de0066b0db40b09243c3845fe810401cf3c7ef will delete the user's `node_modules` folder when the `preinstall` script runs if a NPM Shrinkwrap cache file is not present.

I'm assuming that NPM v3.x doesn't add this file before the `preinstall` scripts are run, which is causing the preinstall script to wipe the `node_modules` folder that `npm install` just created.

I'm not sure what the original intent was here, so I can't recommend a fix.

gkalpak:
I don't see how the changes on #12792 could cause this issue.
The `node_modules/` directory is purged before `npm install`, then `npm install` runs and creates the `node_modules/` directory (installing all necessary modules) and finally the (already existing) `npm-shrinkwrap.json` file is copied over to `node_modules/npm-shrinkwrap.cached.json` (which is created then; it doesn't exist already).

[...]

Issue #11409, opened 2015-03-24
Resource interceptor is not actually an interceptor, just a function that returns what becomes result of resource promise.


KamilSzot:
Here you can see that behaviour of interceptor added to $httpProvider.interceptors is completely different from interceptor declared in resource methods: http://jsfiddle.net/Lzgts/269/

I think we need to update the documentation to reflect this changing the behaviour now might break apps of people who adapted to this behaviour.

That's how I found out about this bug. Behaviour of my app was different when I stopped needing the resource interceptor. After removing it app stopped working properly, but leaving it doing nothing as shown at https://docs.angularjs.org/api/ng/service/$http in section on Interceptors kept the app working properly.

petebacondarwin:
That does seem very confusing. I think we either need to make some changes to `$resource` and how it uses interceptors/promises or we should indeed document this very clearly. Right now I am leaning toward the first option.

gkalpak:
It doesn't really have to do with the interceptors, but with the internal workings of `$resource`.

For "its own reasons", `$resource` uses a default response interceptor that returns `response.resource`. If you overwrite that (using a custom action interceptor), you have to take care of this yourself (or live with the consequences :smiley:).

This has been discussed in more detail in #11201 (expecially https://github.com/angular/angular.js/issues/11201#issuecomment-76367339 and https://github.com/angular/angular.js/issues/11201#issuecomment-76714521).

petebacondarwin:
@gkalpak  thanks for linking to the previous discussion on this. The fact that this has come up again and,  IMO is not very intuitive, I wonder if we ought to make a change to `$resource`.  The comment that chimes with me is (https://github.com/angular/angular.js/issues/11201#issuecomment-76367339):

> This default interceptor is used only if there is no response interceptor specified by the user.

1) Do we really need this default interceptor to return the resource instance rather than the response?
2) Could we always run this default interceptor, either
  a) before any custom interceptor - meaning that the custom interceptor would receive the resource instance rather than the response
  b) after any custom interceptor - meaning that an interceptor would have to make changes to `response.resource` to have an impact outside `$resource`.

2 seems to be fairly arbitrary and in either case would break some uses irreparably.

1 seems to be a breaking change that doesn't really add any value except remove unexpected behaviour; but I would definitely consider it...

gkalpak:
@petebacondarwin, I too agree that we could do better inside `$resource` to avoid surprizes.

[...]

Issue #9334, opened 2014-09-29
ngResource interceptors called inconsistently


jrm-airwatch:
Action-specific interceptors defined in a $resource call are called inconsistently. A response interceptor is called before the success callback, but a responseError interceptor is called after the error callback. Further, the success callback is called even if the response interceptor rejects, and the error callback is called even if the responseError interceptor resolves. And response data is populated before any interceptors run.

This behavior runs contrary to the expectation that an interceptor should transparently intercept results before passing them to the consumer. A more sensible flow would be

http success -> interceptor.response (resolves) -> populate results -> success callback
http success -> interceptor.response (rejects) -> error callback
http error -> interceptor.responseError (resolves) -> populate results -> success callback
http error -> interceptor.responseError (rejects) -> error callback

As it stands, even if a responseError interceptor successfully retries, the error callback is still called, and no response data is populated!

[plunker](http://plnkr.co/RzFFBYnuiM437lzbAkTS?p=preview)

nikrolls:
This is quite a frustrating issue; if it help, here is a plunker: http://plnkr.co/edit/ynHfrkzsqKZbJE2cBZSM?p=preview

Note that:
 - interceptors (labelled as 'before') are called after final callbacks (labelled as 'after')
 - modified payloads (both content and success/rejection) are ignored

gkalpak:
Indeed the concept of `interceptors` in the context of `$resource` is very vague. In the documentation, it is implied that interceptors are essentilly `$http` interceptors, but they are used quite differently. Furthermore, their behavior is inconsistent and not well defined.

Similar issues have been discussed in https://github.com/angular/angular.js/issues/11409, as well.
IMO, `$resource` interceptors should be behave like this:

* They should be applied to the promise returned by the call to `$http` (before anything `$resource`-specific).
* The appropriate `$resource` interceptor should be applied (if specified by the user). Interceptors should be passed the `response` - nothing `$resource`-specific yet.
* If the interceptor rejects, the error callback should be called (and the data should not be updated).
* If the interceptor resolves, `$resource` should update the returned instance based on the content of `response.data` and call the success callback.

Thoughts anyone ?

krotscheck:
I overall agree with the original issue description and @gkalpak's assessment. Also, I'm making the assumption that transformRequest and transformResponse both execute before/after the interceptors are considered, so I'm deliberately not including them.

[...]

PR #13273, opened 2015-11-07
Add request and requestError interception in ngResource

### LINKED ISSUE #5146: ngResource does not apply request interceptor
ngResource removed transformRequest support but does not apply interceptor.request provided in actions. Shouldn't it?

krotscheck:
This patch adds request and requestError interception for ngResource actions,
as per the documentation found for $http interceptors. It is important to
note that returning an error at this stage of the request - before the
call to $http - will completely bypass any global interceptors and/or recovery
handlers, as those are added to a separate context. This is intentional; intercepting
a request before it is passed to the HTTP stack indicates that the resource itself
has made a decision, and that it accepts the responsibility for recovery.

Closes #5146

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -167,9 +167,9 @@ function shallowClearAndCopy(src, dst) {
  *     for more information.
  *   - **`responseType`** - `{string}` - see
  *     [requestType](https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest#responseType).
- *   - **`interceptor`** - `{Object=}` - The interceptor object has two optional methods -
- *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
- *     with `http response` object. See {@link ng.$http $http interceptors}.
+ *   - **`interceptor`** - `{Object=}` - The interceptor object has four optional methods -
+ *     `request`, `requestError`, `response`, and `responseError`. See
+ *     {@link ng.$http $http interceptors} for details.
  *
  * @param {Object} options Hash with custom settings that should extend the
  *   default `$resourceProvider` behavior.  The supported options are:
@@ -633,6 +633,10 @@ angular.module('ngResource', ['ng']).
             var isInstanceCall = this instanceof Resource;
             var value = isInstanceCall ? data : (action.isArray ? [] : new Resource(data));
             var httpConfig = {};
+            var requestInterceptor = action.interceptor && action.interceptor.request ||
+              undefined;
+            var requestErrorInterceptor = action.interceptor && action.interceptor.requestError ||
+              undefined;
             var responseInterceptor = action.interceptor && action.interceptor.response ||
               defaultResponseInterceptor;
             var responseErrorInterceptor = action.interceptor && action.interceptor.responseError ||
@@ -667,7 +671,14 @@ angular.module('ngResource', ['ng']).
               extend({}, extractParams(data, action.params || {}), params),
               action.url);
 
-            var promise = $http(httpConfig).then(function(response) {
+            // Start the promise chain
+            var promise = $q.
+              resolve(httpConfig).
+              then(requestInterceptor).
+              catch(requestErrorInterceptor).
+              then($http);
+
+            promise = promise.then(function(response) {
               var data = response.data;
 
               if (data) {
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -3,7 +3,7 @@
 describe("resource", function() {
 
 describe("basic usage", function() {
-  var $resource, CreditCard, callback, $httpBackend, resourceProvider;
+  var $resource, CreditCard, callback, $httpBackend, resourceProvider, $q;
 
   beforeEach(module('ngResource'));
 
@@ -14,6 +14,7 @@ describe("basic usage", function() {
   beforeEach(inject(function($injector) {
     $httpBackend = $injector.get('$httpBackend');
     $resource = $injector.get('$resource');
+    $q = $injector.get('$q');
     CreditCard = $resource('/CreditCard/:id:verb', {id:'@id.key'}, {
       charge:{
         method:'post',
@@ -978,6 +979,188 @@ describe("basic usage", function() {
       });
     });
 
+
+    describe('requestInterceptor', function() {
+      var rejectReason = {'lol':'cat'};
+      var successSpy, failureSpy;
+
+      beforeEach(function() {
+        successSpy = jasmine.createSpy('successSpy');
+        failureSpy = jasmine.createSpy('failureSpy');
+      });
+
+      it('should allow per action request interceptor that gets full configuration', function() {
+        var CreditCard = $resource('/CreditCard', {}, {
+          query: {
+            method: 'get',
+            isArray: true,
+            interceptor: {
+              request: function(httpConfig) {
+                callback(httpConfig);
+                return httpConfig;
+              }
+            }
+          }
+        });
+
+        $httpBackend.expect('GET', '/CreditCard').respond([{id: 1}]);
+
+        var resource = CreditCard.query();
+        resource.$promise.then(successSpy, failureSpy);
+
+        $httpBackend.flush();
+        expect(callback).toHaveBeenCalledOnce();
+        expect(successSpy).toHaveBeenCalledOnce();
+        expect(failureSpy).not.toHaveBeenCalled();
+
+        expect(callback).toHaveBeenCalledWith({
+          'method': 'get',
+          'url': '/CreditCard'
+        });
+      });
+
+      it('should call $http with the value returned from requestInterceptor', function() {
+        var CreditCard = $resource('/CreditCard', {}, {
+          query: {
+            method: 'get',
+            isArray: true,
+            interceptor: {
+              request: function(httpConfig) {
+                httpConfig.url = '/DebitCard';
+                return httpConfig;
+              }
+            }
+          }
+        });
+
+        $httpBackend.expect('GET', '/DebitCard').respond([{id: 1}]);
+
+        var resource = CreditCard.query();
+        resource.$promise.then(successSpy, failureSpy);
+
+        $httpBackend.flush();
+        expect(successSpy).toHaveBeenCalledOnce();
+        expect(failureSpy).not.toHaveBeenCalled();
+      });
+
+      it('should abort the operation if the requestInterceptor rejects the operation', function() {
+        var CreditCard = $resource('/CreditCard', {}, {
+          query: {
+            method: 'get',
+            isArray: true,
+            interceptor: {
+              request: function() {
+                return $q.reject(rejectReason);
+              }
+            }
+          }
+        });
+
+        var resource = CreditCard.query();
+        resource.$promise.then(successSpy, failureSpy);
+
+        // Make sure all promises resolve.
+        $rootScope.$apply();
+
+        // Ensure the resource promise was rejected
+        expect(resource.$resolved).toBeTruthy();
+        expect(successSpy).not.toHaveBeenCalled();
+        expect(failureSpy).toHaveBeenCalledOnce();
+        expect(failureSpy).toHaveBeenCalledWith(rejectReason);
+
+        // Ensure that no requests were made.
+        $httpBackend.verifyNoOutstandingRequest();
+      });
+
+      it('should call requestErrorInterceptor if requestInterceptor rejects the operation', function() {
+        var CreditCard = $resource('/CreditCard', {}, {
+          query: {
+            method: 'get',
+            isArray: true,
+            interceptor: {
+              request: function() {
+                return $q.reject(rejectReason);
+              },
+              requestError: function(rejection) {
+                callback(rejection);
+                return $q.reject(rejection);
+              }
+            }
+          }
+        });
+
+        var resource = CreditCard.query();
+        resource.$promise.then(successSpy, failureSpy);
+        $rootScope.$digest();
+
+        expect(callback).toHaveBeenCalledOnce();
+        expect(callback).toHaveBeenCalledWith(rejectReason);
+        expect(successSpy).not.toHaveBeenCalled();
+        expect(failureSpy).toHaveBeenCalledOnce();
+        expect(failureSpy).toHaveBeenCalledWith(rejectReason);
+
+        // Ensure that no requests were made.
+        $httpBackend.verifyNoOutstandingRequest();
+      });
+
+      it('should abort the operation if a requestErrorInterceptor rejects the operation', function() {
+          var CreditCard = $resource('/CreditCard', {}, {
+            query: {
+              method: 'get',
+              isArray: true,
+              interceptor: {
+                request: function() {
+                  return $q.reject(rejectReason);
+                },
+                requestError: function(rejection) {
+                  return $q.reject(rejection);
+                }
+              }
+            }
+          });
+
+          var resource = CreditCard.query();
+          resource.$promise.then(successSpy, failureSpy);
+          $rootScope.$apply();
+
+          expect(resource.$resolved).toBeTruthy();
+          expect(successSpy).not.toHaveBeenCalled();
+          expect(failureSpy).toHaveBeenCalledOnce();
+          expect(failureSpy).toHaveBeenCalledWith(rejectReason);
+
+          // Ensure that no requests were made.
+          $httpBackend.verifyNoOutstandingRequest();
+        });
+
+      it('should continue the operation if a requestErrorInterceptor rescues it', function() {
+        var CreditCard = $resource('/CreditCard', {}, {
+          query: {
+            method: 'get',
+            isArray: true,
+            interceptor: {
+              request: function(httpConfig) {
+                return $q.reject(httpConfig);
+              },
+              requestError: function(httpConfig) {
+                return $q.resolve(httpConfig);
+              }
+            }
+          }
+        });
+
+        $httpBackend.expect('GET', '/CreditCard').respond([{id: 1}]);
+
+        var resource = CreditCard.query();
+        resource.$promise.then(successSpy, failureSpy);
+        $httpBackend.flush();
+
+        expect(resource.$resolved).toBeTruthy();
+        expect(successSpy).toHaveBeenCalledOnce();
+        expect(failureSpy).not.toHaveBeenCalled();
+        $httpBackend.verifyNoOutstandingRequest();
+      });
+    });
+
     it('should allow per action response interceptor that gets full response', function() {
       CreditCard = $resource('/CreditCard', {}, {
         query: {
@@ -1318,7 +1501,7 @@ describe("basic usage", function() {
 });
 
 describe('errors', function() {
-  var $httpBackend, $resource, $q;
+  var $httpBackend, $resource, $q, $rootScope;
 
   beforeEach(module(function($exceptionHandlerProvider) {
     $exceptionHandlerProvider.mode('log');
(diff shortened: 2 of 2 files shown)


krotscheck:
Note, this contribution is made under the Hewlett Packard Enterprise Corporate CLA. I'm trying to figure out how to make the bot recognize my second github email address, but the recent separation probably has us in legal limbo for the time being.

krotscheck:
Woot, that worked!

src/ngResource/resource.js
```diff
@@ -593,7 +601,13 @@ angular.module('ngResource', ['ng']).
               extend({}, extractParams(data, action.params || {}), params),
               action.url);
 
-            var promise = $http(httpConfig).then(function(response) {
+            var promise = $q.resolve(httpConfig);
```
gkalpak: Why not directly: `$q.when(requestInterceptor(httpConfig), null, requestErrorInterceptor)`

krotscheck: Well, that's how my brain addressed the problem. First, peel everything into a promise chain. Second, add the desired step into that chain. While I can collapse it easily enough, I personally find it more legible, as each statement is discrete and easily understood.

I don't mind updating the commit though, if there's consensus that your proposed syntax is more desirable.

krotscheck:
As an additional note: The reason I'd like to land this in angular, is because I'd like to build resources who get their root API url from a promise.

1. Resource action is requested (GET /foo/:bar)
2. request interceptor decorates the relative URL with the discovered API endpoint (GET https://example.com/api/v1/foo/:bar)
3. request is made.

Since I live in a multi-api environment, resolving this in the resource made the most sense.

krotscheck:
@petebacondarwin Heya- with 1.5 out now, what are the next steps for this PR? Should I rebase, update, add docs... any guidance?

gkalpak:
@krotscheck, now that 1.5.0 is out (and since this is on the 1.6.x milestone), I will take a look soon and get back to you. You don't need to rebase for now. I'll let you know if you need to update anything.

Thx for working on this !

krotscheck:
@gkalpak Awesome! Looking forward to your comments :)

src/ngResource/resource.js
```diff
@@ -664,7 +672,9 @@ angular.module('ngResource', ['ng']).
               extend({}, extractParams(data, action.params || {}), params),
               action.url);
 
-            var promise = $http(httpConfig).then(function(response) {
+            var promise = $q.when(requestInterceptor(httpConfig), null, requestErrorInterceptor);
```
gkalpak: This will not catch errors thrown in the `requestInterceptor`. The `requestErrorInterceptor` will only be called if `requestInterceptor` returns a rejecting promise.

I think the following is better:

```js
promise.when(httpConfig).then(requestInterceptor).catch(requestErrorInterceptor)
```

src/ngResource/resource.js
```diff
@@ -559,6 +559,10 @@ angular.module('ngResource', ['ng']).
           return ids;
         }
 
+        function defaultRequestInterceptor(httpConfig) {
```
gkalpak: With the change proposed in https://github.com/angular/angular.js/pull/13273/files#r53487712, this is obsolete.

gkalpak:
@krotscheck, I left a couple of comments.

I think it's a good change overall, but I would like to have more tests. E.g.:

* Test that `$http` is called with the value returned from `requestInterceptor`.
* Test that throwing inside the `requestInterceptor` calls the `requestErrorInterceptor`.
* Test that a rejecting `requestInterceptor` aborts the operation.
* Test that a rejecting `requestErrorInterceptor` aborts the operation.
* Test that a `requestErrorInterceptor` can recover from rejection (and the operation continues as usual).
* _...something else that I have missed..._

Feel free to ping me when/if you make any changes, so I can take another look.

--
BTW, there is some ongoing discussion about response interceptors (or interceptors in general) in the context of `$resource` (e.g. in #9334 and elsewhere). Their behavior is currently kind of inconsistent and confusing - I would like to get it right for `v1.6.x`. @krotscheck, if you have any insights, please do chime in :smiley:

krotscheck:
@gkalpak Tests added, feedback incorporated. I also concatenated the promise chain, which had the unexpected impact of creating a gross diff (indentation had to be updated). I can go back and separate the two if you'd like me to.

gkalpak:
@krotscheck, thx ! The diff is fine if you [ignore whitespace][1] :wink:
Taking a look...

[1]: https://github.com/angular/angular.js/pull/13273/files?w=1

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,194 @@ describe("basic usage", function() {
...
+        card.$promise.catch(function(rejection) {
+          rejectResponse = rejection;
+        });
+
+        // Make sure all promises resolve.
+        $rootScope.$apply();
+
+        // Ensure the resource promise was rejected
+        expect(card.$resolved).toBeTruthy();
+        expect(card.$promise.$$state.status).toEqual(2);
```
gkalpak: This is private API surface. We don't need to verify that. (It might change anytime anyway.)

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,194 @@ describe("basic usage", function() {
...
+        expect(rejectResponse).toBe(rejectReason);
+
+        // Ensure that no requests were made.
+        $httpBackend.verifyNoOutstandingRequest();
+      });
+
+      it('should allow per action requestError interceptor', function() {
```
gkalpak: This is essentially the same as the next test. Remove this one.

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,194 @@ describe("basic usage", function() {
...
+                }
+              }
+            }
+          });
+
+          var resource = CreditCard.query();
+          $rootScope.$apply();
+
+          expect(resource.$resolved).toBeTruthy();
+          expect(resource.$promise.$$state.status).toBe(2);
```
gkalpak: Remove this line.

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,194 @@ describe("basic usage", function() {
...
+            }
+          }
+        });
+
+        $httpBackend.expect('GET', '/CreditCard').respond([{id: 1}]);
+
+        CreditCard.query();
+
+        $httpBackend.flush();
+      });
```
gkalpak: I think you should verify that `$resolved` is true etc.

test/ngResource/resourceSpec.js
```diff
@@ -1443,6 +1632,7 @@ describe('cancelling requests', function() {
       });
 
       var creditCard = CreditCard.get();
+      $rootScope.$apply();
```
gkalpak: Is this really needed ? That mght be a BC :(

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,194 @@ describe("basic usage", function() {
       });
     });
 
+
+    describe('requestInterceptor', function() {
+      var rejectReason = {'lol':'cat'};
+      var testError = new Error();
```
gkalpak: This is not used. And I miss the following: _Test that throwing inside the `requestInterceptor` calls the `requestErrorInterceptor`._
Coincidence ? :stuck_out_tongue:

gkalpak:
I left a couple of comments. Additionally, where applicable, I think it's a good idea to use success/error callbacks to verify that the appropriate callback has been called in each case.

test/ngResource/resourceSpec.js
krotscheck: Yes. I believe it's because we're putting a promise between the GET request and the invocation to $http, which has to resolve first. We can wrap creating the requestInterceptor promise into a conditional though, which would remove the need for this line.

gkalpak: How come this is the only test that needs this ? :confused:

krotscheck: ...perhaps a sign that ngResource needs more coverage ;)

krotscheck:
@gkalpak Alright, the most recent patch has the changes you asked for.

krotscheck:
Also, re: my comment about interceptor arrays in  #9334  - i figure we can add that as a followup to this one. i.e. First, make request and response interceptors work. Second, make them accept arrays.

krotscheck:
@gkalpak Heya! Any other changes you'd like?

krotscheck:
@gkalpak @petebacondarwin Heya! Haven't heard much from y'all about this patch. Is it good? Bad? Is there some approval process I still have to go through?

petebacondarwin:
@gkalpak can you take a look at this tomorrow?

Narretz:
I have a few objections against landing this. We have unfortunately too many different concepts relating to request and response / interception / transformation in http and resource:

* Global request and response interceptors that apply to all http requests
* Individual request and response transforms for $http requests
* Individual request and response transforms for $resource requests
* individual response interceptors for $resource which are different from global interceptors, see https://github.com/angular/angular.js/issues/11409

* This commit would add requestInterceptors that seem to work more like global $http interceptors, right?

My concern is that we are adding an API to $resource that adds a feature but makes the behavior in http and resource less consistent than it is already. And once we've added a feature, it's hard to take it back.

I would rather like to see an interceptor system that works the same for $http and $resource. So

* individual response and request interceptors for $http, too (this would also help with #12095
* request and *real* response interceptors for $resource

So I am not against this PR, but I think we should see if we can have a consistent API instead.

krotscheck:
Heya! I'm more or less in agreement with "Consistency is good". I'd be happy to abandon this PR if a competing one comes along that does what you're looking for. Are you aware of any efforts to that end?

gkalpak:
I too agree that we need more consistency in `$resource`, especially `interceptors`. I have outlined my general idea in https://github.com/angular/angular.js/issues/9334#issuecomment-185437094.

IMO, this PR is a step in the right direction. Note that I would land it in 1.6.x only. So, assuming that we should bring consistency to `$resource` interceptors before `1.6.0`, we can iterate over this and break things introduced in this PR, since it won't be released yet.
In that sense, I think it is safe to, as long as we make sure to re-think the whole `interceptors` concept before `1.6.0`.

src/ngResource/resource.js
```diff
@@ -667,7 +671,18 @@ angular.module('ngResource', ['ng']).
               extend({}, extractParams(data, action.params || {}), params),
               action.url);
 
-            var promise = $http(httpConfig).then(function(response) {
+            // Start the promise chain
+            var promise;
+            if (requestInterceptor) {
```
gkalpak: What is there is only `requestErrorInterceptor` ?

gkalpak: Maybe it would be better to be consistent and do the following regardless if there re request interceptors or not:

```js
var promise = $q.
  resolve(httpConfig).
  then(requestInterceptor).
  catch(requestErrorInterecptor).
  then($http);
```

<sub>It might need an extra digest in tests, but this would be the least of the BCs :stuck_out_tongue: </sub>

gkalpak: Or something like that:

```js
var interceptedConfig = (requestInterceptor || angular.identity)(httpConfig);
var promise = $q.resolve(interceptedConfig).catch(requestErrorInterceptor).then($http);
```

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,187 @@ describe("basic usage", function() {
...
+
+      it('should call $http with the value returned from requestInterceptor', function() {
+        var CreditCard = $resource('/CreditCard', {}, {
+          query: {
+            method: 'get',
+            isArray: true,
+            interceptor: {
+              request: function(httpConfig) {
+                httpConfig.url = '/DebitCard';
+                return $q.resolve(httpConfig);
```
gkalpak: Can't you just return `httpConfig` here (without the `$q.resolve()`) ?

test/ngResource/resourceSpec.js
```diff
@@ -1366,6 +1549,78 @@ describe('errors', function() {
...
+          }
+        }
+      });
+
+      var resource = CreditCard.query();
+      resource.$promise.then(successSpy, failureSpy);
+      $rootScope.$apply();
+
+      expect(callback).toHaveBeenCalledOnce();
+      expect(callback.mostRecentCall.args[0]).toBe(rejectReason);
```
gkalpak: You can use our custom `.toHaveBeenCalledWith()` / `.toHaveBeenCalledOnceWith()` matchers for convenience.
(Here and elsewhere.)

test/ngResource/resourceSpec.js
```diff
@@ -1366,6 +1549,78 @@ describe('errors', function() {
...
+    });
+
+    it('should abort the operation if a requestErrorInterceptor throws an exception', function() {
+      var CreditCard = $resource('/CreditCard', {}, {
+        query: {
+          method: 'get',
+          isArray: true,
+          interceptor: {
+            request: function() {
+              return $q.reject(rejectReason);
```
gkalpak: Remove `rejectReason` here, so we can ensure that the failure callback is ineed called with value thrown by `requestError`.

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,187 @@ describe("basic usage", function() {
...
+                callback(rejection);
+                return $q.reject(rejection);
+              }
+            }
+          }
+        });
+
+        var resource = CreditCard.query();
+        resource.$promise.then(successSpy, failureSpy);
+        $rootScope.$apply();
```
gkalpak: Nitpick: Use `$rootScope.$digest()`, when `$apply()` isn't necessary.

gkalpak:
I left some mior comments, but generally LGTM.

WRT https://github.com/angular/angular.js/pull/13273#discussion_r59537498, I think we should do it.
It's better to be consistent, than have people wondering why their tests are not making a request when they add an interceptor. Besides, in most cases people will call `$httpBackend.flush()`, which will take care of `$rootScope.$digest()`, so this shouldn't affect many people.

krotscheck:
Okies, I'm on it.

krotscheck:
Theeeere we go. Was temporarily derailed by #13123, but used the workaround for great justice.

krotscheck:
@gkalpak Hey there! It's been a month since my last patch, which incorporates all the changes you asked for. Is there anything else we're missing?

gkalpak:
@krotscheck: Time :smiley: 

I still wish to get the ngResource interceptors/transforms story fixed for 1.6, but I have to talk people into this :stuck_out_tongue:
I'll do my best!

gkalpak:
BTW, this should also fix #12095. I would be nice to add a testcase for that (i.e. testing that the request interceptor will be able to modify the headers).

krotscheck:
Oh, all you need is time? I had some of that lying around here somewhere... now where did it... hold on... drat!

A few questions:

- What's the cutoff for 1.6? I don't have a lot of extra time, but I may be able to convince someone in the OpenStack community to look at the interceptor implementation.
- When you say: "It would be nice to add a testcase", is that a "Add this testcase or I won't merge it", or is this a "If you get around to it, yay, but otherwise this is fine."?
- What's the process from here? Do you just merge it? Will things get approved in a batch? Do I have to do a rain dance?

petebacondarwin:
The aim is to release 1.6 in September. So there is time (of one kind or another) but perhaps not the kind of time that you need :-)

gkalpak:
> I may be able to convince someone in the OpenStack community to look at the interceptor implementation.

If this is a response to my _"I have to talk people into this"_, I meant I people on the ng1 team :smiley:
(Not that they much talking into - just a tiny bit :smiley: )

--
> is that a "Add this testcase or I won't merge it", or is this a "If you get around to it, yay, but otherwise this is fine."

The latter

--
> What's the process from here?

Review  -->  LGTM  -->  LGT another team member  -->  Merge
(All sprinkled with a good overall strategy/plan for moving forward. But I prefer to make small incremental changes.) 

> Do you just merge it?

We don't merge - we rebase :stuck_out_tongue:

> Will things get approved in a batch?

I hope not.

> Do I have to do a rain dance?

It's not necessary, but people have reported it helps :stuck_out_tongue:

gercheq:
@gkalpak When do you think this will be released? Currently it's not possible to transform the request headers with $resource dynamically. 

```
                        search: {
                            method: 'GET',
                            params: {
                                organizationId: "@organizationId",
                                q: '@q'
                            },
                            transformRequest: function(data, getHeaders) {
                            
                                var token = ContactsContinuationService.getToken(),
                                    headers = getHeaders();
                            
                                headers['Continuation'] = token;
                                headers['RECEP'] = "SUCKS";
                            
                                return data;
                            },

                            url: ApiPath('/organisations/:organizationId/contacts')
                        },
```
Declaration above doesn't work. It transforms the header but the final request does not have these new headers.

I ended up wrapping the $resource with another function as mentioned here : http://stackoverflow.com/questions/20566593/dynamic-resource-headers

gkalpak:
The milestone is `1.6.x (post 1.6.0)`, so after `1.6.0` has been released (probably in 1.6.1).

test/ngResource/resourceSpec.js
```diff
@@ -978,6 +979,188 @@ describe("basic usage", function() {
...
+      });
+
+      it('should allow per action request interceptor that gets full configuration', function() {
+        var CreditCard = $resource('/CreditCard', {}, {
+          query: {
+            method: 'get',
+            isArray: true,
+            interceptor: {
+              request: function(httpConfig) {
+                callback(httpConfig);
```
Narretz: Is the callback defined somewhere?

Narretz:
There's now a merge conflict (due to https://github.com/angular/angular.js/commit/39a90a9c208236b08abde05309b0a0afdcdc4e7f I think). Must be resolved, but should be easy. Can you update the PR @krotscheck ?

krotscheck:
Sadly, I've switched jobs. You'll need to find someone else to update this patch.

Narretz:
Ok. Thanks for the head's up

Narretz:
I wrote this in Slack but it must have been overlooked: I know we talked previously about how the interceptor API is different between ngResource and $http. Did you take this into consideration when approving the PR @gkalpak ? I wonder if we could theoretically move the intereption to $http. Basically, have request-level interceptors in http, which are followed by the app-global interceptors? We don't have to do this in this PR, I just don't want to introduce an API that is blocking us in the future

gkalpak:
Yeah, I have proposed something similar [here][1] and [there][2]. My "problem" with the current implementation is that is is too `$resource`-specific and the interceptor has to know about `$resource`-internals (e.g. manipulate and return `response.resource` as explained [here][3]). This only affects `response` interceptors, so I don't think `requestInterceptors` will pose any problems.

Even if we decided to make this an `$http` feature, this PR would make it more difficult, because the `requestInterceptors` introduced here have the same signature as the per-request `$http` ones.

So, I think we are good.

[1]: https://github.com/angular/angular.js/issues/11409#issuecomment-86698665
[2]: https://github.com/angular/angular.js/issues/9334#issuecomment-185437094
[3]: https://github.com/angular/angular.js/issues/11409#issuecomment-86690672

Narretz:
Regarding `if we decided to make this an $http feature, this PR would make it more difficult, because the requestInterceptors introduced here have the same signature as the per-request $http ones.`

I guess you mean, "would *not*" make it more difficult?
And there's currently no per-request http interceptor, only the global ones, and transformRequest / transformResponse

gkalpak:
That's what I said 😛

gkalpak:
I now realized I have no idea why I didn't put it on the 1.6.0 milestone, as I clearly think this is (tiny) breaking change (see [here][1], [here][2] and [here][3]) 🤔

[1]: https://github.com/angular/angular.js/pull/13273#r53671005
[2]: https://github.com/angular/angular.js/pull/13273#r59537498
[3]: https://github.com/angular/angular.js/pull/13273#issuecomment-209414184

gkalpak:
Here is the rebased version btw: #15674

petebacondarwin:
Closing in favour of #15674

PR #13279, opened 2015-11-07
feat($sanitize): enable amending allowed attributes in $sanitize

majello:
Provide additionalAttributes function in $sanitizeProvider. When called with a comma separated string of attributes, add these attributes to the list of allowed attributes.

This potentially allows the user to introduce changes that break the security guarantees of $sanitize.

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -141,11 +141,14 @@ var $sanitizeMinErr = angular.$$minErr('$sanitize');
  */
 function $SanitizeProvider() {
   var svgEnabled = false;
+  var addAttributes = "";
 
   this.$get = ['$$sanitizeUri', function($$sanitizeUri) {
     if (svgEnabled) {
       angular.extend(validElements, svgElements);
     }
+    var attrMap = toMap(addAttributes);
+    angular.extend(validAttrs, attrMap)
     return function(html) {
       var buf = [];
       htmlParser(html, htmlSanitizeWriter(buf, function(uri, isImage) {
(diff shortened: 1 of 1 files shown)


Narretz:
The problem with an editable white-list is that it is global, so other modules' configs could influence what makes it through sanitize, possibly creating security risks. 
Why do you need configurable attributes? Because there are some attributes missing from the whitelist?
@petebacondarwin wdyt about this one?

majello:
I am aware of that risk. For me what is mostly missing is the style attribute, but there might be others as well. I'm in the early stages of my project and don't have a full handle on the requirements yet. 

Maybe we could have something like a total approved list, and a actual approved list, from which modules could pick, in order to avoid things like onclick handlers. That way there would be a way to control the aggressiveness of sanitize without completely opening up. 

I can't see a way around the globally defined whitelist so far. If anybody has any ideas, I'm willing to put some work into this.

lgalfaso:
I see a few issues with this PR:
* No tests
* Changing this within one app, changes it globally
* Allowing people to whitelist some attributes has the side-effect that may end up allowing everything just to make `$sanitize` work the way they want

The first two points are a must for this to land, the third, I would not be too happy, but willing to compromise

majello:
So I see that this approach is pretty much dead. Could you see an additional parameter to trustAsHtml() working out better? I think a way to loosen/tighten $sanitize in specific cases is desirable in general. If you disagree, I'm happy to roll my own branch for my app, but I'd prefer not to.

Narretz:
Adding a parameter to the sanitize function definitely has the benefits not affecting the whole app.

PR #13281, opened 2015-11-09
docs(ngModel.NgModelController): fix mistake in example

ignovak:
The bodies of `resolved` and `rejected` callbacks were in wrong order.


ignovak:
I signed it!

gkalpak:
I don't think this change is valid.
If the API call results in a normal response (no error), it means that the user exists (i.e. could be retrieved) and thus the validator should fail. If there is an error (e.g. the user was not found), then the validator succeeds, because the username is available.

Thx for taking the time to submit a PR, though :+1:

ignovak:
Oops, that's true. Sorry for carelessness.

PR #13286, opened 2015-11-10
perf(ngModel,ngForm): change ng-model/form controllers to use prototype methods

jbedard:
This makes the largetable-bp ng-model benchmarks 10-15% faster (down 90-100ms for me). The actual controller instantiation doesn't change too much but the overall numbers seem consistently faster, I assume all due to reducing memory usage / gc. Specifically on creation there is ~40% less memory GCed, on destruction there is about ~25% less.

Looking at the heap snapshot this is entirely from having less closures (~70k => 30k closures), other counts are all equal or very close.

Looking at the chrome timeline the number of GC events is about half and the time/gc amounts roughly match the benchpress measurements.

Worth the breaking change? The diff can probably be reduced or split into 2 commits if there's interest...

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -1,6 +1,6 @@
 'use strict';
 
-/* global -nullFormCtrl, -SUBMITTED_CLASS, addSetValidityMethod: true
+/* global -nullFormCtrl, -PENDING_CLASS, -SUBMITTED_CLASS
  */
 var nullFormCtrl = {
   $addControl: noop,
@@ -11,6 +11,7 @@ var nullFormCtrl = {
   $setPristine: noop,
   $setSubmitted: noop
 },
+PENDING_CLASS = 'ng-pending',
 SUBMITTED_CLASS = 'ng-submitted';
 
 function nullFormRenameControl(control, name) {
@@ -61,22 +62,28 @@ function nullFormRenameControl(control, name) {
  */
 //asks for $scope to fool the BC controller module
 FormController.$inject = ['$element', '$attrs', '$scope', '$animate', '$interpolate'];
-function FormController(element, attrs, $scope, $animate, $interpolate) {
-  var form = this,
-      controls = [];
+function FormController($element, $attrs, $scope, $animate, $interpolate) {
+  this.$$controls = [];
 
   // init state
-  form.$error = {};
-  form.$$success = {};
-  form.$pending = undefined;
-  form.$name = $interpolate(attrs.name || attrs.ngForm || '')($scope);
-  form.$dirty = false;
-  form.$pristine = true;
-  form.$valid = true;
-  form.$invalid = false;
-  form.$submitted = false;
-  form.$$parentForm = nullFormCtrl;
+  this.$error = {};
+  this.$$success = {};
+  this.$pending = undefined;
+  this.$name = $interpolate($attrs.name || $attrs.ngForm || '')($scope);
+  this.$dirty = false;
+  this.$pristine = true;
+  this.$valid = true;
+  this.$invalid = false;
+  this.$submitted = false;
+  this.$$parentForm = nullFormCtrl;
 
+  this.$$element = $element;
+  this.$$animate = $animate;
+
+  setupValidity(this);
+}
+
+FormController.prototype = {
   /**
    * @ngdoc method
    * @name form.FormController#$rollbackViewValue
@@ -88,11 +95,11 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
    * event defined in `ng-model-options`. This method is typically needed by the reset button of
    * a form that uses `ng-model-options` to pend updates.
    */
-  form.$rollbackViewValue = function() {
-    forEach(controls, function(control) {
+  $rollbackViewValue: function() {
+    forEach(this.$$controls, function(control) {
       control.$rollbackViewValue();
     });
-  };
+  },
 
   /**
    * @ngdoc method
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngModel.js
```diff
@@ -1075,13 +1093,17 @@ var ngModelDirective = ['$rootScope', function($rootScope) {
...
 
+          function setTouched() {
+            modelCtrl.$setTouched();
+          }
+
           element.on('blur', function(ev) {
             if (modelCtrl.$touched) return;
 
             if ($rootScope.$$phase) {
-              scope.$evalAsync(modelCtrl.$setTouched);
```
jbedard: This is an example of the breaking change...

dcherman: Is it possible to add un-minified only checks to aid in the transition for these methods?

`console.assert(this instanceof NgModelController, 'Some link to the breaking change message');`

It's not worth these checks in the minified build imo since those should be the production assets, but it may be worth it in the debug/unminified builds.

dcherman:
I think if communicated well, this change is worth the break.  For complex apps, the number of NgModelControllers being constructed and torn down on a page can represent a non-trivial amount of memory consumption and GC hits (as your tests show)

Out of curiosity, would this allow engines like v8 to better optimize this controller now that it uses a real prototype, or would have the similarly structured instances been backed by the same hidden class anyway?

Narretz:
I like it, now and then I wonder why it wasn't done like this before. I cannot estimate how many apps will be affected by the breaking change, but I think the perf gains are worth it.

src/ng/directive/form.js
```diff
@@ -132,29 +138,29 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
...
+  $addControl: function(control) {
     // Breaking change - before, inputs whose name was "hasOwnProperty" were quietly ignored
     // and not added to the scope.  Now we throw an error.
     assertNotHasOwnProperty(control.$name, 'input');
-    controls.push(control);
+    this.$$controls.push(control);
 
     if (control.$name) {
-      form[control.$name] = control;
+      this[control.$name] = control;
```
lgalfaso: I know this is unrelated to this change, but this looks dangerous

src/ng/directive/ngModel.js
```diff
@@ -570,34 +592,34 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
     processAsyncValidators();
 
     function processParseErrors() {
```
lgalfaso: can these functions be moved to the prototype?

src/ng/directive/form.js
```diff
@@ -61,22 +61,28 @@ function nullFormRenameControl(control, name) {
...
+  this.$name = $interpolate($attrs.name || $attrs.ngForm || '')($scope);
+  this.$dirty = false;
+  this.$pristine = true;
+  this.$valid = true;
+  this.$invalid = false;
+  this.$submitted = false;
+  this.$$parentForm = nullFormCtrl;
+
+  this.$$element = $element;
+  this.$$animate = $animate;
```
lgalfaso: I find the pattern of exposing services in the instance somehow strange, but find the tradeoff (for using prototype) worth it

lgalfaso:
Overall, I like this PR. It does need a better commit message and a proper breaking change description and migration

src/ng/directive/form.js
jbedard: Yea I don't like it either, but there's really no other option when defining controllers :(

dcherman: FWIW, We've encountered a similar situation in our codebase when using `.controller` directly since the constructor is what has dependencies injected.  In order to work around this, we've resolved to doing something like this:

```js
angular.module('mod').factory('MyController', function($timeout) {
    return class MyController {
        /* @ngInject */
        constructor() {
            this.waitAMinute().then(() => console.log('Done!'));
        }

        waitAMinute() {
            return $timeout(60000);
        }
    }
});
```

That does require one to inject the controller into the directive and pass the function directly to the controller property rather than using the string syntax, and it also won't work with `ng-controller`.  Just throwing that out there in case it inspires any ideas for future refactors.

src/ng/directive/ngModel.js
jbedard: That would require a slightly bigger change because these methods depend on local state defined within `$$runValidators` (such as `doneCallback`). Definitely could be done though...

dcherman:
@lgalfaso Quick thought; right now, we have access to both `this` and `ctrl` in the controllers, so we can detect scenarios where `ctrl !== this`.  Do you think it's worthwhile to add a warning in the unminified build so that if an unbound method was passed around, you get warnings?

That should make it easier to migrate to 1.6 with this breaking change since if you don't get these warnings, it's relatively certain that you won't run into problems.

jbedard:
I've rebased this...

jbedard:
Rebased again. Let me know if there's anything I can do to help get this into 1.6...

Narretz:
@jbedard I think we should merge it. You've never responded to https://github.com/angular/angular.js/pull/13286#discussion_r44686627, do you know what @lgalfaso means by this?

@dcherman How can we log something in the unminfied build and have it removed in the minified buld? Can we use the Closure compiler for this?

dcherman:
@Narretz I'm not that familiar with that Closure Compiler offers in the way of options, but UglifyJS has a `drop_console` option that will remove console statements; that's how we remove chunks of code in our minified artifacts where I work.  Since we're talking about simple function calls with known styles enforced by linting, it would also be relatively easy to write a regex to remove those invocations at build time if an existing option/plugin doesn't work for you.

jbedard:
@Narretz he is referring to the fact that we are blindly setting `this[name]` where `name` is a value from an HTML attribute. What if `name` was "hasOwnProperty" or "$setValidity" etc. This is unrelated to this PR, but it is a good point.

@dcherman I'd rather not add debug-only code like that because many people who use their own (or no) minification won't remove it. An opt-in tool might work, but then no one will use it :|

Narretz:
We've decided that we are not gonna log any warning messages in the console before this change. It's an old-fashioned BC and just have to deal with it (TM) during migration.

Narretz:
(I have rebase this locally, so no need to rebase it again. Just waiting if someone has any last thoughts on this)

jbedard:
I also just rebased this!

:+1: for the BC without any extra warnings

dcherman:
:clap:

PR #13302, opened 2015-11-14
docs(timeout): minor grammar fix

etsuo:


Narretz:
Thanks for the PR! That makes it easier to read but it changes the semantics a bit, because not the value of the promise is resolved, but the promise itself. How about `The promise will be resolved with the return value of the `fn` function`.

etsuo:
Thanks @Narretz - that makes more sense. I'll update my PR.

PR #13309, opened 2015-11-15
docs(guide/Conceptual Overview): $http.success is deprecated

lborgav:
Using the standard then method instead success, because success is deprecated


lborgav:
I signed it!

PR #13318, opened 2015-11-16
fix($compile): support merging special attribute names in `replace` d…

### LINKED ISSUE #13317: fix(Attribute): allow unconventional attribute names on elements
See https://github.com/ngUpgraders/ng-forward/issues/54#issuecomment-157050442

petebacondarwin:
…irectives

When compiling a `replace` directive, the compiler merges the attributes from
the replaced element onto the template element.

Unfortunately, `setAttribute` and other related DOM methods do not allow certain
attribute names - in particular Angular 2 style names such as `(click)` and `[value]`.

This is relevant when using ngForward with Angular Material, since the `mgButton`
directive uses `replace` and in the former you often use `(click)`.

This fixes the problem but for those special attributes the speed is considerably
slow.

Closes #13317

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -989,6 +989,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     function($injector,   $interpolate,   $exceptionHandler,   $templateRequest,   $parse,
              $controller,   $rootScope,   $document,   $sce,   $animate,   $$sanitizeUri) {
 
+    var SIMPLE_ATTR_NAME = /^\w/;
+    var QUOTE_REGEX = /'/g;
+    var specialAttrHolder = document.createElement('div');
     var Attributes = function(element, attributesToCopy) {
       if (attributesToCopy) {
         var keys = Object.keys(attributesToCopy);
@@ -1168,7 +1171,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           if (value === null || isUndefined(value)) {
             this.$$element.removeAttr(attrName);
           } else {
-            this.$$element.attr(attrName, value);
+            if (SIMPLE_ATTR_NAME.test(attrName)) {
+              this.$$element.attr(attrName, value);
+            } else {
+              setSpecialAttr(this.$$element[0], attrName, value);
+            }
           }
         }
 
@@ -1221,6 +1228,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       }
     };
 
+    function setSpecialAttr(element, attrName, value) {
+      specialAttrHolder.innerHTML = "<span " + attrName + "='" + (value || '').replace(QUOTE_REGEX, '&quot;') + "'>";
+      var attribute = specialAttrHolder.firstChild.attributes[0].cloneNode();
+      element.attributes.setNamedItem(attribute);
+    }
 
     function safeAddClass($element, className) {
       try {
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -990,7 +990,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
              $controller,   $rootScope,   $document,   $sce,   $animate,   $$sanitizeUri) {
 
     var SIMPLE_ATTR_NAME = /^\w/;
-    var QUOTE_REGEX = /'/g;
     var specialAttrHolder = document.createElement('div');
     var Attributes = function(element, attributesToCopy) {
       if (attributesToCopy) {
@@ -1229,7 +1228,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     };
 
     function setSpecialAttr(element, attrName, value) {
-      // Attributes names that do not start with letters cannot be set using `setAttribute`
+      // Attributes names that do not start with letters (such as `(click)`) cannot be set using `setAttribute`
       // so we have to jump through some hoops to get such an attribute
       // https://github.com/angular/angular.js/pull/13318
       specialAttrHolder.innerHTML = "<span " + attrName + ">";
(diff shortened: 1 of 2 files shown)


src/ng/compile.js
```diff
@@ -1221,6 +1228,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       }
     };
 
+    function setSpecialAttr(element, attrName, value) {
+      specialAttrHolder.innerHTML = "<span " + attrName + "='" + (value || '').replace(QUOTE_REGEX, '&quot;') + "'>";
+      var attribute = specialAttrHolder.firstChild.attributes[0].cloneNode();
```
gkalpak: [According to MDN][1], `cloneNode()` is supposed to be deprecated.

<sub>
It also says that, since Chrome v45, `Attr` no longer inherits from `Node`, but `cloneNode()` seems to still work on Chrome v46.
</sub>

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Attr#Deprecated_properties_and_methods

petebacondarwin: Yeah, trouble is I can't find any other way of achieving the desired solution. Any other ideas?

petebacondarwin: Ahah how about this:

```js
specialAttrHolder.innerHTML = "<span " + attrName + "='" + (value || '').replace(QUOTE_REGEX, '&quot;') + "'>"
var span = specialAttrHolder.firstChild;
var attribute = span.attributes[0];
span.removeAttribute(attrName);
element.attributes.setNamedItem(attribute);
```

This avoids any deprecated API and could possibly be a lot quicker as we are not cloning anything.

timkindberg:
so this performance hit will only happen in certain scenarios? It will not happen for everything using ng-forward special attributes correct?

src/ng/compile.js
```diff
@@ -989,6 +989,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     function($injector,   $interpolate,   $exceptionHandler,   $templateRequest,   $parse,
              $controller,   $rootScope,   $document,   $sce,   $animate,   $$sanitizeUri) {
 
+    var SIMPLE_ATTR_NAME = /^\w/;
+    var QUOTE_REGEX = /'/g;
```
jbedard: This looks unused...

petebacondarwin: Oh yes. That was from a previous commit in this pr

petebacondarwin:
@jbedard yes. It has no impact for almost all circumstances

jbedard:
That wasn't me asking, but good to know :)

Using `innerHTML` looks painful but I don't see any other way. Except maybe convincing people to not use `replace:true`?

petebacondarwin:
Sorry @jbedard - difficult to keep track on a phone.
@timkindberg - yes you are correct. This only occurs if `replace: true`is used on a directive AND the user of the directive uses "special" attributes such as `(click)` or `[value]`.  If `ng-forward` supported the long form of these attributes, such as `bind-value` and `on-click` then the user can always choose to use that form instead on directives that are `replace: true` and not incur any performance degradation.

I spent ages digging through the DOM API and hacking in different browsers. The final solution we have here is about as fast as I can get and doesn't use any deprecated APIs. But if anyone has time to find a faster solution that would be great.

I'm afraid that we are stuck with `replace: true` for the time being...

lgalfaso:
@petebacondarwin I wonder if we really have to support this. Would it be possible that if someone needs this, then they have to use the extended notation? My rational is that building html by concatenating strings enlarge the surface for mXSS.

petebacondarwin:
@lgalfaso I see where you are coming from about security.

I think that in this case we are safe though (please do check the accuracy of this):

* we are creating a single `div` containing a `span`, which is never connected to the main document. Any dubious HTML that is injected into this div's `innerHTML` will never be executable. (Is that correct?)
* we are using DOM manipulation (`span.attributes[0]`) to get hold of exactly the attribute object that we created in the `innerHTML`. Even if additional HTML was injected into the `innerHTML` it could not bleed through into the destination element.

lgalfaso:
@petebacondarwin this solution still has issues with attributes with special names (tried this with Chrome) http://plnkr.co/edit/0GjwipSakMskljc7X3Bl?p=preview

petebacondarwin:
@lgalfaso - thanks for spotting that. We have to use the `attributes.removeNamedItem(attribute.name)` to get it to work properly. Other combinations seem to fail for one reason or another.

petebacondarwin:
Added a test and a fix for the problem that @lgalfaso noted

lgalfaso:
@petebacondarwin can you please remove all the `console.log`?

petebacondarwin:
The console log are there because there was a failure on Firefox bit I couldn't reproduce it locally. Try pulling the previous commit for the proper solution

petebacondarwin:
Well this is interesting...

Given

```html
<h1 Ω="omega">Hello Plunker!</h1>
```

the following fails only with JQuery on Firefox 

```js
expect(h1.attr('Ω')).toEqual('omega');
```

http://plnkr.co/edit/qrmsTZsW78CYxJSi7Pyv?p=preview

Is that a bug in JQuery @mzgol?

mzgol:
Hmm, interesting, native `getAttribute` works. I'll need to look at it closer. (not today, though)

mzgol:
@petebacondarwin Seems like a bug, I reported it to the jQuery bug tracker: https://github.com/jquery/jquery/issues/2730

mzgol:
OK, please read https://github.com/jquery/jquery/issues/2730#issuecomment-157828916. So, jQuery is using `String#lowercase` on the input which is incorrect according to the spec (citations in the linked comment) but the reason it breaks only in Firefox is that only Firefox is apparently following the spec here...

PR #13320, opened 2015-11-17
Use require.resolve when possible to avoid hard coded module paths.

jdalton:
This PR use `require.resolve` when possible to avoid hard coded module paths. Avoid issues with npm@3 that dedupes/flattens packages deps.


petebacondarwin:
Thanks @jdalton - looks good (although I have not yet tested. We will merge this in after we release 1.5.0)

jdalton:
Rock!

IgorMinar:
this LGTM. @petebacondarwin any reason for not merging this now? Looks pretty harmless.

CI failed, but it looks like a flake. I restarted the build.

petebacondarwin:
The npm install failed locally for me - on Travis it is using the cached node_modules, which is probably why it passed. Also we need to look at moving Travis to a later version of node (currently 0.10.x).
So I don't want to get into this until after 1.5.0 has been released.

IgorMinar:
I see. Ok

Narretz:
Landed as https://github.com/angular/angular.js/commit/6100f10fa9e02bb6b4228e8a553158c928d171d1

PR #13323, opened 2015-11-17
docs(number): fix use of undefined decimalPlaces

rahatarmanahmed:
Replace `decimalPlaces` with `fractionSize`, as `decimalPlaces` isn't defined anywhere and is most likely meant to be `fractionSize`.


rahatarmanahmed:
I signed it!

Issue #13211, opened 2015-10-30
Angular ng-options setting pristine to false on load (post 1.4)


jb-int:
Something about the following combination of "multiple" and "as" in the options expression is making ng-options set pristine to false when the form first loads. Is this an angular bug? Does anyone know of a workaround?

http://plnkr.co/edit/c9ZIgiiisntkJJBeFFdy

Narretz:
Yep, that looks like a bug. You don't need `select as`,  but you need `multiple`. It's got something to do with us checking if the view / model has changed after the options have changed. When that's the case, we call $setViewValue with the new selection, which triggers a call to $setDirty.
As a workaround you could add a directive that runs after ngOptions which forces $pristine = true and $dirty = false after the initial loading.

jb-int:
Thanks for the quick response!

Can anyone tell me how to add a directive that runs after ngOptions?

m-amr:
@Narretz  I think i found the problem
prevoiusValue and nextValue are array for example

    previousValue = [1,2]
    nextValue = [1,2]

The are the same value but the === operator will return false because they are array

    if (!ngModelCtrl.$isEmpty(previousValue)) {
          var nextValue = selectCtrl.readValue();
          if (ngOptions.trackBy ? !equals(previousValue, nextValue) : previousValue !== nextValue) {
            ngModelCtrl.$setViewValue(nextValue);
            ngModelCtrl.$render();
          }
        }

what cause the problem is ngOptions.trackBy is undefined and 
(previousValue !== nextValue) i will be true becuase [1,2] !== [1,2]

why angular are just not use equals() function ??

PR #13326, opened 2015-11-18
fix(ngOptionsDirective): fix form pristine (use !equals function instead of !==  for issue 13211)

m-amr:
This pull request is related to Issue

https://github.com/angular/angular.js/issues/13211

it solves form $pristine and $dirty wrong values in case of multi select element 

prevoiusValue and nextValue are array for example

```
previousValue = [1,2]
nextValue = [1,2]
```

The are the same value but the === operator will return false because they are array
after some debugging i found that these lines cause the problem

```
if (!ngModelCtrl.$isEmpty(previousValue)) {
      var nextValue = selectCtrl.readValue();
      if (ngOptions.trackBy ? !equals(previousValue, nextValue) : previousValue !== nextValue) {
        ngModelCtrl.$setViewValue(nextValue);
        ngModelCtrl.$render();
      }
    }
```

what cause the problem is ngOptions.trackBy is undefined and 
(previousValue !== nextValue) i will be true becuase [1,2] !== [1,2]

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -733,7 +733,8 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         // Check to see if the value has changed due to the update to the options
         if (!ngModelCtrl.$isEmpty(previousValue)) {
           var nextValue = selectCtrl.readValue();
-          if (!equals(previousValue, nextValue)) {
+          var isNotPrimitive = ngOptions.trackBy || multiple;
+          if (isNotPrimitive ? !equals(previousValue, nextValue) : previousValue !== nextValue) {
             ngModelCtrl.$setViewValue(nextValue);
             ngModelCtrl.$render();
           }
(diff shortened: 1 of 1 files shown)


m-amr:
I signed it!

Date: Tue, 17 Nov 2015 17:13:55 -0800

Narretz:
Thanks for the PR.I think we are very specifically only using equals with track by because of performance issues. But I think this is the right place to fix it. We should probably use equals also if the select is multiple.

m-amr:
@Narretz 
Thanks for your feedback
I understand that angular is not always use equals function to compare the values
due to performance issue

i improved my solution 
now we should use equals function to compare values 
in case of  ngOptions.trackBy  or select is multiple (values are not primitive)
else we use equal operator to compare values (values are Primitive)

i added this lined to fix the problem

    var isNotPrimitive = ngOptions.trackBy || multiple;
    if (isNotPrimitive ? !equals(previousValue, nextValue) : previousValue !== nextValue) {
         ngModelCtrl.$setViewValue(nextValue);
         ngModelCtrl.$render();
     }
    
I think this fix the problem and won't cause performance Issue in case select is single value.

Narretz:
Looks good. I'll merge it tomorrow.

PR #13331, opened 2015-11-18
Firefox dual select workaround

rstuart:
Firefox (verified on versions 38..42, Windows and Linux-AMD64) can get
confused about what option is selected after options ngOptions populates
the select control when it has a empty option that is ng-disabled.  It
appears to think both first option added by ngOptions and the one
selected by ngModel are selected.  The trigger seems to be angular
leaving the empty option both selected and disabled.

This patch works around the problem.

### DIFF
--- a/test/ng/directive/selectSpec.js
+++ b/test/ng/directive/selectSpec.js
@@ -1123,5 +1123,18 @@ describe('select', function() {
       }).toThrowMinErr('ng','badname', 'hasOwnProperty is not a valid "option value" name');
     });
 
+    it('should set providedEmptyOption.selected to false', function() {
+      scope.options = ['a'];
+      scope.model = '';
+      compile(
+          '<select ng-model="model" ng-options="opt for opt in options">' +
+            '<option id=empty_option value="">Empty Option</option>' +
+          '</select>');
+      expect(element.find("option").eq(0).attr("selected")).toBe("selected");
+      scope.model = 'a';
+      scope.$digest();
+      expect(element.find("option").eq(0).attr("selected")).toBe(undefined);
+    });
+
   });
 });
--- a/test/ng/directive/selectSpec.js
+++ b/test/ng/directive/selectSpec.js
@@ -1128,12 +1128,13 @@ describe('select', function() {
       scope.model = '';
       compile(
           '<select ng-model="model" ng-options="opt for opt in options">' +
-            '<option id=empty_option value="">Empty Option</option>' +
+            '<option id="empty_option" value="">Empty Option</option>' +
           '</select>');
-      expect(element.find("option").eq(0).attr("selected")).toBe("selected");
+      var opt = element.find("option")[0];
+      expect(!opt.getAttribute("selected")).toBe(false);
       scope.model = 'a';
       scope.$digest();
-      expect(element.find("option").eq(0).attr("selected")).toBe(undefined);
+      expect(!opt.getAttribute("selected")).toBe(true);
     });
 
   });
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -446,6 +446,9 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
       var removeEmptyOption = function() {
         if (!providedEmptyOption) {
           emptyOption.remove();
+        } else {
+          emptyOption.prop('selected', false);
+          emptyOption.attr('selected', false);
         }
       };
 


rstuart:
I signed it!

Narretz:
Thanks for the PR. Do you think you could add a test for this?

rstuart:
It depends on what you want the test to do.  Triggering the bug in Firefox on demand is well near impossible, so I can't submit a test that demonstrates the bug before the patch and shows it's gone after.  But just verifying the code has executed - sure.

Narretz:
We usually need a test that shows the expected behavior, not that code is called. I've seen IE select bugs that were untestable, but this bug seems to have pretty specific prerequesites. Is it possible to put a demo of the problem on plnkr.co or similar? I mean you must have found the bug somehow.

rstuart:
I've set up the trigger conditions as I understand them here: http://plnkr.co/edit/Y0fGXXRZuHUvuYvdMSoQ?p=preview  It doesn't display the problem.  This isn't a surprise.  I found the bug in code that displayed about 6 identically setuo `<select>` elements.  Only one misbehaved.  My reasoning initially followed yours - there must be something special about the particular `<select>` displaying the problem.  I lost 3 days looking for whatever was special - but didn't find it.  Lost is probably the wrong word - it lead me into the guts of angular, and then the patch.

Note the patch is a workaround for a bug in Firefox.  A disabled `<option>` should have no effect on anything.  Given how many instances of Firefox are out there it can't raise it's head too often.  Stepping through writeNgOptionsValue in the Firefox debugger that made it obvious it was a bug in Firefox:

    >> selectElement[0].value 
    "foo"
    >> option.selectValue
    "bar"
    # Press single step to execute: selectElement[0].value = option.selectValue
    >>  selectElement[0].value
    "foo"           # ???
    >> selectElement[0].value =  option.selectValue
    "bar"
    >> selectElement[0].value
    "bar"           # !!!

So just doing the same assignment twice also fixes the problem.  I thought a patch doing the same assignment twice that might a hard sell to angularjs, do I went with that looks like it's the right thing to do - if you aren't going to remove the empty option, you should at least ensure it isn't selected.

I should really report the bug to Mozilla give it's in the nightlies, but that sounds like a lot of work given I can't reproduce it without 10's of thousands of lines of code, and a populated mysql database to drive it.

> We usually need a test that shows the expected behavior, not that code is called. 

I can and would test the code had the right effect, which is to remove the `selected` attribute.  Beyond that it's getting very hard.

rstuart:
I've add a test that verifies the "selected" attribute disappears when it is supposed to.

Narretz:
Thanks for the additional info. Bad that we cannot reproduce the bug. So let me get this straight, if the bug occurs, then ngOptions will set an incorrect value to selected.value? Or is it simply that the disabled empty option still has the "selected" attribute.

test/ng/directive/selectSpec.js
```diff
@@ -1123,5 +1123,19 @@ describe('select', function() {
       }).toThrowMinErr('ng','badname', 'hasOwnProperty is not a valid "option value" name');
     });
 
+    it('should set providedEmptyOption.selected to false', function() {
```
Narretz: ... when a genuine option has been selected

test/ng/directive/selectSpec.js
```diff
@@ -1123,5 +1123,19 @@ describe('select', function() {
...
+      scope.options = ['a'];
+      scope.model = '';
+      compile(
+          '<select ng-model="model" ng-options="opt for opt in options">' +
+            '<option id="empty_option" value="">Empty Option</option>' +
+          '</select>');
+      var opt = element.find("option")[0];
+      expect(!opt.getAttribute("selected")).toBe(false);
```
Narretz: why not `expect(opt.getAttribute("selected")).toBe(true)`

test/ng/directive/selectSpec.js
rstuart: > ngOptions will set an incorrect value to selected.value

Yes, `select.value` being wrong is end result.  More precisely, the value selected by `ng-model` is not reflected in the UI.  But it's not `ng-options` setting it incorrectly, the problem is caused by Firefox apparently ignoring the assignment `selected.value = $modelVaule` under some conditions.

I do not know precisely what those conditions are, but it appears this sequence of events is necessary but not sufficient:

 * We start out with select having one option, which is of course selected.
 * That option (which happens to be angular's EmptyOption) is then disabled (via `ng-disabled` in my case).
 *  `ng-options` then populates the options.  Firefox automagically selects the first one added because the only existing option is disabled.
 * The select controller then assigns `select.value`, but if you print it's value in a debuuger it appears this assignment has been ignored.  If you click the select box in the UI, both the 1st option and the new selected option are highlighted despite the select not being a multiselect.

When I added the code to remove both the `selected` attribute and `selected` property the bug in Firefox is no longer triggered.

test/ng/directive/selectSpec.js
rstuart: > why not expect(opt.getAttribute("selected")).toBe(true)

Indeed. And why not `expect(element.find("option").eq(0).attr("option")).toBe("seleced");`, which is what I had before.  That uses a library to ensure cross browser compatibility, which is how it should be done.  As the commit that follows says, I changed it to what it is now because it didn't work in ie9.

The slightly longer version of the story is appears jqLite() has a bug and returns something different under ie9, and so the travis-ci run failed.  That left me with `getAttribute`, which on browsers I tested returns a variety of values (`"selected"`, `true`, `null`, `undefined`). The lowest common denominator is it returns something truthy or falsey, which is what the code tests for.

Narretz:
We haven't merged this because I had some reservations about fixing this without a clear repro. We recently changed the rendering of options in ngOptions (1.5.5). Can you try and see if this fixes your problem?

petebacondarwin:

PR #13333, opened 2015-11-18
fix($animateCss): respect transition styles already on the element

Narretz:
Previously, $animateCss wouldn't use transition styles that were on the element
before the animation process started. Precisely, transition property, timing-function
and delay were overwritten in the process.

This is a fix for #12656

It's a bit of a special case, because all the transition properties are sitting on the element already before $animateCss is called.

This could also affect the animation property, I haven't checked yet.

Strangely enough, this fix doesn't work when the transition property is `all`, because for some reason $animateCss blocks transition duration etc. when it's all - see https://github.com/angular/angular.js/blob/master/src/ngAnimate/animateCss.js#L596 Removing this doesn't break any tests, though.


test/ngAnimate/animateCssSpec.js
```diff
@@ -1855,7 +1901,7 @@ describe("ngAnimate $animateCss", function() {
 
           var style = element.attr('style');
           expect(style).toContain('3000s');
-          expect(style).toContain('linear');
+          expect(style).toContain('ease');
```
petebacondarwin: why change this?

src/ngAnimate/animateCss.js
```diff
@@ -226,6 +226,7 @@ var DETECT_CSS_PROPERTIES = {
   transitionDuration:      TRANSITION_DURATION_PROP,
   transitionDelay:         TRANSITION_DELAY_PROP,
   transitionProperty:      TRANSITION_PROP + PROPERTY_KEY,
+  transitionTimingFunction: TRANSITION_PROP + TIMING_KEY,
```
gkalpak: See what you did ? Now you have to indent all of them :smile:

matsko:
This looks good, we just need to check the `ease` default value and to update the timings value.

matsko:
Please also squash both commits together since the first one fails without the second one.

matsko:
Excellent work!

Narretz:
Thanks for the review!
I've added two more test:

1. Ensure that timing-function is set to ease if not defined (that should fail if a browser doesn't follow spec)
2. Ensure that transition-property is 'all' if not defined. I had to remove a condition from the code for that to pass.

Narretz:
Reopening to check what travis says.

PR #13340, opened 2015-11-18
docs(ngModel.NgModelController): Example inputs behave the same when …

FesterCluck:
…they shouldn't.

The first input demonstrating $rollbackViewValue is meant to rollback the value, not make it blank. Code in it's current state makes both behave the same.

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -455,7 +455,6 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
    *       $scope.resetWithCancel = function(e) {
    *         if (e.keyCode == 27) {
    *           $scope.myForm.myInput1.$rollbackViewValue();
-   *           $scope.myValue = '';
    *         }
    *       };
    *       $scope.resetWithoutCancel = function(e) {


FesterCluck:
I signed it!

Narretz:
Hmm. If you remove `$scope.myValue = '';` from resetWithCancel, nothing happens to both inputs. That can't be right either. I wonder if something else with the example is broken.

FesterCluck:
The core feature we're aiming at here is when hitting escape after changing
the value, it should reset to the previous value (from the model).

Perhaps a blank initial value is a bad use case. What the demo aims to
demonstrate is that the first function restores the model value to the view
value, while the latter must overwrite the model value. Test it giving
$scope.myValue = 'whatever'; and it will make much more sense.

If it does, I'll update my pull request accordingly.

Narretz:
On further investigation: the use case for rollbackViewValue is actually quite obsolete. If by pressing ESC you set assign a value to the model like "whatever", and then blur the input, it will not set it to the last thing you've entered. Somewhere along the line, it must have been built in that this handled correctly. I will change this example to simply show that $rollbackViewValue resets the view to the last commited value. Thanks for bringing attention to this.

Narretz:
On even further investigation, the described problem does happen, but only when you are trying to set the model to a value that it is already set to, like ''. The thing is that the description of the example is not set up to elicit this behavior.

PR #13341, opened 2015-11-18
Unnecessary input value update

FesterCluck:
Removing lines because they completely confuse the examples. Script for Plunkr examples needs updating as well.


Narretz:
As noted in https://github.com/angular/angular.js/pull/13340#issuecomment-161795404, the example isn't really valid anymore (at least in 1.4). I don't really wnat to nix it from the migration docs, because that may be true for 1.3. Did you encounter this confusing example when updating to 1.3 or 1.4?

Narretz:
I think the current state of the docs is okay.

PR #13343, opened 2015-11-19
Copy performance cherry-picks for 1.3.x

petebacondarwin:
@bourey - please take a look and see if this would help with 1.3 performance.


petebacondarwin:
Wrong base - ignore

PR #13344, opened 2015-11-19
1.3.x copy() performance cherry-picks

petebacondarwin:


petebacondarwin:
@bourey please take a look and see if this would help with 1.3.x performance.

PR #13349, opened 2015-11-19
fix(core): ensure animate runner is the same with and without animations (1.4)

Narretz:
The $$AnimateRunner class is now the same for the core $animate service
and the ngAnimate $animate service. Previously, the core used a different
implementation that didn't match the ngAnimate behavior with regard
to callbacks.


Narretz:
Landed as 937942f5ada6de1bdacdf0ba465f6f118c270119

PR #13350, opened 2015-11-20
docs(error/nobase): making <base> visible in html

anasqadrei:
I couldn't see `<base>` in the html webpage, so I fixed it.


Narretz:
Thanks for the PR. I thought I could get away with `<base> `, but that doesn't work for the first case. I will fix that later.

PR #13352, opened 2015-11-20
Avoid browser input[number] defaults

antonellopasella:
Is there a way to use  <input type="number" ng-model="vm.foo" /> syntax but WITHOUT usa type=number?
I have some quircks with spinner, mouse wheel and so on.

The patched code will works, but I'm lookin for a external solution

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -748,6 +748,7 @@ var inputType = {
       </example>
    */
   'number': numberInputType,
+  'text-number': numberInputType,
 
 
   /**


antonellopasella:
I signed it!

Narretz:
Sorry, I'm not completely sure what this change is about. What is `text-number`? We provide implementations for HTML input types, not for custom input types.
Also, I can't follow this at all: 

> Is there a way to use syntax but WITHOUT usa type=number?
I have some quircks with spinner, mouse wheel and so on.

Narretz:

Issue #13172, opened 2015-10-26
Select with ngRequired not invalid with unmatched Model set from scope


mailinger-mate:
I might have found some inconsistency in select validity. When setting the model to a value not listed in options, the form controller remains valid. I think the right behavior would be an invalid state, as the model is not matching the range of available options.

http://plnkr.co/edit/2qm7P1ag2wMCYOwDGRsF

Narretz:
Oh, yes when the select has required, it should add the required error. Looks like a bug.

Narretz:
The problem is that the select element has its value property set to whatever is set on the scope, even if there's an error somewhere. That makes sense for inputs, because they would otherwise be empty. required uses this value to check if a control is "empty", and if it is, sets the error. Select is a special case, though ... I could see how it would make sense to set the error if the model finds not match in the options. Wdyt @petebacondarwin?

petebacondarwin:
I didn't think we did validation on programmatic changes for any input control?

gkalpak:
@Narretz, just to be clear, are you suggesting that the `required` error is set when the unknown option is selected ?

Narretz:
Yes, that would be the change.

For the other question, we do validate after the model is programmatically
changed, but we do not set the model to undefined if there's an error in
that case.
Am 27.10.2015 19:48 schrieb "Georgios Kalpakas" <notifications@github.com>:

petebacondarwin:
OK, so setting the value of a select so that the unknown option is displayed should cause `required` validation to fail.

Spidy88:
I think rather than labeling this as a bug with `ngRequired`, it should be a feature request for an additional validation directive like `selected-from` or `value-within` that both take an array of valid options.

The challenge I'm facing when thinking about this issue is that `ngOptions` is not the only way to create a select element with options. It's definitely the preferred way but what about custom directives for creating your own options, or the common beginner method of using `ngRepeat`. Both valid ways to create a select statement, and you are still able to use `ngModel` with validators.

The solution this issue is requesting would only support using `ngOptions`. Maybe that's the intent, but I wanted to broach this subject before spending anytime submitting a PR

Narretz:
If we can agree that the select should be invalid when the model is not matched by any option, we should fix this in the pure select, too. Required uses a function in the ngModelController called isEmpty which is called with the current viewValue and when it returns true it will set the required error. Select and ngOptions overwrite this function with their own implementation. To fix this bug we should account for the unmatched case in this logic. 
[...]

PR #13354, opened 2015-11-20
fix(ngRequired): set valid to false if select value is required (value is not in options)

m-amr:
This pull request solve Issue 13172
https://github.com/angular/angular.js/issues/13172

i validate that the view value should be one of the select option
so i overrideng ModelCtrl.$validators.required function in ngOptionsDirective

```
ngModelCtrl.$validators.required = function(modelValue, viewValue) {

      if (!attr.required) {
        return true;
      } else if (ngModelCtrl.$isEmpty(viewValue)) {
        return false;
      }

      var isValidOption = true;
      var viewOptions = [];
      // Get all option and add them to viewOptions array
      angular.forEach(options.items, function(item) {
        viewOptions.push(options.getViewValueFromOption(item));
      });

      // In case of multiple view model is an array so validate all view values
      // if one of them doesn't match set isValidOption to false
      if (multiple) {
        for (var i = 0, length = viewValue.length; i < length; i++) {
          if (viewOptions.indexOf(viewValue[i]) === -1) {
            isValidOption = false;
             break;
          }
        }
      } else {
        if (viewOptions.indexOf(viewValue) === -1) {
          isValidOption = false;
        }
      }

      return isValidOption;
    };
```

we have the following cases

 1 - case  attr.required is false i return true

 2 - case $isEmpty(viewValue) return false

 3 - case is required and not empty now we need to validate that value is one of the options
 and in this case we have two subcases

  a - select in multiple and value is an array
  we check that all the values of the array must be in of the select options to return true

  b - select is a single value so this value must be one of the select options to return true

i think this will solve the problem,


m-amr:
I signed it!

Narretz:
Thanks for the PR. It's going in the right direction, but there are a few things that should be done differently. 
1. It's adding the required validator to every select with ngoptions, which also means it gets called every time the options change. The (ng)Required directive should decide if the validator gets added not the ngoptions directive. We are duplicating code here and would have to change it in two places. => The logic should go into the $isEmpty function. I assume you didn't put it into the isEmpty function because it gets also called in the updateOptions function?

2. It's doing too much work. When validation happens, we should have done all the operations we need to determine if we have a real option or not. Maybe we need to set a flag during updateOptions.

Do you want to take another shot at this?

src/ng/directive/ngOptions.js
```diff
@@ -728,6 +728,39 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+          var isValidOption = true;
+          var viewOptions = [];
+          // Get all option and add them to viewOptions array
+          angular.forEach(options.items, function(item) {
+            viewOptions.push(options.getViewValueFromOption(item));
+          });
+
+          // In case of multiple view model is an array so validate all view values
+          // if one of them doesn't match set isValidOption to false
+          if (multiple) {
```
Narretz: I don't think we want this behavior with `multiple` the required error should only be set if no option is selected. If it's multiple and a at least option is selected, then the required error doesn't make much sense. Maybe we should actually introduce a new kind of error called `optionMismatch` or something like that.

m-amr:
@Narretz 
Thanks for your feedback
I understand the problem and i will try to improve mu solution.

m-amr:
I have improved by solution

1- I have added this isViewOptionValid function to validate if this value is found in select option or not

    function isViewOptionValid(viewValue) {

        var isValidOption = false;
        var viewOptions = [];
        // Get all option and add them to viewOptions array
        angular.forEach(options.items, function(item) {
            viewOptions.push(options.getViewValueFromOption(item));
        });

        // In case of multiple view is an array so validate all view values
        // if one of them match set isValidOption to true
        if (multiple) {
            for (var i = 0, length = viewValue.length; i < length; i++) {
                if (viewOptions.indexOf(viewValue[i]) > -1) {
                    isValidOption = true;
                    break;
                }
            }
        } else {
            if (viewOptions.indexOf(viewValue) > -1) {
                isValidOption = true;
            }
        }

        return isValidOption;
      }

2 - I have override $isEmpty function to check if option is valid or not

     // Copy the implementation of $isEmpty function to be used in overwritten one.
      var $$isEmpty = ngModelCtrl.$isEmpty;
      ngModelCtrl.$isEmpty = function(value) {
        // if value if empty return true
        // or
        // if option is valid return false
        // else if option is not valid return true

        return $$isEmpty(value) || !isViewOptionValid(value);
      };

3 - if select is multiple and value is an array we check that  at least one of the values of the array must be in of the select options to be valid

4 - if have removed 
    
    ngModelCtrl.$isEmpty = function(value) {
      return !value || value.length === 0;
    };

since it is already overwritten in selectDirective with same code in case of multiple

test/ng/directive/ngOptionsSpec.js
```diff
@@ -2532,8 +2532,8 @@ describe('ngOptions', function() {
       expect(scope.value).toBe(false);
 
       scope.$apply('required = true');
-      expect(element).toBeValid();
```
Narretz: Imo, this is not right. required means that at least one option must be selected, it doesn't matter if the value is false. I'll look into why this test is failing.

Narretz:
Thanks for updating this! The test fails because you are not setting isOptionValid to true whenever an option is selected from the ui, i.e. in the readValue fn. I'll take this PR from here, thanks for your work!

m-amr:
@Narretz  Thank for replying
I have tried another solution 
i added this function to options object,
i found it is better to make the options object responsible to tell us it's valid state so we can call this function any time options.isSelectedOptionValid()

     function isSelectedOptionValid() {

          var selectedValue = selectElement.val();

          if (!selectedValue) {
            return false;
          }

          // in case  select is multiple then check
          // if one of the selected values matched with the options  return true
          // or return false if no value match
          if (isArray(selectedValue)) {
            for (var i = 0, length = selectedValue.length; i < length; i++) {
              if (selectValueMap[selectedValue[i]]) {
                return true;
              }
            }
            return false;
          }

          return !!selectValueMap[selectedValue];

        }

I check if the value found in selectValueMap to know if option is valid or not.
and validation is need to be done every time the model change since that it can be changed 
programatically with a value that doesn't match the options.
Thanks a lot for replying and your feedback .

changus:
Hi @Narretz, you mentioned you were taking the PR from here, did you ever commit this?  This seems to still be an issue for me.

Narretz:
@changus You are right. This is still on my radar. I had some other ideas about the implementation, that's why it was never merged. I'll try to get it in until the end of the month.

changus:
@Narretz Awesome, thank you!  I did find another thread:
https://github.com/angular/angular.js/issues/13172 

with this pull request:
https://github.com/angular/angular.js/pull/13962

but I copied those changes into my select.js and did not seem to fix the issue. 

Thanks in advance for looking into it!

Narretz:
We've decided that this should be handled by the application instead of imposing it by default. To make this easier, 1.6.5 will expose info about the unknown option via the selectCtrl: https://github.com/angular/angular.js/commit/6fc0f02ad24a6ac8bef517210e6d7e679c3988fd

PR #13356, opened 2015-11-20
feat(filterFilter): allow overwriting the special `$` property name

### LINKED ISSUE #13313: Unable to watch the changes of the property '$' of an object which is used by filterFilter.
Unable to watch the changes of the property '$' of an object which is used by filterFilter.

```
//object to search 
$scope.search = {'$': 'something...'}

//watch the object
$scope.$watch('search',
  function (newValue, oldValue) {
    //do action 
    results = filter(list, search);
    ...
  }, true);
```
The "$" in the function filterFilter represents any property, however it represents built-in variable in the function equals(watch). 
How to resolve the conflict between them?

filterFilter: https://github.com/angular/angular.js/blob/master/src/ng/filter/filter.js#L235
equals(watch): https://github.com/angular/angular.js/blob/master/src/Angular.js#L992
Smart-Table: https://github.com/lorenzofox3/Smart-Table/blob/master/src/stTable.js#L122

gkalpak:
Previously, the special property name that would match against any
property was hard-coded to `$`.
With this commit, the user can specify an arbitrary property name,
by passing a 4th argument to `filterFilter()`. E.g.:

``` js
var items = [{foo: 'bar'}, {baz: 'qux'}];
var expr  =  {'%': 'bar'};

console.log(filterFilter(items, expr, null, '%'));   // [{foo: 'bar'}]
```

Fixes #13313

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -22,10 +22,11 @@
  *   - `Object`: A pattern object can be used to filter specific properties on objects contained
  *     by `array`. For example `{name:"M", phone:"1"}` predicate will return an array of items
  *     which have property `name` containing "M" and property `phone` containing "1". A special
- *     property name `$` can be used (as in `{$:"text"}`) to accept a match against any
- *     property of the object or its nested object properties. That's equivalent to the simple
- *     substring match with a `string` as described above. The predicate can be negated by prefixing
- *     the string with `!`.
+ *     property name (`$` by default) can be used (e.g. as in `{$: "text"}`) to accept a match
+ *     against any property of the object or its nested object properties. That's equivalent to the
+ *     simple substring match with a `string` as described above. The special property name can be
+ *     overwritten, using the `specialProp` parameter.
+ *     The predicate can be negated by prefixing the string with `!`.
  *     For example `{name: "!M"}` predicate will return an array of items which have property `name`
  *     not containing "M".
  *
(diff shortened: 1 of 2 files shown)


src/ng/filter/filter.js
```diff
@@ -59,6 +60,9 @@
  *     Primitive values are converted to strings. Objects are not compared against primitives,
  *     unless they have a custom `toString` method (e.g. `Date` objects).
  *
+ * @param {string=} specialKey The special property name that matches against any property.
```
petebacondarwin: Can we call this parameter `anyPropertyKey`?

petebacondarwin:
Other than the param name, this all looks good to me. But I think we should wait for 1.5.0 to land now.

gkalpak:
Updated the param name. I'm fine with waiting for 1.5.0 (since this is just an enhancement anyway).
I'm not sure what milestone to set. `1.5.x` ?

Narretz:
Since this not a BC, should we add it to 1.5?

gkalpak:
> But I think we should wait for 1.5.0 to land now.

Maybe I miunderstood @petebacondarwin's comment. He probably meant to merge this in a subsequent `1.5.x` release. He would know best what he meant though :stuck_out_tongue: 

Since this is neither a BC nor a fix, I'm fine either way.

petebacondarwin:
I think I meant that we were about about release 1.5.0 so we should just put it in a 1.5.x release

PR #13363, opened 2015-11-23
fix($animateCss): consider options.delay value for closing timeout

### LINKED ISSUE #13355: ngAnimate: Jumps in transition when 'delay' is set with $animateCss, but not set in CSS
Plnkr: http://plnkr.co/edit/DGI0fUHtzQkyIw6CKuJ6?p=preview

When I run a transition using $animateCss with a delay set in the params, the transition will jump suddenly sometimes, especially when `delay` = `duration`.

I don't think this is supposed to be happening.

This issue is easily fixed by setting `transition-delay` in the object's CSS to **any value at all** (which gets ignored). I am not sure why no-value behaves differently from `transition-delay: [any value]`.

A general workaround for anyone having trouble is to put the following at the top of your CSS:
````
* {
  transition-delay: 0s;
}
````
since this is the default value for `transition-delay` anyway.

### PR DESCRIPTION
**Pull request by Narretz**
Previously, options.delay was only considered when a class added an
extra transition style, because that leads to style recalculation.

Fixes #13355

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -622,7 +622,12 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       }
 
       if (options.delay != null) {
-        var delayStyle = parseFloat(options.delay);
+        var delayStyle;
+        if (typeof options.delay !== "boolean") {
+          delayStyle = parseFloat(options.delay);
+          // number in options.delay means we have to recalculate the delay for the closing timeout
+          maxDelay = Math.max(delayStyle, 0);
+        }
 
         if (flags.applyTransitionDelay) {
           temporaryStyles.push(getCssDelayStyle(delayStyle));
--- a/test/ngAnimate/animateCssSpec.js
+++ b/test/ngAnimate/animateCssSpec.js
@@ -1234,6 +1234,74 @@ describe("ngAnimate $animateCss", function() {
             $timeout.flush();
           }).not.toThrow();
         }));
+
+        it("should consider a positive options.delay value for the closing timeout",
+          inject(function($animateCss, $rootElement, $timeout, $document) {
+
+          var element = jqLite('<div></div>');
+          $rootElement.append(element);
+          jqLite($document[0].body).append($rootElement);
+
+          var options = {
+            delay: 3,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/ngAnimate/animateCss.js [hunk 1]
```diff
@@ -623,6 +623,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
 
       if (options.delay != null) {
         var delayStyle = parseFloat(options.delay);
+        maxDelay = Math.max(options.delay, 0);
```
gkalpak: Shouldn't this rather be `delayStyle` (instead of `options.delay`) ?

Narretz: Hmm. The maxDelay is mainly used for determining the closing timeout. And options.delay overwrites values set in the transition styles.

gkalpak: `delayStyle` is just `parseFloat(options.delay)`.

But thinking about it again, neither option will work as expected in case `options.delay === true` (which is a valid option according to the docs).
From a superficial look, we might need to make something like on [animateCss.js#L849-L850][1] and animateCss.js#L863-L867][2].

[1]: https://github.com/Narretz/angular.js/blob/fix-nganimate-css-13355/src/ngAnimate/animateCss.js#L849-L850
[2]: https://github.com/Narretz/angular.js/blob/fix-nganimate-css-13355/src/ngAnimate/animateCss.js#L863-L867

Narretz: Right, that looks more complex than I initially thought. I'll discuss this with Matias tomorrow.

test/ngAnimate/animateCssSpec.js
```diff
@@ -1234,6 +1234,67 @@ describe("ngAnimate $animateCss", function() {
             $timeout.flush();
           }).not.toThrow();
         }));
+
+        it("should consider a positive options.delay value for the closing timeout",
```
Narretz: I'm not sure why this test fails. It should ensure that the animation ends exactly then when it's supposed to be, not before.

Narretz: It fails with jquery, because the .css implementation is different from jqlite's

PR #13365, opened 2015-11-24
Modified the short date using yyyy

chowlouise:
Improved short date for nl, fr,, en and de

### DIFF
--- a/src/ngLocale/angular-locale_de-de.js
+++ b/src/ngLocale/angular-locale_de-de.js
@@ -90,7 +90,7 @@ $provide.value("$locale", {
     "mediumDate": "dd.MM.y",
     "mediumTime": "HH:mm:ss",
     "short": "dd.MM.yy HH:mm",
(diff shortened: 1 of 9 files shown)


lgalfaso:
Hi, thanks for the PR! The locale information is retrieved and auto-generated from Google closure library. If there is any change, then this change should be performed there (and the changes will get eventually propagated here). Check "Localization Issue" at https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-found-an-issue

Closing this but feel free to comment on the issue when the changes are accepted in the other project and the locale files need to be generated.

chowlouise:
by the way i could not find the files in https://github.com/google/closure-library/tree/master/closure/goog/i18n? which file i need to change in order to achieve the same changes?

lgalfaso:
The file is https://github.com/google/closure-library/blob/master/closure/goog/i18n/datetimesymbols.js

chowlouise:
thanks a lot

Issue #13367, opened 2015-11-24
Possible bug: Optional binding =? not required anymore in 1.4 where it is in 1.3


cburgdorf:
This is a follow up to https://github.com/angular/angular.js/pull/9216#issuecomment-159261159

I noticed that there is a breaking change regarding `=?` between `1.3` and `1.4`.

This example throws with an exception when you try to toggle the second zippy because it doesn't use the open binding and the binding isn't marked as optional. The example is based on 1.3.

http://plnkr.co/edit/Ef2iKOde3TmGzJHRGEhE?p=preview

The same example using 1.4 does not throw however.

http://plnkr.co/edit/cWxO6shBz7OaiewCDMHp?p=preview

petebacondarwin:
It seems that if the value is non-assignable we now never call the watch handler for setting the parent value.

We are still creating the correct `parentSet` function, at https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L2754, which will throw the exception.

And we are still creating the watches correctly at https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L2780

But for some reason the handler never gets fired by the digest. I think this could be some optimisation that sees that the values are never going to change and so doesn't bother watching?

petebacondarwin:
It appears between 1.4.0-beta.2 and 1.4.0-beta.3

petebacondarwin:
`git bisect` tells us that this bug did indeed appear in 6a38dbfd3c34c8f9efff503d17eb3cbeb666d422

cburgdorf:
I wonder if that's really a bug or a feature? The question would be if there are any downsides of the behavior because for this example it turns out as something positive. It makes the example work out of the box without having to know about optional bindings.

lgalfaso:
I think this is an oversight on the PR that @petebacondarwin pointed out.

PR #13373, opened 2015-11-24
fix($parse): handle interceptors with `undefined` expressions

lgalfaso:
When calling `$parse` with `undefined` as the expression and with an interceptor, then when the function is evaluated, then call the interceptor

Closes: #13367


petebacondarwin:
Nice one @lgalfaso - merging (with additional $compile test)

PR #13377, opened 2015-11-25
adding moduleExists to angular

### LINKED ISSUE #13371: feat: method to check whether a certain module exists
Hi,

I would like to request a new feature. I am currently using the following hack to know whether a certain module was already created or not

```javascript
var module;
try { 
  module = angular.module(moduleName);
} catch(err) {
  module = angular.module(moduleName, []);
}
```
Which is of course not ideal and could be simplified if angular does somehow expose `modules`, the registered modules respectively. 

Following the use case where I am using this hack:

```javascript
function tutteliRegisterController(moduleName, dependencies, controllerName, constructor){
    var module;
    try { 
        module = angular.module(moduleName);
        module.requires.push.apply(module.requires, dependencies);
    } catch(err) {
        module = angular.module(moduleName, dependencies);
    }
    module.controller(controllerName, constructor);
}    
```

I am using this function in order that I can separate controllers into different files without the need to bother whether a module (which I am kind of using like a namespace) was already created or not.
This way I can nicely separate the angular registration from the actual functionality like so:

```javascript
tutteliRegisterController(
  'tutteli.controllers',
  ['tutteli.auth'],
  'tutteli.LoginController',
  LoginController
);

LoginController.$inject = ['tutteli.auth.AuthService'];
function LoginController(AuthService){
  //..
}
```

This is surely just a nice to have feature and it works as well without it. But maybe there are more people who would like it. Just as an idea.

robstoll:
the method moduleExists allows to verify whether a module was already created or not. Would fix #13371

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -36,7 +36,7 @@ function setupModuleLoader(window) {
     ensure(angular, 'moduleExists', function(){
       return function moduleExists(name){
         assertNotHasOwnProperty(name, 'module');
-        return modules[name] != undefined;
+        return modules[name] !== undefined;
       };
     });
 
--- a/src/loader.js
+++ b/src/loader.js
@@ -26,15 +26,15 @@ function setupModuleLoader(window) {
   return ensure(angular, 'module', function() {
     /** @type {Object.<string, angular.Module>} */
     var modules = {};
-  
+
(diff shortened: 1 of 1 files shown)


robstoll:
Trying another approach which does not require this feature, see #13371

PR #13383, opened 2015-11-25
fix(ngTransclude): fix case where ngTransclude attribute value equals its key

petebacondarwin:
Some preprocessors such as Jade will automatically provide a value for an attribute
rather than leave it empty. E.g. `<div ng-transclude="ng-transclude">`.
In these situations we still want to use the default transclusion slot.

Closes #12934


Narretz:
LGTM

josip:
:+1: thanks!

PR #13390, opened 2015-11-27
feat($locale): Include original locale ID in $locale

philBrown:
Most systems use _IETF language tag_ codes which are typically a combination of the ISO 639 language code and ISO 3166-1 country code with an underscore or hyphen delimiter. For example `en_US`, `en_AU`, etc.

Whilst the `$locale.id` comes close, the lowercase format makes it impossible to transform to an IETF tag reliably. For example, it would be impossible to deduce `en_Dsrt_US` from `en-dsrt-us`.

It would be very handy for communicating with external systems to be able to pass the IETF tag format, at least with the character casing intact.

### DIFF
--- a/i18n/src/closureI18nExtractor.js
+++ b/i18n/src/closureI18nExtractor.js
@@ -161,6 +161,7 @@ function outputLocale(localeInfo, localeID) {
   if (!localeObj.DATETIME_FORMATS) {
     localeObj.DATETIME_FORMATS = fallBackObj.DATETIME_FORMATS;
   }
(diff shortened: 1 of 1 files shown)


petebacondarwin:
@philBrown this doesn't actually work, since if you look further down the file the `localeObj` is never used. I have an alternative fix, with a unit test that gives you what you need.

petebacondarwin:
@gkalpak points out that your fix does indeed work, but since the code itself is weird, it will be better to fix it in the way I have done in the PR.

PR #13392, opened 2015-11-27
feat(numberFilter.formatNumber): add min group size to number patterns

radmaneshp:
The whole part of Polish numbers are grouped in 3's separated with a space
(and a comma for the decimal separator), but only after there are at least
5 digits in the whole part -
https://pl.wikipedia.org/wiki/Separator_dziesi%C4%99tny:
 3,14
 0,178011392519
 157,3567
 5348,26
 25 772,1
 9 212 345,576213
This change adds a new optional number format property, gMinSize, which if
set causes the whole part of a number to only be grouped if it is greater
than or equal to gMinSize. A separate pull request will be applied to the
I18N project directly to set gMinSize to 5 for Polish locales.

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -183,25 +183,31 @@ function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
     var whole = fraction[0];
     fraction = fraction[1] || '';
 
-    var i, pos = 0,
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Unfortunately we derive our formatting from the Closure-Library and this doesn't not provide the data for this minimum grouping value. So this change will not be usable.

PR #13394, opened 2015-11-27
fix(formatNumber): better corner case handling

### LINKED ISSUE #8674: number formatter outputs NaN.0 for large numbers with decimals

### LINKED ISSUE #10246: Number filter with exponentially small numbers without fractionSize
Currently, if `fractionSize` is not passed, for small numbers like `1e-50` number filter produces string `1e-50`.

It happens because `fractionSize` is undefined here: https://github.com/angular/angular.js/blob/g3_v1_3/src/ng/filter/filters.js#L152

`undefined + 1` is `NaN` and `NaN` can't be bigger than number.

Fraction size is only computed if number has no exponent.

Is it by design?

### LINKED ISSUE #12707: formatNumber produces NaN for numbers near 999999999999999934463

petebacondarwin:
The idea here is that we actually parse the input number ourselves and then do the rounding ourselves manually. This allows us more control and reliability on the output for corner case numbers.

Closes #8674
Closes #12709 
Closes #8705 
Closes #12707 
Closes #10246
Closes #10252

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -1,5 +1,9 @@
 'use strict';
 
+var MAX_DIGITS = 22;
+var DECIMAL_SEP = '.';
+var ZERO_CHAR = '0';
+
 /**
  * @ngdoc filter
  * @name currency
@@ -124,8 +128,6 @@ function currencyFilter($locale) {
      </file>
    </example>
  */
-
-
 numberFilter.$inject = ['$locale'];
 function numberFilter($locale) {
   var formats = $locale.NUMBER_FORMATS;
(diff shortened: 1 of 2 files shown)


Narretz:
I've found another issue that seems to fit the theme: https://github.com/angular/angular.js/issues/10246 Maybe the PR fixes it?

petebacondarwin:
I included the test from #10252 and it passes so I claim that this PR also fixes #10246

petebacondarwin:
This PR is adding approximately 0.3% (181 bytes) to the gzipped and minified code size for `angular.js`.

src/ng/filter/filters.js
```diff
@@ -139,93 +141,186 @@ function numberFilter($locale) {
...
+    while (numStr.charAt(--zeros) == ZERO_CHAR);
+    integerLen = e - i;
+    digits = [];
+    // Convert string to array of digits without leading/trailing zeros.
+    for (j = 0; i <= zeros; i++, j++) {
+      digits[j] = +numStr.charAt(i);
+    }
+  }
+
+  // if the overflow the maximum allowed digits then use an exponent
```
Narretz: Can this be reworded? I think there's a word missing or one too many

src/ng/filter/filters.js
```diff
@@ -139,93 +141,186 @@ function numberFilter($locale) {
...
 
-  var numStr = number + '',
-      formatedText = '',
-      hasExponent = false,
-      parts = [];
+    // determine fractionSize if it is not specified; `+fractionSize` converts it to a number
+    fractionSize = (isUndefined(fractionSize)) ? Math.min(Math.max(minFrac, fractionLen), maxFrac) : +fractionSize;
 
-  if (isInfinity) formatedText = '\u221e';
+    // The indec of the digit to where rounding is to occur
```
Narretz: index

src/ng/filter/filters.js
```diff
@@ -139,93 +141,186 @@ function numberFilter($locale) {
...
 
-  var isNegative = number < 0;
-  number = Math.abs(number);
+  if (i == (zeros = numStr.length)) {
+    // Zero.
+    digits = [0];
+    integerLen = 1;
+  } else {
+    // Determine trailing zeros.
+    while (numStr.charAt(--zeros) == ZERO_CHAR);
```
Narretz: I find this hard too read. What does it even do?

src/ng/filter/filters.js
```diff
@@ -139,93 +141,186 @@ function numberFilter($locale) {
...
+ * @param  {string} numStr The number to parse
+ * @return {object} An object describing this number, containing the following keys:
+ *  - d : an array of digits containing leading zeros as necessary
+ *  - i : the number of the digits in `d` that are to the left of the decimal point
+ *  - e : the exponent for numbers that would need more than `MAX_DIGITS` digits in `d`
+ *
+ */
+function parse(numStr) {
+  var exponent = 0, digits, integerLen;
+  var e, i, j, zeros;
```
Narretz: `e` seems to have a bit more importance than i and j. Can it get a better name?

petebacondarwin:
@Narretz - I updated after your comments and rebased. One more LGTM and I'll merge it.

Narretz:
LGTM!

PR #13400, opened 2015-11-27
feat(ngView): reference resolved locals in scope

shahata:
this will make it easier to use ng-route with component pattern by being able to configure something like `$routeProvider.when('/', {resolve: {items: ($http) => $http.get(...)}, template: '<my-app items="resolve.items" />'});` and not having to configure a dummy controller that puts the resolved values on the scope.

### DIFF
--- a/test/ngRoute/directive/ngViewSpec.js
+++ b/test/ngRoute/directive/ngViewSpec.js
@@ -120,6 +120,47 @@ describe('ngView', function() {
   });
 
 
+  it('should reference resolved locals in scope', function() {
+    module(function($routeProvider) {
+      $routeProvider.when('/foo', {
+        resolve: {
+          name: function() {
+            return 'shahar';
+          }
+        },
+        template: '<div>{{$resolve.name}}</div>'
+      });
+    });
+
+    inject(function($location, $rootScope) {
+      $location.path('/foo');
+      $rootScope.$digest();
+      expect(element.text()).toEqual('shahar');
+    });
+  });
+
+
+  it('should allow to provide an alias for resolved locals using resolveAs', function() {
+    module(function($routeProvider) {
+      $routeProvider.when('/foo', {
+        resolveAs: 'myResolve',
+        resolve: {
+          name: function() {
+            return 'shahar';
+          }
+        },
+        template: '<div>{{myResolve.name}}</div>'
+      });
+    });
+
+    inject(function($location, $rootScope) {
+      $location.path('/foo');
+      $rootScope.$digest();
+      expect(element.text()).toEqual('shahar');
+    });
+  });
+
+
   it('should load content via xhr when route changes', function() {
     module(function($routeProvider) {
       $routeProvider.when('/foo', {templateUrl: 'myUrl1'});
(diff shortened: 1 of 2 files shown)


shahata:
In a nutshell I tried to write a new application with using only module.component and avoiding standalone controllers and ended up writing something like:
```js
$routeProvider.when('/', {
  template: '<my-app items="resolve.items" />',
  resolve: {items: ($http) => $http.get(…)},
  controllerAs: 'resolve',
  controller: function (items) { this.items = items; }
});
```
It is pretty ugly that I need to define the controller only so I can get the resolve results in the template...
With this PR I can do this instead:
```js
$routeProvider.when('/', {
  template: '<my-app items="resolve.items" />',
  resolve: {items: ($http) => $http.get(…)}
});
```

test/ngRoute/directive/ngViewSpec.js
```diff
@@ -120,6 +120,26 @@ describe('ngView', function() {
   });
 
 
+  it('should use default resolve controller when resolve map defined with no controller', function() {
```
gkalpak: The test description is kibd of undescriptive (I wouldn't have guessed what's supposed to do just by reading the description).

petebacondarwin:
How about a `resolvesAs` property to be consistent with `controllerAs` when attaching things to the scope?
Then `module.component` could default to `$resolves`...

shahata:
@petebacondarwin sure, that makes sense. I'll update the PR. Do you prefer resolvesAs over resolveAs?  For me the second sounds better, but maybe that's my broken English :)

petebacondarwin:
Right, so the property that we are aliasing is `resolve` right? So the property that specifies what this alias should be called is `resolveAs`.
You're right.

shahata:
Added `resolveAs`. @petebacondarwin do you think we should consider this change for 1.5? Using angular-route with components is very strange otherwise...

shahata:
oh, I read your original comment again. You mean that this functionality will work only if the route has `resolveAs` and that `module.component` will pass `resolveAs: '$resolve'`? I'm not sure that this is possible since `resolveAs` is a property of the route, not of the component...

petebacondarwin:
I didn't realise that this had nothing to do with `module.component`.

How about:

```
$routeProvider.when('/', {
  template: '<my-app items="$resolve.items" />',
  resolve: {items: ($http) => $http.get(…)}
  resolveAs: '$resolve'
});
```
Then that would mean that a default controller is created that contains all the resolves as properties.
This would not work if a controller was actually specified. In otherwords:

```
$routeProvider.when('/', {
  template: '<my-app items="$resolve.items" />',
  resolve: {items: ($http) => $http.get(…)}
  resolveAs: '$resolve',
  controller: function() { ... }
});
```

would throw an error...

petebacondarwin:
Actually looking at the actual code of the PR - I think that this is good as it is. Let's merge!

gkalpak:
There should be docs for this feature right ?

petebacondarwin:
Agh! Good point @gkalpak

PR #13402, opened 2015-11-28
docs(orderBy): updated controller example

zainengineer:
Updated example which manually injects the filter.
It matches sibling example in functionality.


Narretz:
Thanks, I tweaked it a bit and landed it. Although I wonder why you didn't simply copy the styles from the second example?

PR #13403, opened 2015-11-28
fix for $emit loop bug, when listeners were deregistered

### LINKED ISSUE #13401: the $emit loop is not aware which listener was removed
if $rootScope has 2 listeners attached for one event name, and the first registered listener removes itself, then the second listener will never be called.
this is because the $emit loop processes listeners in a for loop, and whilst processing the first listener, the namedListeners array will be spliced, thus the second listeners index will become 0 instead of 1.

kasparsj:
fixes #13401 the $emit loop is not aware which listener was removed

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -1175,6 +1175,7 @@ function $RootScopeProvider() {
       $emit: function(name, args) {
         var empty = [],
             namedListeners,
+            namedListenersCopy,
             scope = this,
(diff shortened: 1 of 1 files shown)


Narretz:
Hi, thanks for the PR.
The change has unfortunately caused a few tests to fail. And we also need a test for the expected behavior itself.

Narretz:

Issue #12969, opened 2015-09-28
ngMessages(ngAnimate 1.4.5?): Flash of content before animation begins.


topherfangio:
I just noticed a bug which I thought was in Angular Material, but I now believe is in Angular itself.

When the animation for my `ng-messages` begins, there is a flash where you can see all messages before the animation begins. Although the issue is appearing with ngMessages, I think it might be a broader `$animate` issue because I can reproduce this by upgrading only `angular.js` and `angular-animate.js` from 1.4.4 to 1.4.5 (leaving everything else including `angular-messages.js` at 1.4.4).

_Oddly, if I leave either file at 1.4.4 (it doesn't matter which one), the issue is not reproducible, so it seems to be an issue when both files are at version 1.4.5._

**Version Introduced:** 1.4.5?
**Versions Tested:** 1.4.4 (_works here; broken on others_), 1.4.5, 1.4.6, 1.5.0-beta.0
**Browsers:** All (IE Edge, Safari, Chrome/Firefox on both OS X and Windows)

**Steps to Reproduce:**

1. Open https://codepen.io/topherfangio/pen/PPbLaR
1. Click on the input and type a `5` so the new value is `12345`

You will see the errors appear for a split second, disappear, and then they will animate into place.

Again, you can edit the Codepen settings to change either file to 1.4.4 and it fixes the issue.

**Additional Info:**

Might be related to #12825.

I did not see any relevant breaking changes in the changelog between 1.4.4 and 1.4.6 which would account for this. Please let me know if I can provide any additional information.

ThomasBurleson:
@robertmesserle - is this related to any changes you may have made ?
@matsko  - what are your thoughts here ?

robertmesserle:
@ThomasBurleson If this is an AngularJS issue, then no - I have no commits in the angular.js repo.

matsko:
I see what's going on here. I'll have a look later today.

topherfangio:
I reproduced the demo without material just to make sure it wasn't some weird interaction with our CSS/JS. You can view it  below if it's helpful:

https://codepen.io/topherfangio/pen/zvoXjz

kvetis:
It also happens on ngIf in IE and Firefox, but not in Chrome. I've created a JSFiddle that illustrates the difference:

**Reproduction**
In this fiddle, delete the contents of the input box.
http://jsfiddle.net/kvetis/ssnkqog6/3/
[...]

PR #13408, opened 2015-11-29
Feat: ng-[event]-prepare class for structural animations (ngIf etc)

Narretz:
A fix / workaround for #12969

We currently don't have any place in the docs that documents the classes ngAnimate adds. At least I can't find it .:o I think it got lost during 1.4. We should add it back so people can find this feature easier.

Idea by @matsko

### DIFF
--- a/docs/content/guide/animations.ngdoc
+++ b/docs/content/guide/animations.ngdoc
@@ -274,6 +274,37 @@ myModule.directive('my-directive', ['$animate', function($animate) {
 }]);
 ```
 
+## Preventing flicker before an animation starts
+
+When nesting elements with structural animations such as `ngIf` into elements that have class-based
+animations such as `ngClass`, it sometimes happens that before the actual animation starts, there is a brief flicker or flash of content
+where the animated element is briefly visible.
+
+To prevent this, you can apply styles to the `ng-[event]-prepare` class, which is added as soon as an animation is initialized,
+but removed before the actual animation starts (after waiting for a $digest). This class is only added for *structural*
+animations (`enter`, `move`, and `leave`).
+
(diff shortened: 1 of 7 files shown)


Narretz:
@petebacondarwin This is a feature that is needed as a workaround for an ngAnimate issue that can't currently be solved in another way. Should we try to get it in before 1.5.0?

petebacondarwin:
Can we include links in the commit messages to the issues and other commits that this is fixing or related to?

If it a fix to a bug then it is fine to put it in 1.5.0.

It is not clear from the tests how this solve the issue in #12969 - we should have some test or at least a description of the issue and how this PR fixes that in the code and/or commit message. (Or am I just missing that?)

Narretz:
It's not a direct fix, more of a workaround.
Basically, you provide a style (e.g. opacity:0) for ng-enter-prepare that makes sure the element is hidden before the animation actually starts. We cannot directly test this, because the flicker does not always appear.

Narretz:
@petebacondarwin I've added an example to the issue https://github.com/angular/angular.js/issues/12969#issuecomment-160612044

petebacondarwin:
OK, so we also need docs for this feature, especially explaining how it can be used to fix problems that might appear like #12969.

Narretz:
I'll add some docs for it tomorrow

Narretz:
@petebacondarwin I've added docs, fancy taking a look?

petebacondarwin:
Will do. Later today

matsko:
LGTM

Narretz:
@matsko I've noticed that the prepare classes aren't applied to $animateCss based animations. I guess this expected, as they run outside of the animationQueue?

PR #13411, opened 2015-11-30
refactor: single implementation of isPromiseLike()

### LINKED ISSUE #13372: Multiple Different Implementations of isPromiseLike()
[promises-aplus-test-adapter](https://github.com/angular/angular.js/blob/d077966ff1ac18262f4615ff1a533db24d4432a7/lib/promises-aplus/promises-aplus-test-adapter.js) uses the following code to determine if an object is promise-like:

    var isPromiseLike = function isPromiseLike(obj) {return obj && isFunction(obj.then);};

[ngAnimate](https://github.com/angular/angular.js/blob/964a901bd827656cb5eee36d7a204f738d22647b/src/ngAnimate/shared.js) does this:

    var isPromiseLike = function(p) {
        return p && p.then ? true : false;
    };

and [Angular.js](https://github.com/angular/angular.js/blob/4daafd3dbe6a80d578f5a31df1bb99c77559543e/src/Angular.js) does this:

    function isPromiseLike(obj) {
        return obj && isFunction(obj.then);
    }

Perhaps they should all be changed to this:

    function isPromiseLike(obj) {
        return obj && typeof obj.then === 'function';
    }

lucastetreault:
There are multiple different implementations of isPromiseLike(). Change them all to use the one from Angular.js and use the stricter implementation from $q.

Closes #13372

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -674,7 +674,11 @@ function isBoolean(value) {
 
 
 function isPromiseLike(obj) {
-  return obj && isFunction(obj.then);
+  var then;
+  if (isObject(obj) || isFunction(obj)) {
+    then = obj && obj.then;
+  }
+  return isFunction(then) ? true : false;
 }
 
 
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -354,8 +354,8 @@ function qFactory(nextTick, exceptionHandler) {
       var that = this;
       var done = false;
       try {
-        if ((isObject(val) || isFunction(val))) then = val && val.then;
-        if (isFunction(then)) {
+        if(isPromiseLike(val)) {
+          then = val && val.then
           this.promise.$$state.status = -1;
           then.call(val, resolvePromise, rejectPromise, simpleBind(this, this.notify));
         } else {
--- a/lib/promises-aplus/promises-aplus-test-adapter.js
+++ b/lib/promises-aplus/promises-aplus-test-adapter.js
@@ -2,7 +2,6 @@
 'use strict';
 
 var isFunction = function isFunction(value){return typeof value == 'function';};
-var isPromiseLike = function isPromiseLike(obj) {return obj && isFunction(obj.then);};
 var isObject = function isObject(value){return value != null && typeof value === 'object';};
 var isUndefined = function isUndefined(value) {return typeof value === 'undefined';};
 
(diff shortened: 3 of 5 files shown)


src/Angular.js
```diff
@@ -674,7 +674,11 @@ function isBoolean(value) {
 
 
 function isPromiseLike(obj) {
-  return obj && isFunction(obj.then);
+  var then;
```
dcherman: This whole thing can be simplified to

```js
return (isObject(obj) || isFunction(obj)) && isFunction(obj.then);
```

src/ng/q.js
```diff
@@ -354,8 +354,8 @@ function qFactory(nextTick, exceptionHandler) {
       var that = this;
       var done = false;
       try {
-        if ((isObject(val) || isFunction(val))) then = val && val.then;
-        if (isFunction(then)) {
+        if(isPromiseLike(val)) {
+          then = val && val.then
```
dcherman: Why store a reference to `then` from `val` only to do `then.call(val, ...)` a couple lines below?  Seems like you should just be able to do `val.then(...args)` and avoid the `.call` requirement here.  It doesn't look like `then` is used anywhere but there.

jbedard: I think fetching `then` must only be done once. So fetching it inside `isPromiseLike` and again here already breaks things: https://github.com/promises-aplus/promises-spec#notes

dcherman: To avoid it being mutated in-between calls is what their point is, but that cannot reasonably happen here since this happens inside of a single function call and JS is single threaded.

The only way this would even happen is like if they had a getter that returned a different function when the property was accessed which is....nuts.

dcherman: Interestingly enough, I just re-read the spec notes and an accessor mutating the property is what they're trying to guard against, but that seems insane :speak_no_evil: 

In either case, changing this would result in the same number of `.then` property accesses but would be simpler.  The most correct thing would apparently to change `isPromiseLike` to return the `.then` property which you'd then `.call` with `val` here.  That'd be the most spec correct thing, but it just seems unnecessary.

jbedard: Yeah it's super specific, although that does force implementations to be consistent which might be worth it. The only way to do this would be for `isPromiseLIke` to return the `.then` like you mentioned...

PR #13412, opened 2015-11-30
ngResource collection $new method (ActiveRecord style)

AlexVangelov:
Why:
When a collection result is empty, there is no reference to the originated resource object.
Benefits:
Compatible restful models for new and existing resources in resolve blocks.


AlexVangelov:
I signed it!

gkalpak:
@AlexVangelov, what is the usecase for this ? Why can't you just use the resource class ?

AlexVangelov:
@gkalpak 
When a collection is provided by resolve block
and your code is injecting the resolved data value,
if there is at least 1 item you can create a new instance from it, 
but if the collection size is zero, you have to inject the class to create a new instance.
From architecture point of view, you don't have to inject both resolved value and originated class,
since the resolved value already knows where it comes from (unless it's empty).
Borrowed from http://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-build

gkalpak:
> if there is at least 1 item you can create a new instance from it

How can you do that ?

> From architecture point of view, you don't have to inject both resolved value and originated class, since the resolved value already knows where it comes from (unless it's empty)

Well, if the resolved value is just an Array, then it shouldn't know where is came from. If anything, the instances (e.g. array items) know where they came from.
From an architectural point of view, (imo at least) you having an instance of something that you don't know what it is and trying to creat an new instance of the same kind, is not a good idea. Consuming an array of something and creating something are too distinct and decoupled operations. So, if you want to create a new instance of something, it seems reasonable to me that you have to request access (via injection) to its resource class.

The linked example is quite different. You have an association from `Person` to `Pet` and there is a method on `Person` for creating `Pet`s. It's not just a random collection of `Pet`s that knows how to create more `Pet`s. It's the `Person` that know how to do it and that is because a `Person` knows about its associated class, `Pet`.
If you are trying to implement something similar with `$resource`, then class `A` which `has_many` instances of class `B` will need a method for creating `B`s (and that could be implemented through the `B` resource class, because `A` knows exactly what class it is that its associated with).

--
(I'm not sure that was clear enough, but that's more or less how I see it :smiley:)

tl;dr
I think that adding this feature will promote some bad architectural designs and won't offer much benefot over what is already possible today.

Just my 2 cents...

AlexVangelov:
Ok, negotiation :)

> How can you do that ?

You can create a new instance through the constructor of any instance.
I want to provide a reference to this constructor.

What You think about something like:
`value.$constructor = this;`
It is the same as `resolvedCollection[0].constructor`
And this way we can keep it in both cases (isArray: true/false)
Is it acceptable?

You can close the issue, but I want to continue the discussion.

gkalpak:
TBH, I'm not sure how valuable it is to be able to construct something that you don't know what it is. Especially, since that "something", is typically an object with some data and you won't know what data (i.e. properties) it accepts/understands.

Could you share a concrete example of a usecase ? Maybe that'd help convince me :stuck_out_tongue: 

--
In the meantime, I'm moving this to the Ice Box, but will leave it open so others can chime in.

abelinoaleman:
o

AlexVangelov:
It is very valuable! :)
Currently there are 2 reasons to define a resource model and name it.

First is to add some logic inside (fat model thin controller).
Second - to render corresponding html template.

One of the biggest problem working with restful APIs and Angular is that you have to duplicate the server model on the front and keep it in sync. It is easy to build the interface for listing, modifying and adding pets, but ngResource is not ready for that, because my pull request is not merged :)
 
Lets say we have REST API resources cats (/server/cats) and dogs (/server/dogs)
```
templateUrl: url
resolve:
  pets: $resource(url).query().$promise
```
The server is smart enough to realize that the request content-type _application/html_ is for template and _application/json_ is for the data (sure you can still use template cache with the risk to go out of sync).
Our template can use all the power of angular filters to validate the resource without any model code.
But Inside the interface scope we need a methods to modify existing and create new resources.
Here we come... `pets[n]` and `pets.$new` are ready to go instances of $resource.

gkalpak:
It sounds more like `fat model, "fat template", thin controller` ("fat template" as in _having more logic than belongs there_). I mean, your template knows about a `Pet` (e.g. its internal structure), but your controller doesn't ? It doesn't feel right to me.

If you want to keep your logic in your configuration, you can do something like:

```js
var Pet = $resource(url);
...
.when('...', {
  ...
  templateUrl: url
  resolve: {
    Pet: function () { return Pet; },
    pets: function () { return Pet.query().$promise; }
  }
}
```

AlexVangelov:
This reminds me of a joke:
When a developer prepares to sleep, he took two bottles for water - one full and one empty - in case he wants to drink and in case he won't.

gkalpak:
FWIW, you didn't have to close the issue just because I didn't find the feature useful.
There might be different opinions... Just saying :smiley:

AlexVangelov:
Actually I have some concerns about my implementation.
The reference should be available no matter what isArray flag is. 
I appreciate the conversation with you. Following :)

PR #13414, opened 2015-11-30
fix(ngAnimate): don't normalize classes on follow-up animations to jo…

### LINKED ISSUE #13380: ng-show sometimes does not remove ng-hide class after expression becames truthy
I've noticed that in my application ng-show sometimes leaves ng-hide class even after ng-show expression evaluates to true. 

This behavior is not consistent across browsers: In places where I see issues  on Chrome it looks fine on FF, while in other place the situation is opposite.

Even if I retry the same scenario on a single browser, the defect does not occur every time, so this must be some non-deterministic timing issue.

### Scenario & recreation
I've tried to isolate the problem and stripped my app to this simple plunker:
http://plnkr.co/edit/pNUXB6A0gCyfdej8k5Xu?p=preview
I am able to recreate the issue there with my FF ESR 38.4.0, but not every time. 
I'm clicking the button and if the modal appears with both lines of text I need to click Stop/Run in Plunker and try again, for several times. My teammate who is using exactly the same FF version cannot recreate it at all.

### Some more observations:
* Manually changing scope.loadedData (the watched expression there) to a falsy value and then back to the original value causes the div to show up.
* Calling scope.$digest() manually does not help
* replacing ng-show with additional condition in ngClass ('ng-hide': !loadedData) does not help either
* using ng-if instead of ng-show works fine
* debugging reveals that ng-show directive's watch is triggered properly. It calls $animate to remove ng-hide class, but it doesn't take effect.
* ng-show alone works fine, the issue occurs on the element which has both ng-show and ng-class directives

### Affected versions
This is a regression introduced in Angular 1.4.0 (recreated this defect on 1.4.0, 1.4.7, 1.4.8 and 1.5.0-beta.2, but It did not occur on 1.3.20)

Narretz:
…ined animations

This allows follow-up animations to remove a class that is currently
being added in a running animation. 
This previously didn't work if the running animation was a joined animation itself.

Fixes #13380

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -360,19 +360,15 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
           // so an example would involve a leave animation taking over an enter. Then when
           // the postDigest kicks in the enter will be ignored.
           var joinAnimationFlag = isAllowed('join', element, newAnimation, existingAnimation);
-          if (joinAnimationFlag) {
-            if (existingAnimation.state === RUNNING_STATE) {
-              normalizeAnimationOptions(element, options);
-            } else {
-              applyGeneratedPreparationClasses(element, isStructural ? event : null, options);
+          if (joinAnimationFlag && existingAnimation.state < RUNNING_STATE) {
+            applyGeneratedPreparationClasses(element, isStructural ? event : null, options);
 
-              event = newAnimation.event = existingAnimation.event;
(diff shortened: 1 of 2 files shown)


matsko:
So I took an investigation and it looks like the cancellation behaviour for addClass/removeClass doesn't work correctly. Here is a fix with your test spec and the new cancellation code included: https://github.com/angular/angular.js/pull/13472.

cc @petebacondarwin please review.

Narretz:
Superseded by https://github.com/angular/angular.js/pull/13472

PR #13419, opened 2015-12-01
using svg, allow svg attrs also, not only elements

Vnthf:


petebacondarwin:
@Vnthf thanks for this PR. Can you provide an explanation of what is not possible right now with the sanitizer? Then can you ensure that this would not expose a security hole; I expect that such attributes would have been made available already if they would not cause a problem. Finally the PR would need at least one new unit test before it could be merged.

Vnthf:
@petebacondarwin  

If , despite the svgEnabled ( such as graphs) permitted to use svg , and only added to svgElements in validElements svg is not rendered properly.  it will make just svg tag. the contents of the tag is empty because sanitizer is not allowed attr related svg. So attr element must also be allowed , as well as to allow svg. I would like to create unit tests soon.

petebacondarwin:
@Vnthf `validAttrs` already contains the `svgAttrs`. See https://github.com/angular/angular.js/blob/master/src/ngSanitize/sanitize.js#L283-L286

PR #13420, opened 2015-12-01
style($compile,$controller): adding function names for debug/tracing

jbedard:
I find this very useful when profiling or looking at heap dumps. Normally browsers figure out the `var` name in stack traces but when looking at a heap dump things like `var Attributes = function() { .. }` or `return function() { ... }` all show up as "anonymous"...


gkalpak:
LGTM

PR #13422, opened 2015-12-01
fix($compile): allow compile data to be GCed after a non-clone linking

jbedard:
This does two things:
- nulls out some data when a non-cloning ~~transclude~~ link completes to allow mainly the ~~`boundTranscludeFn` func/closure~~ `$compileNodes`/`compositeLinkFn` to be GCed
- throws if that ~~transclude~~ link is executed again since that would either re-link the same element or clone the linked element

### DIFF
--- a/docs/content/error/$compile/multilink.ngdoc
+++ b/docs/content/error/$compile/multilink.ngdoc
@@ -1,0 +1,27 @@
+@ngdoc error
+@name $compile:multilink
+@fullName Linking Element Multiple Times
+@description
+
+This error occurs when a single element is linked more then once.
+
+For example, if an element is compiled and linked twice without cloning:
+```
+  var linker = $compile(template);
+  linker($scope); //=> ok
+  linker($scope); //=> multilink error
+```
+
+Linking an element as a clone multiple times is ok:
+```
+  var linker = $compile(template);
+  linker($scope, function() { ... });     //=> ok
+  linker($scope, function() { ... });     //=> ok
+```
+
+However once an element has been linked it can not be re-linked as a clone:
+```
+  var linker = $compile(template);
+  linker($scope);                       //=> ok
+  linker($scope, function() { ... });   //=> multilink error
+```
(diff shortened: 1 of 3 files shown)


Narretz:
I believe this has the potential to break some people's apps, simply because it's not obvious how you are supposed to use $transclude. But it's good for 1.6.x, with a BC notice.

jbedard:
Anything that breaks was probably already broken, although this might make it more broken (by throwing).

Which makes me think this should also be done on link functions, not only on the transclude functions...

jbedard:
I've added a second commit to do this to all link functions. Since transclude actually just delegates to the linker the second commit actually covers both cases. Is it worth having a transclude specific error message? Or allowing a couple more transclude closures to be GCed in addition to the link ones? I'd so no to the error message, yes to the closures...

jbedard:
The transclusion slot stuff made this a bit more complicated so I've simplified it to only null out the link function data (not transclusion function data - which really just wraps the link function data). This is probably a better first step anyway...

jbedard:
Any thoughts on this for 1.6?

Narretz:
I think we can try it for 1.6. We should probably keep a list of possibly problematic commits for when we will snyc 1.6 to g3 in the future.
Before we merge it, we need an error doc for the multilink error with description and examples what can cause this.

jbedard:
I've added an error doc.

jbedard:
1.6? :grin:

Narretz:
Even though it's unlikely that this is going to break apps, I feel like we should mark this as "fix", just so that it shows up in the changelog, and when people have their app broken, they might see what's causing it.

jbedard:
I've updated the commit message

Narretz:
Finally landed! Thanks for your preserverance jbedard. Due to how easy it is to make this mistake, I'll leave it in 1.6 only (for now). Especially since with some simple directives (such as interpolation), you won't notice any difference in behavior when linking multiple times.

jbedard:
I think just 1.6 is best unless there's a strong reason to put it into 1.5. This can be considered a BC for those using `$compile` incorrectly...

PR #13424, opened 2015-12-01
fix(ngAnimate): ignore children without animation runner when closing…

### LINKED ISSUE #11992: Animate crash: Cannot read property 'runner' of undefined in 1.4.0
```
TypeError: Cannot read property 'runner' of undefined
    at angular-animate.js:2330
    at forEach (angular.js:336)
    at closeChildAnimations (angular-animate.js:2325)
    at queueAnimation (angular-animate.js:2148)
    at Object.$$AnimateQueueProvider.$get.push (angular-animate.js:2040)
    at Object.$AnimateProvider.$get.leave (angular.js:5200)
    at cleanupLastView (angular-route.js:913)
    at angular-route.js:942
    at publicLinkFn (angular.js:7387)
    at $get.boundTranscludeFn (angular.js:7533)`
```

I get it when using it with the following CSS:
```
.hide-fade.ng-hide-add, .hide-fade.ng-hide-remove {
  -webkit-transition:all linear 0.3s;
  -moz-transition:all linear 0.3s;
  -o-transition:all linear 0.3s;
  transition:all linear 0.3s;
  display:block!important;
}
.hide-fade.ng-hide-add.ng-hide-add-active,
.hide-fade.ng-hide-remove {
  opacity:0;
}
.hide-fade.ng-hide-add,
.hide-fade.ng-hide-remove.ng-hide-remove-active {
  opacity:1;
}
```

Does not happen while using 1.3.15

### LINKED ISSUE #13423: Directive: working with html fragment in documentFragment instead of direct modifying DOM
I am very disappointed with the behavior of the angular when it works in a DOM tree.
It inserts and parses directives directly in the HTML, but does not use documentFragment.
Why it can't make something like this?

Directive => inserting innerHTML into documentFragment => doing usual things with it => finding inner directives => parse them => repeat, until all directives into this fragment will be resolved => insert ready-to-use fragment into real tree.

I think, that must be a flag, because some peoples uses templateUrl without caching in production, so there will be a white screen for a log time.

Thanks!

Narretz:
… them

During parent structural animations, ongoing animations on child elements
are closed. These children are looked up by checking the data-ng-animate
attribute. If an element is the clone of an animating element,
it might have this attribute, but no animation runner associated with it,
so we need to ignore it.

Fixes #11992
Closes #13423

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -512,15 +512,15 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       forEach(children, function(child) {
         var state = parseInt(child.getAttribute(NG_ANIMATE_ATTR_NAME));
         var animationDetails = activeAnimationsLookup.get(child);
-        switch (state) {
-          case RUNNING_STATE:
-            animationDetails.runner.end();
-            /* falls through */
-          case PRE_DIGEST_STATE:
-            if (animationDetails) {
+        if (animationDetails) {
+          switch (state) {
+            case RUNNING_STATE:
+              animationDetails.runner.end();
+              /* falls through */
+            case PRE_DIGEST_STATE:
               activeAnimationsLookup.remove(child);
-            }
-            break;
+              break;
+          }
         }
       });
     }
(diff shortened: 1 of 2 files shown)


matsko:
LGTM

PR #13427, opened 2015-12-02
refactor($compile): move setup/get controller methods out of the compile node closure

jbedard:
This moves two methods out of the per-element compile closure.

`getControllers` has no changes

`setupControllers` required the  `newIsolateScopeDirective` param and the use of `hasElementTranscludeDirective` was removed (.data on a comment is a noop).  This can be a little simpler after something such as #13421

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2423,15 +2423,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
           var controllerInstance = $controller(controller, locals, true, directive.controllerAs);
 
-          // For directives with element transclusion the element is a comment,
-          // but jQuery .data doesn't support attaching data to comment nodes as it's hard to
-          // clean up (http://bugs.jquery.com/ticket/8335).
+          // For directives with element transclusion the element is a comment.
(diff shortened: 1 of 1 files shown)


jbedard:
I've split this into two commits so the second only moves the functions and does not change the function bodies.

src/ng/compile.js
```diff
@@ -2285,6 +2214,73 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+
+        var controller = directive.controller;
+        if (controller == '@') {
+          controller = attrs[directive.name];
+        }
+
+        var controllerInstance = $controller(controller, locals, true, directive.controllerAs);
+
+        // For directives with element transclusion the element is a comment.
+        // In this case jQuery .data will not attach any data.
```
gkalpak: This sounds like jqLite **will** attach the data (which is won't afaict).

gkalpak:
LGTM
(This probably needs some refactoring after the recent changes in 3ffdf38, cd21216, 56c3666.)

jbedard:
Rebased and fixed the comment

PR #13428, opened 2015-12-02
Capitalize the name 'JSHint' in the comments

henryw4k:
When referencing JSHint, the first three letters are capitalized. 

https://github.com/jshint/jshint

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1,6 +1,6 @@
 'use strict';
 
-/* We need to tell jshint what variables are being exported */
+/* We need to tell JSHint what variables are being exported */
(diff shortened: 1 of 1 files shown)


src/Angular.js
```diff
@@ -146,13 +146,13 @@ var uppercase = function(string) {return isString(string) ? string.toUpperCase()
 
 
 var manualLowercase = function(s) {
-  /* jshint bitwise: false */
+  /* JSHint bitwise: false */
```
lgalfaso: This capitalization of `JSHint` is not recognized and the code below generates a lint error. Same below

PR #13429, opened 2015-12-02
docs(ngModel.NgModelController): Prevent model overwrite

uglu:
If there are default values in the model, they get overwritten at runtime because the read() function is initiating with a blank element and is assigning that to the model.

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -164,7 +164,7 @@ is set to `true`. The parse error is stored in `ngModel.$error.parse`.
 
               // Specify how UI should be updated
               ngModel.$render = function() {
-                element.html($sce.getTrustedHtml(ngModel.$viewValue || ''));
+                element.html($sce.getTrustedHtml(ngModel.$viewValue || element.html()));
                 read(); // initialize
               };
 
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -165,13 +165,13 @@ is set to `true`. The parse error is stored in `ngModel.$error.parse`.
               // Specify how UI should be updated
               ngModel.$render = function() {
                 element.html($sce.getTrustedHtml(ngModel.$viewValue || ''));
+                read(); // initialize
               };
 
               // Listen for change events to enable binding
               element.on('blur keyup change', function() {
                 scope.$evalAsync(read);
               });
-              read(); // initialize
 
               // Write data to the model
               function read() {


uglu:
Signed

Narretz:
Can you give an example / demo of how this happens? If there's data on the scope, it should always be propagated correctly to the DOM. If there's data in DOM, then I can see errors coming up, but that's also not something angular is expecting.

uglu:
Sorry for the tardy response, i wasn't around the last few days.

So the thing that happens is that if the [contenteditable] element is empty or has some content, that content will replace whatever value the model already has defined.

So if i have `$scope.test='something'` and the view has something like `<div cotntenteditable ng-model="test"></div>`, `$scope.test` will become an empty string instead of filling the `div` with `'something'`

gkalpak:
I don't think the `read()` call needs to be moved inside `$render()` (but maybe the inital call to it needs to be removed - it depends on what we want to achieve).

This boils down to what behaviour we want the example directive to have. Typically, in an Angular app, the data is propagated to the view from the model (and not the other way around). In that case, we should remove the initial call to `read()`, remove the content of `div[contenteditable]` and populate it using `$scope.userContent = 'Change me!'`. (E.g. see this [demo pen][1].)

If the directive is supposed to initialize it's model based on the div's content in the HTML, then we can keep the initial `read()` call. In that case, this special behaviour will be supposed to be a feature of the example directive (and should probably be documented - e.g. add a comment in the code right abive the initial `read()` call).

[1]: https://codepen.io/ExpertSystem/pen/GogoRe?editors=001

Narretz:
Imo, the model should set the value on intialization.

uglu:
Exactly, which wasn't happening. The model was being overwritten by the html.
My latest commit checks first if there is value in the model, if not, adds whatever html there is in the element.

gkalpak:
I would prefer not mixing the two approaches. For this example, I am fine with both (either reading the value from the DOM or setting it from the model), but I suggest not mixing both, as this makes data propagation less straight-forward.

uglu:
Agreed, but i'm not sure how to approach this separation.

gkalpak:
@uglu, I am not sure if I was clear: What I meant is that I would have the directive do one thing (get initial data from the model) **OR** the other (get initial data from the HTML), but not both.

FWIW, here is what I had in mind (in order to imlement the former):

1. Remove the initial call to `read()` (it's not needed any more).
2. Remove the `Change me!` text content from `div[contenteditable]` (since it will b eoverwritten by the model).
3. Initialize `$scope.userContent` to `Change me!` (you can use a controller).

gkalpak:
Thinking about it again. For this simple `[contenteditable]` directive, it makes more sense to read the content off the HTML. Taking that into account and the fact that we should keep things as simple as possible in these examples, I believe it is better to keep things as is.

PR #13430, opened 2015-12-02
docs(guide/Expressions): fixes about void and new operator usage

bhovhannes:
1. `void` operator works in angular expressions in all 1.x versions of Angular. For example: expression {{void(2+3)+'x'}} is being interpolated into 'x' without any errors. Documentation incorrectly states that this is impossible.
2. You cannot create new objects inside Angular expressions. For example: {{ new Date() }} expression fails.

### DIFF
--- a/docs/content/guide/expression.ngdoc
+++ b/docs/content/guide/expression.ngdoc
@@ -34,10 +34,10 @@ Angular expressions are like JavaScript expressions with the following differenc
 
   * **No RegExp Creation With Literal Notation:** You cannot create regular expressions
     in an Angular expression.
-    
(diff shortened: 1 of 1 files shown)


Narretz:
Careful .. void only looks like it works, but it actually "works" because angular expression ignore missing functions / properties. If you defined a "void" fn on the scope, it will be called with 2+3.

The other change is good, though, so I can merge it after you've changed it.

bhovhannes:
Thank you for your reply @Narretz, I've put back deleted part regarding `void` operator.

Narretz:
Landed as https://github.com/angular/angular.js/commit/6752337629e813dc499f23cc0bc96dbb588aed44, thanks!

PR #13431, opened 2015-12-02
feat(ngTransclude & $compile): fix treatment of unfilled optional slots

### LINKED ISSUE #13426: How do you tell if an optional multi transclude property has been set?
Basically I can not find a way to check if an optional transclude was supplied, I want to show hide the parent based on the transclude being set or not.

If there is no way is it possible to get this added? It is a useful feature to along side optional transclude.

Thanks

petebacondarwin:
Closes #13426

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2152,7 +2152,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           transcludeFn = controllersBoundTransclude;
           transcludeFn.$$boundTransclude = boundTranscludeFn;
           // expose the slots on the `$transclude` function
-          transcludeFn.$slots = boundTranscludeFn.$$slots;
+          transcludeFn.isSlotFilled = function(slotName) {
+            return !!boundTranscludeFn.$$slots[slotName];
+          };
         }
 
         if (controllerDirectives) {
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -237,7 +237,8 @@
  *          as those elements need to created and cloned in a special way when they are defined outside their
  *          usual containers (e.g. like `<svg>`).
  *        * See also the `directive.templateNamespace` property.
- *
+ *    The `$transclude` function has a property called `$slots`, which is a hash of slot names to slot transclusion
+ *    functions. If a slot was declared but not filled its value on the `$slots` object will be `null`.
  *
  * #### `require`
  * Require another directive and inject its controller as the fourth argument to the linking function. The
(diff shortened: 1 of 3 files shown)


petebacondarwin:
@PascalPrecht - hey look at this!

PascalPrecht:
@petebacondarwin awesome! Does this also keep the original contents of the DOM where we transclude into?

E.g.
```
<span ng-transclude="fooSlot">This is default DOM and going to be replaced if `fooSlot` is defined, if not this stays here</span>
```

petebacondarwin:
Yep! That is the second commit in the PR.

PascalPrecht:
Argh. I just've just read the commit messages instead of reading the code haha. Awesome! Thanks for the work!

petebacondarwin:
See https://ci.angularjs.org/job/angular.js-pete/989/artifact/build/docs/api/ng/directive/ngTransclude#multi-slot-transclusion
for a running example

src/ng/compile.js
```diff
@@ -2222,13 +2235,15 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
           if (slotName) {
             var slotTranscludeFn = boundTranscludeFn.$$slots[slotName];
-            if (!slotTranscludeFn) {
+            if (slotTranscludeFn) {
+              return slotTranscludeFn(scope, cloneAttachFn, transcludeControllers, futureParentElement, scopeToChild);
+            }
+            if (isUndefined(slotTranscludeFn)) {
```
Narretz: how about just `else`?

Narretz:
`feat($compile): expose transclusion `$$slots` on the `$transclude`
Isn't the commit message a bit contradictory? It's called a feature but `$$slots` indicates a private API.

And a question, is it possible to keep the content with the default behavior (no slots, transclude is true or content)? Since people have requested something similar for the default ngTransclude: https://github.com/angular/angular.js/pull/11839
I think we could extend the 'content' value to make it optional by adding `?` and applying the same behavior.

src/ng/compile.js
petebacondarwin: We have to differentiate between `null` and `undefined`. I will add a comment to clarify.

petebacondarwin:
We could make `$$slots` -> `$slots`.

petebacondarwin:
Regarding default content for the default transclusion, perhaps we could special case `""` in the transclude map:

```js
transclude: {
  'required': 'requiredSlot',
  'optional': '?optionalSlot',
  '': '?' // this would make the default transclusion optional too...
}
```

src/ng/compile.js
Narretz: Okay, I just noticed there's no difference in test behavior.

Narretz:
Then we should also document `$slots`, I guess.

For default transchlusion. Okay, I forgot you can have default transclusion and slots. The proposed API for that looks a bit silly, but it's probably okay. But we can also decide on that after 1.5

petebacondarwin:
Hmm it is a bit awkward to make the default transclusion optional. I think we should punt that feature to post-1.5.0

Narretz:
Looks like the last commit caused some doc e2e tests to fail. :o Other than that, feel free to merge. We can tweak it after the fact.

PR #13433, opened 2015-12-03
feat(ngMock): destroy $rootScope after each test

andyrooger:
Previously $rootScope would be new for each test, but old $rootScopes would never be destroyed.
Now that we are able to destroy the $rootScope, doing so provides an opportunity for code to clean
up things like long-lived event handlers between tests.


petebacondarwin:
Looks good - thanks @andyrooger

andyrooger:
Brilliant. Cheers @petebacondarwin

gkalpak:
Backported to `v1.4.x` as e55829a (as discussed with @petebacondarwin).

PR #13445, opened 2015-12-03
docs($compile): Remove deprecation notice on replace

kentcdodds:
The fact that replace is deprecated in Angular 2 basically means nothing. Pretty much everything's deprecated in Angular 2. This notice discourages people from using features of Angular 1 which could actually serve their purposes best just because they think that it'll get removed from Angular 1.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -289,7 +289,7 @@
  * #### `template`
  * HTML markup that may:
  * * Replace the contents of the directive's element (default).
- * * Replace the directive's element itself (if `replace` is true - DEPRECATED).
+ * * Replace the directive's element itself (if `replace` is true).
  * * Wrap the contents of the directive's element (if `transclude` is true).
  *
  * Value may be:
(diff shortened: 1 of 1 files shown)


Narretz:
Hm.
I've always understood that it's already deprecated. You shouldn't have been using it since what, 1.3?. There are bugs with replace that probably will never be fixed.

kentcdodds:
Hmmm... In that case, it shouldn't even be in the docs...

Narretz:
Why shouldn't it be in the docs? 
Angular 1 has never followed semver that closely. Our minor versions always have nreaking changes, for better of for worse. That's probably why we have deprecations, too.

kentcdodds:
Hmmm... Then I think it'd make more sense if it weren't labeled as being deprecated in 2.0 but that it was deprecated in 1.3... But whatever :-)

Narretz:
I will try to improve the message ... I agree that it's confusing. But deprecation version and removal version isn't the same, I think.

PR #13447, opened 2015-12-04
fix(ngClick): subract processing time of the click action

Tr4v1s:
subract the time it takes for the browser to process the click action for preventing the ghost click. long running processes or slower devices where a lot of rendering is happening would get the ghost click every time.

### DIFF
--- a/src/ngTouch/directive/ngClick.js
+++ b/src/ngTouch/directive/ngClick.js
@@ -57,6 +57,7 @@ ngTouch.directive('ngClick', ['$parse', '$timeout', '$rootElement',
 
   var ACTIVE_CLASS_NAME = 'ng-click-active';
   var lastPreventedTime;
+  var processingTime;
   var touchCoordinates;
   var lastLabelClickCoordinates;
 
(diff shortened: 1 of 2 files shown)


gkalpak:
I feel like this way of calculating the processing time is not reliable, since it depends on factors that vary over time.

E.g. due to the asynchronicity of `$timeout` and `click` events, the current processing load will determine the processing time for a future click event (by which time the processing time may be totally different).

Tr4v1s:
I'm not sure I understand what you mean by "the current processing load will determine the processing time for a future click event". It should calculate independently for each tap, it doesn't store the time and reuse it for future events.

I added a test to show what I mean for the long running process.

Thanks for taking the time to look though :)

gkalpak:
`calculateProcessingTime()` is called upon `touchend`, but `processingTime` (which is used upon `click`) is calculated using a `$timeout` (i.e. asynchronously). There is no guarantee (afaict) that the timeout will have been triggered _before_ the `click` event has been dispatched on the element.

Do I miss something ?

Tr4v1s:
I don't think you are missing anything...

It is true that there is no guarantee(because it depends on the browser for how timeouts are queued up), but in theory it should be added to the timeout queue to be executed before the click(which is something like 300 milliseconds). In the case that the browser executes click without respecting the timeout queue, it has a processing time of 0(which is how it worked before).

I'm really not sure which way to go with this... because timeout and events are so dependent on the browser. I just know that this is an issue that I have run into(only when doing things inefficiently though).

Sorry for the long delay. Holidays...

PR #13449, opened 2015-12-04
docs(ngDisabled): describe your change...

arthur2001:
There is a typo in the doc. It should be 'ng-disabled' and not just 'disabled'.


Narretz:
This is intended, the example is used to show how not to do it. It looks like this is confusing many people, I will open an issue for cleaning this up.

PR #13453, opened 2015-12-05
fix($sanitize): blacklist SVG `<use>` elements

petebacondarwin:
The use element can reference external svg's (same origin) and can include
xlink javascript urls or foreign object that can execute xss.

This change disallows `<use>` elements in sanitized SVG markup.

An example of a malicious SVG document would be:

SVG to sanitize:

```
<svg><use xlink:href="test.svg#xss" /></svg>
```

External SVG file (test.svg)

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg" width="100"
   height="100"
   id="xss">
<a xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="javascript:alert(1)">
  <circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" />
</a>
</svg>
```

Here the SVG to sanitize loads in the `test.svg` file via the `<use>` element.
The sanitizer is not able to parse this file, which contains malicious
executable mark-up.

This can only be taken advantage of if the external file is available via the
same origin restrictions in place.

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -242,7 +242,7 @@ var inlineElements = angular.extend({}, optionalEndTagInlineElements, toMap("a,a
 // They can potentially allow for arbitrary javascript to be executed. See #11290
 var svgElements = toMap("circle,defs,desc,ellipse,font-face,font-face-name,font-face-src,g,glyph," +
         "hkern,image,linearGradient,line,marker,metadata,missing-glyph,mpath,path,polygon,polyline," +
-        "radialGradient,rect,stop,svg,switch,text,title,tspan,use");
+        "radialGradient,rect,stop,svg,switch,text,title,tspan");
 
 // Blocked Elements (will be stripped)
 var blockedElements = toMap("script,style");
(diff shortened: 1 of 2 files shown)


lgalfaso:
LGTM

petebacondarwin:
Easy one :-) Does it need a breaking change notice?

IgorMinar:
Add breaking change note please. Otherwise lgtm

Coridyn:
We use the `<use>` element to reference SVG content defined on the page (e.g. as an alternative to Glyphicons):

```HTML
<a href="#"><svg class="icon"><use xlink:href="#icon-call"></use></svg> Give us a call</a>
```

Would it be possible to opt in to `<use>` elements or share the `$compileProvider.imgSrcSanitizationWhitelist` when santizing this sort of content?

Alternatively, are you aware of any workaround we could use without having to use `trustAsHtml()`?

petebacondarwin:
@Coridyn perhaps you could come up with a custom directive so you are not placing svg directly in the bound HTML content?

Coridyn:
The SVG is coming from user-editable content so I don't think that's possible.

Does running the content through `$compile` has the same security implications as using `trustAsHtml()`?

Issue #13247, opened 2015-11-04
Idea $this in directive locals calls


jpolo:
It would be nice to have a normalized reference to the locals for directive expression.
I've proposed $this but other candidates could be better ($locals, etc...)

Example 
```javascript
angular.directive("fooBar", function () {

   return {
     scope: {
       ngCallback:"&"
     },
     compile: function ($scope) {
       var ctx = {
       foo: "bar"
       };
       ctx.$this = ctx;// Keep a reference to the locals

       $scope.ngCallback(ctx);
     }
   };
})
```
Allows to easily dump available locals
```html
<foo-bar ng-callback="log($this)">
```

Narretz:
Can you give a more complete example? I've a hard time figuring out the use case.

jpolo:
The problem when passing just locals to the context is that there is no way to inspect all locals at once.
If ```log``` l is just a simple forward to ```console.log``` then ```log($this)``` would dump something like ```javascript
{
  foo: "bar"
}
```
That helps a lot when using debugging/using light documented directives.

**Comment by petebacondarwin** [comment]
I see what you mean. You want to have access to a single property on the scope that references the `locals` that were passed to through to the expression.
I.E. when we execute an expression we can provide some locals that are merged with the scope:

```js
var someExpressionFn = $parse('someExpression');
someExpressionFn($scope, locals);
```

You would like us to automatically attach an additional property to the `locals` object that refers to the `locals` object itself.

jpolo:
Exactly sir! Sorry if I wasn't clear before...

petebacondarwin:
I see this as a rare requirement that can be easily implemented on a case by case basis as you show in your example @jpolo. I don't think it is something that we would want to add to all cases, especially as we would have to pollute the scope with yet another special property.

petebacondarwin:
In general, if I have a load of things that I want to pass via locals that can be handled as a single thing by the client then I would set it up so:

```js
$scope.ngCallback({ $event: ctx });
```

This (interestingly) is similar to the form that seems to be appearing in Angular 2 event handling.

jpolo:
In my user need, I am the developer that _uses_ a third party directive ```fooBar```.
In your example, there is no programmatic way to know that the context local is named "$event" because I didn't write this code. My only solution is to read either the code or the doc if present but that is not always the case...

petebacondarwin:
Sure, but the third party should document their directive, no?

petebacondarwin:
[...]

PR #13454, opened 2015-12-06
feat($parse): provide a mechanism to access the locals object

lgalfaso:
Extends the built-in identifiers definitions by adding `$local`. This is a non-assignable reference to the locals object.

Closes: #13247

### DIFF
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -550,8 +550,23 @@ describe('parser', function() {
     });
 
 
-    it('should not confuse `this`, `undefined`, `true`, `false`, `null` when used as identfiers', function() {
-      forEach(['this', 'undefined', 'true', 'false', 'null'], function(identifier) {
+    it('should understand the `$locals` expression', function() {
+      expect(createAst('$locals')).toEqual(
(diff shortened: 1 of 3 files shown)


test/ng/parseSpec.js
```diff
@@ -3589,6 +3604,28 @@ describe('parser', function() {
...
+            $rootScope.$eval('$local.bar = 23');
+          }).toThrow();
+          expect($rootScope.$eval('$local', local)).toBe(local);
+          expect($rootScope.$eval('$local.foo', local)).toBe(42);
+          expect(function() {
+            $rootScope.$eval('$local = {}', local);
+          }).toThrow();
+          expect($rootScope.$eval('$local.bar = 23', local)).toEqual(23);
+          expect(local.bar).toBe(23);
+        }));
```
gkalpak: It would be nice to have a test with `this.$local`.

gkalpak:
I would prefer `$locals` (over `$local`), since this is how we've been referring to it elsewhere (e.g. in the docs).

Other than that (and the test nitpick), it LGTM.

PR #13455, opened 2015-12-06
fix($compile): allow transclude maps to use normalized camelCase element names

### LINKED ISSUE #13439: change suggestion for new Multi-slot transclusion feature
Hi,
I want to suggest that the structure of the transclude object value shown in the example for https://docs.angularjs.org/api/ng/directive/ngTransclude  (v1.5.0-build.4414 (snapshot)) would be the other way around.
i.e. instead of 
```
 transclude: {
        'paneTitle': '?title',
        'paneBody': 'body'
      },
```
it should be (IMHO) 
```
 transclude: {
         'title': '?paneTitle'
         'body': 'paneBody'
      },
```
This would be more consistant with similar maps such as those used in isolated scope and bindToController in the sense that the property name should be the internal element to which the binding happens (the value of the ng-transclude attribute in the template in our case) and the value should say how to resolve it from the "outside".

see also http://stackoverflow.com/questions/22079587/angularjs-transcluding-multiple-sub-elements-in-a-single-angular-directive/34063645#34063645 where I actually suggested that such a feature would be helpful.

petebacondarwin:
The ability to do this was removed at https://github.com/angular/angular.js/commit/c3a2691115b92536fb3d213d0ca16ac68cf32415#diff-a732922b631efed1b9f33a24082ae0dbL1881

On second thoughts we should support both camelCase normalized and exact match elements.

Closes #13439

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -480,11 +480,13 @@
  *
  * **Mult-slot transclusion** is declared by providing an object for the `transclude` property.
  *
- * This object is a map where the keys are the name of the slot to fill and the value is the element selector
- * used to match the HTML to the slot. Only element names are supported for matching. If the element selector
- * is prefixed with a `?` then that slot is optional.
+ * This object is a map where the keys are the name of the slot to fill and the value is an element selector
+ * used to match the HTML to the slot. The element selector can be in normalized camelCase form and will match
+ * elements using any of the standard variants of the camelCase form.
  *
- * For example, the transclude object `{ slotA: '?my-custom-element' }` maps `<my-custom-element>` elements to
+ * If the element selector is prefixed with a `?` then that slot is optional.
+ *
+ * For example, the transclude object `{ slotA: '?myCustomElement' }` maps `<my-custom-element>` elements to
  * the `slotA` slot, which can be accessed via the `$transclude` function or via the {@link ngTransclude} directive.
  *
  * Slots that are not marked as optional (`?`) will trigger a compile time error if there are no matching elements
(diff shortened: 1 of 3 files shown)


src/ng/compile.js
```diff
@@ -480,11 +480,13 @@
  *
  * **Mult-slot transclusion** is declared by providing an object for the `transclude` property.
  *
- * This object is a map where the keys are the name of the slot to fill and the value is the element selector
- * used to match the HTML to the slot. Only element names are supported for matching. If the element selector
- * is prefixed with a `?` then that slot is optional.
+ * This object is a map where the keys are the name of the slot to fill and the value is an element selector
+ * used to match the HTML to the slot. The element selector can be in normalized camelCase form and will match
```
Narretz: We could link to the directive guide/normalization here

Narretz:
I wonder what happens if we have an "svg directive"?  Because some svg elements are camelCased: https://developer.mozilla.org/docs/Web/SVG/Element

Narretz:
Another thing: is it worth mentioning that slot transclusion always transcludes the element and not just its contents?

petebacondarwin:
> Another thing: is it worth mentioning that slot transclusion always transcludes the element and not just its contents?

Yes - I will add that

Narretz:
We just need to make sure this BC doesn't make it into the migration docs. Can you amend the commit message to say this only affects 1.5.0 betas?

Narretz:
Also, the commit message doesn't reflect what the commit actually does. 

Otherwise, LGTM

PR #13456, opened 2015-12-07
Fixed the headings for Changes to your code table

eltronix:


gkalpak:
The resulting HTML was not valid due to multiple `thead` elements ([according to MDN][1] at least - browsers seem to handle the resulting DOM fine).  

I fixed this in #13459 (which also takes care of vertical spacing and header borders), but thx @eltronix for taking the time to submit the PR (and for bringing it to our attention).

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table

eltronix:
Your welcome, thanks for responding.

Issue #13388, opened 2015-11-26
IE11 Error: Invalid argument when parsing a link href containing a percent sign


pneyer:
Hi

Angular cannot render links with percent signs in it when using IE11 as a browser.
A valid email address like "test%me@github.com" inside an <a ng-href="..."> Renders an error to IE11 console "Error: Invalid argument" and the link cannot be rendered.
Example: <a ng-href="mailto:{{'test%me@github.com'}}">a link</a>

Verified in IE11 version: 11.0.9600.18098 using angular 1.4.8

Reproducible here: http://plnkr.co/edit/9ksVLjk6tFpgRPvZAXVq

Other browsers are not affected, i.e. Chrome
![screen shot 2015-11-26 at 16 13 47](https://cloud.githubusercontent.com/assets/9465649/11426014/5652c616-9459-11e5-9861-541c8a8e19e6.png)

Narretz:
Stupid IE! Do you know if this happens in IE lower than 11?

gkalpak:
FYI, it doesn't work on Edge either (but it doesn't throw the error). It just won't set the anchor's `href` attribute.

gutierri:
@Narretz Stupid IE!² This happens in IE 10
![angular_bug](https://cloud.githubusercontent.com/assets/1111356/11435453/eaf9d71a-94c1-11e5-9f69-896262e52c6f.png)

andypatterson:
I did some digging, and this turns out to be annoyingly complex. IE doesn't follow the [URL spec for percent-encoded bytes](https://url.spec.whatwg.org/#percent-encoded-bytes): it follows [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.4) instead (which states that non-escape `%`s should be escaped to `%25`).

I've added a unit test in case this changes in future - to prevent this is IE would be a big-ish rewrite of `urlUtils.urlResolve`

lsblsb:
Same problem here. I stumbled upon this problem, when accessing obfuscated url encoded email-addresses with have a lot of % in it. But - it did not break on some same structured strings before. an error was thrown first when I tried to access this string in a loop: "%0E%13%06%02!%19%04%00%5E%04%00%00%03K%08K". Maybe only specific string-sequences including % are affected?...

gkalpak:
FWIW, this seems to be fixed on Edge 18.

mooncser:
> FWIW, this seems to be fixed on Edge 18.

Is this issue solved in IE11?

PR #13458, opened 2015-12-07
test(booleanAttrsSpec): add unit test for IE11 URL parsing failure

andypatterson:
IE11/10/Edge fail when setting a href to a URL containing a % that isn't a valid escape sequence
Reference: https://github.com/angular/angular.js/issues/13388

### DIFF
--- a/test/ng/directive/booleanAttrsSpec.js
+++ b/test/ng/directive/booleanAttrsSpec.js
@@ -283,6 +283,16 @@ describe('ngHref', function() {
     expect(element.attr('href')).toEqual(undefined);
   }));
 
+  if (msie) {
+    it('should throw error if ng-href contains a non-escaped percent symbol', inject(function($rootScope, $compile) {
+      element = $compile('<a ng-href="http://www.google.com/{{\'a%link\'}}">')($rootScope);
+
+      expect(function() {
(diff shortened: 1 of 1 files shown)


petebacondarwin:
If the point of this is to highlight a bug in IE then we should make that clear in the code as well as the commit and include a link to the issue. Otherwise when this starts to fail we will wonder why it was ever there.

petebacondarwin:
Otherwise LGTM

test/ng/directive/booleanAttrsSpec.js
```diff
@@ -283,6 +283,16 @@ describe('ngHref', function() {
     expect(element.attr('href')).toEqual(undefined);
   }));
 
+  if (msie) {
+    it('should throw error if ng-href contains a non-escaped percent symbol', inject(function($rootScope, $compile) {
+      element = $compile('<a ng-href="http://www.google.com/{{\'a%link\'}}">')($rootScope);
```
SpaRock: undefined

petebacondarwin: ??

SpaRock:
b2b896f
1f8431b

PR #13459, opened 2015-12-07
docs(guide/$location): fix table header formatting

gkalpak:
Al alternative approach to #13456 (with improved spacing and borders).


gkalpak:
**Before:**

before-13459

---
**After:**

after-13459

gkalpak:
Baclported to `v1.4.x` as e26256f.

PR #13460, opened 2015-12-07
fix(input): add missing chars to URL validation regex

### LINKED ISSUE #13379: input[url] validation regex is not valid
In v1.4.8 the validation scheme for `url` inputs was changed, causing valid URLs with semicolons, which previously passed the filter, to become invalid

Our application handles the generation of long urls with messy query strings to be used in advertising, so it's important to us that esoteric characters remain supported. [Semicolons are valid URL characters](http://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid)

You can reproduce the error at the angular code example for a [url input](https://docs.angularjs.org/api/ng/input/input%5Burl%5D), try entering
```
http://www.test.com/asdf;df
http://www.test.com/asdf?asd=f;df=df
```

The fix is simply to add a semicolon to the last chunk of the validation regex [here](https://github.com/angular/angular.js/commit/cb51116dbd225ccfdbc9a565a66a170e65d26331)

andypatterson:
Update the list of permitted chars in URLs

Closes #13379

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -12,7 +12,7 @@
 // Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
 var ISO_DATE_REGEXP = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/;
 // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
-var URL_REGEXP = /^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/]*)?$/;
+var URL_REGEXP = /^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/[\]$'()*,;~]*)?$/;
 var EMAIL_REGEXP = /^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+@[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i;
 var NUMBER_REGEXP = /^\s*(\-|\+)?(\d+|(\d*(\.\d*)))([eE][+-]?\d+)?\s*$/;
 var DATE_REGEXP = /^(\d{4})-(\d{2})-(\d{2})$/;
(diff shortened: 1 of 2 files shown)


src/ng/directive/input.js
```diff
@@ -12,7 +12,7 @@
 // Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
 var ISO_DATE_REGEXP = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/;
 // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
-var URL_REGEXP = /^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/]*)?$/;
+var URL_REGEXP = /^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/\[\]$'\(\)\*\,\;\~]*)?$/;
```
gkalpak: Nitpick, but it's not necessary to escape special characters inside the `[...]` (other than `]` obviously - and sometimes `-` I think). It's not important, but it does make the RegExp more readable.

andypatterson: Whoops, thanks - have removed.

src/ng/directive/input.js
```diff
@@ -12,7 +12,7 @@
 // Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
 var ISO_DATE_REGEXP = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/;
 // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
-var URL_REGEXP = /^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/]*)?$/;
+var URL_REGEXP = /^[A-Za-z][A-Za-z\d.+-]*:\/*(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/\[\]$'\(\)\*,;~]*)?$/;
```
gkalpak: Basically, I meant you don't need to escape **any** of the added characters other than `]`.

gkalpak:
LGTM (other that a minor nitpick).

gkalpak:
Backported to `v1.4.x` as 2995b54.

Thx @andypatterson !

PR #13461, opened 2015-12-07
fix($animateCss): remove animation end event listeners on close

### LINKED ISSUE #10387: Angular-animate.js  /  onAnimationProgress(event) function null error IE
Browser: IE 9, 10, 11
Angular-Animate Version: v1.2.21
Issue:  ev.elaspedTime is null and calling the toFixed function on this causes an exception.

Narretz:
Previously the transition/animation end events were not removed when the
animation was closed. This normally didn't matter, because
the close function knows the animation are closed and won't do work
twice.
However, the listeners themselves do computation that could fail when
the event was missing some data, for example when the event was
triggered instead of natural.

Closes #10387

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -474,6 +474,8 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       var maxDelayTime;
       var maxDuration;
       var maxDurationTime;
+      var startTime;
+      var events = [];
 
       if (options.duration === 0 || (!$sniffer.animations && !$sniffer.transitions)) {
         return closeAndReturnNoopAnimator();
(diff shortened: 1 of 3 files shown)


Narretz:
That isn't good enough, actually. The listeners should be removed inside the close fn, because that is called in at least the following cases: normal end, closing timeout, cancel

matsko:
This looks really good. Please fix up the PR with the spy changes and let's get this green and merge it in.

Narretz:
Thanks for the review. The changes simplfy the code quite a bit! I've also added an expect to see if the on/off calls match transition/animation-end. That way I caught a bug in test - the closing test didn't actually test for animationend.

I will squash the commits if it LGTY.

PR #13462, opened 2015-12-07
fix($resource): remove (broken) support for promises as `timeout`

### LINKED ISSUE #13393: fix($resource): remove (broken) support for _promise as `timeout`_
7170f9d which added support for using a promise as `timeout` in `$resource`, introduced a bug (where you had to re-create the resource class after every cancelled request and failing to do so would abort all subsequent calls). See https://github.com/angular/angular.js/pull/12657#issuecomment-152108887 for more context.

This has been fixed in `v1.5.x` with #13210, but we need to also fix it on `v1.4.x` (#13210 is not suitable for backporting to `v1.4.x` since it adds a new feature and changes the behaviour of `$resource` to some extent).

As discussed in https://github.com/angular/angular.js/pull/13210#issuecomment-159621088, for `v1.4.x` we need to:

- [x] Revert 7170f9d.
- [x] Update the docs to explain that _promise as timeout_ is not supported.
- [x] Log a warning when someone does it.

gkalpak:
This fixes #13393 by:
1. Reverting 7170f9d.
2. Update the docs to note that promise as timeout is not supported.
3. Log a warning when someone tries to use it.

The commits could/should be squashed before merging. We probably need a BC notice (since the broken support for promises as `timeout` has sneaked into `v1.4.8`, right ?

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -568,17 +568,8 @@ angular.module('ngResource', ['ng']).
               undefined;
 
             forEach(action, function(value, key) {
-              switch (key) {
-                default:
-                  httpConfig[key] = copy(value);
-                  break;
-                case 'params':
-                case 'isArray':
-                case 'interceptor':
-                  break;
-                case 'timeout':
-                  httpConfig[key] = value;
-                  break;
+              if (key != 'params' && key != 'isArray' && key != 'interceptor') {
+                httpConfig[key] = copy(value);
               }
             });
 
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -150,8 +150,11 @@ function shallowClearAndCopy(src, dst) {
  *     GET request, otherwise if a cache instance built with
  *     {@link ng.$cacheFactory $cacheFactory}, this cache will be used for
  *     caching.
- *   - **`timeout`** – `{number|Promise}` – timeout in milliseconds, or {@link ng.$q promise} that
- *     should abort the request when resolved.
+ *   - **`timeout`** – `{number}` – timeout in milliseconds.<br />
+ *     **Note:** In contrast to {@link ng.$http#usage $http.config}, {@link ng.$q promises} are
+ *     **not** supported in $resource, because the same value would be used for multiple requests.
+ *     If you need support for cancellable $resource actions, you should upgrade to version 1.5 or
+ *     higher.
  *   - **`withCredentials`** - `{boolean}` - whether to set the `withCredentials` flag on the
  *     XHR object. See
  *     [requests with credentials](https://developer.mozilla.org/en/http_access_control#section_5)
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -368,7 +368,7 @@ angular.module('ngResource', ['ng']).
       }
     };
 
-    this.$get = ['$http', '$q', function($http, $q) {
+    this.$get = ['$http', '$log', '$q', function($http, $log, $q) {
 
       var noop = angular.noop,
         forEach = angular.forEach,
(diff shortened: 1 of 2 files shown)


src/ngResource/resource.js
```diff
@@ -150,8 +150,10 @@ function shallowClearAndCopy(src, dst) {
  *     GET request, otherwise if a cache instance built with
  *     {@link ng.$cacheFactory $cacheFactory}, this cache will be used for
  *     caching.
- *   - **`timeout`** – `{number|Promise}` – timeout in milliseconds, or {@link ng.$q promise} that
- *     should abort the request when resolved.
+ *   - **`timeout`** – `{number}` – timeout in milliseconds.<br />
+ *     **Note:** In contrast to {@link ng.$http#usage $http.config}, {@link ng.$q promises} are
+ *     **not** supported in $resource, because the same value has to be re-used for multiple
```
petebacondarwin: Same value **would** be used?

src/ngResource/resource.js
```diff
@@ -568,17 +570,13 @@ angular.module('ngResource', ['ng']).
...
-                case 'timeout':
-                  httpConfig[key] = value;
-                  break;
+              if ((key === 'timeout') && value && !angular.isNumber(value)) {
+                $log.debug('ngResource:\n' +
+                           '  Only numeric values are allowed as `timeout`.\n' +
+                           '  Promises are not supported in $resource, because the same value ' +
+                           'has to be re-used for multiple requests.');
+              } else if (key != 'params' && key != 'isArray' && key != 'interceptor') {
+                httpConfig[key] = copy(value);
```
petebacondarwin: I think I find the switch statement easier to follow. What is the benefit of changing to if?

petebacondarwin:
Some minor comments but lgtm

src/ngResource/resource.js
gkalpak: I find the `switch` statement better, but it was added as part of 7170f9d (which was reverted).
I aimed for the least amount of change  - it's more obvious if you check the 3rd commit alone (https://github.com/gkalpak/angular.js/commit/ed915fe5bb15a5b7a53a8260e8e14fa7f1c29c6f).

(But the `switch` looks arguably better, so I am fine putting it back.)

src/ngResource/resource.js
gkalpak: I copied it from `v1.5.x`. I should change it there too, right ?

src/ngResource/resource.js
petebacondarwin: Let's use switch! But perhaps add it as a `refact` commit after the revert?

src/ngResource/resource.js
petebacondarwin: If it makes sense to you.

gkalpak: I updated it and changed it on master as well (3694390).

src/ngResource/resource.js
gkalpak: Done.

gkalpak:
@petebacondarwin, I addressed the comments. What about a BC ?

Here's the situation:

- Up to `v1.4.7` (included), using a promise as `timeout` in `$resource`, would silently fail (i.e. have no effect).
- In `v1.4.8`, using a promise as `timeout` will have the (buggy) behaviour described in https://github.com/angular/angular.js/pull/12657#issuecomment-152108887. (I.e. it will work as expected for the first time you resolve the promise and will cancel all subsequent requests after that - one has to re-create the resource class. This is not documented.)
- With this change, using a promise as `timeout` will log a warning and ignore the `timeout` value.

Do we need a BC notice ?

---
It just occurred me: You said we should [_recommend an upgrade to 1.5_][1] in the docs and warning message. 

[1]: https://angular-team.slack.com/archives/angular-1/p1448615582000246

gkalpak:
I updated the docs and the warning message to add a recommendation to upgrade to `v1.5` if one needs support for cancellable `$resource` actions.

petebacondarwin:
Yes, we should put a simple BC notice along the lines of your comment https://github.com/angular/angular.js/pull/13462#issuecomment-162915050

I think in general it is safer to have BCs where they were not needed rather than the other way around

petebacondarwin:
Landed as 474865242c89ba3e8143f0cd52f8c292979ea730 and previous three commits.

PR #13466, opened 2015-12-08
Fix $animate.pin() to work with nested elements

Narretz:
Given the following DOM structure:

```
- e1
  -e2
    -e3 <--- pin this
      e4 <-- animate this
```

the function that is responsible for finding the parent element that is pinned to the app root would find e3, set it as parent, but not indicate that a rootElement had been found.

The PR also contains a second commit that adds a test and fixes a use of minErr inside ngAnimate.


matsko:
The test looks good, but the fix needs to support upper parent pining. Let's chat about this issue during our meeting.

PR #13467, opened 2015-12-08
fix(ngOptions): unset [selected] on unselected empty option

Narretz:
(Cleaned up version of #13331)

Firefox (verified on versions 38..42, Windows and Linux-AMD64) can get
confused about what option is selected after options ngOptions populates
the select control when it has a empty option that is ng-disabled.  It
appears to think both the empty option added by ngOptions and the one
selected by ngModel are selected.  The trigger seems to be angular
leaving the empty option both selected and disabled.

This patch works around the problem.

### DIFF
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -2217,6 +2217,23 @@ describe('ngOptions', function() {
       dealoc(element);
     }));
 
+    // See issue #13331
+    it('should set [selected] to false when a genuine option is selected', function() {
+      scope.options = ['a'];
+      scope.model = '';
+      compile(
+          '<select ng-model="model" ng-options="opt for opt in options">' +
+            '<option value="">Empty Option</option>' +
+          '</select>');
+      var opt = element.find("option")[0];
+      expect(opt.hasAttribute('selected')).toBeTruthy();
+      expect(opt.selected).toBeTruthy();
+      scope.model = 'a';
+      scope.$digest();
+      expect(opt.hasAttribute('selected')).toBeFalsy();
+      expect(opt.selected).toBeFalsy();
+    });
+
   });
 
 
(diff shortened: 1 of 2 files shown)


test/ng/directive/ngOptionsSpec.js
```diff
@@ -2217,6 +2217,21 @@ describe('ngOptions', function() {
       dealoc(element);
     }));
 
+    // See issue #13331
+    it('should set [selected] to false when a genuine option is selected', function() {
+      scope.options = ['a'];
+      scope.model = '';
+      compile(
+          '<select ng-model="model" ng-options="opt for opt in options">' +
+            '<option id="empty_option" value="">Empty Option</option>' +
```
gkalpak: Is the `id` needed here ?

test/ng/directive/ngOptionsSpec.js
```diff
@@ -2217,6 +2217,21 @@ describe('ngOptions', function() {
...
+          '</select>');
+      var opt = element.find("option")[0];
+      expect(opt.getAttribute("selected")).toBeTruthy();
+      scope.model = 'a';
+      scope.$digest();
+      expect(opt.getAttribute("selected")).toBeFalsy();
```
gkalpak: Wouldn't it be better to check with `hasAttribute()` ?
Even an empty (falsy) attribute value (e.g. `selected=""`) might mean the option is selected.

I would also check the property for good measure (here and above).

gkalpak:
It's unfortunate that we can't reproduce the original issue (as described in #13331), but the fix is reasonable anyway.

LGTM (besides some nitpicks in tests).

test/ng/directive/ngOptionsSpec.js
Narretz: Good point on hasAttribute and the selected property. Updated.

petebacondarwin: I think you are right about using `hasAttribute` here since `<option value="value2" selected>Value 2</option>` is a valid selected option. Elsewhere in this spec we just look at the `opt.selected` property instead. Is there some reason why we are checking the attribute rather than the property here?

petebacondarwin:
LGTM - please merge

Narretz:
Got a problem now with IE9, which doesn't report option.selected as falsy. Also, I think we should use removeAttr instead of attr('selected', false). Precisely because selected is an attribute that is "true" by presence.

petebacondarwin:
Closing as the fix is not great and there has been no response for a while from the OP.

PR #13468, opened 2015-12-08
add verticalBufferSize parameter to $swipe.bind

Binvention:
adds a parameter to $swipe.bind to allow for a different vertical buffer separating swipe from scroll. currently swipe vs scroll is determined by the change in y being greater then the change in x

### DIFF
--- a/src/ngTouch/swipe.js
+++ b/src/ngTouch/swipe.js
@@ -139,7 +139,7 @@ ngTouch.factory('$swipe', [function() {
         lastPos = coords;
 
 		//set the buffer size to custom or to the default of 10
-		if  (angular.isNumber(bufferSize)) {
+		if (angular.isNumber(bufferSize)) {
 			MOVE_BUFFER_RADIUS=bufferSize;
 		}
         if (totalX < MOVE_BUFFER_RADIUS && totalY < MOVE_BUFFER_RADIUS) {
--- a/src/ngTouch/swipe.js
+++ b/src/ngTouch/swipe.js
@@ -94,7 +94,7 @@ ngTouch.factory('$swipe', [function() {
      * as described above.
      *
      */
-    bind: function(element, eventHandlers, pointerTypes) {
+    bind: function(element, eventHandlers, pointerTypes,bufferSize) {
       // Absolute total movement, used to control swipe vs. scroll.
       var totalX, totalY;
       // Coordinates of the start position.
@@ -138,6 +138,10 @@ ngTouch.factory('$swipe', [function() {
 
         lastPos = coords;
 
+		//set the buffer size to custom or to the default of 10
+		if(angular.isNumber(bufferSize)){
+			MOVE_BUFFER_RADIUS=bufferSize;
+		}
         if (totalX < MOVE_BUFFER_RADIUS && totalY < MOVE_BUFFER_RADIUS) {
           return;
         }
--- a/src/ngTouch/swipe.js
+++ b/src/ngTouch/swipe.js
@@ -94,7 +94,7 @@ ngTouch.factory('$swipe', [function() {
      * as described above.
      *
      */
-    bind: function(element, eventHandlers, pointerTypes,bufferSize) {
+    bind: function(element, eventHandlers, pointerTypes, bufferSize) {
       // Absolute total movement, used to control swipe vs. scroll.
       var totalX, totalY;
       // Coordinates of the start position.
@@ -139,7 +139,7 @@ ngTouch.factory('$swipe', [function() {
         lastPos = coords;
 
 		//set the buffer size to custom or to the default of 10
-		if(angular.isNumber(bufferSize)){
+		if  (angular.isNumber(bufferSize)) {
 			MOVE_BUFFER_RADIUS=bufferSize;
 		}
         if (totalX < MOVE_BUFFER_RADIUS && totalY < MOVE_BUFFER_RADIUS) {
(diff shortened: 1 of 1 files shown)


Binvention:
okay I signed it

src/ngTouch/swipe.js
```diff
@@ -81,9 +83,10 @@ ngTouch.factory('$swipe', [function() {
      * distance moved in either dimension exceeds a small threshold.
      *
      * Once this threshold is exceeded, either the horizontal or vertical delta is greater.
-     * - If the horizontal distance is greater, this is a swipe and `move` and `end` events follow.
-     * - If the vertical distance is greater, this is a scroll, and we let the browser take over.
-     *   A `cancel` event is sent.
+     * - If the horizontal distance is greater or the vertical distance is less then the vertical buffer,
```
gkalpak: _then_  -->  _than_

src/ngTouch/swipe.js
```diff
@@ -71,6 +71,8 @@ ngTouch.factory('$swipe', [function() {
      * The pointer types that should be used can be specified via the optional
      * third argument, which is an array of strings `'mouse'` and `'touch'`. By default,
      * `$swipe` will listen for `mouse` and `touch` events.
+	 * For vertical swipe motions the fourth optional argument can be used to change the vertical buffer 
+	 * that determines what is a swipe and what is a scroll.
```
gkalpak: Indentation is off.

src/ngTouch/swipe.js
```diff
@@ -81,9 +83,10 @@ ngTouch.factory('$swipe', [function() {
      * distance moved in either dimension exceeds a small threshold.
      *
      * Once this threshold is exceeded, either the horizontal or vertical delta is greater.
-     * - If the horizontal distance is greater, this is a swipe and `move` and `end` events follow.
-     * - If the vertical distance is greater, this is a scroll, and we let the browser take over.
-     *   A `cancel` event is sent.
+     * - If the horizontal distance is greater or the vertical distance is less then the vertical buffer, 
+	 *   this is a swipe and `move` and `end` events follow.
```
gkalpak: Indentation is off.

src/ngTouch/swipe.js
```diff
@@ -81,9 +83,10 @@ ngTouch.factory('$swipe', [function() {
...
+     * - If the vertical distance is greater then the horizontal and the vertical buffer, this is a scroll, and we let the browser take over.
```
gkalpak: _then_  -->  _than_

Also, please, wrap lines at 100 chars.

src/ngTouch/swipe.js
```diff
@@ -103,6 +106,7 @@ ngTouch.factory('$swipe', [function() {
       var lastPos;
       // Whether a swipe is active.
       var active = false;
+	  // Default virtical buffer
```
gkalpak: Indentation is off.
Some code seems to be missing :)

src/ngTouch/swipe.js
```diff
@@ -138,12 +142,16 @@ ngTouch.factory('$swipe', [function() {
 
         lastPos = coords;
 
+		//set the buffer size to custom or to the default of 10
+		if (!angular.isNumber(verticalBufferSize)) {
+			verticalBufferSize=MOVE_BUFFER_RADIUS;
+		}
```
gkalpak: Indentation is off.

Also, we shouldn't check `verticalBuffrSize` on each call. Let's set it once at the top of the `bind()` method.

gkalpak:
Thx @Binvention for submitting a PR. There are still a few things to be taken care of:

- [ ] Our `jscs:src` task is failing due to some trailing whitespace. This needs to be fixed.  
      BTW, it'a a good idea to run all tests locally (i.e. execute `grunt test`), so you can find out (sooner than later) whether your change broke anything.

- [ ] I left some (mostly minor) comments, which need to be addressed.

- [ ] As with every new feature, we need test coverage for the added functionality (i.e. one or more tests showing that the feature works the way it is supposed to and preventing future regressions).

Feel free to ask here, if you need help with any of the above.

--
One last thing:
I was under the impression that that is new parameter would affect the `MOVE_BUFFER_RADIUS` (and thus both `totalX` _and_ `totalY`). Why did you decide to only affect the vertical displacement (aka `totalY`) ?

Binvention:
It was trying to look at possible conflicts in the code and realized that the move buffer radius also determined how much distance it takes before it's considered a swipe or scroll at all. So changing the move buffer radius would increase or decrease the sensitivity of the swipe method but it wouldn't necessarily help determining between a swipe and a scroll

gkalpak:
@Binvention, you can find more info about submitting/updating a PR and about building/testing AngularJS in [here][1] and [there][2].

[1]: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#submit
[2]: https://docs.angularjs.org/misc/contribute

gkalpak:
So is this change about supporting _vertical swipe_ ?
I am a little confused. How can you support both vertical swipe and vertical scroll ? It will be one confusing UX (unless I'm missing something).

@Binvention, could you describe the usecase/how it will be used.

Binvention:
you could support vertical swipe and vertical scroll by creating a buffer size on touch move because then programs that want a vertical swipe can set a high buffer and programs that want a vertical scroll can set a low buffer

Narretz:
I'm closing this PR since the ngTouch module will be deprecated in AngularJS 1.7.0, and we will not do any more work on it. This is because touch handling is a complex topic, and there are stand-alone libraries that are better equipped to handle all the nuances and edge cases. For example, https://hammerjs.github.io/, which is also used by Angular.

PR #13472, opened 2015-12-09
fix(ngAnimate): don't normalize classes on follow-up animations to jo…

### LINKED ISSUE #13339: ngAnimate doesn't remove class which is animating
Removing class which is currently animating doesn't work. 

**Angular Version:** since 1.4.5 till 1.5.0-build.4389
**Browser:** Any
**Reproducable:** almost always (sometimes doesn't reproduce under debugging)
**Operating system:** Any

**Detail**
For the example of bug see [plunker](http://plnkr.co/edit/BsWymZhJ35Ljs9EYSu1Z?p=preview)
By normal it should not have class `pending` at the end.

What happens:
1. Adding class `pending`
2. NgModel adds class `ng-valid-required` (it's joining animation and applies preparation class)
3. Timeout is fullfilled
4. Trying to remove class. 
Inside ngAnimations it checks that element has animations, it cannot be cancelled and skipped, so it's try to join animation. Because element is in running state it tries to normalize options.  Function normalize options cannot find class  `pending` and remove it from options. Remove animation exits as invalid.
5. Class `pending` added to dom node

### LINKED ISSUE #13380: ng-show sometimes does not remove ng-hide class after expression becames truthy

matsko:
…ined animations

This allows follow-up animations to remove a class that is currently
being added.

Fixes #13339
Fixes #13380
Closes #13414

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -12,6 +12,14 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     join: []
   };
 
+  function makeTruthyMapFromKeys(keys) {
+    var map = Object.create(null);
+    forEach(keys, function(key) {
+      map[key] = true;
+    });
+    return map;
+  }
+
   function isAllowed(ruleType, element, currentAnimation, previousAnimation) {
     return rules[ruleType].some(function(fn) {
       return fn(element, currentAnimation, previousAnimation);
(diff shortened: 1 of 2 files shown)


matsko:
@petebacondarwin please review.

Narretz:
We need to update the commit message, it's doing something differen now.

src/ngAnimate/animateQueue.js
```diff
@@ -59,11 +67,38 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
...
+
+    // early detection to save the global CPU shortage :)
+    if ((!isDefined(nA) && !isDefined(nR)) || (!isDefined(cA) && !isDefined(cR))) {
+      return false;
+    }
+
+    var cancelSomething = false;
+
+    cA = cA ? makeTruthyMapFromKeys(cA.split(ONE_SPACE)) : null;
+    cR = cR ? makeTruthyMapFromKeys(cR.split(ONE_SPACE)) : null;
```
petebacondarwin: This repeated logic here could be moved into the `makeTrutyMapFromKeys` and perhaps be renamed to `makeCssClassMap`

src/ngAnimate/animateQueue.js
```diff
@@ -59,11 +67,38 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
...
+      });
+
+      if (cancelSomething) return true;
+    }
+
+    if (cA && nR) {
+      cancelSomething = nR.split(ONE_SPACE).some(function(className) {
+        return cA[className];
+      });
+    }
```
petebacondarwin: The repeated logic here could be moved into a function...

```js
return (hasMatchingClasses(nA, cR)) || (hasMatchingClasses(nR, cA));
```

```js
function hasMatchingClasses(classString, classMap) {
  if (classString && classMap) {
    return classString.split(ONE_SPACE).some(function(className) {
      return classMap[className];
    });
  }
}
```

PR #13477, opened 2015-12-09
docs(ngTransclude): name value pair were the wrong way round

jamesclark92:
The plunker associated with multiple transcludes fails because of this error.

### DIFF
--- a/src/ng/directive/ngTransclude.js
+++ b/src/ng/directive/ngTransclude.js
@@ -126,9 +126,9 @@
  *        return {
  *          restrict: 'E',
  *          transclude: {
- *            'title': '?paneTitle',
- *            'body': 'paneBody',
- *            'footer': '?paneFooter'
+ *            'paneTitle': '?title',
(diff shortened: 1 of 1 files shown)


jamesclark92:
@googlebot I signed it

Narretz:
This has been switched in master, and is coming up in version 1.5.0-rc.0. The left hand side is the target slot in the template, the right hand side is the element name in the directive content.

What was the exact url / version of the docs where this didn't work?

jamesclark92:
@Narretz https://docs.angularjs.org/api/ng/directive/ngTransclude v1.5.0-build.4432(snapshot)

Narretz:
This is caused by how hour docs work. The docs are built from the latest master snapshot, but the actual angular files are sourced from the CDN. The CDN doesn't have the snapshot, so it falls back to the last release. If the master contains changes that the last version does not have, you get a mismatch. This is something we should fix.

PR #13480, opened 2015-12-09
fix($compile): Fix namespace detection for achor elements

feelepxyz:
Fixed a bug where anchor elements with `SVG` in the href attribute got incorrectly detected as svgs. Seems the detection only happens when the anchor element has a child that needs to be transcluded.

Calling `node.toString()` on a anchor returns the `href` value instead of the expected `[object HTMLAnchorElement]`.

Added a test for this edge case.

Plunker to reproduce: http://plnkr.co/edit/OBgCS8phjTwMTFYwhk6D

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1396,7 +1396,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       if (!node) {
         return 'html';
       } else {
-        return nodeName_(node) !== 'foreignobject' && node.toString().match(/SVG/) ? 'svg' : 'html';
+        return nodeName_(node) !== 'foreignobject' && toString.call(node).match(/SVG/) ? 'svg' : 'html';
       }
     }
 
--- a/src/ng/directive/ngInclude.js
+++ b/src/ng/directive/ngInclude.js
@@ -281,7 +281,7 @@ var ngIncludeFillContentDirective = ['$compile',
       priority: -400,
       require: 'ngInclude',
       link: function(scope, $element, $attr, ctrl) {
-        if (/SVG/.test($element[0].toString())) {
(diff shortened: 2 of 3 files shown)


src/ng/compile.js
```diff
@@ -1396,7 +1396,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       if (!node) {
         return 'html';
       } else {
-        return nodeName_(node) !== 'foreignobject' && node.toString().match(/SVG/) ? 'svg' : 'html';
+        return nodeName_(node) !== 'foreignobject' && Object.prototype.toString.call(node).match(/SVG/) ? 'svg' : 'html';
```
lgalfaso: Just
```js
return nodeName_(node) !== 'foreignobject' && toString.call(node).match(/SVG/) ? 'svg' : 'html';
```

harrison:
@lgalfaso sorted. Added a commit for the same fix in ngInclude, but looking at that code, the function `namespaceAdaptedClone` doesn't seem to be defined anywhere in the codebase.. is this directive ever used? :)

lgalfaso:
@harrison the only reference of `namespaceAdaptedClone` is within `src/ng/ngInclude.js`, do not know how this is related to this PR.

One more thing, can you please name the test something like "should handle anchor elements inside svg elements"?

harrison:
@lgalfaso not related just curious if it's useless code:)

The issue happens with the ngIf/transclusion on the `span` which wrongly detects the parent anchor as being an svg. Changed it to `transclusions should handle parent anchor elements`, happy to change.

lgalfaso:
@harrison can you please squash everything into one commit? this makes things easier. Otherwise this looks good and should merge early next week

PR #13483, opened 2015-12-09
fix(ngAria): Apply ARIA attrs correctly

### LINKED ISSUE #11374: [ngAria] aria-required only applied when ng-model has $error.required

### LINKED ISSUE #11830: [ng-aria] 'aria-disabled' attribute must not be set along with the 'disabled'
Hello,

When using angular aria, if the ng-disabled directive is set on an element and the condition is resolved to 'true', the aria-disabled element is also set automatically.

The W3C mentions that the aria-disabled attribute must not set if the disabled attribute (installed by ng-disabled) is already supported by the DOM element (ex: button).

See: http://www.w3.org/TR/aria-in-html/ look for "disabled"

>Use the disabled attribute on any element that is allowed the disabled attribute in HTML5.
>Only use the aria-disabled attribute for elements that are not allowed to have a disabled attribute in HTML5

Regards

### LINKED ISSUE #13078: [ngAria] aria-required is not allowed for input type="checkbox"
We've the problem that ngAria adds attribute aria-required to our checkboxes and W3C Validator rightly complains about that.

What am I doing wrong? Or is it an issue in ngAria?

marcysutton:
BREAKING CHANGE: Where appropriate, ngAria now applies ARIA to custom controls only, not native inputs. Because of this, support for `aria-multiline` on textareas has been removed. 

New support added for ngValue, ngChecked, and ngRequired, along with updated documentation. Also supports ngTrueValue and ngFalseValue as part of ngModel.

Closes #13078
Closes #11374
Closes #11830

cc @petebacondarwin (I got this done finally! :smile:)

### DIFF
--- a/docs/content/guide/accessibility.ngdoc
+++ b/docs/content/guide/accessibility.ngdoc
@@ -33,6 +33,9 @@ Currently, ngAria interfaces with the following directives:
 
  * {@link guide/accessibility#ngmodel ngModel}
  * {@link guide/accessibility#ngdisabled ngDisabled}
+ * {@link guide/accessibility#ngrequired ngRequired}
+ * {@link guide/accessibility#ngvaluechecked ngChecked}
+ * {@link guide/accessibility#ngvaluechecked ngValue}
  * {@link guide/accessibility#ngshow ngShow}
  * {@link guide/accessibility#nghide ngHide}
  * {@link guide/accessibility#ngclick ngClick}
@@ -137,6 +140,27 @@ the keyboard. It is still up to **you** as a developer to **ensure custom contro
 accessible**. As a rule, any time you create a widget involving user interaction, be sure to test 
 it with your keyboard and at least one mobile and desktop screen reader.
 
+<h2 id="ngvaluechecked">ngValue and ngChecked</h2>
+
+To ease the transition between native inputs and custom controls, ngAria now supports
+{@link ng/directive/ngValue ngValue} and {@link ng/directive/ngChecked ngChecked}.
+The original directives were created for native inputs only, so ngAria extends
+support to custom elements by managing `aria-checked` for accessibility.
+
+###Example
+
+```html
+<custom-checkbox ng-checked="val"></custom-checkbox>
+<custom-radio-button ng-value="val"></custom-radio-button>
+```
+
+Becomes:
+
+```html
+<custom-checkbox ng-checked="val" aria-checked="true"></custom-checkbox>
+<custom-radio-button ng-value="val" aria-checked="true"></custom-radio-button>
+```
+
 <h2 id="ngdisabled">ngDisabled</h2>
 
 The `disabled` attribute is only valid for certain elements such as `button`, `input` and
@@ -148,17 +172,36 @@ custom controls to be more accessible.
 ###Example
 
 ```html
-<md-checkbox ng-disabled="disabled">
+<md-checkbox ng-disabled="disabled"></md-checkbox>
 ```
 
 Becomes:
 
 ```html
-<md-checkbox disabled aria-disabled="true">
+<md-checkbox disabled aria-disabled="true"></md-checkbox>
 ```
 
 >You can check whether a control is legitimately disabled for a screen reader by visiting
 [chrome://accessibility](chrome://accessibility) and inspecting [the accessibility tree](http://www.paciellogroup.com/blog/2015/01/the-browser-accessibility-tree/).
+
+<h2 id="ngrequired">ngRequired</h2>
+
+The boolean `required` attribute is only valid for native form controls such as `input` and
+`textarea`. To properly indicate custom element directives such as `<md-checkbox>` or `<custom-input>`
+as required, using ngAria with {@link ng/directive/ngRequired ngRequired} will also add
+`aria-required`. This tells accessibility APIs when a custom control is required.
+
+###Example
+
+```html
+<md-checkbox ng-required="val"></md-checkbox>
+```
+
+Becomes:
+
+```html
+<md-checkbox ng-required="val" aria-required="true"></md-checkbox>
+```
 
 <h2 id="ngshow">ngShow</h2>
 
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -16,19 +16,23 @@
  *
  * For ngAria to do its magic, simply include the module `ngAria` as a dependency. The following
  * directives are supported:
- * `ngModel`, `ngDisabled`, `ngShow`, `ngHide`, `ngClick`, `ngDblClick`, and `ngMessages`.
+ * `ngModel`, `ngChecked`, `ngRequired`, `ngValue`, `ngDisabled`, `ngShow`, `ngHide`, `ngClick`,
+ * `ngDblClick`, and `ngMessages`.
  *
  * Below is a more detailed breakdown of the attributes handled by ngAria:
  *
  * | Directive                                   | Supported Attributes                                                                   |
  * |---------------------------------------------|----------------------------------------------------------------------------------------|
+ * | {@link ng.directive:ngModel ngModel}        | aria-checked, aria-valuemin, aria-valuemax, aria-valuenow, aria-invalid, aria-required, input roles |
  * | {@link ng.directive:ngDisabled ngDisabled}  | aria-disabled                                                                          |
+ * | {@link ng.directive:ngRequired ngRequired}  | aria-required                                                                          |
+ * | {@link ng.directive:ngChecked ngChecked}    | aria-checked                                                                           |
+ * | {@link ng.directive:ngValue ngValue}        | aria-checked                                                                           |
  * | {@link ng.directive:ngShow ngShow}          | aria-hidden                                                                            |
  * | {@link ng.directive:ngHide ngHide}          | aria-hidden                                                                            |
  * | {@link ng.directive:ngDblclick ngDblclick}  | tabindex                                                                               |
  * | {@link module:ngMessages ngMessages}        | aria-live                                                                              |
- * | {@link ng.directive:ngModel ngModel}        | aria-checked, aria-valuemin, aria-valuemax, aria-valuenow, aria-invalid, aria-required, input roles |
- * | {@link ng.directive:ngClick ngClick}        | tabindex, keypress event, button role                                                               |
+ * | {@link ng.directive:ngClick ngClick}        | tabindex, keypress event, button role                                                  |
  *
  * Find out more information about each directive by reading the
  * {@link guide/accessibility ngAria Developer Guide}.
@@ -90,7 +94,6 @@ function $AriaProvider() {
     ariaDisabled: true,
     ariaRequired: true,
     ariaInvalid: true,
-    ariaMultiline: true,
     ariaValue: true,
     tabindex: true,
     bindKeypress: true,
@@ -108,11 +111,10 @@ function $AriaProvider() {
    *  - **ariaDisabled** – `{boolean}` – Enables/disables aria-disabled tags
    *  - **ariaRequired** – `{boolean}` – Enables/disables aria-required tags
    *  - **ariaInvalid** – `{boolean}` – Enables/disables aria-invalid tags
-   *  - **ariaMultiline** – `{boolean}` – Enables/disables aria-multiline tags
    *  - **ariaValue** – `{boolean}` – Enables/disables aria-valuemin, aria-valuemax and aria-valuenow tags
    *  - **tabindex** – `{boolean}` – Enables/disables tabindex tags
-   *  - **bindKeypress** – `{boolean}` – Enables/disables keypress event binding on `&lt;div&gt;` and
-   *    `&lt;li&gt;` elements with ng-click
+   *  - **bindKeypress** – `{boolean}` – Enables/disables keypress event binding on `div` and
+   *    `li` elements with ng-click
    *  - **bindRoleForClick** – `{boolean}` – Adds role=button to non-interactive elements like `div`
    *    using ng-click, making them more accessible to users of assistive technologies
    *
@@ -151,15 +153,15 @@ function $AriaProvider() {
    *
    *```js
    * ngAriaModule.directive('ngDisabled', ['$aria', function($aria) {
-   *   return $aria.$$watchExpr('ngDisabled', 'aria-disabled');
+   *   return $aria.$$watchExpr('ngDisabled', 'aria-disabled', nodeBlackList, false);
    * }])
    *```
    * Shown above, the ngAria module creates a directive with the same signature as the
    * traditional `ng-disabled` directive. But this ngAria version is dedicated to
-   * solely managing accessibility attributes. The internal `$aria` service is used to watch the
-   * boolean attribute `ngDisabled`. If it has not been explicitly set by the developer,
-   * `aria-disabled` is injected as an attribute with its value synchronized to the value in
-   * `ngDisabled`.
+   * solely managing accessibility attributes on custom elements. The internal `$aria` service is
+   * used to watch the boolean attribute `ngDisabled`. If it has not been explicitly set by the
+   * developer, `aria-disabled` is injected as an attribute with its value synchronized to the
+   * value in `ngDisabled`.
    *
    * Because ngAria hooks into the `ng-disabled` directive, developers do not have to do
    * anything to enable this feature. The `aria-disabled` attribute is automatically managed
@@ -167,12 +169,15 @@ function $AriaProvider() {
    *
    * The full list of directives that interface with ngAria:
    * * **ngModel**
+   * * **ngChecked**
+   * * **ngRequired**
+   * * **ngDisabled**
+   * * **ngValue**
    * * **ngShow**
    * * **ngHide**
    * * **ngClick**
    * * **ngDblclick**
    * * **ngMessages**
-   * * **ngDisabled**
    *
    * Read the {@link guide/accessibility ngAria Developer Guide} for a thorough explanation of each
    * directive.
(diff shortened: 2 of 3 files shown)


petebacondarwin:
Well done @marcysutton - thanks. We will get this in after 1.5.0 is released

docs/content/guide/accessibility.ngdoc
```diff
@@ -137,6 +140,26 @@ the keyboard. It is still up to **you** as a developer to **ensure custom contro
 accessible**. As a rule, any time you create a widget involving user interaction, be sure to test 
 it with your keyboard and at least one mobile and desktop screen reader.
 
+<h2 id="ngvaluechecked">ngValue and ngChecked</h2>
+
+To ease the transition between native inputs and custom controls, ngAria now supports
+[ngValue](https://docs.angularjs.org/api/ng/directive/ngValue) and [ngChecked](https://docs.angularjs.org/api/ng/directive/ngChecked). The original directives were created for native inputs only, so ngAria extends
```
gkalpak: These links should be `{@link ng/directive/ngValue}` and `{@link ng/directive/ngChecked}`.

docs/content/guide/accessibility.ngdoc
```diff
@@ -148,18 +171,37 @@ custom controls to be more accessible.
...
 ```
 
 >You can check whether a control is legitimately disabled for a screen reader by visiting
 [chrome://accessibility](chrome://accessibility) and inspecting [the accessibility tree](http://www.paciellogroup.com/blog/2015/01/the-browser-accessibility-tree/).
 
+<h2 id="ngrequired">ngRequired</h2>
+
+The boolean `required` attribute is only valid for native form controls such as `input` and
+`textarea`. To properly indicate custom element directives such as `<md-checkbox>` or `<custom-input>` 
+as required, using ngAria with [ngRequired](https://docs.angularjs.org/api/ng/directive/input)
```
gkalpak: Link: `{@link ng/directive/input}`

gkalpak: Actually, `ngRequired` now has its own docs page, so it could be `{@link ng/directive/ngRequired}`.

docs/content/guide/accessibility.ngdoc
```diff
@@ -148,18 +171,37 @@ custom controls to be more accessible.
...
+will also add `aria-disabled`. This tells accessibility APIs when a custom control is required.
```
gkalpak: _aria-disabled_  -->  _aris-required_

src/ngAria/aria.js
```diff
@@ -235,30 +251,24 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
+              var boolVal = (attr.value == ngModel.$viewValue);
+              elem.attr('aria-checked', boolVal);
+            };
           }
-
-          function ngAriaCheckboxReaction() {
-            elem.attr('aria-checked', !ngModel.$isEmpty(ngModel.$viewValue));
+          function getCheckboxReaction() {
+            return function ngAriaCheckboxReaction(newVal) {
+              var modelValue = attr.ngTrueValue ? ngModel.$viewValue == ngTrueValue : newVal;
```
gkalpak: `ngTrueValue` and `ngFalseValue` are set on the model value (i.e. `ngModel.$modelValue`). The `$viewValue` is always expected to be boolean (i.e. `true`/`false`). I don't think there needs to be any special handling for `ngTrueValue`/`ngFalseValue`.

Do you have a test that failed previously and needs this ?

BTW, by using `==` you are casting `ngTrueValue` to boolean (since `ngModel.$viewValue` is already boolean), which is not what you want to do here :)

src/ngAria/aria.js
```diff
@@ -235,30 +251,24 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
-            }
+            return function ngAriaRadioReaction(newVal) {
+              var boolVal = (attr.value == ngModel.$viewValue);
+              elem.attr('aria-checked', boolVal);
+            };
           }
-
-          function ngAriaCheckboxReaction() {
-            elem.attr('aria-checked', !ngModel.$isEmpty(ngModel.$viewValue));
+          function getCheckboxReaction() {
```
gkalpak: This wrapper function doesn't seem to be necessary here.

src/ngAria/aria.js
```diff
@@ -306,22 +316,17 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
 
-          if (ngModel.$validators.required && shouldAttachAttr('aria-required', 'ariaRequired', elem)) {
-            scope.$watch(function ngAriaRequiredWatch() {
-              return ngModel.$error.required;
-            }, function ngAriaRequiredReaction(newVal) {
-              elem.attr('aria-required', !!newVal);
+          if (!attr.hasOwnProperty('ngRequired') && ngModel.$validators.required
+            && shouldAttachAttr('aria-required', 'ariaRequired', elem, false)) {
+            // ngModel.$error.required is undefined on custom controls
+            scope.$watch(attr['required'], function ngAriaRequiredWatch() {
```
gkalpak: I think it's more efficient to use `attr.$observe('required', ...)`.

test/ngAria/ariaSpec.js
```diff
@@ -104,18 +104,48 @@ describe('$aria', function() {
...
       expect(element.attr('aria-checked')).toBe('false');
     });
 
-    it('should handle checkbox with string model values using ng(True|False)Value', function() {
-      var element = $compile('<input type="checkbox" ng-model="val" ng-true-value="\'yes\'" ' +
+    it('should not handle native checkbox with ngChecked', function() {
+      var element = $compile('<input type="checkbox" ng-checked="val">')(scope);
+
+      scope.$apply('val = true');
+      expect(element.eq(0).attr('aria-checked')).toBeUndefined();
```
gkalpak: Is `.eq(0)` necessary (here and below) ? Why not just `element.attr(...)` ?

test/ngAria/ariaSpec.js
```diff
@@ -104,18 +104,48 @@ describe('$aria', function() {
...
+
+      scope.$apply('val = true');
+      expect(element.eq(0).attr('aria-checked')).toBe('true');
+
+      scope.$apply('val = false');
+      expect(element.eq(0).attr('aria-checked')).toBe('false');
+    });
+
+    it('should handle custom checkbox with string model values using ng(True|False)Value', function() {
+      var element = $compile('<div role="checkbox" ng-model="val" ng-true-value="\'yes\'" ' +
```
gkalpak: If you wnat to be sure that the behavior is tested correctly, you should use a falsy value for `ngTrueValue` and a truthy value for `ngFalseValue` (e.g. `ng-true-value="0" ng-false-value="1"`).

test/ngAria/ariaSpec.js
```diff
@@ -414,49 +423,24 @@ describe('$aria', function() {
...
-      expect(element.attr('aria-required')).toBe('userSetValue');
-
-      compileElement("<textarea ng-model='val' required aria-required='userSetValue'></textarea>");
-      expect(element.attr('aria-required')).toBe('userSetValue');
-
-      compileElement("<select ng-model='val' required aria-required='userSetValue'></select>");
-      expect(element.attr('aria-required')).toBe('userSetValue');
-
-      compileElement("<input ng-model='val' ng-required='true' aria-required='userSetValue'>");
+      compileElement("<div role='checkbox' ng-model='val' aria-required='userSetValue'></div>");
```
gkalpak: Shouldn't this also have `ng-required="true"` ?

test/ngAria/ariaSpec.js
```diff
@@ -414,49 +423,24 @@ describe('$aria', function() {
   describe('aria-required', function() {
```
gkalpak: There is no test verifying that when `ngRequired` evaluates to false, the `aria-required` attribute is removed.

test/ngAria/ariaSpec.js
```diff
@@ -367,15 +372,10 @@ describe('$aria', function() {
...
     it('should not attach aria-disabled', function() {
-      var element = [
-        $compile("<input ng-disabled='val'>")(scope),
-        $compile("<textarea ng-disabled='val'></textarea>")(scope),
-        $compile("<button ng-disabled='val'></button>")(scope),
-        $compile("<select ng-disabled='val'></select>")(scope)
-      ];
+      compileElement('<div ng-disabled="val"></div>');
 
       scope.$apply('val = false');
```
gkalpak: Shouldn't `val` be `true` to test the actual behavior ?

test/ngAria/ariaSpec.js
```diff
@@ -297,52 +320,34 @@ describe('$aria', function() {
...
+        $compile("<input ng-disabled='val'>")(scope),
+        $compile("<textarea ng-disabled='val'></textarea>")(scope),
+        $compile("<button ng-disabled='val'></button>")(scope),
+        $compile("<select ng-disabled='val'></select>")(scope)
+      ];
 
       scope.$apply('val = true');
-      expect(element.attr('aria-disabled')).toBe('true');
+      angular.forEach(element, function(val) {
+        expect(angular.element(val).attr('aria-disabled')).toBeUndefined();
```
gkalpak: I think there is no need to wrap with `angular.element()`.

test/ngAria/ariaSpec.js
```diff
@@ -297,52 +320,34 @@ describe('$aria', function() {
...
+        expect(angular.element(val).attr('disabled')).toBeDefined();
+      });
     });
```
gkalpak: You could also use the `they()` helpers for this test:

```js
they('should not attach itself to native $prop controls', {
  input: '<input ng-disabled="val">',
  textarea: '<textarea ng-disabled="val"></textarea>',
  select: '<select ng-disabled="val"></select>',
  button: '<button ng-disabled="val"></button>'
}, function(tmpl) {
  var element = $compile(tmpl)(scope);  
  scope.$apply('val = true');

  expect(element.attr('disabled')).toBeDefined();
  expect(element.attr('aria-disabled')).toBeUndefined();
});
```

test/ngAria/ariaSpec.js
```diff
@@ -127,7 +157,7 @@ describe('$aria', function() {
...
 
     it('should handle checkbox with integer model values using ngTrueValue', function() {
-      var element = $compile('<input type="checkbox" ng-model="val" ng-true-value="0">')(scope);
+      var element = $compile('<div role="checkbox" ng-model="val" ng-true-value="0">')(scope);
```
gkalpak: So, after some good time wondering how can this test be passing, I realized that `ng-true-value`/`ng-false-value` is only applied to `input` elements (so it has no effect here). This test only passed by chance.

It would fail for example, by changing `ng-true-value="0"` to `ng-true-value="2"` (and `val=0` to `val=2` below).
I am afraid there is no reliable way of determining when a custom checkbox is selected. I would stick to considering it not selected when `ngModel.$viewValue === false`. That is also consistent with `ngAria`'s `ngModelController.$isEmpty()` function. (See [here][1].)

[1]: https://github.com/angular/angular.js/pull/13483/files#diff-91e9129201746ba107348c9a0a7735edR248

src/ngAria/aria.js
gkalpak: See also https://github.com/angular/angular.js/pull/13483/files#r51712702.
Since `ngTrueValue`/`ngFalseValue` are only taken into account on `input` elements (and since `ngAria` doesn't set `aria-required` on `input` elements), we shouldn't take `ngTrue/FalseValue` into account.

docs/content/guide/accessibility.ngdoc
marcysutton: Good catch, although that should say `aria-required`

src/ngAria/aria.js
marcysutton: It was added to be consumed in a watch function on [line 282](https://github.com/angular/angular.js/pull/13483/files#diff-91e9129201746ba107348c9a0a7735edR282).

src/ngAria/aria.js
marcysutton: This support was added to address what I've seen in the wild–developers expect `ngTrueValue` and `ngChecked` to work on custom directives, yet they are only supported on native `input`. Adding support for common use cases requiring ARIA attributes was intentional after working closely with ngAria and Angular Material 1.

docs/content/guide/accessibility.ngdoc
gkalpak: That's what I said :stuck_out_tongue:

src/ngAria/aria.js
gkalpak: Why do you need to call a function that returns the actual function ? Just use `ngAriaCheckboxReaction` directly.
(The same is true for `ngAriaRadioReaction` - unless I'm missing something :smiley:)

src/ngAria/aria.js
gkalpak: `ngChecked` isn't restricted to `input` elements, but `ngTrue/FalseValue` are.

If they are trying to use `ngTrueValue` on a custom checkbox and it is not working (because it's not an `input`), I believe `aria-checked` is the least of their problems :)
Besides, this isn't working correctly (see https://github.com/angular/angular.js/pull/13938/files#diff-604b5030e9e44897f6ba0106fe4d28deR20) - it couldn't be, since we are assuming that `ngTrue/FalseValue` works (and compare against the parsed `ngTrueValue`), when in fact it has no effect.

src/ngAria/aria.js
marcysutton: Got it!

test/ngAria/ariaSpec.js
marcysutton: I don't think `aria-required` has to be removed if `ngRequired` is false, just set to `aria-required="false"`. It isn't a boolean attribute like `required`. I will add a test for that, though. https://www.w3.org/TR/wai-aria/states_and_properties#aria-required

gkalpak: You are right ! A test verifying it's set to `false` sounds good then :)

src/ngAria/aria.js
marcysutton: The use case was the [Angular Material Checkbox](https://material.angularjs.org/latest/demo/checkbox), which uses `ng-true-value` on the second demo. But it still works when I revert this back to the way it was (while still excluding `INPUT`). Therefore I think we should revert this check and avoid going further into the weeds.

docs/content/guide/accessibility.ngdoc
```diff
@@ -148,18 +172,37 @@ custom controls to be more accessible.
...
 ```
 
 >You can check whether a control is legitimately disabled for a screen reader by visiting
 [chrome://accessibility](chrome://accessibility) and inspecting [the accessibility tree](http://www.paciellogroup.com/blog/2015/01/the-browser-accessibility-tree/).
 
+<h2 id="ngrequired">ngRequired</h2>
+
+The boolean `required` attribute is only valid for native form controls such as `input` and
+`textarea`. To properly indicate custom element directives such as `<md-checkbox>` or `<custom-input>`
+as required, using ngAria with {@link ng/directive/ngrequired ngRequired} will also add
```
gkalpak: I'm not sure if `ng/directive/ngrequired` (instead of `ng/directive/ngRequired`) works.

src/ngAria/aria.js
```diff
@@ -16,19 +16,23 @@
...
+ * `ngModel`, `ngChecked`, `ngRequired`, `ngValue`, `ngDisabled`, `ngShow`, `ngHide`, `ngClick`,
+ * `ngDblClick`, and `ngMessages`.
  *
  * Below is a more detailed breakdown of the attributes handled by ngAria:
  *
  * | Directive                                   | Supported Attributes                                                                   |
  * |---------------------------------------------|----------------------------------------------------------------------------------------|
+ * | {@link ng.directive:ngModel ngModel}        | aria-checked, aria-valuemin, aria-valuemax, aria-valuenow, aria-invalid, aria-required, input roles |
  * | {@link ng.directive:ngDisabled ngDisabled}  | aria-disabled                                                                          |
+ * | {@link ng.directive:input ngRequired}       | aria-required                                                                          |
```
gkalpak: _:input_  -->  _:ngRequired_

test/ngAria/ariaSpec.js
```diff
@@ -414,49 +399,28 @@ describe('$aria', function() {
...
-      expect(element.attr('aria-required')).toBe('true');
-
-      scope.$apply("val='input is valid now'");
-      expect(element.attr('aria-required')).toBe('false');
+      expect(element.attr('aria-required')).toBeUndefined();
     });
 
-    it('should attach aria-required to textarea', function() {
-      compileElement('<textarea ng-model="val" required></textarea>');
+     it('should attach to custom controls with ngModel and required', function() {
```
gkalpak: Excess leading whitespace :)

gkalpak:
LGTM (the whitespace can be taken care of upon merging)

src/ngAria/aria.js
gkalpak: Wow, `mdCheckbox` does some [really hacky stuff][1] and applies the `input[type="checkbox"]` pre-linking function. That's how they get `ngTrue/FalseValue` to work for them.
Pretty interesting...

Anyway, the current implementation works with `ngTrue/FalseValue`, so we're good :)

[1]: https://github.com/angular/material/blob/ad9ba5277651423beceb7302197e069047a5105d/src/components/checkbox/checkbox.js#L102-L108

gkalpak:
@marcysutton, (a little too late) but I have some questions regarding dropping `aria-multiline`:

1. Does it mean that it is not necessary to put it on `<textarea>` elements or does the developer have to add it themselves ?
2. Isn't it possible to have custom textboxes (with `[role="textbox"]`) that accept multiline text. Couldn't/Shouldn't `ngAria` set `aria-multiline` in those cases ?

src/ngAria/aria.js
```diff
@@ -198,13 +203,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
-    return $aria.config(normalizedAttr) && !elem.attr(attr);
+  function shouldAttachAttr(attr, normalizedAttr, elem, allowBlacklistEls) {
+    return $aria.config(normalizedAttr) && !elem.attr(attr) && (allowBlacklistEls || !isNodeOneOf(elem, nodeBlackList));
   }
 
   function shouldAttachRole(role, elem) {
+    // if element does not have role attribute
+    // AND element type is equal to role (if custom element has a type equaling shape) <-- remove?
+    // AND element is not INPUT
     return !elem.attr('role') && (elem.attr('type') === role) && (elem[0].nodeName !== 'INPUT');
```
gkalpak: @marcysutton, why are we only checking against `input` ? Shouldn't all native controls (e.g. `textarea`, `select`, `button`) be excluded ?

marcysutton: Yeah, that line should be replaced with `isNodeOneOf(elem, nodeBlackList))`. It has already gone through the `getShape` function but that only checks `type` and `role`.

PR #13485, opened 2015-12-09
fix(Module): allow passing template/templateUrl in array notation

shahata:


gkalpak:
LGTM

src/loader.js
```diff
@@ -362,7 +362,7 @@ function setupModuleLoader(window) {
           component: function(name, options) {
             function factory($injector) {
               function makeInjectable(fn) {
-                if (angular.isFunction(fn)) {
+                if (isFunction(fn) || isArray(fn)) {
```
petebacondarwin: I am pretty sure we cannot rely on being in the closure of angular here.
Although the unit tests will pass, I don't think that `isFunction` and `isArray` will be available in practice.

shahata: `isFunction` is used further down in the same file, so it means we're fine, no? https://github.com/angular/angular.js/blob/master/src/loader.js#L454

petebacondarwin: I would like to see that working in practice before I agree... I wonder if that other use ever worked... :-)

petebacondarwin: That usage was added by an external committer... https://github.com/angular/angular.js/commit/351fe4b79c50a45a11af2fcd2aa7b6fd3b70058d

shahata: If that line didn't work every single service/directive/etc. you define would throw... (btw, it also has a bug of not handling array notation :smile:)

shahata: I also checked in debugger to make sure that it is indeed in the closure

gkalpak: It is also accessing `minErr` near the top of the file. It definitely relies on having access to some stuff. (Whether it _should_ is another story...)

shahata: Ah, but maybe angular-loader.js will have a problem, I'll make sure

Narretz:
I feel like this should be documented, no?

Narretz:
Also, what's the reason why component() itself is not injectable?

gkalpak:
`component()` does not accept a factory function, it gets the _CDO_ (Component Definition Object :smile:) directly.
Although, I've been thinking it might be convenient to overload it, in order to also accept a factory function.

petebacondarwin:
I think it is nice that we are not injecting into the CDO. Anything that needs injection can ask for it.

src/loader.js
lgalfaso: `isFunction` will work because of https://github.com/angular/angular.js/blob/master/src/loader.prefix#L8 `isArray` will not work

shahata: Does anybody have a strong opinion on whether to add `isArray` up there or just use `angular.isFunction` & `angular.isArray`? I don't really care

lgalfaso: if you just do `angular.isArray` you would break the loader as it is meant to be run without angular

Narretz: I think there's an open issue about the loader being broken for certain modules. Maybe that already happens?

petebacondarwin: `Array.isArray` is supported on all the browsers that we support in (outside of v1.2) (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_compatibility) so should just use that in this case...

petebacondarwin: We should definitely have a comment at the top of this file, explaining where we get `isFunction` from.

lgalfaso: @petebacondarwin in 99% of the cases (that is when people use angular and not just the loader), `isFunction` comes from `src/Angular.js` and I doubt that there is a need to explain this. For the 1% of cases that the loader is used stand-alone, then it comes from another place.

Narretz:
I'd still like to see a mention in the docs that you can you can inject any injectables into the template / templateUrl fn. This is far from obvious.

lgalfaso:
@Narretz do you want to open an issue to track that?

Narretz:
@lgalfaso I've added this to the docs just now, but I would really prefer if docs landed as part of the code changes.

Issue #13481, opened 2015-12-09
Injector error can be misleading during module instantiation


robstove:
During module loading there is an opportunity for failure in the runInvokeQueue() function. This can occur when the expected function on the provider does not exist. To provide additional details to the developer I propose checking for undefined value before executing the provider function. This will provide additional details about why the module failed to load.

src\auto\injector.js (starting at line 736)

      function runInvokeQueue(queue) {
        var i, ii;
        for (i = 0, ii = queue.length; i < ii; i++) {
          var invokeArgs = queue[i],
              provider = providerInjector.get(invokeArgs[0]);

          // Undefined provider method check.
          var providerFn = provider[invokeArgs[1]];
          if (providerFn === undefined) {
            throw new Error("Method " + invokeArgs[1] + " was not found for provider " + invokeArgs[0] + ".");
          }

          providerFn.apply(provider, invokeArgs[2]);
        }
      }

Narretz:
Sounds like a good addition. Would you like to open a PR?

robstove:
done

PR #13486, opened 2015-12-09
Handle undefined provider method during module loading

robstove:
https://github.com/angular/angular.js/issues/13481#issuecomment-163435774

### DIFF
--- a/docs/content/error/$injector/unprfn.ngdoc
+++ b/docs/content/error/$injector/unprfn.ngdoc
@@ -1,0 +1,10 @@
+@ngdoc error
+@name $injector:unprfn
+@fullName Undefined Provider Function
+@description
+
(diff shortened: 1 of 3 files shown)


robstove:
I signed it!

Narretz:
Thanks for the PR. There a few things that need to be done before this can be merged:
- [ ]  use the minErr fn instead of Error
- [ ] add a test that shows that this case is caught
- [ ] add a docs error page for this error

Would be cool if you could work on this @robstove If you have questions, feel free to ask.

robstove:
Add the error details however due to the location of this error I wasn't able to find a good way implement a valid test case.

robstove:
Assuming this isn't going anywhere. Closing....

PR #13488, opened 2015-12-10
[don't merge]: fix($animateCss): respect transition styles already on the element

### LINKED ISSUE #12656: ngAnimate assumes transition is linear instead of using the value in CSS

Narretz:
Previously, $animateCss wouldn't use transition styles that were on the element
before the animation process started. Precisely, transition property, timing-function
and delay were overwritten in the process.

Closes #12656
Closes #13333


src/ngAnimate/animateCss.js
```diff
@@ -588,17 +590,10 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
...
-      maxDuration = timings.maxDuration;
-
       var flags = {};
       flags.hasTransitions          = timings.transitionDuration > 0;
       flags.hasAnimations           = timings.animationDuration > 0;
-      flags.hasTransitionAll        = flags.hasTransitions && timings.transitionProperty == 'all';
-      flags.applyTransitionDuration = hasToStyles && (
-                                        (flags.hasTransitions && !flags.hasTransitionAll)
+      flags.applyTransitionDuration = options.duration > 0 || hasToStyles && (flags.hasTransitions
                                          || (flags.hasAnimations && !flags.hasTransitions));
```
gkalpak: I am not sure if this expressionis correct, but if it is, it can be simplified to:

`options.duration > 0 || hasToStyles && (flags.hasTransitions || flags.hasAnimations)`

src/ngAnimate/animateCss.js
```diff
@@ -576,8 +568,18 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
               : gcsLookup.count(cacheKey)
           : 0;
 
-      var isFirst = itemIndex === 0;
+      var timings = computeTimings(node, fullClassName, cacheKey);
+
+      if (options.duration > 0) {
```
gkalpak: What if `options.duration === 0` ?

src/ngAnimate/animateCss.js
gkalpak: Same as above: What if `options.duration === 0` ?
I think we want to support that as well.

test/ngAnimate/animateCssSpec.js
```diff
@@ -2790,7 +2867,7 @@ describe("ngAnimate $animateCss", function() {
         it("should apply a transition duration if the existing transition duration's property value is not 'all'",
           inject(function($animateCss, $rootElement) {
 
-          ss.addRule('.ng-enter', 'transition: 1s linear color');
+          ss.addRule('.ng-enter', 'transition: 1s cubic-bezier(0.25, 0.1, 0.25, 1) color');
```
gkalpak: Was this change necessary ?

test/ngAnimate/animateCssSpec.js
```diff
@@ -634,6 +634,86 @@ describe("ngAnimate $animateCss", function() {
...
+          var options = {
+            to: { background: 'blue' }
+          };
+
+          var animator = $animateCss(element, options);
+          animator.start();
+          triggerAnimationStartFrame();
+
+          // IE reports ease in cubic-bezier form
+          expect(element.css('transition-timing-function')).toBeOneOf('ease', 'cubic-bezier(0.25, 0.1, 0.25, 1)');
```
gkalpak: Nitpick: Since the `cubic-bezier` equivalents of `ease`/`ease-in`/`linear` are used is several tests (and it's not clear why without a comment, nor is it practical to replicate the comment everywhere), it would be nice to have a constant enumeration of the used values at the top and it throughout the tests (e.g. something like `CUBIC_BEZIER_EQUIVALENT.ease`).

test/ngAnimate/animateCssSpec.js
Narretz: Browsers have 'ease' as the default transition-timing value, but one browser, IE I think,reports them as cubic-bezier, all others as ease

Narretz: Ah, yes, but this particular change is probably not necessary.

src/ngAnimate/animateCss.js
Narretz: If the duration is 0, no animation is performed, and no styles are added at all .. but that is only tested for styles that are added to the element, not those that are already on the element

Narretz: The problem is that it's not entirely clear what $animateCss should do with styles that are already on the element when it is called. Should it ignore those or incorporate them? With all this addClass and to properties, it becomes really complex.

gkalpak: Yes, but what if the element has transition styles with `transition-duration > 0` whereas `options.duration === 0`.
E.g. the user wants to have a 0-duration animation (i.e. skip the animation for whatever reason), on an element that would normally animate (because it has `transition-duration` > 0).

Or is it so edgy a case, that we don't need to support it ? :smiley:

src/ngAnimate/animateCss.js
```diff
@@ -292,14 +293,14 @@ function truthyTimingValue(val) {
...
   var value = duration + 's';
-  if (applyOnlyDuration) {
-    style += DURATION_KEY;
-  } else {
-    value += ' linear all';
-  }
+
+  value += ' ' + timings[TRANSITION_TIMING_PROP];
+  value += ' ' + timings[TRANSITION_PROPERTY_PROP];
+  value += timings[TRANSITION_DELAY_PROP] ? ' ' + timings[TRANSITION_DELAY_PROP] + 's' : '';
```
gkalpak: Should these 3 lines use the keys of `DETECT_CSS_PROPERTIES` instead of the values ?
E.g. `timings.transitionDelay` instead of `timings[TRANSITIN_DELAY_PROP]` etc ?

src/ngAnimate/animateCss.js
```diff
@@ -223,12 +223,13 @@ var ELAPSED_TIME_MAX_DECIMAL_PLACES = 3;
...
-  transitionDuration:      TRANSITION_DURATION_PROP,
-  transitionDelay:         TRANSITION_DELAY_PROP,
-  transitionProperty:      TRANSITION_PROP + PROPERTY_KEY,
-  animationDuration:       ANIMATION_DURATION_PROP,
-  animationDelay:          ANIMATION_DELAY_PROP,
-  animationIterationCount: ANIMATION_PROP + ANIMATION_ITERATION_COUNT_KEY
+  transitionDuration:       TRANSITION_DURATION_PROP,
+  transitionDelay:          TRANSITION_DELAY_PROP,
+  transitionProperty:       TRANSITION_PROP + PROPERTY_KEY,
+  transitionTimingFunction: TRANSITION_PROP + TIMING_KEY,
```
gkalpak: These 2 lines shoould use the variables defined in `shared.js`: `TRANSITION_PROPERTY_PROP` and `TRANSITION_TIMING_PROP`

src/ngAnimate/animateCss.js
Narretz: The value in the options always takes precedence. I've tested it, whatever is set in the styles, if you options.duration = 0, no animation is run.

gkalpak: Cool then :)

src/ngAnimate/animateCss.js
Narretz: The "timings" object's keys are the actual style property names for each browser.

src/ngAnimate/animateCss.js
Narretz: Ah, I see the problem. You are right, if an element has transitions defined, but the user set options.duration: 0, then $animateCss will not animate, but that means the transition styles on the element will take over anyway. So it's actually not possible to overwrite an exisiting transition with duration: 0. @matsko do you think that should be possible?.

matsko:
Mobile Safari is still broken on this one.

petebacondarwin:
Moving to 1.5.x as it not going to make it into 1.5.0

PR #13489, opened 2015-12-10
fix(ngOptions): don't skip optgroup elements with value === ''

### LINKED ISSUE #13487: Select ng-options with "group by" does not clear list on update (IE11)
Having  a `<select>` and `ng-options` with a `group by` clause, after an update of the `list` (e.g. using a `filter`) the rendered list (of `<optgroup>`/`<option>`) isn't cleared but gets longer and longer after each update (IE11).

```html
<select size="20" ng-model="selected" 
  ng-options="i as i.name group by i.cls for i in list | filter:filter">
</select>
```

**Steps to reproduce**:
1. Open https://jsfiddle.net/gk8xeupm/ with IE11
2. Type "foo" into the filter text field
3. Clear text field
4. Type "bar" into the filter text field
5. Clear text field
6. Watch the list becoming longer and longer

**AngularJS**: 1.4.8
**Reproducable**: always
**Operating System**: Windows 10
**Browsers**: IE version 11.11.10586.0

Narretz:
Internet explorer 11 returns '' for optgroup elements without a value attribute.
We only want to skip option elements with value ''

Fixes #13487

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -643,7 +643,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
                 (current === emptyOption_ ||
                 current === unknownOption_ ||
                 current.nodeType === NODE_TYPE_COMMENT ||
-                current.value === '')) {
+                (nodeName_(current) === 'option' && current.value === ''))) {
             current = current.nextSibling;
           }
         }
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -1555,23 +1555,26 @@ describe('ngOptions', function() {
 
       expect(element).toEqualSelectValue(scope.selected);
 
-      var zero = jqLite(element.find('optgroup')[0]);
-      var b = jqLite(zero.find('option')[0]);
-      var e = jqLite(zero.find('option')[1]);
+      var optgroups = element.find('optgroup');
+      expect(optgroups.length).toBe(3);
+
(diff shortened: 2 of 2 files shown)


test/ng/directive/ngOptionsSpec.js
```diff
@@ -1584,6 +1626,79 @@ describe('ngOptions', function() {
...
+      expect(delta.text()).toEqual('Delta');
+
+      // New group
+      three = optgroups.eq(2);
+      expect(three.children('option').length).toBe(1);
+
+      epsilon = three.find('option').eq(0);
+      expect(three.attr('label')).toEqual('three');
+      expect(epsilon.text()).toEqual('Epsilon');
+    });
```
gkalpak: Nice test :)

src/ng/directive/ngOptions.js
```diff
@@ -643,7 +643,7 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
                 (current === emptyOption_ ||
                 current === unknownOption_ ||
                 current.nodeType === NODE_TYPE_COMMENT ||
-                current.value === '')) {
+                (current.nodeName == 'OPTION' && current.value === ''))) {
```
gkalpak: Nitpick1: Use `===` instead of `==`.
Nitpick2: Might be better to use `nodeName_(current) === 'option'` for consistency.

gkalpak:
Other than a couple of nitpicks, it LGTM (if Travis is happy) :+1:

Narretz:
Thx for the review @gkalpak

PR #13490, opened 2015-12-10
docs: revamp information about interpolation

Narretz:
- Move interpolation info from Directive guide into new interpolation guide
- Add information about boolean attributes to interpolation guide
- remove wroong examples from prefixed boolean attribute docs, link
  to interpolation guide instead
- mention additional examples for attributes that benefit from ngAttr

I felt this was a good idea because currently interpolation information has no real place in the guide. So these docs try to bring the available info together and add some other stuff that has come up recently (ngAttr, mixing expressions and interpolation)

### DIFF
--- a/docs/content/guide/interpolation.ngdoc
+++ b/docs/content/guide/interpolation.ngdoc
@@ -1,0 +1,142 @@
+@ngdoc overview
+@name Interpolation
+@sortOrder 275
+@description
+
+# Interpolation and data-binding
+
+Interpolation markup with embedded @link {guide/expressions expressions} is used by Angular to
+provide data-binding to text nodes and attribute values.
+
+An example of interpolation is shown below:
+
+```html
+<a ng-href="img/{{username}}.jpg">Hello {{username}}!</a>
+```
+
+### How text and attribute bindings work
+
+During the compilation process the {@link ng.$compile compiler} uses the {@link ng.$interpolate $interpolate}
+service to see if text nodes and element attributes contain interpolation markup with embedded expressions.
+
+If that is the case, the compiler adds an interpolateDirective to the node and
+registers {@link ng.$rootScope.Scope#$watch watches} on the computed interpolation function,
+which will update the corresponding text nodes or attribute values as part of the
+normal {@link ng.$rootScope.Scope#$digest digest} cycle.
+
+Note that the interpolateDirective has a priority of 100 and sets up the watch in the preLink function.
+
+### Binding to boolean attributes
+
+Attributes such as `disabled` are called `boolean` attributes, because their presence means `true` and
+their absence means `false`. We cannot use normal attribute bindings with them, because the HTML
+specification does not require browsers to preserve the values of boolean attributes. This means that
+if we put an Angular interpolation expression into such an attribute then the binding information
+would be lost, because the browser ignores the attribute value.
+
+In the following example, the interpolation information would be ignored and the browser would simply
+interpret the attribute as present, meaning that the button would always be disabled.
+
+ ```html
+  Disabled: <input type="checkbox" ng-model="isDisabled" />
+  <button disabled="{{isDisabled}}">Disabled</button>
+```
+
+For this reason, Angular provides special `ng`-prefixed directives for the following boolean attributes:
+{@link ngDisabled `disabled`}, [@link ngRequired `required`}, [@link ngSelected `selected`},
+{@link ngChecked `checked`}, {@link ngReadonly `readOnly`} , and [@link ngOpen `open`}.
+
+These directives take an expression inside the attribute, and set the corresponding boolean attribute
+to true when the expression evaluates to truthy.
+
+ ```html
+  Disabled: <input type="checkbox" ng-model="isDisabled" />
+  <button ng-disabled="isDisabled">Disabled</button>
+```
+
+### `ngAttr` for binding to arbitrary attributes
+
+Web browsers are sometimes picky about what values they consider valid for attributes.
+
+For example, considering this template:
+
+```html
+<svg>
+  <circle cx="{{cx}}"></circle>
+</svg>
+```
+
+We would expect Angular to be able to bind to this, but when we check the console we see
+something like `Error: Invalid value for attribute cx="{{cx}}"`. Because of the SVG DOM API's
+restrictions, you cannot simply write `cx="{{cx}}"`.
+
+With `ng-attr-cx` you can work around this problem.
+
+If an attribute with a binding is prefixed with the `ngAttr` prefix (denormalized as `ng-attr-`)
+then during the binding it will be applied to the corresponding unprefixed attribute. This allows
+you to bind to attributes that would otherwise be eagerly processed by browsers
+(e.g. an SVG element's `circle[cx]` attributes). When using `ngAttr`, the `allOrNothing` flag of
+{@link ng.$interpolate $interpolate} is used, so if any expression in the interpolated string
+results in `undefined`, the attribute is removed and not added to the element.
+
+For example, we could fix the example above by instead writing:
+
+```html
+<svg>
+  <circle ng-attr-cx="{{cx}}"></circle>
+</svg>
+```
+
+If one wants to modify a camelcased attribute (SVG elements have valid camelcased attributes),
+such as `viewBox` on the `svg` element, one can use underscores to denote that the attribute to bind
+to is naturally camelcased.
+
+For example, to bind to `viewBox`, we can write:
+
+```html
+<svg ng-attr-view_box="{{viewBox}}">
+</svg>
+```
+
+The following attributes are also known to cause problems when used with normal bindings:
+
+- **size** in `<select>` elements (see [Github issue 1619](https://github.com/angular/angular.js/issues/1619))
+- **placeholder** in `<textarea>` in Internet Explorer 10/11 (see [Github issue 5025](https://github.com/angular/angular.js/issues/5025))
+
+
+### Embedding interpolation markup inside expressions
+
+Angular directives take either expressions or interpolation markup with embedded expressions. So the
+following example which embeds interpolation inside an expression is a bad practice:
+
+```html
+<div ng-show="form{{$index}}.$invalid"></div>
+```
+
+You should instead delegate the computation of complex expressions to the scope, like this:
+
+```html
+<div ng-show="getForm($index).$invalid"></div>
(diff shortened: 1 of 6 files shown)


docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+
+ ```html
+  <div ng-init="isDisabled = false">
+    <button disabled="{{isDisabled}}">Disabled</button>
+  </div>
+```
+
+For this reason, Angular provides special `ng`-prefixed directives for the following boolean attributes:
+{@link ngDisabled `disabled`}, [@link ngSelected `selected`}, {@link ngChecked `checked`},
+{@link ngReadonly `readOnly`} , and [@link ngOpen `open`}.
```
gkalpak: `ngMultiple` and of course `ngRequired` :)

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+
+These directives take an expression inside the attribute, and set the corresponding boolean attribute
+to true when the expression evaluates to truthy.
+
+ ```html
+  <div ng-init="isDisabled = false">
```
gkalpak: Woohoo, an anti-pattern :)

How about: 

```html
<div>
  Disabled: <input type="checkbox" ng-model="isDisabled" />
  <button ng-disabled="isDisabled">Click me</button>
</div>
```

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+### How text and attribute bindings work
+
+During the compilation process the {@link ng.$compile compiler} uses the {@link ng.$interpolate $interpolate}
+service to see if text nodes and element attributes contain interpolation markup with embedded expressions.
+If that is the case, the compiler adds an interpolationDirective to the node and
+registers {@link ng.$rootScope.Scope#$watch watches} on the computed interpolation function,
+which will update the corresponding text nodes or attribute values as part of the
+normal {@link ng.$rootScope.Scope#$digest digest} cycle.
+
+### Binding to boolean attributes
```
gkalpak: This section is very useful for sure, but I don't think it is related to interpolation.
If there isn't a better place to put it though, it's much better here than nowhere :)

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+you to bind to attributes that would otherwise be eagerly processed by browsers
+(e.g. an SVG element's `circle[cx]` attributes). When using `ngAttr`, the `allOrNothing` flag of
+{@link ng.$interpolate $interpolate} is used, so if any expression in the interpolated string
+results in `undefined`, the attribute is removed and not added to the element.
+
+For example, we could fix the example above by instead writing:
+
+```html
+<svg>
+  <circle ng-attr-cx="cx"></circle>
```
gkalpak: Shouldn't it be `ng-attr-cx="{{ cx }}"` ?

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+```
+
+If one wants to modify a camelcased attribute (SVG elements have valid camelcased attributes),
+such as `viewBox` on the `svg` element, one can use underscores to denote that the attribute to bind
+to is naturally camelcased.
+
+For example, to bind to `viewBox`, we can write:
+
+```html
+<svg ng-attr-view_box="viewBox">
```
gkalpak: _viewBox_  -->  _{{ viewBox }}_

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+</svg>
+```
+
+The following attributes are also known to cause problems with normal bindings:
```
gkalpak: Could this be more clear ? What do you mean _"cause problems with normal bindings"_.

I guess it means one can't use interpolation directly on the attribute value and should use `ngAttr` instead, but could we be more explicit ?

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+
+- **size** in `<select>` elements
+- **placeholder** in `<textarea>` in Internet Explorer 10/11
+
+
+### Embedding interpolations inside expression
```
gkalpak: Not sure if it was intentional, but `expressions` sounds better to be.

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+
+Angular directives take either expressions or interpolation markup with embedded expressions. So the
+following example which embeds interpolation inside an expression is a bad practice:
+
+```html
+<div ng-messages="items[{{index}}]"></div>
+```
+
+You should instead delegate the computation of complex expressions to the scope, like this:
```
gkalpak: Like what ?

docs/content/guide/interpolation.ngdoc
```diff
@@ -0,0 +1,125 @@
...
+The following attributes are also known to cause problems with normal bindings:
+
+- **size** in `<select>` elements
+- **placeholder** in `<textarea>` in Internet Explorer 10/11
+
+
+### Embedding interpolations inside expression
+
+Angular directives take either expressions or interpolation markup with embedded expressions. So the
+following example which embeds interpolation inside an expression is a bad practice:
```
gkalpak: I would take it a step further and say that it might not work as expected (and even if it works now, it might break in the future).

gkalpak: Sorry, I just saw it is mentioned below.

src/ng/compile.js
```diff
@@ -1228,7 +1228,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
        * @param {string} key Normalized key. (ie ngAttribute) .
        * @param {function(interpolatedValue)} fn Function that will be called whenever
                 the interpolated value of the attribute changes.
-       *        See the {@link guide/directive#text-and-attribute-bindings Directives} guide for more info.
+       *        See the {@link guide/interpolation#how-text-and-attribute-bindings-work Directives}
```
gkalpak: _Directives_  -->  _Interpolation_ (?)

docs/content/guide/interpolation.ngdoc
gkalpak: OK, it was copied over - but I would still change it.

src/ng/compile.js
gkalpak: You could also put "guide" inside the curly braces for consistency with other occurrences.

src/ng/directive/attrs.js
```diff
@@ -248,13 +229,12 @@
...
  * @description
- * The HTML specification does not require browsers to preserve the values of boolean attributes
- * such as readonly. (Their presence means true and their absence means false.)
- * If we put an Angular interpolation expression into such an attribute then the
- * binding information would be lost when the browser removes the attribute.
- * The `ngReadonly` directive solves this problem for the `readonly` attribute.
- * This complementary directive is not removed by the browser and so provides
- * a permanent reliable place to store the binding information.
+ *
+ * Sets the `readOnly` attribute on the element, if the expression inside `ngReadOnly` is truthy.
```
gkalpak: _ngReadOnly_  -->  _ngReadonly_

gkalpak:
Nice change !

A few suggestions (not directly related with this PR):

1. Add documentation for `ngRequired` and `ngMultiple` as standalone directives.
 2. Maybe remove the `ng-` prefixed versions of boolean attributes from the lists of attributes supported by the `input`/`select`/`textarea` directives. It's not like the controls make anything special to support the `ng`-prefixed versions.

docs/content/guide/interpolation.ngdoc
Narretz: Yes, it's not strictly related. But one of the uses of interpolation is setting dynamic attribute values. And in that way it's relevant I'd say.

docs/content/guide/interpolation.ngdoc
Narretz: ngMultiple doesn't exist actually. At least not in attrs.js (also no docs). I think this is because binding to multiple in select is not supported, and in inputs it's not yet supported.

+1 for ngRequired

docs/content/guide/interpolation.ngdoc
Narretz: There are issues for these two, maybe I can just link them?

docs/content/guide/interpolation.ngdoc
Narretz: +1

docs/content/guide/interpolation.ngdoc
Narretz: Oops!

src/ng/compile.js
Narretz: +1

docs/content/guide/interpolation.ngdoc
Narretz: Yes. I thought ngAttr works like the other ng- prefixed directives, but it's a separate mechanism.

docs/content/guide/interpolation.ngdoc
gkalpak: Works for me (for lack of a better suggestion :smiley:).

docs/content/guide/interpolation.ngdoc
gkalpak: Right. It is a `BOOLEAN_ATTR`, but I missed that we explicitly skip `multiple` in 'attrs.js'. So no `ngMultiple`.

docs/content/guide/interpolation.ngdoc
gkalpak: Probably works for me.

Narretz:
Yes, the docs for ngRequired (and also ngMinlength and ngMaxlength) are overdue. They weren't always standalone directives, so that's why they are missing.

Narretz:
I've updated the docs and addressed your comments, and also added docs for ngRequired. PTAL. Removing the ng- prefixed directives from hte inputs - I thought about it, but to be consistent we would need to remove the unprefixed attributes too, because they invoke the same directive.

Narretz:
Landed as https://github.com/angular/angular.js/commit/8863836cd4b82c8f0f9d72de4323aace28402c38

Issue #13391, opened 2015-11-27
Node is not defined


rodoabad:
I'm getting this error when I'm trying to run Angular in jsdom.

```D:\Web\GitHub\angular-test-examples\node_modules\angular\angular.js:2789
var jqLiteContains = Node.prototype.contains || function(arg) {
                     ^

ReferenceError: Node is not defined
```

```javascript
'use strict';

var jsdom = require('jsdom').jsdom;

var document = global.document = jsdom('<html><head></head><body></body></html>');
var window = global.window = document.defaultView;
global.navigator = window.navigator = {};

window.mocha = {};
window.beforeEach = global.beforeEach;
window.afterEach = global.afterEach;

require('angular');
```

Narretz:
Doesn't that mean jsdom doesn't provide the Node global?

rodoabad:
Yeah. I provided Node and it worked.

neha-nsharma:
I am new to this, how do you set Node in jsdom?

anotomix:
Just install lower than 1.4.8 angular version, then it should work.

mato75:
Is there a fix for angular 1.5.7

gkalpak:
Just make sure `Node` is available on `window`.

ankakusu:
@gkalpak Can you give an example for that? How do we be sure that `Node` is available on `window`?

gkalpak:
@ankakusu , it depends on where the problem comes from. Usually `window.Node` is available in all browser environments.

E.g. @mato75's problem was that they were trying to use Angular in the context of a web worker (see https://github.com/angular/angular.js/issues/13442#issuecomment-233558421). Other people had similar problems when using `jsDom` (and not configuring it properly.

Normally, there shouldn't be an issue with `window.Node`. If you have problems, you need to find out what causes it.

PR #13492, opened 2015-12-10
fix(jqLite): do not break if `Node` is not available

### LINKED ISSUE #13442: Uncaught ReferenceError: Node is not defined
angular.js:2789 Uncaught ReferenceError: Node is not defined

Updated from 1.4.7 to 1.4.8

gkalpak:
Fixes #13442

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -256,9 +256,10 @@ function jqLiteParseHTML(html, context) {
 
 
 // IE9-11 has no method "contains" in SVG element and in Node.prototype. Bug #10259.
-var jqLiteContains = Node.prototype.contains || function(arg) {
+var jqLiteContains = function(parentNode, childNode) {
   // jshint bitwise: false
-  return !!(this.compareDocumentPosition(arg) & 16);
+  return parentNode.contains ? parentNode.contains(childNode)
+                             : !!(parentNode.compareDocumentPosition(childNode) & 16);
   // jshint bitwise: true
 };
 
@@ -824,7 +825,7 @@ function specialMouseHandlerWrapper(target, event, handler) {
   var related = event.relatedTarget;
   // For mousenter/leave call the handler if related is outside the target.
   // NB: No relatedTarget if the mouse left/entered the browser window
-  if (!related || (related !== target && !jqLiteContains.call(target, related))) {
+  if (!related || (related !== target && !jqLiteContains(target, related))) {
     handler.call(target, event);
   }
 }


gkalpak:
Does anyone have any idea how/if we can test that it works if `Node` is not availalbe on the global context ?

petebacondarwin:
Did we identify for which browser this breaks? If so then we should be able to reproduce the problem in a unit test easily enough by including that browser in our testing...

gkalpak:
If I understood the usecase in #13442 correctly, this isn't fixing a browser issue, but some specific setup issue (with requirejs, maybe server-side stuff), where `Node` was not available in the global context. I guess we can't replicate that in tests. 

The original fix that was solved by 941c1c3 (which introduced the dependecy on `Node`) was affecting IE9-11. I tried on IE11 and there are indeed unit tests that fail without that fix, but pass with this commit.

We are already testing against IE9-11, so we would be good to go in that respect.

petebacondarwin:
I would rather not do the check on every call to `jqLiteContains`...  How about the following?

```js
var jqLiteContains = $window.Node && $window.Node.prototype.contains || function(arg) {
...
```

gkalpak:
I thought about that too, but I wanted to use the "native" `contains` as much as possible. E.g. when `Node` is undefined at the time of defining `jqLiteContains` (for whatever reason, which isn't necessarily that `.contains()` is not supported by the environment), it will use our custom method and not the native one (even if it's available on `target`).

Besides, in the majority of cases (where `Node.prototype.contains` will be available), you will still be calling `jqLiteContains.call(target, ...)` which I figured wouldn't be much of an optimization over `target.contains && target.contains(...)`. (I am just making assumptions here, I didn't run any benchmarks, since it isn't a critical path anyway).

--
But, yeah, I could have gone with your version, I just had to pick one.
I'm fine changing it.

petebacondarwin:
I don't understand when `window.Node` could be `undefined` only temporarily.

gkalpak:
I'm not sure either (maybe @jlmitch5 or @neha-nsharma can shed more light on this).
What I suspect is that (somehow) the `window` is set up at a later time (after `angular.js` has been executed). See https://github.com/angular/angular.js/issues/13442#issuecomment-162537547.

Maybe using a JS DOM implementation for testing (e.g. [jsdom][1]).

[1]: https://github.com/tmpvar/jsdom

gkalpak:
I signed it (?)

petebacondarwin:
I am loathe to land this without a more concrete example of where this is failing.

gkalpak:
@petebacondarwin, it was also reported in #13391. It seems indeed to be failing with `jsdom` (among other things ?).

petebacondarwin:
It looks to me that you can define a window object via jsdom and add it to the global context (https://github.com/tmpvar/jsdom#for-the-hardcore-jsdomjsdom and https://github.com/tmpvar/jsdom#creating-a-browser-like-window-object).

and that Node ought to be available on that window (https://github.com/tmpvar/jsdom/blob/2184995dc6531cb92d219497c8f88c9b7f088bc1/lib/jsdom/living/node.js).

If not then I would raise a bug in jsdom.

Can anyone provide an actual running example of this problem?

gkalpak:
@petebacondarwin, according to https://github.com/angular/angular.js/issues/13442#issuecomment-165174783 (and subsequent discussion), it seems that changing `Node.prototype.contains` to `window.Node.ptototype.contains` solves the issue.

It sounds a reasonable change to me. WDYT ?

petebacondarwin:
I need to see an actual real running example where this is breaking.

petebacondarwin:
Closing in favour of a deeper investigation from @mgol of the other possible ways that using `jsdom` or similar could break AngularJS 1.

PR #13493, opened 2015-12-10
fix(ngInclude): add a return in a promise handler

lionep:
Since bluebird 3, when a promise is created within another promise handler, and nothing is returned from it, the library prints a warning.

See related issue https://github.com/petkaantonov/bluebird/issues/846

### DIFF
--- a/src/ng/directive/ngInclude.js
+++ b/src/ng/directive/ngInclude.js
@@ -252,6 +252,7 @@ var ngIncludeDirective = ['$templateRequest', '$anchorScroll', '$animate',
 
               currentScope.$emit('$includeContentLoaded', src);
               scope.$eval(onloadExp);
+              return null;
             }, function() {
               if (thisChangeId === changeCounter) {
                 cleanupLastIncludeContent();


lionep:
I signed it

lionep:
I signed it!

gkalpak:
You should add a comment explaining what's the purpose of this return, or it is very likely to be "optimized away". 

I am pretty confident this new bluebird 3.x behaviour will affect other places in the codebase.

Narretz:
So this only happens if you replace $q with bluebird. We could add a comment or a test for this, but it's a bit hm.

Is the warning part of a newer promise spec?

gcoupelant:
Yes, it happens when $q is overridden with Bluebird, with [angular-bluebird-promise](https://github.com/mattlewis92/angular-bluebird-promises) for example.
It's not a Promise spec, but Bluebird specific, [as explained here](http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-none-were-returned-from-it).

In our case, the promise is created by `$animate.enter` [line 247](https://github.com/lionep/angular.js/blob/patch-1/src/ng/directive/ngInclude.js#L247).

Narretz:
I'm gonna close this as we do not officially support bluebird as a replacement promise library. Should we do so in the future, we will have to address this more thoroughly in the whole code base anyway.

PR #13495, opened 2015-12-10
fix(ngAnimate): do not use event.timeStamp anymore for time tracking

### LINKED ISSUE #13494: Do not use event.timeStamp anymore
With the forthcoming releases for Firefox 44 and new versions of Chromium and Webkit, the use of `event.timeStamp` will yield a different value compared to what it did before. Therefore for this reason, all versions of ngAnimate will make use of `Date.now()` to leverage timing values within animations.

Original bug report:
https://bugzilla.mozilla.org/show_bug.cgi?id=1231619#c3

Changes to Chromium:
https://code.google.com/p/chromium/issues/detail?id=160524

matsko:
Due to recent changes in Chrome, Firefox and Webkit use of the
event.timeStamp value will lead to unpredictable behaviour due to
precision changes. Therefore it's best to stick entirely to use
`Date.now()` when it comes to confirming the end of transition-
ending values. See #13494 for more info.

Applies to 1.2, 1.3, 1.4 and 1.5.

Closes #13494

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -792,7 +792,10 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       function onAnimationProgress(event) {
         event.stopPropagation();
         var ev = event.originalEvent || event;
-        var timeStamp = ev.$manualTimeStamp || ev.timeStamp || Date.now();
+
+        // we now always use `Date.now()` due to the recent changes with
+        // event.timeStamp in Firefox, Webkit and Chrome (see #13494 for more info)
+        var timeStamp = ev.$manualTimeStamp || Date.now();
 
         /* Firefox (or possibly just Gecko) likes to not round values up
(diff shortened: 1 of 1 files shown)


petebacondarwin:
LGTM

majido:
It may actually be best to use ```performance.now()``` instead of ```Date.now()```. That provides a more accurate time and more importantly perhaps it is not subject to systems time skews making it more appropriate to measure time periods such as this. That means a bit more work to to change the startTime and perhaps some of your tests.

petebacondarwin:
@majido - Although since `Performance.now()` is not available on all the platforms we support, we would have to have a fallback to `Date.now()` anyway.

Narretz:
The only browser who doesn't support it is IE9 (obviously). But we can still switch to performance.now later.

petebacondarwin:
Let's go with `Date.now()` for the moment and reconsider `Performance.now()` if issues appear.

Issue #13496, opened 2015-12-10
CSS animations not completed properly on Chrome 49 Canary and FF nightly


majido:
**Context**: There is an ongoing effect to change Event.timeStamp from epoch time to a high-res timestamp. This meas that Event.timeStamp is no longer going to be comparable to Date.now() due to different time origins but rather may be compared to performance.now(). This feature is currently enabled by default in [Chrome 49 Canary](https://www.chromestatus.com/features/5523910145605632) (all platforms) and Firefox 44 developer edition (on certain platforms).

**Issue**: In angular, the logic for triggering animation completion for css animations depends on comparing Date.now() (i.e., startTime) with incoming animation event timestamp which is causing issues for certain types of animations ([example](http://plnkr.co/edit/Tr1nd5r0gAyy3eH3Z7Zh?p=preview).

Here is the original report on [Firefox bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1231619#c3) which has some additional details.

**Potential fix**: A simple fix is to use Date.now() instead of event.timeStamp [here](https://github.com/angular/angular.js/blob/20604e7fc4f69ecfafbd8d0c1fdc70d478075c3a/src/ngAnimate/animateCss.js#L808). This is already the fallback option which is in fact being used for Firefox (FF has 0 for timestamp!). It should not have any really impact given that the current Event.timestamp is not much different from Date.now() when event handler is invoked.

petebacondarwin:
Duplicate of #13494

PR #13497, opened 2015-12-10
fix: Use Date.now() instead of Event.timeStamp to detect animation end

majido:
Chrome and Firefox are planning to change event.timeStamp to be a
high resolution time instead of epoch time. This means that the
event.timeStamp will not longer be compatible with Date.now()[1, 2]

Instead we can use Date.now() at the moment when the event is
received by event listener instead.

This addresses  https://github.com/angular/angular.js/issues/13496

[1] http://crbug.com/538600
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1231619#c3


petebacondarwin:
Duplicate of #13495 
Sorry and thanks @majido

PR #13498, opened 2015-12-10
docs($routeProvider): document `resolveAs` and assiging `resolve` map on scope

gkalpak:
Related to #13400.


shahata:
It might be nice to explain that the reasoning behind this feature is to allow passing the resolved locals as attributes to components. Short example:
```js
$routeProvider.when('/', {
  template: '<my-app items="resolve.items" />',
  resolve: {items: ($http) => $http.get(…)}
});
```

gkalpak:
Good suggestion, @shahata. I added another commit with examples with components and other minor improvements. PTAL

PR #13502, opened 2015-12-11
Update animateCss.js to not remove events from elements

josephperrott:
animateCss currently removes all events from and element during close().  We should check if any animationend events have occured by checking the length of the events array, rather than its truthiness.

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -750,7 +750,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
         }
 
         // Remove the transitionend / animationend listener(s)
-        if (events) {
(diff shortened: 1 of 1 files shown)


gkalpak:
@josephperrott, thx for submitting this. Before we can merge it though, there needs to be a test that shows that the issue is fixed and prevent future regressions (plus what @googlebot said regarding the [CLA][1]).

[1]: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#cla

Narretz:
@josephperrott How did you find this bug? The code only removes all events if the animation hasn't started yet. In all other cases it works.

Narretz:
And again with a test: https://github.com/angular/angular.js/pull/13514

PR #13504, opened 2015-12-11
revert: fix($animateCss): remove animation end event listeners on close

IgorMinar:
This reverts commit c98e08fd871fa5df39c66e9700345c2ae0f21bc0.

This commit was identified as incompatible with ng-material at Google
and is causing broken builds there. Proper fix to be investigated once
the immediate regression is addressed.

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -474,8 +474,6 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       var maxDelayTime;
       var maxDuration;
       var maxDurationTime;
-      var startTime;
-      var events = [];
 
       if (options.duration === 0 || (!$sniffer.animations && !$sniffer.transitions)) {
         return closeAndReturnNoopAnimator();
@@ -749,11 +747,6 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
           options.onDone();
         }
 
-        // Remove the transitionend / animationend listener(s)
-        if (events) {
-          element.off(events.join(' '), onAnimationProgress);
-        }
-
         // if the preparation function fails then the promise is not setup
         if (runner) {
           runner.complete(!rejected);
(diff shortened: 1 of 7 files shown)


IgorMinar:
do not merge anything into 1.4.x branch including this PR.

IgorMinar:
I'll merge this in when we are ready.

@petebacondarwin, @Narretz, @gkalpak please cherry pick the test fix to master and verify why the lint check didn't catch ddescribe during CI.

gkalpak:
Looking into the `ddescribe` issue.

gkalpak:
(Indeed, it was merged _without_ `ddecribe` in master (6858caf), but _with_ `ddescribe` in v1.4.x (85e392f)).

For some reason, I can't seem to find any travis jobs forthe commits between 20604e7 and 8f0b482 (on master) and between 33cc75e and 1329e0f (on v1.4.x). If I am not missing something, this means that there was no "CI" for the following commits:

**master:** 6858caf, bc41ad8, 6428ed5
**v1.4.x:** 7caf913, 0c1b54f, 85e392f, c98e08f (these include the commit that introduced the `ddescribe`)

--
The 2 builds on v1.4.x after the missed/skipped commits should have caught the `ddescribe` error, but they didn't get to run the `ci-checks` task, because they errored one step before that (on the `test:unit` task) because of timing out issues.

Maybe it is related to this Travis issue: https://www.traviscistatus.com/incidents/b3xhcrr7g46d

gkalpak:
@IgorMinar, by _"please cherry pick the test fix to master"_ I assume you only meant the `ddescribe` test fix, not the reverted `$animateCss` tests. If you meant something else, just lmk ;)

src/ngAnimate/animateCss.js
```diff
@@ -749,11 +747,6 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
           options.onDone();
         }
 
-        // Remove the transitionend / animationend listener(s)
-        if (events) {
-          element.off(events.join(' '), onAnimationProgress);
```
gkalpak: Nt sure if it is related to the "breakage", but this line can cause trouble if `events` is an empty array. This is because of how `jqLiteOn` and `jqLiteOff` work (I didn't look into jQuery's implementation).

With an empty `events` array, we will be calling `element.on('', onAnimationProgress)` and `element.off('', onAnimationProgress)`. Due to how `.on()` and `.off()` are implemented, these two calls will result in:

1. `.on(...)`: Register a listener for the `''` (empty string) event (won't ever fire - but that's fine).
2. `.off(...)`: Deregister **all** listeners, because: `.off('', ...)` <==> `.off()`.
    `jqLiteOff` only check if the `event` is truthy (and `''` isn't) and if not, it thinks it is a no-argument call and deregisters all listeners.

gkalpak: If that is indeed the issue, I think it makes more sense to fix `jqLiteOff` instead of fixing it here.
BTW, prior to `v1.3.0-beta.19`, `jqLiteOff` did the right thing (using `isUndefined(type)` instead of `!type`), until someone ["optimized" it][1] &trade; :stuck_out_tongue: 

[1]: https://github.com/angular/angular.js/commit/abb17cce8b459e4646d1c2a2428b691c3d95fb4c

petebacondarwin:
Good work @gkalpak .

src/ngAnimate/animateCss.js
gkalpak: DISCLAIMER: I haven't investigated if `events` _can_ be empty (or if it will always have atleast one entry), but there's a good chance it can, since someone already noticed it: #13502.

Narretz:
Thanks @gkalpak. If jquery also removes all events when argument is '' then there's no use in fixing it in jqlite. I'll check.

gkalpak:
Yes, we needs to verify what jQuery does (but I would be very surprised if they did it as `jqLiteOff`).

Narretz:
The results are in ... jquery doesn't remove any events if .off is called with empty string. So we need to fix jqlite, I'm gonna do that now. We can of course also make the code in animateCss more explicit. Wdyt?

petebacondarwin:
@Narretz thanks

Narretz:
Scratch that, actually, jQuery also removes all events when called with '', so the fix must go directly into animateCss

IgorMinar:
ok. this is good to be merged now, but there are some commits in the v1.4.x branch that might need to be reverted before we land this PR. commits to revert (in this order): 194ade2, 1329e0f.

I talked to @petebacondarwin about this and either him or I will do this cleanup next week. We are not under time pressure to do this as long as v1.4.x remains frozen.

Narretz:
@IgorMinar I have a fix for the animation issue, so it doesn't necessarily have to be reverted. Or is there another reason why we can't just fix it?

gkalpak:
@IgorMinar, are you sure 194ade2 is the one that needs to be reverted ?

IgorMinar:
yes - the two commits listed are in v1.4.x branch but not in google3. g3
now contains an emergency rollback (revert) of the animation commit in this
PR.

Do not touch v1.4.x - pete or me will do the cleanup next week.

IgorMinar:
landed as 194ade2...bab3069 (I had to force push to get g3 and github in sync)

gkalpak:
Is `v1.4.x` still "frozen" ?

PR #13507, opened 2015-12-11
Bugfix - Typos

jutaz:
This PR fixes few typo found around the place

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ install:
   # Log HTTP requests
   - npm config set loglevel http
   - npm install -g npm@2.5
-  # Instal npm dependecies and ensure that npm cache is not stale
+  # Install npm dependencies and ensure that npm cache is not stale
   - npm install
 
 before_script:
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -167,7 +167,7 @@ var ANIMATE_TIMER_KEY = '$$animateCss';
  * ```
  *
  * To actually start the animation we need to run `animation.start()` which will then return a promise that we can hook into to detect when the animation ends.
- * If we choose not to run the animation then we MUST run `animation.end()` to perform a cleanup on the element (since some CSS classes and stlyes may have been
+ * If we choose not to run the animation then we MUST run `animation.end()` to perform a cleanup on the element (since some CSS classes and styles may have been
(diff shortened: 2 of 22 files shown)


jutaz:
I signed it!

.travis.yml
```diff
@@ -47,7 +47,7 @@ install:
   # Log HTTP requests
   - npm config set loglevel http
   - npm install -g npm@2.5
-  # Instal npm dependecies and ensure that npm cache is not stale
+  # Instal npm dependencies and ensure that npm cache is not stale
```
Narretz: There' also *instal* -> *install*

jutaz: Haha, missed this one!

jutaz:
Amended to include `Instal` -> `Install`

src/ngAnimate/animateCss.js
```diff
@@ -261,7 +261,7 @@ function computeCssStyles($window, element, properties) {
       }
 
       // by setting this to null in the event that the delay is not set or is set directly as 0
-      // then we can still allow for zegative values to be used later on and not mistake this
+      // then we can still allow for negative values to be used later on and not mistake this
```
gkalpak: Maybe `zegative` means `zero or negative` :stuck_out_tongue:

jutaz: But from what I understand, zero values are being set to `null`, right? So in that case zero is not allowed :wink:

gkalpak: Yes, but it could be bug in code and not a typo in docs :stuck_out_tongue:
Just kidding...

gkalpak:
Nice one !

I favour (pun intended) `cancelled` / `behaviour` over `canceled` / `behavior`, but that's unimportant :)

src/ngAnimate/animateCss.js
jutaz: Haha, that would be bad

jutaz:
> I favour (pun intended) cancelled / behaviour over canceled / behavior, but that's unimportant :)

Yeah, but given that this project is `EN_us`, it makes sense, as `behaviour` and `cancelled` are more from `EN_gb` locale :wink:

gkalpak:
BTW this is not "backportable" to `v1.4.x`.
If you care to resolve the conflicts and submit a PR against the `v1.4.x` branch (since you seem to be really serious about typos), I'd be happy to merge :smiley:

Thx for all the typo crashing so far :+1:

PR #13511, opened 2015-12-11
fix(jqLite): don't deregister all listeners when calling off with ''

Narretz:
.off should only deregister all events when the first argument is undefined.

### DIFF
--- a/test/jqLiteSpec.js
+++ b/test/jqLiteSpec.js
@@ -1345,6 +1345,29 @@ describe('jqLite', function() {
       expect(mouseoverSpy).not.toHaveBeenCalled();
     });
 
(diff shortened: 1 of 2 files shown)


test/jqLiteSpec.js
```diff
@@ -1345,6 +1345,29 @@ describe('jqLite', function() {
       expect(mouseoverSpy).not.toHaveBeenCalled();
     });
 
+    it('should not deregister any listeners when called with empty string', function() {
```
Narretz: Should we test for other falsy values, too? (false, null)

Narretz:
Actually, jQuery also removes all event when called with '', so the fix must go directly into animateCss

PR #13513, opened 2015-12-11
fix(ngMock): trigger digest with `$httpBackend.verifyNoOutstandingRequest`

### LINKED ISSUE #13506: $httpBackend.verifyNoOutstandingRequest fails to throw when digest is not triggered
**Overview of the Issue**
In unit test, I use `$httpBackend.verifyNoOutstandingRequest();` to make sure that no request was done. The test also passes when the request is done, because `verifyNoOutstandingRequest` does not trigger digest (in contrary to `flush` and `verifyNoOutstandingExpectation`) which leads to unpropagated [_Unexpected request_](https://github.com/angular/angular.js/blob/75e876424da5f569481488d03cf3a61441341513/src/ngMock/angular-mocks.js#L1237) error.
There is a [comment](https://github.com/angular/angular.js/commit/ea6fc6e69c2a2aa213c71ed4e917a0d54d064e4c#commitcomment-11571705) asking whether calling digest should be added to `verifyNoOutstandingRequest` as well, but with no response.

**Use Case** - testing that `$http` request is _not_ being done
**Angular Version** - 1.4.8
**Reproduce the Error**
```js
    it('should not allow submitting invalid form', function() {
        var ctrl = createCtrl();
        ctrl.submit();
        $httpBackend.verifyNoOutstandingRequest();
    });
```

katemihalikova:
To be consistent with `$httpBackend.verifyNoOutstandingExpectation` and `$httpBackend.flush`
Fixes #13506

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1742,7 +1742,8 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
    *   afterEach($httpBackend.verifyNoOutstandingRequest);
    * ```
    */
-  $httpBackend.verifyNoOutstandingRequest = function() {
+  $httpBackend.verifyNoOutstandingRequest = function(digest) {
+    if (digest !== false) $rootScope.$digest();
     if (responses.length) {
       throw new Error('Unflushed requests: ' + responses.length);
     }


katemihalikova:
Signed, thx.

glebec:
If #14235 is accepted we should add similar documentation to this parameter.

gkalpak:
I added a test, tweaked the commit message (to add a more detailed description and a breaking change notice) and landed.

Thx @katemihalikova!

PR #13514, opened 2015-12-11
fix($animateCss): don't unregister events when empty

Narretz:


src/ngAnimate/animateCss.js
```diff
@@ -750,7 +750,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
         }
 
         // Remove the transitionend / animationend listener(s)
-        if (events) {
+        if (events && events.length) {
```
gkalpak: Nitpick: `events` will always be truthy, since it is initialized to `[]`.

gkalpak:
We need to make sure that we will not call `.on()` if `events` is empty, because then event listeners will be registered for `''` (empty string) and they will never be deregistered.
I haven't checked if there are other conditions that ensure we never reach the `.on(events.join(' '), ...)` call, but if there aren't, we should check explicitly. (In fact, it would be better to check explicitly eitherway, unless I missed something obvious.)

Narretz:
Hmm. I think we should only test behavior. The unit tests should not be concerned with the events variable or if it can be empty or not. I realize that keeping implementation out of unit tests isn't always easy. For example, with $animateCss, you have to base the tests on the assumption that an animation will only actually start after a requestAnimationFrame.

I will have another look at the code and see if there are other cases where the animation process gets aborted before the events are registered.

src/ngAnimate/animateCss.js
Narretz: Not a nitpick, good catch!

gkalpak:
By "check" I didn't mean in unit tests, I meant in the actual code that sets up the listeners.
I.e. change [this line][1]:

```js
// Before:
element.on(events.join(' '), onAnimationProgress);
// ^-- If events is empty, a listener will be registered for '' (and never deregistered)

// After:
if (events.length) element.on(events.join(' '), onAnimationProgress);
```

[1]: https://github.com/Narretz/angular.js/blob/dc88816bdb7057f432eca4f5629290ef5e07608f/src/ngAnimate/animateCss.js#L963

Narretz:
Ok, I see what you mean. I don't think events can be empty when listeners are added, but the `if` can't hurt.

src/ngAnimate/animateCss.js
```diff
@@ -749,8 +749,8 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
           options.onDone();
         }
 
-        // Remove the transitionend / animationend listener(s)
-        if (events) {
+        if (events && events.length) {
```
Narretz: It's actually possible that events is undefined, because if no node is detected, the animation is closed before a bunch of variables are even declared.

gkalpak: Right, I missed that :disappointed:

petebacondarwin:
LGTM

PR #13517, opened 2015-12-11
chore(travis/build.sh): move ddescribe check before unit tests

Narretz:
This was changed as part of https://github.com/angular/angular.js/pull/9792.
While the logic is sound that style errors shouldn't block tests from
running, ddescribe should run before the tests.
Otherwise, when Travis exits with a warning after some browsers have run,
ddescribe doesn't get run and it doesn't immediately become apparent
that not all tests have run.

### DIFF
--- a/scripts/travis/build.sh
+++ b/scripts/travis/build.sh
@@ -5,7 +5,9 @@ set -e
 export BROWSER_STACK_ACCESS_KEY=`echo $BROWSER_STACK_ACCESS_KEY | rev`
 export SAUCE_ACCESS_KEY=`echo $SAUCE_ACCESS_KEY | rev`
 
-if [ $JOB = "unit" ]; then
+if [ $JOB = "ci-checks" ]; then
+  grunt ci-checks
+elif [ $JOB = "unit" ]; then
   if [ "$BROWSER_PROVIDER" == "browserstack" ]; then
     BROWSERS="BS_Chrome,BS_Safari,BS_Firefox,BS_IE_9,BS_IE_10,BS_IE_11,BS_iOS"
   else
@@ -14,7 +16,6 @@ if [ $JOB = "unit" ]; then
 
   grunt test:promises-aplus
   grunt test:unit --browsers $BROWSERS --reporters dots
-  grunt ci-checks
   grunt tests:docs --browsers $BROWSERS --reporters dots
 elif [ $JOB = "docs-e2e" ]; then
   grunt test:travis-protractor --specs "docs/app/e2e/**/*.scenario.js"
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ branches:
 
 env:
   matrix:
+    - JOB=ci-checks
     - JOB=unit BROWSER_PROVIDER=saucelabs
     - JOB=docs-e2e BROWSER_PROVIDER=saucelabs
     - JOB=e2e TEST_TARGET=jqlite BROWSER_PROVIDER=saucelabs
@@ -51,11 +52,7 @@ install:
   - npm install
 
 before_script:
-  - mkdir -p $LOGS_DIR
-  - ./scripts/travis/start_browser_provider.sh
-  - npm install -g grunt-cli
-  - grunt package
-  - ./scripts/travis/wait_for_browser_provider.sh
+  - ./scripts/travis/before_build.sh
 
 script:
   - ./scripts/travis/build.sh
--- a/scripts/travis/before_build.sh
+++ b/scripts/travis/before_build.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
+mkdir -p $LOGS_DIR
+
+if [ $JOB != "ci-checks" ]; then
+  echo "start_browser_provider"
+  ./scripts/travis/start_browser_provider.sh
+fi
+
+npm install -g grunt-cli
+
+if [ $JOB != "ci-checks" ]; then
+  grunt package
+  echo "wait_for_browser_provider"
+  ./scripts/travis/wait_for_browser_provider.sh
+fi


scripts/travis/build.sh
```diff
@@ -12,6 +12,7 @@ if [ $JOB = "unit" ]; then
     BROWSERS="SL_Chrome,SL_Safari,SL_Firefox,SL_IE_9,SL_IE_10,SL_IE_11,SL_iOS"
   fi
 
+  grunt ddescribe-iit
```
gkalpak: TBH, I would prefer running all ci-checks (except for `jscs`) first and just run `jscs` after the unit tests.

Gruntfile.js
```diff
@@ -355,6 +355,6 @@ module.exports = function(grunt) {
   grunt.registerTask('minify', ['bower','clean', 'build', 'minall']);
   grunt.registerTask('webserver', ['connect:devserver']);
   grunt.registerTask('package', ['bower','clean', 'buildall', 'minall', 'collect-errors', 'docs', 'copy', 'write', 'compress']);
-  grunt.registerTask('ci-checks', ['ddescribe-iit', 'merge-conflict', 'jshint', 'jscs']);
+  grunt.registerTask('ci-checks', ['merge-conflict', 'jshint', 'jscs']);
```
gkalpak: It was convenient having all these tests in a single task, so one can run them before pushing to master with a single command. I would prefer to leave `ci-checks` as is and create a new task excluding the task that we choose (be it `ddescribe-iit` or `jscs`).

Narretz:
I've created a new task that includes ddescribe-iit, jshint and merge-conflicts. jscs is run after the tests run.

Gruntfile.js
```diff
@@ -355,6 +355,7 @@ module.exports = function(grunt) {
   grunt.registerTask('minify', ['bower','clean', 'build', 'minall']);
   grunt.registerTask('webserver', ['connect:devserver']);
   grunt.registerTask('package', ['bower','clean', 'buildall', 'minall', 'collect-errors', 'docs', 'copy', 'write', 'compress']);
-  grunt.registerTask('ci-checks', ['ddescribe-iit', 'merge-conflict', 'jshint', 'jscs']);
+  grunt.registerTask('ci-pre-checks', ['ddescribe-iit', 'jshint', 'merge-conflict']);
+  grunt.registerTask('ci-checks', ['merge-conflict', 'jshint', 'jscs']);
```
gkalpak: Was removing `ddescribe-iit` from this task intended ?

Gruntfile.js
```diff
@@ -355,6 +355,7 @@ module.exports = function(grunt) {
   grunt.registerTask('minify', ['bower','clean', 'build', 'minall']);
   grunt.registerTask('webserver', ['connect:devserver']);
   grunt.registerTask('package', ['bower','clean', 'buildall', 'minall', 'collect-errors', 'docs', 'copy', 'write', 'compress']);
-  grunt.registerTask('ci-checks', ['ddescribe-iit', 'merge-conflict', 'jshint', 'jscs']);
+  grunt.registerTask('ci-pre-checks', ['ddescribe-iit', 'jshint', 'merge-conflict']);
```
gkalpak: Was re-arranging them inteded ? I think `jshint` will err if there are `merge` artifacts left over, so having `merge-conflict` run before `jshint`, might produce more meaningful errors (but I haven't really tested it :smiley:).

Narretz:
Updated. Good catches!

scripts/travis/build.sh
```diff
@@ -12,9 +12,10 @@ if [ $JOB = "unit" ]; then
     BROWSERS="SL_Chrome,SL_Safari,SL_Firefox,SL_IE_9,SL_IE_10,SL_IE_11,SL_iOS"
   fi
 
+  grunt ci-pre-checks
   grunt test:promises-aplus
   grunt test:unit --browsers $BROWSERS --reporters dots
-  grunt ci-checks
+  grunt jcsc
```
gkalpak: _jcsc_  -->  _jscs_  (:smiley:)

gkalpak:
LGTM (with typo fixed) :+1:

petebacondarwin:
I think it would be even better if we create a new job on travis that only runs the ci-checks.
This would mean that we could see if tests are passing even if the ci-checks are failing, while having a clear failure for the build as a whole if the ci-checks fail.

Narretz:
Ok, sounds reasonable. The only downside I see is that it will increase the average build run time by a few minutes.

petebacondarwin:
Not necessarily as they are run in parallel - also the ci-checks wouldn't use up tunnels as it is not doing any karma or protractor

gkalpak:
If the ci-checks are failing it means that one of the following is happening:

1. `ddescribe-iit` fails - In that case, tests passing is not that informative.
2. `merge-conflict` fails - Most probably the source code is not parsable. Even if it (accidentally) is, there will need to be changes to remove the merge artifacts, so the tests might fails anyway.
3. `jshint` fails - There's a chance that the tests passing is informative, but not guarranteed.

So, I am not sure we gain much :smiley: `jscs` is the only one that is not directly related to the tests passing//failing, but that is run at the end.

<sub>Just saying...</sub>
</sub>(Either way is fine by me.)</sub>

petebacondarwin:
`ci-checks` also runs `jscs`.

The reason this is helpful is that when debugging it is helpful to send a commit with a ddescribe block to Travis to run just that block. But if the ddescribe test kills the build early you don't see if the test would have passed.

Narretz:
When you are working on a fix, it's also nice to see at a glance if your tests failed, or if you just made a style error.

Narretz:
I've created a new sh file which contains all the logic for the before_script block. However, I am getting this error for all jobs:

> /home/travis/build.sh: line 45: ./scripts/travis/before_build.sh: Permission denied

The file isn't much different than the others. Aynone have an idea?

gkalpak:
Could it be that directory you are trying to create ?

Narretz:
The file was created witt the wrong permissions ... I've added the new job, PTAL

petebacondarwin:
I was going to suggest a bit of `chmod` was needed :-)

gkalpak:
LGTM

petebacondarwin:
Merging but tweaking the error message here: https://github.com/angular/angular.js/pull/13517/files#diff-014afcca28a4c58f73f0c7973192510cR35

petebacondarwin:
Landed as 0b94e8a8bf313c0279f3f3b6516f123a32402d0e

Now checking that it works on 1.2, 1.3 and 1.4

petebacondarwin:
Landed as

* 34e5623542f293428627eaa11a66dd7dbb6db473 on 1.2
* 7fa13c3da2c5a41b9bcc8962e51b0e0f770e0dc5 on 1.3
* 7882c1c6aeb1dcfe89fe34e32dfa75dc81e8ae96 on 1.4

petebacondarwin:
Thanks @Narretz

PR #13522, opened 2015-12-12
fix(ngRepeat): Check own object property

recursivefunk:


Narretz:
Care to explain why?

jray:
I signed it.

jray:
@Narretz sure, without the check the loop will iterate over every property in the object's prototype chain. It appears a CI test is failing though, I'll have to look into it!

Narretz:
Looks like our unit tests don't like this change at all ... ;)
lastBlockMap is actually created as Object.create(null), doesn't that mean it has no prototype at all? The code comments say exactly that it safeguards against prototypically inherited properties.
If you could give a clear use case example where this problem happens for you, that would be great.

jray:
Ah yes an ```Object.create(null)``` instantiation would negate the need for this. It's not an issue I've faced in general - from what I understand if Objects accumulate a nested hierarchy of prototypal properties, searching down the prototype chain would negatively affect performance. Doesn't seem to be an issue here, though. I'll close this out - thanks for taking a look!

PR #13525, opened 2015-12-13
fix($compile): Add missing variable declaration

lgalfaso:


lgalfaso:
@petebacondarwin this should be backported, can you ping me when the 1.4 branch is open for commits?

Issue #13510, opened 2015-12-11
fix($controller): support instantiation of ES6 classes


petebacondarwin:
See https://github.com/angular/angular.js/pull/12598#issuecomment-163738832

The `$controller` service has two paths for creating instances of a controller. In the `later == true` path we are not using `$injector.instantiate`, which has recently been fixed to support ES6 classes.

We ought to refactor `$controller` to use `$injector.instantiate` in both paths for consistency, if this is possible.

lgalfaso:
Raw facts:

* A class constructors must be called using `new` (and this behavior cannot be replaced with `Function.apply` as there would be no `new.target` defined) https://people.mozilla.org/~jorendorff/es6-draft.html#sec-ecmascript-function-objects-construct-argumentslist-newtarget
* There is a currently defined behavior of `bindToController` that the values to be bound to the controller, are present during the controller initialization https://docs.angularjs.org/api/ng/service/$compile#-bindtocontroller-


These two things, AFAICT, are incompatible, and a breaking change will be needed to add support for ES6 classes and `bindToController`.

Proposed solution:
1. Change the behavior of `bindToController` and remove the mandate that the values must be present during the controller initialization. This is be a breaking change.
2. Recommend developers to use setters to detect when values are bound


An alternative to *2* is
3. When using `bindToController`, if the class contains a method called *[[TBD]]*, then, after the initial values are set, this function is called


I would like to avoid *3* as it creates some complexities

gkalpak:
TBH, I was in favor of something like (3).

The problems I see with (1) and (2):

**a.** It will be a major BC and one that will cause silent failures i most cases, so it will be even more difficult for people to migrate.

**b.** Using setters adds quite some amount of boilerplate.

**c.** Even with setters, some (not uncommon) scenarios might be difficult or inefficient to implement. E.g. quite often, I have used the scope bindings during controller initialization for one-off configuration (e.g. if `this.someProp` equals `X`, then operate in `Y` mode etc). This won't be so easy to do using setters, especially when multiple properties need to be taken into account together. You'd have to keep track of what properties have been set (until all properties you are intrested in are set), you should defer all dependent operations etc.
It's doable, but it won't be pleasant :)

--
What I think might be a good compromise is:

**a.** Having a special method, that will be called when the bindings are initialized. That way we can instantiate the object (using a constructor function or class), assign the properties to it and call it's _special method_. 

[...]

PR #13540, opened 2015-12-14
feat($compile): Allow ES6 classes as controllers with bindToController: true

lgalfaso:
Modify `$injector.invoke` so ES6 classes would be invoked using `new`

Closes: #13510

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -4235,6 +4235,53 @@ describe('$compile', function() {
     });
 
 
+    it('should eventually expose isolate scope variables on ES6 class controller with controllerAs when bindToController is true', function() {
+      if (!/chrome/i.test(navigator.userAgent)) return;
+      /*jshint -W061 */
+      var controllerCalled = false;
+      module(function($compileProvider) {
+        $compileProvider.directive('fooDir', valueFn({
+          template: '<p>isolate</p>',
+          scope: {
+            'data': '=dirData',
+            'str': '@dirStr',
+            'fn': '&dirFn'
+          },
+          controller: eval(
+            "class Foo {" +
+            "  constructor($scope) {}" +
+            "  check() {" +
+            "    expect(this.data).toEqualData({" +
+            "      'foo': 'bar'," +
+            "      'baz': 'biz'" +
+            "    });" +
+            "    expect(this.str).toBe('Hello, world!');" +
+            "    expect(this.fn()).toBe('called!');" +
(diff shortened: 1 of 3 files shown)


lgalfaso:
Do not know how to fix this flake in the tests
```
IE 9.0.0 (Windows 7) filters date should ignore falsy inputs FAILED
	Error: [$injector:modulerr] Failed to instantiate module ng due to:
	Out of memory
```

test/ng/compileSpec.js
```diff
@@ -4235,6 +4235,53 @@ describe('$compile', function() {
...
+        $compileProvider.directive('fooDir', valueFn({
+          template: '<p>isolate</p>',
+          scope: {
+            'data': '=dirData',
+            'str': '@dirStr',
+            'fn': '&dirFn'
+          },
+          controller: eval(
+            "class Foo {" +
+            "  constructor($scope) {}" +
```
gkalpak: Is `$scope` needed here ?

gkalpak:
LGTM

Something for a follow-up PR:
Maybe it's a good idea to avoid intializing the bindings prior to instantiating a class-based controller, since we know they won't be used (i.e. the return value will be different than `self`).

test/ng/compileSpec.js
lgalfaso: Not really, will remove it

lgalfaso:
@petebacondarwin do you think this should go into 1.5 ?

bisubus:
@lgalfaso There's a significant problem with this approach, I've covered it in `this comment` and fought it in `that extension` (you may [go by the implementation](https://github.com/ex-machine/ng-classified/blob/1.0.2/src/ng-classified.js#L54) and see how clumsy it became with multiple checks and flags).

The source of the problem is that this

    function isClass(func) {
      return typeof func === 'function'
        && /^class\s/.test(Function.prototype.toString.call(func));
    }

will work for native classes and will fail for transpiled ones (with Babel).

lgalfaso:
I am landing this as this PR already has enough to handle all current browsers that natively support `class`.

lgalfaso:
@bisubus I am not sure how this affects Babel at all. The current mechanism that forces a `this` into the constructor function is indistinguishable for Babel as there is no support for `new.target`. Would it be possible for you to create a plunker that uses the latest snapshot and shows this issue?

petebacondarwin:
This change caused IE9 to run out of memory.

PR #13543, opened 2015-12-15
fix(ngInclude): do not compile template if original scope is destroyed

### LINKED ISSUE #13515: ngInclude compiles template even after scope is destroyed
With slow internet connection template may load significant time. And scope may be destroyed before template is loaded (e.g. some section is hidden by user action).

ngInclude doesn't check whether scope is destroyed and compiles the template, that may lead to several problems:
Memory leak: compiled template and scope are never destroyed, destroy event will never fire.
If some directive inside ngInclude template requires parent directive from outside of ngInclude, it will throw a js error because outer directive is already destroyed.

The error occurs with angularjs 1.2.29 and 1.4.8.

Possible fix would be checking whether scope is not destroyed before compiling template or cancelling template load promise when scope is destroyed.

This plunker demonstrates the issue: http://plnkr.co/edit/20PSyn65ggL6FPYsLcJX
Ben Nadel also mentioned this issue in this post: http://www.bennadel.com/blog/2891-nginclude-asynchronous-template-life-cycle-bug-in-angularjs.htm

zgmnkv:
With slow internet connection scope may be destroyed before template is loaded.
Previously in this case ngInclude compiled template that leaded to memory leaks
and errors in some cases.

Closes #13515

### DIFF
--- a/src/ng/directive/ngInclude.js
+++ b/src/ng/directive/ngInclude.js
@@ -232,6 +232,8 @@ var ngIncludeDirective = ['$templateRequest', '$anchorScroll', '$animate',
             //set the 2nd param to true to ignore the template request error so that the inner
             //contents and scope can be cleaned up.
             $templateRequest(src, true).then(function(response) {
+              if (scope.$$destroyed) return;
+
               if (thisChangeId !== changeCounter) return;
               var newScope = scope.$new();
               ctrl.template = response;
@@ -253,6 +255,8 @@ var ngIncludeDirective = ['$templateRequest', '$anchorScroll', '$animate',
               currentScope.$emit('$includeContentLoaded', src);
               scope.$eval(onloadExp);
             }, function() {
+              if (scope.$$destroyed) return;
+
               if (thisChangeId === changeCounter) {
                 cleanupLastIncludeContent();
                 scope.$emit('$includeContentError', src);
--- a/test/ng/directive/ngIncludeSpec.js
+++ b/test/ng/directive/ngIncludeSpec.js
@@ -398,6 +398,26 @@ describe('ngInclude', function() {
   });
 
 
+  it('should not compile template if original scope is destroyed', function() {
+    module(function($provide) {
+      $provide.decorator('$compile', function($delegate) {
+        return jasmine.createSpy('$compile').andCallFake($delegate);
+      });
+    });
+    inject(function($rootScope, $httpBackend, $compile) {
+      $httpBackend.when('GET', 'url').respond('template text');
+      $rootScope.show = true;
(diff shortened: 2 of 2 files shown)


src/ng/directive/ngInclude.js
```diff
@@ -253,6 +255,8 @@ var ngIncludeDirective = ['$templateRequest', '$anchorScroll', '$animate',
               currentScope.$emit('$includeContentLoaded', src);
               scope.$eval(onloadExp);
             }, function() {
+              if (scope.$$destroyed) return;
```
lgalfaso: looks like this code path is not tested

lgalfaso:
otherwise, LGTM

src/ng/directive/ngInclude.js
zgmnkv: Yes, it is not tested. I think it is quite internal logic, the aim is not to perform unnecessary cleanup. By the time when scope is destroyed element should be also removed, so no need to call cleanupLastIncludeContent, and $emit on destroyed scope is useless. There are no changes visible from outside.
If test is required, I can try to spy on scope and element and check that scope.$destroy, scope.$emit and element.remove are not being called.

lgalfaso: I do not think there is a need for a lot of spying, you can have another test just like the one you have but
* same template but without the `ng-if`
* creating a new scope from `$rootScope`, linking the template to it
* registering a listener
* destroying the scope
* doing the http flush

zgmnkv: Do you mean listener to $includeContentError event? What should I expect in this case? this event will not fire anyway, because scope removes all listeners on destroy.
Or do you mean that we shouldn't make any expectations, just check that this code works?

lgalfaso: I think you are right, and nothing else is needed.

zgmnkv:
So would it be merged? As far as I understand we agreed that no more test is needed.
CI seems to have failed due to some internal error
`15 Dec 10:30:41 - Response: {"error": "Too many active org tunnels: 30 >= 30"}.`

lgalfaso:
Sorry for the delay, this landed into master

zgmnkv:
Thanks! Is it possible to merge it to stable branches 1.2, 1.3, 1.4 ? Should I create PRs to relevant branches?

lgalfaso:
I would be ok with 1.4 (even when I would like to avoid this as we are already in the final stages for a 1.5 release). All the other branches are only receiving security fixes or big regressions fixes.

@petebacondarwin WDYT?

gkalpak:
Backported to `v1.4.x` as 9590bcf (since @petebacondarwin said we should also backport _"any bug fixes that do not change public API or make breaking change"_).

zgmnkv:
Thanks guys

PR #13544, opened 2015-12-15
fix(input): fix URL validation being too strict

### LINKED ISSUE #13528: URL_REGEXP is not accurate
@andrewaustin @petebacondarwin

This commit broke the URL validation: https://github.com/angular/angular.js/commit/ffb6b2fb56d9ffcb051284965dd538629ea9687a

For example this URL (which uses a curly brackets macro syntax and should be a perfectly valid URL):
```
http://ads.xxxxx.com/xxxxxx/?pid=1234&pos=c&r={TIMESTAMP}
```

console

Also check this
http://stackoverflow.com/questions/5984116/regular-expression-for-w3c-compliant-urls

gkalpak:
Background:
Prior to ffb6b2f, there was a bug in `URL_REGEXP`, trying to match the hostname as `\S+` (meaning
any non-space character). This resulted in never actually validating the structure of the URL (e.g.
segments such as port, path, query, fragment).
Then ffb6b2f and subsequently e4bb838 fixed that bug, but revealed `URL_REGEXP`'s "strictness" wrt
certain parts of the URL.

Since browsers are too lenient when it comes to URL validation, this commit relaxes the "strictness"
of `URL_REGEXP`, focusing more on the general structure, than on the specific characters allowed in
each segment.

Note 1: `URL_REGEXP` still seems to be stricter than browsers in some cases.
Note 2: Browsers don't always agree on what is a valid URL and what isn't.

Fixes #13528

### DIFF
--- a/test/helpers/privateMocks.js
+++ b/test/helpers/privateMocks.js
@@ -11,7 +11,7 @@ function baseThey(msg, vals, spec, itFn) {
   var valsIsArray = angular.isArray(vals);
 
   angular.forEach(vals, function(val, key) {
-    var m = msg.replace(/\$prop/g, angular.toJson(valsIsArray ? val : key));
+    var m = msg.split('$prop').join(angular.toJson(valsIsArray ? val : key));
     itFn(m, function() {
       /* jshint -W040 : ignore possible strict violation due to use of this */
       spec.call(this, val);
--- a/test/ng/directive/inputSpec.js
+++ b/test/ng/directive/inputSpec.js
@@ -2535,22 +2535,115 @@ describe('input', function() {
 
 
     describe('URL_REGEXP', function() {
-      /* global URL_REGEXP: false */
-      it('should validate url', function() {
-        // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
-        expect(URL_REGEXP.test('http://server:123/path')).toBe(true);
-        expect(URL_REGEXP.test('https://server:123/path')).toBe(true);
-        expect(URL_REGEXP.test('file:///home/user')).toBe(true);
-        expect(URL_REGEXP.test('mailto:user@example.com?subject=Foo')).toBe(true);
-        expect(URL_REGEXP.test('r2-d2.c3-p0://localhost/foo')).toBe(true);
-        expect(URL_REGEXP.test('abc:/foo')).toBe(true);
-        expect(URL_REGEXP.test('http://example.com/path;path')).toBe(true);
-        expect(URL_REGEXP.test('http://example.com/[]$\'()*,~)')).toBe(true);
-        expect(URL_REGEXP.test('http:')).toBe(false);
-        expect(URL_REGEXP.test('a@B.c')).toBe(false);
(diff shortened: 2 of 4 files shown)


gkalpak:
For anyone interested, I used [this pen][1] to visuallize how each browser's validation compares to ours for the URLs used in the test.

[1]: https://codepen.io/ExpertSystem/pen/ZQQbVN?editors=001

test/ng/directive/inputSpec.js
```diff
@@ -2534,23 +2534,116 @@ describe('input', function() {
...
-        expect(URL_REGEXP.test('r2-d2.c3-p0://localhost/foo')).toBe(true);
-        expect(URL_REGEXP.test('abc:/foo')).toBe(true);
-        expect(URL_REGEXP.test('http://example.com/path;path')).toBe(true);
-        expect(URL_REGEXP.test('http://example.com/[]$\'()*,~)')).toBe(true);
-        expect(URL_REGEXP.test('http:')).toBe(false);
-        expect(URL_REGEXP.test('a@B.c')).toBe(false);
-        expect(URL_REGEXP.test('a_B.c')).toBe(false);
-        expect(URL_REGEXP.test('0scheme://example.com')).toBe(false);
-        expect(URL_REGEXP.test('http://example.com:9999/``')).toBe(false);
+    ddescribe('URL_REGEXP', function() {
```
petebacondarwin: `ddescribe` - tut tut!

test/helpers/privateMocks.js
```diff
@@ -11,7 +11,7 @@ function baseThey(msg, vals, spec, itFn) {
   var valsIsArray = angular.isArray(vals);
 
   angular.forEach(vals, function(val, key) {
-    var m = msg.replace(/\$prop/g, angular.toJson(valsIsArray ? val : key));
+    var m = msg.split('$prop').join(angular.toJson(valsIsArray ? val : key));
```
petebacondarwin: what is the motivation for this change?

petebacondarwin:
The regex looks good to me. Couple of minor comments to address.

test/ng/directive/inputSpec.js
gkalpak: Was testing our CI :stuck_out_tongue:

test/helpers/privateMocks.js
gkalpak: It's described in the commit message of the [first commit][1].

Basically, if your replacement (`val` or `key`) contains `$&`, it will be replaced with `$prop`.
Eg. `'Testing $prop'.replace(/\$prop/g, 'string with $&') === 'Testing string with $prop'`.

[1]: https://github.com/gkalpak/angular.js/commit/031303d3271905faf265470a6effe12ee7eebf6e

gkalpak: I just realized there a spec for the privateMocks. I'll add a test.

gkalpak:
@petebacondarwin, I added a test to the first commit. PTAL

petebacondarwin:
LGTM

PR #13546, opened 2015-12-15
docs: add docs for ngPattern, ngMinlength, ngMaxlength

### LINKED ISSUE #9991: Standalone docs for (ng)Pattern
`ngPattern` docs are currently only available for the different input types, but has been standalone for quite some time. The `pattern`attribute is also recognized as a directive (undocumented), and there's currently also a subtle difference between `pattern` and `ngPattern`, so this should definitely be documented.

Narretz:
Closes #9991

### DIFF
--- a/src/ng/directive/validators.js
+++ b/src/ng/directive/validators.js
@@ -19,7 +19,74 @@ var requiredDirective = function() {
   };
 };
 
+/**
+ * @ngdoc directive
+ * @name ngPattern
+ *
+ * @description
+ *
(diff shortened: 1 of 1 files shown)


petebacondarwin:
@Narretz - the Travis failures are real. Some of the e2e tests are failing because of this change.

Narretz:
Landed as https://github.com/angular/angular.js/commit/8863836cd4b82c8f0f9d72de4323aace28402c38

gkalpak:
:+1:

PR #13551, opened 2015-12-15
fix(linky): do not throw on non-string values

### LINKED ISSUE #13547: ngSanitize linky filter throws error if `text` is truthy but not a string
http://jsfiddle.net/an710xcL/16/

Stacktrace thrown in that fiddle:
```
TypeError: l.match is not a function
    at https://code.angularjs.org/1.4.8/angular-sanitize.min.js:15:169
    at fn (eval at <anonymous> (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:13365:15), <anonymous>:4:159)
    at regularInterceptedExpression (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:14446:55)
    at Object.expressionInputWatch [as get] (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:14347:26)
    at Scope.$digest (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:15888:40)
    at Scope.$apply (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:16160:24)
    at bootstrapApply (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:1679:15)
    at Object.invoke (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:4523:17)
    at doBootstrap (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:1677:14)
    at bootstrap (http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular.js:1697:12)
```

If looping an array or dictionary and linkying each value, it would be nice if mixed types didn't fail. 

A simple workaround for now is to cast before passing to linky, e.g:
```
var text = 42;
$filter('linky')(''+text);
```

gkalpak:
Fixes #13547

### DIFF
--- a/src/ngSanitize/filter/linky.js
+++ b/src/ngSanitize/filter/linky.js
@@ -133,9 +133,10 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
   var LINKY_URL_REGEXP =
         /((ftp|https?):\/\/|(www\.)|(mailto:)?[A-Za-z0-9._%+-]+@)\S*[^\s.;,(){}<>"\u201d\u2019]/i,
       MAILTO_REGEXP = /^mailto:/i;
+  var isString = angular.isString;
 
   return function(text, target, attributes) {
-    if (!text) return text;
+    if (!text || !isString(text)) return text;
     var match;
     var raw = text;
     var html = [];
--- a/test/ngSanitize/filter/linkySpec.js
+++ b/test/ngSanitize/filter/linkySpec.js
@@ -20,6 +20,14 @@ describe('linky', function() {
     expect(linky(undefined)).not.toBeDefined();
   });
 
+  it('should return non-string, truthy values unchanged', function() {
+    expect(linky(true)).toBe(true);
+    expect(linky(42)).toBe(42);
+    expect(linky([])).toEqual([]);
+    expect(linky({})).toEqual({});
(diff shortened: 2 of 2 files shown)


gkalpak:
I am not sure we want to fix this actually. We've been moving in the opposite direction with other filters lately (e.g. `filterFilter` and `orderBy` throwing if not passed an array).

In any case, I think it's better to either explicitly throw (as with other filters) or return the input unchanged, if it's not a string.

lgalfaso:
I am leaning towards not making this change, but I am fine if enough people wants this change

gkalpak:
@lgalfaso, do you think we should explicitly throw (a la https://github.com/angular/angular.js/blob/v1.5.0-rc.0/src/ng/filter/filter.js#L136) ?

lgalfaso:
@gkalpak with the right BC notice, yes, I think that that way is much better

petebacondarwin:
OK @gkalpak - let's go with throwing in non-sensical cases...

gkalpak:
Closing in favor of #13693.
@lgalfaso, @petebacondarwin, feel free to take a look at #13693 :smile:

PR #13557, opened 2015-12-16
Updating example to use standard promise syntax

mkalish:
Examples were using deprecated .success and .error rather .then


mkalish:
I signed it!

PR #13568, opened 2015-12-17
chore(jenkins): run Jenkins builds on Node 4 (via nvm)

petebacondarwin:


scripts/jenkins/set-node-version.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Install nvm for this shell
+source ~/.nvm/nvm.sh
```
Narretz: Does jenkins provide this?

petebacondarwin: It is manually installed on the server.

scripts/jenkins/master.sh
```diff
@@ -17,14 +17,7 @@ function init {
 
 function build {
   cd ../..
-
-  if [[ $NO_TEST == "true" ]]; then
```
gkalpak: If the `--no-test` option is not used any more, should we remove it from `ARG_DEFS` [above][1] ?

[1]: https://github.com/angular/angular.js/pull/13568/files#diff-8ba5664b63cf24ac7a492e66ec3364fcR7

scripts/jenkins/set-node-version.sh
```diff
@@ -0,0 +1,7 @@
...
+
+# Use node.js at 4.2.x
+nvm install 4.2
```
gkalpak: Is `nvm install ...` sufficient or do we also need `nvm use ...` ?

scripts/jenkins/master.sh
petebacondarwin: Oh yes. Good spot

scripts/jenkins/set-node-version.sh
petebacondarwin: `nvm install` is sufficient.

scripts/jenkins/master.sh
petebacondarwin: Added at 4bcb307abc29d35fa1fa3a6fd3ab1ef39dd52576 (and similar commits in each of the other branches).

scripts/jenkins/set-node-version.sh
mzgol: Could we print the nvm/Node/npm versions used? Could be useful for debugging.

scripts/jenkins/set-node-version.sh
mzgol: `4` would be enough, the latest `4.x.y` release is an LTS, it so happens it's `4.2.y` in fact.

scripts/jenkins/set-node-version.sh
petebacondarwin: I think that it does do that automatically...

scripts/jenkins/set-node-version.sh
petebacondarwin: From the node.js docs:

> **Minors:** Minors are the addition and refinements of APIs or subsystems. They do not generally change APIs or introduce backwards compatible breaking changes, except where unavoidable. These are mostly completely additive releases.

I think it is safer to stick with a fixed minor release for automatic updates

mzgol: OK. It doesn't matter here anyway as there will be no Node.js 4.3.

petebacondarwin: I see: https://github.com/nodejs/LTS/

PR #13569, opened 2015-12-17
chore(GruntFile): check angularFiles (update)

petebacondarwin:


petebacondarwin:
@matsko - can you take a look at these changes to your PR?

lib/grunt/validate-angular-files.js
```diff
@@ -0,0 +1,56 @@
...
+          if (!isHiddenFile && !detectedFiles[filePath]) {
+            grunt.log.error(filePath + " exists in the local file structure but isn't used by any module.");
+            errorsDetected = true;
+          }
+        }
+      });
+    });
+
+    if (errorsDetected) {
+      throw new Error("Not all files were properly detected in the local file structure.");
```
gkalpak: This error message is not 100% accurate (as the errors might be due to files in the local file structure not being used).

gkalpak:
LGTM (despite the inconsistent use of `'/"` :stuck_out_tongue:)

matsko:
@petebacondarwin LGTM. Please upload as is. I will fix the error message that @gkalpak pointed out to something better in a follow-up commit.

PR #13576, opened 2015-12-17
feat(ngMock): add support for `$animate.closeAndFlush()`

### LINKED ISSUE #13005: $animate.flush() does not flush $animateCss animations in queue
We have an issue in https://github.com/angular-ui/bootstrap/pull/4493 where it turns out that `$animate.flush()` does not flush the animations queued up by $animateCss, so the `expandDone` callback in the promise chain never fires in the collapse directive, which is causing some testing problems.

In particular, the `.then` callback in the promise returned by `$animateCss(...).start()` never gets called because there is no mechanism in ngMock for flushing out the queued animation.

matsko:
Use `$animate.closeAndFlush()` to close all running animations.

Closes #13005


src/ngAnimate/animateCss.js
```diff
@@ -456,7 +456,7 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
       var node = getDomNode(element);
       if (!node
           || !node.parentNode
-          || !$animate.enabled()) {
+          || !$$animateQueue.enabled()) {
```
gkalpak: Why this change ?

Narretz: ngMock decorates the $animate service with the new closeAndFlush method. Both $animateCss and $$animateJs have their own closeAndFlush implementations so, $animate requires both services. Since $animateCss also requires $animate, this would create a circular dependency.

src/ngAnimate/animateJs.js
```diff
@@ -63,8 +65,32 @@ var $$AnimateJsProvider = ['$animateProvider', function($animateProvider) {
         applyAnimationClasses(element, options);
       }
 
+      function close(success) {
```
gkalpak: `success` isn't used.

gkalpak:
LGTM (other than the unused `success` argument).

PR #13578, opened 2015-12-18
fix(ngAnimate): only copy over the animation options once

matsko:
A bug material has exposed that ngAnimate makes a copy of
the provided animation options twice. By making two copies,
the same DOM operations are performed during and at the end
of the animation. Therefore if the CSS classes being added/
removed contain existing transition code then this will lead
to rendering issues.

### DIFF
--- a/test/ng/animateCssSpec.js
+++ b/test/ng/animateCssSpec.js
@@ -31,6 +31,28 @@ describe("$animateCss", function() {
       expect(copiedOptions).toEqual(initialOptions);
     }));
 
+    it("should not create a copy of the provided options if they have already been prepared earlier",
+      inject(function($animateCss, $$rAF) {
+
+      var options = {
+        from: { height: '50px' },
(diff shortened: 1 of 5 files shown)


matsko:
@petebacondarwin or @Narretz please review.

matsko:
1.4.9 and 1.5.0-rc.1

test/ng/animateCssSpec.js
```diff
@@ -31,6 +31,28 @@ describe("$animateCss", function() {
       expect(copiedOptions).toEqual(initialOptions);
     }));
 
+    it("should not create a copy of the provided options if they have already been prepared earlier",
```
Narretz: Is it possible to provoke the unwanted render behavior instead? That the options are copied seems like an implementation detail to me.

Narretz:
I got an integration test for this and will add this during the WE.

PR #13579, opened 2015-12-18
refactor($parse): remove unused variables

thorn0:


gkalpak:
This seems indeed a left over from _the big `$parse` refactoring_ (https://github.com/angular/angular.js/commit/0d424263ead16635afb582affab2b147f8e71626).
@lgalfaso, can you confirm these aren't needed ?

lgalfaso:
landed as dec8a0eb72594577b05108bb92cf2e2f34c1ce24

gkalpak:
Backported to `v1.4.x` as 689c01f.

Issue #13566, opened 2015-12-17
transclude: true is probably a bad default for component helper


sarod:
PR #12933 introduces the new component helper which is great.
However I found that using `transclude:true` as a default  is probably a mistake as:
* Most component don't need it. And it's find to enable it when you need it.
* It doesn't play well with some native directive also requiring transclusion (namely ngSwitchWhen)

For instance when creating a simple component with 
```js
module.component('myComp', 
  {
    template:'<div></div>'
  }
);
```
And use it in a template like this:
```html
<div ng-switch="x">
  <my-comp ng-switch-when="1"></my-comp>
  <div ng-switch-when="2"></div>
</div>
```

I get the following error:
```
[$compile:multidir] Multiple directives [ngSwitchWhen, myComp (module: myApp)] asking for transclusion on: <my-comp ng-switch-when="1">
```

petebacondarwin:
@sarod - thanks for this feedback.
A simple workaround for your case is to set transclude to be false:

```
module.component('myComp', 
  {
    transclude: false,
    template:'<div></div>'
  }
);
```

I am willing to consider that many components do not require transclusion. Moreover that if you want multislot transclusion then you would still need to provide the `transclude` property.

So perhaps it would be better to default `transclude` to `false`.

What do others think? @shahata ?

gkalpak:
Indeed `false` might be a more reasonable default for `transclude`.

Not directly related to this, but (off the top of my head) shouldn't `transclude: 'element'` and `transclude: true` be able to co-exist on an element ?

shahata:
We decided to set `transclude` to `true` under the assumption that it has no unwanted side effects. If that is not the case we should either reconsider or find a way to work around those side effects.

Narretz:
`transclude: false` is a good idea. Maybe we should also suggest that component directives should not have other directives on them? Moving to components, I'd personnally keep them separate from each other / other directives.

lgalfaso:
I think that `transclude: false` is a safer default

petebacondarwin:
OK, good call @sarod - let's change the default to `false`. This will be a breaking change in RC1.

@Narretz - although we should discourage multiple components on a single element, it seems like a common case to use a component on an element that has a structural directive, such as ng-if, ng-repeat, ng-swtich-when and so on. It just happens that these guys all use transclusion too.

As @gkalpak points out - perhaps we ought to allow element transclusion and content transclusion on the same element...

PR #13581, opened 2015-12-18
fix(loader): use false as default value of transclude in component helper

sarod:
Use false as default value of transclude in component helper.
The change is motivated by the fact that using transclude:true when not necessary made component
not usable in conjunction with structural directives requiring transclusion like ng-switch-when.

Closes: #13566

BREAKING CHANGE:
This is a breaking change compared to 1.5.0.rc.0 where transclude was true by default.


petebacondarwin:
LGTM! Thanks @sarod

PR #13592, opened 2015-12-19
Fix ngMock window.inject() stack trace reporting on PhanthomJS for 1.4.x branch

jurko-gospodnetic:
Fixes & adds a tests for issue #13591 for the angular `1.4.x` branch.

This does the same work as pull request #13593 does on the angular `master` branch.

Issue #13591, opened 2015-12-19
ngMock window.inject() does not get current stack trace correctly on PhantomJS and IE10+


jurko-gospodnetic:
**Overview of the Issue**
`ngMock` `window.inject()` function collects the current stack trace by expecting to find it in a locally created `Error` object's `stack` property. This however does not hold true in case when using a browser that fills an `Error` object's `stack` information only when that object gets thrown.

**Motivation for or Use Case**
The bug was found while researching why karma reports invalid error information when using PhantomJS in cases like this:
```
it('the holy crusade', inject(function() {
  var x = {};
  x.holyGrail();
}));
```
where the ngMock `inject()` implementation would incorrectly add the word `undefined` (read too early from the `Error` object's `stack` property) to the end of the collected error stack trace information, thus causing the main error description (gotten by parsing the collected error stack trace information, assuming it has specific structure) to be reported back to karma as `undefined`.

**Angular Version(s)**
- reproduced using angular version `1.4.8`
- current `master` version still has the exact same code so should be affected by the same bug

**Browsers and Operating System**
- reproduced using PhantomJS browser
- if [MSDN docs](https://msdn.microsoft.com/en-us/library/windows/apps/hh699850.aspx) are to be believed, should not work when used in IE browsers as well
- reproduced using `Ubuntu 15.10`, but should not be OS specific

**Reproduce the Error**
Here's a test for the angular ngMock test suite that can be used to reproduce the error:
```
      describe('when called outside of test spec context and inject callback throws an Error', function () {
        function testCaller() {
          return inject(function () {
            throw new Error();
          });
        }
        var throwErrorFromInjectCallback = testCaller();

        it('should update thrown Error stack with inject call location', function () {
          try {
            throwErrorFromInjectCallback();
          } catch (e) {
            expect(e.stack).toMatch('testCaller');
          }
        });
      });
```

Note that this whole functionality is completely untested and, at least to me, seems buggy or unfinished, but I'll open separate issues for that, while this one is targeted specifically to the error stack information constructed in `ngMock`'s `window.inject()`.

**Suggest a Fix**
- after creating the `Error` object in question, throw it and catch it in case it does not already have `stack` information stored on it

jurko-gospodnetic:
Pull request #13592 is a fix for this issue for the `1.4.x` branch.
Pull request #13593 is a fix for this issue for the `master` branch.

jurko-gospodnetic:
ping?

PR #13593, opened 2015-12-20
Fix ngMock window.inject() stack trace reporting on PhanthomJS for master branch

jurko-gospodnetic:
Fixes & adds a tests for issue #13591 for the angular `master` branch.

This does the same work as pull request #13592 does on the angular `1.4.x` branch.

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2899,6 +2899,15 @@ angular.mock.$RootScopeDecorator = ['$delegate', function($delegate) {
   window.inject = angular.mock.inject = function() {
     var blockFns = Array.prototype.slice.call(arguments, 0);
     var errorForStack = new Error('Declaration Location');
+    // some browsers, e.g. PhanthomJS, do not set a new error object's stack
+    // information until it has been thrown
+    if (!errorForStack.stack) {
+      try {
+        throw errorForStack;
+      } catch (e) {
+        errorForStack = e;
+      }
+    }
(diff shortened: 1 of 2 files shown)


jurko-gospodnetic:
Could this get merged? It's a simple bug fix, amply commented and with relevant tests added.

jurko-gospodnetic:
Could this pull request be merged? It's a simple bug-fix, well commented and with relevant tests added. Leaving it hanging here just makes it attract conflicts. :worried:

gkalpak:
Cleaned it up a bit and merged. @jurko-gospodnetic, it would be great if you could make sure it works as expected on PhantomJS.

jurko-gospodnetic:
@gkalpak hi & thanks for merging.

I ran the code through PhantomJS and it works as expected. I also ran angular's `ngMock` related `test:jqlite` test suite using PhantomJS and it passes and some custom added `console.log()` calls report not having stack information before throwing the error and having it after throwing & catching it, as expected.

There are some other parts that fail when you run the full angular test suite fails on PhantomJS, but I did not want to get into that at this time.

Btw. nice catch with throwing and catching not constructing a different error instance but updating the `stack` in the original error object.

gkalpak:
Thx for getting back !

> There are some other parts that fail when you run the full angular test suite fails on PhantomJS

Yeah, PhantomJS is not use of the officially supported browsers, so we don't run tests against it and our testsuite is not 100% "PhantomJS-compatible". But many people are using it in their projects without a problem (including myself).

Issue #13594, opened 2015-12-20
ngMock window.inject() does not add inject() call location info to errors thrown from repeated inject() callback calls


jurko-gospodnetic:
**Overview of the Issue**
`ngMock` `window.inject()` function has some seemingly quick-fix functionality adding the `window.inject()` call location stack trace information to any errors thrown by the callback passed to the `window.inject()` call. This however has a bug in that it only adds this information to errors thrown from the first callback invocation, but not for any later ones, while there are typical scenarios when a callback given to it may be invoked multiple times during a single test run.

**Motivation for or Use Case**
Here's a typical scenario where a callback passed to `ngMock`'s `window.inject()` function gets called multiple times in a single test suite:
```
describe('I\'m a little tea pot', function() {
  beforeEach(inject(function() {
    throw new Error();
  }));
  it('first test spec', function () {});
  it('second test spec', function () {});
});
```
Here the first test spec's error information will include the `window.inject()` call location stack trace information, while the second test spec's will not. And if for some reason only the second test spec throws an error, then no `window.inject()` call location stack trace information will be reported back to the user at all.

**Angular Version(s)**
- reproduced using angular `1.4.8` & `master` versions

**Browsers and Operating System**
- reproducible using any browser supporting JavaScript error stack information, e.g. Chrome, Firefox & Opera, IE10+ & PhantomJS
- reproduced using `Ubuntu 15.10`, but should not be OS specific

**Reproduce the Error**
Just run the use case presented in the **Motivation for or Use Case** section above.

**Suggest a Fix**
- don't clear the `errorForStack` information after the first call to the `window.inject()` callback

jurko-gospodnetic:
Pull request #13597 is a fix for this issue for the `1.4.x` branch.
Pull request #13598 is a fix for this issue for the `master` branch.

jurko-gospodnetic:
ping?

Narretz:
Thanks for raising this issue. It's not super high priority, so we'll look into it when we have the time.

jurko-gospodnetic:
Heh, yeah... but pull requests resolving it have been prepared... so I don't think applying it should take much time. :smile: Kind of a low hanging fruit at the moment while, as the time goes by, its pull requests will just collect conflicts with time.

Please let me know if there is anything else I can do to do get this merged.

Many thanks,
  Jurko Gospodnetić

Narretz:
Looks like the PRs have been merged!

Narretz:
[...]

PR #13598, opened 2015-12-20
Fix ngMock window.inject() error stack trace reporting on repeated or non-initial injection function calls for the master branch (issue #13594)

jurko-gospodnetic:
Fixes & adds tests for issue #13594 for the angular `master` branch.

Does the same work as pull request #13597 does on the angular `1.4.x` branch.

`window.inject()` function updates stack trace information for `Error` objects thrown from injection functions passed to it with stack trace for the `window.inject()` calling location. However this functionality was broken and did not work correctly in the following cases:
- when the same injection function gets called multiple times
- when more than one injection function is passed to a single `window.inject()` call and a non-initial one throws an `Error`

This pull request actually builds on and thus includes pull request #13593, as newly added test specs are integrated into the test structure set up by the base pull request. That's why you should first review & merge that pull request, or ignore its commits when reviewing this one.

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2960,8 +2960,6 @@ angular.mock.$RootScopeDecorator = ['$delegate', function($delegate) {
             throw new ErrorAddingDeclarationLocationStack(e, errorForStack);
           }
           throw e;
-        } finally {
-          errorForStack = null;
         }
       }
     }
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -972,6 +972,7 @@ describe('ngMock', function() {
             var throwingInjectingCall = testInjectCaller();
             throwingInjectingCall.setThrow(true);
 
+            // regression test for issue #13591 when run on IE10+ or PhantomJS
             it('should update thrown Error stack trace with inject call location', function() {
               try {
                 throwingInjectingCall();
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -932,19 +932,28 @@ describe('ngMock', function() {
       // function returned by inject(), when called outside of test spec
       // context, may have stored state so do not reuse the result from this
       // call in multiple test specs
-      function testInjectCaller() {
-        var shouldThrow;
-        // using an extra internalInjectCaller() wrapper here avoids stack trace
-        // constructed by some browsers (e.g. FireFox) from containing the name
-        // of the external caller function
-        var injectingCall = (function internalInjectCaller() {
-          return inject(function() {
-            if (shouldThrow)
+      function testInjectCaller(injectionFunctionCount) {
+        var shouldThrow = [];
+        // using an extra named function wrapper around the Error throw avoids
+        // stack trace constructed by some browsers (e.g. FireFox) from
+        // containing the name of the external caller function
+        function injectionFunction(index) {
+          return function() {
+            if (shouldThrow[index])
               throw new Error();
-          });
-        })();
-        injectingCall.setThrow = function(value) {
-          shouldThrow = value;
+          };
+        }
+        var injectionFunctions = [];
+        for (var i = 0; i < (injectionFunctionCount || 1); ++i) {
+          injectionFunctions.push(injectionFunction(i));
+        }
+        var injectingCall = inject.apply(window, injectionFunctions);
+        injectingCall.setThrow = function(index, value) {
+          if (!isDefined(value)) {
+            value = index;
+            index = 0;
+          }
+          shouldThrow[index] = value;
         };
         return injectingCall;
       }
@@ -998,6 +1007,22 @@ describe('ngMock', function() {
               }
               try {
                 injectingCall();  // repeated failing call
+              } catch (e) {
+                expect(e.stack).toMatch('testInjectCaller');
+              }
+            });
+          });
+
+          describe('when called outside of test spec context with multiple injected functions', function() {
+            var injectingCall = testInjectCaller(2);
+
+            // regression test for issue #13594
+            // regression test for issue #13591 when run on IE10+ or PhantomJS
+            it('should update thrown Error stack when second injected function fails', function() {
+              injectingCall.setThrow(0, false);
+              injectingCall.setThrow(1, true);
+              try {
+                injectingCall();
               } catch (e) {
                 expect(e.stack).toMatch('testInjectCaller');
               }
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -915,50 +915,67 @@ describe('ngMock', function() {
         }).toThrow('test message');
       }));
 
-      describe('error stack trace when called outside of spec context', function() {
-        // - Chrome, Firefox, Edge, Opera give us the stack trace as soon as an Error is created
-        // - IE10+, PhantomJS give us the stack trace only once the error is thrown
-        // - IE9 does not provide stack traces
-        var stackTraceSupported = (function() {
-          var error = new Error();
-          if (!error.stack) {
-            try {
-              throw error;
-            } catch (e) {}
-          }
-
-          return !!error.stack;
-        })();
-
-        function testCaller() {
-          return inject(function() {
-            throw new Error();
-          });
+      // - Chrome, Firefox, Edge, Opera give us the stack trace as soon as an Error is created
+      // - IE10+, PhantomJS give us the stack trace only once the error is thrown
+      // - IE9 does not provide stack traces
+      var stackTraceSupported = (function() {
+        var error = new Error();
+        if (!error.stack) {
+          try {
+            throw error;
+          } catch (e) {}
         }
-        var throwErrorFromInjectCallback = testCaller();
 
-        if (stackTraceSupported) {
-          describe('on browsers supporting stack traces', function() {
-            it('should update thrown Error stack trace with inject call location', function() {
-              try {
-                throwErrorFromInjectCallback();
-              } catch (e) {
-                expect(e.stack).toMatch('testCaller');
-              }
-            });
+        return !!error.stack;
+      })();
+
+      function testInjectCaller() {
+        var shouldThrow;
+        var injectingCall = (function internalInjectCaller() {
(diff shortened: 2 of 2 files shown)


Narretz:
Hi, I took a quick look at your PRs. Looking good so far. However, with all these commits it's hard to see what exactly belongs together What I would like to see is one commit per fix - that means the commit includes the fix, the tests and any refactorings. If you can create a new PR for the master branch with these changes, that would be great.

jurko-gospodnetic:
Pull request #13593, that this one is based on, has only two commits:
- a5a913a25f6ac2ef7f6a1718ee1332eea47cfa88 - making the fix
- e3eb4bd736dd8190339fab7e8ff49904fb49bb3f - adding tests

So that one should be reviewed & merged first.

What is left for this pull request are:

- 1 test refactoring commit (0c1b078b8744a94a7ba950f68a3c6436c237917b) intentionally split up to make it easier to review
- 1 fix commit (1c238a12e49bd4fac28dd8626e249c8f5cdbaa37)
- 3 comment adding commits (9b635bb25b511fc89a3ba3d749e576b93c174b52, 3244cf9a0aa2b914f231c3e89fb902c42fe6cf6d & 7ad56950c877e05ae5fd1915771ce6cb29367e2f)
- 2 test adding commits (563ca466098188ed7932355e2052ace3124ffc84 & 4e5b9d7ba4368264817d1af37fe73e09988320f1), each adding a single test related to this fix

Now, if you really feel that would help in some way, we can easily just squash it all together into:
- 1 commit for pull request #13593 (from its 2 commits)
- 1 commit for this pull request (from its 7 commits)

though you can see the effects of all of these squashed together easily on the `Files changed` tab already, and I personally find it easier to analyse the changes later on if needed when they are split up into such separate, stand-alone & commented commits.

In the end - it is just one fix per pull request as it is now.

Narretz:
I've taken another look at this. Sorry for the slow turnaround. The problem in https://github.com/angular/angular.js/pull/13593 I understand, but for these changes I can't really see the real world scenario where it would happen that you have multiple calls to inject() inside a single spec ( usually the spec wraps a single `it` fn, so you only have one inject per spec). I would like to see a real-world example of this being a problem first before merging it.

jurko-gospodnetic:
## Scenario 1
As a project's test suite grows, I find that it is often the case where some test utility functions get created. And some of them make use of specific angular-based services that they need to `inject`. In such cases it is typically useful to have that test utility perform the inject internally instead of requiring that the called pass it in explicitly. And when you have that tool, and use it in your test, you easily get in a situation where you make an `inject` call in the test call itself, and you make the `inject` call inside the test utility as well.

## Scenario 2
One of the test code organization patterns I've often encountered is this:

```
describe('some test suite', function () {
    var myService;

    beforeEach(inject(function (_myService_) {
        myService = _myService_;
    }));

    it('some spec', function () {
        ...make use of both myService...
    });

    it('another spec', inject(function (anotherService) {
        ...make use of both myService & anotherService...
    }));
});
```

In both of these scenarios you get multiple inject calls within the context of a single test spec.

And really, even if it were the case, the behaviour as implemented right now is inconsistent - it seems to say - doing this is allowed, but it is not allowed, or rather, you can do it, and it will most likely work, but maybe it won't work exactly what you expect, and we don't define or document what will happen. :smile:

An alternative to this patch would be reporting an error if an `inject()` call is made multiple times within the context of running a single test spec, which really seems like a very big (and undocumented) restriction.

Narretz:
The second scenario makes perfect sense. I just had a hard time visualizing the setup from the abstract tests. While trying to recreate the scenario, I've come across another issue. In the following plnkr, I have the example setup you provided: 
http://plnkr.co/edit/0edB849phWov4cRJCf9Q?p=preview

two specs, one inject in a beforeEach and another one in the second spec. Both inject the same service that is throwing on a fn that is called inside the spec. However, in the first spec, we don't get the inject part in the stacktrace (so no inject location is provided), but in the second spec we have it. From what I understand we should have the inject stack trace in both cases. Do you agree?

I have included your fix in the plnkr, so this is an issue that happens with your fix too.

jurko-gospodnetic:
Busty atm, but certainly curious and will take a look. If you like we can sync up over Skype (my login there is `jurkog`) or hangout or whatever and take a look at this together over the weekend.

Narretz:
@jurko-gospodnetic Sry, I didn't have time on the weekend, but we can sync on the coming weekend if you want

jurko-gospodnetic:
Cool, just contact me over Skype when you're online. :+1:

jurko-gospodnetic:
@Narretz I took a look at that plunkr you left a link to, and I don't really see anything wrong with the behaviour shown.

> However, in the first spec, we don't get the inject part in the stacktrace (so no inject location is provided), but in the second spec we have it. From what I understand we should have the inject stack trace in both cases. Do you agree?

Nope. The behaviour seems correct. The first spec does not fail inside any service injection function, while the second one does. Also, the behaviour presented in that plunkr does not really have anything to do with the changes done in this pull request as neither do any of the failing/throwing service injection functions get called multiple times nor do any of the `inject()` calls get passed multiple service injection callbacks.

There are actually two `inject()` calls done there:

1. The first `inject()` call made outside any specific test spec context, at the time when jasmine is still collecting all the test suites & specs
    1. it does not actually do any injections at that time, but instead returns a function that will do the `myService` injection and store the injected service in the `myService` variable, once it gets called
    1. that function is passed on into `beforeEach()` to have jasmine call it automatically at the start of each of our test specs
    1. since that service injection never fails, this whole initial `inject()` call is pretty much of no concern to us
    1. note that this constructed service injection function thus gets called twice - once for each test spec, and so, had it failed without the patch prepared in this pull request, its call location information would not have been included in the call stack information displayed for the second test case
1. The second `inject()` call also made outside any specific test spec context, at the time when jasmine is still collecting all the test suites & specs
    1. it does not actually do any injections at that time, but instead returns a function that will do the `$q` injection and then make the failing `myService.get()` call
    1. that function is passed on into the `it()` call as the implementation function for our second test case, to be called by jasmine at the appropriate time

and here is a more detailed description of what happens when each of our failing test specs get run:

1. The first failing test spec makes no additional `inject()` calls in it
    1. it just calls the injected service function `myService.get()`, which throws an Error
    1. the call stack presented for this test spec seems correct and lists:

        > this.get@http://run.plnkr.co/U1KJKxif15akUge5/app.js:9:25

        service function making the actual Error throw

        > @http://run.plnkr.co/U1KJKxif15akUge5/appSpec.js:39:15

        point in our test case function where we called the `myService.get()` function

        > attemptSync@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:1741:9
        > QueueRunner.prototype.run@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:1729:9
        > QueueRunner.prototype.execute@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:1714:5
        > ...

        the rest of jasmine's infrastructural kludge needed to run our test spec function
    1. there is no additional call stack information for any `inject()` call because the test failure did not occur inside any service injection function
1. The second failing test spec also makes no additional `inject()` calls, and instead just calls our previously registered service injection function (constructed by the second `inject()` call)
    1. the injected `$q` service never gets used for anything and so is of no concern to us
    1. the call to `myService.get()` throws an error and we get the expected call stack listed:

        > this.get@http://run.plnkr.co/U1KJKxif15akUge5/app.js:9:25

        service function making the actual Error throw

        > @http://run.plnkr.co/U1KJKxif15akUge5/appSpec.js:44:17

        point in our test case function where we called the `myService.get()` function

        > invoke@https://code.angularjs.org/snapshot/angular.js:4658:16
        > workFn@http://run.plnkr.co/U1KJKxif15akUge5/angular-mocks.js:2832:11

        angular's service injection function that got returned by the second `inject()` call - remember, that is what we registered as our test spec's implementation function, and the fact that there are two layers in the stack for it is just an internal angular.js detail as it actually creates another internal wrapper for that function

        > attemptSync@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:1741:9
        > QueueRunner.prototype.run@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:1729:9
        > QueueRunner.prototype.execute@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:1714:5
        > ...

        the rest of jasmine's infrastructural kludge needed to run our test spec function
    1. plus we get additional call stack information appended to the end (separated by a new-line) identifying the `inject()` call that constructed the service injection function inside which our error occurred:

        > angular.mock.inject@http://run.plnkr.co/U1KJKxif15akUge5/angular-mocks.js:2791:25
        
        point inside angular.js's `inject()` implementation where this call stack was actually collected, when we made the `inject()` call in question (we could tweak the displayed stack to remove this line if needed, but that was not done in the original code and thus is out of scope for this specific pull request)
        
        > @http://run.plnkr.co/U1KJKxif15akUge5/appSpec.js:42:67
        
        point in our test code where we made the inject() call that constructed the failing service injection function (in this case it has been done inline with the jasmine `it()` call defining the test spec in question so there is not much real use in this `inject()` call stack, but it would have come in useful had we called constructed the service injection function earlier and possibly reused it in different test specs)
        
        > addSpecsToSuite@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:725:9
        > Env/this.describe@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:695:7
        > jasmineInterface.describe@http://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.js:2969:14
        > ...
        
        jasmine's plumbing for collecting the list of all the defined test suites & specs, as a part of which our `inject()` call was made

To see the difference with or without the patch applied, you might want to replace the second test case with something like this:

```
describe('Mercury', function () {
    beforeEach(inject(function($q) {
        service.get();
    }));
    
    it('Venus', function () {});
    it('Earth', function () {});
});
```

Which, without the patch applied, should make:

1. the `Venus` test spec display the `inject()` call location information
1. the `Earth` test spec NOT display the `inject()` call location information

while with the patch applied, both should correctly display the `inject()` call location information

Issue #10149, opened 2014-11-20
$provide.decorator to extend directive scope no longer works in 1.3.x


sumigoma:
I'm using a directive that has an isolated scope, but I would like to expose an additional member so that I can bind to its value from the parent scope. Since the directive is from a third party I would like to do so without modifying the original code. In Angular 1.2.x, I used to be able to do *this* to decorate the existing directive with a new two-way binding.

```javascript
app.config(function($provide) {
    $provide.decorator('mycustomDirective',['$delegate','$controller', function($delegate, $controller) {
      
        var directive = $delegate[0];
        angular.extend(directive.scope, { 
            othervar: "=" 
        }); 
        
        return $delegate;  
    }]);
});   
```

However, in Angular 1.3.x, extending the scope this way no longer seems to work. This is evident in this **simplified plunkr**: [http://plnkr.co/edit/bwgwGm6IcoRfLjIL8K9X?p=info](http://plnkr.co/edit/bwgwGm6IcoRfLjIL8K9X?p=info). 

Changing the script tag to 1.3.x breaks the behavior. It seems all other properties of the directive can still be modified using $provide.decorator, but the scope in particular no longer seems to get updated.

Is this a bug? What changed in Angular 1.3.x to cause this specific behavior? And what is the recommended way of extending the scope of an existing directive (with an isolated scope)?

yangchristian:
+1. My use case is around directive templates: I'm using a 3rd party directive and want to alter some standard markup in its template to work with other directives. Here's the concrete scenario since that last sentence was a bit abstract:

- The 3rd party directive template has some simple `<a>` tags with an `href` binding, e.g. `<a href="{{ href }}">`
- I want to supply a [ui-router](https://github.com/angular-ui/ui-router) `state`, e.g. `<a ui-sref="{{ state }}">`
- I used to decorate the directive like OP, now I'm using the `.$$isolateBindings` workaround

arjunasuresh3:
+1

iainreid820:
As @sumigoma noted, appending 'Directive' after youe directive name allows you to then decorate that directive. I'm currently using this method in Angular 1.4.x.

Wlada:
+1

newmesiss:
as I solve this if I need to execute a method in controlling father?
which is ativado with ng-click within the directive used

nmccready:
[...]

PR #13605, opened 2015-12-21
Three small changes on compile

lgalfaso:
- A small refactor so no `<span>` elements are added to text element
- A fix so it is possible to add decorators to directives that modify the directive scope
- A style fix

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -778,6 +778,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
   // The assumption is that future DOM event attribute names will begin with
   // 'on' and be composed of only English letters.
   var EVENT_HANDLER_ATTR_REGEXP = /^(on[a-z]+|formaction)$/;
+  var bindingCache = createMap();
 
   function parseIsolateBindings(scope, directiveName, isController) {
     var LOCAL_REGEXP = /^\s*([@&]|=(\*?))(\??)\s*(\w*)\s*$/;
@@ -785,6 +786,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     var bindings = {};
 
     forEach(scope, function(definition, scopeName) {
+      if (definition in bindingCache) {
+        bindings[scopeName] = bindingCache[definition];
+        return;
+      }
       var match = definition.match(LOCAL_REGEXP);
 
       if (!match) {
@@ -802,6 +807,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         optional: match[3] === '?',
         attrName: match[4] || scopeName
       };
+      if (match[4]) {
+        bindingCache[definition] = bindings[scopeName];
+      }
     });
 
     return bindings;
@@ -894,11 +902,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
                 directive.name = directive.name || name;
                 directive.require = directive.require || (directive.controller && directive.name);
                 directive.restrict = directive.restrict || 'EA';
-                var bindings = directive.$$bindings =
-                    parseDirectiveBindings(directive, directive.name);
-                if (isObject(bindings.isolateScope)) {
-                  directive.$$isolateBindings = bindings.isolateScope;
-                }
                 directive.$$moduleName = directiveFactory.$$moduleName;
                 directives.push(directive);
               } catch (e) {
@@ -2315,6 +2318,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
                  directive.restrict.indexOf(location) != -1) {
               if (startAttrName) {
                 directive = inherit(directive, {$$start: startAttrName, $$end: endAttrName});
+              }
+              if (!directive.$$bindings) {
+                var bindings = directive.$$bindings =
+                    parseDirectiveBindings(directive, directive.name);
+                if (isObject(bindings.isolateScope)) {
+                  directive.$$isolateBindings = bindings.isolateScope;
+                }
               }
               tDirectives.push(directive);
               match = directive;
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -871,7 +871,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    *    {@link guide/directive} for more info.
    * @returns {ng.$compileProvider} Self for chaining.
    */
-   this.directive = function registerDirective(name, directiveFactory) {
+  this.directive = function registerDirective(name, directiveFactory) {
     assertNotHasOwnProperty(name, 'directive');
     if (isString(name)) {
       assertValidDirectiveName(name);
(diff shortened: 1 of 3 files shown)


lgalfaso:
Given that it is somehow late for 1.5, I would like to have @petebacondarwin comment before merging this.

src/ng/compile.js
```diff
@@ -1319,13 +1318,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         // modify it.
         $compileNodes = jqLite($compileNodes);
       }
-      // We can not compile top level text elements since text nodes can be merged and we will
-      // not be able to attach scope data to them, so we will wrap them in <span>
```
gkalpak: OOC, what has changed and this comment is no longer valid ?
Why did we need to wrap them in `<span>`, but not any more ?

test/ng/compileSpec.js
```diff
@@ -367,27 +367,28 @@ describe('$compile', function() {
...
-      expect(text[0].nodeName).toEqual('SPAN');
-      expect(element.find('span').text()).toEqual('A<a>B</a>C');
-    }));
-
 
-    it('should not wrap root whitespace text nodes in spans', function() {
+    it('should not wrap root text nodes in spans', function() {
       element = jqLite(
         '<div>   <div>A</div>\n  ' + // The spaces and newlines here should not get wrapped
         '<div>B</div>C\t\n  ' +  // The "C", tabs and spaces here will be wrapped
```
gkalpak: The comment is no longer valid (and potentially confusing).

test/ng/compileSpec.js
```diff
@@ -6273,7 +6305,7 @@ describe('$compile', function() {
         inject(function(log, $rootScope, $compile) {
```
gkalpak: Totally unrelated, but `$log` is redundant here.

test/ng/compileSpec.js
```diff
@@ -5949,6 +5950,37 @@ describe('$compile', function() {
       });
 
 
+      it('should not merge text elements from transcluded content', function() {
+        module(function() {
+          directive('foo', valueFn({
+            transclude: 'content',
```
gkalpak: <sub>I wonder if `transclude: 'content'` ever had a special meaning...</sub>

src/ng/compile.js
```diff
@@ -785,6 +786,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     var bindings = {};
 
     forEach(scope, function(definition, scopeName) {
+      if (scopeName in bindingCache) {
```
gkalpak: Since `bindingCache` is global (as in "$compiler-wide"), what if different isolate bindings, have different `definition` for the same `scopeName` ?
Am I missing something ?

lgalfaso: I think you are right, let me fix this

src/ng/compile.js
lgalfaso: There are two parts:
* In theory, you can break things up if you do a *hand* transclude and then call `normalize` on an ascendant node. All the built-in directives do not suffer from this issue as they add a comment node between in between the transcluded content
* Given a text node that is a root node of transcluded content, you will not be able to call `scope` to get the element scope. Old parts of the code relied on this, but this is no longer true

test/ng/compileSpec.js
lgalfaso: in the old days, `transclude: 'content'` is what now is called `transclude: true`. Many tests still keep this old nomenclature

test/ng/compileSpec.js
lgalfaso: Thanks!

test/ng/compileSpec.js
lgalfaso: thanks!

test/ng/compileSpec.js
```diff
@@ -7427,6 +7459,51 @@ describe('$compile', function() {
...
+        $provide.decorator('fooDirective', function($delegate) {
+          var directive = $delegate[0];
+          directive.scope.something = '=';
+          directive.template = '<span>{{something}}</span>';
+          return $delegate;
+        });
+      });
+      inject(function($compile, $rootScope) {
+        element = $compile('<div><div foo something="abc"></div></div>')($rootScope);
+        $rootScope.abc = "bar";
```
gkalpak: Nitpick: _"bar"_  -->  _'bar'_ (for consistency)

test/ng/compileSpec.js
```diff
@@ -7427,6 +7459,51 @@ describe('$compile', function() {
...
+        $rootScope.$digest();
+        expect(element.text()).toBe('bar');
+      });
+    });
+
+
+    it('should distinguis different bindings with the same binding name', function() {
```
gkalpak: _distinguis_  -->  _distinguish_

test/ng/compileSpec.js
lgalfaso: done

test/ng/compileSpec.js
lgalfaso: done

test/ng/compileSpec.js
gkalpak: I meant _double quotes_ --> _single quotes_ (not _abc_ ---> _bar_).

But nevermind. My battle against double/single quote consistency is a lost cause anyway :stuck_out_tongue: 
They're everywhere...

test/ng/compileSpec.js
```diff
@@ -4193,7 +4194,7 @@ describe('$compile', function() {
       dealoc(element);
       compile('<div><span my-component>');
       expect(componentScope).not.toBe(firstComponentScope);
-      expect(componentScope.$$isolateBindings).toBe(first$$isolateBindings);
+      expect(componentScope.$$isolateBindings).toEqual(first$$isolateBindings);
```
gkalpak: Is this necessary ? I would expct that the `bindingCache` makes sure the same object is returned ?

gkalpak:
The 3rd commit message is a bit confusing. Did you mean "decorators" (instead of "interceptors").

LGTM otherwise (with https://github.com/angular/angular.js/pull/13605#discussion_r48242247 addressed if necessary)
:+1: :+1:

test/ng/compileSpec.js
lgalfaso: this is the comparison of the object that contains all the bindings. In this case, the object will be different but the content of both objects will have the same object instances as content.

gkalpak: Right !

In that case, it might be useful to have a test that verifies that `bindingCache` works (i.e. returns the same object for the same scope binding definition (when `attrName` is defined)).
I'm not sure though, since it might fall under the "implementation detail" category and shouldn't be tested for explicitly.

Leaving it up to you :smiley:

lgalfaso: I think this is an implementation detail

jbedard: I think that implementation detail is important. That whole point of moving `parseDirectiveBindings` out of the link phase was to avoid recreating the same objects over and over again (memory usage/gc). Currently this change is partially undoing that enhancement which is why this test has to be changed.

Can we just put an `if` statement around the `parseDirectiveBindings` call instead of changing this? Or would that break the fix for #10149?

lgalfaso: > That whole point of moving parseDirectiveBindings out of the link phase was to avoid recreating the same objects over and over again (memory usage/gc).

`parseDirectiveBindings` is not called during link time, but during compile time. The cost of creating one new object is negligible compared to everything else that is going on at compile time.

> Can we just put an if statement around the parseDirectiveBindings call instead of changing this?

I do not understand this part, can you please elaborate?

jbedard: Somehow only do the `parseDirectiveBindings` once basically...

lgalfaso: Added an if, not really a fan of the change but if it makes things faster

Narretz:
The refactor commit should be labeled "fix", otherwise it won't show up in the changelog (automatically). Refactor doesn't match anyway, it's definitely adding a change to the output of an API.

lgalfaso:
> The refactor commit should be labeled "fix", otherwise it won't show up in the changelog (automatically). Refactor doesn't match anyway, it's definitely adding a change to the output of an API.

Makes sense, will do so when merging

> Commit 1 and 3 can also be backported if I see correctly.

Makes sense, will backport these two

lgalfaso:
Landed in master with
a021a376fc8a4676dfded794cf3395a5668e31ab
7617c08da69a0d447507dadd0ef41d2415462ca7
1061c56fe159691112bc81ca34573ea9681a7255

and backported into 1.5 with
ece293a3635f3150c28852c85d06c8de5e461001
0728cc2f2bb04d5dbdfca41f3afacea16c75ee07

petebacondarwin:
Cool +1

gkalpak:
It seems that somehow the BC made it into `v1.5.x` and consequently into `v1.5.1` :(

lgalfaso:
@gkalpak can you explain this a little more? 
The BC landed in master with https://github.com/angular/angular.js/commit/7617c08da69a0d447507dadd0ef41d2415462ca7 and this change is not in 1.5.0 https://github.com/angular/angular.js/blob/v1.5.0/src/ng/compile.js#L1526

gkalpak:
@lgalfaso, (as you already know by now) it is in `v1.5.1` ([compile.js#L1532][1]), as it was accidentally "backported" (sort of). It has been reverted on `v1.5.x` and all is good now :smiley: 

[1]: https://github.com/angular/angular.js/blob/v1.5.1/src/ng/compile.js#L1532

PR #13608, opened 2015-12-22
docs($interpolateProvider): describe your change...

sakkeerhussain:
App name was different in two parts in example code.

### DIFF
--- a/src/ng/interpolate.js
+++ b/src/ng/interpolate.js
@@ -44,7 +44,7 @@ $interpolateMinErr.interr = function(text, err) {
       this.label = "This binding is brought you by // interpolation symbols.";
   });
 </script>
-<div ng-app="App" ng-controller="DemoController as demo">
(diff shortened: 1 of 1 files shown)


src/ng/interpolate.js
```diff
@@ -44,7 +44,7 @@ $interpolateMinErr.interr = function(text, err) {
       this.label = "This binding is brought you by // interpolation symbols.";
   });
 </script>
-<div ng-app="App" ng-controller="DemoController as demo">
+<div ng-app="customInterpolationApp" ng-controller="DemoController as demo">
```
gkalpak: I wonder why the protractor test below didn't catch this. Not very reassuring... :)

sakkeerhussain: I signed it!

Thanks,
Sakkeer Hussain

Narretz:
The example processor is adding the module defined by `<example module="customInterpolationApp">` as the ng-app on the body tag. So the right fix is to remove the ng-app from the index.html

Issue #13228, opened 2015-11-02
docs($compile): usage of bindToController as object is undocumented


gjungb:
Since version 1.4 it's possible to use ```bindToController``` as an object like so:

```
bindToController: {
  text: '@text',
  obj: '=obj',
  expr: '&expr'
},
scope: {}
```

This fact is not reflected in the documentation.

Narretz:
Yeah, that's right! Would you like to submit a PR that adds the info?

Nazanin1369:
@Narretz Putting up a PR today for this

PR #13625, opened 2015-12-24
docs($compile): provide usage of bindToController as an object

Nazanin1369:
Related to: #13228


Narretz:
Thanks for the PR. The info isn't really complete, though and I'd like to check if bindToController with hash is doing the "right thing" at the moment, before merging this: https://github.com/angular/angular.js/issues/13658

PR #13626, opened 2015-12-24
fix removeFromArray

someone235:
removeFromArray check if `val >= 0` before `arr.splice(index, 1)`, instead of `index >= 0`

### DIFF
--- a/src/ngAnimate/shared.js
+++ b/src/ngAnimate/shared.js
@@ -118,7 +118,7 @@ function pendClasses(classes, fix, isPrefix) {
 
 function removeFromArray(arr, val) {
   var index = arr.indexOf(val);
-  if (val >= 0) {
+  if (index >= 0) {
     arr.splice(index, 1);
   }
 }


someone235:
I signed it!

someone235:
I've signed the cia

lgalfaso:
Hi, can you please:
* Squash this into one commit
* Create a test that shows the error and that this fixes

Thanks!

Narretz:
removeFromArray is only used once, here: https://github.com/angular/angular.js/blob/9c7c494c3e4f325d7db3269c689ddfe5f78e0ec2/src/ngAnimate/animateCss.js#L846

This is used to store the `animation-play-state` style in the temporaryStyles, so it can be cleaned up when the animation ends. However, the style is already cleaned up at this point (before temporaryStyles): https://github.com/angular/angular.js/blob/9c7c494c3e4f325d7db3269c689ddfe5f78e0ec2/src/ngAnimate/animateCss.js#L725

So, yes there's a bug here, which could possibly remove a temporary style from an element if an animation is resumed.

We should simply remove this block: https://github.com/angular/angular.js/blob/9c7c494c3e4f325d7db3269c689ddfe5f78e0ec2/src/ngAnimate/animateCss.js#L844-L846

PR #13634, opened 2015-12-26
feat(q): add isPromiseLike

wesleycho:
- Add `isPromiseLike` public method for determining whether value is a promise-like object

This addreses #13372.

My only concern with this implementation is that IMO this doesn't really belong on $q, but maybe on the `angular` global itself.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -674,7 +674,7 @@ function isBoolean(value) {
 
 
 function isPromiseLike(obj) {
-  return obj && isFunction(obj.then);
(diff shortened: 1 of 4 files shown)


lgalfaso:
I expect that this change makes $q fail the Promise/A+ compliance test

wesleycho:
Removed the change to $q's $$resolve method - it does take away some of the motivation for this method being implemented though.

Some caveats - I'm not sure whether this function should be exposed on $q or `angular`. Another thing is that the change in ngAnimate appears to not be used, but if it were, it would have perf implications.

PR #13641, opened 2015-12-28
fix(copy): add support for String/Boolean/Number object types

jbedard:
By modifying `isString`/`isBoolean`/`isNumber` this also effects everywhere those are used, which I think is good? But could also effect performance in those areas if called an extreme number of times.

Fixes the original issue mentioned in #13398

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -881,22 +881,10 @@ function copy(source, destination) {
     }
 
     var needsRecurse = false;
-    var destination;
+    var destination = copyType(source);
 
-    if (isArray(source)) {
-      destination = [];
-      needsRecurse = true;
-    } else if (isTypedArray(source)) {
-      destination = new source.constructor(source);
-    } else if (isDate(source)) {
-      destination = new Date(source.getTime());
-    } else if (isRegExp(source)) {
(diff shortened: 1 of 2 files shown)


lgalfaso:
I am not sure about this PR change to isXXX can we refactor this so
toString is called once on the object?

jbedard:
I've updated the PR to only make the `toString` call once.

Original commit for reference: 5dc679db11d3b4efdaa048a7c46e710fe9504d24

src/Angular.js
```diff
@@ -886,16 +886,7 @@ function copy(source, destination) {
...
-      destination = new source.constructor(source);
-    } else if (isDate(source)) {
-      destination = new Date(source.getTime());
-    } else if (isRegExp(source)) {
-      destination = new RegExp(source.source, source.toString().match(/[^\/]*$/)[0]);
-      destination.lastIndex = source.lastIndex;
-    } else if (isFunction(source.cloneNode)) {
-        destination = source.cloneNode(true);
-    } else {
+    } else if (!(destination = copyType(source))) {
```
lgalfaso: I think this will not work as intended with `new Boolean(false)` nor with `new String('')`

jbedard:
Updated again

lgalfaso:
LGTM

PR #13650, opened 2015-12-29
refactor: Refactor the equals function for readability

TehShrike:
- Turn an unnecessary nested if block into an &&
- Change the isRegExp block to use the same "if the second argument is the wrong type, return false" pattern as all the other blocks

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -969,38 +969,37 @@ function equals(o1, o2) {
   if (o1 === null || o2 === null) return false;
   if (o1 !== o1 && o2 !== o2) return true; // NaN === NaN
   var t1 = typeof o1, t2 = typeof o2, length, key, keySet;
-  if (t1 == t2) {
-    if (t1 == 'object') {
-      if (isArray(o1)) {
-        if (!isArray(o2)) return false;
-        if ((length = o1.length) == o2.length) {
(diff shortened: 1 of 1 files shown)


gkalpak:
LGTM (does exactly as the description says - git diffing adds much more noise :smiley:)

TehShrike:
It's true, the [no-whitespace diff](https://github.com/angular/angular.js/pull/13650/files?w=1) is a lot shorter.

src/Angular.js
```diff
@@ -969,38 +969,37 @@ function equals(o1, o2) {
...
-              isDefined(o2[key]) &&
-              !isFunction(o2[key])) return false;
         }
         return true;
       }
+    } else if (isDate(o1)) {
+      if (!isDate(o2)) return false;
+      return equals(o1.getTime(), o2.getTime());
+    } else if (isRegExp(o1)) {
+      if (!isRegExp(o2)) return false;
```
lgalfaso: [nit] I find `return isRegExp(o2) && o1.toString() == o2.toString();` easier to read, but not a blocker

TehShrike: Me too, but every other `if (some type)` block used the other pattern, and the function as a whole seemed easier to read when they all followed the same pattern.

PR #13651, opened 2015-12-29
docs - grammar change - tutorial/6 - Templating Links

ammills01:
Corrected the grammar on line 62 by adding the word 'an' which forced me to move 'only' down to line 63.


ammills01:
I signed it!

PR #13655, opened 2015-12-30
docs(loader): use type in return tag

ulitink:
I've found that `@returns {angular.Module}` was intentionally changed to `@returns {module}` by commit 3f98d6ac990fd514065bfff04711ac97a5a2ae4d, but this looks incorrect to me. `angular.Module` is a valid type, could you please explain why is it not suitable?


petebacondarwin:
I think you are right!

Issue #13653, opened 2015-12-30
$q: Rejected Promises with non rejection callback should report the rejection reason in the console.


xieranmaya:
Eg: If I write the code below:
```js
var promise2 = Promise.resolve(8).then(function(value){
  alter(value)
})
```
Notice the `alter`(should be `alert`) function which is not exist and will cause an error which will be caught and adopted by the promise returned by the last then which is the `promise2`.

If you run this code in console, you'll see something like `Uncaught (in promise) ReferenceError: alter is not defined(…)` in console.
The primitive Promise in ES6 will report rejected reason in console when the promise has no rejection handler.

![image](https://cloud.githubusercontent.com/assets/2993947/12047348/df4dd4ae-af04-11e5-8c47-edad65f0381f.png)

But if you change the `Promise` to `$q` in the above code snippet, you won't see the error reported in console. Which is bacsuse $q did not do this.

Sometimes if the promise returned by the last `then` method in promise chain and something went wrong in the last `then` method, the promise will got rejected, if the promise implementation do not report the error in the console, the developers would never know the error unless he/she dive deep in the code...

On the contrary, Q's `done` method can report the error(but if you don't chain a `done` method in the promise chain, it won't), and Bluebird will always report the error(and some useful warnings) in console if a rejected promise has no rejection handler.

Q(notice the `done` method):
![image](https://cloud.githubusercontent.com/assets/2993947/12047376/64e6224c-af05-11e5-93df-8928a3ba2240.png)

Bluebird(with warnings):
![image](https://cloud.githubusercontent.com/assets/2993947/12047355/0d25dfe8-af05-11e5-8b24-762494f7a617.png)

$q(do not report error on rejected promise):
![image](https://cloud.githubusercontent.com/assets/2993947/12047565/75d55f20-af08-11e5-9dee-fd8048b47bb8.png)

$q(always report):
![image](https://cloud.githubusercontent.com/assets/2993947/12047572/ac3ee18a-af08-11e5-9e0b-f23001c9bc84.png)

Eh...as I wrote this, I found that since angular 1.4.x, it starts to report the error or exception caught in the promise but even if the exception will be processed in the subsequent promise chain, it still report the error which is not a desired behavior:
![image](https://cloud.githubusercontent.com/assets/2993947/12047479/4084e882-af07-11e5-82ee-7c26cf5bd37e.png)

[...]

Issue #7992, opened 2014-06-26
$q reports errors in promise callbacks even when they are caught


domenic:
Plunker: http://plnkr.co/edit/5vMj4AJibWcnTKYR5z73?p=preview

I have caught the error and used it to display useful information to my user in the UI. However, Angular still logs a big ol' stack trace to the console.

This isn't a spec compliance issue (since the error isn't thrown, just logged), but it's a developer ergonomics issue. I don't understand how using $q on any reasonably large promise project would not just result in a console filled with spurious errors.

As always, the correct time to report a rejection as unhandled is debatable. Bluebird does it at the end of turn, whereas Q doesn't do it at all, relying on `.done()` instead. But this isn't even about unhandled rejections---it's about all rejections!

The Bluebird strategy seems like a pure win: it would remove this spurious reporting, but keep it for any cases that could potentially be real.

just-boris:
Now I can silently rethrow error using `$q.reject()`

```js
$http.get().catch(function(err) {
   if(!canBeHandled) {
       // propagate error to further promises
       return $q.reject(err);
   }
});
```

But that means that all my services/controllers which use `$http` should include `$q`. This is a bit annoying. It would be better to propagate error using `throw` as well as other promise libraries do.

```js
$http.get().catch(function(err) {
   if(!canBeHandled) {
       // causes call to $exceptionHandler
       throw new Error(err);
   }
});
```
Now it calls `$exceptionHandler` and that means that it can't catch and suppress that error in further promises in chain.

op1ekun:
@just-boris I totally agree :+1: 

I run into this problem while simplifying services based on `$http` (I was using Bluebird for a while, but calling `$apply` on consumer's side was cumbersome).

I spent some time to figure out that `throw new Error()` doesn't reject $http promise.
Quite annoying...

Thankfully the project is quite fresh so refactoring is not a major issue.
However I can feel the pain of ppl throwing errors in their `$http` services.

xieranmaya:
What is thought to be the desired behavior is that it only report rejected promises' errors when there is no rejection handler registered on the promise(that is, never called then(*,handler) on it), which is like Q or Bluebird or ES6 primitive Promise.

#13653

PR #13662, opened 2016-01-02
feat($q): report promises with non rejection callback

lgalfaso:
Rejected promises that do not have a callback to handle the rejection report this to $exceptionHandler so they can be logged to the console.

BREAKING CHANGE

Tests that depend on specific order or number of messages in $exceptionHandler
will need to handle rejected promises report.

Closes: #13653
Closes: #7992

### DIFF
--- a/src/ng/interval.js
+++ b/src/ng/interval.js
@@ -188,6 +188,8 @@ function $IntervalProvider() {
       */
     interval.cancel = function(promise) {
       if (promise && promise.$$intervalId in intervals) {
+        // Interval cancels should not report as unhandled promise.
+        intervals[promise.$$intervalId].promise.catch(noop);
         intervals[promise.$$intervalId].reject('canceled');
         $window.clearInterval(promise.$$intervalId);
         delete intervals[promise.$$intervalId];
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -216,21 +216,58 @@
  *
  * @returns {Promise} The newly created promise.
  */
+/**
+ * @ngdoc provider
+ * @name $qProvider
+ *
+ * @description
+ */
 function $QProvider() {
-
+  var errorOnUnhandledRejections = true;
   this.$get = ['$rootScope', '$exceptionHandler', function($rootScope, $exceptionHandler) {
     return qFactory(function(callback) {
       $rootScope.$evalAsync(callback);
-    }, $exceptionHandler);
+    }, $exceptionHandler, errorOnUnhandledRejections);
   }];
+
+  /**
+   * @ngdoc method
+   * @name $qProvider#errorOnUnhandledRejections
+   * @kind function
+   *
+   * @description
+   * Retrieves or overrides whether to generate an error when a rejected promise is not handled.
+   *
+   * @param {boolean=} value Whether to generate an error when a rejected promise is not handled.
+   * @returns {boolean|ng.$qProvider} Current value when called without a new value or self for
+   *    chaining otherwise.
+   */
+  this.errorOnUnhandledRejections = function(value) {
+    if (isDefined(value)) {
+      errorOnUnhandledRejections = value;
+      return this;
+    } else {
+      return errorOnUnhandledRejections;
+    }
+  };
 }
 
 function $$QProvider() {
+  var errorOnUnhandledRejections = true;
   this.$get = ['$browser', '$exceptionHandler', function($browser, $exceptionHandler) {
     return qFactory(function(callback) {
       $browser.defer(callback);
-    }, $exceptionHandler);
+    }, $exceptionHandler, errorOnUnhandledRejections);
   }];
+
+  this.errorOnUnhandledRejections = function(value) {
+    if (isDefined(value)) {
+      errorOnUnhandledRejections = value;
+      return this;
+    } else {
+      return errorOnUnhandledRejections;
+    }
+  };
 }
 
 /**
@@ -239,10 +276,14 @@ function $$QProvider() {
  * @param {function(function)} nextTick Function for executing functions in the next turn.
  * @param {function(...*)} exceptionHandler Function into which unexpected exceptions are passed for
  *     debugging purposes.
+ @ param {=boolean} errorOnUnhandledRejections Whether an error should be generated on unhandled
+ *     promises rejections.
  * @returns {object} Promise manager.
  */
-function qFactory(nextTick, exceptionHandler) {
+function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
   var $qMinErr = minErr('$q', TypeError);
+  var queueSize = 0;
+  var checkQueue = [];
 
   /**
    * @ngdoc method
@@ -307,28 +348,62 @@ function qFactory(nextTick, exceptionHandler) {
     pending = state.pending;
     state.processScheduled = false;
     state.pending = undefined;
-    for (var i = 0, ii = pending.length; i < ii; ++i) {
-      deferred = pending[i][0];
-      fn = pending[i][state.status];
-      try {
-        if (isFunction(fn)) {
-          deferred.resolve(fn(state.value));
-        } else if (state.status === 1) {
-          deferred.resolve(state.value);
-        } else {
-          deferred.reject(state.value);
+    try {
+      for (var i = 0, ii = pending.length; i < ii; ++i) {
+        state.pur = true;
+        deferred = pending[i][0];
+        fn = pending[i][state.status];
+        try {
+          if (isFunction(fn)) {
+            deferred.resolve(fn(state.value));
+          } else if (state.status === 1) {
+            deferred.resolve(state.value);
+          } else {
+            deferred.reject(state.value);
+          }
+        } catch (e) {
+          deferred.reject(e);
+          exceptionHandler(e);
         }
-      } catch (e) {
-        deferred.reject(e);
-        exceptionHandler(e);
+      }
+    } finally {
+      --queueSize;
+      if (errorOnUnhandledRejections && queueSize === 0) {
+        nextTick(processChecksFn());
       }
     }
   }
 
+  function processChecks() {
+    while (!queueSize && checkQueue.length) {
+      var toCheck = checkQueue.shift();
+      if (!toCheck.pur) {
+        toCheck.pur = true;
+        var errorMessage = 'Possibly unhandled rejection: ' + toDebugString(toCheck.value);
+        exceptionHandler(errorMessage);
+      }
+    }
+  }
+
+  function processChecksFn() {
+    return function() { processChecks(); };
+  }
+
+  function processQueueFn(state) {
+    return function() { processQueue(state); };
+  }
+
   function scheduleProcessQueue(state) {
+    if (errorOnUnhandledRejections && !state.pending && state.status === 2 && !state.pur) {
+      if (queueSize === 0 && checkQueue.length === 0) {
+        nextTick(processChecksFn());
+      }
+      checkQueue.push(state);
+    }
     if (state.processScheduled || !state.pending) return;
     state.processScheduled = true;
-    nextTick(function() { processQueue(state); });
+    ++queueSize;
+    nextTick(processQueueFn(state));
   }
 
   function Deferred() {
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -706,13 +706,14 @@ angular.module('ngResource', ['ng']).
               return $q.reject(response);
             });
 
-            promise['finally'](function() {
+            promise = promise['finally'](function(response) {
               value.$resolved = true;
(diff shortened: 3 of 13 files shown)


src/ngResource/resource.js
```diff
@@ -702,12 +702,13 @@ angular.module('ngResource', ['ng']).
               return $q.reject(response);
             });
 
-            promise.finally(function() {
+            promise = promise.finally(function(response) {
```
jbedard: The `.finally` callback doesn't take any arguments and the result is ignored unless it's a promise (and even when it's a promise the result of that promise is ignored). I think this arg and the return can be dropped.

lgalfaso: Fact: If a `resource` call is canceled the internal promise is rejected

The previous code did the following:

`internalPromise.finally(...)` this creates a promise that I will call `X`
`internalPromise.then(...)` this second promise is returned. I will call this promise `Y`

If `internalPromise` is rejected, then it will reject `X` and `Y`. The user can handle the rejection of `Y` as this is what they receive, but they can do nothing about `X`. This leaves us with 2 options

1/ Change the code to
```js
promise.finally(...).then(noop, noop);
promise = promise.then(...);
```

This will handle any rejections from `X`

2/ What is in this patch

```js
promise = promise.finally(...).then(...);
```

The second option has the benefit of not creating another promise

jbedard: That all makes sense... but `response` is undefined 100% of the time because the `.finally` callback doesn't take any arguments. You can drop the `response` arg and return statement...

src/ng/interval.js
```diff
@@ -188,6 +188,8 @@ function $IntervalProvider() {
       */
     interval.cancel = function(promise) {
       if (promise && promise.$$intervalId in intervals) {
+        // Interval cancels should not report as unhandled promise.
+        intervals[promise.$$intervalId].promise.then(noop, noop);
```
gkalpak: Isn't this just `promise.then(noop, noop)` ? (Of course the user can mess things up by attaching the `$$intervalId` to an arbitrary promise, but it's (hopefully) unlikely.)

Unrelated to this PR, but any idea why we mark the `promise` parameter as optional ?

test/ngResource/resourceSpec.js
```diff
@@ -1052,7 +1052,8 @@ describe("basic usage", function() {
     it('should call the error callback if provided on non 2xx response', function() {
       $httpBackend.expect('GET', '/CreditCard/123').respond(ERROR_CODE, ERROR_RESPONSE);
 
-      CreditCard.get({id:123}, callback, errorCB);
+      var ccs = CreditCard.get({id:123}, callback, errorCB);
```
gkalpak: I certainly don't like the fact that (even if an error callback is provided) an error will be reported in the console.

lgalfaso: I agree, but the alternative would be
https://github.com/lgalfaso/angular.js/commit/9f7ed380da3491b2a353f67f41c3f92774713b49

Do you think the alternative is best?

gkalpak: Yes, I think an alternative that doesn't report an error when an error-callback is used is better.
lgalfaso@9f7ed38 isn't ideal either, as it's a BC :(

lgalfaso:
@gkalpak added an option to disable this behavior. PTAL.

src/ng/q.js
```diff
@@ -217,20 +217,51 @@
...
+    }, $exceptionHandler, errorOnUnhandledRejections);
   }];
+
+  /**
+   * @ngdoc method
+   * @name $qProvider#errorOnUnhandledRejections
+   * @kind function
+   *
+   * @description
+   * Retrieves or overrides on wheter to generate an error when a rejected promise is not handled.
```
gkalpak: _on wheter_  -->  _whether_ (?)

src/ng/q.js
```diff
@@ -217,20 +217,51 @@
...
+   *
+   * @param {boolean=} value Wheter to generate an error when a rejected promise is not handled.
```
gkalpak: _Wheter_  -->  _Whether_

src/ng/q.js
```diff
@@ -307,28 +342,62 @@ function qFactory(nextTick, exceptionHandler) {
...
+  function processChecksFn() {
+    return function() { processChecks(); };
+  }
+
+  function processQueueFn(state) {
+    return function() { processQueue(state); };
+  }
+
   function scheduleProcessQueue(state) {
+    if (!state.pending && state.status === 2 && !state.pur) {
```
gkalpak: Since `errorOnUnhandledRejections` doesn't change at runtime, it's better to avoid doing unnecessary work.
E.g. check for `errorOnUnhandledRejections` here, instead of pushing into `checkQueue` and scheduling `processChecksFn`, where we simply empty `checkQueue` (if `errorOnUnhandledrejections` is false).

src/ng/q.js
```diff
@@ -307,28 +342,62 @@ function qFactory(nextTick, exceptionHandler) {
...
+          exceptionHandler(e);
         }
-      } catch (e) {
-        deferred.reject(e);
-        exceptionHandler(e);
+      }
+    } finally {
+      --queueSize;
+      if (queueSize === 0) {
+        nextTick(processChecksFn());
```
gkalpak: Here too it might be better to not call this is `errorOnUnhandledRejections` is false.
<sub>(And why not simply `nextTick(processChecks)` ?)</sub>

test/ng/qSpec.js
```diff
@@ -1586,6 +1598,8 @@ describe('q', function() {
       var rejectedPromise = q.reject('rejected');
       expect(rejectedPromise['finally']).not.toBeUndefined();
       expect(rejectedPromise['catch']).not.toBeUndefined();
+      rejectedPromise.then(noop, noop);
```
gkalpak: I like `.catch(noop)` better (here and below) as it's more concoise and communicates the purpose a little better.

src/ngResource/resource.js
```diff
@@ -721,7 +718,13 @@ angular.module('ngResource', ['ng']).
                 (success || noop)(value, response.headers);
                 return value;
               },
-              responseErrorInterceptor);
+              responseErrorInterceptor || error ?
+              function(response) {
+                (error || noop)(response);
+                (responseErrorInterceptor || noop)(response);
+                return response;
```
gkalpak: Isn't this changing the current behavior ?
I think we should either return the return value of `responseErrorInterceptor` or `$q.reject(response)`.

Or maybe something like:

```js
promise = promise.then(
  function(response) {
    var value = responseInterceptor(response);
    (success || noop)(value, response.headers);
    return value;
  }, function(response) {
    (error || noop)(response);
    return $q.reject(response);
  }).catch(responseErrorInterceptor);
```

src/ngResource/resource.js
```diff
@@ -701,12 +701,9 @@ angular.module('ngResource', ['ng']).
               response.resource = value;
 
               return response;
-            }, function(response) {
-              (error || noop)(response);
-              return $q.reject(response);
             });
 
-            promise['finally'](function() {
+            promise = promise['finally'](function() {
```
gkalpak: OOC, does this make any difference (in this particular case, where the code inside the `finally` callback is synchronous) ?

gkalpak:
I left some comments. Some additional thoughts:

1. If we have this enabled by default, shouldn't we also document the extra errors as a BC.
    E.g. if my otherwise perfect production application started loggin warnings in my clients console, I would like a big, bold heads-up, so I can take appropriate meassures.
2. Same as with `$timeout/$interval`, an explicitly cancelled `$resource` request (using `$cancelRequest`) should **not** report an error, imo.
3. To avoid surprises, we should be more explicit - both in the **docs** and in **tests** - about cases that we are handling internally (e.g. `$timeout/$interval#cancel()`, `$resource#$cancelRequest()`?).

src/ng/q.js
lgalfaso: Makes sense

src/ng/q.js
lgalfaso: for `$$q`, `nextTick` will add a `$timeout`, if `queueSize != 0` then we know we still have to wait before doing the checks.

src/ngResource/resource.js
lgalfaso: The code inside needs to run once the http request is resolved or canceled.

src/ngResource/resource.js
lgalfaso: The proposed alternative has the issue that if `responseErrorInterceptor` is undefined, then `return $q.reject(response);` will be unhandled.

lgalfaso: The alternative would be to change `return $q.reject(response);` to `return responseErrorInterceptor ? $q.reject(response) : response;` that I think it is a little bit too much

lgalfaso:
Updated the PR with the changes. Made the change to `$resource#$cancelRequest()` so it follows the same logic that is present for `$timeout/$interval#cancel()`.

That said, I do not think that there is a need to add any special documentation for this case.

src/ngResource/resource.js
gkalpak: I meant (assuming that the finally callback does not return a promise and does not throw any errors), does it make any difference to have:

```js
promise.finally(...)
promise.then(...)
```
vs
```js
promise = promise.finally(...)
promise.then(...)
```

But thinking about it again, the former has the issue that is creates a new promise that will remain unhandled if `$http` rejects (as you explained in another comment).

src/ng/q.js
```diff
@@ -217,20 +217,51 @@
...
   this.$get = ['$rootScope', '$exceptionHandler', function($rootScope, $exceptionHandler) {
     return qFactory(function(callback) {
       $rootScope.$evalAsync(callback);
-    }, $exceptionHandler);
+    }, $exceptionHandler, errorOnUnhandledRejections);
   }];
+
+  /**
+   * @ngdoc method
+   * @name $qProvider#errorOnUnhandledRejections
```
gkalpak: I wonder where this method shows up in the docs (considering there are no docs for `$qProvider`).
I think there should "standalone" docs for `$qProvider` (even if just containing this method), but it can be tackled in a follow-up PR.

src/ng/q.js
```diff
@@ -217,20 +217,51 @@
...
+   * @kind function
+   *
+   * @description
+   * Retrieves or overrides on whether to generate an error when a rejected promise is not handled.
```
gkalpak: Is the `on` unnecessary here (or do  miss something) ?

test/ng/qSpec.js
gkalpak: Seems like you missed a couple of `then(noop, noop)`s here and [there][1].
OOC, is there anything special about them, or just an oversight ?

[1]: https://github.com/angular/angular.js/pull/13662/files#diff-0cfb390955e411a4a5437fa337d52481R2046

gkalpak:
I left a couple of minor comments, but it LGTM otherwise.

I'm not sure about the default value of the flag. I guess `true` is OK although it might scare people at first when they migrate to 1.6.x (I don't think anyone is handling all their rejections in tests for example).
Maybe we could even backport it to 1.5.x with the flag set to `false` by default (but it's not necessary).

lgalfaso:
Added a page for `qProvider`, fixed the documentation issue and changed the remaining `.then(noop, noop)` to `catch(noop)`

lgalfaso:
About backporting this to 1.5. I am fine as long as the default value is `false`. This ways, people can check this before 1.6 is out.

lgalfaso:
landed into master as c9dffde1cb167660120753181cb6d01dc1d1b3d0 and 71cf28cf06295e0936f706c048bc07e6c963acc4

src/ng/q.js
```diff
@@ -307,28 +348,62 @@ function qFactory(nextTick, exceptionHandler) {
...
+    }
+  }
+
+  function processChecks() {
+    while (!queueSize && checkQueue.length) {
+      var toCheck = checkQueue.shift();
+      if (!toCheck.pur) {
+        toCheck.pur = true;
+        var errorMessage = 'Possibly unhandled rejection: ' + toDebugString(toCheck.value);
+        exceptionHandler(errorMessage);
```
amalitsky: Out of curiosity, why `errorMessage` isn't wrapped by `new Error()` so that stack trace is available?

gkalpak: Good question 🤔

mgol: @amalitsky This code path is only used if the value thrown in a promise callback was not an error. We could construct an error object here but it wouldn't be of much value as its stack would end up here and not in the place from where the error comes.

To be useful, only errors need to be thrown in promise callbacks and not other values. Which is a good idea in general.

amalitsky: @mgol, wouldn't path also have a reference to the promise invocation?

PR #13663, opened 2016-01-02
fix(select): re-define ngModelCtrl.$render in the select postLink fn

### LINKED ISSUE #13583: Option element inside optgroup is not correctly marked as selected
If select tag contains multiple optgroup elements then options inside optgroups other than the first one are not correctly marked as selected. 

__See:__  http://codepen.io/anon/pen/jWqWBd
__Reproducable:__ always
__Browsers:__ Chrome ver. 47.0.2526.80 m
__Operating system:__ Windows 7

Occurs on version 1.4.8. The issue is not replicable on version 1.4.7 and the previous ones.

Narretz:
Previously, the `$render` function was re-defined in the `select` directive's
`preLink` function. When a `select` element is compiled, every `option`
element inside it is linked and registered with the `selectCtrl`, which
calls `$render` to update the selected `option`. `$render` calls `selectCtrl.writeValue`,
which adds an unknown `option` in case no option is selected. In cases where
`optgroup` elements are followed by a line-break, adding the unknown `option`
confuses the html compiler and makes it call the link function of the following
`option` with a wrong element, which means this option is not correctly
registered.
Since manipulation of the DOM in the `preLink` function is wrong API usage,
the problem cannot be fixed in the compiler.

With this commit, the `$render` function is not re-defined until the `select` directive's
`postLink` function, at which point all `option` elements have been linked
already.

The commit also changes the `toEqualSelectWithOptions` matcher to
take selected options in groups into account.

Closes #13583


test/ng/directive/selectSpec.js
```diff
@@ -198,6 +239,50 @@ describe('select', function() {
...
+
+      $compile(select)(scope);
+      scope.$apply();
+
+      expect(select).toEqualSelectWithOptions({'first':['A'], 'second': [['B']]});
+      dealoc(select);
+    });
+
+
+    it('should only call selectCtrl.writeValue after a digest has occured', function() {
```
Narretz: This tests the behavior explicitly, while the first test only tests the functionality. I like this one though, because it makes sure we don't call writeValue for each linked option during compilation.

PR #13672, opened 2016-01-04
fix($animateCss): remove animation end event listeners on close

Narretz:
Previously the transition/animation end events were not removed when the
animation was closed. This normally didn't matter, because
the close function knows the animations are closed and won't do work
twice.
However, the listeners themselves do computation that could fail when
the event was missing some data, for example when the event was
triggered instead of natural.

This commit includes the fix for a bug that was introduced by this change
and landed on master separately: d041081add76ef9d896cbc29f76a54baa3315135


Narretz:
Landed.

PR #13673, opened 2016-01-04
docs(tutorial/2): e2e test was missing filename

monkpit:
added `__`test/e2e/scenarios.js`:__` to denote which file we should change to add the behavioral tests.


monkpit:
I signed it :hammer:

PR #13681, opened 2016-01-05
docs($compile): add docs for bindToController with object hash

### LINKED ISSUE #13228: docs($compile): usage of bindToController as object is undocumented

Narretz:
Closes #13228
Closes #13625

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -212,9 +212,22 @@
  *
  *
  * #### `bindToController`
- * When an isolate scope is used for a component (see above), and `controllerAs` is used, `bindToController: true` will
+ * This property is used to bind scope properties directly to the controller. It can be either
+ * `true` or an object hash with the same format as the `scope` property. Additionally, a controller
+ * alias must be set, either by using `controllerAs: 'myAlias'` or by specifying the alias in the controller
+ * definition: `controller: 'myCtrl as myAlias'`.
+ *
+ * When an isolate scope is used for a directive (see above), `bindToController: true` will
  * allow a component to have its properties bound to the controller, rather than to scope. When the controller
  * is instantiated, the initial values of the isolate scope bindings are already available.
+ *
+ * It is also possible to set `bindToController` to an object hash with the same format as the `scope` property.
+ * This will set up the scope bindings to the controller directly. Note that `scope` can still be used
+ * to define which kind of scope is created. By default, no scope is created. Use `scope: {}` to create an isolate
+ * scope (useful for component directives).
+ *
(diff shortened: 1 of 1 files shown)


petebacondarwin:
We can add the bit about bound properties being available in a later commit. This particular PR LGTM right now.

src/ng/compile.js
```diff
@@ -212,10 +212,22 @@
  *
  *
  * #### `bindToController`
- * When an isolate scope is used for a component (see above), and `controllerAs` is used, `bindToController: true` will
+ * This property can be used to bind scope properties directly to the controller. It can be either
+ * `true` or an object hash with the same format as the `scope` property. Additionally,
+ * `controllerAs` must be used for it to work.
```
gkalpak: This makes it sound like the `controllerAs` property must be used (which is not the case).
To avoid confusion, I would re-phrase this to make it clear that a controller alias must be specified (either using `controllerAs` or `controller: '... as ...'`).

src/ng/compile.js
```diff
@@ -212,10 +212,22 @@
...
+ *
+ * When an isolate scope is used for a directive (see above), `bindToController: true` will
```
gkalpak: This is not specific to `isolate` scopes.

src/ng/compile.js
```diff
@@ -212,10 +212,22 @@
...
  * allow a component to have its properties bound to the controller, rather than to scope. When the controller
  * is instantiated, the initial values of the isolate scope bindings are already available.
  *
+ * It is also possible to set `bindToController` to an object hash with the same format as the `scope` property.
+ * This will set up the scope bindings to the controller directly. Note that `scope` can still be used
+ * to define which kind of scope is created. By default, no scope is created. Use `{}` to create an isolate
```
gkalpak: I would replace `` `{}` `` with `` `scope: {}` `` to avoid confusion.

src/ng/compile.js
Narretz: But how can you bind to the controller if there's nothing to bind to (i.e. if you have scope: true, then there are no bindings, right)?

src/ng/compile.js
Narretz: Good catch.

src/ng/compile.js
gkalpak: Good point :grin:

PR #13686, opened 2016-01-06
test: fix ms edge failures

gkalpak:


test/ng/compileSpec.js
```diff
@@ -231,11 +231,26 @@ describe('$compile', function() {
...
-    function assertIsValidSvgCircle(elem) {
+    function assertIsValidSvgCircle(elem, possiblyUnsizedParent) {
       expect(isUnknownElement(elem)).toBe(false);
       expect(isSVGElement(elem)).toBe(true);
+
       var box = elem.getBoundingClientRect();
-      expect(box.width === 0 && box.height === 0).toBe(false);
+      var hasWidthOrHeight = (box.width !== 0) || (box.height !== 0);
+
+      if (possiblyUnsizedParent && !hasWidthOrHeight) {
```
lgalfaso: this change defeats the porpoise of the test itself

test/ng/directive/inputSpec.js
```diff
@@ -1196,7 +1196,7 @@ describe('input', function() {
 
       it('should validate if max is empty', function() {
         $rootScope.maxVal = undefined;
-        $rootScope.value = new Date(9999, 11, 31, 23, 59, 59);
+        $rootScope.value = new Date(3000, 11, 31, 23, 59, 59);
```
lgalfaso: an explanation would be nice

test/ngAnimate/animateCssSpec.js
```diff
@@ -28,9 +35,19 @@ describe("ngAnimate $animateCss", function() {
 
   var ss, prefix, triggerAnimationStartFrame;
   beforeEach(module(function() {
-    return function($document, $window, $sniffer, $$rAF, $animate) {
+    return function($document, $sniffer, $$rAF, $animate) {
       prefix = '-' + $sniffer.vendorPrefix.toLowerCase() + '-';
-      ss = createMockStyleSheet($document, $window);
+      ss = createMockStyleSheet($document);
+      ss.addPossiblyPrefixedRule = function(selector, styles) {
```
lgalfaso: any reason why this is not done here
https://github.com/angular/angular.js/pull/13686/files#diff-40c3b4c40fc4d81bed628855f8d24f26R67 ?

lgalfaso:
I know that this touches several tests, but can you please squash this into one commit?

test/ng/compileSpec.js
gkalpak: I was not sure what the purpose of the test was (not to mention it has a comment that says it might be redundant), so I am open to alternative suggestions.

TBH, I am not too familiar with SVG, so I don't really know why MS Edge behaves differently than other browsers (although it theoretically supports `foreignObject`). It isn't Angular-related, btw, because the same behavior can be reproduced with "raw" HTML.

test/ng/directive/inputSpec.js
gkalpak: I didn't feel like an explanation was required in the actual code, so I put it on the commit message instead. The problem is, I don't really know why MS Edge doesn't like the original Data, nor could I reproduce it independently (outside of the tests), but I figured it is not worth investigating further.

test/ng/compileSpec.js
lgalfaso: I think it is better to put an `if (isEdge)` condition on the test and do not run the test there at all than this change. I understand that this is an issue with Edge, but we should not try to test something else because of a browser

test/ng/directive/inputSpec.js
gkalpak: It probably has something to do with how MS Edge handles UTC -> local datetime conversion while overflowing 9999. E.g. with my UTC+2 timezone, the UTC value `9999-12-31T21:59:59.999` (which corresponds to `9999-12-31T23:59:59.999` local time) is fine, but the UTC value `9999-12-31T22:00:00.000` (which corresponds to `+10000-01-01T00:00:00.000` local time) "confuses" MS Edge.

It may also have something to do with ISO 8601 defaulting to min. 4 digits per year, but _permitting the expansion of the year representation but only by prior agreement between the sender and the receiver_ (whatever that means).

test/ngAnimate/animateCssSpec.js
gkalpak: I thought about that too. I chose to make the change locally, since it isn't currently used outside of that file, but I am fine putting it in `createMockStyleSheet()` if you think it's better.

gkalpak:
@lgalfaso, thx for the review. I prefer leaving it as separate commits for now (so it's easier to review related changes in the proper context) and squash upon merging.

mgol:
Good job! LGTM.

test/ng/directive/inputSpec.js
petebacondarwin: This worries me...

test/ng/compileSpec.js
petebacondarwin: I think I am with @lgalfaso on this unless we can explain better why it is happening this way.

test/ngAnimate/animateCssSpec.js
lgalfaso: yes, please move it

gkalpak: Done in https://github.com/gkalpak/angular.js/commit/a2d5ef7092cc0321ba5458744940bf01edc50a49.

test/ng/compileSpec.js
gkalpak: I wrapped the test in `if (!isEdge)` in https://github.com/gkalpak/angular.js/commit/f7f44176b3d20c375300f2a28701f8d74fce5440.

test/ng/directive/inputSpec.js
gkalpak: I looked a little more into this: MS Edge does not seem to support years > 9999 after all.
([Demo][1]: it's not possible to set the year to > 9999 on Edge)

So, I think this test fix is fine (since `9999` doesn't have any significant value for the test - we just need a later date).

--
Yet, while investigating this, I realized that our internal implementation of date-family inputs, do not support years > 9999.
I will make an issue for that (unless it's so on purpose).

[1]: https://codepen.io/ExpertSystem/pen/YwQWRx?editors=001

petebacondarwin:
LGTM - @lgalfaso any further comments? If there are no more comments then let's land this before RC1

lgalfaso:
LGTM

petebacondarwin:
Please merge

gkalpak:
Backported to `v1.4.x` as da04571 and 0a641c0.

PR #13692, opened 2016-01-06
refactor: move component helper to the $compileProvider

Narretz:
The Module.component() now delegates to $compileProvider.component().
This has the following benefits:
- when using only the loader, we are not accessing out of scope variables / functions
- components can be registered via $compileProvider
- docs are a bit easier to find

/cc @petebacondarwin @shahata


test/ng/compileSpec.js
```diff
@@ -9276,4 +9276,159 @@ describe('$compile', function() {
...
+      module('my');
+
+      inject(function($compile, $rootScope, log) {
+        element = $compile('<my-component></my-component>')($rootScope);
+        expect(element.find('div').text()).toEqual('SUCCESS');
+        expect(log).toEqual('OK');
+      });
+    });
+
+    it('should register a directive via $compileProvider.component()', function() {
```
Narretz: new test

test/ng/compileSpec.js
```diff
@@ -9276,4 +9276,159 @@ describe('$compile', function() {
...
     });
   });
+
+  describe('component helper', function() {
+    it('should return the module', function() {
+      var myModule = angular.module('my', []);
+      expect(myModule.component('myComponent', {})).toBe(myModule);
+    });
+
+    it('should register a directive', function() {
```
Narretz: this test is new

src/ng/compile.js
```diff
@@ -915,6 +915,162 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+   *   });
+   * ```
+   *
+   * <br />
+   * See also {@link ng.$compileProvider#directive $compileProvider.directive()}.
+   */
+  this.component = function registerComponent(name, options) {
+    function factory($injector) {
+      function makeInjectable(fn) {
+        if (isFunction(fn) || Array.isArray(fn)) {
```
gkalpak: Better to use `isArray()` now that we're here.

src/ng/compile.js
```diff
@@ -915,6 +915,162 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+        }
+      }
+
+      var template = (!options.template && !options.templateUrl ? '' : options.template);
+      return {
+        controller: options.controller || function() {},
+        controllerAs: identifierForController(options.controller) || options.controllerAs || name,
+        template: makeInjectable(template),
+        templateUrl: makeInjectable(options.templateUrl),
+        transclude: options.transclude === undefined ? false : options.transclude,
```
gkalpak: Minor, but `options.transclude` or `options.transclude || false` would work as well (now that `false` is the default).

gkalpak:
A couple of minor nitpicks, but LGTM otherwise.

<sub>This will (obviously) affect the fix for https://github.com/angular/batarang/issues/275.</sub>

petebacondarwin:
LGTM - merging with tweaks suggested by @gkalpak and some minor docs changes

PR #13696, opened 2016-01-06
fix(ngAnimate): allow event listeners on document in IE

### LINKED ISSUE #13548: $animate.on: javascript error with IE11 (Angular v1.4.8)
Hi,
$animate.on generates a javascript error with IE11.
All is ok with Chrome and Firefox.
Plunker here (to test with IE):
http://plnkr.co/edit/jtokwmcZRiG0vxkiL6mE?p=preview

Open Developer Tools with F12 and click on "Visible".
The error is:
TypeError: L’objet ne gère pas la propriété ou la méthode "contains"
at Anonymous function (https://code.angularjs.org/1.4.8/angular-animate.js:2146:11)

In line 2146 in angular-animate.js (v1.4.8), there is:
```
function findCallbacks(parent, element, event) {
    ...
	forEach(entries, function(entry) {
          if (entry.node.contains(targetNode)) {
            matches.push(entry.callback);
```
IMHO, this is a problem with method "contains" with IE.
Node.contains is not implemented on Node.
See:
https://connect.microsoft.com/IE/Feedback/Details/785343

But, I haven't found a patch or a workaround for this bug.

Philippe

Narretz:
Fixes #13548

Should we put this contains method in the global closure so we don't have to define it twice?

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -171,6 +171,13 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       return mergeAnimationOptions(element, options, {});
     }
 
+    // IE9-11 has no method "contains" in SVG element and in Node.prototype. Bug #10259.
+    var contains = Node.prototype.contains || function(arg) {
+      // jshint bitwise: false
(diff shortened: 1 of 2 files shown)


petebacondarwin:
LGTM - merging

PR #13697, opened 2016-01-06
fix($injector): workaround for MS Edge class detection

lgalfaso:
Fix for MS Edge class detection

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -826,8 +826,10 @@ function createInjector(modulesToLoad, strictDi) {
       if (msie <= 11) {
         return false;
       }
+      // Workaround for MS Edge.
+      // Check https://connect.microsoft.com/IE/Feedback/Details/2211653
       return typeof func === 'function'
-        && /^class\s/.test(Function.prototype.toString.call(func));
+        && /^(?:class\s|constructor\()/.test(Function.prototype.toString.call(func));
     }
 
     function invoke(fn, self, locals, serviceName) {
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -1,5 +1,7 @@
 'use strict';
 
+/* globals support: false */
+
 describe('injector', function() {
   var providers;
   var injector;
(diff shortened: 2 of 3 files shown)


lgalfaso:
@petebacondarwin can you take a look at this? Did some small so we do not have to keep track of what browser supports what ES6 feature

test/auto/injectorSpec.js
```diff
@@ -247,30 +249,37 @@ describe('injector', function() {
...
+      if (support.fatArrow) {
         it('should take args before first arrow', function() {
           expect(annotate(eval('a => b => b'))).toEqual(['a']);
         });
+      }
 
+      if (support.classes) {
         it('should be possible to instantiate ES6 classes', function() {
           // Only Chrome (not even the FF we use) supports ES6 classes.
           if (!/chrome/i.test(navigator.userAgent)) return;
```
petebacondarwin: @lgalfaso should this chrome check be removed?

petebacondarwin:
One question, otherwise LGTM

test/auto/injectorSpec.js
lgalfaso: you are right

src/auto/injector.js
```diff
@@ -826,8 +826,10 @@ function createInjector(modulesToLoad, strictDi) {
       if (msie <= 11) {
         return false;
       }
+      // Workaround for MS Edge.
+      // Check https://connect.microsoft.com/IE/Feedback/Details/2211653
       return typeof func === 'function'
-        && /^class\s/.test(Function.prototype.toString.call(func));
+        && /^(?:class\s|constructor\()/.test(Function.prototype.toString.call(func));
```
petebacondarwin: Actually I think we need to test for `constructor(` only if we are on Edge, since it is possible that someone might have a function actually called `constructor`.

petebacondarwin: @lgalfaso explained that:

```js
(function constructor() {}).toString() === "function constructor() {}"
```

So this is not an issue

PR #13701, opened 2016-01-07
style(*): remove unused variables

jbedard:
... and in some cases make use of unused vars, and rename a duplicate var.

It would be nice of some of these could be caught by the build...

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -265,9 +265,9 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
       };
       ngModelSet = function($scope, newValue) {
         if (isFunction(parsedNgModel($scope))) {
-          invokeModelSetter($scope, {$$$p: ctrl.$modelValue});
+          invokeModelSetter($scope, {$$$p: newValue});
         } else {
-          parsedNgModelAssign($scope, ctrl.$modelValue);
+          parsedNgModelAssign($scope, newValue);
         }
       };
     } else if (!parsedNgModel.assign) {
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -644,7 +644,7 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
         setValidity(name, undefined);
         validatorPromises.push(promise.then(function() {
           setValidity(name, true);
-        }, function(error) {
+        }, function() {
           allValid = false;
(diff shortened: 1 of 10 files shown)


src/ng/templateRequest.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var $compileMinErr = minErr('$compile');
+var $templateRequestMinErr = minErr('$compile');
```
gkalpak: You could also just reuse the already created `$compileMinErr`.

gkalpak:
LGTM

src/ng/templateRequest.js
jbedard: I figured this file shouldn't depend on compile.js like that. Might want to also remove the `'$compile'` string but I figured that would be another change...

src/ng/directive/ngModel.js
```diff
@@ -265,9 +265,9 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
       };
       ngModelSet = function($scope, newValue) {
         if (isFunction(parsedNgModel($scope))) {
-          invokeModelSetter($scope, {$$$p: ctrl.$modelValue});
+          invokeModelSetter($scope, {$$$p: newValue});
         } else {
-          parsedNgModelAssign($scope, ctrl.$modelValue);
+          parsedNgModelAssign($scope, newValue);
```
petebacondarwin: @lgalfaso are these two changes correct?

lgalfaso: this is fine as `ngModelSet` is only called from
https://github.com/jbedard/angular.js/blob/unused/src/ng/directive/ngModel.js#L755

petebacondarwin:
In that case LGTM

Narretz:
jshint has an unused rule. Let's see what happens when we use that.

jbedard:
Currently that doesn't work very well because of how variables are shared across files and jshint is run on the src files. If jshint was instead run on the resulting angular.js it would work, or maybe if shared variables are marked as globals or the rule is disabled for those vars?

Narretz:
Many files use jshint global already. Some variables actually show up as undefined if they are not marked global. But maybe that's just my personal setup.

Narretz:
Ah, I confused unused with undef ... So yes, directives, providers, services all trigger unused because they are only used in AngularPublic.js. This file already has a global definition, it's just that a lot of functions and variables are missing.
And for some reason, the angular api functions are also marked as unused, even though they are marked global in the src\jshint file

PR #13708, opened 2016-01-07
fix(isArrayLike): recognize empty instances of an Array subclass

### LINKED ISSUE #13560: ng-repeat loops over Array Like objects instance methods when length is 0
Starting in angular 1.4.8 Array Like objects, ie Collection Classes which extend Array behave strangely in ng-repeat.

ng-repeat works correctly when length is > 0 and only loops over elements in the array but when length == 0 any non array element properties are looped over instead.

```
function BaseCollection(){}
BaseCollection.prototype = Array.prototype;
var child = new BaseCollection();
child.someProperty1 = true
child.someProperty2 = true
```

Even tho child is empty with length of 0 ng-repeat will loop over the 2 non element properties.

Works correctly in angular 1.4.7 but problem is reproducible in 1.4.8 and 1.5.0-rc.0

gkalpak:
Fixes #13560

### DIFF
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -1218,17 +1218,38 @@ describe('angular', function() {
     });
 
     it('should return true if passed a nodelist', function() {
-      var nodes = document.body.childNodes;
-      expect(isArrayLike(nodes)).toBe(true);
+      var nodes1 = document.body.childNodes;
+      expect(isArrayLike(nodes1)).toBe(true);
+
+      var nodes2 = document.getElementsByTagName('nonExistingTagName');
+      expect(isArrayLike(nodes2)).toBe(true);
     });
 
     it('should return false for objects with `length` but no matching indexable items', function() {
-      var obj = {
+      var obj1 = {
         a: 'a',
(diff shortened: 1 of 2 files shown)


test/AngularSpec.js
```diff
@@ -1218,17 +1218,38 @@ describe('angular', function() {
...
+      expect(isArrayLike(obj2)).toBe(false);
+    });
+
+    it('should return true for empty instances of an Array subclass', function() {
+      function ArrayLike() {}
+      ArrayLike.prototype = Array.prototype;
+
+      var arrLike = new ArrayLike();
+      expect(arrLike.length).toBe(0);
+      expect(isArrayLike(arrLike)).toBe(true);
```
gkalpak: This is actually the only test that fails without this change, but I added the other exectations to help catch future regressions.

petebacondarwin:
OK, I think this LGTM

petebacondarwin:
What do you think @lgalfaso ?

lgalfaso:
LGTM

PR #13711, opened 2016-01-08
feat(ngMock.$componentController): add helper to instantiate controllers for components

### LINKED ISSUE #13683: REQUEST FOR FEEDBACK: `angular.component()` - auto creating a controller
As discussed with @shahata and @petebacondarwin - 

The problem I run into when writing lots of "component like" directives is the unnecessarily repeated configuration over and over again.

### It's redundant to have a separate controller declaration.

Registering controllers by name in a component based architecture is useful only for unit tests, where you'd want to use `$controller` to create an instance of the component's controller.

But registering it over and over again is redundant and can be done by the `.component`.

### We should have one `.comp.js` file

If we use `.component()`, it should auto create a new controller, registered with the same component name (because it too must be unique).

So we should end up with one file where the `.component()` becomes a true "annotation like" thing, something like: 

```js
angular
 .module('myModule')
 .component('productList', {
   controller: ProductListController,
   templateUrl: 'product-list.comp.html'
  })

function ProductListController(){

}

$controller('productList') // <-- should be an instance of ProductListController

```

I'd love to hear the cons for this approach

What do you think?

petebacondarwin:
Closes #13683

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2161,6 +2161,28 @@ angular.mock.$ControllerDecorator = ['$delegate', function($delegate) {
   };
 }];
 
+/**
+ * @ngdoc service
+ * @name $componentController
+ * @description
+ * A service that can be used to create instances of component controllers.
+ * @param {string} componentName the name of the component whose controller we want to instantiate
+ * @param {Object} locals Injection locals for Controller.
+ * @param {Object=} bindings Properties to add to the controller before invoking the constructor. This is used
+ *                           to simulate the `bindToController` feature and simplify certain kinds of tests.
+ * @return {Object} Instance of requested controller.
+ */
+angular.mock.$ComponentControllerProvider = ['$compileProvider', function($compileProvider) {
+  return {
+    $get: ['$controller', function($controller) {
+      return function $componentController(componentName, locals, bindings, ident) {
+        var controller = $compileProvider.$$componentControllers[componentName];
+        return $controller(controller, locals, bindings, ident);
+      };
+    }]
+  };
+}];
+
 
 /**
  * @ngdoc module
@@ -2184,7 +2206,8 @@ angular.module('ngMock', ['ng']).provider({
   $log: angular.mock.$LogProvider,
   $interval: angular.mock.$IntervalProvider,
   $httpBackend: angular.mock.$HttpBackendProvider,
-  $rootElement: angular.mock.$RootElementProvider
+  $rootElement: angular.mock.$RootElementProvider,
+  $componentController: angular.mock.$ComponentControllerProvider
 }).config(['$provide', function($provide) {
   $provide.decorator('$timeout', angular.mock.$TimeoutDecorator);
   $provide.decorator('$$rAF', angular.mock.$RAFDecorator);
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2166,18 +2166,24 @@ angular.mock.$ControllerDecorator = ['$delegate', function($delegate) {
  * @name $componentController
  * @description
  * A service that can be used to create instances of component controllers.
+ * <div class="alert alert-info">
+ * Be aware that the controller will be instantiated and attached to the scope as specified in
+ * the component definition object. That means that you must always provide a `$scope` object
+ * in the `locals` param.
+ * </div>
  * @param {string} componentName the name of the component whose controller we want to instantiate
  * @param {Object} locals Injection locals for Controller.
  * @param {Object=} bindings Properties to add to the controller before invoking the constructor. This is used
  *                           to simulate the `bindToController` feature and simplify certain kinds of tests.
+ * @param {string=} ident Override the property name to use when attaching the controller to the scope.
  * @return {Object} Instance of requested controller.
  */
 angular.mock.$ComponentControllerProvider = ['$compileProvider', function($compileProvider) {
   return {
     $get: ['$controller', function($controller) {
       return function $componentController(componentName, locals, bindings, ident) {
-        var controller = $compileProvider.$$componentControllers[componentName];
-        return $controller(controller, locals, bindings, ident);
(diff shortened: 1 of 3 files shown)


test/ngMock/angular-mocksSpec.js
```diff
@@ -1857,6 +1857,65 @@ describe('ngMock', function() {
...
+          controller: 'TestController'
+        });
+      });
+      inject(function($componentController, $rootScope) {
+        var $scope = {};
+        var ctrl = $componentController('test', {$scope: $scope, a: 'A', b: 'B'}, { x: 'X', y: 'Y' });
+        expect(ctrl).toEqual({$scope: $scope, a: 'A', b: 'B', x: 'X', y: 'Y' });
+      });
+    });
+  });
```
gkalpak: I'd add a test with `controller: 'TestController as test'`

test/ngMock/angular-mocksSpec.js
```diff
@@ -1857,6 +1857,65 @@ describe('ngMock', function() {
...
+      }
+      module(function($compileProvider) {
+        $compileProvider.component('test', {
+          controller: TestController
+        });
+      });
+      inject(function($componentController, $rootScope) {
+        var $scope = {};
+        var ctrl = $componentController('test', {$scope: $scope, a: 'A', b: 'B'}, { x: 'X', y: 'Y' });
+        expect(ctrl).toEqual({$scope: $scope, a: 'A', b: 'B', x: 'X', y: 'Y' });
```
gkalpak: Super-nitpicking: Inconsistent spacing before/after `{` / `}` (affects other tests too) :stuck_out_tongue:

<sub>(Sorry, I couldn't help it :flushed:)</sub>

src/ngMock/angular-mocks.js
```diff
@@ -2161,6 +2161,28 @@ angular.mock.$ControllerDecorator = ['$delegate', function($delegate) {
...
+ * @param {string} componentName the name of the component whose controller we want to instantiate
+ * @param {Object} locals Injection locals for Controller.
+ * @param {Object=} bindings Properties to add to the controller before invoking the constructor. This is used
+ *                           to simulate the `bindToController` feature and simplify certain kinds of tests.
+ * @return {Object} Instance of requested controller.
+ */
+angular.mock.$ComponentControllerProvider = ['$compileProvider', function($compileProvider) {
+  return {
+    $get: ['$controller', function($controller) {
+      return function $directiveController(componentName, locals, bindings, ident) {
```
gkalpak: Nitpicking: `$directiveController` is a somewhat misleading name.

Narretz:
Wouldn't it be nice to have that for plain directives, too?

gkalpak:
There will be some slight inconsistency between using `controller: 'SomeCtrl as identifier'` and `controller: 'SomeCtrl'` or `controller: SomeCtrl`. In the former case, the controller will be added as an alias on the scope, whereas in other two cases, it won't (when usign `$componentController`, I mean). Likewise, failing to provide a `$scope` in `locals` will throw an error in the former case (because it will try to assign the controller instance on the scope), but not in the other two cases.

test/ngMock/angular-mocksSpec.js
petebacondarwin: If you like but the implementation is just proxying to `$controller` anyway so it doesn't test any new paths.

test/ngMock/angular-mocksSpec.js
petebacondarwin: Right! Interesting that JSCS doesn't pick that up

petebacondarwin:
@Narretz - it is much more complicated to do it for plain directives because we don't actually call the directive factory function until the directive is matched in some HTML. This means that the controller is not easily available. Components are simpler because the controller is not wrapped in a factory function.

petebacondarwin:
>There will be some slight inconsistency between using controller: 'SomeCtrl as identifier' and controller: 'SomeCtrl' or controller: SomeCtrl. In the former case, the controller will be added as an alias on the scope, whereas in other two cases, it won't (when usign $componentController, I mean). Likewise, failing to provide a $scope in locals will throw an error in the former case (because it will try to assign the controller instance on the scope), but not in the other two cases.

@gkalpak this is also true of `ngMock.$controller` right now.

test/ngMock/angular-mocksSpec.js
gkalpak: JSCS does not check for this. We'd need either the [disallowSpacesInsideObjectBrackets][1] rule or the [requireSpacesInsideObjectBrackets][2] rule, which we don't have.

I prefer disallowing spaces (which is more consistent with our disallowing spaces inside square brackets and parenthesis). Should I add a rule for object brackets ?

[1]: http://jscs.info/rule/disallowSpacesInsideObjectBrackets
[2]: http://jscs.info/rule/requireSpacesInsideObjectBrackets

gkalpak:
@petebacondarwin, true, but things are more explicit there. You are are the one passing the argument to `$controller()` so the outcome should be less surprizing. In the case of `$componentController()` this happens indirectly; I would expect the same outcome when instantiating a component controller.

I wouldn't go to great lengths trying to "fix" this. If there's no easy fix (e.g. extracting the controller name and ignoring the alias or something), I would at least document that when using `controller: 'SomeCtrl as identifier'`, you need to supply a `$scope` in `locals`.

No biggie anyway - it only affects test code. Feel free to leave as is, if you think it's OK.
LGTM otherwise :+1:

petebacondarwin:
@gkalpak - OK, I see what you mean. But since this is for people testing controllers that (hopefully) they have actually written, then it shouldn't be *too* surprising!

petebacondarwin:
Actually in light of the latest discussions about `controller as` we should probably provide the `ident` param, if it is not explicitly provided, to be in line with the values of the component definition. I.E. if none is given the controller is attached to the scope via `$scope.$ctrl`.

gkalpak:
Now it LRGTM ('R' for "really") :smiley:

Issue #13714, opened 2016-01-08
Remove Default from ngList


Binvention:
The default on ngList of ", " prevents the ability of the attribute to use an empty string as a delimiter which can be useful if you want a list of every character in the text (including line breaks, spaces, tabs, etc). if the current default is removed then an empty string will be possible (in fact it will be the default)

gkalpak:
It doesn't seem reasonable to have the empty string be the default delimiter, but we should allow the user to specify it (via `ng-list=""`).

m-amr:
@gkalpak 

I think there is noway to distinguish between ng-list and ng-list = ""
because they both will return an empty string as a value

     element.attr(attr.$attr.ngList)
     // it will always return empty sting in case of
     // <input ng-list> or <input ng-list="">

I think we have this two options 
1 - add another attr ng-list-no-separator
2 - add a value like ng-list="no-separator"

i think we can't make no separator a default instead of ", " because it will break the
already written application that depends on that behaviour,

Binvention:
I know that the default of ", " over writes if you try and out the delimiter as "". Removing the current delimiter would make the default "" since an attribute that isn't assigned a value returns "". I already tried it in my own code and it does work just by removing the default. And any other delimiter could still be assigned using ng-list="delimiter"

m-amr:
@Binvention ,
 but what if someone depends that the default value is ", " it will breaks his application
your solution looks logically right but will not have a backward compatibility
changing a default value will break application that depends on this default behaviour.

Binvention:
Yes it will lose backwards compatibility but any way of solving it while keeping the default will require more code. And it isn't very clear how much that default is even used. I tried looking up the origins for the default but was unable to find it. I don't think it would be a huge change for anyone, just allowing for one more option, but there isn't any way to tell. So it's a balance between excess code that might not be necessary, increased capability, and backwards compatibility that we don't know we need.

gkalpak:
I imagine that `', '` is a very reasonable default. I expect it to be used most of the time. We shouldn't be making a change that (a) breaks existing apps (i.e. raises the barrier to upgrade) and (b) require more code for the most common case.
Besides, I consider `''` a very rare usecase. 

[...]

PR #13725, opened 2016-01-09
fix(ngListDiretive): add ngListNoSeparator to split list by no separator

m-amr:


gkalpak:
Fixes #13714.

m-amr:
@Narretz I have updated the documentation, please check it.
Thanks.

src/ng/directive/ngList.js
```diff
@@ -19,6 +20,12 @@
  *   when joining the list items back together) and whitespace around each list item is stripped
  *   before it is added to the model.
  *
+ * The The behaviour of the directive is affected by the use of the `ngListNoSeparator` attribute.
```
shoja: "The The" to "The"

Narretz:
See here why we are not going to implement this: https://github.com/angular/angular.js/issues/13714#issuecomment-225170719

PR #13727, opened 2016-01-09
fix(ngListDiretive): add missing '>' to textarea closing tag ngListSpec

m-amr:


gkalpak:
Backported to `v1.4.x` as 7f1cd3e.

PR #13732, opened 2016-01-10
feat(ngMock.$componentController): not increase extra bytes in angular.min.js

### LINKED ISSUE #13683: REQUEST FOR FEEDBACK: `angular.component()` - auto creating a controller

drpicox:
I made some changes to the solution of @petebacondarwin that closes #13683 
Basically:
- I have removed the `$$componentControllers` map from `$controlerProvider`
- I have used `$injector` to get the component directive structure

The objective of these changes are:
- do not increase extra bytes the angular.min.js

### DIFF
--- a/docs/app/src/examples.js
+++ b/docs/app/src/examples.js
@@ -24,7 +24,26 @@ angular.module('examples', [])
 
 
 .factory('openPlunkr', ['formPostData', '$http', '$q', function(formPostData, $http, $q) {
-  return function(exampleFolder, clickEvent) {
+
+    var COPYRIGHT = 'Copyright ' + (new Date()).getFullYear() + ' Google Inc. All Rights Reserved.\n'
+     + 'Use of this source code is governed by an MIT-style license that\n'
+     + 'can be found in the LICENSE file at http://angular.io/license';
+    var COPYRIGHT_JS_CSS = '\n\n/*\n' + COPYRIGHT + '\n*/';
+    var COPYRIGHT_HTML = '\n\n<!-- \n' + COPYRIGHT + '\n-->';
+    function getCopyright(filename) {
+      switch (filename.substr(filename.lastIndexOf('.'))) {
+        case '.html':
+          return COPYRIGHT_HTML;
+        case '.js':
+        case '.css':
+          return COPYRIGHT_JS_CSS;
+        case '.md':
+          return COPYRIGHT;
+      }
+      return '';
+    }
+
+    return function(exampleFolder, clickEvent) {
 
     var exampleName = 'AngularJS Example';
     var newWindow = clickEvent.ctrlKey || clickEvent.metaKey;
@@ -67,7 +86,7 @@ angular.module('examples', [])
         var postData = {};
 
         angular.forEach(files, function(file) {
-          postData['files[' + file.name + ']'] = file.content;
+          postData['files[' + file.name + ']'] = file.content + getCopyright(file.name);
         });
 
         postData['tags[0]'] = "angularjs";
--- a/src/ng/animateCss.js
+++ b/src/ng/animateCss.js
@@ -15,10 +15,14 @@ var $CoreAnimateCssProvider = function() {
   this.$get = ['$$rAF', '$q', '$$AnimateRunner', function($$rAF, $q, $$AnimateRunner) {
 
     return function(element, initialOptions) {
-      // we always make a copy of the options since
-      // there should never be any side effects on
-      // the input data when running `$animateCss`.
-      var options = copy(initialOptions);
+      // all of the animation functions should create
+      // a copy of the options data, however, if a
+      // parent service has already created a copy then
+      // we should stick to using that
+      var options = initialOptions || {};
+      if (!options.$$prepared) {
+        options = copy(options);
+      }
 
       // there is no point in applying the styles since
       // there is no animation that goes on at all in
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -928,6 +928,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     return this;
   };
 
+  this.$$componentControllers = createMap();
   /**
    * @ngdoc method
    * @name $compileProvider#component
@@ -1052,6 +1053,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    * See also {@link ng.$compileProvider#directive $compileProvider.directive()}.
    */
   this.component = function registerComponent(name, options) {
+    var controller = options.controller || function() {};
+    var ident = identifierForController(options.controller) || options.controllerAs || '$ctrl';
+    this.$$componentControllers[name] = { controller: controller, ident: ident};
+
     function factory($injector) {
       function makeInjectable(fn) {
         if (isFunction(fn) || isArray(fn)) {
@@ -1065,8 +1070,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       var template = (!options.template && !options.templateUrl ? '' : options.template);
       return {
-        controller: options.controller || function() {},
-        controllerAs: identifierForController(options.controller) || options.controllerAs || '$ctrl',
+        controller: controller,
+        controllerAs: ident,
         template: makeInjectable(template),
         templateUrl: makeInjectable(options.templateUrl),
         transclude: options.transclude,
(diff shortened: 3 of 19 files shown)


petebacondarwin:
Nice one @drpicox - I will tweak this (due to refactorings I did already) and merge it on top of what I just landed.

drpicox:
Thanks, and thanks also for your code!  (Oh, that's the reason of conflicts)
That was the kind of implementation that I had in mind, it's not the better, 
- ¿what happens if there are multiple directives with the same name? 
- ¿is it ok allow to use it on directives? 


Btw, I do not understand why says cla:no, I have signed it on Jun 06, 2015 04:43 PDT

> It looks like you've already signed this CLA. If you'd like to edit your contact information, you may do so below.
>
> Date Signed
> Jun 06, 2015 04:43 PDT

petebacondarwin:
It says no because there are commits from you and me; and it is wants to check that you aren't just pushing someone else's code without their permission

petebacondarwin:
We can't use this on directives because the injector will only return a factory function for directives.

petebacondarwin:
One can only have a single component directive on an element, so having multiple component directives with the same name would not make sense.

I am not sure what the injector will do if there are multiple directives with the same name.

drpicox:
So it seems that's ok.
Thanks.

gkalpak:
The problem with this change is that it can't defferentiate between components and non-component directives and assumes.

If there are non-component directives that have the same name as a component, then the order of registration will affect the behavior of `$componentController` (and that's not good).

--
Maybe we could add a private (as in $$-prefixed) property on components, so we can tell them apart.

petebacondarwin:
I agree that there are some potential issues with this approach. Let me have a play today.

drpicox:
@gkalpak it is what I expected, but imho components are directives (sugar syntax), in fact they are registered as directives:

https://github.com/angular/angular.js/blob/dd14e0c44d2963d217cd4eb28f1ad6e6a643d63f/src/ng/compile.js#L1092

```javascript
 this.component = function registerComponent(name, options) {
    ...
    function factory($injector) {
      ...
      return {
        controller: controller,
        controllerAs: ident,
        template: makeInjectable(template),
        templateUrl: makeInjectable(options.templateUrl),
        transclude: options.transclude,
        scope: {},
        bindToController: options.bindings || {},
        restrict: options.restrict || 'E'
      };
    }
    ...
    return this.directive(name, factory);
  };
```

Programmers might build "_components_" manually with `module.directive`, in such cases it still works.

If there is a directive with the same name than a component, it means that we have actually two directives with the same name. May be in this case the helper function should throw an error.

If it would change in the future, and components become not implemented through directives, it may broke many codes, like many tests, and them must be fixed.

The cool part is that we can add an alias for free called `$directiveController`.

gkalpak:
@drpicox, the "problem" is that multiple directives with the same name is a supported feature (and quite a useful one in several cases). That is the reason why it is not easy to implement this feature for "non-component" directives with predictable behavior.

By "non-component" I mean directives that are not registered using the `component()` helper. Components (directives registered with the `component()` helper) are indeed "plain old" directives, but they are guaranteed to have certain attributes (e.g. `restrict: 'E'`, isolate scope, a controller, a controller alias (an a specific default fallback alias) etc). These attributes allow the implementation of `$componentController` to be predictable in all cases.

drpicox:
@gkalpak for what I understand you are not telling that the suggested implementation for testing is wrong, but the way to implement the directive is. I explain:

- you say that components are guaranteed to have the following properties: restricted to 'E', isolated (it will change in 1.5.0-rc1), with a controller, a controller alias, ... 
- it seems that you imply that you cannot have two components with the same name
- currently with module.component you can change restricted, make non isolated components, make components with replicated names, ...

This code is correct in the current implementation:

```javascript
app.component('myComponent', {
  controller: MyControllerAspect1,
});
app.component('myComponent', {
  controller: MyControllerAspect2,
  restrict: 'A',
});
```

In this case, the current implementation of `$componentController` for 1.5.0-rc1 is not working properly, because it takes one only of them (the last implemented).

This code is [not working](https://docs.angularjs.org/error/$compile/multidir?p0=myDirective&p1=%20(module:%20plunker)&p2=myDirective&p3=%20(module:%20plunker)&p4=%27myDirective%27%20controller&p5=%3Cmy-directive%3E) in 1.4.8: 

```javascript
app.directive('myDirective', function() { 
  return {
    controller: MyControllerAspect1,
    controllerAs: 'aspect1',
  };
});
app.directive('myDirective', function() { 
  return {
    controller: MyControllerAspect2,
    controllerAs: 'aspect2',
  };
});
```

In order to avoid multiple controllers for the same component `restrict` must be forced always to `'E'`. 

If that arrives to happen, then the implementation of `$componentController` should be to look into the list of directives and take the controller of the directive with restrict:'E' or without restrict.

The other point to consider is what happens with [decorators](https://docs.angularjs.org/api/auto/service/$provide#decorator). It is legal to do:

```javascript
app.decorator('myComponentDirective', function($delegate) {
  var directive = $delegate[0];
  directive.controller = MyControllerAspect3;
  return $delegate;
});
```

In such case, the original solution that I try to amend, in addition to save bytes in the angular.min.js, it also is getting the right controller, a controller that must satisfy the tested interface.

So, there are the following ways to proceed:

1. Modify semantics to force all components restrict to 'E' (which seems reasonable to me),
2. Modify the way of creating components to avoid create two components with the same name (code into angular.min.js is required),
3. Accept that can be multiple components with the same name

How to implement `$componentController` in each case:

- Case 1: components must be 'E'
  `$componentController`: Instead of getting the first element from directives it has to look for the directive with no restrict and controller or restrict containing 'E' and controller. It must fail if two directives satisfies this condition (as it would fail the template compilation).

- Case 2: no two components with the same name
  `compile.js`: add an structure and control to fail if two components with the same name are added
  `$componentController`: same implementation to deal with `module.decorator` thing

- Case 3: as currently implemented
  `$componentController`: add one more argument to request the restrictness of the controller, by default 'E'. Then follow as case 1 but with the selected restrictness.

- Case 4: none of above
  That is the option that I said in previous comments: failing if there are more than one directive with the same name it was intended to be consistent in the testing results. The idea is that you want to put multiple components, directives, whatever over the same name (something not common), is up to you to decide how to test it and you can tweak it as much as you like, so you will probably do not use `$componentController`.
  
@petebacondarwin and @gkalpak any suggestion for which case should be applied?

gkalpak:
@drpicox, indeed I missed that you can have several components with the same name. In that case, `$componentController` may not bahave as expected.

Maybe it's a good idea to only allow one component per "restrict" type (maybe even only allow `E` and `A`). The only usecase that `A` is useful is for elements that only accept specific type of children (e.g. `table` cannot have `<my-component>` as a child, but it can have `<tr my-component>`).
Still, it it worth considering to only allow `restrict: 'E'` for component and handle such case using non-component directives.

I think we need to re-think `$componentController` :confused:

drpicox:
Worse that that, rethink `component`. 
Once `component` is defined, solve `$componentController` is just straight forward.

petebacondarwin:
OK, so we should restrict `mod.component` only to elements. That should fix most of the problems, since it would not be possible to have a running application which has an element that matches two components.

petebacondarwin:
Then the `$componentController` could filter out non-element directives. There should only be one of those. If there was not exactly one, then it should throw an error.

drpicox:
So Case 1. I'll try to do the changes later today.

drpicox:
Fixed. 
It should be ok now.

petebacondarwin:
@drpicox I guess you are trying to rebase?

drpicox:
Yep!
https://github.com/angular/angular.js/pull/13742

drpicox:
Rebase is my pending lesson in git.

Issue #13664, opened 2016-01-02
REQUEST FOR FEEDBACK: `angular.component()` - default directive controller name


petebacondarwin:
We should use a consistent default value for the name of a component's directive controller when it is attached to the scope. See https://github.com/angular/angular.js/issues/10007#issuecomment-166704255

Currently we are defaulting to the canonical name of the component. This is not ideal as

a) component names can become long and unwieldy for use in a template
b) it is more complicated to automatically update the template to be used in Angular 2, where the context **is** the controller.

The criteria for the name are:

1) it must be the same for all components
2) it must start with `$`
3) it must be short (2-4 chars)

In addition the name should represent what is actually being published to the scope.

Some of previous suggestions include:

* `vm` - this is the commonly used name in many applications but the controller is not necessarily a "view model"
* `$comp` - this is the current suggestion from the team but can be confused with compare and is not that short
* `$ctrl` - this can be confused with input ConTRoL elements
* `$this` - the controller is not really `this` in the template, since the context is still actually the scope

drpicox:
c) programmers are tempted to use `scope:true` and access to ancestor controllers directly.

petebacondarwin:
@drpicox - I am tempted to say that we ban `isolate: false` for components created using this helper.

drpicox:
I agree, after considering this:

* ```isolate: true``` when restrict is 'E': for me they are truly components, in where $ctrl notation has full meaning
* ```isolate: false``` when restrict is 'A': for me they are _decorators_, a kind of enhancer of the existing components, in this case $ctrl collides so the standard nomenclature it's fine

But I consider that the second case is better to do with directives, _decorators_ are not frequent, usually low level, and not suited for juniors.

So probably is good idea to ban `isolate: false`.

Anyway, a function to get the controller of the 'E' component of the current entity should be good idea, specially to write generic _decorators_.

Narretz:
I like `$cmp`, but find `$comp` even better, because it's clearer.

mgol:
> I like `$cmp`, but find `$comp` even better, because it's clearer.

I like `$comp`. Whenever I see `$cmp` I think "compare".

tabanliviu:
different suggestion: why not have the name of the component as the controller instance name?
ex: <user-profile> -> scope.userProfile

mgol:
@tabanliviu That's how it's currently done on `master` but @petebacondarwin mentioned in this very issue, in the first post, why a common name is better.

[...]

PR #13736, opened 2016-01-11
fix($controller): allow identifiers containing `$`

gkalpak:
As discussed in https://github.com/angular/angular.js/issues/13664#issuecomment-170536024.

### DIFF
--- a/src/ng/controller.js
+++ b/src/ng/controller.js
@@ -3,7 +3,7 @@
 var $controllerMinErr = minErr('$controller');
 
 
-var CNTRL_REG = /^(\S+)(\s+as\s+(\w+))?$/;
+var CNTRL_REG = /^(\S+)(\s+as\s+([\w$]+))?$/;
 function identifierForController(controller, ident) {
   if (ident && isString(ident)) return ident;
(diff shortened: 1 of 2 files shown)


src/ng/controller.js
```diff
@@ -3,7 +3,7 @@
 var $controllerMinErr = minErr('$controller');
 
 
-var CNTRL_REG = /^(\S+)(\s+as\s+(\w+))?$/;
+var CNTRL_REG = /^(\S+)(\s+as\s+([\w$]+))?$/;
```
petebacondarwin: Perhaps only allow `$` at the beginning or is that being too strict?

gkalpak: Yeah, it crossed my mind, but I thought it was too strict. It complicates things unnecessarily, I think.
(Note that it's already possible to use many more characters using `controllerAs`.)

gkalpak:
@petebacondarwin, do you think we should backport it to `v1.4.x` ?

petebacondarwin:
Yep

gkalpak:
Backported to `v1.4.x` as 2563ff7.

PR #13739, opened 2016-01-11
test: fix ms edge failures

gkalpak:
This is a backport of #13686 to v1.4.x.
It includes exactly the same changes with the exception of the fixes to `$sanitizeSpec`, which do not apply to `v1.4.x` (and were omitted). Unit-tests pass on MS Edge.


gkalpak:
Merged as da04571 and 0a641c0.

Issue #13603, opened 2015-12-21
module.component should restrict to 'AE' by default


davidreher:
Hey,

I tried using the new 1.5 component router and component api. While doing this I created my components without an explicit restrict attribute. [The loader restricts them to 'E' by default](https://github.com/angular/angular.js/blob/master/src/loader.js#L426). 

The component router (for angular1?) creates new components [via attributes syntax](https://github.com/angular/angular/blob/master/modules/angular1_router/src/ng_outlet.ts#L158).

Please change the default restriction to 'AE' or change the component router ;)

fesor:
I think that this is issue of angular/router rather that angular. Components should be only elements (by default).

davidreher:
I just saw, that [pull request 5278](https://github.com/angular/angular/pull/5278) fixes this issue ... Hope it gets merged soon ;)

Narretz:
Please don't close the issue, this is usually where most of the discussion happens.

davidreher:
Hmm, I thought it would be best to not have this discussion in multiple threads (here, in the angular2 repo bug and the pull request) ...

petebacondarwin:
Components should be E only - we need to fix the router.

PR #13741, opened 2016-01-11
`component()` helper changes

petebacondarwin:


gkalpak:
[The docs][1] still seem to imply that `restrict: 'E'` is just a configurable default (rather than an enforced value). Maybe we should be more explicit about it being non-configurable (e.g. remove the `default bahavior like` part).

LGTM otherwise.

[1]: https://github.com/angular/angular.js/pull/13741/files#diff-a732922b631efed1b9f33a24082ae0dbR983

petebacondarwin:
OK, I'll improve the docs.

Mischi:
removing the ability to set the `restrict` option makes angular.component()'s unusable with the new ngComponentRouter until the router will support `elements` next to `attributes` ...

See #13603

petebacondarwin:
@Mischi that is actually a problem with the router that needs to be fixed. @btford ?

AIMMOTH:
Why remove element? How do you preserve DOM structure which is important for SEO?

petebacondarwin:
@AIMMOTH not sure I understand what you mean.

AIMMOTH:
@petebacondarwin Recommendations is that the DOM structure should be arranged for SEO. For instance <article></article> should be used. If this article is a component, it was possible to add the component as attribute <article component></article> but now you need to destroy DOM structure with <component></component>

Or have I missed something?

PR #13744, opened 2016-01-12
use forEach instead of Object.keys in compiler Attributes constructor

jiubao:
use forEach instead of Object.keys in compiler Attributes constructor

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1193,13 +1193,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     var specialAttrHolder = document.createElement('div');
     var Attributes = function(element, attributesToCopy) {
       if (attributesToCopy) {
-        var keys = Object.keys(attributesToCopy);
-        var i, l, key;
-
-        for (i = 0, l = keys.length; i < l; i++) {
-          key = keys[i];
-          this[key] = attributesToCopy[key];
-        }
+        var self = this;
+        forEach(attributesToCopy, function(value, key) {
+          self[key] = value;
+        });
       } else {
         this.$attr = {};
       }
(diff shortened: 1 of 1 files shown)


jiubao:
@googlebot I signed it!

gkalpak:
I believe `Object.keys() + for-loop` was used for performance reasons.
What is the reason (other than readability maybe) for changing it ?

jiubao:
@gkalpak:
1. Are there so many cases that `attributesToCopy` has a deep prototype inheritance structure?
2. Suppose it is the case, why not refactor `forEach` to use `Object.keys` instead of `for in` in the last two `else`:
```javascript
function forEach(obj, iterator, context) {
    ...
    } else if (typeof obj.hasOwnProperty === 'function') {
      // Slow path for objects inheriting Object.prototype, hasOwnProperty check needed
      for (key in obj) {
        if (obj.hasOwnProperty(key)) {
          iterator.call(context, obj[key], key, obj);
        }
      }
    } else {
      // Slow path for objects which do not have a method `hasOwnProperty`
      for (key in obj) {
        if (hasOwnProperty.call(obj, key)) {
          iterator.call(context, obj[key], key, obj);
        }
      }
    }
    ...
```
so that it does not have the `for in` performance issue, meanwhile get `forEach` benefits:
* support `obj.forEach`:
```javascript
    } else if (obj.forEach && obj.forEach !== forEach) {
        obj.forEach(iterator, context, obj);
    }
```
* may be faster in `isBlankObject` case:
```javascript
    } else if (isBlankObject(obj)) {
      // createMap() fast path --- Safe to avoid hasOwnProperty check because prototype chain is empty
      for (key in obj) {
        iterator.call(context, obj[key], key, obj);
      }
    }
```
)

petebacondarwin:
Here is the relevant commit: https://github.com/jiubao/angular.js/commit/54fa16e45d8769ce6708a28388326db0eea53c7e

`attributesToCopy` always has extra properties up the prototype chain.

jiubao:
OK, thanks for the explains!

PR #13752, opened 2016-01-12
docs(input[number]): describe your change...

zachlysobey:
Changed `input[type=number]` min and max attributes to take `number` rather than `string`.

Apologies if I'm confused about this, perhaps since they can take dates or numbers it needs to be a string? (or something?)


zachlysobey:
@robwormald [2:46 PM] 
I think that's an attr which is a string anyway

Issue #5893, opened 2014-01-20
feat(directive): Add ability to inject required controllers into the controller function


maklemenz:
**quick overview**
You have the directives A, B and C
B requires A, 
C requires B

The methods in C need to call methods in the controller of B.
The methods in the controller of B need call methods in the controller of A.

The controller of B cannot know the controller of A because it can only be passed into the link-function of B.

**more details**
http://stackoverflow.com/questions/21231294/angularjs-inject-required-directive-controller-into-the-controller-instead-of-t/21231428

Actually there is a workaround that is too complicated for such an easy problem.

Even the ngModelController does use another workaround with a feature that seems to be undocumented: the binding of directive controllers to the directive-elements data property:
https://github.com/angular/angular.js/blob/5adea0ba64b758f8f5700da73f6faca8ae1983be/src/ng/directive/input.js#L969

thorn0:
> I feel like that is a reasonable api and it doesn't complicate the DDO further.

Actually, it simplifies the DDO. Compare the two snippets I wrote above.

ewinslow:
@kentcdodds see #2095

kentcdodds:
@thorn0, I was referring to it making the DDO more complex for `module.directive` users, not `angular.component`. Either way :+1:

@ewinslow, I decided against asynchronous directives. There are several reasons it just wouldn't work well.

piernik:
I think that injecting required controller as local is "have to" in new version.

SonofNun15:
:+1:
We are wrestling with this same issue. We try to do very little in link functions because they are harder to test. We tried putting `scope.myController = requiredController` in the prelink function so that the controller of the required directive would be available in the child directive's controller, but it turns out that the controller runs before both the pre AND post link functions, so this isn't going to work.

piernik:
Any news on this?

programmist:
+1.

As a work-around I'm constructing the controller and adding it to the scope within the link function, but that's pretty ugly.  It still beats unit testing link functions though.

programmist:
+1

djdmbrwsk:
+1

hardsetting:
+1

thorn0:
To simplify the implementation, we can introduce a restriction that only the controllers from the element's parents (`^^` and `?^^`) can be injected.

ciccia:
+1

abyx:
:+1:

DumboJet:
+1

[...]

PR #13760, opened 2016-01-13
feat($compile): add `controllersReady` option to `component` helper

petebacondarwin:
A potential solution to #5893


petebacondarwin:
I spoke with @Narretz about this and we have a better idea...

PR #8401, opened 2014-07-29
feat(getControllers): allow objects for require

PatrickJS:
allowing objects, similar to $Q.all, read out more clearly for example
current ngModelDirective

``` javascript
var ngModelDirective = function() {
  return {
    restrict: 'A',
    require: ['ngModel', '^?form', '^?ngModelOptions'],
    controller: NgModelController,
    link: {
      pre: function(scope, element, attr, ctrls) {
        // Pass the ng-model-options to the ng-model controller
        if (ctrls[2]) {
          ctrls[0].$options = ctrls[2].$options;
        }

        // notify others, especially parent forms

        var modelCtrl = ctrls[0],
            formCtrl = ctrls[1] || nullFormCtrl;

        formCtrl.$addControl(modelCtrl);

        scope.$on('$destroy', function() {
          formCtrl.$removeControl(modelCtrl);
        });
      },
      post: function(scope, element, attr, ctrls) {
        var modelCtrl = ctrls[0];
        if (modelCtrl.$options && modelCtrl.$options.updateOn) {
          element.on(modelCtrl.$options.updateOn, function(ev) {
            scope.$apply(function() {
              modelCtrl.$$debounceViewValueCommit(ev && ev.type);
            });
          });
        }

        element.on('blur', function(ev) {
          scope.$evalAsync(function() {
            modelCtrl.$setTouched();
          });
        });
      }
    }
  };
};
```

ngModelDirective with a require object

``` javascript
var ngModelDirective = function() {
  return {
    restrict: 'A',
    require: {
      ngModel: 'ngModel',
      form: '^?form',
      ngModelOptions: '^?ngModelOptions'
    },
    controller: NgModelController,
    link: {
      pre: function(scope, element, attr, ctrls) {
        // Pass the ng-model-options to the ng-model controller
        if (ctrls.ngModelOptions) {
          ctrls.ngModel.$options = ctrls.ngModelOptions.$options;
        }

        // notify others, especially parent forms

        var modelCtrl = ctrls.ngModel,
            formCtrl = ctrls.form || nullFormCtrl;

        formCtrl.$addControl(modelCtrl);

        scope.$on('$destroy', function() {
          formCtrl.$removeControl(modelCtrl);
        });
      },
      post: function(scope, element, attr, ctrls) {
        var modelCtrl = ctrls.ngModel;
        if (modelCtrl.$options && modelCtrl.$options.updateOn) {
          element.on(modelCtrl.$options.updateOn, function(ev) {
            scope.$apply(function() {
[...]

PR #13763, opened 2016-01-13
`$onInit` and friends

petebacondarwin:


src/ng/compile.js
```diff
@@ -255,12 +265,22 @@
...
  * Require another directive and inject its controller as the fourth argument to the linking function. The
- * `require` takes a string name (or array of strings) of the directive(s) to pass in. If an array is used, the
- * injected argument will be an array in corresponding order. If no such directive can be
- * found, or if the directive does not have a controller, then an error is raised (unless no link function
- * is specified, in which case error checking is skipped). The name can be prefixed with:
+ * `require` property can be a string, an array or an object:
+ * * a **string** containing the name of the directive to pass to the linking function
+ * * an **array** containing the names of directives to pass to the linking function. The argument passed to the
+ * linking function will be an array of controllers in the same order as the names in the `require` property
+ * * an **object** whose property values are the names of the directibes to pass to the linking function. The argument
```
gkalpak: directibes  -->  directives

src/ng/compile.js
```diff
@@ -255,12 +265,22 @@
...
+ * passed to the linking function will also be an object with matching keys, whose values will be the corresponding
+ * controller.
```
gkalpak: controller  -->  controllers (?)

src/ng/compile.js
```diff
@@ -2386,8 +2412,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             removeControllerBindingWatches =
               initializeDirectiveBindings(controllerScope, attrs, controller.instance, bindings, controllerDirective);
           }
+
+          if (isObject(controllerDirective.require) && !isArray(controllerDirective.require)) {
+            var controllerObj = getControllers(name, controllerDirective.require, $element, elementControllers);
+            extend(controller.instance, controllerObj);
```
gkalpak: Doesn't this mean that controllers whose constructors return a value won't be properly handled ?
Do we really need to set the required controllers as properties on the requiring controller ? How about just passing the object as argument to `$onInit` ?

test/ng/compileSpec.js
```diff
@@ -9435,5 +9530,5 @@ describe('$compile', function() {
         }));
       });
     });
-  });
+   });
```
gkalpak: I am on mobile (so I can't be sure), but tjis change doesn't seem right...

gkalpak:
With the exception of https://github.com/angular/angular.js/pull/13763#discussion_r49637240 (and a couple of nitpicks), this L(pretty)GTM :+1:

src/ng/compile.js
petebacondarwin: By the time we get here we have already dealt with constructors who return a value (around lines 2406-2414). I have added a test to prove this:

petebacondarwin: I didn't like to add parameters to `$onInit`. If we don't want to automatically bind then, instead, we should be injecting things into the constructor of the controller, like in ng2... But that is done differently for non-ancestor components

test/ng/compileSpec.js
```diff
@@ -4244,6 +4244,22 @@ describe('$compile', function() {
       if (!/chrome/i.test(navigator.userAgent)) return;
```
lgalfaso: unrelated to this issue, but can we change this to `if (!support.classes) return;` ?

petebacondarwin: On it [sic]!

petebacondarwin: Actually, it is not so easy @lgalfaso since the latest Safari (9.0.2) is even worse for `toString` on a a class instance.

src/ng/compile.js
```diff
@@ -255,12 +265,26 @@
...
+ * * a **string** containing the name of the directive to pass to the linking function
+ * * an **array** containing the names of directives to pass to the linking function. The argument passed to the
+ * linking function will be an array of controllers in the same order as the names in the `require` property
+ * * an **object** whose property values are the names of the directives to pass to the linking function. The argument
+ * passed to the linking function will also be an object with matching keys, whose values will hold the corresponding
+ * controllers.
+ *
+ * If the `require` property is an object and the directive provides a controller, then the required controllers are
+ * bound to the controller using the keys of the `require` property. See the {@link $compileProvider#component} helper
+ * for an example of how this can be used.
```
Narretz: Should note that the controllers are only available once $onInit has been called, no?

src/ng/compile.js
```diff
@@ -255,12 +265,26 @@
  *    The `$transclude` function also has a method on it, `$transclude.isSlotFilled(slotName)`, which returns
  *    `true` if the specified slot contains content (i.e. one or more DOM nodes).
  *
+ * The controller can provide the following methods that act as life-cycle hooks:
```
Narretz: Should we mention that this is basically a new phase that happens before both pre and postLink?

Narretz:
Now that require can be an object, should we pass the ctrls as an object to the link function like described here: https://github.com/angular/angular.js/pull/8401?

Narretz:
Don't forget that these issues are fixed by the PR:
https://github.com/angular/angular.js/pull/6040
https://github.com/angular/angular.js/issues/5893

petebacondarwin:
> Now that require can be an object, should we pass the ctrls as an object to the link function like described here: #8401? (Yes from me)

@Narretz One of the commits in this PR does actually do that - but only when the `require` property is an object itself.

src/ng/compile.js
gkalpak: I meant if one of the **required** controllers returns a value, then we might extend **this** controller with the wrong value (depends on the order of execution of the directives).

(Your test proves that it works if **this** controller returns a value.)

petebacondarwin: You are right! I have fixed it and added another test in 1d18df20fe6dd0ab50f9285f03ff979f302f5874

gkalpak: LGTM

src/ng/compile.js
petebacondarwin: Updated

src/ng/compile.js
petebacondarwin: Updated

src/ng/compile.js
```diff
@@ -2388,6 +2494,21 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
         }
 
+        // Bind the required controllers to the controller, if `require` is an object
```
Narretz: ... and bindToController is truthy

src/ng/compile.js
```diff
@@ -218,8 +218,18 @@
  * definition: `controller: 'myCtrl as myAlias'`.
  *
  * When an isolate scope is used for a directive (see above), `bindToController: true` will
- * allow a component to have its properties bound to the controller, rather than to scope. When the controller
- * is instantiated, the initial values of the isolate scope bindings will be available if the controller is not an ES6 class.
+ * allow a component to have its properties bound to the controller, rather than to scope.
+ *
+ * After the controller is instantiated, the initial values of the isolate scope bindings will bound to the controller
```
gkalpak: will bound  -->  will be? bound

src/ng/compile.js
```diff
@@ -255,12 +265,28 @@
...
+ * passed to the linking function will also be an object with matching keys, whose values will hold the corresponding
+ * controllers.
+ *
+ * If the `require` property is an object and `bindToController` is truthy, then the required controllers are
+ * bound to the controller using the keys of the `require` property. This binding occurs after all the controllers
+ * have been constructed but before `$onInit` is called.
+ * See the {@link $compileProvider#component} helper for an example of how this can be used.
+ *
+ * If no such required directive(s) can be found, or if the directive does not have a controller, then an error is
+ * raised (unless no link function is specified, in which case error checking is skipped). The name can be prefixed with:
```
gkalpak: Is error checking indeed skipped if no link function is specified ?

src/ng/compile.js
gkalpak: Nitpick: Comment could also mention "bindToController" as a requirement.

test/ng/compileSpec.js
```diff
@@ -4280,6 +4283,7 @@ describe('$compile', function() {
         element = $compile('<div foo-dir dir-data="remoteData" ' +
                                  'dir-str="Hello, {{whom}}!" ' +
                                  'dir-fn="fn()"></div>')($rootScope);
+        expect(Controller.prototype.$onInit).toHaveBeenCalled();
         element.data('$fooDirController').check();
```
gkalpak: Isn't this redundant now ?

test/ng/compileSpec.js
```diff
@@ -4287,6 +4291,7 @@ describe('$compile', function() {
     });
 
 
+
```
gkalpak: Too much whitespace ?

Narretz:
I just noticed the component method doesn't accept the templateNamespace property, is this intentional?

test/ng/compileSpec.js
```diff
@@ -5307,6 +5338,195 @@ describe('$compile', function() {
...
+      angular.module('my', [])
+        .directive('me', function() {
+          return {
+            restrict: 'E',
+            scope: {},
+            controllerAs: '$ctrl',
+            require: { container: '^parent', friend: 'sibling' },
+            bindToController: true,
+            controller: MeController,
+            controllerAs: '$ctrl'
```
gkalpak: Two `controllerAs` occurrences (where 0 would be sufficient ?).

test/ng/compileSpec.js
```diff
@@ -5307,6 +5338,195 @@ describe('$compile', function() {
...
+      inject(function($compile, $rootScope, meDirective) {
+        element = $compile('<parent><me sibling></me></parent>')($rootScope);
+        expect(MeController.prototype.$onInit).toHaveBeenCalled();
+        expect(parentController).toEqual(jasmine.any(ParentController));
+        expect(siblingController).toEqual(jasmine.any(SiblingController));
+      });
+    });
+
+
+    it('should not bind required controllers bindToController is falsy', function() {
```
gkalpak: Missing "if".

test/ng/compileSpec.js
```diff
@@ -5307,6 +5338,195 @@ describe('$compile', function() {
...
+      inject(function($compile, $rootScope, meDirective) {
+        element = $compile('<parent><me sibling></me></parent>')($rootScope);
+        expect(meController.$onInit).toHaveBeenCalled();
+        expect(parentController).toEqual(jasmine.any(ParentController));
+        expect(siblingController).toEqual(jasmine.any(SiblingController));
+      });
+    });
+
+
+    it('should bind required controllers to controllers that return an explicit constructor return value', function() {
```
gkalpak: Nitpick: I'd add a priority to `me` (1 would do), in order to make the test more explicit (and future-proof).

Explanation:
Currently, this test relies on the fact that the directives have default priorities, which means that the order of execution of directives on the same element is determined by their alphabetical order. Thus `me` gets a "higher
 priority than `sibling` and its controller is instantiated first.
<sub>(I said it's a nitpick :stuck_out_tongue: )</sub>

src/ng/compile.js
petebacondarwin: indeed - the error checking is inside `getControllers` which is not called if no link function (and now also no bindToController+require-as-object)

src/ng/compile.js
petebacondarwin: done

src/ng/compile.js
petebacondarwin: done

test/ng/compileSpec.js
petebacondarwin: yep! removed

test/ng/compileSpec.js
petebacondarwin: removed

test/ng/compileSpec.js
petebacondarwin: fixed

test/ng/compileSpec.js
petebacondarwin: OK, fair enough

src/ng/compile.js
gkalpak: I meant don't we need to mention that error-checking is skipped if no link function **AND** no bindToController+requireAsObject ?

petebacondarwin: fixed

test/ng/compileSpec.js
petebacondarwin: Done

petebacondarwin:
> I just noticed the component method doesn't accept the templateNamespace / multiElement property, is this intentional?

@Narretz - I think we should try to keep the component API as simple as possible. These two are fairly uncommon, right? If people really need them they can complain and we can add them later.

Narretz:
@petebacondarwin Sounds good

Narretz:
There's a real travis failure btw:
> Chrome 45.0.2454 (Linux 0.0.0) $compile isolated locals should eventually expose isolate scope variables on ES6 class controller with controllerAs when bindToController is true FAILED

Although the test wasn't introduced by this PR.

petebacondarwin:
@Narretz I have fixed that failing test.

test/ng/compileSpec.js
lgalfaso: it looks like there are at least 2 issues reported on JavascriptCore (Safari JS engine) regarding this
https://bugs.webkit.org/show_bug.cgi?id=149743
https://bugs.webkit.org/show_bug.cgi?id=144285

I think that some other strategy. Meanwhile, just keep the old code

ev45ive:
This looks very nice and clean. Should be like this from beggining. Very nice work. Do you think $onInit with bound required controllers and component() syntax will get on board with 1.5 release? I'd love to use it in production - it would clean up a lot of messy directive code. :-)

PR #13769, opened 2016-01-14
fix($compile): handle boolean attributes in `@` bindings

### LINKED ISSUE #13767: input[checkbox] doesn't catch checked in scope directive
The attribute "checked" is not recovered in the scope of a directive.
This seems broken since version 1.4.4
https://plnkr.co/edit/0j7Y9mkeP4m2waCLYLxU?p=preview
Working on 1.3.15
https://plnkr.co/edit/GN4K5EdtfHxPpe7x3OCc?p=preview

petebacondarwin:
Closes #13767

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2969,10 +2969,15 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               }
             });
             attrs.$$observers[attrName].$$scope = scope;
-            if (isString(attrs[attrName])) {
+            lastValue = attrs[attrName];
+            if (isString(lastValue)) {
               // If the attribute has been provided then we trigger an interpolation to ensure
               // the value is there for use in the link fn
-              destination[scopeName] = $interpolate(attrs[attrName])(scope);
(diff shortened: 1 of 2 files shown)


petebacondarwin:
To be backported to 1.4

Narretz:
Looks good. However, can you add a comment that explains why we have to do that? I'm not convinced it's angular that is modifying these boolean attributes. Isn't it the browser that converts the content of boolean attributes to "true"?

petebacondarwin:
Here is plunker that shows the problem is fixed by this PR: https://plnkr.co/edit/mk9BJODWJdYFhMpw3nGa?p=preview

petebacondarwin:
@Narretz here is the line where the boolean conversion happens I believe: https://github.com/angular/angular.js/blob/master/src/ng/directive/attrs.js#L339

Narretz:
You are right, thanks for the link. Can't hurt to put it in a little comment, though ;)

petebacondarwin:
Will do!

PR #13773, opened 2016-01-15
fix(form): make ngForm $pristine after nested control.$setPristine() (counter version)

### LINKED ISSUE #13715: ngForm stays $dirty after control.$setPristine()
Hi there. Frustrating problem. in an ngForm, if in my controller I set one of the form's controls to $dirty (via control.$setDirty()), then in the controller set the same control to $pristine, the form remains $dirty.

1) on state load, form is $pristine
2) via another event, call controller function that does control.$setDirty() - form is now $dirty
3) via another event, call controller function that does control.$setPristine() - form is still $dirty

I'm pretty sure this shouldn't happen, and is getting kludgy to fix. I'm using Angular 1.4.8.

Plunkr:
https://plnkr.co/edit/6noEtj6vbMkqxiKK6Yp7?p=preview

Thanks

martinschulzofficial:
When calling $setPristine on the nested form or control,
form becomes $pristine of all the nested controls are pristine

Closes #13715

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -5,6 +5,7 @@
 var nullFormCtrl = {
   $addControl: noop,
   $$renameControl: nullFormRenameControl,
+  $$updatePristine: noop,
   $removeControl: noop,
   $setValidity: noop,
   $setDirty: noop,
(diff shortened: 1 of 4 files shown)


test/ng/directive/formSpec.js
```diff
@@ -714,7 +714,8 @@ describe('form', function() {
       expect(form.$error.maxlength[0].$name).toBe('childform');
 
       inputController.$setPristine();
-      expect(form.$dirty).toBe(true);
+      // this assertion prevents to propagate prestine to the parent form
+      // expect(form.$dirty).toBe(true);
```
gkalpak: You should change this to `expect(form.$dirty).toBe(false);` and remove the subsequent call to `form.$setPristine()`.

test/ng/directive/formSpec.js
```diff
@@ -1043,6 +1044,80 @@ describe('form', function() {
...
+      expect(form).not.toBePristine();
+
+      childFormCtrl1.$setPristine();
+      scope.$apply();
+      expect(form).not.toBePristine();
+
+      childFormCtrl2.$setPristine();
+      scope.$apply();
+      expect(form).toBePristine();
+    });
```
gkalpak: I'd add a test to make sure a change propagates from a control through it's parent form to the outer-most form.

src/ng/directive/form.js
```diff
@@ -254,19 +255,46 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
...
+    var isPristine = true,
+        controlsLength = controls.length,
+        i;
+
+    for (i = 0; i < controlsLength; i++) {
+      if (!controls[i].$pristine) {
+        isPristine = false;
+        break;
+      }
+    }
```
gkalpak: Nitpicking: The above lines can be "summarized" as

`var isPristine = controls.every(function(control) { return control.$pristine; });`

gkalpak:
It generally LGTM (with a couple of nitpicks), BUT:

We need to properly handle added/removed controls.

linoleum-js:
Now it uses internal counter. It's a bit complicated, because we have to divide `$setPristine` propagtion into *capturing* and *bubbling*.
Also I'm not 100% sure that idempotence test cases are sufficient (I think this is crucial).

gkalpak:
I think it's best to have the two alternative approaches as two independent PRs (so we can review/update/decide upon separately).
Could you split revert the second commit and put it into a separate PR ?

Thx @linoleum-js for working on this, btw :+1:

petebacondarwin:
Has this been split?

lucasmaj (approved):

PR #13784, opened 2016-01-17
Increase performance when checking if element is animatable

Narretz:
Built upon PR #13783

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -571,6 +571,13 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       return getDomNode(nodeOrElmA) === getDomNode(nodeOrElmB);
     }
 
+    /**
+     * This fn returns false if any of the following is true:
+     * a) animations on any parent element are disabled, and animations on the element aren't explicitly allowed
+     * b) a parent element has an ongoing structural animation, and animateChildren is false
+     * c) the element is not a child of the body
+     * d) the element is not a child of the $rootElement
+     */
     function areAnimationsAllowed(element, parentElement, event) {
       var bodyElement = jqLite($document[0].body);
       var bodyElementDetected = isMatchingElement(element, bodyElement) || element[0].nodeName === 'HTML';
@@ -604,10 +611,12 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
         if (!parentAnimationDetected) {
           var parentElementDisabled = disabledElementsLookup.get(parentNode);
 
-          // disable animations if the user hasn't explicitly enabled animations on the
-          // current element
           if (parentElementDisabled === true && elementDisabled !== false) {
+            // disable animations if the user hasn't explicitly enabled animations on the
+            // current element
             elementDisabled = true;
+            // element is disabled via parent element, no need to check anything else
+            break;
           } else if (parentElementDisabled === false) {
             elementDisabled = false;
           }
@@ -625,9 +634,15 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
         if (parentAnimationDetected && animateChildren === false) break;
 
         if (!bodyElementDetected) {
-          // we also need to ensure that the element is or will be apart of the body element
+          // we also need to ensure that the element is or will be a part of the body element
           // otherwise it is pointless to even issue an animation to be rendered
           bodyElementDetected = isMatchingElement(parentElement, bodyElement);
+        }
+
+        if (bodyElementDetected && rootElementDetected) {
+          // If both body and root have been found, any other checks are pointless,
+          // as no animation data should live outside the application
+          break;
         }
 
         if (!rootElementDetected) {
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -624,23 +624,20 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
         // there is no need to continue traversing at this point
         if (parentAnimationDetected && animateChildren === false) break;
 
-        if (!rootElementDetected) {
-          // angular doesn't want to attempt to animate elements outside of the application
-          // therefore we need to ensure that the rootElement is an ancestor of the current element
-          rootElementDetected = isMatchingElement(parentElement, $rootElement);
-          if (!rootElementDetected) {
-            parentHost = parentElement.data(NG_ANIMATE_PIN_DATA);
-            if (parentHost) {
-              parentElement = parentHost;
-              rootElementDetected = true;
-            }
-          }
-        }
-
         if (!bodyElementDetected) {
           // we also need to ensure that the element is or will be apart of the body element
           // otherwise it is pointless to even issue an animation to be rendered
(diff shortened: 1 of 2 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -1543,6 +1543,49 @@ describe("animations", function() {
       });
     });
 
+    they('should not animate an element when the pinned element (which is the $prop element), is pinned to an element that is not a child of the $rootElement',
+      ['same', 'parent', 'grandparent'],
+      function(elementRelation) {
+        inject(function($animate, $compile, $document, $rootElement, $rootScope) {
```
gkalpak: `$compile` is not used.
(Yes, in other tests too :smiley:)

test/ngAnimate/animateSpec.js
```diff
@@ -1543,6 +1543,49 @@ describe("animations", function() {
...
+        $rootScope.$digest();
+        expect(capturedAnimation).toBeFalsy();
+
+        $animate.pin(pinElement, pinTargetElement);
+
+        $animate.addClass(animateElement, 'blue');
+        $rootScope.$digest();
+        expect(capturedAnimation).toBeFalsy();
+
+        dealoc(pinElement);
```
gkalpak: Why do we only dealoc `pinElement` ?
What about `pinTargetElement` or `innerParent`.

(Other tests seem inconsistent too (unless I'm missing something.)

src/ngAnimate/animateQueue.js
```diff
@@ -579,6 +586,11 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
       var animateChildren;
       var elementDisabled = disabledElementsLookup.get(getDomNode(element));
 
+      if (elementDisabled === true) {
+        // If animations on the element have been disabled explicitly, no need to traverse the parents
+        return false;
+      }
```
gkalpak: AFAICT, `elementDisabled` could be still set to `false` on [line 626][1], so this change is either invalid or also a fix (in addition to perf improvement). I believe it is the latter (in which case a test should be added).

[1]: https://github.com/angular/angular.js/pull/13784/files#diff-c15d87c15661cba4a721501ba4fda5b2R626

test/ngAnimate/animateSpec.js
```diff
@@ -1543,6 +1543,49 @@ describe("animations", function() {
       });
     });
 
+    they('should not animate an element when the pinned element (which is the $prop element), is pinned to an element that is not a child of the $rootElement',
```
gkalpak: This line is too long :smiley:"

src/ngAnimate/animateQueue.js
```diff
@@ -624,25 +638,29 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
...
-              rootElementDetected = true;
-            }
+        if (!rootElementDetected && !parentHost) {
+          // If no rootElement is detected, and no parentHost has been found already,
+          // check if the parentElement is pinned to another element
+          parentHost = parentElement.data(NG_ANIMATE_PIN_DATA);
+          if (parentHost) {
+            // The pin target element becomes the parent element
+            parentElement = parentHost;
+            rootElementDetected = isMatchingElement(parentElement, $rootElement);
```
gkalpak: This change assumes that there can only be one "pinning" on the path from any element to the rootElement.
E.g. one can't pin element A to element B and element B to element C (where only C is a child of the rootElement).

Although this was not explicitly supported (nor do I know why one might want to do it), it is a potential BC.

src/ngAnimate/animateQueue.js
```diff
@@ -624,25 +638,29 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
...
-          // angular doesn't want to attempt to animate elements outside of the application
-          // therefore we need to ensure that the rootElement is an ancestor of the current element
-          rootElementDetected = isMatchingElement(parentElement, $rootElement);
-          if (!rootElementDetected) {
-            parentHost = parentElement.data(NG_ANIMATE_PIN_DATA);
-            if (parentHost) {
-              parentElement = parentHost;
-              rootElementDetected = true;
-            }
+        if (!rootElementDetected && !parentHost) {
```
gkalpak: This change assumes that there can only be one "pinning" on the path from any element to the rootElement.
E.g. one can't pin element A to element B and element B to element C (where only C is a child of the rootElement).

Although this was not explicitly supported (nor do I know why one might want to do it), it is a potential BC.

Narretz: Good catch. I think we can add this as a reasonable breaking change. So far, pin() was broken for more cases than it was working for, so adoption doesn't seem to high.

test/ngAnimate/animateSpec.js
Narretz: Really? Shouldn't jshint / jscs report this then? I thought we were at 200 chars.

src/ngAnimate/animateQueue.js
Narretz: But line 626 is inside the loop, and this check comes before the loop. Also, this is checked before areAnimationsAllowed, too. The lookup here was added later to catch elements that were explicitly allowed.

Narretz: I'll remove this, because this is checked already at https://github.com/angular/angular.js/pull/13784/files#diff-c15d87c15661cba4a721501ba4fda5b2R342

Narretz:
@gkalpak I've reverted the change, and recursive pinning is now allowed again. We can always change that later. And I've found and fixed another bug with the host element detection. Can you please review again?

test/ngAnimate/animateSpec.js
gkalpak: Wasn't it at 100 chars ? Did that change and I missed that ?
Imo, 100 chars makes sense because it makes side-by-side diffing much easier.

test/ngAnimate/animateSpec.js
```diff
@@ -1543,34 +1547,92 @@ describe("animations", function() {
       });
     });
 
-    it('should adhere to the disabled state of the hosted parent when an element is pinned',
-      inject(function($animate, $compile, $document, $rootElement, $rootScope) {
+    they('should not animate an element when the pinned element (which is the $prop element), is pinned to an element that is not a child of the $rootElement',
+      ['same', 'parent', 'grandparent'],
+      function(elementRelation) {
+        inject(function($animate, $compile, $document, $rootElement, $rootScope) {
```
gkalpak: `$compile` is not used (here and below).

test/ngAnimate/animateSpec.js
```diff
@@ -1543,34 +1547,92 @@ describe("animations", function() {
...
+        expect(capturedAnimation).toBeFalsy();
 
-      dealoc(element);
-    }));
+        dealoc(pinElement);
+      });
+    });
+
+    they('should adhere to the disabled state of the hosted parent when the $prop element is pinned',
+      ['parent'],
```
gkalpak: Why `they` if just using one value ?

Narretz:
@gkalpak Thanks for the review. Should be all good now.

gkalpak:
LGTM :smiley:

PR #13787, opened 2016-01-18
fix($animate): cancel fallback timeout when animation ends normally

Narretz:


Narretz:
@matsko This is something I noticed by chance. I wonder if it was intentional that the timeout is not explicitly canceled, but that it cleans up after itself?

Narretz:
@matsko Did you have time to look at this (or ask Mishko)?

PR #13792, opened 2016-01-18
perf($compile): avoid needless overhead when wrapping text nodes

dcherman:
This commit positively affects performance in two main ways:

1,  When wrapping text nodes in the compile step, we do not need the overhead
of the `forEach` function versus a normal for loop since we do not make
use of the closure for anything.
2.  When actually wrapping the node, we can completely bypass jqLite which
avoids several function calls and the overhead of cloning the wrapper node
which we already know to be unique.

Tests in applications show about an 83% decrease in time spent in this
specific loop.  During the bootstrap phase of one large application at my company, the time spent in this loop went from about 60ms down to 9-10ms since this is a pretty hot piece of code.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1494,13 +1494,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         // modify it.
         $compileNodes = jqLite($compileNodes);
       }
+
+      var NOT_EMPTY = /\S+/;
+
       // We can not compile top level text elements since text nodes can be merged and we will
       // not be able to attach scope data to them, so we will wrap them in <span>
-      forEach($compileNodes, function(node, index) {
-        if (node.nodeType == NODE_TYPE_TEXT && node.nodeValue.match(/\S+/) /* non-empty */) {
-          $compileNodes[index] = jqLite(node).wrap('<span></span>').parent()[0];
+      for (var i = 0, len = $compileNodes.length; i < len; i++) {
+        var domNode = $compileNodes[i];
+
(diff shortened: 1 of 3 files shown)


dcherman:
Not sure how to re-trigger the CI tests, but the failure looks to be a Selenium//Webdriver problem versus an actual failure.

Narretz:
Thansk for the PR! I've restarted the test.
@mgol Is this something you'd be interested in reviewing?

src/ng/compile.js
```diff
@@ -1494,13 +1494,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
       // not be able to attach scope data to them, so we will wrap them in <span>
-      forEach($compileNodes, function(node, index) {
-        if (node.nodeType == NODE_TYPE_TEXT && node.nodeValue.match(/\S+/) /* non-empty */) {
-          $compileNodes[index] = jqLite(node).wrap('<span></span>').parent()[0];
+      for (var i = 0, len = $compileNodes.length; i < len; i++) {
+        var domNode = $compileNodes[i];
+
+        if (domNode.nodeType === NODE_TYPE_TEXT && domNode.nodeValue.match(NOT_EMPTY) /* non-empty */) {
+          jqLiteWrapNode(domNode, document.createElement('span'));
+          $compileNodes[i] = domNode.parentNode;
```
petebacondarwin: These two lines can actually be:

```js
jqLiteWrapNode(domNode, $compileNodes[i] = document.createElement('span'));
```

petebacondarwin:
Other than that, it looks good to me. Merging

petebacondarwin:
Landed as 92e4801d88fbe9b7ef719fd3d0175d85420e1cc4

mgol:
:+1:

PR #13798, opened 2016-01-19
fix(form): make ngForm $pristine after nested control.$setPristine() (brute force version)

### LINKED ISSUE #13715: ngForm stays $dirty after control.$setPristine()

martinschulzofficial:
When calling $setPristine on the nested form or control,
form becomes $pristine if all the nested controls are pristine

Closes #13715

### DIFF
--- a/src/ng/directive/form.js
+++ b/src/ng/directive/form.js
@@ -279,16 +279,9 @@ function FormController(element, attrs, $scope, $animate, $interpolate) {
 
   // Private API: update form pristine-ness
   form.$$updatePristine = function() {
-    var isPristine = true,
-        controlsLength = controls.length,
-        i;
-
-    for (i = 0; i < controlsLength; i++) {
-      if (!controls[i].$pristine) {
-        isPristine = false;
-        break;
-      }
-    }
+    var isPristine = controls.every(function(control) {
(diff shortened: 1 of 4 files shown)


petebacondarwin:
Thanks for this @linoleum-js but I think that a better algorithm would be needed. The suggestion of a counter would work well. Do you want to try to implement that instead?

linoleum-js:
@petebacondarwin I already did https://github.com/angular/angular.js/pull/13773. @gkalpak  ask me to split to solutions into different PLs

petebacondarwin:
Thanks.  It is good to cross reference these things.

PR #13801, opened 2016-01-19
chore($compile): remove an unused dependency

### LINKED ISSUE #13791: $compile: `$document` is unused
Currently, `$document` is injected into compile, however it is entirely unused; the global `document` is always referenced instead.  Should the `$document` variable be used in all of these locations, or should we consider removing the `$document` dependency?

I'm happy to make a PR for whichever way you feel this should go.

dcherman:
Fixes #13791


gkalpak:
Backported to `v1.4.x` as c429ad8.

PR #13802, opened 2016-01-19
feat(component): log a warning for unsupported options in CDO

gkalpak:


petebacondarwin:
This is going to add quite a bit to the size of `angular.min.js`. Perhaps we can find a way to make the message shorter - perhaps just provide a URL to some docs?

gkalpak:
Hm...good point about the file size. The problem with pointing to a specific URL is that we'd need to produce it based on the Angular version used (as the docs might change from version to version).

How about dropping "ALIASED" (moving them into "UNSUPPORTED") and have on message like:

```
'NOTICE: Unsupported options in component "' + componentName + '":\n  ' + unsupportedOpts.join(', ')
```

petebacondarwin:
Actually, I thought about this a bit more. Really this is the kind of thing for which we created [ngHint](https://github.com/angular/angular-hint). I think we should either add a plugin for that or just do nothing. People will soon realise that their directives are not working as planned. @Narretz new `component` guide has a nice table that makes it quite clear what options are in and what are out

gkalpak:
Also makes sense. Closing...

PR #13803, opened 2016-01-19
removed ignored 'placeholder' attribute

nexus-uw:
input[type="date"] does not support the placeholder attribute (ref: http://www.w3schools.com/tags/att_input_placeholder.asp).

Also I present this as a contender for the smallest possible PR to this repo.

note: if this PR is acceptable, I expand it to include the remaining ignored placeholders in this file.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -182,7 +182,7 @@ var inputType = {
        <form name="myForm" ng-controller="DateController as dateCtrl">
           <label for="exampleInput">Pick a date in 2013:</label>
           <input type="date" id="exampleInput" name="input" ng-model="example.value"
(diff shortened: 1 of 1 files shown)


Narretz:
Yes, this change is acceptable! Before we can merge however, we need for you the following things:
- [ ] sign the CLA
- [ ] change the commit message to `docs(input): remove ignored placeholder attribute`
- [ ] make sure it's all one commit

gkalpak:
I think we still want to keep the placeholder, because some browsers (including Firefox) do not support `type="date"` and fallback to `type="text"`. In such cases, the placeholder is very helpful imo.

Narretz:
@gkalpak Ah, yes you are right. I guess this can't be merged then :(

nexus-uw:
poop. 

well thanks for the quick review.

PR #13814, opened 2016-01-21
fix(ngAnimate): properly cancel-out previously running class-based animations

### LINKED ISSUE #10156: Flickering with ngAnimate and ngShow
Please take a look at this jsfiddle: http://jsfiddle.net/1op08u1o/4/ in IE 11. If you click on link 2, then quickly click on link 1 sometimes you can see a flicker where the orange block is bumped down by the green box before the orange block disappears.

This only seems to happen in IE11, and only if I include the ngAnimate module. Even without css transitions and just showing or hiding there is still a flicker. It may take a few tries to see the flicker happen. I'm using angular 1.3.3

matsko:
Prior to this fix the addition and removal of a CSS class via
ngAnimate would cause flicker effects because $animate was unable
to keep track of the CSS classes once they were applied to the
element. This fix ensures that ngAnimate always keeps a reference
to the classes in the currently running animation so that cancelling
works accordingly.

Closes #10156
Closes #13814

Please merge for 1.4 and 1.5.

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -42,22 +42,21 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     });
   }
 
-  function hasAnimationClasses(options, and) {
-    options = options || {};
-    var a = (options.addClass || '').length > 0;
-    var b = (options.removeClass || '').length > 0;
+  function hasAnimationClasses(animation, and) {
+    var a = (animation.addClass || '').length > 0;
+    var b = (animation.removeClass || '').length > 0;
     return and ? a && b : a || b;
   }
 
   rules.join.push(function(element, newAnimation, currentAnimation) {
     // if the new animation is class-based then we can just tack that on
-    return !newAnimation.structural && hasAnimationClasses(newAnimation.options);
+    return !newAnimation.structural && hasAnimationClasses(newAnimation);
   });
 
   rules.skip.push(function(element, newAnimation, currentAnimation) {
     // there is no need to animate anything if no classes are being added and
     // there is no structural animation that will be triggered
-    return !newAnimation.structural && !hasAnimationClasses(newAnimation.options);
+    return !newAnimation.structural && !hasAnimationClasses(newAnimation);
   });
 
   rules.skip.push(function(element, newAnimation, currentAnimation) {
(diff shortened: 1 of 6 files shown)


Narretz:
UNfortunately, I'm not seeing that the issue is fixed: http://plnkr.co/edit/HIaO54UfXl9z674KSzZV?p=preview

Steps to repdroduce:
1. You are initially on link 1 (green)
2. Click on link 2
3. Immediately click on link 1

What should we see?
The orange show animation should be briefly visible, then it should be hidden immediately (no animation / transition style on the hide animtion) and the green box should fade in.

What do we see?
The orange box doesn't hide immediately. It lingers around for a split second before disappearing.

The whole thing doesn't happen when you switch the link slowly.
So I wonder if it's something else instead.

matsko:
@Narretz try this demo: http://plnkr.co/edit/GwDlruHuZs4ZKFP7kqC9?p=preview. I think your animate.js might have not been the right build.

Narretz:
@matsko Oh, do I look stupid now! I've only built the unminified version, but tested it with the minified, which was older of course. Let's just pretend this didn't happen, okay? :wink:

matsko:
@Narretz I won't tell anyone :P

vitaly-t:
Great! Will it get into 1.3?

Narretz:
We don't backport to 1.3 anymore. And I think the bug was introduced in 1.4
anyway. Wha are you still on 1.3?
Am 22.01.2016 02:36 schrieb "Vitaly Tomilov" <notifications@github.com>:

Narretz:
@vitaly-t Okay, I see it has been reported first for 1.3, but we don't support 1.3 - only for major breaking stuff.

src/ngAnimate/animateQueue.js
```diff
@@ -374,7 +379,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
             close();
             return runner;
           } else {
-            mergeAnimationOptions(element, existingAnimation.options, options);
+            mergeAnimationDetails(element, existingAnimation.options, options);
```
gkalpak: How come we don't need to pass the animation object, but the options (considering all other invocation changed) ?

src/ngAnimate/animateQueue.js
```diff
@@ -362,11 +359,19 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
...
         runner: runner
       };
 
+      if (!hasExistingAnimation) {
+        // normalization in this case means that it removes redundant CSS classes that
+        // already exist (addClass) or do not exist (removeClass) on the element
+        normalizeAnimationDetails(element, newAnimation);
+      }
+
       if (hasExistingAnimation) {
```
gkalpak: This can be an `else` block to the `if` above.

Narretz: This is fixed in the updated PR: https://github.com/angular/angular.js/pull/13822

PR #13822, opened 2016-01-22
fix(ngAnimate): properly cancel-out previously running class-based an…

### LINKED ISSUE #10156: Flickering with ngAnimate and ngShow

Narretz:
…imations

Prior to this fix the addition and removal of a CSS class via
ngAnimate would cause flicker effects because $animate was unable
to keep track of the CSS classes once they were applied to the
element. This fix ensures that ngAnimate always keeps a reference
to the classes in the currently running animation so that cancelling
works accordingly.

Closes #10156
Closes #13814

This updated PR adds a new test that ensures the buggy scenario is correctly reproduced. We need to actively start the animation by flushing a requestAnimationFrame, otherwise the addClass / removeClass values are not correct.

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -42,22 +42,21 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     });
   }
 
-  function hasAnimationClasses(options, and) {
-    options = options || {};
-    var a = (options.addClass || '').length > 0;
-    var b = (options.removeClass || '').length > 0;
+  function hasAnimationClasses(animation, and) {
+    var a = (animation.addClass || '').length > 0;
+    var b = (animation.removeClass || '').length > 0;
     return and ? a && b : a || b;
   }
 
   rules.join.push(function(element, newAnimation, currentAnimation) {
     // if the new animation is class-based then we can just tack that on
-    return !newAnimation.structural && hasAnimationClasses(newAnimation.options);
+    return !newAnimation.structural && hasAnimationClasses(newAnimation);
   });
 
   rules.skip.push(function(element, newAnimation, currentAnimation) {
     // there is no need to animate anything if no classes are being added and
     // there is no structural animation that will be triggered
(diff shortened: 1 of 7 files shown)


src/ngAnimate/animateQueue.js
```diff
@@ -374,11 +373,10 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
             close();
             return runner;
           } else {
-            mergeAnimationOptions(element, existingAnimation.options, options);
+            mergeAnimationDetails(element, existingAnimation.options, options);
```
gkalpak: How come we don't need to pass the animation object, but the options (considering all other invocation changed) ?

Narretz: Gooood question

test/ngAnimate/integrationSpec.js
```diff
@@ -25,6 +25,32 @@ describe('ngAnimate integration tests', function() {
     ss.destroy();
   });
 
+ it('should cancel a running and started removeClass animation when a follow-up addClass animation adds the same class',
```
petebacondarwin: what does it mean to be "running and started"?

Narretz: Yes, another very good question. Internally, there's a running state, but that can mean two things: first, the animation process waits for the next digest to happen; then it flushes an requestAnimationFrame (for the reflow) to trigger the actual animation start. The previous test didn'T flush the rAF, and so the scenario wasn't created.

Narretz:
@gkalpak I've updated the merge code and added a test for this path. PTAL.

gkalpak:
LGTM

PR #13824, opened 2016-01-22
feat($resource): add support for timeout in cancellable actions

mrac:
Old behavior: actions can be either cancellable or have a numeric timeout.
When having both defined, cancellable was ignored.
With this commit: it's possible for actions to have both cancellable:true
and numeric timeout defined.

Example usage:

``` js
  get: {
    method: 'GET',
    cancellable: true,
    timeout: 10000
  }
});

var currentPost = Post.get({id: 1});
...
// the next request can cancel the previous one
currentPost.$cancelRequest();
currentPost = Post.get({id: 2});

// any of those requests will also timeout, if the response
// doesn't come within 10 seconds
```

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -593,7 +593,6 @@ angular.module('ngResource', ['ng']).
           } else {
             if(cancellable) {
               simulatedTimeout = action.timeout;
-              delete action.timeout;
             }
           }
 
@@ -665,11 +664,7 @@ angular.module('ngResource', ['ng']).
               httpConfig.timeout = timeoutDeferred.promise;
 
               if(simulatedTimeout) {
-                simulatedTimeoutPromise = $timeout(function() {
-                  if(timeoutDeferred) {
-                    timeoutDeferred.resolve();
-                  }
-                }, simulatedTimeout);
+                simulatedTimeoutPromise = $timeout(timeoutDeferred.resolve, simulatedTimeout);
               }
             }
 
@@ -723,6 +718,7 @@ angular.module('ngResource', ['ng']).
                 value.$cancelRequest = angular.noop;
                 timeoutDeferred = httpConfig.timeout = null;
                 $timeout.cancel(simulatedTimeoutPromise);
+                simulatedTimeoutPromise = null;
               }
             });
 
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -65,6 +65,7 @@ function shallowClearAndCopy(src, dst) {
  * @requires $http
  * @requires ng.$log
  * @requires $q
+ * @requires $timeout
  *
  * @description
  * A factory which creates a resource object that lets you interact with
@@ -160,7 +161,6 @@ function shallowClearAndCopy(src, dst) {
  *     will be cancelled (if not already completed) by calling `$cancelRequest()` on the call's
  *     return value. Calling `$cancelRequest()` for a non-cancellable or an already
  *     completed/cancelled request will have no effect.<br />
- *     **Note:** If a timeout is specified in millisecondes, `cancellable` is ignored.
  *   - **`withCredentials`** - `{boolean}` - whether to set the `withCredentials` flag on the
  *     XHR object. See
  *     [requests with credentials](https://developer.mozilla.org/en/http_access_control#section_5)
@@ -416,7 +416,7 @@ angular.module('ngResource', ['ng']).
       }
     };
 
-    this.$get = ['$http', '$log', '$q', function($http, $log, $q) {
+    this.$get = ['$http', '$log', '$q', '$timeout', function($http, $log, $q, $timeout) {
 
       var noop = angular.noop,
         forEach = angular.forEach,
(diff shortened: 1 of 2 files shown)


src/ngResource/resource.js
```diff
@@ -586,9 +590,11 @@ angular.module('ngResource', ['ng']).
                          'requests, you should use the `cancellable` option.');
               delete action.timeout;
             }
-            cancellable = angular.isDefined(action.cancellable) ? action.cancellable :
-                         (options && angular.isDefined(options.cancellable)) ? options.cancellable :
-                         provider.defaults.cancellable;
+          } else {
+            if(cancellable) {
+              simulatedTimeout = action.timeout;
+              delete action.timeout;
```
gkalpak: Why delete it here ? Couldn't we just let it be ?

src/ngResource/resource.js
```diff
@@ -707,6 +722,7 @@ angular.module('ngResource', ['ng']).
               if (!isInstanceCall && cancellable) {
                 value.$cancelRequest = angular.noop;
                 timeoutDeferred = httpConfig.timeout = null;
+                $timeout.cancel(simulatedTimeoutPromise);
```
gkalpak: Also set `simulatedTimeoutPromise` to null (just in case).

src/ngResource/resource.js
```diff
@@ -575,7 +575,11 @@ angular.module('ngResource', ['ng']).
 
         forEach(actions, function(action, name) {
           var hasBody = /^(POST|PUT|PATCH)$/i.test(action.method);
-          var cancellable = false;
+          var cancellable = angular.isDefined(action.cancellable) ? action.cancellable :
+            (options && angular.isDefined(options.cancellable)) ? options.cancellable :
+              provider.defaults.cancellable;
+
+          var simulatedTimeout;
```
gkalpak: `simulatedTimeout` is a little confusing imo (what is "simulated" about it).
<sub>(On the other hand, I'm notiriously bad with names :smiley:)</sub>

src/ngResource/resource.js
```diff
@@ -656,6 +663,14 @@ angular.module('ngResource', ['ng']).
             if (!isInstanceCall && cancellable) {
               timeoutDeferred = $q.defer();
               httpConfig.timeout = timeoutDeferred.promise;
+
+              if(simulatedTimeout) {
+                simulatedTimeoutPromise = $timeout(function() {
+                  if(timeoutDeferred) {
+                    timeoutDeferred.resolve();
+                  }
+                }, simulatedTimeout);
```
gkalpak: This could be just `simulatedTimeoutPromise = $timeout(timeoutDeferred.resolve, simulatedTimeout);`.

gkalpak:
A couple of minor comments, but LGTM otherwise.

src/ngResource/resource.js
mrac: I was trying to find a different name, not to confuse with timeout which is passed to $http. There are quite a few variables that deal with $http timeout.

So it for me it means "non-$http" timeout.

src/ngResource/resource.js
mrac: Yes we don't need to delete it. It gets overwritten anyway later:
```
httpConfig.timeout = timeoutDeferred.promise;
```

src/ngResource/resource.js
mrac: Yes, but I wanted to be sure, timeoutDeferred is not null already, so extra if() here. But may be it's obvious.

src/ngResource/resource.js
mrac: Ok.

src/ngResource/resource.js
mrac: You are right, timeoutDeferred.resolve can be evaluated instantly (and not inside the timeout handler).

src/ngResource/resource.js
gkalpak: I don't think it is overwritten. `action.timeout` is not the same as `httpConfig.timeout`.
But we shouldn't delete it if we don't have to :smiley:

gkalpak:
@mrac, thx for working on this :sparkles: :+1: :sparkles: 
I will take it from here (due to tight time constraints to catch the 1.5.0 train).

BTW (so you know for the next time), there were some legitimate failures in tests - nothing major just a couple of whitespace issues that JSCS (rightfully) complaints about :smiley:
(You can see them on Travis, e.g. [here][1] - or you can also run the tests locally with `grunt test` (to run all the tests) or `grunt ci-checks`/`grunt test:unit`/`grunt test:e2e` etc (to run a specific subset).)

[1]: https://travis-ci.org/angular/angular.js/jobs/104696540

mrac:
@gkalpak great, thanks! Good to hear it's going to be in 1.5.0.
Yeah, I ran unit tests and e2e tests as described here https://docs.angularjs.org/misc/contribute. Apparently ```grunt test``` does much more ;) Thanks for the info.

mrac:
@gkalpak It's nice to have it released already, thanks!

Still here remains a problem of how to distinguish between cancelled requests and timed-out requests. The latter might need different handling in most cases. Apparently both produce the same error response `{status: -1, statusText: "", ...}`

The solution would be to allow to pass a statusConfig to timeout-promise resolve value in $http. It could produce different error status and statusText. Currently timeout-promise resolve value is ignored entirely.

gkalpak:
I don't think it's high priority, but I would be happy to review any PRs if anyone decided to submit them :wink:

mrac:
@gkalpak Great! So take a look:
feat($http,$resource): statusText resolved in timeout promise #14021

PR #13826, opened 2016-01-22
fix(ngSanitize): Blacklist the attribyte `usemap`

lgalfaso:
Given that the attribute `name` is blacklisted, the attribute `usemap` should be
blacklisted too.

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -254,7 +254,7 @@ var validElements = angular.extend({},
                                    optionalEndTagElements);
 
 //Attributes that have href and hence need to be sanitized
-var uriAttrs = toMap("background,cite,href,longdesc,src,usemap,xlink:href");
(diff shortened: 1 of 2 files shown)


Narretz:
This has landed in master, I guess we should backport to 1.4?

petebacondarwin:
Difficulty is that it is a BC for 1.4

Narretz:
Ok, then let's just keep it as it is in 1.4

breadtk:
FWIW the change log calls this out simply as a breaking change. Perhaps it should be relabeled as a "Security Note" to denote its importance?

breadtk:
Thanks @gkalpak!

marekciupak:
Since this one wasn't applied to 1.2.x branch, should I consider 1.2.x version as vulnerable?

gkalpak:
@marekciupak, you are right, this should be backported to 1.2.x :+1: 
On it...

Issue #13827, opened 2016-01-22
nonassign $compile error message is vague


wesleycho:
Oftentimes the error message thrown by https://github.com/angular/angular.js/blob/d98c5f03a4cee88a3f0a383c217e3b2f84bbaa25/src/ng/compile.js#L3122-L3124 is vague, and is hard to determine exactly what is wrong.

For example, the error message I am seeing is `Error: [$compile:nonassign] http://errors.angularjs.org/1.4.9/$compile/nonassign?p0=undefined&p1=foo`, which does not tell me the attribute name.

Tweaking the text on this to output the attribute name could be a huge boon, and a minimal addition - I can open a PR for this if desired.

PR #13828, opened 2016-01-22
feat(compile): improve nonassign error message

wesleycho:
- Improve error message to mention attribute the expression errored on

This implements #13827.


gkalpak:
LGTM. Please, add `Fixes #...` at the end of the comit message, so the related issues will be automatically closed.

(I also think it's preferrable to have such change as `docs`, rather than `feat`.)

wesleycho:
Done

PR #13831, opened 2016-01-23
docs(tutorial/12 - Applying Animations): enter

salamanderrex:
Should be enter not click. Since we define the css trigger at move, not click.


salamanderrex:
I signed it!

gkalpak:
Backported to `v1.4.x` as da03497.

PR #13837, opened 2016-01-25
fix($location): do not decode `%2F` codes in the path back to `/` characters

### LINKED ISSUE #13245: Potential Bug with URL Hash being decoded (Tested on Angular 1.4.2 & 1.4.7)
After browsing the code and reading http://tools.ietf.org/html/rfc3986#section-2.4 I'm still not sure if this is intended behavior. This is happening for us in one specific instance where we have Angular being used by a component on our page. A 3rd party script sets window.location.hash, and Angular's $locationWatch picks up the url and the url is decoded... When the 3rd party js script grabs the window.location.hash value, its now decoded and the third party script expects it to be encoded still which creates a bug. The reason it occurs in this case is that the third party script is splitting on the "/" character, so 942520H PROCAT SE KOHLER CV680 W/52 SIDE DISCHARGE gets encoded to /s/TPI/942502G_PROCAT_26HP_KAW_W%2F%2F61_SIDE_DISCHARGE//1
Angular modifies it to /s/TPI/942502G_PROCAT_26HP_KAW_W//61_SIDE_DISCHARGE//1. Which is put into json to send over the wire... So the decoding in the hash breaks that process...

Here is an example setting the window.location.hash = "%27" notice that it gets decoded...

5a84f580-821e-11e5-8df0-555485426a96

From what I could glean from the spec this shouldn't happen...

"The characters slash ("/") and question mark ("?") are allowed to
represent data within the fragment identifier. Beware that some
older, erroneous implementations may not handle this data correctly
when it is used as the base URI for relative references (Section
5.1)."

I did find this in the source... looks like I'm hitting the or of the else, seems to happen across all browsers ;)

} else {
// - reloadLocation is needed as browsers don't allow to read out
// the new location.href if a reload happened.
// - the replacement is a workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=407172
return reloadLocation || location.href.replace(/%27/g,"'");
}

petebacondarwin:
If the user has purposefully encoded forward slashes into their $location path
(i.e. the bit after the hashbang), then we should not forceably decode it,
since if you want forward slashes you are allowed to place them in an
unencoded form straight into a hash segment of a URL

Closes #13245

### DIFF
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -79,6 +79,7 @@
     "parseKeyValue": false,
     "toKeyValue": false,
     "encodeUriSegment": false,
+    "decodeUriSegment": false,
     "encodeUriQuery": false,
     "angularInit": false,
     "bootstrap": false,
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -74,6 +74,7 @@
   parseKeyValue: true,
   toKeyValue: true,
   encodeUriSegment: true,
+  decodeUriSegment: true,
   encodeUriQuery: true,
   angularInit: true,
   bootstrap: true,
@@ -1338,9 +1339,21 @@ function toKeyValue(obj) {
  */
 function encodeUriSegment(val) {
   return encodeUriQuery(val, true).
+             replace(/%252F/gi, '%2F').
              replace(/%26/gi, '&').
              replace(/%3D/gi, '=').
              replace(/%2B/gi, '+');
+}
+
+
+/**
+ * We need our custom method because our segments may contain encoded forward slashes that should
+ * not be decoded.
+ */
+function decodeUriSegment(val) {
+  val = val.replace(/%2F/gi, '%252F');
+  val = decodeURIComponent(val);
+  return val;
 }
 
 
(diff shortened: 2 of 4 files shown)


petebacondarwin:
The main concern with this is that I had to change a previously passing test. But I think that this test was actually unnecessarily tight.

Narretz:
LGTM

gkalpak:
Since we are using `encodeUriSegment` on the `hash`, shouldn't we also use `decodeUriSegment` for the `hash` as well ?

petebacondarwin:
Oh man! This location stuff is too convoluted. I am now not clear when we want to use our custom encoding/decoding and when not. I think it works at the moment, mostly be chance!

christopherthielen:
The special slash handling in paths has been affecting ui-router also. See this recent analysis ticket: https://github.com/angular-ui/ui-router/issues/2598

I think $location tries a little too hard to be helpful with the slashes and encoding, and we lose the capability to manage it ourselves. 

It would be great if this were addressed, but I'd be concerned about introducing new behavior this late in the game.  Perhaps an opt-in location service that allowed a bit more "raw" access would mitigate that possibility.

abirmingham:
Any update on this? +1 @christopherthielen suggesting more configurability (or opt out of the $location service entirely).

petebacondarwin:
The current state of affairs is this PR: https://github.com/angular/angular.js/pull/16354. See the link in the PR to follow the reasoning...

PR #13848, opened 2016-01-26
fix($compile): properly denormalize templates when only one of the start/end symbols is different

gkalpak:
Previously, if either of the start/end interpolation symbols remained unchanged (i.e. `{{` or `}}`), then directive templates would not be denormalized properly. Changing only one of the start/end symbols (but not both) is an uncommon but legitimate usecase.

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -3137,6 +3137,54 @@ describe('$compile', function() {
     });
 
 
+    it('should support custom start interpolation symbol, even when `endSymbol` doesn\'t change',
+      function() {
+        module(function($compileProvider, $interpolateProvider) {
+          $interpolateProvider.startSymbol('[[');
+          $compileProvider.directive('myDirective', function() {
+            return {
+              template: '<span>{{ hello }}|{{ hello | uppercase }}</span>'
+            };
+          });
+        });
+
+        inject(function($compile, $rootScope) {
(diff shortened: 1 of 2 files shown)


test/ng/compileSpec.js
```diff
@@ -3137,6 +3137,54 @@ describe('$compile', function() {
     });
 
 
+    it('should support custom start interpolation symbol, even when `endSymbol` doesn\'t change',
```
Narretz: `endSymbol is the default`?

Narretz:
LGTM with a small style suggestion

test/ng/compileSpec.js
petebacondarwin: What @gkalpak means is that these test that the functionality works even if you change **only one** of `startSymbol` and `endSymbol`

petebacondarwin:
LGTM

test/ng/compileSpec.js
Narretz: I know, I just thought it was more explicit to state that the other symbol remains the default symbol. As is works for me though.

gkalpak:
Backported to `v1.4.x` as 2d44a68.

test/ng/compileSpec.js
gkalpak: I meant to change it as @Narretz suggested upon merge, but forgot to amend the commit before pushing :disappointed:

PR #13850, opened 2016-01-26
fix($parse): Preserve expensive checks when runnning $eval inside an expression

lgalfaso:
When running an expression with expensive checks, there is a call to `$eval` or `$evalAsync`
then that expression is also evaluated using expensive checks

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1757,9 +1757,18 @@ function $ParseProvider() {
           csp: noUnsafeEval,
           expensiveChecks: true
         };
+    var runningChecksEnabled = false;
 
-    return function $parse(exp, interceptorFn, expensiveChecks) {
+    $parse.$$runningExpensiveChecks = function() {
+      return runningChecksEnabled;
+    };
+
+    return $parse;
+
+    function $parse(exp, interceptorFn, expensiveChecks) {
       var parsedExpression, oneTime, cacheKey;
+
+      expensiveChecks = expensiveChecks || runningChecksEnabled;
 
       switch (typeof exp) {
         case 'string':
@@ -1786,6 +1795,9 @@ function $ParseProvider() {
             } else if (parsedExpression.inputs) {
               parsedExpression.$$watchDelegate = inputsWatchDelegate;
             }
+            if (expensiveChecks) {
+              parsedExpression = expensiveChecksInterceptor(parsedExpression);
+            }
             cache[cacheKey] = parsedExpression;
           }
           return addInterceptor(parsedExpression, interceptorFn);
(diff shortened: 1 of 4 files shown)


gkalpak:
Aren't `$watch` and `$watchCollection` also affected ?
LGTM otherwise.

gkalpak:
So, this works for `$$postDigest()`, `$apply()`, `$watch()` and `$watchCollection()`, but it doesn't seem to work for `$applyAsync()`.

I would still add tests for all of those (better safe than sorry :smiley:). E.g.:

```js
scope.foo = {w: $window};

expect(function() {
  scope.$eval($parse('$$postDigest($eval("foo.w && true"))', null, true));
}).toThrow();
expect(function() {
  scope.$eval($parse('$apply("foo.w && true")', null, true));
}).toThrow();

expect(function() {
  scope.$eval($parse('$applyAsync("foo.w && true")', null, true));
  scope.$digest();
}).toThrow();
expect(function() {
  scope.$eval($parse('$watch("foo.w && true")', null, true));
  scope.$digest();
}).toThrow();
expect(function() {
  scope.$eval($parse('$watchCollection("foo.w && true", $eval)', null, true));
  scope.$digest();
}).toThrow();
```

src/ng/parse.js
```diff
@@ -1788,15 +1797,28 @@ function $ParseProvider() {
...
         default:
-          return addInterceptor(noop, interceptorFn);
+          return expensiveChecksInterceptor(addInterceptor(noop, interceptorFn), expensiveChecks);
       }
-    };
+    }
+
+    function expensiveChecksInterceptor(fn, expensiveChecks) {
+      if (!expensiveChecks) return fn;
+      return function expensiveCheckFn(scope, locals, assign, inputs) {
```
jbedard: Won't this break `$$watchDelegate` and `inputs` on `fn`?

lgalfaso: @jbedard you are right, this needs more work

jbedard:
How about disallowing execution of nested `$parse`ed statements. Would that be an option?

lgalfaso:
> How about disallowing execution of nested $parseed statements. Would that be an option?

That would be too big of a breaking change

jbedard:
People actually do that? I've never even thought of it. Unless there is a less obvious way that I haven't thought of...

lgalfaso:
Updated the PR and fixed the comments. PTAL

src/ng/parse.js
```diff
@@ -1788,6 +1797,7 @@ function $ParseProvider() {
             }
             cache[cacheKey] = parsedExpression;
           }
+          parsedExpression = expensiveChecksInterceptor(parsedExpression, expensiveChecks);
```
gkalpak: Couldn't we move this inside the if-block ?

gkalpak: Also, this would hide properties like `assign`, `constant`, `inputs` and `literal`.

petebacondarwin: Also, wouldn't it make sense to cache the `parsedExpression` that has already had the expensive checks interceptor added, if the cache is `cacheExpensive`?

gkalpak: Actually, that's what I meant by "move this inside the if-block". (Inside and _before_ `cache[cacheKey] = parsedExpression`.)
Thx for making it clear, @petebacondarwin :smiley:

lgalfaso: > Also, this would hide properties like assign, constant, inputs and literal

Given that the expensive checks is a private API, I think this is not an issue. Anyhow, made the change.

> Couldn't we move this inside the if-block ?

Yep :)

petebacondarwin:
LGTM

test/ng/rootScopeSpec.js
```diff
@@ -1387,7 +1387,7 @@ describe('Scope', function() {
       expect(child.log).toBe('child context');
     }));
 
-    it('should operate only with a single queue across all child and isolate scopes', inject(function($rootScope) {
+    it('should operate only with a single queue across all child and isolate scopes', inject(function($rootScope, $parse) {
       var childScope = $rootScope.$new();
```
arashkbanan: Hi

gkalpak:
@petebacondarwin, should this be backported to `v1.4.x` ?

petebacondarwin:
@gkalpak - yes I would say so. Please do make it happen :-)

gkalpak:
Backported to `v1.4.x` as 96d62cc.

PR #13852, opened 2016-01-26
fix(ngTouch): remove ngClick override

### LINKED ISSUE #3296: ngMobile / directive / ngClick.js - Dramatic performance
I had to stop using mobile ngClick, because of its problems with performance.

In my case, the 300 elements that have nested 3-4 ngClick, while adding new elements, performance (especially in chrome) tend to zero - its not possible to work with the application then. What does not happen if you use a standard ngClick.

### LINKED ISSUE #3347: ngMobile's overridden ngClick directive fails to bust the click event
I've already reported this issue 21 days ago, here: #3132 and unfortunately it got tacked onto the back of another issue which resulted in it being closed. So here we go again :)

Here is an example of the problem: http://embed.plnkr.co/HQvrYQ/preview

The problem is that the tap event might have caused the DOM to change, so that the click fires in the same position but something else is there now which causes the new DOM element to incorrectly receive the event.

This issue is noticeable on iOS Safari, inside an iOS UIWebView, Android Default Browser (not Chrome) and inside an Android WebView. This does not seem to be an issue with Chrome on Android.

### LINKED ISSUE #3999: ng-touch, mouse event and 'ng-click-active'
I am building angularjs app that works on both touch and mouse environment. I am pleased to use ng-touch for touch support. I am using 'ng-click-active' for styling pressed state and it works well on touch device. But it does not works well on desktop and mouse environment. 'ng-click-active' is work differently from a  :active selector.

- 'ng-click-active' should not be disappeared if mouse down and hold on the element. (chrome and ie)
- 'ng-click-active' should not be disappeared if mouse down and move on the element.

below code makes the problem. 

    element.on('mousemove mouseup', function(event) {
       element.removeClass(ACTIVE_CLASS_NAME);
     });

chrome trigger mousemove event even if mouse does not move.

### LINKED ISSUE #4030: ngTouch: ng-click directive is broking default behaviour of child elements

### LINKED ISSUE #4428: clientX/clientY missing from event in ngTouch on iOS7
When using jquery1.9 , ngTouch under iOS7 the event for touchStart and touchEnd have undefined clientX/clientY  . The actual values are found in event.originalEvent.changedTouches[0].clientX and event.originalEvent.changedTouches[0].clientY .

The effect is the "fast" click is never called or protected as the dist calculation is always NaN .

### LINKED ISSUE #5307: Touch event on a link issue

### LINKED ISSUE #6001: angular-touch breaks click events
Hi,

**Overview of the issue**:
Without jQuery, all click listeners got an empty ```event``` object if the link is annotated with ```ng-click```.

```
<a data-ng-click="reportClick()" data-test-listen-to-click>Link with directive</a>
```

Leads to:

```
testListenToClick#onClick; event: 
Object {preventDefault: function, stopPropagation: function}
```

This for example breaks angular-ui-router: https://github.com/angular-ui/ui-router/issues/827

**Angular Version(s)**
1.2.10

Chrome 32 with touch emulation, Chrome Mobile 32

**Reproduce the error**
http://jsfiddle.net/scheffield/55Skr/

**Related issues**
I might be wrong, but this issue looks like it has the same reason: #5307

### LINKED ISSUE #6251: ngClick using ngTouch fires twice
ngTouch 1.2.1 on Safari iOS 7 fires a second time by tapping wherever on the screen:

http://jsfiddle.net/coma/2hWWa/embedded/result/

http://stackoverflow.com/questions/21708730/why-are-multiple-click-events-fired-when-using-ngtouch/21766071#21766071

### LINKED ISSUE #6432: ngTouch: iOs form elements loose focus when tapped to enter text (textareas,inputs)

### LINKED ISSUE #7935: ngTouch - Keep track of touch by identifier
I am currently developing a multitouch application and tried to use `ngTouch` here. I stumbled upon an implementation which can handle only one touch at a time. I am wondering if this is done by purpose as i dont know about the side effects of my idea.

Right now ngTouch alters the `ng-click` directive, so it will work with touch events, but the implementation works with only one touch at a time.

Any OS, any Browser
Angular and ngTouch Version 1.2.17

I created a plunkr to demonstate, please use a touch device with multiple touches:
http://plnkr.co/edit/s2vkc4neFzha6nsBSGQG

If you hold one button, the other one will not work anymore, altough the `touchstart` and `touchend` events are triggered. I think this is due to 3 reasons:

1. The `touchstart` event handler on the element takes the first touch in the list of `touches` instead of the `targetTouches`. If there is another touch in progress it will take it as the `touchstart` coordinates which is wrong  (at least semantically)  since this is not the touch which occured.

2.  The `preventGhostClick` function can only handle one event too.

3. The `tapping` property gets reset if **any** touch move event happens

In my case i am trying to use a 2D-Joystick which gets dragged around all the time. Additionally i want to trigger some other stuff like a horn. I looked for a solution to get this behavior, but had no success, the only thing left would be to attach the touch handlers manually. 

I would be glad to provide a pull request if wished.

Note: Currently there is a problem if jQuery included, the starting coordinates are NaN for the touchstart. See https://github.com/angular/angular.js/pull/4583

### LINKED ISSUE #9744: ng-click working wrong when -webkit-overflow-scrolling: touch
Dears,
I have a page full of links ng-clicks, that we can scroll down : 

overflow-y: scroll;
-webkit-overflow-scrolling: touch

I've noticed something strange (both iPhone 4 and Android Samsung Galasy S4) : 

when you touch the page to scroll (touchstart on a first ng-click directive)
and the scroll has enough speed
and you touch again to stop the scroll (touchstart on a second ng-click directive)

then it opens the first ng-click default action on the touchend.

We would expect nothing or the second ng-click to take action. Because the first one has been forgotten for a while.

Could you please give any help on this issue ?

Thanks

### LINKED ISSUE #9872: ngTouch ngClick doesn't bubble
When using ngClick on mobile, ngTouch
>prevents the following click event from propagating.

And the custom click event it triggers doesn't propagate.

This creates inconsistent behavior in that on a computer, the click event continues to bubble/propagate, but on mobile, it stops. This means that any directives higher in the dom that listen for element clicks fail.

See [my SO question](http://stackoverflow.com/questions/26685498/ngtouch-ngclick-doesnt-bubble) for more details.

### LINKED ISSUE #10366: ng-touch - Child element triggers ng-click for parent
Example of problem below:

http://codepen.io/bzamora/pen/OPyogw

I am using ng-touch in my AngularJS app to provide touch events. I am using the ng-touch built in ng-click-active class to animate a tile when clicked. All is good so far, no issues.

There is also a checkbox inside the tile that is only displayed when the user is in "edit" mode. At this point, I don't want the clicking of the checkbox to trigger the parent's element click. I've tried canceling the click when in edit mode ng-click="inEditMode === false && doStuff() and also using $event.stopPropagation() in the checkbox to try to stop the click event from propagating up to the tile. Nothing has worked so far...

Thanks for any help!

Also asked in SO: http://stackoverflow.com/questions/27272989/ng-touch-child-element-triggers-ng-click-for-parent

### LINKED ISSUE #10918: 'ng-click-active' remains stuck
This is hard to track and thus hard to fix as it doesn't happen every time, but can be definitely reproduced on desktop Chrome and mobile browsers (Chrome, Safari, Android browser).

- When element with ng-click defined and '.ng-click-active' style defined in CSS is clicked on, it gets class 'ng-click-active' as long as element is in pressed/touched state
- Once click/touch has finished, 'ng-click-active' class should be removed from the element.

However, this is sometimes not the case and element remains in 'clicked' state with class still on long after click had finished.
Unfortunately it's not easy to reproduce this as it doesn't happen every time, but it's pretty annoying.

### LINKED ISSUE #11197: ngTouch in iPad did not improve tap handling
I know mobile safari handles ng-mouseover, ng-click and ng-mouseleave separately, but I supposed that ngTouch would fix this.
Here's a plunker with my testings. In mobile safari, ng-mouseover triggers first and then ng-click.
http://plnkr.co/edit/DcA73t1VSdrcQtGDCCRB?p=preview

### LINKED ISSUE #11261: ngTouch check for blur() function before using it.

### LINKED ISSUE #11358: ngTouch doesn't work

### LINKED ISSUE #12082: Click not fired with ng-touch 1.3.16

### LINKED ISSUE #12150: Calling element.click() causes "$apply already in progress" with ng-touch
With angular-touch installed and working fine, when calling `element.click()` in the process of an $apply (where `element` is a plain vanilla DOM element, gotten from something like `$event.target`) causes an error `$apply already in progress`.

Presumably this is because ng-touch registers a click handler on each and every element, which calls $apply again, which is no good for Angular.

I see two solutions possible:
1. Fix angular-touch so that it doesn't do an $apply (for example, using $timeout is common practise)
2. Fix Angular to allow calling $apply while another $apply is already in progress. This one makes the most sense to me. If an $apply is already happening, calling it again don't need to do anything at all.

All this is on Angular 1.4.1, angular-touch 1.4.1, no jQuery, and Firefox 38.

### LINKED ISSUE #12153: ngTouch prevents focusing inputs at or below ngClick
You can see [my comment on this commit here](https://github.com/angular/angular.js/commit/0bbd20f255b2954b5c41617fe718cf6eca36a972#commitcomment-11730613) to explain the problem with that commit further. But, I truly don't understand how in an attempt to bust the redundant, original/"ghost click" click event (because it's being done faster than the original thanks to this wonderful work) we would also want to prevent the very likely *intended* behavior of also focusing a clicked focusable element. After all, it isn't as if this focusing behavior is also redundant; ngTouch's ngClick doesn't also focus the element if focusable when it triggers the click event. But before I contribute either a removal of this blurring "feature", or the addition of a speedier focusing *feature*, do you all agree that something must be done?

Because FYI, (and I'm sure I'm not the only one) my organization is relying my own copy of Angular *just* to remove this one line of code. (...which happened to break the login for my company's site, as well as all other inputs in all other dialogs, and even gave pause to my CTO that we should even be utilizing Angular to the extent we are considering this bug was because of migrating to 1.4.0 - though technically a minor release, it's hardly a *minor* release)

### LINKED ISSUE #12392: Native click event of inner element busted by outer ngClick element
When there is a nested structure, e.g.

    <div ng-click="action()">
        <button my-click>mybutton</my-button>
        <p>lipsum....</p>
    </div>

where `my-click` is a directive that have native click binding:

    link: function(scope, el, attrs){
        el.on('click', function(ev){ 
           //...some other action
           // ev.stopPropagation() won't work as ngTouch use event capture
        });
    }

The native click event won't call when click on the button. The click event is busted. The outer div has raised `touchstart` and `touchend`  and thus busted all internal click event from nested element. You cannot stop the event propagation from button to div as it is capturing mode.

demo at http://plnkr.co/edit/NQ09S0EGqoiPs6QTIhLY?p=preview

### LINKED ISSUE #12455: Indeterminate ngModel behavior when updateOn is blur
The example for the issue described here can be checked on [CodePen](http://codepen.io/robmazan/pen/MwzjEa?editors=101).

Let's assume you have the following setup:

  - An input field with `ngModel` on it, having `ng-model-options="{updateOn: 'blur'}"`
  - A button with `ngClick` on it, where the handler function using the value of the model from scope

Executing the following steps has different results on mouse and on touch devices:

  - Edit the input value
  - Click on the button directly (without clicking anywhere else first)

With mouse based input the `ngClick` handler can access the current input value on scope, and it is the same value that you can see in the input field.

On touch devices however, when the `ngClick` handler is executed the value is not updated yet (see the CodePen example with Chrome DevTools device emulation to see it happening).

### LINKED ISSUE #12734: Incorrect events order with ngTouch
```html
<element>
    <input/>
</element>
```

```js
document.addEventListener('click', click, true);
document.addEventListener('touchend', touchend, true);
element.addEventListener('mousedown', mousedown, true);
element.addEventListener('touchstart', touchstart, true);
element.addEventListener('blur', blur, true);
inputElement.addEventListener('focus', focus);
```

mouse click
```
["mousedown","click","focus"] 
```
touch without ngTouch
```
["touchstart","touchend","mousedown","click","focus"] 
```
touch with ngTouch. Incorrect order!
```
["touchstart","touchend","mousedown","focus","click","blur"]
```

### LINKED ISSUE #13287: link selection doesn't work on IOS safari
let me explain, when a long click on a link i expect this to happen :
screenshot_11_10_15__11_17_am

I'm doing `<a  ng-href="{{url}}">{{url}}</a>` on Firefox+Chrome android and it work perfectly. But on iOS safari the link is not recognize.

Is it an issue, or am I doing it wrong ?

### LINKED ISSUE #13558: Drop-down lists in modals and dropdowns don't work correctly on iPad if ngTouch is a dependency
If you use a native drop-down list (<select>) inside a dropdown or a modal from angular-bootstrap and if your application has a dependency on ngTouch (1.4.3), the drop-down lists don't work as expected. When you click a drop-down list, the list with options blinks most of the time, making it very hard to choose an option. If you remove the ngTouch dependency, the drop-down lists work fine.

I created a plunkr which demonstrates this issue: http://plnkr.co/edit/fsU1zKpwalrxEEelsBka.
I'm using safari on an iPad air 5 (iOS 9.2).

Could you please provide a solution for this issue?
Thank you!

Narretz:
This commit removes the ngClick directive from the ngTouch module.
The directive was conceived to remove the 300ms delay
for click events on mobile browsers, by sending a synthetic click event on touchstart.
It also tried to make sure that the original click event that the browser sends after 300ms
was "busted", so that no redundant "ghost-clicks" appear.

There are various reasons why this feature has been removed:
- "This is an ugly, terrible hack!" (says so in the source)
- It is plagued by various bugs that are hard to fix / test for all platforms (see below)
- Simply including ngTouch activates the ngClick override, which means even if you simply want
  to use ngSwipe, you may break parts of your app
- There exist alternatives for removing the 300ms delay, that can be used very well with Angular:
  [FastClick](https://github.com/ftlabs/fastclick), [Tappy!](https://github.com/filamentgroup/tappy/)
  (There's also hammer.js for touch events / gestures)
- The 300ms delay itself is on the way out - Chrome and Firefox for Android remove the 300ms delay
  when the usual `<meta name="viewport" content="width=device-width">` is set. In IE, the
  `touch-action` css property can be set to `none` or `manipulation` to remove the delay. Finally,
  since iOs 8, Safari doesn't delay "slow" taps anymore. There are some caveats though, which can be
  found in this excellent article on which this summary is based: http://developer.telerik.com/featured/300-ms-click-delay-ios-8/

Note that this change does not affect the `ngSwipe` directive.

Issues with interactive elements (input, a etc.) when parent element has ngClick:
Closes #13287
Closes #13558
Closes #12392
Closes #12153
Closes #11358
Closes #6432
Closes #12867
Closes #7231
Closes #13213
Closes #12082
Closes #4030
Closes #12545
Closes #6001
Closes #5307

Other issues:
- incorrect event order
- incorrect event propagation
- ghost-clicks / failing clickbusting with corner cases
- browser specific bugs
- et al.

Closes #12734
Closes #12455
Closes #11197
Closes #11261
Closes #12150
Closes #10918
Closes #10366
Closes #9872
Closes #9744
Closes #6251
Closes #4428
Closes #3999
Closes #3347
Closes #3296
Closes #7935
Closes #13272
Closes #13447
Closes #13122
Closes #12317
Closes #11577
Closes #11342
Closes #10211
Closes #9724
Closes #7134
Closes #6330
Closes #3447

BREAKING CHANGE:

The `ngClick` override from the `ngTouch` module has been removed. This means that on
touch-based devices, users might experience a 300ms delay before an ngClick event is fired.

If you rely on this behavior, consider using [FastClick](https://github.com/ftlabs/fastclick) or
[Tappy!](https://github.com/filamentgroup/tappy/). Also note that modern browsers remove the 300ms
delay under some circumstances:
- Chrome and Firefox for Android remove the 300ms delay when the well-known `<meta name="viewport" content="width=device-width">` is set
- Internet Explorer removes the delay when  `touch-action` css property is set to `none` or `manipulation`
- Since iOs 8, Safari removes the delay on so-called "slow taps"

See this article for more info on the topic: http://developer.telerik.com/featured/300-ms-click-delay-ios-8/


Narretz:
- [ ] add BC to migration guide after merge

Do we need any other doc changes?

wesleycho:
This also fixes some bugs that occur with third party angular libraries due to the non-emission of the click event when using ngTouch (since the click event never really happens, but touchstart/touchend do, so the click "event" doesn't properly bubble up).

Narretz:
We are going with deprecation of ngTouch.ngClick and disabling it by default instead: https://github.com/angular/angular.js/pull/13857

PR #13861, opened 2016-01-27
fix($http): properly synchronize $browser outstandingRequestCount

### LINKED ISSUE #13782: $http doesn't immediately increment $brower's outstandingRequestCount. Protractor issues result.

Thanks.

wbyoko:
The issue is that using $http doesn't update $browser.outstandingRequestCount synchronously so that $browser.notifyWhenNoOutstandingRequests waits accordingly.
Configuring this synchronization with the promise chain updates the outstandingRequestCount correctly.

Closes #13782

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -375,8 +375,8 @@ function $HttpProvider() {
    **/
   var interceptorFactories = this.interceptors = [];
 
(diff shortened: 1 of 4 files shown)


gkalpak:
Closing, since this is essentially the same as #13862 (just against a different branch).
Once fixed in `master`, we can cherry-pick the commit from `master` into `v1.4.x`.

Thx @wbyoko !

wbyoko:
Sounds Good.

PR #13862, opened 2016-01-27
fix($http): include interceptors as a part of $browser's outstandingRequest

### LINKED ISSUE #13782: $http doesn't immediately increment $brower's outstandingRequestCount. Protractor issues result.

wbyoko:
The issue is that using $http doesn't update $browser.outstandingRequestCount synchronously so that $browser.notifyWhenNoOutstandingRequests waits accordingly.
Configuring this synchronization with the promise chain updates the outstandingRequestCount correctly.

Fixes #13782

### DIFF
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1866,6 +1866,24 @@ describe('$http', function() {
         expect(paramSerializer({foo: 'foo', bar: ['bar', 'baz']})).toEqual('bar=bar&bar=baz&foo=foo');
       });
     });
+
+    describe('$browser outstandingRequests', function() {
+      var $browser;
+
+      beforeEach(inject(['$browser', function(browser) {
+        $browser = browser;
+      }]));
+
+      it('should update $brower outstandingRequestCount', function() {
+        $httpBackend.when('GET').respond(200);
+
+        expect($browser.$$outstandingRequestCount).toBe(0);
+        $http({method: 'GET', url: '/some'});
+        expect($browser.$$outstandingRequestCount).toBe(1);
+        $httpBackend.flush();
+        expect($browser.$$outstandingRequestCount).toBe(0);
+      });
+    });
   });
 
 
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1867,21 +1867,22 @@ describe('$http', function() {
       });
     });
 
-    describe('$browser outstandingRequests', function() {
-      var $browser;
+    describe('$browser\'s outstandingRequestCount', function() {
+      var incOutstandingRequestCountSpy, completeOutstandingRequestSpy;
 
-      beforeEach(inject(['$browser', function(browser) {
-        $browser = browser;
-      }]));
+      beforeEach(inject(function($browser) {
+        incOutstandingRequestCountSpy = spyOn($browser, '$$incOutstandingRequestCount').andCallThrough();
+        completeOutstandingRequestSpy = spyOn($browser, '$$completeOutstandingRequest').andCallThrough();
+      }));
 
-      it('should update $brower outstandingRequestCount', function() {
+      it('should update $browser outstandingRequestCount', function() {
         $httpBackend.when('GET').respond(200);
 
-        expect($browser.$$outstandingRequestCount).toBe(0);
+        expect(incOutstandingRequestCountSpy).not.toHaveBeenCalled();
         $http({method: 'GET', url: '/some'});
-        expect($browser.$$outstandingRequestCount).toBe(1);
+        expect(incOutstandingRequestCountSpy).toHaveBeenCalledOnce();
         $httpBackend.flush();
-        expect($browser.$$outstandingRequestCount).toBe(0);
+        expect(completeOutstandingRequestSpy).toHaveBeenCalledOnce();
       });
     });
   });
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1871,15 +1871,27 @@ describe('$http', function() {
       var incOutstandingRequestCountSpy, completeOutstandingRequestSpy;
 
       beforeEach(inject(function($browser) {
-        incOutstandingRequestCountSpy = spyOn($browser, '$$incOutstandingRequestCount').andCallThrough();
-        completeOutstandingRequestSpy = spyOn($browser, '$$completeOutstandingRequest').andCallThrough();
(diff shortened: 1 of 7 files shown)


test/ng/httpSpec.js
```diff
@@ -1866,6 +1866,24 @@ describe('$http', function() {
...
     });
+
+    describe('$browser outstandingRequests', function() {
+      var $browser;
+
+      beforeEach(inject(['$browser', function(browser) {
+        $browser = browser;
+      }]));
+
+      it('should update $brower outstandingRequestCount', function() {
```
gkalpak: _$brower_  -->  _$browser_

test/ng/httpSpec.js
```diff
@@ -1866,6 +1866,24 @@ describe('$http', function() {
         expect(paramSerializer({foo: 'foo', bar: ['bar', 'baz']})).toEqual('bar=bar&bar=baz&foo=foo');
       });
     });
+
+    describe('$browser outstandingRequests', function() {
+      var $browser;
+
+      beforeEach(inject(['$browser', function(browser) {
```
gkalpak: We usually don't use inline array notation in tests. You can do:

```js
beforeEach(inject(function(_$browser_) {
  $browser = _$browser_;
}));
```

wbyoko:
@gkalpak @Narretz Removed the changes to ngMock's $browser in favor of using spy. Also added a test for failure as well. Would you think more tests are needed?

gkalpak:
@wbyoko, I was trying to reproduce the problem in an e2e test, but couldn't. Do you happen to have a minimal reproduction (even a flaky one) ?

wbyoko:
This fiddle I included in #13782 shows the issue. http://jsfiddle.net/awtwff3x/

You could also use an interceptor with a setTimeout which resolves a promise. 

We do have request interceptors in our project which may be why we would see this issue.

do you add e2e tests in `test/e2e/tests` ?

gkalpak:
@wbyoko, thx (it seems like I missed that).

Regarding your question: One way to add e2e tests is to add a `xyzSpec.js` file in `test/e2e/tests/` and a corresponding "tiny app" in `test/e2e/fixtures/xyz/` (where you normally put an `index.html` and a `script.js` file). There is a `sample` you can take a look at.

gkalpak:
@wbyoko, this test access `$browser.notifyWhenNoOutstandingRequests()` directly (i.e. without protractor).
Do you have any protractor tests that fail due to this issue ? I still can't reproduce it with protractor :disappointed:

wbyoko:
@gkalpak I added an e2e test. Uses the setTimeout method in an interceptor and doesn't pass without included updates to http.

I can clean this up and squash / recreate this commit. Would you think this e2e test is needed?

wbyoko:
That test would shouldn't pass on the master branch right now.

gkalpak:
I'd like to have an actual e2e test that showcases the issue. Let me take a look and get back to you.

wbyoko:
@gkalpak Thanks. I ran this test on the master branch to make sure that it didn't pass.

And this showcases the fix by including interceptors as a part of the outstanding request which normally protractor wasn't waiting for. 

I wouldn't understand how the e2e test is accessing `$browser.notifyWhenNoOutstandingRequests()` directly.

Thanks again.

gkalpak:
Protractor uses `$$testability.whenStable()` to detect stability before performing any action (e.g. finding elements on a page etc). `$$testability.whenStable()` uses `$browser.notifyWhenNoOutstandingRequests()`.

wbyoko:
But with the e2e test it's going through protractor to do it. 

So the fiddle calls it directly, the e2e test doesn't.

It seems actual issue that $browser's outstandingRequest should be including interceptors because otherwise using them has the potentially to make e2e tests inconsistent. 

Thanks.

wbyoko:
@gkalpak any updates on what you found?

Thanks.

gkalpak:
@wbyoko, I didn't get a chance to properly look into it yet. I'll get back to this right after `v1.5.0` is out (probably today), so we'll get it into `v1.5.1`.

wbyoko:
@gkalpak sounds good. let me know when you have. 

Thanks.

gkalpak:
AFAICT, the only way to reproduce it in an e2e test (or in a real app), is by using a non-Angular service inside of an interceptor (e.g. using `window.setTimeout` instead of `$timeout`). This is expected though, because native services to not rely on `$browser.defer`, so there is no way for `$browser` to know that something is pending (another good reason to use built-in services instead of native ones).

TBH, I am kind of leaning towards considering this a "works as expected" (unless of course we can come up with an actual reproduction using built-in services).
Note that `$browser` is a private service, so although it is possible to reproduce the issue by calling `$browser.notifyWhenNoOutstandingRequests()` directly, this is not a supported usecase.

I'd like to know what others think about it.
/cc @lgalfaso, who has expressed interest in this :smiley:

wbyoko:
@gkalpak If i can clarify what the issue is now after our discussion and investigation, it is that using interceptors creates a real possibility of e2e tests becoming flaky.

I can't identify the exact threshold but adding the setTimeout with a delay is what can make tests fail every time. Whereas if the outstanding request incorporated the entire http promise chain (including interceptors), the setTimeout e2e tests pass every time

Protractor says that it will wait for $http requests to resolve, which is not happening consistently if interceptors are included.

Wouldn't the expected behavior be that protractor remain consistent when interceptors are utilized?

wbyoko:
I don't see any non-angular services in the interceptors used in the project where this issue was found. 

Updating the $browser outstanding request calls in the same way outlined in this pr, had tests with $http requests pass without the need of `browser.sleep`.

gkalpak:
@wbyoko, can you share an example of a flaky test that doesn't use non-Angular services ?

wbyoko:
Well the project does use some third party modules, like ui-router. I wouldn't know if ui-router could be the issue but the tests are straight forward.

wbyoko:
I'm looking on the protractor site and see that some people are having similar issues with $http requests.

https://github.com/angular/protractor/issues/789

/cc @juliemr, who may know if more have run into this issue.

wbyoko:
This issue seems to havean e2e test that's failing due to $http.

https://github.com/angular/protractor/issues/1641

This issue has a app as a part of the ticket

https://github.com/0xR/protractor-ngroute-bug

The repo uses $browser directly, but is a complete application.

PR #13867, opened 2016-01-28
Update YTL to TRY

abides:
ISO 4217 code   TRY (Numeric 949) ref:https://en.wikipedia.org/wiki/Turkish_lira


gkalpak:

Thx for the PR though :)

PR #13868, opened 2016-01-28
Turkish shortdate format change

abides:
DD.MM.YYYY, or "DD <name of the month> YYYY" ref: https://en.wikipedia.org/wiki/Date_and_time_notation_in_Turkey

It shoulde be dot seperator not space in shortdate format


gkalpak:

PR #13869, opened 2016-01-28
fix($http): immediatelly increment $browser's `outstandingRequestCount`

### LINKED ISSUE #13782: $http doesn't immediately increment $brower's outstandingRequestCount. Protractor issues result.

gkalpak:
This allows protractor to more reliably detect when all outstanding requests have been completed.

Fixes #13782
Closes #13862

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -375,8 +375,8 @@ function $HttpProvider() {
    **/
   var interceptorFactories = this.interceptors = [];
 
-  this.$get = ['$httpBackend', '$$cookieReader', '$cacheFactory', '$rootScope', '$q', '$injector',
-      function($httpBackend, $$cookieReader, $cacheFactory, $rootScope, $q, $injector) {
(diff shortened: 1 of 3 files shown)


gkalpak:
This is an alternative approach to #13862.

It attempts to decrement the counter a little closer to where it was done before (i.e. before the response interceptors) - not that it should matter afaict - and has a few more tests.

arashkbanan:
@arashkbanan

arashkbanan:
@fahimeh

Narretz:
I like it.
Why didn't it work before though? What's the problem with $httpBackend? I think that should go into the commit message.

gkalpak:
The problem (as mentioned in https://github.com/angular/angular.js/issues/13782#issuecomment-172376703) was that `$httpBackend` was not called until after the promise chain (including the request interceptors) had been processed.
At the very least this requested a $digest, but also note that some request interceptors could return promises (that where also resolved async).

I'll update the commit message.

PR #13871, opened 2016-01-28
fix($parse): Copy `inputs` for expressions with expensive checks

lgalfaso:


gkalpak:
LGTM

gkalpak:
Backported to `v1.4.x` as 0b7fff3.

PR #13877, opened 2016-01-28
Improve animation guide docs

Narretz:


docs/content/guide/animations.ngdoc
```diff
@@ -274,6 +274,98 @@ myModule.directive('my-directive', ['$animate', function($animate) {
...
 ```
 
+## Animations on app bootstrap / page load
+
+By default, animations are disabled when the Angular app {@link guide/bootstrap bootstraps}. If you are using the {@link ngApp} directive,
+this happens in the `DOMContentLoaded` event, so immediately after the page has been loaded.
+Animations are disabled, so that UI and content are instantly visible. Otherwise, with many animations on
+the page, the loading process may become too visually overwhelming, and the performance may suffer.
+
+Internally, this is done by waiting for two digests - that is usually enough to wait until all templates
```
gkalpak: I think we wait until all templates have been downloaded and **then** we wait for one digest more (plus the one that is already in progress).

docs/content/guide/animations.ngdoc
```diff
@@ -274,6 +274,98 @@ myModule.directive('my-directive', ['$animate', function($animate) {
...
+/&#42; prefixed with animate- &#42;/
+.animate-fade-add.animate-fade-add-active {
+  transition:1s linear all;
+  opacity:0;
+}
+```
+
+The classNameFilter approach generally applies the biggest speed boost, because the matching is
+done before any other animation disabling strategies are checked. However, once a class has been
+excluded via the classNameFilter, it's not possible to enable animations for this element at runtime.
```
gkalpak: Nitpicking: "has been excluded via the classNameFilter" sounds like the elements matching the RegExp will **not** be animated (while it's the othr way around: only those matching the RegExp will be animated).

docs/content/guide/animations.ngdoc
```diff
@@ -274,6 +274,98 @@ myModule.directive('my-directive', ['$animate', function($animate) {
...
+
+### At runtime: {@link ng.$animate#enabled $animate.enabled()}
+
+This function can be used to enable / disable animations in two different ways:
+
+With a single `boolean` argument, it enables / disables animations globally: `$animate.enabled(false)`
+disables all animations in your app.
+
+When the second argument is an native DOM or jQuery element, the function enables / disables
```
gkalpak: _an native_  -->  _a native_

docs/content/guide/animations.ngdoc
```diff
@@ -274,6 +274,98 @@ myModule.directive('my-directive', ['$animate', function($animate) {
...
+animations on this element *and all its children*: `$animate.enabled(false, myElement)`. This is the
+most flexible way to change the animation state. For example, even if you have used it to disable
+animations on a parent element, you can still re-enable it for a child element. And compared to the
+`classNameFilter`, you can change the animation status at runtime instead of only in the config.
```
gkalpak: _in the config_  -->  _during config_ (?)

docs/content/guide/animations.ngdoc
```diff
@@ -305,6 +397,47 @@ In that case, you can add styles to the CSS that make sure the element stays hid
...
+
+.spinner.ng-animate {
+    transition: 0s none;
+    animation: 0s none;
+}
+
+If you do have CSS transitions / animations defined for the animation events, make sure they have higher priority
+than any styles that are independent from ngAnimate.
+
+You can also use one of the two other {@link guide/animations#how-to-selectively-enable-disable-and-skip-animations strategies to disable animations}
```
gkalpak: Missing `.` at the end.

docs/content/guide/animations.ngdoc
```diff
@@ -305,6 +397,47 @@ In that case, you can add styles to the CSS that make sure the element stays hid
...
+    animation: rotating 2s linear infinite;
+}
+```
+
+Now, when the `ngIf` changes, `ngAnimate` will see the spinner animation and use
+it to animate the `enter`/`leave` event, which doesn't work because
+the animation is infinite. The element will still be added / removed after a timeout, but there will be a
+noticable delay.
+
+This might also happen  because some third-party frameworks place animation duration defaults
```
gkalpak: Double space between `happen` and `because` :grin:

src/ng/directive/ngRepeat.js
```diff
@@ -164,11 +164,11 @@
...
  *
  * @animations
- * **.enter** - when a new item is added to the list or when an item is revealed after a filter
- *
- * **.leave** - when an item is removed from the list or when an item is filtered out
- *
- * **.move** - when an adjacent item is filtered out causing a reorder or when the item contents are reordered
+ * | Animation                        | Occurs                              |
+ * |----------------------------------|-------------------------------------|
+ * | {@link ng.$animate#enter .enter} | when a new item is added to the list or when an item is revealed after a filter |
```
gkalpak: There is a `.` before the names (e.g. `.enter` vs `enter`). In previous tables there was no `.`

src/ng/directive/ngShowHide.js
```diff
@@ -89,8 +89,10 @@ var NG_HIDE_IN_PROGRESS_CLASS = 'ng-hide-animate';
  * property to block during animation states--ngAnimate will handle the style toggling automatically for you.
  *
  * @animations
- * addClass: `.ng-hide` - happens after the `ngShow` expression evaluates to a truthy value and the just before contents are set to visible
- * removeClass: `.ng-hide` - happens after the `ngShow` expression evaluates to a non truthy value and just before the contents are set to hidden
+ * | Animation                        | Occurs                              |
+ * |----------------------------------|-------------------------------------|
+ * | {@link $animate#addClass addClass} `.ng-hide`  | after the `ngShow` expression evaluates to a truthy value and the just before contents are set to visible |
```
gkalpak: It's the other way around: `removeClass` when truthy, `addClass` when falsy.
(There is also a similat section in `ngHide` (also wrong), that could use a table :smiley:)

src/ng/directive/ngSwitch.js
```diff
@@ -27,8 +27,10 @@
  * </div>
 
  * @animations
- * enter - happens after the ngSwitch contents change and the matched child element is placed inside the container
- * leave - happens just after the ngSwitch contents change and just before the former contents are removed from the DOM
+ * | Animation                        | Occurs                              |
+ * |----------------------------------|-------------------------------------|
+ * | {@link ng.$animate#enter enter} enter | after the ngSwitch contents change and the matched child element is placed inside the container |
```
gkalpak: There's an extra `enter` after `}`.

src/ngRoute/directive/ngView.js
```diff
@@ -19,8 +19,10 @@ ngRouteModule.directive('ngView', ngViewFillContentFactory);
  * Requires the {@link ngRoute `ngRoute`} module to be installed.
  *
  * @animations
- * enter - animation is used to bring new content into the browser.
- * leave - animation is used to animate existing content away.
+ * | Animation                        | Occurs                              |
+ * |----------------------------------|-------------------------------------|
+ * | {@link ng.$animate#enter enter}  | when the expression changes, on the new view |
+ * | {@link ng.$animate#leave leave}  | when the expression changes, on the old view |
```
gkalpak: I would change the description to something like: _when the new/old view is inserted to/removed from the DOM_

src/ngAnimate/ngAnimateSwap.js
```diff
@@ -10,11 +10,17 @@
...
+ * common usecase for this directive is a rotating banner or slider component which
  * contains one image being present at a time. When the active image changes
  * then the old image will perform a `leave` animation and the new element
  * will be inserted via an `enter` animation.
  *
+ * @animations
+ * | Animation                        | Occurs                               |
+ * |----------------------------------|--------------------------------------|
+ * | {@link ng.$animate#enter enter}  | when the expression changes, on the new element |
+ * | {@link ng.$animate#leave leave}  | when the expression changes, on the old element |
```
gkalpak: Here too, I would change the description to something like: _when the new/old element is inserted to/removed from the DOM_

Narretz:
thanks @gkalpak - I've updated and addressed your changes.

src/ng/directive/ngShowHide.js
```diff
@@ -85,12 +85,14 @@ var NG_HIDE_IN_PROGRESS_CLASS = 'ng-hide-animate';
...
+ * Keep in mind that, as of AngularJS version 1.3, there is no need to change the display
  * property to block during animation states--ngAnimate will handle the style toggling automatically for you.
  *
  * @animations
- * addClass: `.ng-hide` - happens after the `ngShow` expression evaluates to a truthy value and the just before contents are set to visible
- * removeClass: `.ng-hide` - happens after the `ngShow` expression evaluates to a non truthy value and just before the contents are set to hidden
+ * | Animation                        | Occurs                              |
+ * |----------------------------------|-------------------------------------|
+ * | {@link $animate#addClass addClass} `.ng-hide`  | after the `ngShow` expression evaluates to a non truthy value and just before the contents are set to hidden |
+ * | {@link $animate#removeClass removeClass}  `.ng-hide`  | after the `ngShow` expression evaluates to a truthy value and the just before contents are set to visible |
```
gkalpak: _the just before contents_  -->  _just before the contents_

src/ng/directive/ngShowHide.js
```diff
@@ -249,12 +251,15 @@ var ngShowDirective = ['$animate', function($animate) {
...
+ * Keep in mind that, as of AngularJS version 1.3, there is no need to change the display
  * property to block during animation states--ngAnimate will handle the style toggling automatically for you.
  *
  * @animations
- * removeClass: `.ng-hide` - happens after the `ngHide` expression evaluates to a truthy value and just before the contents are set to hidden
- * addClass: `.ng-hide` - happens after the `ngHide` expression evaluates to a non truthy value and just before the contents are set to visible
+ * | Animation                        | Occurs                              |
+ * |----------------------------------|-------------------------------------|
+ * | {@link $animate#addClass addClass} `.ng-hide`  | after the `ngShow` expression evaluates to a non truthy value and just before the contents are set to hidden |
+ * | {@link $animate#removeClass removeClass}  `.ng-hide`  | after the `ngShow` expression evaluates to a truthy value and the just before contents are set to visible |
```
gkalpak: This table refers to `ngShow`. Should be converted to `ngHide`.

PR #13878, opened 2016-01-28
fix(select): handle corner case of adding options via a custom directive

### LINKED ISSUE #13874: chromeHack method throws TypeError: optionElement[0].hasAttribute is not a function
Hi,

I have a custom directive that replaces with template `<option ng-repeat>` and this ends up throwing an exception in chromeHack on line 28664:

```
TypeError: optionElement[0].hasAttribute is not a function
    at chromeHack (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:28664:24)
    at self.addOption (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:28745:5)
    at self.registerOption (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:28792:12)
    at http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:29133:22
    at invokeLinkFn (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:9039:9)
    at nodeLinkFn (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:8533:11)
    at compositeLinkFn (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:7929:13)
    at nodeLinkFn (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:8528:24)
    at compositeLinkFn (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:7929:13)
    at nodeLinkFn (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.js:8528:24) <!-- ngRepeat: item in data -->
```

**Reproducable**: always
**Angular Version(s)**: 1.4.8, 1.4.9
**Browsers and Operating System**:  Chrome Version 47.0.2526.111 m, Windows 10 64-bit
**Reproduce the Error**: http://plnkr.co/edit/UejSXqwfq320wtzN7HWG

gkalpak:
Under specific circumstances (e.g. adding options via a directive with `replace: true` and a structural directive in its template), an error occurred when trying to call `hasAttribute()` on a comment node (which doesn't support that method).
This commit fixes it by first checking if `hasAttribute()` is available.

Fixes #13874

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -81,7 +81,7 @@ var SelectController =
   // Tell the select control that an option, with the given value, has been added
   self.addOption = function(value, element) {
     // Skip comment nodes, as they only pollute the `optionsMap`
-    if (element.prop('nodeType') === NODE_TYPE_COMMENT) return;
+    if (element[0].nodeType === NODE_TYPE_COMMENT) return;
 
     assertNotHasOwnProperty(value, '"option value"');
     if (value === '') {
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -6,9 +6,8 @@ function chromeHack(optionElement) {
   // Workaround for https://code.google.com/p/chromium/issues/detail?id=381459
   // Adding an <option selected="selected"> element to a <select required="required"> should
   // automatically select the new element
-  var optionNode = optionElement[0];
-  if (optionNode.hasAttribute && optionNode.hasAttribute('selected')) {
-    optionNode.selected = true;
+  if (optionElement[0].hasAttribute('selected')) {
+    optionElement[0].selected = true;
   }
 }
 
@@ -81,6 +80,9 @@ var SelectController =
 
   // Tell the select control that an option, with the given value, has been added
   self.addOption = function(value, element) {
+    // Skip comment nodes, as they only pollute the `optionsMap`
+    if (element.prop('nodeType') === NODE_TYPE_COMMENT) return;
+
     assertNotHasOwnProperty(value, '"option value"');
     if (value === '') {
       self.emptyOption = element;
@@ -453,7 +455,6 @@ var optionDirective = ['$interpolate', function($interpolate) {
     restrict: 'E',
     priority: 100,
     compile: function(element, attr) {
-
       if (isDefined(attr.value)) {
         // If the value attribute is defined, check if it contains an interpolation
         var interpolateValueFn = $interpolate(attr.value, true);
@@ -467,7 +468,6 @@ var optionDirective = ['$interpolate', function($interpolate) {
       }
 
       return function(scope, element, attr) {
-
         // This is an optimization over using ^^ since we don't want to have to search
         // all the way to the root of the DOM for every single option element
         var selectCtrlName = '$selectController',
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -6,8 +6,9 @@ function chromeHack(optionElement) {
   // Workaround for https://code.google.com/p/chromium/issues/detail?id=381459
   // Adding an <option selected="selected"> element to a <select required="required"> should
   // automatically select the new element
-  if (optionElement[0].hasAttribute('selected')) {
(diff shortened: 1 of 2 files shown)


test/ng/directive/selectSpec.js
```diff
@@ -44,6 +44,17 @@ describe('select', function() {
         }
       };
     });
+
+    $compileProvider.directive('myOptions', function() {
+      return {
+        scope: {myOptions: '='},
+        replace: true,
+        template:
+            '<option value="{{ ::option.value }}" ng-repeat="option in ::myOptions">' +
```
Narretz: Is there a reasonw why this uses one-time binding?

test/ng/directive/selectSpec.js
```diff
@@ -611,6 +622,24 @@ describe('select', function() {
...
+    it('should not break when adding options via a directive with `replace: true` '
+        + 'and a structural directive in its template',
+      function() {
+        scope.options = [
+          {value: '1', label: 'Option 1'},
+          {value: '2', label: 'Option 2'},
+          {value: '3', label: 'Option 3'}
+        ];
+
+        expect(function() {
```
Narretz: I would prefer that we check the validity of the select itself: `expect(element).toEqualSelect([unknownValue()], '1', '2', '3');`

src/ng/directive/select.js
```diff
@@ -6,8 +6,9 @@ function chromeHack(optionElement) {
   // Workaround for https://code.google.com/p/chromium/issues/detail?id=381459
   // Adding an <option selected="selected"> element to a <select required="required"> should
   // automatically select the new element
-  if (optionElement[0].hasAttribute('selected')) {
-    optionElement[0].selected = true;
+  var optionNode = optionElement[0];
+  if (optionNode.hasAttribute && optionNode.hasAttribute('selected')) {
```
Narretz: Can we instead filter comment nodes in the addOption function? They pollute the optionsMap. Doesn't really do anything, but it's cleaner.

test/ng/directive/selectSpec.js
gkalpak: No particular reason - force of habbit :)

test/ng/directive/selectSpec.js
gkalpak: Will change it.

src/ng/directive/select.js
gkalpak: Done

gkalpak:
@Narretz, I made the suggested changes. PTAL

src/ng/directive/select.js
```diff
@@ -80,6 +80,9 @@ var SelectController =
 
   // Tell the select control that an option, with the given value, has been added
   self.addOption = function(value, element) {
+    // Skip comment nodes, as they only pollute the `optionsMap`
+    if (element.prop('nodeType') === NODE_TYPE_COMMENT) return;
```
Narretz: element[0].nodeType?

Narretz:
small suggestion, otherwise LGTM

src/ng/directive/select.js
gkalpak: That was my first "reaction", but then I thought: "What if `element[0]` is undefined ?"
So, instead of `element[0] && element[0].nodeType`, I went for `element.prop(...)` (I'm not even sure if the element passed to the linking function can be empty...)

gkalpak: I guess it can't. Fixed

Narretz: AddOption is a private Api so we guarantee that element will always be jq
wrapped. I'm a bit worried about the overhead of calling prop on every
option.
Am 29.01.2016 13:28 schrieb "Georgios Kalpakas" <notifications@github.com>:

gkalpak:
Backported to `v1.4.x` as df6e731.

Thx for the review, @Narretz.

PR #13886, opened 2016-01-29
fix(input): re-validate when partially editing date-family inputs

### LINKED ISSUE #12207: Changing only part of date (input) does not change form's pristine state (Chrome)

jbedard:
This checks for changes to the badInput/typeMismatch validity flags on keydown/wheel/mousedown events to try to mark date-like inputs as $dirty when they have been partially entered but are still incomplete/invalid.

This has a couple advantages/disadvantages compared to #12902 explained at https://github.com/angular/angular.js/pull/12902#issuecomment-174226116.

Fixes #12207

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -32,6 +32,12 @@ var WEEK_REGEXP = /^(\d{4})-W(\d\d)$/;
 var MONTH_REGEXP = /^(\d{4})-(\d\d)$/;
 var TIME_REGEXP = /^(\d\d):(\d\d)(?::(\d\d)(\.\d{1,3})?)?$/;
 
+var PARTIAL_VALIDATION_EVENTS = 'keydown wheel mousedown';
+var PARTIAL_VALIDATION_TYPES = createMap();
+forEach('date,datetime-local,month,time,week'.split(','), function(type) {
+  PARTIAL_VALIDATION_TYPES[type] = true;
+});
+
 var inputType = {
 
   /**
@@ -1118,6 +1124,8 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     });
   }
 
+  var timeout;
+
   var listener = function(ev) {
     if (timeout) {
       $browser.defer.cancel(timeout);
@@ -1147,8 +1155,6 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
   if ($sniffer.hasEvent('input')) {
     element.on('input', listener);
   } else {
-    var timeout;
-
     var deferListener = function(ev, input, origValue) {
       if (!timeout) {
         timeout = $browser.defer(function() {
@@ -1179,6 +1185,26 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
   // if user paste into input using mouse on older browser
   // or form autocomplete on newer browser, we need "change" event to catch it
   element.on('change', listener);
+
+  // Some native input types (date-family) have the ability to change validity without
+  // firing any input/change events.
+  // For these event types, when native validators are present and the browser supports the type,
+  // check for validity changes on various DOM events.
+  if (PARTIAL_VALIDATION_TYPES[type] && ctrl.$$hasNativeValidators && type === attr.type) {
+    element.on(PARTIAL_VALIDATION_EVENTS, function(ev) {
+      if (!timeout) {
+        var validity = this[VALIDITY_STATE_PROPERTY];
+        var origBadInput = validity.badInput;
+        var origTypeMismatch = validity.typeMismatch;
+        timeout = $browser.defer(function() {
+          timeout = null;
+          if (validity.badInput !== origBadInput || validity.typeMismatch !== origTypeMismatch) {
+            listener(ev);
+          }
+        });
+      }
+    });
+  }
 
   ctrl.$render = function() {
     // Workaround for Firefox validation #12102.
--- a/test/ng/directive/inputSpec.js
+++ b/test/ng/directive/inputSpec.js
@@ -1943,6 +1943,83 @@ describe('input', function() {
     });
   });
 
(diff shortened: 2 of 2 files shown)


src/ng/directive/input.js
```diff
@@ -1180,6 +1186,26 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+  // firing any input/change events.
+  // For these event types, when native validators are present and the browser supports the type,
+  // check for validity changes on various DOM events.
+  if (PARTIAL_VALIDATION_TYPES[type] && ctrl.$$hasNativeValidators && type === attr.type) {
+    element.on(PARTIAL_VALIDATION_EVENTS, function(ev) {
+      if (!timeout) {
+        var validity = this[VALIDITY_STATE_PROPERTY] || {};
+        var origBadInput = validity.badInput;
+        var origTypeMismatch = validity.typeMismatch;
+        timeout = $browser.defer(function() {
```
Narretz: What's the timeout for?

Narretz:
Definitely simpler.
But I don't particularly like that we move validation logic into the inout. Makes it more compley than it already is.
The $dirty problem could also be solved by calling $setDirty in the new event listener, though.

src/ng/directive/input.js
jbedard: When a `keydown` event fires the input has not yet changed. This might not be necessary for the other `PARTIAL_VALIDATION_EVENTS` though. The other option is using `keyup` but then changes won't be detected if the key is held for a while. Basically the same logic used for the no-input-event case (IE).

jbedard:
That is a good point... but I think this is still more change detection then validation logic, it just happens to detect those changes using the `validity` state.

jbedard:
The tests are currently failing on all the non-chrome browsers. Anyone have an opinion for how to fix that? Just put a giant `if (chrome)` around them all? :|

test/ng/directive/inputSpec.js
```diff
@@ -1918,6 +1918,83 @@ describe('input', function() {
     });
   });
 
+  ['month', 'week', 'time', 'date', 'datetime-local'].forEach(function(inputType) {
+    if (jqLite('<input type="' + inputType + '">').attr('type') !== inputType) {
```
gkalpak: This fails to detect the supporting browsers. You should use `.prop()` instead.
The `type` attribute will always have the requested type, but the `type` property will be set to `test` if the browser doesn't support the type.

gkalpak:
@jbedard, they are expected to fail on browsers that don't support the input type, but the code fails to properly detect support. See https://github.com/angular/angular.js/pull/13886/files#r51847302.

gkalpak:
BTW, at the moment only Chrome, Opera and Edge support date inputs.

src/ng/directive/input.js
```diff
@@ -1180,6 +1186,26 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
   // or form autocomplete on newer browser, we need "change" event to catch it
   element.on('change', listener);
 
+  // Some native input types (date-family) have the ability to change validity without
+  // firing any input/change events.
+  // For these event types, when native validators are present and the browser supports the type,
+  // check for validity changes on various DOM events.
+  if (PARTIAL_VALIDATION_TYPES[type] && ctrl.$$hasNativeValidators && type === attr.type) {
+    element.on(PARTIAL_VALIDATION_EVENTS, function(ev) {
+      if (!timeout) {
```
gkalpak: The fact that we are reusing the same timeout with the `else` brach above, is not good I think.
Since the callbacks are conditionally calling `listener`, there might be a timeout registered, so we don't register another one, but that other timeout's callback may end up not calling `listener` (while our callback would).

jbedard: I've had that concern as well for a while now with `keydown` vs `cut`/`paste` events in the IE case (no `input` event).

But in this case, with current browser date-like input support, I don't think there will be any issues. I don't think there is any browser that will *not* have the `input` event but *will* have native date-like inputs.

I do think this should change at some point though, even if just for the IE case...

gkalpak: True. Having separate timeouts would still be "cleaner", but even the shared timeout isn't indeed expected to create any issues.

src/ng/directive/input.js
```diff
@@ -1180,6 +1186,26 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+        var validity = this[VALIDITY_STATE_PROPERTY] || {};
```
gkalpak: Since we have checked that `$$hasNativeValidators` is true, is the `|| {}` part necessary ?

gkalpak:
A couple of minor comments (that could be ignored as well) :smiley:
LGTM otherwise :+1:

src/ng/directive/input.js
jbedard: I copied this from `badInputChecker`. I assume it's not needed there either. Should I remove both?

gkalpak: In `badInputChecker`, we check `element[0].validity` and then call `element.prop(VALIDITY_STATE_PROPERTY)` asynchronousy. I am not sure if it is possible for the node inside `element` to be replaced (e.g. by another directive). If not, then I guess we could change that too.

It's slightly different here, since the listener is attached to the underlying DOM node, so it's guarranteed not to change.

jbedard: Good points. I've dropped the `|| {}`...

gkalpak:
Backported to `v1.5.x` (e383804) and `v1.4.x` (02929f8).

PR #13887, opened 2016-01-29
fix(dateFilter, input): fix Date parsing in IE/Edge when timezone offset contains `:`

### LINKED ISSUE #13880: Angular date filter in IE
Hi,

I have an issue with date filter and timezone in IE 11.

I created a plunker. 
http://plnkr.co/FzVbYl

The filter is not formatting the date-time based on the timezone.

Example, if you change your system time and run the plunker, the time is the local time of browser. Which is incorrect.

This works btw in chrome, safari and firefox.

gkalpak:
When `Date.parse`-ing a date string, IE and Edge don't recognize the timezone offset in the format `+HH:mm` (but only without the `:`). According to [the spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15), the timezone offset should contain `:`. The [ISO 8601 Standard](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC) allows both forms (with and without `:`).
Although the `Date` implementation in JavaScript does not 100% follow the ISO 8601 Standard (it's just _based on it_), all other browsers seem to recognize both forms as well.

Fixes #13880

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1215,7 +1215,10 @@ function fromJson(json) {
 }
 
 
+var ALL_COLONS = /:/g;
 function timezoneToOffset(timezone, fallback) {
+  // IE/Edge do not "understand" colon (`:`) in timezone
+  timezone = timezone.replace(ALL_COLONS, '');
   var requestedTimezoneOffset = Date.parse('Jan 01, 1970 00:00:00 ' + timezone) / 60000;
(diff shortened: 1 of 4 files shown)


Narretz:
LGTM

gkalpak:
Backported to `v1.4.x` as 571afd6.

PR #13888, opened 2016-01-29
chore(travis): update Chrome to 47, FF to 43, and add Safari 9

Narretz:


scripts/travis/before_build.sh
```diff
@@ -11,8 +11,11 @@ fi
 
 npm install -g grunt-cli
 
-if [ $JOB != "ci-checks" ]; then
-  grunt package
+if [ $JOB != "ci-checks"]; then
+  if [ $JOB != "unit" ]; then
+    # unit tests test against unminified code
```
gkalpak: What is the reason for this change ?

Narretz: I'm just trying some optimizations. Unit tests don't need package, and that shaves off a good minute of the unit tests. Not terribly important since e2e tests take longer anyway, but nice to have if you just want to know the unit tests pass.

gkalpak: Ideally, we should test against both unminified and minified files :)

PR #13890, opened 2016-01-29
doc Fix typo component.ngdoc

DrDanRyan:
The wrong field name was being passed into the `$ctrl.update` call in `heroDetail.html` resulting in the wrong behavior (`name` was being updated instead of `location`)


DrDanRyan:
I signed it!

PR #13896, opened 2016-01-30
docs(error/inprog): describe your change...

smithl:
small typo: a -> an (before $apply)


gkalpak:
It depends on how you read `$apply`. For some it is "dollar apply", for others (including me) it's just "apply". I'd merge this, if only the CLA was signed... :wink:

rerdos:
@gkalpak @smithl I think the best solution would be `a(n) $apply`.

smithl:
@gkalpak @rerdos - Fair enough, I was concerned that it wasn't consistent with the two other instances on the same page, where it says "an apply." I am now noticing one more example on the same page where its "within a $apply/$digest" - this one should probably be changed as well.

gkalpak:
I fixed the other occurrence while merging. Thx !
Backported to `v1.4.x` as afb298b.

PR #13901, opened 2016-01-30
docs($http): reword of XSRF attack overview

vucalur:
Previous version emphasised "gaining user's private data".
While this perfectly describes JSON vulnerability (which is based on XSRF),
data theft suits XSS more.
Pure XSRF is more about performing requests that have side effects.


gkalpak:
I reworded it a bit and merged. Thx !
Backported to `v1.4.x` as 8dc4c75.

PR #13903, opened 2016-01-30
docs(angular.isBoolean): describe your change...

cynthia-wi:
As an Angular developer, I would like to see angular.isBoolean in the documentation since it is one of the five primitive data types.


Narretz:
We've decided that we are not gonna expose this function via documentation. While it's a very small change, we don't want to give the impression that we are going to expose more angular helper functions.

Issue #13735, opened 2016-01-11
fix(input[date]): support years with more than 4 digits


gkalpak:
Currently, (mainly because of the REGEXPs used for parsing dates, e.g. [here][L13] and [there][L29-L32]), date-family inputs do not support years with more than 4 digits (they consider the value invalid and the parser returns `undefined`).

According to the [W3 spec][1], the value of a date-family input should be a "valid _full-date_ as defined in [[RFC 3339]][2], with the additional qualification that **the year component is four or more digits** representing a number greater than 0" <sub>(emphasis mine)</sub>.

Additionally, [ISO 8601][3] allows (upon agreement of the involved parties) to have _expanded representation_ of years (with more than 4 digits); see for example sections 2.3.8 and 4.1.2.4.

--
Current browser support:

* **Chrome**, **Opera**: Support date inputs and years greater than 9999.

* **Firefox**, **IE**: Do not support date inputs, but the user can still enter the dates in ISO format and `ngModel` parses them correctly

* **Edge**: Supports date inputs, but **not** years greater than 9999.

* **Safari**: ???

There's nothing we can do for MS Edge on the Angular side, but the users of other browsers will still benefit from adding support for years greater than 9999.

[1]: http://www.w3.org/TR/html-markup/input.date.html
[2]: http://www.w3.org/TR/html-markup/references.html#refsRFC3339
[3]: http://dotat.at/tmp/ISO_8601-2004_E.pdf

[L13]: https://github.com/angular/angular.js/blob/8ccc0547a8838454300e2dc75bab7fafdf22d21f/src/ng/directive/input.js#L13
[L29-L32]: https://github.com/angular/angular.js/blob/8ccc0547a8838454300e2dc75bab7fafdf22d21f/src/ng/directive/input.js#L29-L33

Narretz:
Safari 9 has no special date input handling. ios9 has a datepicker input, and you can't actually enter the date by hand (at least not in Saucelabs).

PR #13905, opened 2016-01-31
fix(input[date]): support years with more than 4 digits

LeeAdcock:
Change to regular expressions used for matching ISO date strings. Increases the number of digits allowed in a year from exactly four to instead be four or more. Also matches on the entire string, resolving ambiguity around extra characters at the start of the year, or after the timezone.

Closes https://github.com/angular/angular.js/issues/13735

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -10,7 +10,7 @@
 */
 
 // Regex code was initially obtained from SO prior to modification: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
-var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d(:?[0-5]\d)?|Z)$/;
+var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+(?:[+-][0-2]\d(?::?[0-5]\d)?|Z)$/;
 // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
 // Note: We are being more lenient, because browsers are too.
 //   1. Scheme
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -9,8 +9,8 @@
   ngModelMinErr: false,
 */
 
-// Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
-var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)$/;
+// Regex code was initially obtained from SO prior to modification: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
+var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d(:?[0-5]\d)?|Z)$/;
 // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
 // Note: We are being more lenient, because browsers are too.
 //   1. Scheme
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -10,7 +10,7 @@
 */
 
 // Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
-var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)$/
+var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)$/;
 // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
 // Note: We are being more lenient, because browsers are too.
 //   1. Scheme
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -10,7 +10,7 @@
 */
 
 // Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
-var ISO_DATE_REGEXP = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/;
+var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)$/
 // See valid URLs in RFC3987 (http://tools.ietf.org/html/rfc3987)
 // Note: We are being more lenient, because browsers are too.
 //   1. Scheme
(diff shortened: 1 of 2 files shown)


test/ng/directive/inputSpec.js
```diff
@@ -657,6 +657,18 @@ describe('input', function() {
...
     });
 
+    it('should allow four or more digits in year', function() {
+      var inputElm = helper.compileInput('<input type="month" ng-model="value"  ng-model-options="{timezone: \'UTC\'}"/>');
+
+      helper.changeInputValueTo('10123-03');
+      expect(+$rootScope.value).toBe(Date.UTC(10123, 2, 1, 0, 0, 0));
+
+      $rootScope.$apply(function() {
+        $rootScope.value = new Date(Date.UTC(10123, 2, 1, 0, 0, 0));
```
gkalpak: You should use a different value.

test/ng/directive/inputSpec.js
```diff
@@ -856,6 +868,17 @@ describe('input', function() {
...
     });
 
+    it('should allow four or more digits in year', function() {
+      var inputElm = helper.compileInput('<input type="week" ng-model="value"  ng-model-options="{timezone: \'UTC\'}"/>');
+
+      helper.changeInputValueTo('10123-W03');
+      expect(+$rootScope.value).toBe(Date.UTC(10123, 0, 21));
+
+      $rootScope.$apply(function() {
+        $rootScope.value = new Date(Date.UTC(10321, 0, 21));
```
gkalpak: You should use a different value.

test/ng/directive/inputSpec.js
```diff
@@ -1734,6 +1757,19 @@ describe('input', function() {
...
     );
 
+    it('should allow four or more digits in year', function() {
+      var inputElm = helper.compileInput('<input type="date" ng-model="value" ng-model-options="{timezone: \'UTC\'}" />');
+
+        helper.changeInputValueTo('10123-01-01');
+        expect(+$rootScope.value).toBe(Date.UTC(10123, 0, 1, 0, 0, 0));
+
+        $rootScope.$apply(function() {
+          $rootScope.value = new Date(Date.UTC(10123, 0, 1, 0, 0, 0));
```
gkalpak: You should use a different value.

gkalpak:
There's no test for `datetime-local` :smiley:

test/ng/directive/inputSpec.js
```diff
@@ -1941,6 +1977,111 @@ describe('input', function() {
...
+        ['2010-01-T00:00:00.0000+01:01', false],    // day too few digits
+        ['2010-01-1T00:00:00.0000+01:01', false],   // day too few digits
+        ['2010-01-41T00:00:00.0000+01:01', false],  // day value too large
+        ['2010-01-200T00:00:00.0000+01:01', false], // day value too many digits
+
+        // Validate time
+        ['2010-01-01', false],                      // time must be specified
+        ['2010-01-0101:00:00.0000+01:01', false],   // missing date time seperator
+        ['2010-01-01V01:00:00.0000+01:01', false],  // invalid date time seperator
+        ['2010-01-01T01-00-00.0000+01:01', false],  // time must use period seperator
```
gkalpak: Actually, it must use colon separator.

test/ng/directive/inputSpec.js
```diff
@@ -1941,6 +1977,111 @@ describe('input', function() {
...
+        // Validate timezone
+        ['20123-06-15T00:00:00.0000', false],       // timezone must be specified
+
+        // Validate timezone offset
+        ['2010-06-15T00:00:00.0000+01:01', true],
+        ['2010-06-15T00:00:00.0000-01:01', true],
+        ['2010-06-15T00:00:00.0000~01:01', false],  // timezone must have valid postive/negative indicator
+        ['2010-06-15T00:00:00.000001:01', false],   // timezone must have valid postive/negative indicator
+        ['2010-06-15T00:00:00.0000+00:01Z', false], // timezone invalid trailing characters
+        ['2010-06-15T00:00:00.0000+00:01 ', false], // timezone invalid trailing characters
```
gkalpak: Actually, according to the standard, the timezone offset can have one of three forms:

`+/-HH:mm`, `+/-HHmm`, `+/-HH`

(Feel free to add suppory for those as well :smiley:)

LeeAdcock:
Thank you for the feedback, I appreciate it.

src/ng/directive/input.js
```diff
@@ -9,8 +9,8 @@
   ngModelMinErr: false,
 */
 
-// Regex code is obtained from SO: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
-var ISO_DATE_REGEXP = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/;
+// Regex code was initially obtained from SO prior to modification: https://stackoverflow.com/questions/3143070/javascript-regex-iso-datetime#answer-3143231
+var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d(:?[0-5]\d)?|Z)$/;
```
gkalpak: Since we are not using the groups anywhere, it would be better for performance to make them non-capturing.
(As an added bonus, you'll have the chance to write a RegExp that includes `?::?` - not somethng you do everyday :grin:)

gkalpak: BTW, I know I proposed this change, but there is currently an issue with IE/Edge (they can't parse a timezone offset when there is no colon), so additional steps might be needed. That said, I'm not sure when/if `ISO_DATE_REGEXP` is used after all, so let me look into it some more :)

LeeAdcock: Likely to be a problem if the input model is populated directly with an ISO formatted text string, instead of from the browser native picker.

LeeAdcock: I took a look at Edge in Saucelabs and found the same result when using the Date constructor for parsing both '2010-06-15T00:00:00.0000+**01:30**' as well as '2010-06-15T00:00:00.0000+**0130**'.

Is there a bug report for the Edge issue you referenced?

gkalpak: I don't know if there is a bug report - it just came up a few days ago and I added a workaround to fix a similar issue in #13887.

gkalpak: > Likely to be a problem if the input model is populated directly with an ISO formatted text string

The `ISO_DATE_REGEXP` is only used in an `ngModelController` parser, meaning it will only be applied when parsing the `$viewValue` to transform it to a `$modelValue`. The only case where the `$viewValue` can be an ISO-8601 date string, is on browsers that do not natively support `input[date]` and the users enter it by hand (which is a rather unlikely interaction imo).

Anyway, assuming we still need `ISO_DATE_REGEXP`, what needs to be done in order to properly support all valid timezone offsets on all supported browsers, is "fixing up" the date string before passing it to `new Date()`. That would include adding a colon if missing and add `00` minutes if omitted.

Thinking about it again, this is totally unrelated to the initial intent of this PR, so it's not a good idea to mix them. I suggest, leaving only changes to add support for >4-digit years in this PR and if you fancy feel free to submit another one with the required changes to properly support all valid timezone offset formats in another PR.

Sorry, I had you make those changes and now telling they don't belong here :disappointed:

Narretz:
@gkalpak Can you take another look at this?

gkalpak:
@Narretz, will do. Thx for the heads-up ! (I hadn't notice the updates.)

test/ng/directive/inputSpec.js
```diff
@@ -1941,6 +1989,125 @@ describe('input', function() {
...
+
+      they('should validate date: $prop', dates, function(item) {
+        var date = item[0];
+        var valid = item[1];
+
+        /* global ISO_DATE_REGEXP: false */
+        expect(ISO_DATE_REGEXP.test(date)).toBe(valid);
+      });
+
+      it('should be non-capturing', function() {
```
gkalpak: This test is not needed. Capturing vs Non-capturing is just an implementation detail.

gkalpak:
LGTM (expect for one minor comment).
Thx for working on this one, @LeeAdcock :+1:

test/ng/directive/inputSpec.js
LeeAdcock: Happy to make this change, but would point out that if having the regex be non-capturing is important for performance, we should have a test that validates it. The regex was originally capturing, but that mistake wasn't noticed until this unit test was written. It's not unlikely that the next person that modifies it after us will unintentionally make it capturing again, unless there is a test enforces this style of implementation for performance gains.

gkalpak: This is still an implementation detail. E.g. if we decide to add support for all timezone formats, this will change: we'll need some capturing groups. Unless it is identified as a performance issue (which it won't - it's more of a theoretical performance gain, unlikely to be noticable), it remains an implementation detail.

So, while it's nice to avoid capturing groups if you don't need them, having a test for it will only make it more difficult to add functionality later. Unit tests should assert the functionality/behavior, not the performance characteristics, imo.

LeeAdcock:
Thanks for the feedback. The implementation test has been removed.

gkalpak:
Backported to `v1.5.x` as d76951f.

PR #13907, opened 2016-01-31
docs(tutorial/7 - Routing

ZvikaZ:
Fixed typo in the URL at the beginning


gkalpak:
Thx for the PR, but I believe the URL was correct. It uses Angular's hashbang mode for client-side routing, so the path to the route appears in the hash fragment of the URL (_after_ the actual path, in this case: `/app/index.html`).

PR #13909, opened 2016-01-31
fix($rootScope): Set no context when calling listeners

lgalfaso:
When calling a listener, do not expose the inner workings with `this`.

### DIFF
--- a/test/ng/rootScopeSpec.js
+++ b/test/ng/rootScopeSpec.js
@@ -117,6 +117,20 @@ describe('Scope', function() {
     }));
 
 
+    it('should not expose the `inner working of watch', inject(function($rootScope) {
+      function Getter() {
+        expect(this).toBeUndefined();
+        return 'foo';
+      }
+      function Listener() {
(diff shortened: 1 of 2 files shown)


gkalpak:
The same is true for `watch.get`. Other than that LGTM.

test/ng/rootScopeSpec.js
```diff
@@ -117,6 +117,20 @@ describe('Scope', function() {
...
 
+    it('should not expose the `inner working of watch', inject(function($rootScope) {
+      function Getter() {
+        expect(this).toBeUndefined();
+        return 'foo';
+      }
+      function Listener() {
+        expect(this).toBeUndefined();
+      }
+      if (msie < 10) return;
```
gkalpak: OOC, what is the problem with IE9 ?

lgalfaso: There is no `"strict mode"` is IE9, so `this` is `window`

gkalpak:
Backported to `v1.4.x` as ab5c769.

PR #13913, opened 2016-01-31
fix(formatNumber): allow negative fraction size

jbedard:
This worked before 6a0686d434c41445c50b2d9669073802ede77b3b. The other option would be to not support this and just convert `fractionSize < 0` to `0`.

This solution could probably be improved though...

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -221,18 +221,37 @@ function roundNumber(parsedNumber, fractionSize, minFrac, maxFrac) {
     var digit = digits[roundAt];
 
     if (roundAt > 0) {
-      digits.splice(roundAt);
+      // Drop fractional digits beyond `roundAt`
+      digits.splice(Math.max(parsedNumber.i, roundAt));
+
+      // Set non-fractional digits beyond `roundAt` to 0
+      for (var j = roundAt; j < digits.length; j++) {
+        digits[j] = 0;
+      }
     } else {
       // We rounded to zero so reset the parsedNumber
+      fractionLen = Math.max(0, fractionLen);
       parsedNumber.i = 1;
-      digits.length = roundAt = fractionSize + 1;
-      for (var i=0; i < roundAt; i++) digits[i] = 0;
+      digits.length = Math.max(1, roundAt = fractionSize + 1);
+      digits[0] = 0;
+      for (var i = 1; i < roundAt; i++) digits[i] = 0;
     }
 
-    if (digit >= 5) digits[roundAt - 1]++;
+    if (digit >= 5) {
+      if (roundAt - 1 < 0) {
+        for (var k = 0; k > roundAt; k--) {
+          digits.unshift(0);
+          parsedNumber.i++;
+        }
+        digits.unshift(1);
+        parsedNumber.i++;
+      } else {
+        digits[roundAt - 1]++;
+      }
+    }
 
     // Pad out with zeros to get the required fraction length
-    for (; fractionLen < fractionSize; fractionLen++) digits.push(0);
+    for (; fractionLen < Math.max(0, fractionSize); fractionLen++) digits.push(0);
 
 
     // Do any carrying, e.g. a digit was rounded up to 10
--- a/test/ng/filter/filtersSpec.js
+++ b/test/ng/filter/filtersSpec.js
@@ -86,6 +86,26 @@ describe('filters', function() {
       expect(num).toBe('123.100');
     });
 
+    it('should work with negative fractionSize', function() {
+      expect(formatNumber(49, pattern, ',', '.', -2)).toBe('0');
+      expect(formatNumber(50, pattern, ',', '.', -2)).toBe('100');
+      expect(formatNumber(51, pattern, ',', '.', -2)).toBe('100');
+      expect(formatNumber(1234, pattern, ',', '.', -1)).toBe('1,230');
+      expect(formatNumber(1234.567, pattern, ',', '.', -1)).toBe('1,230');
+      expect(formatNumber(1235, pattern, ',', '.', -1)).toBe('1,240');
+      expect(formatNumber(1235, pattern, ',', '.', -2)).toBe('1,200');
+      expect(formatNumber(1235, pattern, ',', '.', -3)).toBe('1,000');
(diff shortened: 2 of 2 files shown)


src/ng/filter/filters.js
```diff
@@ -221,18 +221,32 @@ function roundNumber(parsedNumber, fractionSize, minFrac, maxFrac) {
...
+      // Set non-fractional digits beyond `roundAt` to 0
+      for (var j=roundAt; j < digits.length; j++) {
+        digits[j] = 0;
+      }
     } else {
       // We rounded to zero so reset the parsedNumber
       parsedNumber.i = 1;
-      digits.length = roundAt = fractionSize + 1;
-      for (var i=0; i < roundAt; i++) digits[i] = 0;
+      digits.length = Math.max(1, roundAt = fractionSize + 1);
```
gkalpak: AFAICT, `fractionSize + 1` can never be > 1, because:

```
fractionSize + 1 > 1 <=>
fractionSize > 0     <=>
fractionSize + parsedNumber.i > 0 (because `parsedNumber.i > 0`) <=>
roundedAt > 0 (on line 220) <=>
We wouldn't be on this branch of the if-else block
```

gkalpak: Hm...it seems that `parsedNumber.i` **can** be < 1, so ignore the above :)

test/ng/filter/filtersSpec.js
```diff
@@ -86,6 +86,20 @@ describe('filters', function() {
...
+      expect(formatNumber(1234.567, pattern, ',', '.', -1)).toBe("1,230");
+      expect(formatNumber(1235, pattern, ',', '.', -1)).toBe("1,240");
+      expect(formatNumber(1235, pattern, ',', '.', -2)).toBe("1,200");
+      expect(formatNumber(1235, pattern, ',', '.', -3)).toBe("1,000");
+      expect(formatNumber(1235, pattern, ',', '.', -4)).toBe("0");
+      expect(formatNumber(1236, pattern, ',', '.', -1)).toBe("1,240");
+      expect(formatNumber(1000, pattern, ',', '.', -1)).toBe("1,000");
+      expect(formatNumber(1, pattern, ',', '.', -1)).toBe("0");
+      expect(formatNumber(1, pattern, ',', '.', -2)).toBe("0");
+      expect(formatNumber(9, pattern, ',', '.', -1)).toBe("10");
```
gkalpak: Although the behavior of negative `fractionSize` isn't defined somewhere (afaik), there are some cases that behave differently that I expect:

**(A)**
```js
expect(formatNumber(1000, pattern, ',', '.', -4)).toBe("0");   // But the actual value is `0.000`
```

and

**(B)**
```js
expect(formatNumber(51, pattern, ',', '.', -2)).toBe('100');      // But the actual value is `10`
expect(formatNumber(501, pattern, ',', '.', -3)).toBe('1,000');   // But the actual value is `10`
```

Basically, there are two issues (both appearing when using a negative `fractionSize` whose absolute value equals the number of integer digit in the number):

* **(A)**: When the number ends with zeros (as in `1000`), unneeded "decimal" zeros are appended at the end of the result.

* **(B)**: When the first digit of the number is >= 5, it is always rounded up to `10`.

src/ng/filter/filters.js
```diff
@@ -221,18 +221,32 @@ function roundNumber(parsedNumber, fractionSize, minFrac, maxFrac) {
...
+      digits[0] = 0;
+      for (var i=1; i < roundAt; i++) digits[i] = 0;
     }
 
-    if (digit >= 5) digits[roundAt - 1]++;
+    if (digit >= 5) {
+      if (roundAt - 1 < 0) {
```
gkalpak: Inserting the following lines here, solves problem (A) described in https://github.com/angular/angular.js/pull/13913/files#r51409840:

```js
for (var k = 0; k > roundAt; k--) {
  digits.unshift(0);
  parsedNumber.i++;
}
```

src/ng/filter/filters.js
```diff
@@ -221,18 +221,32 @@ function roundNumber(parsedNumber, fractionSize, minFrac, maxFrac) {
...
-      digits.splice(roundAt);
+      // Drop fractional digits beyond `roundAt`
+      digits.splice(Math.max(parsedNumber.i, roundAt));
+
+      // Set non-fractional digits beyond `roundAt` to 0
+      for (var j=roundAt; j < digits.length; j++) {
+        digits[j] = 0;
+      }
     } else {
       // We rounded to zero so reset the parsedNumber
```
gkalpak: Inserted the following line here, solves problem (B) described in https://github.com/angular/angular.js/pull/13913/files#r51409840:

```js
fractionLen = Math.max(0, fractionLen);
```

gkalpak:
Other than the 2 issues pointed in the comments, it LGTM.

jbedard:
Added tests and your fixes. Thanks for showing the exact fixes!

src/ng/filter/filters.js
```diff
@@ -221,18 +221,37 @@ function roundNumber(parsedNumber, fractionSize, minFrac, maxFrac) {
     var digit = digits[roundAt];
 
     if (roundAt > 0) {
-      digits.splice(roundAt);
+      // Drop fractional digits beyond `roundAt`
+      digits.splice(Math.max(parsedNumber.i, roundAt));
+
+      // Set non-fractional digits beyond `roundAt` to 0
+      for (var j=roundAt; j < digits.length; j++) {
```
gkalpak: Interesting that JSCS doesn't complain about no space around `=` (I thought it would) :)

gkalpak:
LGTM

src/ng/filter/filters.js
jbedard: Updated! Line 237 also updated, note that that one had no spaces before this commit as well...

gkalpak: Yeah, I didn't mean you had to change it necessarily - I was just surprised with JSCS.
But look how much better it looks now :stuck_out_tongue:

Narretz:
@gkalpak do you want to merge?

gkalpak:
@Narretz, on it ! Thx for the reminder :smiley:

gkalpak:
Backported to `v1.5.x` as e046c17.

PR #13921, opened 2016-02-02
docs($compile): refine explanation of `=`-bindings

thorn0:
The current version of this paragraph is in many ways inaccurate and confusing.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -166,35 +166,38 @@
  * is bound to the parent scope, via matching attributes on the directive's element:
  *
  * * `@` or `@attr` - bind a local scope property to the value of DOM attribute. The result is
- *   always a string since DOM attributes are strings. If no `attr` name is specified  then the
- *   attribute name is assumed to be the same as the local name.
- *   Given `<widget my-attr="hello {{name}}">` and widget definition
- *   of `scope: { localName:'@myAttr' }`, then widget scope property `localName` will reflect
- *   the interpolated value of `hello {{name}}`. As the `name` attribute changes so will the
- *   `localName` property on the widget scope. The `name` is read from the parent scope (not
- *   component scope).
+ *   always a string since DOM attributes are strings. If no `attr` name is specified then the
+ *   attribute name is assumed to be the same as the local name. Given `<my-component
+ *   my-attr="hello {{name}}">` and the isolate scope definition `scope: { localName:'@myAttr' }`,
+ *   the directive's scope property `localName` will reflect the interpolated value of `hello
+ *   {{name}}`. As the `name` attribute changes so will the `localName` property on the directive's
+ *   scope. The `name` is read from the parent scope (not the directive's scope).
  *
- * * `=` or `=attr` - set up bi-directional binding between a local scope property and the
- *   parent scope property of name defined via the value of the `attr` attribute. If no `attr`
- *   name is specified then the attribute name is assumed to be the same as the local name.
- *   Given `<widget my-attr="parentModel">` and widget definition of
- *   `scope: { localModel:'=myAttr' }`, then widget scope property `localModel` will reflect the
- *   value of `parentModel` on the parent scope. Any changes to `parentModel` will be reflected
- *   in `localModel` and any changes in `localModel` will reflect in `parentModel`. If the parent
- *   scope property doesn't exist, it will throw a NON_ASSIGNABLE_MODEL_EXPRESSION exception. You
- *   can avoid this behavior using `=?` or `=?attr` in order to flag the property as optional. If
- *   you want to shallow watch for changes (i.e. $watchCollection instead of $watch) you can use
- *   `=*` or `=*attr` (`=*?` or `=*?attr` if the property is optional).
+ * * `=` or `=attr` - set up a bidirectional binding between a local scope property and an expression
+ *   passed via the attribute `attr`. The expression is evaluated in the context of the parent scope.
+ *   If no `attr` name is specified then the attribute name is assumed to be the same as the local
+ *   name. Given `<my-component my-attr="parentModel">` and the isolate scope definition `scope: {
+ *   localModel: '=myAttr' }`, the property `localModel` on the directive's scope will reflect the
+ *   value of `parentModel` on the parent scope. Changes to `parentModel` will be reflected in
+ *   `localModel` and vice versa. Optional attributes should be marked as such with a question mark:
+ *   `=?` or `=?attr`. If the binding expression is non-assignable, or if the attribute isn't
+ *   optional and doesn't exist, an exception ({@link error/$compile/nonassign `$compile:nonassign`})
+ *   will be thrown on discovering changes to the local value since it'll be impossible to sync it
+ *   with the parent scope in this case. By default, the {@link ng.$rootScope.Scope#$watch `$watch`}
+ *   method is used for tracking changes, and the equality check is based on object identity.
+ *   However, if an object literal or an array literal is passed as the binding expression, the
+ *   equality check is done by value (the {@link angular.equals} function is used). It's also
+ *   possible to watch the evaluated value shallowly with {@link ng.$rootScope.Scope#$watchCollection
+ *   `$watchCollection`}: use `=*` or `=*attr` (`=*?` or `=*?attr` if the attribute is optional).
  *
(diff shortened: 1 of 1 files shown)


src/ng/compile.js
```diff
@@ -174,17 +174,18 @@
...
- *   `scope: { localModel:'=myAttr' }`, then widget scope property `localModel` will reflect the
- *   value of `parentModel` on the parent scope. Any changes to `parentModel` will be reflected
- *   in `localModel` and any changes in `localModel` will reflect in `parentModel`. If the parent
- *   scope property doesn't exist, it will throw a NON_ASSIGNABLE_MODEL_EXPRESSION exception. You
- *   can avoid this behavior using `=?` or `=?attr` in order to flag the property as optional. If
- *   you want to shallow watch for changes (i.e. $watchCollection instead of $watch) you can use
- *   `=*` or `=*attr` (`=*?` or `=*?attr` if the property is optional).
+ * * `=` or `=attr` - set up a bidirectional binding between a local scope property and an expression
+ *   passed via the attribute `attr`. The expression is evaluated in the context of the parent scope and
+ *   can be either an assignable expression or a literal value. If no `attr` name is specified then the
```
gkalpak: Since you are at it, it might be a good idea so somehow mention that if a non-assignable expression is passed, then `$compile:nonassign` will be thrown when trying to assign to it.

src/ng/compile.js
```diff
@@ -174,17 +174,18 @@
...
+ *   attribute name is assumed to be the same as the local name. Given `<widget my-attr="parentModel">`
+ *   and the widget definition `scope: { localModel:'=myAttr' }`, the property `localModel` on the widget's
+ *   scope will reflect the value of `parentModel` on the parent scope. Changes to `parentModel` will be
+ *   reflected in `localModel` and vice versa. If the attribute doesn't exist, an exception will be thrown
+ *   ({@link error/$compile/nonassign $compile:nonassign}). You can avoid this behavior and flag the
+ *   property as optional using `=?` or `=?attr`. By default, the {@link ng.$rootScope.Scope#$watch} method
+ *   is used, the equality check is done by reference. If you need to watch a collection for changes using
```
gkalpak: _is used, the equality_  -->  _is used and the equality_ (?)

gkalpak: "watch a collection" is a little vague (users might wonder if a collection is something angular specific). I slightly preferred the previous phrasing.

thorn0: The previous phrasing is really misleading. "If you want to _shallow_ watch for changes" sounds like by default the watch is deep whereas it's not the case.

gkalpak: Fair enough :smiley:

thorn0:
Is it looking better now?

src/ng/compile.js
```diff
@@ -174,17 +174,20 @@
...
- *   can avoid this behavior using `=?` or `=?attr` in order to flag the property as optional. If
- *   you want to shallow watch for changes (i.e. $watchCollection instead of $watch) you can use
- *   `=*` or `=*attr` (`=*?` or `=*?attr` if the property is optional).
+ * * `=` or `=attr` - set up a bidirectional binding between a local scope property and an expression
+ *   passed via the attribute `attr`. The expression is evaluated in the context of the parent scope and
+ *   must be either an assignable expression or a literal value. If no `attr` name is specified then the
+ *   attribute name is assumed to be the same as the local name. Given `<widget my-attr="parentModel">`
+ *   and the widget definition `scope: { localModel:'=myAttr' }`, the property `localModel` on the widget's
+ *   scope will reflect the value of `parentModel` on the parent scope. Changes to `parentModel` will be
+ *   reflected in `localModel` and vice versa. If the attribute doesn't exist, an exception will be thrown
```
gkalpak: Actually, I think it will only throw if you try to **assign** and the expression is non-assignable or the attribute is missing.

thorn0: Right. I believe that's exactly what I wrote.

gkalpak: I'm not sure I was clear. I meant it will throw if:

1. The expression is non-assignable and you try to assign to it.
2. The attribut is missing **and you try to assign to it**.

I believe that the part in bold is missing from the description.

thorn0: Ah, I see. You're right. But is it worth going into these details when all we need here is just to give a notion of optional bindings? Won't it overcomplicate the explanation?

However, I found another interesting detail missing from this text. Namely, when the expression is an object or array literal, the watch is deep.

thorn0: I mean, if the attribute is optional, please mark it as such. The idea should be like this. It doesn't really matter how and when exactly it fails if you didn't mark it.

thorn0:
@gkalpak pls have a look

src/ng/compile.js
```diff
@@ -174,17 +174,22 @@
...
- *   `scope: { localModel:'=myAttr' }`, then widget scope property `localModel` will reflect the
- *   value of `parentModel` on the parent scope. Any changes to `parentModel` will be reflected
- *   in `localModel` and any changes in `localModel` will reflect in `parentModel`. If the parent
- *   scope property doesn't exist, it will throw a NON_ASSIGNABLE_MODEL_EXPRESSION exception. You
- *   can avoid this behavior using `=?` or `=?attr` in order to flag the property as optional. If
- *   you want to shallow watch for changes (i.e. $watchCollection instead of $watch) you can use
- *   `=*` or `=*attr` (`=*?` or `=*?attr` if the property is optional).
+ * * `=` or `=attr` - set up a bidirectional binding between a local scope property and an expression
+ *   passed via the attribute `attr`. The expression is evaluated in the context of the parent scope and
+ *   must be either an assignable expression or a literal value. If no `attr` name is specified then the
```
gkalpak: It is not required to be an assignable expression. Only if you try to assign to it :)

thorn0: Done

Narretz:
Looks good. One suggestion, can you change all instances of widget to directive? I widget is a relict of very old times, iirc, back before the term directive was introduced.

src/ng/compile.js
```diff
@@ -174,17 +174,22 @@
...
+ *   If no `attr` name is specified then the attribute name is assumed to be the same as the local
+ *   name. Given `<widget my-attr="parentModel">` and the widget definition `scope: { localModel:
+ *   '=myAttr' }`, the property `localModel` on the widget's scope will reflect the value of
+ *   `parentModel` on the parent scope. Changes to `parentModel` will be reflected in `localModel`
+ *   and vice versa. Optional attributes should be marked as such with a question mark: `=?` or
+ *   `=?attr`. If the binding expression is non-assignable, or if the attribute isn't optional and
+ *   doesn't exist, an exception ({@link error/$compile/nonassign `$compile:nonassign`}) will be
+ *   thrown on discovering changes to the local value since it'll be impossible to sync it with the
+ *   parent scope in this case. By default, the {@link ng.$rootScope.Scope#$watch `$watch`} method is
+ *   used for tracking changes, and the equality check is done by reference. However, if an object
```
Narretz: Personally, I find it clearer to say that the equality check is based on object identity. I guess it means the same though.

thorn0:
> can you change all instances of widget to directive?

What about "component"? It's already called so a bit above:

> **`{...}` (an object hash):** A new "isolate" scope is created for the directive's element. [...] This is useful when creating reusable components...

Narretz:
I would be okay with `<my-component my-attr="parentModel">`, but `directive definition` and `directive's scope`. Component is a rather specific term. I think the part you quoted shows that. Components imply a certain modularity, but directives can also add events etc. without being components.

thorn0:
OK. Eliminated widgets and replaced 'by reference' with 'based on object identity'.

But I also don't like that it says "local scope", "directive's scope" everywhere whereas the same binding mechanism is used for controllers.

Narretz:
I think that's okay. The text is very information dense anyway. bindToController should make clear that it changes this behavior.

thorn0:
I think bindings should get a higher level section of their own, like [Transcusion](https://docs.angularjs.org/api/ng/service/$compile#transclusion). It can be done later though.

gkalpak:
LGTM

Narretz:
@gkalpak Can you merge? Then I can rebase the one-way PR.

gkalpak:
@Narretz: Merged.
(Backported to `v1.4.x` as 8a1f600.)

PR #13922, opened 2016-02-02
refactor(injector): remove unneeded JSHint directives

thorn0:
The difference between applying a constructor with and without parentheses is merely syntactic.


gkalpak:
This was supposed to be part of #13785, but since it's not certain that #13785 will land, let's merge this :)

Issue #13854, opened 2016-01-27
Feature request: make one-way expression bindings for components


thorn0:
See http://teropa.info/blog/2015/10/18/refactoring-angular-apps-to-components.html#replace-two-way-binding-with-one-way-binding

There seems to be little value in the ability of components to change values in the parent scope via two-way `=`-bindings. `@`-bindings are one-way, but they can be used only for string values. It'd be logical to introduce a new kind of one-way bindings for expressions of any type or to change the behavior of `=`-bindings for components (defined with the `component` method).

17cupsofcoffee:
:+1: This would make it a lot easier to migrate components across to Angular 2, as well - the less we're relying on two-way binding, the less we'll have to refactor later on. At the minute I'm just trying to enforce 'don't mutate the inputs' as a code convention, but it'd be great to not have to rely on that.

Narretz:
There's actually a PR for that: https://github.com/angular/angular.js/pull/12835
I personally like the idea.

robstove:
Nice start. Out of curiosity, what is the reasoning for using '#' for specifying the binding mode?

thorn0:
I believe it's still not late for the 1.5 release to include this. It'd be really consistent with the rest of the changes related to the component approach. Please consider extracting it from the ice box before the release.

Let me quote the description of the Ice Box milestone:

> Milestone for valid issues and bugs that we have no aspirations on fixing any time soon because they are not affecting primary use-cases and are not aligned with our goals.

I can't agree that it's not aligned with the current goal of **migration facilitation**.

robstove:
@thorn0 +1

Narretz:
We'll talk about this in the meeting today.

Regarding the symbol for it, there's nothing set, but `>` is favorable. `-` is also okay, but might be confused with the dash delimiter. `#` has no real connection to bindings.

thorn0:
@Narretz That's great. As for the argument of 'further complicating the DDO API', the API docs can be made to appear less complicated by adding a standard highlighted paragraph 'Best Practice', which would recommend using these new bindings for inputs and `&` (or events?) for outputs.

robstove:
Is this being targeted for 1.5?

Narretz:
@robstove 

[...]

PR #13928, opened 2016-02-02
feat($compile): add one-way binding to the isolate scope defintion

Narretz:


src/ng/compile.js
```diff
@@ -3032,6 +3052,29 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+            parentGet = $parse(attrs[attrName]);
+
+            destination[scopeName] = parentGet(scope);
+
+            function onParentValueChange(newParentValue) {
+              destination[scopeName] = newParentValue;
+            }
+
+            if (definition.collection) {
+              removeWatch = scope.$watchCollection($parse(attrs[attrName]), onParentValueChange);
```
gkalpak: Couldn't `parentGet` be used here (to avoid double parsing) ?

test/ng/compileSpec.js
```diff
@@ -3533,6 +3533,14 @@ describe('$compile', function() {
             optreference: '=?',
             colref: '=*',
             colrefAlias: '=* colref',
+            owRef: '<',
+            owRefAlias: '< owRef',
+            owReference: '<',
```
gkalpak: Why both `owRef` and `owReference` ?

src/ng/compile.js
```diff
@@ -186,6 +186,27 @@
...
+ *   two caveats:
+ *     1. one-way binding does not copy the value from the parent to the isolate scope, it simply
+ *     sets the same value. That means if your bound value is an object, changes to its properties
+ *     in the isolate scope will be reflected in the parent scope.
+ *     2. one-way binding watches changes to the **identity** of the parent value. That is important should
+ *     you replace a bound object in the isolated scope. In that case, you must also replace your
+ *     object in ther parent scope with a new one, otherwise the one-way binding won't pick up changes.
+ *
+ *   One-way binding is useful if you do not plan to propagate changes to your isolated scope bindings
+ *   back to the parent. It does not however not make this completely impossible.
```
gkalpak: _not ... not_  -->  _not ..._

test/ng/compileSpec.js
```diff
@@ -4219,6 +4227,299 @@ describe('$compile', function() {
     });
 
 
+    describe('one-way binding', function() {
+      it('should update isolate when the identity of origin changes', inject(function() {
+        compile('<div><span my-component ow-ref="obj">');
+        expect(componentScope.owRef).toBe(undefined);
+        expect(componentScope.owRefAlias).toBe(componentScope.owRef);
```
gkalpak: This only checks that `componentScope.owRefAlias` is `undefined`.
Shouldn't there be similar expectations below ?

test/ng/compileSpec.js
```diff
@@ -4219,6 +4227,299 @@ describe('$compile', function() {
     });
 
 
+    describe('one-way binding', function() {
+      it('should update isolate when the identity of origin changes', inject(function() {
```
gkalpak: Unecessary `inject()` (here and below) ?

src/ng/compile.js
thorn0: I'm reading and re-reading the Caveat No 2 and can't get it. Why do I have to replace something in the parent scope? Isolated components don't care about and don't touch the parent scope.

Narretz: Hehe, yeah the explanation was problematic to write. Basically, if you replace the bound object in the isolate scope, changing only a property of the object in the parent scope won't update the child scope, because the watch on the parent scope only picks up when the identity of the parent scope value changes.

src/ng/compile.js
```diff
@@ -3032,6 +3052,29 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+            parentGet = $parse(attrs[attrName]);
+
+            destination[scopeName] = parentGet(scope);
+
+            if (definition.collection) {
+              removeWatch = scope.$watchCollection(parentGet), function onParentValueChange(newParentValue) {
+                destination[scopeName] = newParentValue;
+              });
+            } else {
+              removeWatch = scope.$watch(parentGet), function onParentCollectionValueChange(newParentValue) {
```
thorn0: The names of the watch listener functions should switch places.

src/ng/compile.js
thorn0: BTW, please have a look at #13921. The paragraph about `=`-bindings, which you took as a starting point, is pretty outdated and inaccurate.

src/ng/compile.js
Narretz: Do we even need to parse this? We are just passing the string to watch, which parses this anyway. Yeah, I think I was confused about what's going oon in the two-way binding

thorn0: I think we should parse it to init the child value. However, you can do this
```js
scope.$watchCollection(parentGet, ...)
```
to not parse twice.

src/ng/compile.js
```diff
@@ -3032,6 +3052,29 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+            parentGet = $parse(attrs[attrName]);
+
+            destination[scopeName] = parentGet(scope);
+
+            if (definition.collection) {
+              removeWatch = scope.$watchCollection(attrs[attrName], function onParentValueChange(newParentValue) {
+                destination[scopeName] = newParentValue;
+              });
+            } else {
+              removeWatch = scope.$watch(attrs[attrName], function onParentCollectionValueChange(newParentValue) {
```
campersau: Are the function names for the watch callbacks swapped?
Should't it be `onParentValueChange` here and `onParentCollectionValueChange` above?

src/ng/compile.js
```diff
@@ -186,6 +186,27 @@
  *   you want to shallow watch for changes (i.e. $watchCollection instead of $watch) you can use
  *   `=*` or `=*attr` (`=*?` or `=*?attr` if the property is optional).
  *
+ * * `<` or `<attr` - set up one-way (one-directional) binding between a local scope property and the
```
petebacondarwin: I think the character should be `>` giving more feeling of forward movement

thorn0: As I wrote in #13854:

> `bindings: { propertyName: "<attrName" }`
> Which means: the value is copied from the attribute `attrName` to the property `propertyName`.

test/ng/compileSpec.js
Narretz: They are necessary. (many other tests have them). I  think it's because the $rootScope etc. "globals" are set up in a module'S return fn.

test/ng/compileSpec.js
Narretz: Updated

src/ng/compile.js
petebacondarwin: Trouble is that in my mind `<attrName` looks like a broken HTML element

petebacondarwin: How about `<=`? It has the same connatation of moving from attribute to property (even more so) it has an element of `=` about it and looks nice `propertyName: '<=attrName'`

thorn0: Having this connotation is not as important as **not having** the opposite. As for multi-character sigils, I think they would look too scary in combination with `?` and `*`:
```js
bindings: { propertyName: "<=*?attrName" } // oh...
```

petebacondarwin: I'm happy to go with the consensus and get this in :-)

test/ng/compileSpec.js
```diff
@@ -3533,6 +3533,13 @@ describe('$compile', function() {
             optreference: '=?',
             colref: '=*',
             colrefAlias: '=* colref',
+            owRef: '<',
+            owRefAlias: '< owRef',
+            owOptref: '<?',
+            owOptrefAlias: '<? owOptref',
+            owOptreference: '<?',
```
gkalpak: OOC, what is the purpose of `owOptref` + `owOptreference` ?

test/ng/compileSpec.js
```diff
@@ -4219,6 +4226,303 @@ describe('$compile', function() {
...
+        expect(componentScope.owRef).toBe($rootScope.obj);
+
+        componentScope.owRef.mark = 789;
+        $rootScope.$apply();
+        expect($rootScope.obj).toEqual({mark: 789});
+        expect(componentScope.owRef).toBe($rootScope.obj);
+      }));
+
+
+      it('should not throw on non assignable changes', inject(function() {
```
gkalpak: Did you mean "non assignable expressions" ?

test/ng/compileSpec.js
```diff
@@ -4219,6 +4226,303 @@ describe('$compile', function() {
...
+      }));
+
+
+      it('should update isolate scope when "<"-bound NaN changes', inject(function() {
+        $rootScope.num = NaN;
+        compile('<div my-component ow-ref="num"></div>');
+        var isolateScope = element.isolateScope();
+        expect(isolateScope.owRef).toBeNaN();
+
+        $rootScope.$apply(function(scope) { scope.num = 64; });
```
gkalpak: People don't seem to prefer it, but many `assignment + $apply()` or `$apply(callback)` can be shortened as `$apply('expresion')`. E.g.: `$rootScope.$apply('num = 64')`. Just saying... (it doesn't make any difference anyway :smiley:)

test/ng/compileSpec.js
```diff
@@ -4219,6 +4226,303 @@ describe('$compile', function() {
...
+        it('should not complain when the component changes', inject(function() {
+          compile('<div><span my-component ow-ref="{name: name}">');
+
+          $rootScope.name = 'a';
+          $rootScope.$apply();
+          componentScope.owRef = {name: 'b'};
+          componentScope.$apply();
+          expect(componentScope.owRef).toEqual({name: 'b'});
+          expect($rootScope.name).toBe('a');
+        }));
```
gkalpak: I think it would be worth having a test for what happens after changing a property on `$rootScope` after `componentScope.owRef = {name: 'b'}`.
If I am not mistaken, there is no way to overwrite `componentScope.owRef` when using a literal object, right ?

test/ng/compileSpec.js
```diff
@@ -4219,6 +4226,303 @@ describe('$compile', function() {
...
+
+
+        it('should work for primitive literals', inject(function() {
+          test('1', 1);
+          test('null', null);
+          test('undefined', undefined);
+          test("'someString'", 'someString');
```
gkalpak: What about `true/false` ?

test/ng/compileSpec.js
Narretz: Came from a c&p'd test, but `owOptreference` doesn't look like it's needed.

test/ng/compileSpec.js
Narretz: Probably

test/ng/compileSpec.js
Narretz: This is actually the only test in this block that uses this, because it was added later, all the others use

    $rootScope.num = 64;.
    $rootScope.$apply();

test/ng/compileSpec.js
Narretz: I believe so, yes. That's the same with two-way data-binding, the literal always wins.

Narretz: Actually, the test shows currently shows that it is possible.

thorn0: It's possible even with `=`-bindings now. If a literal is passed, they work like they're one-way.

test/ng/compileSpec.js
Narretz: Can do

test/ng/compileSpec.js
Narretz: Ah, I overlooked a small detail. Yes, I can add an expectation for what happens when I change name to 'c' on the origin scope (Spoilers, it works)

Narretz: Done

thorn0:
Currently, if an object literal or an array literal is passed as the expression for a two-way (`=`) binding, the set up watch is deep (uses `angular.equals`). The implementation in this PR doesn't do this with the new one-way bindings. IMO it'd more consistent if one-way bindings behaved exactly like two-way bindings, just without probability for the parent to get modified by them (with the known caveats).

test/ng/compileSpec.js
gkalpak: I mean

```js
$rootScope.num = 64;
$rootScope.$apply();
```

can be written much more conviently as

```js
$rootScope.$apply('num = 64');
```

But for some reason, everyone seems to avoid it :smiley:
Really, I don't know why I keep bringing this up, since it doesn't make any difference (in tests).

test/ng/compileSpec.js
```diff
@@ -4219,6 +4223,308 @@ describe('$compile', function() {
...
+          $rootScope.$apply();
+          componentScope.owRef = {name: 'b'};
+          componentScope.$apply();
+
+          expect(componentScope.owRef).toEqual({name: 'b'});
+          expect($rootScope.name).toBe('a');
+
+          $rootScope.name = 'c';
+          $rootScope.$apply();
+          expect(componentScope.owRef).toEqual({name: 'c'});
```
gkalpak: I expected it to remain `{name: 'b'}`. Interesting...

test/ng/compileSpec.js
```diff
@@ -4219,6 +4223,308 @@ describe('$compile', function() {
...
+            $rootScope.$apply();
+            expect(componentScope.owRef).toBe(literalValue);
+            dealoc(element);
+          }
+        }));
+
+        describe('optional one-way binding', function() {
+          it('should update local when origin changes', inject(function() {
+            compile('<div><span my-component ow-optref="name">');
+            expect(componentScope.owOptRef).toBe(undefined);
```
gkalpak: _owOptRef_ --> _owOptref_

test/ng/compileSpec.js
```diff
@@ -4219,6 +4223,308 @@ describe('$compile', function() {
...
+            expect(componentScope.owOptRefAlias).toBe(componentScope.owOptRef);
```
gkalpak: _owOptRefAlias_  -->  _owOptrefAlias_
_owOptRef_  -->  _owOptref_

src/ng/compile.js
```diff
@@ -3032,6 +3053,29 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+              attrs[attrName] = void 0;
+            }
+            if (optional && !attrs[attrName]) break;
+
+            parentGet = $parse(attrs[attrName]);
+
+            destination[scopeName] = parentGet(scope);
+
+            if (definition.collection) {
+              removeWatch = scope.$watchCollection(attrs[attrName], function onParentCollectionValueChange(newParentValue) {
```
gkalpak: I think we can reuse `parentGet` (here and below) to avoid re-parsing the expression.

test/ng/compileSpec.js
gkalpak: Aha, parsed expression with `inputs` and `$$watchDelegates`. Nice :smiley:

Narretz: That's why it works? I'm actually confused by this right now.

src/ng/compile.js
Narretz: It works, but I wonder if $parse isn't doing extra work? Or does it detect an already parsed expression?

Narretz:
Hm, something strange is going on. The watchAction functions fire twice every time there's an apply. I don't think this should happen. Maybe that's also why the literals work even though they shouldn't.

src/ng/compile.js
robstove: What was the reason `-` didn't make sense?

gkalpak:
A couple of minor comments, but LGTM otherwise, expect for the fact that we should be more detailed on what is updated when and what is the difference between `*` and no `*`.

test/ng/compileSpec.js
gkalpak: Yes. Parsing `{name: name}` the resulting parsed expression has `inputs`, so it has a `$$watchDelegate` that will re-evaluate the expression whenever the inputs change value (in this case, `inputs` is the `name` property on `$rootScope`).
So, whenever `$rootScope.name` changes, the expression will be evaluated and will return a new object, thus the change will be propagated to the child.

Narretz: So this works, but when you are going deeper, you need objectEquality watch - I've added that right now.

src/ng/compile.js
```diff
@@ -3032,6 +3053,29 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+              attrs[attrName] = void 0;
+            }
+            if (optional && !attrs[attrName]) break;
+
+            parentGet = $parse(attrs[attrName]);
+
+            destination[scopeName] = parentGet(scope);
+
+            if (definition.collection) {
+              removeWatch = scope.$watchCollection(parentGet, function parentCollectionValueWatchAction(newParentValue) {
```
lgalfaso: I do not understand what is the use case to use `$watchCollection`. We are not copying things from one side to the other, so I would just drop support for `'<*'`, have the minimal viable feature and cut this release.

If, later on, people ask for this feature, then we can take the time to think if this is the way that best fits the needs.

gkalpak: I agree with @lgalfaso.

The only difference with `<*` is when the parent will be "synced" to the child and this only matters, if the child has set its property to a new instance. Since I can't think of any good usecase for setting the child to a new instance (knowing that at any point the parent could overwrite that value), I believe that `<*` will only cause confusion.

Keeping this feature as simple as possible (which means simpler semantics, less corner-cases, less unexpected behaviors) is pretty important imo.

As @lgalfaso said, we can always add stuff/features, but removing isn't that simple. Let's start simple.
<sub>(BTW, as @lgalfaso pointed out, there is no `*` in ng2, so I think it helps to be consistent.)</sub>

thorn0: @lgalfaso Well spotted!

Narretz:
I've 
- removed the collectionWatch functionality
- updated the description to incorporate changes made by @thorn0 
- added watching literals by object equality

src/ng/compile.js
```diff
@@ -186,6 +186,30 @@
  *   you want to shallow watch for changes (i.e. $watchCollection instead of $watch) you can use
  *   `=*` or `=*attr` (`=*?` or `=*?attr` if the property is optional).
  *
+ * * `<` or `<attr` - set up a one-way (one-directional) binding between a local scope property and an
+ *   expression passed via the attribute `attr`. The expression is evaluated in the context of the
+ *   parent scope. If no `attr` name is specified then the attribute name is assumed to be the same as the
+ *   local name. You can also make the binding optional by adding `?`: `<?` or `<?attr`.
+ *
+ *   For example, given `<dir my-attr="parentModel">` and directive definition of
```
gkalpak: _dir_  -->  _my-component_ (to be consistent with #13921)

src/ng/compile.js
```diff
@@ -186,6 +186,30 @@
...
+ *   `scope: { localModel:'<myAttr' }`, then the isolated scope property `localModel` will reflect the
+ *   value of `parentModel` on the parent scope. Any changes to `parentModel` will be reflected
+ *   in `localModel`, but changes in `localModel` will not reflect in `parentModel`. There are however
+ *   two caveats:
+ *     1. one-way binding does not copy the value from the parent to the isolate scope, it simply
+ *     sets the same value. That means if your bound value is an object, changes to its properties
+ *     in the isolated scope will be reflected in the parent scope (because both reference the same object).
+ *     2. one-way binding watches changes to the **identity** of the parent value. That means the
+ *     {@link ng.$rootScope.Scope#$watch`$watch`} on the parent value only fires if the reference
```
gkalpak: Missing space before `` `$watch` ``.

src/ng/compile.js
```diff
@@ -186,6 +186,30 @@
...
+ *     to the value has changed. In most cases, this should not be of concern, but can be important
+ *     to know if you one-way bind to an object, and then replace that object in the isolated scope.
+ *     If you now change a property of the object in
+ *     your parent scope, the change will not be propagated to the isolated scope, because the identity
+ *     of the object has not changed. Instead you must assign a new object.
```
gkalpak: I would change _"the identity of the object has not changed"_ to _"the identity of the object in he parent scope has not changed"_.

gkalpak:
LGTM (expect for some minor docs typos/fixes).
After this gets merged, it might be a goood idea to add an example in the Directives guide.

src/ng/compile.js
jbedard: I just found a use case for this today while using a 2-way binding (and wishing it was a 1-way binding)... when doing things like `<element binding="array | orderBy:'foo'">` using a `=` binding causes an infinite digest while using `=*` works. But `=*` or `<*` really doesn't seem like a great solution for that, it would be great if there was some other way. Maybe things like `filter` and `orderby` should somehow return the same instance each time if it hasn't changed?

lgalfaso: @jbedard when you have `<element binding="array | orderBy:'foo'">` we are not watching the result of the array after it went thru the filter, but before. Anyhow, if the behavior of `filter` and `orderBy` should change, then that should be another issue's discussion

jbedard: > we are not watching the result of the array after it went thru the filter, but before

I don't think that's right. Just doing `$rootScope.$watch("a | orderBy"); $rootScope.$apply("a = []")` causes an infinite digest.  I believe `=*` was created specifically to avoid this. However I think `=*` is a bad solution, it would be better if $parse/$watch handled this somehow...

PR #13936, opened 2016-02-03
docs(guide/accessibility): fix links

gkalpak:


gkalpak:
Backported to `v1.4.x` as 8519b8a.

PR #13943, opened 2016-02-04
Docs: Convert comp guide to one-way bindings, update iscp error

Narretz:


petebacondarwin:
LGTM as long as you are sure that the cases where you have changed it to `<` are definitely valid.

Narretz:
@petebacondarwin What do you mean by valid? < can be used for everything, imo. I've only used @ because it doesn't two-way bind. But with < we don't have to interpolate.

petebacondarwin:
I mean that does the change actually improve the use case? Does it take into account that although the binding is now one-way, we can still change properties of objects, etc, etc.

docs/content/error/$compile/iscp.ngdoc
```diff
@@ -12,9 +12,11 @@ myModule.directive('directiveName', function factory() {
     scope: {
       'attrName': '@', // OK
       'attrName2': '=localName', // OK
+      'attrName2': '<?localName', // OK
       'attrName3': 'name',    // ERROR: missing mode @&=
       'attrName4': ' = name', // ERROR: extra spaces
```
gkalpak: I just noticed this. I think space is allowed around the symbols.

docs/content/error/$compile/iscp.ngdoc
```diff
@@ -12,9 +12,11 @@ myModule.directive('directiveName', function factory() {
     scope: {
       'attrName': '@', // OK
       'attrName2': '=localName', // OK
+      'attrName2': '<?localName', // OK
```
gkalpak: I think it's better to increment all attrName suffixes (e.g. here: _attrName2_  -->  _attrName3_).

Narretz:
Hmm ..yeah, I don't mention that you can still update the parent expression if you are not careful. Maybe I should. One thing that's now easier is that we don't use @ for strings anymore (which I used previously because they don't bind to the parent scope)

docs/content/guide/component.ngdoc
```diff
@@ -109,13 +109,16 @@ change will be reflected in the parent component. For components however, only t
 the data should modify it, to make it easy to reason about what data is changed, and when. For that reason,
 components should follow a few simple conventions:
 
-  - Inputs are realized with `@` and `=` bindings
+  - Inputs are realized with `<` bindings. The `<` symbol denotes {@link $compile#-scope- one-way bindings} which are
```
gkalpak: I don't feel we should promote `<` over `@`. `@` use `$observe` which could be efficient in many cases.

docs/content/guide/component.ngdoc
```diff
@@ -303,12 +306,16 @@ application, every view is a component:
 ```
 <br />
 When using {@link ngRoute.$routeProvider $routeProvider}, you can often avoid some
-boilerplate, by assigning the resolved dependencies directly to the route scope:
+boilerplate, by assigning the resolved dependencies directly to the route scope. Since 1.5,
+ngRoute automatically assigns the resolves to the route scope property `$resolve` (you can also
+configure that property). When using components, you can take advantage of this and pass resolves
```
gkalpak: Maybe: _that property_  -->  _the property name_

docs/content/guide/component.ngdoc
```diff
@@ -252,8 +255,8 @@ it upwards to the heroList component, which updates it.
     templateUrl: 'editableField.html',
     controller: EditableFieldController,
     bindings: {
-      fieldValue: '@',
-      fieldType: '@',
+      fieldValue: '<',
+      fieldType: '<?',
```
petebacondarwin: Do you think that `fieldType` should stay as `@` or maybe `@?`?

gkalpak: E.g. this change will create one extra `$watch`.

docs/content/guide/component.ngdoc
gkalpak: I'm fine replacing `=` with `<`, but I would leave `@` for where it is appropriate.

Narretz: Hm, yes it's true that @ without an interpolation is faster than <. But otherwise you just need @ if you really care about the value being set in the attribute.

docs/content/guide/component.ngdoc
Narretz: In this specific case. Otherwise the number of watches is the same, no (1 for interpolation, 1 for < binding)

docs/content/guide/component.ngdoc
Narretz: Another advantage of < (or = for that matter) is that the type of object is preserved, while @ only supports strings. Static literals are also unwatched automatically, so there's no real performance benefit afaik.

Narretz: Another advantage of < (or = for that matter) is that the type of object is preserved, while @ only supports strings. Static literals are also unwatched automatically, so there's no real performance benefit afaik.

gkalpak: You are right. For the type of things that can be passed with `@`, there shouldn't be any perf gain over `<` (except for some absolutely negligible initial setup (watching + immediately unwatching)).

So, I'm fine with replacing `@` with `<` (but one has to wrap the constant values in quotes).

petebacondarwin: My concern was really that `@` should be used when you are most likely not to pass an expression but a simple string. By changing it from `@` to `<` (or `=`) your attribute must change from `<editable-field field-type="text">` to `<editable-field field-type="'text'">` (note the extra quotes).

docs/content/error/$compile/iscp.ngdoc
Narretz: That'S right! I moved this into the OK section

Narretz:
I re-added `@` and added some hopefully clarifying info about one-way binding.

petebacondarwin:
LGTM

PR #13945, opened 2016-02-04
fix($sniffer): fix history sniffing in Chrome Packaged Apps

### LINKED ISSUE #11932: history.pushState is not available in packaged apps.
I get this error message when I run my chrome apps since m43 of chrome.  Seems related to this change. https://codereview.chromium.org/988293002

It doesn't have an negative effect on my apps so far, but I only get it when I use angular (3.15) in my chrome apps.

I saw a related closed issue.
https://github.com/angular/angular.js/issues/1358

I'm using the angular-csp.css file in my app.

gkalpak:
Although `window.history` is present in the context of Chrome Packaged Apps, it is not allowed to
access `window.history.pushState` or `window.history.state`, resulting in errors when trying to
"sniff" history support.
This commit fixes it by detecting a Chrome Packaged App (through the presence of
`window.chrome.app.runtime`). Note that `window.chrome.app` is present in the context of "normal"
webpages as well, but it doesn't have the `runtime` property, which is only available to packaged
apps (e.g. see https://developer.chrome.com/apps/api_index).

Fixes #11932
## 

The first commit is just some cleanup (that makes the diff look worse than it actually is).
Included it separately, so the fix is easier to review - could be squashed upon merging.

### DIFF
--- a/src/ng/sniffer.js
+++ b/src/ng/sniffer.js
@@ -17,6 +17,8 @@
 function $SnifferProvider() {
   this.$get = ['$window', '$document', function($window, $document) {
     var eventSupport = {},
+        isChromePackagedApp = $window.chrome && $window.chrome.app && $window.chrome.app.runtime,
+        hasHistoryPushState = $window.history && $window.history.pushState,
         android =
           toInt((/android (\d+)/.exec(lowercase(($window.navigator || {}).userAgent)) || [])[1]),
         boxee = /Boxee/i.test(($window.navigator || {}).userAgent),
@@ -57,11 +59,13 @@ function $SnifferProvider() {
       // http://code.google.com/p/android/issues/detail?id=17471
       // https://github.com/angular/angular.js/issues/904
 
+      // Chrome Packaged Apps are not allowed to access `history.pushState`. They can be detected by
+      // the presence of `chrome.app.runtime` (see https://developer.chrome.com/apps/api_index)
       // older webkit browser (533.9) on Boxee box has exactly the same problem as Android has
       // so let's not use the history API also
       // We are purposefully using `!(android < 4)` to cover the case when `android` is undefined
       // jshint -W018
-      history: !!($window.history && $window.history.pushState && !(android < 4) && !boxee),
+      history: !!(!isChromePackagedApp && hasHistoryPushState && !(android < 4) && !boxee),
       // jshint +W018
       hasEvent: function(event) {
         // IE9 implements 'input' event it's so fubared that we rather pretend that it doesn't have
--- a/src/ng/sniffer.js
+++ b/src/ng/sniffer.js
@@ -17,8 +17,10 @@
 function $SnifferProvider() {
   this.$get = ['$window', '$document', function($window, $document) {
     var eventSupport = {},
+        // Chrome Packaged Apps are not allowed to access `history.pushState`. They can be detected by
+        // the presence of `chrome.app.runtime` (see https://developer.chrome.com/apps/api_index)
         isChromePackagedApp = $window.chrome && $window.chrome.app && $window.chrome.app.runtime,
-        hasHistoryPushState = $window.history && $window.history.pushState,
+        hasHistoryPushState = !isChromePackagedApp && $window.history && $window.history.pushState,
         android =
           toInt((/android (\d+)/.exec(lowercase(($window.navigator || {}).userAgent)) || [])[1]),
         boxee = /Boxee/i.test(($window.navigator || {}).userAgent),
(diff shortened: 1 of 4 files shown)


src/ng/sniffer.js
```diff
@@ -17,6 +17,8 @@
 function $SnifferProvider() {
   this.$get = ['$window', '$document', function($window, $document) {
     var eventSupport = {},
+        isChromePackagedApp = $window.chrome && $window.chrome.app && $window.chrome.app.runtime,
+        hasHistoryPushState = $window.history && $window.history.pushState,
```
Narretz: Just an idea, but could we use `'pushstate' in $window.history`?

gkalpak: Yes, we can. But what is the benefit ? We want to return `false` in Chrome Packaged Apps (and `'pushState' in window.history` returns true, although we are not allowed to access `history.pushState` or `histroy.state`).

(BTW, this fix is not correct, I need to update it.)

Narretz: Ah, okay, I see

gkalpak:
I pushed a 3rd commit that prevents logging exceptions in the console. PTAL

src/ng/sniffer.js
```diff
@@ -61,7 +65,7 @@ function $SnifferProvider() {
       // so let's not use the history API also
       // We are purposefully using `!(android < 4)` to cover the case when `android` is undefined
       // jshint -W018
-      history: !!($window.history && $window.history.pushState && !(android < 4) && !boxee),
+      history: !!(hasHistoryPushState && !(android < 4) && !boxee),
```
petebacondarwin: why move the `$window.history....` checks out of this expression? Is Chrome more special that `android < 4` or `boxee`?

I would expect it is clearer to simply change this to:

```js
!!(!isChromePackagedApp && $window.history && $window.history.pushState && !(android < 4) && !boxee),
```

test/ng/snifferSpec.js
```diff
@@ -12,14 +11,84 @@ describe('$sniffer', function() {
...
     it('should be true if history.pushState defined', function() {
-      expect(sniffer({history: {pushState: noop, replaceState: noop}}).history).toBe(true);
+      var mockWindow = {
+        history: {
+          pushState: noop,
+          replaceState: noop
+        }
+      };
+
+      expect(sniffer(mockWindow).history).toBe(true);
```
petebacondarwin: this is just code style change right?

test/ng/snifferSpec.js
```diff
@@ -1,10 +1,9 @@
 'use strict';
 
 describe('$sniffer', function() {
-
   function sniffer($window, $document) {
     /* global $SnifferProvider: false */
-    $window.navigator = {};
+    $window.navigator = $window.navigator || {};
```
petebacondarwin: why do the tests now need this?

petebacondarwin:
A few minor comments but generally looks good to me.

test/ng/snifferSpec.js
gkalpak: Yes, there a first commit which just makes _"the structure and layout of tests more consistent"_: https://github.com/angular/angular.js/pull/13945/commits/566b31ce7e470cbafcf6dfee38280bce45f01501

test/ng/snifferSpec.js
gkalpak: I just "enhanced" the helper function, so it can be reused in more tests.

src/ng/sniffer.js
gkalpak: > Is Chrome more special that android < 4 or boxee

Yeah, Chrome is actually very close to my heart :stuck_out_tongue: 

I put it this way for readability (avoiding `!!(!` and hitting the 100 chars per line limit, thus having to wrap). 
I'm fine having it either way.

gkalpak: Plus, if I have `isChromePackagedApp` and `hasHistoryPushState` close together, it's easier for reference wrt to the comment explaining why Chrome Packaged Apps need special handling wrt to checking the existence of `history.pushState`.

petebacondarwin: OK, how about:

```
// Chrome Packaged Apps are not allowed to access `history.pushState`. They can be detected by
// the presence of `chrome.app.runtime` (see https://developer.chrome.com/apps/api_index)
isChromePackagedApp = $window.chrome && $window.chrome.app && $window.chrome.app.runtime,
hasHistoryPushState = $window.history && $window.history.pushState,
```

then later:

```
history: !isChromePackagedApp && !!hasHistoryPushState && !(android < 4) && !boxee,
```

gkalpak: The thing is that we shouldn't even try to touch `history.pushState` in packaged apps (or there'll be errors).
That's wh ywe need the `!isChromePackagedApp` short-circuiting at the beginning of `hasHistoryPushState`.

petebacondarwin: oh :-(

petebacondarwin:
LGTM - squash and merge

PR #13949, opened 2016-02-05
docs(guide/migration): document a BC to `ngAria` (d06431e)

gkalpak:


gkalpak:
The BC description is a little vague (and there is no guidelines on how to modify ones code), because I am not sure myself ([yey][1]) :smiley:

[1]: https://github.com/angular/angular.js/pull/13483#issuecomment-180243067

Narretz:
The way I understand it is that this should not affect any apps, because it corrects that aria behavior? It's just technically a BC because the aria-multiline support has been removed.

gkalpak:
I was confused, because [this MDN article][1] says you should put `aria-multiline` on `<textarea>`, but after further reading, I believe it's incorrect (possibly outdated).
There's a second part: Previously, the `aria-multiline` attribute was added to elements with a `type` or `role` of `textbox` (possibly custom multiline text input fields). I am not sure, if removing support for those was an oversight and if it was intentional then does it mean that the user should add it manually (or maybe that it's not possible to have a custom multilne text input (for reasons I am not aware of)).

[1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_textbox_role

gkalpak:
After talking with @marcysutton, it is indeed not needed to add `aria-multiline` on `<textarea` elements.
Also, as she correctly pointed out, there is no way to tell if a custom input with `[role="textbox"]` is multiline or not, so we shouldn't add `aria-multiline` there either.

I've update the PR. PTAL
/cc @marcysutton

docs/content/guide/migration.ngdoc
```diff
@@ -114,6 +114,27 @@ Objects considered array-like include: arrays, array subclasses, strings, NodeLi
...
+Due to [d06431e](https://github.com/angular/angular.js/commit/d06431e5309bb0125588877451dc79b935808134),
+the `ngAria`-enhanced directives (e.g. `ngModel`, `ngDisabled` etc) will not apply ARIA attributes
+to native inputs, unless necessary. Previously, ARIA attributes were always applied to native
+inputs, despite this being unnecessary in most cases.
+This change will not affect the accessibility of your applications (since native inputs are
+accessible by default), but if you relied on ARIA attributes being present on native inputs (for
+whatever reason), you'll have to add and update them manually.
+
+Additionally, the `aria-multiline` attribute, which was previously added to elements with a `type`
+or `role` on `textbox`, will not be added anymore, since there is no way `ngAria` can tell if the
```
Narretz: of `textbox`?

docs/content/guide/migration.ngdoc
```diff
@@ -114,6 +114,27 @@ Objects considered array-like include: arrays, array subclasses, strings, NodeLi
...
+textbox element is multiline or not.
+If you relied on `aria-multiline="true"` being automatically added by `ngAria`, you need to apply it
+yourself. E.g. change your code from `<div role="textbox" ng-model="..." ...>` to
+`<div role="textbox" ng-model="..." ... aria-multiline="true">`.
+
+In the context of `ngAria`, elements considered "native inputs" include:
```
Narretz: This could also go after the " ... being unnecessary in most cases."

Narretz:
One typo, another suggestion, otherwise LGTM

docs/content/guide/migration.ngdoc
gkalpak: Done

docs/content/guide/migration.ngdoc
gkalpak: Right :)

gkalpak:
Thx, @Narretz !
(Backported to `v1.5.x` as 0cc636b.)

PR #13952, opened 2016-02-05
improve even odd checking and avoid substr call where unnecessary

marijaselakovic:
I would propose two slight optimizations:

1) You should avoid call to substr if not necessary, so instead vendorPrefix.substr(0, 1), use vendorPrefix[0] which is cheaper and semantically equivalent. 

2) As explained in [https://github.com/angular/angular.js/pull/4359](4359), adapt even odd checking in ngClass

### DIFF
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -69,7 +69,7 @@ function classDirective(name, selector) {
         }
 
         function ngClassWatchAction(newVal) {
-          if (selector === true || scope.$index % 2 === selector) {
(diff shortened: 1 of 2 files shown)


marijaselakovic:
I signed it!

gkalpak:
There is a jsHint check failure (see https://travis-ci.org/angular/angular.js/jobs/107253414).
You need to disable jsHint's `bitwise` check for that line.

Narretz:
Landed as 7d056c1 and 18ec663

PR #13954, opened 2016-02-05
Adds new book to index.ngdoc booklist

jfmercer:
Professional AngularJS is, IME, one of the best books out there on Angular. It would make a nice addition to the booklist.


jfmercer:
I signed it!

PR #13958, opened 2016-02-06
refactor(equalsFunction): divide internal implemenation

### LINKED ISSUE #13957: Refactor complex angular equals function

m-amr:
extract complex logic to named methods
fixes #13957

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -977,44 +977,112 @@ function shallowCopy(src, dst) {
  * @returns {boolean} True if arguments are equal.
  */
 function equals(o1, o2) {
-  if (o1 === o2) return true;
-  if (o1 === null || o2 === null) return false;
-  if (o1 !== o1 && o2 !== o2) return true; // NaN === NaN
-  var t1 = typeof o1, t2 = typeof o2, length, key, keySet;
-  if (t1 == t2 && t1 == 'object') {
-    if (isArray(o1)) {
-      if (!isArray(o2)) return false;
-      if ((length = o1.length) == o2.length) {
-        for (key = 0; key < length; key++) {
-          if (!equals(o1[key], o2[key])) return false;
-        }
-        return true;
-      }
-    } else if (isDate(o1)) {
-      if (!isDate(o2)) return false;
-      return equals(o1.getTime(), o2.getTime());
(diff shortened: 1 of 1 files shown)


m-amr:
I don't know test fails here
I run tests on my local machine and get this.

Finished in 711.017 seconds
296 tests, 762 assertions, 0 failures

Shutting down selenium standalone server.
[launcher] 0 instance(s) of WebDriver still running
[launcher] chrome #1 passed

Done, without errors

gkalpak:
I don't feel refactoring everything out into functions makes it much more clear (if at all). 
Also, note this may have perf implications, considering `equals` is used in some critical paths.

lgalfaso:
Hi, thanks for the PR! Unfortunately I fully agree with @gkalpak and I find that this change does not make `angular.equals` any simpler to understand.

m-amr:
@gkalpak @lgalfaso I respect your decision 
 but the equals was a very complex function, with a lot of cases but now after
refactoring it just.

      function equals(o1, o2) {
          if (areBothEqualValue(o1, o2)) return true;
          if (isAnyOneHasValueEqualNull(o1, o2)) return false;
          if (areBothValuesEqualNaN(o1, o2)) return true;
          if (areBothObjectsEqual(o1, o2)) return true;
          return false;
      }

I didn't change any logic or and all cases are checked in the same order as before.
I refactored it by extracting functions with a meaningful names. to make it tell the story to reader
now it clear that we are doing the following steps
1 - check if their values are equal using areBothEqualValue(o1, o2) function
2 - check if one of them is null using AnyOneHasValueEqualNull(o1, o2) function
3 - check if their values equal NAN using areBothValuesEqualNaN(o1, o2) function
4 - check if two objects are using equal areBothObjectsEqual(o1, o2) function

I think using a meaningful function names make it easier to understand and better than adding
inline code comments that have the risk to be not updated with the code.
based on http://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882/ref=sr_1_1?ie=UTF8&qid=1454769207&sr=8-1&keywords=clean+code
chapter 4 (Comments ) 

I think it now easier to understand and easier to change too.

PR #13959, opened 2016-02-06
docs(guide/migration): minor english tweak

prayagverma:
Replace `a` with `an`


gkalpak:
Backported to `v1.5.x` as 12f0e07.

PR #13962, opened 2016-02-06
f[eat|ix](select): support values added by ngValue

### LINKED ISSUE #6297: ng-model boolean value binding behaviour between jQuery and jqLite (ng-value with option problem)
Hi,

We are using following code to bind ng-model on boolean value:
```
    <select ng-model="filter.active">
        <option value=""/>
        <option value="true">True selected</option>
        <option value="false">False selected</option>
    </select>
```

Here is a fiddle explaining different behaviour if jquery is present or is not present on page:
working : http://jsfiddle.net/8byx5/
broken with jquery : http://jsfiddle.net/8byx5/1/

What do you think about this behaviour ?

### LINKED ISSUE #9842: multiple {{}} interpolations inside <option> are causing value created by ngValue to be replaced by <option>'s text
See http://plnkr.co/edit/MXxLZ5wkVj4Cq2e5qQ4e?p=preview

##### works fine

```html
    <select ng-model="baz.val">
      <option 
        ng-value="foo.val" 
        ng-repeat="foo in bar">
        {{ foo.name + ' ' + foo.second_name }}
      </option>
    </select> 

```

##### ng-value replaced in this case

```html
    <select ng-model="baz.val">
      <option 
        ng-value="foo.val" 
        ng-repeat="foo in bar">
        {{ foo.name }} {{ foo.second_name }}
      </option>
    </select> 

```

also it's happening only if you use `ng-value`, it works fine when you use `value="{{ val }}"`, see https://github.com/angular/angular.js/issues/9840

Narretz:
select elements with ngModel will now set ngModel to option values added by ngValue.
This allows setting values of any type without the use of ngOptions.

Interpolations inside attributes can only be strings, but the ngValue directive uses attrs.$set,
which does not follow any type restriction. Any $observe on the value attribute will therefore receive
the original value (result of ngValue expression). However, when a user selects an option, the browser
sets the select value to the actual option's value attribute, which is still always a string.
For that reason, when option are added by ngValue, we set the hashed value of the original value in
the value attribute and store the actual value in an extra map. When the select value changes, we
read access the actual value via the hashed select value.

Closes #9842
Closes #6297
- [ ] Test that it takes precedence over interpolated option / option text
- [x] Test for multiple selections
- [x] Fix failing Chrome tests
- [ ] Docs

Talking points:
- The implementation works reasonably well, although it's a bit ~~hacky~~special
- With ngValue, option values are hash values, not the real values (which makes sense, but might be confusing)
- ~~afaik angular 2 suffers from the same problem as ng1 right now~~

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -20,6 +20,7 @@ var SelectController =
 
   // If the ngModel doesn't get provided then provide a dummy noop version to prevent errors
   self.ngModelCtrl = noopNgModelController;
+  self.multiple = false;
 
   // The "unknown" option is one that is prepended to the list if the viewValue
   // does not match any of the options. When it is rendered the value of the unknown
@@ -91,7 +92,9 @@ var SelectController =
     }
     var count = optionsMap.get(value) || 0;
     optionsMap.put(value, count + 1);
-    self.ngModelCtrl.$render();
+    // Only render at the end of a digest. This improves render performance when many options
+    // are added during a digest and ensures all relevant options are correctly marked as selected
+    scheduleRender();
   };
 
   // Tell the select control that an option, with the given value, has been removed
@@ -115,6 +118,15 @@ var SelectController =
   };
 
 
+  var renderScheduled = false;
+  function scheduleRender() {
+    if (renderScheduled) return;
+    renderScheduled = true;
+    $scope.$$postDigest(function() {
+      renderScheduled = false;
+      self.ngModelCtrl.$render();
+    });
+  }
 
   var updateScheduled = false;
   function scheduleViewValueUpdate(renderAfter) {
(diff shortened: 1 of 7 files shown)


test/ng/directive/selectSpec.js
```diff
@@ -1224,5 +1236,202 @@ describe('select', function() {
...
+            expect(scope.selected).toBe(prop);
+          }
+
+          expect(element.find('option').eq(0).prop('selected')).toBe(true);
+          expect(element.find('option').eq(1).val()).toBe('string:UPDATEDVALUE');
+
+          scope.selected = 'UPDATEDVALUE';
+          scope.$digest();
+
+          // expect(element.find('option').eq(0).prop('selected')).toBe(true); not selected in Chrome?
```
Narretz: I wanted to fix this, but I'm not sure it's necessary. I don't think browsers have to set *selected* when the value of the select changes. It's strange though that is set before changing the select value via model update.

src/ng/directive/select.js
```diff
@@ -116,9 +120,26 @@ var SelectController =
 
   self.registerOption = function(optionScope, optionElement, optionAttrs, interpolateValueFn, interpolateTextFn) {
 
-    if (interpolateValueFn) {
+    if (interpolateValueFn === true) {
```
petebacondarwin: I am not keen on using `true` as a special flag here.
Perhaps we could just add yet another parameter to the `registerOption` method?

petebacondarwin: Or perhaps we don't need to do that at all, we could just check whether `optionAttrs.ngValue` is defined?

petebacondarwin:
This test fails:

```js
      it('should work update the value as a result of changes to the options', function() {
        var A = { name: 'A'}, B = { name: 'B'}, C = { name: 'C'};
        scope.options = [A, B, C];
        scope.obj = {};
        compile(
          '<select ng-model="obj.value">' +
            '<option ng-repeat="option in options" ng-value="option">{{$index}}</option>' +
          '</select>'
        );

        var optionElements = element.find('option');
        expect(optionElements.length).toEqual(4);
        browserTrigger(optionElements.eq(0));

        optionElements = element.find('option');
        expect(optionElements.length).toEqual(3);
        expect(scope.obj.value).toBe(A);

        scope.options.shift();
        scope.$digest();

        optionElements = element.find('option');
        expect(optionElements.length).toEqual(3);
        expect(scope.obj.value).toBe(null);
      });
```

Narretz:
Have you tried if that works with regular interpolated options? Because I don't think it does

petebacondarwin:
It certainly works with `ngOptions`, which is what we are trying to emulate, right?

Narretz:
@petebacondarwin I've updated the PR again with the changes we talked about (support for disabled options, and model updates when the options change / are removed / get disabled / get added).

src/ng/directive/select.js
```diff
@@ -113,36 +131,128 @@ var SelectController =
...
+  function updateModelAfterOptionChange(renderAfter) {
+    if (self.multiple) {
+      if (!handleMultipleChanges) {
+        handleMultipleChanges = true;
+      } else {
+        $scope.$$postDigest(function() {
+          handleMultipleChanges = false;
+          self.ngModelCtrl.$setViewValue(self.readValue());
+          if (renderAfter) self.ngModelCtrl.$render();
+        });
```
petebacondarwin: This pattern of running something once in the postDigest keeps appearing. I wonder if we should abstract it out into $scope?

petebacondarwin:
I think you might have nailed it but it is the kind of change that I want to spend some time thinking about corner cases...

petebacondarwin:
I did author some of these commits and I am OK with that :-)

src/ng/directive/select.js
```diff
@@ -14,10 +14,15 @@ var SelectController =
         ['$element', '$scope', function($element, $scope) {
 
   var self = this,
-      optionsMap = new HashMap();
+      optionsMap = new HashMap(),
+      handleMultipleDestroy = false; // Flag to run an update to the model after selected options
```
petebacondarwin: no longer needed?

Narretz:
I'll change the tests and then add some docs. Should I mentiond that performance in IE is going to be very bad with 3000+ options?

petebacondarwin:
Why not

Narretz:
While fixing up the PR, I found a behavior in ngOptions and select that I find a bit confusing: You have a that select is multiple and the model contains two values, one of which is matched by an option, while the other isn't.
What happens is that the matching option gets selected, and the model keeps both values. Imo, I would expect no option to be selected, as the model isn't completely matched by the select options. Thoughts?

Narretz:
So in https://github.com/angular/angular.js/issues/13172 we decided that if *no* option at all can be selected, then the select should be invalid (required). 

Doesn't say anything about a partial mismatch, though. We could decide not to select anything when not all options are matched, but I'm not convinced that this is a good idea.

Theoretically, setting the model to match the available options is the developers responsibility, so if that happens it's your own fault. However, it's not very easy to check if a select's options will match the model. We could expose the `selectCtrl.hasOption` function and make it work with ngOptions, so that users could manually check if a specific $viewValue is matched by their select, see (https://github.com/angular/angular.js/issues/12915)

Narretz:
I'm also fine with making a decision on this later. As far as I can tell, the described behavior hasn't bothered anyone yet.
@petebacondarwin I need your perf commit as part of the commit that handles multiple selects correctly. Because otherwise, the current behavior doesn't work as expected. If the model has two values set, but only one is matched by options, then without the "perf" fix, the option will set the model to the one value only. Delaying the update after render handles this.

gkalpak:
:sparkles: :tada: :sparkles:

PR #13964, opened 2016-02-07
docs(error/nobase): tag base wasn't displayed properly

saamed:
Replaced < with &lt and > with &gt;

### DIFF
--- a/docs/content/error/$location/nobase.ngdoc
+++ b/docs/content/error/$location/nobase.ngdoc
@@ -4,7 +4,7 @@
 @description
 
 If you configure {@link ng.$location `$location`} to use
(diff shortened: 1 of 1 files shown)


saamed:
I signed it!

petebacondarwin:
It looks like it is rendering properly for me...

screen shot 2016-02-08 at 02 15 25

You should not need to escape HTML characters in code blocks.

Narretz:
@petebacondarwin In the code block above the Description, there's a gap where `<base>` should be. I tried to fix this before. Let me see if I still have a branch with a fix.

petebacondarwin:
OK but that is not the problem that this PR is trying to fix

Narretz:
See https://github.com/angular/angular.js/pull/14033

PR #13970, opened 2016-02-07
docs(guide/component): add replace attribute to comparison

Puigcerber:
Add replace attribute to comparison between Directive definition and Component definition.


petebacondarwin:
Thanks for the contribution @Puigcerber 
Replace is deprecated. We wish it wasn't there. We don't want to advertise it in the documentation.
So I'm afraid that we not going to merge this PR.

Puigcerber:
No problem! I thought that might be the reason but still it was weird not to see it in the comparison. Cheers.

PR #13975, opened 2016-02-08
docs(guide/Unit Testing): add info about promises

Narretz:
Related #1915

### DIFF
--- a/docs/content/guide/unit-testing.ngdoc
+++ b/docs/content/guide/unit-testing.ngdoc
@@ -430,5 +430,13 @@ If your directive uses `templateUrl`, consider using
 to pre-compile HTML templates and thus avoid having to load them over HTTP during test execution.
 Otherwise you may run into issues if the test directory hierarchy differs from the application's.
 
+## Testing Promises
+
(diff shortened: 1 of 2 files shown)


docs/content/guide/unit-testing.ngdoc
```diff
@@ -430,5 +430,12 @@ If your directive uses `templateUrl`, consider using
 to pre-compile HTML templates and thus avoid having to load them over HTTP during test execution.
 Otherwise you may run into issues if the test directory hierarchy differs from the application's.
 
+## Testing Promises
+
+When testing promises, it's important to know that the resolution of promises is tied to the {@link ng.$rootScope.Scope#$digest digest cycle}.
+That means a promise's `then`, `catch` and `finally` callback functions are only called after a digest has run.
+In tests, you can trigger a digest by calling a scope's {@link ng.$rootScope.Scope#$apply `$apply`function}.
```
gkalpak: Missing whitespace before "function".

gkalpak: Since many tests don't need a scope, I would say that the way to trigger a digest is by means of `$rootScope.$digest()`. I'm fine mentioning both was (`$rootScope.$digest()` and `someScope.$apply()`), but I would definitely mention the former.

PR #13977, opened 2016-02-08
refact($componentController): don't return an object from the provide…

### LINKED ISSUE #13969: ngMock fails at $componentController after update from 1.4.8 to 1.5.0
i have no idea whats up?
```
Error: [$injector:modulerr] Failed to instantiate module ngMock due to:
Error: [$injector:pget] Provider '$componentController' must define $get factory method.
```

FYI: tests are also working in 1.4.9

after some digging:
https://github.com/angular/angular.js/blob/v1.5.0/src/ngMock/angular-mocks.js#L2186
shouldn't this read `this.$get = ....`? after fixing that in the dist file, the tests are working

also, to clarify, I am running tests on module, which only consists of services and factories

petebacondarwin:
…r constructor function

Closes #13969

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2182,29 +2182,27 @@ angular.mock.$ControllerDecorator = ['$delegate', function($delegate) {
  * @return {Object} Instance of requested controller.
  */
 angular.mock.$ComponentControllerProvider = ['$compileProvider', function($compileProvider) {
-  return {
-    $get: ['$controller','$injector', function($controller,$injector) {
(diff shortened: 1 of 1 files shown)


Narretz:
Should there be some sort of test for this?

gkalpak:
It's just a refactor, it works either way (so the fact that the previous tests are passing is all we need).
(BTW, I believe the correct `type` in the commit message is `refactor` (not `refact`) :smiley:)

LGTM

petebacondarwin:
I fixed the commit message

Issue #6266, opened 2014-02-14
ngRepeat: property names starting with "$" will not be rendered


mgmeyers:
When using ngRepeat to iterate over an object, it will not output any key-value pair where the key starts with a dollar sign ($).

Here is a demo:
http://jsfiddle.net/takvg/4/

And, the actual code in question can be found here: https://github.com/angular/angular.js/blob/master/src/ng/directive/ngRepeat.js#L284

``` javascript
if (isArrayLike(collection)) {
  collectionKeys = collection;
  trackByIdFn = trackByIdExpFn || trackByIdArrayFn;
} else {
  trackByIdFn = trackByIdExpFn || trackByIdObjFn;
  // if object, extract keys, sort them and use to determine order of iteration over obj props
  collectionKeys = [];
  for (key in collection) {
    if (collection.hasOwnProperty(key) && key.charAt(0) != '$') {
      collectionKeys.push(key);
    }
  }
  collectionKeys.sort();
}
```

Specifically:

``` javascript
key.charAt(0) != '$'
```

I'm assuming this is to prevent angular's internal properties from being iterated over by ngRepeat. I would suggest removing this limitation, but I don't know enough about the internals or the reasoning behind this to know what effect that might have.

sunnypatel:
Has anyone found a workaround to this?

frfancha:
@sunnypatel Workaround is easy: just make your ng-repeat on Object.keys(myObject)

Dhumez-Sebastien:
@sunnypatel here is example of @frfancha : https://jsfiddle.net/takvg/5/

startswithaj:
+1

uriva:
+1

gaurav5430:
is there a workaround for key-value pairs , where one of the keys is $ , and we want to display both key and value. 

(in the example fiddle provided by Dhumez, we are only displaying keys)

augimidge:
@gaurav5430 here is a workaround for key-value pairs.
https://jsfiddle.net/takvg/8/

Narretz:
This would require a change in angular.equals, which could possibly affect many apps negatively at this point of the AngularJS lifecycle. Closing as wontfix

Padge91:
In case anyone else stumbles upon this issue, this is my workaround:

I just added a null character at the front of the String value used for the key...
`if (keyName.length > 0 && keyName[0]==="$"){
keyName = keyName.replace("$", "\0$");
}`

Seems to work just fine.

PR #13978, opened 2016-02-08
 	docs(ngRepeat): mention limitations of object iteration

Narretz:
Issue is tracked in issue Related #6266

### DIFF
--- a/src/ng/directive/ngRepeat.js
+++ b/src/ng/directive/ngRepeat.js
@@ -36,17 +36,23 @@
  * <div ng-repeat="(key, value) in myObj"> ... </div>
  * ```
  *
- * You need to be aware that the JavaScript specification does not define the order of keys
- * returned for an object. (To mitigate this in Angular 1.3 the `ngRepeat` directive
- * used to sort the keys alphabetically.)
+ * However, there are a limitations compared to array iteration:
  *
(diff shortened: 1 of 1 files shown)


src/ng/directive/ngRepeat.js
```diff
@@ -36,17 +36,23 @@
...
+ * - `ngRepeat` will silently *ignore* object keys starting with `$`, because
+ *   it's a prefix used by Angular for public (`$`) and private (`$$`) properties.
+ *
+ * - The built-in filters {@link ng.orderBy orderBy} and {@link ng.filter filter} do not work with
+ *   objects, and will throw if used with one.
+ *
+ * If you are hitting any of these limitations, the recommended workaround is to convert your object into an array
+ * that is sorted into the order that you prefer before providing it to `ngRepeat`. You could
  * do this with a filter such as [toArrayFilter](http://ngmodules.org/modules/angular-toArrayFilter)
  * or implement a `$watch` on the object yourself.
```
gkalpak: I wonder what it means to _"implement a `$watch` on the object yourself"_. Any idea ?

gkalpak:
LGTM

src/ng/directive/ngRepeat.js
Narretz: Maybe a watch that detects an object and then transforms it into an array?

PR #13979, opened 2016-02-08
 	docs(*): clarify limitations of app bootstrapping 

Narretz:


docs/content/guide/bootstrap.ngdoc
```diff
@@ -142,6 +142,17 @@ This is the sequence that your code should follow:
   2. Call {@link angular.bootstrap} to {@link compiler compile} the element into an
   executable, bi-directionally bound application.
 
+## Things to keep in mind
+
+There a few things to keep in mind regardless of automatic or manual bootstrapping:
+
+- There is no explicit support for bootstrapping more than one AngularJS applications on a page, and
+  doing this may cause bugs and/or memory leaks.
+- Likewise, apps cannot be nested within each other.
```
gkalpak: I am not sure what is ment by "nested", but you can have something like:

```html
<div app-1-root>
  <!-- Inside app-1 -->
  <div ng-non-bindable>
    <div app-2-root>
      <!-- Inside app-2 -->
    </div>
  </div>
  <!-- Inside app-1 -->
</div>
```

docs/content/guide/bootstrap.ngdoc
```diff
@@ -142,6 +142,17 @@ This is the sequence that your code should follow:
   2. Call {@link angular.bootstrap} to {@link compiler compile} the element into an
   executable, bi-directionally bound application.
 
+## Things to keep in mind
+
+There a few things to keep in mind regardless of automatic or manual bootstrapping:
+
+- There is no explicit support for bootstrapping more than one AngularJS applications on a page, and
```
gkalpak: Unless you have proof, I don't believe it :stuck_out_tongue:

docs/content/guide/bootstrap.ngdoc
Narretz: Oh, interesting. I copied this from ngApp, so it apparently only applies to ngApp, not manual bootstrapping

src/Angular.js
```diff
@@ -1545,16 +1552,23 @@ function angularInit(element, bootstrap) {
...
- * Note that Protractor based end-to-end tests cannot use this function to bootstrap manually.
- * They must use {@link ng.directive:ngApp ngApp}.
+ * For more information, see the {@link guide/bootstrap Bootstrap guide}.
  *
  * Angular will detect if it has been loaded into the browser more than once and only allow the
  * first loaded script to be bootstrapped and will report a warning to the browser console for
  * each of the subsequent scripts. This prevents strange results in applications, where otherwise
  * multiple instances of Angular try to work on the DOM.
  *
+ * Note that Protractor based end-to-end tests cannot use this function to bootstrap manually.
```
gkalpak: This could as well be put into a warning alert imo.

docs/content/guide/bootstrap.ngdoc
gkalpak: OK, so the fact that we don't run any tests for multiple apps per page (and how could we since Protractor doesn't support manual bootstraping :smiley:), doesn't leave as the option to say it is supported. But I find this statement a little too discouraging.

Also, the [`ngApp` docs][1] say:

> To run multiple applications in an HTML document you must manually bootstrap them using angular.bootstrap instead.

Which gives the impression, it **is** possible/supported to run multiple apps on a page.

--
Maybe this could be re-phrased to something more into the direction: _"It might work, but it's untested, there's shared resources and you might run into prolems (especially with more complex apps), so use at your own risk"_

I don't feel too strongly about it - feel free to ignore this.

[1]: https://docs.angularjs.org/api/ng/directive/ngApp

docs/content/guide/bootstrap.ngdoc
gkalpak: Yeah, you can only auto-bootstrap one app. You need to manaully bootstrap the rest.
I haven't used it with any complex nested app, but simple have worked without problems.

But still, it's untested :disappointed:

Narretz:
thx @gkalpak I updated the wording for multiple apps per page.

PR #13988, opened 2016-02-09
chore(build): ensure that version pattern is matched accordingly

matsko:


Narretz:
I assume this fixes a bug with the g3 sync?

petebacondarwin:
@matsko - I think this is now out of date right?

PR #13993, opened 2016-02-10
decodeURIComponent - prevent URIError

modmac:
Replace decodeURIComponent with tryDecodeURIComponent to prevent throwing "URIError: malformed URI sequence".

https://github.com/angular/angular.js/blob/master/src/Angular.js#L1275

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -37,10 +37,10 @@ function parseAppUrl(relativeUrl, locationObj) {
     relativeUrl = '/' + relativeUrl;
   }
   var match = urlResolve(relativeUrl);
-  locationObj.$$path = decodeURIComponent(prefixed && match.pathname.charAt(0) === '/' ?
+  locationObj.$$path = tryDecodeURIComponent(prefixed && match.pathname.charAt(0) === '/' ?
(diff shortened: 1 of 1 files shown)


mod3st:
I signed it!

2016-02-10 15:34 GMT+01:00 googlebot <notifications@github.com>:

gkalpak:
I think this will break the contract that `$location.path()` or `$location.hash()` return a string.
OOC, under what circumstances can this error been thrown ? Do you have a reproduction.

There should also be some tests, verifying the new behavior and guarding against regressions.

mod3st:
Yes, try to request a page with a #% hash in the URL.

Example:
https://console.developers.google.com/freetrial?newProject=false#%

Maybe in this example, if decodeURIComponent throws an error, we should catch it and encode hash/uri to keep wrong characters (ex. single %) as encoded UTF?

gkalpak:
Aha, I see the error.

TBH, trying to guess and fix unencoded UTF characters is a business I'd rather not be into :)
That said, silently failing on invalid URLs, won't be fun to debug. I think it's better to throw here.

@petebacondarwin, wdyt ?

PR #13996, opened 2016-02-10
feat(ngAria[ngClick]): anchor should treat keypress as click

### LINKED ISSUE #11053: ngAria should make ng-click as accessible for anchor tags as it does for DIV/LI
Changes for #10388 got rid of the extra events for `BUTTON` and `A` tags, but I think more could be done to make the accessibility more consistent.  The browser already triggers the click event for `A` tags on `enter` if the `href` attribute is set, but never triggers the click event on `space`.  

I think ngAria should hook keypress and trigger ng-click bindings on `space`, and trigger on `enter` if no `href` attribute is set.

This would result in ng-click behaving the same on `BUTTON` (implemented by the browser), `A` (browser + this change), `DIV`, and `LI` - triggering on click, `enter`, or `space`.

LeeAdcock:
Enhances ngAria to trigger ngClick for 'space' keypresses on anchor and button elements, and
for 'enter' keypresses on anchor elements when there is no value for the href attribute.

Closes https://github.com/angular/angular.js/issues/11053

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -385,9 +385,9 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
           if ($aria.config('bindKeypress')) {
             elem.on('keypress', function(event) {
               var keyCode = event.which || event.keyCode;
-              var hasHref = (attr.href || attr.xlinkHref);
-              if ((keyCode === 32 || (keyCode === 13 && !hasHref)) && !attr.ngKeypress) {
-                elem[0].click();
+              var hasHref = attr.href || attr.xlinkHref;
+              if ((keyCode===13 || keyCode === 32) && !hasHref && !attr.ngKeypress) {
+                scope.$apply(callback);
               }
 
               function callback() {
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -377,6 +377,28 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             });
           }
         }
+
+        if ($aria.config('bindKeypress') && isNodeOneOf(elem, ['A', 'BUTTON'])) {
+          elem.on('keypress', function(event) {
+            var keyCode = event.which || event.keyCode;
+
+            if (elem[0].nodeName==='A') {
+              var hasHref = elem.attr('href') != null && elem.attr('href') != '';
+              if ((keyCode === 13 && !hasHref) || keyCode === 32) {
+                scope.$apply(callback);
+              }
+            } 
+            else if (elem[0].nodeName==='BUTTON') {
+              if (keyCode === 32) {
+                scope.$apply(callback);
+              }
+            }
+            
+            function callback() {
+              fn(scope, { $event: event });
+            }
+          });
+        }
       };
     }
   };
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -386,7 +386,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             elem.on('keypress', function(event) {
               var keyCode = event.which || event.keyCode;
               var hasHref = attr.href || attr.xlinkHref;
-              if ((keyCode===13 || keyCode === 32) && !hasHref && !attr.ngKeypress) {
+              if ((keyCode === 32 || keyCode === 13) && !hasHref && !attr.ngKeypress) {
                 scope.$apply(callback);
               }
 
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -378,25 +378,23 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
           }
         }
 
-        if ($aria.config('bindKeypress') && isNodeOneOf(elem, ['A', 'BUTTON'])) {
-          elem.on('keypress', function(event) {
-            var keyCode = event.which || event.keyCode;
-
-            if (elem[0].nodeName === 'A') {
-              var hasHref = elem.attr('href') != null && elem.attr('href') !== '';
-              if ((keyCode === 13 && !hasHref) || keyCode === 32) {
+        if (elem[0].nodeName === 'A') {
+          if ($aria.config('bindRoleForClick') && !attr.href && !attr.xlinkHref && !attr.role) {
+              elem.attr('role', 'link');
+          }
+          if($aria.config('bindKeypress')) {
+            elem.on('keypress', function(event) {
+              var keyCode = event.which || event.keyCode;
+              var hasHref = (attr.href || attr.xlinkHref);
+              if (keyCode === 32 && !hasHref && !attr.ngKeypress) {
                 scope.$apply(callback);
               }
-            } else if (elem[0].nodeName === 'BUTTON') {
-              if (keyCode === 32) {
-                scope.$apply(callback);
-              }
-            }
 
-            function callback() {
-              fn(scope, { $event: event });
-            }
-          });
+              function callback() {
+                fn(scope, { $event: event });
+              }
+            });
+          }
         }
       };
     }
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -386,7 +386,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             elem.on('keypress', function(event) {
               var keyCode = event.which || event.keyCode;
               var hasHref = (attr.href || attr.xlinkHref);
-              if (keyCode === 32 && !hasHref && !attr.ngKeypress) {
+              if ((keyCode === 32 || (keyCode === 13 && !hasHref)) && !attr.ngKeypress) {
                 scope.$apply(callback);
               }
 
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -382,18 +382,17 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
           elem.on('keypress', function(event) {
             var keyCode = event.which || event.keyCode;
 
-            if (elem[0].nodeName==='A') {
-              var hasHref = elem.attr('href') != null && elem.attr('href') != '';
+            if (elem[0].nodeName === 'A') {
+              var hasHref = elem.attr('href') != null && elem.attr('href') !== '';
               if ((keyCode === 13 && !hasHref) || keyCode === 32) {
                 scope.$apply(callback);
               }
-            } 
-            else if (elem[0].nodeName==='BUTTON') {
+            } else if (elem[0].nodeName === 'BUTTON') {
               if (keyCode === 32) {
                 scope.$apply(callback);
               }
             }
-            
+
             function callback() {
               fn(scope, { $event: event });
             }
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -387,7 +387,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
               var keyCode = event.which || event.keyCode;
               var hasHref = (attr.href || attr.xlinkHref);
               if ((keyCode === 32 || (keyCode === 13 && !hasHref)) && !attr.ngKeypress) {
-                scope.$apply(callback);
+                elem[0].click();
               }
 
               function callback() {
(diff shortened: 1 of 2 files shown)


src/ngAria/aria.js
```diff
@@ -377,6 +377,27 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             });
           }
         }
+
+        if ($aria.config('bindKeypress') && isNodeOneOf(elem, ['A', 'BUTTON'])) {
```
gkalpak: Actually, we don't want to do anything with `button`s, just `a`s. The browser handles pressing both ENTER and SPACE on buttons.

gkalpak: You should also check that `attr.ngKeypress` is falsy.

src/ngAria/aria.js
```diff
@@ -377,6 +377,27 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
+          elem.on('keypress', function(event) {
+            var keyCode = event.which || event.keyCode;
+
+            if (elem[0].nodeName === 'A') {
+              var hasHref = elem.attr('href') != null && elem.attr('href') !== '';
```
gkalpak: You just have to check if `elem.attr('href')` is falsy.

BTW, if we need to also support SVG, then we need to check `xlink:href` (instead of `href`).
(You can take a look at the `htmlAnchorDirective` in the source code for more details.)

src/ngAria/aria.js
```diff
@@ -377,6 +377,27 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
+              if ((keyCode === 13 && !hasHref) || keyCode === 32) {
```
gkalpak: The browser handles ENTER for anchor. We just need to handle SPACE (and I would still check that it has no href).

gkalpak:
I left some comments (that should be easy to address). Additonally, according to https://github.com/angular/angular.js/issues/11053#issuecomment-87104961, we should also bind the `role` to `link` on `<a>` elements (if the relevant setting in `$aria` is enabled).

@LeeAdcock, feel free to ping when/if you've made any changes, so I can take another look. Thx !

src/ngAria/aria.js
LeeAdcock: The ENTER doesn't seem to be handled for anchors when href is falsy. I think we need to trigger the click when ENTER and no href. The SPACE is not handled regardless of the href.

LeeAdcock:
Thank you for your feedback. I'm made the suggested changes, but I believe there's still a little bit of work to go.  

- In the case that href (and xlink:href) is falsy the browser doesn't react to ENTER key presses. I believe we need to be watching for these in the keypress binding.
- The SPACE doesn't have any impact on links where href is defined. I believe we do need to be watching for SPACE key presses regardless of the href.
- I'm not sure executing the ng-click is the best way to provide full accessibility. Even when watching for SPACE key presses for anchors with hrefs, the ENTER is interpreted as a click and follows the link, while SPACE only executes ng-click. It would seem both should be a click and follow the link.
- This isn't the documented use of bindRoleForClick and bindKeypress, I assume I should update the documentation to include details about anchor handling, versus creating new config flags.

Please let me know which direction you'd like me to go.

LeeAdcock:
I continued to iterate on these changes, and believe the PR now aligns with the intent of the issue and provides what I would have expected from ngAria for enhanced accessibility for links. This results in the same behavior for mouse clicks, the space key, and the enter key. The ```role='link'``` attribute is now also defined when needed.

I would suggest that the previous behavior might be more than an inconvenience.  We have written our own custom directives to work around these shortcomings, since the current Angular behavior fails accessibility standards.

Feedback welcome, @gkalpak.

src/ngAria/aria.js
```diff
@@ -377,6 +377,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
+          }
+          if ($aria.config('bindKeypress')) {
+            elem.on('keypress', function(event) {
+              var keyCode = event.which || event.keyCode;
+              var hasHref = (attr.href || attr.xlinkHref);
+              if ((keyCode === 32 || (keyCode === 13 && !hasHref)) && !attr.ngKeypress) {
+                elem[0].click();
+              }
+
+              function callback() {
```
campersau: It looks like `callback` is never used inside this code path.

src/ngAria/aria.js
```diff
@@ -377,6 +377,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             });
           }
         }
+
+        if (elem[0].nodeName === 'A') {
+          if ($aria.config('bindRoleForClick') && !attr.href && !attr.xlinkHref && !attr.role) {
+              elem.attr('role', 'link');
+          }
+          if ($aria.config('bindKeypress')) {
+            elem.on('keypress', function(event) {
```
marcysutton: I think we'd be better off with `keydown` here (and elsewhere, actually)

marcysutton:
Thanks for working on this! Keep in mind the expected keyboard interaction for links is ENTER, not SPACE. That is more of a typical button interaction.

src/ngAria/aria.js
gkalpak: Indeed switching to `keydown` seems more consistent with how native buttons work.
Yet, I suggest we make the change from `keypress` to `keydowm` in another PR (since it's used to several places) and possibly tests.

Also, does it mean that `bindKeypress` should be renames `bindKeydown` (which is a breaking change) ?
Leaving the name `bindKeypress` but binding to `keydown` isn't ideal either :smiley:
Maybe we could deprecate `bindKeypress` in favor of `bindKeydown` and remove it in `v1.6.x`...

gkalpak: I opened #14063 to track this.

src/ngAria/aria.js
```diff
@@ -377,6 +377,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             });
           }
         }
+
+        if (elem[0].nodeName === 'A') {
+          if ($aria.config('bindRoleForClick') && !attr.href && !attr.xlinkHref && !attr.role) {
+              elem.attr('role', 'link');
+          }
+          if ($aria.config('bindKeypress')) {
```
gkalpak: To avoid unnecessary work, we should check for `!attr.ngKeypress` here (instead of inside the listener below).

src/ngAria/aria.js
```diff
@@ -377,6 +377,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
+            elem.on('keypress', function(event) {
+              var keyCode = event.which || event.keyCode;
+              var hasHref = (attr.href || attr.xlinkHref);
```
gkalpak: This is not exactly what we are doing in the built-in `htmlAnchorDirective` (in core) and there's a very slight chance this will not do the expected thing. This bug is so unlikely to _bite_ anyone though, that I'm fine leaving it as is (for the sake of simplicity).

src/ngAria/aria.js
```diff
@@ -377,6 +377,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
+              if ((keyCode === 32 || (keyCode === 13 && !hasHref)) && !attr.ngKeypress) {
```
gkalpak: I don't think we should do anything (regardless the button clicked), if `hasHref` is truthy. What we are trying to do here is to "fix" accessibility on apps that (for whatever reason) are trying to use `<a>` elements as buttons (in which case they shouldn't provide an `href` value).

If someone applies both `href` and `ngClick`, I'm afraid they're on their own. Besides, different browsers handle the situation differently (e.g. IIRC Chrome first changes the URL and then executes the click handler and Firefox does the opposite (or it could be the other way around)).

`ngAria` should not be into that business. All we should be concerned with here is making `<a ng-click="...">...` behave like a "real" button wrt keyboard input.

src/ngAria/aria.js
```diff
@@ -377,6 +377,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
+                elem[0].click();
```
gkalpak: I'm torn about this. I think I slightly lean towards keeping things consistent among all non-button elements (e.g. `<div>` or `<li>`). Not sure...

gkalpak: @LeeAdcock, what did you have in mind about this change ? In what way is this preferrable to just calling `fn` ?
BTW, if this ends up sticking around, then the callback below should be removed (as mentioned by @campersau).

gkalpak:
> Keep in mind the expected keyboard interaction for links is ENTER, not SPACE. That is more of a typical button interaction.

But native buttons do respond to SPACE too. So, our "custom button band-aid" should handle that as well, right ?

src/ngAria/aria.js
LeeAdcock: I was aiming to make the behavior the same for ```<a>``` elements across the varying types of input. So regardless if the user clicks with the mouse, or uses the keyboard or screen reader to press SPACE or ENTER, the behavior should be the same. If someone is using ```fn``` on anchor tags with the ```hrefs``` defined, then the two keys behave differently - one follows the link, while the other executes the ```ng-click``` and stops. Using ```.click``` resolves this inconsistency.  I guess I leave it to you to tell me if this is a scenario we are not concerned with. 

Perhaps as another feature enhancement, ngAria should be applying behavior to ```ng-href``` to ensure anchor keyboard input is consistent for SPACE and ENTER.

gkalpak: Taking https://github.com/angular/angular.js/pull/13996#discussion_r53084886 into consideration, we should only concern ourselves with `<a>` without `href`/`xlinkHref`, which would remove the ambiguity.

And then, the only difference would be that `fn(...)` would not trigger other `click` handlers on the element, while `.click()` would. And since (1) the user has already specified `ngClick` on the element and (2) it's good to be consistent with other non-button elements (e.g. `<div>` or `<li>`), I think it's better to use `fn(...)`.

src/ngAria/aria.js
marcysutton: Thanks @gkalpak, that can definitely happen in a separate PR.

src/ngAria/aria.js
```diff
@@ -377,6 +377,25 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
...
           }
         }
+
+        if (elem[0].nodeName === 'A') {
+          if ($aria.config('bindRoleForClick') && !attr.href && !attr.xlinkHref && !attr.role) {
+              elem.attr('role', 'link');
+          }
+          if ($aria.config('bindKeypress')) {
+            elem.on('keypress', function(event) {
+              var keyCode = event.which || event.keyCode;
```
marcysutton: This section looks an awful lot like https://github.com/LeeAdcock/angular.js/blob/ngclick-accessible-11053/src/ngAria/aria.js#L367 above. I wonder if we could combine those two sections somehow, perhaps with a reusable function?

test/ngAria/ariaSpec.js
```diff
@@ -755,6 +755,38 @@ describe('$aria', function() {
...
+    it('should not bind keypress to anchor elements when keypress is not falsy', function() {
+      compileElement('<a ng-keypress="event = null" ngclick="event = $event">{{event.type}}{{event.keyCode}}</a>');
+      expect(element.text()).toBe('');
+      element.triggerHandler({ type: 'keypress', keyCode: 32 });
+      expect(element.text()).toBe('');
+      element.triggerHandler({ type: 'keypress', keyCode: 13 });
+      expect(element.text()).toBe('');
+    });
+
+    it('should not bind keypress to anchor elements when href is not falsy', function() {
```
marcysutton: Do people really use `ng-href` and `ng-click` together? :eyes:

test/ngAria/ariaSpec.js
```diff
@@ -755,6 +755,38 @@ describe('$aria', function() {
       element.triggerHandler({ type: 'keypress', keyCode: 13 });
       expect(element.text()).toBe('');
     });
+
+    it('should apply role to anchor elements without href ', function() {
+      compileElement('<a ng-click="event = $event">{{event.type}}{{event.keyCode}}</a>');
```
marcysutton: Why wouldn't you use a button element for this? I don't understand the use case.

PR #14001, opened 2016-02-10
checking if request is made from file protocol avoid reading XSRF cookie

MoatazAbdallh:
Checking if request is made from file protocol avoid reading XSRF cookie which cause issue with OS which serve from file protocol as WebOS.

### DIFF
--- a/src/ng/cookieReader.js
+++ b/src/ng/cookieReader.js
@@ -22,7 +22,7 @@ function $$CookieReader($document) {
       return str;
     }
   }
-  //Checking if cookie is supported as there are some old browsers & some OS not support cookie.
+  //Checking if cookie is supported as there are some old browsers & some OS as WebOS not support cookie.
   try {
     rawDocument.cookie;
   }
@@ -31,7 +31,7 @@ function $$CookieReader($document) {
   }
   return function() {
     var cookieArray, cookie, i, index, name;
-    var currentCookieString = cookieSupported ? rawDocument.cookie || '' : '';
+    var currentCookieString = (cookieSupported && rawDocument.cookie) || '';
 
     if (currentCookieString !== lastCookieString) {
       lastCookieString = currentCookieString;
--- a/src/ng/cookieReader.js
+++ b/src/ng/cookieReader.js
@@ -13,6 +13,7 @@ function $$CookieReader($document) {
   var rawDocument = $document[0] || {};
   var lastCookies = {};
   var lastCookieString = '';
+  var cookieSupported = true;
 
   function safeDecodeURIComponent(str) {
     try {
@@ -21,10 +22,16 @@ function $$CookieReader($document) {
       return str;
     }
   }
-
+  //Checking if cookie is supported as there are some old browsers & some OS not support cookie.
+  try {
+    rawDocument.cookie;
+  }
+  catch (e) {
+    cookieSupported = false;
+  }
   return function() {
     var cookieArray, cookie, i, index, name;
-    var currentCookieString = rawDocument.cookie || '';
+    var currentCookieString = cookieSupported ? rawDocument.cookie || '' : '';
 
     if (currentCookieString !== lastCookieString) {
       lastCookieString = currentCookieString;
--- a/src/ng/urlUtils.js
+++ b/src/ng/urlUtils.js
@@ -81,7 +81,16 @@ function urlResolve(url) {
       : '/' + urlParsingNode.pathname
   };
 }
-
+/**
+ * Parse a request URL
+ *
+ * @param {string|object} url The url of the request as a string that will be resolved
+ * or a parsed URL object.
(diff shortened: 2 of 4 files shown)


mo3taz-abdallh:
I signed CLA

src/ng/urlUtils.js
```diff
@@ -94,3 +94,15 @@ function urlIsSameOrigin(requestUrl) {
   return (parsed.protocol === originUrl.protocol &&
           parsed.host === originUrl.host);
 }
+
+  /**
+  * Parse a request URL & determine if its file protocol to avoid $$cookieReader of XSRF token in case of file.
+  *
+  * @param {string|object} requestUrl The url of the request as a string that will be resolved
+  * or a parsed URL object.
+  * @returns {boolean} Whether the request is for the same origin as the application document.
```
gkalpak: This line is incorrect.

src/ng/urlUtils.js
```diff
@@ -94,3 +94,15 @@ function urlIsSameOrigin(requestUrl) {
...
+  */
+function urlIsFromFile(requestUrl) {
+  var parsed = (isString(requestUrl)) ? urlResolve(requestUrl) : requestUrl;
```
gkalpak: Let's avoid parsing the URL twice (once in `urlIsSameOrigin` and one here).
Let's parse it once and pass tge parsed URL to both functions.

gkalpak:
We also need tests verifying the new behavior (and guarding against future regressions).

BTW, could you give more details on the issue ? Why is it a problem trying to access the cookie in WebOS ?

src/ng/urlUtils.js
mo3taz-abdallh: I would edit it & have one unique function for parsing

mo3taz-abdallh:
WebOS serve from file:// protocol which through an exception in $$cookieReader() function i could send to you debugging exception, this occurs on all TVs that have WebOS, which i guess it should handled as frequency high.

gkalpak:
Could you identify which line exaclty throughs inside `$$cookieReader` (or post the complete stack trace along with the version of angular you use) ?

mo3taz-abdallh:
$$CookieReader throw the following exception 
Error: SecurityError: DOM Exception 18
in the following line
var currentCookieString = rawDocument.cookie || '';
as said previously serving is done in WebOS from File:// protocol

gkalpak:
After some Google-searching, it seems that this exception can be also thrown in (some older versions of) some browsers, when cookies are disabled.

It might make more sense not to special-case this for the `file` protocol, but trying to detect if cookies are supported during the initialization of the `$$CookieReader`. E.g.:

```js
var notSupported = false;
try { rawDocument.cookie; } catch (err) { notSupported = true; }
```

--
@mo3taz-abdallh, wdyt ?

mo3taz-abdallh:
I didn't know if there are old browsers didn't support cookies but i faced that WebOs serve from file protocol which not support rawDocument.cookie but i have reverted the previous commit and pre-checking for cookie support in $$CookieReader.

src/ng/cookieReader.js
```diff
@@ -21,10 +22,16 @@ function $$CookieReader($document) {
       return str;
     }
   }
-
+  //Checking if cookie is supported as there are some old browsers & some OS not support cookie.
```
gkalpak: Add a mention to WebOS in the comment.

src/ng/cookieReader.js
```diff
@@ -21,10 +22,16 @@ function $$CookieReader($document) {
...
+  try {
+    rawDocument.cookie;
+  }
+  catch (e) {
+    cookieSupported = false;
+  }
```
gkalpak: Nitpicking: A newline before and after the try-catch block would be nice.

src/ng/cookieReader.js
```diff
@@ -21,10 +22,16 @@ function $$CookieReader($document) {
...
   return function() {
     var cookieArray, cookie, i, index, name;
-    var currentCookieString = rawDocument.cookie || '';
+    var currentCookieString = cookieSupported ? rawDocument.cookie || '' : '';
```
gkalpak: Nitpicking: IMO `(cookieSupported && rawDocument.cookie) || ''` reads better.

gkalpak:
I left some minor comments (mostly nitpicks).
There need to be some tests, before we can merge this.

mo3taz-abdallh:
Ok wait for testing from yours side, then merge.
Regards

gkalpak:
@mo3taz-abdallh, I meant that you need to write some tests (that verify the fix and prevent from future regressions). We can't merge a PR without accompanying tests.

gkalpak:
Closing since we can't merge this without accompanying tests.
Feel free to re-open when you make the necessary changes.

Issue #8608, opened 2014-08-14
Feature request: Pass params object to $resource param if it's a function


philBrown:
The doco for ngResource states that _paramDefaults_ and action `params` may be functions however there is no useful data passed to these functions.

It would be handy to at least have the `actionParams` passed in. For example, change this line in `extractParams`

```
if (isFunction(value)) { value = value(); }
```

to

```
if (isFunction(value)) { value = value(actionParams); }
```

philBrown:
@SekibOmazic you don't have to use the function argument. You can just leave it as

    group: function() {
        return currentGroup;
    }
    
The point is that the action parameters would be there if you do need to use them.

SekibOmazic:
@philBrown Yes, in that case they'd be useful.

gkalpak:
I don't see the usefulness in passing `actionParams`. It's just a definition object (saying where to look up the values for each param) usually.

@philBrown, what's the usecase for this ?
IMO, it would much more interesting to pass the actual `data` and/or the `params` passed to the method call as arguments. WDYT ?

philBrown:
@gkalpak it's been a while now so I can't remember exactly what my original use-case was but it was something along the lines of creating dynamic param values based on the other params, ie having the param function aware of the call-time state.

I agree that having access to the `data` as well could be beneficial.

gkalpak:
@philBrown, `actionParams` (as you suggested) does not really contain "call-time state". It's the static params (usually just references to item properties).

Passing the `params` and `data` should give some "call-time" context. Luckily, there is a PR for this #14003 :smiley:

Narretz:
We have changed this a long time ago to pass the `data`. Not the params though. But since this is a low frequency issue, I'm going to close it.

PR #14003, opened 2016-02-11
feat(resource): Pass parameter object to params that are functions

LeeAdcock:
When defining default or action-specific resource parameters using functions, the parameter object will be passed into each of the functions. This parameter object can then optionally be used when deriving and returning the parameter value. 

Closes https://github.com/angular/angular.js/issues/8608

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -134,7 +134,7 @@ function shallowClearAndCopy(src, dst) {
  *   - **`params`** – {Object=} – Optional set of pre-bound parameters for this action. If any of
  *     the parameter value is a function, it will be executed every time when a param value needs to
  *     be obtained for a request (unless the param was overridden). The function will be passed two
- *     parameters, a hash of the parameter keys with values and the body for `POST`, `PUT`, and 
+ *     parameters, a hash of the parameter keys with values and the body for `POST`, `PUT`, and
  *    `PATCH` requests. The second parameter will be undefined for all other requests.
  *   - **`url`** – {string} – action specific `url` override. The url templating is supported just
  *     like for the resource-level urls.
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -102,7 +102,9 @@ function shallowClearAndCopy(src, dst) {
  * @param {Object=} paramDefaults Default values for `url` parameters. These can be overridden in
  *   `actions` methods. If a parameter value is a function, it will be executed every time
  *   when a param value needs to be obtained for a request (unless the param was overridden). The
- *   function will be passed a hash of the parameter keys and values.
+ *   function will be passed two parameters, a hash of the parameter keys with values and the body
+ *   for `POST`, `PUT`, and `PATCH` requests. The second parameter will be undefined for all other
+ *   requests.
  *
  *   Each key value in the parameter object is first bound to url template if present and then any
  *   excess keys are appended to the url search query after the `?`.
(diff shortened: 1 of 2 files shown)


gkalpak:
From a quick look, I'm afraid this feature will introduce some non-determinism, which makes it kind of not-so-useful.

I.e. the order in which the `actionParams` properties are traversed (which is browser dependent) in combination with the order of params can affect the output of a function.

Maybe we should pass the `params` and `data` passed to the call of the method ?

LeeAdcock:
The actionParams passed to the functions is a separate data structure from the one used to store the function outputs.  For that reason, the functions won't have access to the return values from other functions, and the values they have won't change depending on execution order.

I had looked at a different implementation which did ```extend({}, actionParams, ids)``` so returned values would be passed to the next function, but as you suggested that added non-deterministic issues, which would be more harmful than helpful. Even if forced a deterministic output by doing something executing a specific order, its just not developer-friendly, yuk.

I can play around with passing the ```data``` and see if that might be more useful.

gkalpak:
> The actionParams passed to the functions is a separate data structure from the one used to store the function outputs.

You are right; I missed that. In that case, it's even less useful, since it will mostly contain property names, not actual values. I don't think there is any benefit in setting static values in `actionParams` (just use a different URL).

I do see merit in passing the `data` and/or `params` objects, though.
@LeeAdcock, feel free to try that out and ping me if you make any changes.

LeeAdcock:
I agree the params object seems like of little value, unless you had to do some sort of value transform and wanted to put the logic in default params to consume values from the action params. The data object would only be useful for value transforms for POST, PUT, PATCH...for all other verbs it is undefined.

Perhaps the most flexible would be to pass both, as you suggested. I would suggest we pass the param object first since it is more likely to have values defined than the data object.

```
if (isFunction(value)) { value = value(actionParams, data); }
```

Thoughts @gkalpak?

gkalpak:
@LeeAdcock, sounds good ! (Exactly what I has in mind :wink:)

LeeAdcock:
Implemented as we discussed, thanks for the feedback @gkalpak

gkalpak:
@LeeAdcock, I see you pass `actionParams` and `data`. I actually had `params` and `data` in mind (but `params` is not even passed to `extractParams()` - that could change of course :wink:).

My reasoning is that `paramDefaults` are available at "definition" time (most of the time `actionParams` too, but not always), so the developer has access to them already. It is `params` that is not available until "call" time.

--
There is also another "issue" with the current implementation (prior to this PR).
Consider the following example:

```js
var Thing = $resource(url, { /* paramDefaults */ }, {
  customAction: {
    params: { /* actionParams */ },
    ...
  }
});

var thing = Thing.get({ /* params */ }, ...);
```

Here is what we are doing:

1. Extend `paramDefaults` with `actionParams`.
2. Determine the values for the resulting object (including calling any functions and looking up `@` values in `data`). This is done inside the [`extractParams()`][1] function.
3. Then we return `ids` and we extend it with `params` [here][2].

Since on step (3) we might be overwriting some values computed in step (2), we are doing some potentially unnecessary work (calling functions and looking up properties in `data` for keys that will be overwritten right after).

--
Here is an implementation that takes care of both issues/concerns mentioned above:

**(1)**
Replace [this line][2] with:

```js
extractParams(data, action.params, params),
```

**(2)**
Change the implementation of [`extractParams`][1] to:

```js
function extractParams(data, actionParams, params) {
  var ids = {};

  mergedParams = extend({}, paramDefaults, actionParams, params);
  forEach(mergedParams, function(value, key) {
    if (!params || !params.hasOwnProperty(key)) {
      if (isFunction(value)) value = value(mergedParams, data);
      if (value && value.charAt && value.charAt(0) === '@') {
        value = lookupDottedPath(data, value.substr(1));
      }

      ids[key] = value;
    }
  });

  return ids;
}
```

WDYT, @LeeAdcock ?

BTW, thx for working on this and other stuff (and bearing with my feedback :grin:) !
Really appreciate it :+1: 

[1]: https://github.com/LeeAdcock/angular.js/blob/params-to-resource-8608/src/ngResource/resource.js#L555
[2]: https://github.com/LeeAdcock/angular.js/blob/params-to-resource-8608/src/ngResource/resource.js#L672

LeeAdcock:
I like this, and I agree it seems more likely to be useful than the original implementation with ```defaultParams```/```actionParams```.  Thank you for continuing to point me in the right direction, I appreciate it :).

This approach does seem to introduce one change in behavior that might not have been intended. By doing calling ```extend({}, paramDefaults, actionParams, params)``` before processing the functions and  ```@``` references, it is now possible to provide functions and references as ```params``` values, where they were not supported previously.  Although this doesn't seem particularly useful, I don't see where it is a problem either. It makes an infinite recursion problem possible, but seems like developers have to be really trying in order to encounter it.

I've made the change and am working through updating the tests.

gkalpak:
> it is now possible to provide functions and references as params values

The `!params.hasOwnProperty(key)` check was added exactly to prevent that change in behavior. Did I miss something ?

Narretz:
We have changed this a long time ago to pass the `data`. Not the params though. But since this is a low frequency issue, I'm going to close it.

PR #14004, opened 2016-02-11
Update controller can be function

ashmna:
Controller can be function, this case needed when implemented page permissions


gkalpak:
Tgx for the PR, but this change is incorrect. If `controller` is a function, then it is assumed to be the constructor function for the controller. If you need to check foe permissions, you can do it using the `resolve`.

Issue #14006, opened 2016-02-11
remaining call to console.log breaks ie9


lucienbertin:
in [angular.bind.js](https://github.com/angular/angular.js/blob/master/src/angular.bind.js) there is a call to `console.log` but `window.console` only exists when the devTools are open in ie9 _[reference](http://stackoverflow.com/questions/5472938/does-ie9-support-console-log-and-is-it-a-real-function)_

PR #14007, opened 2016-02-11
only call console.log when window.console exists

lucienbertin:
issue #14006 - `window.console` only exists in IE 8 & 9 when the devtools are open

### DIFF
--- a/src/angular.bind.js
+++ b/src/angular.bind.js
@@ -1,6 +1,8 @@
 if (window.angular.bootstrap) {
   //AngularJS is already loaded, so we can return here...
-  console.log('WARNING: Tried to load angular more than once.');
+  if (!!window.console) {
+    console.log('WARNING: Tried to load angular more than once.');
+  }
   return;
 }
 


lucienbertin:
I signed it!

src/angular.bind.js
```diff
@@ -1,6 +1,8 @@
 if (window.angular.bootstrap) {
   //AngularJS is already loaded, so we can return here...
-  console.log('WARNING: Tried to load angular more than once.');
+  if (!!window.console) {
```
Narretz: Why not `window.console`?

Narretz:
Ideally, we should have an e2e test for this. Do you want to give it a shot? There's a sample e2e test in test/e2e/sample.

src/angular.bind.js
lucienbertin: that's personal preference. I prefer to test if something is `true` rather than `truthy` ; `!!window.console` is a boolean while `window.console` is _something_

I don't think it has any impact on anything so if you prefer `if (window.console)` i'll change it

Narretz: If you are doing !! then you are still relying on console being truthy though

lucienbertin: yeah but my `if` is testing a boolean, not a _something_ that is truthy or falsy

when i read this code
```js
var foo, bar;
...
if (foo) { ... }
if (!!bar) { ... }
```

I assume foo is a `boolean` while bar is a `something`

as i said it's just personal preference

----- 

fun fact with [momentjs](http://momentjs.com/), `moment.duration(0)` is truthy while being equal to 0

```js
moment.duration(0) == 0; // true
moment.duration(0) === 0; // false
!!moment.duration(0); // true
!!0; // false, obviously
```

lucienbertin:
I tried for the last 2 hours to make grunt work, following the [angular contribute doc](https://docs.angularjs.org/misc/contribute) but it just don't want to register there is a `Gruntfile.js` in the dir

I dont know whats causing it, i'll let you handle the e2e test

I'll try to make it work for my next contribution

----------

i'm running a command prompt on win 10 as admin and typing

```
npm install grunt grunt-cli -g
grunt package
```

and getting

```
Fatal error: Unable to find local grunt
```

when the `dir` cmd gives me

```
Mode                LastWriteTime         Length Name
----                -------------         ------ ----
... ellipsis...
-a----        2/11/2016   4:36 PM           7170 gdocs.js
-a----        2/11/2016   4:36 PM          10729 Gruntfile.js
-a----        2/11/2016   4:36 PM            610 init-repo.sh
... ellipsis...
```

I've worked with grunt professionally for the past year, it's not my first walk around the park

Narretz:
Grunt must be installed without the -g flag. Only grunt-cli must be
installed globally
Am 11.02.2016 19:13 schrieb "lucienbertin" <notifications@github.com>:

lucienbertin:
```
> grunt package
Loading "Gruntfile.js" tasks...ERROR
>> Error: Cannot find module 'shelljs'
Warning: Task "package" not found. Use --force to continue.

Aborted due to warnings.
```

I'm giving up here on the e2e test

also grunt is referenced in `package.json` so it should get installed when i run `npm install`, i shouldn't have to run `npm install grunt`

Narretz:
Sorry about that. It's possible that the latest node / npm versions are not quite compatible yet with our ancient setup.

PR #14008, opened 2016-02-11
docs: allow plnkr links to open in new window

### LINKED ISSUE #8328: feat(docs): plnkr examples should open in new window
Previously, "Edit in Plunker" links from the API reference docs would open plunker in a new window. However, pop up blockers would frequently block the new window unless a user explicitly allowed popups from the domain. Even manually clicking through to the link would not populate the plunker with the correct template because the feature relied on posting FormData to plunker which would be clobbered by the popup blocker. This was temporarily fixed by https://github.com/angular/angular.js/pull/8327 which posts the form in the same window. 

Presumably, the pop-up blocker was not allowing the new window to open because the submission of the form was not happening within the context of a user event (the click), but was instead waiting on some data to be loaded over XHR, then submitted to plunker.

A better user experience would be to have the form and form-date pre-populated on the page, and immediately submit the form within `ngClick`.

Narretz:
(with ctrl + click)
We could also make this the default, but I don't know if popup-blockers will still block this

Closes #8328

### DIFF
--- a/docs/app/src/docs.js
+++ b/docs/app/src/docs.js
@@ -1,12 +1,10 @@
 angular.module('DocsController', [])
 
 .controller('DocsController', [
-          '$scope', '$rootScope', '$location', '$window', '$cookies', 'openPlunkr',
+          '$scope', '$rootScope', '$location', '$window', '$cookies',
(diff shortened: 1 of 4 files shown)


gkalpak:
:+1:

PR #14013, opened 2016-02-11
feat(dateFilter): add support for STANDALONEMONTH in format (`LLLL`)

### LINKED ISSUE #13999: Date filter (still) shows incorrect month names for Slavic languages
As a follow up of this issue:
https://github.com/angular/angular.js/issues/10247

Languages like Polish and Croatian have different versions of months names depending on the context.

The Angular's localization files now have both versions present ("MONTH" and "STANDALONEMONTH"). But the date filter doesn't seems to be always using the right ones, actually it still only seems to use the "MONTH" version at all times

For example if you use "MMMM" to just get the month's name it should show the version from "STANDALONEMONTH".

Would it be possible to at least make the plain "MMMM" one behave correctly? Because if it does then AngularUI's datepicker will work correctly too. Thanks.

Or introduce a new code, like five M's ('MMMMM') or something for the standalone version so people can choose themselves what version should be used.

I'm on 1.4.9 by the way.

gkalpak:
Fixes #13999

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -358,10 +358,11 @@ function dateGetter(name, size, offset, trim) {
   };
 }
 
-function dateStrGetter(name, shortForm) {
+function dateStrGetter(name, shortForm, formatProp) {
   return function(date, formats) {
     var value = date['get' + name]();
-    var get = uppercase(shortForm ? ('SHORT' + name) : name);
+    var propName = formatProp || name;
+    var get = uppercase(shortForm ? ('SHORT' + propName) : propName);
 
     return formats[get][value];
   };
@@ -423,6 +424,7 @@ var DATE_FORMATS = {
    MMM: dateStrGetter('Month', true),
     MM: dateGetter('Month', 2, 1),
      M: dateGetter('Month', 1, 1),
+  LLLL: dateStrGetter('Month', false, 'StandaloneMonth'),
     dd: dateGetter('Date', 2),
      d: dateGetter('Date', 1),
     HH: dateGetter('Hours', 2),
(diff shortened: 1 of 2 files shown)


src/ng/filter/filters.js
```diff
@@ -358,10 +358,11 @@ function dateGetter(name, size, offset, trim) {
   };
 }
 
-function dateStrGetter(name, shortForm) {
+function dateStrGetter(name, shortForm, formatProp) {
   return function(date, formats) {
     var value = date['get' + name]();
-    var get = uppercase(shortForm ? ('SHORT' + name) : name);
+    var propName = formatProp || name;
+    var get = uppercase(shortForm ? ('SHORT' + propName) : propName);
```
lgalfaso: If we change this to
```js
function dateStrGetter(name, shortForm, propPrefix) {
...
var get = uppercase((propPrefix || '') + (shortForm ? 'Short' : '') + name);
...
```

And also change the line below to
```js
LLLL: dateStrGetter('Month', false, 'Standalone'),
```

Then, in the future, we can implement *LLL* by just doing
```js
LLL: dateStrGetter('Month', true, 'Standalone'),
```

[To change *Standalone* to a boolean is also fine]

src/ng/filter/filters.js
```diff
@@ -423,6 +424,7 @@ var DATE_FORMATS = {
    MMM: dateStrGetter('Month', true),
     MM: dateGetter('Month', 2, 1),
      M: dateGetter('Month', 1, 1),
+  LLLL: dateStrGetter('Month', false, 'StandaloneMonth'),
```
lgalfaso: StandAlone (as two words)

src/ng/filter/filters.js
gkalpak: Good point ! (It was late :stuck_out_tongue:)
I went with the boolean (but it's easy to switch to a string prefix if needed).

src/ng/filter/filters.js
gkalpak: Changed to boolean :smiley:

lgalfaso:
LGTM

gkalpak:
Backported to `v1.5.x` as 3e5b25b.

PR #14016, opened 2016-02-12
docs(error/nobase): show <base> tag

remy:
The <base> tag was missing from the doc page - this change (presumes) to fix that and escape the tag in the title.

screen shot 2016-02-12 at 11 39 29


remy:
I signed it!

gkalpak:
Backported to `v1.5.x` as b42147a (although I don't think it's necessary).
There is also another issue in on the next line, but it needs to be tackled differently (will take care of it in another PR).

PR #14021, opened 2016-02-12
feat($http,$resource): statusText resolved in timeout promise

mrac:
Timeout-promise resolved value will be passed as response error statusText.

``` js
  var promise = $timeout(function() {
    return "something";
  }, 500);

  $http({
    method: 'GET',
    url: '...',
    timeout: promise
  }).catch(function(err) {
    console.log(err.status);  // -1
    console.log(err.statusText);  // something
  });
```

This feature will also help distinguish cancelled $resource requests
from timed-out $resource requests. Before, both cases resulted
with the same error object:
`{status: -1, statusText: '', ...}`
thus impossible to handle both cases in different way.

Now cancelled requests will come with statusText='cancelled',
and timed-out requests with statusText='timeout' in the error object.

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1350,11 +1350,7 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
         for (var i = 0, ii = responses.length; i < ii; i++) {
           if (responses[i] === handleResponse) {
             responses.splice(i, 1);
-            if (statusText) {
-                callback(-1, undefined, '', statusText);
-            } else {
-                callback(-1, undefined, '');
-            }
+            callback(-1, undefined, '', statusText);
             break;
           }
         }
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1346,11 +1346,15 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
                  copy(response[3] || ''));
       }
 
-      function handleTimeout() {
+      function handleTimeout(statusText) {
         for (var i = 0, ii = responses.length; i < ii; i++) {
           if (responses[i] === handleResponse) {
             responses.splice(i, 1);
-            callback(-1, undefined, '');
+            if (statusText) {
+                callback(-1, undefined, '', statusText);
+            } else {
+                callback(-1, undefined, '');
+            }
             break;
           }
         }
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -658,7 +658,9 @@ angular.module('ngResource', ['ng']).
               httpConfig.timeout = timeoutDeferred.promise;
 
               if (numericTimeout) {
-                numericTimeoutPromise = $timeout(timeoutDeferred.resolve, numericTimeout);
+                numericTimeoutPromise = $timeout(function() {
+                    timeoutDeferred.resolve("timeout");
+                }, numericTimeout);
               }
             }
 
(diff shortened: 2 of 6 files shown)


src/ngMock/angular-mocks.js
```diff
@@ -1346,11 +1346,15 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
                  copy(response[3] || ''));
       }
 
-      function handleTimeout() {
+      function handleTimeout(statusText) {
         for (var i = 0, ii = responses.length; i < ii; i++) {
           if (responses[i] === handleResponse) {
             responses.splice(i, 1);
-            callback(-1, undefined, '');
+            if (statusText) {
```
gkalpak: Why not just always call `callback(-1, undefined, '', statusText)` ?

src/ngResource/resource.js
```diff
@@ -658,7 +658,9 @@ angular.module('ngResource', ['ng']).
               httpConfig.timeout = timeoutDeferred.promise;
 
               if (numericTimeout) {
-                numericTimeoutPromise = $timeout(timeoutDeferred.resolve, numericTimeout);
+                numericTimeoutPromise = $timeout(function() {
+                    timeoutDeferred.resolve("timeout");
+                }, numericTimeout);
```
gkalpak: This can be also written as `numericTimeoutPromise = $timeout(timeoutDeferred.resolve, numericTimeout, false, 'timeout');`

gkalpak: But...you know...it's better to stay as is :disappointed_relieved:

test/ngResource/resourceSpec.js
```diff
@@ -1544,7 +1544,13 @@ describe('cancelling requests', function() {
       }
     });
 
-    CreditCard.get();
+    var creditCard = CreditCard.get();
+
+    creditCard.$promise.catch(function(err) {
```
gkalpak: It would be better to also verify that the function has been called (e.g. by using a spy).
(Here and in the tests below.)

gkalpak:
Overall looks good. I left a couple of comments.

src/ngMock/angular-mocks.js
mrac: Because some unit tests use spy's ```toHaveBeenCalledWith()``` method which is strict regarding the number of parameters, and I didn't want to break the old API. With ```undefined``` as the forth parameter those tests will fail, so may be some third-party modules would break if they test for it. It's minor however, and if you agree, we can just change it.

test/ngResource/resourceSpec.js
mrac: thnx. I pushed a fix

src/ngMock/angular-mocks.js
gkalpak: The advice in the docs is that _"you should never need to use this service directly"_ and it's use as a function isn't even documented (even more so the arguments of the `done` callback).
I really think it's a non-issue, so go ahead and use the simpler version :smile:

gkalpak:
@mrac , you never pinged me and I missed your latest change.
LGTM (expect that this needs to be rebased and the test have to be converted to Jasmine 2).

Narretz:
Is this a BC? The statusText is now something that it wasn't before in some cases. Could break some logic / tests.

gkalpak:
In that sense it can be thoght of as a BC, I guess.

nicolas-berezin:
@Narretz @gkalpak 
Hey guys,
Seems like this issue still exists on the latest AngularJS build.
If ngResource is configured as:
`$resource("url", {}, {
            list: {
                method: 'GET',
                cancellable: true,
                timeout: 1000
            }`
 So if it is cancellable and has a numeric timeout at the same time, the response object will always have 
`xhrStatus: 'abort'` even if the request was cancelled by timeout.
There is no way to distinguish was it cancelled by calling `$cancelRequest()` or it was cancelled by timeout.

PR #14023, opened 2016-02-12
fix(ngValue): set value property instead of attribute

### LINKED ISSUE #13984: ngValue one way binding?
Hi.

I'm trying to use ng-value in an input box as a one way binding. But the thing is, when the model changes, the input box is not reflecting the new value. I'm debugging the ngValue directive and i can see that the watcher is triggered, under the hood the value is being set with setAttribute (using element.attr). That is setting the value of the attribute, but it is not changin the view.

I created this fiddler, and I see this strange behaviour. Just have a look at the value reflected in the view and the value in the console.debug. Is this normal behaviour? Can anyone explain this to me?

https://jsfiddle.net/upqeksf0/
This is in Chrome 46, I haven't checked other browsers.

m-amr:
in case of input element set value usering property instead of attribute
fixed #13984

### DIFF
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -140,6 +140,7 @@
     "jqLiteParseHTML": false,
     "jqLiteWrapNode": false,
     "getBooleanAttrName": false,
+    "getValueAttrName": false,
     "getAliasedAttrName": false,
     "createEventHandler": false,
     "JQLitePrototype": false,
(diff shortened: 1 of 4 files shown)


gkalpak:
I don't think this is the right fix. The behavior of `$set` should not be changed imo.
The change must be local to `ngValue`.

m-amr:
@gkalpak , I tried to solve the problem inside the $set, but based on your opinion
That the change must be local to ngValue in this pull request
https://github.com/angular/angular.js/pull/14031

gkalpak:
Closing in favor of #14031.

PR #14026, opened 2016-02-12
fix(dateFilter): Correctly format BC years

lgalfaso:
- Correctly format BC years
- Fix a function name collition
- Allow TzDate to use BC years

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -331,11 +331,15 @@ function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
   }
 }
 
-function padNumber(num, digits, trim) {
+function padNumber(num, digits, trim, negWrap) {
   var neg = '';
-  if (num < 0) {
-    neg =  '-';
-    num = -num;
+  if (num < 0 || (negWrap && num <= 0)) {
+    if (negWrap) {
+      num = -num + 1;
+    } else {
+      num = -num;
+      neg = '-';
+    }
   }
   num = '' + num;
   while (num.length < digits) num = ZERO_CHAR + num;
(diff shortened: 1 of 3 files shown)


src/ng/filter/filters.js
```diff
@@ -331,11 +331,15 @@ function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
...
 }
 
-function padNumber(num, digits, trim) {
+function padNumber(num, digits, trim, negWrap) {
   var neg = '';
   if (num < 0) {
-    neg =  '-';
-    num = -num;
+    if (negWrap) {
+      num = -num + 1;
```
gkalpak: Is this `+1` necessary ? I think it's not consistent with how browsers treat negative years.

lgalfaso: Year 0 is 1 BC, year -1 is 2 BC

gkalpak: I know, but it is weird, because `0000-01-01` will be `0 BC` and `-0001-01-01` will be `2 BC`. It's kind of a gray area, so I'm not too bothered.

I jsut realized: what if `G` is not included in the format string ?
Then `-0026-01-01` will be `January 1st, 27`, right ? :confused:

lgalfaso: Sorry, I misunderstood your original point.

> I just realized: what if G is not included in the format string ?
> Then -0026-01-01 will be January 1st, 27, right ?

Yes, but I would say this is expected. If there is a need to reflect BC years as 0/negative, then we have to add support for the pattern *u...*

lgalfaso:
Updated the PR with the year 0 fix

gkalpak:
LGTM

lgalfaso:
Landed with
9421674dad0e5aa9703449174490f1b985858460

and backported to 1.5 with
e36205f5af82b69362def7d2b6eeeb038f592311

PR #14031, opened 2016-02-13
fix(ngValue): set value property instead of attr

### LINKED ISSUE #13984: ngValue one way binding?

m-amr:
in case of input element set value using property instead of attribute
fixed #13984

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1701,6 +1701,10 @@ var CONSTANT_VALUE_REGEXP = /^(true|false|\d+)$/;
  * so that when the element is selected, the {@link ngModel `ngModel`} of that element is set to
  * the bound value.
  *
+ * It can be used to achieve one-way binding of a given expression to the value of `element`,
+ * so that when the expression change the element is set to the bound value.
+ *
+ *
  * `ngValue` is useful when dynamically generating lists of radio buttons using
  * {@link ngRepeat `ngRepeat`}, as shown below.
  *
@@ -1711,7 +1715,7 @@ var CONSTANT_VALUE_REGEXP = /^(true|false|\d+)$/;
  *
  * @element input
  * @param {string=} ngValue angular expression, whose value will be bound to the `value` attribute
- *   of the `input` element
+ * and `value` property of the element
  *
  * @example
     <example name="ngValue-directive" module="valueExample">
@@ -1750,24 +1754,33 @@ var CONSTANT_VALUE_REGEXP = /^(true|false|\d+)$/;
     </example>
  */
 var ngValueDirective = function() {
+  //helper function
+  var updateElementValue = function(element, attr, value) {
+    /**
+     *  input use 'value attribute' as default value if 'value property has no value'
+     *  but once set the 'value property' it will not read value from 'value attribute'
+     *  so we set both the value attribute and property here to avoid this problem.
+     */
+    attr.$set('value', value);
+    element.prop('value', value);
+  };
+
   return {
     restrict: 'A',
     priority: 100,
     compile: function(tpl, tplAttr) {
       if (CONSTANT_VALUE_REGEXP.test(tplAttr.ngValue)) {
         return function ngValueConstantLink(scope, elm, attr) {
-          attr.$set('value', scope.$eval(attr.ngValue));
+          var value = scope.$eval(attr.ngValue);
+          updateElementValue(elm, attr, value);
         };
       } else {
         return function ngValueLink(scope, elm, attr) {
           scope.$watch(attr.ngValue, function valueWatchAction(value) {
-            attr.$set('value', value);
+            updateElementValue(elm, attr, value);
           });
         };
       }
     }
   };
 };
-
-
-
--- a/test/ng/directive/inputSpec.js
+++ b/test/ng/directive/inputSpec.js
@@ -3046,6 +3046,22 @@ describe('input', function() {
       expect(inputElm[0].getAttribute('value')).toBe('something');
     });
 
+    they('should update the $prop "value" property and attribute after change the "value" property', {
+      input: '<input type="text" ng-value="value">',
+      textarea: '<textarea ng-value="value"></textarea>'
+    }, function(tmpl) {
+      var element = helper.compileInput(tmpl);
+
+      helper.changeInputValueTo('newValue');
+      expect(element[0].value).toBe('newValue');
+      expect(element[0].getAttribute('value')).toBeNull();
+
+      $rootScope.$apply(function() {
+        $rootScope.value = 'anotherValue';
+      });
+      expect(element[0].value).toBe('anotherValue');
+      expect(element[0].getAttribute('value')).toBe('anotherValue');
+    });
 
     it('should evaluate and set constant expressions', function() {
       var inputElm = helper.compileInput('<input type="radio" ng-model="selected" ng-value="true">' +


src/ng/directive/input.js
```diff
@@ -1756,12 +1756,25 @@ var ngValueDirective = function() {
...
         return function ngValueConstantLink(scope, elm, attr) {
-          attr.$set('value', scope.$eval(attr.ngValue));
+          var value = scope.$eval(attr.ngValue);
+          /**
+           *  input use 'value attribute' as default value if 'value property has no value'
+           *  but once set the 'value property' it will not read value from 'value attribute'
+           *  so we set both the value attribute and property here to avoid this problem.
+           */
+          attr.$set('value', value);
+          elm[0].value = value;
```
gkalpak: Could you extract these two lines (along with the comment above) into a helper function to avoid duplication.

gkalpak:
The docs also need to be updated to reflect the changes.

I would still add a breaking change note, since we are breaking a (documented-ish) assumption.
I can't think of any usecase, where this change would indeed break an app, but it is a breaking change.

m-amr:
@gkalpak 

1. I have extracted a method to update the value to prevent code duplication .
2. I have updating the commit message with the breaking change.
3. I can't found anything to be update in the documentation. 
because the default behaviour of the ng-value is to update the input when the model change and this is a bug. please tell me if documentation is need to be updated ? 

Thanks for your feedback.

gkalpak:
WRT to the docs, e.g. soewhere it says that the resulting _"value will be bound to the `value` attribute of the input element_", which is not accurate anymore.

The docs need a more thorough make-over anyway, because they mention `option` and `input[radio]` elements, but with this change `ngValue` can be used for more purposes, so I can take care of the docs update while mergin or later.

--
It LGTM.
@Narretz, since you were working on some `ngValue` stuff, do you have any objection/concerns about this ?

Narretz:
I would like to see the setter code to be more precise in terms of which elements need attribute set and which need property set. So basically the setter fn should be set up based on the element type. I assume the value property should be set for textarea and input[text], while option and input[radio|checkbox] are good with attribute only.
And the docs must be updated. Atm, the docs are pretty explicit that the directive is only for radio/checkbox and option elements.
So the change is actually a feature and not a fix.

m-amr:
@Narretz I think that the setter is setting the property in case of boolean values, i tried ti fix it in this pull request  https://github.com/angular/angular.js/pull/14023

Narretz:
@m-amr I'm not sure I follow. Input elements / controls cannot have boolean values. Their values are always string values. Angular models are a different situation, but they generally follow this rule, with exceptions.

m-amr:
@Narretz 
sorry I didn't mean that their values are boolean, i mean that they had properties that is boolean values

A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

        var BOOLEAN_ATTR = {};
           forEach('multiple,selected,checked,disabled,readOnly,required,open'.split(','), function(value) {
             BOOLEAN_ATTR[lowercase(value)] = value;
         });
        var BOOLEAN_ELEMENTS = {};
        forEach('input,select,option,textarea,button,form,details'.split(','), function(value) {
            BOOLEAN_ELEMENTS[value] = true;
        });

so if the in case of boolean attribute it sets the property

         $set: function(key, value, writeAttr, attrName) {
        // TODO: decide whether or not to throw an error if "class"
        //is set through this function since it may cause $updateClass to
        //become unstable.

        var node = this.$$element[0],
            booleanKey = getBooleanAttrName(node, key),
            aliasedKey = getAliasedAttrName(key),
            observer = key,
            nodeName;

        if (booleanKey) {
          this.$$element.prop(key, value);
          attrName = booleanKey;
        } else if (aliasedKey) {
          this[aliasedKey] = value;
          observer = aliasedKey;
        }

so it will call this.$$element.prop(key, value); in case of boolean attribute.

m-amr:
Sorry i had done git --amend to wrong commit, i have reset it and push it.

src/ng/directive/input.js
```diff
@@ -1750,24 +1750,33 @@ var CONSTANT_VALUE_REGEXP = /^(true|false|\d+)$/;
...
 var ngValueDirective = function() {
+  //helper function
+  var updateElementValue = function(element, attr, value) {
+    /**
+     *  input use 'value attribute' as default value if 'value property has no value'
+     *  but once set the 'value property' it will not read value from 'value attribute'
+     *  so we set both the value attribute and property here to avoid this problem.
+     */
+    attr.$set('value', value);
+    element[0].value = value;
```
Narretz: Should use `prop()`

test/ng/directive/inputSpec.js
```diff
@@ -3046,6 +3046,26 @@ describe('input', function() {
...
+    }, function(tmpl) {
+      var element = helper.compileInput(tmpl);
+
+      $rootScope.$apply(function() {
+        $rootScope.value = 'something';
+      });
+      expect(element[0].value).toBe('something');
+      expect(element[0].getAttribute('value')).toBe('something');
+
+      helper.changeInputValueTo('newValue');
```
Narretz: There is no expect after his change, so either it's redundant or an expect is missing

m-amr: This is not 'redundant' it important to test that the 
"input binding to the model" is still working after user edited the input, without it we are not testing the Issue 
if "value property"  has no value, it will read the value from "value attribute" 
The Issue happened when user type something in the input,
see the following step
1. User edit the input
2 . then change the model value programmatically by changing the scope.
3. The input is not updated will the new value of the model

I am trying to simulating what happen in the Issue https://github.com/angular/angular.js/issues/13984

do i need to add expect after it ?

Narretz: Yes, you should add an expect. Otherwise you can't be sure that the input / scope has really been changed.

gkalpak: @m-amr, basically this is redundant in this test, because we have already tested that the `value` property _does_ have a value. If you wanted to test the scenario your describe, you should edit the input, before setting the value (maybe in a separate test).

<sub>FWIW, I don't think an extra `expect` is needed.</sub>

m-amr:
@Narretz @gkalpak 
Could you please give it another look, i have done the following
 1. use prop function to set the value instead of element.value
 2. remove redundant code from test
 3. i had updated the docs

Thanks.

src/ng/directive/input.js
```diff
@@ -1701,6 +1701,11 @@ var CONSTANT_VALUE_REGEXP = /^(true|false|\d+)$/;
  * so that when the element is selected, the {@link ngModel `ngModel`} of that element is set to
  * the bound value.
  *
+ * It can be used to achieve one-way binding of a given expression to the value of
+ * {@link input[text] `input[text]`} or {@link textarea `textarea`}. so that when the expression change
```
gkalpak: Why only `input[text]`. Won't it work with any input type ?

src/ng/directive/input.js
```diff
@@ -1711,7 +1716,7 @@ var CONSTANT_VALUE_REGEXP = /^(true|false|\d+)$/;
  *
  * @element input
  * @param {string=} ngValue angular expression, whose value will be bound to the `value` attribute
- *   of the `input` element
+ * and `value` property of the `input` or `textarea` element
```
gkalpak: I don't think `input` was meant as `<input>` (it would be wrong). Maybe we should just say `element`, since it can be `input`, `textarea`, `option` (anything actually).

test/ng/directive/inputSpec.js
```diff
@@ -3046,6 +3046,22 @@ describe('input', function() {
       expect(inputElm[0].getAttribute('value')).toBe('something');
     });
 
+    they('should update the $prop "value" property and attribute after change the "value" property', {
+      input: '<input type="text" ng-value="value">',
+      textarea: '<textarea ng-value="value">></textarea>'
```
gkalpak: There's an extra `>`

gkalpak:
> I would like to see the setter code to be more precise in terms of which elements need attribute set and which need property set. So basically the setter fn should be set up based on the element type. I assume the value property should be set for textarea and input[text], while option and input[radio|checkbox] are good with attribute only.

I'd rather set both and not worry about browser quirks (unless this "strategy" would lead to problems).
@Narretz, do you know of any usecase, where setting both would be a problem ?

src/ng/directive/input.js
m-amr: I have to change it to be input type

test/ng/directive/inputSpec.js
m-amr: i had fixed it fixed.

src/ng/directive/input.js
m-amr: I have change it, i am now using 'element'  instead of input and textarea

src/ng/directive/input.js
```diff
@@ -1701,6 +1701,10 @@ var CONSTANT_VALUE_REGEXP = /^(true|false|\d+)$/;
  * so that when the element is selected, the {@link ngModel `ngModel`} of that element is set to
  * the bound value.
  *
+ * It can be used to achieve one-way binding of a given expression to the value of
+ * {@link input `input`}. so that when the expression change the element is set to the bound value.
```
gkalpak: `.`  -->  `,`

m-amr: Done.

Narretz:
@gkalpak I don't know of any way this could cause problems. I was just conservative, I guess. 
The BC notice needs some love, because it's a bit hard to understand, otherwise LGTM.

PR #14033, opened 2016-02-13
docs(errorDisplay): encode `<` and `>` in error messages

gkalpak:
When an error message contains an HTML string (e.g. `$location:nobase` containing `<base>`), it was interpreted as a literal HTML element, instead of text. Error messages are not expected to render as HTML, but we still need to use `.html()` in `errorDisplay`, so that the links created by `errorLinkFilter` are properly displayed.
This commit solves this issue by replacing `<`/`>` with `&lt;`/`&gt;`.

Related to #14016.
## 

There is also another commit that adds some tests for the `errors` module of the docs app.

### DIFF
--- a/docs/app/src/errors.js
+++ b/docs/app/src/errors.js
@@ -13,9 +13,9 @@ angular.module('errors', ['ngSanitize'])
   };
 
   return function (text, target) {
-    var targetHtml = target ? ' target="' + target + '"' : '';
-
     if (!text) return text;
+
+    var targetHtml = target ? ' target="' + target + '"' : '';
 
     return $sanitize(text.replace(LINKY_URL_REGEXP, function (url) {
       if (STACK_TRACE_REGEXP.test(url)) {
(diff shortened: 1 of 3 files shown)


gkalpak:
Before:
location nobase before

After:
location nobase after

docs/app/src/errors.js
```diff
@@ -34,6 +34,10 @@ angular.module('errors', ['ngSanitize'])
 
 
 .directive('errorDisplay', ['$location', 'errorLinkFilter', function ($location, errorLinkFilter) {
+  var encodeAngularBrackets = function (text) {
```
campersau: *encode***Angular***Brackets* -> *encode***Html***Brackets* ?

petebacondarwin: By Angular, @gkalpak means `<` and `>` right?

petebacondarwin:
This looks good but I wonder if we could save some complexity by simply using a binding (which should do the escaping for us) rather than writing to `element.html`?

docs/app/src/errors.js
gkalpak: I think "angle brackets" might be more "correct" (and certainly less confusing).

gkalpak:
@petebacondarwin, we need the HTML (because we want to render `<a>` elements), so not sure what you have in mind.

TBH, a much simpler workaround would be to change the error message to not include `<base>` (but just `base` for example), but who likes workarounds :)

gkalpak:
Merged as 321180a and 9c7c494.

petebacondarwin:
Nice thanks

PR #14035, opened 2016-02-14
feat($cacheFactory): add 'has' method to $cacheFactory cache

### LINKED ISSUE #8692: Implement $cacheFactory.Cache#has method
When checking in #8214 I was honestly surprised we don't expose a way to check if a key is in the cache.

It would be helpful to have a simple implementation, like `has: function(key) { return key in data; }` or something.

@IgorMinar this is a public API change so I didn't just write a patch, but what do you think? It's pretty trivial and better than testing if the value of a key is `undefined`

ryanhart2:
Add 'has' method to $cacheFactory cache that returns boolean to indicate if key exists in cache.
If the cache is a Least Recently Used (LRU) cache i.e. a capacity has been specified, a call to `has(key)` does not 'refresh' the key as a call to `get(key)` would.

Closes #8692


gkalpak:
I wonder what is the usecase for this ?

The problem with this is that if you intend to retrieve the value immediately, then you should just retrieve it in one pass (instead of checking if it is there and then retrieving it).
If you don't intend to retrieve the value immediately, then there is chance that the previous return value of `.has()` is no longer valid by the time you try to retrieve it (e.g. the item has been removed).
If you are trying to use the presence of a specific key in the cache as a flag, then you there is probably a better way :)

Am I missing something ?

ryanhart2:
To be honest, I didn't have a use case myself. I was just addressing issue #8692 opened by @caitp.

gkalpak:
Aha, I'm putting this to Ice Box then. We can revisit when/if we have a compelling usecase for this :smiley:

Narretz:
The demand for this is really low, and it's not really a must have feature. So let's not do it.

PR #14050, opened 2016-02-15
Update README.md

alex-teren:


gkalpak:
I hastily merged this without noticing there is no signed CLA :grin:
@alex-teren, could you please sign the CLA now ?

<sub>I know it is a tiny change, but we require a signed CLA for all code that gets merged.</sub>

PR #14051, opened 2016-02-15
feat(filter): uncurrency filter

ErinsMatthew:
New filter to convert formatted currency string to unformatted number.
This code is heavily based on [Open Exchange Rates's
accounting.js](https://github.com/openexchangerates/accounting.js)
unformat function which is Copyright (c) 2014 Open Exchange Rates.


Narretz:
I think that would be good as a 3rd party module instead of a part of core.

petebacondarwin:
I agree - this is not for core. Thanks for the submission @ErinsMatthew. I would love to see this published as a 3rd party module.

Issue #14054, opened 2016-02-16
ngMessages documentation is hard to follow


benelliott:
Opening issue for imminent PR

Narretz:
I've left some comments in your PR. I'll close this issue since it's not very useful without a description of the problems the docs have.

PR #14056, opened 2016-02-16
docs(ngMessages): Clarify ngMessages docs

benelliott:
Clarification of the core functionality of the ngMessages module

Addresses #14054


benelliott:
I signed it!

benelliott:
I've signed it again!

src/ngMessages/messages.js
```diff
@@ -24,10 +24,16 @@ var jqLite = angular.element;
  * `ngMessage` and `ngMessageExp` directives.
  *
  * # Usage
- * The `ngMessages` directive listens on a key/value collection which is set on the ngMessages attribute.
- * Since the {@link ngModel ngModel} directive exposes an `$error` object, this error object can be
- * used with `ngMessages` to display control error messages in an easier way than with just regular angular
- * template directives.
+ * The `ngMessages` directive listens on a key/value collection which is set on the ngMessages attribute of a container element.
```
Narretz: This reads like the ngMessages attribute is on a container element that is not the same as the ngMessages element

src/ngMessages/messages.js
```diff
@@ -24,10 +24,16 @@ var jqLite = angular.element;
...
+ * There should be one such element for each field whose error messages are to be displayed.
```
Narretz: This sentence is confusing. Either make it clear what this means, or remove it

src/ngMessages/messages.js
```diff
@@ -24,10 +24,16 @@ var jqLite = angular.element;
...
+ * The container element should have a series of child elements which each correspond to a possible error message for that field.
```
Narretz: It's not only error messages, it can be any string/message. The most common use is simply for $error elements.

src/ngMessages/messages.js
```diff
@@ -44,25 +50,34 @@ var jqLite = angular.element;
...
  * ```
  *
- * Now whatever key/value entries are present within the provided object (in this case `$error`) then
- * the ngMessages directive will render the inner first ngMessage directive (depending if the key values
- * match the attribute value present on each ngMessage directive). In other words, if your errors
- * object contains the following data:
+ * In order to show error messages corresponding to `myField` we first create a container element with an `ngMessages` attribute
+ * set to the `$error` object owned by the `myField` input in our `myForm` form.
+ *
+ * Within this container we create separate elements for each of the possible errors that `myField` could be annotated with. Exactly
```
Narretz: `annotated` is a poor description. It's simply "an element for each error myField could have"

src/ngMessages/messages.js
```diff
@@ -44,25 +50,34 @@ var jqLite = angular.element;
...
+ * which error a particular element corresponds to is declared using the `ngMessage` attribute - for example, `ng-message="required"`
+ * specified that this particular element should be displayed when the corresponding field is `required` but has no value.
```
Narretz: This is imo too specific and could simply be "should be displayed when `myField` has the required error."

src/ngMessages/messages.js
```diff
@@ -44,25 +50,34 @@ var jqLite = angular.element;
...
+ *
+ * By default, `ngMessages` will only display one error message for a particular field at any one time. Whether or not an error (child)
```
Narretz: "at any time"

src/ngMessages/messages.js
```diff
@@ -44,25 +50,34 @@ var jqLite = angular.element;
...
+ * element will be displayed is determined by whether the corresponding property in the error object is `true`, and the position of
```
Narretz: This sentence is too long and confusing.

src/ngMessages/messages.js
```diff
@@ -44,25 +50,34 @@ var jqLite = angular.element;
...
+ * that element relative to other relevant error elements in the container - error elements which appear first in the DOM will be
+ * displayed above others in cases where more than one such element is deemed to be relevant.
+ *
+ * Given the following error object for our example (which informs us that the field `myField` is currently erroneous in terms of both its
```
Narretz: " ... currently has the required and minlength errors"

src/ngMessages/messages.js
```diff
@@ -44,25 +50,34 @@ var jqLite = angular.element;
...
+ * `required` and its `minlength` atrributes):
  *
  * ```javascript
  * <!-- keep in mind that ngModel automatically sets these error flags -->
  * myField.$error = { minlength : true, required : true };
  * ```
+ * The `required` message will be displayed first to the user since it appears before the `minlength` message in the DOM.
```
Narretz: This is actually a bad example ( and has been before your changes), because the minlength error is not set when nothing is entered into input. email + minlength is probably a better idea.

Narretz:
The idea to make this clear is good, but atm the changes need some work. I particularly don't like the wording "container" since it makes it sound like it's a different element than the `ngMessages` element.

benelliott:
Thanks for the feedback, please take a look at the changes and let me know. I'll try to fix the CLA again in the meantime...

benelliott:
I've signed it... again....

src/ngMessages/messages.js
```diff
@@ -24,45 +24,64 @@ var jqLite = angular.element;
...
  * ```
  *
- * Now whatever key/value entries are present within the provided object (in this case `$error`) then
- * the ngMessages directive will render the inner first ngMessage directive (depending if the key values
- * match the attribute value present on each ngMessage directive). In other words, if your errors
- * object contains the following data:
+ * In order to show error messages corresponding to `myField` we first create an element with an `ngMessages` attribute
+ * set to the `$error` object owned by the `myField` input in our `myForm` form.
+ *
+ * Within this element we then create separate elements for each of the possible errors that the user's submission for
```
Narretz: that the user's submission for  `myField` could have -> that `myField` could have

src/ngMessages/messages.js
```diff
@@ -24,45 +24,64 @@ var jqLite = angular.element;
...
+ * `myField` could have. The `ngMessage` attribute is used to declare which element(s) will appear for which error
+ * - for example, setting `ng-message="required"` specifies that this particular element should be displayed when there
+ * is no value present for the required field `myField` (because the key `required` will be `true` in the object
+ * `myForm.myField.$error`).
+ *
+ * By default, `ngMessages` will only display one message for a particular key/value collection at any time. If more
```
Narretz: "If more than one message is applicable then the decision of which message to show is" -> "If more than one message (or error) key is currently true, then which element is shown is"

src/ngMessages/messages.js
```diff
@@ -24,45 +24,64 @@ var jqLite = angular.element;
...
+ * than one message is applicable then the decision of which message to show is determined by the order of messages in
+ * the HTML template code (messages declared first are prioritised). This mechanism means the developer does not have to
+ * prioritise messages using custom JavaScript code.
+ *
+ * Given the following error object for our example (which informs us that the field `myField` is currently erroneous
+ * in terms of both its `required` and `email` atrributes):
```
Narretz: atrributes -> attributes

src/ngMessages/messages.js
```diff
@@ -24,45 +24,64 @@ var jqLite = angular.element;
...
+ * - for example, setting `ng-message="required"` specifies that this particular element should be displayed when there
+ * is no value present for the required field `myField` (because the key `required` will be `true` in the object
+ * `myForm.myField.$error`).
+ *
+ * By default, `ngMessages` will only display one message for a particular key/value collection at any time. If more
+ * than one message is applicable then the decision of which message to show is determined by the order of messages in
+ * the HTML template code (messages declared first are prioritised). This mechanism means the developer does not have to
+ * prioritise messages using custom JavaScript code.
+ *
+ * Given the following error object for our example (which informs us that the field `myField` is currently erroneous
```
Narretz: currently erroneous in terms of both its `required` and `email` atrributes -> "currently has both the `required` and `email` errors"

src/ngMessages/messages.js
```diff
@@ -24,45 +24,64 @@ var jqLite = angular.element;
  * `ngMessage` and `ngMessageExp` directives.
  *
  * # Usage
- * The `ngMessages` directive listens on a key/value collection which is set on the ngMessages attribute.
- * Since the {@link ngModel ngModel} directive exposes an `$error` object, this error object can be
- * used with `ngMessages` to display control error messages in an easier way than with just regular angular
- * template directives.
+ * The `ngMessages` directive allows individual keys in a key/value collection to each be associated with an element
```
Narretz: "individual" and "each" can be dropped. 
I would also say "associated with a child element"

src/ngMessages/messages.js
```diff
@@ -24,45 +24,64 @@ var jqLite = angular.element;
...
+ * (or 'message') that will show or hide based on the truthiness of that key's value in the collection. A common use
+ * case for `ngMessages` is for annotating form inputs with error messages using the `$error` object exposed by the
```
Narretz: "is for annotating form inputs with error message" -> is to display error messages for inputs

Narretz:
I've left a few more comments to make the text a little less wordy and verbose, otherwise it looks good.
It also looks like you committed the last commit with another email address and that's why the  CLA is not okay again,

benelliott:
I tried to fix the email address issue by applying the script at https://help.github.com/articles/changing-author-info/ but it's broken the entire commit history. You'll have to bear with me while I fix this!

benelliott:
I have closed this PR as it was a bit of a mess, please see #14103 for further discussion.

PR #14059, opened 2016-02-16
docs(guide/Components): Added missed fieldType integration

AlexanderTserkovniy:
In the file `name="editableField.html"` was missed the `fieldType` property integration.

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -282,7 +282,7 @@ it upwards to the heroList component, which updates the original data.
 </file>
 <file name="editableField.html">
   <span ng-switch="$ctrl.editMode">
-    <input ng-switch-when="true" type="text" ng-model="$ctrl.fieldValue">
+    <input ng-switch-when="true" type="{{::$ctrl.fieldType}}" ng-model="$ctrl.fieldValue">
     <span ng-switch-default>{{$ctrl.fieldValue}}</span>
   </span>
   <button ng-click="$ctrl.handleModeChange()">{{$ctrl.editMode ? 'Save' : 'Edit'}}</button>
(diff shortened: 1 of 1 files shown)


AlexanderTserkovniy:
I signed it!

docs/content/guide/component.ngdoc
```diff
@@ -282,7 +282,7 @@ it upwards to the heroList component, which updates the original data.
 </file>
 <file name="editableField.html">
   <span ng-switch="$ctrl.editMode">
-    <input ng-switch-when="true" type="text" ng-model="$ctrl.fieldValue">
+    <input ng-switch-when="true" type="{{::$ctrl.fieldType}}" ng-model="$ctrl.fieldValue">
```
gkalpak: I'd remove the `::` (so we don't introduce another concept and because there is nothing preventing the `fieldType` to change dynamically :smiley:)

AlexanderTserkovniy: Hmm, is not it the best practice for performance?
I mean in the example the `fieldType` is not changed dynamically so it makes it more transparent, what is wanted to be achieved.
Finally, for the `::` I can add a note, with link to read about it.

gkalpak: > Hmm, is not it the best practice for performance?

It is, but only when the usecase allows it. This is avery simple demo, of course, but in a real up, having a very generic, reusable `editableField` you'd probably want to allow consumers to change the field type dynamically.

In any case, the main purpose of the is to showcase components and their usage and related use-patterns and to that end I believe it's best to keep other concepts to a minimum. I wouldn't oppose if `::` was aready used elsewhere (e.g. I'm sure there are places where is this guide where it is much more appropriate - but this is not the point), but introducing it here unnecessarily introduces yet another concept.

So, I would prefer it gone (but it's not super important, so feel free to leave it as is if you feel strongly about it :smiley:)

AlexanderTserkovniy: Ahaha, thanks! I will remove it. Because I like to keep everything clean and easy :)

gkalpak:
Backported to `v1.5.x` as 2ea7b06.

PR #14064, opened 2016-02-17
fix(copy): add support for copying `Blob` objects

### LINKED ISSUE #9669: $httpBackend cannot mock binary blob responses at IE11 (InvalidStateError)
I found this issue running tests with karma at IE11 / Win8.1.

When i try to mock a blob response with `$httpBackend.expect`, i got `InvalidStateError`.

I dig and found that this happens because of the `angular.copy` which clones fake response data.

If i hack `angular.copy` with something like this the error stops.

```js
if (isBlob(source)) {
    destination = new Blob([source]);
} else if (isArray(source)) {
```

gkalpak:
Although `copy()` does not need to (and never will) support all kinds of objects, there is a (not uncommon) usecase for supporting `Blob` objects:

`ngMock`'s `$httpBackend` will return a copy of the response data (so that changes in one test won't affect others). Since returning `Blob` objects in response to HTTP requests is a valid usecase and since `ngMocks`'s `$httpBackend` will use `copy()` to create a copy of that data, it is reasonable to support `Blob` objects.
(I didn't run any benchmarks, but the additional check for the type of the copied element should have negligible impact, compared to the other stuff that `copy()` is doing.)

Fixes #9669

### DIFF
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -228,6 +228,18 @@ describe('angular', function() {
       }
     });
 
+    it('should handle Blob objects', function() {
+      if (typeof Blob !== 'undefined') {
+        var src = new Blob(['foo'], {type: 'bar'});
+        var dst = copy(src);
+
+        expect(dst).not.toBe(src);
+        expect(dst.size).toBe(3);
+        expect(dst.type).toBe('bar');
+        expect(isBlob(dst)).toBe(true);
+      }
+    });
+
     it("should throw an exception if a Uint8Array is the destination", function() {
       if (typeof Uint8Array !== 'undefined') {
         var src = new Uint8Array();
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -1012,7 +1012,7 @@ describe('ngMock', function() {
 
 
     it('should be able to handle Blobs as mock data', function() {
-      if (typeof Blob !== 'udefined') {
+      if (typeof Blob !== 'undefined') {
         var mockBlob = new Blob(['{"foo":"bar"}'], {type: 'application/json'});
 
         hb.when('GET', '/url1').respond(200, mockBlob, {});
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -1011,6 +1011,26 @@ describe('ngMock', function() {
     });
 
 
+    it('should be able to handle Blobs as mock data', function() {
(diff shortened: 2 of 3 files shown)


test/AngularSpec.js
```diff
@@ -228,6 +228,18 @@ describe('angular', function() {
...
 
+    it('should handle Blob objects', function() {
+      if (typeof Blob !== 'undefined') {
+        var src = new Blob(['foo'], {type: 'bar'});
+        var dst = copy(src);
+
+        expect(dst).not.toBe(src);
+        expect(dst.size).toBe(3);
+        expect(dst.type).toBe('bar');
+        expect(isBlob(dst)).toBe(true);
```
jurko-gospodnetic: Have you checked if this test perhaps passes and gives a false positive without the applied `copyType()` patch? I seem to recall when running into this issue first that the cloned object was getting reported as a Blob even though it was not actually one, and I could only detect the difference by attempting to make a call to one of its specific functions like `slice()` or by attempting to read the blob's content by using the `FileReader` interface (either of which then failed in some internal implementation, in different ways depending on the browser used).

jurko-gospodnetic:
Approved. Tested it, and it works as advertised for me.  :+1:

Not sure how you report review approvals here. :smile:

Any chance this can be back-ported to the 1.4.x branch?

lgalfaso:
LGTM

test/AngularSpec.js
gkalpak: BTW, I don't think `typeof` returns `Blob` for blobs. In latest Chrome and Firefox at least, it return `object`.
(Maybe what you had in mind is that `expect(typeof dst).toBe(typeof src)` would indeed pass even without the fix (because both would return `object`).)

test/ngMock/angular-mocksSpec.js
```diff
@@ -1011,6 +1011,26 @@ describe('ngMock', function() {
     });
 
 
+    it('should be able to handle Blobs as mock data', function() {
+      if (typeof Blob !== 'udefined') {
```
gkalpak: `udefined` (lol)

test/AngularSpec.js
jurko-gospodnetic: I'm fairly certain I got Blob with PhantomJS. Don't recall what actual user browsers gave me there. But it does not matter in the end. Should work like a charm now. :smiley_cat:

gkalpak:
Backported to `v1.5.x` (e9d579b) and `v1.4.x` (863a423).

PR #14065, opened 2016-02-17
fix(ngAria): bind to `keydown` instead of `keypress` in `ngClick`

### LINKED ISSUE #14063: fix(ngAria): bind to `keydown` instead of `keypress` in `ngClick`
`ngAria`'s `ngClick` (among other things) tries to emulate the behavior of native buttons for non-button elements (e.g. `<div>`, `<li>` etc) used as button. Using native buttons is strongly recommended, but we should still try to make things accessible, even if the developer chooses to not follow the best practice :smiley:

Currently, we detect "clicks" by listening to `keypress` events. Switching to `keydown` instead, would make the behavior more consistent with how the browser fires click events on native buttons.

Also, discussed in https://github.com/angular/angular.js/pull/13996/files#r52934300.

LeeAdcock:
The ngAria keyboard support for html elements used as button now binds to the keydown
event instead of keypress to better emulate behavior of native buttons.

BREAKING CHANGE: The ngAria configuration flag that controls binding to keyboard input
has been renamed from `bindKeypress` to `bindKeydown` so that it accurately describes the
implemented behavior.

Closes #14063

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -114,7 +114,7 @@ function $AriaProvider() {
    *  - **ariaValue** – `{boolean}` – Enables/disables aria-valuemin, aria-valuemax and aria-valuenow tags
    *  - **tabindex** – `{boolean}` – Enables/disables tabindex tags
    *  - **bindKeydown** – `{boolean}` – Enables/disables keydown event binding on `div` and
-   *    `li;` elements with ng-click
+   *    `li` elements with ng-click
    *  - **bindRoleForClick** – `{boolean}` – Adds role=button to non-interactive elements like `div`
    *    using ng-click, making them more accessible to users of assistive technologies
    *
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -96,7 +96,7 @@ function $AriaProvider() {
     ariaInvalid: true,
     ariaValue: true,
     tabindex: true,
-    bindKeypress: true,
+    bindKeydown: true,
     bindRoleForClick: true
   };
 
@@ -113,8 +113,8 @@ function $AriaProvider() {
    *  - **ariaInvalid** – `{boolean}` – Enables/disables aria-invalid tags
    *  - **ariaValue** – `{boolean}` – Enables/disables aria-valuemin, aria-valuemax and aria-valuenow tags
    *  - **tabindex** – `{boolean}` – Enables/disables tabindex tags
-   *  - **bindKeypress** – `{boolean}` – Enables/disables keypress event binding on `div` and
-   *    `li` elements with ng-click
+   *  - **bindKeydown** – `{boolean}` – Enables/disables keydown event binding on `div` and
+   *    `li;` elements with ng-click
    *  - **bindRoleForClick** – `{boolean}` – Adds role=button to non-interactive elements like `div`
    *    using ng-click, making them more accessible to users of assistive technologies
    *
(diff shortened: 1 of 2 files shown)


test/ngAria/ariaSpec.js
```diff
@@ -630,32 +630,18 @@ describe('$aria', function() {
...
-      divElement.triggerHandler({type: 'keypress', which: 32});
-      liElement.triggerHandler({type: 'keypress', which: 32});
+      divElement.triggerHandler({type: 'keydown', which: 32});
+      liElement.triggerHandler({type: 'keydown', which: 32});
 
       expect(clickFn).toHaveBeenCalledWith('div');
       expect(clickFn).toHaveBeenCalledWith('li');
     });
 
-    it('should not override existing ng-keypress', function() {
```
gkalpak: Don't remove this test, just change it to use `keydown` instead.

src/ngAria/aria.js
```diff
@@ -364,8 +364,8 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             elem.attr('tabindex', 0);
           }
 
-          if ($aria.config('bindKeypress') && !attr.ngKeypress) {
-            elem.on('keypress', function(event) {
+          if ($aria.config('bindKeydown') && !attr.ngKeypress) {
```
gkalpak: `attr.ngKeypress` should be also changed to `attr.ngKeydown`.
I wonder if it's a good idea to check for both (`ngKeydow` and `ngKeypress`).

LeeAdcock:
Thank you for your feedback @gkalpak, I've implemented changes as you suggested.

gkalpak:
There is one remaining accurrence of `keypress` that needs to be changed to `keydown` in [here][1].

--
And I still wonder if we should leave not do anything if the element does already have an `ngKeydown`/`ngKeypress`/`ngKeyup` (instead of only checking for `ngKeydown`). I mean, if the developer has placed both `ngClick` and one of the keyboard event handling directives on the element, they are clearly handling the keyboard interaction themselves, right.

I believe it would be best to leave them alone in that case, since they seem to know what they're doing.
WDYT @marcysutton ?

[1]: https://github.com/LeeAdcock/angular.js/blob/ngclick-keydown-14063/src/ngAria/aria.js#L35

LeeAdcock:
The more I think about it, the more I agree with you.  It's clear the developer already has specific behavior in mind, that they have already implemented.   Also, our willingness in this PR to switch from one key event to another signifies our opinion that in this context, they are not significantly different. It seems reasonable that we would not bind to key events if the user has already defined customer key event behavior.

LeeAdcock:
@gkalpak any other feedback here?

gkalpak:
@LeeAdcock, I didn't have time to take a look yet - will do in the next days.

marcysutton:
Looks good to me other than the [documentation comment](https://github.com/LeeAdcock/angular.js/blob/ngclick-keydown-14063/src/ngAria/aria.js#L35) @gkalpak pointed out. I think it should just say "keydown event" since that's all it wires up–not keydown and keyup.

gkalpak:
I would change `bindKeyEvents` to `bindKeydown` and the cos to just mention `keydown event` (because that's what we are binding to), but other than that LGTM (with a proper BC notice).

gkalpak:
I made the necessary changes and merged :tada:
Thx @LeeAdcock for working on this!

PR #14070, opened 2016-02-17
fix($compile): handle boolean attributes in `@` bindings

jannic:
Commit db5e0ff handles initial values of boolean attributes. The same
change needs to be applied inside the attrs.$observe() call.

(I'm not familiar with the angular.js codebase, so this may not be the right way to fix the issue. What I verified is that this fixes a regression with some production code which worked with angular.js 1.2.28 and 1.3.20 but failed with 1.4.7.
So feel free to replace this PR with some better fix, as you like.)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3120,7 +3120,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               destination[scopeName] = attrs[attrName] = void 0;
             }
             attrs.$observe(attrName, function(value) {
-              if (isString(value)) {
+              if (isString(value) || isBoolean(value)) {
                 var oldValue = destination[scopeName];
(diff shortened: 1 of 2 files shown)


gkalpak:
Good catch :+1: (This has been on my TODO list for a while :smiley:)
There are still a couple of things to do, before we can merge this:

1. Sign the CLA.
2. Add one or more tests, that verifies the fix (and prevents future regressions).

<sub>Feel free to ping me when/if you make any changes, so I can take another look :wink:</sub>

jannic:
@gkalpak just signed the CLA a few minutes ago. A test was already added back in February.

jannic:
Rebased to current master and manually resolved conflicts with 874c0fdc

PR #14071, opened 2016-02-17
perf($animate): listen for document visibility changes

### LINKED ISSUE #14066: Slow ng-repeat because of AnimateRunner even without ngAnimate
I'm trying to print a lot of elements to the DOM, but it is painfully slow.
However, the reason is caused by AnimateRunner and nothing else.

_Note: I do not want to filter the elements further, but rather get to the bottom of why an un-needed component is stalling the entire browser_

I have disabled ngAnimate, the page has no animations, yet the slowest part of the whole process is it's complete function.

Please see the profile below

Your suggestions are greatly appreciated.

Narretz:
Accessing the document for the hidden state is costly for
platforms like Electron.

Closes #14066


src/ng/animateRunner.js
```diff
@@ -28,8 +28,8 @@ var $$AnimateAsyncRunFactoryProvider = function() {
 };
 
 var $$AnimateRunnerFactoryProvider = function() {
-  this.$get = ['$q', '$sniffer', '$$animateAsyncRun', '$document', '$timeout',
-       function($q,   $sniffer,   $$animateAsyncRun,   $document,   $timeout) {
+  this.$get = ['$q', '$sniffer', '$$animateAsyncRun', '$document', '$$isDocumentHidden', '$timeout',
+       function($q,   $sniffer,   $$animateAsyncRun,   $document, $$isDocumentHidden, $timeout) {
```
gkalpak: `$document` is not needed anymore. (You can also align the arguments for consistency.)

src/ng/animateRunner.js
```diff
@@ -85,7 +85,7 @@ var $$AnimateRunnerFactoryProvider = function() {
 
         // the document may not be ready or attached
         // to the module for some internal tests
-        if (doc && doc.hidden) {
+        if ($$isDocumentHidden()) {
```
gkalpak: Instead of doing this in every tick (and for every runner), coudn't we listen for an event and update the function as necessary ?

src/ngAnimate/animateQueue.js
```diff
@@ -96,9 +96,9 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
   });
 
   this.$get = ['$$rAF', '$rootScope', '$rootElement', '$document', '$$HashMap',
-               '$$animation', '$$AnimateRunner', '$templateRequest', '$$jqLite', '$$forceReflow',
+               '$$animation', '$$AnimateRunner', '$templateRequest', '$$jqLite', '$$forceReflow', '$$isDocumentHidden',
        function($$rAF,   $rootScope,   $rootElement,   $document,   $$HashMap,
-                $$animation,   $$AnimateRunner,   $templateRequest,   $$jqLite,   $$forceReflow) {
+                $$animation,   $$AnimateRunner,   $templateRequest,   $$jqLite,   $$forceReflow, $$isDocumentHidden) {
```
gkalpak: Let's be consistent and align the arguments :grin:

ArekSredzki:
@Narretz These changes should also be propagated to ngAnimate as the issue continues to exist if it is enabled.

src/ng/animateRunner.js
Narretz: tick is only called when the animation completes, so only once per runner. I'm not sure how costly one function call is. Event listener is also possible. but I don't want to add another one over the one we have in $$isDocumentHidden. So maybe we could move the actual async tick into this service where it will be updated by the listener. Wdyt?

gkalpak: Ah, I confused this with the `$$rAF`'s tick. Since it's indeed called one, I guess it's fine as it is.

Narretz:
@ArekSredzki the runner is the same with or without ngAnimate included, and the only other playe where hidden is checked is the queue, so this should be covered

Narretz:
It looks like the test errors are because adding the event listener populates the jqlite cache. Maybe this causes another memory leak.

ArekSredzki:
@Narretz I had checked it before you amended your commit and the problem was still very apparent with ngAnimate included... please check now, or I will on monday :)

Narretz:
I've only changed a few cosmetic things in the ammended commit.
Like I said, these are the only two instances where document.hidden is accessed. With ngAnimate included, the whole process is slow (even without animations) for different reasons. (Mainly it needs to walk the DOM to the top to see if animations are possible / allowed). I assume this is what you are seeing. Or are you seeing he exact same functions being slow?

ArekSredzki:
@Narretz Sorry for the sluggish response. You are correct, and I believe that this solves the issue!
When do you think it can be released?

Thanks!

test/ng/compileSpec.js
```diff
@@ -7245,22 +7245,22 @@ describe('$compile', function() {
         });
 
         inject(function($compile, $rootScope) {
-          expect(jqLiteCacheSize()).toEqual(0);
+          var cacheSize = jqLiteCacheSize();
```
mgol: What's the reason for this change?

Narretz: This and a few other tests assume that the cacheSize is 0 after bootstrap. The listener for visibilitychange on the document adds a cache entry. So I thought it makes more sense to not assume a specific cacheSize on bootstrap.

mgol: OK, makes sense.

mgol:
LGTM.

Narretz:
@mgol I had to change the expects for the add/removeEventListener calls to account for the fact that jquery 2.1 always calls them with the third argument set to false. Is this a functional change in jquery 2.1?

mgol:
jQuery has always passed the 3rd argument (`useCapture`) to `addEventListener` because it used to be required in Firefox some time ago. jQuery 1.12 & 2.2 have stopped passing this argument as it's not been needed for a long time (IE9+ don't require it).

The way jQuery calls addEventListener is not its public API so we were free to change it even in a patch release. I adjusted jqLite code to do the same but only on `master`, `v1.5.x` continues to pass `useCapture` just in case. You'll need to adjust the expectations when cherry-picking to `v1.5.x`.

Narretz:
Thanks for the explanation! I've adjusted the tests to simply check the first two arguments and ignore the third.

Issue #14030, opened 2016-02-13
Implement lifecycle hooks for directive/component controllers (ng2-style)


gkalpak:
I open this issue in order to track/discuss the implementation of lifecucle hooks to directive/component controllers.

--
### Context

In our attempt to improve the upgrade path to Angular 2, it makes sense to allow developers structure their directive/component controllers is a way that is as easy as possible to migrate them to Angular 2 (ideally, there should be no modification of the controllers code necessary).
As part of this attempt, it will be useful to implement some of the [lifecycle hooks available in Angular 2][1]. (Obviously, it isn't necessary (or possible) to implement all of them, due to differences between ng1/ng2 internals.)

--

### Lifecycle Hooks

- [x] `OnInit`: Implemented in `v1.5.0` as `$onInit`.
- [ ] `OnDestroy`: Useful (and probably straightforward to implement). Discussed in #14020.
- [ ] ~~`DoCheck`~~: Not applicable.
- [ ] `OnChanges`: ~~Not applicable.~~ Applicable.
- [ ] `AfterContentInit`: Not sure how much more useful than `AfterViewInit` (in the context of ng1).
- [ ] ~~`AfterContentChecked`~~: Not applicable.
- [ ] `AfterViewInit`: Useful. Discussed in #13991.
- [ ] ~~`AfterViewChecked`~~: Not applicable.

[1]: https://angular.io/docs/ts/latest/guide/lifecycle-hooks.html

thorn0:
Why is `OnChanges` not applicable?

gkalpak:
I'm not sure what I was thinking at the time. Maybe because it's tied to Inputs/Outputs and ng1 doesn't have the same semantics. It would probably be reasonably feasible to implement for '@' and '<' as inputs and '&' as outputs (with a small overhead), but with two-way binding it gets really muddy.

But I definitely be convinced otherwise. One of the main purposes of opening the issue was to get the discussion going.

thorn0:
I'm not sure if 'OnChanges` works with outputs at all. Even though [the docs](https://angular.io/docs/ts/latest/guide/lifecycle-hooks.html) state otherwise:
> `ngOnChanges` - called when an input or output binding value changes

mikeerickson:
I am unable to get $onChanges to fire regardless what I try.  To further complicate matters, the documentation is VERY vague and there are basically NO examples, etc. online (I am trying to get it working with ES6 modules).

[...]

Issue #13991, opened 2016-02-10
FormController: attach to component controller?


royaldark:
The new component support in 1.5 is great, but it seems to interact badly with FormController. With a very simple component like such:
```js
someModule.component('customComponent', {
  template: '<form name="$ctrl.form">Is form valid? {{ $ctrl.form.$valid }}</form>',
  controller: ['$scope', function($scope) {
    this.$onInit = function() {
      console.log("FormController:");
      console.log(this.form);
    };
  }]
});
```
all seems well on-screen, as it will correctly show `Is form valid? true`. However, the `console.log` in the customComponent controller can't find the FormController that should be bound to it via `name="$ctrl.form"`. I also checked `$scope.form` and `$scope.$ctrl.form` to make sure it wasn't a `this`-binding issue, but no luck.

Here is a Plunker showing the issue in action: http://plnkr.co/edit/P8gKwz6koHeoXRHdhoGl

Am I missing something obvious?

gkalpak:
The only guarantee for the `$onInit` hook is that the **required** controllers will be available. The `FormController` you are trying to access is not required and it uses a different mechanism, so I am afraid you are "stuck" with using a post-link function for grabbing that (or try to access it asynchronously if you know your template doesn't load any partial from the server).

--
Maybe we could add an `$afterViewInit` hook or something, that would be called after the post-linking phadmse.

@petebacondarwin, wdyt ?

jonricaurte:
@gkalpak That would be awesome if you could add an $afterViewInit hook after post linking. I think after seeing more of 1.5, people are going to be requesting more hooks that are included in angular 2 because it is easier to manage your code in the controller vs in a directive especially if you are using ES2015 which make writing elaborate directives more of pain and would ease the transition to angular 2 even further.

Narretz:
The form is also set on the scope - there's no way (yet) to attach it to the controller.

gkalpak:
(Not that it's related to the lifecycle hook, but) it **is** possible by using `name="$ctrl.form"` (as in OP's example), since `$ctrl` is just a reference to the controller on the scope.

The important thing about the hook is that it will be called after everything is set (either on the scope or on the controller), so the user doesn't try to access properties that don't yet exist.

Narretz:
Oh, yes, obviously. Thanks for correcting that.

petebacondarwin:
Sounds interesting. Is that a hook in A2? What are its exact semantics? Does someone want to knock up a PR?

[...]

PR #14080, opened 2016-02-18
feat(ngAsDirective): new as directive to publish component controllers into current scope 

drpicox:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)** Feature.

**What is the current behavior? (You can also link to an open issue here)**: 
With restrict you can reference parent controllers from a children, but if you want to reference children components you need to find them with _jqLike_ and then get the controller.

**What is the new behavior (if this is a feature change)?**: 
It adds a new directive that assigns the controller of the current DOM element component to any expression. If the element is destroyed (ex: it was inside ngIf or ngRepeat) it assigns a null.

``` html
<toggle ng-as="burger"></toggle>
<menu ng-if="burger.isVisible"></menu>
...
<countdown-timer ng-as="$ctrl.countdownTimer"></countdown-timer>
<button ng-click="$ctrl.relaunch()"></button>
```

No.

**Other information**:

I present in this PR the implementation of a directive that I use in almost all my projects (and other projects in which I'm consultant) since the last year.
### Basic use

It is really simple: it publishes the controller of component in the current scope. Ex:

``` html
<toggle ng-as="myToggle"></toggle>
<button ng-click="myToggle.toggle();">Toggle</button>
<p ng-show="myToggle.isOpened()">The toggle is active</p>
```

so it can be accessed from enclosing components. 
Until this point, it is a kind of custom `controllerAs` property for parent scope.
### Using with parent components

In addition, as following the recommendation did by @petebacondarwin in https://github.com/angular/angular.js/issues/10007#issuecomment-168425866 it is able to assign it to any value, event enclosing controller. Ex:

``` html
<toggle ng-as="$ctrl.myToggle"></toggle>
```
### ~~Keep track of DOM elements~~

It also serves to look for exact DOM elements instead of relaying in _jqLite_ searches. In case that there is no a controller, the as directive assigns the current DOM element to the expression set.

``` html
    <br>From <a href="https://developer.mozilla.org/en/docs/Web/HTML/Element/audio">here</a>.<br>
    <player></player>
    <script>
        angular.module('demo').component('player', {
            template: '' +
                '<audio as="$ctrl.audio" src="http://developer.mozilla.org/@api/deki/files/2926/=AudioTest_(1).ogg"></audio>' +
                '<button ng-click="$ctrl.play()">Play</button>' +
                '',
            controller: function() {
                this.play = function() {
                    this.audio.play();
                };
            },
        });
    </script>
```
#### Why?

ngController is great for SEO webpages, but it has the inconvenience that it does not integrates well with bindings and many things have to be done "manually". For example:

``` html
    <script>
        angular
            .module('demo', [])
            .service('booksService', function() {
                this.getBook = function(id) {
                    return { name: 'name'+id, abstract: 'abstract...'+id };
                };
            });
    </script>

    <div ng-controller="BookDetailController as bookDetailCtrl" book-id="1">
        <h1>{{bookDetailCtrl.book.name}}</h1>
        <p>{{bookDetailCtrl.book.abstract}}</p>
    </div>
    <script>
        angular.module('demo').controller('BookDetailController', function(booksService,$attrs) {
            this.book = booksService.getBook($attrs.bookId);
        });
    </script>
```

In this example it needs to access to `$attrs` manually to find the bookId. You can imagine more complex functionalities, but I think that this is just enough to expose.

Now, imagine that we do a directive to handle the same behaviour:

``` html
    <static-book-detail book-id="2">
        <h1>{{bookDetailCtrl.book.name}}</h1>
        <p>{{bookDetailCtrl.book.abstract}}</p>
    </static-book-detail>
    <script>
        angular.module('demo').directive('staticBookDetail', function() {
            return {
                scope: true,
                controller: function(booksService) {
                    this.book = booksService.getBook(this.bookId);
                },
                controllerAs: 'bookDetailCtrl',
                bindToController: {
                    bookId: '@',
                },
            };
            this.book = booksService.getBook($attrs.bookId);
        });
    </script>
```

It has a conceptual problem: you must know by heart that static-book-detail publishes the `bookDetailCtrl`. 

Alternatively you can use a component, which always use `$ctrl` like:

``` html
    <p>Does not works as expected:</p>
    <static-book-component book-id="3">
        <h1>{{$ctrl.book.name}}</h1>
        <p>{{$ctrl.book.abstract}}</p>
    </static-book-component>
    <script>
        angular.module('demo').component('staticBookComponent', {
            bindings: {
                bookId: '@',
            },
            controller: function(booksService) {
                this.book = booksService.getBook(this.bookId);
            },
        });
    </script>
```

but this naive solution does not works because `$ctrl` is not defined in the same scope that code relies.
#### Popossal

Finally, using as you can have this solution that seems cleaner and more easy to understand:

``` html
    <p>It dows works (and no implicit):</p>
    <static-book-component ng-as="$ctrl" book-id="4">
        <h1>{{$ctrl.book.name}}</h1>
        <p>{{$ctrl.book.abstract}}</p>
    </static-book-component>

    <script>
        angular.module('demo').component('staticBookComponent', {
            bindings: {
                bookId: '@',
            },
            controller: function(booksService) {
                this.book = booksService.getBook(this.bookId);
            },
        });
    </script>
```

_Additional Notes_: `ngAs` directive behaviour is very close to **Angular2** `#` template variables. It also copies concepts from _RiotJS_ and _Polymer_ (at least first versions).


test/ng/directive/asSpec.js
```diff
@@ -0,0 +1,87 @@
...
+      var scope = $rootScope.$new();
+      var $ctrl = {undamaged: true};
+      scope.$ctrl = $ctrl;
+      dealoc($compile('<my-component as="$ctrl.myComponent"></my-component>')(scope));
+
+      expect(scope.$ctrl).toBe($ctrl);
+      expect(scope.$ctrl.undamaged).toBe(true);
+      expect($ctrl.myComponent).not.toBeUndefined();
+      expect($ctrl.myComponent && $ctrl.myComponent.property).toBe(expected);
+      scope.$destroy();
```
gkalpak: No need to destroy the scope (or dealoc the element, I believe).

test/ng/directive/asSpec.js
```diff
@@ -0,0 +1,87 @@
...
+        myComponentController = this;
+      }
+    });
+    module('my');
+
+    inject(function($compile, $rootScope) {
+      var scope = $rootScope.$new();
+      dealoc($compile('<my-component as="bar.myComponent"></my-component>')(scope));
+
+      expect(scope.bar).not.toBeUndefined();
```
gkalpak: This is not needed (it's implied by the next statement).

test/ng/directive/asSpec.js
```diff
@@ -0,0 +1,87 @@
...
+      var scope = $rootScope.$new();
+      dealoc($compile('<my-directive as="bar.myDirective"></my-directive>')(scope));
+
+      expect(scope.bar).not.toBeUndefined();
+      expect(scope.bar && scope.bar.myDirective).toBe(myDirectiveController);
+      scope.$destroy();
+    });
+  });
+
+  it('should bind non HTMLElements elements into the current scope if no component', inject(function($compile, $rootScope) {
```
gkalpak: Intead of injecting `$rootScope` and `$compile` every time, it's cleaner to do it in a `beforeEach()` block.

gkalpak: _if no component_  -->  _if no controller_ (?)

src/ng/directive/as.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * This directive allows to access to the controller of components in your
+ * current template or from current controller.
+ *
+ * @example
+ * ### Access to component controller
+ * This example demonstrates basic publishing of the component controller into the scope.
+ * <example name="asDirectiveComponentExample" module="asComponentExample">
+ *   <file name="index.html">
+ *     <script>
```
gkalpak: Why not put this in a different `script.js` file ?

src/ng/directive/as.js
```diff
@@ -0,0 +1,90 @@
...
+ *       '',
+ *     controller: function() {
+ *       this.play = function() {
+ *         this.audio.play();
+ *       };
+ *     }
+ *   });
+ * ```
+ *
+ * Note that angular does not allows to access DOM elements directly
```
gkalpak: This is only true when expensive checks are on, which happens on event-directives (such as `ngClick`), but not on other expressions.
Exposing DOM elements on the scope ia really bad idea and a big security concern, so we should definitely not do this.

src/ng/directive/as.js
```diff
@@ -0,0 +1,90 @@
...
+ * from templates, you need a controller to manipulate them.
+ *
+ */
+var componentAsDirective = ['$parse', function($parse) {
+  return {
+    restrict: 'A',
+    compile: function(tElement, tAttrs) {
+      // gets the theoretical controller name, converts <some-thing> into "someThing")
+      var controllerName = tElement[0].tagName.toLowerCase()
+        .replace(/[^\w]\w/g, function(a) { return a.slice(1).toUpperCase(); });
```
gkalpak: This is not taking all possible cases into account (e.g. whe using `data-/x-` prefixes).
You can use `tAttrs.$normalize()` instead.

src/ng/directive/as.js
```diff
@@ -0,0 +1,90 @@
...
+      var setter = $parse(tAttrs.as).assign;
+
+      return function(scope, iElement, iAttrs) {
+        // gets the controller of the current element (see jqLiteController for details)
+        var controller = iElement.data('$' + controllerName + 'Controller');
+
+        if (controller) {
+          setter(scope, controller);
+        } else {
+          setter(scope, iElement[0]);
```
gkalpak: Really bad idea (see https://github.com/angular/angular.js/pull/14080/files#r53457925).

test/ng/directive/asSpec.js
drpicox: Ops! Right! Each test have a new shiny $rootScope, so not even new required.

About dealoc a I do not have clear if I must add or remove it, I found the following:

- https://github.com/angular/material/issues/2250
- https://github.com/angular/angular.js/blob/1061c56fe159691112bc81ca34573ea9681a7255/test/ng/compileSpec.js#L738-L748

And tests from:

- https://github.com/angular/angular.js/blob/1061c56fe159691112bc81ca34573ea9681a7255/test/ng/compileSpec.js#L10001-L10015
are using the global "element" variable cleaned by:
- https://github.com/angular/angular.js/blob/1061c56fe159691112bc81ca34573ea9681a7255/test/ng/compileSpec.js#L145

But I'm not sure if the main afterEach is executed after the whole describe, or after each piece. Or may be it is a bug.

But if you are confident that it is not necessary I'll remove it.

src/ng/directive/as.js
drpicox: Ok.
In the beggining I never did that, but NG2 does something like this (it is not scope, but binds DOMElements to Controllers), and it is very handy: in the template you know how many elements you have, where are them, how they interact, and basically you attach them to the controller. Security issues are controlled by parser.
Anyway, the nice alternative is to make nice small directives to wrap native elements like audio/video/... (so there is only one element to jqLite find) and then attach their controllers.

gkalpak:
This is interesting, but still needs some work imo.

A few thoughts:

1. It seems quite similar to ng2's `var`/`#` for defining template-local variables.
    Maybe it's a good idea to rename it to `var`, maybe it isn't (because it's not exactly the same).

2. It's not a good idea to expose DOM elements in the template.
    (It's different in ng2, because there is better isolation there - no scope hierarchy).

3. It doesn't clean up the scope when the component is destroyed/removed.

src/ng/directive/as.js
drpicox: Ok, I'll correct it. It seems right.

El dv, 19 febr 2016 a les 13:59 Georgios Kalpakas <notifications@github.com>
va escriure:

src/ng/directive/as.js
gkalpak: > NG2 does something like this (it is not scope, but binds DOMElements to Controllers)

Angular 2 has better isolation, so it's not as much of a concern. Here we are attaching it on scope and there is the scope hierarchy, meaning that malicious code could reach the DOM element from virtually anyware on the page.
So, we don't have a way to have "template-local" variables as in ng2.

--
> Security issues are controlled by parser

Only when expensive checks are on, but we can't controller in what kind of expressions the malicious code would end up to. So, the parser doesn't have our back here.

--
> the nice alternative is to make nice small directives to wrap native elements like audio/video/

Sounds about right :smiley:

test/ng/directive/asSpec.js
drpicox: It has a problem. It is modifying the module definition in each test
(creating a special directive case in each test). So it can inject nothing
before module definition. I copied this solution from .component testing.
(I do not put the example because I'm on mobile now)

El dv, 19 febr 2016 a les 13:55 Georgios Kalpakas <notifications@github.com>
va escriure:

src/ng/directive/as.js
drpicox: Ok, I'll add the test and fix it.

El dv, 19 febr 2016 a les 14:08 Georgios Kalpakas <notifications@github.com>
va escriure:

test/ng/directive/asSpec.js
gkalpak: No, I'm not confident, that's why I wrote _"(or dealoc the element, I believe)"_. TBH, when `dealoc` is needed and when it isn't is still a big mystery for me :).

gkalpak: After some investigation it seems that `testabilityPatch` applies an `afterEach` to all tests, that automatically cleans up `$rootElement` and `$rootScope`. This means that you don't have to manually call `dealoc` if at least one of the following conditions holds:

1. Any elements you create are in the DOM tree of the `$rootElement` (e.g. you call `$rootElement.append(yourCreatedElement)`).
2. No other scope than `$rootScope` is sued for linking.

E.g. in this test, if you keep using `$rootScope.$new()`, then you need to `dealoc` (either the element or the scope) - yup, scopes can be `dealoc`'d too :smiley:
If you switch to using `$rootScope`, then I think you don't need `dealoc`ing.

The good thing is that if you don't clean up properly, you'll get some warnings in the console, so you know something's wrong.

test/ng/directive/asSpec.js
gkalpak: You are right, I missed that. BTW, you don't need to create a new module with `angular.module(...)`. (The `.component()` tests do it for different reasons.)

You might be able to simplify the tests by creating the component in a `beforeEach()` block, but it's unimportant at this point. It's fine as it is.

drpicox:
Thanks for aaaaalllll the feedback!

About thoughts:

1. **Naming**:
    - I originally named it `id` (like old polymer versions or riotjs), but I found that some of my colleagues found it confusing (in addition to any problems with the actual meaning of id attribute)
    - I tried `var` for a while, as you suggested, but it semantically seems to mean a new variable inside the scope, not accessible from controller. I tried variants `prop` , `let` , ... and at the end it was more confusing. What happens when I want to use it inside a ngRepeat, what happens when I want to cross a scope and register it into the controller or "_main template scope_", ...
   - Finally I started to use the `as` notation, my colleagues found it very straightforward, and it seems to go in the same direction that `controllerAs`. In addition, as explained before, thanks to https://github.com/angular/angular.js/issues/10007#issuecomment-168425866 I created a explicit version which fits to all cases: set into controller, local variable inside ngRepeat, and so on.

2. **DOMElements**: removed, you're right endeed.

3. **Destroyed/Removed**: Ops! Thanks again! Never found the case before, but you're completely right. ¿What happens when the element is destroyed? (for example inside an ngIf) If it is recreated again it is all right, but if it is removed, it stops working properly. I'll add the test case and then nullify the controller assignation.

Thanks a lot for your feedback! :smile:

petebacondarwin:
Although I really quite like the idea, and I can see a number of potentially nice uses for it, such as comparison validators, such as password confirmation, I am really nervous about how easy it would be to cause memory leaks, if one of these references was held on to after the DOM element containing the original value was destroyed.

I think that Angular 2 can get away with this because its vars are only available to the containing template. Is that right or do they have additional mechanisms in place?

drpicox:
### Nullification

I have already applied the nullification of the controller saved value when its component is destroyed (it overwrites the previously published controller object by null). So it does not matter if any parent scope holds the object where it is assigned, it is nullified. Beyond that point, I believe that if it is misused it should be a programmers fault. From my point of view, it is like using `require`, you can register the child controller into the parent directive: and you can mess it up if you want.

An excerpt from the test (notice that it crosses ng-if scope):

```javascript
it('should nullify the variable once the component is destroyed', function() {
  $rootScope.$ctrl = {};
  $compile('<div ng-if="!nullified"><my-component as="$ctrl.myComponent"></my-component></div>')($rootScope);
  $rootScope.$apply('nullified = false');
  expect($rootScope.$ctrl.myComponent).toBe(myComponentController);

  $rootScope.$apply('nullified = true');
  expect($rootScope.$ctrl.myComponent).toBe(null);
});
```

Anyway it is a good idea to think a little bit more about possible memory leaks.


### `as` a new relationship direction

The idea of the `as` directive is to be the complement of `require`. It gives the opposite direction of creating relationships between directives. Without it, some implementations are really weird.

One typical use of the `as` directive is for graphic designers, they can add the logic for toggles, accordions, and so one directly in the template.

```html
<toggle as="myToggle"></toggle>
<button ng-click="myToggle.toggle();">Toggle</button>
<p ng-show="myToggle.isOpened()">The toggle is active</p>
```

The other use, which makes more sense that just a helper for graphic designers, is to enable interaction with children directives, in short: call children methods. 
Interactions from parent directives to children directives are through bindings and changing binding values, and children directives can interact with parent directives are:
- modifying `'='` binding values,
- invoking `'&'` methods,
- through `require` which allows to call to parent directive methods.

So the overall idea is enable parent controllers to call children directive methods, and use existing directives from the template itself.


### "_Deprecation_" of ngController and non-template directives & transclusion

One more strong point is the possibility to express in a easy to understand way when the controller of a directive is exposing its interface. _Note: I'm not proposing deprecate ngController, but directives are more flexible some times_.

In this case you know that `staticBookDetailCtrl` is referring to `static-book-detail` because the name matches (it is a kind of _implicit_ value). But beyond that, if no code is available, you should not notice that there is a `staticBookDetailCtrl` available.

```html
    <static-book-detail book-id="2">
        <h1>{{staticBookDetailCtrl.book.name}}</h1>
        <p>{{staticBookDetailCtrl.book.abstract}}</p>
    </static-book-detail>
```

But, in this other example:

```html
    <static-book-component as="bookCtrl" book-id="4">
        <h1>{{bookCtrl.book.name}}</h1>
        <p>{{bookCtrl.book.abstract}}</p>
    </static-book-component>
```

It is just straight forward, you know for sure, without doubt, that you have a `bookCtrl` which is the actual controller of the `staticBookComponent`.

In addition, it also has the good point that it does not matter if you change your directive internal semantics about scopes. I explain it:
- the first example (the _implicit_ version) it only works with `scope:true` 
- if you change the `scope` to `{}` it stops working in all places where the directive is being used
- if you decide to use transclusion in the first example (in order to use a template), it will stop working once again.

So, what about static(landing) pages matters using `as` seems a real good strategy.


### Other frameworks

If you take a look to Polymer it uses `id="foo"` + `this.$.foo` to access to child elements. Example:

- https://github.com/PolymerElements/iron-ajax

```
<iron-request id="xhr"></iron-request>
...
this.$.xhr.send({url: url, params: params});
```

Some time ago I've seen that riot.js have been doing the same, but I cannot find it now.

VictorDeBlas:
+1

Good work!

petebacondarwin:
OK I think we are going the right direction here... Let me review properly later this week

dcherman:
As a slight variation on this, what if we were to allow an expression rather than a simple property binding?

I find that a registration/de-registration pattern is very common in components that I wrote in order to work around ngRepeat/ngIf/other asynchronous behaviors where child directives come in at a later date

As a result, having this be an expression could simplify those use cases while still allowing this one:

```js
as="myToggle = $child"
```

However a registration function could also be invoked

```js
as="toggleDidArrive($child)"
```

The latter would not be possible without watching the property.

When we figure out the logic for the `$onDestroy` functionality in controllers, we could implement the inverse and get the full pattern.

gkalpak:
> I find that a registration/de-registration pattern is very common in components that I wrote in order to work around ngRepeat/ngIf/other asynchronous behaviors where child directives come in at a later date

I don't see why this will be an issue with the `as` directive. `as` will only expose the controller of the element it is on, so we don't care about children being initialized. And if it's on a `transclude: 'element'` directive, it will only get compiled when the actual element (not the comment) is inserted into the DOM.

drpicox:
I'll add some extra few tests later:

- what happens if it is in the same place that an ngIf (ngIf is
transcluding the element)?
- check that it works well with $onInit.

drpicox:
I have already done some tests about.

After reading the last @dcherman suggestion it seems that there can be some problems with `$destroy` when transcluding type `element`. In one hand as directive has no transclusion, so, itself it is not affected by this, in the other hand, I added the edge case in which the same component has an ngIf and an as:

```javascript
    it('should nullify be compatible with $element transclusion', function() {
      $rootScope.$ctrl = {};
      $compile('<my-component as="$ctrl.myComponent" ng-if="!nullified"></my-component>')($rootScope);

      $rootScope.$apply('nullified = true');
      expect($rootScope.$ctrl.myComponent).toBeUndefined();
      $rootScope.$apply('nullified = false');
      expect($rootScope.$ctrl.myComponent).toBe(myComponentController);
      $rootScope.$apply('nullified = true');
      expect($rootScope.$ctrl.myComponent).toBe(null);
    });
```

It work well as expected.

In the other hand, one of my claims about this directive is the possibility to replace ngController and supporting the case of transclussio, but there was no tests about it, I added them.

I have also considered @dcherman to transform the `as` directive into some kind of `ngInit`. I do not like the idea: the name seems to deceive (it is not the behavior of as in other contexts), and we can combine it with ngInit to achieve the same result. So considering this I tried:

  - Using `$onInit`: I thought that it was executed after the building of the component was completed, it is not, it is executed before compiling the component template. So it is not useful. May be the behavior of `$onInit` can be change to solve this, or may be we can add another callback like `$onReady` 

  - Using `ngInit`: It should be possible to use ngInit to trigger the initialization callback. Unfortunately it does not work well with the current implementation:

  ```html
  <!-- is failing -->
  <my-component as="myComp" ng-init="$ctrl.componentDidArrive(myComp)"></my-component>
  ```

  Because ngInit executes in pre-link with priority:450. I have tested (not committed) asDirective to work in pre-link with priority 475 and then it works. An alternative may be:

  ```html
  <my-component1 as="$ctrl.myComp1"></my-component1>
  <my-component2 as="$ctrl.myComp2"></my-component2>
  <div ng-init="$ctrl.$onReady()"></div>
  ```

May be it makes sense to change `asDirective` to work in pre-link with a priority compatible with ngInit. Both are assignation value directives. Or may be, not changing it, we can avoid dynamic registering inside a ngRepeat in which elements are added but not removed (in this case I think that `require` is safer because you know your parent)

  ```html
  <!-- this is dangerous if compatible with ngInit -->
  <my-component as="myComp" ng-init="$ctrl.add(myComponent)" ng-repeat="item in items"></my-component>
  <button ng-click="items.push('new')">Add</button>
  <button ng-click="items.shift()">Remove</button>
  ```

About `$onInit` or `$onReady` may be it is worth to analyze its convenience. Many times, working with components like directives, I implemented a post-link function that just calls $ctrl.activate(), so I can handle things after all template components are properly initialized.

gkalpak:
I still don't get the issues (probably missing something).
@dcherman, could you provide a failing testcase, where the current implementation would fail ?

drpicox:
About the proposal of @dcherman of making `as` behaviour `ngInit` like, I looked to in progress issues regarding `$onInit` and I have found the following:

  - Implement lifecycle hooks for directive/component controllers (ng2-style): https://github.com/angular/angular.js/issues/14030
  
  > it will be useful to implement some of the lifecycle hooks available in Angular 2

  It proposes add two hooks: 

  > AfterContentInit: Not sure how much more useful than AfterViewInit (in the context of ng1).
  > AfterViewInit: Useful. Discussed in #13991.

The case of #13991 seems to be the one closer to this feature, so it seems that `$afterViewInit` should be the hook required to start using child controllers.

drpicox:
Uhmmm, I'm not sure what to do with form directive:

```javascript
// this test passes
it('should be compatible with form', inject(function($compile, $rootScope) {
    $rootScope.$ctrl = {};
    $compile('<form as="$ctrl.form"></form>')($rootScope);
    expect($rootScope.$ctrl.form.$valid).toBe(true);
  }));
```

```javascript
  // this test fails
  it('should be compatible with ngForm', inject(function($compile, $rootScope) {
    $rootScope.$ctrl = {};
    $compile('<ng-form as="$ctrl.form"></ng-form>')($rootScope);
    expect($rootScope.$ctrl.form.$valid).toBe(true);
  }));
``` 

The reason is that ngForm is using an undocumented feature of `.directive` that allows to change the name of the "_directive_": `name: 'form'`. I have looked in all the angular source code and it seems that there are not more examples of this.

I can add this code to `asDirective` inside `compile` function, but it does not feels right for me:

```javascript
      // special case: ngForm is the only known directive that uses an undocumented
      // feature of .directive able to change the name of the directive
      if (controllerName === 'ngForm') {
        controllerName = 'form';
      }
```

I may ignore it because:
- you can always use `name` of form that does the trick, but this is also a kind of exception
- `as` directive is intended only for components
- may modify compile register the controller with both names (which I think that really makes sense, creates a more predictable code)

Or may be look for other solution, like registering entity directive controllers with always the same name, so none of the text transformations are required.

dcherman:
@gkalpak http://jsfiddle.net/rogqp47y/17/ is probably the fiddle you need to check out since it demonstrates the problem with depending on the `$destroy` event on an element.

For the record, I don't believe this to be a very common pattern at all, but I believe in being thorough in addressing edge cases even if the answer ends up being "we don't care about that one".

dcherman:
Also, we should be consistent in whatever decision we make here.  In #14127, it seems as if we may be leaning towards depending on scope destruction as the indicator of "this directive was destroyed".  If that's the case, we should use the same strategy here when when we assign null to the expression.

@petebacondarwin Agree?

src/ng/directive/as.js
```diff
@@ -0,0 +1,75 @@
...
+    restrict: 'A',
+    compile: function(tElement, tAttrs) {
+      // gets the expected controller name, converts <x-some-thing> into "someThing")
+      var controllerName = tAttrs.$normalize(tElement[0].tagName.toLowerCase())
+        .replace(/[^\w]\w/g, function(a) { return a.slice(1).toUpperCase(); });
+
+      // special case: ngForm is the only known directive that uses an undocumented
+      // feature of .directive able to change the name of the directive
+      if (controllerName === 'ngForm') {
+        controllerName = 'form';
```
dcherman: @drpicox It's funny that you special cased form actually; form already has a method of attaching the form controller to a scope property per the documentation.  If we introduce this, we have duplicate ways of accomplishing the same thing; I wonder if we should plan for a deprecation.

src/ng/directive/as.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      // get the setter for the as attribute
+      var setter = $parse(tAttrs.as).assign;
+
+      return function(scope, iElement, iAttrs) {
+        // gets the controller of the current element (see jqLiteController for details)
+        var controller = iElement.data('$' + controllerName + 'Controller');
```
dcherman: This should use the `.controller` method probably

src/ng/directive/as.js
gkalpak: Form's version does more things under the hood, so I think we should keep it.

src/ng/directive/as.js
gkalpak: Off the top of my head `controller()` uses `inheritedData`, which is not what we want.

:+1: for `jqLite.data()`.

drpicox: Exactly, thanks for the tip of jqLite.

drpicox: Ops!
Tests with jqLite are fine, but jQuery tests are not.
```javascript
        // this is failing with test:jquery, working with test:jqLite
        var controller = jqLiteData(iElement[0], '$' + controllerName + 'Controller');
```

Although this is working:
```javascript
       // working in both cases
       var controller = jqLite(iElement[0]).data('$' + controllerName + 'Controller');
```

After some research in the code I found the following:
(see https://github.com/angular/angular.js/blob/4883e957971f1c551a0f560427bba0224f291449/src/ng/compile.js#L2333-L2342 )

```javascript
  // compile.js 
          if (controllerResult !== controller.instance) {
            // If the controller constructor has a return value, overwrite the instance
            // from setupControllers
            $element.data('$' + controllerDirective.name + 'Controller', controllerResult);
            removeControllerBindingWatches && removeControllerBindingWatches();
            removeControllerBindingWatches =
              initializeDirectiveBindings(controllerScope, attrs, controller.instance, bindings, controllerDirective);
          }
```

And the `$element` that it is using is the same `$element` that postLink function seems to be receiving:
( see https://github.com/angular/angular.js/blob/4883e957971f1c551a0f560427bba0224f291449/src/ng/compile.js#L2381-L2391 )

```javascript
        // POSTLINKING
        for (i = postLinkFns.length - 1; i >= 0; i--) {
          linkFn = postLinkFns[i];
          invokeLinkFn(linkFn,
              linkFn.isolateScope ? isolateScope : scope,
              $element,
              attrs,
              linkFn.require && getControllers(linkFn.directiveName, linkFn.require, $element, elementControllers),
              transcludeFn
          );
        }
```

After doing some checking, `jqLite()` is a wrapper for `jQuery()` when jQuery is present (basically I checked the versions and data function):

```javascript
   // if they were different
   var jqElement = jqLite(iElement[0]);
   expect(jqElement.data).not.toBe(iElement.data);
   expect(jqElement.jquery).not.toBe(iElement.jquery);
   // both fails always
```

My conclussion, as I understand, when jQuery is loaded, I guess because of coding coherence, `compile` uses `jQuery.data()` instead of real `jqLite.data()`. If I'm try to get the data from the `jqLiteExpandoStore` it will fail because it was never user to save this information.

So it seems that it would not be changed.

dcherman: Yea, don't use `jqLiteData` because as you found out, that uses the internal jqLite data store at all times.  If instead you use `jqLite.data`, it will use the appropriate storage if jQuery is loaded since jqLite is replaced with jQuery in those cases.

drpicox:
I was checking the result of the last commit and it says:

> NoSuchWindowError: The driver is not focused on a window. You must switch to a window before proceeding. (WARNING: The server did not provide any stacktrace information)

```
Running "protractor:travis" (protractor) task
[launcher] Running 3 instances of WebDriver
......F.FF.FFFFFFFFFF.F.FFFFFFFFFF.FFFFF.FFFF.FFFFFFFFFF.FFFF.F.FFFF.FFFF.FFF.FFFF.FFFF.FFFF.FFF.FFFF.FFF.FFFFF.FFF.FFFFF.FFF.FFFF.FF.FFFFFFFFF.F.FFFFFFFF.FF.FFFFFF.FF.FFFFFFFFFF.FF.FF
------------------------------------
[safari 7 OS X 10.9 #3] PID: 4386
[safari 7 OS X 10.9 #3] Using SauceLabs selenium server at http://ondemand.saucelabs.com:80/wd/hub
[safari 7 OS X 10.9 #3] FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
[safari 7 OS X 10.9 #3] 
[safari 7 OS X 10.9 #3] Failures:
[safari 7 OS X 10.9 #3] 
[safari 7 OS X 10.9 #3]   1) App where angular is loaded more than once should have the interpolated text
[safari 7 OS X 10.9 #3]   at 0.485s [Sat, 05 Mar 2016 22:20:19 GMT]
[safari 7 OS X 10.9 #3]    Message:
[safari 7 OS X 10.9 #3]      NoSuchWindowError: The driver is not focused on a window. You must switch to a window before proceeding. (WARNING: The server did not provide any stacktrace information)
[safari 7 OS X 10.9 #3] Command duration or timeout: 6 milliseconds
[safari 7 OS X 10.9 #3] Build info: version: '2.48.0', revision: 'b7b081a', time: '2015-10-07 15:48:56'
[safari 7 OS X 10.9 #3] System info: host: 'itako55048.prod.miso', ip: '172.20.55.48', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.9.5', java.version: '1.8.0_66'
[safari 7 OS X 10.9 #3] Driver info: org.openqa.selenium.safari.SafariDriver
[safari 7 OS X 10.9 #3] Capabilities [{browserName=safari, takesScreenshot=true, javascriptEnabled=true, version=7.1.7, cssSelectorsEnabled=true, platform=MAC, secureSsl=true}]
[safari 7 OS X 10.9 #3] Session ID: null
[safari 7 OS X 10.9 #3]    Stacktrace:
[safari 7 OS X 10.9 #3]      NoSuchWindowError: The driver is not focused on a window. You must switch to a window before proceeding. (WARNING: The server did not provide any stacktrace information)
[safari 7 OS X 10.9 #3] Command duration or timeout: 6 milliseconds
[safari 7 OS X 10.9 #3] Build info: version: '2.48.0', revision: 'b7b081a', time: '2015-10-07 15:48:56'
[safari 7 OS X 10.9 #3] System info: host: 'itako55048.prod.miso', ip: '172.20.55.48', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.9.5', java.version: '1.8.0_66'
[safari 7 OS X 10.9 #3] Driver info: org.openqa.selenium.safari.SafariDriver
[safari 7 OS X 10.9 #3] Capabilities [{browserName=safari, takesScreenshot=true, javascriptEnabled=true, version=7.1.7, cssSelectorsEnabled=true, platform=MAC, secureSsl=true}]
[safari 7 OS X 10.9 #3] Session ID: null
[safari 7 OS X 10.9 #3]     at Array.forEach (native)
[safari 7 OS X 10.9 #3] From: Task: Asynchronous test function: beforeEach()
[safari 7 OS X 10.9 #3] Error
[safari 7 OS X 10.9 #3]     at [object Object].<anonymous> (/home/travis/build/angular/angular.js/test/e2e/tests/angular-already-loadedSpec.js:2:3)
[safari 7 OS X 10.9 #3]     at Object.<anonymous> (/home/travis/build/angular/angular.js/test/e2e/tests/angular-already-loadedSpec.js:1:63)
[safari 7 OS X 10.9 #3] 
[safari 7 OS X 10.9 #3]   2) App where angular is loaded more than once should have the interpolated text
[safari 7 OS X 10.9 #3]   at 0.485s [Sat, 05 Mar 2016 22:20:19 GMT]
[safari 7 OS X 10.9 #3]    Message:
[safari 7 OS X 10.9 #3]      NoSuchWindowError: The driver is not focused on a window. You must switch to a window before proceeding. (WARNING: The server did not provide any stacktrace information)
[safari 7 OS X 10.9 #3] Command duration or timeout: 5 milliseconds
...
```

I tried to run locally on my computer with Safari 9.0.3 (Mac OS X 10.11.3), I had problems with cookies, https://github.com/karma-runner/karma/issues/878 and missing webdriver plugin, but it passed.

¿How can I request a rebuild of the last commit?

petebacondarwin:
@drpicox you can trigger a rebuild on Travis by simply amending the last commit and force pushing...

```bash
git commit --amend
git push --force-with-lease
```

But I will trigger a rebuild for you now directly.

drpicox:
Thanks!
Good to know!

src/ng/directive/as.js
drpicox: I thought a lot about duplication, and at the end I arrived to many conclusions/reasonings including this one:

- I would have never thought about using name="" to assign a form controller into another controller, it just seems out of scope of name="" (although I have no other mechanism, probably I would have used the `$scope`, in fact I have seen this solution plenty of times)
- If I had as="" to assign controller into the "template" controller, I wouldn't have hesitated about using as=""

So at least from that perspective, _name_ property and _as_ directive seems to have different intentions, at least for the API consumer, and it seems reasonable to have both of them available in forms.

petebacondarwin:
Looks green now

drpicox:
Yes! Thank you.
Anyway, I have seen now the comment of @dcherman that $destroy should follow the same pattern of https://github.com/angular/angular.js/pull/14127#issuecomment-192734803 (`$onDestroy`). I agree, because it would be a surprise for the programmer if not. So... pushing again (sorry).

drpicox:
I have updated the tests to add more testing about transclude: 'element'.

About transclusion of controllers:
- Before calling $transclude the element is represented by a comment,
- For what I have understood in comments inside `compile.js` jQuery has a limitation that cannot attach data to comments,
- It seems that the transcluded element controller is attached ($element.data) to the created element with $transclude, not to the comment

About priority:
- If the priority of asDirective is exactly equal or greater than the one of the component with transclude element, it receives a comment element in the link function, without data of any kind
- If asDirective has lower priority that the one of the component with transclude element, it is called as many times as the $transclude is called and it receives in the link function the created element, with the controller attached

About scope.$on('$destroy') or element.on('$destroy'): both works with the same exact results in the 14 tests that there are of as directive with both versions: jQuery and jqLite.

Concluding:
- I kept `scope.$on('$destroy')` to be consistent with the `$onDestroy` hook
- I set priority -1 to asDirective, so it will work as expected in any case

drpicox:
I have updated this issue and I have (tried to) summarize everything in this page:

- renamed from as="" to ng-as="", so it makes clear that it is an angular directive

- as requested by @gkalpak it **does not** assigns DOM element

- removed the case for ngForm. name="" does the same and more things (thanks for pointing this out)

- it handles properly destruction/remove in all cases (including transclussion), consistent with $onDestroy hook

- @dcherman proposed to do something like: `as="myLocal = $child"` but it does not handles destruction and we have lifecycle hooks , I believe that goes in another direction

- added a test to verify that it works as expected with http templates

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,144 @@
...
+      var setter = $parse(tAttrs.ngAs).assign;
+
+      return function(scope, iElement, iAttrs) {
+        // gets the controller of the current element (see jqLiteController for details)
+        var controller = iElement.data('$' + controllerName + 'Controller');
+        setter(scope, controller);
+
+        // when the element is removed, remove it from the scope assignment (nullify it)
+        scope.$on('$destroy', function() {
+          setter(scope, null);
```
dcherman: You may want to execute the expression first to ensure you should still do this:

`if (expr(scope) === controller) { ... }`

I ended up writing this code in my own lib since I wanted it prior to this PR being accepted and noticed that with something like ui-router where multiple instances of a view may briefly exist during a state transition, you can end up in scenarios where a new/different instance of a controller gets assigned to this expression prior to the `$destroy` hook being run.  As a result, you unexpectedly end up with null/undefined rather than the new controller.

If that's kind of unclear, let me know and I'll try to make a reproducer to show what I mean later.

drpicox: Good point! Thanks!

Enter/leave animations usually duplicate elements during transitions (for example ngView itself). Because you can write `ng-as="$ctrl.myComponent"` you cannot protect yourself with a new scope. 
Leaving element already existed and entering element is created, so order is maintained. It seems that adding the condition is good enough.

jbedard:
Isn't it weird having internal component state/methods on the controller and providing a way (`ng-as`) for others to access that controller? It seems like internal state and exposed API/state should be separate. For example components often have `$onChanges` which even has the single-$ making that method look public, and this PR exposes that object outside the component.

Is that the direction angular2 is going?

dcherman:
@jbedard I think this solves a similar problem to what `ViewChild` does in Angular 2

https://angular.io/docs/ts/latest/api/core/index/ViewChild-var.html

jbedard:
But is there any way to separate the internal state/methods from the object exposed through this though? If this is equivalent to angular 2 then I'm all for it, just curious...

drpicox:
As I have stated before, it is the same that it happening in RiotJS, Agular2, and Polymer. (In fact Polymer this were the whole DOM element, including the API for that element).
This lack of separation of internal state/methods bothered me, but after using Polymer I have realized that it keeps a simplicity that helps. For my experience, new programmers feels more comfortable understanding the whole controller as a single entity that not as 2 or 3 entities sharing state. 
See https://angular.io/docs/ts/latest/cookbook/component-communication.html#!#parent-interacts-with-child-via-local-variable- for a working example in Angular2.

Aaron-Pool:
Really like this, any updates on when it may be merged in?

petebacondarwin:
@jbedard do you want to shepherd this feature into 1.6?

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
...
+      var getter = $parse(tAttrs.ngAs);
+      var setter = getter.assign;
+
+      return function(scope, iElement, iAttrs) {
+        // gets the controller of the current element (see jqLiteController for details)
+        var controller = iElement.data('$' + controllerName + 'Controller');
+        setter(scope, controller);
+
+        // when the element is removed, remove it from the scope assignment (nullify it)
+        scope.$on('$destroy', function() {
```
jbedard: I wonder if this should be `$element.on('$destroy', ...)`? Similar to what the comment implies.

I think when using scope it is possible that the element/component gets removed but the `scope` we are referencing here does not? Maybe it would only happen if manually removed from the DOM?

We seem to be inconsistent in other spots today... `ng-model` uses scope destroy, `formDirective` uses element destroy, `SelectController` uses both. IMO the element one makes more sense, especially when the directive does not have a `scope: true|{}`.

Opinions?

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
...
+    restrict: 'A',
+    compile: function(tElement, tAttrs) {
+      // gets the expected controller name, converts <x-some-thing> into "someThing")
+      var controllerName = tAttrs.$normalize(tElement[0].tagName.toLowerCase());
+
+      // get the setter for the as attribute
+      var getter = $parse(tAttrs.ngAs);
+      var setter = getter.assign;
+
+      return function(scope, iElement, iAttrs) {
```
jbedard: What is the `i` prefix? I think we should try to be consistent with other directives with normally do `scope, element, attrs`. Same with the `compile` args.

src/ng/directive/ngAs.js
dcherman: I think we decided that `$scope` represents the lifecycle of a component/directive, not the element itself.  This is consistent with the decision made for the `$onDestroy` hook as well as the `.component` helper enforcing an isolate scope.  As pointed out earlier, you can't use the element with `transclude: 'element'` directives, though that's not the common use case.

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
...
+ *   </file>
+ * </example>
+ */
+var ngAsDirective = ['$parse', function($parse) {
+  return {
+    priority: -1,
+    restrict: 'A',
+    compile: function(tElement, tAttrs) {
+      // gets the expected controller name, converts <x-some-thing> into "someThing")
+      var controllerName = tAttrs.$normalize(tElement[0].tagName.toLowerCase());
```
jbedard: I'm not really sure why, but everywhere else we use `.nodeName` instead of `.tagName`. Half of those places also use the `nodeName_(el|$el)` helper which I'd vote for here.

You should also use `directiveNormalize` directly instead of off the `tAttrs`.

```javascript
var controllerName = directiveNormalize(nodeName_(tElement));
```

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
...
+ *      });
+ *   </file>
+ * </example>
+ */
+var ngAsDirective = ['$parse', function($parse) {
+  return {
+    priority: -1,
+    restrict: 'A',
+    compile: function(tElement, tAttrs) {
+      // gets the expected controller name, converts <x-some-thing> into "someThing")
```
drpicox: `data-some-thing` `x-some-thing` `some:thing` `some-thing` are normalized to `someThing`

¿Which is the extra closing brace?

src/ng/directive/ngAs.js
jbedard: I think the `$isolateScope` represents the lifecycle of a component, not the parent `$scope` (which I think is being used in this case)?

src/ng/directive/ngAs.js
drpicox: Changed.

src/ng/directive/ngAs.js
jbedard: There's a `)` at the end of the comment...?

src/ng/directive/ngAs.js
jbedard: Someone might remove the element + destroy the isolated scope of the component. I think that would be valid and a proper cleanup, but would not trigger `$destroy` on this parent `$scope`.

So we didn't use `$element.on('$destroy', ..)` due to element transclusion? Would that not work if this directive has lower priority then the transclusion directive?

drpicox: Tested and it works.

```javascript
it('should nullify be compatible with $element transclusion', function() {
      var template = '<my-component ' +
        'ng-as="$ctrl.myComponent" ' +
        'ng-if="!nullified"' +
        '></my-component>';
      $rootScope.$ctrl = {};
      $compile(template)($rootScope);

      $rootScope.$apply('nullified = true');
      expect($rootScope.$ctrl.myComponent).toBeUndefined();
      $rootScope.$apply('nullified = false');
      expect($rootScope.$ctrl.myComponent).toBe(myComponentController);
      $rootScope.$apply('nullified = true');
      expect($rootScope.$ctrl.myComponent).toBe(null);
});
```

src/ng/directive/ngAs.js
drpicox: Ops, thanks!

src/ng/directive/ngAs.js
drpicox: Changed.

src/ng/directive/ngAs.js
drpicox: I have the test:

```javascript
    it('should nullify the variable once the component is destroyed manually', function() {
      var template = '<my-component ng-as="$ctrl.myComponent"></my-component>';
      $rootScope.$ctrl = {};
      $compile(template)($rootScope);
      expect($rootScope.$ctrl.myComponent).toBe(myComponentController);

      myComponentController.destroyManually();
      expect($rootScope.$ctrl.myComponent).toBe(null);
    });
```

In this case `$element.on('$destroy', ...)` works but `scope.$on('$destroy', ...)` fails as expected.

dcherman: @drpicox In the cases I've needed to access the scope for whatever reason, I often cheat and assign it to a property on a controller, then just fetch the controller.  It's gross, but works in both debug and production modes.

@jbedard I forget where this discussion occurred so I can't reference it, but this has come up before and I think the consensus was that the developer is doing it wrong.  In that unit test, it's linking the compiled component against the `$rootScope`, essentially tying that component to the lifecycle of `$rootScope`.  In order to do this correctly, they should have created a new scope using `$rootScope.$new()` and destroyed it at the same time that they wanted to programatically.  This mirrors what `ngIf`, `ngRepeat`, etc. do internally with transclusion and why the use case just works when using those directives.

jbedard: What if I do it "correctly" but with an isolated scope?

That aside, even if it does mean the developer is doing something wrong... is there an issue with using the element event instead of scope event?

jbedard: If `$element` is a comment then `$element.controller()` won't work either, so I don't think scope vs element $destroy matter in that case.

dcherman: That's a good point as of today.  It actually would work with `jqLite`, and an argument could be made that jQuery should change `acceptData` for the non-compat build, but that's all in the future and maybes.

In that case, I don't have a strong argument one way or another at the moment.

jbedard: I think jQuery and jqLite are the same:
https://github.com/angular/angular.js/blob/v1.5.8/src/jqLite.js#L188-L193
https://github.com/jquery/jquery/blob/3.1.1/src/data/var/acceptData.js

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
+'use strict';
+
+/**
+ * @ngdoc directive
+ * @name ngAs
+ * @restrict A
+ *
+ * @description
+ * The `ngAs` attribute tells Angular to assign element component controller
```
jbedard: "**the** element component controller to ~~a~~ **the** given property"?

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
...
+ * to a given property.
+ *
+ * Using this directive you can use the controller of existing components
+ * in your template (children components).
+ *
+ * If the children component is destroyed
```
jbedard: Not really sure what our conventions are here, but this line looks pretty short...?

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
...
+ * a `null` is assigned to the property.
```
jbedard: Probably drop the "a"?

src/ng/directive/ngAs.js
jbedard: "children" can probably be dropped from this sentence

src/ng/directive/ngAs.js
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * @example
+ * ### Use inside the scope
+ * This example shows how the controller of the component toggle
+ * is reused in the template through the scope to use its logic.
+ * <example name="ngAsDirectiveComponentExample" module="ngAsExample">
+ *   <file name="index.html">
+ *     <toggle ng-as="myToggle"></toggle>
+ *     <button ng-click="myToggle.toggle()">Toggle</button>
+ *     <div ng-show="myToggle.isOpen()">You are using a children component to show it.</div>
```
drpicox: From this point of view yes.
Probably I should add a parent component to make it more explicit that it is the parent template who creates the relationships.

src/ng/directive/ngAs.js
drpicox: @dcherman what about this: http://jsfiddle.net/xmw5yney/1/

The problem is not that I have used $rootScope, the point is that components have their own scope, so, it is not a requirement for the user to create a new scope. 
The example that you shows you are removing the scope of the parent "component" not the scope of the children component. So I think that you example is not valid.

drpicox:
I have improved documentation, examples, and tests.

petebacondarwin:
@drpicox we finally got round to discussing this in our team meeting today. It looks like we are in agreement that this is a useful feature that we would like to merge. Thanks for your patience and efforts.

There is one outstanding behaviour that we would like to tighten up.

In Angular the `#` syntax restricts access to the variable to the template on which it is defined. We would like to emulate this by changing the directive as follows:

* restrict the expression that can be given to the `ngAs` directive to a single property name, which means it can only be attached directly to a scope.
* apply this expression to the nearest isolated scope containing the current element. This would effectively mean that the value is put directly onto the scope of the current template.

We would also like to point out in the documentation that the developer should not try to access this value from code, (i.e. controller methods) since this allows too much freedom to cause problems in the app; and also it is not how it works in Angular (which is something we are trying to emulate here).

@jbedard is going to keep track of this and help review and get it merged in. If @drpicox, you don't have time or inclination to complete this, do let him know and he would be happy to finish it up.

dcherman:
@petebacondarwin Sure, but Angular 2 also allows `ViewChild` or `ContentChild` which one could argue is what this feature is equivalent to since `#` is documented as giving access to the element instance versus the controller instance.  If we're considering this feature the equivalent of `ViewChild`, then it makes perfect sense to allow us to store the result on our controller.

drpicox:
It seems sound.
I really appreciate all the efforts that you have done to get a good and
safe definition.

So in short:

[ ] do not parse ngAs but use it as symbol name
[ ] add a check to the symbol name to be a valid symbol, ej: regex + error
message if it is not  (like ngController or ngRepeat)
[ ] do not attach into the current scope but inside the nearest isolated
scope (i.e. scope: {})

The second point has a problem: ngRepeat (scope: true). It could happen
that someone put the ngAs inside it. In this case there are five possible
behaviours: 1) attach one random -nothing ensured-, 2) attach only the
first, 3) attach the last, 4) throw an error. Given the situation with
ngAnimate (two instances of the same element coexists for a short time), 1)
probably is the most likely solution. In this case nothing has to be
changed.


I expect to complete the changes tomorrow.

petebacondarwin:
What happens in Angular if `#` is used inside `ngFor`?

petebacondarwin:
Oh! It seems that hash variables do not pass the `ngFor` https://plnkr.co/edit/nwhzkRWkStcNU1LKXgKq?p=preview
This is probably because the inside of the ngFor element is actually inside a `<template>` so it is not in the same context.

Do we therefore need to stop at the containing scope, rather than the containing isolated scope?

jbedard:
Stopping at the containing scope probably means we have to drop the "single property name" idea... :(

dcherman:
@jbedard Have we addressed the idea that we're talking about two different concepts?  The `#` in Angular2 is related to storing the element instance, this is related to the controller which corresponds to the component instance in Angular 2, so it's more closely related to `ViewChild` than it is `#`, at least as I understand those two concepts.

petebacondarwin:
I don't believe that this is about`ViewChild` etc. From the [anguar.io docs](https://angular.io/docs/ts/latest/guide/template-syntax.html#!#ref-vars):

> A template reference variable is a reference to a DOM element or directive within a template.
> It can be used with native DOM elements but also with Angular components — in fact, it will work with any custom web component.

petebacondarwin:
`ViewChild` is a "live" query of the children found in a component's template. As matching children get added or removed the query is updated (as an observable emission).

drpicox:
I just did a similar plunkr: https://embed.plnkr.co/FVRCLo/

It shows that # has the exact same behaviour than we would expect of ngAs with the current implementation if we limit the name to one symbol. It stops in the first scope, isolated or not.

gkalpak:
> @petebacondarwin: Oh! It seems that hash variables do not pass the `ngFor`
> @jbedard: Stopping at the containing scope probably means we have to drop the "single property name" idea... :(

Another idea for keeping it as consistent as possible with Angular, would be to attach it to the first isolate scope or the first scope of a structural directive (which ever is found first). But this would need additional changes (e.g. to somehow "mark" the scopes of structural directives).

Other than that, if I had to choose between (a) the first isolate scope and (b) the first scope, I would prefer (b) and document it, because it would be closer to Angular in most cases. I mean, most non-isolate scopes nowadays belong to structural directives.

drpicox:
I agree with you @gkalpak. I also would go for (b).

gkalpak:
In the interest of making this as close as possible to Angular, I also suggest the following changes:

1. Rename the directive to `ngRef`. (This will also be similar to the `ngBindon` construct.)
   - **Angular**: `#foo` or `ref-foo`
   - **AngularJS**: `ng-ref="foo"` 

2. If there is no element directive with a controller on the element, then bind the underlying Node instead.

drpicox:
The renaming looks fine.

gkalpak:
Let's wait for @jbedard or others to chime in on my suggestions, before you do unnecessary work :smiley:

petebacondarwin:
I have one more concern about the mapping to Angular `#` functionality, which is that it allows references to be attached to different directives on the element via their `exportAs` property. See https://angular.io/docs/ts/latest/guide/forms.html#!#show-and-hide-validation-error-messages for an example of usage with `ngModel`.

It may be that we don't support this and restrict our version to only accessing the component on an element or the element itself, if no component exists. But we ought to make users aware of the difference; and maybe consider an addition to the syntax such as `ngRef="form:ngForm"` or similar.

jbedard:
Sorry I'm not fully up to speed with all the `@ViewChild`, `@ContentChild` stuff. But those seem to be different then `#` / `ref-*`?

From all the comments above...
1. rename to `ng-ref` (:+1: from me)
2. we need to either:
2.1. support non-simple names
2.2. don't support things like `ng-repeat`
2.3. add more complexity to determine the first parent "structural directive"
3. Angular supports multiple of these on an element

@gkalpak is there any way to determine the first parent "structural directive"? If that was easy, without having to create any new API to declare something "structural", then it would be nice.  But otherwise I'm fine with just support non-simple names and using the scope like this PR currently does, but we should add a note in the docs to discourage use from outside the template.

@petebacondarwin for supporting multiple of these on a single element... How about something like `ng-ref-thing-referenced="myRef"`.  Which would assign`$element.controller("thingReferenced"))` to `myRef`, and because each attribute name will be unique you can have as many as you want on an element. I think it's the opposite of angular2 which has `#myRef="thingReferenced"` but if `myRef` can have dots in it then I think we'd have to (see other discussion). Note that `ng-ref="myRef"` would still use the element name falling back to the DOM itself. We'd probably only do the simple `ng-ref=` for now, but something to think about...

drpicox:
I did not know the functionality of exportAs of Angular2, thanks, it looks very useful.

My original idea of ng-as was to provide a simile with `ng-controller="MyController as myController"`: `<my-component ng-as="myComponent">`.

Form was not compatible because it uses a non-documented directive configuration (name:). But It did not bothered me because current main target were component controllers.

Allowing to specify the controller name, things like this are possible: 

```html
<p my-tooltip="tooltip text" ng-ref="myTooltip as tooltip">some text</p>
<buton ng-click="tooltip.show()">Show</button>
```

but 1) this is the reverse of the notation that Angular2 uses, and 2) a separator should be required to use many of them (cannot use two attributes with the same name in the same entity).

To support multiple references in the same element, may be something like this notation should be good:

```html
<my-component ng-ref="component"></my-component>
  ...and...
<my-component my-tooltip="my tooltip text" my-scroll-area 
  ng-ref="{
    tooltip: 'myTooltip',
    scrollArea: 'myScrollArea',
    component: 'myComponent'
  }"></my-component>
```

_Structural directives_: I think that each directive that defines a new scope, transclusion regions, ... smells to structural directives. Unfortunately AngularJs1 have nothing to make it visible in the template.

_Complex names_: It seems that Angular2 does not support them. Doing tests it allows to use dot, but then you cannot use them (may be is a bug, https://plnkr.co/edit/SJKsK0?p=preview ). 

_ViewChild_, _ContentChild_: there was confusion because you can achieve something similar using p <child-component ng-as="$ctrl.childComponent">`. If we do not allow this kind of expressions, they are not relevant. May be in the future, in addition to "require:", "children:" or similar can be added.

`ng-ref-whatever`: I think that it cannot be solved with a directive and compile must be modified. If this is the case and this is the notation to follow, then, may be we could use directly `#` notation.

petebacondarwin:
I think there are decreasing gains in supporting more and more of the Angular syntax. I think the most impact can come from

* allowing one reference per element
* connecting to the (only) component controller on that element, or falling back to the element itself
* applying the property to the scope that contains the element

Are there use cases that this does not support that are significant?

drpicox:
I have been using my own ngAs for more than 2 years, and I learned that:

  - Attribute directives: I never needed to use it with attribute directives, for example: `<div tooltip="msg">...</div>` can be rewritten into `<tooltip text="msg"><div>...</div></tooltip>`, and this second solution is usually more elegant and maintainable.

  - DOM elements: It has been very handy to reference DOM elements, like `<object>`, to access to some features like play music, but I drop the support because it was considered unsafe: https://github.com/angular/angular.js/pull/14080#issuecomment-186210871 , and https://github.com/angular/angular.js/blob/v1.5.10/src/ng/parse.js#L102 

  - Scope containing the element: It is very handy to have access directly from the component controller, not only from $scope. A great example (but not limited to) are forms: it is very handy do `<form name="$ctrl.form"`, and then `if (this.form.$valid) { ... }` (ex: https://plnkr.co/edit/ozdKq6?p=preview ) Without it, I have been seen coders doing all kind of crazy stuff:  double validate everything, using jQuery to detect ng-invalid classes... 


So for me:

1) ok not reference attribute directive controllers
2) I would find nice to reference DOM elements (and useful)
3) I would like to attach them to the controller

Any of them can be added in the future without breaking changes.
In case of 3) Angular2 has ViewChild that works different, but ngForm[name] is already allowing this, it is handy, and programmers always can inject `$scope` :S

gkalpak:
WRT to my concern about exposing the DOM element in https://github.com/angular/angular.js/pull/14080#issuecomment-186210871, I think it is no longr valid, since we have now removed the sandbox and we are not in the business of preventing access to "dangerous" stuff.

dcherman:
@drpicox Here's one more use case: requiring child controllers for content that might be moved outside of the normal DOM structure in a `postLink` function.

While this may seem simple at first, consider if you're trying to do something like absolutely position a tooltip outside of the boundaries of an element containing a `transform` CSS rule.  Normally you'd use `position: fixed` for that so that the layout root is the body, however `transform` creates a local layout root, so you can't do that.  The only option is to position the element outside of the one with the transform, normally just at the end of the body, however this breaks the ability to find a child controller contained inside the content of that tip (as well as breaking `require` on the child, but that's a different issue than is being addressed here).

It also makes unit testing extremely nice in some cases, though I understand that's not the primary use case here :)

drpicox:
Yes, unit testing is just straightforward.

drpicox:
A small clarification, when I said "Any of them can be added in the future without breaking changes", I also mean that adding them now is like open the box of pandora, go back is a breaking change.

So, if it is unclear how to proceed, I can go for the basic one with illegal variable name checking (if not checked, adding new features would be breaking changes).

But to make things easier:

2) 

```html
<audio ng-ref="$ctrl.audioPlayer">
	<source src="pop.mp3" type="audio/mpeg">
</audio>
```

3) 

```html
<toggle ng-ref="toggle">...</toggle>
<div ng-init="$ctrl.toggle = toggle"></div>
```

```javascript
function MyCompCtrl($scope) {
  this.proceed = function() {
    $scope.toggle.close();
  };
}
```

drpicox:
I have added three different commits, each one for one part of the implementation.
I have also updated the branch to the current master for better control of errors in 1. and 2.

gkalpak:
My thoughts:

1. So, what do you suggest for cases where multiple directives have controllers? 😕 Note that even collecting all available directive controllers is complicated (unless this is somehow tied into the compiler). I think I am more keen on only considering the controller attached to the node-name.

2. But Angular does it :grin: I think the closer we can get to Angular, the better for the users.

3. I agree. I don't like this either. I mean, ideally, I would like it to be assigned on the controller that corresponds to the temlate in which `ng-ref` appears, but given this is currently not possible, doing the simplest thing is the best alternative imo (see also https://github.com/angular/angular.js/pull/14080#issuecomment-276298504).

4. Creating multiple refs of elements of the same kind shouldn't be an issue. You choose the ref/property name, so you can give each a different one. Again, I believe we should stick as close as possible to the Angular behavior.

Narretz:
My thoughts on your thoughts :)

1. For multiple directives, we could use an object where the keys are the directives and the values are the reference expressions. This would need to be decided before we merge it, because it could collide with the ref expression (i.e we need a way to decide if the value of ngRef is the map or the reference).

2. Didn't know Angular publishes the element, so I'm fine with keeping it Angular

3. So binding to any expression?

Narretz:
Close to landing as #16511

PR #14088, opened 2016-02-19
fix(ngRoute): allow `ngView` to be included in an asynchronously loaded template

### LINKED ISSUE #1213: Initial route update doesn't happen if ngView in a template loaded by ngInclude
I can't seem to reproduce this on jsFiddle (mhevery says it's because you need real async with delay to reproduce, but the only way I know to include templates for ngInclude in jsFiddle are inline script tags), but here's how to set it up with two local files:

### index.html
```html
<html>
  <head>
    <script src="http://docs.angularjs.org/angular-1.0.1.min.js"></script>
      angular.module('myApp', [], function($routeProvider) {
        $routeProvider.when('/topic', {
          controller: TopicsCtrl,
          template: 'Controller: {{name}}'
        });
        $routeProvider.when('/topic/:topicId', {
          controller: TopicCtrl,
          template: 'Controller: {{name}}<br />Topic ID: {{params.topicId}}'
        });
        $routeProvider.otherwise({redirectTo: '/topic'});
      });

      function TopicsCtrl($scope) {
        $scope.name = 'TopicsCtrl';
      };

      function TopicCtrl($scope) {
        $scope.name = 'TopicCtrl';
        $scope.params = $routeParams;
      };
    </script>
  </head>
  <body ng-app="myApp">
    <div ng-include src="'app.html'"></div>
  </body>
</html>
```

### app.html
```html
<a href="#/topic">All topics</a>
<a href="#/topic/1">Topic 1</a>
<a href="#/topic/2">Topic 2</a>
<a href="#/topic/3">Topic 3</a>
<hr />
<div ng-view></div>
```

## Expected results:
Page loads, redirects to #/topic, and shows 'Controller: TopicCtrl' in the view.

## Actual results:
Page loads, does not redirect.  If you click any link, the proper view is loaded.  However, if you then reload, the view will not be properly loaded.

If you move the ng-view into `index.html` it works fine.

### LINKED ISSUE #6812: using ng-view inside another directive breaks refresh and back button
I hit an interesting problem yesterday with AngularJS views. They (the views) where working when clicking on a link, but not working when accessed directly, or when the back button was used

After quite a bit of debugging, I was able to track the problem to the fact that if I placed the ng-view directive inside another directive, the refresh and back button would break (although it would work ok for links and direct browser url manipulation).

Here is a blog post I just published with more details/screenshots about this issue: http://blog.diniscruz.com/2014/03/problem-with-angularjs-ng-view-it.html.

For reference here is the C# UnitTest that I wrote to replicate the issue:

        [Test]
        public void Page_Tbot_View_Bug()
        {                             
            var urlTemplate  = "{0}/tbot_v2/{1}";
            var expectedHtml = "This is a test";

            Action<string,bool> runTest = 
                    (pageUrl, hasExpectedHtml) =>
                        {
                            var url         = urlTemplate.format(TargetServer, pageUrl);                            
                            ie.open(url);                                                        
                            "Current Url: {0}".info(ie.url());
                            "Current HTML {0}".info(ie.html()); 
                            Assert.AreEqual(hasExpectedHtml, html().contains(expectedHtml));                            
                        };
            
            
            runTest("default.htm#test" , false);        // default html has the ng-view inside a directive       
            runTest("users.htm#test"   , true);         // default html has the ng-view on this page            
        }        

Note: I looked at the other route and ng-view issues (like https://github.com/angular/angular.js/issues/2100 and https://github.com/angular/angular.js/issues/3825) but they don't seem to cover this scenario

gkalpak:
During it's linking phase, `ngView` relies on the info provided in `$route.current` for instantiating the initial view. `$route.current` is set in the callback of a listener to `$locationChangeSuccess`, which is registered during the instantiation of the `$route` service.

Thus, it is crucial that the `$route` service is instantiated before the initial `$locationChangeSuccess` is fired. Since `ngView` declares `$route` as a dependency, the service is instantiated in time if `ngView` is present during the initial load of the page.

Yet, in cases where `ngView` is included in a template that is loaded asynchronously (e.g. in another directive's template), the directive factory might not be called soon enough for `$route` to be instantiated before the initial `$locationChangeSuccess` event is fired.

This commit fixes it, by always instantiating `$route` up front, during the initialization phase.

Fixes #1213
Fixes #6812

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -17,7 +17,11 @@
  */
  /* global -ngRouteModule */
 var ngRouteModule = angular.module('ngRoute', ['ng']).
-                        provider('$route', $RouteProvider),
+                        provider('$route', $RouteProvider).
+                        // Ensure `$route` will be instantiated in time to capture the initial
+                        // `$locationChangeSuccess` event. This is necessary in case `ngView` is
+                        // included in an asynchronously loaded template.
+                        run(['$route', angular.noop]),
     $routeMinErr = angular.$$minErr('ngRoute');
 
 /**
--- a/test/ngRoute/directive/ngViewSpec.js
+++ b/test/ngRoute/directive/ngViewSpec.js
@@ -1027,3 +1027,34 @@ describe('ngView animations', function() {
     ));
   });
 });
+
+describe('ngView in async template', function() {
+  beforeEach(module('ngRoute'));
+  beforeEach(module(function($compileProvider, $provide, $routeProvider) {
+    $compileProvider.directive('asyncView', function() {
+      return {templateUrl: 'async-view.html'};
+    });
+
+    $provide.decorator('$templateRequest', function($timeout) {
+      return function() {
+        return $timeout(angular.identity, 500, false, '<ng-view></ng-view>');
+      };
+    });
+
+    $routeProvider.when('/', {template: 'Hello, world !'});
+  }));
+
+
+  it('should work correctly upon initial page load',
+    // Injecting `$location` here is necessary, so that it gets instantiated early
+    inject(function($compile, $location, $rootScope, $timeout) {
+      var elem = $compile('<async-view></async-view>')($rootScope);
+      $rootScope.$digest();
+      $timeout.flush(500);
+
+      expect(elem.text()).toBe('Hello, world !');
+
+      dealoc(elem);
+    })
+  );
+});
--- a/test/ngRoute/routeSpec.js
+++ b/test/ngRoute/routeSpec.js
@@ -23,6 +23,21 @@ describe('$route', function() {
     dealoc(element);
   });
 
+  it('should be loaded upon initial load (even if `ngView` is loaded async)', function() {
+    module(function($routeProvider) {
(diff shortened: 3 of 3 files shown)


Narretz:
Would be cool to get this fixed! Does this commit address the problems the first fix for this had? Namely that this caused lots of test failures in a google app because the order of some events changed?

gkalpak:
@Narretz, I'm not aware of the first fix you mention, but I'm quite confident, there won't be any issues with event order. This fix doesn't change the order of any events, it just ensures `$route` is instantiated early enough to catch the "interesting" events.

Narretz:
I found this comment: https://github.com/angular/angular.js/issues/1213#issuecomment-55499226

gkalpak:
Ah, right, I forgot about that. No, this fix just ensures `$route` is instantiated early (which could be done by any other service/controller/directive or `run` block anyway), so it set's up it's listeners for `$location` events in time.

Narretz:
Ok, then it LGTM.

gkalpak:
Backported to `v1.5.x` (8237482) and `v1.4.x` (88322c1).

PR #14096, opened 2016-02-20
docs(guide/Components): Clear the "require" issue

YonatanKra:
It was unclear that the required controllers are available only after the `$onInit` event.


YonatanKra:
I signed it!

gkalpak:
I tweaked it a bit and merged. Thx !
Backported to `v1.5.x` as 67f3e76.

PR #14098, opened 2016-02-20
fix(ngMock): prevent memory leak due to data attached to `$rootElement`

### LINKED ISSUE #14094: Memory leak in tests due to `$injector` being attached to `$rootElement`
#14034 introduced the following memory leak in tests using `angular-mocks`:

The `$injector` is attached to the `$rootElement` ([like this][1]) and is never cleaned up.
Since a new `$rootElement` is created for every test, the leak causes Karma to crash on large test-suites.

[1]: https://github.com/angular/angular.js/commit/88bb5518eb8b860e961e17d71f9ff638227854c8#diff-2a255ed5e9564e25ce6eb711b604f40fR2093

gkalpak:
Starting with 88bb551, `ngMock` will attach the `$injector` to the `$rootElement`, but will never clean it up, resulting in a memory leak. Since a new `$rootElement` is created for every test, this leak causes Karma to crash on large test-suites.
The problem was not detected by our internal tests, because we do our own clean-up in
`testabilityPatch.js`.

This commit prevents the memory leak, by cleaning up all data attached to `$rootElement` after each test.

Fixes #14094

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2602,16 +2602,10 @@ if (window.jasmine || window.mocha) {
     currentSpec = null;
 
     if (injector) {
-      var cleanUpElems = [originalRootElement];
-      var $rootElement = injector.get('$rootElement');
-      if ($rootElement !== originalRootElement) cleanUpElems.push($rootElement);
-      cleanUpElems.forEach(function(elem) {
-        // The `$rootElement` might not have been created or
-        // a mocked `$rootElement` might not have the standard methods
-        if (elem && elem.off) elem.off();
-        if (elem && elem.removeData) elem.removeData();
-      });
-      if (window.jQuery) window.jQuery.cleanData(cleanUpElems);
+      var cleanUpElems = [originalRootElement[0]];
+      var rootNode = injector.get('$rootElement')[0];
+      if (rootNode && (rootNode !== originalRootElement[0])) cleanUpElems.push(rootNode);
+      angular.element.cleanData(cleanUpElems);
 
       injector.get('$rootScope').$destroy();
     }
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -127,12 +127,12 @@ angular.mock.$Browser = function() {
 };
 angular.mock.$Browser.prototype = {
 
-/**
-  * @name $browser#poll
-  *
-  * @description
-  * run all fns in pollFns
-  */
+  /**
+   * @name $browser#poll
+   *
+   * @description
+   * run all fns in pollFns
+   */
   poll: function poll() {
     angular.forEach(this.pollFns, function(pollFn) {
       pollFn();
@@ -2089,10 +2089,12 @@ angular.mock.$RAFDecorator = ['$delegate', function($delegate) {
 /**
  *
  */
+var originalRootElement;
 angular.mock.$RootElementProvider = function() {
-  this.$get = function() {
-    return angular.element('<div ng-app></div>');
-  };
+  this.$get = ['$injector', function($injector) {
+    originalRootElement = angular.element('<div ng-app></div>').data('$injector', $injector);
+    return originalRootElement;
+  }];
 };
 
 /**
@@ -2600,7 +2602,17 @@ if (window.jasmine || window.mocha) {
     currentSpec = null;
 
     if (injector) {
-      injector.get('$rootElement').off();
+      var cleanUpElems = [originalRootElement];
+      var $rootElement = injector.get('$rootElement');
+      if ($rootElement !== originalRootElement) cleanUpElems.push($rootElement);
+      cleanUpElems.forEach(function(elem) {
+        // The `$rootElement` might not have been created or
+        // a mocked `$rootElement` might not have the standard methods
+        if (elem && elem.off) elem.off();
+        if (elem && elem.removeData) elem.removeData();
+      });
+      if (window.jQuery) window.jQuery.cleanData(cleanUpElems);
+
       injector.get('$rootScope').$destroy();
     }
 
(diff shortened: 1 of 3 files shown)


test/ngMock/angular-mocksSpec.js
```diff
@@ -2408,9 +2408,134 @@ describe('ngMockE2E', function() {
...
+
 describe('make sure that we can create an injector outside of tests', function() {
   //since some libraries create custom injectors outside of tests,
   //we want to make sure that this is not breaking the internals of
   //how we manage annotated function cleanup during tests. See #10967
   angular.injector([function($injector) {}]);
 });
+
+
+describe('don\'t leak memory between tests', function() {
```
petebacondarwin: I prefer describe block descriptions to be nouns of things that are being tested...

But I see you are following the pattern above :-(

petebacondarwin:
So the idea of the tests is that you have to run a test beforehand that sets up the spies so that you can check to see that the root element was in fact cleaned up?
I think that the wording of the describe and it blocks could be tweaked so that we remember what they are doing in 12 months time.

test/ngMock/angular-mocksSpec.js
```diff
@@ -2408,9 +2408,134 @@ describe('ngMockE2E', function() {
...
+      expect(prevCleanDataSpy.callCount).toBe(2);
+
+      var cleanUpElems = prevCleanDataSpy.calls[1].args[0];
+      expect(cleanUpElems.length).toBe(1);
+      expect(cleanUpElems[0][0]).toBe(prevRootElement[0]);
+    }
+  });
+});
+
+describe('don\'t leak memory between tests with mocked `$rootScope`', function() {
```
petebacondarwin: should this be "with mocked `$rootElement`"?

test/ngMock/angular-mocksSpec.js
```diff
@@ -2408,9 +2408,134 @@ describe('ngMockE2E', function() {
...
+    if (jQuery) {
+      // One call is made by `testabilityPatch`'s `dealoc()`
+      // We want to verify the subsequent call, made by `angular-mocks`
+      expect(prevCleanDataSpy.callCount).toBe(2);
+
+      var cleanUpElems = prevCleanDataSpy.calls[1].args[0];
+      expect(cleanUpElems.length).toBe(1);
+      expect(cleanUpElems[0][0]).toBe(prevRootElement[0]);
+    }
+  });
```
petebacondarwin: Do we need to do anything to tidy up the spies that we are holding on to?

test/ngMock/angular-mocksSpec.js
gkalpak: I don't like the descriptions either. I'll tweak them.

test/ngMock/angular-mocksSpec.js
gkalpak: Theoretically, no. Jasmine will restore the original objects after each test. We hold on to the spies, but the original object is restored. When the `define` blocks are done, the references to those spies are useless (and should be garbage collected), but don't affect the environment in any way.

test/ngMock/angular-mocksSpec.js
gkalpak: Yup :)

src/ngMock/angular-mocks.js
```diff
@@ -2600,7 +2602,17 @@ if (window.jasmine || window.mocha) {
...
+      var cleanUpElems = [originalRootElement];
+      var $rootElement = injector.get('$rootElement');
+      if ($rootElement !== originalRootElement) cleanUpElems.push($rootElement);
+      cleanUpElems.forEach(function(elem) {
+        // The `$rootElement` might not have been created or
+        // a mocked `$rootElement` might not have the standard methods
+        if (elem && elem.off) elem.off();
+        if (elem && elem.removeData) elem.removeData();
+      });
+      if (window.jQuery) window.jQuery.cleanData(cleanUpElems);
```
jbedard: I think this can be `jqLite.cleanData(cleanUpElements)`. Then you might not need the `off` or `removeData` calls above?

petebacondarwin: I was wondering about that too.

gkalpak: I copied this over from `testabilityPatch.js`, but I think you are right.

From a quick look, I alos think there might be a leak here (and in `testabilityPatch.js`):
Passing an array of jqLite/jQuery elements won't properly clean up. `cleanData` expects an array of "raw" node elements.

I'll try this out.

gkalpak:
I has to rebase this on top of master (because the commit this is supposed to fix was reverted).

I incorporated the original commit (attaching `$injector` to `$rootScope`) in this PR's 1st commit. (Other than that, I haven't done any other changes in the 1st commit, so you don't have to re-review the other parts.)

I added a 2nd commit, that addressed your feedback (except for https://github.com/angular/angular.js/pull/14098#discussion_r53570545 - I'll play with that next).
I tried to make the test descriptions as clear as possible and add comments explaining parts of the test logic that felt unclear, but I'm open to suggestions for better explanations :smiley:

gkalpak:
I added a 3rd commit that uses `jqLite.cleanData(...)`.

src/ngMock/angular-mocks.js
jbedard: The `jqLite` version of `cleanData` was added [fairly recently](https://github.com/angular/angular.js/commit/96288d02d316ff1508b2e0673f34291f9abd1c7f#diff-8e0d9e5c18fe4f1050484080525d4367R192)...

gkalpak:
And a 4th commit (fixup 3) that fixes a couple of corner cases.
<sub>(BTW, the commit messages contain more details about the contents of each fixup commit.)</sub>

petebacondarwin:
LGTM - please squash and merge.

src/ngMock/angular-mocks.js
gkalpak: Yeah, that would explain it :)

gkalpak:
Backported to `v1.5.x` (75373dd) and `v1.4.x` (571e323).

PR #14101, opened 2016-02-21
Filter number:  fix for the output of -Infinity

nonopolarity:


jianlin:
I signed the developer agreement.
> Thanks, your CLA has been submitted.

Narretz:
Thanks - I've landed this commit as part of the other PR. In the future, I recommend that you open different branches for each separate fix. If they are all on master, then newer PRs will include all commits that come before it, which makes it a bit confusing. It's also more difficult for you to sync your fork with the master branch.

PR #14103, opened 2016-02-22
docs(ngMessages): clarify ngMessages docs with clearer example

benelliott:
Recreated PR #14056 due to issues with CLA


Narretz:
Thanks @benelliott, this has landed. I've made a few minor adjustments, I hope that's okay. If you want to work on this further, I think a good addition would be a runnable example in addition to the static code blocks.

benelliott:
Great, thanks! I'll look into that.

PR #14107, opened 2016-02-22
fix(ngMock): don't break if `$rootScope.$destroy()` is not a function

### LINKED ISSUE #14106: `ngMock` breaks during clean-up with mocked `$rootScope`
#13433 introduced `$rootScope` clean-up after each test. The implementation assumes that `$rootScope` will always have a `$destroy()` method, which might not be true for a mocked/decorated `$rootScope`, resulting in an error.

Also discussed in https://github.com/angular/angular.js/issues/13985#issuecomment-186409320.

gkalpak:
Previously, `angular-mocks` was calling `$rootScope.$destroy()` after each test as part of it's cleaning up, assuming that it was always available. This could break if `$rootScope` was mocked and the mocked version didn't provide the `$destroy()` method.
This commit prevents the error by first checking that `$rootScope.$destroy` is present.

Fixes #14106

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2612,8 +2612,10 @@ if (window.jasmine || window.mocha) {
       }
       angular.element.cleanData(cleanUpNodes);
 
+      // Ensure `$destroy()` is available, before calling it
+      // (a mocked `$rootScope` might not implement it (or not even be an object at all))
       var $rootScope = injector.get('$rootScope');
-      if ($rootScope.$destroy) $rootScope.$destroy();
+      if ($rootScope && $rootScope.$destroy) $rootScope.$destroy();
     }
 
     // clean up jquery's fragment cache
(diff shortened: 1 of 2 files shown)


src/ngMock/angular-mocks.js
```diff
@@ -2612,7 +2612,8 @@ if (window.jasmine || window.mocha) {
       }
       angular.element.cleanData(cleanUpNodes);
 
-      injector.get('$rootScope').$destroy();
+      var $rootScope = injector.get('$rootScope');
+      if ($rootScope.$destroy) $rootScope.$destroy();
```
Narretz: can you add a comment about this being possibly undefined due to mocking?

gkalpak: Done

Narretz:
LGTM

gkalpak:
Backported to `v1.5.x` (50ed871) and `v1.4.x` (24a7f28).

Issue #13951, opened 2016-02-05
[TO DISCUSS] Provide a method to get a list of all defined controllers


Vadorequest:
## Problem
There is no way to detect whether a controller has been defined or not without instanciate it, which isn't a good solution. See discussion on SO: https://stackoverflow.com/questions/19734565/how-to-check-if-an-angularjs-controller-has-been-defined

## Use case
Say that we have an angular app which is a generic product which also deals with specific content based on a customer. Say that we have generic controllers that are use by default, except when a specific controller with the customer *suffix* exists and we want to use that controller, but only if it exists. Otherwise we would use the generic controller. (Basically a controller fallback)

In this case we need to check if the specific controller exists and if it doesn't then use the generic one.

### Example
```coffeescript
angular.module('app').config ($stateProvider, ClientServiceProvider) ->
  _ClientService = ClientServiceProvider.$get()

    $stateProvider
    .state 'app',
        controller: _ClientService.resolveClientCtrl 'AppCtrl'
```
The `resolveClientCtrl` method would check whether or not a specific controller exists or use the provided controller. (this is an example of use, for a specific need)

## Proposition 1
Angular already has a list of controllers used internally which aren't shared out of its scope. What we could do is simply add a method that provides the list of defined controller as an *array of strings*.

A solution close to this has been proposed here on SO: http://stackoverflow.com/a/19736852/2391795

## Proposition 2
We could also just add a method which doesn't provide the list of controllers but check whether a controller exists or not.

---

This has been asked two years ago and has more than 5k views in one SO question. I believe this should be in angular "core".

If one of those propositions is accepted (or both) then I'll make a PR.

Narretz:
Thanks for the detailed writeup. What I wonder:
Is there a specific reason why you need different controllers? Couldn't you delegate the task of deciding based on whatever input to a simple controller, which then calls a service based on the input / state?

Adding a `has` method would probably the best way to resolve this issue. But I'm not convinced that there's a pressing need.

Vadorequest:
Well, first I think that if so many people have viewed that question it means that it's not just for my specific use but because there is a greater need. I just gave one example of use.

And yes. We don't want to check inside the same controller whether it is client A, B or C. Because we could have hundreds of them and the code would become unmaintainable. So we need to separate those.

A `has` method would do the trick for my specific need indeed. That would be enough. I'm trying to figure out if there is use case where we would need a complete list of all controllers name, like in an array. That would allow more things actually, that would offer more possibilities.

[...]

Issue #13505, opened 2015-12-11
Upgrade build setup to node 5 + update all npm dependencies (especially karma)


IgorMinar:
also can someone please prepare PRs to make master, 1.4.x, 1.3.x compatible with node 5? do not merge them, just prepare the PRs please. Ideally these three branches should have the same build setup and be always updated in sync so that it’s easy to jump between branches and cherry-pick commits
- [ ] master
- [ ] 1.4.x
- [ ] 1.3.x

IgorMinar:
this is not time-critical but we should do it soonish. on or before Dec 21 is good.

gkalpak:
Node 5 comes with npm 3, which breaks our "purge node_modules in preinstall" workflow. This is tracked in https://github.com/angular/angular.js/issues/13123.

The same issue should also affect ng2 (see https://github.com/angular/angular/issues/5172). @IgorMinar, any ideas on how to tackle it on ng2 ?

IgorMinar:
I'll work on nmp3 upgrade of ng2 in late december.

If this is blocking then let's upgrade to node 4 first now and once ng2 is
npm3 compatible we'll reuse the same approach here.

Narretz:
Why do we need it for 1.3? I'd say more people are using 1.2 than 1.3.

petebacondarwin:
I think it is reasonable to ensure that our main branches are all building on the same environment.
1.3 is a low priority for us but I think that there are still quite a few Google apps using that version. @IgorMinar could probably provide more detail.

petebacondarwin:
Here is a PR for using node 4 on Travis: https://github.com/angular/angular.js/pull/13552

Narretz:
We are on node 6 right now :D

PR #14109, opened 2016-02-22
Add a public "has" method to the Controller class, to check the existence of a given controller

Vadorequest:
I haven't written any test. The `npm install` failed here and I've been unable to test anything. (Seems like it fails on gyp-rebuild with my WIndows 8.1...)

The PR is quite small. if someone with a proper setup could just add some test for it, it would be nice.

https://github.com/angular/angular.js/issues/13951

### DIFF
--- a/src/ng/controller.js
+++ b/src/ng/controller.js
@@ -29,6 +29,15 @@ function $ControllerProvider() {
 
   /**
    * @ngdoc method
+   * @name $controllerProvider#has
+   * @param {string} name Controller name to check.
+   */
+  this.has = function(name) {
+    return controllers.hasOwnProperty(name);
+  };
+
+  /**
+   * @ngdoc method
    * @name $controllerProvider#register
    * @param {string|Object} name Controller name, or an object map of controllers where the keys are
    *    the names and the values are the constructors.
--- a/test/ng/controllerSpec.js
+++ b/test/ng/controllerSpec.js
@@ -57,6 +57,21 @@ describe('$controller', function() {
       expect(ctrl instanceof BarCtrl).toBe(true);
     });
 
+    it('should return true when having an existing controller, should return false otherwise', function() {
+      $controllerProvider.register('FooCtrl', noop);
+      $controllerProvider.register('BarCtrl', ['dep1', 'dep2', noop]);
+      $controllerProvider.register({
+        'BazCtrl': noop,
+        'QuxCtrl': ['dep1', 'dep2', noop]
+      });
+
+      expect($controllerProvider.has('FooCtrl')).toBe(true);
+      expect($controllerProvider.has('BarCtrl')).toBe(true);
+      expect($controllerProvider.has('BazCtrl')).toBe(true);
+      expect($controllerProvider.has('QuxCtrl')).toBe(true);
+
+      expect($controllerProvider.has('UnknownCtrl')).toBe(false);
+    });
 
     it('should allow registration of controllers annotated with arrays', function() {
       var FooCtrl = function($scope) { $scope.foo = 'bar'; },


src/ng/controller.js
```diff
@@ -29,6 +29,15 @@ function $ControllerProvider() {
 
   /**
    * @ngdoc method
+   * @name $controllerProvider#has
+   * @param (string) name Controller name to check.
```
gkalpak: Use `{...}` instead of `(...)`

src/ng/controller.js
```diff
@@ -29,6 +29,15 @@ function $ControllerProvider() {
...
+   */
+  this.has = function(name){
+    return (name && controllers[name])? true : false;
```
gkalpak: Use `controllers.hasOwnProperty(name)` to avoid looking up the prototype chain.

gkalpak:
I left a couple of minor comments. There are also some styling issues that JSCS complains about (you can take a look [here][1]).

Could you also add `Fixes #13951` at the end of your commit message ? This way, the related issue will be automatically closed when this PR is merged.

--
Regarding `npm install`:
Indeed `gyp-rebuild` will throw lots of read in your face, while trying to rebuild some modules on Windows. This doesn't necessarily mean the installation failed (it will just fall back to pre-built artifacts).
Are you sure `npm install` failed ? (Even if it has, the `gyp-rebuild` failures might not be the cause.)

Regarding tests:
Even if you are not able to run the tests locally, you could still add them and push the changes to this PR, so CI can run them for you. It's not necessary though (and it's tooo slow and obviously less than ideal :smiley:).

[1]: https://travis-ci.org/angular/angular.js/jobs/111074010

Vadorequest:
@gkalpak Thanks for feedback. I've pushed changes. I'm installing the environment on an Ubunutu now. I don't know what happened about gyp-rebuild but many node modules weren't properly installed after the red messages, I assumed it didn't go well. :)

(I forgot about `Fixes #13951` but since I'll eventually have to `rebase -i` I'll add it later anyway.

gkalpak:
Great ! The code LGTM. Let us know if you run into any difficulties setting up the testing environment.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties

Vadorequest:
I just pushed some tests, hope they'll pass. Not sure aboute the `it` messages though.
I definitely ran into issues with the setup, I got red messages again after a `npm install`, but all packages seemed to be installed, then the whole `node_modules` folder got deleted. I really don't get it. Seems like an exception has been thrown somewhere and the whole thing got cancelled, I tried twice, same result.

So, CI's got to do the testing part :)

gkalpak:
Are you by any chance using npm >=3.0.0 ? That would explain the `node_modules/` directory getting deleted after modules are installed. It's because the execution order of pre/post-install scripts wrt to other operations has changed in npm 3.
Our setup is currently not compatible with npm 3 (and even if you worked around that, I doubt all our dependencies are). This is tracked in #13505 btw.

gkalpak:
There were some [failures][1] :smiley:
Taking a closer look...

[1]: https://travis-ci.org/angular/angular.js/builds/111180072

test/ng/controllerSpec.js
```diff
@@ -57,6 +57,27 @@ describe('$controller', function() {
...
 
+    it('should return true when having an existing controller', function() {
+      var FooCtrl = function($scope) { $scope.foo = 'foo'; },
+          scope = {},
+          ctrl;
+
+      $controllerProvider.register({FooCtrl: FooCtrl});
+
+      ctrl = $controller('FooCtrl', {$scope: scope});
+      expect(ctrl.has('FooCtrl')).toBe(true);
```
gkalpak: It's `$controllerProvider` that has the `.has()` method. Not specific controller instances.
This test could look like:

```js
$controllerProvider.register('FooCtrl', noop);

expect($controllerProvider.has('FooCtrl')).toBe(true);
expect($controllerProvider.has('BarCtrl')).toBe(false);
```

Or if we wanted to be more thorough, something like:

```js
$controllerProvider.register('FooCtrl', noop);
$controllerProvider.register('BarCtrl', ['dep1', 'dep2', noop]);
$controllerProvider.register({
  'BazCtrl': noop,
  'QuxCtrl: ['dep1', 'dep2', noop]
});

expect($controllerProvider.has('FooCtrl')).toBe(true);
expect($controllerProvider.has('BarCtrl')).toBe(true);
expect($controllerProvider.has('BazCtrl')).toBe(true);
expect($controllerProvider.has('QuxCtrl')).toBe(true);

expect($controllerProvider.has('UnknownCtrl')).toBe(false);
```

And basically, the next test would be redundant.

test/ng/controllerSpec.js
```diff
@@ -57,6 +57,19 @@ describe('$controller', function() {
...
+      $controllerProvider.register({FooCtrl: FooCtrl});
+      expect($controllerProvider.has('FooCtrl')).toBe(true);
+    });
+
+    it('should return false when not having a non existing controller', function() {
+      var FooCtrl = function($scope) { $scope.foo = 'foo'; };
+
+      $controllerProvider.register({FooCtrl: FooCtrl});
+      expect($controllerProvider.has('BarCtrl')).toBe(false);
+    });
```
gkalpak: I would basically merge both tests, into something like:

```js
$controllerProvider.register('FooCtrl', noop);

expect($controllerProvider.has('FooCtrl')).toBe(true);
expect($controllerProvider.has('BarCtrl')).toBe(false);
```

Or if we wanted to be more thorough, something like:

```js
$controllerProvider.register('FooCtrl', noop);
$controllerProvider.register('BarCtrl', ['dep1', 'dep2', noop]);
$controllerProvider.register({
  'BazCtrl': noop,
  'QuxCtrl': ['dep1', 'dep2', noop]
});

expect($controllerProvider.has('FooCtrl')).toBe(true);
expect($controllerProvider.has('BarCtrl')).toBe(true);
expect($controllerProvider.has('BazCtrl')).toBe(true);
expect($controllerProvider.has('QuxCtrl')).toBe(true);

expect($controllerProvider.has('UnknownCtrl')).toBe(false);
```

gkalpak:
@Vadorequest, do you think it would be useful to have the method available on `$controller` as well ?

Vadorequest:
@gkalpak No, I don't think the controller needs to have a `has` method, the provider makes much more sense.

test/ng/controllerSpec.js
Vadorequest: I don't understand what `noop` is here, should I replace it by `{}`?

gkalpak: Sorry, too much internals :grin:
`noop` is basically `angular.noop`, which is basically `function() {}` (a no-op function). You can use `noop` as a shorthand for an empty constructor function. For this functionality, you don't need to pass any arguments to the controller, nor does the controller need to have any logic.

Vadorequest:
Tests passed :) Cool.

And yes, I'm using node 5.5, that's why it fails during the setup.

Vadorequest:
Using node 3.3.1 it worked well, I've been able to run the tests and everything looks good.
The PR has been squashed and is ready to be merged.

test/ng/controllerSpec.js
```diff
@@ -57,6 +57,24 @@ describe('$controller', function() {
       expect(ctrl instanceof BarCtrl).toBe(true);
     });
 
+    it('should return true when having an existing controller, should return false otherwise', function() {
+      var FooCtrl = function($scope) { $scope.foo = 'foo'; },
+          BarCtrl = function($scope) { $scope.bar = 'bar'; };
```
gkalpak: We don't need these two controllers, do we ?

gkalpak:
@Vadorequest, only a minor clean-up comment. LGTM otherwise :+1:

test/ng/controllerSpec.js
Vadorequest: Indeed! I just removed them.

gkalpak:
Th @Vadorequest :+1: 
Backported to `v1.5.x` as bb9575d.

Vadorequest:
Thanks to you @gkalpak :)
Does that mean that it can be used in the latest version righ now?

gkalpak:
It will be available from the upcoming `v1.5.1` release onwards. Right now you can try it out with the [latest snapshot][1], which is basically the code urrently on master.
(NOTE: The snapshot changes with every commit to master and might contain breaking changes. It is **not** recommened for use in production :smiley: - it's just an easy way to try something out before it's released.)

[1]: https://code.angularjs.org/snapshot/

Vadorequest:
Thanks ;)

PR #14114, opened 2016-02-23
feat($compile): add custom annotations to the controller

petebacondarwin:
  feat
  inconvenient
  add custom annotations to the controller of component directives
  Nope
- **Other information**:

This means that we can access these annotations, such as
`$routeConfig` and `$routerCanActivate` without highjacking
the `ng` module.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1106,6 +1106,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     forEach(options, function(val, key) {
       if (key.charAt(0) === '$') {
         factory[key] = val;
+        controller[key] = val;
       }
     });
 
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -10044,6 +10044,22 @@ describe('$compile', function() {
       }));
     });
 
+    it('should add additional annotations to the controller constructor', function() {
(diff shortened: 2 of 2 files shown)


Narretz:
LGTM. What is the router doing with these annotations?

petebacondarwin:
The router needs to extract this metadata. Previously, the annotations were added to the factory function, but this meant that we had to intercept calls to `module.directive`. With this change, we can use the `$injector` to get the controller constructor and then access the annotations that way.

petebacondarwin:
And it allows us to get rid of a load of horrible router code. See https://github.com/petebacondarwin/ng1-component-router-demo/commit/9a3656e6f843dfa17b32e19d78a0ca3c0bb6be08

petebacondarwin:
Landed with an update to the docs for the `component()` helper function.

gseabrook:
I'm interested to know if I am configuring my components incorrectly as this change broke my app. My component config looks like

angular
        .module('menu')
        .component('menu', {
            controller: 'MenuController',
            templateUrl: './components/menu/menu.html',
            $canActivate: function() {
                return true; //real code omitted for brevity
            }
        });

However when "controller[key]" = val runs on line 1109 the controller variable is a string and so it fails

petebacondarwin:
Ahah! Good catch! That is a bug.

PR #14116, opened 2016-02-23
Wait for browser async 1.5

### LINKED ISSUE #789: Protractor does not wait to resolve promises on routes
Hey,

I'm having a few issues getting protractor to wait for my dependencies to load in the route.

The basic code I have is this (which works fine in Angular itself)

```
$routeProvider.when('/login, {
    templateUrl: 'template.html',
    controller: 'LoginCtrl',
    resolve: {
         dependencies: function($q, $rootScope) {
          var def = $q.defer();
          yepnope([
            {
                load: [
                    'dependencies.js',
                ],
                complete: function () {
                    $rootScope.$apply(function () {
                        def.resolve();
                    });
                }
            }
        ]);
        return def.promise;
        }
     }
 });
```

Protractor doesn't wait for the route to resolve and doesn't even display any content from the route. If the resolve dependency is removed the route successfully loads (but the test fails as the dependency isn't loaded).

I've tried it with $timeout testing and that works fine, and as this isn't a $http request, is there something I'm missing, or is this just not supported?

Thanks

### LINKED ISSUE #13782: $http doesn't immediately increment $brower's outstandingRequestCount. Protractor issues result.

heathkit:
Makes whenStable() wait for outstanding browser requests asynchronously. If it waits synchronously, it will miss $http calls made in the current digest cycle.

Fixes #13782

I think this would also fix angular/protractor#789

### DIFF
--- a/test/ng/testabilitySpec.js
+++ b/test/ng/testabilitySpec.js
@@ -189,10 +189,35 @@ describe('$$testability', function() {
 
   describe('waiting for stability', function() {
     it('should process callbacks immediately with no outstanding requests',
-      inject(function($$testability) {
+      inject(function($$testability, $timeout) {
         var callback = jasmine.createSpy('callback');
-        $$testability.whenStable(callback);
-        expect(callback).toHaveBeenCalled();
+        runs(function() {
+          $$testability.whenStable(callback);
+          $timeout.flush();
+        });
+
+        waitsFor(function() {
+          return callback.calls.length > 0;
+        }, "The callback should be called.", 500);
+      }));
+
+    it('should wait for new $http calls asynchronously',
+      inject(function($$testability, $httpBackend, $http, $timeout) {
+        var callback = jasmine.createSpy('callback');
+        runs(function() {
+          $httpBackend.when('GET').respond(200);
+
+          $http.get('');
+          $$testability.whenStable(callback);
+          expect(callback).not.toHaveBeenCalled();
+
+          $httpBackend.flush();
+          $timeout.flush();
+        });
+
+        waitsFor(function() {
+          return callback.calls.length > 0;
+        }, "The callback should be called.", 500);
       }));
   });
 });
--- a/test/ng/testabilitySpec.js
+++ b/test/ng/testabilitySpec.js
@@ -188,7 +188,7 @@ describe('$$testability', function() {
   });
 
   describe('waiting for stability', function() {
-    it('should process callbacks immediately with no outstanding requests',
+    it('should process callbacks without waiting if there are no outstanding requests',
       inject(function($$testability, $timeout) {
         var callback = jasmine.createSpy('callback');
         runs(function() {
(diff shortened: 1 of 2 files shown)


gkalpak:
(Note to self: This is similar to #13862 and #13869.)

test/ng/testabilitySpec.js
```diff
@@ -189,10 +189,35 @@ describe('$$testability', function() {
 
   describe('waiting for stability', function() {
     it('should process callbacks immediately with no outstanding requests',
-      inject(function($$testability) {
+      inject(function($$testability, $timeout) {
```
mgol: The description of this test looks incorrect now (as nothing gets invoked immediately).

mgiambalvo:
I made the test description a little clearer since it's no longer really immediate.

test/ng/testabilitySpec.js
```diff
@@ -188,11 +188,36 @@ describe('$$testability', function() {
...
-    it('should process callbacks immediately with no outstanding requests',
-      inject(function($$testability) {
+    it('should process callbacks without waiting if there are no outstanding requests',
+      inject(function($$testability, $timeout) {
         var callback = jasmine.createSpy('callback');
-        $$testability.whenStable(callback);
-        expect(callback).toHaveBeenCalled();
+        runs(function() {
+          $$testability.whenStable(callback);
+          $timeout.flush();
```
gkalpak: Why is this needed ? Wouldn't `$rootScope.$digest()` suffice ?

test/ng/testabilitySpec.js
```diff
@@ -188,11 +188,36 @@ describe('$$testability', function() {
...
 
   describe('waiting for stability', function() {
-    it('should process callbacks immediately with no outstanding requests',
-      inject(function($$testability) {
+    it('should process callbacks without waiting if there are no outstanding requests',
+      inject(function($$testability, $timeout) {
         var callback = jasmine.createSpy('callback');
-        $$testability.whenStable(callback);
-        expect(callback).toHaveBeenCalled();
+        runs(function() {
```
gkalpak: Since unit tests run synchronously, you shouldn't need async tests (using `runs`/`waitsFor`). Have you tried without ?
(BTW, note that we have updated to Jasmine v2 now, so these tests won't run anymore (since `runs`/`waitsFor` is a Jasmine 1.x thing). But you shouldn't need them anyway (hopefully)...)

gkalpak:
@mgiambalvo, sorry for the late response. Could you address my comments (which apply to both tests btw) and ping me ?

TBH, I don't think this would fix https://github.com/angular/protractor/issues/789. Have you tried it (OOC) ?
If you fancy, feel free to try to create an e2e test as well (take a look at https://github.com/angular/angular.js/pull/13862 to get an idea how to create an e2e test for Angular's testsuite).

gkalpak:
I tried this out and verified that it doesn't indeed fix neither #13782 nor angular/protractor#789.

PR #14119, opened 2016-02-23
docs(guide/Internet Explorer Compatibility): Add ng-attr-type workaround for buttons in IE

### LINKED ISSUE #14117: IE11: button type with interpolated value is ignored

Bug (with workaround).  Though I couldn't find any info on this behavior on the internet and it seems like it'd be an important "gotcha" for a lot of people that should really be documented somewhere.

Using `<button type="{{someInterpolatedThing}}">` doesn't seem to play nicely with IE11 (didn't test past versions).  It sets the type to `submit` regardless of what the interpolated value comes out to.

* **Steps to reproduce and if possible a minimal demo of the problem**

[Here's a pen showing the behavior](http://codepen.io/anon/pen/jWgGPq?editors=1010).  Open in Chrome, you get button type of `button`.  Open in IE11, you get button type of `submit`.

Workaround: Using `ng-attr-type="{{someInterpolatedThing}}"` fixes it.  I guess if this is the only way to workaround it, it might be useful having this as an additional example in the ngAttr docs since the canonical one there right now is `<circle cx>` which makes it seem like it might be primarily for svg.

IE11 interpolates the value and sets the type of the button

Cross-browser consistency (if possible) or at least let people know about it in the docs (if not possible)

* **Which version of Angular, and which browser and OS does this issue affect?** Did this work in previous
versions of Angular? Please also test with the latest stable ánd snapshot versions.

1.5.0 on IE11

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix)

SteveShaffer:
Resolves #14117

### DIFF
--- a/docs/content/guide/ie.ngdoc
+++ b/docs/content/guide/ie.ngdoc
@@ -29,3 +29,6 @@ To ensure your Angular application works on IE please consider:
 
 1. Use `ng-style` tags instead of `style="{{ someCss }}"`. The latter works in Chrome and Firefox
    but does not work in Internet Explorer <= 11 (the most recent version at time of writing).
+2. For the `type` attribute of buttons, use `ng-attr-type` tags instead of 
+   `type="{{ someExpression }}"`.  If using the latter, Internet Explorer overwrites the expression
+   with `type="submit"` before Angular has a chance to interpolate it.


SteveShaffer:
I signed it!

docs/content/guide/ie.ngdoc
```diff
@@ -28,4 +28,7 @@ browsers, but it is up to you to test and decide whether it works for your parti
 To ensure your Angular application works on IE please consider:
 
 1. Use `ng-style` tags instead of `style="{{ someCss }}"`. The latter works in Chrome and Firefox
-   but does not work in Internet Explorer <= 11 (the most recent version at time of writing).
\ No newline at end of file
+   but does not work in Internet Explorer <= 11 (the most recent version at time of writing).
+2. For buttons, use `ng-attr-type` tags instead of `type="{{ someExpression }}"`.
```
Narretz: "Forthe type attribute of buttons, use ..."

SteveShaffer:
Incorporated comment and squashed commits

PR #14122, opened 2016-02-24
docs($http): add a note about modifying data in transformRequest

### LINKED ISSUE #12468: TransformRequest in $http is changing scope
I'm sending scope data trough $http and change that data with TransformRequest before it reaches the API.
When the data is send, an $digest cycle occurs and changes also the scope data.
Note that there is a watcher on scope.value

Here is a basic example of how I change the data to send:

```
var config = {
  method : 'POST',
  url : "APIurl",
  data : 3,
  transformRequest : function(data){

    return data * 2;
  }
};

$http(config)
.then(function result(){
//Nothing is done is here
}, function(error){
//Nothing is done is here
});
```

scope that is send:

```
scope.value = 3
send scope.value and transform it (multiply by 2)
scope.value = 6 //without applying a change on the scope in the result function
//scope.value should stay equal to 3
```

Is this a normal behaviour?
Is there a way to transform the data I send trough http, but keep the scope as it is?

When debugging in chrome, I see the scope is changed when this piece of angularJS code is triggered:

I set /* **Comments** */ to comment code

```
traverseScopesLoop:
do { // "traverse the scopes" loop
if ((watchers = current.$$watchers)) {
  // process our watches
  length = watchers.length;
  while (length--) { /* **I assume it passes trough all watchers** */
    try {
      watch = watchers[length];
      // Most common watches are on primitives, in which case we can short
      // circuit it with === operator, only when === fails do we use .equals
      if (watch) {
        if ((value = watch.get(current)) !== (last = watch.last) &&
            !(watch.eq /* **watch.equal gets triggered. I think it is false** */
                ? equals(value, last)
                : (typeof value === 'number' && typeof last === 'number'
                   && isNaN(value) && isNaN(last)))) {
          dirty = true;
          lastDirtyWatch = watch;
          watch.last = watch.eq ? copy(value, null) : value; /* **watch.eq is false, value is applied**  */
          watch.fn(value, ((last === initWatchVal) ? value : last), current); /* **Goes trough this 2 times, the 2nd time the scope is changed** */
          if (ttl < 5) { /* **It tries this condition and goes back to watch.fn** */
            logIdx = 4 - ttl;
            if (!watchLog[logIdx]) watchLog[logIdx] = [];
            watchLog[logIdx].push({
              msg: isFunction(watch.exp) ? 'fn: ' + (watch.exp.name || watch.exp.toString()) : watch.exp,
              newVal: value,
              oldVal: last
            });
          }
        } else if (watch === lastDirtyWatch) {
          // If the most recently dirty watcher is now clean, short circuit since the remaining watchers
          // have already been tested.
          dirty = false;
          break traverseScopesLoop;
        }
      }
    } catch (e) {
      $exceptionHandler(e);
    }
  }
}   
```

Narretz:
  docs update
  See #12468

Closes #12468

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -550,6 +550,15 @@ function $HttpProvider() {
      * the transformed value (`function(data, headersGetter, status)`) or an array of such transformation functions,
      * which allows you to `push` or `unshift` a new transformation function into the transformation chain.
      *
+     * <div class="alert alert-warning">
+     * **Note:** Angular does not make a copy of the `data` parameter before it is passed into the `transformRequest` pipeline.
+     * That means changes to the properties of `data` are not local to the transform function (since Javascript passes objects by reference).
+     * For example, when calling `$http.get(url, $scope.myObject)`, modifications to the object's properties in a transformRequest
+     * function will be reflected on the scope and in any templates where the object is data-bound.
+     * To prevent his, transform functions should have no side-effects.
+     * If you need to modify properties, it is recommended to make a copy of the data, or create new object to return.
+     * </div>
+     *
      * ### Default Transformations
      *
      * The `$httpProvider` provider and `$http` service expose `defaults.transformRequest` and


src/ng/http.js
```diff
@@ -550,6 +550,15 @@ function $HttpProvider() {
      * the transformed value (`function(data, headersGetter, status)`) or an array of such transformation functions,
      * which allows you to `push` or `unshift` a new transformation function into the transformation chain.
      *
+     * <div class="alert alert-warning">
+     * **Note:** Angular does not make a copy of the `data` parameter before it is passed into the `transformRequest` pipeline.
+     * That means changes to the data are not local to the transform function (since Javascript passes objects by reference).
```
petebacondarwin: changes to properties on the data object are not local

src/ng/http.js
```diff
@@ -550,6 +550,15 @@ function $HttpProvider() {
...
+     * For example, if you are calling `$http.get(url, $scope.myObject)`, and then modify the object's properties in a transformRequest
+     * function, the changes will be reflected on the scope and in any templates where the object is data-bound.
+     * If you are modifying properties, it is therefore recommended to make a copy of the data in the transform function,
+     * or create new object to return.
+     * </div>
+     *
```
petebacondarwin: Is it worth also mentioning that transformRequest and transformResponse functions ought to be "pure", i.e. have no side-effects?

Narretz: Can do, as a general recommendation

src/ng/http.js
Narretz: Ok

PR #14125, opened 2016-02-24
ngOptions: always set selected attribute for selected options

Narretz:
ngOptions wouldn't set the "selected" attribute, if the option was automatically preselected by the browser.


Narretz:
Can you review this? Atm, the test would pass with jqlite without the patch, but it would fail in jquery. I think we can handle the jqlite at a later point, as the issue is quite low impact anyway.

gkalpak:
LGTM

PR #14131, opened 2016-02-25
Component router guide

petebacondarwin:

docs update!
- **Other information**:

So this is still a draft but it could be good enough to merge and iterate on.
The big problem is that, at the moment, I had to include the whole of the router code in the runnable example.

You can see a demo of the guide here: https://ci.angularjs.org/job/angular.js-pete/1073/artifact/build/docs/guide/component-router


gkalpak:
Looks good overall (there are a few typis here and there, but the diff is not displayed to comment :disappointed:)

It might be better to remove the Crisis Center bits from the example, since it's not mentioned in the guide. (Maybe provide the whole app (**with** Crisis Center) at the end as a more complete example.

--
Regarding including the whole router code in the example, it's ugly, but I can live with it as a temporary work-around.
When the issue with `v0.2.x` being marked as pre-release (instead of latest) is resolved, we could use https://npmcdn.org maybe

src/ngComponentRouter/Router.js
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * @ngdoc directive
+ * @name  ngOutlet
+ * @description
+ *
+ */
+
+/**
+ * @name ngLink
```
brandonroberts: The sample code used here for configuring routes and the ngLink is incorrect.

petebacondarwin: Yes - work in progress :-)

btford:
FYI I published `@angular/router` v0.2.0 from master– you should be able to use that now.

petebacondarwin:
Yes!

petebacondarwin:
npmcdn doesn't seem to cope with the scoped package `@angular/router`

gkalpak:
I can see the files just fine (e.g. [angular_1_router.js][1]).
[1]: https://npmcdn.com/@angular/router@0.2.0/angular1/angular_1_router.js

PR #14132, opened 2016-02-25
refactor($compile): Create non-descriptive comments when debugInfoEnabled is false

### LINKED ISSUE #8722: how i can remove the comments created by ng-if and ng-repeat?
Is there a way to prevent Angular from creating "helper" HTML comments? For example,

```<div ng-include="myTemplate"></div>```
Will transform into something like

```
<!-- ngInclude: 'hurr-durr.html' -->
<div ng-include="myTemplate"></div>
```
How do I stop this?

### PR DESCRIPTION
**Pull request by lgalfaso**
  Refactor on generated comments used by `$compile`
  Comments contain the directive name and expression as part of their data.
  When debugInfoEnabled is false, the comments will have empty `data`. When debugInfoEnabled
  is true, the current behavior is kept.
  No
- **Other information**:

When debugInfoEnabled is `false` when comments generated by transclusions, ngIf,
ngRepeat and ngSwitch will not contain any information about the directive nor
the expression associated with it.

Closes: #8722

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1508,6 +1508,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       safeAddClass($element, isolated ? 'ng-isolate-scope' : 'ng-scope');
     } : noop;
 
+    compile.$$createComment = function(directiveName, comment) {
+      var content = '';
+      if (debugInfoEnabled) {
+        content = ' ' + (directiveName || '') + ': ' + (comment || '') + ' ';
+      }
+      return document.createComment(content);
+    };
+
     return compile;
 
     //================================
@@ -2054,8 +2062,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             terminalPriority = directive.priority;
             $template = $compileNode;
             $compileNode = templateAttrs.$$element =
-                jqLite(document.createComment(' ' + directiveName + ': ' +
-                                              templateAttrs[directiveName] + ' '));
+                jqLite(compile.$$createComment(directiveName, templateAttrs[directiveName]));
             compileNode = $compileNode[0];
             replaceWith(jqCollection, sliceArgs($template), compileNode);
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by gkalpak** [comment]
`ngMessagesInclude` creates a similar comment. Maybe we could use `$createComment()` there too.

Other than that it LGTM if you can make JSCS happy :wink:

File: src/ng/compile.js [hunk 1]
```diff
@@ -1508,6 +1508,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       safeAddClass($element, isolated ? 'ng-isolate-scope' : 'ng-scope');
     } : noop;
 
+    compile.$createComment = function(directiveName, comment) {
```
jbedard: Should this be `$$createComment` (double $) like the other methods above?

lgalfaso:
> `ngMessagesInclude` creates a similar comment. Maybe we could use `$createComment()` there too.

thought about it, but that would imply that it would need to check that this function is present... maybe not that bad

src/ng/compile.js
lgalfaso: if it is `$$createComment` then for sure we should not use it in `ngMessagesInclude`

gkalpak:
External modules (such as `ngMessages`) are only guaranted to work if used with the same version of core. There are several things that break (in several modules) if used with the incorrect version. 

So, `angular-messges.js v1.5.1` should only be used with `angular.js v1.5.1`. So, there shouldn't be an issue.

src/ng/compile.js
gkalpak: That's actually a good idea :)

lgalfaso: mmm... ok, updated the PR

lgalfaso:
> External modules (such as ngMessages) are only guaranted to work if used with the same version of core. 

but people are not that careful, and breaking things is always bad

Narretz:
Letting ngMessages use $createComent is a good idea. We've made a similar change in 1.4.9 where some function where moved from ngAnimate to ng, which made both incompatible with other 1.4.x versions. This is expected and we should take advantage of it here.

lgalfaso:
landed with 5e37b2a
backported to 1.5.x with 8237482

PR #14135, opened 2016-02-26
WIP fix(ngRoute): Attach resolved locals before controller instantiation

### LINKED ISSUE #14134: $resolve is not added to the $scope until after the controller is instantiated
**Note: for support questions, please use one of these channels:** https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.

BUG

See my question and the accepted answer at http://stackoverflow.com/questions/35637853/access-angular-route-scope-property-resolve-within-controller-instead-of-view/35647226#35647226

The new $resolve property on the scope added in #134400 works fine when referenced from within the view template. However, the property is not available on `$scope` within a controller unless a `$timeout` is used. Additionally, even using `$timeout` (maybe as a result of using `controllerAs` syntax?) the value can be printed to the console but still does not propagate to the view.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
https://jsfiddle.net/96dwgjea/1/

The `$resolve` property should be put on the scope prior to the controller being instantiated.

It would be even better if `$resolve` was available to inject directly into the controller, so using the `controllerAs` syntax I could just do:

    app.config(function($routeProvider) {
      $routeProvider.when('/test'. {
        templateUrl: 'test.html',
        controller: 'TestController',
        controllerAs: 'vm',
        resolve: {
          'stuff': function() { return 'stuff'; },
          'otherstuff': function() { return 'other stuff'; }
        }
      });
    });

    app.controller(function($resolve) {
      this.stuff = $resolve.stuff;
      this.otherstuff = $resolve.otherstuff;
    });

Instead of injecting the whole `$scope`:

    app.controller(function($scope) {
      this.stuff = $scope.$resolve.stuff;
      this.otherstuff = $scope.$resolve.otherstuff;
    });

Many times I need/want to make mutated copies of the resolved properties for use in the view while still retaining the full data set in the controller. I can easily do this the old way, by injecting each individual resolve property, but I have controllers that have a lot of parameters due to having a bunch of resolve properties (think a view with a form containing a bunch of `<select>`s populated with arrays obtained from factories). It would be much cleaner if all of these properties were injectable in one parameter rather than n parameters.

* **Which version of Angular, and which browser and OS does this issue affect?**
1.5.0 - tested on Chrome 48.0.2564.116 m

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix)
I will submit a PR to fix the issue of needing the `$timeout`. The issue of having `$resolve` as an injectable would obviously need further discussion.

jdforsythe:
  Bug fix. Fixes #14134 
  #14134
  Add the `$resolve` property to the route scope prior to instantiating the controller
  no
- **Other information**:

Resolved locals are attached to route scope as `$resolve` or the
`resolveAs` string prior to instantiation of the routes controller
so the properties are available in the controller as `$scope.$resolve`.

Fixes #14134

### DIFF
--- a/src/ngRoute/directive/ngView.js
+++ b/src/ngRoute/directive/ngView.js
@@ -268,6 +268,8 @@ function ngViewFillContentFactory($compile, $controller, $route) {
 
       var link = $compile($element.contents());
 
+      scope[current.resolveAs || '$resolve'] = locals;
+
       if (current.controller) {
         locals.$scope = scope;
         var controller = $controller(current.controller, locals);
@@ -277,7 +279,6 @@ function ngViewFillContentFactory($compile, $controller, $route) {
         $element.data('$ngControllerController', controller);
         $element.children().data('$ngControllerController', controller);
       }
-      scope[current.resolveAs || '$resolve'] = locals;
 
       link(scope);
     }
--- a/test/ngRoute/directive/ngViewSpec.js
+++ b/test/ngRoute/directive/ngViewSpec.js
@@ -160,6 +160,61 @@ describe('ngView', function() {
     });
   });
 
+  it('should reference resolved locals via $scope.$resolve in controller', function() {
+    var controllerScope,
+        Ctrl = function($scope) {
+          controllerScope = $scope;
+          controllerScope.mutated = $scope.$resolve.name + 'bar';
+        };
+
+    module(function($routeProvider) {
+      $routeProvider.when('/foo', {
+        resolve: {
+          name: function() {
+            return 'foo';
+          }
+        },
+        template: '<div>{{mutated}}</div>',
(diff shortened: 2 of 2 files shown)


jdforsythe:
[WIP] because I was unsure where to document this new behavior. If the team has any suggestions, I'd like to hear them. The current documentation for `$resolve` is in the components section, but that didn't seem to make sense for using the property within any generic route controller. Does it make sense to put it in the di doc?

Narretz:
I assume this is superseded by https://github.com/angular/angular.js/pull/14136

Narretz:
That's actually a standalone issue.

PR #14136, opened 2016-02-26
ngRoute: pass $resolve as local to route controller

jdforsythe:
  feat(ngRoute): add `$resolve` (or `resolveAs` string) as injectable into the controller
  Currently the resolved locals are added as a property on the route scope. #14135 would allow those to be accessed in the controller by injecting the `$scope`. 
  This PR would allow injecting just `$resolve` (or whatever `resolveAs` is set to) in the case that the `controllerAs` syntax is being used and the developer wants access to the resolved locals without injecting the entire `$scope`
  No
- **Other information**:
  Due to the nature of this PR (needing the resolved locals to be available at the time the controller is instantiated), it is dependent on the same change as #14135 and thus was branched from there. It wouldn't make sense to accept this PR and not #14135, but in the case that it's decided this is not wanted, I separated this code.

### DIFF
--- a/src/ngRoute/directive/ngView.js
+++ b/src/ngRoute/directive/ngView.js
@@ -268,6 +268,8 @@ function ngViewFillContentFactory($compile, $controller, $route) {
 
       var link = $compile($element.contents());
 
+      scope[current.resolveAs || '$resolve'] = locals;
+
       if (current.controller) {
         locals.$scope = scope;
         var controller = $controller(current.controller, locals);
@@ -277,7 +279,6 @@ function ngViewFillContentFactory($compile, $controller, $route) {
         $element.data('$ngControllerController', controller);
         $element.children().data('$ngControllerController', controller);
       }
-      scope[current.resolveAs || '$resolve'] = locals;
 
       link(scope);
     }
--- a/src/ngRoute/directive/ngView.js
+++ b/src/ngRoute/directive/ngView.js
@@ -262,15 +262,17 @@ function ngViewFillContentFactory($compile, $controller, $route) {
     priority: -400,
     link: function(scope, $element) {
       var current = $route.current,
-          locals = current.locals;
+          locals = current.locals,
+          resolveAs = current.resolveAs || '$resolve';
 
       $element.html(locals.$template);
 
       var link = $compile($element.contents());
 
-      scope[current.resolveAs || '$resolve'] = locals;
+      scope[resolveAs] = locals;
 
       if (current.controller) {
+        locals[resolveAs] = locals;
         locals.$scope = scope;
         var controller = $controller(current.controller, locals);
         if (current.controllerAs) {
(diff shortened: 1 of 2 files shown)


jdforsythe:
I'd appreciate some guidance from the team on where to put the documentation for this feature. As with #14135 I'm thinking in the di docs, but if there's a better place, let me know.

jdforsythe:
I don't think the Travis failure has anything to do with the PR. It's failing on the Sauce Labs error:
`Response: {"error": "Too many active org tunnels: 30 >= 30"}`.

I closed and reopened the PR so it would try again, and it failed again. Locally all tests are passing.

Narretz:
Yes, the errors are unrelated.
As noted in #14134, we currently don't feel that this change is super important.
Actually, your PR introduces two different changes.
1. assign $resolve to the scope before the controller is instantiated
2. pass $resolve as a local into the controller.

While I can understand (1), I'm stronly against feature (2). You are already passing each resolve property as a local intro the controller, there's no need to additionally pass the $resolve.

jdforsythe:
The point of this would be to change this:

    /* jshint maxparams: 5 */
    app.controller('TestController', ['local_1', 'local_2', 'local_3', 'local_4', 'local_5', function(local_1, local_2, local_3, local_4, local_5) {
      this.local_1 = local_1;
      this.local_2 = local_2;
      this.local_3 = local_3;
      this.local_4 = local_4;
      this.local_5 = local_5;
    }]);

into this:

    app.controller('TestController', ['$resolve', function($resolve) {
      this.local_1 = $resolve.local_1;
      this.local_2 = $resolve.local_2;
      this.local_3 = $resolve.local_3;
      this.local_4 = $resolve.local_4;
      this.local_5 = $resolve.local_5;
    }]);

There are cases where you may have a long list of locals injected (as stated above, a bunch of `<select>`s for a form, for instance, but there are many examples in some of our apps). This not only makes is look much cleaner but stops jshint from screaming. We want to typically enforce a max number of parameters, but there are many cases where we must use the local comment to override that, and any time we add another local we have to change the max params option. I understand there are not many new features getting added into 1.x and certainly you're not accepting a major BC like only using `$resolve` instead of injecting each local individually, but that probably would have been the better option from the beginning. I can't see any time you want a resolved local to not be injected, so this would have simplified things from the beginning by just automatically injecting everything as a property of `$resolve`.

gkalpak:
@jdforsythe, isn't that already possible using `$q.all(...)` ? ([Demo pen][1])

jdforsythe:
@gkalpak Link?

petebacondarwin:
I think that @gkalpak is referring to this?

```
controller: function($route) {
  this.$resolve = $route.current.locals;
}
```

gkalpak:
I added the link above. @petebacondarwin 's approach works just as well, of course :)

PR #14159, opened 2016-03-02
fix(ngRoute): make route.resolve count as a pending request

sjelin:

Make angular count pending [`resolve`](https://github.com/angular/angular.js/blob/master/src/ngRoute/route.js#L96) values as pending request, for the purposes of making `whenStable` more accurate.

Currently, if someone uses promises to asynchronously set a `resolve`, this is not counted as a pending request.  The result is that callbacks to `whenStable` can be invoked before these variables are available.  This was causing protractor users problems (see https://github.com/angular/protractor/issues/789#issuecomment-190983200 for details)

Waiting for these variables now counts as a pending request.

I guess you could imagine a breaking change where someone wanted a `whenStable` callback to be invoked before the variables resolved.

Not sure tests are necessary for such a small change.  And since this brings behavior in line with user expectations not sure about the docs either.
- **Other information**:

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -7,6 +7,7 @@
 var isArray;
 var isObject;
 var isDefined;
+var noop;
 
 /**
  * @ngdoc module
@@ -54,6 +55,7 @@ function $RouteProvider() {
   isArray = angular.isArray;
   isObject = angular.isObject;
   isDefined = angular.isDefined;
+  noop = angular.noop;
 
   function inherit(parent, extra) {
     return angular.extend(Object.create(parent), extra);
@@ -350,7 +352,8 @@ function $RouteProvider() {
                '$injector',
                '$templateRequest',
                '$sce',
-      function($rootScope, $location, $routeParams, $q, $injector, $templateRequest, $sce) {
+               '$browser',
+      function($rootScope, $location, $routeParams, $q, $injector, $templateRequest, $sce, $browser) {
 
     /**
      * @ngdoc service
@@ -680,6 +683,8 @@ function $RouteProvider() {
 
         var nextRoutePromise = $q.resolve(nextRoute);
 
+        $browser.$$incOutstandingRequestCount();
+
         nextRoutePromise.
           then(getRedirectionData).
           then(handlePossibleRedirection).
@@ -700,6 +705,13 @@ function $RouteProvider() {
             if (nextRoute === $route.current) {
               $rootScope.$broadcast('$routeChangeError', nextRoute, lastRoute, error);
             }
+          }).finally(function() {
+            // Because `commitRoute()` is called from a `$rootScope.$evalAsync` block (see
+            // `$locationWatch`), this `$$completeOutstandingRequest()` call will not cause
+            // `outstandingRequestCount` to hit zero.  This is important in case we are redirecting
+            // to a new route which also requires some asynchronous work.
+
+            $browser.$$completeOutstandingRequest(noop);
           });
       }
     }
--- a/test/e2e/fixtures/ng-route-promise/index.html
+++ b/test/e2e/fixtures/ng-route-promise/index.html
@@ -1,0 +1,9 @@
+<html ng-app="lettersApp">
+  <body>
+    <div ng-view></div>
+
+    <script src="angular.js"></script>
+    <script src="angular-route.js"></script>
+    <script src="script.js"></script>
+  </body>
+</html>
--- a/test/e2e/fixtures/ng-route-promise/script.js
+++ b/test/e2e/fixtures/ng-route-promise/script.js
@@ -1,0 +1,43 @@
+'use strict';
+
+angular.
+  module('lettersApp', ['ngRoute']).
+  config(function($routeProvider) {
+    $routeProvider.
+      otherwise(resolveRedirectTo('/foo1')).
+      when('/foo1', resolveRedirectTo('/bar1')).
+      when('/bar1', resolveRedirectTo('/baz1')).
+      when('/baz1', resolveRedirectTo('/qux1')).
+      when('/qux1', {
+        template: '<ul><li ng-repeat="letter in $resolve.letters">{{ letter }}</li></ul>',
+        resolve: resolveLetters()
+      }).
+      when('/foo2', resolveRedirectTo('/bar2')).
+      when('/bar2', resolveRedirectTo('/baz2')).
+      when('/baz2', resolveRedirectTo('/qux2')).
+      when('/qux2', {
+        template: '{{ $resolve.letters.length }}',
+        resolve: resolveLetters()
+      });
+
+    // Helpers
+    function resolveLetters() {
+      return {
+        letters: function($q) {
+          return $q(function(resolve) {
+            window.setTimeout(resolve, 1000, ['a', 'b', 'c', 'd', 'e']);
+          });
+        }
+      };
+    }
+
+    function resolveRedirectTo(path) {
+      return {
+        resolveRedirectTo: function($q) {
+          return $q(function(resolve) {
+            window.setTimeout(resolve, 250, path);
+          });
+        }
+      };
+    }
+  });
--- a/test/e2e/tests/ng-route-promise.spec.js
+++ b/test/e2e/tests/ng-route-promise.spec.js
@@ -1,0 +1,33 @@
+'use strict';
+
+describe('ngRoute promises', function() {
+  beforeEach(function() {
+    loadFixture('ng-route-promise');
+  });
+
+  it('should wait for route promises', function() {
+    expect(element.all(by.tagName('li')).count()).toBe(5);
+  });
+
+  it('should time out if the promise takes long enough', function() {
+    // Don't try this at home kids, I'm a protractor dev
+    browser.manage().timeouts().setScriptTimeout(1500);
+    browser.waitForAngular().then(function() {
+      fail('waitForAngular() should have timed out, but didn\'t');
+    }, function(error) {
+      expect(error.message).toContain('Timed out waiting for asynchronous Angular tasks to finish');
+    });
+  });
+
+  it('should wait for route promises when navigating to another route', function() {
+    browser.setLocation('/foo2');
+    expect(element(by.tagName('body')).getText()).toBe('5');
+  });
+
+  afterEach(function(done) {
+    // Restore old timeout limit
+    browser.getProcessedConfig().then(function(config) {
+      return browser.manage().timeouts().setScriptTimeout(config.allScriptsTimeout);
+    }).then(done);
+  });
+});
--- a/test/ngRoute/routeSpec.js
+++ b/test/ngRoute/routeSpec.js
@@ -2082,4 +2082,188 @@ describe('$route', function() {
       expect(function() { $route.updateParams(); }).toThrowMinErr('ngRoute', 'norout');
     }));
   });
+
+  describe('testability', function() {
+    it('should wait for $resolve promises before calling callbacks', function() {
+      var deferred;
+
+      module(function($provide, $routeProvider) {
+        $routeProvider.when('/path', {
+          template: '',
+          resolve: {
+            a: function($q) {
+              deferred = $q.defer();
+              return deferred.promise;
+            }
+          }
+        });
+      });
+
+      inject(function($location, $route, $rootScope, $httpBackend, $$testability) {
+        $location.path('/path');
+        $rootScope.$digest();
+
+        var callback = jasmine.createSpy('callback');
+        $$testability.whenStable(callback);
+        expect(callback).not.toHaveBeenCalled();
+
+        deferred.resolve();
+        $rootScope.$digest();
+        expect(callback).toHaveBeenCalled();
+      });
+    });
+
+    it('should call callback after $resolve promises are rejected', function() {
+      var deferred;
+
+      module(function($provide, $routeProvider) {
+        $routeProvider.when('/path', {
+          template: '',
+          resolve: {
+            a: function($q) {
+              deferred = $q.defer();
+              return deferred.promise;
+            }
+          }
+        });
+      });
+
+      inject(function($location, $route, $rootScope, $httpBackend, $$testability) {
+        $location.path('/path');
+        $rootScope.$digest();
+
+        var callback = jasmine.createSpy('callback');
+        $$testability.whenStable(callback);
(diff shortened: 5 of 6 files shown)


gkalpak:
(Note to self: Related to #13782.)

gkalpak:
@sjelin, this seems reasonable. It needs rebasing and definitely some tests (ideally both unit and e2e).
(I wouldn't consider it a breaking change, but this is debatable :smile:)

Would you be interested in moving this forward?

sjelin:
Hi, sorry, yeah - I'll take a look.

sjelin:
Update: I'm back on this.  Just rebased, I'll add some tests after lunch

sjelin:
@gkalpak I finally added tests!  Sorry for not prioritizing this

src/ngRoute/route.js
```diff
@@ -786,7 +788,10 @@ function $RouteProvider() {
         if (angular.isDefined(template)) {
           locals['$template'] = template;
         }
-        return $q.all(locals);
+        return $q.all(locals).then(function(locals) {
+          $browser.$$completeOutstandingRequest(function() {});
+          return locals;
+        });
```
gkalpak: This should be `$.all(locals).finally(function() { $browser.$$completeOutstandingRequest(function() {}); });`.
Because now you will never reach 0 outstanding requests once a resolve fails in the app.

src/ngRoute/route.js
```diff
@@ -786,7 +788,10 @@ function $RouteProvider() {
         if (angular.isDefined(template)) {
           locals['$template'] = template;
         }
-        return $q.all(locals);
+        return $q.all(locals).then(function(locals) {
+          $browser.$$completeOutstandingRequest(function() {});
```
gkalpak: I would use `angular.noop` here. Actually, I would initialize `noop = angular.noop` (e.g. the same as `isArray`) above.

test/ng/testabilitySpec.js
```diff
@@ -194,5 +194,33 @@ describe('$$testability', function() {
         $$testability.whenStable(callback);
         expect(callback).toHaveBeenCalled();
       }));
+
+    it('should wait for route promises before calling callbacks', function() {
```
gkalpak: This test does not belong here. It should be moved to `routeSpec.js`.

gkalpak: You should also add a test for where a `resolve` fails/throws.

gkalpak (changes requested):
I left some comments. We also miss an e2e test.
<sub>(@sjelin, feel free to ping me if you need any help.)</sub>

sjelin:
@gkalpak all comments addressed

gkalpak (changes requested):
While taking another look, I realized that `ngRoute` has one more async feature (that might not be there when this PR was first created) which is worth taking into account: `resolveRedirectTo`

Basically, it is possible to decide about a redirection asynchronously (possibly after some delay). It sounds reasonable to take that delay into account too.

Thus, I suggest expanding the outstanding request increment/decrement "scope" to [this block][1], instead of just around `resolveLocals`. Maybe something like:

```js
function decrementRequestCount() {
  $browser.$$completeOutstandingRequest(noop);
}

...

$browser.$$incOutstandingRequestCount();

nextRoutePromise.
  then(getRedirectionData).
  then(handlePossibleRedirection).
  then(function(keepProcessingRoute) {
    return keepProcessingRoute && nextRoutePromise.
      then(resolveLocals).
      then(function(locals) {
        // after route change
        if (nextRoute === $route.current) {
          if (nextRoute) {
            nextRoute.locals = locals;
            angular.copy(nextRoute.params, $routeParams);
          }

          $rootScope.$broadcast('$routeChangeSuccess', nextRoute, lastRoute);
        }
      });
  }).
  catch(function(error) {
    if (nextRoute === $route.current) {
      $rootScope.$broadcast('$routeChangeError', nextRoute, lastRoute, error);
    }
  }).
  finally(decrementRequestCount);
```

WDYT, @sjelin?

[1]: https://github.com/sjelin/angular.js/blob/3fa796fc566658e2e71c71ab3c9d13a12268de76/src/ngRoute/route.js#L686-L706

gkalpak:
(Also, the Travis failures seem legit.)

src/ngRoute/route.js
```diff
@@ -786,7 +790,9 @@ function $RouteProvider() {
         if (angular.isDefined(template)) {
           locals['$template'] = template;
         }
-        return $q.all(locals);
+        return $q.all(locals).finally(function(locals) {
```
sjelin: woops 😅

test/e2e/fixtures/ng-route-promise/index.html
```diff
@@ -0,0 +1,12 @@
+<html ng-app="lettersApp">
+  <head>
+    <meta charset="utf-8">
+    <title>Angular.js Example</title>
+    <script src="angular.js"></script>
+    <script src="angular-route.js"></script>
+    <script src="script.js"></script>
+  </head>
```
sjelin: Just habit - I'll remove it

test/e2e/fixtures/ng-route-promise/script.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+var lettersApp = angular.module('lettersApp', ['ngRoute']);
+
+lettersApp.config(function($routeProvider) {
+  $routeProvider.
+    when('/', {
+      template: '<ul><li ng-repeat="letter in letters">{{letter}}</li><ul>',
+      controller: 'LettersCtrl',
```
sjelin: Will do!  An yeah, not having arrow functions is the worse 😢

sjelin (commented):
I like the idea of expending the scope of this change to include `resolveRedirectTo`, but I'm not sure how to handle the following race condition:


1. A route uses `resolveRedirectTo`, causing `$$incOutstandingRequestCount` to be called
2. Protractor queries for stability, getting enqueued by `whenStable`
3. `resolveRedirectTo` resolves to a new route, thereby calling `$$completeOutstandingRequest`, and thereby calling  the `whenStable` callbacks.  Protractor will get the signal in a few milliseconds, since it takes time for things to go through the selenium server 
4. The new route uses `resolve`
5. Protractor, thinking the page is stable, executes some webdriver commands commands even though `resolve` has not yet resolved
6. `resolve` resolves, but too late

gkalpak:
> I like the idea of expending the scope of this change to include resolveRedirectTo, but I'm not sure how to handle the following race condition: ...

This doesn't seem to be an issue, because of the relative timing of when things happen. The redirection is correctly taken into account (see the test in #15587).

I have incorporated the proposed changes into #15587 (along with a fix in the second test). Feel free cherry-pick the changes into this PR.

sjelin:
Oh, because javascript is synchronous huh?  I get it now

test/ngRoute/routeSpec.js
```diff
@@ -2082,4 +2082,58 @@ describe('$route', function() {
...
+        $$testability.whenStable(callback);
+        expect(callback).not.toHaveBeenCalled();
+
+        deferred.resolve();
+        $rootScope.$digest();
+        expect(callback).toHaveBeenCalled();
+      });
+    });
+
+    it('should call callback after route promises are rejected', function() {
```
sjelin: @gkalpak I wrote this unit test for the race condition I was worried about, and it does appear to be a problem

test/e2e/tests/ng-route-promise.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    loadFixture('ng-route-promise');
+  });
+
+  it('should wait for promises in resolve blocks', function() {
+    expect(element.all(by.tagName('li')).count()).toBe(5);
+  });
+
+  it('should time out if the promise takes long enough', function() {
+    // Don't try this at home kids, I'm a protractor dev
+    browser.manage().timeouts().setScriptTimeout(1100);
```
sjelin: @gkalpak oddly, this e2e test works fine event though the unit test fails.  I have a theory, which is that a *second* `waitForAngular()` is being introduced somewhere unexpected.  But I'm not sure where.

test/ngRoute/routeSpec.js
```diff
@@ -2082,4 +2082,144 @@ describe('$route', function() {
...
+        $$testability.whenStable(callback);
+        expect(callback).not.toHaveBeenCalled();
+
+        deferred.reject();
+        $rootScope.$digest();
+        expect(callback).toHaveBeenCalled();
+      });
+    });
+
+    it('should wait for all route promises before calling callbacks', function() {
```
sjelin: @gkalpak  This is the unit test which I wrote to test for the race condition.  This unit tests fails, but this doesn't appear to be a problem with real browsers.  I did essentially the same thing in a real browser, and it correctly called the callback only after both promises were resolved.  I'll look into the error in the unit testing setup after lunch

test/e2e/tests/ng-route-promise.spec.js
gkalpak: Sounds reasonable. Doesn't every `Element[Array]Finder` (or whatever they are called) - i.e. `element(...)/element.all(...)` implicitly `waitForAngular()` (I was under that impression)? That would explain it.

test/ngRoute/routeSpec.js
```diff
@@ -2082,4 +2082,152 @@ describe('$route', function() {
...
+
+      inject(function($browser, $location, $route, $rootScope, $httpBackend, $$testability) {
+        $location.path('/foo');
+        $rootScope.$digest();
+
+        var callback = jasmine.createSpy('callback');
+        $$testability.whenStable(callback);
+        expect(callback).not.toHaveBeenCalled();
+
+        // ngRoute code is wrapped in a $browser.defer call (via $rootScope.evalAsync), which in
```
sjelin: @gkalpak this comment is the result of my investigation

test/e2e/tests/ng-route-promise.spec.js
```diff
@@ -0,0 +1,31 @@
...
+
+  it('should time out if the promise takes long enough', function() {
+    // Don't try this at home kids, I'm a protractor dev
+    browser.manage().timeouts().setScriptTimeout(1500);
+    browser.waitForAngular().then(function() {
+      fail('waitForAngular() should have timed out, but didn\'t');
+    }, function(error) {
+      var errorRe = new RegExp(
+          'Timed out waiting for asynchronous Angular tasks to finish after ' +
+          '(?:1\\.5 seconds|15\\d\\d ?ms)\\.', '');
```
gkalpak: I know you took this from my PR, but this RegExp is testing more than is necessary (and might give false negatives after a future update). I think it suffixes to have:

```js
expect(error.message).toContain('Timed out waiting for asynchronous Angular tasks to finish');
```

test/e2e/tests/ng-route-promise.spec.js
```diff
@@ -0,0 +1,31 @@
...
+      expect(error.message).toMatch(errorRe);
+    });
+  });
+
+  afterEach(function(done) {
+    // Restore old timeout limit
+    browser.getProcessedConfig().then(function(config) {
+      return browser.manage().timeouts().setScriptTimeout(config.allScriptsTimeout);
+    }).then(function() { done(); });
```
gkalpak: Couldn't this just be `.then(done)`?

test/ngRoute/routeSpec.js
gkalpak: Where does the `$rootScope.$evalAsync()` come from?

gkalpak: Answering my own question: It comes from `$location` (see [here][1]).
So, you were spot on 👍 (which is a good thing, because it confirms that the feature works out-of-the-box outside of unit tests).

[1]: https://github.com/angular/angular.js/blob/752f411c8b3495620226e45815807ff798496801/src/ng/location.js#L991

gkalpak:
So, this generally LGTM (now that we have confirmed that the failures where due to unit-test mocks).

Ideally, I would apply some more changes (mainly to tests): 927f355b
<sub>(I don't feel too strongly about it though - feel free to ignore.)</sub>

test/e2e/tests/ng-route-promise.spec.js
sjelin: Probably.  In some jasmine setups, passing parameters to the `done` callback indicates a failed test.  This isn't an issue anymore, but I guess it's just a habit of mine now.

test/ngRoute/routeSpec.js
sjelin: I'll mention `$location` too :smile:

sjelin:
@gkalpak I adopted your changes but then made some extremely minor ones of my own: https://github.com/angular/angular.js/pull/14159/commits/3b106ae97ee7dea991eab44b4a5d5033aa0d9b75

1. I wrapped the callback in `$$completeOutstandingRequest` in a `try` block, which is how it's done in production
2. I moved some comments around.  I wanted some of the explanation to be in `src/ngRoute/route.js` so that no one looked at that code and wondered "Why is this never causing `outstandingRequestCount` to hit zero?"
3. In production, I routed all modifications to `outstandingRequestCount` through `$$incOutstandingRequestCount` or `$$completeOutstandingRequest`.  This doesn't fix anything but it just seemed like a good refactor

src/ng/browser.js
```diff
@@ -321,7 +321,7 @@ function Browser(window, document, $log, $sniffer) {
    */
   self.defer = function(fn, delay) {
     var timeoutId;
-    outstandingRequestCount++;
+    self.$$incOutstandingRequestCount();
```
gkalpak: I don't think this buys us much and is inconsistent with the rest of the file (which either accesses/modifies `outstandRequestCount` directly or through (the private) `completeOutstandingRequest()` function).

gkalpak (approved):

gkalpak:
LGTM (Nit: I don't particularly like the 3rd change :grin:)

<sub>@googlebot, I confirm that I am okay with my commits being contributed to this fine project :smiley:</sub>

src/ng/browser.js
sjelin: I disagree but I'll revert this change anyway :stuck_out_tongue:

sjelin:
@gkalpak alright, should be good to go!  I don't have permission to actually do the merge though

gkalpak:
Awesome! I can take it from here. Thx for bearing with me to the end :smiley:

PR #14166, opened 2016-03-03
perf(ngAnimate): cache repeated calls to addClass/removeClass

### LINKED ISSUE #14165: ngAnimate doesn't cache failed animations to avoid addClass/removeClass
**Note: for support questions, please use one of these channels:** https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.
- **Do you want to request a _feature_ or report a _bug_?**
- **What is the current behavior?**
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
  https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:yBpEi4).
- **What is the expected behavior?**
- **What is the motivation / use case for changing the behavior?**
- **Which version of Angular, and which browser and OS does this issue affect?** Did this work in previous
  versions of Angular? Please also test with the latest stable and snapshot versions.
- **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix)

matsko:
Applies to 1.4.x and 1.5.x and master

Closes #14165
Closes #14166

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -303,33 +303,6 @@ function getCssTransitionDurationStyle(duration, applyOnlyDuration) {
   return [style, value];
 }
 
-function createLocalCacheLookup() {
-  var cache = Object.create(null);
-  return {
-    flush: function() {
-      cache = Object.create(null);
-    },
-
-    count: function(key) {
-      var entry = cache[key];
-      return entry ? entry.total : 0;
-    },
-
-    get: function(key) {
-      var entry = cache[key];
-      return entry && entry.value;
-    },
-
-    put: function(key, value) {
-      if (!cache[key]) {
-        cache[key] = { total: 1, value: value };
-      } else {
-        cache[key].total++;
-      }
-    }
-  };
-}
-
 // we do not reassign an already present style value since
 // if we detect the style property value again we may be
 // detecting styles that were added via the `from` styles.
@@ -348,26 +321,15 @@ function registerRestorableStyles(backup, node, properties) {
 }
 
 var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
-  var gcsLookup = createLocalCacheLookup();
-  var gcsStaggerLookup = createLocalCacheLookup();
-
-  this.$get = ['$window', '$$jqLite', '$$AnimateRunner', '$timeout',
+  this.$get = ['$window', '$$jqLite', '$$AnimateRunner', '$timeout', '$$animateCache',
                '$$forceReflow', '$sniffer', '$$rAFScheduler', '$$animateQueue',
-       function($window,   $$jqLite,   $$AnimateRunner,   $timeout,
+       function($window,   $$jqLite,   $$AnimateRunner,   $timeout,   $$animateCache,
                 $$forceReflow,   $sniffer,   $$rAFScheduler, $$animateQueue) {
 
     var applyAnimationClasses = applyAnimationClassesFactory($$jqLite);
 
-    var parentCounter = 0;
-    function gcsHashFn(node, extraClasses) {
-      var KEY = "$$ngAnimateParentKey";
-      var parentNode = node.parentNode;
-      var parentID = parentNode[KEY] || (parentNode[KEY] = ++parentCounter);
-      return parentID + '-' + node.getAttribute('class') + '-' + extraClasses;
-    }
-
-    function computeCachedCssStyles(node, className, cacheKey, properties) {
-      var timings = gcsLookup.get(cacheKey);
+    function computeCachedCssStyles(node, className, cacheKey, allowInvalid, properties) {
+      var timings = $$animateCache.get(cacheKey);
 
       if (!timings) {
         timings = computeCssStyles($window, node, properties);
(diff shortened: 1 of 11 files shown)


matsko:
Don't merge this yet. There's a cache issue.

matsko:
OK now it's fixed.

Narretz:
Wow that's a ton of changes. I'll try to review today. However could you update the commit message with an explanation of the issue and the fix?

test/ngAnimate/animationSpec.js
```diff
@@ -977,11 +1010,12 @@ describe('$$animation', function() {
...
       inject(function($$animation, $rootScope, $animate) {
         element.addClass('four');
+        parent.append(element);
 
         var completed = false;
         $$animation(element, 'event', {
-          from: { background: 'red' },
-          to: { background: 'blue', 'font-size': '50px' }
+          from: { height: '100px' },
+          to: { height: '200px', 'font-size': '50px' }
```
petebacondarwin: @matsko why change these styles? It seems arbitrary.

test/ngAnimate/integrationSpec.js
```diff
@@ -499,6 +571,45 @@ describe('ngAnimate integration tests', function() {
...
+        expect(classRemoveSpy.callCount).toBe(2);
+
+        expect(classAddSpy.calls[0].args[1]).toBe('ng-animate');
+        expect(classAddSpy.calls[1].args[1]).toBe('ng-enter');
+        expect(classRemoveSpy.calls[0].args[1]).toBe('ng-enter');
+        expect(classRemoveSpy.calls[1].args[1]).toBe('ng-animate');
+
+        expect(element.children().length).toBe(100);
+      });
+    });
```
petebacondarwin: Can you explain exactly what you are testing here?
In what way does this test invalid animations. Actually, what do you mean by invalid animations?

My meagre understanding is that you are adding 100 items to an array. The `.animate-me` class is not applied to any of the elements in the DOM so there should be no animation, right?

src/ngAnimate/animateCss.js
```diff
@@ -376,20 +338,26 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
         }
       }
 
+      // if an GCS call doesn't return anything valid for the animation then we
+      // should mark that so that repeated classAdd/removeRemove calls are skipped
```
Narretz: Did you mean `addClass/removeClass`?

src/ngAnimate/animateCss.js
```diff
@@ -520,7 +487,6 @@ var $AnimateCssProvider = ['$animateProvider', function($animateProvider) {
 
       var preparationClasses = [structuralClassName, addRemoveClassName].join(' ').trim();
       var fullClassName = classes + ' ' + preparationClasses;
-      var activeClasses = pendClasses(preparationClasses, ACTIVE_CLASS_SUFFIX);
```
Narretz: Where did this go? Was it unused?

src/ngAnimate/animation.js
```diff
@@ -219,7 +235,32 @@ var $$AnimationProvider = ['$animateProvider', function($animateProvider) {
         // we need to sort each of the animations in order of parent to child
         // relationships. This ensures that the child classes are applied at the
         // right time.
-        $$rAFScheduler(sortAnimations(toBeSortedAnimations));
+        var finalAnimations = sortAnimations(toBeSortedAnimations);
+        for (var i = 0; i < finalAnimations.length; i++) {
```
Narretz: Is it worth it to have this loop where all it does it filter out the first (or only) animation and to not give it a prepareClass?

test/ngAnimate/animateCacheSpec.js
```diff
@@ -0,0 +1,96 @@
...
+    var invalidEntry = { someEssentialProperty: false };
+
+    $$animateCache.put('key1', validEntry, true);
+    $$animateCache.put('key2', invalidEntry, false);
+
+    expect($$animateCache.containsCachedValidAnimation('key1')).toBe(true);
+    expect($$animateCache.containsCachedValidAnimation('key2')).toBe(false);
+  }));
+
+  it('should consider a cached entry valid if it does not exist in the cache', inject(function($$animateCache) {
```
Narretz: Do you mena `It should consider a key valid if it does not exist in the cache`?  Because a cached entry that doesn't exist in the cache is not possible.

test/ngAnimate/animationSpec.js
```diff
@@ -515,11 +522,24 @@ describe('$$animation', function() {
       }));
 
 
-      they('should add the preparation class before the $prop-animation is pushed to the queue',
+      they('should only apply the ng-$prop-prepare class if they are a child animation',
```
Narretz: `it is` or `there a child animations`

test/ngAnimate/integrationSpec.js
Narretz: I agree that the descriptions needs to be better. Fom what I can see, the new mechanism uses a cache key that combines the animation details and the parent element of the animated element. An invalid animation is one that doesn't have actual animations. $animateCss reads the animation styles at runtime, and if there's no animation or transition duration, there is no animation.
Now if you have 100 elements in a repeater, then all of them have the same parent and (probably) the same classes. If the first element doesn't have a transition / animation, then the next element doesn't have one either, so we can skip getComputedStyle, applying animation classes etc. 

If that's how it's done, then it's a clever way to increase performance when rendering large ngRepeats with ngAnimate, but with no actual animations.
But how would that affect animations if someone applies styles to a specific index in the list? As far as I can see, the animation wouldn't be detected.

Narretz:
Oh and @matsko can you add a demo that shows the speed up that this introduces?

test/ngAnimate/animationSpec.js
matsko: When the element is attached to a parent element then the colors are actually applied and getComputedStyle (which is what element.css in jQuery uses) actually figures out what the color is. Beforehand it was just using the inline style. Since we now deal with actually getting the RGB value for the color it was easier just to use a more stable value like a height measurement value.

Narretz:
Here's a plnkr to test the performance: http://plnkr.co/edit/bnaaaHbWkyQULeoOOg2X?p=preview

There's a simple ngRepeat that an ngClass on it. In my tests, ng animate with this test is about 30% faster.

However, if you include the `<input>` in the ngIf, the performance becomes pretty atrocious with both versions. There seems to be another problem there.

Regarding animating nth-child(x) elements in a repeater, that doesn't seem to work anyway: http://plnkr.co/edit/klCK5E22YPJ9Zfbz2meF?p=preview

PR #14179, opened 2016-03-04
docs(ng/compile.js): $onInit doesn't actually wait for the bindings to be initialized

cozmy:
The current docs implies that $onInit is called after all the directive's
bindings are initialized which is not the case. $onInit is called after
all required controllers of a directive are available.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -248,14 +248,11 @@
  * allow a component to have its properties bound to the controller, rather than to scope.
  *
  * After the controller is instantiated, the initial values of the isolate scope bindings will be bound to the controller
- * properties. You can access these bindings once they have been initialized by providing a controller method called
- * `$onInit`, which is called after all the controllers on an element have been constructed and had their bindings
- * initialized.
+ * properties.
  *
  * <div class="alert alert-warning">
  * **Deprecation warning:** although bindings for non-ES6 class controllers are currently
- * bound to `this` before the controller constructor is called, this use is now deprecated. Please place initialization
- * code that relies upon bindings inside a `$onInit` method on the controller, instead.
+ * bound to `this` before the controller constructor is called, this use is now deprecated.
  * </div>
  *
  * It is also possible to set `bindToController` to an object hash with the same format as the `scope` property.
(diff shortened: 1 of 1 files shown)


gkalpak:
What do you mean by _"which is not the case"_ ?
AFAIK, it _is_ indeed the case :smiley:

Can you give an example of what you mean ?

cozmy:
The documentation seems unclear to me when it uses the expression *“the bindings are initialized”*. What does this imply - that the bindings are anything but undefined?

Expressions such as *“Please place initialization code that relies upon bindings inside a $onInit method on the controller, instead."* make me think that $onInit will run when all non-optional bindings are defined.

Please see my attached example [here](https://jsfiddle.net/8s44rxfv/). $onInit runs immediately, it doesn’t even wait for a digest cycle to give the opportunity to asyncData to be “initialized”.

From the source code we can clearly observe that $onInit is simply called only after "the required controllers are binded to the controller, if `require` is an object and `bindToController` is truthy" and nothing more. This is why I believe the words "binding" and "initialized" shouldn't be used in the documentation of $onInit.

Narretz:
I agree that the docs are not super clear about this. $onInit guarantess that the initial values of the bindings are available. This is necessary because we cannot have the bindings ready for ES6 class constructors. In that case you have to use $onInit. If you are using a normal fn, you don't have to use $onInit. onInit also does not guarantee that any async binding data is available, because we have no way of knowing when the data is actually ready. This is only for binding data that is already available when the element is linked.

cozmy:
*"I agree that the docs are not super clear about this."* - that's why I believe we should clear this ambiguity. In the end, if $onInit has no link with the bindings whatsoever, why should it be specified in the documentation that "You can access these bindings once they have been initialized by providing a controller method called $onInit", "place initialization code that relies upon bindings inside a $onInit method" or *"$onInit - Called on each controller after all the controllers on an element have been constructed **and had their bindings initialized**"*.

In my opinion, if this detail remains unclear from the early stage of Angular 1.5, it can lead to even more confusion later on. Other sources have already have erroneous info [If your directive/component controller has a method called $onInit() the compiler will now call it after the component has been initialized and all its bindings have been set up.](http://angularjs.blogspot.ro/2016/02/angular-150-ennoblement-facilitation.html), [Angular 2 has the ngOnInit method, a great lifecycle callback that will help us transition from Angular 1.x. ngOnInit is called right after the directive’s data-bound properties have been checked for the first time, and before any of its children have been checked. It is invoked only once when the directive is instantiated - **much like in this Angular 1.x implementation.**](https://toddmotto.com/on-init-require-object-syntax-angular-component/) or even [my answer here](https://toddmotto.com/on-init-require-object-syntax-angular-component/) - which I should soon fix.

gkalpak:
@cozmy, I still don't get why you think it is unrelated to bindings. It is indeed executed **after** the bindings have been initialized (and in that respect it's very similiar to ng2's `ngOnInit`).

Note it sayd "initialized" not "stabilized" or something. Bindings initialization means:
If you request that `yourCtrl.foo` be bound to `parentScope.bar`, then Angular set's up watchers to update the values (as necessary depending on the type of binding, e.g. 2-way, 1-way, @-binding etc) and initializes `yourCtrl.foo` with `parentScope.bar` (whatever `parentScope.bar` is at the moment).

This is necessary, because (as @Narretz mentioned), we can't have your controllers properties initialized in the constructor if you are using a class (we can still do that for _plain old_ ES5 constructor functions). So we have to first create an instance of the controller class, then apply the bindings and then notify the constroller instance via `$onInit()`.

cozmy:
@gkalpak Thanks for clearing that up!

PR #14199, opened 2016-03-08
fix($interpolate): always unescape escaped interpolation markers

### LINKED ISSUE #14196: Escaped interpolation markers are not unescaped if there are no interpolated expressions in the same text

Bug

See http://plnkr.co/edit/b5PYWDpsYAoZ5LcDz6o7

 - the first div correctly turns the escaped interpoltion markers back into `{{ ... }}`
 - the second div doesn't, because there is no interpolation used elsewhere in the element text

 - escaped interpolation markers should always be unescaped (after interpolation)

Currently it's impossible to escape `{{ ... }}` and have it always render correctly (without resorting to hacks like always inserting an empty interpolation at the start of each value).

* **Which version of Angular, and which browser and OS does this issue affect?** Did this work in previous
versions of Angular? Please also test with the latest stable and snapshot versions.

I wasn't able to find a nightly link, but looking at the source it hasn't changed recently.

gkalpak:
Previously, whenever `mustHaveExpression` was true (e.g. when compiling a text node),
`$interpolate` would not unescape the escaped interpolation markers if there were no actual
interpolation expressions in the same string.
This commit fixes the problem, by always unescaping the escaped markers (if any).

Due to always checking for the presence of escaped interpolation markers, there is a slight
performance hit.

Fixes #14196

### DIFF
--- a/src/ng/interpolate.js
+++ b/src/ng/interpolate.js
@@ -97,12 +97,19 @@ function $InterpolateProvider() {
 
 
   this.$get = ['$parse', '$exceptionHandler', '$sce', function($parse, $exceptionHandler, $sce) {
-    var startSymbolLength = startSymbol.length,
+    var EVERY_CHAR = /./g,
+        startSymbolLength = startSymbol.length,
         endSymbolLength = endSymbol.length,
-        escapedStartRegexp = new RegExp(startSymbol.replace(/./g, escape), 'g'),
-        escapedEndRegexp = new RegExp(endSymbol.replace(/./g, escape), 'g');
+        escapedStartSymbol = startSymbol.replace(EVERY_CHAR, escapeForString),
+        escapedEndSymbol = endSymbol.replace(EVERY_CHAR, escapeForString),
+        escapedStartRegexp = new RegExp(startSymbol.replace(EVERY_CHAR, escapeForRegex), 'g'),
+        escapedEndRegexp = new RegExp(endSymbol.replace(EVERY_CHAR, escapeForRegex), 'g');
 
-    function escape(ch) {
+    function escapeForString(ch) {
+      return '\\' + ch;
+    }
+
+    function escapeForRegex(ch) {
       return '\\\\\\' + ch;
     }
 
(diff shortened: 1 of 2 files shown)


gkalpak:
I wonder if this is a BC. 

Possible breakages:

1. Even if there are no actual interpolation expressions, interpolating strings with escaped markers will return an interpolation function. Previously it would return `undefined` (and not unescape the escaped markers).
2. If you relied on the current (documented!) behavior, of not unescaping escaped markers on strings with no actual interpolation expressions.

lgalfaso:
This PR brings a new level of complexity when people want to have the
literal of value `"\{\}"` as you would need to write `\\{\\{`, this is `\`
should be escaped but only if it would get confused with an interpolation.
This is, it is not possible to write a context free function that takes one
char at a time and returns the escaped char.

This is, we should unescape everything (not only `\{`) or keep the current
behavior. The former would be a BC.

gkalpak:
Maybe I didn't understand well, but with the current behavior, `\{\{` might or might not be unescaped depending on whether an actual intetpolation exists in the same string. Doesn't it mean that a context free function won't work now either ?

lgalfaso:
@gkalpak agree, but this PR **moves** the issue from one pattern to another :)

petebacondarwin:
I think this looks good to me. The issue that @lgalfaso points out is already a problem for text blocks that contain interpolation expressions anyway. This just brings consistency between blocks that do and do not contain interpolation expressions.

petebacondarwin:
Breaking change? Well at the moment I would argue that the escaping is broken, since it is inconsistent and I doubt that anyone is really relying on these escaped sequences **not** being unescaped in text blocks that do not contain interpolation expressions.

lgalfaso:
@petebacondarwin if this PR lands, how would you write `\{\{`?

petebacondarwin:
I am not sure. But however you do it after this PR it will be the same as you would do it before the PR, in the case that there was some interpolated expression in the same text block.

One ugly way would be to do...

```
{{ "\\{\\{" }}
```

lgalfaso:
I think everybody agrees that the current is less than ideal. I would not
call it broken, but find it normal that others do think it is broken. Now,
this is what I find strange. We know that there are some limitation and
this PR removes some of them, but at the same time it adds new limitations
that worked fine before.

petebacondarwin:
The new limitation that worked before was that this PR fixes the inconsistency that if a text node did not contain any interpolations then we would not try to unescape our publicly documented escape sequence for curly braces.

I would say that this was such a weird inconsistency that no one could have been using the escaping effectively, since their server-side escaping mechanism must have been able to ascertain whether a set of double curlies, which were supposed to be rendered as-is, appeared in a text block that actually contained a set of double curlies, which were supposed to be interpolated.

I agree that we need a clear way of escaping the escape sequence.

PR #14202, opened 2016-03-09
feat($location): default hashPrefix to '!'

### LINKED ISSUE #13812: $location.hash inserts two hash signs
When on the url `index.html` when I set the hash at some point to, say, `"foo=bar"`, the url gets two hashes in it.

So, as part of a click handler or something (doesn't really matter):
```
$location.hash("foo=bar");
```
The url becomes:
```
index.html##foo=bar
```
But we should expect:
```
index.html#foo=bar
```
This is Angular 1.4.8 on Firefox 43 on Windows 10.
Also happens in Chrome 47, IE11, and Edge 13.

petebacondarwin:
The $location service is designed to support hash prefixed URLs
for cases where the browser does not support HTML5 push-state navigation.

The Google Ajax Crawling Scheme expects that local paths within a SPA start
with a hash-bang (e.g. `somedomain.com/base/path/#!/client/side/path`).

The `$locationProvide` allows the application developer to configure the
hashPrefix, and it is normal to set this to a bang '!', but the default
has always been the empty string ''.

This has caused some confusion where a user is not aware of this feature
and wonders why adding a hash value to the location (e.g. `$location.hash('xxx')`)
results in a double hash: `##xxx`.

This commit changes the default value of the prefix to '!', which is more
natural and expected.

See https://developers.google.com/webmasters/ajax-crawling/docs/getting-started

Closes #13812

BREAKING CHANGE

The hash-prefix for `$location` hash-bang URLs has changed from the empty
string "" to the bang "!". If your application does not use HTML5 mode
or is being run on browsers that do not support HTML5 mode, and you have
not specified your own hash-prefix then client side URLs will now contain
a "!" prefix. For example, rather than `mydomain.com/#/a/b/c` will become
`mydomain/#!/a/b/c`.

If you actually wanted to have no hash-prefix then you should configure
this by adding a configuration block to you application:

```
appModule.config(['$locationProvider', function($locationProvider) {
  $locationProvider.hashPrefix("");
}]);
```

### DIFF
--- a/test/ng/locationSpec.js
+++ b/test/ng/locationSpec.js
@@ -13,6 +13,28 @@ describe('$location', function() {
   });
 
 
+  describe('defaults', function() {
+    it('should have hashPrefix of "!"', function() {
+      initService({});
+      inject(
+        initBrowser({ url: 'http://host.com/base/index.html', basePath: '/base/index.html' }),
+        function($location) {
+          $location.path('/a/b/c');
+          expect($location.absUrl()).toEqual('http://host.com/base/index.html#!/a/b/c');
+        });
+    });
+
+    it('should not be html5 mode', function() {
+      initService({});
+      inject(
+        initBrowser({ url: 'http://host.com/base/index.html', basePath: '/base/index.html' }),
+        function($location) {
+          $location.path('/a/b/c');
+          expect($location.absUrl()).toContain('#!');
+        });
+    });
+  });
+
   describe('File Protocol', function() {
     /* global urlParsingNode: true */
     var urlParsingNodePlaceholder;
@@ -676,10 +698,10 @@ describe('$location', function() {
                 $location.path('/').replace();
             }
           });
-          expect($browser.url()).toEqual('http://server/base/#/home');
+          expect($browser.url()).toEqual('http://server/base/#!/home');
           $rootScope.$digest();
           expect(handlerCalled).toEqual(true);
-          expect($browser.url()).toEqual('http://server/base/#/');
+          expect($browser.url()).toEqual('http://server/base/#!/');
         }
       );
     });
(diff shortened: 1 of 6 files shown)


test/ng/locationSpec.js
```diff
@@ -13,6 +13,28 @@ describe('$location', function() {
...
+      initService({});
+      inject(
+        initBrowser({ url: 'http://host.com/base/index.html', basePath: '/base/index.html' }),
+        function($location) {
+          $location.path('/a/b/c');
+          expect($location.absUrl()).toEqual('http://host.com/base/index.html#!/a/b/c');
+        });
+    });
+
+    it('should not be hashbang mode', function() {
```
gkalpak: Isn't this test identical with the above ?

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+// Place your settings in this file to overwrite default and user settings.
```
gkalpak: Was thus file commited accidentally ?

aciccarello:
Will old links get redirected to the new #! url?

`somedomain.com/base/path/#/client/side/path` -> `somedomain.com/base/path/#!/client/side/path`

gkalpak:
@aciccarello, nope. That's part of the breaking change.
You should either migrate your URLs to including the `!` or fallback to the old behavior by explicitly setting the hashPrefix to the empty string.

.vscode/settings.json
petebacondarwin: Doh! That's teach me to start using Visual Studio Code

test/ng/locationSpec.js
petebacondarwin: The body of the test may be identical but the sentiment is different.

gkalpak:
Implementation-wise it LGTM. I'm not yet convinced about this change (from a theoretical point of view), but it sure doesn't hurt (and it's easy to "work around" it for backward compatibility).

Assuming this won't change anything in terms of how search engines treat `#!/path` vs `#/path` URLs, I'm fine merging this :smiley:

test/ng/locationSpec.js
petebacondarwin: Actually the description of this test should be `should not be htm5mode`

PR #14218, opened 2016-03-11
docs(error/numfmt): describe your change...

wowczarczyk:

docs update

Now the formatter function is used in place of with parser function and vice versa

Functions have been swapped

No
- **Other information**:

Formatter function seems to be be swapped with parser function

### DIFF
--- a/docs/content/error/ngModel/numfmt.ngdoc
+++ b/docs/content/error/ngModel/numfmt.ngdoc
@@ -44,10 +44,10 @@ directive to convert it into the format the `input[number]` directive expects.
       require: 'ngModel',
       link: function(scope, element, attrs, ngModel) {
         ngModel.$parsers.push(function(value) {
-          return '' + value;
+          return parseFloat(value, 10);
         });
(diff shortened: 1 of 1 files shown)


voodoorider:
I signed it!

pt., 11.03.2016 o 10:43 użytkownik googlebot <notifications@github.com>
napisał:

Narretz:
That change is not right.The directive adds an additional parser / formatter that converts the input into a value that the original number type parsers / formatters expect. If you open the example in a plnkr, and make your changes, you will get the numfmt error again,

voodoorider:
Still I find it a little odd and counter-intuitive that the formatter function returns numeric value and parser function returns string in terms of ```<input type="number">```

Narretz:
I think this is explained in the error message. This example directive is only needed if you want your models for input[number] to be strings. Maybe we describe better what actually happens? How would you change the description?

voodoorider:
Well after giving it a second though I think it is OK. Sorry for the confusion.

PR #14221, opened 2016-03-12
Fix broken expectation in Safari for serializing invalid dates.

### LINKED ISSUE #13415: Date Filter: Safari throws error if the date is an Invalid Date.
Safari's `JSON.stringify` doesn't handle invalid dates correctly. It throws a `RangeError: Invalid Date`.
Whenever you try to interpolate an `Invalid Date` with the date filter, it breaks.
`{{ 1/0 | date }}` http://plnkr.co/edit/z8xKljNoR1GdsMQypzlq?p=preview

In https://github.com/angular/angular.js/blob/master/src/ng/interpolate.js
`parseStringifyInterceptor` > `stringify` > `toJson` > `JSON.stringify`

This issue might be too specific for a fix to be included in Angular though. Thoughts ?
I'd be glad to provide a PR.
Something along :
```
if(isNaN(date.getTime())){
  return '';
}
```
here https://github.com/angular/angular.js/blob/master/src/ng/filter/filters.js#L489

BobChao87:

Bug fix.

Fixes #13415. Current behaviour is that Safari throws a render error instead of correctly handling Invalid Dates (such as `new Date(1/0)`).

New behaviour is that Safari performs the same as other browsers, by serializing the invalid date or any `RangeError`) to `'null'`.

No.

- [Yes] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [Yes] Tests for the changes have been added (for bug fixes / features)
- [Unnecessary] Docs have been added / updated (for bug fixes / features)

**Other information**:

Added test to prevent regression. No docs need updating since no behaviour changed and no new features were added.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1195,7 +1195,12 @@ function toJson(obj, pretty) {
   if (!isNumber(pretty)) {
     pretty = pretty ? 2 : null;
   }
-  return JSON.stringify(obj, toJsonReplacer, pretty);
+  try {
+    return JSON.stringify(obj, toJsonReplacer, pretty);
+  } catch (RangeError) {
+    // In Safari, 'Invalid Date's throw a range error. 'null' is returned in all other browsers.
+    return 'null';
+  }
 }
 
 
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1169,6 +1169,8 @@ function toJsonReplacer(key, value) {
     val = '$DOCUMENT';
   } else if (isScope(value)) {
     val = '$SCOPE';
+  } else if (isDate(value) && isNaN(value.getMilliseconds())) {
+    val = null;
   }
 
   return val;
@@ -1195,12 +1197,7 @@ function toJson(obj, pretty) {
   if (!isNumber(pretty)) {
     pretty = pretty ? 2 : null;
   }
-  try {
-    return JSON.stringify(obj, toJsonReplacer, pretty);
-  } catch (RangeError) {
-    // In Safari, 'Invalid Date's throw a range error. 'null' is returned in all other browsers.
-    return 'null';
-  }
+  return JSON.stringify(obj, toJsonReplacer, pretty);
 }
 
 
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -2060,7 +2060,7 @@ describe('angular', function() {
     });
 
     it('should serialize invalid dates to null', function() {
-      expect(toJson(new Date(1 / 0))).toEqual('null');
+      expect(toJson({when: new Date(1 / 0)})).toEqual('{"when":null}');
     });
   });
 
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -2060,7 +2060,7 @@ describe('angular', function() {
     });
 
     it('should serialize invalid dates to null', function() {
-      expect(toJson(new Date(1/0))).toEqual('null');
+      expect(toJson(new Date(1 / 0))).toEqual('null');
     });
   });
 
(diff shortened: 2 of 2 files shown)


BobChao87:
Sorry about the spam of commits. I'm not sure what I'm doing with git. It's all new to me.

I'm not sure what to do about the CLA issue since @googlebot claims it found a CLA for me, but not for the committing party, but I am the only committing party.

petebacondarwin:
The build is green now (it was just Travis being flaky)
I will find out what to do about the CLA

gkalpak:
The CLA issue is probably because at some point you merged master into your branch.
You could try to rebase removing that commit (and optionally squach your other commits into one).

src/Angular.js
```diff
@@ -1195,7 +1195,12 @@ function toJson(obj, pretty) {
   if (!isNumber(pretty)) {
     pretty = pretty ? 2 : null;
   }
-  return JSON.stringify(obj, toJsonReplacer, pretty);
+  try {
+    return JSON.stringify(obj, toJsonReplacer, pretty);
+  } catch (RangeError) {
+    // In Safari, 'Invalid Date's throw a range error. 'null' is returned in all other browsers.
+    return 'null';
```
gkalpak: This means that the whole Object will be serialized as `"null"` even if one sub-property is an invalid date.
This is inconsistent with other browsers and doesn't feel right either.

I think it is more appropriate to handle this inside the `toJsonReplacer()` function.

test/AngularSpec.js
```diff
@@ -2058,6 +2058,10 @@ describe('angular', function() {
     it('should serialize undefined as undefined', function() {
       expect(toJson(undefined)).toEqual(undefined);
     });
+
+    it('should serialize invalid dates to null', function() {
+      expect(toJson(new Date(1 / 0))).toEqual('null');
```
gkalpak: To cover more cases, put the invalid date inside an object (e.g. `toJson({when: new Date(1 / 0)})`).

IgorMinar:
I was able to manually verify the CLA signature. I'll file a bug against the bot owners to support verification for users with hidden email addresses.

petebacondarwin:
@BobChao87 - we have sorted out the CLA problems; can you address the most recent comments then we can get this merged.

BobChao87:
That should do it. The replacer was a good idea, I was clearly too focused on the exact example provided. Thanks @IgorMinar for the CLA fixes.

BobChao87:
Checks for the invalid date worked outside. They don't seem to work inside the `toJsonReplacer` because the date isn't passed in as a Date. I'll try to figure something out.

gkalpak:
Indeed. I couldn't figure something out. Maybe this problem is not solveable (at least not without having to traverse the whole object which is obviously out of question).

Or maybe you have better luck figuring something out...

Narretz:
I think as long as the issue can't be fixed satisfactorily, we'll put this into the regular 1.5 milestone.

petebacondarwin:
Here is a hack that fixes this problem:

```js
var DatetoJSON = Date.prototype.toJSON;
Date.prototype.toJSON = function() {
  try {
    return DatetoJSON.call(this);
  } catch(e) {
    if (e instanceof RangeError) {
      return null;
    }
    throw e;
  }
};
```

The question is how best to sniff for Safari and only do it for that browser?

petebacondarwin:
But since this sort of changing built in global objects is not nice, I think we cannot do this in Angular.
It is a perfectly reasonable workaround for your own apps though...

petebacondarwin:
I'll document this as a known issue, with this workaround over the weekend.
But I think we will have to leave this as it is and close this PR and the related issue.

PR #14226, opened 2016-03-13
chore(*): Upgrade to Jasmine 2.4

lgalfaso:
Chore

All tests run with Jasmine 1.3

All tests run with Jasmine 2.4

No

**Other information**:

Highlights:
in Jasmine, `this` is no longer jasmine.Spec. Removed the tests that check this.
New mechanism to run async tests as Jasmine 2 removed `runs`, `waits` and `waitsFor`
The functions `iit`, `ddescribe` and `tthey` were renamed `fit`, `fdescribe` and `fthey`
  as the originals came from Karma, Karma no longer bundles Jasmine and the new function
  name comes from Jasmine
7e916455b36dc9ca4d4afc1e44cade90006d00e3 was partially reverted. Kept the test, remove the code changes


lgalfaso:
This PR is big and given that it changes many files, can get easily rotten, so it would be ideal if this can be reviewed fast.

lgalfaso:
There is a pending task to upgrade grunt-ddescribe-iit to support for `fit`, `fdescribe` and `fthey`

lgalfaso:
I do not have access to IE9, can someone check why there is an error when running `tests:modules` ?

Narretz:
@lgalfaso The complete log shows an error for FIrefox in tests:modules:
```
Firefox 43.0.0 (Linux 0.0.0) ERROR
  Error: [ng:btstrpd] App Already Bootstrapped with this Element '&lt;div ng-app=""&gt;'
  http://errors.angularjs.org/1.5.1-build.20899+sha.0082806/ng/btstrpd?p0=%26lt%3Bdiv%20ng-app%3D%22%22%26gt%3B
  at /home/travis/build/angular/angular.js/build/angular.js:68
.................
Firefox 43.0.0 (Linux 0.0.0): Executed 17 of 713 ERROR (0.126 secs / 0.027 secs)
```

IE 9 is fine:
```
IE 9.0.0 (Windows 7 0.0.0): Executed 713 of 713 SUCCESS (2.888 secs / 2.348 secs)
```

Narretz:
Why was https://github.com/lgalfaso/angular.js/commit/7e916455b36dc9ca4d4afc1e44cade90006d00e3 reverted? Because we don't support these browsers?

lgalfaso:
> Why was lgalfaso@7e91645 reverted? Because we don't support these browsers?

Jasmine 2.4 `toEqual` checks that actual and expected are of the same type. 7e91645 changes the type of what is thrown.

Narretz:
Ok, thanks. Can you add that explanation to the commit message?

test/loaderSpec.js
```diff
@@ -47,22 +47,29 @@ describe('module loader', function() {
...
+      ['$provide', 'constant', createArgumentsOf(['abc', 123])],
+      ['$provide', 'decorator', createArgumentsOf(['dk', 'dv'])],
+      ['$provide', 'provider', createArgumentsOf(['sk', 'sv'])],
+      ['$provide', 'factory', createArgumentsOf(['fk', 'fv'])],
+      ['$provide', 'service', createArgumentsOf(['a', 'aa'])],
+      ['$provide', 'value', createArgumentsOf(['k', 'v'])],
+      ['$filterProvider', 'register', createArgumentsOf(['f', 'ff'])],
+      ['$compileProvider', 'directive', createArgumentsOf(['d', 'dd'])],
+      ['$compileProvider', 'component', createArgumentsOf(['c', 'cc'])],
+      ['$controllerProvider', 'register', createArgumentsOf(['ctrl', 'ccc'])]
```
petebacondarwin: Is it possible to just use `jasmine.objectContaining` for these tests?

petebacondarwin:
@lgalfaso can you confirm that after the change to the angular mock module tests still work on PhantomJS and Safari 7?  This is the only part that affects people outside of this project so it would be good to make sure we are not breaking anything.

lgalfaso:
Reverted the changes on angular.mocks.js, made the changes on loaderSpec.js, fixed the issue with Firefox and now the test are back to green.

PTAL

test/helpers/privateMocksSpec.js
```diff
@@ -9,7 +9,7 @@ describe('private mocks', function() {
         spyOn(window, 'it');
 
         they('should do stuff with $prop', ['a', 'b', 'c']);
-        expect(window.it.calls.length).toEqual(3);
```
Narretz: All of these could be converted instead to `.calls.count()).toEqual(...` This would make it similar to what is was before.

Narretz:
Since we now move to jasmine 2.4, I think we should also change the docs to recommend jasmine 2.4 instead of 1.3, espepcially here: https://github.com/angular/angular.js/blob/a4e60cb6970d8b6fa9e0af4b9f881ee3ba7fdc99/docs/content/guide/unit-testing.ngdoc and https://github.com/angular/angular.js/blob/a4e60cb6970d8b6fa9e0af4b9f881ee3ba7fdc99/docs/content/guide/e2e-testing.ngdoc The unit tests in the guide don't actually run, so they have to be checked for correctness with jasmine 2.4. @lgalfaso do you want to make these changes? I can also do it if you want.

test/helpers/privateMocksSpec.js
lgalfaso: I still think the new way is better and the way that is present in Jasmine documentation

lgalfaso:
Updated the docs, did not run the tests in the guide, but in theory it
should be possible for users to use the version from Jasmine that they like
the most.

PR #14227, opened 2016-03-13
feat($compile): support dynamic transclusion slots

odolha:
New feature: dynamic transclusion slots.

Currently (1.5.x), a directive/component with transclusion slots must specify up-front the slots, using the transclude: { ... } approach.

With this PR, it is possible for consumers of such a directive/component to specify the transclusion slots themselves (in a special attribute called 'transclude-slots' which will be parsed before the component is compiled). This allows more abstract and flexible components - see the new example added in ngTransclude docs.

This PR doesn't introduce any breaking change.

Reference issue (starting point for PR work): http://stackoverflow.com/questions/35165202/angular-multiple-transclude-dynamic-number-of-elements.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -536,14 +536,16 @@
  * Testing Transclusion Directives}.
  * </div>
  *
- * There are three kinds of transclusion depending upon whether you want to transclude just the contents of the
- * directive's element, the entire element or multiple parts of the element contents:
+ * There are four kinds of transclusion depending upon whether you want to transclude just the contents of the
+ * directive's element, the entire element, multiple parts of the element contents, or multiple parts dynamically:
  *
  * * `true` - transclude the content (i.e. the child nodes) of the directive's element.
  * * `'element'` - transclude the whole of the directive's element including any directives on this
  *   element that defined at a lower priority than this directive. When used, the `template`
  *   property is ignored.
  * * **`{...}` (an object hash):** - map elements of the content onto transclusion "slots" in the template.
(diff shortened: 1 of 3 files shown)


Narretz:
I'm not a big fan, as it makes the API even more complex and I don't really see a big requirement of having an arbitrary number of transclusion slots. But let's see what others have to say.

odolha:
Indeed it makes it a bit more complex.

Yet, in all fairness - multi-slot transclusion itself is still new, makes things more complex but it was really necessary (in my opinion).

Personally I've encountered the requirements for this to be dynamic very soon after migrating to 1.5 and. One use case is explained in the example I provided.

Currently we duplicate some code because this feature is missing, particualary template code.

We considered various other approaches, particulary using 'require' and extending controllers.

The only feasable and readable approach we found so far is to extract parts of the abstract component into other smaller components (e.g. the header and content cells of a table) and then re-asamble them into the specific needs of each case.

Of course other more unintuitve ways to avoid code duplication are possible (e.g. writing some content manipulation directives), but we prefer a clean, clear and simple direction.

If anyone has other suggestions on how to solve the use case without dynamic slots, let's discuss.

dapriett:
+1

nmindz:
+1. This would reduce a few templates by aprox. 80% of it's original length if I were able to use an arbitrary number of transclusion slots w/o being bound by the `transcluyde: { ... }` syntax.

rodonako:
+1

premkumar-gg:
+1

PR #14231, opened 2016-03-14
add comma to docs

nonopolarity:
just a doc change

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -514,7 +514,7 @@ var locationPrototype = {
    * @param {string|Object.<string>|Object.<Array.<string>>} search New search params - string or
    * hash object.
    *
-   * When called with a single argument the method acts as a setter, setting the `search` component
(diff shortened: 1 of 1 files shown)


robinj:
I think that, because of the second comma in the sentence, the comma that you are adding doesn't make sense. It makes "the method acts as a setter" seem like a parenthetical component.

jianlin:
that's because in Strunk and White, it mentioned that when there is a light pause in the sentence, then the comma belongs there.  

If the sentence is:

> When called with a single argument the method acts as a setter.

It can be:

> When called with a single argument, the method acts as a setter.

so if we add the further explanation part:  

> setting the `search` component.

then it is

> the method acts as a setter, setting the `search` component.

making it

> When called with a single argument, the method acts as a setter, setting the `search` component.

Narretz:
Strunk & White isn't always the best advice wrt style issues. I think it's better to leave it as-is. But thanks for the PR, keep them coming

PR #14233, opened 2016-03-14
fix(ngOptions): select disabled option when set from model

### LINKED ISSUE #12756: `select` with disabled option in model: value not showing in closed state
#### Overview of the Issue

When using `select` with ngOptions' `disable` keyword, and the model value is one that is disabled through the ngOptions directive, the select element doesn't show the model value.

The UI should show the value, but upon opening the select dropdown, the value should appear disabled.

What actually happens is, that the select shows as if there was no value in the model.

#### Motivation for or Use Case

A model value can come from a record in a database, e.g. an option that was valid but no longer. In this case, the select doubles as display and input element. Due to this, it doesn't display a disabled value.

#### Angular Version(s)

1.4.5 (as `disable` is new to 1.4)

#### Browsers and Operating System
Tested on OS X 10.10 Chrome, FF, Safari 
#### Reproduce the Error
http://plnkr.co/edit/unjDAvAUnBimn29URkSP?p=preview
#### Related Issues
Possibly #11762

Narretz:
bug fix

See #12756

Probably not, when we understand it as a bug fix.

**Other information**:

When a disabled option is set as the ngModel, ngOptions will now select the option
in the select element, which will set select.val() to the option hash value and visually
show the option value / label as selected in the select box. Previously, disabled
options forced the unknown value.
The previous behavior is inconsistent with both default HTML behavior and select with
ngModel but without ngOptions. Both allow disabled values to be selected programmatically.

Common use cases for this behavior include options that where previously valid, but have
been disabled, and cannot be selected again.

Fixes #12756

### DIFF
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -771,7 +771,7 @@ describe('ngOptions', function() {
       });
 
 
-      it('should not select disabled options when model changes', function() {
+      it('should select disabled options when model changes', function() {
         scope.options = [
           { name: 'white', value: '#FFFFFF' },
           { name: 'one', value: 1, unavailable: true },
@@ -792,11 +792,14 @@ describe('ngOptions', function() {
         scope.$apply('selected = 1');
         options = element.find('option');
 
-        expect(element.val()).toEqualUnknownValue('?');
-        expect(options.length).toEqual(5);
-        expect(options.eq(0).prop('selected')).toEqual(true);
+        // jQuery returns null for val() when the option is disabled, see
+        // https://bugs.jquery.com/ticket/13097
+        expect(element[0].value).toBe('number:1');
+        expect(options.length).toEqual(4);
+        expect(options.eq(0).prop('selected')).toEqual(false);
+        expect(options.eq(1).prop('selected')).toEqual(true);
         expect(options.eq(2).prop('selected')).toEqual(false);
-        expect(options.eq(4).prop('selected')).toEqual(false);
+        expect(options.eq(3).prop('selected')).toEqual(false);
       });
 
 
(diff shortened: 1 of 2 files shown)


test/ng/directive/ngOptionsSpec.js
```diff
@@ -2515,37 +2521,6 @@ describe('ngOptions', function() {
       expect(element.find('option')[1].selected).toBeTruthy();
     });
 
-    it('should not write disabled selections from model', function() {
```
gkalpak: What is the duplicate test ?

gkalpak:
Are you sure we don't need to also ignore `.disabled` in `selectCtrl.readValue()` ?

Also, I can't decide if this is a BC or not ? Sure the new behavior seems more consistent and the old one was not documented, but still...

test/ng/directive/ngOptionsSpec.js
Narretz: This one :) It's a duplicate of https://github.com/angular/angular.js/pull/14233/files#diff-2cda0d7a6d8ec47da3560fbe59a634e3R850

Narretz:
Hmm ... readValue is called when the select changes, so on user input. At this point, the selected option shouldn't be disabled anyay, becasue the browser prevents selecting a disabeld option. So it's an additional check that would only come into play when someone selectes a disabled option programmtically and then triggers change. I think the check is okay.

I'm fine with putting it into 1.6, but I also think that this is the expected behavior. I also don't think that anyone's application could break by showing the disabled value. The model is the same in both cases.

Narretz:
@petebacondarwin wdyt, 1.5 / 1.4 or 1.6 with Breaking Change notice?

gkalpak:
> Hmm ... readValue is called when the select changes, so on user input.

AFAICT, `readValue()` is read inside `updateOptions()` which is invoked whenever the collection returnd by `ngOptions.getWatchables()` changes. I suspect that under certain circumstances, a change could be incurred programmatically to the collection returned by `ngOptions.getWatchables()`, but I don't have evidence (and not because I didn't try :stuck_out_tongue:).

So, this is just a hunch. And even if I proved correct, it would only affect some very cornery corner cases, so feel free to ignore.

--
On the other hand, `readValue` reads the current value of the `<select>` element and tries to match it to an option item. I don't think we should care much, how the value ended up into the `<select>` (be it through user-interaction or other means (e.g. rogrammatically)) - although most browsers don't allow selecting a disabled option.
This would also be more consistent with `select without ngOptions` (which just returns the value of the `<select>` element, not caring how it came to be.

Theoretically thinking about it, the `disable when` construct's purpose should be to set the `disabled` property of the respective `<option>` element. How the underlying implementation chooses to handle this (be it a native `<select>` or even a custom one) is up to them.

--
That's all too theoretical and minor though. I don't feel strongly either way.
(Besides I might be missing some other side effects of returning disabled options from `readValue()`.)

petebacondarwin:
Given that this is really only a visual change let's put it in 1.5.x

petebacondarwin:
Please merge @Narretz

PR #14234, opened 2016-03-14
fix(ngAnimate.$animate): remove animation callbacks when the element …

Narretz:
fixes a small memory leak and fixes test to ensure expected behavior

$animate.on() listeners are not removed when the element is removed from the DOM

**Other information**:

…is removed

The test for this didn't actually test the listener removal. The addClass animation after
the element removal didn't start because the enter animation was still in progress.


gkalpak:
LGTM

PR #14236, opened 2016-03-14
docs(ng.$log): show how log methods are bound to console

glebec:

Docs update

The fact that `$log` methods are applied to the `console` regardless of invocation context is useful but not noted anywhere.

An example is provided of how this pre-binding streamlines the use of `$log` methods as handler functions, e.g. for `$http` promises.

No.

**Other information**:

N/A.

### DIFF
--- a/src/ng/log.js
+++ b/src/ng/log.js
@@ -10,6 +10,14 @@
  * into the browser's console (if present).
  *
  * The main purpose of this service is to simplify debugging and troubleshooting.
+ * Also, all `$log` methods are applied directly to `console`, streamlining using
+ * `$log` methods as handler functions (e.g. for {@link ng.$http $http}):
+ *
+ * ```js
+ *   $http.get('/api/widgets')
+ *   .then(function (response) { use(response); })
+ *   .catch($log.error) // replaces `console.error.bind(console)` or a wrapping function
+ * ```
  *
  * The default is to log `debug` messages. You can use
  * {@link ng.$logProvider ng.$logProvider#debugEnabled} to change this.


gkalpak:
I don't think we want to document this (so people start relying on it).
It is an implementation detail and we should retain the flexibility to change it any time.

glebec:
That's fair. For me it's one of the only reasons I use `$log` over `console`, so it seemed like a feature, even if accidental. But I can also accept that maybe it's not intended to be a feature. My questions would then be:

* How likely will it be that in the future, the `$log` methods will use the `this` keyword? If the answer is "conceivably that will happen," that lends more credence to not documenting this otherwise-useful characteristic of the `$log` service. If the answer is "almost certainly not," I would posit that this documentation provides a compelling use case for `$log`.
* If you *do* incorporate this PR, test cases might serve as an effective guard against accidentally or casually removing this functionality, at least not without altering the documentation to match and versioning it as a breaking change. Would that make this PR more or less palatable from your perspective?

Regards, —G.

gkalpak:
> How likely will it be that in the future, the `$log` methods will use the `this` keyword?

Neither very likely, nor very unlikely. There have always been thoughts about providing a more powerful implementation for `$log` (one that might more heavily rely on prototype and `this`) - not very likely to happen, but there have been discussions :stuck_out_tongue: Or maybe we decide to delegate directly to `console` for whatever reason. Furthermore, if people start relying on `$log` methods not being context-sensitive, then all 3rd-party implementation that decorate or eplace `$log` will need to also ensure they have "context-insensitive" methods.

> test cases might serve as an effective guard

It's not about accidentally breaking it, it's more about being free to break it. By exposing implementation details, we tie ourselves to specific implementations, which makes it harder to swap them for something better.

--
Also note we are discussing about `$log.error` vs `$log.error.bind($log)` - it's really a minimal overhead :smiley:

I'll leave this open though - maybe be someone disagrees...

glebec:
> It's not about accidentally breaking it, it's more about being free to break it.

I hear you. Okay, I don't feel very strongly about it in any case, it's a small thing. ;-) Cheers, —G.

petebacondarwin:
Closing as won't fix. Thanks for the PR @glebec

PR #14237, opened 2016-03-14
docs(guide/Component Router): describe your change...

Josh68:

Docs update

Method $routerCanActivate lifecycle hook does not appear to exist for components in 1.x

The $canActivate hook on the CDO appears to be the only way to control route resolution for components in 1.x

No

**Other information**:

NOTES:

From what I can tell, the only way of "resolving" for a component route is in the component's definition object and the method is $canActivate, not $routerCanActivate. Adding a $routerCanActivate alternative would be helpful as a method that can be called on the constructor.

I did find a helpful min-safe pattern that might be worth documenting. This is the ability to inject into the CDO's $canActivate method by putting to-be-injected module strings AFTER the default parameters ($nextInstruction and $prevInstruction), followed by the callback with the injected modules as the 3rd - xx arguments. This way I can reference a service to check authentication status (for example), returning a promise that will resolve to true or false and thereby either permit the route to proceed or kill it.

However, the best feature would be to add some way to reference a routeProvider, as with ngRoute, for global $canActivate patterns. A pattern I had successfully used with ngRoute was to extend the "when" method to do my authentication check and abort any route that had a configuration setting like "requiresAuth." To be able to do the same with component-based routes would be helpful, but I haven't figured out how. Thanks.

### DIFF
--- a/docs/content/guide/component-router.ngdoc
+++ b/docs/content/guide/component-router.ngdoc
@@ -956,9 +956,9 @@ respectively.
 
 **How do I prevent navigation from occurring?**
 
-Each **Component** can provide the `$routerCanActivate` and `$routerCanDeactivate` **Lifecycle Hooks**. The
-`$routerCanDeactivate` hook is an instance method on the **Component**. The `$routerCanActivate` hook is a
-static method defined on either the **Component Definition Object** or the **Component's** constructor function.
+Each **Component** can provide the `$canActivate` and `$routerCanDeactivate` **Lifecycle Hooks**. The
+`$routerCanDeactivate` hook is an instance method on the **Component**. The `$canActivate` hook is used as a
+static method defined on the **Component Definition Object**.
 
 The **Router** will call these hooks to control navigation from one **Route** to another. Each of these hooks can
 return a `boolean` or a Promise that will resolve to a `boolean`.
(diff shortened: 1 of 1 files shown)


Josh68:
I have signed the CLA. Thanks,

Josh Schneider

petebacondarwin:
While this change is accurate, right now. I actually want to change the hook's name to be `$routerCanActivate` to keep it consistent with the other hooks.

Josh68:
I kind of thought that might be the case. As an option, maybe it should be made very clear on the new documentation (like right at the top) what the exact status of ngcomponentrouter is. I'm sure I'm not alone in trying to setup a new 1.5.x project and take advantage of components and the component router, but the router itself seems to be an afterthought to 2.x development. I'm sure the team is ridiculously busy, so if you have any recommendations for other mods to the 1.5.x component router docs, I will put in a different PR. Thanks.

petebacondarwin:
OK, let's merge this since it is the status quo. There will have to be a breaking change announcement with a rename of the hook anyway

Josh68:
Thanks, Pete.

On Wed, Mar 16, 2016 at 9:50 AM, Pete Bacon Darwin <notifications@github.com

PR #14242, opened 2016-03-15
fix(ngMessages): don't crash when first transcluded message is removed

### LINKED ISSUE #14183: App freezes when using a directive that transcludes ng-message
I'm trying to create a directive that contains some common css classes and hide/show logic that I use for all my `ng-messages`. it looks like this

```
var StandardMessagesComponent = function() {
    
}

angular.module('myApp').component('standardMessages', {
    controller: StandardMessagesComponent,
    controllerAs: 'ctrl',
    templateUrl: 'standardMessages.html',
    transclude: true,
    bindings: {
        form: '<',
        field: '@'
    }
});
```

```
<ng-messages for="ctrl.form[ctrl.field].$error" role="alert" class="standard-messages"
             ng-if="ctrl.form[ctrl.field].$error && (ctrl.form.$submitted || ctrl.form[ctrl.field].$touched)"
>
    <ng-transclude></ng-transclude>
</ng-messages>
```

It is causing my app to freeze when the following is true:
-there is a validation error causing one of the `ng-message`s to be shown
-there are multiple `ng-message` elements
-The user navigates to another page

The following plunkr reproduces the issue: https://plnkr.co/edit/Qi2JQ4RioAOxZG98uCou?p=preview
To crash the app click the input and then click the "other page" link.

I'm using Angular 1.5, angular-messages 1.5, and ui-router 0.2.18. This issue exists in Chrome and Firefox but works fine in IE11 (although it freezes for a second)

Narretz:

**Other information**:

Under specific circumstances, ngMessages would go into an infinite loop and crash the
browser / page:
- At least two ngMessage elements are wrapped inside another element (e.g. ngTransclude)
- The first message is currently visible
- The first message is removed (e.g. when the whole ngMessages element is removed by an ngIf)

When a message is removed, it looks for a previous message - in this specific case it would misidentify
the second message for a previous message, which would then cause the first message to be marked as the
second message's next message, resulting in an infinite loop, and crash.

This fix ensures that only messages with an index lower than the starting message are considered
previous elements.

Fixes #14183

@gkalpak you triaged the original issue, do you mind reviewing this?

### DIFF
--- a/test/ngMessages/messagesSpec.js
+++ b/test/ngMessages/messagesSpec.js
@@ -485,6 +485,126 @@ describe('ngMessages', function() {
     });
   });
 
+  describe('ngMessage nested nested inside elements', function() {
+
+    it('should not crash or leak memory when the messages are transcluded, the first message is ' +
+      'visible, and ngMessages is removed by ngIf', function() {
+
+      module(function($compileProvider) {
+        $compileProvider.directive('messageWrap', function() {
+          return {
+            transclude: true,
+            scope: {
+              col: '=col'
+            },
+            template: '<div ng-messages="col"><ng-transclude></ng-transclude></div>'
+          };
+        });
+      });
+
+      inject(function($rootScope, $compile) {
+
+        element = $compile('<div><div ng-if="show"><div message-wrap col="col">' +
+                           '        <div ng-message="a">A</div>' +
+                           '        <div ng-message="b">B</div>' +
+                           '</div></div></div>')($rootScope);
+
+        $rootScope.$apply(function() {
+          $rootScope.show = true;
+          $rootScope.col = {
+            a: true,
+            b: true
+          };
+        });
+
+        expect(messageChildren(element).length).toBe(1);
+        expect(trim(element.text())).toEqual('A');
+
+        $rootScope.$apply('show = false');
+
+        expect(messageChildren(element).length).toBe(0);
+      });
+    });
+
+
+    it('should not crash when the first of two nested messages is removed', function() {
+      inject(function($rootScope, $compile) {
+
+        element = $compile(
+          '<div ng-messages="col">' +
+            '<div class="wrapper">' +
+              '<div remove-me ng-message="a">A</div>' +
+              '<div ng-message="b">B</div>' +
+            '</div>' +
+          '</div>'
+        )($rootScope);
+
+        $rootScope.$apply(function() {
+          $rootScope.col = {
+            a: true,
+            b: false
+          };
+        });
+
+        expect(messageChildren(element).length).toBe(1);
+        expect(trim(element.text())).toEqual('A');
+
+        var ctrl = element.controller('ngMessages');
+        var deregisterSpy = spyOn(ctrl, 'deregister').and.callThrough();
+
+        var nodeA = element[0].querySelector('[ng-message="a"]');
+        jqLite(nodeA).remove();
+        $rootScope.$digest(); // The next digest triggers the error
+
+        // Make sure removing the element triggers the deregistration in ngMessages
+        expect(trim(deregisterSpy.calls.mostRecent().args[0].nodeValue)).toBe('ngMessage: a');
+        expect(messageChildren(element).length).toBe(0);
+      });
+    });
+
+
+    it('should not crash, but show deeply nested messages correctly after a message ' +
+      'has been removed', function() {
+      inject(function($rootScope, $compile) {
+
+        element = $compile(
+          '<div ng-messages="col" ng-messages-multiple>' +
+            '<div class="another-wrapper">' +
+              '<div ng-message="a">A</div>' +
+              '<div class="wrapper">' +
+                '<div ng-message="b">B</div>' +
+                '<div ng-message="c">C</div>' +
+              '</div>' +
+              '<div ng-message="d">D</div>' +
+            '</div>' +
+          '</div>'
+        )($rootScope);
+
+        $rootScope.$apply(function() {
+          $rootScope.col = {
+            a: true,
+            b: true
+          };
+        });
+
+        expect(messageChildren(element).length).toBe(2);
+        expect(trim(element.text())).toEqual('AB');
+
+        var ctrl = element.controller('ngMessages');
+        var deregisterSpy = spyOn(ctrl, 'deregister').and.callThrough();
+
+        var nodeB = element[0].querySelector('[ng-message="b"]');
+        jqLite(nodeB).remove();
+        $rootScope.$digest(); // The next digest triggers the error
+
+        // Make sure removing the element triggers the deregistration in ngMessages
+        expect(trim(deregisterSpy.calls.mostRecent().args[0].nodeValue)).toBe('ngMessage: b');
+        expect(messageChildren(element).length).toBe(1);
+        expect(trim(element.text())).toEqual('A');
+      });
+    });
+  });
+
   describe('when including templates', function() {
     they('should work with a dynamic collection model which is managed by ngRepeat',
       {'<div ng-messages-include="...">': '<div ng-messages="item">' +
@@ -690,6 +810,37 @@ describe('ngMessages', function() {
       expect(messageChildren(element).length).toBe(1);
       expect(trim(element.text())).toEqual("C");
     }));
+
+
+    it('should properly detect a previous message, even if it was registered later',
+      inject(function($compile, $rootScope, $templateCache) {
+        $templateCache.put('include.html', '<div ng-message="a">A</div>');
+        var html =
+            '<div ng-messages="items">' +
+              '<div ng-include="\'include.html\'"></div>' +
+              '<div ng-message="b">B</div>' +
+              '<div ng-message="c">C</div>' +
+            '</div>';
+
+        element = $compile(html)($rootScope);
+        $rootScope.$apply('items = {b: true, c: true}');
+
+        expect(element.text()).toBe('B');
+
+        var ctrl = element.controller('ngMessages');
+        var deregisterSpy = spyOn(ctrl, 'deregister').and.callThrough();
+
+        var nodeB = element[0].querySelector('[ng-message="b"]');
+        jqLite(nodeB).remove();
+
+        // Make sure removing the element triggers the deregistration in ngMessages
+        expect(trim(deregisterSpy.calls.mostRecent().args[0].nodeValue)).toBe('ngMessage: b');
+
+        $rootScope.$apply('items.a = true');
+
+        expect(element.text()).toBe('A');
+      })
+    );
 
   });
 
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -410,6 +410,13 @@ angular.module('ngMessages', [])
 
          $scope.$watchCollection($attrs.ngMessages || $attrs['for'], ctrl.render);
 
+         // If the element is destroyed, proactively destroy all the currently visible messages
+         $element.on('$destroy', function() {
+           forEach(messages, function(item) {
+             item.message.detach();
+           });
+         });
+
          this.reRender = function() {
            if (!renderLater) {
              renderLater = true;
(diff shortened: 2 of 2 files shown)


test/ngMessages/messagesSpec.js
```diff
@@ -412,6 +412,90 @@ describe('ngMessages', function() {
   }));
 
 
+  it('should not crash when the messages are transcluded, the first message is visible, and ngMessages is removed by ngIf', function() {
```
gkalpak: This line is too long. Please wrap at 100 chars :grin:

test/ngMessages/messagesSpec.js
```diff
@@ -412,6 +412,90 @@ describe('ngMessages', function() {
...
+          required: true,
+          extra: true
+        };
+      });
+
+      expect(messageChildren(element).length).toBe(1);
+      expect(trim(element.text())).toEqual('Fill in the text field.');
+
+      $rootScope.$apply(function() {
+        $rootScope.col.required = true;
```
gkalpak: This is redundant.

test/ngMessages/messagesSpec.js
```diff
@@ -412,6 +412,90 @@ describe('ngMessages', function() {
...
 
 
+  it('should not crash when the messages are transcluded, the first message is visible, and ngMessages is removed by ngIf', function() {
+
+    module(function($compileProvider) {
+      $compileProvider.directive('messageWrap', function() {
+        return {
+          transclude: true,
+          scope: {
+            col: '=col'
```
gkalpak: OOC, what does `col` stand for ?

gkalpak:
I believe this fix could result in different behavior, since now the order of messages is determined by the order of registration and not their place in the DOM. E.g.:

```js
<ng-messages for="someObj">
  <include-messages>
    <!-- Some directive with a `templateUrl` containing:
      <ng-message when="error1">...</ng-message>
      <ng-message when="error2">...</ng-message>
    -->
  </include-messages>
  <ng-message when="error3">...</ng-message>
  <ng-message when="error4">...</ng-message>
</ng-messages>
```

--
@Narretz, can you confirm that this is indeed tha case ?
If yes, can we avoid it ?

test/ngMessages/messagesSpec.js
Narretz: collection, I think

test/ngMessages/messagesSpec.js
Narretz: maxlength in jshint is 200, but I'll do it for you :)

Narretz:
@gkalpak Thanks for looking into this. As far as I can see, I don't think the order is compromised. I've added a test to make sure it doesn't happen. Does the test reflect the case you had in mind? You can also add another test, don't be shy ;>

I've also fixed a memory leak with ngTransclude in the test that happens when ngIf destroys the ngMessages container. If only one message is currently visible, ngMessages deregisters this message and then tries to render the second message. The second message never gets its destroy handler triggered, but somehow holds onto the transclusion scope, which can't be collected then. I'm not sure if that's an issue in the real world, though.

gkalpak:
The problem becomes obvious when removing one of the messages. Here is a failing testcase:

```js
  it('should properly detect the previous message, even if it was registered later',
    inject(function($compile, $rootScope, $templateCache) {
      $templateCache.put('include.html', '<div ng-message="a">A</div>');
      var html =
          '<div ng-messages="items">' +
            '<div ng-include="\'include.html\'"></div>' +
            '<div ng-message="b">B</div>' +
            '<div ng-message="c">C</div>' +
          '</div>';

      element = $compile(html)($rootScope);
      $rootScope.$apply('items = {b: true, c: true}');

      expect(element.text()).toBe('B');

      var nodeB = element[0].querySelector('[ng-message="b"]');
      angular.element(nodeB).remove();
      $rootScope.$apply('items.a = true');

      expect(element.text()).toBe('A');
    })
  );
```

(Removing the element like this is kind of contrived, but I am pretty sure the same issues could be reproduced with `ngIf` and a remote template for example.)

Narretz:
Thanks for the test!
This is definitely a problem. I've pushed a new commit that makes this and the original test pass, but it only works if there's just one level of nesting when you remove the element directly:

```
        '<div ng-messages="col">' +
          '<div class="wrapper">' +
            '<div remove-me ng-message="required">Fill in the text field.</div>' +
            '<div ng-message="extra">Extra error message.</div>' +
          '</div>' +
        '</div>'
```

More levels of nesting still produce the same problem:
        
       '<div ng-messages="col">' +
          '<div class="another-wrapper">' +
            '<div class="wrapper">' +
              '<div remove-me ng-message="required">Fill in the text field.</div>' +
              '<div ng-message="extra">Extra error message.</div>' +
            '</div>' +
            '<div ng-message="different">Different error message.</div>' +
          '</div>' +
        '</div>'

It's quite difficult to maintain a notion of previous element, when the nesting becomes more complex.
But the problem will only happen if someone removes a single message from the DOM, something that shouldn't be necessary. The original issue - removing the complete ngMessages element with an ngIF - is already fixed by removing all message nodes when the ngMessages element is destroyed.
So we basically two options - just fix the original issue, or improve the querying for previousNodes. Maybe a top-down search is better then searching from the comment element itself.

gkalpak:
TBH, I'm still not sure what caused the original issue. I read your description in the commit message:

> When a message is removed, it looks for a previous message - in this specific case it would misidentify the second message for a previous message, which would then cause the first message to be marked as the second message's next message, resulting in an infinite loop, and crash.

But I haven't looked into why/how can the 2nd message be identified as previous to the 1st one (is it because it becomes the `ctrl.head`) ? And even then, why does the 1st message become the `next` for the 2nd one ?
Maybe we could fix that logic.

Narretz:
So the problem is in this function: https://github.com/angular/angular.js/blob/aa077e81129c740041438688dff2e8d20c3d7b52/src/ngMessages/messages.js#L444

When you have DOM like this: 
        '<div ng-messages="col">' +
          '<div class="wrapper">' +
            '<div remove-me ng-message="required">Fill in the text field.</div>' +
            '<div ng-message="extra">Extra error message.</div>' +
          '</div>' +
        '</div>'

and the "required" message element gets removed, findPreviousMessage looks upwards in the DOM to find a previous message. Since required has no previousSibling, the wrapper element is the next prevNode: https://github.com/angular/angular.js/blob/aa077e81129c740041438688dff2e8d20c3d7b52/src/ngMessages/messages.js#L459. The code then assigns the last child of the wrapper as the prevNode, and since this is the "extra" message it is mis-identied as a previous node. Later, the "required" message is set as the next message for the "required" message: https://github.com/angular/angular.js/blob/aa077e81129c740041438688dff2e8d20c3d7b52/src/ngMessages/messages.js#L485

The problem is really that the nesting inside ngMessages can be arbitrarily complicated, but this wasn't really planned, because you should only use ng-message and ng-message-include. I can understand if someone has an ng-transclude, but more nesting shouldn't really be necessary.

Searching for previous nodes can definitely be improved, but I'm not sure if it's worthwhile. One possibility is to start the search from the ngMessages element, and stop when you find the passed in the comment element.

gkalpak:
From a quick look, it seems that the actual bug is on [messages.js#L455][1]: It shouldn't look into the `comment` element's parent, because it has already looked in there (since it started at `comment`.

I haven't tried it, but I believe that initializing `parentLookup` with `comment.parentNode` on [messages.js#L446][2] should fix the bug. E.g.:

```js
var parentLookup = [comment.parentNode];
```

[1]: https://github.com/angular/angular.js/blob/aa077e81129c740041438688dff2e8d20c3d7b52/src/ngMessages/messages.js#L455
[2]: https://github.com/angular/angular.js/blob/aa077e81129c740041438688dff2e8d20c3d7b52/src/ngMessages/messages.js#L446

Narretz:
This fix works if there's only one level of nesting. It's the same as doing:
```
             if (prevNode.childNodes.length &&
                prevNode !== comment.parentNode && // this is new
                parentLookup.indexOf(prevNode) == -1
              ) {
               parentLookup.push(prevNode);
               prevNode = prevNode.childNodes[prevNode.childNodes.length - 1];
             } else {
               prevNode = prevNode.previousSibling || prevNode.parentNode;
             }
```

But both fixes don't work if there's a DOM like this:

        '<div ng-messages="col">' +
          '<div class="another-wrapper">' +
            '<div class="wrapper">' +
              '<div remove-me ng-message="required">Fill in the text field.</div>' +
              '<div ng-message="extra">Extra error message.</div>' +
            '</div>' +
            '<div ng-message="different">Different error message.</div>' +
          '</div>' +
        '</div>'

This is probably a very unusual case, but this will also cause an infinite loop, so we should probably fix it. Maybe merge the patch now and then work on this later?

gkalpak:
Taking a more careful look, the problem seems to be that once we traverse all `previousSiblings`, we go to the `parentNode` and from there we start at the last `childNode` (which is essentially a `nextSibling` to the original `comment`.

Basically, if we reach at a `parentNode` after having examined one of its `childNodes`, we should never go back in, since it is guaranteed that we won't find an actual "previous message" in there.

This is how `findPreviousMessage` should look like:

```js
function findPreviousMessage(parent, comment) {
  var prevNode = comment;
  var parentLookup = [];
  while (prevNode && prevNode !== parent) {
    var prevKey = prevNode.$$ngMessageNode;
    if (prevKey && prevKey.length) {
      return messages[prevKey];
    }

    // dive deeper into the DOM and examine its children for any ngMessage
    // comments that may be in an element that appears deeper in the list
    if (prevNode.childNodes.length && parentLookup.indexOf(prevNode) === -1) {
      parentLookup.push(prevNode);
      prevNode = prevNode.childNodes[prevNode.childNodes.length - 1];
    } else if (prevNode.previousSibling) {
       prevNode = prevNode.previousSibling;
    } else {
      prevNode = prevNode.parentNode;
      parentLookup.push(prevNode);
    }
  }
}
```

Narretz:
> Taking a more careful look, the problem seems to be that once we traverse all previousSiblings, we go to the parentNode and from there we start at the last childNode (which is essentially a nextSibling to the original comment.

That's what I was trying to explain before. But it's probably easier experienced than explained

And your solution seems to work!

Narretz:
I'Ve updated the PR with your fix, PTAL

test/ngMessages/messagesSpec.js
```diff
@@ -412,6 +412,111 @@ describe('ngMessages', function() {
...
+
+      var nodeA = element[0].querySelector('[ng-message="a"]');
+      nodeA.remove();
+
+      expect(messageChildren(element).length).toBe(0);
+    });
+  });
+
+
+  it('should show deeply nested messages correctly after a message has been removed', function() {
```
gkalpak: Is this test failing without the fix ?

test/ngMessages/messagesSpec.js
```diff
@@ -412,6 +412,111 @@ describe('ngMessages', function() {
   }));
 
 
+  it('should not crash or leak memory when the messages are transcluded, the first message is' +
```
gkalpak: Missing whitespace after `is`

test/ngMessages/messagesSpec.js
gkalpak: Makes sense :)

gkalpak:
I left a couple of minor comments (plus the errors caught by Travis - good job Travis).

LGTM otherwise :+1:

test/ngMessages/messagesSpec.js
Narretz: Actually ... since I converted the tests to remove the element via .remove() directly instead of inside of a watcher, the tests don't fail anymore without the patch. This issue is killing me ...

Narretz: That's because the `$element.on('$destroy')` is only triggered by a jqLite(element).remove

Narretz:
I fixed up the tests, so that they actually fail without the fix, and put them into their own describe. I think that's it. This issue already made my shortlist for "worst issue of 2016"

gkalpak:
LGTM

<sub>(And I also think this issue was fun :smiley:)</sub>

PR #14250, opened 2016-03-16
feat($sce): handle URLs through the $sce service, plus allow concatenation in that $sce contexts.

rjamet:

This is an improvement to  uniformize the secure context code, by making the $sce handle URL-context in roughly the same way it handles HTML-context. That also required a more silent change, to have $sce-contexts where concatenation of values is allowed (described below).

Currently, the whole $sce.URL context is unused. Dynamically-set URLs that are sanitized through various codepaths (mostly in compile.js, $set function). 

Usually, if one wants to force a value through Angular's security mechanisms, the $sce.trustAs function should be used. There's no straightforward way to do it in a single place for URLs, as `<a href="{{sce.trustAsUrl(foo)}}">` sanitizes silently its link. 

Also, an attribute being a $sce context means that concatenations are blocked: `<iframe src="generate.php?{{params}}">` would throw for instance. However, since it's not a $sce context, `` is fine.

The new behavior makes Angular handle URLs with the $sce service, similarly to HTML context. $sce.trustAsUrl is now useful, as it allows bypassing the sanitization. If one uses plain strings in URL-context attributes, they will be sanitized through the existing $$sanitizeUri service.

I also have tweaked $interpolate to work as expected with concatenated values in select secure contexts. In the listed contexts (URL only for now, though that could be expanded to anything), trustAs'd values with no concatenations are not sanitized (eg `"{{trustedVar}}"` works as you'd expect), and anything else is downgraded to string, concatenated if needed, and then passed to $sce.getTrusted that handles the sanitization as usual. For instance, `"java{{trustAsUrl('script:evil();')}}"` is sanitized as it should.

Yes: I've merged both URL sanitization methods in $$sanitizeUri (since there's a single $sce.URL context). This will break the applications that edit one of these whitelists, but I believe there's security benefits in doing so, as they will be able to use trustAsUrl in specific parts of their application instead of having to blanket-approve the special scheme they rely on. Otherwise, I believe the changes are backwards-compatible unless you're really hooking into Angular internals.

**Other information**:

I'm from Google security, but I'm not right all the time, so please ask about anything: I might have missed something, or this might not be something you want. I had to modifiy components I'm not that familiar with. Things to check:
- If you had special reasons to split the URL sanitization in two parts, I haven't been able to find it. I just merged the whitelists. 
- I'm not entirely sure of the $interpolate change. I've had several issues with infinite digests with arrays while developping, the latest changes seem to pass, but there might still be dragons hidden in there.
- I'm not sure either I got all the doc modified. 
- Finally, $sce.HTML / $sce.RESOURCE_URL could also be concatenated. At least for HTML it works, and you could have something like the commented-out test in sceSpecs.js, but I'm not sure it brings much value here, so I've left it disabled for now. RESOURCE_URL could follow the same path with either completely-trusted URLs, or concatenated-then-treated-as-string URLs that would be checked against the whitelist/blacklist mechanism.


rjamet:
I broke IE tests with "access denied" errors... I'll take a deeper look sometime this week or the following one.

test/ng/directive/booleanAttrsSpec.js
```diff
@@ -191,34 +191,32 @@ describe('ngSrc', function() {
   }));
 
 
-  if (msie) {
-    it('should update the element property as well as the attribute', inject(
-        function($compile, $rootScope, $sce) {
-          // on IE, if "ng:src" directive declaration is used and "src" attribute doesn't exist
-          // then calling element.setAttribute('src', 'foo') doesn't do anything, so we need
-          // to set the property as well to achieve the desired effect
+  it('should update the element property as well as the attribute', inject(
```
rjamet: This is an observable change : I'm really not sure why these weren't canonicalized on IE before, but now they are, and it just seems like the right behavior.

rjamet:
I finally got this to work! There's another very minor change: IE props for URLs are now canonicalized, and I assume sanitized. I'm not entirely sure of the origin of the previous behavior, nor if that matters much?

rjamet:
(And the CI looks good, except mobile Safari that crashed :/)

src/ng/compile.js
```diff
@@ -1458,11 +1429,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
         nodeName = nodeName_(this.$$element);
 
-        if ((nodeName === 'a' && (key === 'href' || key === 'xlinkHref')) ||
-            (nodeName === 'img' && key === 'src')) {
-          // sanitize a[href] and img[src] values
-          this[key] = value = $$sanitizeUri(value, key === 'src');
```
gkalpak: Afaict, this change means that people using `attrs.$set(key, value)` will **not** benefit from sanitization (for `<a href/xlink:href` or `<img src/ng-src`) any more. Is that correct?

src/ng/compile.js
```diff
@@ -2956,6 +2929,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           (tag !== "img" && (attrNormalizedName === "src" ||
                             attrNormalizedName === "ngSrc"))) {
         return $sce.RESOURCE_URL;
+      } else if (tag === "img" && (attrNormalizedName === "src" ||
```
gkalpak: If #14019 is landed, `audio`, `video` and `source` should also be added to the list, right?

src/ng/compile.js
```diff
@@ -2956,6 +2929,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           (tag !== "img" && (attrNormalizedName === "src" ||
```
gkalpak: See https://github.com/angular/angular.js/pull/14250/files#r75310645.

src/ng/interpolate.js
```diff
@@ -301,27 +307,54 @@ function $InterpolateProvider() {
...
+
       // Concatenating expressions makes it hard to reason about whether some combination of
       // concatenated values are unsafe to use and could easily lead to XSS.  By requiring that a
-      // single expression be used for iframe[src], object[src], etc., we ensure that the value
-      // that's used is assigned or constructed by some JS code somewhere that is more testable or
-      // make it obvious that you bound the value to some user controlled value.  This helps reduce
-      // the load when auditing for XSS issues.
-      if (trustedContext && concat.length > 1) {
-          $interpolateMinErr.throwNoconcat(text);
-      }
```
gkalpak: Why remove the checks at this point and defer them inside `compute()` instead?
Are we gaining anything?

src/ng/interpolate.js
```diff
@@ -301,27 +307,54 @@ function $InterpolateProvider() {
...
       if (!mustHaveExpression || expressions.length) {
         var compute = function(values) {
           for (var i = 0, ii = expressions.length; i < ii; i++) {
             if (allOrNothing && isUndefined(values[i])) return;
             concat[expressionPositions[i]] = values[i];
           }
-          return concat.join('');
+
+          if (contextAllowsConcatenation) {
+            if (singleExpression) {
```
gkalpak: There doesn't seem to be any benefit in having two distinct code paths (that essentially do the same thing).
(Keeping track of `singleExpression` seems unnecessary as well.)

src/ng/interpolate.js
```diff
@@ -338,7 +371,7 @@ function $InterpolateProvider() {
 
               return compute(values);
             } catch (err) {
-              $exceptionHandler($interpolateMinErr.interr(text, err));
+              $exceptionHandler(err);
```
gkalpak: I don't like this change either :smiley:

src/ng/compile.js
rjamet: Yes. I assumed users shouldn't use these and instead rely on attributes at the template level.

src/ng/compile.js
rjamet: Yup. Security-wise, these three won't execute script, so that doesn't matter too much, but if a user decides to have a more restricted URL sanitization for various reasons, they'd probably like to have the right contexts set in attributes. I'm thinking for instance of webmail apps: if they want to prevent third-party trackers, they could tighten $sce.getTrustedUrl to only accept untrusted urls if they're same-origin.

src/ng/interpolate.js
rjamet: Yeah, that's weird, I have no idea why I changed that. Fixed, and the tests that broke with it.

src/ng/interpolate.js
rjamet: Readability, mostly: see the other comment below.

src/ng/interpolate.js
rjamet: That's mostly for readability: I was struggling when writing this part to understand the data flows and make them clearer, that's why I preferred to split them so I could comment the paths inline. I'm also reusing `singleExpression` below, in parseStringifyInterceptor (to keep the value of "{{foo}}" from being stringified). 

That being said, I'm not that happy with it: if you have suggestions of how to write this better, I'm all ears :/

src/ng/compile.js
rjamet: Done.

rjamet:
(running into Node trouble, I'll update when I manage to check tests, sorry about that)

src/ng/interpolate.js
rjamet: I remember why now : this wraps noconcat errors in an interpolate:interr error, which seems sane, so I've kept the wrapping as it was before this PR.

rjamet:
Ok, this is ready for review, I sorted my issues and tests are green :)

koto:
Are there any remaining concerns for merging this? It seems useful.

petebacondarwin:
If this has a breaking change then we won't get it into 1.6.0 now. We can merge it into master once 1.6.0 has been released.

petebacondarwin (approved):
I have downloaded and rebased this locally - not a trivial task since this PR is getting quite old.
It looks good to me but since this is all hitting security sensitive code, I would like two more LGTM from core AngularJS members.

src/ngSanitize/sanitize.js
```diff
@@ -158,7 +158,7 @@ function $SanitizeProvider() {
     return function(html) {
       var buf = [];
       htmlParser(html, htmlSanitizeWriter(buf, function(uri, isImage) {
```
rjamet: I don't think so, it was only used to decide which sanitization method to use.

petebacondarwin: thanks. Waiting on a last review from @IgorMinar  and @gkalpak before merging. Did you @rjamet test this on Google3 already?

rjamet: I'll take a look today, but it's at least breaking for custom whitelists since the two functions got merged: it's trivial to fix though. I might find other issues though (I'll update tonight). Should that wait for a minor version bump?

gkalpak: > Should that wait for a minor version bump?

Yes, as a breaking change this will be merged only into master (which will eventually become 1.7.0; probably some time around May).

gkalpak:
> If you had special reasons to split the URL sanitization in two parts, I haven't been able to find it. I just merged the whitelists.

A little background:

1. First only `a[href]` was sanitized: 9532234
2. Then `img[src]` was sanitized based on the same whitelist: 1adf29a
3. Then `a[href]` and `img[src]` were sanitized based on different whitelists: 3e39ac7

The main reason for the separate whitelists afaict is the ability to trust different schemes / URLs. For example, by default, `a[href]` allows `mailto` and `tel` schemes, while `img[src]` allows `blob` and `data:image`. It should also allow more granular control/auditing from developers: trusting on source for imahes only or trusting another source for links only etc.

It seems useful to be able to sanitize `a[href]` and `img[src]` from different whitelists (but then again I am no security expert :smiley:)

@rjamet, what are your thoughts on this?

src/ng/sanitizeUri.js
```diff
@@ -3,67 +3,41 @@
...
-   *    chaining otherwise.
-   */
-  this.aHrefSanitizationWhitelist = function(regexp) {
-    if (isDefined(regexp)) {
-      aHrefSanitizationWhitelist = regexp;
-      return this;
-    }
-    return aHrefSanitizationWhitelist;
-  };
+  var uriSanitizationWhitelist = /^\s*((https?|ftp|file|blob|tel|mailto):|data:image\/)/i;
```
rjamet: I'm wrong there, blob shouldn't be safe for navigation. A link to a blob with user-controlled contents and a dangerous MIME type will run in the origin that created it.

rjamet:
The initial thought was to align with Closure and simplify, but it's probably not a great reason to break things... How about aligning with Angular?(https://github.com/angular/angular/blob/master/modules/%40angular/platform-browser/src/security/url_sanitizer.ts)

I think that the distinction is not necessary, the safe subset is fairly much the same. There's some nonsensical choices (img src=tel://), but they're safe, except blob that I shouldn't have included there. 

Overall, my feeling is that a configurable whitelist is only useful in very specific cases, and most apps shouldn't configure it: for instance, if your platform has unusual conventions that don't follow the rest of the world (https://github.com/angular/angular/issues/12840 comes to mind). Something like tel:// or market:// urls shouldn't be allowed app-wide, in general cases. And now, with that PR, URLs can be trustAs'd, so in most cases a trustMarketUrl-filter to allow these in specific bindings makes more sense to me.

src/ng/compile.js
rjamet: Bumping this: there's actually a lot of uses of attrs.$set that would be relevant, and probably a bunch of vulnerabilities created by this PR. But then, as far as I can tell, URLs are the only things sanitized at this level, all the rest isn't :/

src/ngSanitize/sanitize.js
rjamet: In that case I can handle and document the google3-specific issues, since that's likely to be plumbed differently internally. There's still going to be a few issues for everyone:

- if we keep this common whitelist model, all URL bindings of blob urls that were previously fine will break without $sce.trustAsUrl, and I don't think we can figure these out beforehand.
- whitelist customizations will break (renamed)
- things plugging directly into Angular's $set for URLs will lose sanitization (in line with other contexts).

petebacondarwin:
Closing in favour of #16378 , which is rebased and does not require breaking change to white lists.

PR #14259, opened 2016-03-17
fix(ngAnimate): run structural animations with cancelled out class ch…

### LINKED ISSUE #14249: Bug: ngAnimation classes skipped on enter when using ng-if on form that contains required email input
Hey!

When you have a form that is shown using `ngIf` and contain an `<input type="email" ng-model="someModel" required>`, ngAnimate classes are not added when the `ngIf` statement becomes truthy.

**Steps to reproduce**
[Link to demo](http://codepen.io/poacher2k/pen/zqNOxo)
1. Create a form that is shown with `ngIf` and contains an `<input type="email" ng-model="someModel" required>`
2. Make `ngIf` statement truthy

ngAnimate classes should be added when `ngIf` statement becomes truthy.

Consistent behavior 

**Affected:**
This occured when I tested with Chrome and Firefox on Linux, Safari on OSX, and Chrome / IE on Win7, both angular 1.5.x stable and snapshot.

**Other information**
It works as expected if:
1. `type="email"` is changed to something else
2. `required` is removed
3. `ngShow / ngHide` is used instead of `ngIf`

Narretz:
Bugfix

No

**Other information**:

…anges

When multiple animations on the same element are queued before a $digest passes,
the animator tries to create as few actual animations as possible by joining / canceling
redundant animations. Class-based animations for example are cancelled when the classes that
are added and removed are the same, and the result is no class-change. This however must only
happen if there's no structural animation currently queued.

Fixes #14249

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -1104,7 +1104,8 @@ describe("animations", function() {
         $animate.removeClass(element, 'active-class');
         $rootScope.$digest();
 
-        expect(doneHandler).toHaveBeenCalled();
+        // true = rejected
+        expect(doneHandler).toHaveBeenCalledWith(true);
       }));
 
       it('should cancel the previously running removeClass animation if a follow-up addClass animation is using the same class value',
@@ -1123,7 +1124,8 @@ describe("animations", function() {
         $animate.addClass(element, 'active-class');
         $rootScope.$digest();
 
-        expect(doneHandler).toHaveBeenCalled();
+        // true = rejected
+        expect(doneHandler).toHaveBeenCalledWith(true);
       }));
 
       it('should merge a follow-up animation that does not add classes into the previous animation (pre-digest)',
@@ -1198,6 +1200,29 @@ describe("animations", function() {
 
         expect(capturedAnimation[2].addClass).toBe('blue');
       }));
+
+      it('should NOT cancel a previously joined addClass+structural animation if a follow-up ' +
+        'removeClass animation is using the same class value (pre-digest)',
+        inject(function($animate, $rootScope) {
+
+        var runner = $animate.enter(element, parent);
+        $animate.addClass(element, 'active-class');
+
+        var doneHandler = jasmine.createSpy('enter done');
+        runner.done(doneHandler);
+
+        expect(doneHandler).not.toHaveBeenCalled();
+
+        $animate.removeClass(element, 'active-class');
+        $rootScope.$digest();
+
+        expect(capturedAnimation[1]).toBe('enter');
+        expect(capturedAnimation[2].addClass).toBe(null);
+        expect(capturedAnimation[2].removeClass).toBe(null);
+
+        expect(doneHandler).not.toHaveBeenCalled();
+      }));
+
     });
 
     describe('should merge', function() {
--- a/test/ngAnimate/integrationSpec.js
+++ b/test/ngAnimate/integrationSpec.js
@@ -756,5 +756,43 @@ describe('ngAnimate integration tests', function() {
         expect(child.attr('style')).toContain('50px');
       });
     });
+
+
+    it('should execute the enter animation on a <form> with ngIf that has an ' +
+      '<input type="email" required>', function() {
+
+      var animationSpy = jasmine.createSpy();
+
+      module(function($animateProvider) {
+        $animateProvider.register('.animate-me', function() {
+          return {
+            enter: function(element, done) {
+              animationSpy();
+              done();
+            }
+          };
+        });
+      });
+
+      inject(function($animate, $rootScope, $compile) {
+
+        element = jqLite(
+          '<div>' +
+            '<form class="animate-me" ng-if="show">' +
+              '<input ng-model="myModel" type="email" required />' +
+            '</form>' +
+          '</div>');
+
+        html(element);
+
+        $compile(element)($rootScope);
+
+        $rootScope.show = true;
+        $rootScope.$digest();
+
+        $animate.flush();
+        expect(animationSpy).toHaveBeenCalled();
+      });
+    });
   });
 });
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -82,6 +82,11 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
   });
 
   rules.cancel.push(function(element, newAnimation, currentAnimation) {
+    // cancel the animation if classes added / removed in both animation cancel each other out,
+    // but only if the current animation isn't structural
+
+    if (currentAnimation.structural) return false;
+
     var nA = newAnimation.addClass;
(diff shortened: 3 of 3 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -1198,6 +1200,28 @@ describe("animations", function() {
 
         expect(capturedAnimation[2].addClass).toBe('blue');
       }));
+
+      it('should NOT cancel a previously joined addClass+structural animation if a follow-up removeClass animation is using the same class value (pre-digest)',
```
gkalpak: <sub><sub><sub>Too long ?</sub></sub></sub>

test/ngAnimate/animateSpec.js
```diff
@@ -1198,6 +1200,28 @@ describe("animations", function() {
...
+        expect(doneHandler).not.toHaveBeenCalled();
+
+        $animate.removeClass(element, 'active-class');
+        $rootScope.$digest();
+
+        expect(capturedAnimation[1]).toBe('enter');
+        expect(capturedAnimation[2].addClass).toBe(null);
+        expect(capturedAnimation[2].removeClass).toBe(null);
+
+        expect(doneHandler).not.toHaveBeenCalledWith(true);
```
gkalpak: MAybe this could be more explicit. Should it not have been called at all or or have been called but without `true` ?

test/ngAnimate/integrationSpec.js
```diff
@@ -756,5 +756,38 @@ describe('ngAnimate integration tests', function() {
         expect(child.attr('style')).toContain('50px');
       });
     });
+
+
+    it('should execute the enter animation on a <form> with ngIf that has a <input> of type="email" that is required', function() {
```
gkalpak: <sub><sub><sub>Too long...</sub></sub></sub>

test/ngAnimate/integrationSpec.js
```diff
@@ -756,5 +756,38 @@ describe('ngAnimate integration tests', function() {
...
+              done();
+            }
+          };
+        });
+      });
+
+      inject(function($animate, $rootScope, $compile) {
+        var form;
+
+        element = jqLite('<div><form class="animate-me" ng-if="show"><input ng-model="myModel" type="email" required /></form></div>');
```
gkalpak: <sub><sub><sub>Too long...</sub></sub></sub>

test/ngAnimate/integrationSpec.js
gkalpak: Could you describe the underlying conditions.
I assume this is not specific to `form`/`ngIf`/`email`/`required`, but some combination of `enter` animation with `addClass`/`removeClass` animations (such as those done by validators). But what exactly ?

test/ngAnimate/animateSpec.js
Narretz: Too small? ;)

test/ngAnimate/animateSpec.js
Narretz: Good point. I think "notHaveBeenCalled" is intended.

test/ngAnimate/integrationSpec.js
Narretz: This test replicates the actual case from the original issue, the test in animateSpec.js is the "condensed" version. In practice it's pretty rare, because you actually need all these exact directives. The root cause lies in the way the email input sets / unsets the errors / error classes. Because for some reason, email and url inputs set ctrl.$$parserType even though they don't have a parser. That adds some unnessecary class additions / removals which trigger this bug. I actually meant to open a PR for fixing this, but forgot.

Narretz:
@gkalpak can you take another look at this?

test/ngAnimate/animateSpec.js
```diff
@@ -1201,7 +1201,8 @@ describe("animations", function() {
         expect(capturedAnimation[2].addClass).toBe('blue');
       }));
 
-      it('should NOT cancel a previously joined addClass+structural animation if a follow-up removeClass animation is using the same class value (pre-digest)',
+      it('should NOT cancel a previously joined addClass+structural animation if a follow-up' +
```
gkalpak: Missing whitespace after `follow-up`.

test/ngAnimate/animateSpec.js
```diff
@@ -1198,6 +1200,29 @@ describe("animations", function() {
 
         expect(capturedAnimation[2].addClass).toBe('blue');
       }));
+
+      it('should NOT cancel a previously joined addClass+structural animation if a follow-up' +
```
gkalpak: Missing whitespace after `follow-up`.

test/ngAnimate/integrationSpec.js
```diff
@@ -756,5 +756,45 @@ describe('ngAnimate integration tests', function() {
         expect(child.attr('style')).toContain('50px');
       });
     });
+
+
+    it('should execute the enter animation on a <form> with ngIf that has an' +
```
gkalpak: Missing whitespace after `an`.

test/ngAnimate/integrationSpec.js
```diff
@@ -756,5 +756,45 @@ describe('ngAnimate integration tests', function() {
...
+            '</form>' +
+          '</div>');
+
+        html(element);
+
+        $compile(element)($rootScope);
+
+        $rootScope.show = true;
+        $rootScope.$digest();
+        form = element.find('form');
```
gkalpak: This doesn't seem to be used.

test/ngAnimate/integrationSpec.js
```diff
@@ -756,5 +756,45 @@ describe('ngAnimate integration tests', function() {
...
+        });
+      });
+
+      inject(function($animate, $rootScope, $compile) {
+        var form;
+
+        element = jqLite(
+          '<div>' +
+            '<form class="animate-me" ng-if="show">' +
+              '<input ng-model="myModel" type="email" required />' +
```
gkalpak: This test will be ineffective if that other PR lands that will change the `$$parserName` for `input[email]`.
It would be better to have a test that explicitly reproduces the previously problematic situation.

Narretz: The test in animateSpec.js explicitly reproduces setting the classes. I always try to have one test that reproduces the actual directives etc. and one that uses the involved APIs directly. And I'm not sure about the test being ineffective, it's just very specific (and we don't know if the behavior will break at a later point in time again, maybe for other reasons)

gkalpak: Fair enough !
Since there is already a test that reproduces the issue using the low-level APIs, I'm fine with this :smiley:

gkalpak:
Other than the minor comments (whitespaces and unused var?), it LGTM :+1:

Narretz:
Landed as  cc1de81f5e2867e9ee6db9ecb2a4fd9d19fa2588 and c7813e9ebf793fe89380dcad54e8e002fafdd985

PR #14264, opened 2016-03-17
Fix NgModelController: Update example code

yilinglu:
Example code update

No

- [ x] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ x] Docs have been added / updated (for bug fixes / features)

**Other information**:

Before this change, both input field will be set to blank when escape key is pressed down. However, according to the documentation for this example, the intent is to roll back the view value on the second input field when escape key is pressed. This change enables the rollback of view value as it the example intended.

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -475,8 +475,9 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
    *           e.preventDefault();
    *           if (rollback) {
    *             $scope.myForm[value].$rollbackViewValue();
-   *           }
-   *           $scope.model[value] = '';
+   *           }else{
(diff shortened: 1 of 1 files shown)


gkalpak:
> both input field will be set to blank when escape key is pressed

I see something different. If you follow the instructions in the example, you will notice that the first input is left untouched when pressing <kbd>esc</kbd> and the model is updated with the field's value on blur. With second field, the field's value is correctly reset when pressing <kbd>esc</kbd> and everything works as intended (even after you blur the field).

**Note:** In order to see the described differences, the initial model values must be `''` (as mentioned in the example description).

--
Closing as it works as expected (but thx for taking the time to submit a PR :+1:)

Issue #14268, opened 2016-03-18
Controller functions stringified twice when directives are compiled, even in strict DI mode


robertknight:
Whilst investigating poor performance when instantiating a component, I was surprised to see `Function.prototype.toString()` showing up in the profile taking a significant amount of time (about 2ms per instance of a component in a list of potentially hundreds of items) despite the app using strict DI.

It appears that when instantiating a component, Angular calls `Function.prototype.toString` on every controller twice. The first time is in `setupControllers` due to what looks like a typo in `controller == '@'` and the second is to determine whether a function is an ES2015 class (!!), which seems silly considering that the proportion of users using native classes in production apps will be tiny.

In testing, it appears that stringifying a large function is somewhat more expensive in Firefox than Chrome/Node/IE.

**Steps to reproduce:**

1. Open this Plunker: http://plnkr.co/edit/RbdcKbymucZWvOF4V6Co
2. Open dev tools and refresh, observe 2 traces from a monkey-patched `Function.prototype.toString` invoked on `AppController`

Narretz:
> and the second is to determine whether a function is an ES2015 class (!!), which seems silly considering that the proportion of users using native classes in production apps will be tiny.

We want to support this, why do you consider this silly? This is going to be used more and more.

> The first time is in setupControllers due to what looks like a typo in controller == '@' 

This is to support an old API that `ngController` uses. I think it's undocumented. Maybe we don't need it anymore.

Narretz:
Or use `controller === '@'` ....

Narretz:
It's also worth a shot to raise the issue of toString being slow in Firefox itself.

robertknight:
> We want to support this, why do you consider this silly? This is going to be used more and more.

Apologies, perhaps I should rephrase, I think it is very unfortunate to be paying a performance penalty for something that very few users are using.

> It's also worth a shot to raise the issue of toString being slow in Firefox itself.

I can do that, though I don't know whether it is reasonable to expect that stringifying a function will always be very fast. It looks like Firefox might be disposing of the original source in some contexts, possibly as an optimization? - https://dxr.mozilla.org/mozilla-central/source/js/src/jsfun.cpp#936

Narretz:
Can you test if you see an actual improvement when you change `==` to `===` here https://github.com/angular/angular.js/blob/82a4545e772400127cc909ff44f95809e8790bae/src/ng/compile.js#L2488?

[...]

PR #14271, opened 2016-03-18
perf($compile): use strict comparison for controller === '@'

Narretz:
perf improvement

No

**Other information**:

In the DDO, controller can be '@', which means the controller name
is taken from the directive attribute. This is undocumented and internally
only used by ngController. There seems to be no case where converting the
controller function to a string would actually be necessary.

Related #14268

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2484,7 +2484,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         };
 
         var controller = directive.controller;
-        if (controller == '@') {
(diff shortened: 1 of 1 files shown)


dcherman:
Kinda sorta related, but why isn't strict equality being enforced by a linter?

gkalpak:
@Narretz: LGTM

@dcherman: Good question :)

Narretz:
LAnded as https://github.com/angular/angular.js/commit/bbd3db14f857aab996ad129f2f15ca6348e9fd9f

PR #14275, opened 2016-03-18
test(helpers): fix error message generation for `toHaveBeenCalledOnce[With]` matchers

gkalpak:
This commit fixes the error messages for 2 of our internal matchers (`toHaveBeenCalledOnce()` and `toHaveBeenCalledOnceWith()`). The previous format (returning an array containing both the "normal" and the negative error messages), is not supported in Jasmine 2.4. It will always concatenate the messages.

Related to #14226.

<sub>There is also a 2nd commit (to be squashed when merging) with minor fixes (typos and testcase descriptions) and a couple of TODOs.</sub>

### DIFF
--- a/test/helpers/matchers.js
+++ b/test/helpers/matchers.js
@@ -131,25 +131,38 @@ beforeEach(function() {
       return {
         compare: function(actual) {
           if (arguments.length > 1) {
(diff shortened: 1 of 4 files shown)


Narretz:
LGTM. I wonder though if we can test our custom matchers? Testception?

PR #14282, opened 2016-03-20
refactor(): delete a usless conditional sentence

CarlosMiguelCuevas:

refactor

the same

no

**Other information**:

delete a conditional sentences. it is marked with a TODO tag to be deleted.

### DIFF
--- a/src/ngAnimate/animation.js
+++ b/src/ngAnimate/animation.js
@@ -348,8 +348,6 @@ var $$AnimationProvider = ['$animateProvider', function($animateProvider) {
         // may attempt more elements, but custom drivers are more particular
         for (var i = drivers.length - 1; i >= 0; i--) {
           var driverName = drivers[i];
-          if (!$injector.has(driverName)) continue; // TODO(matsko): remove this check
-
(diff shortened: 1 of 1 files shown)


src/ngAnimate/animation.js
```diff
@@ -348,8 +348,6 @@ var $$AnimationProvider = ['$animateProvider', function($animateProvider) {
         // may attempt more elements, but custom drivers are more particular
         for (var i = drivers.length - 1; i >= 0; i--) {
           var driverName = drivers[i];
-          if (!$injector.has(driverName)) continue; // TODO(matsko): remove this check
```
gkalpak: @matsko, any idea why you had marked this for removal ?

petebacondarwin: Pretty sure this was only here for debugging. If the `driverName` refers to a driver that has not be registered with the `injector` then that is surely an error.

We register only two drivers in the `ngAnimate` module:

* `$$animateCssDriver`
* `$$animateJsDriver`

We do this from inside their respective providers, so it is not possible for our code to break this.

PR #14286, opened 2016-03-21
fix($compile): workaround a GC bug in Chrome < 50

### LINKED ISSUE #14041: ng-repeat-start error with angular 1.5.0
Hello
I have an empty array that will use ng-repeat-start to render HTML.But I got some error when click button to add item to array. As below photo

angular-1 5 0-ng-repeat-start-error
ps: 

I've tried kind of various to find out.Finally I found below code will cause this error.

> app.directive("ngRepeat", ['$animate', function ($animate) {
                return function ($scope, $element) {
                   $animate.enabled($element, false);
                };
            }]);

At beginning I thought it caused by  $animate.enabled. after I empty the function as below code that still error.
> app.directive("ngRepeat", [ function () {
                return function ($scope, $element) {
                };
            }]);

Ps:1.This problem just occurs when I upgrade to 1.5.0 
     2.I use chrome 64 bit.This error will not occurs on IE 11 or IE Edge

simple example
[plnkr](http://plnkr.co/edit/eIoapMVC9LQAAO2RCPTj?p=preview)

Please help the Issue,thank you very much.

dcherman:
In the version of V8 used in Chrome < 50, the parent of template nodes for
`transclude: "element"` directives would be improperly garbage collected
despite still having been referenced via `parentNode`.

No tests are included with this change since this is not a deterministic bug and would require triggering a GC sweep which would involve running the test suite with v8 natives enabled.

Fixes #14041

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2067,6 +2067,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             compileNode = $compileNode[0];
             replaceWith(jqCollection, sliceArgs($template), compileNode);
 
+            // Support: Chrome < 50
+            // https://github.com/angular/angular.js/issues/14041
+
+            // In the versions of V8 prior to Chrome 50, the document fragment that is created
+            // in the `replaceWith` function is improperly garbage collected despite still
+            // being referenced by the `parentNode` property of all of the child nodes.  By adding
+            // a reference to the fragment via a different property, we can avoid that incorrect
+            // behavior.
+            $template[0].$$parentNode = $template[0].parentNode;
+
             childTranscludeFn = compilationGenerator(mightHaveMultipleTransclusionError, $template, transcludeFn, terminalPriority,
                                         replaceDirective && replaceDirective.name, {
                                           // Don't pass in:


dcherman:
For reasons mentioned in the original message, no tests were included, however I cannot reproduce the bug anymore with this patch.

gkalpak:
Crazy how you tracked this down :smiley:

lgalfaso:
Is there a bug opened in Chrome that tracks this? If there is one, can this information be added as a comment in the code.

dcherman:
@lgalfaso So before making this PR I tried to find any issue or commit on the Chromium, Blink, or v8 bug trackers/repos that I could reference in a comment or the commit message, but I couldn't find anything.  I think chances are high that it's a commit in v8 that fixed the actual bug since it's GC related, but no message that has the word `GC`, `Garbage`, or `DOM` stands out as being related to this.

Easiest way to figure it out would end up being to git-bisect v8 and re-run the test case to find the commit where it gets fixed, but I don't have a Chromium/v8 build environment and it may not be worth the effort.

lgalfaso:
Hey, what I was looking for is more basic:
- In the comment, it claims that this bug only affects Chrome < 50. Is this
fixed in Chrome 50?
- If yes, done!
- If no, open a bug in Chrome with this example and add a reference in the
code to the bug. If someone looks at the code, they will know why it is
there and when it is safe to delete it

gkalpak:
@lgalfaso, there is "empirical evidence" that it's fixed in Chrome 50 (but no hard proof :stuck_out_tongue:).

lgalfaso:
@gkalpak empirical evidence is good enough :)

dcherman:
Yep, I tested the original un-patched reproducer in both Chrome 50 and 51 (current canary).  I cannot reproduce in either version, so somewhere between v8 versions 4.9.385.28 and 5.0.71, there was a commit that fixed this problem.

Narretz:
Thanks @dcherman!

gkalpak:
@Narretz, shouldn't this also be backported to `v1.5.x` (so it's included in the upcoming release) ?

Narretz:
Absolutely yes. Will do
Am 23.03.2016 21:26 schrieb "Georgios Kalpakas" <notifications@github.com>:

PR #14293, opened 2016-03-21
perf($parse): Inline constants

lgalfaso:
Performance

Constants are not inlined in the function generated by `$parse`

Constants are inlined in the function generated by `$parse`

No

**Other information**:

Inline constants definitions in function calls, array definitions and object values.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -924,7 +924,7 @@ ASTCompiler.prototype = {
     case AST.Literal:
       expression = this.escape(ast.value);
       this.assign(intoId, expression);
-      recursionFn(expression);
+      recursionFn(intoId || expression);
       break;
     case AST.UnaryExpression:
       this.recurse(ast.argument, undefined, undefined, function(expr) { right = expr; });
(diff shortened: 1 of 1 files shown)


jbedard:
Have an example of how this changes the generated function?

lgalfaso:
For the expression `[1, {foo: "bar"}, 1 + 2]` it changes it from
```
// After some reordering and cleanup
var v1 = 1;
var v2 = "bar";
var v3 = {foo: v2};
var v4 = 1;
var v5 = 2;
var v6 = plus(v4, v5);
var v7 = [v1, v3, v6];
return v7;
```

to
```
return [1, {foo: "bar"}, plus(1, 2)];
```

Expression parts that are not constants, did not change at all.

jbedard:
Nice! :smile:

Narretz:
@lgalfaso Can you put this exampe into the commit message? Makes it easier to grok it later

lgalfaso:
Fixed the commit message and landed as bd7d5f6345439aa2d1da708ffee20b4c565131d4

PR #14302, opened 2016-03-22
feat($compile): add more lifecycle hooks to directive controllers

### LINKED ISSUE #13991: FormController: attach to component controller?

### LINKED ISSUE #14020: 1.5 Feature: Backport ngOnDestroy (as $onDestroy for example) exactly as ngOnInit ($onInit)
In new 1.5.0 release is possible to use [$onInit](https://toddmotto.com/on-init-require-object-syntax-angular-component) hook which is similar to `ngOnInit ` from [Angular 2 lifecycle hooks](https://angular.io/docs/ts/latest/guide/lifecycle-hooks.html). It will help with keeping controller's  constructors clean, with initialization logic moved to the `$onInit` method.
I think worth consideration is backport also `ngOnDestroy`, as `$onDestroy` for example.
### Use Case:
In recent project in my company on of the business rules is to send request to device every second (which we have done with timeouts). We [try to stay away as possible from injecting `$scope` to controllers](https://github.com/johnpapa/angular-styleguide#style-y031), but timeouts needs to be cleared when user opens new view or hits back button on mobile device etc... So we solved this with `$scope.$on`:

```javascript
$scope.$on('$destroy', function () {
      viewModel.removeIntervals();
});
```
And it's only one direct usage of `$scope` at this moment. If we have `$onDestroy` hook, code would be like:

```javascript
viewModel.$onDestroy =  function () {
      viewModel.removeIntervals();
});
```

And we can get rid off `$scope` injection completely. Also, it let us keep code closer to Angular 2 syntax.

Thank you in advance for looking on this post.

Best regards,
Radek.

### LINKED ISSUE #14030: Implement lifecycle hooks for directive/component controllers (ng2-style)

petebacondarwin:

Feature

Missing lifecycle hooks for directive controllers

Add these lifecycle hooks

Nope

**Other information**:

This change adds in the following new lifecycle hooks, which map in some
way to those in Angular 2:
- `$onChanges(changesObj)` - Called whenever one-way bindings are updated. The `changesObj` is a hash whose keys
  are the names of the bound properties that have changed, and the values are an object of the form
  `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
  cloning the bound value to prevent accidental mutation of the outer value.
- `$onDestroy` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
  external resources, watches and event handlers.
- `$afterViewInit` - Called after this controller's element and its children been linked. Similar to the post-link
  function this hook can be used to set up DOM event handlers and do direct DOM manipulation.
  Note that child elements that contain `templateUrl` directives will not have been compiled and linked since
  they are waiting for their template to load asynchronously and their own compilation and linking has been
  suspended until that occurs.

Closes #14127
Closes #14030
Closes #14020
Closes #13991

### DIFF
--- a/docs/content/error/$compile/infchng.ngdoc
+++ b/docs/content/error/$compile/infchng.ngdoc
@@ -1,0 +1,30 @@
+@ngdoc error
+@name $compile:infchng
+@fullName Unstable `$onChanges` hooks
+@description
+
+This error occurs when the application's model becomes unstable because some `$onChanges` hooks are causing updates which then trigger
+further calls to `$onChanges` that can never complete.
+Angular detects this situation and prevents an infinite loop from causing the browser to become unresponsive.
+
+For example, the situation can occur by setting up a `$onChanges()` hook which triggers an event on the component, which subsequently
+triggers the component's bound inputs to be updated:
+
+```html
+<c1 prop="a" on-change="a = -a"></c1>
+```
+
+```js
+function Controller1() {}
+Controller1.$onChanges = function() {
+  this.onChange();
+};
+
+mod.component('c1', {
+  controller: Controller1,
+  bindings: {'prop': '<', onChange: '&'}
+}
+```
+
+The maximum number of allowed iterations of the `$onChanges` hooks is controlled via TTL setting which can be configured via
+{@link ng.$compileProvider#onChangesTtl `$compileProvider.onChangesTtl`}.
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -147,6 +147,30 @@ components should follow a few simple conventions:
       }
     ```
 
+- **Components have a well-defined lifecycle**
+Each component can implement "lifecycle hooks". These are methods that will be called at certain points in the life
+of the component. The following hook methods can be implemented:
+
+  * `$onInit()` - Called on each controller after all the controllers on an element have been constructed and
+    had their bindings initialized (and before the pre &amp; post linking functions for the directives on
+    this element). This is a good place to put initialization code for your controller.
+  * `$onChanges(changesObj)` - Called whenever one-way bindings are updated. The `changesObj` is a hash whose keys
+    are the names of the bound properties that have changed, and the values are an object of the form
+    `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
+    cloning the bound value to prevent accidental mutation of the outer value.
+  * `$onDestroy()` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
+    external resources, watches and event handlers.
+  * `$postLink()` - Called after this controller's element and its children have been linked. Similar to the post-link
+    function this hook can be used to set up DOM event handlers and do direct DOM manipulation.
+    Note that child elements that contain `templateUrl` directives will not have been compiled and linked since
+    they are waiting for their template to load asynchronously and their own compilation and linking has been
+    suspended until that occurs.
+    This hook can be considered analogous to the `ngAfterViewInit` and `ngAfterContentInit` hooks in Angular 2.
+    Since the compilation process is rather different in Angular 1 there is no direct mapping and care should
+    be taken when upgrading.
+
+By implementing these methods, you component can take part in its lifecycle.
+
 - **An application is a tree of components:**
 Ideally, the whole application should be a tree of components that implement clearly defined inputs
 and outputs, and minimize two-way data binding. That way, it's easier to predict when data changes and what the state
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -293,9 +293,23 @@
  *    `true` if the specified slot contains content (i.e. one or more DOM nodes).
  *
  * The controller can provide the following methods that act as life-cycle hooks:
- * * `$onInit` - Called on each controller after all the controllers on an element have been constructed and
+ * * `$onInit()` - Called on each controller after all the controllers on an element have been constructed and
  *   had their bindings initialized (and before the pre &amp; post linking functions for the directives on
  *   this element). This is a good place to put initialization code for your controller.
+ * * `$onChanges(changesObj)` - Called whenever one-way (`<`) or interpolation (`@`) bindings are updated. The
+ *   `changesObj` is a hash whose keys are the names of the bound properties that have changed, and the values are an
+ *   object of the form `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component
+ *   such as cloning the bound value to prevent accidental mutation of the outer value.
+ * * `$onDestroy()` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
+ *   external resources, watches and event handlers. Note that components have their `$onDestroy()` hooks called in
+ *   the same order as the `$scope.$broadcast` events are triggered, which is top down. This means that parent
+ *   components will have their `$onDestroy()` hook called before child components.
+ * * `$postLink()` - Called after this controller's element and its children have been linked. Similar to the post-link
+ *   function this hook can be used to set up DOM event handlers and do direct DOM manipulation.
+ *   Note that child elements that contain `templateUrl` directives will not have been compiled and linked since
+ *   they are waiting for their template to load asynchronously and their own compilation and linking has been
+ *   suspended until that occurs.
+ *
  *
  * #### `require`
  * Require another directive and inject its controller as the fourth argument to the linking function. The
@@ -1207,6 +1221,36 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     return debugInfoEnabled;
   };
 
+
+  var TTL = 10;
+  /**
+   * @ngdoc method
+   * @name $compileProvider#onChangesTtl
+   * @description
+   *
+   * Sets the number of times `$onChanges` hooks can trigger new changes before giving up and
+   * assuming that the model is unstable.
+   *
+   * The current default is 10 iterations.
+   *
+   * In complex applications it's possible that dependencies between `$onChanges` hooks and bindings will result
+   * in several iterations of calls to these hooks. However if an application needs more than the default 10
+   * iterations to stabilize then you should investigate what is causing the model to continuously change during
+   * the `$onChanges` hook execution.
+   *
+   * Increasing the TTL could have performance implications, so you should not change it without proper justification.
+   *
+   * @param {number} limit The number of `$onChanges` hook iterations.
+   * @returns {number|object} the current limit (or `this` if called as a setter for chaining)
+   */
+  this.onChangesTtl = function(value) {
+    if (arguments.length) {
+      TTL = value;
+      return this;
+    }
+    return TTL;
+  };
+
   this.$get = [
             '$injector', '$interpolate', '$exceptionHandler', '$templateRequest', '$parse',
             '$controller', '$rootScope', '$sce', '$animate', '$$sanitizeUri',
@@ -1215,6 +1259,36 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
     var SIMPLE_ATTR_NAME = /^\w/;
     var specialAttrHolder = document.createElement('div');
+
+
+
+    var onChangesTtl = TTL;
+    // The onChanges hooks should all be run together in a single digest
+    // When changes occur, the call to trigger their hooks will be added to this queue
+    var onChangesQueue;
+
+    // This function is called in a $$postDigest to trigger all the onChanges hooks in a single digest
+    function flushOnChangesQueue() {
+      try {
+        if (!(--onChangesTtl)) {
+          // We have hit the TTL limit so reset everything
+          onChangesQueue = undefined;
+          throw $compileMinErr('infchng', '{0} $onChanges() iterations reached. Aborting!\n', TTL);
+        }
+        // We must run this hook in an apply since the $$postDigest runs outside apply
+        $rootScope.$apply(function() {
+          for (var i = 0, ii = onChangesQueue.length; i < ii; ++i) {
+            onChangesQueue[i]();
+          }
+          // Reset the queue to trigger a new schedule next time there is a change
+          onChangesQueue = undefined;
+        });
+      } finally {
+        onChangesTtl++;
+      }
+    }
+
+
     function Attributes(element, attributesToCopy) {
       if (attributesToCopy) {
         var keys = Object.keys(attributesToCopy);
@@ -2360,10 +2434,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
         });
 
-        // Trigger the `$onInit` method on all controllers that have one
+        // Handle the init and destroy lifecycle hooks on all controllers that have them
         forEach(elementControllers, function(controller) {
-          if (isFunction(controller.instance.$onInit)) {
-            controller.instance.$onInit();
+          var controllerInstance = controller.instance;
+          if (isFunction(controllerInstance.$onInit)) {
+            controllerInstance.$onInit();
+          }
+          if (isFunction(controllerInstance.$onDestroy)) {
+            controllerScope.$on('$destroy', function callOnDestroyHook() {
+              controllerInstance.$onDestroy();
+            });
           }
         });
 
@@ -2399,6 +2479,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               transcludeFn
           );
         }
+
+        // Trigger $postLink lifecycle hooks
+        forEach(elementControllers, function(controller) {
+          var controllerInstance = controller.instance;
+          if (isFunction(controllerInstance.$postLink)) {
+            controllerInstance.$postLink();
+          }
+        });
 
         // This is the function that is injected as `$transclude`.
         // Note: all arguments are optional!
@@ -2995,6 +3083,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     // only occurs for isolate scopes and new scopes with controllerAs.
     function initializeDirectiveBindings(scope, attrs, destination, bindings, directive) {
       var removeWatchCollection = [];
+      var changes;
       forEach(bindings, function initializeBinding(definition, scopeName) {
         var attrName = definition.attrName,
         optional = definition.optional,
@@ -3010,6 +3099,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             }
             attrs.$observe(attrName, function(value) {
               if (isString(value)) {
+                var oldValue = destination[scopeName];
+                recordChanges(scopeName, value, oldValue);
                 destination[scopeName] = value;
               }
             });
@@ -3081,6 +3172,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             destination[scopeName] = parentGet(scope);
 
             removeWatch = scope.$watch(parentGet, function parentValueWatchAction(newParentValue) {
+              var oldValue = destination[scopeName];
+              recordChanges(scopeName, newParentValue, oldValue);
               destination[scopeName] = newParentValue;
             }, parentGet.literal);
 
@@ -3100,6 +3193,33 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             break;
         }
       });
+
+      function recordChanges(key, currentValue, previousValue) {
+        if (isFunction(destination.$onChanges) && currentValue !== previousValue) {
+          // If we have not already scheduled the top level onChangesQueue handler then do so now
+          if (!onChangesQueue) {
+            scope.$$postDigest(flushOnChangesQueue);
+            onChangesQueue = [];
+          }
+          // If we have not already queued a trigger of onChanges for this controller then do so now
+          if (!changes) {
+            changes = {};
+            onChangesQueue.push(triggerOnChangesHook);
+          }
+          // If the has been a change on this property already then we need to reuse the previous value
+          if (changes[key]) {
+            previousValue = changes[key].previousValue;
+          }
+          // Store this change
+          changes[key] = {previousValue: previousValue, currentValue: currentValue};
+        }
+      }
+
+      function triggerOnChangesHook() {
+        destination.$onChanges(changes);
+        // Now clear the changes so that we schedule onChanges when more changes arrive
+        changes = undefined;
+      }
 
       return removeWatchCollection.length && function removeWatches() {
         for (var i = 0, ii = removeWatchCollection.length; i < ii; ++i) {
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -3515,6 +3515,391 @@ describe('$compile', function() {
     });
   });
 
+  describe('controller lifecycle hooks', function() {
+
+    describe('$onInit', function() {
+
+      it('should call `$onInit`, if provided, after all the controllers on the element have been initialized', function() {
+
+        function check() {
+          /*jshint validthis:true */
+          expect(this.element.controller('d1').id).toEqual(1);
+          expect(this.element.controller('d2').id).toEqual(2);
+        }
+
+        function Controller1($element) { this.id = 1; this.element = $element; }
+        Controller1.prototype.$onInit = jasmine.createSpy('$onInit').and.callFake(check);
+
+        function Controller2($element) { this.id = 2; this.element = $element; }
+        Controller2.prototype.$onInit = jasmine.createSpy('$onInit').and.callFake(check);
+
+        angular.module('my', [])
+          .directive('d1', valueFn({ controller: Controller1 }))
+          .directive('d2', valueFn({ controller: Controller2 }));
+
+        module('my');
+        inject(function($compile, $rootScope) {
+          element = $compile('<div d1 d2></div>')($rootScope);
+          expect(Controller1.prototype.$onInit).toHaveBeenCalledOnce();
+          expect(Controller2.prototype.$onInit).toHaveBeenCalledOnce();
+        });
+      });
+    });
+
+
+    describe('$onDestroy', function() {
+
+      it('should call `$onDestroy`, if provided, on the controller when its scope is destroyed', function() {
+
+        function TestController() { this.count = 0; }
+        TestController.prototype.$onDestroy = function() { this.count++; };
+
+        angular.module('my', [])
+          .directive('d1', valueFn({ scope: true, controller: TestController }))
+          .directive('d2', valueFn({ scope: {}, controller: TestController }))
+          .directive('d3', valueFn({ controller: TestController }));
+
+        module('my');
+        inject(function($compile, $rootScope) {
+
+          element = $compile('<div><d1 ng-if="show[0]"></d1><d2 ng-if="show[1]"></d2><div ng-if="show[2]"><d3></d3></div></div>')($rootScope);
+
+          $rootScope.$apply('show = [true, true, true]');
+          var d1Controller = element.find('d1').controller('d1');
+          var d2Controller = element.find('d2').controller('d2');
+          var d3Controller = element.find('d3').controller('d3');
+
+          expect([d1Controller.count, d2Controller.count, d3Controller.count]).toEqual([0,0,0]);
+          $rootScope.$apply('show = [false, true, true]');
+          expect([d1Controller.count, d2Controller.count, d3Controller.count]).toEqual([1,0,0]);
+          $rootScope.$apply('show = [false, false, true]');
+          expect([d1Controller.count, d2Controller.count, d3Controller.count]).toEqual([1,1,0]);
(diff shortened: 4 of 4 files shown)


docs/content/guide/component.ngdoc
```diff
@@ -147,6 +147,27 @@ components should follow a few simple conventions:
...
+    are the names of the bound properties that have changed, and the values are an object of the form
+    `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
+    cloning the bound value to prevent accidental mutation of the outer value.
+  * `$onDestroy` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
+    external resources, watches and event handlers.
+  * `$afterViewInit` - Called after this controller's element and its children been linked. Similar to the post-link
+    function this hook can be used to set up DOM event handlers and do direct DOM manipulation.
+    Note that child elements that contain `templateUrl` directives will not have been compiled and linked since
+    they are waiting for their template to load asynchronously and their own compilation and linking has been
+    suspended until that occurs.
```
mgol: Is it also how it happens now if you use regular postLink functions in `module.directive`-defined directives?

docs/content/guide/component.ngdoc
```diff
@@ -147,6 +147,27 @@ components should follow a few simple conventions:
...
     ```
 
+- **Components have a well-defined lifecycle
+Each component can implement "lifecycle hooks", which are methods that will be called at certain points in the life
+of the component. The following hook methods can be implemented:
+
+  * `$onInit()` - Called on each controller after all the controllers on an element have been constructed and
+    had their bindings initialized (and before the pre &amp; post linking functions for the directives on
+    this element). This is a good place to put initialization code for your controller.
+  * `$onChanges(changesObj)` - Called whenever one-way bindings are updated. The `changesObj` is a hash whose keys
```
mgol: I love it (as well as `$onDestroy`); more reasons to stop injecting `$scope` into controllers.

src/ng/compile.js
```diff
@@ -3090,6 +3119,27 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         }
       });
 
+      function recordChanges(key, currentValue, previousValue) {
+        if (isFunction(destination.$onChanges)) {
+          // If we have not already schedule the onChanges hook then do so now
```
mgol: `s/schedule/scheduled/`

mgol:
I'd add tests that ensure `$onChanges` is **not** called on changes in other kinds of bindings.

docs/content/guide/component.ngdoc
petebacondarwin: Yes

petebacondarwin:
Fixed the comment typo and added tests @mgol PTAL

docs/content/guide/component.ngdoc
```diff
@@ -147,6 +147,27 @@ components should follow a few simple conventions:
...
+    are the names of the bound properties that have changed, and the values are an object of the form
+    `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
+    cloning the bound value to prevent accidental mutation of the outer value.
+  * `$onDestroy` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
+    external resources, watches and event handlers.
+  * `$afterViewInit` - Called after this controller's element and its children been linked. Similar to the post-link
```
gkalpak: _been linked_  -->  _have been linked_

mgol:
LGTM (once the typo @gkalpak noticed is fixed)

docs/content/guide/component.ngdoc
gkalpak: TBH, I feel that this last limitation makes this hook pretty usel...ehm...much less useful than the corresponding ng2 hook :(

--
I assume, with the current implementation, this hook is very similar to the post-link function, which means that:

(a) It doesn't offer new functionality for ng1 users.
(b) It doesn't help much with migrating but for very simple components with inline templates. Even caching the templates won't help, right ?

These are just assumptions atm (I haven't looked at the code yet), but if they are true, I'm a little skeptical about this hook.

src/ng/compile.js
```diff
@@ -293,9 +293,21 @@
  *    `true` if the specified slot contains content (i.e. one or more DOM nodes).
  *
  * The controller can provide the following methods that act as life-cycle hooks:
- * * `$onInit` - Called on each controller after all the controllers on an element have been constructed and
+ * * `$onInit()` - Called on each controller after all the controllers on an element have been constructed and
  *   had their bindings initialized (and before the pre &amp; post linking functions for the directives on
  *   this element). This is a good place to put initialization code for your controller.
+ * * `$onChanges(changesObj)` - Called whenever one-way bindings are updated. The `changesObj` is a hash whose keys
```
gkalpak: Nitpick: I would put the `<` symbol there somewhere, because `@` bindings are also called "one-way" in many blogsposts/tutorials and this might confuse users.

src/ng/compile.js
```diff
@@ -293,9 +293,21 @@
...
+ *   are the names of the bound properties that have changed, and the values are an object of the form
+ *   `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
+ *   cloning the bound value to prevent accidental mutation of the outer value.
+ * * `$onDestroy` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
+ *   external resources, watches and event handlers.
+ * * `$afterViewInit` - Called after this controller's element and its children been linked. Similar to the post-link
```
gkalpak: Again: _**have** been linked_

src/ng/compile.js
petebacondarwin: OK

docs/content/guide/component.ngdoc
mgol: In my current project I avoid these problems by not using `templateUrl` but relying on Webpack being able to import HTML files using the ES6 import syntax. Unfortunately you can't do things like that without any preprocessing done so it can't be a general solution.

petebacondarwin: You are right @gkalpak, which is why I was suggesting that we call it `$afterLink` instead.
It does, though, have value for components, where there is no way right now to hook into the post-link function.

src/ng/compile.js
```diff
@@ -3101,6 +3130,27 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+            scope.$$postDigest(triggerOnChangesHook);
+          }
+          // Store this change
+          changes[key] = {previousValue: previousValue, currentValue: currentValue};
+        }
+      }
+
+      function triggerOnChangesHook() {
+        // We must run this hook in an apply since the $$postDigest runs outside apply
+        scope.$apply(function() {
```
gkalpak: Doesn't this mean there will be one extra digest for each controller (that has at least one changed binding) per digest ?

docs/content/guide/component.ngdoc
petebacondarwin: I am becoming very keen on the webpack `template: require('some.template.html)` approach :-)

src/ng/compile.js
petebacondarwin: Yes you are right.
The alternative is to use `$applyAsync` but this has the downside of delaying the call to `$onChanges` by maybe 100ms.

What is better?

docs/content/guide/component.ngdoc
gkalpak: Yeah, I only wish TypeScript had something similar :(
But in the meantime, I believe most developers follow the "best practice" of caching their templates with `$templateCache` in a build step. Besides a component doesn't know what it's children (or its grandchildren) do wrt template vs templateUrl.

I am afraid this will confuse people and create some headaches while migrating.

--
If we can't/won't make it work similarly to ng2, it might be btter to name is `$postLink` and just provide it as a way to access the post-link function while using the `.component()` helper, as @petebacondarwin suggested.

src/ng/compile.js
gkalpak: I think we could move the login that calls `triggerOnChangeHook` inside `$compile` (I mean make it a compiler-local thing) and store the changes for **each** controller there, so we only trigger one extra `$digest`.

test/ng/compileSpec.js
```diff
@@ -3515,6 +3515,202 @@ describe('$compile', function() {
...
+        inject(function($compile, $rootScope) {
+
+          element = $compile('<div><d1 ng-if="show[0]"></d1><d2 ng-if="show[1]"></d2><div ng-if="show[2]"><d3></d3></div></div>')($rootScope);
+
+          $rootScope.$apply('show = [true, true, true]');
+          var d1Controller = element.find('d1').controller('d1');
+          var d2Controller = element.find('d2').controller('d2');
+          var d3Controller = element.find('d3').controller('d3');
+
+          expect([d1Controller.count, d2Controller.count, d3Controller.count]).toEqual([0,0,0]);
```
gkalpak: `[0,0,0]` ? What were you doing JSCS ? How did you let this happen ? :disappointed:

gkalpak:
I am also wondering if applying `$onChange` to `@` bindings would be closer to ng2 (iirc it will).

test/ng/compileSpec.js
petebacondarwin: That style of array appears throughout this file

gkalpak: That's why I'm so disappointed with JSCS :stuck_out_tongue:

docs/content/guide/component.ngdoc
```diff
@@ -147,6 +147,30 @@ components should follow a few simple conventions:
       }
     ```
 
+- **Components have a well-defined lifecycle
```
gkalpak: Missing closing `**`.

docs/content/guide/component.ngdoc
```diff
@@ -147,6 +147,30 @@ components should follow a few simple conventions:
...
+  * `$onInit()` - Called on each controller after all the controllers on an element have been constructed and
+    had their bindings initialized (and before the pre &amp; post linking functions for the directives on
+    this element). This is a good place to put initialization code for your controller.
+  * `$onChanges(changesObj)` - Called whenever one-way bindings are updated. The `changesObj` is a hash whose keys
+    are the names of the bound properties that have changed, and the values are an object of the form
+    `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
+    cloning the bound value to prevent accidental mutation of the outer value.
+  * `$onDestroy` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
+    external resources, watches and event handlers.
+  * `$postLink` - Called after this controller's element and its children been linked. Similar to the post-link
```
gkalpak: _been_  -->  _have been_

src/ng/compile.js
```diff
@@ -293,9 +293,21 @@
...
  * The controller can provide the following methods that act as life-cycle hooks:
- * * `$onInit` - Called on each controller after all the controllers on an element have been constructed and
+ * * `$onInit()` - Called on each controller after all the controllers on an element have been constructed and
  *   had their bindings initialized (and before the pre &amp; post linking functions for the directives on
  *   this element). This is a good place to put initialization code for your controller.
+ * * `$onChanges(changesObj)` - Called whenever one-way (`<`) bindings are updated. The `changesObj` is a hash
+ *   whose keys are the names of the bound properties that have changed, and the values are an object of the form
+ *   `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
+ *   cloning the bound value to prevent accidental mutation of the outer value.
+ * * `$onDestroy` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
```
gkalpak: There is no `()`  after `$onDestroy` (and `$postLink`) for consistency.

src/ng/compile.js
```diff
@@ -474,7 +486,7 @@
  *
  *   * `iElement` - instance element - The element where the directive is to be used. It is safe to
  *     manipulate the children of the element only in `postLink` function since the children have
- *     already been linked.
+ *     already have been linked.
```
gkalpak: Wrong _been_. We don't need _have_ here.

src/ng/compile.js
```diff
@@ -1215,6 +1227,24 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
     var specialAttrHolder = document.createElement('div');
+
+    // The onChanges hooks should all be run together in a single digest
+    // When changes occur, the call to trigger their hooks will be added to this queue
+    var onChangesQueue;
+
+    // This function is called in a $$postDigest to trigger all the onChanges hooks in a single digest
+    function flushOnChangesQueue() {
+      // We must run this hook in an apply since the $$postDigest runs outside apply
+      $rootScope.$apply(function() {
```
gkalpak: I would only run a digest if there are actually entries in `onChangesQueue`.

src/ng/compile.js
```diff
@@ -1215,6 +1227,24 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+        for (var i = 0, ii = onChangesQueue.length; i < ii; ++i) {
```
gkalpak: What about changes detect during this digest ?

test/ng/compileSpec.js
```diff
@@ -3515,6 +3515,284 @@ describe('$compile', function() {
...
+          // We add this watch after the compilation to ensure that it will run after the binding watchers
+          // therefore triggering the thing that this test is hoping to enfore
+          $rootScope.$watch('a', function(val) { $rootScope.b = val * 2; });
+
+          // There should be no changes initially
+          expect(log).toEqual([]);
+
+          // Update val to trigger the onChanges
+          $rootScope.$apply('a = 42');
+          // Now we the change should have the real previous value (undefined), not the intemediate one (42)
```
gkalpak: -we
intemediate  -->  intermediate

test/ng/compileSpec.js
```diff
@@ -3515,6 +3515,284 @@ describe('$compile', function() {
...
+          expect(log).toEqual([{prop: {previousValue: undefined, currentValue: 126}}]);
+
+          // Clear the log
+          log = [];
+
+          // Update val to trigger the onChanges
+          $rootScope.$apply('a = 7');
+          // Now we the change should have the real previous value (126), not the intemediate one, (91)
```
gkalpak: Same typos as above.

src/ng/compile.js
gkalpak: It might be a good idea to document (here or elsewhere) that parent->child execution order.

src/ng/compile.js
petebacondarwin: Doh!

src/ng/compile.js
petebacondarwin: The `flushOnChangesQueue()` is only called if there are entries

src/ng/compile.js
petebacondarwin: Good call...
All the items in the onChanges queue will be flushed before the new digest is triggered.

The sequence looks like:

* some change 1
* some change 2
* $digest
  * schedule flushOnChangesQueue
  * queue onChanges 1
  * queue onChanges 2
* $$postDigest
  * flushOnChangesQueue
    * $apply
      * trigger onChanges 1
        * some change 1a
      * trigger onChanges 2
        * some change 2a
      * $digest
        * schedule onChangesFlush
        * queue onChanges 1
        * queue onChanges 2
      * $$postDigest
        * flushOnChangesQueue
          * $apply
            * trigger onChanges 1
            * trigger onChanges 2
          * $digest

I think this means that we are OK (except for the potential for deep recursion). I'll make a test

src/ng/compile.js
gkalpak: Right :grin:

src/ng/compile.js
petebacondarwin: Added a test @gkalpak PTAL

src/ng/compile.js
```diff
@@ -293,9 +293,23 @@
  *    `true` if the specified slot contains content (i.e. one or more DOM nodes).
  *
  * The controller can provide the following methods that act as life-cycle hooks:
- * * `$onInit` - Called on each controller after all the controllers on an element have been constructed and
+ * * `$onInit()` - Called on each controller after all the controllers on an element have been constructed and
  *   had their bindings initialized (and before the pre &amp; post linking functions for the directives on
  *   this element). This is a good place to put initialization code for your controller.
+ * * `$onChanges(changesObj)` - Called whenever one-way (`<`) bindings are updated. The `changesObj` is a hash
```
gkalpak: Now we need to change this again to also mention `@` :grin:

src/ng/compile.js
```diff
@@ -3101,6 +3152,33 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         }
       });
 
+      function recordChanges(key, currentValue, previousValue) {
+        if (isFunction(destination.$onChanges) && currentValue !== previousValue) {
+          // If we have not already scheduled the top level onChangesQueue handler then do so now
+          if (!onChangesQueue) {
+            scope.$$postDigest(flushOnChangesQueue);
```
dcherman: Why `$$postDigest` versus `$evalAsync`?  Didn't pull this down to test, but `$evalAsync` would feel like the more correct method to use here so that we continue to live within the same digest cycle and don't accidentally violate the max TTL due to handling record changes

src/ng/compile.js
```diff
@@ -2360,10 +2392,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
         forEach(elementControllers, function(controller) {
-          if (isFunction(controller.instance.$onInit)) {
-            controller.instance.$onInit();
+          var controllerInstance = controller.instance;
+          if (isFunction(controllerInstance.$onInit)) {
+            controllerInstance.$onInit();
+          }
+          if (isFunction(controllerInstance.$onDestroy)) {
+            controllerScope.$on('$destroy', function callOnDestroyHook() {
+              controllerInstance.$onDestroy();
```
dcherman: This might be a ridiculous edge case, but is there any concern about removal of the hook during the lifetime of the controller?  `$onDestroy` may not necessarily exist on the instance in the future even though it does right now.

src/ng/compile.js
petebacondarwin: We want all the changes in the current digest to settle before we trigger the onChanges hook.
For this to happen we need the current digest to have completed, which is why postDigest is a good place. Perhaps you can convince me otherwise?

src/ng/compile.js
petebacondarwin: For the moment I don't want to support dynamic changes to the component hook methods: neither lazy definition nor removal. If this is a really common use case we could consider implementing it later without introducing a breaking change.

src/ng/compile.js
gkalpak: So, this looks good apart from the infinite loop issue.

Ideally, there should be some mechanism in place to prevent freezing the whole app, because someone misused `$onChanges` (a TTL-like mechanism maybe).
If not (e.g. it turns out to be too complicated), we should at least document very clearly the dangers of making changes that would trigger another `$onChanges` round. from inside an `$onChanges()` callback.

gkalpak: So, this looks good apart from the infinite loop issue.

Ideally, there should be some mechanism in place to prevent freezing the whole app, because someone misused `$onChanges` (a TTL-like mechanism maybe).
If not (e.g. it turns out to be too complicated), we should at least document very clearly the dangers of making changes that would trigger another `$onChanges` round. from inside an `$onChanges()` callback.

gkalpak:
LGTM with https://github.com/angular/angular.js/pull/14302/files#r57215106 (somehow) addressed
:+1:

src/ng/compile.js
petebacondarwin: It is not an infinite loop, it is an infinite recursion, which will trigger a `RangeError` exception that will prevent the browser from freezing and indicate to the developer that they have done a bad thing.

Added a test to demonstrate.

petebacondarwin:
I've added time-to-live constraints for onChanges hooks, which should address @gkalpak's last concern

docs/content/error/$compile/infchng.ngdoc
```diff
@@ -0,0 +1,24 @@
...
+Angular detects this situation and prevents an infinite loop from causing the browser to become unresponsive.
+
+For example, the situation can occur by setting up a `$onChanges()` hook which triggers an event on the component, which subsequently
+triggers the component's bound inputs to be updated:
+
+```html
+<c1 prop="a" on-change="a = -a"></c1>
+```
+
+```js
```
Narretz: Can you add the component definition object? Makes it easier to understand if you know the bindings are `bindings: {'prop': '<', onChange: '&'}`

src/ng/compile.js
```diff
@@ -1221,6 +1221,34 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+   *
+   * In complex applications it's possible that dependencies between `$onChanges` hooks and bindings will result
+   * in several iterations of calls to these hooks. However if an application needs more than the default 10
+   * iterations to stabilize then you should investigate what is causing the model to continuously change during
+   * the `$onChanges` hook execution.
+   *
+   * Increasing the TTL could have performance implications, so you should not change it without proper justification.
+   *
+   * @param {number} limit The number of `$onChanges` hook iterations.
+   */
```
gkalpak: Missing `@returns` docs.

src/ng/compile.js
```diff
@@ -1221,6 +1221,34 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+  this.onChangesTtl = function(value) {
+    if (arguments.length) {
+      TTL = value;
+    }
```
gkalpak: I suggest returning `this` for consistency. (And then also document it :smiley:)

src/ng/compile.js
```diff
@@ -3156,6 +3190,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         if (isFunction(destination.$onChanges) && currentValue !== previousValue) {
           // If we have not already scheduled the top level onChangesQueue handler then do so now
           if (!onChangesQueue) {
+            // Initialize the TTL tracker for tracking stack overflow
+            if (isUndefined(onChangesTtl)) { onChangesTtl = TTL; }
```
gkalpak: Shouldn't it somehow be reset (at some point) ?

docs/content/error/$compile/infchng.ngdoc
petebacondarwin: Done

src/ng/compile.js
petebacondarwin: Done

src/ng/compile.js
petebacondarwin: Done

src/ng/compile.js
```diff
@@ -1207,6 +1221,36 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+   *
+   * In complex applications it's possible that dependencies between `$onChanges` hooks and bindings will result
+   * in several iterations of calls to these hooks. However if an application needs more than the default 10
+   * iterations to stabilize then you should investigate what is causing the model to continuously change during
+   * the `$onChanges` hook execution.
+   *
+   * Increasing the TTL could have performance implications, so you should not change it without proper justification.
+   *
+   * @param {number} limit The number of `$onChanges` hook iterations.
+   * @returns {number|object} the current limit (or `this` if called as a setting for chaining)
```
gkalpak: _setting_  -->  _setter_

src/ng/compile.js
```diff
@@ -1215,6 +1259,31 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
     var SIMPLE_ATTR_NAME = /^\w/;
     var specialAttrHolder = document.createElement('div');
+
+
+
+    var onChangesTtl;
```
gkalpak: With the current implementation couldn't we initialize `onChangeTtl` here ?

src/ng/compile.js
```diff
@@ -1215,6 +1259,31 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+    // The onChanges hooks should all be run together in a single digest
+    // When changes occur, the call to trigger their hooks will be added to this queue
+    var onChangesQueue;
+
+    // This function is called in a $$postDigest to trigger all the onChanges hooks in a single digest
+    function flushOnChangesQueue() {
+      if (!(onChangesTtl--)) {
+        throw $compileMinErr('infchng', '{0} $onChanges() iterations reached. Aborting!\n', TTL);
```
gkalpak: I believe we should increment `onChangesTtl` before throwing.

src/ng/compile.js
dcherman: No need anymore FWIW.  The addition of the other TTL mechanism mitigated most of my concern, though it'd be sweet if the existing TTL could have been passed along rather then implementing another one so that the max TTL can be defined in one place.

src/ng/compile.js
petebacondarwin: true

src/ng/compile.js
petebacondarwin: OK!

petebacondarwin: Actually we need to reset it to the original value TTL

src/ng/compile.js
```diff
@@ -1215,6 +1259,32 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+
+    var onChangesTtl = TTL;
+    // The onChanges hooks should all be run together in a single digest
+    // When changes occur, the call to trigger their hooks will be added to this queue
+    var onChangesQueue;
+
+    // This function is called in a $$postDigest to trigger all the onChanges hooks in a single digest
+    function flushOnChangesQueue() {
+      if (!(onChangesTtl--)) {
+        onChangesTtl = TTL;
```
gkalpak: I don't think this is correct. We should just `onChangesTtl++`, because the exception will be caught in this digest's `catch` block, so the rest of the ++ (on https://github.com/angular/angular.js/pull/14302/files#diff-a732922b631efed1b9f33a24082ae0dbR1284) will be executed anyway.

src/ng/compile.js
```diff
@@ -3101,6 +3190,34 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         }
       });
 
+      function recordChanges(key, currentValue, previousValue) {
+        if (isFunction(destination.$onChanges) && currentValue !== previousValue) {
+          // If we have not already scheduled the top level onChangesQueue handler then do so now
+          if (!onChangesQueue) {
+            // Initialize the TTL tracker for tracking stack overflow
```
gkalpak: This comment is no longer needed.

src/ng/compile.js
petebacondarwin: Hi @doberman can you clarify what you mean here?
Are you saying that the use of EvalAsync is not needed any more?

I did consider piggybacking on the rootscope TTL but since it is not actually related to the digest cycle I  thought we should keep them independent

PR #14308, opened 2016-03-23
fix(ngMessages): create new scope for ngMessage, clean it up correctly

### LINKED ISSUE #14307: ngMessages with bound message doesn't remove watch when message is removed
We have ngMessage texts which can change, and so are bound to model values rather than being hardcoded into the view.

eg.

```
<div ng-messages="form.input.$error">
    <div ng-message="myError">{{errorMsg}}</div>
</div>
```

Every time the message is shown, it adds a new watch which is not removed when the message is removed, but stays added until the parent ng-messages element is destroyed.

JSFiddle showing the problem is available here https://jsfiddle.net/jwalkersl/Lumjd9tm/

Narretz:
bug fix

#14307

ngMessage creates a child scope instead of reusing the ngMessages scope.

Possibly. Needs investigation.

**Other information**:

Previously, ngMessage elements used the same scope as ngMessages. When ngMessage
has interpolation in the textContent, then removing the message would not remove
the watcher from the scope - it would only be removed when the whole ngMessages
element was removed.

This commit changes the ngMessage transclude function to create a new child scope
instead, which can be destroyed safely when the message element is removed and
the message is detached

Fixes #14307

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -684,7 +684,7 @@ function ngMessageDirectiveFactory() {
           },
           attach: function() {
             if (!currentElement) {
-              $transclude(scope, function(elm) {
+              $transclude(function(elm, newScope) {
                 $animate.enter(elm, null, element);
                 currentElement = elm;
 
@@ -700,6 +700,7 @@ function ngMessageDirectiveFactory() {
                     ngMessagesCtrl.deregister(commentNode);
                     messageCtrl.detach();
                   }
+                  newScope.$destroy();
                 });
               });
             }
--- a/test/ngMessages/messagesSpec.js
+++ b/test/ngMessages/messagesSpec.js
@@ -609,6 +609,34 @@ describe('ngMessages', function() {
     });
   });
 
+
+  it('should clean-up the ngMessage scope when a message is removed',
+    inject(function($compile, $rootScope) {
+
+      var html =
+          '<div ng-messages="items">' +
+            '<div ng-message="a">{{forA}}</div>' +
+          '</div>';
+
+      element = $compile(html)($rootScope);
+      $rootScope.$apply(function() {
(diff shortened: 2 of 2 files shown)


test/ngMessages/messagesSpec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-describe('ngMessages', function() {
+fdescribe('ngMessages', function() {
```
gkalpak: Ooops

test/ngMessages/messagesSpec.js
```diff
@@ -485,6 +485,51 @@ describe('ngMessages', function() {
     });
   });
 
+
+  it('should clean-up the ngMessage scope when a message is removed',
+    inject(function($compile, $rootScope) {
+      function countWatchers() {
```
gkalpak: There is `$rootScope.$countWatchers()` (see https://docs.angularjs.org/api/ngMock/type/$rootScope.Scope).

src/ngMessages/messages.js
```diff
@@ -654,16 +654,17 @@ function ngMessageDirectiveFactory() {
...
           test: function(name) {
             return contains(records, name);
           },
           attach: function() {
             if (!currentElement) {
-              $transclude(scope, function(elm) {
+              $transclude(function(elm, newScope) {
                 $animate.enter(elm, null, element);
                 currentElement = elm;
+                currentScope = newScope;
```
gkalpak: Another alternative, is to bind to the element's `$destroy` event and clean up from there:

```js
elm.on('$destroy', function () { newScope.$destroy(); });
```


It has the benefits that you don't need to keep track of the `currentScope` and you are "safe" in case the element is somehow removed without going through the `detach()` method (although that's probably not possible, but better safe than sorry).

PR #14313, opened 2016-03-24
Chore: add jshint eqeqeq

Narretz:
chore

we don't enforce strict comparison for our src files

now we do!

No

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1259,7 +1259,7 @@ function createDateParser(regexp, mapping) {
       // When a date is JSON'ified to wraps itself inside of an extra
       // set of double quotes. This makes the date parsing code unable
       // to match the date string and parse it as a date.
-      if (iso.charAt(0) == '"' && iso.charAt(iso.length - 1) == '"') {
+      if (iso.charAt(0) === '"' && iso.charAt(iso.length - 1) === '"') {
         iso = iso.substring(1, iso.length - 1);
       }
       if (ISO_DATE_REGEXP.test(iso)) {
@@ -1476,6 +1476,8 @@ function radioInputType(scope, element, attr, ctrl) {
 
   ctrl.$render = function() {
     var value = attr.value;
+    // Strict comparison would cause a BC
+    /* jshint eqeqeq:false */
     element[0].checked = (value == ctrl.$viewValue);
   };
 
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -258,6 +258,8 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
           }
 
           function getRadioReaction(newVal) {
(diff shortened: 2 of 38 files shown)


src/ng/directive/input.js
```diff
@@ -1476,6 +1476,8 @@ function radioInputType(scope, element, attr, ctrl) {
 
   ctrl.$render = function() {
     var value = attr.value;
+    // Strict comparison would cause a BC
```
gkalpak: Are you sure ? From a (super) quick look, we seems to be updating `$viewValue` with `attr.value` and here we are comparing it with `attr.value` again.

Narretz: I've added a test here: https://github.com/angular/angular.js/pull/14313/files#diff-ce6d0f141f60d25eede7abf044d04b93 Without strict comparison, you can set your model to `0` and it will still check a radio button that has `value="0"`(string)

Narretz:
There are a few test failures in ngMessageFormat, but only when running tests:modules ...

src/ngAria/aria.js
```diff
@@ -258,6 +258,8 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
           }
 
           function getRadioReaction(newVal) {
+            // Strict comparison would cause a BC
```
gkalpak: It's inheriting the behavior from core, right ?

Narretz: Yep!

test/ng/parseSpec.js
```diff
@@ -1739,6 +1739,7 @@ describe('parser', function() {
         expect(scope.$eval("!true")).toBeFalsy();
         expect(scope.$eval("1==1")).toBeTruthy();
         expect(scope.$eval("1==true")).toBeTruthy();
+        expect(scope.$eval('1!=true')).toBeFalsy();
```
gkalpak: :+1:

gkalpak:
LGTM

--
I think this is my favorit PR ever :smile: 
(It will only come second to the (future) one that replaces double-quotes with single-quotes (where appropriate). :stuck_out_tongue:

Narretz:
@gkalpak Could you look at the commit I've added?

src/ngMessageFormat/messageFormatParser.js
```diff
@@ -189,7 +189,7 @@ MessageFormatParser.prototype.ruleInsideString = function ruleInsideString() {
         position.line, position.column, this.text);
   }
   var chars = match[0];
-  if (match == this.stringQuote) {
+  if (match[0] === this.stringQuote) {
```
gkalpak: Nitpick: Either use `chars` or remove it (in the line above), since it's not used elsewhere.

gkalpak:
LGTM (with one nitpick).

PR #14314, opened 2016-03-24
chore(angularFiles): add missing files

Narretz:
ngMessageFormat test files were missing from angularTest
ngMessages and ngAnimate were missing from karmaModules. The test
files were using two APIs that are not exposed in the module tests.
jqLiteHasClass has been changed to use angular.element(..).hasClass()
instead, and for ngMessagesSpec a trim() fn has been added


test/helpers/matchers.js
```diff
@@ -218,7 +218,7 @@ beforeEach(function() {
           };
           var classes = clazz.trim().split(/\s+/);
           for (var i = 0; i < classes.length; ++i) {
-            if (!jqLiteHasClass(actual[0], classes[i])) {
+            if (!angular.element(actual[0]).hasClass(classes[i])) {
```
Narretz: I guess this is a bit heavier than calling jqLiteHasClass directly, so it might make the tests run slower.

gkalpak: It seems that `actual` is already a jqLite/jQuery element. We could probably do `actual.hasClass()`

Narretz: Doesn't the identity of atual depend on what is passed to `expect`?

gkalpak: Yes. `actual` is what we pass to `expect`.

PR #14322, opened 2016-03-26
perf(injector): cache the results of the native class detection check

thorn0:

Performance improvement

Relatively expensive `Function.prototype.toString` is called every time a function is checked for being a native class. See #14268.

The results of the check are cached in the `$$ngIsClass` property of the function.

no

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -837,13 +837,16 @@ function createInjector(modulesToLoad, strictDi) {
 
     function isClass(func) {
       // IE 9-11 do not support classes and IE9 leaks with the code below.
-      if (msie <= 11) {
+      if (msie <= 11 || typeof func !== 'function') {
         return false;
       }
-      // Workaround for MS Edge.
-      // Check https://connect.microsoft.com/IE/Feedback/Details/2211653
-      return typeof func === 'function'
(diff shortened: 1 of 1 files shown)


lgalfaso:
I would call the property `$$ngIsClass` just to be sure there is no collision with another library that is also trying to identity if a given function is a class. Otherwise LGTM

thorn0:
Done

lgalfaso:
landed as 5ceb5dbfa6d9b6d15232a1f5c767b2f431325948

gkalpak:
@lgalfaso, should we backport this to `v1.5.x` ?

lgalfaso:
@gkalpak I lean towards not backporting this, as we are adding a property that may or may not be expected. That said, I would not oppose to this being backported.

gkalpak:
I don't feel strongly about it either. But if we consider the new property a potential BC, maybe we should mention it as such.

thorn0:
Adding this property to a constructor by hand seems to be a way to turn off the deprecated behavior of the bindings even when native classes aren't used. That's something I'd like to do in my projects, so please backport this PR to 1.5. Or probably let's just add an option to $compileProvider for disabling that behavior .

PR #14323, opened 2016-03-26
feat($compile): add `isFirstChange()` method to onChanges object

### LINKED ISSUE #14318: $onChanges isn't called on the initialization of the component

Feature (I guess)

`$onChanges` isn't called on the initialization of the component.

The motivation is to match the behavior of Angular 2. In A2, `ngOnChanges` is called on the initialization of the component too. And the change object includes the `isFirstChange` method, which in this case returns `true`.

A plunk that shows how `ngOnChanges` works in Angular 2: http://plnkr.co/edit/kIf8QgrFxtENzvr6V5TZ?p=preview

petebacondarwin:

Feature

Fixes #14318

Adds `isFirstChange()` method to changes objects

No - although there is now an extra onChanges call if items are not updated during the first digest since the bindings were defined.

**Other information**:

Closes #14318

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -156,7 +156,7 @@ of the component. The following hook methods can be implemented:
     this element). This is a good place to put initialization code for your controller.
   * `$onChanges(changesObj)` - Called whenever one-way bindings are updated. The `changesObj` is a hash whose keys
     are the names of the bound properties that have changed, and the values are an object of the form
-    `{ currentValue: ..., previousValue: ... }`. Use this hook to trigger updates within a component such as
+    `{ currentValue, previousValue, isFirstChange() }`. Use this hook to trigger updates within a component such as
     cloning the bound value to prevent accidental mutation of the outer value.
   * `$onDestroy()` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
     external resources, watches and event handlers.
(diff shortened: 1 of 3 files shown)


thorn0:
Is `$onChanges` called before or after `$onInit`?

gkalpak:
@thorn0, before `$onInit`. Isn't that how ng2 does it as well ?

@petebacondarwin, LGTM (despite some whitespace inconsistencies :stuck_out_tongue: )

petebacondarwin:
@thorn0 - it seems that A2 does it this way: https://angular.io/docs/ts/latest/api/core/OnChanges-interface.html

@gkalpak - it passes JSCS :-P

petebacondarwin:
@gkalpak - fixed the white-space incongruities

thorn0:
> it seems that A2 does it this way

Although it does, this order doesn't look reasonable to me. :) But that's an issue for the A2 project...

petebacondarwin:
@thorn0 - I think the idea is that you can choose to ignore this "pre-init" ngChanges by checking the `isFirstChange()` method. But maybe you are suggesting that they only added that BECAUSE it was too difficult to ensure that this first change hook was not called.

thorn0:
I just have doubts about the idea of calling anything on a not fully initialized component. As for the initial call of `$onChanges` in general, I think it's a good thing. For example, for use cases when the component has 'calculated' properties that depend on bindings.

Issue #14120, opened 2016-02-24
transition events not always fired


wesleycho:
**Note: for support questions, please use one of these channels:** https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.

* **Do you want to request a *feature* or report a *bug*?**

Bug

* **What is the current behavior?**

In certain situations, when the user switches to a different tab, the `transitionend` event is not fired - this is a documented browser defect with Firefox [here](https://bugzilla.mozilla.org/show_bug.cgi?id=683696), and seems to be reproducible in Chrome at the least.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:yBpEi4).

Not minimal unfortunately, but [here](http://plnkr.co/edit/pGRtrLoA7NOWfeOAQYnW?p=preview) is an example from UI Bootstrap. Source code for component can be found [here](https://github.com/angular-ui/bootstrap/blob/master/src/carousel/carousel.js) - in particular, the culprit is that [this listener never fires](https://github.com/angular-ui/bootstrap/blob/master/src/carousel/carousel.js#L208-L221) due to this browser defect, which causes the callback to never execute.

* **What is the expected behavior?**

The `transitionend` event fires and the slides continue rotating since the `$currentTransition` internal state is reset.

* **What is the motivation / use case for changing the behavior?**

My proposal is to have ngAnimate use the computed style information to manually fire the `transitionend` event if it is never fired to cause the listeners to be resolved - a timeout for the transition time period + 1 ms would be satisfactory to me, which would prevent library maintainers to have to do an extra `getComputedStyle` call & an extra calculation of the transition time length. This is a hack, but this is the least onerous solution IMO.

* **Which version of Angular, and which browser and OS does this issue affect?** Did this work in previous
versions of Angular? Please also test with the latest stable ánd snapshot versions.

Angular 1.4+ when using $animateCss instead of $animate. This affects at the least Chrome & Firefox across all OSs at the least.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix)

Narretz:
animateCss actually already has a timeout that should end the animation after (I think) duration * 1.5. That the end listener isn't executed seems like a bug, though.

Narretz:
[...]

PR #14341, opened 2016-03-29
fix: call $animate.on() listeners when document is hidden

Narretz:
Regression fix

See #14120

No

**Other information**:

The first commit fixes a very rare case of callbacks being called in the wrong order, and has nothing to do with the regression (but it was found while fixing it)

### DIFF
--- a/test/ngAnimate/animateSpec.js
+++ b/test/ngAnimate/animateSpec.js
@@ -2323,6 +2323,36 @@ describe("animations", function() {
         expect(classSpy).not.toHaveBeenCalled();
       }));
 
+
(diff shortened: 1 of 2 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -2240,19 +2240,87 @@ describe("animations", function() {
         $rootScope.$digest();
         $animate.flush();
 
-        expect(moveSpy.callCount).toBe(1);
```
matsko: Is there a reason why we're avoiding this?

matsko:
LGTM

PR #14342, opened 2016-03-29
docs(guide/Forms): describe your change...

DrDanRyan:
Typo bug fix

The last example does not function due to mis-capitalization of directive name and element attribute

Fix broken example

Don't think so

**Other information**:

Typos in contentEditable directive example

### DIFF
--- a/docs/content/guide/forms.ngdoc
+++ b/docs/content/guide/forms.ngdoc
@@ -480,7 +480,7 @@ The following example shows how to add two-way data-binding to contentEditable e
     <pre>model = {{content}}</pre>
 
     <style type="text/css">
-      div[contentEditable] {
+      div[content-editable] {
         cursor: pointer;
         background-color: #D0D0D0;
       }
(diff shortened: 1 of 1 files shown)


docs/content/guide/forms.ngdoc
```diff
@@ -488,7 +488,7 @@ The following example shows how to add two-way data-binding to contentEditable e
   </file>
 
   <file name="script.js">
-    angular.module('form-example2', []).directive('contenteditable', function() {
+    angular.module('form-example2', []).directive('contentEditable', function() {
```
gkalpak: This would look for a `content-editable` attribute (or element), not `contenteditable`.
(Note that HTML is case-insensitive.)

docs/content/guide/forms.ngdoc
```diff
@@ -480,7 +480,7 @@ The following example shows how to add two-way data-binding to contentEditable e
     <pre>model = {{content}}</pre>
 
     <style type="text/css">
-      div[contentEditable] {
+      div[content-editable] {
```
gkalpak: This will not match our element, which has a `contentEditable` attribute.

gkalpak:
Thx for the PR, but I'm afraid we'll have to pass this time. The changes will actually break the example (see inline comments).
Thx for taking the time anyway. Looking forward to your next PR :stuck_out_tongue:

PR #14345, opened 2016-03-30
refactor(jshint): don't assume browser-only globals

### LINKED ISSUE #13442: Uncaught ReferenceError: Node is not defined

mgol:

Feature: We're no longer assuming browser globals in `src/**/*.js` which will prevent issues like #13442 from happening in the future.

Currently browser globals are assumed. This sometimes breaks tools like jsdom that make them available under `window` but not as globals; see e.g. #13442.

For the browser these changes shouldn't matter. In other environments browser globals are now taken from the `window` variable instead of the global.

No.

**Other information**:

Fixes #13442

### DIFF
--- a/docs/gulpfile.js
+++ b/docs/gulpfile.js
@@ -1,5 +1,9 @@
 "use strict";
 
+var fs = require('fs');
+var _ = require('lodash');
+var stripJsonComments = require('strip-json-comments');
+
 var gulp = require('gulp');
 var log = require('gulp-util').log;
 var concat = require('gulp-concat');
@@ -24,6 +28,23 @@ var src = 'app/src/**/*.js';
 var ignoredFiles = '!src/angular.bind.js';
 var assets = 'app/assets/**/*';
 
+
+var getJshintConfig = function(filepath) {
+    return JSON.parse(stripJsonComments(fs.readFileSync(filepath, {encoding: 'utf-8'})));
+};
+
+var getMergedJshintConfig = function(filepath) {
+  // "extends" doesn't work in configuration passed by an object, we need to do the extending ourselves.
+  var config = getJshintConfig(filepath);
+  var baseConfig = getJshintConfig('../.jshintrc-base');
+  _.merge(config, baseConfig);
+  delete config.extends;
+
+  // Examples don't run in strict mode; accept that for now.
+  config.strict = false;
+
+  return config;
+};
 
 var copyComponent = function(component, pattern, sourceFolder, packageFile) {
   pattern = pattern || '/**/*';
@@ -90,17 +111,35 @@ gulp.task('assets', ['bower'], function() {
 
 gulp.task('doc-gen', ['bower'], function() {
   var dgeni = new Dgeni([require('./config')]);
-  return dgeni.generate().catch(function(error) {
+  return dgeni.generate().catch(function() {
     process.exit(1);
   });
 });
 
 // JSHint the example and protractor test files
 gulp.task('jshint', ['doc-gen'], function() {
-  gulp.src([outputFolder + '/ptore2e/**/*.js', outputFolder + '/examples/**/*.js'])
-    .pipe(jshint())
-    .pipe(jshint.reporter('jshint-stylish'))
-    .pipe(jshint.reporter('fail'));
+  var examplesConfig = getMergedJshintConfig('../docs/app/test/.jshintrc');
+  // Some tests use `alert` which is not assumed to be available even with `"browser": true`.
+  examplesConfig.globals.alert = false;
+
+  var protractorConfig = getMergedJshintConfig('../docs/app/e2e/.jshintrc');
+
+  return merge(
+    gulp.src([
+      outputFolder + '/examples/**/*.js',
+      '!' + outputFolder + '/examples/**/protractor.js',
+    ])
+      .pipe(jshint(examplesConfig))
+      .pipe(jshint.reporter('jshint-stylish'))
+      .pipe(jshint.reporter('fail')),
+    gulp.src([
+      outputFolder + '/ptore2e/**/*.js',
+      outputFolder + '/examples/**/protractor.js',
+    ])
+      .pipe(jshint(protractorConfig))
+      .pipe(jshint.reporter('jshint-stylish'))
+      .pipe(jshint.reporter('fail'))
+  );
 });
 
 
--- a/src/angular.prefix
+++ b/src/angular.prefix
@@ -3,4 +3,4 @@
  * (c) 2010-2016 Google, Inc. http://angularjs.org
  * License: MIT
  */
-(function(window, document, undefined) {
+(function(window) {
--- a/src/module.prefix
+++ b/src/module.prefix
@@ -3,4 +3,4 @@
  * (c) 2010-2016 Google, Inc. http://angularjs.org
  * License: MIT
  */
-(function(window, angular, undefined) {
+(function(window, angular) {
--- a/src/ngScenario/angular.prefix
+++ b/src/ngScenario/angular.prefix
@@ -3,5 +3,5 @@
  * (c) 2010-2016 Google, Inc. http://angularjs.org
  * License: MIT
  */
-(function(window, document){
+(function(window){
   var _jQuery = window.jQuery.noConflict(true);
--- a/.jshintrc-base
+++ b/.jshintrc-base
@@ -1,19 +1,24 @@
 {
   "bitwise": true,
+  "esversion": 6,
   "immed": true,
   "newcap": true,
   "noarg": true,
   "noempty": true,
   "nonew": true,
-  "trailing": true,
   "maxlen": 200,
   "boss": true,
   "eqnull": true,
   "expr": true,
-  "globalstrict": true,
   "laxbreak": true,
   "loopfunc": true,
+  "strict": "global",
   "sub": true,
   "undef": true,
-  "indent": 2
+  "indent": 2,
+
+  "globals": {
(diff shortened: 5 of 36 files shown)


mgol:
i also updated JSHint.

mgol:
@petebacondarwin I see JSHint failures in the docs gulp task. Which JSHint config is used to lint `build/docs/**/*.js`?

petebacondarwin:
@mgol I guess it is just the top level `.jshintrc` in the root of the project.

petebacondarwin:
If you want we could copy a docs specific `.jshintrc` into the `build/docs` folder as part of the `gulp assets` task.

mgol:
Where are globals like `by` taken from? It surprised me to see errors like:

    'by' is not defined

gkalpak:
`by` should be from Protractor

--
WRT `.jshintrc`, I believe it's better to take a more granular, specifying globals in the directories they are needed. E.g. don't have something in the top level if we only need it in `src/` or `test/` etc.

petebacondarwin:
@gkalpak - in that case we need dgeni to create and provide such a file for every example that contains e2e tests.

gkalpak:
Why is that ? Can't we just put the required globals in a parent directory's `.jshintrc` ?

petebacondarwin:
OK, so I forgot that the e2e tests all go in their own folder.
The relevant folders we have are:

```
docs
 - js
 - examples
 - ptore2e
```

So we should copy a baseline `.jshintrc` in the `docs` folder, then perhaps a specialized one inside the `ptoe2e`.

mgol:
> So we should copy a baseline .jshintrc in the docs folder

We can extend a common one, no need to copy all the settings. :)

lgalfaso:
Would it be possible to use `$window` in the cases that this is possible?

mgol:
@lgalfaso It would but I'm pretty sure it'd break a lot of tests as people are mocking `$window` and not providing all necessary fields like `setTimeout` so I don't think that's something we should attempt in 1.5.x.

Also, we're currently providing globals `window` & `document` as [parameters to the factory](https://github.com/angular/angular.js/blob/b54634d37b69bd3f7f563d0718873fe9f50eb2d1/src/angular.suffix#L5); we'd need to drop `document` from there and always use `$window.document`.

We might consider all that for 1.6 but I wanted to land something in 1.5.x and it seems too risky to me. WDYT?

lgalfaso:
Fair point, let's keep window

gkalpak:
@mgol, now that you mention it, couldn't we just change `document` to `window.document` in [angular.suffix#L5][1] and leave it as just `document` in most other places ?

[1]: https://github.com/angular/angular.js/blob/b54634d37b69bd3f7f563d0718873fe9f50eb2d1/src/angular.suffix#L5

mgol:
@gkalpak I've already talked with @petebacondarwin about that. :) I plan to apply this change.

mgol:
So, I've been wondering about this change. Since I'm removing `"browser": true` I need to specify each browser global that should be used directly in `.jshintrc`. But that would mean I need to add `document` there as well. If we go this route, we might also want to add `window.setTimeout` & friends to the factory parameters in https://github.com/angular/angular.js/blob/b54634d37b69bd3f7f563d0718873fe9f50eb2d1/src/angular.suffix#L5 but then they'd need to be added there as well... Perhaps we need one base `.jshintrc` for all source files that would declare these assumed browser globals? Or maybe we should just prefix all browser globals with `window.` and not do all that? Thoughts?

petebacondarwin:
Don't we have that situation right now?
The `.jshintrc` in the `src` folder is responsible (really) for the `ng` folder source files and the top level source files such as `Angular.js` and `jqLite.js`. Then each of the modules has its own `.jshintrc`. Each of these extends the `jshintrc-base` file.

mgol:
The `.jshintrc-base` file is a base for the whole project but I want the few browser globals to only be defined in code run in browsers, not e.g. Node. We could perhaps get a separate `.jshintrc-base` created under `src/` that would extend the main base config & add those globals from the factory and then extend this file?

mgol:
OTH, as @lgalfaso mentioned, ideally we'd use `$window` where appropriate but we can't do it that way before 1.6.0. But if we want to do it in the end, keeping anything except `window` in the factory seems backwards. So we should decide if we want to change it in 1.6.0 and if the answer's yes I'd remove `document` from the factory & just prefix every non-core JS browser global with `window.`.

petebacondarwin:
@mgol I think that your suggestion based on moving to using `$window` everywhere in 1.6 and prefixing all browser globals with `window` in 1.5 sounds reasonable. Do you want to have a go at that and see if it turns up any difficulties?

mgol:
@petebacondarwin on it!

mgol:
PR updated.

mgol:
All tests have passed this time.

petebacondarwin:
Sweet!

docs/gulpfile.js
```diff
@@ -88,7 +109,7 @@ gulp.task('assets', ['bower'], function() {
 });
 
 
-gulp.task('doc-gen', ['bower'], function() {
+gulp.task('doc-gen', ['bower'], function(cb) {
```
petebacondarwin: I don't think we need the callback here

src/angular.prefix
```diff
@@ -3,4 +3,4 @@
  * (c) 2010-2016 Google, Inc. http://angularjs.org
  * License: MIT
  */
-(function(window, document, undefined) {
+(function(window) {
```
petebacondarwin: I think that having `undefined` as a parameter, which is not assigned, is actually a clever way to ensure that we have the "real" `undefined` inside the closure. So I am for leaving this as a parameter.

src/module.prefix
```diff
@@ -3,4 +3,4 @@
  * (c) 2010-2016 Google, Inc. http://angularjs.org
  * License: MIT
  */
-(function(window, angular, undefined) {
+(function(window, angular) {
```
petebacondarwin: see my comment on `undefined` above

src/ngScenario/angular.prefix
```diff
@@ -3,5 +3,5 @@
  * (c) 2010-2016 Google, Inc. http://angularjs.org
  * License: MIT
  */
-(function(window, document){
+(function(window){
```
petebacondarwin: perhaps we should include `undefined` here too?

src/angular.prefix
mgol: What would it guard against, though? We've used to have the same parameter in jQuery but we removed it; if someone created a variable `undefined` that is not equal to `undefined`, basically everything would be broken.

docs/gulpfile.js
mgol: Right; I forgot to remove it after backtracking from a previous idea. :)

src/angular.prefix
mgol: IOW, shadowning `undefined` with sth else is more or less equivalent to shadowing any other builtin like `Object`, `Array`, `Object.prototype` methods etc. There are lots of ways in which someone can create a broken environment in which Angular breaks. I don't think it buys us much to defend just against shadowing `undefined` and it's impossible to guard against everything.

petebacondarwin:
```
var old_undefined = undefined;
undefined = {};
function moo(undefined) {
  console.log(undefined === old_undefined);
}
moo();
```

mgol: I know what it does. :) What I'm saying is that it doesn't really buy us much; there are thousands other APIs that, if messed with, will break Angular and we don't protect against such changes. Moreover, I think it's more likely that someone will break `Object.prototype` for us than that they'll shadow `undefined`. I feel this `undefined` parameter is only providing a false sense of security.

petebacondarwin: It saves us from using `void 0` everywhere, no?
I agree it doesn't save us from the other stuff, but we do have various bits of defensive code in place in the case that people do override parts of the `Object.prototype`.
What is the benefit in removing it?

mgol: > It saves us from using `void 0` everywhere, no?

I'd just rely on the `undefined` from the outer scope. Note that it's not even possible in ES5 to overwrite `undefined`, it's immutable even in sloppy mode (the only difference is that in strict mode trying to overwrite `undefined` throws while in sloppy mode it's just a noop). The only way this could backfire is if Angular was loaded not via a regular script tag but wrapped in a separate closure which shadowed `undefined`. But that requires modifying the source and if someone does that we've already lost the battle.

> I agree it doesn't save us from the other stuff, but we do have various bits of defensive code in place in the case that people do override parts of the Object.prototype.

Since the global `undefined` is immutable if Angular is loaded normally so that its `undefined` points to the global one no one is able to modify that; shadowing is also impossible after the fact.

> What is the benefit in removing it?

There is no huge benefit, perhaps a tiny size decrease. If you feel strongly about it, I'll restore it. :) I just think it doesn't really solve any problem.

petebacondarwin: OK, you convinced me :-)

petebacondarwin:
OK, LGTM - please merge

gkalpak:
:+1:

Regarding the using `$window` in `1.6.x`, won't that breal lots of tests relying on a mocked `$window` ?

mgol:
@gkalpak https://github.com/angular/angular.js/pull/14345#issuecomment-204503298 :P 

That's why I didn't want to land it for 1.5. As for 1.6, we can still discuss.

PR #14348, opened 2016-03-30
fix(loader): module.decorator order of operations is now irrelevant

### LINKED ISSUE #12382: module.decorator() requires explicit order of operations.
All of the other module-level API methods - service, factory, run, config, etc - can execute in any arbitrary order. However, the new .decorator() method *has to run* after the target service is defined. This is because the feature has been merely piped into the $provide.decorator() method, which used to be required to run in the configuration phase, after all services have been defined. Now that it can be called at any time, the necessary $get() method doesn't always exist. 

As such, this fails:

angular.module( "Foo" ).decorator( "greet", fn ); // error, greetProvider cannot be found.
angular.module( "Foo" ).service( "greet", fn );

... but this works:

angular.module( "Foo" ).service( "greet", fn );
angular.module( "Foo" ).decorator( "greet", fn );

If you could just queue up the .decorator() calls and run them as the first part of the config-queue-flush, I think this would solve the dependency on order-of-operations.

I don't know much about contributing to Angular, so in the future, hopefully I can submit this as part of a PR.

robertmirro:
Bug fix

See #12382 

module.decorator declarations are now invoked via the configBlocks queue to ensure that the provider being decorated has been declared regardless of the order in which module.decorator is declared.

Possibly.  If the same provider is decorated multiple times via a combination of module.decorator and $provide.decorator (via module.config), the decoration order may change now depending on declaration order of the module methods.

**Other information**:

I didn't ngdoc the added `queue` param of `invokeLaterAndSetModuleName` because this fn isn't publically exposed and `queue` wasn't doc'd for `invokeLater`.

module.decorator() is now processed via the configBlocks order of operations and:
- no longer throws if declared before the provider being decorated
- guarantees the correct provider will be decorated when multiple,
  same-name providers are defined

Closes: #12382

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -203,7 +203,7 @@ function setupModuleLoader(window) {
            * @description
            * See {@link auto.$provide#decorator $provide.decorator()}.
            */
-          decorator: invokeLaterAndSetModuleName('$provide', 'decorator'),
+          decorator: invokeLaterAndSetModuleName('$provide', 'decorator', configBlocks),
 
           /**
            * @ngdoc method
@@ -349,10 +349,11 @@ function setupModuleLoader(window) {
          * @param {string} method
          * @returns {angular.Module}
          */
-        function invokeLaterAndSetModuleName(provider, method) {
+        function invokeLaterAndSetModuleName(provider, method, queue) {
+          if (!queue) queue = invokeQueue;
           return function(recipeName, factoryFunction) {
             if (factoryFunction && isFunction(factoryFunction)) factoryFunction.$$moduleName = name;
-            invokeQueue.push([provider, method, arguments]);
+            queue.push([provider, method, arguments]);
             return moduleInstance;
           };
         }
--- a/test/loaderSpec.js
+++ b/test/loaderSpec.js
@@ -48,7 +48,6 @@ describe('module loader', function() {
     expect(myModule.requires).toEqual(['other']);
     expect(myModule._invokeQueue).toEqual([
       ['$provide', 'constant', jasmine.objectContaining(['abc', 123])],
-      ['$provide', 'decorator', jasmine.objectContaining(['dk', 'dv'])],
       ['$provide', 'provider', jasmine.objectContaining(['sk', 'sv'])],
       ['$provide', 'factory', jasmine.objectContaining(['fk', 'fv'])],
       ['$provide', 'service', jasmine.objectContaining(['a', 'aa'])],
@@ -60,9 +59,21 @@ describe('module loader', function() {
     ]);
     expect(myModule._configBlocks).toEqual([
       ['$injector', 'invoke', jasmine.objectContaining(['config'])],
+      ['$provide', 'decorator', jasmine.objectContaining(['dk', 'dv'])],
       ['$injector', 'invoke', jasmine.objectContaining(['init2'])]
     ]);
     expect(myModule._runBlocks).toEqual(['runBlock']);
+  });
+
+
+  it("should not throw error when `module.decorator` is declared before provider that it decorates", function() {
+    angular.module('theModule', []).
+      decorator('theProvider', function($delegate) { return $delegate; }).
+      factory('theProvider', function() { return {}; });
+
+    expect(function() {
+      createInjector(['theModule']);
+    }).not.toThrow();
   });
 
 


Narretz:
Thanks for the PR. I do think we should add a BC note to be on the safe side. Can you please add one, that describes the new behavior, how it's different from the old behavior and how this might break existing code? Small code example (that shows how order affects execution) would also be nice. You can have a look at the Changelog or the Migration guide to see how a BC note should look.

robertmirro:
How?  Amend commit 478650c and re-push?  If so, won't that wipe out your comments above?

Narretz:
Yes amend and force push

Narretz:
The comments are independent of that. And comments on changes are also preserved if the code doesn't change

robertmirro:
Thanks!  Please let me know if the new commit is not what you had in mind.

Narretz:
The commit message looks great, very nice!

I would only change the names `anotherDecoratorFn` and `theDecoratorFn` to `provideDecoratorFn` and `moduleDecoratorFn` to make it easier to see the new order.

I'd also like to see a test that ensures defining a decorator before its provider works.

robertmirro:
I've made the following changes:
- relocated the `Closes` tag to come before `BREAKING CHANGE:` in the commit message
- renamed fn name examples in the commit message
- created an additional unit test


Regarding the additional unit test...

I included the additional unit test in `loaderSpec` (as opposed to `injectorSpec`) because the failure described via #12382 is relevant to `module`.

However, I'm not entirely convinced this test is necessary because this PR is meant to move the processing of `module.decorator` declarations to the `configBlocks` queue. The originally modified unit test in `loaderSpec` proves that change works.


Regarding the `Closes` tag... 

Neither the [Contributing to AngularJS](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#footer) nor the [AngularJS Git Commit Message Conventions](https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#heading=h.uh9jqqu4fjy3) documents seem to specify that the `Closes #<issue number>` tag **needs** to come before `BREAKING CHANGE:`.  In fact, the outline section displayed at the top of [AngularJS Git Commit Message Conventions](https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#) seems to indicate _Referencing issues_ comes last in the _Message footer_.

Many BC commits seem to follow this pattern: [d06431e](https://github.com/angular/angular.js/commit/d06431e5309bb0125588877451dc79b935808134), [98c2db7](https://github.com/angular/angular.js/commit/98c2db7f9c2d078a408576e722407d518c7ee10a), [98528be](https://github.com/angular/angular.js/commit/98528be311b48269ba0e15ba4e3e2ad9b89693a9), [983b059](https://github.com/angular/angular.js/commit/983b0598121a8c5a3a51a30120e114d7e3085d4d), etc...

Maybe the docs can be updated if format is critical and the above is no longer acceptable?

Thanks!

Narretz:
There's a typo in the new loaderSpec test that makes the tests fail.
I know that the test you changed validates the correctness of the fix, but it's important (to me) to have a test that exactly tests this change. The changed test doesn't make it obvious that the order doesn't matter now.

Wrt to the commit message guidelines, putting Closes #... before the BC notice makes the CL output prettier. But we could add that to the guidelines.

robertmirro:
Oops, regarding the typo...

I had been testing via `grunt test:unit` all along.  

I changed naming conventions that are used in the test at the last moment and then ran `grunt test` as a final sanity check before my commit.  

Unfortunately the `package` task that is invoked before `test.unit` stalls on `minall` (more specifically, I had to remove the `parseext` item from `min` in order to resolve the stall in the test process), so I didn't noticed that `test.unit` was never re-run.

This behavior occurs on Win7 Pro 64-bit.

gkalpak:
@robertmirro: Are you sure it "stalls". It usually takes a lot of time and CPU (and can give the impression that it's frozen), but its not.

robertmirro:
@gkalpak I just waited 5-ish minutes. That feels "stalled" to me, but I'm open to using a different word to describe this behavior.  :smile: 

With `parseext` **included**, `minall` never completes after `angular.min.js` is generated:
http://screencast.com/t/dhEVZizmxc25

With `parseext` **commented out**, `minall` completes and the next task (`collect-errors`) begins:
http://screencast.com/t/tD7StOaSML

gkalpak:
That's odd (unless it just needs more time on your machine - depending on the available resources).
The tasks are running fine on my Windows 10 :smiley:

If I were you I would (1) try giving it some more time and (2) removing `node_modules` and re-installing (I guess that's the "switch it off and back on" strategy in the Node.js world :stuck_out_tongue:)

robertmirro:
Being that I'm no stranger to anomalies such as this, I'll choose option (3)... Leaving `parseext` commented out while simultaneously attempting to forget that this issue probably only exists on my machine.  :stuck_out_tongue_winking_eye: 

Thanks though!

gkalpak:
Yeah that would work as well, I guess :smiley:

robertmirro:
@Narretz Hi.  Why was this PR closed with no merge?

PR #14355, opened 2016-03-31
fix($compile): ensure that `$onChanges` hook is called correctly

petebacondarwin:

fix

the `$onChanges` hook is not being called correctly

**Other information**:

Due to the way that we instantiate controllers, the `$onChanges` hook
was not always available at the time we were trying to trigger the initial
call to this hook. For instance, if the hook was actually defined inside
the constructor function.

This commit fixes that but also fixes the fact that the initial call was
being made in the postDigest anyway, which was incorrect because the
it should have been made before the `$onInit` call.

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -3668,17 +3668,25 @@ describe('$compile', function() {
           // Setup the directive with two bindings
           element = $compile('<c1 prop1="val" prop2="val2" other="val3" attr="{{val4}}"></c1>')($rootScope);
 
-          // There should be no changes initially
-          expect(log).toEqual([]);
+          expect(log).toEqual([
+            {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@mgol does this fix your problem too?

mgol:
I'll have a look tomorrow.

test/ng/compileSpec.js
```diff
@@ -3735,7 +3743,10 @@ describe('$compile', function() {
           $rootScope.$watch('a', function(val) { $rootScope.b = val * 2; });
 
           // There should be no changes initially
```
mgol: This comment no longer applies.

mgol:
After removing the one outdated comment, LGTM. :+1:

PR #14361, opened 2016-04-01
fix(injector): make sure we call Function.prototype.toString not an override

a510:

Bug fix

the injector uses fn.toString to extract function arguments

the injector uses Function.prototype.toString.call to avoid calling an override

No

**Other information**:

use Function.prototype.toString.call instead of fn.toString, because the user might override the toString function of the "fn" instance, without knowing it's being used by the injector.


lgalfaso:
Hi @a510, the change looks fine, but a test would be mostly welcome.

lgalfaso:
and this should be marked as `fix` not as `refactor`

PR #14363, opened 2016-04-02
chore(package.json): remove unused grunt-jasmine-node

thorn0:
At least, it looks unused.

### DIFF
--- a/npm-shrinkwrap.clean.json
+++ b/npm-shrinkwrap.clean.json
@@ -3943,10 +3943,6 @@
         }
       }
     },
-    "grunt-jasmine-node": {
-      "version": "0.1.0",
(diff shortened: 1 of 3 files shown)


lgalfaso:
`npm run test-i18n` ?

thorn0:
Does is use grunt?

gkalpak:
`npm run test-i18n` is using `jasmine-node` directly ([package.json#L20-L21][1]).

[1]: https://github.com/angular/angular.js/blob/bd0915c4007dcc5b0cae6968598731117c510ffa/package.json#L20-L21

Narretz:
If Travis passes, does that mean we can safely remove it?

Narretz:
Do we need to update the shrinkwrap for this, too?

gkalpak:
I think we do.

thorn0:
done

PR #14365, opened 2016-04-04
docs(guide/Components): add note to output events

machineloop:
Docs clarification and additional line added to example

Currently the example for using output events doesn't include an attribute binding example from a parent component/directive. Others are finding the lack of example confusing here:  http://stackoverflow.com/questions/35559354/the-in-angular-1-5-components/35560611

- Include line of code pulled from full example demonstrating Output event's use in parent component
- Add note clarifying camelCase to snake-case requirement to use the Output event callback feature on components

No

**Other information**:

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -133,8 +133,13 @@ components should follow a few simple conventions:
     For a deletion, that means the component doesn't delete the `hero` itself, but sends it back to
     the owner component via the correct event.
     ```html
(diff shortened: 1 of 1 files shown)


Narretz:
Thanks for the PR. I will add the additional code example and a less verbose HTML comment about snake case. This is nothing new for components, binding names have always been camelCase in the definition and snake case in templates. These are also not "real" events, but the usual & bindings which execute an expression in the parent (input) scope. This is different from angular2. (These bindings don't bubble up for example).

PR #14366, opened 2016-04-04
Fix(ng-animate-ref): copy contextual styling to clone

### LINKED ISSUE #12402: ng-animate-ref - Clone doesn't copy contextual styling of original
When an ng-animate-ref element is cloned before an animation, I would expect it to look exactly the same as its source element. Instead, when the source element has styling applied because of the specific context it is in (e.g. ```.context .source-element { background: green; }```), this styling is not applied to the cloned element.

This makes the animation weird: it first snaps to the styling the .source-element would have when in no context, and then animates to its new context. Example: http://plnkr.co/edit/AXzKmQcE5BUx4NXg4cCS?p=preview. When clicking an element, it first snaps to a blue background, before animating to a black background. I would expect it to animate directly from a green to a black background.

This can be worked around by simply applying all styling directly to the source element to be animated (e.g. ```.source-element-in-context { background: green; }``` (see this plunker: http://plnkr.co/edit/Odho9WnoOm9JFv3gzzZl?p=preview), but it makes much more sense to me when angular itself applies all styling of the source element to the cloned element.

NickBolles:
Fix(ng-animate-ref): copy contextual styling to clone

#12402 Elements "Snap" to non-context aware state before animating.

http://plnkr.co/edit/OLhhDaz8awp5wzLoeAr1?p=preview

The style context is now copied to the cloned element. 
http://plnkr.co/edit/EvBFwIGD40pHxewXCYGY?p=preview

No.

**Other information**:
I am unsure if tests should be created, or maybe more accurately how to create a test for this.
The same goes for Docs. Except that this seems to be the expected behavior. Maybe a note should be made in the https://docs.angularjs.org/api/ngAnimate#how-is-the-morphing-handled- paragraph though

Copy all contextual styling to the cloned element.
The list of properties that are copied is derived from
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties

Closes #12402

### DIFF
--- a/src/ngAnimate/animateCssDriver.js
+++ b/src/ngAnimate/animateCssDriver.js
@@ -9,12 +9,120 @@ var $$AnimateCssDriverProvider = ['$$animationProvider', function($$animationPro
   var NG_OUT_ANCHOR_CLASS_NAME = 'ng-anchor-out';
   var NG_IN_ANCHOR_CLASS_NAME = 'ng-anchor-in';
 
+
+  var ANIMATED_PROPS = [
+    'MozOutlineRadius',
+    'MozOutlineRadiusBottomleft',
+    'MozOutlineRadiusBottomright',
+    'MozOutlineRadiusTopleft',
+    'MozOutlineRadiusTopright',
+    'WebkitTextStroke',
+    'WebkitTextStrokeColor',
+    'WebkitTouchCallout',
+    'backdropFilter',
+    'background',
+    'backgroundColor',
+    'backgroundPosition',
+    'backgroundSize',
+    'border',
+    'borderBottom',
+    'borderBottomColor',
+    'borderBottomLeftRadius',
+    'borderBottomRightRadius',
+    'borderBottomWidth',
+    'borderColor',
+    'borderLeft',
+    'borderLeftColor',
+    'borderLeftWidth',
+    'borderRadius',
+    'borderRight',
+    'borderRightColor',
+    'borderRightWidth',
+    'borderTop',
+    'borderTopColor',
+    'borderTopLeftRadius',
+    'borderTopRightRadius',
+    'borderTopWidth',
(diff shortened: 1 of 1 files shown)


gkalpak:
I'm afraid copying over all those properties might be a little overkill.
That said, you can still use CSS to have the appropriate style. (It's just more boilerplate, right ?)

NickBolles:
What do you mean by "you can still use CSS to have the appropriate style"? 
The point of this PR is to not have to know the styles on the element and to avoid issues with contextual styles. 

Another example. ng-style causes this snapping behavior when going to or from ng-style.
http://plnkr.co/edit/4QQ1RhHOsQ0w6lLKrJRY?p=preview

Compared to the modified:
http://plnkr.co/edit/TIQR2lv0Wa3zAwqeEhy0?p=preview

gkalpak:
@NickBolles, I knwo what the point fo this PR is. What I'm saying is that this PR provides a convenience for something that is already possible using CSS (i.e. duplicate the styling for `.record[ng-animate-ref]` elements).

Obviously, having the contextual styles applied automatically is a big improvement, but there is also a significant cost which I am not sure is worth it. (I'm not saying it isn't, I'm just a bit sceptical.)

NickBolles:
@gkalpak gotcha. I agree that there is a pretty significant cost. 

Is there any way to conditionally do this. Such as the promise un wrapping that was removed? 

Or possible to have my own decorator to do this? Would that be the better route?

gkalpak:
I don't think there is a way to use your own implementation (unless you provide your own `$animate` service, which is obviously not realistic).

The more I think about it, the more I think the best route is to not have conditional styling.
As suggested in the docs, you should rely on simple classes for applying the styling for anchor elements.

gkalpak:
Maybe @matsko or @Narretz have any insights on this.

PR #14367, opened 2016-04-04
Pr 11547

### LINKED ISSUE #1934: Ability to react to progress events of $http XHR

petebacondarwin:

Closes #11547
Closes #1934

**Other information**:

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2010,6 +2010,17 @@ function MockXhr() {
   };
 
   this.abort = angular.noop;
+
+  this.$$events = {};
+  this.addEventListener = function(name, listener) {
+    if (angular.isUndefined(this.$$events[name])) this.$$events[name] = [];
+    this.$$events[name].push(listener);
+  };
+
+  this.upload = {
+    $$events: {},
+    addEventListener: this.addEventListener
+  };
 }
 
 
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1322,11 +1322,14 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
   }
 
   // TODO(vojta): change params to: method, url, data, headers, callback
-  function $httpBackend(method, url, data, callback, headers, timeout, withCredentials, responseType) {
+  function $httpBackend(method, url, data, callback, headers, timeout, withCredentials, responseType, eventHandlers, uploadEventHandlers) {
 
     var xhr = new MockXhr(),
         expectation = expectations[0],
         wasExpected = false;
+
+    xhr.$$events = eventHandlers;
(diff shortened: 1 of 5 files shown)


src/ngMock/angular-mocks.js
```diff
@@ -2006,6 +2009,17 @@ function MockXhr() {
   };
 
   this.abort = angular.noop;
+
+  this.$$events = {};
+  this.addEventListener = function(name, listener) {
+    if (angular.isUndefined(this.$$events[name])) this.$$events[name] = [];
+    this.$$events[name].push(listener);
+  };
```
gkalpak: I am kind of confused with this snippet:

When we create a `MockXhr`, we set `this.$$events = {}`.
Then inside the mock `$httpBackend`, right after we create `xhr = new MockXhr()`, we overwrite `xhr.$$events = eventHandlers` (where `eventHandlers` is of type `{[key: string]: function}`).
Then, when `addEventListener` is called on a `MockXhr` instance, we check create an array on `this.$$events[name]` and push the listener.

I would expect the tests to fail, but they obviously don't, so what do I miss ?

petebacondarwin: I agree this is not optimal.
The `MockXhr` is used in two different scenarios:
* In the `$httpBackend` tests we instantiate `MockXhr` objects and pass them through to the `$httpBackend` service, which does call `addEventListener`. This allows us to test the "real" use of the `$httpBackend` service.
* In the `$http` tests, we have a mock `$httpBackend` service (from `ngMock`), which happens to hold its own instance of `MockXhr`. In these tests we never call `addEventListener` and I just chose to use the same `$$events` object for holding our mock event handlers for the test

gkalpak: I see :confused:
<sub>(Clarification comments welcome :smile:</sub>

gkalpak:
LGTM

benjamingr:
I'd like to see more tests for this but overall nice work :)

petebacondarwin:
@benjamingr - what further tests would you like to see?

benjamingr:
Well, the only test right now is that events are passed in one way - there is a single test, I'd like to se more coverage.

I'm also not sure how users are supposed to mock it, that should probably be documented.

petebacondarwin:
@benjamingr could you be more explicit. What code paths are not being covered?

petebacondarwin:
@benjamingr Can you create a new issue for improving the documentation of mocking the events so that we can track it?

PR #14386, opened 2016-04-06
ngTranscludeLocals: assign values into transcluded scopes

drpicox:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**: it creates a new feature that allows to define ("_leak_") scope values in the transcluded templates

**What is the current behavior? (You can also link to an open issue here)**: you have to use `$parent` to access some values and it can be dangerous or complicated.

**What is the new behavior (if this is a feature change)?**: it adds an attribute to ng-transclude called ng-transclude-locals that allows to define values that can be setted to trancluded template.

  Example:

``` html
  <!-- HeroListComponent.tpl.html -->
  <div ng-repeat="hero in $ctrl.heros">
    <ng-transclude ng-transclude-locals="{hero: hero}"></ng-transclude>
  </div>

  <!-- OverviewHeroList.tpl.html -->
  <ul>
  <hero-list>
    <li>{{hero.name}}</li>
  </hero-list>
  </ul>
```

**Does this PR introduce a breaking change?**: introduce breaking changes or introduce the same breaking changes, or less, that the existence of an attribute called ng-transclude-slot does. Less because if it defines an object, it might overwrite locals. But I have prefixed it as ng-transclude so it is unlikely to have a colision.

**Other information**:

**Overview of the PR**: it allows to "_leak_" variables into the transcluded elements scopes:

  Example:

``` html
  <!-- HeroListComponent.tpl.html -->
  <div ng-repeat="hero in $ctrl.heros">
    <ng-transclude ng-transclude-locals="{hero: hero}"></ng-transclude>
  </div>

  <!-- OverviewHeroList.tpl.html -->
  <ul>
  <hero-list>
    <li>{{hero.name}}</li>
  </hero-list>
  </ul>
```

**Motivation of the PR**: although it can be done now using `$parent`:

``` html
  <!-- HeroListComponent.tpl.html -->
  <div ng-repeat="hero in $ctrl.heros">
    <ng-transclude></ng-transclude>
  </div>

  <!-- OverviewHeroList.tpl.html -->
  <ul>
  <hero-list>
    <li>{{$parent.hero.name}}</li>
  </hero-list>
  </ul>
```

It is not a recommended solution, because it can turn really complicated and we need to know when new scopes are created, ex:

``` html
  <!-- HeroListComponent.tpl.html -->
  <div ng-repeat="hero in $ctrl.heros">
    <ng-transclude></ng-transclude>
  </div>

  <!-- HeroStarsList.tpl.html -->
  <hero-list>
    <div ng-if="$parent.hero.stars">
      <nice-box>
        Stars: {{$parent.$parent.hero.stars}}
      </nice-box>
    </div>
  </hero-list>
```

or an alternative solution that can be achieved with ngInit (does not seems nice also):

``` html
<!-- HeroListComponent.tpl.html -->
<div ng-repeat="hero in $ctrl.heros">
  <ng-transclude></ng-transclude>
</div>

<!-- HeroStarsList.tpl.html -->
<hero-list>
  <span ng-init="stars = $parent.hero.stars"></span>
  <div ng-if="stars">
    <nice-box>
      Stars: {{stars}}
    </nice-box>
  </div>
</hero-list>
```

**Alternative**: use a basic variable, example `$locals`:

``` html
  <!-- HeroListComponent.tpl.html -->
  <div ng-repeat="hero in $ctrl.heros">
    <ng-transclude></ng-transclude>
  </div>

  <!-- HeroStarsList.tpl.html -->
  <hero-list>
    <div ng-if="$locals.hero.stars">
      <nice-box>
        Stars: {{$locals.hero.stars}}
        <!-- ¿what happens here? ^^^ -->
      </nice-box>
    </div>
  </hero-list>
```

But it still have expressivity issues similar to $parent.

I'm open to suggestions. Anyway, I think that this feature is very interesting, I do not like the idea of using `$parent` (to avoid it I have replicated some components, but first extracting common factor). 
I think that this feature will allow to create better components with lower coupling between information retrieval and information presentation (aka smart/dumb components).


Narretz:
Hi, I don't understand the use case for this. In all the tests the "local" values that are passed into the transcluded scope are bound to the directive. So all I'm seeing is:

```
scope: {
  param: '<'
}

...

$rootScope.status = 'whatever'

```html
<argument-directive param="{local: status}">{{local}}</argument-directive>
```

But why not do 
```html
<argument-directive param="{local: status}">{{status}}</argument-directive>
```

ngTranscluded content doesn't use the scope of the directive, that's the whole thing about it. So I don't see how this feature adds anything to transclusion. But maybe I don't understand the use case.

drpicox:
Hi @Narretz the idea of this PR is be able to add context to transcluded templates.

Normally, when you _transclude a template once there is no problem_, and everything is simple and straightforward as you have shown.
But when you _transclude a template twice problems start_, usually you need to tell to the transcluded template which number of transclusion, properties that might expect, etc...

I have proposed this PR because in the past I had already built alternate ngTransclude directives to handle this (including multi-transclussion), and in any other cases I have done or seen many workarounds to do similar things through more complex or _dangerous_ (`$parent`) approaches.

The idea is convert transclusion into a first order parameter for components, so at the end parameters available for components would be:
- _constants_ (@)
- variables ('<' or '=')
- _triggers_ (&)
- templates (transclusions)
Notice that **currently _triggers_ can be injected with `locals`** from the component, **the idea is to give the same flexibility to templates**. 

The applications that I had for these are from pagers or searchers up to highly configurable payment modals. Just imagine a payment screen with multiple packs available with a selector that shows different information depending the kind of product, country or even currency, in such cases, transcluded templates are one parameter more for the component. It tries to substitute `$compile` calls required in such case by just the transclusion mechanism existing in angular.

If you want to see a more detailed example I have wrote the following plnkr:

- https://plnkr.co/edit/gu4DbH0v65eEDMB2Bs47?p=preview

The code is the following:

```html
  <h3>Brief</h3>
  <item-list-repeat><item-brief item="$parent.item"></item-brief></item-list-repeat>
  <h3>Detail</h3>
  <item-list-repeat><item-detail item="$parent.item"></item-detail></item-list-repeat>
```

```javascript
   .component('itemListRepeat', {
     transclude: true,
     template: ''+
       '<div>Page: {{$ctrl.page}}</span>'+
       '<div ng-repeat="item in $ctrl.items | limitTo:$ctrl.limit:$ctrl.begin track by item.id">'+
       '<ng-transclude></ng-transclude>'+
       '</div>'+
       '<div><button ng-click="$ctrl.next()">Next</button></div>'+
       '',
     controller: function(items) {
       this.items = items;
       this.page = 1;
       this.limit = 2;
       this.begin = 0;
       this.next = function() {
         this.begin = this.begin + this.limit;
         if (this.begin >= items.length) {
           this.begin = 0;
         }
         this.page = Math.floor(this.begin / this.limit) + 1;
       };
     }
   })
   .component('itemBrief', {
     bindings: { item: '<' },
     template: '<div class="item-brief">Item: {{$ctrl.item.id}}</div>'
   })
   .component('itemDetail', {
     bindings: { item: '<' },
     template: ''+
       '<h4><item-brief item="$ctrl.item"></h4>' +
       '<ul>'+
       '<li>Short: {{$ctrl.item.n}}</li>' +
       '<li>Long: {{$ctrl.item.t}}</li>' +
       '/<ul>'+
       ''
   })
```

About tests, my real target is the following:

```javascript
    it('should be compatible with ngRepeat', function() {
      element = $compile('<list-component param="\'ctx\'">[{{position}}] {{item}} ({{param}})</list-component>')($rootScope);

      $rootScope.$apply();
      expect(element.text()).toBe('[1] first (ctx)[2] second (ctx)[3] third (ctx)');
    });
```

I'm sorry to not have been more clear in the initial comment.

drpicox:
Trying to replicate a real example:

```html
<!-- A modal to buy gold coins -->
<payment-modal packs="$ctrl.goldPacks">
  <payment-modal-pack>
     <gold-pack pack="pack" payment-method="selectedPaymentMethod"></gold-pack-cost>
  </payment-modal-pack>
  <payment-modal-terms>
    <premium-terms pack="selectedPack" payment-method="selectedPaymentMethod"></premium-cost>
  <payment-modal-terms>
  <payment-modal-extra>
    <gold-autorecharge extras="extras"></gold-autorecharge>
  </payment-modal-extra>
</payment-modal>
```

```html
<!-- A modal to buy premium membership -->
<payment-modal  packs="$ctrl.premiumPacks">
  <payment-modal-pack>
     <premium-pack pack="pack" payment-method="selectedPaymentMethod"></premium-pack-cost>
  </payment-modal-pack>
  <payment-modal-terms>
    <premium-terms pack="selectedPack" payment-method="selectedPaymentMethod"></premium-cost>
  <payment-modal-terms>
</payment-modal>
```

petebacondarwin:
I think what we are talking about here is a kind of container directive, where the content of the element on which the directive lives, contains a "template" to be used inside the directive's template to modify the look of the template, but which is actually in the context of the directive's inner scope rather than the outer scope.

Is that right or something more subtle?

drpicox:
Yes, it is almost as you say.

The problem that I have is that inner templates needs to access to _both contexts_:
- outer context because might contain configurations & other values
- inner context because it provides to you about container data (like current repeated element)

To solve this duality, I have used ng-transclude as source of controlled inner context leaking: it does not leaks everything, just what the container component expects to share. So in some way it becomes one more interface element (just like bindings).

**Alternative**:

With multislot transclusion I can try to make it more explicit (I have to work with the notation):

```html
<payment-modal  packs="$ctrl.premiumPacks">
  <payment-modal-pack bind-innerscope="pack">
     <premium-pack pack="pack.pack" payment-method="pack.selectedPaymentMethod"></premium-pack>
  </payment-modal-pack>
  ...
</payment-modal>
```

Without multi-slot it should be a notation on the container directive itself, which might be confusing or introduce breaking changes. 
In this alternative probably it would only apply on multi-transclude (although only one transclusion could being used).

petebacondarwin:
I feel that this could be an interesting solution...
Is there anything similar in Angular 2?

drpicox:
I have been looking for it but I didn't found anything similar. 
There is a very little information about how how <ng-content> works in Angular2 and I also have very little experience in ng2 (yet). It should be nice to have a solution that will be able to work in both sides.

lgalfaso:
Hi, would like to start saying that I see some value in adding a functionality that will allow a transclusion slot to have some local bindings. That said, I do not like the syntax nor the implementations :(

I find this syntax nicer

```html
<!-- HeroListComponent.tpl.html -->
<div ng-repeat="oneHero in $ctrl.heros">
  <ng-transclude hero="oneHero"></ng-transclude>
</div>

<!-- OverviewHeroList.tpl.html -->
<ul>
  <hero-list>
    <li>{{hero.name}}</li>
  </hero-list>
</ul>
```

but for this to work, there is a need to specify the binding on this slot somehow in the directive, that implies that the directive definition would need to be like
```js
.directive('heroList', function(){
  return {
    transclude: true,
    transcludeBinding: {
      hero: '='
    },
    [...]
  };
});
```

The reasons being
* With the original approach, the developer using the directive `heroList` will need to look at the directive template to know that there will be a local named `hero`. With the new approach, the developer will have to look for `transcludeBinding` in the directive definition. With the later approach is easier to digest and to automatically generate documentation 
* The orignal approach makes the assumption that one binding type fits all

drpicox:
I like the proposal, but I do not know how it will work with multi-transclusion in which each slot has different number of parameters. 

In fact the proposal that I have done values must be ready before execute the transclusion, which can be a problem.

I'll try to find an alternative proposal in the next days that at least would be compatible with '<' and '&'.

petebacondarwin:
How about a more simplistic solution? See #14799

drpicox:
Attaching $ctrl to the current scope is already covered by #14080 and it is not enough: if a transclude section is repeated we want access to variables like $index which cannot be fixed inside the controller.

Because of this it is not necessary '&' bindings, $ctrl functions are accessible.

And I do not have an alternative to @lgalfaso proposal because I do not like pollute the directive definition object and because each transclusion part may have its own requirements. 
(also because I've been too busy)

petebacondarwin:
How about we expose a controller `NgTranscludeController` on the `ngTransclude` directive, which has a property called `$locals`, which are exposed onto the transcluded scope?
Then people can create their own directives that "require" this controller and attach their own data...
This is similar to the situation we have for validators, parsers and serializers for `NgModelController`.

gkalpak:
Would it be possible to expose a property from the scope in order to avoid the overhead of a controller?

petebacondarwin:
@gkalpak can you elaborate?

gkalpak:
Something like [this][1]. The only difference with the current `ngTransclude` implementation is:

```diff
 function ngTranscludeCloneAttachFn(clone, transcludedScope) {
   if (clone.length) {
     $element.append(clone);
+    if ($scope.$transcludeLocals) {
+    // OR:
+    // if ($scope.hasOwnProperty('$transcludeLocals')) {
+      transcludedScope.$transcludeLocals = $scope.$transcludeLocals;
     }
   } else {
     ...
```

[1]: https://codepen.io/ExpertSystem/pen/qNaMmJ?editors=0011

petebacondarwin:
We can actually achieve exactly the behaviour of this PR without changing core at all.
Simply adding in a new directive `ngTranscludeLocals`, which admittedly has to make the slightly naughty use of `scope.$$childHead`:

http://plnkr.co/edit/pxEIDXxOnuutXgGkvyYl?p=preview

```js
.directive('ngTranscludeLocals', function() {
  return function link(scope, element, attr) {
    scope.$watchCollection(attr.ngTranscludeLocals, function(newLocals, oldLocals) {
      var transcludeScope = scope.$$childHead;
      if (oldLocals && (newLocals !== oldLocals)) {
        angular.forEach(oldLocals, function(val, local) { delete transcludeScope[local]; });
      }
      if (newLocals) angular.forEach(newLocals, function(val, local) { transcludeScope[local] = val; });
   });
  };
});
```

drpicox:
I like the idea to put the functionality in its own directive. Using $$childHead may be is a little bit ugly, but if it inside the "core", I think that it is ok and better to have it inside the transclude directive.

If you let me I'll modify the PR in that way. I hope to have it this Friday.

About `$transcludeLocals` option I think that should not be enough because some of the locals are `let` fashion and living only inside ngFor regions (for example, `$index`)

drpicox:
- I have moved it into its own directive as suggested at https://github.com/angular/angular.js/pull/14386#issuecomment-227508851

- After lot of thinking binding of kind of '&' and others are not necessary, with #14080 we can already have access to the controller of the directive

- It allows to pass different sets of values to each transclusion template as they are needed (a header transclusion, an element transclusion, ...)

- It does not adds more configuration into the directive DDO, it should require any mechanism to match each set of "bindings" to each transclusion instance inside the directive template

So... now it lgtm

src/ng/directive/ngTranscludeLocals.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * @ngdoc directive
+ * @name ngTranscludeLocals
+ * @restrict A
+ *
+ * @description
+ * Directive that allows to define local variables for a given transclusion scope.
+ *
+ * It works with {@link ng.directive:ngTransclude ngTransclude} and it
+ * expects an object which define the local varibles for the transcluded template.
```
gkalpak: _define_  -->  _defines_
__varibles_  -->  _variables_

src/ng/directive/ngTranscludeLocals.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @element ng-transclude
```
gkalpak: I don't think this is accurate. It can be on any element (as long as it has `ngTransclude` on it).

src/ng/directive/ngTranscludeLocals.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {expression} ngTranscludeLocals an object with pairs of key-value
+ *                                        in which the key is the variable name inside the template
+ *                                        and the value is the value for those variables
+ *
+ * @example
+ * ### Transclude with locals
+ * This example demonstrates basic transclusion of local context into a transcluded template.
+ * <example name="simpleTranscludeExample" module="transcludeLocalsExample">
+ *   <file name="index.html">
+ *     <script>
```
gkalpak: Why not split it out into a `script.js` file?

src/ng/directive/ngTranscludeLocals.js
```diff
@@ -0,0 +1,70 @@
...
+ * Directive that allows to define local variables for a given transclusion scope.
+ *
+ * It works with {@link ng.directive:ngTransclude ngTransclude} and it
+ * expects an object which define the local varibles for the transcluded template.
+ *
+ * @element ng-transclude
+ *
+ * @param {expression} ngTranscludeLocals an object with pairs of key-value
+ *                                        in which the key is the variable name inside the template
+ *                                        and the value is the value for those variables
```
gkalpak: I think we should be more explicit about `value` being an expression.

src/ng/directive/ngTranscludeLocals.js
```diff
@@ -0,0 +1,70 @@
...
+ *           template: '<div ng-repeat="hero in $ctrl.heros">' +
+ *                       '<ng-transclude ng-transclude-locals="{hero: hero, position:$index+1}"></ng-transclude>' +
+ *                     '</div>',
+ *           controller: function() {
+ *             this.heros = ['Superman','Batman','Spawn'];
+ *           }
+ *         });
+ *     </script>
+ *     <div>
+ *       <ul>
```
gkalpak: It makes more sense to move the `<ul>` inside `heroList`'s template.

src/ng/directive/ngTranscludeLocals.js
```diff
@@ -0,0 +1,70 @@
...
+ * </example>
+
+ */
+var ngTranscludeLocalsMinErr = minErr('ngTranscludeLocals');
+var ngTranscludeLocalsDirective = function() {
+  return {
+    restrict: 'A',
+    link: function ngTranscludeLocalsLink($scope, $element, $attrs) {
+      $scope.$watchCollection($attrs.ngTranscludeLocals, function(newLocals, oldLocals) {
+        var transcludeScope = $scope.$$childHead;
```
gkalpak: Is it possible that this is undefined/null?

test/ng/directive/ngTranscludeLocalsSpec.js
```diff
@@ -0,0 +1,80 @@
...
+      $rootScope.$apply();
+      expect(element.text()).toBe('-2');
+
+      delete $rootScope.status;
+      $rootScope.$apply();
+      expect(element.text()).toBe('-');
+    });
+
+    it('should be compatible with ngRepeat', function() {
+      element = $compile('<transclude-list-items-component param="\'ctx\'">[{{position}}] {{item}} ({{param}})</transclude-list-items-component>')($rootScope);
```
gkalpak: <sub>Please wrap all lines at 100 chars max if possible :grin:</sub>

test/ng/directive/ngTranscludeLocalsSpec.js
```diff
@@ -0,0 +1,80 @@
...
+
+      $rootScope.$apply();
+      expect(element.text()).toBe('[1] first (ctx)[2] second (ctx)[3] third (ctx)');
+    });
+
+});
```
gkalpak: I think we need more tests. E.g. what happens with multi-slot transclusion? What about optional, empty slots?

src/ng/directive/ngTranscludeLocals.js
drpicox: That's right, good point.

src/ng/directive/ngTranscludeLocals.js
drpicox: I have changed it to "Object"

src/ng/directive/ngTranscludeLocals.js
drpicox: Mainly because I copy pasted the example from other directive.
But, I was trying to make easier to understand, but I think that did not.

src/ng/directive/ngTranscludeLocals.js
drpicox: Uhmmm that means that the example is not good enough.

src/ng/directive/ngTranscludeLocals.js
drpicox: Right, if not used with a scope that has no childs, it fails.
Adding test and solution (ignore and do nothing).

test/ng/directive/ngTranscludeLocalsSpec.js
drpicox: I respond below

drpicox:
Multi-slot transclusion fails because the $scope is shared for all of the template, and $$childHead is pointing to the first private scope (it does not matter if it is a transclusion or not).

So what I did is to add a scope:true to the directive, and now it works in all cases.

Unfortunately this is not the most elegant solution either. Ideally I should use element.isolateScope() which gives the real child scope, but is not available when debugInfo is disabled.
May be should be better return to the original implementation inside ngTransclude.

As side effect of the current implementation, something that I do not like, if you have a directive with isolated scope, you can mess with it and place values. But I'm don't think that it could be a big deal.

petebacondarwin:
@mhevery is not keen on this change. And so far we haven't found a reasonable analogy in Angular 2.
This means that we will probably have to put this PR on hold for now.

drpicox:
It is ok. Usually I (and many other people) fix it using templates as strings and compiled with $compile, it is not elegant but works.
About NG2 I need more practice just to know if it can solve this problem with any other mechanism.

¿Did you have seen the example that I have added? 
I have found the way to make explicit the name of the locals to publish in the transcluded template:

```
<example name="simpleTranscludeExample" module="transcludeLocalsExample">
  <file name="index.html">
    <aside>
      <b>Heros</b>:
      <foreach-hero hero-as="hero">
        {{hero}}
      </foreach-hero>
    </aside>
    <div class="card-container">
      <foreach-hero hero-as="hero" position-as="rank">
        <div class="card">
          <h3>#{{rank}}</h3>
          <h5>{{hero}}</h5>
        </div>
      </foreach-hero>
    </div>
  </file>
  <file name="example.js">
    angular.module('transcludeLocalsExample', [])
      .component('foreachHero', {
        transclude: true,
        bindings: { heroAs:'@', positionAs:'@', },
        template: '<span ng-repeat="hero in $ctrl.heros">' +
                    '<ng-transclude ng-transclude-locals="{'+
                        '[$ctrl.heroAs]: hero, '+
                        '[$ctrl.positionAs]: $index+1 '+
                     '}"></ng-transclude>' +
                  '</span>',
        controller: function() {
          this.heros = ['Superman','Batman','Spawn'];
        }
      });
  </file>
  <file name="style.css">
    aside {
      padding: 10px;
      background: #fefefe;
    }
    .card {
      float: left;
      border-radius: 5px;
      background: #eee;
      padding: 0 15px;
      margin: 6px;
      min-width: 8em;
    }
  </file>
  <file name="protractor.js" type="protractor">
    it('should have transcluded all repeated elements with different local', function() {
      var heros = element.all(by.repeater('hero in $ctrl.heros'));
      expect(heros.get(0).getText()).toEqual('Superman');
      expect(heros.get(1).getText()).toEqual('Batman');
      expect(heros.get(2).getText()).toEqual('Spawn');
      expect(heros.get(3).getText()).toEqual('#1\nSuperman');
      expect(heros.get(4).getText()).toEqual('#2\nBatman');
      expect(heros.get(5).getText()).toEqual('#3\nSpawn');
     });
  </file>
</example>
```

petebacondarwin:
@ashclarke - exactly. We do not promise to keep private API from changing without notice. So if you do use such an API then it might suddenly break when you do an update.

ashclarke:
@petebacondarwin - Thanks for the clarification!

petebacondarwin:
Closing as a won't fix. Thanks @drpicox for your work on this.

PR #14400, opened 2016-04-09
perf(ngOptions): use documentFragment to populate select

### LINKED ISSUE #12076: ngOptions slow in IE with large arrays
I have a 500+ element array which I want to bind to a `<select>` element.
In Chrome, this works fine, adding the array takes a fraction of a second.
In IE, it takes multiple seconds to add the array, during which time the UI is unresponsive.

Looking at timelines, it appears that Chrome recalculates styles after all `<option>` elements are added, but IE calculates styles and layouts after each `<option>` is added.

Chrome:
Chrome Timeline
IE:
IE Timeline

Maybe using `documentFragments` could fix this? 

Plunker: http://plnkr.co/edit/ULtEHtcD5La4Ax7hd5kY?p=preview

### LINKED ISSUE #13607: <select> with multiple becomes unresponsible after changing contents of ngOptions list in IE11
There seems to be a couple of issues listed for similiar issues but I just updated to 1.4.8 and still have this issue.

I have two of these in a directive to allow users to move items from one list to the other (both lists are subsets of a bigger list of items, and are filtered so no items can be in both lists).
```
<select id="multiSelectAvailable" 
ng-model="selected.available" 
multiple="multiple" 
size="{{available.length}}" 
ng-options="e as e[displayAttr] for e in available">
</select>

<select id="multiSelectSelected" 
ng-model="selected.current" 
multiple="multiple" 
size="{{model.length}}" 
ng-options="e as e[displayAttr] for e in model">
</select>
```

Sometimes when setting the list of items one or both of the selects becomes unresponsive. After changing the contents a couple of times it starts working again. However it could just be a coincidence.

** Edit **
Removed the comment about Swedish characters as that seems to not have an impact.

Narretz:
perf/fix

- ngoptions is slow in ie with many options
- strange freezes in ie in some cases

a tiny one possibly - now the select element is completely emptied before the options are appended

**Other information**:

This changes the way option elements are generated when the ngOption collection changes.
Previously, we would re-use option elements when possible (updating their text and
label). Now, we first remove all currently displayed options and the create new options for the
collection. The new options are first appended to a documentFragment, which is in the end appended
to the selectElement.

Using documentFragment improves render performance in IE with large option collections
(> 100 elements) considerably.

Always creating new options fixes issues in IE where the select would become unresponsive to user
input.

Fixes #13607
Fixes #12076

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -245,7 +245,7 @@ var NG_OPTIONS_REGEXP = /^\s*([\s\S]+?)(?:\s+as\s+([\s\S]+?))?(?:\s+group\s+by\s
 // jshint maxlen: 100
 
 
-var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
+var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, $document, $parse) {
 
   function parseOptionsExpression(optionsExp, selectElement, scope) {
 
@@ -432,7 +432,10 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
 
       var options;
       var ngOptions = parseOptionsExpression(attr.ngOptions, selectElement, scope);
-
+      // This stores the newly created options before they are appended to the select.
+      // Since the contents are removed from the fragment when it is appended,
+      // we only need to create it once.
+      var listFragment = $document[0].createDocumentFragment();
 
       var renderEmptyOption = function() {
         if (!providedEmptyOption) {
@@ -581,6 +584,8 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
         emptyOption = jqLite(optionTemplate.cloneNode(false));
       }
 
+      selectElement.empty();
+
       // We need to do this here to ensure that the options object is defined
       // when we first hit it in writeNgOptionsValue
       updateOptions();
@@ -589,6 +594,12 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
       scope.$watchCollection(ngOptions.getWatchables, updateOptions);
 
       // ------------------------------------------------------------------ //
+
+      function addOptionElement(option, parent) {
+        var optionElement = optionTemplate.cloneNode(false);
+        parent.appendChild(optionElement);
+        updateOptionElement(option, optionElement);
+      }
 
 
       function updateOptionElement(option, element) {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I checked this with the ngOptions benchmark the project and there is no discernable difference in speed on Chrome, Firefox or Safarin on OS/X. So I am happy to go with using this much simpler (and less buggy when it comes to IE) algorithm.

petebacondarwin:
@Narretz please merge into master and 1.5.x

src/ng/directive/ngOptions.js
```diff
@@ -606,73 +608,35 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
-
-      function updateOptions() {
-
-        var previousValue = options && selectCtrl.readValue();
 
         options = ngOptions.getOptions();
 
         var groupMap = {};
-        var currentElement = selectElement[0].firstChild;
+        var listFragment = $document[0].createDocumentFragment();
```
Narretz: Just noticed that it should be possible to create the documentFragment only once per ngOptions, as the fragment contents are removed once it has been appended to a container

src/ng/directive/ngOptions.js
```diff
@@ -606,73 +608,35 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+        // We must remove all current options, but cannot simply set innerHTML = null
+        // since the providedOption might have an ngIf on it that inserts comments which we must
+        // preserve
+        // Instead iterate over the current option elements and remove them or their optgroup
+        // parents
+        if (options) {
+          for (var i = options.items.length - 1; i >= 0; i--) {
+            var option = options.items[i];
+            if (option.group) {
+              jqLiteRemove(option.element.parentNode);
```
gkalpak: So, if there is a group, we don't need to preserve ngIf comments ?
How come ?

src/ng/directive/ngOptions.js
```diff
@@ -686,53 +650,32 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
+              listFragment.appendChild(groupElement);
 
               // Update the label on the group element
               groupElement.label = option.group;
 
               // Store it for use later
               group = groupMap[option.group] = {
-                groupElement: groupElement,
-                currentOptionElement: groupElement.firstChild
+                groupElement: groupElement
```
gkalpak: Minor: Since now this object only holds `groupElement`, we can avoid creating a new object and j7st store the `groupElement` directly.

src/ng/directive/ngOptions.js
```diff
@@ -686,53 +650,32 @@ var ngOptionsDirective = ['$compile', '$parse', function($compile, $parse) {
...
-            group.currentOptionElement = optionElement.nextSibling;
 
           } else {
 
-            // This option is not in a group
-            optionElement = addOrReuseElement(selectElement[0],
-                                              currentElement,
-                                              'option',
-                                              optionTemplate);
+            optionElement = optionTemplate.cloneNode(false);
```
gkalpak: Nitpick: this path is almost identical with above (only the parentElement changes). We could factor it out of the `if-else` block.

src/ng/directive/ngOptions.js
Narretz: Not sure if I understand how this could happen. ngOption generated options can't have ngIf, and empty / unknown options are not part the options.items. So whatever we remove here will not touch any comments.

gkalpak: Also minor, but it seems that we will be calling `jqLiteRemove()` multiple times on the same optgroup element. Maybe we should avoid that.

gkalpak: I see about the ngIf. So badically only the empty option can have an ngIf. Makes sense. 
There is a typo in the comment: `providedOption`  -->  `providedEmptyOption` :smiley:

gkalpak:
So, except for a couple of minor (ignorable) comments it LGTM.
I wonder if it might break people's test (the same way it broke ours).

src/ng/directive/ngOptions.js
Narretz: It doesn't lookt like the muliple jqLiteRemove affects performance too much, so I would like to keep it like that for simplicity

Narretz:
Wrt to the tests - maybe. But the test was only working before because it relied on the implementation (option elements are reused). And tests shouldn't do that.

PR #14401, opened 2016-04-09
chore(travis): test on IE Edge

### LINKED ISSUE #13687: Run travis tests on Microsoft Edge
- [x] update to karma 0.13.x (Edge poses as webkit browser, confusing karma)
- [ ] ~~install a launcher for Edge? See issue: https://github.com/karma-runner/karma-ie-launcher~~/issues/35
- [ ] update karma conf
- [ ] update travis build.sh

Narretz:
Closes #13687

### DIFF
--- a/karma-shared.conf.js
+++ b/karma-shared.conf.js
@@ -74,6 +74,12 @@ module.exports = function(config, specificOptions) {
         platform: 'Windows 8.1',
         version: '11'
       },
+      'SL_EDGE': {
+        base: 'SauceLabs',
+        browserName: 'microsoftedge',
+        platform: 'Windows 10',
+        version: '14'
+      },
       'SL_iOS': {
         base: 'SauceLabs',
         browserName: 'iphone',
@@ -119,6 +125,13 @@ module.exports = function(config, specificOptions) {
         browser_version: '11.0',
         os: 'Windows',
         os_version: '8.1'
+      },
+      'BS_EDGE': {
+        base: 'BrowserStack',
+        browser: 'edge',
+        browser_version: '14',
+        os: 'Windows',
+        os_version: '10'
       },
       'BS_iOS_8': {
         base: 'BrowserStack',
--- a/test/ng/directive/inputSpec.js
+++ b/test/ng/directive/inputSpec.js
@@ -5,6 +5,9 @@
 describe('input', function() {
   var helper = {}, $compile, $rootScope, $browser, $sniffer, $timeout, $q;
 
+  // UA sniffing to exclude Edge from some date input tests
+  var isEdge = /\bEdge\//.test(window.navigator.userAgent);
+
   generateInputCompilerHelper(helper);
 
   beforeEach(inject(function(_$compile_, _$rootScope_, _$browser_, _$sniffer_, _$timeout_, _$q_) {
@@ -688,18 +691,20 @@ describe('input', function() {
       expect($rootScope.form.alias.$error.month).toBeTruthy();
     });
 
-    it('should allow four or more digits in year', function() {
-      var inputElm = helper.compileInput('<input type="month" ng-model="value"  ng-model-options="{timezone: \'UTC\'}"/>');
 
-      helper.changeInputValueTo('10123-03');
-      expect(+$rootScope.value).toBe(Date.UTC(10123, 2, 1, 0, 0, 0));
+    if (!isEdge) {
+      it('should allow four or more digits in year', function() {
+        var inputElm = helper.compileInput('<input type="month" ng-model="value"  ng-model-options="{timezone: \'UTC\'}"/>');
 
-      $rootScope.$apply(function() {
-        $rootScope.value = new Date(Date.UTC(20456, 3, 1, 0, 0, 0));
+        helper.changeInputValueTo('10123-03');
+        expect(+$rootScope.value).toBe(Date.UTC(10123, 2, 1, 0, 0, 0));
+
+        $rootScope.$apply(function() {
+          $rootScope.value = new Date(Date.UTC(20456, 3, 1, 0, 0, 0));
+        });
+        expect(inputElm.val()).toBe('20456-04');
       });
-      expect(inputElm.val()).toBe('20456-04');
-    });
-
+    }
 
     it('should only change the month of a bound date', function() {
       var inputElm = helper.compileInput('<input type="month" ng-model="value" ng-model-options="{timezone: \'UTC\'}" />');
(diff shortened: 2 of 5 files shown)


Narretz:
Edge times out on one single test in the whole suite! And it makes no sense why exactly this test. There are also some input and ngTouch failures( the latter actually abort the test run completely ...)

jdalton:
Hey @Narretz! 

Just a heads up this test/sniff should be revisited on the next Edge release.

We recently ran into this [onboarding Angular into our Edge regression testing](https://channel9.msdn.com/events/Build/2016/P513) and filed an internal bug.
The good news is it was [patched recently](https://github.com/Microsoft/ChakraCore/pull/752).

test/ng/directive/inputSpec.js
```diff
@@ -657,17 +657,20 @@ describe('input', function() {
       expect($rootScope.form.alias.$error.month).toBeTruthy();
     });
 
-    it('should allow four or more digits in year', function() {
-      var inputElm = helper.compileInput('<input type="month" ng-model="value"  ng-model-options="{timezone: \'UTC\'}"/>');
 
-      helper.changeInputValueTo('10123-03');
```
AmazingJaze: @Narretz when you were preparing this change, were you able to determine the root cause of these inputSpec failures?  I've been looking at these failures recently as part of the work @jdalton mentioned to include AngularJS tests in our Edge Regression testing.  I have concluded that this line `helper.changeInputValueTo('10123-03');` is failing to set the `.value` property on the HTML element in Edge, which is what results in the eventual failure, but I had to stop before getting to "why" the value property isn't getting set. 

If there is an Edge bug behind this I'd love to make sure it gets filed.

Narretz: @AmazingJaze I assumed that Edge simply doesn't allow setting years with more than 4 digits in all the date input elements, so when you enter `10123-03` etc. the input has no value (because it's invalid) and we don't get a value to work with. But I haven't actually checked if this is an Edge bug.

gkalpak: It is indeed the case that Edge does not support dates with years past 9999.

The spec does explicitly allow years with more than 4 digits (but not all browsers support it).
See #13735 for some context.

Narretz: Looking at this plnkr, http://plnkr.co/edit/TkRo7GD139Pab0qC5ZEm?p=preview it seems to be the case that there is no way to enter a year with more than 4 digits in Edge currently. In Chrome, I can always use the keyboard to input the parts of the date, and so I can also enter 5 digit dates. In Edge it looks like I can only select a date from the picker. Is this correct? This seems like a UX issue to me.

Narretz:
Hi @jdalton thanks for the heads-up! Do you know when a new stable version is scheduled for release?

test/ng/directive/inputSpec.js
gkalpak: AFAIK, it is indeed not possible to set the date without using the picker (unless you open up dev tools and set the `value` property directly). I guess it is a conscious choice (albeit one I'm not a big fan of :stuck_out_tongue:).

jdalton:
> Do you know when a new stable version is scheduled for release?

Not at the moment. I don't believe the patch has made it into our insider builds yet.

test/ng/directive/inputSpec.js
AmazingJaze: Thanks for the tips @Narretz and @gkalpak. I will follow up with the EdgeHTML folks internally about both of these issues. Cheers :beers:

jdalton:
@Narretz just to confirm the timeout is the same issue we patched do you have a screenshot of the fail/hang.

Narretz:
@jdalton Here's the console output from such a failed test: https://s3.amazonaws.com/archive.travis-ci.org/jobs/121916115/log.txt (Search for "heartbeat")

AmazingJaze:
@Narretz did you encounter any trouble getting [$$cookieReader](https://github.com/angular/angular.js/blob/master/test/ng/cookieReaderSpec.js) and [$$cookieWriter](https://github.com/angular/angular.js/blob/master/test/ngCookies/cookieWriterSpec.js) tests to successfully pass on Edge?

When I try to run Angular.js tests in Edge, those tests would constantly fail for me but I was able to get them working with a small change to the test code. Should I send a PR?

Narretz:
@AmazingJaze No, there weren't any problems with the cookie services. But I was only able to test on microsoftedge 25.10586 (in Saucelabs) so maybe it's a problem introduced in later versions. You can open a PR and I'll take a look!

mgol:
@Narretz there is no "IE Edge", it's just "Edge" or "Microsoft Edge"; commit messages should be updated.

Have you tried it again recently? Are there any further problems?

Narretz:
Here's the last test that is currently failing in Edge which I don't know how to handle: https://github.com/angular/angular.js/blob/136a42abc162a64ac57ff45417553b58a031990b/test/AngularSpec.js#L1776
It concerns the auto-bootstrap prevention we built in.
The problem is that Edge doesn't set the `origin` property on `a` elements, at least not in Microsoft Edge 38.14393.0.0, EdgeHTML 14.14393.
MDN indicates that `origin` should be available, but it's unclear since which version: https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/origin (check history of this page)

I assume we can try to test again when the Creators Update goes live later in April.

jdalton:
@Narretz I don't believe that will be resolved with the Creators update since it's not resolved in the insider preview. I did find that this worked though:
```js
new URL(someAnchorElement).origin
```

Narretz:
Hi @jdalton , thanks for looking into this. Unfortunately in my local version, `new URL(someAnchorElement).origin` doesn't work for our use case, it returns `null`.
We need to origin to decide if we should allow auto-bootstrap of an AngularJS app in an extension context.

```
    link.href = src.value // this is something like "ms-browser-extension://something";
    var url = new URL(link);
    // url.origin is null
```

So in short, if the link has the protocol `ms-browser-extension:`, `origin` returns `null`. It should return `ms-browser-extension://something` (I've also tested with all other vendor prefixed proctocols and `browserext:`, none work in Edge)

Should I open an issue for that?

And regarding the `origin` prop on `a` elements, should I open a ticket too?

jdalton:
> And regarding the origin prop on a elements, should I open a ticket too?

I opened an internal issue for that, but an external to link to wouldn't hurt.

> Should I open an issue for that?

Yes, for the `new URL(...).origin` not returning correctly for ms-extensions.

Regarding your detection. Isn't the protocol something you can use to as well when deciding to allow auto-bootstrapping?

Narretz:
It seems like returning "null" from the scheme `ms-browser-extension` is in line with the spec: https://url.spec.whatwg.org/#concept-url-origin. Chrome and Firefox currently return `ms-browser-extension://` and would then be in violation of the spec. It looks like they handle any non-special scheme as they do for `http` etc, whereas Edge treats everything like "Otherwise" except when it's special.

petebacondarwin:
We need to check the origin because we want to allow autobootstrap within an extension only if the extension is loading the angular.js script from the same extension. It is not enough to know that we are in an extension. If we cannot identify the origin then cannot assert that we are not crossing from one extension to another.

The simple workaround (and best practice) is to always manually bootstrap AngularJS within an extension. This will always work on all browsers.

Until we can find a way to identify that the angular.js script is being loading into the same extension from which it is being sourced, we will not be able to allow auto bootstrapping within MS Edge extensions.

petebacondarwin (approved):
LGTM

mgol (changes requested):

mgol:
Remember to change the commit message when landing! (you could also update the title of this PR to say just "Edge" instead of "IE Edge").

petebacondarwin:
Feel free to follow @mgol's comment and merge this into master @Narretz

karma-shared.conf.js
```diff
@@ -74,6 +74,12 @@ module.exports = function(config, specificOptions) {
         platform: 'Windows 8.1',
         version: '11'
       },
+      'SL_EDGE': {
+        base: 'SauceLabs',
+        browserName: 'microsoftedge',
+        platform: 'Windows 10',
+        version: '14'
+      },
```
Narretz: Added

PR #14402, opened 2016-04-09
fix($compile): do not use `noop()` as controller for multiple components

### LINKED ISSUE #14391: Potential bug in registerComponent
### Description
Bug exists on angular 1.5.3 and even on master branch. 

In compile.js at the beginning of registerComponent (angular.component) method there is a assignment:

`var controller = options.controller || noop;`

Some lines later (after factory function definition) the `controller` variable is modified by assigning component options starting with `$` sign. 
```
forEach(options, function(val, key) {
  if (key.charAt(0) === '$') {
    factory[key] = val;
    // Don't try to copy over annotations to named controller
    if (isFunction(controller)) controller[key] = val;
  }
});
```
This may modify an angular method `noop` if component doesn't have a controller, which should not have happened.

### How I found it?
I came across this problem using ngComponentRouter 2.0.0 with Angular 1.5.3 when I had two components without controllers defined. In this situation when factory function is executed, the default controller (noop) has $routeConfig property from the later one defined component (because registerComponent modify the same object: `noop` function). Result: ngComponentRouter couldn't find properly defined route.

### Fix proposal
Probably the fix is fairly simple. Use `function() {}` instead of `noop`:

`var controller = options.controller || function() {};`

This will guarantee new function object at any situation.

gkalpak:
Bug fix

#14391

A new function is used as controller for each component that does not specify one.

No

**Other information**:
Currently, custom annotations are copied from the CDO onto the controller constructor.
Using `noop()` when no controller has been specified, pollutes it with custom annotations and makes one component's annotations available to all other components that have `noop()` as their controller.

Fixes #14391

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -10494,7 +10494,7 @@ describe('$compile', function() {
     });
 
     it('should expose additional annotations on the directive definition object', function() {
-      var myModule = angular.module('my', []).component('myComponent', {
+      angular.module('my', []).component('myComponent', {
         $canActivate: 'canActivate',
         $routeConfig: 'routeConfig',
         $customAnnotation: 'XXX'
@@ -10510,7 +10510,7 @@ describe('$compile', function() {
     });
 
     it('should support custom annotations if the controller is named', function() {
-      var myModule = angular.module('my', []).component('myComponent', {
+      angular.module('my', []).component('myComponent', {
         $customAnnotation: 'XXX',
         controller: 'SomeNamedController'
       });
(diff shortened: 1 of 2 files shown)


test/ng/compileSpec.js
```diff
@@ -10522,6 +10522,24 @@ describe('$compile', function() {
       });
     });
 
+    it('should provide a new empty controller if none is specified', function() {
+      angular.
+        module('my', []).
+        component('myComponent1', {$customAnnotation1: 'XXX'}).
+        component('myComponent2', {$customAnnotation2: 'YYY'});
+      module('my');
```
Narretz: There should be a newline before and after this line for readability.

Narretz:
LGTM, apart from a nitpick

PR #14404, opened 2016-04-09
perf(ngClass): refactor to optimize the case of static map of classes with large objects

drpicox:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**: performance improvement

**What is the current behavior? (You can also link to an open issue here)**: it makes copies of very large objects:

``` javascript
    it('should not copy verylargeobject', inject(function($rootScope, $compile) {
      var getProp = jasmine.createSpy('getProp');
      var verylargeobject = {};
      Object.defineProperty(verylargeobject, 'prop', {
        get: getProp,
        enumerable: true
      });
      element = $compile('<div ng-class="{foo: verylargeobject}"></div>')($rootScope);
      $rootScope.verylargeobject = verylargeobject;
      $rootScope.$digest();

      expect(getProp).not.toHaveBeenCalled();
    }));
```

**What is the new behavior (if this is a feature change)?**: the user perceives no change but that it is faster with very large objects

**Does this PR introduce a breaking change?**: no

**Other information**:
This PR responds to https://github.com/angular/angular.js/pull/14394#issuecomment-207812797 
It is alternative to https://github.com/angular/angular.js/pull/14394

### DIFF
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -8,32 +8,33 @@
 
 function classDirective(name, selector) {
   name = 'ngClass' + name;
-  return ['$animate', function($animate) {
+  return ['$animate','$parse', function($animate,$parse) {
     return {
       restrict: 'AC',
       link: function(scope, element, attr) {
-        var oldVal;
+        var oldClasses;
 
-        scope.$watch(attr[name], ngClassWatchAction, true);
+        var classFn = $parse(attr[name], arrayClasses);
+        scope.$watch(classFn, ngClassWatchAction, true);
 
         attr.$observe('class', function(value) {
-          ngClassWatchAction(scope.$eval(attr[name]));
+          ngClassWatchAction(classFn(scope));
         });
 
 
         if (name !== 'ngClass') {
           scope.$watch('$index', function($index, old$index) {
-            /* eslint-disable no-bitwise */
+            // eslint-disable-next-line no-bitwise
             var mod = $index & 1;
+            // eslint-disable-next-line no-bitwise
             if (mod !== (old$index & 1)) {
-              var classes = arrayClasses(scope.$eval(attr[name]));
               if (mod === selector) {
-                addClasses(classes);
+                ngClassWatchAction(classFn(scope));
               } else {
-                removeClasses(classes);
+                removeClasses(oldClasses);
+                oldClasses = [];
               }
             }
-            /* eslint-enable */
           });
         }
 
@@ -80,18 +81,13 @@ function classDirective(name, selector) {
         function ngClassWatchAction(newVal) {
           // eslint-disable-next-line no-bitwise
           if (selector === true || (scope.$index & 1) === selector) {
-            var newClasses = arrayClasses(newVal || []);
-            if (!oldVal) {
+            var newClasses = (newVal || []).join(' ').split(' ');
+            if (!oldClasses) {
               addClasses(newClasses);
-            } else if (!equals(newVal,oldVal)) {
-              var oldClasses = arrayClasses(oldVal);
+            } else if (!arrayEqualClasses(oldClasses, newClasses)) {
               updateClasses(oldClasses, newClasses);
             }
-          }
-          if (isArray(newVal)) {
-            oldVal = newVal.map(function(v) { return shallowCopy(v); });
-          } else {
-            oldVal = shallowCopy(newVal);
+            oldClasses = shallowCopy(newClasses);
           }
         }
       }
@@ -112,24 +108,46 @@ function classDirective(name, selector) {
     }
 
     function arrayClasses(classVal) {
-      var classes = [];
+      var classes;
       if (isArray(classVal)) {
-        forEach(classVal, function(v) {
-          classes = classes.concat(arrayClasses(v));
-        });
-        return classes;
-      } else if (isString(classVal)) {
-        return classVal.split(' ');
+        classes = classVal.map(classNames);
       } else if (isObject(classVal)) {
-        forEach(classVal, function(v, k) {
-          if (v) {
-            classes = classes.concat(k.split(' '));
+        classes = [];
+        forEach(classVal, function(val, key) {
+          if (val) {
+            classes.push(key);
+          } else if (isUndefined(val)) {
+            // This case is for one time binding:
+            // ::expression are evaluated until no undefineds are found
+            // if no undefined is placed inside the array,
+            // it would assume that the value is fully computed and will fail
+            classes.push(val);
           }
         });
-        return classes;
+      } else if (isString(classVal)) {
+        classes = [classVal];
+      } else {
+        classes = classVal;
       }
-      return classVal;
+
+      return classes;
     }
+
+    function classNames(values) {
+      if (isObject(values)) {
+        return Object.keys(values).filter(function(k) {
+          return values[k];
+        }).join(' ');
+      } else {
+        return values;
+      }
+    }
+
+    function arrayEqualClasses(a1, a2) {
+      return a1 === a2 ||
+        isArray(a1) && isArray(a2) && a1.join(' ') === a2.join(' ');
+    }
+
   }];
 }
 
--- a/test/ng/directive/ngClassSpec.js
+++ b/test/ng/directive/ngClassSpec.js
@@ -299,6 +299,54 @@ describe('ngClass', function() {
     expect(e2.hasClass('D')).toBeFalsy();
   }));
 
+  it('should compute class names when changing row parity and class value',
+    inject(function($rootScope, $compile) {
+      element = $compile('<ul>' +
+        '<li ng-repeat="i in arr track by i" ng-class-even="cls"></li>' +
+        '</ul>')($rootScope);
+
+      $rootScope.$apply(function() {
+        $rootScope.arr = ['a','b'];
+        $rootScope.cls = 'even';
+      });
+      var e3 = jqLite(element[0].childNodes[3]);
+      expect(e3).toHaveClass('even');
+
+      $rootScope.$apply(function() {
+        $rootScope.arr = ['b'];
+        $rootScope.cls = 'off';
+      });
+      var e1 = jqLite(element[0].childNodes[1]);
+      expect(e3[0]).toBe(e1[0]); // make sure same instance of ngClassEven
+      expect(e3).not.toHaveClass('even');
+      expect(e3).not.toHaveClass('off');
+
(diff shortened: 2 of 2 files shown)


test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,78 @@ describe('ngClass', function() {
     expect(e2.hasClass('even')).toBeTruthy();
     expect(e2.hasClass('odd')).toBeFalsy();
   }));
+
+  it('should do value stabilization as expected when one-time binding', inject(function($rootScope, $compile) {
```
gkalpak: Please, wrap lines at 100 chars.

test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,78 @@ describe('ngClass', function() {
...
+    element = $compile('<div ng-class=" :: className"></div>')($rootScope);
+    $rootScope.$digest();
+
+    $rootScope.className = 'foo';
+    $rootScope.$digest();
+    expect(element.hasClass('foo')).toBe(true);
```
gkalpak: Use the `toHaveClass()` custom matcher (here and elsewhere).

test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,78 @@ describe('ngClass', function() {
...
+
+    $rootScope.className = 'bar';
+    $rootScope.$digest();
+    expect(element.hasClass('foo')).toBe(true);
+  }));
+
+  it('should do value remove the watcher as expected when one-time binding', inject(function($rootScope, $compile) {
```
gkalpak: This test is redundat (the above tests is enough to verify the "one-time-bindness".
There should be also tests for `::{foo: bar}` and `::[...]` though. (I'm not sure the array case is supported by one-time binding, but the object case should be.)

drpicox:
For one time binding is getting hard, I just cannot previously convert to a plain string because of:

```javascript
  it('stabilizes when in a literal are not undefineds', inject(function($rootScope, $compile) {
    $rootScope.foo = true;
    element = $compile('<div ng-class="::{foo:foo,bar:bar}"></div>')($rootScope);
    $rootScope.$digest();
    expect(element).toHaveClass('foo');
    $rootScope.bar = true;
    $rootScope.$digest();
    expect(element).toHaveClass('bar');
    $rootScope.bar = false;
    $rootScope.$digest();
    expect(element).toHaveClass('bar');
  }));
```

In the current implementation the watcher watches a string which cannot contain undefineds. I will change the strategy. Fortunately it is only one level:

```javascript
  // this test fails (nested undefineds)
  it('stabilises when nested?', inject(function($rootScope, $compile) {
    $rootScope.foo = true;
    element = $compile('<div ng-class="::[{foo:foo,bar:bar}]"></div>')($rootScope);
    $rootScope.$digest();
    expect(element).toHaveClass('foo');
    $rootScope.bar = true;
    $rootScope.$digest();
    expect(element).toHaveClass('bar');
    $rootScope.bar = false;
    $rootScope.$digest();
    expect(element).toHaveClass('bar');
  }));
```

from `parse.js` (see that isAllDefined is one level, not nested):

```javascript
    function oneTimeLiteralWatchDelegate(scope, listener, objectEquality, parsedExpression) {
      var unwatch, lastValue;
      return unwatch = scope.$watch(function oneTimeWatch(scope) {
        return parsedExpression(scope);
      }, function oneTimeListener(value, old, scope) {
        lastValue = value;
        if (isFunction(listener)) {
          listener.call(this, value, old, scope);
        }
        if (isAllDefined(value)) {
          scope.$$postDigest(function() {
            if (isAllDefined(lastValue)) unwatch();
          });
        }
      }, objectEquality);

      function isAllDefined(value) {
        var allDefined = true;
        forEach(value, function(val) {
          if (!isDefined(val)) allDefined = false;
        });
        return allDefined;
      }
    }
```

drpicox:
@gkalpak can you take a look to:

- https://travis-ci.org/angular/angular.js/pull_requests

it seems that the following test fails randomly for Firefox:

```
[firefox 28 #2] ..F...............................................................................................................................................................
[firefox 28 #2] 
[firefox 28 #2] Failures:
[firefox 28 #2] 1) $anchorScroll with `yOffset` should automatically scroll when navigating to a URL with a hash
[firefox 28 #2]   Message:
[firefox 28 #2]     Expected #anchor-5's top to be 50, but it was 208
[firefox 28 #2]   Stack:
[firefox 28 #2]     Error: Failed expectation
[firefox 28 #2]         at /home/travis/build/angular/angular.js/test/e2e/tests/anchor-scroll.spec.js:113:32
[firefox 28 #2]         at /home/travis/build/angular/angular.js/test/e2e/tests/anchor-scroll.spec.js:160:7
```

gkalpak:
Yeah, I've seen it too. I thought these were flakes. Maybe #14912 will make them go away.

drpicox:
Indeed. Firefox 28 is quite old :D

gkalpak:
It's green now.

Narretz:
Did this LGTY back in July @gkalpak ?

Narretz:
@drpicox can you please rebase this?

gkalpak:
> Did this LGTY back in July @gkalpak ?

Not if I didn't say so :smiley:
(I got caught up in other stuff and I never gave this a final review. I'll take a look asap.)

drpicox:
Ops!

I'll do it now.
David

drpicox:
Rebased. Waiting for travis tests.

drpicox:
Green.

gkalpak (changes requested):
There a couple of issues with this implementation.

TBH, I feel like there should be a more straight-forward way to achieve the same. I'll poke around.

src/ng/directive/ngClass.js
```diff
@@ -8,92 +8,93 @@
...
+          function ngClassWatchAction(newVal) {
+            // eslint-disable-next-line no-bitwise
+            if (selector === true || (scope.$index & 1) === selector) {
+              var newClasses = (newVal || []).join(' ').split(' ');
+              if (!oldClasses) {
+                addClasses(newClasses);
+              } else if (!classArrayEquals(oldClasses, newClasses)) {
+                updateClasses(oldClasses, newClasses);
+              }
+              oldClasses = shallowCopy(newClasses);
```
drpicox: I'll try to write a test to make it fail and fix it.

test/ng/directive/ngClassSpec.js
```diff
@@ -424,6 +424,111 @@ describe('ngClass', function() {
...
+  it('should track changes of mutating object inside an array',
+    inject(function($rootScope, $compile) {
+      $rootScope.classVar = [{orange: true}];
+      element = $compile('<div ng-class="classVar"></div>')($rootScope);
+      $rootScope.$digest();
+
+      $rootScope.classVar[0].orange = false;
+      $rootScope.$digest();
+
+      expect(element).not.toHaveClass('orange');
```
drpicox: ok

test/ng/directive/ngClassSpec.js
```diff
@@ -424,6 +424,111 @@ describe('ngClass', function() {
...
+      expect(element).toHaveClass('foo');
+
+      $rootScope.className = 'bar';
+      $rootScope.$digest();
+      expect(element).toHaveClass('foo');
+      expect(element).not.toHaveClass('bar');
+    })
+  );
+
+  it('should do value remove the watcher when static map one-time binding',
```
drpicox: thanks

test/ng/directive/ngClassSpec.js
```diff
@@ -424,6 +424,111 @@ describe('ngClass', function() {
...
+      $rootScope.$digest();
+      expect(element).toHaveClass('foo');
+
+      $rootScope.className = 'bar';
+      $rootScope.$digest();
+      expect(element).toHaveClass('foo');
+    })
+  );
+
+  it('should do value remove the watcher when static array one-time binding',
```
drpicox: thanks

src/ng/directive/ngClass.js
```diff
@@ -8,92 +8,93 @@
...
       restrict: 'AC',
-      link: function(scope, element, attr) {
-        var oldVal;
-
-        scope.$watch(attr[name], ngClassWatchAction, true);
-
-        attr.$observe('class', function(value) {
-          ngClassWatchAction(scope.$eval(attr[name]));
+      compile: function(tElement, tAttr) {
+        var classFn = $parse(tAttr[name], function(value) {
```
gkalpak: The overhead should be negligible (and not enough to justify the breaking change).

<sub>(Note that `$parse` has a cache, so it doesn't have to actually lex/parse the same expressiom twice. It only has to retrieve the parsed expression and apply the interceptor.)</sub>

drpicox: I'll put it again inside a link function. 
I suggest that it can be considered a "major upgrade" to do such refactor of all parse expression code from link functions to compile functions, in order to improve performance when ng-repeat & others are involved.

src/ng/directive/ngClass.js
drpicox: Trying to find a test to make it break, I found a case that breaks angular in older versions (at least since 1.3.8 -I didn't tried more-). Is the case of the first scope.$watch that you comment. 
I have the plnkr here: https://plnkr.co/edit/W1ck8dS08TCJpirWiGVA?p=preview
I solve it here, but I do not know if I should create another PR for solving it in the current code, what did you suggest?

I tried to find a failing test for the shallow copy inside the condition so I should move it. But I have found that moving the shallow copy outside the condition interferes with the $index $watch and make the new test fails.

src/ng/directive/ngClass.js
```diff
@@ -8,92 +8,93 @@
...
-              if (mod === selector) {
-                addClasses(classes);
-              } else {
-                removeClasses(classes);
-              }
-            }
-            /* eslint-enable */
-          });
-        }
+          scope.$watch(classFn, ngClassWatchAction, classArrayEquals);
```
drpicox: I'll check, I was wrong. It takes classArrayEquals, using true, all tests passes as expected, no extra copies performed (I have also check the $watch code).
So it is updated, Thanks.

gkalpak:
I have an alternative PR (#15228) which includes several refactorings, the fix mentioned above and some perf improvements (that should cover the ones in this PR and more).

I have "borrowed" the tests and part of the implementation :grin:

WDYT?

drpicox:
You have done a really good job 👍

PR #14405, opened 2016-04-09
Fix ngClass mixed array/object when object instance does not changes

drpicox:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)** bugfix

**What is the current behavior? (You can also link to an open issue here)**: when the expression is [{}] it does not tracks changes inside the object

``` javascript
  it('should support changing multiple classes via variable array mixed with conditionally via a map', inject(function($rootScope, $compile) {
    $rootScope.classVar = ['', {orange: true}];
    element = $compile('<div class="existing" ng-class="classVar"></div>')($rootScope);
    $rootScope.$digest();

    $rootScope.classVar[1].orange = false;
    $rootScope.$digest();

    expect(element.hasClass('orange')).toBeFalsy(); /// <<--- this fails before this patch
  }));
```

**What is the new behavior (if this is a feature change)?**: now it tracks changes correctly

**Does this PR introduce a breaking change?**: nop

**Other information**:

In this plnkr you can see the failure:
- http://plnkr.co/edit/f0V9RVDGREetSW00TTRZ?p=preview 

https://github.com/angular/angular.js/pull/14404 also fixes this problem by doing a refactor. (it tries to be a performance optimization) 
This PR can be combined into https://github.com/angular/angular.js/pull/14394 if you ask.

### DIFF
--- a/test/ng/directive/ngClassSpec.js
+++ b/test/ng/directive/ngClassSpec.js
@@ -410,16 +410,20 @@ describe('ngClass', function() {
     expect(e2.hasClass('odd')).toBeFalsy();
   }));
 
-  it('should support changing multiple classes via variable array mixed with conditionally via a map', inject(function($rootScope, $compile) {
-    $rootScope.classVar = ['', {orange: true}];
-    element = $compile('<div class="existing" ng-class="classVar"></div>')($rootScope);
-    $rootScope.$digest();
+  it('should support mixed array/object variable with a mutating object',
+    inject(function($rootScope, $compile) {
+      element = $compile('<div ng-class="classVar"></div>')($rootScope);
 
-    $rootScope.classVar[1].orange = false;
-    $rootScope.$digest();
+      $rootScope.classVar = [{orange: true}];
+      $rootScope.$digest();
+      expect(element).toHaveClass('orange');
 
-    expect(element.hasClass('orange')).toBeFalsy();
-  }));
+      $rootScope.classVar[0].orange = false;
+      $rootScope.$digest();
+
+      expect(element).not.toHaveClass('orange');
+    })
+  );
 
 });
 
--- a/test/ng/directive/ngClassSpec.js
+++ b/test/ng/directive/ngClassSpec.js
@@ -409,6 +409,18 @@ describe('ngClass', function() {
     expect(e2.hasClass('even')).toBeTruthy();
     expect(e2.hasClass('odd')).toBeFalsy();
   }));
+
+  it('should support changing multiple classes via variable array mixed with conditionally via a map', inject(function($rootScope, $compile) {
+    $rootScope.classVar = ['', {orange: true}];
+    element = $compile('<div class="existing" ng-class="classVar"></div>')($rootScope);
+    $rootScope.$digest();
+
+    $rootScope.classVar[1].orange = false;
+    $rootScope.$digest();
+
+    expect(element.hasClass('orange')).toBeFalsy();
+  }));
+
 });
 
 describe('ngClass animations', function() {
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -78,7 +78,11 @@ function classDirective(name, selector) {
               updateClasses(oldClasses, newClasses);
             }
           }
-          oldVal = shallowCopy(newVal);
+          if (isArray(newVal)) {
+            oldVal = newVal.map(function(v) { return shallowCopy(v); });
+          } else {
+            oldVal = shallowCopy(newVal);
+          }
         }
       }
     };


drpicox:
As I commented in #14394 there is this case that are not resolved here and it has two solutions. The problem:

```javascript
  // this test actually passes
  it('should support mixed nested array and object leafs variable', inject(function($rootScope, $compile) {
    $rootScope.classVar = ['red', ['green', {blue: true}]];
    element = $compile('<div class="existing" ng-class="classVar"></div>')($rootScope);
    $rootScope.$digest();

    expect(element.hasClass('red')).toBeTruthy();
    expect(element.hasClass('green')).toBeTruthy();
    expect(element.hasClass('blue')).toBeTruthy();
  }));

  // this test fails even with this patch
  it('should support mixed nested array and object leafs variable with mutating object', inject(function($rootScope, $compile) {
    $rootScope.classVar = ['red', ['green', {blue: true}]];
    element = $compile('<div class="existing" ng-class="classVar"></div>')($rootScope);
    $rootScope.$digest();

    $rootScope.classVar[1][1].blue = false;

    expect(element.hasClass('red')).toBeTruthy();
    expect(element.hasClass('green')).toBeTruthy();
    expect(element.hasClass('blue')).toBeFalsy();
  }));
```

Solutions:

1. **Do not accept nested array syntax**: it would be "**breaking changes**" so any current implementation that it is using it. But it is not likely because it is not documented and the tracking of changes is not working properly. It have to be checked what happens in Angular2, because Angular2 uses this concept of nested array to express Injection and Component dependencies it is possible that feels comfortable for the programmer having nested arrays.

2. **Deep solve change tracking only for arrays**: it will not break the implementation and only will cost nothing (except when using nested arrays because the checking would be done properly). In such case the implementation of #14404 is already solving this.

¿Which do you consider more acceptable?

gkalpak:
The docs are very explicit about what types of expressions are supported in `ngClass`. Nested arrays are **not** supported, so if anyone uses them, they are out of luck imho.

According to the docs, the expression used with `ngClass` should evaluate to one of 3 things:

1. A string.
2. An object.
3. An array containing strings and/or objects.

(Where _objects_ are treated as key-value pairs, with keys denoting classes and values determining the presense or absense of the corresponding key/class.)

test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,18 @@ describe('ngClass', function() {
     expect(e2.hasClass('even')).toBeTruthy();
     expect(e2.hasClass('odd')).toBeFalsy();
   }));
+
+  it('should support changing multiple classes via variable array mixed with conditionally via a map', inject(function($rootScope, $compile) {
```
gkalpak: Pleasd wrap lines at 100 chars.

test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,18 @@ describe('ngClass', function() {
...
+    $rootScope.classVar = ['', {orange: true}];
```
gkalpak: Is the `''` necessary ?

test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,18 @@ describe('ngClass', function() {
...
+    element = $compile('<div class="existing" ng-class="classVar"></div>')($rootScope);
```
gkalpak: Remove the `class="existing"` part. It's better to keep the test code as simple as possible.

test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,18 @@ describe('ngClass', function() {
...
+    $rootScope.$digest();
+
```
gkalpak: Add an expectation to make sure `orange` is present (just in case :smiley:).

test/ng/directive/ngClassSpec.js
```diff
@@ -409,6 +409,18 @@ describe('ngClass', function() {
...
+    $rootScope.classVar[1].orange = false;
+    $rootScope.$digest();
+
+    expect(element.hasClass('orange')).toBeFalsy();
```
gkalpak: Use the `toHaveClass()` custom matcher.

gkalpak:
I left a couple of minor comments wrt the test, but LGTM otherwise. :+1: 
Feel free to ping me once you address them.

drpicox:
This case is getting hard:

```javascript
  it('stabilizes when in a literal are not undefineds', inject(function($rootScope, $compile) {
    $rootScope.foo = true;
    element = $compile('<div ng-class="::{foo:foo,bar:bar}"></div>')($rootScope);
    $rootScope.$digest();
    expect(element).toHaveClass('foo');
    $rootScope.bar = true;
    $rootScope.$digest();
    expect(element).toHaveClass('bar');
    $rootScope.bar = false;
    $rootScope.$digest();
    expect(element).toHaveClass('bar');
  }));
```

Because the watcher watches a string which cannot contain undefineds. I'll change the strategy.

drpicox:
@gkalpak I think that is now right. 
Just ask and I'll squash all commits in one.

gkalpak:
It's OK, I'll squash while merging. Thx :+1:

PR #14410, opened 2016-04-11
Make numeric work for all languages

AnObstacle:
According to wiki: https://en.wikipedia.org/wiki/Decimal_mark, it work for all languages except Arabic, Persian and braille (witch are Unicode).

Now typing 3,3 is equal to 3.3.


gkalpak:
The problem with this change is that some browsers (e.g. IE) return the raw content of the input as its value (instead of normalizing the decimal separator according to the user's locale for example).
E.g. if I enter `1,000` in an input field, I might mean 1000 or 1.0 (depending on my locale).

Currently, the input is marked an invalid, so there is no ambiguity. With the proposed change, it would be converted to 1.0, which is a bug.

So, thx for the PR, we are going to pass this time :+1:
Keep'em coming though :smiley:

PR #14223, opened 2016-03-12
docs(tutorial/2 - Angular Templates): describe your change...

unseen1980:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
docs update

**What is the current behavior? (You can also link to an open issue here)**
https://docs.angularjs.org/tutorial/step_02

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change?**
No

**Please check if the PR fulfills these requirements**
- [v ] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**Other information**:

Reference to wrong file

PR #14178, opened 2016-03-04
docs(tutorial/0 - Bootstrapping): describe your change...

tbwork:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
  NO
- **What is the current behavior?** (You can also link to an open issue here)
- **What is the new behavior (if this is a feature change)?**
- **Does this PR introduce a breaking change?**

NO
- **Please check if the PR fulfills these requirements**
- [ ] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- **Other information**:

As describe below "The `ng-app` attribute represents an Angular directive named `ngApp` (Angular uses
  `spinal-case` for its custom attributes and `camelCase` for the corresponding directives
  which implement them).". `ng-app` is an attribute, not a directive.

PR #13834, opened 2016-01-24
docs(tutorial): update PhoneCat tutorial to Angular 1.5, components and style guide

teropa:
Bring the PhoneCat tutorial up to speed with Angular 1.5 and the modern
component style. Introduce components instead of bare ngControllers from
the get-go. Add an additional chapter about organizing the code in feature modules
and using the style guide.

This also makes the guide compatible with the Angular 2 upgrade tutorial
on angular.io.

Closes #13312
Related to angular/angular-phonecat#289

PR #13632, opened 2015-12-25
docs(tutorial/step-5): .success has been deprecated

wytesk133:

PR #14416, opened 2016-04-12
docs(tutorial): update to use `v1.5.x` and best practices

### LINKED ISSUE #12755: docs(tutorial): Code samples use deprecated $http promise methods
The tutorial ([website](https://docs.angularjs.org/tutorial), [repo](https://github.com/angular/angular.js/blob/master/docs/content/tutorial/)) looks like it might be due for at least a few minor changes. In particular, starting in [step 5](https://docs.angularjs.org/tutorial/step_05#controller), it sets up a call using the $http service like this:
```
phonecatApp.controller('PhoneListCtrl', function ($scope, $http) {
  $http.get('phones/phones.json').success(function(data) {
    $scope.phones = data;
  });

  $scope.orderProp = 'age';
});
```
When, according to the most recent docs, [we aren't supposed to do that anymore](https://docs.angularjs.org/api/ng/service/$http#deprecation-notice). I'd be happy to move everything into using the `.then()` method, but this would also require fixing the [angular-phonecat](https://github.com/angular/angular-phonecat) repo that's used in the tutorial, and I'm not sure if there's a particular process in place for coordinating changes so they line up properly. (Also, to be honest, I have no idea how you've been dealing with pushing changes through all the phonecat "releases" retroactively without someone dying in a rebasing accident.)

### LINKED ISSUE #13312: Phonecat tutorial does not follow style guide
The [Phonecat tutorial](https://docs.angularjs.org/tutorial) does not really follow the conventions laid out by the [Angular Styleguide](https://github.com/johnpapa/angular-styleguide).

In particular, I think the following rules should be applied:

* [Rule of 1](https://github.com/johnpapa/angular-styleguide#style-y001)
* [Folders-by-feature structure](https://github.com/johnpapa/angular-styleguide#style-y152)
* [controllerAs View Syntax](https://github.com/johnpapa/angular-styleguide#style-y030)

It also might not hurt to explicitly refer to the style guide in the tutorial.

### LINKED ISSUE #13623: phonecat setup instructions seems to be out of date
First of all, I am a Visual Studio / .NET MVC / jQuery / Chrome developer and I am new to node and npm and git.  
I want and need to learn to develop with angularjs and I have the patience to dig in and succeed but the phonecat application set up instructions are not working for me.

Initially, I am having problems setting up phonecat.  I cannot run git due to certificate problem:

C:\WebProjects>git clone --depth=14 https://github.com/angular/angular-phonecat.git

Cloning into 'angular-phonecat'...
fatal: unable to access 'https://github.com/angular/angular-phonecat.git/': SSL certificate problem: self signed certificate in certificate chain

My workaround, in which I have little confidence, except that it did download a lot of files including the package.json:
C:\WebProjects>git -c http.sslVerify=false clone --depth=14 https://github.com/angular/angular-phonecat.git

Then npm failed, looking for python which I do not have and is not mentioned as a prerequisite in the instructions.

I got a lot of messages about the wrong versions.  The warnings are unnecessary because I have much newer versions than these very old instructions.

C:\WebProjects\angular-phonecat>npm installnpm WARN engine karma@0.12.37: wanted: {"node":">=0.8 <=0.12 || >=1 <=2"} (current: {"node":"4.2.3","npm":"2.14.7"})

Why is this code so old that it is looking for version 0.x and I have version 4.x?
Had the same warnings with node looking for version 0.x and I have 4.x.

> angular-phonecat@0.0.0 postinstall C:\WebProjects\angular-phonecat
> bower install
bower SELF_SIGNED_CERT_IN_CHAIN      Request to https://bower.herokuapp.com/packages/jquery failed: self signed certificate in certificate chain

I received a bunch of errors like this and I feel discouraged and I am not going to continue with the phonecat tutorial.

Not being experienced with node and npm and python this is surprising, confusing and disappointing.  
It makes me think that the phonecat sample app is too old to bother with.  Please either update phonecat or replace it.

I am not having a good user experience setting up phonecat with these challenges.  
I wish someone would test these instructions and update them.

Sincerely, Joe Kahl

gkalpak:
This is a major re-structuring of the tutorial app's codebase, aiming at applying established best practices (in terms of file naming/layout and code organization) and utilizing several new features and enhancements (most notably components) introduced in recent versions of Angular (especially v1.5).

Apart from the overall changes, two new chapters were introduced: one on components and one on code organization.
## 

In the process, several other things were (incidentally) taken care of, including:
- Dependencies were upgraded to latest versions.
- Animations were polished.
- Outdated links were updated.
- The app's base URL was changed to `/` (instead of `/app/`).

BTW, this has been tested with the following versions of Node (on Windows 10) and everything worked fine:
- 0.11.16
- 4.2.6
- 4.4.2
- 5.10.0
## 

This was inspired by (and loosely based on) #13834.
Again, mad props to @teropa for leading the way :)
## 

**Note:**
The old version of the tutorial, that is compatible with Angular version 1.4 or older, has been saved on the `pre-v1.5.0-snapshot` branch of [angular-phonecat](https://github.com/angular/angular-phonecat). The `v1.4.x` version of the tutorial should be pointed to that branch instead of `master`.
## 

Related to angular/angular-phonecat#326.
Related to angular/angular-seed#341.

---

Fixes #12755
Fixes #13312
Fixes #13623
Fixes #13632

Closes #8952
Closes #11726
Closes #12946
Closes #12947
Closes #13198
Closes #13284
Closes #13834
Closes #14178
Closes #14223


gkalpak:
This has to get merged simultaneously with corresponding PRs on:

Repo | PRs
--- | ---
_angular-phonecat_ | [#326][phonecat-326]
_angular-seed_ | [#329][seed-329], [#333][seed-333]
_angular.io_ | TBD (see [Upgrading from 1.x][ng2-upgrade])

--
Things that still need to be fixed:

- [ ] The images under `docs/img/tutorial` must be updated to match the new text/code.

--
Things that should/could be cosidered (potentially after this PR has been merged merged):

- [ ] Update the tutorial (and the necessary docs-app directives) on the `v1.4.x` branch, which will still contain the old version of the tutorial, to point to the `pre-v1.5.0-snapshot` branch of [angular-phonecat][phonecat].
- [ ] Adding an "advanced" section at the end, either pointing to resources (e.g. relevant Developer Guide sections) or as subsequent steps. Example topics:
    * Directives/DDO, 2-way/1-way/1-time bindings
    * Forms (+ `ngMessages`)
    * `ngComponentRouter`
    * (multi-slot) transclusion
    * Task runners/Build tools and best practices
    * Testing animations

- [ ] To discuss: Is it better to import the `core` module from the main `phonecatApp` module only, or on a per-view basis ?
- [ ] To discuss: Is it better to split up the CSS per component or keep it in a couple top-level files (e.g. one for animations and one for the rest) ? What about animation-related JS code ?
- [ ] To discuss: Do we want to mention/use Protractor's `directConnect: true` config option ?


[phonecat]: https://github.com/angular/angular-phonecat
[phonecat-326]: https://github.com/angular/angular-phonecat/pull/326
[ng2-upgrade]: https://angular.io/docs/ts/latest/guide/upgrade.html
[seed-329]: https://github.com/angular/angular-seed/pull/329
[seed-333]: https://github.com/angular/angular-seed/pull/333

teropa:
@gkalpak I can look into getting angular.io updated. I'll need to look into this branch in more detail to see the extent of the changes need to that guide, but in any case I'll bring it up with the docs team tomorrow.

gkalpak:
@teropa: Thx, that would be great ! Also, as you look through it, any feedback is more than welcome :smiley:

BTW, I tried to follow the structure that you used in the Upgrade Guide more closely, so hopefully you will be able to get rid of some of the noise in the guide and focus more on the new/different concepts.

--
Big :+1: for your work on both #13834 and the [Upgrade Guide][1].

[1]: https://angular.io/docs/ts/latest/guide/upgrade.html

docs/content/tutorial/index.ngdoc
```diff
@@ -62,24 +65,26 @@ a few git commands.
 ### Install Git
 
 You can download and install Git from http://git-scm.com/download. Once installed, you should have
-access to the `git` command line tool.  The main commands that you will need to use are:
+access to the `git` command line tool. The main commands that you will need to use are:
+
+* `git clone ...`: Clone a remote repository onto your local machine.
+* `git checkout ...`: Check out a particular branch or a tagged version of the code to hack on.
```
petebacondarwin: There is an annoying issue about corporate firewalls blocking git

https://github.com/angular/angular-phonecat/issues/141
and probably
https://github.com/angular/angular-phonecat/issues/235

Can we add this information into the guide?

docs/content/tutorial/index.ngdoc
```diff
@@ -129,30 +131,32 @@ Once you have Node.js installed on your machine, you can download the tool depen
...
-- [Http-Server][http-server] - simple local static web server
-- [Karma][karma] - unit test runner
-- [Protractor][protractor] - end to end (E2E) test runner
+* [Bower][bower] - client-side code package manager
+* [Http-Server][http-server] - simple local static web server
+* [Karma][karma] - unit test runner
+* [Protractor][protractor] - end-to-end (E2E) test runner
 
-Running `npm install` will also automatically use bower to download the Angular framework into the
+Running `npm install` will also automatically use bower to download the AngularJS framework into the
```
petebacondarwin: how about we ditch bower altogether and just use npm to do the installation of angular libraries?

docs/content/tutorial/index.ngdoc
```diff
@@ -190,103 +194,111 @@ application during development. Start the web server by running:
...
 
-You will need to have Java present on your dev machine to allow the Selenium standalone to be started.
-Check if you already have java installed by opening a terminal/command line window and typing
-'''
-java -version
-'''
-If java is already installed and exists in the PATH then you will be shown the version installed,
-if, however you receive a message that "java is not recognized as an internal command or external
-command" you will need to install [java].
+  If JDK is not already install, you can download it [here][jdk-download].
```
petebacondarwin: If JDK is not already **installed**

docs/content/tutorial/step_00.ngdoc
```diff
@@ -7,130 +7,143 @@
...
 ```
 npm install
 ```
 
-To see the app running in a browser, open a *separate* terminal/command line tab or window, then
-run `npm start` to start the web server. Now, open a browser window for the app and navigate to
-<a href="http://localhost:8000/app/" target="_blank" title="Open app on localhost">`http://localhost:8000/app/`</a>
+To see the app running in a browser, open a _separate_ terminal/command line tab or window, then run
+`npm start` to start the web server. Now, open a browser window for the app and navigate to
+http://localhost:8000/index.html.
```
petebacondarwin: Should we consider using `lite-server` instead of `http-server` since it has a couple of nice features like automatic rewriting to index.html (to support HTML5 mode) and also it starts up the browser when you start the server.

docs/content/tutorial/step_00.ngdoc
```diff
@@ -7,130 +7,143 @@
...
-<body>
-
-  <p>Nothing here {{'yet' + '!'}}</p>
-
-</body>
+<html lang="en" ng-app="phonecatApp">
+  <head>
+    <meta charset="utf-8">
+    <title>My HTML File</title>
+    <link rel="stylesheet" href="bower_components/bootstrap/dist/css/bootstrap.css" />
```
petebacondarwin: `<link>` tags are empty, we should not use a self-closing tag

docs/content/tutorial/step_01.ngdoc
```diff
@@ -12,11 +12,11 @@ dynamically display the same result with any set of data.
 
 In this step you will add some basic information about two cell phones to an HTML page.
 
-- The page now contains a list with information about two phones.
+* The page now contains a list with information about two phones.
 
 <div doc-tutorial-reset="1"></div>
 
-
+<br />
```
petebacondarwin: `<br>` tags are also empty

docs/content/tutorial/step_02.ngdoc
```diff
@@ -49,128 +50,150 @@ __`app/index.html`:__
 </html>
 ```
 
-We replaced the hard-coded phone list with the {@link ng.directive:ngRepeat ngRepeat directive}
-and two {@link guide/expression Angular expressions}:
+We replaced the hard-coded phone list with the {@link ng.directive:ngRepeat ngRepeat} directive and
+two {@link guide/expression Angular expressions}:
```
petebacondarwin: You should be able to simply write `{@link ngRepeat}` here, unless you don't want the `ngRepeat` to be styled as code, in which case you can write `{@link ngRepeat ntRepeat}`

docs/content/tutorial/step_03.ngdoc
```diff
@@ -1,223 +1,279 @@
...
+  </p>
+  <p>
+    (**Note:** Directives are an advanced topic, so you might want to postpone studying them, until
+               you have mastered the basics.)
+  </p>
+</div>
 
-* Data-binding: This is one of the core features in Angular. When the page loads, Angular binds the
-name of the input box to a variable of the same name in the data model and keeps the two in sync.
+To create a component, we use the `.component()` method of an {@link module Angular module}. We must
```
petebacondarwin: perhaps a link to the component helper here?

docs/content/tutorial/step_03.ngdoc
```diff
@@ -1,223 +1,279 @@
...
+For reasons already mentioned (and for other reasons that are out of the scope of this tutorial), it
+is considered a good practice to avoid using the scope directly. We can (and should) use our
+controller instance; i.e. assign our data and methods on properties of our controller (the "`this`"
+inside the controller constructor), instead of directly to the scope.
 
-__`test/e2e/scenarios.js`:__
+From the template, we can refer to our controller instance using an alias. This way, the context of
+evaluation for our expressions is even more clear. By default, components use `$ctrl` as the
+controller alias, but we can override it, should the need arise. (It is a very uncommon need though,
+so we will be sticking with `$ctrl` most of the time.)
```
petebacondarwin: I would not bother with the sentence in parentheses

docs/content/tutorial/step_03.ngdoc
```diff
@@ -1,223 +1,279 @@
...
-      query.sendKeys('motorola');
-      expect(phoneList.count()).toBe(2);
-    });
+<br />
+**`app/phone-list.component.js`:**
+
+```js
+// Register `phoneList` component, along with its associated controller and template
+angular.
+  module('phonecatApp').
```
petebacondarwin: is it officially recommended practice to "reopen" the module to add in the component?
Given that you are creating a new file anyway, why not have a one module = one file bijection.
The benefits of this are:
* you don't have to sort the files when building a distribution file
* you can easily move a file to another project without having to bring along its module definition

docs/content/tutorial/step_03.ngdoc
```diff
@@ -1,223 +1,279 @@
...
+  beforeEach(module('phonecatApp'));
 
-### Display Query in Title
-Let's see how we can get the current value of the `query` model to appear in the HTML page title.
+  // Test the controller
+  describe('PhoneListController', function() {
 
-* Add an end-to-end test into the `describe` block, `test/e2e/scenarios.js` should look like this:
+    it('should create a `phones` model with 3 phones', inject(function($componentController) {
+      var ctrl = $componentController('phoneList', {$scope: {}});
```
petebacondarwin: `{$scope: {}}` not needed any more!!

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
-```
 
-We made the following changes to the `index.html` template:
+## Organizing by Feature
 
-* First, we added a `<select>` html element named `orderProp`, so that our users can pick from the
-two provided sorting options.
+So, now that we learned we should put everything in its own file, our `app/` directory will soon be
+full with dozens of files and specs (remember we keep our unit test files next to the corresponding
+source code files). What's more important, logically related files will not be groupped together; it
```
petebacondarwin: groupped -> grouped

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+will be really difficult of locate all files related to a specific section of the application and
+make a change or fix a bug.
 
-<img class="diagram" src="img/tutorial/tutorial_04.png">
+So, what shall we do?
 
-* We then chained the `filter` filter with {@link ng.filter:orderBy `orderBy`}
-filter to further process the input into the repeater. `orderBy` is a filter that takes an input
-array, copies it and reorders the copy which is then returned.
+Well, we are going to group our files into directories _by feature_. For example, since we have a
```
petebacondarwin: I am pretty sure that directories is a Windows thing :-)
Shouldn't we stick with folders throughout instead?

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+section in our application that lists phones, we will put all related files into a `phone-list/`
+directory under `app/`. We are soon to find out that certain features are used across different
+parts of the application. We will put those inside `app/core/`.
 
-Angular creates a two way data-binding between the select element and the `orderProp` model.
-`orderProp` is then used as the input for the `orderBy` filter.
+<div class="alert alert-info">
+  <p>
+    Other typical names for our `core` directory are `common` and `components`. The latter is kind
```
petebacondarwin: I sometimes use `shared` too

petebacondarwin: what does the "style-guide" have to say about this?

petebacondarwin: at one point people were also playing with the name `ng_modules` or `angular_modules` but this was deemed too verbose.

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+Based on what we have discussed so far, here is our directory/file layout for the `phoneList`
+"feature":
+
+```
+  app/
+    phone-list/
+      phone-list.component.js
+      phone-list.component.spec.js
+    app.js
+```
```
petebacondarwin: Perhaps it would be nice for this directory structure to show the `core` folder too?

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
 
-As we discussed in the section about data-binding and the repeater in step 3, whenever the model
-changes (for example because a user changes the order with the select drop down menu), Angular's
-data-binding will cause the view to automatically update. No bloated DOM manipulation code is
-necessary!
 
+## Using Modules
 
+As previously mentioned, one of the benefits of having a modular architecture is code reuse &mdash;
+not only inside the same application, but across application too. There is one final step in making
```
petebacondarwin: across application**s**

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+will make all entities registered on `phoneList` available on `phonecatApp` as well.
 
-The unit test now verifies that the default ordering property is set.
+<div class="alert alert-info">
+  <p>
+    Don't forget to also update your `index.html` adding a `<script>` tag for each JacaScript file
+    we have created. This might seem tedious, but is totally worth it.
+  </p>
+  <p>
+    In a production-ready application, you are better off concatenating and minifying all your
```
petebacondarwin: I think this reads slightly better:

> In a production-ready application, you would concatenate and minify all your JavaScript files anyway...

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
 ```
 
+By passing `phoneList` inside the dependencies array when defining the `phonecatApp` module, Angular
+will make all entities registered on `phoneList` available on `phonecatApp` as well.
 
-The unit test now verifies that the default ordering property is set.
+<div class="alert alert-info">
+  <p>
+    Don't forget to also update your `index.html` adding a `<script>` tag for each JacaScript file
+    we have created. This might seem tedious, but is totally worth it.
```
petebacondarwin: It might be worth mentioning (since you've done it this way) that the module files must be loaded before component files.

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+which are basically fragments of HTML code that dictate how our data is laid out and presented to
+the user. In {@link step_03 step 3}, we saw how we can specify the template for a component as a
+string using the `template` property of the CDO (Component Definition Object).
+Having HTML code in a string isn't ideal, especially for bigger templates. It would be much better,
+if we could have our HTML code in `.html` files. This way, we would get all the support our
+IDE/editor has to offer (e.g. HTML-specific color-highlighting and auto-completion) and also keep
+our component definitions cleaner.
 
+So, while it's perfectly fine to keep our component templates inline (using the `template` property
+of the CDO), we decided to use an external template for our `phoneList` component. In order to
```
petebacondarwin: How about?

> we are going to use

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+denote that we are using an external template, we use the `templateUrl` property and specify the URL
+that our template will be loaded from. Since we want to keep our template close to where the
+component is defined, we place it inside `app/phone-list/`.
 
-Let's turn our attention to the end-to-end test.
+We copied the contents of the `template` property (the HTML code) into
```
petebacondarwin: Is the language of the tutorial past tense: we did this, we did that?
Or is it imperative: do this or do that?

I am inclined toward the latter.

In which case this should be :

> Copy the contents of the `template` ...

Does that sound right to you? If so can you check this styling throughout?

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
-      query.sendKeys('tablet'); //let's narrow the dataset to make the test assertions shorter
 
-      expect(getNames()).toEqual([
-        "Motorola XOOM\u2122 with Wi-Fi",
-        "MOTOROLA XOOM\u2122"
-      ]);
+## Final Directory/File Layout
 
-      element(by.model('orderProp')).element(by.css('option[value="name"]')).click();
+After all the refactorings that took place, this is how our application looks like from the outside:
```
petebacondarwin: I think this should be either

> this is how our application looks from the outside

or

> this is what our application looks like from the outside

I would prefer the former

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+```
+  app/
+    phone-list/
+      phone-list.component.js
+      phone-list.component.spec.js
+      phone-list.module.js
+      phone-list.template.html
+    app.css
+    app.module.js
+    index.html
```
petebacondarwin: this is missing the core directory

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
-* Add an `{{orderProp}}` binding into the `index.html` template to display its current value as
-text.
+});
+```
+
+If not already done so, run the tests (using the `npm test` command) and verify that they still
+pass.
+
+<div class="alert alert-success">
+  One of the great things about tests is the confidence the provide, when refactoring your
```
petebacondarwin: confidence **they** provide

docs/content/tutorial/step_04.ngdoc
```diff
@@ -1,199 +1,310 @@
...
+  application. It's easy to break something as you start moving files around and re-arranging
+  modules. Having a good test coverage is the quickest, easiest and most reliable way of knowing
```
petebacondarwin: > Having good test coverage

docs/content/tutorial/step_04.ngdoc
petebacondarwin: I just realised that perhaps this bit is past tense because each step might already have some files setup for the developer already?

docs/content/tutorial/step_05.ngdoc
```diff
@@ -1,277 +1,168 @@
...
+  of the model.
 
-You can create your own services, and in fact we will do exactly that in step 11. As a naming
-convention, Angular's built-in services, Scope methods and a few other Angular APIs have a `$`
-prefix in front of the name.
+  <img  class="diagram" src="img/tutorial/tutorial_05.png">
 
-The `$` prefix is there to namespace Angular-provided services.
-To prevent collisions it's best to avoid naming your services and models anything that begins with a `$`.
+* Use of the `filter` filter: The {@link ng.filter:filter filter} function uses the `$ctrl.query`
```
petebacondarwin: pretty sure this link can be more simply: `{@link filter}`

docs/content/tutorial/step_05.ngdoc
```diff
@@ -1,277 +1,168 @@
...
-### A Note on Minification
+# Testing
 
-Since Angular infers the controller's dependencies from the names of arguments to the controller's
-constructor function, if you were to [minify](http://goo.gl/SAnnsm) the JavaScript code for
-`PhoneListCtrl` controller, all of its function arguments would be minified as well, and the
-dependency injector would not be able to identify services correctly.
+In previous steps, we learned how to write and run unit tests. Unit tests are perfect for testing
+controllers and other parts of our application written in JavaScript, but they can't easily
+test DOM manipulation or the wiring of our application. For these, an end-to-end (E2E) test is a
```
petebacondarwin: might be worth mentioning templates here:

> can't easily test templates, DOM manipulation or wiring of our application

petebacondarwin: actually unless you defined wiring earlier on, I think this is a bit of a vague term. Perhaps `bindings` would be better?

docs/content/tutorial/step_05.ngdoc
```diff
@@ -1,277 +1,168 @@
...
-```
+Even though the syntax of this test looks very much like our controller unit test written with
+Jasmine, the E2E test uses APIs of [Protractor][protractor]. Read about the Protractor APIs in the
+[Protractor API Docs][protractor-docs].
 
-You should now see the following output in the Karma tab:
-
-<pre>Chrome 22.0: Executed 2 of 2 SUCCESS (0.028 secs / 0.007 secs)</pre>
+Much like Karma is the test runner for unit tests, we use Protractor to run E2E tests. Try it with
+`npm run protractor`. E2E tests are slow, so unlike with unit tests, Protractor will exit after the
```
petebacondarwin: > E2E tests take time

Sounds less negative :-)

docs/content/tutorial/step_05.ngdoc
```diff
@@ -1,277 +1,168 @@
...
+tests run and will not automatically rerun the test suite on every file change.
+To rerun the test suite, execute `npm run protractor` again.
 
+<div class="alert alert-info">
+  **Note:** You must ensure your application is being served via a web server to test with
+  protractor. You can do this by using `npm start`.
```
petebacondarwin: In a different terminal/console

docs/content/tutorial/step_06.ngdoc
```diff
@@ -1,108 +1,239 @@
...
+
+  <img class="diagram" src="img/tutorial/tutorial_06.png">
 
-We also added phone images next to each record using an image tag with the {@link
-ng.directive:ngSrc ngSrc} directive. That directive prevents the
-browser from treating the Angular `{{ expression }}` markup literally, and initiating a request to
-an invalid URL `http://localhost:8000/app/{{phone.imageUrl}}`, which it would have done if we had
-only specified an attribute binding in a regular `src` attribute (`<img src="{{phone.imageUrl}}">`).
-Using the `ngSrc` directive prevents the browser from making an http request to an invalid location.
+* We then chained the `filter` filter with the {@link ng.filter:orderBy orderBy} filter to further
```
petebacondarwin: `{@link orderBy}`?

docs/content/tutorial/step_06.ngdoc
```diff
@@ -1,108 +1,239 @@
...
+
+  // Load the module that contains the `phoneList` component before each test
+  beforeEach(module('phoneList'));
+
+  // Test the controller
+  describe('PhoneListController', function() {
+    var ctrl;
+
+    beforeEach(inject(function($componentController) {
+      ctrl = $componentController('phoneList', {$scope: {}});
```
petebacondarwin: No need for the `{$scope: {}}` any more :-)

docs/content/tutorial/step_06.ngdoc
```diff
@@ -1,108 +1,239 @@
...
+        }
+
+        query.sendKeys('tablet'); // Let's narrow the dataset to make the test assertions shorter
+
+        expect(getNames()).toEqual([
+          'Motorola XOOM\u2122 with Wi-Fi',
+          'MOTOROLA XOOM\u2122'
+        ]);
+
+        element(by.model('$ctrl.orderProp')).element(by.css('option[value="name"]')).click();
```
petebacondarwin: Is it worth extracting out the protractor queries here:

```
var orderSelect = element(by.model('$ctrl.orderProp'));
var nameOption = orderSelect.element(by.css('option[value="name"]'));
nameOption.click();
```

docs/content/tutorial/step_06.ngdoc
```diff
@@ -1,108 +1,239 @@
...
 You can now rerun `npm run protractor` to see the tests run.
 
 
 # Experiments
 
-* Replace the `ng-src` directive with a plain old `src` attribute. Using tools such as Firebug,
-or Chrome's Web Inspector, or inspecting the webserver access logs, confirm that the app is indeed
-making an extraneous request to `/app/%7B%7Bphone.imageUrl%7D%7D` (or
-`/app/{{phone.imageUrl}}`).
+<div></div>
```
petebacondarwin: Is this to force some whitespace?

docs/content/tutorial/step_07.ngdoc
```diff
@@ -1,380 +1,311 @@
...
 @description
 
 <ul doc-tutorial-nav="7"></ul>
 
 
-In this step, you will learn how to create a layout template and how to build an app that has
-multiple views by adding routing, using an Angular module called 'ngRoute'.
+Enough of building an app with three phones in a hard-coded dataset! Let's fetch a larger dataset
+from our server using one of Angular's built-in {@link guide/services services} called
+{@link ng.$http $http}. We will use Angular's {@link guide/di dependency injection (DI)} to provide
```
petebacondarwin: `{@link ng.$http}` will format the `$http` as an inline code block. As you have it now the $http will look like normal text. What were you aiming for?

docs/content/tutorial/step_07.ngdoc
```diff
@@ -1,380 +1,311 @@
...
+browser and our app, they both look the same. For the sake of simplicity, we will use JSON files
+in this tutorial.)
 
-When the application bootstraps, Angular creates an injector that will be used to find and inject all
-of the services that are required by your app. The injector itself doesn't know anything about what
-`$http` or `$route` services do. In fact, the injector doesn't even know about the existence of these services
-unless it is configured with proper module definitions.
+The `$http` service returns a {@link ng.$q promise object}, which has a `then()` method. We call
+this method to handle the asynchronous response and assign the phone data to the controller, as a
+model called `phones`. Notice that Angular detected the JSON response and parsed it for us into the
```
petebacondarwin: model -> property ?

docs/content/tutorial/step_07.ngdoc
```diff
@@ -1,380 +1,311 @@
...
+`data` property of the `response` object passed to our callback!
 
-The injector only carries out the following steps :
+Since we are making the assignment of the `phones` model in a callback function, where the `this`
```
petebacondarwin: model -> property

docs/content/tutorial/step_07.ngdoc
```diff
@@ -1,380 +1,311 @@
...
-  'phonecatControllers'
-]);
-
-...
+.component('phoneList', {..., controller: ['$http', function PhoneListController($http) {...}]});
 ```
 
-Notice the second argument passed to `angular.module`, `['ngRoute', 'phonecatControllers']`. This
-array lists the modules that `phonecatApp` depends on.
+From this point onwards, we are going to use the inline method in the tutorial. With that in mind,
```
petebacondarwin: When defining (and annotating) controllers inline, does that make putting methods on the controller's prototype more tricky?

petebacondarwin: Do we want people to start thinking about component controllers more as classes than constructor functions?

docs/content/tutorial/step_07.ngdoc
```diff
@@ -1,380 +1,311 @@
...
+we create the `PhoneListController`, we need to tell the testing harness to expect an incoming
+request from the controller. To do this we:
 
-Because our example app is relatively small, we'll just add all of our controllers to the
-`phonecatControllers` module.
+* Inject the `$httpBackend` service into the `beforeEach()` function. This is a
+  {@link ngMock.$httpBackend mock version} of the service that in a production environment
+  facilitates all XHR and JSONP requests. The mock version of this service allows us to write tests
+  without having to deal with native APIs and the global state associated with them — both of which
+  make testing a nightmare.
```
petebacondarwin: And also overcomes the asynchronous nature of these calls, which slows down unit tests

docs/content/tutorial/step_08.ngdoc
```diff
@@ -1,197 +1,119 @@
...
-    </dl>
-  </li>
-    ...
-  <li>
-    <span>Additional Features</span>
-    <dd>{{phone.additionalFeatures}}</dd>
+...
+<ul class="phones">
+  <li ng-repeat="phone in $ctrl.phones | filter:$ctrl.query | orderBy:$ctrl.orderProp" class="thumbnail">
+    <a href="#/phones/{{phone.id}}" class="thumb">
```
petebacondarwin: the default hash prefix is now `!` so these links should probably be: `#!/phones/{{phone.id}}`

docs/content/tutorial/step_08.ngdoc
```diff
@@ -1,197 +1,119 @@
...
 
+    it('should render phone specific links', function() {
+      var query = element(by.model('$ctrl.query'));
+      query.sendKeys('nexus');
 
-    it('should display nexus-s page', function() {
-      expect(element(by.binding('phone.name')).getText()).toBe('Nexus S');
+      element.all(by.css('.phones li a')).first().click();
+      browser.getLocationAbsUrl().then(function(url) {
+        expect(url).toBe('/phones/nexus-s');
```
petebacondarwin: I think that `getLocationAbsUrl` will actually return something with `http://localhost...` etc at the beginning

docs/content/tutorial/step_09.ngdoc
```diff
@@ -1,144 +1,414 @@
...
 
 <ul doc-tutorial-nav="9"></ul>
 
-In this step you will learn how to create your own custom display filter.
 
-* In the previous step, the details page displayed either "true" or "false" to indicate whether
-certain phone features were present or not. We have used a custom filter to convert those text
-strings into glyphs: ✓ for "true", and ✘ for "false". Let's see what the filter code looks like.
+In this step, you will learn how to create a layout template and how to build an application that
+has multiple views by adding routing, using an Angular module called {@link ngRoute ngRoute}.
```
petebacondarwin: this form of the link will not format the ngRoute as code block. Is that what you want?

docs/content/tutorial/step_09.ngdoc
```diff
@@ -1,144 +1,414 @@
...
+When the application bootstraps, Angular creates an injector that will be used to find and inject
+all of the services that are required by your application. The injector itself doesn't know anything
+about what the `$http` or `$route` services do. In fact, the injector doesn't even know about the
+existence of these services, unless it is configured with proper module definitions.
+
+The injector only carries out the following steps:
+
+* Load the module definition(s) that you specify in your application.
+* Register all Providers defined in these module definition(s).
+* When asked to do so, inject a specified function and any necessary dependencies (services), that
```
petebacondarwin: This doesn't read that well. Perhaps...

> When asked to do so, lazily instantiate services and their dependencies, via their Providers, a parameters to an injectable function.

docs/content/tutorial/step_09.ngdoc
```diff
@@ -1,144 +1,414 @@
...
+Providers are objects that provide (create) instances of services and expose configuration APIs,
+that can be used to control the creation and runtime behavior of a service. In case of the `$route`
+service, the `$routeProvider` exposes APIs that allow you to define routes for your application.
+
+<div class="alert alert-warning">
+  **Note:** Providers can only be injected into `config` functions. Thus you could not inject
+  `$routeProvider` into `PhoneListController` at runtime.
+</div>
+
+Angular modules solve the problem of removing global state from the application and provide a way
```
petebacondarwin: global variables ?

docs/content/tutorial/step_09.ngdoc
```diff
@@ -1,144 +1,414 @@
...
+any modification.
+
+<br />
+**`karma.conf.js`:**
 
 ```js
-describe('filter', function() {
+    files: [
+      'bower_components/angular/angular.js',
+      'bower_components/angular-resource/angular-resource.js',
```
petebacondarwin: should be

```
'bower_components/angular-resource/angular-route.js',
```

right?

docs/content/tutorial/step_09.ngdoc
```diff
@@ -1,144 +1,414 @@
...
-      expect(checkmarkFilter(true)).toBe('\u2713');
-      expect(checkmarkFilter(false)).toBe('\u2718');
-    }));
+```js
+  ...
+
+  it('should redirect `index.html` to `index.html#/phones', function() {
+    browser.get('index.html');
+    browser.getLocationAbsUrl().then(function(url) {
+      expect(url).toBe('/phones');
```
petebacondarwin: absolute URL?

docs/content/tutorial/the_end.ngdoc
```diff
@@ -9,11 +9,18 @@ previous steps using the `git checkout` command.
...
 
-When you are ready to start developing a project using Angular, we recommend that you bootstrap
-your development with the [angular-seed](https://github.com/angular/angular-seed) project.
+When you are ready to start developing a project using AngularJS, we recommend that you bootstrap
+your development with the [angular-seed project][angular-seed].
 
-We hope this tutorial was useful to you and that you learned enough about Angular to make you want
-to learn more. We especially hope you are inspired to go out and develop Angular web apps of your
-own, and that you might be interested in {@link misc/contribute contributing} to Angular.
+We hope this tutorial was useful to you and that you learned enough about AngularJS to make you want
```
petebacondarwin: Angular

docs/content/tutorial/the_end.ngdoc
```diff
@@ -9,11 +9,18 @@ previous steps using the `git checkout` command.
...
+to learn more. We especially hope you are inspired to go out and develop Angular web applications of
+your own, and that you might be interested in {@link misc/contribute contributing} to AngularJS.
```
petebacondarwin: I think we should use Angular throughout

docs/content/tutorial/step_14.ngdoc
```diff
@@ -0,0 +1,565 @@
...
+
+<ul doc-tutorial-nav="14"></ul>
+
+
+[bower]: http://bower.io/
+[caniuse-css-animation]: http://caniuse.com/#feat=css-animation
+[caniuse-css-transitions]: http://caniuse.com/#feat=css-transitions
+[jquery]: https://jquery.com/
+[jquery-animate]: https://api.jquery.com/animate/
+[webplatform-animations]: https://docs.webplatform.org/wiki/css/properties/animations
```
petebacondarwin: I assume that you didn't make any significant changes to this module?

docs/content/tutorial/step_13.ngdoc
```diff
@@ -0,0 +1,321 @@
...
+
+<br />
+**`app/core/phone/phone.module.js`:**
+
+```js
+angular.module('core.phone', ['ngResource']);
+```
+
+<br />
+**`app/core/phone/phone.service.js`:**
```
teropa: Should it be `phone.factory.js`?

petebacondarwin: IMO it should be `service` since that it what it is, in the end. The fact that we just the `factory()` helper to define the service is an implementation detail that it not what we are trying to convey here.

teropa: OK, makes sense. johnpapa/angular-styleguide uses `.factory.js` a lot but I think I actually like this more.

docs/content/tutorial/index.ngdoc
gkalpak: I thought about it, but (afaik) we would need a "build" step that copies the appropriate files from `node_modules/` into `app/`. So, it decided it would be preferrable to keep things simple.

docs/content/tutorial/index.ngdoc
gkalpak: Fixed

docs/content/tutorial/step_00.ngdoc
gkalpak: The version of `http-server` we are using now also supports opening the browser.
Since we are not using HTML5 mode in the tutorial (I would rather keep it that way, in order to avoid discussing one more concept (server-side configuration for HTML5 mode)), I don't think `lite-server` offers any advantage over `http-server`. We _could_ switch, but we need to also update `angular-phonecat` and `angular-seed`, so I would avoid it if there is no compelling reason.

How about we do it in a subsequent PR ?

docs/content/tutorial/step_00.ngdoc
gkalpak: `<link>` (and ``) are void elements, which according to [the spec][1] are allowed to have a `/` (it's optional for HTML 5). I guess it's a remnant from my XHTML days, but I find it more readable with the slash, so I always use it.

I'm fine removing it, if you feel strongly about it.

[1]: https://www.w3.org/TR/html5/syntax.html#syntax-start-tag

docs/content/tutorial/step_01.ngdoc
gkalpak: Same as above.

docs/content/tutorial/step_02.ngdoc
gkalpak: At first, I styled them as code, but it turned out it was very hard to distinguish clickable code (links) from plain code, so I switched to rely on non-code link styling for clickable/linkable stuff. I will use the simpler form though (`{@link ngRepeat ngRepeat}`).

docs/content/tutorial/step_03.ngdoc
gkalpak: Added

docs/content/tutorial/step_03.ngdoc
gkalpak: Removed

docs/content/tutorial/step_03.ngdoc
gkalpak: Removed !

docs/content/tutorial/step_04.ngdoc
gkalpak: Fixed

docs/content/tutorial/step_04.ngdoc
gkalpak: Hm...based on (unofficial) resources like [this][1], [this][2] or [this][3], I'd say it's definitely not a Windows thing - or at least that `folder` is definitely a Windows thing :stuck_out_tongue: 
In addition, based on the fact that we have CWD (current working directory) and CD (change directory) and not CWF of CF, my understanding is that `directory` is the technically correct term (especially when dealing with command line stuff).

But, I'm actually not too confident on that :stuck_out_tongue:
If you think people will be more familiar with the term `folder`, I'm fine changing it.

[1]: http://superuser.com/questions/169457/directory-vs-folder
[2]: http://superuser.com/questions/187900/what-is-the-difference-between-a-directory-and-a-folder
[3]: http://stackoverflow.com/questions/1191381/english-ui-terminology-directory-or-folder

docs/content/tutorial/step_04.ngdoc
gkalpak: The de facto [styleguide][1] for ng1 uses `core`, but states that you can choose the name that makes sense to you (e.g. common, shared). Based on recent work on `angular-cli` (which is supposed to be based on the new (unreleased) styleguide), I think `shared` will be the new recommendation.

I chose `core` because that is what was used by @teropa and the guide on [angular.io](https://angular.io).
I don't think it makes much of a difference anyway. Anything is better than `components` at this point :stuck_out_tongue: 

I've added a mention to `shared`.

[1]: https://github.com/johnpapa/angular-styleguide/tree/master/a1#naming

docs/content/tutorial/step_04.ngdoc
gkalpak: In this listings (there are more below), I show the current structure after a specific refactoring step (e.g. move feature into own directory, make it its own module, use external template etc).
But even in the final structure shown below, there is no `core/` directory yet - it is createdin the next step, when we create a "shared" feature (the checkmark filter).

I think it would be confusing to show a directory (or folder :stuck_out_tongue:) that we haven't created yet, or to create a directory that we don't need. I can still add it if you feel strongly about it :stuck_out_tongue:

docs/content/tutorial/step_04.ngdoc
gkalpak: Fixed

docs/content/tutorial/step_04.ngdoc
gkalpak: Changed

docs/content/tutorial/step_04.ngdoc
gkalpak: SGTM - changed

docs/content/tutorial/step_04.ngdoc
gkalpak: Yes, this is also the current style. At the top of each step, a highlevel overview of the changes are presented along with a link for the full diff on github and then below, we exlain what we did and highlight the most important changed (thus the past tense).

docs/content/tutorial/step_04.ngdoc
gkalpak: Changed

docs/content/tutorial/step_04.ngdoc
gkalpak: It's because there isn't one yet (see https://github.com/angular/angular.js/pull/14416/files/0234f66cdb8703393ab24bffc49a467b48e912e7#r61547667).

docs/content/tutorial/step_04.ngdoc
gkalpak: Fixed

docs/content/tutorial/step_04.ngdoc
gkalpak: Fixed

docs/content/tutorial/step_05.ngdoc
gkalpak: In this case, we need the `ng.filter:` part, because there is also `guide/filter`, so dgeni doesn't know what to pick for `filter`. (And as mentioned above, I chose not to format links as code for "distinguishability".)

docs/content/tutorial/step_05.ngdoc
gkalpak: Good point. (Since `bindings` sounded more narrow-scoped than intended,) I changed it to: 

> can't easily test templates, DOM manipulation or interoperability of components and services

WDYT ?

docs/content/tutorial/step_05.ngdoc
gkalpak: The glass is half-full, right :smiley:
Changed

docs/content/tutorial/step_05.ngdoc
gkalpak: I changed the whole message to:

>   **Note:** In order for protractor to access and run tests against your application, it must be
  served via a web server. In a different terminal/command line window, run `npm start` to fire up
  the web server.

WDYT ?

docs/content/tutorial/step_06.ngdoc
gkalpak: Removed the `ng.filter:` prefix.

docs/content/tutorial/step_06.ngdoc
gkalpak: Removed !

docs/content/tutorial/step_06.ngdoc
gkalpak: Done

docs/content/tutorial/step_06.ngdoc
gkalpak: `<ul>` right after a heading (`# Experiments`) have some different styling. Inserting a `<div>` to avoid `h1 + ul` selectors (iirc). Not ideal, I know :smiley:

docs/content/tutorial/step_07.ngdoc
gkalpak: Normal (link) text :smiley:

docs/content/tutorial/step_07.ngdoc
gkalpak: This is what they are called in the current version of the tutorial (I guess trying to make the connection with the MVC paradigm more obvious).
Changed it to `property`.

docs/content/tutorial/step_07.ngdoc
gkalpak: Changed.

docs/content/tutorial/step_07.ngdoc
gkalpak: I don't think annotation style makes much difference. Inline definition does indeed make it more difficult to put stuff on the prototype, but refactoring them is trivial as well: E.g.:

```js
.component('myComp', {
  templateUrl: '...',
  bindings: {...},
  controller: function MyCompController(...) {
    this.someProp = ...;
    this.someMethod = ...;
    ...
  }
})

// to

.component('myComp', {
  templateUrl: '...',
  bindings: {...},
  controller: MyCompController
})

function MyCompController(...) {
  this.someProp = ...;
  ...
}
MyCompController.prototype.someMethod = ...;
...
```

--
Personally, I don't see much benefit in using the prototype for controllers and services. And by the time people get to bother with the difference of defining methods in the constructor vs on the prototype, classes will be all over the place :smiley:

I like how the inline definition looks much more similar to a class (with TypeScript annotation in ng2). I.e.:

```js
.component('myComp', {
  templateUrl: '...',
  bindings: {...},
  controller: function MyCompController(...) {
    this.someProp = ...;
    this.someMethod = ...;
    ...
  }
})

// looks at lot like:

@Component(
  selector: 'myComp',
  templateUrl: '...',
  inputs: [...]
}
class MyCompController {
  this.someProp = ...;
  someMethod() {...}
  ...
}
```

--
Most importantly, for a tutorial, I believe simpler is better.
(Plus, I could definitely do without having to exlain JS prototypal inheritance :grin:)

docs/content/tutorial/step_07.ngdoc
gkalpak: Added

docs/content/tutorial/step_08.ngdoc
gkalpak: It's not the default on `1.5.x` that the tutorial is using. Something for the next major tutorial update once `1.6.0` is out :stuck_out_tongue:

docs/content/tutorial/step_08.ngdoc
gkalpak: Based on the fact that the test passes, it doesn't :stuck_out_tongue: 
I think there might be a bug in protractor: https://github.com/angular/protractor/blob/7b0302c952f200231957c3857050e0fa8d5b09d3/lib/clientsidescripts.js#L651-L658

If there is `angular.getTestability()`, is returns `$$tstability`'s `getLocation()` (which returns `$location.url()`).
If there isn't `angular.getTestability()`, then it returns `$locaton.absUrl()`. I don't know under what circumstances there might be no `angular.getTestability()` though.

gkalpak: It seems that it's been like this since Protractor `v1.1.1` (unless we change what `$$testability.getLocation()` returns in core): https://github.com/angular/protractor/commit/0e9c2f49dbe8cdbe56a2ce270f504e26cea0a111

gkalpak: No, it's been like this since the beginning of `$$testability`: https://github.com/angular/angular.js/commit/85880a64900fa22a61feb926bf52de0965332ca5

gkalpak: I submitted angular/protractor#3185

docs/content/tutorial/step_09.ngdoc
gkalpak: Yup :smiley:

docs/content/tutorial/step_09.ngdoc
gkalpak: Changed

docs/content/tutorial/step_09.ngdoc
gkalpak: "State" sounds cooler, but changed it anyway :stuck_out_tongue:

docs/content/tutorial/step_09.ngdoc
gkalpak: See https://github.com/angular/angular.js/pull/14416/files/0234f66cdb8703393ab24bffc49a467b48e912e7#r61557295.

docs/content/tutorial/step_09.ngdoc
gkalpak: Right :smiley:

docs/content/tutorial/step_14.ngdoc
gkalpak: Define "significant" :stuck_out_tongue: 
IIRC I changed the wording here and there, added some experiments and removed CSS prefixes that aren't necessary when using latest versions of browsers (I left a note that older browsers might need prefixes).

docs/content/tutorial/the_end.ngdoc
gkalpak: I used `AngularJS` when I felt that the framework as a whole was referred to or where a distinction with Angular2 (or AngularDart) should be made and used just `Angular` in all other places. (FWIW, I have thought about every occurence and made a conscious choice for `Angular` vs `AngularJS` - "conscious" doesn't necessarily mean "most appropriate" :stuck_out_tongue: )

If you think it's better to use `Angular` thoughout, I'm fine as well.

docs/content/tutorial/the_end.ngdoc
gkalpak: See above.

docs/content/tutorial/step_03.ngdoc
gkalpak: I don't think it is explicitly mentioned anywhere, but this seems to be what john's styleguide is going throughout.
This is how I do it too fwiw :stuck_out_tongue:

Pros and cons of one module per file:

> you don't have to sort the files when building a distribution file

:+1: 

> you can easily move a file to another project without having to bring along its module definition

The idea is that you are creating a module for the "least viable feature" - i.e. the smaller chunk of functionality that does something worth moving around or re-using in another project. So, there is no point in copying just the controller file of a component over to another project - you would copy the whole component directory instead(including the controller, controller spec, the module definition, the template etc).

> Too many modules

:-1:

> I don't think it's promoted as a good practice anywhere (at least I haven't seen it), so there might be few projects following this convention out there (I'm just speculating here).

:-1: 

--
Could go either way (but only because I'm too tempted by not having to sort the files when bundling/testing :stuck_out_tongue: )
I also have a slight concern about the overhead of creating a module for everything. I think there is a (tiny) bit of work involved in loading a module, which could add up.

Leaving it up to you :grin:

docs/content/tutorial/step_04.ngdoc
gkalpak: I added a note.

docs/content/tutorial/index.ngdoc
gkalpak: Added

docs/content/tutorial/the_end.ngdoc
petebacondarwin: It's just that this is what I was told to do by a Googler... possibly Igor or Naomi.

gkalpak:
I added three more commits:

- 9e13bd0: Merged a commit from @petebacondarwin that added SVG format for images.
- d68265f: Updated the tutorial images to match the new content of the tutorial (also fixed up and re-instated 3 hidden images).
- 0002c05: Addressed feedback/comments on the original commit. The equivalent changes have been made to  angular/angular-phonecat#326 (where necessary).

@petebacondarwin, PTAL :smiley:

docs/content/tutorial/step_00.ngdoc
petebacondarwin: OK stick with `http-server`

docs/content/tutorial/index.ngdoc
petebacondarwin: We are going to add a `postinstall` hook to copy angular files from `node_modules` to `app/lib`

docs/content/tutorial/step_04.ngdoc
petebacondarwin: OK

docs/content/tutorial/step_04.ngdoc
petebacondarwin: `core` is fine

docs/content/tutorial/step_04.ngdoc
petebacondarwin: So you went with directory throughout :-) OK

docs/content/tutorial/step_03.ngdoc
petebacondarwin: Ohhh! It's so hard to decide...

I am generally not in favour of reopening anything. Especially old milk bottles!

So I think I would rather go with one module per file:

* I don't believe that there is not so much of a performance penalty, because there is a similar cost to looking up the module when reopening.
* I believe that having small modules that explicitly declare their dependencies makes them a bit more self documenting.
* My feeling is that if things are so tightly bound that they should be in a single module then they should be in the same file??

If you do feel strongly @gkalpak  then don't do it. But if you are not bothered then I would be delighted to see this changed :-)

docs/content/tutorial/step_04.ngdoc
petebacondarwin: OK!

docs/content/tutorial/step_05.ngdoc
petebacondarwin: OK

docs/content/tutorial/step_06.ngdoc
petebacondarwin: OK

docs/content/tutorial/step_07.ngdoc
petebacondarwin: OK. Sounds good!

docs/content/tutorial/step_08.ngdoc
petebacondarwin: Oh! Right.
Perhaps we should set the value to `"!"` in a config block in this tutorial?

docs/content/tutorial/step_08.ngdoc
petebacondarwin: OK. So if that issue gets resolved will we have to change our tests here?

docs/content/tutorial/step_09.ngdoc
petebacondarwin: There is plenty of global state in an Angular app even without global variables :-)
So you've still got plenty of cool

docs/content/tutorial/the_end.ngdoc
petebacondarwin: Leave them as you have them and see if anyone on the Google team notices :-) (and complains)

docs/content/tutorial/the_end.ngdoc
petebacondarwin: Seen

petebacondarwin:
@gkalpak OK, so I have run through all our comment conversations.
I will do a walk through of the actual tutorial tomorrow and see if it all makes sense.
Thanks for all your work on this!

docs/content/tutorial/index.ngdoc
gkalpak: We decided to go for npm + a `postinstall` script that copies the appropriate files.

docs/content/tutorial/step_08.ngdoc
gkalpak: If it gets resolved it will be a BC for Protractor (so it has to come with a major version update).
If we update to that major version, then yes, we will need to update our tests.

petebacondarwin:
I found the bit about `git://` and proxies. Great! I think it would benefit from a bit of a heading or title in the alert box, to make it easier to find if you are having problems. Perhaps : **Firewall/Proxy Issues**?

petebacondarwin:
It may be worth putting another alert box after the first `npm start` explaining what the error message means if you already have a server running on port 8000

petebacondarwin:
It might be worth saying that updating webdriver can take quite a long time (minutes on my machine) and if you Ctrl-C out of it then you get weird errors when trying to run protractor later.

docs/content/tutorial/step_00.ngdoc
petebacondarwin: Although lite-server also offers live-reloading :-)

petebacondarwin:
In step-5 the protractor tests fail because they are trying to access `localhost:8000/index.html` whereas the correct url is `localhost:8000/app/index.html`.

I believe the simplest fix is to change the `browser.get` to:

```
    beforeEach(function() {
      browser.get('app/index.html');
    });
```

Another alternative fix is to change the baseUrl in the protractor config:

```
  baseUrl: 'http://localhost:8000/app/',
```

petebacondarwin:
I take that last comment back. I see that `npm start` has changed to run the local web server from the `app` folder. So this is fine

petebacondarwin:
In step-6 we need to change the headings from "template" and "controller" to "component template" and "component controller" in keeping with step-5

petebacondarwin:
Similarly in step-7 the "controller" heading should change to "component", I believe

petebacondarwin:
And for subsequent steps

petebacondarwin:
By the way I love the new images but perhaps the gradient fills are a bit cheesy?

petebacondarwin:
The next step button on step-12 takes you to the end.

docs/content/tutorial/step_00.ngdoc
gkalpak: I have lready added this to my post-merge tasks :)

gkalpak:
I added a new commit that should address all above comments except for the image gradient cheesy-ness (because I am not sure what it means). Is it [like this][1] or like too yellow ?

[1]: http://el.urbandictionary.com/define.php?term=cheesy

docs/content/tutorial/index.ngdoc
gkalpak: In order to avoid the complexity of having a node script that uses some additional module (e.g. `fs-extra`) in order to copy the files etc, we decided to move `index.html` to the root and reference the files inside `node_module/` directly.

petebacondarwin:
By cheesy I meant a bit 1980s in style. I realise that you were following the form of the current app.
But I think it would benefit with a flatter look.

docs/content/tutorial/step_05.ngdoc
```diff
@@ -141,6 +141,12 @@ To rerun the test suite, execute `npm run protractor` again.
   the web server.
 </div>
 
+<div class="alert alert-warning">
+  The web server is configured to use port 8000. If the port is already in use (for example by
+  another instance of a running web server) you will get an `EADDRINUSE` error. Make sure the port 
+  is available before running `npm start`.
+</div>
+
```
petebacondarwin: I think this warning should also appear in the introduction when setting up the environment. At least that is where I encountered this error.

gkalpak: I put it here, because we already have 8-9 alerts in the introduction. I can move it there if think it's better. (One more won't hurt I guess :smiley:)

gkalpak: Do you think would be better to move all the warnings in the introduction in a separate sections (at the bottom of introduction).

petebacondarwin: That might work. A kind of setup troubleshooting section?

gkalpak:
Added a couple more fixups that move warning in a "Common Issues" section (in the introduction) and use `!` as hash-prefix (angular-phonecat and angular-seed have been updated accordingly).
/cc @petebacondarwin

gkalpak:
Tasks for after getting this is:

- [ ] Make appropriate modifications to the 1.4.x tutorial (see https://github.com/angular/angular.js/pull/14416#issuecomment-208844996).
- [ ] Use a more flat look in images - remove gradient.
- [ ] Switch to `lite-server` (with auto-reloading).

petebacondarwin:
LGTM

PR #14424, opened 2016-04-13
revert: "fix(ngRoute): allow `ngView` to be included in an asynchronously loaded template"

### LINKED ISSUE #14337: Default route gets resolved in test code when calling $httpBackend.flush() or scope.$apply()

Report a breaking change (or a bug).

When upgrading from 1.5.0 to 1.5.1 or newer, calling `$httpBackend.flush()` or `scope.$apply()` in test code causes the default route to resolve automatically. This can lead unexpected requests and break the tests. This seems to be due to change in [ngRoute](https://github.com/angular/angular.js/commit/8237482d49e76e2c4994fe6207e3c9799ef04163).

**If the current behavior is a bug, please provide the steps to reproduce**

Can be reproduced with for example [AngularJS Phonecat demo](https://github.com/angular/angular-phonecat) by updating bower.json to use Angular 1.5.1 and running the tests.

gkalpak:
Bug fix (regression).

Eagerly loading `$route`, could break tests, because it might request the root or default route
template (something `$httpBackend` would know nothing about).
See #14337.

Tests don't break.
Also, this was added as a known issue in `ngView` docs.

No

**Other information**:
This reverts commit 5e37b2a.

It will be re-applied for `v1.6.x`, with a breaking change notice and possibly a way to disable
the feature is tests.

Fixes #14337

### DIFF
--- a/src/ngRoute/directive/ngView.js
+++ b/src/ngRoute/directive/ngView.js
@@ -26,11 +26,12 @@ ngRouteModule.directive('ngView', ngViewFillContentFactory);
  *
  * The enter and leave animation occur concurrently.
  *
- * @knownIssue If `ngView` is contained in an asynchronously loaded templated (i.e. not directly
- *             inside your main HTML file, then you need to make sure that `$route` is instantiated
- *             in time to capture the initial `$locationChangeStart` event and load the appropriate
- *             view. One way to ensure `$route` will be instantiated in time, it to have it as a
- *             dependency in a `.run` block: `myModule.run(['$route', function() {}]);`
+ * @knownIssue If `ngView` is contained in an asynchronously loaded template (e.g. in another
+ *             directive's templateUrl or in a template loaded using `ngInclude`), then you need to
+ *             make sure that `$route` is instantiated in time to capture the initial
+ *             `$locationChangeStart` event and load the appropriate view. One way to achieve this
+ *             is to have it as a dependency in a `.run` block:
+ *             `myModule.run(['$route', function() {}]);`
  *
  * @scope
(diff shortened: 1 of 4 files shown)


src/ngRoute/directive/ngView.js
```diff
@@ -26,6 +26,12 @@ ngRouteModule.directive('ngView', ngViewFillContentFactory);
  *
  * The enter and leave animation occur concurrently.
  *
+ * @knownIssue If `ngView` is contained in an asynchronously loaded templated (i.e. not directly
```
Narretz: template~~d~~
And I would list the usual use-cases instead  (e.g. in an ngInclude or in a directive's templateUrl)

src/ngRoute/directive/ngView.js
```diff
@@ -26,6 +26,12 @@ ngRouteModule.directive('ngView', ngViewFillContentFactory);
...
+ *             inside your main HTML file, then you need to make sure that `$route` is instantiated
+ *             in time to capture the initial `$locationChangeStart` event and load the appropriate
+ *             view. One way to ensure `$route` will be instantiated in time, it to have it as a
```
Narretz: it to have -> is to have

src/ngRoute/directive/ngView.js
gkalpak: Done

src/ngRoute/directive/ngView.js
gkalpak: Done

Narretz:
LGTM

PR #14428, opened 2016-04-13
docs($compileProvider): document 'require' for angular.Component

ademuri:


ademuri:
This commit has the wrong email for my CLA - I'll try again in a minute.

PR #14429, opened 2016-04-13
docs($compileProvider): document 'require' in angular.Component

ademuri:
jsdoc update

No

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1047,6 +1047,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    *      See {@link ng.$compile#-bindtocontroller- `bindToController`}.
    *    - `transclude` – `{boolean=}` – whether {@link $compile#transclusion content transclusion} is enabled.
    *      Disabled by default.
+   *    - `require` - `{Object<string, string>=}` - requires the controllers of other directives and binds them to
+   *      this component's controller. The object keys specify the property names under which the required
+   *      controllers (object values) will be bound. See {@link ng.$compile#-require- `require`}.
    *    - `$...` – additional properties to attach to the directive factory function and the controller
    *      constructor function. (This is used by the component router to annotate)
    *


ademuri:
Note: I was unable to build the doc to test that it works, because the build appears to already be broken:

$ grunt package --verbose
Initializing
Command-line options: --verbose

Reading "Gruntfile.js" Gruntfile...OK

<...snip.../>

Running "docs" task
[14:08:13] Working directory changed to ~/angular/angular.js/docs
[14:08:14] Using gulpfile ~/angular/angular.js/docs/gulpfile.js
[14:08:14] Starting 'bower'...
[14:08:14] Starting 'build-app'...
[14:08:14] Finished 'bower' after 372 ms
[14:08:14] Starting 'assets'...
[14:08:15] Starting 'doc-gen'...
error:   Error processing docs:  TypeError: object is not a function
    at Function.v.extend.async (/usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/node_modules/validate.js/validate.js:196:14)
    at /usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/lib/Dgeni.js:166:23
    at _fulfilled (/usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/node_modules/q/q.js:834:54)
    at self.promiseDispatch.done (/usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/node_modules/q/q.js:863:30)
    at Promise.promise.promiseDispatch (/usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/node_modules/q/q.js:796:13)
    at /usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/node_modules/q/q.js:857:14
    at runSingle (/usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/node_modules/q/q.js:137:13)
    at flush (/usr/local/google/home/ademuri/angular/angular.js/node_modules/dgeni/node_modules/q/q.js:125:13)
    at process._tickCallback (node.js:415:13)
Warning: doc generation failed Use --force to continue.

Aborted due to warnings.

src/ng/compile.js
```diff
@@ -1047,6 +1047,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    *      See {@link ng.$compile#-bindtocontroller- `bindToController`}.
    *    - `transclude` – `{boolean=}` – whether {@link $compile#transclusion content transclusion} is enabled.
    *      Disabled by default.
+   *    - `require` - `{!Object|undefined}` - requires the controllers of other directives and binds them to
```
gkalpak: Why `!Object|undefined` instead of `object=` (as other parameters above) ?
Besides, it can be a string or an array as well.

ademuri:
Changed to use '=' instead of '|undefined'.

Use 'Object' since 'object' isn't a real Closure type.

gkalpak:
Yeah, casing was not my point. I thought both work (I used `object` for consistency).
In any case, Object is not the only permitted type. It can also be either a string or an Array.

ademuri:
I don't think it makes sense to pass a string or array here. For a directive, those are used to specify controllers that get injected into the link function. For a component, there is no link function. The object is used to specify a mapping from the required controller to the name that it is bound to on the component's controller.

See https://docs.angularjs.org/guide/component

gkalpak:
I thought you could still use a string or Array to specify depenencies (as in: only use this component inside a XYZ directive etc). As it turns out, these restrictions are only enforced if the required controllers are requested (e.g. in a link function or in controllers for `require: {}`).
Since components can't have a link function, string or Array `requires` are passed to the DDO, but have no effect, so they are essentially ignored.

LGTM then :+1:

Issue #14426, opened 2016-04-13
Unexpected behavior in unit-tests using $componentController and &-bindings to pass data


mjwwit:
**Do you want to request a *feature* or report a *bug*?**
Bug with using `$componentController`, `&`-bindings, and passing arguments.

**What is the current behavior?**
Using `$componentController` to instantiate the controller of a component that has an `&`-binding used to pass data back to its user results in unexpected argument data being passed to the function.

***Example***
Given:
```javascript
  angular.module('heroApp', [])
    .component('heroDetail', {
      template: '<div></div>',
      bindings: {
        hero: '<',
        onDelete: '&'
      }
    });
```
The current behavior using `$componentController` is:
```javascript
    it('should call the onDelete binding when a hero is deleted', function() {
      var deleteSpy = jasmine.createSpy('deleteSpy');
      component = $componentController('heroDetail',
        {$scope: scope},
        {
          hero: {name: 'Wolverine'},
          onDelete: deleteSpy
        }
      );

      component.onDelete({hero: component.hero});
      // Spy is called with object map, instead of just the "hero" value
      expect(deleteSpy).toHaveBeenCalledWith({hero: component.hero});
    });
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**
Here is a minimal demo, comparing `$componentController` and `$compile` to unit-test the `onDelete` binding: [Demo](http://jsbin.com/nolaga/3/edit?js,output)

**What is the expected behavior?**
Given the same "heroDetail" component as above, one would expect the following test to succeed:
```javascript
    it('should call the onDelete binding when a hero is deleted', function() {
      var deleteSpy = jasmine.createSpy('deleteSpy');
      component = $componentController('heroDetail',
        {$scope: scope},
        {
          hero: {name: 'Wolverine'},
          onDelete: deleteSpy
        }
      );

      component.onDelete({hero: component.hero});
      // Spy is called with the object map, instead of just the "hero" value
      expect(deleteSpy).toHaveBeenCalledWith(component.hero);
    });
```
This would make `$componentController`'s behavior consistent with `$compile`'s behavior (and the real world).

**What is the motivation / use case for changing the behavior?**
[...]

PR #14432, opened 2016-04-14
fix($componentController): expressions passed to $componentController '&'-bindings are now parsed

mjwwit:
Bug fix

Behavior is explained in issue #14426.

The new behavior is the ability to pass expressions to `$componentController` `&`-bindings. Example:

``` javascript
  it('should call the onDelete binding when a hero is deleted', function() {
    scope.deleteFn = jasmine.createSpy('deleteSpy');
    component = $componentController('heroDetail',
      {$scope: scope},
      {hero: hero, onDelete: 'deleteFn(hero)'}
    );

    component.onDelete({hero: component.hero});
    expect(spy('deleteSpy')).toHaveBeenCalledWith(component.hero);
  });
```

The expression is parsed in a way similar to the parsing of a binding during the compilation of a component element.

No.

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -484,9 +484,10 @@ describe('component: heroDetail', function() {
   });
 
   it('should call the onDelete binding when a hero is deleted', function() {
+    scope.deleteFn = jasmine.createSpy('deleteSpy');
     component = $componentController('heroDetail',
       {$scope: scope},
-      {hero: hero, onDelete: jasmine.createSpy('deleteSpy')}
+      {hero: hero, onDelete: 'deleteFn(hero)'}
     );
 
     component.onDelete({hero: component.hero});
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2207,7 +2207,7 @@ angular.mock.$ControllerDecorator = ['$delegate', function($delegate) {
  * @return {Object} Instance of requested controller.
  */
 angular.mock.$ComponentControllerProvider = ['$compileProvider', function($compileProvider) {
-  this.$get = ['$controller','$injector', '$rootScope', function($controller, $injector, $rootScope) {
+  this.$get = ['$controller','$injector', '$parse', '$rootScope', function($controller, $injector, $parse, $rootScope) {
     return function $componentController(componentName, locals, bindings, ident) {
       // get all directives associated to the component name
       var directives = $injector.get(componentName + 'Directive');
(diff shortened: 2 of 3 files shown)


mjwwit:
I signed it!

drpicox:
It's not properly a bugFix, its a new feature. And I do not see the point, it makes testing more costly. It is the same for the case that you have exposed:

```javascript
  it('should call the onDelete binding when a hero is deleted', function() {
    component = $componentController('heroDetail',
      {$scope: scope},
      {hero: hero, onDelete:  jasmine.createSpy('deleteSpy')}
    );

    component.onDelete({hero: component.hero});
    expect(spy('deleteSpy')).toHaveBeenCalledWith(component.hero);
  });
```

In addition your solution is not accurate, it does not takes into account what happens about other cases:

```javascript
bindToController: {
  fireSelect: '@onSelect',
},
```

But it should be interesting to add the testing that I have wrote here.

mjwwit:
It's either a bugfix in the angular code or a bugfix for the docs. The case you mention above (the one from the docs) does not work.
```javascript
  it('should call the onDelete binding when a hero is deleted', function() {
    component = $componentController('heroDetail',
      {$scope: scope},
      {hero: hero, onDelete:  jasmine.createSpy('deleteSpy')}
    );

    component.onDelete({hero: component.hero});

    // This expectation will fail, as `deleteSpy` is called with the `$parse` locals context
    expect(spy('deleteSpy')).toHaveBeenCalledWith(component.hero);

    // This expectation will succeed, but is not representative for the way your bound function
    // will be called in the real world
    expect(spy('deleteSpy')).toHaveBeenCalledWith({ hero: component.hero });
  });
```
My fix (or feature, depending on how you look at it) does not necessarily make testing more costly. Only expression bindings are parsed. You can still pass a function reference into the binding directly and skip the parsing.

My fix does fall short when it comes to named bindings, thanks for noticing! I will update the PR to add this.

mjwwit:
I fixed the named binding issue, it now works fine. I also added a test case to prove it.

drpicox:
Indeed docs are wrong:

```js
  it('should call the onDelete binding when a hero is deleted', function() {
    component = $componentController('heroDetail',
      {$scope: scope},
      {hero: hero, onDelete: jasmine.createSpy('deleteSpy')}
    );

    component.onDelete({hero: component.hero});
    expect(spy('deleteSpy')).toHaveBeenCalledWith({hero: component.hero});
  });
```

IMHO I would appreciate a PR to fix docs (it is a pitty that there are not unit tests for the docs). But add a test first to mocksSpec.js, please.

About not to change `$componentController` is that semantics can be a little weird for `<`  and  `=` if we try to make them consistent. Now they are expected to take values as is, but if `&` takes a string and parses it from scope, ¿what to expect from `<` and `=` in such cases? Example:

```javascript
    component = $componentController('colorShowroom',
      { $scope: scope},
      { color: 'red' } // bindings: { color: '<' }
    );
```
So in this case, what is red? ¿the string value expected for color or the value of red inside the `$scope`?

If it is not changed, it might be confusing, some bindings getting parsed and others not.

And even worst, $componentController is analogous to $controller, both have the same interface and both create an instance of a $controller, why one version behaves different from the other? 
If you want to test the component (not the directive), you might do: `$compile('<color-showroom color="foo"></color-showroom>`)($rootScope)`, and no need for compute bindings or other stuff.

gkalpak:
I think the current behavior is better. The docs should be fixed though.
The `bindings` argument or `$componentController`, should be passed the function that gets assigned on the controller. All the parsing (and other safeguards we have in place) that happens as part of parsing a directive's bindings are out of scope here.

When testing the controller, you should be only concerned with the part of the API that interacts with it, in this case, your controller only knows it gets an `onDelete` function and calls it with `{hero: ...}`.
That is what you need to test.

Thx for working on this, but I'm going to close it, since the current behavior is the intended one.
It's a good spot, though, and you are more than welcome to submit a new request with the docs fix.
:+1:

Issue #14206, opened 2016-03-09
Properties set before constructor is called when using bindToController


suedama1756:
**Note: for support questions, please use one of these channels:** https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.

* **Do you want to request a *feature* or report a *bug*?**

BUG

* **What is the current behavior?**

When using bindToController properties are being set before the constructor is invoked. 

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:yBpEi4).

```
bindToController: {
	min: '@',
}

class MyController {
  private _min : number; 
  constructor() {
     this._min = 0;
  }

  get min() {
    return this._min;
  }
 
  set min(value : number) {
     this._min = value;
  } 
}
```
* **What is the expected behavior?**

Constructor should be called before properties are set.

* **What is the motivation / use case for changing the behavior?**

Behaviour in properties setters that rely on values initialized in the constructor fail.

* **Which version of Angular, and which browser and OS does this issue affect?** Did this work in previous
versions of Angular? Please also test with the latest stable and snapshot versions.

Windows 7, TypeScript 1.8.2, chrome 49.0.2623.87, Angular 1.5.0

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix)

suedama1756:
The constructor is called in the following lines  after the properties have been set in nodeLinkFn.

var controllerResult = controller();
if (controllerResult !== controller.instance) {

 The properties setters created using Object.defineProperty are invoked even though the constructor has not yet been invoked. This is problematic if the property setters rely on values initialized in the constructor (in my case throwing exceptions).

gkalpak:
This actually a feature not a bug :)

We are binding the properties, so you can work right away from the constructor with them.
(**Note:** This only works with regular functions not classes (but your TypeScript Class obviously gets transpiled to a plain old ES5 constructor function).)

Maybe you could check if `_min` is undefined before assigning a default value in the constructor.

suedama1756:
With respect, I would suggest breaking the semantics of classes is a BUG and not a feature. As ECMA 6 becomes standard with property getters and setters this type of issue will become more common. It is called a "constructor" after all and controllers in all other areas of angular would have their constructors called before any other interaction (AFAIK). 

[...]

PR #14439, opened 2016-04-14
fix(ngMock#$controller): properly assign bindings to all types of controllers (e.g. class-based)

### LINKED ISSUE #14437: $componentController doesn't bind properties when component uses ES6 controller class
Bug

Bindings set up with $componentController are undefined when using a ES6 class as the controller. See an example at http://jsbin.com/qaxivowuni/1/edit?js,output. Switching from the Es6Controller to the Es5Controller causes the binding to be defined.

The bindings are defined regardless of how the controller class is created.

Unable to use ES6 features.

Angular 1.5 / Chrome 49.0.2623.87 / Mac OS 10.11.4

gkalpak:
Bug fix (I think).

The bindings are not assigned to the final instances of controllers specified as Classes or returning a value from their constructor.
See #14437.

The bindings are assigned to any type of controller, similar to how `$compile` would do it.

Not really.

**Other information**:
The decorated version of `$controller` is able to assign bindings to a controller instance prior to
instantiation, emulating the behavior of `$compile` with directive controllers.

There are cases, that the actual controller instance is different than the pre-populated one (e.g.
when the controller constructor function returns a value or when the controller is an ES2015 Class).
While `$compile` accounts for such situation, by re-asigning the bindings after the controller has
been instantiated, `ngMock`'s `$controller` didn't.

This commit fixes it, by re-applying the bindings if the actual controller instance is different
than the original one (similar to how `$compile` would do it).

Fixes #14437

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2181,9 +2181,15 @@ angular.mock.$RootElementProvider = function() {
 angular.mock.$ControllerDecorator = ['$delegate', function($delegate) {
   return function(expression, locals, later, ident) {
     if (later && typeof later === 'object') {
-      var create = $delegate(expression, locals, true, ident);
-      angular.extend(create.instance, later);
-      return create();
+      var instantiate = $delegate(expression, locals, true, ident);
+      angular.extend(instantiate.instance, later);
+
+      var instance = instantiate();
+      if (instance !== instantiate.instance) {
+        angular.extend(instance, later);
+      }
+
+      return instance;
     }
     return $delegate(expression, locals, later, ident);
   };
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -1,5 +1,7 @@
 'use strict';
 
+/* globals support: false */
+
 describe('ngMock', function() {
   var noop = angular.noop;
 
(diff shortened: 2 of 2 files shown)


Narretz:
Safari 9 doesn't like this:

```
Safari 9.0.0 (Mac OS X 10.11.0) ngMock $controllerDecorator should support assigning bindings to class-based controller FAILED

	TypeError: Cannot call a class constructor (line 1)

	invoke@/home/travis/build/angular/angular.js/src/auto/injector.js:867:24

	$controllerInit@/home/travis/build/angular/angular.js/src/ng/controller.js:147:40

	/home/travis/build/angular/angular.js/src/ngMock/angular-mocks.js:2187:33

	/home/travis/build/angular/angular.js/test/ngMock/angular-mocksSpec.js:1933:33

	invoke@/home/travis/build/angular/angular.js/src/auto/injector.js:867:24

	workFn@/home/travis/build/angular/angular.js/src/ngMock/angular-mocks.js:2965:26

	inject@/home/travis/build/angular/angular.js/src/ngMock/angular-mocks.js:2934:46

	/home/travis/build/angular/angular.js/test/ngMock/angular-mocksSpec.js:1932:15

	inject@/home/travis/build/angular/angular.js/src/ngMock/angular-mocks.js:2927:34

	/home/travis/build/angular/angular.js/test/ngMock/angular-mocksSpec.js:1932:15
```

Narretz:
We can test on 9.1 though, maybe it was fixed by this version( I doubt it)

gkalpak:
Yeah, I saw it failed. I pushed a potential fix, but it failed again and I'm not sure why (I can't test locally on Windows).

I'll try to track yhis down via SauceLabs' manual testing latet (or tomorrow).

Narretz:
Isn't that the same problem with have in compile with Safari?

gkalpak:
No idea :)

Narretz:
I was thinking about this: https://github.com/angular/angular.js/pull/13785

gkalpak:
@Narretz, it seems like it. So, does it mean that `$compile` won't currently work with class-based controllers on Safari 8/9.0.0 ?

gkalpak:
That is indeed the case. I changed the class-based test to only run on Chrome (as is done i the `$compile` tests).

drpicox:
I'm thinking that may be it is also related to: https://github.com/angular/angular.js/issues/14206

Narretz:
LGTM

tuhoojabotti:
I still seem to have this issue with $componentController and angular 1.5.8 and ES6 classes, neither locals or bindings are being set on the controller instance.

mgol:
@tuhoojabotti It works fine for me, I have multiple tests that rely on that. If something's broken for you, please open a new issue.

tuhoojabotti:
@mgol I just realized it's because of my test setup. I do not use ngMock because I can't get past calling flush on every http request, so I have my own mock module that exposes some ngMock providers, but I forgot to add the $controllerDecorator, that's why the bindings aren't being set.

PR #14443, opened 2016-04-15
docs(guide/component): fix for component controller unit-test example

mjwwit:
Docs update.

Error in docs is described in #14426.

No.

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -233,7 +233,7 @@ it upwards to the heroList component, which updates the original data.
 </file>
 
 <file name="heroDetail.js">
-  function HeroDetailController($scope, $element, $attrs) {
+  function HeroDetailController() {
     var ctrl = this;
 
     ctrl.update = function(prop, value) {
@@ -455,7 +455,7 @@ The examples use the [Jasmine](http://jasmine.github.io/) testing framework.
 describe('component: heroDetail', function() {
   var component, scope, hero, $componentController;
 
-  beforeEach(module('simpleComponent'));
+  beforeEach(module('heroApp'));
 
   beforeEach(inject(function($rootScope, _$componentController_) {
     scope = $rootScope.$new();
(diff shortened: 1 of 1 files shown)


docs/content/guide/component.ngdoc
```diff
@@ -233,9 +233,13 @@ it upwards to the heroList component, which updates the original data.
 </file>
 
 <file name="heroDetail.js">
-  function HeroDetailController($scope, $element, $attrs) {
+  function HeroDetailController($scope) {
```
gkalpak: Do we need `$scope` here ? Did you leave it on purpose ?

docs/content/guide/component.ngdoc
```diff
@@ -233,9 +233,13 @@ it upwards to the heroList component, which updates the original data.
...
     var ctrl = this;
 
+    ctrl.hero = ctrl.hero || {};
+    ctrl.hero.name = ctrl.hero.name || undefined;
+    ctrl.hero.location = ctrl.hero.location || 'unknown';
```
gkalpak: Why did you make these changes ? They don't seem necessary.

docs/content/guide/component.ngdoc
mjwwit: I made the absolute minimum number of changes to make the unit tests pass. In the unit tests the `$componentController` is provided with a `$scope` value, but not with `$element` or `$attrs`.
The `$scope` service is not being used in the controller though, so it might as well be removed.

If I'm not wrong, a PR was merged the other day removing the requirement to always provide `$scope` in `$componentController` locals, so it can be removed from there as well.
I can update the PR if you like.

docs/content/guide/component.ngdoc
mjwwit: Those changes were tested in the unit tests. The tests were failing because the actual functionality was missing. I chose to update the component instead of removing the failing test. If needed, I can remove the test and the functionality.

docs/content/guide/component.ngdoc
gkalpak: You are right. Please, do remove the unnecessary `$scope` and update the test to not provide one.

docs/content/guide/component.ngdoc
gkalpak: Really ? It's a shame we are not actually running these tests  :disappointed:

mjwwit:
I've removed the `$scope` references, and removed an outdated comment.

Isn't there some way you can get the unit tests to automatically run from docs using dgeni? It's already parsing the code snippets and putting them into files. Maybe it's possible to generate a karma config and run tests as well.

gkalpak:
We are running the e2e tests from the docs. Not sure if we are already doing the same for the guide.
Being able to run unit tests would be nice, but it won't be trivial I guess.

docs/content/guide/component.ngdoc
gkalpak: TBH, I would rather remove this initialization and the corresponding test.
It doesn't add much value and doesn't seem like a good practice either. If nothing else, such initialization should be happening in an `$onInit` hook ?

WDYT ?

docs/content/guide/component.ngdoc
```diff
@@ -464,8 +468,7 @@ describe('component: heroDetail', function() {
   }));
 
   it('should set the default values of the hero', function() {
-    // It's necessary to always pass the scope in the locals, so that the controller instance can be bound to it
-    component = $componentController('heroDetail', {$scope: scope});
+    component = $componentController('heroDetail', {});
```
gkalpak: Nitpick: I prefer to omit the unnecessary `locals` argument altogether (or pass `null` where we want a 3rd argument).

gkalpak:
Except for the fact that I am not a big fan of the initialization logic (see https://github.com/angular/angular.js/pull/14443/files#r60963112), it LGTM.
Feel free to also remove unnecessarily injected arguments to other controllers.

docs/content/guide/component.ngdoc
mjwwit: I agree. I'll remove the test and the logic.

PR #14447, opened 2016-04-15
chore(gitignore): exclude .log and .stackdump files

Narretz:
Ignore those pesky .log files that are often generated by node and .stackdump, which is created by ConEmu.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ npm-debug.log
 /tmp/
 /scripts/bower/bower-*
 .vscode
+*.log
+*.stackdump


gkalpak:
Besides `npm-debug.log` what other node-related `.log` files are there ?
I don't remember seeing any.

Narretz:
It's possible that they are coming from ConEmu too. I just deleted them all, so I can't say for sure :)

gkalpak:
LGTM (I guess there is no .log that we would want to keep, wherever it's coming from :smiley:)

Issue #10859, opened 2015-01-25
Grunt tasks slow due to synchronous network calls


pkozlowski-opensource:
Our build is, unfortunately, not lightening-fast and part of the problem is [that line](https://github.com/angular/angular.js/blob/9f7c5ceba74a5e22a706b38dc7f538b6199cb89b/Gruntfile.js#L5) that is executed on each and every Grunt run. This line will trigger call a network call to Google's CDN every time, while probably only needed during the release time.

We should see which expensive operations could deferred to the very last moment (that is, executed as part of a task that needs them instead of executing those operations systematically)

Narretz:
We've introdcued a flag to disable this call: https://github.com/angular/angular.js/commit/6a134605f1904098022f4e25ade67d0a91a52621
Turns out it's not that time intensive anyway

PR #14458, opened 2016-04-18
chore(grunt): only get versions when actually needed

Narretz:
chore

See #10859


petebacondarwin:
It feels kind of awkward that the utility is writing back into the config.
Can't we make the calls to get the version info lazy, so that they only run expensive requests when they are called and then cache them for future calls?

petebacondarwin:
Closing for now as we have a flag for blocking these requests.

Issue #11945, opened 2015-05-26
$http service: specification of `status` in error case


jgehrcke:
https://docs.angularjs.org/api/ng/service/$http clearly states that the `status` argument to the `success` handler is between 200 and 299. In all other cases, the `error` handler is called. However, there is no statement about the values that `status` can assume within the error handler. I guess that it is either `0` (if no back-end response arrived, for whatever reason) or the HTTP status indicated by the back-end. I hope no other values are allowed (such as `null`). Is that correct? Is that guaranteed? If it is, it should be documented, correct?

Thanks,

Jan-Philip

gkalpak:
I think, the only guarantee is that (if it's a number) it will be <= 0. Other than that, it comes directly from the `XMLHttpRequest`'s `status` property, which [according to MDN][1] should be an _unsigned short_, but I am not sure what some weird backend might do (and what browsers would accept).

It's extremely unlikely that you get a non-numeric value (but it's not something Angular-specific really).

[1]: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#xmlhttprequest-status

jgehrcke:
You meant write `>= 0`, right?

So, if AngularJS guarantees to not touch and just forward `XMLHttpRequest.status`, then things are clear enough. The behavior of this flag is specified [here](http://www.w3.org/TR/XMLHttpRequest/#the-status-attribute
): 
```text
The status attribute must return the result of running these steps:

    If the state is UNSENT or OPENED, return 0.

    If the error flag is set, return 0.

    Return the HTTP status code.
```

In other words, one can safely expect this to be an integer, and checking for 0 is a useful way to catch all errors that were not produced by the back-end itself.

gkalpak:
@jgehrcke, yeah I meant `>= 0`. 

<sub>BTW, I kind of lied, Angular does a tiny little manipulation of `XMLHttpRequest.status`: It passes it through `Math.max(status, 0)` (in order to _"normalize internal statuses"_).</sub>
<sub>This has the side-effect that `status` is parsed as number, so things might get weird if the backend returns a non-numeric status value. (But if you have a non-number-status-returning backend, things are weird already, so...)</sub>

--
**Bottom line:**
It is (reasonably) safe to assume, that the `status` passed to the error handler will be a non-negative integer outside of the [200, 300) range.

zachcarroll:
@gkalpak @jgehrcke 
[...]

PR #14459, opened 2016-04-18
docs($http): Add explanation for error status

AbhishekGarg:
Add explanation for error status codes in $http call and line about normalization to zero in case of negative status code.
This alleviates confusion about status codes for which error callback would be called and final values to be expected from an erroneous call.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -451,9 +451,11 @@ function $HttpProvider() {
      *   - **statusText** – `{string}` – HTTP status text of the response.
      *
      * A response status code between 200 and 299 is considered a success status and
-     * will result in the success callback being called. Note that if the response is a redirect,
-     * XMLHttpRequest will transparently follow it, meaning that the error callback will not be
-     * called for such responses.
+     * will result in the success callback being called. Response status code other than
+     * the mentioned ones is considered as error status. The error callback is called in such case.
+     * Also, any negative status code is normalized to zero. Other codes are returned as such.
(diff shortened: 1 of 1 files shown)


gkalpak:
@AbhishekGarg, we need a signed CLA before we can merge this.
Also, please add `Fixes #11945` at the end of your commit message, so that issue it closed when/if this is merged.

AbhishekGarg:
I have signed the CLA.
@gkalpak I'll make the suggested changes and changes for successful build asap.

AbhishekGarg:
@gkalpak Updated commit message and changed for travis-ci check.
Could you please have a look?

src/ng/http.js
```diff
@@ -451,9 +451,11 @@ function $HttpProvider() {
      *   - **statusText** – `{string}` – HTTP status text of the response.
      *
      * A response status code between 200 and 299 is considered a success status and
-     * will result in the success callback being called. Note that if the response is a redirect,
-     * XMLHttpRequest will transparently follow it, meaning that the error callback will not be
-     * called for such responses.
+     * will result in the success callback being called. Response status code other than
+     * the mentioned ones is considered as error status. The error callback is called in such case.
+     * Also, any negative status code is normalized to zero. Other codes are returned as such.
```
gkalpak: Only codes < -1 are normalized. -1 is returned as is (and usually, means aborted, e.g. via timeout).

gkalpak:
I tweaked it a bit and merged. Thx !

Issue #12572, opened 2015-08-13
feat(orderBy): Support natural sort


gabrielmaldi:
This is a feature request. I searched issues and pull request and didn't find anything, sorry if this is a dupe.

I think that many of us would like to be able to ask `orderBy` to sort strings in a "natural" way (i.e. `foo 92` "is smaller than" `foo 147`, as opposed to the current lexicographical order which sorts `1` before `9`).

I'd be willing to submit a pull request but first I'd like to know if this is a feature the Angular team would like to add, discuss the possible syntax, and get some guidelines (I took a look at `orderByFilter` and noticed it's really optimized for performance).

Thanks!

lgalfaso:
This is one of those things and always found a reasonable workaround (eg
for this case, (+somefield.substr(4))), but would be open to it if there is
enough community interest
>
>

Ericson578:
Maybe a flag to force orderBy to convert input into integers?

frankweb:
@Narretz I'd like to know your oppinion about this feature, I could create a new filter.

Ericson578:
I've definitely run into this problem before, where I have a field that I want to order by in an ng-repeat, and the field is a number but for a variety of reasons is stored in js as a string. I've fixed the problem by writing a custom filter that converts to int before comparing using parseInt(num, 2), but would be nice if there was already a filter that did this.

gkalpak:
Is it just me or does this sound like a better fit for an external module (rather than a core feature) ?

gkalpak:
Now that #14468 has landed in master, it will be even easier to do this by providing a custom comparator function. I'm going to close this for now, since it seems like a better fit for an external module.

PR #14468, opened 2016-04-19
feat(orderBy): add support for custom comparators

### LINKED ISSUE #13238: Ordering string which are case sensitive
Hi,

when I want to sort some strings with respected lower and upper case I'm forced to use my own function. Is there any way to use `orderBy` in this case?

https://github.com/angular/angular.js/blob/v1.5.0-beta.1/src/ng/filter/orderBy.js#L275

### LINKED ISSUE #14455: orderBy  not supported persian language
filter orderBy not supported persian language, 
orderBy ignore  letters "گ چ پ ژ".

gkalpak:
Feature

The comparator used by `orderBy` is hard-coded. 

The user can specify a custom comparator function.

No

**Other information**:
With this change the 3rd argument is interpreted as a comparator function, used to compare the values returned by the predicates. Leaving it empty falls back to the default comparator and a value of `false` is a special case that also falls back to the default comparator but reverses the order. Thus the new implementation is backwards compatible.

This commit also expands the documentation to cover the algorithm used to sort elements and adds a few more unit and e2e tests (unrelated to the change).

Helps with #12572 (maybe this is as close as we want to get).

Fixes #13238
Fixes #14455

Closes #5123
Closes #8112
Closes #10368

### DIFF
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -6,44 +6,118 @@
  * @kind function
  *
  * @description
- * Orders a specified `array` by the `expression` predicate. It is ordered alphabetically
- * for strings and numerically for numbers. Note: if you notice numbers are not being sorted
- * as expected, make sure they are actually being saved as numbers and not strings.
- * Array-like values (e.g. NodeLists, jQuery objects, TypedArrays, Strings, etc) are also supported.
+ * Returns an array containing the items from the specified `collection`, ordered by a `comparator`
+ * function based on the values computed using the `expression` predicate.
  *
- * @param {Array} array The array (or array-like object) to sort.
- * @param {function(*)|string|Array.<(function(*)|string)>=} expression A predicate to be
- *    used by the comparator to determine the order of elements.
+ * The `collection` can be an Array or array-like object (e.g. NodeList, jQuery object, TypedArray,
+ * String, etc).
+ *
+ * The `expression` can be a signle predicate or a list of predicates, each serving as a tie-breaker
+ * for the preceeding one. The `expression` is evaluated against each item and the output is used
+ * for comparing with other items.
+ *
+ * The comparison is done using the `comparator` function. If none is specified, a default, built-in
+ * comparator is used (see below for details). Passing `true` as the comparator has a special
+ * meaning: It will use the built-in comparator, but sort in reverse order.
+ *
+ * ### Under the hood
+ *
+ * Ordering the specified `collection` happens in two phases:
+ *
+ * 1. All items are passed through the predicate (or predicates), and the returned values are saved
+ *    along with their type (`string`, `number` etc). For example, an item `{name: 'foo'}`, passed
+ *    through a predicate that extracts the value of the `name` property, would be transformed to:
+ *    ```
+ *    {
+ *      value: 'foo',
+ *      type: 'string',
+ *      index: ...
+ *    }
+ *    ```
+ * 2. The comparator function is used to sort the items, besed on the derived values, types and
+ *    indices.
+ *
+ * If you use a custom comparator, it will be called with pairs of objects of the form
+ * `{value: ..., type: '...', index: ...}` and is expected to return `0` if the objects are equal
+ * (as far as the comparator is concerned), `-1` if the 1st one should be ranked higher than the
+ * second, or `1` otherwise.
+ *
+ * In order to ensure that the sorting will be deterministic across platforms, if none of the
+ * specified predicates can distinguish between two items, `orderBy` will automatically introduce a
+ * dummy predicate that returns the item's index as `value`.
+ * (If you are using a custom comparator, make sure it can handle this case as well.)
+ *
+ * Finally, in an attempt to simplify things, if a predicate returns an object as the extracted
+ * value for an item, `orderBy` will try to convert that object to a primitive value, before passing
+ * it to the comparator. The following rules govern the conversion:
+ *
+ * 1. If the object has a `valueOf()` method that returns a primitive, its return value will be
+ *    used instead.<br />
+ *    (If the object does have a `valueOf()` method, but it returns another object, then the
+ *    returned object will be used in subsequent steps.)
+ * 2. If the object has a custom `toString()` method (i.e. not the one inherited from `Object`) that
+ *    returns a primitive, its return value will be used instead.<br />
+ *    (If the object does have a `toString()` method, but it returns another object, then the
+ *    returned object will be used in subsequent steps.)
+ * 3. No conversion; the object itself is used.
+ *
+ * ### The default comparator
+ *
+ * The default, built-in comparator should be sufficient for most usecases. In short, it compares
+ * numbers numerically, strings alphabetically (and case-insensitively), for objects falls back to
+ * using their index in the original colection, and sorts values of different types by type.
+ *
+ * More specifically, it follows these steps to determine the relative order of items:
+ *
+ * 1. If the compared values are of different types, compare the types themselves alphabetically.
+ * 2. If both values are of type `string`, compare them alphabetically in a case- and
+ *    locale-insensitive way.
+ * 3. If both values are objects, compare their indices instead.
+ * 4. Otherwise, return:
+ *    -  `0`, if the values are equal (by strict equality comparison, i.e. using `===`).
+ *    - `-1`, if the 1st value is "less than" the 2nd value (compared using the `<` operator).
+ *    -  `1`, otherwise.
+ *
+ * **Note:** If you notice numbers not being sorted as expected, make sure they are actually being
+ *           saved as numbers and not strings.
+ *
+ * @param {Array|ArrayLike} collection - The collection (array or array-like object) to sort.
+ * @param {(Function|string|Array.<Function|string>)=} expression - A predicate (or list of
+ *    predicates) to be used by the comparator to determine the order of elements.
  *
  *    Can be one of:
  *
- *    - `function`: Getter function. The result of this function will be sorted using the
- *      `<`, `===`, `>` operator.
- *    - `string`: An Angular expression. The result of this expression is used to compare elements
- *      (for example `name` to sort by a property called `name` or `name.substr(0, 3)` to sort by
- *      3 first characters of a property called `name`). The result of a constant expression
- *      is interpreted as a property name to be used in comparisons (for example `"special name"`
- *      to sort object by the value of their `special name` property). An expression can be
- *      optionally prefixed with `+` or `-` to control ascending or descending sort order
- *      (for example, `+name` or `-name`). If no property is provided, (e.g. `'+'`) then the array
- *      element itself is used to compare where sorting.
- *    - `Array`: An array of function or string predicates. The first predicate in the array
- *      is used for sorting, but when two items are equivalent, the next predicate is used.
+ *    - `Function`: A getter function. Ths function will be called with each item as argument and
+ *      the return value will be used for sorting.
+ *    - `string`: An Angular expression. This expression will be evaluated against each item and the
+ *      result will be used for sorting. For example, use `name` to sort by a property called `name`
+ *      or `name.substring(0, 3)` to sort by the first 3 characters of the `name` property.<br />
+ *      (The result of a constant expression is interpreted as a property name to be used for
+ *      comparison. For example, use `"special name"` to sort by a property called `special name`.)<br />
+ *      An expression can be optionally prefixed with `+` or `-` to control the sorting direction,
+ *      ascending or descending. For example, `+name` or `-name`. If no property is provided, (e.g.
+ *      `'+'` or `'-'`), the collection element itself is used in comparisons.
+ *    - `Array`: An array of function and/or string predicates. If a predicate cannot determine the
+ *      relative order of two items, the next predicate is used as a tie-breaker.
  *
- *    If the predicate is missing or empty then it defaults to `'+'`.
+ * **Note:** If the predicate is missing or empty then it defaults to `'+'`.
  *
- * @param {boolean=} reverse Reverse the order of the array.
- * @returns {Array} Sorted copy of the source array.
+ * @param {(boolean|Function)=} comparator - The comparator function used to determine the relative
+ *    order of value pairs. If omitted, a built-in comparator will be used. Passing `true` will use
+ *    the built-in comparator, but reverse the order.
+ *
+ * @returns {Array} - The sorted array.
  *
  *
  * @example
- * The example below demonstrates a simple ngRepeat, where the data is sorted
- * by age in descending order (predicate is set to `'-age'`).
- * `reverse` is not set, which means it defaults to `false`.
-   <example module="orderByExample">
+ * The example below demonstrates a simple ngRepeat, where the data is sorted by age in descending
+ * order (expression is set to `'-age'`). The `comparator` is not set, which means it defaults to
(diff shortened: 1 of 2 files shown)


test/ng/filter/orderBySpec.js
```diff
@@ -157,6 +183,38 @@ describe('Filter: orderBy', function() {
     });
 
 
+    it('should support a custom comparator', function() {
```
Narretz: I feel that the test is a bit too clever with "ownerD" being the nerd in question. ;)

Narretz:
What if you want to have a custom comparator and the option to easily reverse the order?

test/ng/filter/orderBySpec.js
gkalpak: I would go as far as calling it "nerdy" :stuck_out_tongue: 

I added a bunch of other tests, so it's not so necessary any more - I just kept it around because I liked it. I'm fine removing it if you think it's too confusing.

gkalpak:
This now ready for review :smile:

I updated the docs and added more tests. Especially wrt to _how_ we sort things (beyond the: _"numbers numerically, strings alphabetically"_). I believe that if people are to rely on some ordering algorithm, they have the right to know what that is.

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
  * @example
- * The example below demonstrates a simple ngRepeat, where the data is sorted
- * by age in descending order (predicate is set to `'-age'`).
- * `reverse` is not set, which means it defaults to `false`.
-   <example module="orderByExample">
+ * The example below demonstrates a simple ngRepeat, where the data is sorted by age in descending
+ * order (expression is set to `'-age'`). The `comparator` is not set, which means it defaults to
+ * the built-in comparator.
+ *
+   <example module="orderByExample1">
```
Narretz: Can you name this and the other examples (name="desriptiveName"). That makes them easier to find in the build folder.

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
+ * Returns an array containing the items from the specified `collection`, ordered by a `comparator`
+ * function based on the values computed using the `expression` predicate.
  *
- * @param {Array} array The array (or array-like object) to sort.
- * @param {function(*)|string|Array.<(function(*)|string)>=} expression A predicate to be
- *    used by the comparator to determine the order of elements.
+ * The `collection` can be an Array or array-like object (e.g. NodeList, jQuery object, TypedArray,
+ * String, etc).
+ *
+ * The `expression` can be a signle predicate or a list of predicates, each serving as a tie-breaker
```
Narretz: signle => single
and I would add a comma after or, because at the moment, it's not super clear to what "each" refers

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
+ *
+ * If you use a custom comparator, it will be called with pairs of objects of the form
+ * `{value: ..., type: '...', index: ...}` and is expected to return `0` if the objects are equal
+ * (as far as the comparator is concerned), `-1` if the 1st one should be ranked higher than the
+ * second, or `1` otherwise.
+ *
+ * In order to ensure that the sorting will be deterministic across platforms, if none of the
+ * specified predicates can distinguish between two items, `orderBy` will automatically introduce a
+ * dummy predicate that returns the item's index as `value`.
+ * (If you are using a custom comparator, make sure it can handle this case as well.)
```
Narretz: "as well" is redundant, imo

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
+ *
+ * Finally, in an attempt to simplify things, if a predicate returns an object as the extracted
+ * value for an item, `orderBy` will try to convert that object to a primitive value, before passing
+ * it to the comparator. The following rules govern the conversion:
+ *
+ * 1. If the object has a `valueOf()` method that returns a primitive, its return value will be
+ *    used instead.<br />
+ *    (If the object does have a `valueOf()` method, but it returns another object, then the
```
Narretz: ", but it returns" -> "that returns"

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
+ *    returned object will be used in subsequent steps.)
+ * 2. If the object has a custom `toString()` method (i.e. not the one inherited from `Object`) that
+ *    returns a primitive, its return value will be used instead.<br />
+ *    (If the object does have a `toString()` method, but it returns another object, then the
```
Narretz: same as above

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
- *    used by the comparator to determine the order of elements.
+ * The `collection` can be an Array or array-like object (e.g. NodeList, jQuery object, TypedArray,
+ * String, etc).
+ *
+ * The `expression` can be a signle predicate or a list of predicates, each serving as a tie-breaker
+ * for the preceeding one. The `expression` is evaluated against each item and the output is used
+ * for comparing with other items.
+ *
+ * The comparison is done using the `comparator` function. If none is specified, a default, built-in
+ * comparator is used (see below for details). Passing `true` as the comparator has a special
```
Narretz: I think it makes sense to say here briefly that strings are sorted alphabetically and numbers numerically (the most common use cases)

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
+ * 2. If the object has a custom `toString()` method (i.e. not the one inherited from `Object`) that
+ *    returns a primitive, its return value will be used instead.<br />
+ *    (If the object does have a `toString()` method, but it returns another object, then the
+ *    returned object will be used in subsequent steps.)
+ * 3. No conversion; the object itself is used.
+ *
+ * ### The default comparator
+ *
+ * The default, built-in comparator should be sufficient for most usecases. In short, it compares
+ * numbers numerically, strings alphabetically (and case-insensitively), for objects falls back to
```
Narretz: Does that mean if the predicate is an object, the default ordering maintains the original order?

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
- *      (for example `name` to sort by a property called `name` or `name.substr(0, 3)` to sort by
- *      3 first characters of a property called `name`). The result of a constant expression
- *      is interpreted as a property name to be used in comparisons (for example `"special name"`
- *      to sort object by the value of their `special name` property). An expression can be
- *      optionally prefixed with `+` or `-` to control ascending or descending sort order
- *      (for example, `+name` or `-name`). If no property is provided, (e.g. `'+'`) then the array
- *      element itself is used to compare where sorting.
- *    - `Array`: An array of function or string predicates. The first predicate in the array
- *      is used for sorting, but when two items are equivalent, the next predicate is used.
+ *    - `Function`: A getter function. Ths function will be called with each item as argument and
```
Narretz: Ths => this

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,118 @@
...
+ *      the return value will be used for sorting.
+ *    - `string`: An Angular expression. This expression will be evaluated against each item and the
+ *      result will be used for sorting. For example, use `name` to sort by a property called `name`
+ *      or `name.substring(0, 3)` to sort by the first 3 characters of the `name` property.<br />
+ *      (The result of a constant expression is interpreted as a property name to be used for
```
Narretz: Isn't that the same as in the sentence above? ( For example, use `name` to sort by a property called `name`)

src/ng/filter/orderBy.js
```diff
@@ -103,100 +210,332 @@
...
+         expect(firstName.getText()).toBe('John');
+         expect(lastName.getText()).toBe('Julie');
+       });
+     </file>
+   </example>
+ * <hr />
+ *
+ * @example
+ * If you have very specific requirements about the way items are sorted, you can pass your own
+ * comparator function. For example, you might need to compare some strings is a locale-sensitive
```
Narretz: is => in

Narretz:
Can you also mention in the docs that orderBy can only order alphabetically by latin characters? Actually, which characters are supported anyway?

src/ng/filter/orderBy.js
gkalpak: I changed it to: _"make sure it can handle this predicate as well"_

src/ng/filter/orderBy.js
gkalpak: Changed it to: _"If the object has a `valueOf()` method that returns..."_

src/ng/filter/orderBy.js
gkalpak: Same as above.

src/ng/filter/orderBy.js
gkalpak: If the predicate returns an object for two elements, there these two elements will retain their original relative ordering.
So, yes :smiley:

src/ng/filter/orderBy.js
gkalpak: There is a subtle difference:

If the Angular expression `name` is evaluated against an object, it will be evaluated to the object's `name` proeprty. (This is standard Angular behavior.)
If the Angular expression `"special name"` is evaluated against an object, typically it would be evaluated to the string `"special name"` (because it's a constant expression - it doesn't access the evluation context). But here, it is special cased: It will actually evaluate to the object's `special name` property.

I guess it was introduced back when `this` was not an lias to the evaluation context in Angular expressions. Nowadays, you can use `this['special name']`, so we don't need the special treating, but I guess we need the backwards compatibility, so...

gkalpak:
@Narretz, I have addressed your comments. PTAL

Regarding the "supported" characters:
It uses the default `<` comparison (so it's not latin characters only). It is just that it doesn't do locale-sensitive comparison (and in some locales it makes a difference). The "locale-insensitive-ness" of the default comparator is already mentioned - other than that all characters are supported (in a sense).

Narretz:
Thanks for the update.
I remember we talked about how to keep the API backwards compatible in a sane way. Did you change anything about this since?

gkalpak:
This is already backwards compatible - but you are right we did talk about something:
Making it easy to have a custom comparator and still be able to specify ascending/descending order.

We settled for `orderBy(collection[, expression[, reverse[, comparator]]])`. I'll make the changes.

gkalpak:
@Narretz, I moved `comparator` as a 4th argument. Now you can specify `reverse` and `comparator` independently. PTAL

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,121 @@
...
+ *      relative order of two items, the next predicate is used as a tie-breaker.
  *
- *    If the predicate is missing or empty then it defaults to `'+'`.
+ * **Note:** If the predicate is missing or empty then it defaults to `'+'`.
  *
- * @param {boolean=} reverse Reverse the order of the array.
- * @returns {Array} Sorted copy of the source array.
+ * @param {boolean=} reverse - If `true`, reverse the sorting order.
+ * @param {(Function)=} comparator - The comparator function used to determine the relative order of
+ *    value pairs. If omitted, a built-in comparator will be used.
```
Narretz: Is there more than one built-in comparator? Or should it be "the default comparator"?

src/ng/filter/orderBy.js
Narretz: I feel like this doesn't really explain that "name" is a property on the scope, and that the value of that property determines which property of the item is used to order the collection.

src/ng/filter/orderBy.js
```diff
@@ -103,100 +212,333 @@
...
+       });
+     </file>
+   </example>
+ * <hr />
+ *
+ * @example
+ * If you have very specific requirements about the way items are sorted, you can pass your own
+ * comparator function. For example, you might need to compare some strings in a locale-sensitive
+ * way. (When specifying a custom comparator, you also need to pass a value for the `reverse`
+ * argument - passing `false` retains the default sorting order, i.e. ascending.)
```
Narretz: Actually, all falsy values. Specifically, undefined can be used (and I would use that if I don't care about a parameter).

src/ng/filter/orderBy.js
```diff
@@ -6,44 +6,121 @@
  * @kind function
  *
  * @description
- * Orders a specified `array` by the `expression` predicate. It is ordered alphabetically
- * for strings and numerically for numbers. Note: if you notice numbers are not being sorted
- * as expected, make sure they are actually being saved as numbers and not strings.
- * Array-like values (e.g. NodeLists, jQuery objects, TypedArrays, Strings, etc) are also supported.
+ * Returns an array containing the items from the specified `collection`, ordered by a `comparator`
+ * function based on the values computed using the `expression` predicate.
```
Narretz: A small example of what the result of the expression might be would be good (in most cases it's the property of an object)

Narretz:
I left a few smaller comments where I think the docs could be clarified. I've also opened a PR in your repo that adds descriptions ot the examples
Otherwise LGTM. Maybe @petebacondarwin can give it another look, too.

test/ng/filter/orderBySpec.js
```diff
@@ -237,6 +218,346 @@ describe('Filter: orderBy', function() {
           {foo: 1, bar: 8}, {foo: 1, bar: 5}, {foo: 1, bar: 2}
           ]);
     });
+
+
+    describe('(reversing order)', function() {
+      it('should not reverse collection if `comparator` param is falsy',
```
petebacondarwin: `comparator` param -> `reverse` param

test/ng/filter/orderBySpec.js
```diff
@@ -237,6 +218,346 @@ describe('Filter: orderBy', function() {
...
+          expect(orderBy(items, expr, 1)).toEqual(sorted);
+          expect(orderBy(items, expr, 'reverse')).toEqual(sorted);
+          expect(orderBy(items, expr, {})).toEqual(sorted);
+          expect(orderBy(items, expr, [])).toEqual(sorted);
+          expect(orderBy(items, expr, noop)).toEqual(sorted);
+        }
+      );
+
+
+      it('should reverse collection if `comparator` param is `true`, even without an `expression`',
```
petebacondarwin: `comparator` param -> `reverse` param

petebacondarwin:
Apart from the couple of nitpicks in docs and tests, this LGTM.
Don't forget to fix up the commit message.

src/ng/filter/orderBy.js
gkalpak: That's not what it is supposed to explain :smiley: It is supposed to explain what the string `name` (i.e. `'name'`) would work.

I admit this is not super clear - hopefully the examples clear things up. I added the surrounding quotes and renamed `name` to `label`, to avoid the confusion of _"the `name` property"_ vs _"the property name"_ (i.e. the name of some property).
We can always improve the docs once merged :smiley:

src/ng/filter/orderBy.js
gkalpak: There's only one. Changed it to `the`.

src/ng/filter/orderBy.js
gkalpak: We have documented this as an (optional) boolean. So, although we treat all falsy values as `false` internally, I prefer to use a boolean in the example.

gkalpak:
Fixed and merged! Thx @Narretz and @petebacondarwin for the reviews :+1:

PR #14471, opened 2016-04-20
fix passing eventHandlers when passThrough

andreyjkee:
**Bug fix**
This commit fix passing eventHandlers and uploadEventHandlers when using angular.mock

Need to fix function arguments in tests: angular.js/test/ngMock/angular-mocksSpec.js


andreyjkee:
I signed it!

Narretz:
Thanks for the PR! Good catch. Can you please update these 2 tests, too? https://github.com/angular/angular.js/blob/e14b486e07be49cdc2113f6c1a5c22e8d101a34f/test/ngMock/angular-mocksSpec.js#L2135-L2150

andreyjkee:
yes, tests updated

PR #14493, opened 2016-04-22
fix(ngSrcset): ignore undefined ng-srcset directive

### LINKED ISSUE #14470: "TypeError: Cannot read property 'split' of undefined" when using ng-attr-srcset with potentially null value
Bug report.

We're seeing an internal Angular error:

	TypeError: Cannot read property 'split' of undefined

when trying to use ng-attr-srcset, where the expression can be potentially undefined.   To reproduce,:

        

Normally, ng-attr-defined attributes are supposed to be ignored if the associated value is undefined, but in this case, we're hitting an error:

        // split srcset into tuple of uri and descriptor except for the last item
        var rawUris = trimmedSrcset.split(pattern);

where trimmedSrcset is undefined under the above situation.

ng-attr is breaking with regards to "srcset".  It should be consistent with ng-attr's behavior of undefined for other attribute names.  What's happening is an internal error that should be fixed, since it's in the framework rather than application code.

Angular 1.5.

rphv:
This is a bug fix.

Currently, a TypeError is thrown when data-ng-attr-srcset is undefined.

Now, the attribute is ignored if it's undefined.

No.

**Other information**:

Closes #14470

### DIFF
--- a/test/ng/directive/ngSrcsetSpec.js
+++ b/test/ng/directive/ngSrcsetSpec.js
@@ -30,7 +30,7 @@ describe('ngSrcset', function() {
   }));
 
   it('should not throw an error if undefined', inject(function($rootScope, $compile) {
-    element = $compile('<img data-ng-attr-srcset="{{undefined}}">')($rootScope);
+    element = $compile('<img ng-attr-srcset="{{undefined}}">')($rootScope);
     $rootScope.$digest();
   }));
 });
--- a/test/ng/directive/ngSrcsetSpec.js
+++ b/test/ng/directive/ngSrcsetSpec.js
@@ -30,10 +30,8 @@ describe('ngSrcset', function() {
   }));
 
   it('should not throw an error if undefined', inject(function($rootScope, $compile) {
-    $rootScope.imageUrl = {};
-    element = $compile('<img ng-srcset="{{undefined}}">')($rootScope);
+    element = $compile('<img data-ng-attr-srcset="{{undefined}}">')($rootScope);
     $rootScope.$digest();
-    expect(element.attr('srcset')).toBe(undefined);
   }));
 });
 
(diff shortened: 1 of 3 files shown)


test/ng/directive/ngSrcsetSpec.js
```diff
@@ -28,5 +28,12 @@ describe('ngSrcset', function() {
     $rootScope.$digest();
     expect(element.attr('srcset')).toBe('http://example.com/image1.png 1x,unsafe:javascript:doEvilStuff() 2x');
   }));
+
+  it('should not throw an error if undefined', inject(function($rootScope, $compile) {
```
gkalpak: I didn't try it, but I am pretty sure this test passes without the fix.
Can you confirm that's inded tha case and change the test, so it fails without the fix ?

gkalpak:
Thx @rphv for working on this.
Could you address my comment ? (Feel free to ping me once you have, so I can take another look.)

test/ng/directive/ngSrcsetSpec.js
rphv: Yes, you're right. That test wasn't testing the fix.

I've updated the test and verified that it fails without the fix.

rphv:
Test updated.

On Fri, Apr 22, 2016 at 3:57 AM, Georgios Kalpakas <notifications@github.com

gkalpak:
@rphv, since this is not `ng-attr-` specific (and there is no guarantee that its implementatation will keep exercising the same code paths in the future), could you also add a test that call `attrs.$set('srcset', undefined)` directly on an `` element ?

(Also, you don't have to use `data-` prefixes in tests - they are just adding clutter.)

rphv:
Thank you for your advice and attention.

I'm not sure if the new test I added is in the right place or in the
accepted style of the project. However, it does fail without the fix.

gkalpak:
Tweaked the test a bit and merged.
Thx for working on this :+1:

PR #14496, opened 2016-04-22
fix($templateRequest): Handle empty templates in cache correctly.

### LINKED ISSUE #14479: empty template are treated as not exist in templateCache

empty template are treated as not exist in templateCache

create a empty template, use template cache. and it is treated as not cached

should be treated as cached

i think this is a bug.

multiple version include the latest

in $TemplateRequestProvider

this condition
if (!isString(tpl) || !$templateCache.get(tpl))

should be if (!isString(tpl) || !isUndefiend($templateCache.get(tpl)))

rjamet:

Bug fix.

Empty templates in $templateCache are not considered as in cache for $templateRequest security purposes, so they trigger $sce checks. Specifically, $templateRequest calls $templateCache.get, but looks for the falsiness of the return value instead of it being defined. 

Empty templates in $templateCache are now considered as trusted as any other in $templateCache.

No.

- [x] The commit message follows our guidelines: https://github.com/angular/angular.--set-upstreamjs/blob/master/CONTRIBUTING.md#commit-message-format
- [] Docs have been added / updated (for bug fixes / features)

I'm not sure docs are necessary, as this bug wasn't documented.

**Other information**:

Fixes #14479.

### DIFF
--- a/src/ng/templateRequest.js
+++ b/src/ng/templateRequest.js
@@ -68,14 +68,8 @@ function $TemplateRequestProvider() {
       // are included in there. This also makes Angular accept any script
       // directive, no matter its name. However, we still need to unwrap trusted
       // types.
-      if (!isString(tpl)) { // Safe types need unwrapping.
+      if (!isString(tpl) || isUndefined($templateCache.get(tpl))) {
         tpl = $sce.getTrustedResourceUrl(tpl);
-      } else { // Strings can be tested against the cache.
-        var cacheAnswer = $templateCache.get(tpl);
-        if (typeof cacheAnswer === 'undefined') {
-          // It is not in the cache, hence we don't trust it and it has to go through the $sce.
-          tpl = $sce.getTrustedResourceUrl(tpl);
-        } // Anything else means the cache has a record of it, so no need to check.
       }
 
       var transformResponse = $http.defaults && $http.defaults.transformResponse;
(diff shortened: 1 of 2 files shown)


src/ng/templateRequest.js
```diff
@@ -68,8 +68,14 @@ function $TemplateRequestProvider() {
       // are included in there. This also makes Angular accept any script
       // directive, no matter its name. However, we still need to unwrap trusted
       // types.
-      if (!isString(tpl) || !$templateCache.get(tpl)) {
```
gkalpak: Couldn't you just change this line to `... || isUndefined($templateCache.get(tpl))...` and leave everythng else unchanged ?

test/ng/templateRequestSpec.js
```diff
@@ -158,6 +158,37 @@ describe('$templateRequest', function() {
...
+      $rootScope.$digest();
+    }).toThrow();
+
+    $templateCache.put('tpl.html', null); // makes no sense, but it's been added, so trust it.
+    expect(function() {
+      $templateRequest('tpl.html');
+      $rootScope.$digest();
+    }).not.toThrow();
+
+    $templateCache.put('tpl.html', ''); // should work (empty template)
```
gkalpak: Since you have already called `$templateCache.put()` previously, this test is not so reliable.
I would either empty the cache first ot use the `they` helper to make each case its own testcase.

src/ng/templateRequest.js
rjamet: The reasoning is that people asked for wrong fixes several times, so my original code was probably confusing. I've unrolled the if to accomodate clearer comments, but that can be reverted if you prefer.

test/ng/templateRequestSpec.js
rjamet: Yup, good point. Fixed with deleteAll after each of the blocks.

PR #14497, opened 2016-04-22

jkrth617:
docs: the error message wasn't helpful

the error message reads as
"Argument 'mainController' is not a"
just a fragment sentence with not a lot of insight

I added a 'usually happens when' section to point out that they may have forgot an argument

no

**Other information**:

the message just said 'bad' but didn't give any insight on what that meant. I added a 'usually happens' section because I found myself making this mistake and seeing a nothing error

### DIFF
--- a/docs/content/error/ng/areq.ngdoc
+++ b/docs/content/error/ng/areq.ngdoc
@@ -6,3 +6,8 @@
 AngularJS often asserts that certain values will be present and truthy using a
 helper function. If the assertion fails, this error is thrown. To fix this problem,
 make sure that the value the assertion expects is defined and truthy.
+
+This usually happens when you forget to pass in both arguments to a .controller function or .service function for example.  
+Remember to pass both the name of the component as a string and then the second argument as the functionality with all  
+dependencies injected using either the inline style or the '.$inject=' style.
+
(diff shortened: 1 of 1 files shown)


jkrth617:
I have signed

gkalpak:
If you literally see _"Argument 'mainController' is not a"_, it seems like an error (which we should fix if possible). Can you post a reproduction ?

Speculating about possible errors (such as included `<script>` tags) doesn't sound very helpful either.
I would include a couple minimal examples to better illustrate what you mean by _"as the functionality with all dependencies injected using either the inline style or the '.$inject=' style"_.

<sub>I'm afraid this is going to be more confusing than helpful in its current state.</sub>

petebacondarwin:
Looking through the code there is no where that would trigger this error with the message cut off like that.
I wonder if you were looking on a narrow screen and we clipped the message using CSS?

petebacondarwin:
Hmm so are not clipping that message - it is set to `height: auto` and `white-space: pre-wrap`

petebacondarwin:
I suspect that you got the following error, which was probably due to the controller not being defined or misspelled (capitalisation?):

https://docs.angularjs.org/error/ng/areq?p0=mainController&p1=not%20a%20function,%20got%20undefined

We don't have a reproduction for the problem and there has been no reply for a month so I am closing this one.  @jkrth617 if you can provide more information, either comment here or open a new issue.

PR #14499, opened 2016-04-23
fix($brower): set the url even if the browser transforms it

### LINKED ISSUE #14427: IE 11 confuses $browser.url(), causing $location to go mad (infinite digest cycle, ...)

A *bug* in $browser.url() that causes $location to go mad (infinite digest cycle, spontaneous navigation to previous views, ...).

If `$browser.url()` is passed a url, it executes the following code:

```
      // Don't use history API if only the hash changed
      // due to a bug in IE10/IE11 which leads
      // to not firing a `hashchange` nor `popstate` event
      // in some cases (see #9143).
        cacheState();
      } else {
        if (!sameBase || pendingLocation) {
          pendingLocation = url;
        }
        if (replace) {
          location.replace(url);
        } else if (!sameBase) {
          location.href = url;
        } else {
          location.hash = getHash(url);
        }
        if (location.href !== url) {
          pendingLocation = url;
        }
```

If `sameBase` and `sameState` are both true, and the caller has set `replace`, this invokes `location.href=url`. If `url` contains a substring that can be interpreted as a HTML entity reference, [Internet Explorer 11 often resolves that entity reference](https://connect.microsoft.com/IE/feedback/details/1040980/bug-in-ie-which-interprets-document-location-href-as-html) (in our case, the url parameter `&not-before-policy=0` was replaced by `¬-before-policy=0`. Therefore, `location.href !== url`, and `pendingLocation` is set.

As a consequence, querying `$browser.url()`, which is implemented by 

```
    // getter
    } else {
      // - pendingLocation is needed as browsers don't allow to read out
      //   the new location.href if a reload happened or if there is a bug like in iOS 9 (see
      //   https://openradar.appspot.com/22186109).
      // - the replacement is a workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=407172
      return pendingLocation || location.href.replace(/%27/g,"'");
    }
```

yields the original url, which is probably fine.

However, future calls to `$browser.url()` to actually change the url satisfy the condition 

therefore do not update `pendingLocation`, and `$browser.url()` continues to return the old url.

This is turn horribly confuses $location, because it activates the watch:

```
    $rootScope.$watch(function $locationWatch() {
      var oldUrl = trimEmptyHash($browser.url());
      var newUrl = trimEmptyHash($location.absUrl());
      var oldState = $browser.state();
      var currentReplace = $location.$$replace;
      var urlOrStateChanged = oldUrl !== newUrl ||
        ($location.$$html5 && $sniffer.history && oldState !== $location.$$state);

      if (initializing || urlOrStateChanged) {
        initializing = false;

        $rootScope.$evalAsync(function() {
          var newUrl = $location.absUrl();
          var defaultPrevented = $rootScope.$broadcast('$locationChangeStart', newUrl, oldUrl,
              $location.$$state, oldState).defaultPrevented;

          // if the location was changed by a `$locationChangeStart` handler then stop
          // processing this location change
          if ($location.absUrl() !== newUrl) return;

          if (defaultPrevented) {
            $location.$$parse(oldUrl);
            $location.$$state = oldState;
          } else {
            if (urlOrStateChanged) {
              setBrowserUrlWithFallback(newUrl, currentReplace,
                                        oldState === $location.$$state ? null : $location.$$state);
            }
            afterLocationChange(oldUrl, oldState);
          }
        });
      }
```

which always schedules an asyncTask to `$broadcast` the corresponding events, causing an infinite digest cycle.

Moreover, when `$browser` invokes `onUrlChange` in response to a browser event, it passes `$browser.url()`, which causes $location to spontaneously navigate back to the original url rather than the url the browser displays.

In IE 11:
```
$browser.url('oauthCallback#state=xxx&not-before-policy=0');
$browser.url('oauthCallback#state=xxx&not-before-policy=0');
$browser.url('someOtherPage');
$browser.url(); // should return 'someOtherPage', but returns  `oauthCallback...`
```

Note that IE 11 performing HTML escaping in URLs is nondeterministic. On some environments, it occurs nearly every time, in others hardly ever. To get a deterministic reproducer, you'd have to mock `location`.

Obviously, invoking the setter of `$browser.url()` should cause the getter to report the new value. 

Our users rather dislike infinite digest loops.

We tested with angular 1.5.0, but the code in master looks unchanged. We suspect the regression was introduced by 8d39bd8abf423517b5bff70137c2a29e32bff76d

bedag-moo:

Fixes #14427

See #14427

Features should not be affected - unless you consider inifinite digest a feature ;-)

No as far as I know. 

**Other information**:

While $location expects that $browser stores the URL unchanged, some browsers transform the URL
when setting or defer the acutal update. To work around this, $browser.url() kept the unchanged
URL in pendingLocation. However, it failed to update pendingLocation in all code paths, causing
$browser.url() to sometimes incorrectly report previous URLs, which horribly confused $location.
This fix ensures that pendingLocation is always updated if set, causing url() to report the
current url.

Fixes #14427

### DIFF
--- a/src/ng/browser.js
+++ b/src/ng/browser.js
@@ -153,7 +153,7 @@ function Browser(window, document, $log, $sniffer) {
         // Do the assignment again so that those two variables are referentially identical.
         lastHistoryState = cachedState;
       } else {
-        if (!sameBase || pendingLocation) {
+        if (!sameBase) {
           pendingLocation = url;
         }
         if (replace) {
@@ -166,6 +166,9 @@ function Browser(window, document, $log, $sniffer) {
         if (location.href !== url) {
           pendingLocation = url;
         }
+      }
+      if (pendingLocation) {
+        pendingLocation = url;
       }
       return self;
     // getter
--- a/test/ng/browserSpecs.js
+++ b/test/ng/browserSpecs.js
@@ -92,6 +92,7 @@ function MockWindow(options) {
     },
     replaceState: function(state, title, url) {
       locationHref = url;
+      if (!options.updateAsync) committedHref = locationHref;
       mockWindow.history.state = copy(state);
     }
   };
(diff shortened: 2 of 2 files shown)


bedag-moo:
I signed the CLA.

bedag-moo:
Sorry for missing semicolon, the commit should have fixed it.

It seems Travis is still building the old version, though. No clue why. Does it not support amending pull requests?

Is there anything else I can do to accelerate acceptance of this pull request? (It's a blocking bug for one of our projects, and I'd rather like to minimize the time where I must maintain a fork of angular.)

bedag-moo:
It's been two weeks ... is there anything I can do to accelerate the review process, @gkalpak?

bedag-moo:
@petebacondarwin might be best qualified to review this change, since he introduced the `pendingLocation` variable in 8d39bd8abf423517b5bff70137c2a29e32bff76d. Could you take a look?

petebacondarwin:
I'll take a look next week

petebacondarwin:
@bedag-moo - really nice PR. `$location`/`$browser` have some of the most complicated brain-bending bits of Angular bugs.  I think this looks good so I am going to merge it into master and 1.5.x

Martinspire:
Isn't there a typo in the title and now in the readme?

petebacondarwin:
Fixed the CHANGELOG at f58d4fb7482f0c0884641654d301373268ff2d6b
Thanks @Martinspire

PR #14509, opened 2016-04-25
perf($provider): simplifies the condition for $get value

piotrkabacinski:

Simplifies the condition for $get value.

Behavior is the same. PR has optimization effect.

**Other information**:

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -710,7 +710,7 @@ function createInjector(modulesToLoad, strictDi) {
 
   function factory(name, factoryFn, enforce) {
     return provider(name, {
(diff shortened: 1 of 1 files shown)


piotrkabacinski:
@googlebot I signed it!

gkalpak:
This change isn't valid. It is neither a perf improvement, nor does it retain the same behavior:

- `enforce !== false` means the if `enforce` is not explicitly set to false, then it is assumed true.
- `enforce` means that any falsy value (including `undefined`) is interpreted as `false`. This means that if omitted, `false` will be assumed (which is **not** the intention).

But thx for taking the time to submit a PR anyway. Maybe next time :wink:

piotrkabacinski:
@gkalpak - thanks for the feedback and clearing up this case :)

PR #14514, opened 2016-04-26
docs(guide/Developer Guide): Fixed formatting with subtitles

tupini07:
On the [main page](https://docs.angularjs.org/guide) of the developer guide:

The "Books" and "Videos" subtitles had no space between text and the '#' so it didn't render as a subtitle in the web page.

It should be:
### Videos

Instead of:


tupini07:
I signed it!

PR #14519, opened 2016-04-26
perf($animate): listen for document visibility changes

### LINKED ISSUE #14066: Slow ng-repeat because of AnimateRunner even without ngAnimate

Narretz:

**Other information**:

perf(ngAnimate): listen for document visibility changes

Accessing the document for the hidden state is costly for
platforms like Electron. Instead, listen for visibilitychange
and store the state.

(#14071)
Closes #14066


src/ng/document.js
```diff
@@ -30,3 +30,29 @@ function $DocumentProvider() {
...
+    var hidden = doc && doc.hidden;
+
+    $document.on('visibilitychange', changeListener);
+
+    $rootScope.$on('$destroy', function() {
+      $document.off('visibilitychange', changeListener);
+    });
+
+    function changeListener() {
+      hidden = doc.hidden;
```
campersau: You are checking if `doc` is available in line 42 but not here. Is this correct? Is the check above necessary?

Narretz: I think the check was originally introduced by @matsko in case the document isn't available. I assume that when the document isn't available I can't add a listener anyway.

test/ng/documentSpec.js
```diff
@@ -27,3 +27,29 @@ describe('$document', function() {
     });
   });
 });
+
+
+describe('$$isDocumentHidden', function() {
+  it('should listen on the visibilitychange event', function() {
+    var doc;
```
gkalpak: Unused

test/ng/documentSpec.js
```diff
@@ -27,3 +27,29 @@ describe('$document', function() {
...
+
+    var spy = spyOn(document, 'addEventListener').and.callThrough();
+
+    inject(function($$isDocumentHidden, $document) {
+      expect(spy.calls.mostRecent().args[0]).toBe('visibilitychange');
+      expect(spy.calls.mostRecent().args[1]).toEqual(jasmine.any(Function));
```
gkalpak: Since we don't really care if it was the most recent, you could also usr `toHaveBeenCalledWith()`.

test/ng/documentSpec.js
```diff
@@ -27,3 +27,29 @@ describe('$document', function() {
...
+      expect($$isDocumentHidden()).toBeFalsy(); // undefined in browsers that don't support visibility
```
gkalpak: This seems slightly unrelated to the rest of the testcase. Could as well go into it's own testcase.

test/ng/documentSpec.js
```diff
@@ -27,3 +27,29 @@ describe('$document', function() {
     });
   });
 });
+
+
+describe('$$isDocumentHidden', function() {
```
gkalpak: I think there should be a test tgat shows the value changing when `visibilitychange` is fired.

asnar:
On Apr 27, 2016 2:41 AM, "Martin Staffa" <notifications@github.com> wrote:
>
> What kind of change does this PR introduce? (Bug fix, feature, docs
update, ...)
>
> What is the current behavior? (You can also link to an open issue here)
>
> What is the new behavior (if this is a feature change)?
>
> Does this PR introduce a breaking change?
>
> Please check if the PR fulfills these requirements
>
> The commit message follows our guidelines:
https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format

test/ng/documentSpec.js
Narretz: I used that before, but there's a difference between jquery 2.1 and jqlite, jquery calls addEventListener with the third argument set to false, jqlite doesn't set the argument. That's why I used mostRecent.

test/ng/documentSpec.js
Narretz: I agree, but I found it quite complicated to do that. Because you'd have to mock the document (because hidden isn't writable), and still have it accept event listeners and fire events.

test/ng/documentSpec.js
gkalpak: So, _that's_ what you were talking about with @mgol :)
OK then :+1:

test/ng/documentSpec.js
gkalpak: I don't think you need all that. I think all you need it mocking `document` wrt the `hidden` attribute (e.g. making it a getter/setter). The events are dispatched through `$document`, so you can trigger them using `triggerHandler('visibilitychange')`.

But this is aff the top of my head, so I might be missing something. If it's not too complicated I'd say let's add some.
If it's too complicated, it's fine leaving it out.

Narretz:
Superseded by #14568. Turns out you can't squash the PR when your try to merge from the same branch in your fork.

PR #14527, opened 2016-04-27
fix(ng-bind-html): deep watch the actual $sce value, not its toString.

### LINKED ISSUE #14526: ngBindHtml does not work if SafeHtml types don't have a toString function
If you have a `$sce` service whose objects do not return their value in `toString`, `ngBindHtml` will never receive updates because it relies on the `toString` value for change detection. That's pretty common because `toString` by default just returns `[object Object]`, and for optimized binaries your optimizer might strip a `toString` method.

Using `toString` in `ngBindHtml` is overall a bit hacky, it runs counter to AngularJS' normal change detection. The fix is to just use a deep watch.

mprobst:
Custom $sce implementations might not have a `toString()`, or it might be
compiled away in non-debug mode. Deep watching the object is the correct fix
with Angular's regular watch semantics.

The performance of this should be equivalent - `toString()` on objects usually
touches all fields, so a deep object watch will be in the same ballpark, except
that it avoids the (potentially big) string allocation.

Fixes #14526.

### DIFF
--- a/src/ng/directive/ngBind.js
+++ b/src/ng/directive/ngBind.js
@@ -188,8 +188,9 @@ var ngBindHtmlDirective = ['$sce', '$parse', '$compile', function($sce, $parse, 
     restrict: 'A',
     compile: function ngBindHtmlCompile(tElement, tAttrs) {
       var ngBindHtmlGetter = $parse(tAttrs.ngBindHtml);
-      var ngBindHtmlWatch = $parse(tAttrs.ngBindHtml, function getStringValue(value) {
-        return (value || '').toString();
+      var ngBindHtmlWatch = $parse(tAttrs.ngBindHtml, function sceValueOf(val) {
+        // Unwrap the value to compare the actual inner safe value, not the wrapper object.
+        return $sce.valueOf(val);
       });
       $compile.$$addBindingClass(tElement);
 
@@ -197,9 +198,9 @@ var ngBindHtmlDirective = ['$sce', '$parse', '$compile', function($sce, $parse, 
         $compile.$$addBindingInfo(element, attr.ngBindHtml);
 
         scope.$watch(ngBindHtmlWatch, function ngBindHtmlWatchAction() {
-          // we re-evaluate the expr because we want a TrustedValueHolderType
-          // for $sce, not a string
-          element.html($sce.getTrustedHtml(ngBindHtmlGetter(scope)) || '');
+          // The watched value is the unwrapped value. To avoid re-escaping, use the direct getter.
+          var value = ngBindHtmlGetter(scope);
+          element.html($sce.getTrustedHtml(value) || '');
         });
       };
     }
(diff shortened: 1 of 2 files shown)


mprobst:
@rjamet

src/ng/directive/ngBind.js
```diff
@@ -188,19 +188,16 @@ var ngBindHtmlDirective = ['$sce', '$parse', '$compile', function($sce, $parse,
...
-        return (value || '').toString();
-      });
       $compile.$$addBindingClass(tElement);
 
       return function ngBindHtmlLink(scope, element, attr) {
         $compile.$$addBindingInfo(element, attr.ngBindHtml);
 
-        scope.$watch(ngBindHtmlWatch, function ngBindHtmlWatchAction() {
+        scope.$watch(ngBindHtmlGetter, function ngBindHtmlWatchAction(newVal) {
           // we re-evaluate the expr because we want a TrustedValueHolderType
```
IgorMinar: update the comment please

IgorMinar:
lgtm

src/ng/directive/ngBind.js
mprobst: fixed.

gkalpak:
FWIW, I don't think performance will be the same (in all cases). Running `angular.equals()` on every digest is more expensive that comparing strings. But `ngBindHtml` shouldn't be so common, so the overall impact won't make any practical difference.

I wonder whether it would be worth it to use `toString()` if the built-in `$sceDelegate` is used or to always use `valueOf()` (with swallow $watch) instead of `toString()` ?

Anyway, I'll merge this once Travis is green and we can make further changes in subsequent PRs.

gkalpak:
The failures are legitimate. Basically, due to how `equals()` works (and how `$sceDelegate` builds trusted types), any vaues returned by `$sce.trustAsHtml('...')` are considered identical.

gkalpak:
We can fix this by exposing the trusted value as a public property on `$sceDelegate`'s wrapper type, but this might break existing implementation of custom `$sceDelegate`s.

petebacondarwin:
@mprobst can you take a look at @gkalpak's comments?

rjamet:
Exposing the internal trusted value is what $sce.valueOf does, and it returns non-sce types as-is. It's part of the public interface of the $sce, so I think it's the right tool here. (code: https://github.com/angular/angular.js/blob/master/src/ng/sce.js#L330)

mprobst:
@rjamet thanks for the tip with `valueOf`, that's indeed the right approach. I've amended the PR. I also added a unit test for ngBindHtml updating values on digest loops, i.e. the that the specific watch getter that it uses actually fires.

gkalpak:
LGTM

gkalpak:
I tweaked the commit message a bit (because it was describing the previous approach with deep-watching) and merged.

mprobst:
Thanks @gkalpak!

PR #14531, opened 2016-04-28
fix($injector): add workaround for class stringification in Chrome v50/51

### LINKED ISSUE #14240: Native class syntax breaks invoking directive controller
Bug

Initializing controller created with native class syntax breaks directive compilation. Problem here is with pushing bindings from directive to self initalized controller instance before constructor is invoked, but ES2015 class syntax specification forbids using constructor without "new" operator. 

Codepen example: http://codepen.io/smalluban/pen/eZBeNr?editors=1012

Console output:

```
console_runner-ba402f0….js:1 TypeError: Class constructor Controller cannot be invoked without 'new'
    at Object.invoke (angular.js:4604)
    at extend.instance (angular.js:9855)
    at nodeLinkFn (angular.js:8927)
    at compositeLinkFn (angular.js:8226)
    at compositeLinkFn (angular.js:8229)
    at publicLinkFn (angular.js:8106)
    at angular.js:1696
    at Scope.$eval (angular.js:16820)
    at Scope.$apply (angular.js:16920)
    at bootstrapApply (angular.js:1694)
```

Pushing bindings to class constructor can not be fixed. With ES5 ordinary function everything works fine, but using class syntax is not safe if you don't use transpiler(like Babel). Only way is to deprecate it and push values after class is constructed, but I have no idea if this is possible (due to fact of using this functionality).

There is a lot of articles where we are encourage to use class syntax for creating controllers. But using classes is danger, because application will break if you use native classes. 

In my opinion using Babel to transpile class syntax is temporary. Most modern browsers already support classes, so using Angular 1 in hybrid application could be used without transpilation.

Angular: 1.5.0
Browser: Every browser where class syntax is supported: Chrome 49+, Edge 13+, Firefox 45+, Safari 9+.

gkalpak:
Bug fix.

In Chrome v50, the `$injector` fails to properly detect classes and throws when trying to `invoke` a class.
See #14240 (especially https://github.com/angular/angular.js/issues/14240#issuecomment-197418167 and https://github.com/angular/angular.js/issues/14240#issuecomment-214039468).

Classes are detected and `$injetor.invoke`d correctly.

No

**Other information**:
Everything works correctly on v52 without the fix. Not sure about v51.
We can't run the tests on CI, because the problem appears only when creating a class directly (not through `eval()`). I added a couple of commented out tests, so it is easy to test once Chrome v52 is released (as stable). (I left a TODO note to remove them.)

Partially fixes #14240.

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -70,12 +70,16 @@ var FN_ARG = /^\s*(_?)(\S+?)\1\s*$/;
 var STRIP_COMMENTS = /((\/\/.*$)|(\/\*[\s\S]*?\*\/))/mg;
 var $injectorMinErr = minErr('$injector');
 
-function extractArgs(fn) {
+function stringifyFn(fn) {
   // Support: Chrome 50-51 only
   // Creating a new string by adding `' '` at the end, to hack around some bug in Chrome v50/51
   // (See https://github.com/angular/angular.js/issues/14487.)
   // TODO (gkalpak): Remove workaround when Chrome v52 is released
-  var fnText = Function.prototype.toString.call(fn).replace(STRIP_COMMENTS, '') + ' ',
+  return Function.prototype.toString.call(fn) + ' ';
+}
+
+function extractArgs(fn) {
+  var fnText = stringifyFn(fn).replace(STRIP_COMMENTS, ''),
       args = fnText.match(ARROW_ARG) || fnText.match(FN_ARGS);
   return args;
 }
@@ -849,7 +853,7 @@ function createInjector(modulesToLoad, strictDi) {
       if (!isBoolean(result)) {
         // Workaround for MS Edge.
         // Check https://connect.microsoft.com/IE/Feedback/Details/2211653
-        result = func.$$ngIsClass = /^(?:class\s|constructor\()/.test(Function.prototype.toString.call(func));
+        result = func.$$ngIsClass = /^(?:class\s|constructor\()/.test(stringifyFn(func));
       }
       return result;
     }
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -283,6 +283,14 @@ describe('injector', function() {
         it('should take args before first arrow', function() {
           expect(annotate(eval('a => b => b'))).toEqual(['a']);
         });
+
+        // Support: Chrome 50-51 only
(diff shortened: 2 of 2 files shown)


test/auto/injectorSpec.js
```diff
@@ -283,6 +283,14 @@ describe('injector', function() {
         it('should take args before first arrow', function() {
           expect(annotate(eval('a => b => b'))).toEqual(['a']);
         });
+
+        // Support: Chrome 50-51 only
```
petebacondarwin: +1 :-)

petebacondarwin:
Assuming the CI build passes this LGTM

src/auto/injector.js
```diff
@@ -849,7 +853,7 @@ function createInjector(modulesToLoad, strictDi) {
       if (!isBoolean(result)) {
         // Workaround for MS Edge.
         // Check https://connect.microsoft.com/IE/Feedback/Details/2211653
-        result = func.$$ngIsClass = /^(?:class\s|constructor\()/.test(Function.prototype.toString.call(func));
+        result = func.$$ngIsClass = /^(?:class\s|constructor\()/.test(stringifyFn(func));
```
thorn0: In the regex, let's replace `\s` with `\b`. Otherwise it doesn't work in the situations like following:
1. `class{ ... }`
2. `class/* foo */ Bar { ... }`

src/auto/injector.js
```diff
@@ -849,7 +853,7 @@ function createInjector(modulesToLoad, strictDi) {
       if (!isBoolean(result)) {
         // Workaround for MS Edge.
         // Check https://connect.microsoft.com/IE/Feedback/Details/2211653
```
mgol: Maybe you could change the comment at the same time to:
```
// Support: Edge 12-13 only
// See https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/6156135/
```

It's fixed in the Edge preview and the bug has moved.

src/auto/injector.js
gkalpak: I meant to do it as part of this PR, but forgot.
I will submit a new one.

src/auto/injector.js
gkalpak: Too late :stuck_out_tongue: 
Will do in a follow-up PR.

gkalpak:
Follow-up: #14533

PR #14533, opened 2016-04-28
fix($injector): fix class detection RegExp

gkalpak:
Bug fix

ES6 classes are not properly detected unless `class` is followed by a whitespace character.

ES6 classes are properly detected, even if class is **not** followed by a whitespace character.
E.g. `class{}` or `class/* ... */ {}`.

No

**Other information**:
Mentioned in https://github.com/angular/angular.js/pull/14531#discussion_r61410683.

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -851,9 +851,9 @@ function createInjector(modulesToLoad, strictDi) {
       }
       var result = func.$$ngIsClass;
       if (!isBoolean(result)) {
-        // Workaround for MS Edge.
-        // Check https://connect.microsoft.com/IE/Feedback/Details/2211653
-        result = func.$$ngIsClass = /^(?:class\s|constructor\()/.test(stringifyFn(func));
+        // Support: Edge 12-13 only
+        // See: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/6156135/
+        result = func.$$ngIsClass = /^(?:class\b|constructor\()/.test(stringifyFn(func));
       }
       return result;
     }
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -302,6 +302,26 @@ describe('injector', function() {
           expect(instance.aVal()).toEqual('a-value');
         });
 
+        if (/chrome/.test(navigator.userAgent)) {
+          they('should detect ES6 classes regardless of whitespace/comments ($prop)', [
+            'class Test {}',
+            'class Test{}',
+            'class //<--ES6 stuff\nTest {}',
+            'class//<--ES6 stuff\nTest {}',
+            'class {}',
+            'class{}',
+            'class //<--ES6 stuff\n {}',
+            'class//<--ES6 stuff\n {}',
+            'class/* Test */{}',
+            'class /* Test */ {}'
+          ], function(classDefinition) {
+            var Clazz = eval('(' + classDefinition + ')');
+            var instance = injector.invoke(Clazz);
+
+            expect(instance).toEqual(jasmine.any(Clazz));
+          });
+        }
+
         // Support: Chrome 50-51 only
         // TODO (gkalpak): Remove when Chrome v52 is relased.
         // it('should be able to invoke classes', function() {


src/auto/injector.js
```diff
@@ -851,9 +851,10 @@ function createInjector(modulesToLoad, strictDi) {
       }
       var result = func.$$ngIsClass;
       if (!isBoolean(result)) {
+        // Support: Edge 12-13 only
         // Workaround for MS Edge.
```
mgol: This line seems kind of moot now.

test/auto/injectorSpec.js
```diff
@@ -302,6 +302,19 @@ describe('injector', function() {
           expect(instance.aVal()).toEqual('a-value');
         });
 
+        they('should detect ES6 classes regardless of whitespace/comments ($prop)', [
+          'class Test {}',
+          'class Test{}',
+          'class {}',
+          'class{}',
+          'class/* Test */{}'
```
mgol: I'd add:
```js
class // aaa
Test {}
```
and:
```js
class // aaa
{}
```
and:
```js
class /* Test */ Test {}
```

src/auto/injector.js
gkalpak: Removed

test/auto/injectorSpec.js
gkalpak: Added

gkalpak:
There seem to be some failures on Safari. I'll have to investigate further :unamused:

gkalpak:
So, basically, we can't detect classes in Safari 9 :disappointed: 
This is a known issue (see e.g. #13785).

What should we do about it ? Skip the tests on Safari ?

mgol:
> What should we do about it ? Skip the tests on Safari ?

Yeah, I think so. Classes are not supported everywhere yet, it's natural some browsers might have buggy support for some time; people relying on classes in Angular will need to restrict support to well-behaving environments.

Note that Safari Technology Preview stringifies classes correctly so Safari 10 (coming out in ~6 months if the past is any indication) will be fine.

gkalpak:
Yeah, I meant what should we do about the tests (which ae failing on CI - as expected).
I will restrict them to Chrome for now (unless someone has a better idea).

mgol:
@gkalpak Why just to Chrome? What about Firefox?

gkalpak:
Firefox (at least up to v46) does not stringify classes in a way that allows class detection :disappointed: 

```
> (class Test { constructor(a) { this.a = a; } aMethod() { return this.a; } }).toString()
> "function Test(a) {
  "use strict";
   this.a = a; }"

**Comment by mgol** [comment]
Yeah, they have an open bug for it, it's not fixed even on Nightly (v49)
yet: https://bugzilla.mozilla.org/show_bug.cgi?id=1216630

File: test/auto/injectorSpec.js [hunk 3]
```diff
@@ -302,6 +302,26 @@ describe('injector', function() {
           expect(instance.aVal()).toEqual('a-value');
         });
 
+        if (/chrome/.test(navigator.userAgent)) {
```
dcherman: Why wrap this in a browser sniff for Chrome only?  Since this test doesn't look browser specific and is just more about the injector doing the right thing, I think it'd make more sense to feature detect the support for ES6 classes

```js
  var supportsClasses = (function() {
    try {
      eval("class C {}");
      return true;
    } catch (err) {
      return false;
    }
  }());

  if (supportsClasses) { ... }
```

mgol: The situation in tests is a little different; in jQuery e.g. we don't sniff in source but if a specific bug exists in a specific browser & version we do sniff in tests. This is not only because it's simpler to do but also to rule out the possibility that a broken support check will make the tests be skipped - we do want to verify that in these specific browsers they're being run.

I agree with you here, though as we're talking about a bleeding-edge feature that browsers are only implementing now and more & more of them will have it in a working state. That's different.

gkalpak: I generally agree, but there is not much we can do here unfortunately. There is more context in the comments (e.g. starting at https://github.com/angular/angular.js/pull/14533#issuecomment-215488679).

PR #14537, opened 2016-04-29
directives: added drag-drop events

MrOutput:

**Other information**:

### DIFF
--- a/src/ng/directive/ngEventDirs.js
+++ b/src/ng/directive/ngEventDirs.js
@@ -46,7 +46,7 @@ var forceAsyncEvents = {
   'focus': true
 };
 forEach(
-  'click dblclick mousedown mouseup mouseover mouseout mousemove mouseenter mouseleave keydown keyup keypress submit focus blur copy cut paste'.split(' '),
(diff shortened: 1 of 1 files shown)


gkalpak:
The problem with dnd related events is that they usually happen multiple times per second and are not suitable for integration with the $digest loop. Maybe just add the few that make sense (e.g. `drop`).

MrOutput:
gotcha, it is extremely helpful to have as an ngDirective. Perhaps, with these directives the digest loop should not fire for them? Is there some sort of exclusion array with the names of events to exclude, to stop the 250ms fire, from triggering the cycle?

gkalpak:
There is no exclusion array. But you can always create your own directives :smiley:

Narretz:
The fact that we have directives for mousemover etc. contradicts @gkalpak's arguments a bit. The logic is however that you usually have a more involved setup for these events anyway, so the ngEvent directives are probably not too useful in a lot of cases.

gkalpak:
Some `ng-mouse` directives aren't appropriate either - it would be a breaking change to remove them thought :smiley:

Narretz:
Let's not do it. These events are too specific to be very useful in the current design of the ngEvent directives.

eislasq:
I also have sugested D&D #12469 about 1 year ago and was rejected. 
But now I have published as a library https://github.com/eislasq/drag-and-drop

PR #14544, opened 2016-04-29
docs($location): describe your change...

FGasper:
docs update

inaccurate docs

correct docs

no

**Other information**:

docs for return of path() inaccurately describe the function’s return when a value is passed in.

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -482,7 +482,7 @@ var locationPrototype = {
    * ```
    *
    * @param {(string|number)=} path New path
-   * @return {string} path
+   * @return {(string|object)} path if called with no parameters, or `$location` if called with a parameter
    */
   path: locationGetterSetter('$$path', function(path) {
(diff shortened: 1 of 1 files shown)


FGasper:
I signed it!

src/ng/location.js
```diff
@@ -482,7 +482,7 @@ var locationPrototype = {
    * ```
    *
    * @param {(string|number)=} path New path
-   * @return {string} path
+   * @return {(string|object)} path if called with no parameters, or `$location` if called with a parameter
```
Narretz: I don't think objects are supported per se, they just work because we call toString on path, which was added here: https://github.com/angular/angular.js/issues/7054. I think we shouldn't have built in handling for intergers, as the docs clearly say the path must be a string.

Narretz: Disregard my last comment, I was in the wrong line.

Issue #14534, opened 2016-04-28
postDigestQueue poor performances


AntoineEsteve:
One of our page is very heavy. To decrease our watcher count and to accelerate the angular digest cycle we are using a lot the On-Time-Binding syntax `::`. We are also using [angular-bind-notifier][1] to avoid unnecessary watchs over our expressions on this page. 

This strategy allowed us to reduce the angular digest cycle considerably. 

But this strategy has **one down side**:  It uses the **`$$postDigest`** (`postDigestQueue`) to **`unwatch`** the expression after it has been evaluated successfully.

**So what ?**

At the end of the digest, angular will run through the `postDigestQueue`. As we used a lot of On-Time-Binding expressions, our `postDigestQueue` can grow up to more than 100 000 queued tasks.

The problem is that angular uses the [following code][3] to loop over the queue:

    while (postDigestQueue.length) {
      try {
        postDigestQueue.shift()();
      } catch (e) {
        $exceptionHandler(e);
      }
    }

> The shift method removes the element at the zeroeth index and shifts
> the values at consecutive indexes down, then returns the removed
> value. 
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

Yes, `Array.prototype.shift()` is a very expensive call when there is a lot of elements in the array.

Our digest cycle will sometimes take more than 20 seconds because of that.

When we change the previous code with the following one, it is way faster:

    for (var i = 0; i < postDigestQueue.length; i++) {
      try {
        postDigestQueue[i]();
      } catch (e) {
        $exceptionHandler(e);
      }
    }
    postDigestQueue.length = 0;

**Is there a reason why they did that ? 
Shouldn't we use that much the one time binding ?** 

I could see one reason: if a task add itself a need task in the queue. Is is possible ? (`$$postDigest` is a private queue) An answer would be to use `pop`instead of `shift` if the order of execution is not important, but is it ?

**Edit:** The order seems important because the `postDigestQueue` is used with the animations.

  [1]: https://github.com/kasperlewau/angular-bind-notifier
  [2]: https://github.com/angular/angular.js/blob/master/src/ng/rootScope.js
  [3]: https://github.com/angular/angular.js/blob/master/src/ng/rootScope.js#L851

Narretz:
One way to see if a change to the code causes any problems is to make the change in the code and the run the tests to see if any of them fail. You can open a PR with the change, and Travis will run the tests if you can't do it locally.

[...]

PR #14545, opened 2016-04-29
perf($rootScope): make postDigestQueue more efficient

thorn0:

Performance improvement

#14534

Using [this implementation](http://code.stephenmorley.org/javascript/queues/) of queue instead of an array.

`postDigestQueue` is available as `$rootScope.$$postDigestQueue`. So it can be considered a breaking change.

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -769,15 +769,19 @@ function $RootScopeProvider() {
           dirty = false;
           current = target;
 
-          while (asyncQueue.length) {
+          // It's safe for asyncQueuePosition to be a local variable here because this loop can't
+          // be reentered recursively. Calling $digest from a function passed to $applyAsync would
+          // lead to a '$digest already in progress' error.
+          for (var asyncQueuePosition = 0; asyncQueuePosition < asyncQueue.length; asyncQueuePosition++) {
             try {
-              asyncTask = asyncQueue.shift();
+              asyncTask = asyncQueue[asyncQueuePosition];
               asyncTask.scope.$eval(asyncTask.expression, asyncTask.locals);
             } catch (e) {
               $exceptionHandler(e);
             }
             lastDirtyWatch = null;
           }
+          asyncQueue.length = 0;
 
           traverseScopesLoop:
           do { // "traverse the scopes" loop
@@ -848,13 +852,15 @@ function $RootScopeProvider() {
 
         clearPhase();
 
-        while (postDigestQueue.length) {
+        // postDigestQueuePosition isn't local here because this loop can be reentered recursively.
+        while (postDigestQueuePosition < postDigestQueue.length) {
           try {
-            postDigestQueue.shift()();
+            postDigestQueue[postDigestQueuePosition++]();
           } catch (e) {
             $exceptionHandler(e);
           }
         }
+        postDigestQueue.length = postDigestQueuePosition = 0;
       },
 
 
@@ -1308,6 +1314,8 @@ function $RootScopeProvider() {
     var asyncQueue = $rootScope.$$asyncQueue = [];
     var postDigestQueue = $rootScope.$$postDigestQueue = [];
     var applyAsyncQueue = $rootScope.$$applyAsyncQueue = [];
+
+    var postDigestQueuePosition = 0;
 
     return $rootScope;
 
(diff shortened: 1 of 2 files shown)


gkalpak:
This looks pretty good to me. I think they tiny extra bytes are totally worth it if we can get a speed boost on every `$digest`.

OOC, have you done any benchmarks ?

thorn0:
There is a benchmark on [the page](http://code.stephenmorley.org/javascript/queues/) from where I took the implementation of queue.

jbedard:
Since `postDigestQueue` is only used in one location and it is emptied each and every time... why not just loop through it and then clear it once at the end?
```java
for (var i=0, len=postDigestQueue.length; i<len; i++) {
   try {
      postDigestQueue[i]();
   } catch (e) {
       $exceptionHandler(e);
   }
}
postDigestQueue.length = 0;
```
That seems simpler and does only one `.length = 0` instead of `.slice(offset)` log(n) times?

thorn0:
Because this loop can be called recursively inside `postDigestQueue[i]();`.

jbedard:
Ah, no wonder that simple solution was never used ^_^

lgalfaso:
This change looks interesting, but the code at http://code.stephenmorley.org/javascript/queues/ is not MIT licensed, so it cannot be merged into Angular.

thorn0:
On the other hand, it's quite difficult to implement queue in such a way that it wouldn't resemble and look like a work derived from this CC0 licensed code.

thorn0:
@lgalfaso I rewrote it a bit. Please have a look. Now `slice` isn't called unless the underlying array is longer than 200, which makes it even faster because previously `slice` was called too often for short queues.

gkalpak:
Similar to what @jbedard proposed, couldn't we just empty the queue when it's empty ?
I mean, we know that it is always traversed until it's emptied and previous values are already being set to `null` (so the memory consumption shouldn't be a concern).

Not a big deal, but it could save a few `.slice()` calls for really large queues.

petebacondarwin:
I would be interested in benchmarking this against a linked list queue.

thorn0:
@petebacondarwin Linked list seems to be a bit slower than that queue. You can get the script I used for benchmarking from [here](https://gist.github.com/thorn0/b6856656b3c7684d370f9967692d95ce).

thorn0:
@gkalpak Although my benchmark script doesn't show any improvement from doing this, I moved the `slice` call to a separate method `purge`, which has to be called manually.

Also I used this queue implementation for `asyncQueue` and `applyAsyncQueue`.

dcherman:
As a play on what @jbedard suggested, what if you looped through the `$$postDigestQueue` but stored the variable pointing to head in a shared closure so that each call to `$digest` would start draining the queue at the appropriate location?

```js
// At the top of the `$RootScopeProvider` closure
var postDigestQueuePosition = 0;

while(postDigestQueuePosition < postDigestQueue.length) {
  try {
    postDigestQueue[postDigestQueuePosition++]();
  } catch (e) {
    $exceptionHandler(e);
  }
}

postDigestQueue.length = postDigestQueuePosition = 0;
```

How does that compare perf wise to the queue?  It's also a more minimalistic change

thorn0:
@dcherman You're right. What you wrote makes perfect sense. I've updated the commit. Of course, it's better than the queue as it doesn't involve any calls to methods like `slice`. Your solution retains the references to processed items until the array is cleared, but I'm not sure whether it's important.

jbedard:
Nice :+1:

petebacondarwin:
This looks pretty good now. Thanks @thorn0

src/ng/rootScope.js
```diff
@@ -848,13 +849,14 @@ function $RootScopeProvider() {
 
         clearPhase();
 
-        while (postDigestQueue.length) {
+        while (postDigestQueuePosition < postDigestQueue.length) {
```
gkalpak: Why not keep `postDigestQueuePosition` local and use a for loop as with `asyncQueue` above ?

thorn0: See above. This loop can be reentered recursively.

gkalpak: You are right. Which makes me wonder: Could the same happen with the `asyncQueue` ?

dcherman: No - the `asyncQueue` is drained during the `$digest` cycle where `postDigestQueue` is drained outside of it.  As a result, you can't recurse into it since you'd have gotten a digest in progress error.

thorn0: I've added these explanations as comments in the code.

PR #14560, opened 2016-05-04
update npm install instructions.

markdesign:
current npm install instruction does not work:
`npm install @angular/router --save`

working version is:
`npm install @angular/router@0.2.0 --save`


petebacondarwin:
@markdesign thanks for this submission. Unfortunately you never signed the CLA so we weren't able to use it. I fixed the issue with a different commit instead.

PR #14571, opened 2016-05-06
fix($LogProvider) : Allow console methods overriding

floribon:

Bug fix

If a user override a console method on its personal code, such as `console.error`

```
console.error = (function(cerror) {
  return function() {
    const args = Array.from(arguments);
    args.unshift('[my custom prefix] ');
    cerror.apply(console, args);
  };
})(console.error);
```

And then Angular logs something, such as an error:

```
scope.$apply(function() { throw new Error('boom'); })
// logs "boom"
```

We don't see the `[my custom prefix]` appended to the error message. The original `console.error` is being used.

This tiny change will make Angular use the overriden console methods.

```
scope.$apply(function() { throw new Error('boom'); })
// logs "[my custom prefix]  boom"
```

No

**Other information**:

If a user overrides the methods console.log, console.error, etc, this
change will use them rather than the default ones.

### DIFF
--- a/src/ng/log.js
+++ b/src/ng/log.js
@@ -150,7 +150,8 @@ function $LogProvider() {
           forEach(arguments, function(arg) {
             args.push(formatError(arg));
           });
-          return logFn.apply(console, args);
+          // Using logFn here prevents custom console methods overriding
+          return console[type].apply(console, args);
         };
       }
 


floribon:
I signed it! 👍

gkalpak:
First if we were to make this change, we should be consistent and do the same thing on IE as well.

But I'm not too eager to support this usecase - it doesn't seem like a good idea to dynamically overwrite global objects. If this is something specific to your Angular app, you should be decorating the `$log` service instead.

In any case, you can achieve the same result by overwriting the methods before loading Angular.

floribon:
Alright, agreed, I as a user should find a work around. I'll check the `$log` service, thanks.
I'm going to close it then.

Issue #13590, opened 2015-12-19
Directive's controller setter called before class instantiation (view attributes assigned before initialization)


CodySchaaf:
This seems to be intentional in order to allow access to bound variables during a directive's controller instantiation, but it is really unintuitive.

As you can see in this [codepen](http://codepen.io/codyschaaf/pen/jWWPEq) this can cause issues with variables that should have been instantiated in the constructor not being available when a method is called. 

Basically in my situation we have a controller that needs to notify its link function when a bound variable is changed (widget.title), to do this I set up a setter for the variable that calls notify when it is called. 

So the parent directive can change the title, then the setter is called which notifies the widget's link function. Although since the parentDir changes the title in it's constructor we get a console error where notify is called before the array of subscribers is set up. You can uncomment the code to see the order of the console logs. This code would allow my version to work, although it would end up masking other issues. 

The code responsible for this is [here](https://github.com/angular/angular.js/blob/d077966ff1ac18262f4615ff1a533db24d4432a7/src/ng/controller.js#L117)

Following this pattern:

``` ts
var instance = Object.create(directiveDefinition.controller.prototype);

instance.widgetTitle = "New Title"

directiveDefinition.controller.apply(instance);
```

Is there an existing pattern for getting around this issue? Seems sort of like a bug, although it would probably require a pretty big breaking change that would make a lot of people angry.

gkalpak:
Indeed this seems to be an unwanted side-effect of our way of instantiating controllers with bindings.
Unfortunately, "fixing" this would break other (much more common) usecases, as you already mentioned.

So, I'm afraid we'll have to stick with the current implementation and people using setters with side-effects should handle this "manually" (unless someone can come up with a clever work-around).

Narretz:
Since 1.6, the bindings are after the controller's constructor is called, which should resolve this issue.

Issue #12865, opened 2015-09-16
Prototype Accessors Called Before Constructor/Injection.


rcollette:
I'm seeing a behavior that seems to be both undocumented, and leads to an "impedance mismatch" with TypeScript, which I think the Angular team is generally interested in.

In AngularJs, prototype accessors  of a directive's controller are called before the controller's constructor.  I am not sure where this is documented or why it happens this way.   If an accessor needs to use an injected service, you have to abandon that first set call.   I don't know if returning from a setter and doing nothing has ramifications in a real world setting, but in a simple example it seems to solve the problem.   "Prototype accessors" (my terminology may be incorrect) are what are generated by TypeScript and I believe are what ES6 get/set accessors do.

```javascript
  var TestController1 = (function() {
    function TestController1(testService) {
      console.log("TestController1 constructor");
      this.testService = testService;
    }
    //This is how TypeScript generates properties.
    Object.defineProperty(TestController1.prototype, "aprop", {
      get: function() {
        console.log("TestController1 Get Value:",this._aProperty);
        return this._aProperty;
      },
      set: function(value) {
        console.log("TestController1 setter called with value: ", value);
        if(!this.testService){
          //in order to make this work I have to exit on the first call.
          console.log("TestController1: exiting, no service yet");
          return;
        }
        this._aProperty = this.testService.doSomething(value);
      },
      enumerable: true,
      configurable: true
    });
    return TestController1;
  })();
```

To make the "first" (which may not actually be the first) set call work,  I can define a property on _this captured in the constructor.  In this case the accessor has access to the injected service it needs.  But now I'm back to the bloated `defineProperty(ies)` JS syntax rather than being able to use TypeScript's get/set syntax.

```javascript
  var TestController2 = (function() {
    function TestController2(testService) {
      console.log("TestController2 constructor");
      var _this = this;
      this.testService = testService;
      
      //This is how I would typically create properties with angular
      Object.defineProperty(_this, "aprop", {
        get: function() {
          console.log("TestController2 Get Value:",this._aProperty);
          return this._aProperty;
          return _this._aProperty;
        },
        set: function(value) {
          console.log("TestController2 setter called with value: ", value);
          this._aProperty = _this.testService.doSomething(value);
        },
        enumerable: true,
        configurable: true
      });
    }
    return TestController2;
  })();
```
Console output:
```
script.js:61 TestController2 constructor
script.js:28 TestController1 setter called with value:  x
script.js:31 TestController1: exiting, no service yet
[...]

PR #14580, opened 2016-05-10
fix($compile): don't run unnecessary update to one-way bindings

### LINKED ISSUE #14546: Outer change is stronger than the inner one when they are both changed and using one-way binding
Bug (?)

I'm not sure if bug or by design, but when using one-way (`<`), if both the outer and inner scope have been changed, the outer is stronger, **even if the inner's change happened after the outer one**.
http://jsbin.com/zunoraquju/1/edit?html,js,output

I would simply expect that the last change is stronger, that's what happens till now with two-way (`=`). You can change my attached example to `=` and it will work.

Actually a common case - if the passed value is `X` I want to change it to be `Y` (in constructor / `$onInit`) and display it, but preserve the one-way so the outer scope won't get the update.

`1.5+`

[Relevant test](https://github.com/angular/angular.js/commit/4ac23c0ac59c269d65b7f78efec75d060121bd18#diff-348c2f3781ed66a24894c2046a52c628R4268) (I would expect it to be `misko`)

petebacondarwin:

Bug Fix

The watch handler was triggering on its first invocation, even though
its change had already been dealt with when setting up the binding.

The handler does not trigger an update on its first invocation

Nope - I don't think so. Although I had to modify a bunch of tests to ensure that a digest occurred between compilation and the first checks.

  ~~\- [ ] Docs have been added / updated (for bug fixes / features)~~

**Other information**:

Closes #14546

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -3823,6 +3823,7 @@ describe('$compile', function() {
 
           $rootScope.$apply('a = 7');
           element = $compile('<c1 prop="a" attr="{{a}}"></c1>')($rootScope);
+
           expect(log).toEqual([
             {
               prop: jasmine.objectContaining({currentValue: 7}),
@@ -3862,6 +3863,7 @@ describe('$compile', function() {
         inject(function($compile, $rootScope) {
           $rootScope.$apply('a = 7');
           element = $compile('<c1 prop="a" attr="{{a}}"></c1>')($rootScope);
+
           expect(log).toEqual([
             {
               prop: jasmine.objectContaining({currentValue: 7}),
@@ -4744,6 +4746,7 @@ describe('$compile', function() {
     describe('one-way binding', function() {
       it('should update isolate when the identity of origin changes', inject(function() {
         compile('<div><span my-component ow-ref="obj">');
+
         expect(componentScope.owRef).toBeUndefined();
         expect(componentScope.owRefAlias).toBe(componentScope.owRef);
 
@@ -4780,6 +4783,7 @@ describe('$compile', function() {
 
       it('should update isolate when both change', inject(function() {
         compile('<div><span my-component ow-ref="name">');
+
         $rootScope.name = {mark:123};
         componentScope.owRef = 'misko';
 
@@ -4796,6 +4800,101 @@ describe('$compile', function() {
         expect(componentScope.owRefAlias).toBe($rootScope.name);
       }));
 
+      describe('initialization', function() {
+        var component, log;
+
+        beforeEach(function() {
+          log = [];
+          angular.module('owComponentTest', [])
+            .component('owComponent', {
+              bindings: { input: '<' },
+              controller: function() {
+                component = this;
+                this.input = 'constructor';
+                log.push('constructor');
+
+                this.$onInit = function() {
+                  this.input = '$onInit';
+                  log.push('$onInit');
+                };
+
+                this.$onChanges = function(changes) {
+                  if (changes.input) {
+                    log.push(['$onChanges', changes.input]);
+                  }
+                };
+              }
+            });
+        });
+
+        it('should not update isolate again after $onInit if outer has not changed', function() {
+          module('owComponentTest');
+          inject(function() {
+            $rootScope.name = 'outer';
+            compile('<ow-component input="name"></ow-component>');
+
+            expect($rootScope.name).toEqual('outer');
+            expect(component.input).toEqual('$onInit');
+
+            $rootScope.$apply();
+
+            expect($rootScope.name).toEqual('outer');
+            expect(component.input).toEqual('$onInit');
+
+            expect(log).toEqual([
+              'constructor',
+              ['$onChanges', jasmine.objectContaining({ currentValue: 'outer' })],
+              '$onInit'
+            ]);
+          });
+        });
+
+        it('should update isolate again after $onInit if outer has changed (before initial watchAction call)', function() {
+          module('owComponentTest');
+          inject(function() {
+            $rootScope.name = 'outer1';
+            compile('<ow-component input="name"></ow-component>');
+
+            expect(component.input).toEqual('$onInit');
+            $rootScope.$apply('name = "outer2"');
+
+            expect($rootScope.name).toEqual('outer2');
+            expect(component.input).toEqual('outer2');
+            expect(log).toEqual([
+              'constructor',
+              ['$onChanges', jasmine.objectContaining({ currentValue: 'outer1' })],
+              '$onInit',
+              ['$onChanges', jasmine.objectContaining({ currentValue: 'outer2', previousValue: 'outer1' })]
+            ]);
+          });
+        });
+
+        it('should update isolate again after $onInit if outer has changed (before initial watchAction call)', function() {
+          angular.module('owComponentTest')
+            .directive('changeInput', function() {
+              return function(scope, elem, attrs) {
+                scope.name = 'outer2';
+              };
+            });
+          module('owComponentTest');
+          inject(function() {
+            $rootScope.name = 'outer1';
+            compile('<ow-component input="name" change-input></ow-component>');
+
+            expect(component.input).toEqual('$onInit');
+            $rootScope.$digest();
+
+            expect($rootScope.name).toEqual('outer2');
+            expect(component.input).toEqual('outer2');
+            expect(log).toEqual([
+              'constructor',
+              ['$onChanges', jasmine.objectContaining({ currentValue: 'outer1' })],
+              '$onInit',
+              ['$onChanges', jasmine.objectContaining({ currentValue: 'outer2', previousValue: 'outer1' })]
+            ]);
+          });
+        });
+      });
 
       it('should not break when isolate and origin both change to the same value', inject(function() {
         $rootScope.name = 'aaa';
@@ -4819,7 +4918,6 @@ describe('$compile', function() {
         $rootScope.name = {mark:123};
         compile('<div><span my-component ow-ref="name">');
 
-        $rootScope.$apply();
         expect($rootScope.name).toEqual({mark:123});
         expect(componentScope.owRef).toBe($rootScope.name);
         expect(componentScope.owRefAlias).toBe($rootScope.name);
@@ -4836,7 +4934,6 @@ describe('$compile', function() {
         $rootScope.obj = {mark:123};
         compile('<div><span my-component ow-ref="obj">');
 
-        $rootScope.$apply();
         expect($rootScope.obj).toEqual({mark:123});
         expect(componentScope.owRef).toBe($rootScope.obj);
 
@@ -4849,6 +4946,7 @@ describe('$compile', function() {
 
       it('should not throw on non assignable expressions in the parent', inject(function() {
         compile('<div><span my-component ow-ref="\'hello \' + name">');
+
         $rootScope.name = 'world';
         $rootScope.$apply();
         expect(componentScope.owRef).toBe('hello world');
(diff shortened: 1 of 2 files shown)


test/ng/compileSpec.js
```diff
@@ -3756,6 +3760,7 @@ describe('$compile', function() {
         module('my');
         inject(function($compile, $rootScope) {
           element = $compile('<c1 prop1="val"></c1>')($rootScope);
+          $rootScope.$apply();
```
gkalpak: Why is this needed ?

test/ng/compileSpec.js
```diff
@@ -4796,6 +4814,35 @@ describe('$compile', function() {
...
+          });
+        module('owComponentTest');
+        inject(function() {
+          $rootScope.name = 'outer';
+          compile('<ow-component input="name"></ow-component>');
+          $rootScope.$apply();
+
+          expect($rootScope.name).toEqual('outer');
+          expect(component.input).toEqual('$onInit');
+        })
```
gkalpak: Missing semicolon.

test/ng/compileSpec.js
```diff
@@ -4796,6 +4814,35 @@ describe('$compile', function() {
...
+            bindings: { input: '<' },
+            controller: function() {
+              component = this;
+              this.input = 'constructor';
+              this.$onInit = function() {
+                this.input = '$onInit';
+              };
+              this.$onChanges = function(changes) {
+                if (changes.input) {
+                  log.push(changes.input);
```
gkalpak: You are pushing to `log`, but never checking what was actually pushed.

petebacondarwin: Fixed.

test/ng/compileSpec.js
petebacondarwin: If we have no digest before changing the value of `val` then effectively the system thinks that `val` has not changed, so `$onChanges` is not called.

gkalpak:
One testcase I believe should work, but doesn't, is when the parent value changes after initializing the bindings but before the first call to the one-way binding watcher callback.
Here are a couple of unit tests demonstrating the issue. They are essentially equivalent, the first being more straight-forward and the latter being more "real-world-ish".

```js
      it('should update isolate again after $onInit if outer has changed (before initial watchAction call)', function() {
        var log = [];
        var component;
        angular.module('owComponentTest', [])
          .component('owComponent', {
            bindings: { input: '<' },
            controller: function() {
              component = this;
              this.input = 'constructor';
              this.$onInit = function() {
                this.input = '$onInit';
              };
              this.$onChanges = function(changes) {
                if (changes.input) {
                  log.push(changes.input);
                }
              };
            }
          });
        module('owComponentTest');
        inject(function() {
          $rootScope.name = 'outer1';
          compile('<ow-component input="name"></ow-component>');

          expect(component.input).toEqual('$onInit');
          $rootScope.$apply('name = "outer2"');

          expect($rootScope.name).toEqual('outer2');
          expect(component.input).toEqual('outer2');
        });
      });

      it('should update isolate again after $onInit if outer has changed (before initial watchAction call)', function() {
        var log = [];
        var component;
        angular.module('owComponentTest', [])
          .directive('changeInput', function() {
            return function(scope, elem, attrs) {
              scope.name = 'outer2';
            };
          })
          .component('owComponent', {
            bindings: { input: '<' },
            controller: function() {
              component = this;
              this.input = 'constructor';
              this.$onInit = function() {
                this.input = '$onInit';
              };
              this.$onChanges = function(changes) {
                if (changes.input) {
                  log.push(changes.input);
                }
              };
            }
          });
        module('owComponentTest');
        inject(function() {
          $rootScope.name = 'outer1';
          compile('<ow-component input="name" change-input></ow-component>');

          expect(component.input).toEqual('$onInit');
          $rootScope.$digest();

          expect($rootScope.name).toEqual('outer2');
          expect(component.input).toEqual('outer2');
        });
      });
```

gkalpak:
BTW, in order to support these usecases (and fix the original bug), we could change `if (oldValue === newValue) return` to:

```js
if (oldValue === newValue) {
  if (oldValue === initialValue) return;
  oldValue = destination[scopeName];
}
```

gkalpak:
BTW, in order to support these usecases (and fix the original bug), we could change the code from:

```js
destination[scopeName] = parentGet(scope);
...

removeWatch = scope.$watch(parentGet, function parentValueWatchAction(newValue, oldValue) {
  if (oldValue === newValue) return;
  ...
```

to:

```js
var initialValue = destination[scopeName] = parentGet(scope);
...

removeWatch = scope.$watch(parentGet, function parentValueWatchAction(newValue, oldValue) {
  if (oldValue === newValue) {
    if (oldValue === initialValue) return;
    oldValue = initialValue;
  }
  ...
```

test/ng/compileSpec.js
```diff
@@ -4796,6 +4800,96 @@ describe('$compile', function() {
...
+            $rootScope.name = 'outer';
+            compile('<ow-component input="name"></ow-component>');
+
+            expect($rootScope.name).toEqual('outer');
+            expect(component.input).toEqual('$onInit');
+            expect(log).toEqual([
+              'constructor',
+              ['$onChanges', jasmine.objectContaining({ currentValue: 'outer' })],
+              '$onInit'
+            ]);
```
gkalpak: Wouldn't this test pass even without the fix ? (I haven't tested it, but I think so.)
I think it's a good idea to add a `$rootScope.$digest()` and another round of expectation to make sure that nothing is overwritten when the watch callback is executed.

test/ng/compileSpec.js
```diff
@@ -4796,6 +4800,96 @@ describe('$compile', function() {
...
+          });
+        });
+
```
gkalpak: <sub>Inconsistent number of newlines between tests :grin:</sub>

gkalpak:
A couple of minor test-related comments.
Otherwise LGTM :+1:

test/ng/compileSpec.js
petebacondarwin: OK. It did fail before the fix. But then I might have changed it again since then.

test/ng/compileSpec.js
petebacondarwin: Shock horror! Isn't there a JSCS rule for that??

test/ng/compileSpec.js
petebacondarwin: It totally fails as it is without the fix. But I will add the extra digest and check

khoomeister:
Hey, this just broke a whole bunch of our unit tests which I was hoping to get some clarification on.

So right now, we have something like this:

```typescript
export someComponent = {
  ...

  set config(config: Config) {
    if (!this.state) return
    this.state.config = config
  }
}

angular.component('someComponent', someComponent)
```

The reason for the `if (!this.state) return` line is because [prototype accessors are called before the constructor](https://github.com/angular/angular.js/issues/12865), and we need to do a number of things in the constructor (not least setting up `this.state`) before the config setter is called again.

We can't move the constructor code since we rely on injected dependencies to do some setup.  Perhaps there's a better pattern for what we want to do but not sure how our use case can be accomplished.  Do you have any ideas on what to do here?

gkalpak:
TBH, I don't get what the problem is. Could you share some more code and/or tests.

khoomeister:
Hey @gkalpak the issue we're encountering is basically very similar to #13590 - I guess we just need to work around it.

khoomeister:
Just to give you some context, setters got called twice on startup (once before the component controller instantiates and once after).  We relied on the behavior that setters got called after instantiation since dependencies are available at that point.  Since this change returns early for matching values, our setter doesn't execute.

khoomeister:
OK I see the angular way of doing this is to hook into `$onChanges` :smiley: - thx!

gkalpak:
Glad you got it working with `$onChanges` :smiley:
I'm still not sure what your code does and how it is afected by this, but you might also want to look into `$onInit`.

bcherny:
@gkalpak I think what @khoomeister is saying is that in 1.5.5, the order of calls was setter-constructor-setter. So on the first fire `this.state` was undefined and the setter returned early, then the constructor defined `this.state`, then the setter fired again (successfully this time, because `this.state` was now defined).

In 1.5.6, that 2nd call to the setter never happens, so the setter is never fully executed.

This is a *breaking change* for anyone that relies on something being defined on `this` in a setter!

gkalpak:
@bcherny, it will be easier to investigate with an actual example, but this is expected bahavior as far as I understand. The fact that it worked differently in 1.5.5 (i.e. overwriting the inner value even though the outer vaue hadn't changed) was a bug. This commit is restoring the expacted/correct behavior and thus is not considered a breaking change (although it would break a usecase that relied on the buggy behavior).

The proper way to ensure something is run after the bindings have been initialized is using an appropriate hook (e.g. `$onInit` or `$onChanges` - depending on the usecase).

bcherny:
@gkalpak Demo here - http://codepen.io/anon/pen/EyVyzB

gkalpak:
@bcherny, fwiw it is expected behavior. The actual "problem" with your usecase is not how one-way binding watches are triggered, but the fact that we are pre-assigning the binding values on the controller instance before actually instantiating it.

Although this is supposed to be a feature, this has also created its fair share of headaches as well, so it is scheduled for deprecation/removal in future versions. (We might even provide a way to opt-out in 1.5.x as well.)
Note, that the behavior will be different, depending on whether or not your controller is recognized as a class (vs plain old constructor function).

bcherny:
A backwards compatible way to make it work (at least in my case) would be getting rid of the 1st binding fire (caused by pre-assigning) rather than the 2nd binding fire -- is that what you mean?

So in 1.5.5 it was:

`call setter - call constructor - call setter`

In 1.5.6 it's:

`call setter - call constructor`

And what I'm proposing is:

`call constructor - cal setter`

petebacondarwin:
`call constructor - call setter` is what we are aiming for in 1.6 but this would certainly break even more apps than the current change in 1.5.6

bcherny:
Good to hear - at least for me, that would be an improvement!

gkalpak:
Note that this is the behavior you might get today if your controller is a class (a real class).
I think it would be nice to introduce the new behavior as an opt-in in 1.5.x, so people that are eager for this get a chance to try it out early and we gather some feedback before 1.6 is out.
Something like:

```js
$compileProvider.enableBindingsPreAssignment(false);
```

The feature would be enabled by default on 1.5.x (so backwards compatible), but can be turned off.
Then on 1.6.x we will have it disabled by default (optionally allowing developers to turn it on).

petebacondarwin:
@gkalpak I agree. But I think it should be simply:

```
$compileProvider.preAssignBindings(false);
```

RobbertWolfs:
Is there any progress on this?
Because we just upgraded from 1.5.5 to 1.5.8 and we also have binding issues..

We have a component that requires a boolean value but it is always set to the default value.. not the one we use in the binding..

Some code: 

**in our test:**
`
   it('it has a required binding', () => {
               $scope.required = true';
               element = createDirective('<image-upload required="required"></image-upload>', $scope);
               controller = element.controller('imageUpload'); 
               expect(controller.required).to.eql(true);
});
`

**in our component**
`
   class ImageUploadController {
       // @ngInject
       constructor() {
           // DEFAULT BINDING
           this.required = false;
       }
   }
 
   export default {
       template: `my template`,
       controller: ImageUploadController,
       bindings: {
           required: '<?',
       },
   };
`

gkalpak:
@RobbertWolfs, in case you haven't noticed, the related PR has been merged and the `preAssignBindingsEnabled` configuration switch will be available in an upcoming 1.5.x release :tada:

PR #14589, opened 2016-05-11
fix($parse): call once stable bind-once expressions with filter

### LINKED ISSUE #14583: in a one-time filtered expression (:: func() | filter), the filter is invoked the second time even though its input has not changed
https://plnkr.co/edit/S9VhW6TdvkvtpGhgQBKQ?p=preview

for a regular expression the filter is invoked only once

lgalfaso:
fix, but given the nature I would submit this for 1.6 only

stateless filters on bind-once expression are called multiple times.

stateless filters on bind-once expression are called once.

No, but developers might expect the multiple calls (even when they should not)

**Other information**:

When an expression has the following:
- Bind-once
- A stateless filter at the end
- Optionally an expression interceptor

Then call the filter in the expression once.

Closes: #14583

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -2050,14 +2050,20 @@ function $ParseProvider() {
       }, listener, objectEquality, prettyPrintExpression);
     }
 
-    function oneTimeWatchDelegate(scope, listener, objectEquality, parsedExpression) {
+    function oneTimeWatchDelegate(scope, listener, objectEquality, parsedExpression, prettyPrintExpression) {
       var unwatch, lastValue;
-      return unwatch = scope.$watch(function oneTimeWatch(scope) {
+      if (parsedExpression.inputs) {
+        return unwatch = inputsWatchDelegate(scope, oneTimeListener, objectEquality, parsedExpression, prettyPrintExpression);
+      } else {
+        return unwatch = scope.$watch(oneTimeWatch, oneTimeListener, objectEquality);
+      }
+      function oneTimeWatch(scope) {
         return parsedExpression(scope);
-      }, function oneTimeListener(value, old, scope) {
+      }
+      function oneTimeListener(value, old, scope) {
         lastValue = value;
         if (isFunction(listener)) {
-          listener.apply(this, arguments);
+          listener(value, old, scope);
         }
         if (isDefined(value)) {
           scope.$$postDigest(function() {
@@ -2066,7 +2072,7 @@ function $ParseProvider() {
             }
           });
         }
-      }, objectEquality);
+      }
     }
 
     function oneTimeLiteralWatchDelegate(scope, listener, objectEquality, parsedExpression) {
@@ -2076,7 +2082,7 @@ function $ParseProvider() {
       }, function oneTimeListener(value, old, scope) {
         lastValue = value;
         if (isFunction(listener)) {
-          listener.call(this, value, old, scope);
+          listener(value, old, scope);
         }
         if (isAllDefined(value)) {
           scope.$$postDigest(function() {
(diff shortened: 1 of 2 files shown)


src/ng/parse.js
```diff
@@ -2050,14 +2050,20 @@ function $ParseProvider() {
...
+      }
+      function oneTimeWatch(scope) {
         return parsedExpression(scope);
-      }, function oneTimeListener(value, old, scope) {
+      }
+      function oneTimeListener(value, old, scope) {
         lastValue = value;
         if (isFunction(listener)) {
-          listener.apply(this, arguments);
+          listener(value, old, scope);
```
jbedard: This (and the removal of `.call` below) seem to be unrelated to the rest of the PR?

I think the change is ok since it is equivalent to the [normal listener invoking](https://github.com/angular/angular.js/blob/4487d4a8a477ad440500e3574c141da48d13ab3d/src/ng/rootScope.js#L803). But I think the idea of doing .call/apply here was to avoid duplicating that logic.

lgalfaso: `this` is always undefined, so the old form adds nothing.

jbedard: The old form just preserved whatever the normal listener invoking did, so the logic (or lack of) was not duplicated. I'm fine with the change, but it seems unrelated to the actual fix?

src/ng/parse.js
```diff
@@ -2050,14 +2050,20 @@ function $ParseProvider() {
...
     }
 
-    function oneTimeWatchDelegate(scope, listener, objectEquality, parsedExpression) {
+    function oneTimeWatchDelegate(scope, listener, objectEquality, parsedExpression, prettyPrintExpression) {
       var unwatch, lastValue;
-      return unwatch = scope.$watch(function oneTimeWatch(scope) {
+      if (parsedExpression.inputs) {
+        return unwatch = inputsWatchDelegate(scope, oneTimeListener, objectEquality, parsedExpression, prettyPrintExpression);
+      } else {
+        return unwatch = scope.$watch(oneTimeWatch, oneTimeListener, objectEquality);
```
gkalpak: Why not use `parsedExpression` directly ?

test/ng/parseSpec.js
```diff
@@ -3232,6 +3232,18 @@ describe('parser', function() {
           expect(fn()).toEqual(undefined);
         }));
 
+        it('should invoke a stateless filter once when the parsed expression has an interceptor',
+           inject(function($parse, $rootScope) {
+          var countFilter = jasmine.createSpy();
+          var interceptor = jasmine.createSpy();
+          countFilter.and.callThrough();
```
gkalpak: Is this necessary ?

src/ng/parse.js
lgalfaso: `parsedExpression` has `$$watchDelegate` :)

test/ng/parseSpec.js
lgalfaso: In a strict sense, no, but there is some special logic for one time binding and when something is evaluated to `undefined` and I want to be sure that whoever reads the test, that we are not running into that case.

gkalpak: But, how is `.and.callThrough()` changing anything. You don't have an actual function to call through anyway.

lgalfaso: I do not understand this. `countFilter` is registered as a filter and then this filter is used in the expression `foo() | count`. `foo` is a function that returns `1` and i want to be sure that the expression evaluates to something other than `undefined`

gkalpak: `countFilter` is initialized as `jasmine.createSpy()` (2 lines above). So you are basically doing `jasmine.createSpy().and.callThrough()`. What you are calling through to is `function() {}` which should return undefined no matter what (e.g. see [demo][1]).

[1]: https://codepen.io/ExpertSystem/pen/LNoPqe?editors=0010

lgalfaso: you are right, my mistake. Fixed.

gkalpak:
Should we backport this to 1.5.x (since it's a fix) ?

lgalfaso:
> Should we backport this to 1.5.x (since it's a fix) ?

I lean towards not to backport this. The reason being that it does add some behavior change.

PR #14593, opened 2016-05-12
feat($resource): Allow $resource paramDefaults to accept a promise.

kevmo314:

New feature.

$resource paramDefaults accepts either a value or a function.

$resource paramDefaults now accepts either a value or a function or a promise.

No.

**Other information**:

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -552,8 +552,9 @@ angular.module('ngResource', ['ng']).
           actionParams = extend({}, paramDefaults, actionParams);
           forEach(actionParams, function(value, key) {
             if (isFunction(value)) { value = value(); }
-            ids[key] = value && value.charAt && value.charAt(0) === '@' ?
-              lookupDottedPath(data, value.substr(1)) : value;
+            ids[key] = $q.when(
+              value && value.charAt && value.charAt(0) === '@' ?
+                lookupDottedPath(data, value.substr(1)) : value);
           });
           return ids;
         }
@@ -663,11 +664,13 @@ angular.module('ngResource', ['ng']).
             }
 
             if (hasBody) httpConfig.data = data;
-            route.setUrlParams(httpConfig,
-              extend({}, extractParams(data, action.params || {}), params),
-              action.url);
-
-            var promise = $http(httpConfig).then(function(response) {
+            var promise = $q.all(extractParams(data, action.params || {})).then(function(ids) {
+              route.setUrlParams(httpConfig,
+                extend({}, ids, params),
+                action.url);
+              return $http(httpConfig);
+            });
+            promise = promise.then(function(response) {
               var data = response.data;
 
               if (data) {
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -102,6 +102,7 @@ function shallowClearAndCopy(src, dst) {
  * @param {Object=} paramDefaults Default values for `url` parameters. These can be overridden in
  *   `actions` methods. If a parameter value is a function, it will be executed every time
  *   when a param value needs to be obtained for a request (unless the param was overridden).
+ *   If a parameter value is a promise, the request will execute when the promise is resolved.
  *
  *   Each key value in the parameter object is first bound to url template if present and then any
  *   excess keys are appended to the url search query after the `?`.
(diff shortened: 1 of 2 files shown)


kevmo314:
I'm a Googler, I added my @google.com address to my github account.

Narretz:
I don't feel comfortable adding another level of async to an already async API. What's the general use case for this feature? How often do you have dynamic param defaults that must be retrieved async?

kevmo314:
I have a case where I need a user id to be sent in a `$resource`, however this user id must be fetched from another API request and it can't be embedded in the page as the page is static. Since the `$resource` cannot wait for the credentials to arrive, there's a race condition as a result if the credentials arrive before the resource is needed. Both calls are made before the first user interaction.

This seems like a rather common scenario, e.g. loading the posts for a given user, and from what I could tell the common solution is to write a service wrapper around the resource to verify it's authenticated first since there's no way to make the authentication call synchronous. However, this seems like it's literally just reimplementing everything `$resource` provides. Do you have any other suggestions for how to mitigate the race condition?

gkalpak:
There are some failing tests btw...

src/ngResource/resource.js
```diff
@@ -663,11 +665,13 @@ angular.module('ngResource', ['ng']).
...
 
             if (hasBody) httpConfig.data = data;
-            route.setUrlParams(httpConfig,
-              extend({}, extractParams(data, action.params || {}), params),
-              action.url);
-
-            var promise = $http(httpConfig).then(function(response) {
+            var promise = $q.all(extractParams(data, action.params || {})).then(function(ids) {
+              route.setUrlParams(httpConfig,
+                extend({}, ids, params),
```
gkalpak: I don't think the first `{}` argument is needed any more.
(In fact, I don't think it was needed before anyway.)

test/ngResource/resourceSpec.js
```diff
@@ -623,34 +623,39 @@ describe("basic usage", function() {
 
 
   it('should support dynamic default parameters (global)', function() {
-    var currentGroup = 'students',
-        Person = $resource('/Person/:group/:id', { group: function() { return currentGroup; }});
+    inject(function($q) {
```
gkalpak: Nitpick: Wrap the test funnction with `inject()` and avoid the extra indentation. E.g.:

```js
it('...', inject(function($q) {...});
```

gkalpak:
One way would be to indeed have a wrapper service that retrieves the current user first and then calls the appropriate `$resource` method, but that would be very inconvenient.

I don't think this is a common situation though. I, at least, have never run into this issue, because typically, retrieving the current user is a prerequisite for loading a route (e.g. in a route's `resolve` dependincies) or kicking off some initialization.

kevmo314:
Before I go ahead and fix those tests, is this something the angular team would be open to having or is it too much of an edge case?

gkalpak:
I do think it's a little bit of an edge case but not too much :smile:
It is already possible to solve this in different ways, but having async default params could be cleaner in specific scenarios.

Since it doesn't change the current behavior (for users not relying on promises), I'd be fine adding this.

Narretz:
I personally wouldn't introduce it since it's an uncommon case and we'd have to support it in the future. @petebacondarwin wdyt?

kevmo314:
Quick ping for thoughts?

petebacondarwin:
I don't have a problem with sticking another promise in front of the request. It's async promises anyway right?

But here's an idea...

We have `$resource` response interceptors, right? What if we implemented request interceptors too?
Then you could write an interceptor that checked all the params for promises and waited for them all to complete before making the request.

This would seem to be a more general solution to the problem.

gkalpak:
I'm all for `$resource` request interceptors, but I am not sure they are a suitable solution for this kind of task - it might be a little low level.

Here's a PR for request interceptors: #14683 
Building the params into the URL happens before it gets passed to the interceptor (and I think it makes sense that way). WDYT ?

petebacondarwin:
OK, unless @Narretz can convince me otherwise I think this would be an acceptable feature

Narretz:
@petebacondarwin My objection is that it adds to the API surface even though the use case is relatively rare (something like this has never been requested before). So even though it's a simple change, it would benefit only very few users and we'd have to support it in the future.
Further, since we aren't sure where to to with interceptors in resource, we could possibly create an obstacle with this PR for the future.

PR #14597, opened 2016-05-12
docs(angular.noop): returns undefined

tdelmas:
Doc : Explicitly say that angular.noop returns undefined


Narretz:
I think if nothing is returned, you don't explicitly have to document that.

tdelmas:
Thank you for the clarification.

PR #14598, opened 2016-05-12
fix(filters): remove unreachable code

lipeno:
Remove "break" as it is unreachable because of the conditional statements.

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -242,7 +242,6 @@ function deepCompare(actual, expected, comparator, matchAgainstAnyProp, dontMatc
       } else {
         return comparator(actual, expected);
       }
-      break;
     case 'function':
       return false;
     default:


lipeno:
I signed it!

Narretz:
jshint doesn't like that. You'll have to add a comment that says that you deliberately left out the break statement because it's unreachable because of the else. Although I'm not sure that makes the code easier to understand.

gkalpak:
TBH, I always like to put `break` statements (even if unreachable), unless there is a `return` statement at the end of the `case` block. It is safer (e.g. against future refactorings) and as @Narretz said makes the code clearer (i.e. I don't have to go through three different `if-else` branches with loops etc to find out if it will fall through to the next `case` block).

And since `jsHint` also likes `break` statements, I think we win :stuck_out_tongue:

--
Is this causing you any issues (e.g. some IDE complaining about unreachable statements) or are you just trying to do the "right<super>&trade;</super>" thing ?

Narretz:
I'm going to close this issue because we haven't got any feedback. Feel free to reopen this issue if you can provide new feedback.

PR #14604, opened 2016-05-13
docs(ngMock): use master installation template

stevemao:
**What kind of change does this PR introduce?**

docs update

```
Google CDN e.g. {% code %}"//ajax.googleapis.com/ajax/libs/angularjs/X.Y.Z/"{% endcode %}
```

```
Google CDN e.g.
"//ajax.googleapis.com/ajax/libs/angularjs/X.Y.Z/angular-mock.js"
```

No

**Other information**:

The inline installation can't display the proper context. On the website, it shows something like

```
Google CDN e.g. {% code %}"//ajax.googleapis.com/ajax/libs/angularjs/X.Y.Z/"{% endcode %}
```

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2251,49 +2251,6 @@ angular.mock.$ComponentControllerProvider = ['$compileProvider', function($compi
  *
  *
  * <div doc-module-components="ngMock"></div>
- *
- * @installation
- *
- *  <p>First, download the file:</p>
- *  <ul>
-      <li>
-        <a href="https://developers.google.com/speed/libraries/devguide#angularjs">Google CDN</a> e.g.
-        {% code %}"//ajax.googleapis.com/ajax/libs/angularjs/X.Y.Z/"{% endcode %}
-      </li>
-      <li>
-        <a href="https://www.npmjs.com/">NPM</a> e.g.
-        {% code %}npm install {$ doc.packageName $}@X.Y.Z{% endcode %}
-      </li>
-      <li>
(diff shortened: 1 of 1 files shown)


Narretz:
Hm, that's bad. But we cannot use the default installation notes because they will make it look like you should include ngMocks into your application code wheras it should be included in your test runner script.

stevemao:
Oh well. The markup is wrong. I'm not familiar with the syntax. How do I fix it? 😄

Narretz:
We can either use static version numbers or change how the template is
handled.

gkalpak:
Basically, we want to let nunjucks parse the `docs.installation` [here][1].
\cc @petebacondarwin :grin:

[1]: https://github.com/angular/angular.js/blob/3b5751dce8d6c699dc76e47cfa544c30b38b9771/docs/config/templates/api/module.template.html#L10

gkalpak:
@petebacondarwin, this is still not 100% fixed. Things like `{$ doc.packageName $}`, `{$ doc.packageFile $}` are still literally appearing in the output.

petebacondarwin:
I think that 921f80e1eabf04ecd4dcfda85e7ac801a8f17728 fixes these outstanding issues.

stevemao:
So we don't need context to feed the templates but just hardcode the actual file name and versions?

gkalpak:
Generally, he installation instructions are automatically generated based on a template (and fed with dynamical values, e.g. package name, file name etc). There are a couple of cases (including `ngMock`) where we need specific, non-generic installation instruction, so we provide them directly. In such cases, we can (and should) pass the exact content, including package name and file name.

PR #14610, opened 2016-05-13
fix($compile): correctly denormalize templates that do not use standard interpolation symbols

petebacondarwin:

FIX

See #6493

Fixed

No

**Other information**:

*\* STILL NEED TO ADD DOCS **

In order to support third party modules that do not use the same interpolation
symbols as the main app, we implemented denormalization (see dfe99836c).

This required that 3rd party modules always used the standad `{{` and `}}`
interpolation symbols, so that we could correctly denormalise them to the
current app's symbols.

The problem with this became apparent when an app template using new symbols
inadvertently contained one of the standard interpolation symbols.

E.g. `<div data-context='{"context":{"id":3,"type":"page"}}">`

The double closing curly braces were being incorrectly denormalised.

This commit fixes this by allowing the compiler to be configured to know
what symbols are being used in templates from a given module.

Now modules can specify that what interpolation symbols they use and the
compiler will denormalize appropriately.

Closes #6493
Closes #6453

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -3155,23 +3155,52 @@ describe('$compile', function() {
       });
     });
 
-    it('should allow modules to specify what interpolation symbol is used in templates', function() {
+    it('should denormalise interpolation symbols in templates correctly, if different to the current symbols', function() {
       angular.module('symbol-test', [])
         .directive('myDirective', function() {
           return {
-            template: '<span foo=\'{"ctx":{"id":3}}\'></span>'
+            template: '<span foo=\'{"ctx":{"id":3}}\'>[[1 + 2]]</span>'
           };
+        })
+        .config(function($compileProvider) {
+          $compileProvider.moduleSymbols('symbol-test', '[[', ']]');
         });
 
-      module('symbol-test', function($interpolateProvider, $compileProvider) {
-        $interpolateProvider.startSymbol('##');
-        $interpolateProvider.endSymbol(']]');
-        $compileProvider.moduleSymbols('symbol-test', '##', ']]');
+      module('symbol-test');
+
+      inject(function($compile) {
+        element = $compile('<div><div my-directive></div></div>')($rootScope);
+        expect(element.children('div').children('span').attr('foo')).toBe('{"ctx":{"id":3\\}\\}');
+        expect(element.text()).toEqual('{{1 + 2}}');
+        $rootScope.$apply();
+        expect(element.text()).toEqual('3');
+      });
+    });
+
+
+    it('should escape the current interpolation symbols in templates, before denormalising, if the template module symbols are different', function() {
+      angular.module('symbol-test', [])
+        .directive('myDirective', function() {
+          return {
+            template: '<span>{[1 + 2]}[[3 + 4]]</span>'
+          };
+        })
+        .config(function($compileProvider) {
+          // Specify that the templates in this module use `[[` and `]]` interpolation symbols
+          $compileProvider.moduleSymbols('symbol-test', '[[', ']]');
+        });
+
+      module('symbol-test', function($interpolateProvider) {
+        // Specify that this app uses `{[` and `]}` as interpolation symbols
+        $interpolateProvider.startSymbol('{[');
+        $interpolateProvider.endSymbol(']}');
       });
 
       inject(function($compile) {
         element = $compile('<div><div my-directive></div></div>')($rootScope);
-        expect(element.children('div').children('span').attr('foo')).toBe('{"ctx":{"id":3}}');
+        expect(element.text()).toEqual('\\{\\[1 + 2\\]\\}{[3 + 4]}');
+        $rootScope.$apply();
+        expect(element.text()).toEqual('{[1 + 2]}7');
       });
     });
 
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -3155,6 +3155,26 @@ describe('$compile', function() {
       });
     });
 
+    it('should allow modules to specify what interpolation symbol is used in templates', function() {
+      angular.module('symbol-test', [])
+        .directive('myDirective', function() {
+          return {
+            template: '<span foo=\'{"ctx":{"id":3}}\'></span>'
+          };
+        });
+
+      module('symbol-test', function($interpolateProvider, $compileProvider) {
+        $interpolateProvider.startSymbol('##');
+        $interpolateProvider.endSymbol(']]');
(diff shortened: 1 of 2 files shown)


test/ng/compileSpec.js
```diff
@@ -3155,6 +3155,26 @@ describe('$compile', function() {
...
+    it('should allow modules to specify what interpolation symbol is used in templates', function() {
+      angular.module('symbol-test', [])
+        .directive('myDirective', function() {
+          return {
+            template: '<span foo=\'{"ctx":{"id":3}}\'></span>'
+          };
+        });
+
+      module('symbol-test', function($interpolateProvider, $compileProvider) {
+        $interpolateProvider.startSymbol('##');
```
lgalfaso: these two lines should not be there, as with them, the test makes no sense.

lgalfaso:
I see this as a feature. @petebacondarwin would it be possible to know why you think it is a fix?

petebacondarwin:
To clarify this a bit more. If I change my interpolation symbols (in my app) to `[[` and `]]` then I would expect that a component with this template

```
<div>{a:{b:42}}</div>
```

should output

```
<div>{a:{b:42}}</div>
```

But it doesn’t. See http://plnkr.co/edit/s9KVOZ9srU0JfkQYPrCf?p=preview

lgalfaso:
Ok, but then I think that should be true in the test even without the
configuration changes to $interpolateProvider

petebacondarwin:
Without the changes to the interpolate provider then no denormalisation occurs so the bug doesn't occur in the same way. If we have not changed the symbols when we accept that we should escape `}}` ourselves if they appear in the template as content.

lgalfaso:
> Without the changes to the interpolate provider then no denormalisation occurs so the bug doesn't occur in the same way. If we have not changed the symbols when we accept that we should escape }} ourselves if they appear in the template as content.

Agree, but then `$compileProvider.moduleSymbols('symbol-test', '##', ']]');` does **not** mean "The interpolation symbols for all templates in this module are `##` and `]]`" that is what most people would expect.

petebacondarwin:
Ahah! @lgalfaso so what you are saying is that we have another issue:
if the symbols in a template do not match those in the application, then we need to escape any instance of the application symbols, in the templates, too.

lgalfaso:
implementation details a side, yes, the issue is the one stated.

petebacondarwin:
@lgalfaso here are better tests:

```js
    it('should denormalise interpolation symbols in templates correctly, if different to the current symbols', function() {
      angular.module('symbol-test', [])
        .directive('myDirective', function() {
          return {
            template: '<span foo=\'{"ctx":{"id":3}}\'>[[1 + 2]]</span>'
          };
        });

      module('symbol-test', function($compileProvider) {
        $compileProvider.moduleSymbols('symbol-test', '[[', ']]');
      });

      inject(function($compile) {
        element = $compile('<div><div my-directive></div></div>')($rootScope);
        expect(element.children('div').children('span').attr('foo')).toBe('{"ctx":{"id":3}}');
        expect(element.text()).toEqual('{{1 + 2}}');
        $rootScope.$apply();
        expect(element.text()).toEqual('3');
      });
    });


    it('should escape the current interpolation symbols in templates, before denormalising, if the template module symbols are different', function() {
      angular.module('symbol-test', [])
        .directive('myDirective', function() {
          return {
            template: '<span>{[1 + 2]}[[3 + 4]]</span>'
          };
        });

      module('symbol-test', function($interpolateProvider, $compileProvider) {
        $interpolateProvider.startSymbol('{[');
        $interpolateProvider.endSymbol(']}');
        $compileProvider.moduleSymbols('symbol-test', '[[', ']]');
      });

      inject(function($compile) {
        element = $compile('<div><div my-directive></div></div>')($rootScope);
        expect(element.text()).toEqual('\\{\\[1 + 2\\]\\}{[3 + 4]}');
        $rootScope.$apply();
        expect(element.text()).toEqual('3');
      });
    });
```

The second one fails at the moment.

petebacondarwin:
I added these tests and updated the code to make them pass.

lgalfaso:
@petebacondarwin I think that the solution will have to be somehow more complex. Eg

```js
it('this works', function() {
  angular.module('symbol-test', [])
    .directive('myDirective', function() {
      return {
        template: '<span foo=\'{"ctx":{"id":3}}\'>Hi</span>'
      };
    });

  module('symbol-test');
  inject(function($compile) {
    element = $compile('<div><div my-directive></div></div>')($rootScope);
    $rootScope.$apply();
    expect(element.find('span').attr('foo')).toEqual('{"ctx":{"id":3}}');
  });
});

it('so this should work too', function() {
  angular.module('symbol-test', [])
    .directive('myDirective', function() {
      return {
        template: '<span foo=\'{"ctx":{"id":3}}\'>Hi</span>'
      };
    })
    .config(function($compileProvider) {
      $compileProvider.moduleSymbols('symbol-test', '[[', ']]');
    });

  module('symbol-test');
  inject(function($compile) {
    element = $compile('<div><div my-directive></div></div>')($rootScope);
    $rootScope.$apply();
    expect(element.find('span').attr('foo')).toEqual('{"ctx":{"id":3}}');
  });
});
```

gkalpak:
I just realized there is another issue with template denormalization: We are not taking `[ng-non-bindable]` into account - we still replace the interpolation symbols :disappointed: 
(Or is this a feature ?)

test/ng/compileSpec.js
```diff
@@ -3155,6 +3155,55 @@ describe('$compile', function() {
...
+        });
+
+      module('symbol-test');
+
+      inject(function($compile) {
+        element = $compile('<div><div my-directive></div></div>')($rootScope);
+        expect(element.children('div').children('span').attr('foo')).toBe('{"ctx":{"id":3\\}\\}');
+        expect(element.text()).toEqual('{{1 + 2}}');
+        $rootScope.$apply();
+        expect(element.text()).toEqual('3');
```
gkalpak: What about `.attr('foo')` ? I believe it is still escaped and I think this is not what we want...

petebacondarwin: That is what @lgalfaso was referring to in his test: https://github.com/angular/angular.js/pull/14610#issuecomment-219272549

:-(

petebacondarwin:
Hmm looks like this could be more painful than it is worth...

lgalfaso:
> Hmm looks like this could be more painful than it is worth...

yep

petebacondarwin:
Perhaps another one for the "known issues" pile?

gkalpak:
> another one

More like "a couple more" :stuck_out_tongue:

PR #14625, opened 2016-05-17
fix(aria/ngModel): do not overwrite the default `$isEmpty()` method for checkboxes

### LINKED ISSUE #14621: ngAria ngModel directive breaks when used with terminal priority 1<p<201 directive

Bug

When using the ngAria module with a terminal, priority 1<p<201 directive on the same element as an ngModel directive an error occurs breaking the compile step.

1. Include ngAria
2. Define directive with `terminal: true` and a `priority` 1<p<201
3. Apply defined directive and ngModel to the same element
4. Run application, error occurs

See http://plnkr.co/edit/amf20s9OovC3vYzjOKhU

No error should occur.

1.4.8 does not have the issue. 1.5.5 and the snapshot have it.

Issue appears to have introduced in the commit a460ce4665b187db7f8dc7df6a0c3caea999c1e2 with a change to src/ngAria/aria.js

gkalpak:
Bug fix(es)

`ngAria` overwrites `NgModelController.$isEmpty()` for custom checkboxes (only considering `false` as empty), despite the fact that custom checkboxes don't necessarily have custom parsers setting falsy values to `false`. (E.g. the checkbox is considered "checked" even if value is `undefined` (because `undefined !== false`.)
Also, `ngModel` breaks when manually compiling from a terminal directive with lower priorty (see #14621).

Yes!

**Other information**:
Previously, `ngAria` would overwrite the default `ngModelController.$isEmpty()` method for custom
`checkbox`-shaped controls (e.g. `role="checkbox"` or `role="menuitemcheckbox"`), using the same
implementation as `input[checkbox]` (i.e. `value === false`). While this makes sense for
`input[checkbox]` which also defines a custom parser, it doesn't make sense for a custom `checkbox`
out-of-the-box. For example, an unintialized value (`undefined`) would make the checkbox appear as
"checked".

If the user wants to provide a custom parser (e.g. setting falsy values to `false`), then they
should also provide a custom `$isEmpty()` method.

As a side effect, this commit solves issue #14621. (We could have solved it in different ways.)

Fixes #14621

BREAKING CHANGE:

Custom `checkbox`-shaped controls (e.g. checkboxes, menuitemcheckboxes), no longer have a custom
`$isEmpty()` method on their `NgModelController` that checks for `value === false`. Unless
overwritten, the default `$isEmpty()` method will be used, which treats `undefined`, `null`, `NaN`
and `''` as "empty".

**Note:** The `$isEmpty()` method is used to determine if the checkbox is checked ("not empty" means
          "checked") and thus it can indirectly affect other things, such as the control's validity
          with respect to the `required` validator (e.g. "empty" + "required" --> "invalid").

Before:

``` js
var template = '<my-checkbox role="checkbox" ng-model="value"></my-checkbox>';
var customCheckbox = $compile(template)(scope);
var ctrl = customCheckbox.controller('ngModel');

scope.$apply('value = false');
console.log(ctrl.$isEmpty());   //--> true

scope.$apply('value = true');
console.log(ctrl.$isEmpty());   //--> false

scope.$apply('value = undefined'/* or null or NaN or '' */);
console.log(ctrl.$isEmpty());   //--> false
```

After:

``` js
var template = '<my-checkbox role="checkbox" ng-model="value"></my-checkbox>';
var customCheckbox = $compile(template)(scope);
var ctrl = customCheckbox.controller('ngModel');

scope.$apply('value = false');
console.log(ctrl.$isEmpty());   //--> false

scope.$apply('value = true');
console.log(ctrl.$isEmpty());   //--> false

scope.$apply('value = undefined'/* or null or NaN or '' */);
console.log(ctrl.$isEmpty());   //--> true
```
## 

If you want to have a custom `$isEmpty()` method, you need to overwrite the default. For example:

``` js
.directive('myCheckbox', function myCheckboxDirective() {
  return {
    require: 'ngModel',
    link: function myCheckboxPostLink(scope, elem, attrs, ngModelCtrl) {
      ngModelCtrl.$isEmpty = function myCheckboxIsEmpty(value) {
        return !value;   // Any falsy value means "empty"

        // Or to restore the previous behavior:
        // return value === false;
      };
    }
  };
})
```

### DIFF
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -123,29 +123,31 @@ describe('$aria', function() {
       expect(element.attr('aria-checked')).toBe('false');
     });
 
-    it('should rely on the `$isEmpty()` method', function() {
-      compileElement('<div role="checkbox" ng-model="val"></div>');
-      var ctrl = element.controller('ngModel');
-      ctrl.$isEmpty = function(value) {
-        return value === 'not-checked';
-      };
+    it('should use `$isEmpty()` to determine if the checkbox is checked',
+      function() {
+        compileElement('<div role="checkbox" ng-model="val"></div>');
+        var ctrl = element.controller('ngModel');
+        ctrl.$isEmpty = function(value) {
+          return value === 'not-checked';
+        };
 
-      scope.$apply('val = true');
-      expect(ctrl.$modelValue).toBe(true);
-      expect(element.attr('aria-checked')).toBe('true');
+        scope.$apply('val = true');
+        expect(ctrl.$modelValue).toBe(true);
+        expect(element.attr('aria-checked')).toBe('true');
 
-      scope.$apply('val = false');
-      expect(ctrl.$modelValue).toBe(false);
-      expect(element.attr('aria-checked')).toBe('true');
+        scope.$apply('val = false');
+        expect(ctrl.$modelValue).toBe(false);
+        expect(element.attr('aria-checked')).toBe('true');
 
-      scope.$apply('val = "not-checked"');
-      expect(ctrl.$modelValue).toBe('not-checked');
-      expect(element.attr('aria-checked')).toBe('false');
+        scope.$apply('val = "not-checked"');
+        expect(ctrl.$modelValue).toBe('not-checked');
+        expect(element.attr('aria-checked')).toBe('false');
 
-      scope.$apply('val = "checked"');
-      expect(ctrl.$modelValue).toBe('checked');
-      expect(element.attr('aria-checked')).toBe('true');
-    });
+        scope.$apply('val = "checked"');
+        expect(ctrl.$modelValue).toBe('checked');
+        expect(element.attr('aria-checked')).toBe('true');
+      }
+    );
 
     it('should not handle native checkbox with ngChecked', function() {
       var element = $compile('<input type="checkbox" ng-checked="val">')(scope);
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -84,7 +84,6 @@ describe('$aria', function() {
     });
   });
 
-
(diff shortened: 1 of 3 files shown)


gkalpak:
I wonder if it would make more sense to provide a custom `$isEmpty` method to checkboxes, that interprets all falsy values as "empty"...

Narretz:
We could imo check emptiness based on the checked property, no? This would change the semantics (as the model value is currently validated), but I think it makes sense, for checkboxes specifically. I mean they are empty when unchecked almost by definition ...

gkalpak:
I was referring to custom checkboxes (I think :smiley:), which don't necessarily have a `checked` property. For `input[checkbox]` we are already doing the right thing (basing emptiness on the value of the `checked` property.

But, I am still not sure if it is a good idea to provide a custom `$isEmpty` function for custom checkboxes. Or at least we should document it properly, because it is happening under the hood and we don't mention it anywhere.

test/ngAria/ariaSpec.js
```diff
@@ -115,13 +114,37 @@ describe('$aria', function() {
...
+      compileElement('<div role="checkbox" ng-model="val"></div>');
+
+      scope.$apply('val = "checked"');
+      expect(element.attr('aria-checked')).toBe('true');
+
+      scope.$apply('val = null');
+      expect(element.attr('aria-checked')).toBe('false');
+    });
+
+    it('should rely on the `$isEmpty()` method', function() {
```
Narretz: Rely on it in which case? This could be more verbose.

Narretz:
You are right, the real checkbox uses checked. I'm fine with removing the custom isEmpty.

gkalpak:
@Narretz, is this a LGTM? :stuck_out_tongue: 
<sub>(I fixed the test description btw.)</sub>

Narretz:
Yes, LGTM :D

PR #14627, opened 2016-05-17
fix($compile): don't allow constructor to overwrite input bindings

petebacondarwin:

FIX

Changes to input bound properties inside the controller constructor are overwriting the
initial value of the input binding.

We now reassign these properties after calling the constructor.

No - previous releases of Angular 1.5.x over-wrote these properties both
after the constructor and after $onInit.

  ~~\- [ ] Docs have been added / updated (for bug fixes / features)~~

**Other information**:

Previously any input bindings would be reassigned after the constructor
and the `$onInit` hook had been called because of an invalid extra update
in the input binding watch handler.

In 304796471292f9805b9cf77e51aacc9cfbb09921 we removed this extra update,
which resulted in changes to the controller properties in either the
constructor or the `$onInit` hook to be kept.

It turns out that the fact that we can overwrite the input bindings in the
constructor is invalid. Really the bindings should have been assigned
only after the constructor is called but due to a historical feature that
enabled developers to access the bindings from within the constructor this
was possible.

To prevent this we now assign the bindings before calling the constructor
and then reassign them after the constructor but before the call to `$onInit`.

This is the behaviour that we want in the long term, where we will remove
the feature of pre-binding properties before calling the constructor any
way.

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -4827,6 +4827,31 @@ describe('$compile', function() {
             });
         });
 
+
+        it('should reassign bindings to the controller before $onInit, in case the constructor has overwritten them', function() {
+          angular.module('owComponentTest', [])
+            .component('owComponent', {
+              bindings: { input: '<' },
+              controller: function() {
+                component = this;
+                this.input = 'constructor-changed';
+                this.$onInit = function() {
+                  log.push('$onInit: input=' + this.input);
+                };
+              }
+            });
+
+          module('owComponentTest');
+          inject(function() {
+            $rootScope.name = 'outer';
+            compile('<ow-component input="name"></ow-component>');
+            expect(component.input).toEqual('outer');
+            expect(log).toEqual([
+              '$onInit: input=outer'
+            ]);
+          });
+        });
+
         it('should not update isolate again after $onInit if outer has not changed', function() {
           module('owComponentTest');
           inject(function() {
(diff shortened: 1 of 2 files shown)


test/ng/compileSpec.js
```diff
@@ -4827,6 +4827,31 @@ describe('$compile', function() {
...
+                  log.push('$onInit: input=' + this.input);
+                };
+              }
+            });
+
+          module('owComponentTest');
+          inject(function() {
+            $rootScope.name = 'outer';
+            compile('<ow-component input="name"></ow-component>');
+            expect(component.input).toEqual('outer');
```
gkalpak: TBH, I wouldn't expect this...

IgorMinar: why? the expectation is that when onInit runs we initialize the bindings overwriting anything that the developer set in the constructor.

IgorMinar:
lgtm

test/ng/compileSpec.js
gkalpak: Because currently the properties are supposed to be initialized _before_ instantiating the constructor (so that they are available in the constructor function. The user can use the values, but also overwrite them.

With this change, we are overwriting them again after the constructor function has been called.
This is a BC imo.

petebacondarwin:
I agree with @gkalpak - I think that this PR actually confuses things.
We should document clearly that changing bindings inside either the constructor or the `$onInit` hook will overwrite the initial value from the binding.

petebacondarwin:
Also this PR was failing on a bunch of tests, which to fix would require changes to tests, which also indicates BC

PR #14633, opened 2016-05-18
fix(ngAnimate): safe-guard against missing document

Narretz:
A different fix for the mocked document problem (in 1.4.x) that also fixes another possibly breaking access to the document.

One caveat: since we are now failing gracefully, and a missing document means no animations fire, tests that test animations and mock out document will not run animations, and this will be opaque to developers. However, I assume that most tests that mock document are actually very basic, as lots of other things should break when $document is mocked

Please check if the PR fulfills these requirements

```
The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
```

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -103,6 +103,9 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     var activeAnimationsLookup = new $$HashMap();
     var disabledElementsLookup = new $$HashMap();
     var animationsEnabled = null;
+    // $document might be mocked and won't include a real document.
+    // Providing an empty object will prevent property read errors
+    var rawDocument = $document[0] || {};
 
     function postDigestTaskFactory() {
       var postDigestCalled = false;
@@ -331,7 +334,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
 
       var isStructural = ['enter', 'move', 'leave'].indexOf(event) >= 0;
 
-      var documentHidden = animationsEnabled && $document[0].hidden;
+      var documentHidden = rawDocument.hidden;
 
       // this is a hard disable of all animations for the application or on
       // the element itself, therefore  there is no need to continue further
@@ -583,7 +586,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
      * d) the element is not a child of the $rootElement
      */
     function areAnimationsAllowed(element, parentElement, event) {
-      var bodyElement = jqLite($document[0].body);
+      var bodyElement = jqLite(rawDocument.body);
       var bodyElementDetected = isMatchingElement(element, bodyElement) || element[0].nodeName === 'HTML';
       var rootElementDetected = isMatchingElement(element, $rootElement);
       var parentAnimationDetected = false;
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -103,9 +103,9 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     var activeAnimationsLookup = new $$HashMap();
     var disabledElementsLookup = new $$HashMap();
     var animationsEnabled = null;
-    // $document might be mocked and won't include a real document.
-    // Providing an empty object will prevent property read errors
-    var rawDocument = $document[0] || {};
+    // $document might be mocked out in tests and won't include a real document.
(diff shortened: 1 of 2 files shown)


src/ngAnimate/animateQueue.js
```diff
@@ -331,7 +334,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
 
       var isStructural = ['enter', 'move', 'leave'].indexOf(event) >= 0;
 
-      var documentHidden = animationsEnabled && $document[0].hidden;
+      var documentHidden = rawDocument.hidden;
```
ThomasBurleson: Do you know need to include `animationsEnabled` ?

petebacondarwin: @ThomasBurleson - that was a hack by @IgorMinar  to get the tests passing in G3

petebacondarwin:
LGTM

src/ngAnimate/animateQueue.js
Narretz: You mean "not need"? animationsEnabled was added by Igor to fix a test in the g3 sync, but as I wrote in the commit message, this doesn't actually prevent the access to the document in all cases. If animations are enabled, $document[0] is still checked, and it might still be undefined

src/ngAnimate/animateQueue.js
```diff
@@ -583,7 +586,7 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
      * d) the element is not a child of the $rootElement
      */
     function areAnimationsAllowed(element, parentElement, event) {
-      var bodyElement = jqLite($document[0].body);
+      var bodyElement = jqLite(rawDocument.body);
```
gkalpak: IIRC there is one more access to `$document[0].body`. Should we change that too ?

Narretz: In this file?

src/ngAnimate/animateQueue.js
```diff
@@ -103,6 +103,9 @@ var $$AnimateQueueProvider = ['$animateProvider', function($animateProvider) {
     var activeAnimationsLookup = new $$HashMap();
     var disabledElementsLookup = new $$HashMap();
     var animationsEnabled = null;
+    // $document might be mocked and won't include a real document.
```
IgorMinar: "... mocked out in tests ..."

IgorMinar:
does it really make sense to run animations when $document is mocked out? what's the use-case for that?

I think that all of the animation code assumes that either the animations enabled and document is not mocked out (or is mocked out by a special animations-compatible mock used for unit testing of the animations code) or the animations are disabled.

With all these changes we are adding a forth scenario - animations are enabled but $document is mocked out by third-party that is not aware of the $document requirements by the animation code.

IgorMinar:
I think the assumption we should make is that if someone mocks out $document then they don't want animations to run.

src/ngAnimate/animateQueue.js
gkalpak: No, in [animateCssDriver.js][1].

[1]: https://github.com/Narretz/angular.js/blob/d9c1d330e35af76f1de7ecb59adbf317769b43e1/src/ngAnimate/animateCssDriver.js#L22

Narretz:
We can make this assumption but this might not necessarily be what someone else expects because it's not clear that animations need a document. But i like it since it makes the code simpler. 
The bigger issue is that once we start to account for a missing or insufficiently mocked out document there's a lot more to change.

jbedard:
Shouldn't tests just mock the document properly? Seems weird putting that mock in the actual code...

Narretz:
To clarify, even in the current form, without a document no animations are performed, as the areAnimationsAllowed check won't find a body element: https://github.com/angular/angular.js/pull/14633/files#diff-c15d87c15661cba4a721501ba4fda5b2R589

So I'd be okay with making this explicit if we find no document in the first place.

petebacondarwin:
I just realised that this actually checks the opposite... if the document is mocked out (and `rawDocument` is set to `{}`) then we **will indeed** try to run the animations.

Perhaps we should set `rawDocument` to `{ hidden: true }` if `$document[0]` is undefined?

Narretz:
Yes that would short-circuit the animations earlier and be  more explicit. Sounds good to me.

Narretz:
I've updated the PR to use an object {hidden: true} when no document is available. This will prevent the animator from attempting to run any animations

matsko:
LGTM

PR #14635, opened 2016-05-18
feat($compile): improve filter system of directive

jacopomaroli:
Feature

when you

```
$compile(element, undefined, undefined, 'ngController')($scope);
$scope.$apply();
```

only a single directive can be excluded

with this patch we could have more control on what directives we want to compile.
An example could be:

```
element.attr("ng-include", "template");
$compile(element, undefined, undefined, function(name){ return name == 'ngInclude'; })($scope);
$scope.$apply();
```

so we could compile just the directives we need. maybe just those attached runtime.
this way we can avoid re-attaching controllers or other ng-directives.

No. the code provided is backward compatible

**Other information**:

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2670,7 +2670,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
      */
     function addDirective(tDirectives, name, location, maxPriority, ignoreDirective, startAttrName,
                           endAttrName) {
-      if (name === ignoreDirective) return null;
+      if (typeof ignoreDirective === 'string' && name === ignoreDirective) return null;
+      if (typeof ignoreDirective === 'function' && !ignoreDirective.call(null, name)) return null;
+
       var match = null;
       if (hasDirectives.hasOwnProperty(name)) {
         for (var directive, directives = $injector.get(name + Suffix),


jacopomaroli:
I signed it!

src/ng/compile.js
```diff
@@ -2670,7 +2670,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
      */
     function addDirective(tDirectives, name, location, maxPriority, ignoreDirective, startAttrName,
                           endAttrName) {
-      if (name === ignoreDirective) return null;
+      if (typeof ignoreDirective === 'string' && name === ignoreDirective) return null;
+      if (typeof ignoreDirective === 'function' && !ignoreDirective.call(null, name)) return null;
```
gkalpak: Why not just `ignoreDirective(name)` ?

gkalpak: Also, I find it unintuitive that if a function named `ignoreDirective()` returns `true`, then the directive is not ignored.
I would expect the opposite.

gkalpak:
Since this is a private feature (not documented, not intended for general "consumption") and compiling specific directives isn't something we are currently supporting (which means it could lead to all sorts of hard-to-debug errors), I don't think we should add this.

I think most of the time you can achieve the same result with a high-priority terminal directive and the `maxPriority` argument.

src/ng/compile.js
jacopomaroli: I thought we might like a value for "this" but I wanted to discuss with the community about the best option before choosing.
As of now just `ignoreDirective(name)` is perfectly fine.

Also I realized exposing the full parameters list to the function might be a good idea, like:

`if (typeof ignoreDirective === 'function' && !ignoreDirective(tDirectives, name, location, maxPriority, ignoreDirective, startAttrName, endAttrName)) return null;`

Someone might need also other parameters like "location" in case he wants to process only the class directives or the attribute ones.

jacopomaroli:
It was mainly intended for already compiled elements which you add a new directive.

e.g. an element might not have a ng-include on load but you might like adding it runtime and execute just that directive.
Is it possible to achieve this with a high priority terminal directive?

src/ng/compile.js
jacopomaroli: About `ignoreDirective()` returning `true` It was easier for me thinking it like a positive filter returning true if the name was ok.
But since it's called ignoreDirective you're right, we should keep it consistent.

gkalpak:
> e.g. an element might not have a ng-include on load but you might like adding it runtime and execute just that directive.
Is it possible to achieve this with a high priority terminal directive?

If you want to be able to add the extra directive at an arbitrary time, then no, it isn't possible.
What is possible, is delaying the compilation until you have decided what to do (i.e. add or not add a directive, set an attribute value etc).

The problem with first compiling some directives and later re-compiling for a specific additional directive  only is that some features may not work as expected (e.g. what if the new directive has `replace: true`, or what if it requires an isolate scope while the element already has one, or what if it specifies a template while the element already has one, or what if it requires a sibling controller etc).

petebacondarwin:
I agree with @gkalpak that this is a feature that could cause a huge amount of problems given the complexity of what directives can actually do to the DOM and the scope tree.

jacopomaroli:
isn't "replace" in directive flagged as deprecated and to remove? [https://code.angularjs.org/1.3.0-beta.11/docs/api/ng/service/$compile#-replace-deprecated-will-be-removed-in-next-major-release-](url)

I don't know angular's internals very well but I think it should all be already managed. If the new directive specifies a template and one's existing it will be replaced by the one specified in the compiled directive, if it requires an isolate scope the current one will be overridden (we could use a $destroy on the current element's scope) and only the controller not ignored could be required.

@petebacondarwin: if we have a well defined number of test case I'll try to provide some.

As of today, If I understand correctly, if we want to keep the scope tree consistent the only option is to `scope.$destroy();` and recompile the element with the new directives added because of scope's dependency?

thanks

Narretz:
Almost the same as https://github.com/angular/angular.js/pull/12682, which also has a test. Still not a big fan, though

petebacondarwin:
> As of today, If I understand correctly, if we want to keep the scope tree consistent the only option is to scope.$destroy(); and recompile the element with the new directives added because of scope's dependency?

That is the correct approach.

petebacondarwin:
@jacopomaroli thanks for sending in this PR. We feel that it is a rare use case that is not even one that we would recommend for Angular 1 so we do not intend to implement this. There are a few workarounds that have been discussed in the comments.

PR #14638, opened 2016-05-19
Add sudo before apt-get

sssemil:
docs

### DIFF
--- a/docs/content/tutorial/index.ngdoc
+++ b/docs/content/tutorial/index.ngdoc
@@ -110,7 +110,7 @@ suggested solution is to also install the `nodejs-legacy` apt package, which ren
 `nodejs`.
 
 ```
-apt-get install nodejs-legacy npm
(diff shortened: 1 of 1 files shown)


sssemil:
I signed it!

Narretz:
Since I'm not using Linux / MacOs I can't say if this is "correct". I do believe that sudo should only be used when appropriate, and not generally. I think this is then out of scope for the docs, since we cannot know if someone needs sudo or not.

gkalpak:
I'm (mainly) on Windows too, but I think `apt-get install` does indeed need `sudo` (except maybe for very specific, uncommon setups).

That said, we are about to merge #14416, which will overwrite this change anyway.
@sssemil, if you are confident `sudo` is necessary, please check back in a few days and submit a new PR against the updated tutorial.

Thx !

sssemil:
Cool, I'm sure that apt-get requires sudo, so I'll reopen it later. Cheers.

PR #14640, opened 2016-05-19
fix(ngMessagesInclude): do not compile template if scope is destroyed

### LINKED ISSUE #12695: ngMessagesInclude processed after parent ngMessages is removed from DOM
As the title says, ngMessagesInclude processed after parent ngMessages is removed from DOM. It fails because angular can't find parent controller for linking inner messages. It doesn't happen when `ngMessage`'s are inlined inside the `ngMessages` element, or while parent element is still present on DOM.

Have an ng-messages like this:

```html
<div ng-if="myForm.$submitted" ng-messages="myForm.myName.$error">
  <div ng-messages-include="messages.html"></div>
</div>
```
Submission is triggered by:

```js
this.change = function(myForm) {
  var self = this;
  myForm.$setSubmitted();    // event A
  $scope.$applyAsync(function () {
    self.currentStep = 'step2.html';     // event B
  });
};
```

At event (A), `messages.html` is downloaded. Because it uses `$templateRequest`, linking is stalled until server returns. Meanwhile, event (B) proceeds with the removal of `ngMessages` and its children from DOM. Later, ngMessagesInclude content arrives and angular will compile it. But at that point there's no way to retrieve `ngMessages` controller anymore.

Plnkr: http://plnkr.co/edit/0GOOLERvfj7n9vDVEyZp?p=preview

Most of the time, it will happen together with an `ngIf` condition.

Tested against version 1.4.4.

I've found a way to avoid that:

```js
ng-if="myForm.$submitted && myForm.myName.$invalid" 
```

To solve that at component side, the `require: '^^ngMessages'` could be made optional and check that condition during linking.

jpekkala:
Bug fix

https://github.com/angular/angular.js/issues/12695

**Other information**:

Messages imported with ngMessagesInclude are loaded asynchronously and they can arrive after the ngMessages element has already been removed from DOM. Previously such messages were still compiled and that caused a $compile:ctreq error. Now they are silently ignored if ngMessagesInclude's scope has already been destroyed.

Closes #12695

### DIFF
--- a/test/ngMessages/messagesSpec.js
+++ b/test/ngMessages/messagesSpec.js
@@ -842,6 +842,33 @@ describe('ngMessages', function() {
       })
     );
 
+    it('should not compile template if scope is destroyed', function() {
+      module(function($provide) {
+        $provide.decorator('$compile', ['$delegate', function($delegate) {
+          var result = jasmine.createSpy('$compile').and.callFake($delegate);
+          result.$$createComment = $delegate.$$createComment;
+          return result;
+        }]);
+      });
+      inject(function($rootScope, $httpBackend, $compile) {
+        $httpBackend.expectGET('messages.html').respond('<div ng-message="a">A</div>');
+        $rootScope.show = true;
+        var html =
+            '<div ng-if="show">' +
+              '<div ng-messages="items">' +
+                '<div ng-messages-include="messages.html"></div>' +
+              '</div>' +
+            '</div>';
+
+        element = $compile(html)($rootScope);
+        $rootScope.$digest();
+        $rootScope.show = false;
+        $rootScope.$digest();
+        $compile.calls.reset();
+        $httpBackend.flush();
+        expect($compile).not.toHaveBeenCalled();
+      });
+    });
   });
 
   describe('when multiple', function() {
--- a/test/ngMessages/messagesSpec.js
+++ b/test/ngMessages/messagesSpec.js
@@ -842,14 +842,7 @@ describe('ngMessages', function() {
       })
     );
 
-    it('should not compile template if scope is destroyed', function() {
-      module(function($provide) {
-        $provide.decorator('$compile', ['$delegate', function($delegate) {
-          var result = jasmine.createSpy('$compile').and.callFake($delegate);
-          result.$$createComment = $delegate.$$createComment;
-          return result;
-        }]);
-      });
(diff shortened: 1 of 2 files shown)


test/ngMessages/messagesSpec.js
```diff
@@ -842,6 +842,33 @@ describe('ngMessages', function() {
...
+              '</div>' +
+            '</div>';
+
+        element = $compile(html)($rootScope);
+        $rootScope.$digest();
+        $rootScope.show = false;
+        $rootScope.$digest();
+        $compile.calls.reset();
+        $httpBackend.flush();
+        expect($compile).not.toHaveBeenCalled();
```
Narretz: Instead of spying on compile, I'd rather see an expectation for not throwing, or a check that validates the DOM structure is as we expect it. The call to compile is a bit to implementation specific for me

test/ngMessages/messagesSpec.js
```diff
@@ -842,6 +842,33 @@ describe('ngMessages', function() {
       })
     );
 
+    it('should not compile template if scope is destroyed', function() {
+      module(function($provide) {
+        $provide.decorator('$compile', ['$delegate', function($delegate) {
+          var result = jasmine.createSpy('$compile').and.callFake($delegate);
```
gkalpak: `return spyOn($delegate).and.callThrough();` is more straight-forward.

jpekkala:
Thanks for the reviews @Narretz and @gkalpak. An expectation for not throwing was my first idea too but after seeing the analogous test in ngInclude, I decided to do it the same way as they had.

This is how I would write the test with the expectation for not throwing:
```
it('should not throw if scope has been destroyed when template request is ready',
  inject(function($rootScope, $httpBackend, $compile) {
    $httpBackend.expectGET('messages.html').respond('<div ng-message="a">A</div>');
    $rootScope.show = true;
    var html =
        '<div ng-if="show">' +
          '<div ng-messages="items">' +
            '<div ng-messages-include="messages.html"></div>' +
          '</div>' +
        '</div>';

    element = $compile(html)($rootScope);
    $rootScope.$digest();
    $rootScope.show = false;
    $rootScope.$digest();
    expect(function() {
      $httpBackend.flush();
    }).not.toThrow();
}));
```
Should I update the pull request?

Narretz:
In that case, can you just add the not.toThroe expectation?

PR #14656, opened 2016-05-24
feat($compile): backport $doCheck

zbjornson:
Feature

If implemented, $doCheck will not be invoked

$doCheck will be invoked on every digest cycle.

No

**Other information**:

This mirrors the current actual behavior of Angular 2:
- During the initial compile, the sequence is `$onChanges`, `$onInit`, `$doCheck`.
- During subsequent digests, the sequence is `$doCheck`, `$onChanges`, `$doCheck`.
  (See the [live peekaboo demo from the docs](https://angular.io/resources/live-examples/lifecycle-hooks/ts/plnkr.html) to see the sequence.)
- Per my comments [here](https://github.com/angular/angular/issues/6810#issuecomment-220889605), `ngDoCheck` in Angular 2 just seems to be a hook into the digest cycle that supplements, but does not replace/override, the default change detector. The Angular 2 [docs](https://angular.io/docs/ts/latest/api/core/DoCheck-interface.html) are incorrect ([issue](https://github.com/angular/angular/issues/7307)): `$onChanges` is indeed called, and the code in `doCheck` does not override the default CD.

/cc @gkalpak

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -300,6 +300,12 @@
  *   `changesObj` is a hash whose keys are the names of the bound properties that have changed, and the values are an
  *   object of the form `{ currentValue, previousValue, isFirstChange() }`. Use this hook to trigger updates within a
  *   component such as cloning the bound value to prevent accidental mutation of the outer value.
+ * * `$doCheck()` - Called on each turn of the digest cycle. Provides an opportunity to detect and act on
+ *   changes. Any actions that you wish to take in response to the changes that you detect must be
+ *   invoked from this hook; implementing this has no effect on when `$onChanges` is called. For example, this hook
+ *   could be useful if you wish to perform a deep equality check, or to check a Date object, changes to which would not
+ *   be detected by Angular's change detector and thus not trigger `$onChanges`. This hook is invoked with no arguments;
+ *   if detecting changes, you must store the previous value(s) for comparison to the current values.
  * * `$onDestroy()` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
  *   external resources, watches and event handlers. Note that components have their `$onDestroy()` hooks called in
  *   the same order as the `$scope.$broadcast` events are triggered, which is top down. This means that parent
@@ -2483,6 +2489,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           if (isFunction(controllerInstance.$onInit)) {
             controllerInstance.$onInit();
           }
+          if (isFunction(controllerInstance.$doCheck)) {
+            controllerInstance.$doCheck();
+          }
           if (isFunction(controllerInstance.$onDestroy)) {
             controllerScope.$on('$destroy', function callOnDestroyHook() {
               controllerInstance.$onDestroy();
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I think we need to get a better understanding of what `DoCheck` means in Angular 2 before we can make a decision about this.

zbjornson:
Yeah, I'd love to have one of the devs who worked on this part confirm or correct what I've found (@vicb?).

Commented in https://github.com/angular/angular/issues/7307 regarding the effect of ChangeDetectionStrategy. Especially because CDS doesn't exist in Angular 1, this still appears to be a good reflection of Angular 2.

dcherman:
From what I understand of `doCheck` in ng2, it's intended to completely replace Change Detection which means you must then implement it yourself within the `ngDoCheck` callback.

The Angular1 equivalent would pretty much be if `$doCheck` was implemented, then you don't run any of the watchers on the current scope and allow `$doCheck` to determine what's done, but that's a much larger change than is being proposed here.

zbjornson:
@dcherman is that based on internal design knowledge? That behavior is what the docs suggest, but they're confirmed to be at least partly incorrect (#7307) and the angular 2 directives that implement ngDoCheck (e.g. ngFor) are built around ngDoCheck just being a hook into the digest cycle (i.e. nothing in angular itself consumes the result of ngDoCheck -- any actions that need to be taken as a result of detected changes need to be taken in ngDoCheck itself). Possibility that ngDoCheck is not finished though...?

dcherman:
@zbjornson Not at all; that point is just from both tinkering with ng2 / reading the docs and points that I picked up during ngConf earlier this month.  It's entirely possible that the description is no longer accurate.

zbjornson:
@petebacondarwin My [PR to correct the ng2 `ngDoCheck` docs](https://github.com/angular/angular/pull/8807) was accepted yesterday by someone who seems knowledgeable about the lifecycle hooks. The docs (when released) now describe the actual ng2 behavior, which is what I mirrored in this backport. Does that help with the decision?

mazzur:
@petebacondarwin @zbjornson Is there any progress on this issue? This hook seems to be quite vital in attempts to get rid of $scope in components when migrating to angular 1.5 providing the ability to deep watch object properties supplementing capabilities of the $onChanges hook.

src/ng/compile.js
```diff
@@ -300,6 +300,9 @@
  *   `changesObj` is a hash whose keys are the names of the bound properties that have changed, and the values are an
  *   object of the form `{ currentValue, previousValue, isFirstChange() }`. Use this hook to trigger updates within a
  *   component such as cloning the bound value to prevent accidental mutation of the outer value.
+ * * `$doCheck()` - Called on each digest cycle. Provides an opportunity to detect and act on
```
petebacondarwin: Perhaps to make it even more clear?

```
Called on each turn of the digest loop
```

src/ng/compile.js
```diff
@@ -300,6 +300,9 @@
...
+ *   changes. Any actions that you wish to take in response to the changes that you detect must be
+ *   invoked from this hook; implementing this has no effect on when `$onChanges` is called.
```
petebacondarwin: An example of a use case would be helpful. Perhaps a "deep" equality check or a check of a Date object that might be changed in internally?

petebacondarwin: It should probably also be made clear that the implementer of this method would be responsible for tracking the "previous" values for comparison.

zbjornson:
@petebacondarwin do those doc revisions read better now? Thanks for the review.

petebacondarwin:
There were a couple of issues this the way this is implemented.
I have fixed this and added a runnable example here: https://github.com/angular/angular.js/pull/14811
Please take a look

PR #3332, opened 2013-07-24
Route continues to be processed even if providing a redirectTo

jardilio:
If a route defines a redirectTo value, the current route should stop processing the route and instead wait to process the route of the redirect. 

Currently, what is happening is that Angular detects a redirectTo value and updates $location, but then continues processing the route that was intended to be redirected. Templates are loaded, resolves are processed, ng-view then updates the view with a new template and instantiates the controller all for a route that should have been redirected.

A common use case for assigning a function to the redirectTo is to validation authentication, permission check, or some other logic to determine if the route should continue or redirect else where. In the end, this happens, but in between, unexpected results may occur by updating the view and instantiating controllers for logic that wasn't expected to be executed.

http://plnkr.co/edit/8QlA0ouuePH3p35Ntmjy?p=preview

PR #14658, opened 2016-05-24
Jardilio route redirect fix

petebacondarwin:
See #3332

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -608,15 +608,8 @@ function $RouteProvider() {
         $q.when(nextRoute).
           then(function() {
             if (nextRoute) {
-              var locals = angular.extend({}, nextRoute.resolve),
-                  template, templateUrl;
-
-              angular.forEach(locals, function(value, key) {
-                locals[key] = angular.isString(value) ?
-                    $injector.get(value) : $injector.invoke(value, null, null, key);
-              });
-
-              template = getTemplateFor(nextRoute);
+              var locals = resolveLocalsFor(nextRoute);
+              var template = getTemplateFor(nextRoute);
               if (angular.isDefined(template)) {
                 locals['$template'] = template;
               }
@@ -638,6 +631,16 @@ function $RouteProvider() {
             }
           });
       }
+    }
+
+    function resolveLocalsFor(route) {
+      var locals = angular.extend({}, route.resolve);
+      angular.forEach(locals, function(value, key) {
+        locals[key] = angular.isString(value) ?
+            $injector.get(value) :
+            $injector.invoke(value, null, null, key);
+      });
+      return locals;
     }
 
 
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -616,19 +616,7 @@ function $RouteProvider() {
                     $injector.get(value) : $injector.invoke(value, null, null, key);
               });
 
-              if (angular.isDefined(template = nextRoute.template)) {
-                if (angular.isFunction(template)) {
-                  template = template(nextRoute.params);
-                }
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I accidentally changed the authorship of the first commit. I will change back before merging...

test/ngRoute/routeSpec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-describe('$route', function() {
+fdescribe('$route', function() {
```
gkalpak: :scream:

src/ngRoute/route.js
```diff
@@ -588,46 +588,25 @@ function $RouteProvider() {
...
+                       .search(nextRoute.params)
                        .replace();
+              newUrl = $location.url();
             } else {
-              $location.url(nextRoute.redirectTo(nextRoute.pathParams, $location.path(), $location.search()))
-                       .replace();
+              newUrl = nextRoute.redirectTo(nextRoute.pathParams, $location.path(), $location.search());
+              $location.url(newUrl).replace();
+            }
+            if (isDefined(newUrl) && url !== newUrl) {
```
gkalpak: I thought `null` would behave the same as `undefined`.

gkalpak:
Where are the docs ? :stuck_out_tongue:

gkalpak:
I would also mention `resolve` amoong the things that don't get processed in the BC notice.

src/ngRoute/route.js
Narretz: Should all falsy values abort the redirect?

test/ngRoute/routeSpec.js
```diff
@@ -1061,6 +1061,55 @@ describe('$route', function() {
             .toEqual(['http://server/#!/bar/id3?extra=eId', true, null]);
       });
     });
+
+    it('should not instantiate controller or process template for a redirected route', function() {
```
Narretz: What about `resolve`?

petebacondarwin:
So at the moment null works differently so this pr does not change that. We should consider making null work like undefined in a follow up pr

petebacondarwin:
I'll add checks commit message and docs for resolve

src/ngRoute/route.js
petebacondarwin: currently `null` and `""` are not treated like `undefined` and so this PR doesn't change that (to minimize BC at this stage).

test/ngRoute/routeSpec.js
petebacondarwin: Oops

petebacondarwin:
@gkalpak and @Narretz PTAL!

Narretz:
LGTM

Narretz:
Don't forget to change back the authorship though!

src/ngRoute/route.js
```diff
@@ -136,6 +136,12 @@ function $RouteProvider() {
    *      The custom `redirectTo` function is expected to return a string which will be used
    *      to update `$location.path()` and `$location.search()`.
    *
+   *      Routes that specify `redirectTo` will not have their controllers, template functions
+   *      or resolves called, the `$location` will be changed to the redirect url and route
+   *      processing will stop. The exception to this is if the `redirectTo` is a function that
+   *      returns `undefined`. In this case the route transition occurs as though their was no
```
gkalpak: _their_  -->  _there_

petebacondarwin: Aghast!

gkalpak:
The commit message mentions that `null` cancels the redirection. Remove it when merging to avoid confusion.

LGTM

petebacondarwin:
Oops. Well I guess I have to fix the `null` path now then, eh?

PR #3711, opened 2013-08-23
fix: replace x.indexOf(y)==0 with x.lastIndexOf(y,0)===0

spamdaemon:
Replace the pattern x.indexOf(y)===0 with x.lastIndexOf(y,0) since that
does not have to check the entire string x for that substring.
This fix was triggered when found a test like this x.substr(0,2)=='$$'.

I ran all the unit tests and e2e tests, but did not add any new tests. I was thinking of creating an angular.startsWith function and can do so if desired, but ultimately the number of such prefix tests is rather small so that I decided against it.

I found this little nugget at stack-overflow:  http://stackoverflow.com/questions/646628/javascript-startswith

PR #14660, opened 2016-05-24
refactor(*): faster check for a string starting with a substring

petebacondarwin:
Thanks to @spamdaemon for the original PR (#3711) to make this improvement.

In naive tests on Chrome I got the following results:

```
             Matches   Misses
indexOf      33ms      1494ms
lastIndexOf  11ms      11ms
```

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -48,17 +48,20 @@ function parseAppUrl(relativeUrl, locationObj) {
   }
 }
 
+function startsWith(haystack, needle) {
+  return haystack.lastIndexOf(needle, 0) === 0;
+}
 
 /**
  *
- * @param {string} begin
- * @param {string} whole
- * @returns {string} returns text from whole after begin or undefined if it does not begin with
- *                   expected string.
+ * @param {string} base
+ * @param {string} url
+ * @returns {string} returns text from `url` after `base` or `undefined` if it does not begin with
+ *                   the expected string.
  */
-function beginsWith(begin, whole) {
-  if (whole.indexOf(begin) === 0) {
-    return whole.substr(begin.length);
+function stripBaseUrl(base, url) {
+  if (startsWith(url, base)) {
+    return url.substr(base.length);
   }
 }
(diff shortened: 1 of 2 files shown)


petebacondarwin:
This is mostly to trigger Travis

petebacondarwin:
Landed as a4e4feed909408ff837ac328371d0fad6ca49e9d

src/ng/location.js
```diff
@@ -909,7 +912,7 @@ function $LocationProvider() {
     // update $location when $browser url changes
     $browser.onUrlChange(function(newUrl, newState) {
 
-      if (isUndefined(beginsWith(appBaseNoFile, newUrl))) {
+      if (isUndefined(stripBaseUrl(appBaseNoFile, newUrl))) {
```
lgalfaso: Better
if (!startsWith(...)) {

?

src/ng/location.js
```diff
@@ -48,17 +48,20 @@ function parseAppUrl(relativeUrl, locationObj) {
   }
 }
 
+function startsWith(haystack, needle) {
```
lgalfaso: I wonder if we should move this to `src/Angular.js` as it can be quite handy

jbedard: I'd think a generic helper like this should. Could even use [String#startsWith](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) when available?

petebacondarwin: One refactoring at a time :-)
At this stage this would only be used here and in $sanitize. So I felt it was simpler right now not to move it out.
But sure, it would be fine to do so.

src/ng/location.js
petebacondarwin: That would be good. Easy PR for you @lgalfaso  :-)

lgalfaso: Ok cfc8b41771e6f2f5bcf692400e960144009b1f18

PR #14665, opened 2016-05-25
feat(ngAria[ngClick]): preventDefault on pressing Space so the page doesn't scroll

stowball:
This is a bug fix for the ngAria ngClick directive.

When a user presses space on a "normal" `button`, the page should not scroll. However, with a "fake", `[role="button"]`, pressing space doesn't `preventDefault()`, so it scrolls the page.

You can see the bad behavior here: https://plnkr.co/edit/SPxQ0X2RQvvfvAmDVTNA?p=preview 

This PR adds `event.preventDefault();` so that pressing Space doesn't activate the native browser function.

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -377,6 +377,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             elem.on('keypress', function(event) {
               var keyCode = event.which || event.keyCode;
               if (keyCode === 32 || keyCode === 13) {
+                event.preventDefault();
                 scope.$apply(callback);
               }
 


marcysutton:
Seems straight forward enough for me. I verified it still works when the div has an interactive child element: https://plnkr.co/edit/SPxQ0X2RQvvfvAmDVTNA?p=preview

But it highlights another problem, which I will file as an issue: ngAria needs to check for at least one interactive child element before binding `role="button"` and `tabindex="0"` because otherwise you get interactive elements wrapped around other interactive elements. This is a problem for screen reader users, and why a `<button>` element wrapping other `<button>` elements is invalid HTML.

src/ngAria/aria.js
```diff
@@ -377,6 +377,7 @@ ngAriaModule.directive('ngShow', ['$aria', function($aria) {
             elem.on('keypress', function(event) {
```
tybiwety23: 32

gkalpak:
I've somehow missed that :smiley: LGTM, except for the lack of tests.
@stowball, would you be interested in adding some?

tybiwety23:
Leave a commement

stowball:
I'm new to testing. How would we go about testing the whether the window scroll position changed?

gkalpak:
It would be a little involved, because you would have to make sure there is room to scroll and then check the windows `scrollTop` probably. But I don't think you have to. You can just check that `preventDefault()` is called on the dispatched event (for <kbd>ENTER</kbd>, <kbd>SPACEBAR</kbd> only).

PR #14671, opened 2016-05-25
docs(tutorial/Tutorial): First page of documents detail addition regarding node js

MrChriZ:
docs update

No

**Other information**:
On my system at least the node command did not work in my bash shell after freshly installing Node.JS. 
After restarting bash it must have picked up an updated path variable.

### DIFF
--- a/docs/content/tutorial/index.ngdoc
+++ b/docs/content/tutorial/index.ngdoc
@@ -104,7 +104,7 @@ If you want to run the preconfigured local web server and the test tools then yo
 
 You can download a Node.js installer for your operating system from https://nodejs.org/en/download/.
 
-Check the version of Node.js that you have installed by running the following command:
+Check the version of Node.js that you have installed by running the following command (may require restarting of console window if freshly installed):
 
 ```
 node --version


MrChriZ:
I signed it

gkalpak:
I feel that this kind of detail is out of scope for the tutorial (not to mention that different OS/versions may have different requirements).

Thx for taking the time to submit a PR :+1:
Keep 'em coming !

MrChriZ:
Thanks Georgios - just for reference though the very next line is OS dependent and therefore beyond the scope of the tutorial:"In Debian based distributions, there might be a name clash with another utility called node. The suggested solution is to also install thenodejs-legacy apt package, which renames node to nodejs.apt-get install nodejs-legacy npm
nodejs --version npm --version"

So maybe those lines should be deleted too?

gkalpak:
I think those were added at a time when Node.js wasn't so popular. TBH, I currently don't use the `nodejs-legacy` package on Debian - not sure if it's necessary any more.

If you ask me, I would gladly remove that comment too (because out of scope), but others might object :grin:

Issue #14673, opened 2016-05-25
ng-click causes digest already in progress when triggering with jquery


markgardner:
**_Note_: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a _feature_ or report a _bug_?**
Bug.

**What is the current behavior?**
When triggering an event using jquery within a ng-click handler trigger, it can cause a digest already in progress error.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**
This plnkr should illustrate a possible problem. However, I'm sure this happens often when integrating 3rd-party libaries into angular.
http://plnkr.co/edit/VAXafyll7hTXR5BR5FrR

**What is the expected behavior?**
Ideally, the common ng event directives would be able to call their handlers whether there is an active digest cycle or not.

**What is the motivation / use case for changing the behavior?**
I'm experiencing this problem in my work's app with a few components as we are growing our event handlers.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**
This appears to not be specific to any particular browser / OS.

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
I have a potential PR that I can submit with this issue.

PR #14674, opened 2016-05-25
fix(ngEventDirs): check for $rootScope.$$phase in event handler and d…

markgardner:
Bug fix

#14673

Digest cycle already in progress error can inadvertently be caused when triggering an
element's click event while within an active digest cycle. This is due to the ngEventsDirs
event handler always calling $rootScope.$apply regardless of the status of $rootScope.$$phase.
Checking the phase and calling the function immediately if within an active digest cycle
will prevent the problem without reducing current functionality.

This will change ngClick and the like to check for $rootScope.$$phase to determine if it should call the callback function without entering a new $apply.

Nope

**Other information**:

### DIFF
--- a/src/ng/directive/ngEventDirs.js
+++ b/src/ng/directive/ngEventDirs.js
@@ -63,10 +63,12 @@ forEach(
               var callback = function() {
                 fn(scope, {$event:event});
               };
-              if (forceAsyncEvents[eventName] && $rootScope.$$phase) {
+              if (!$rootScope.$$phase) {
+                scope.$apply(callback);
+              } else if (forceAsyncEvents[eventName]) {
                 scope.$evalAsync(callback);
               } else {
-                scope.$apply(callback);
+                callback();
               }
             });
           };
--- a/test/ng/directive/ngEventDirsSpec.js
+++ b/test/ng/directive/ngEventDirsSpec.js
@@ -150,4 +150,46 @@ describe('event directives', function() {
     }));
 
   });
+
+  describe('click', function() {
+
+    it('should call the listener synchronously if inside of $apply',
+        inject(function($rootScope, $compile) {
+      var watchedVal;
+
+      element = $compile('<button type="button" ng-click="click()">Button</button>')($rootScope);
+      $rootScope.$watch('value', function(newValue) {
+        watchedVal = newValue;
+      });
+      $rootScope.click = jasmine.createSpy('click').and.callFake(function() {
+        $rootScope.value = 'newValue';
+      });
+
+      $rootScope.$apply(function() {
+        element.triggerHandler('click');
+      });
+
+      expect($rootScope.click).toHaveBeenCalledOnce();
+      expect(watchedVal).toEqual('newValue');
+    }));
+
+    it('should call the listener synchronously if outside of $apply',
+        inject(function($rootScope, $compile) {
+      var watchedVal;
+
+      element = $compile('<button type="button" ng-click="click()">Button</button>')($rootScope);
+      $rootScope.$watch('value', function(newValue) {
+        watchedVal = newValue;
+      });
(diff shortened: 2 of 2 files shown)


src/ng/directive/ngEventDirs.js
```diff
@@ -65,6 +65,8 @@ forEach(
               };
               if (forceAsyncEvents[eventName] && $rootScope.$$phase) {
                 scope.$evalAsync(callback);
+              } else if ($rootScope.$$phase) {
```
gkalpak: In order to avoid double checking `$rootScope.$$phase` and simplify, I might change the whole block to:

```js
if (!$rootScope.$$phase) {
  scope.$apply(callback);
} else if (forceAsyncEvents[eventName]) {
  scope.$evalAsync(callback);
} else {
  callback();
}
```

gkalpak:
Technically speaking, there are ways to avoid this (e.g. wrapping the call to `click()` in a `$timeout` or preventing event propagation) and having an ngClick handler on an element that calls `.click()` on a sibling element, while their parent element also has an ngClick handler isn't a very common usecase :smiley:

But then again, why not ? It has minimal overhead and is a reasonable change :+1: 

LGTM

src/ng/directive/ngEventDirs.js
markgardner: Great point, I've updated the commit to reflect this.

markgardner:
@gkalpak very true, I have had this happen a couple of times and normally worked around it.

jbedard:
I think the reason this was never done is because it makes `element.click()` unpredictable. With this change the event handler might be async or might not. If it is async then things like `event.preventDefault()` or `event.stopPropagation()` won't work, event order might become unpredictable etc.  Blur and focus were special cases because they are sometimes automatically triggered by the browser when manipulating the dom (at least blur is).

markgardner:
@jbedard that makes sense that would be a concern. Though I don't think this change will effect that. The biggest difference is whether it will use $rootScope.$apply or a regular function call to call the event handler. Both should be synchronous if I understand the code correctly.

jbedard:
My bad then! I guess it's only the `forceAsyncEvents` have that behavior and this doesn't change that. I wonder if it is worth moving the `expect($rootScope.click).toHaveBeenCalledOnce()` into the `$apply` to verify this?

gkalpak:
Previously if an non-async ng-event was fired while a `$digest` was in progress, there would be an error. With this PR, such calls will execute the callback directly (without `$apply()`).

So, the only change in terms of behavior is that calls that would previously throw, now won't.

--
Moving the expectation inside the `$apply` block won't hurt, but it doesn't buy us anything either. It only verifies half of the story (events fired while `$digest` in progress) - it doesn't tell us anything about events fired outside of a `$digest`.

I would be fine adding adding another test that verifies the synchronous nature of both types of callback calls (events fired inside and outside of `$digest`).

markgardner:
I've updated the PR to include another test to cover triggering an event outside of a digest cycle.

test/ng/directive/ngEventDirsSpec.js
```diff
@@ -150,4 +150,38 @@ describe('event directives', function() {
...
+    it('should call the listener synchronously if inside of $apply',
+        inject(function($rootScope, $compile) {
+      element = $compile('<button type="button" ng-click="click()">Button</button>')($rootScope);
+      $rootScope.click = jasmine.createSpy('click');
+
+      $rootScope.$apply(function() {
+        element.triggerHandler('click');
+      });
+
+      expect($rootScope.click).toHaveBeenCalledOnce();
```
gkalpak: This doesn't actually check that it is called synchronously (e.g. it might be called outside of the `$apply()` block).
You could use the same approach as below (using a `$watch`).

markgardner: I've updated the PR to use the same checks for both tests

Issue #14683, opened 2016-05-26
ngChange: allow locals to be passed from ngModel


dcherman:
***Note*: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**

**What is the expected behavior?**

**What is the motivation / use case for changing the behavior?**

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

gkalpak:
It would have to pass through several layers to get to `ngChange` :smiley:
$setViewValue --> debounce --> parse --> validate --> $viewChangeListeners --> ngChange

I'm not sure it's worth the trouble. I wonder if it would be more straight-forward to have a custom directive and augment the `NgModelController`.

I'm not saying I couldn't get convinved otherwise...

<sub>(Whatever you do, don't call it `$locals` though :stuck_out_tongue: </sub>

dcherman:
@gkalpak It's actually not bad at all.  I wrote up a quick PR that accomplishes what I described in #14684.  If it seems reasonable, I'll clean it up with tests and docs and get it merge ready.

PR #14684, opened 2016-05-27
feat(ngModel): support locals being passed through to ngChange

dcherman:
This is a WIP proposal for #14683.  If the approach seems acceptable, I'll write up accompanying unit tests and documentation.

### DIFF
--- a/src/ng/directive/ngChange.js
+++ b/src/ng/directive/ngChange.js
@@ -72,7 +72,7 @@ var ngChangeDirective = valueFn({
   require: 'ngModel',
   link: function(scope, element, attr, ctrl) {
     ctrl.$viewChangeListeners.push(function() {
-      scope.$eval(attr.ngChange);
+      scope.$eval(attr.ngChange, ctrl.$ngChangeLocals);
     });
   }
 });
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -760,6 +760,8 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
         $exceptionHandler(e);
       }
     });
+
+    this.$ngChangeLocals = undefined;
   };
 
   /**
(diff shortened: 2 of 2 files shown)


gkalpak:
I'm not a big fan of this tbh. Things I don't like:

1. It doesn't feel like a common feature. Besides it is doable outside the core.
2. The concept of `locals` is not the most trivial one to explain. This adds an overhead to simple concepts, such as `ngChange`.

--
Anyway, it is a tiny change with minimal overhead. So, if others think this "view value metadata" feature would be useful, I won't stand in your way :stuck_out_tongue:

--
Things I would consider:

1. Making the metadata available via locals as `$metadata` or `$viewValueMetadata` or something. This would use the same mechanism under the hood, but users don't have to understand scope evaluation and locals.
2. Passing the metadata as argument to all `$viewChangeListeners`.

dcherman:
It's more common for implementers of library controls with complex behaviors.

The two primary use cases that have come up for us are:

1.  Selectable Behaviors: When a user shift clicks or ctrl clicks multiple items, the resulting ngModel value is the value extracted from an expression on those items.  It's useful to provide additional information about the operation though such as the value of the item that was actually clicked; that metadata is used to provide behaviors like scrolling the item fully into view in case it was partially clipped.

2.  Drag and Drop - A similar thing, the ngModel value might be the values of the item(s) that were dropped, however additional metadata about the operation allows you to customize the behavior such as removing the item from the source if you're dragging between two lists.

These are admittedly not the common use case, but I figured it'd be worth seeing if others would find value in this.

petebacondarwin:
I also don't feel very comfortable with this feature.
It would seem to me that a better place to do this is in a `$parser`. Then you can pass a complex object, containing the actual main value and any metadata, to the `$setViewValue` method and the `$parser` can do any special work that it wants before pulling out the main value and passing that down the pipeline.

petebacondarwin:
Closing due to lack of response

PR #14685, opened 2016-05-27
fix(input): listen on "change" for radio and checkbox

### LINKED ISSUE #4516: Bootstrap 3 javascript checkbox and radio button toggle prevents angular's ng-model from updating on input
When using Bootstrap 3's javascript buttons for checkboxes, the checkbox's ng-model does not get updated. Looking into the source code for both I know why.

Bootstrap 3 html for checkbox button group:

``` HTML
<div class="btn-group" data-toggle="buttons">
  <label class="btn btn-primary">
    <input type="checkbox"> Option 1
  </label>
  <label class="btn btn-primary">
    <input type="checkbox"> Option 2
  </label>
  <label class="btn btn-primary">
    <input type="checkbox"> Option 3
  </label>
</div>
```

Bootstrap 3 javascript for button toggle:

``` JavaScript
Button.prototype.toggle = function () {
        var $parent = this.$element.closest('[data-toggle="buttons"]')

        if ($parent.length) {
            var $input = this.$element.find('input')
              .prop('checked', !this.$element.hasClass('active'))
              .trigger('change')
            if ($input.prop('type') === 'radio') $parent.find('.active').removeClass('active')
        }

        this.$element.toggleClass('active')
    }
```

Angular source for checkbox (radio button is similar):

``` JavaScript
function checkboxInputType(scope, element, attr, ctrl) {
        var trueValue = attr.ngTrueValue,
            falseValue = attr.ngFalseValue;

        if (!isString(trueValue)) trueValue = true;
        if (!isString(falseValue)) falseValue = false;

        element.on('click', function () {
            scope.$apply(function () {
                ctrl.$setViewValue(element[0].checked);
            });
        });

        ctrl.$render = function () {
            element[0].checked = ctrl.$viewValue;
        };

        ctrl.$formatters.push(function (value) {
            return value === trueValue;
        });

        ctrl.$parsers.push(function (value) {
            return value ? trueValue : falseValue;
        });
    }
```

The problem is Angular's checkbox is listening for the `click` event whereas Bootstrap's button handles the `click` event and fires the `change` event on the checkbox input.

I wrote a simple directive that does exactly what Angular's checkbox `click` event listener does but instead listens for the `change` event.

``` JavaScript
appModule.directive('mccCheckboxDetectChange', [function mccCheckboxDetectChange() {

    return {
        replace: false,
        require: 'ngModel',
        scope: false,
        link: function (scope, element, attrs, ngModelCtrl) {
            element.on('change', function () {
                scope.$apply(function () {
                    ngModelCtrl.$setViewValue(element[0].checked);
                });
            });
        }
    };
}]);
```

Is there a reason why Angular listens for the `click` event for radio and checkboxes instead of the `change` event? Should I create a pull request and change the `click` event listener to be a `change` event listener? This would be my first pull request.

Thanks!
Philip

### LINKED ISSUE #14667: ng-model gets updated before ng-click on input radio
**_Note_: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

Bug

ng-model gets updated before ng-click on input radio and there is no way to prevent the default action

It should be a simple way to prevent default action.

Browsers: Chrome, IE11
Angularjs: 1.5.0

Narretz:
bug fix

radio and checkbox listen on "click" which can cause unforeseen behavior

radio and checkbox listen on "change"

Yes, for certain unit-test behaviors

**Other information**:

input[radio] and inout[checkbox] now listen on the change event instead
of the click event. This fixes issue with 3rd party libraries that trigger
a change event on inputs, e.g. Bootstrap 3 custom checkbox / radio button
toggles.
It also makes it easier to prevent specific events that can cause a checkbox / radio
to change, e.g. click events. Previously, this was difficult because the custom click
handler had to be registered before the input directive's click handler.

It is possible that radio and checkbox listened to click because IE8 has
broken support for listening on change, see http://www.quirksmode.org/dom/events/change.html

Closes #4516
Closes #14667

BREAKING CHANGE:

input[radio] and input[checkbox] now need to be attached to the document to propagate events
correctly. This should only be of concern in unit-tests that compile input elements and
trigger click events on them. This is because we now listen to the change event which
gets automatically triggered by browsers when a checkbox or radio is clicked. However,
this may fail in some browsers when the elements are not attached to the document.

Before:

``` js
    it('should update the model', inject(function($compile, $rootScope) {
      var inputElm = $compile('<input type="checkbox" ng-model="checkbox" />')($rootScope);

      browserTrigger(inputElm[0], 'click');
      expect($rootScope.checkbox).toBe(true);
    });
```

With this patch, `$rootScope.checkbox` might not be true, because the click event
hasn't triggered the change event. To make the test, work append the inputElm to the app's
$rootElement, and the $rootElement to the $document:

After:

``` js
    it('should update the model', inject(function($compile, $rootScope, $rootElement, $document) {
      var inputElm = $compile('<input type="checkbox" ng-model="checkbox" />')($rootScope);

      $rootElement.append(inputElm);
      $document.append($rootElement);

      browserTrigger(inputElm[0], 'click');
      expect($rootScope.checkbox).toBe(true);
    });
```

### DIFF
--- a/test/BinderSpec.js
+++ b/test/BinderSpec.js
@@ -401,12 +401,17 @@ describe('Binder', function() {
     expect(optionC.text()).toEqual('C');
   }));
 
-  it('ItShouldSelectTheCorrectRadioBox', inject(function($rootScope, $compile) {
+  it('ItShouldSelectTheCorrectRadioBox', inject(function($rootScope, $compile, $rootElement, $document) {
     element = $compile(
       '<div>' +
         '<input type="radio" ng-model="sex" value="female">' +
         '<input type="radio" ng-model="sex" value="male">' +
       '</div>')($rootScope);
+
+    // Append the app to the document so that "click" on a radio/checkbox triggers "change"
+    // Support: Chrome, Safari 8, 9
+    jqLite($document[0].body).append($rootElement.append(element));
+
     var female = jqLite(element[0].childNodes[0]);
     var male = jqLite(element[0].childNodes[1]);
 
--- a/test/helpers/testabilityPatch.js
+++ b/test/helpers/testabilityPatch.js
@@ -319,7 +319,7 @@ function generateInputCompilerHelper(helper) {
         };
       });
     });
-    inject(function($compile, $rootScope, $sniffer) {
+    inject(function($compile, $rootScope, $sniffer, $document, $rootElement) {
 
       helper.compileInput = function(inputHtml, mockValidity, scope) {
 
@@ -340,6 +340,11 @@ function generateInputCompilerHelper(helper) {
 
         // Compile the lot and return the input element
         $compile(helper.formElm)(scope);
+
+        $rootElement.append(helper.formElm);
+        // Append the app to the document so that "click" on a radio/checkbox triggers "change"
+        // Support: Chrome, Safari 8, 9
+        jqLite($document[0].body).append($rootElement);
 
         spyOn(scope.form, '$addControl').and.callThrough();
         spyOn(scope.form, '$$renameControl').and.callThrough();
--- a/test/ng/directive/inputSpec.js
+++ b/test/ng/directive/inputSpec.js
@@ -3974,7 +3974,7 @@ describe('input', function() {
 
   describe('radio', function() {
 
-    it('should update the model', function() {
+    they('should update the model on $prop event', ['click', 'change'], function(event) {
       var inputElm = helper.compileInput(
           '<input type="radio" ng-model="color" value="white" />' +
           '<input type="radio" ng-model="color" value="red" />' +
@@ -3990,7 +3990,8 @@ describe('input', function() {
       expect(inputElm[1].checked).toBe(true);
       expect(inputElm[2].checked).toBe(false);
 
-      browserTrigger(inputElm[2], 'click');
+      if (event === 'change') inputElm[2].checked = true;
+      browserTrigger(inputElm[2], event);
       expect($rootScope.color).toBe('blue');
     });
 
@@ -4089,6 +4090,23 @@ describe('input', function() {
       expect(inputElm.hasClass('ng-invalid')).toBe(false);
       expect(inputElm.hasClass('ng-pristine')).toBe(false);
       expect(inputElm.hasClass('ng-dirty')).toBe(false);
+    });
+
+
+    they('should update the model on $prop event', ['click', 'change'], function(event) {
+      var inputElm = helper.compileInput('<input type="checkbox" ng-model="checkbox" />');
+
+      expect(inputElm[0].checked).toBe(false);
+
+      $rootScope.$apply('checkbox = true');
+      expect(inputElm[0].checked).toBe(true);
+
+      $rootScope.$apply('checkbox = false');
+      expect(inputElm[0].checked).toBe(false);
+
+      if (event === 'change') inputElm[0].checked = true;
+      browserTrigger(inputElm[0], event);
+      expect($rootScope.checkbox).toBe(true);
     });
 
 
--- a/test/ng/directive/ngRepeatSpec.js
+++ b/test/ng/directive/ngRepeatSpec.js
@@ -354,7 +354,7 @@ describe('ngRepeat', function() {
     });
 
 
-    it('should iterate over object with changing primitive property values', function() {
+    it('should iterate over object with changing primitive property values', inject(function($rootElement, $document) {
       // test for issue #933
 
       element = $compile(
@@ -364,6 +364,10 @@ describe('ngRepeat', function() {
               '<input type="checkbox" ng-model="items[key]">' +
             '</li>' +
           '</ul>')(scope);
+
+      // Append the app to the document so that "click" on a radio/checkbox triggers "change"
+      // Support: Chrome, Safari 8, 9
+      jqLite($document[0].body).append($rootElement.append(element));
 
       scope.items = {misko: true, shyam: true, zhenbo:true};
       scope.$digest();
@@ -395,7 +399,7 @@ describe('ngRepeat', function() {
       expect(element.find('input')[0].checked).toBe(false);
       expect(element.find('input')[1].checked).toBe(true);
       expect(element.find('input')[2].checked).toBe(true);
-    });
+    }));
   });
 
   describe('alias as', function() {
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1820,7 +1820,7 @@ function radioInputType(scope, element, attr, ctrl) {
     }
   };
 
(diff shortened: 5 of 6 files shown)


test/helpers/testabilityPatch.js
```diff
@@ -381,6 +381,9 @@ function generateInputCompilerHelper(helper) {
         // Compile the lot and return the input element
         $compile(helper.formElm)(scope);
 
+        $rootElement.append(helper.formElm);
+        jqLite($document[0].body).append($rootElement);
```
gkalpak: OOC, why is this needed ?

test/ng/directive/inputSpec.js
```diff
@@ -2948,7 +2948,8 @@ describe('input', function() {
       expect(inputElm[1].checked).toBe(true);
       expect(inputElm[2].checked).toBe(false);
 
-      browserTrigger(inputElm[2], 'click');
+      if (event === 'change') inputElm[2].checked = true;
```
gkalpak: This looks fishy...

gkalpak: But then I guess you can't do much...
<sub>
Maybe we should document that for custom checkboxes/radios, where the change event is triggered manually, you need to ensure `.checked` has been set first. Or maybe it's obvious and doesn't need documentation..
</sub>

test/helpers/testabilityPatch.js
gkalpak: OK, I see why it is needed. Is it necessary on all browsers or some of them ? Which ones ?

test/ng/directive/inputSpec.js
Narretz: This is expected, I guess. This is exactly what bootstrap 3 does, see the example code here: https://github.com/angular/angular.js/issues/4516#issue-21221793

gkalpak:
Generally, LGTM.

I am a little worried of what might unforseeably break as a result of this change. For example, usecases depending on the relative order of event callbacks (e.g. `click` listeners registered after our listener and expecting the model to have been updated).

We should mention it in the commit message imo.
There might be other cases that I missed.

--
Talking about BC messages, mentioning `browserTrigger` would be confusing (since it Angular internal stuff). I would just replace it with a comment, e.g.: `// Somehow trigger a `click` event on the element`.

test/helpers/testabilityPatch.js
Narretz: Some browsers don't trigger the "change" event on "click" when the input isn't attached to the document. This may be a bug, but it also makes sense. This is why the BC should only appear in (unit )tests.

Narretz:
You are right, this could break some other cases.
And I always thought browserTrigger was external ... comes from working on core so much!

test/helpers/testabilityPatch.js
Narretz: If we use browserTrigger, it is needed in Chrome. It works in Firefox. Haven't tested other browsers. If you use the native element.click() fn, you need it in both FF and Chrome

Narretz:
I've fixed the test and updated the commit message / BC note

gkalpak:
Something is still failing on mobile Safari :disappointed:

Narretz:
@gkalpak Yeah I noticed that. Is probably gonna be a pain in the ass to fix.

test/helpers/testabilityPatch.js
petebacondarwin: We recently changed `browserTrigger` to manually raise events if the node is not connected to a document - https://github.com/angular/angular.js/commit/c8bfbfc5d7d66b4913874ae2f1c0df6c0dea7857 - would that solve this problem?

Narretz: Possibly, I'd have to test. It's not so much an issue, rather a slight annoyance. The PR still isn't grren because an ngTouch test fails on Safari iOS - related to the clickbusting code in ngClick. I don't have an iOs device, so I can't test this.

mgol:
@Narretz This just hit me in a production-running app that also used FastClick. FastClick morphs click handlers into touchstart ones with some hacks around it and in 50% of cases this broke the radios on the site in iOS 10.

Narretz:
Intereesting. Maybe this will pass in ios? Because so far ios8 is blocking this ...

mgol:
Angular 2 supports iOS 7-10 so I'd rather not drop iOS 8 support in ng1; we
should try to cover the same browsers IMO...

Narretz:
@petebacondarwin @gkalpak I've updated this (BC) PR for 1.7.0. Can you give it another look? Or probably a new look since it's from last year. ;)
Specifically I updated the commit message to include more info about possible Breaking Changes and info about which browsers are affected by the event propagation issue.

test/BinderSpec.js
```diff
@@ -401,12 +401,15 @@ describe('Binder', function() {
...
 
-  it('ItShouldSelectTheCorrectRadioBox', inject(function($rootScope, $compile) {
+  it('ItShouldSelectTheCorrectRadioBox', inject(function($rootScope, $compile, $rootElement, $document) {
     element = $compile(
       '<div>' +
         '<input type="radio" ng-model="sex" value="female">' +
         '<input type="radio" ng-model="sex" value="male">' +
       '</div>')($rootScope);
+
+    jqLite($document[0].body).append($rootElement.append(element));
```
gkalpak: Maybe leave a comment here why this is needed.

test/ng/directive/ngRepeatSpec.js
```diff
@@ -365,6 +365,8 @@ describe('ngRepeat', function() {
             '</li>' +
           '</ul>')(scope);
 
+      jqLite($document[0].body).append($rootElement.append(element));
```
gkalpak: I think a comment would be nice here too.

gkalpak (approved):
LGTM 👍 

I am not thrilled about the fact that clicking on an detached element does not trigger a `change` event (but this is a browser/spec issue, not a problem with this PR 😞). I have briefly thought about changing `browserTrigger()` to also emit a `change` event if an detached checkbox/radiobutton's `checked` property is changed as a result of ` click` event, but it felt too automagical :smiley:

A couple of commit message nitpicks:
- Mention "instead of `click`" in the commit message subject (to make it obvious for those readng through the changelog - it now sounds more like `change` is listened to _in addition_ to `click`).
- _you have don't have to attach_ --> _you don't have to attach_
- Put injectables and function names in `` `...` `` (e.g. `$rootElement`, `$document`, `triggerHandler()`).

Narretz:
I Opened an issue on Chromium for future reference: https://bugs.chromium.org/p/chromium/issues/detail?id=773680

PR #14687, opened 2016-05-27
fix($compile): turn link[href] into a RESOURCE_URL context.

rjamet:

Bug fix.

There's no $sce context for link[href], meaning a <link rel="foo" href="{{bar}}" /> tag will have no $sce protection (might have link sanitization though, I'm not sure of that). That's fine for several rel types, but at least stylesheet and import can run script in your origin, which corresponds to a RESOURCE_URL context.

Turn it into a RESOURCE_URL context. Note that this is an overcorrection, sadly, but the security level of this attribute depends on the rel attribute, and I don't know if we can make context-aware decisions in this type of situation. 

RESOURCE_URLs are allowed to be interpolated if the value is $sce-blessed, or for plain strings if they're from the same origin.

Angular2 is being fixed to do the same with https://github.com/angular/angular/pull/8869/files#diff-5ffbf51559339aaa2db503c37cc3a658R60.

Most apps keep the default whitelist, so it will break apps with dynamically-set link[href]s pointing to other origins.

No comprehensive list of contexts in the docs to update, they just say the $sce.RESOURCE_URL type is "For URLs that are not only safe to follow as links, but whose contents are also safe to include in your application. Examples include ... ".

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2953,6 +2953,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       // maction[xlink:href] can source SVG.  It's not limited to <maction>.
       if (attrNormalizedName === "xlinkHref" ||
           (tag === "form" && attrNormalizedName === "action") ||
+          // links can be stylesheets or imports, which can run script in the current origin
+          (tag === "link" && attrNormalizedName === "href") ||
           (tag !== "img" && (attrNormalizedName === "src" ||
                             attrNormalizedName === "ngSrc"))) {
         return $sce.RESOURCE_URL;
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@mprobst can you give a view on this - since you are working on the A2 version?

mprobst:
Yes, this is correct. Thanks for the fix Raphael!

gkalpak:
Isn't this a BC ?

mprobst:
This is a breaking change. LGTMs though.

gkalpak:
LGTM2 - I'm just saying we should update the commit message with a BC notice (including instructions on how to update old code).

Issue #14657, opened 2016-05-24
limitTo is using isArray


bathos:
***Note*: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

The limitTo filter performs a hard `isArray` check and exits early if this evaluates to false, returning the whole input. Note that the `isArray` function is `Array.isArray`, which only returns true for objects which are *"precisely"* instance of Array, even if they are in fact otherwise *instances of* Array.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**

https://github.com/angular/angular.js/blob/master/src/ng/filter/limitTo.js#L111

**What is the expected behavior?**

I would expect one of the following:

1. Most natural would be for limitTo to use the same criteria as ngRepeat itself to determine whether its input is an ordered, indexed iterable. NgRepeat uses `isArrayLike` to make the same judgment and limitTo is a natural companion to ngRepeat. However it is conceivable that not supporting objects which merely have array-like interfaces is intentional, because limitTo uses `slice` -- not that this couldn't be changed, but it is more involving.

2. That it not ignore Array instances that are not themselves exotic.

**What is the motivation / use case for changing the behavior?**

As far as I can tell, there's no reason to be restricting limitTo this way, and most likely the `isArray` usage is an artifact of the time this was first written, because this distinction would not have mattered much a few years ago.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**

I came across the problem in 1.4.7, but it appears to be unchanged in the most recent code as well.

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

In general, I would suggest `Array.isArray` is being misused. It's a sort of technical check that returns true [only for array exotics](http://www.ecma-international.org/ecma-262/6.0/#sec-isarray), meaning objects created either with literal syntax or the Array constructor, which have exotic proxy traps and which cannot otherwise be identified with total certainty by any other means. Misleading name aside, it is not really for checking if something "is an array" in the regular sense; the way to do this is `instanceof Array`, as with any other constructor.

[...]

PR #14694, opened 2016-05-31
feat(limitTo): add support for array-likes

wesleycho:
- Add support for array-like objects

This implements #14657.

### DIFF
--- a/src/ng/filter/limitTo.js
+++ b/src/ng/filter/limitTo.js
@@ -9,9 +9,10 @@
  * Creates a new array or string containing only a specified number of elements. The elements
  * are taken from either the beginning or the end of the source array, string or number, as specified by
  * the value and sign (positive or negative) of `limit`. If a number is used as input, it is
- * converted to a string.
+ * converted to a string. Array-like values (e.g. NodeLists, jQuery objects, TypedArrays, Strings, etc)
+ * are also supported.
  *
- * @param {Array|string|number} input Source array, string or number to be limited.
+ * @param {Array} input Array or array-like to be limited.
  * @param {string|number} limit The length of the returned array or string. If the `limit` number
  *     is positive, `limit` number of items from the beginning of the source array/string are copied.
  *     If the number is negative, `limit` number  of items from the end of the source array/string
@@ -108,19 +109,25 @@ function limitToFilter() {
     if (isNaN(limit)) return input;
 
     if (isNumber(input)) input = input.toString();
-    if (!isArray(input) && !isString(input)) return input;
+    if (!isArrayLike(input)) return input;
 
     begin = (!begin || isNaN(begin)) ? 0 : toInt(begin);
     begin = (begin < 0) ? Math.max(0, input.length + begin) : begin;
 
     if (limit >= 0) {
-      return input.slice(begin, begin + limit);
+      return sliceFn(input, begin, begin + limit);
     } else {
       if (begin === 0) {
-        return input.slice(limit, input.length);
+        return sliceFn(input, limit, input.length);
       } else {
-        return input.slice(Math.max(0, begin + limit), begin);
+        return sliceFn(input, Math.max(0, begin + limit), begin);
       }
     }
   };
+}
+
+function sliceFn(input, begin, end) {
+  if (isString(input)) return input.slice(begin, end);
+
+  return slice.call(input, begin, end);
 }
--- a/test/ng/filter/limitToSpec.js
+++ b/test/ng/filter/limitToSpec.js
@@ -4,12 +4,26 @@ describe('Filter: limitTo', function() {
   var items;
   var str;
   var number;
+  var arrayLike;
   var limitTo;
 
   beforeEach(inject(function($filter) {
     items = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h'];
     str = "tuvwxyz";
     number = 100.045;
+    arrayLike = {
+      0: 'a',
+      1: 'b',
+      2: 'c',
+      3: 'd',
+      4: 'e',
+      5: 'f',
+      6: 'g',
+      7: 'h',
+      get length() {
+        return Object.keys(this).length - 1;
+      }
+    };
     limitTo = $filter('limitTo');
   }));
 
@@ -21,6 +35,8 @@ describe('Filter: limitTo', function() {
     expect(limitTo(str, '3')).toEqual("tuv");
     expect(limitTo(number, 3)).toEqual("100");
     expect(limitTo(number, '3')).toEqual("100");
+    expect(limitTo(arrayLike, 3)).toEqual(['a', 'b', 'c']);
+    expect(limitTo(arrayLike, '3')).toEqual(['a', 'b', 'c']);
   });
 
   it('should return the first X items beginning from index Y when X and Y are positive', function() {
@@ -28,6 +44,8 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, '3', 3)).toEqual(['d', 'e', 'f']);
     expect(limitTo(str, 3, 3)).toEqual("wxy");
     expect(limitTo(str, '3', '3')).toEqual("wxy");
+    expect(limitTo(arrayLike, 3, 3)).toEqual(['d', 'e', 'f']);
+    expect(limitTo(arrayLike, '3', '3')).toEqual(['d', 'e', 'f']);
   });
 
   it('should return the first X items beginning from index Y when X is positive and Y is negative', function() {
@@ -35,6 +53,8 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, '3', -3)).toEqual(['f', 'g', 'h']);
     expect(limitTo(str, 3, -3)).toEqual("xyz");
     expect(limitTo(str, '3', '-3')).toEqual("xyz");
+    expect(limitTo(arrayLike, 3, '-3')).toEqual(['f', 'g', 'h']);
+    expect(limitTo(arrayLike, '3', -3)).toEqual(['f', 'g', 'h']);
   });
 
   it('should return the last X items when X is negative', function() {
@@ -44,6 +64,8 @@ describe('Filter: limitTo', function() {
     expect(limitTo(str, '-3')).toEqual("xyz");
     expect(limitTo(number, -3)).toEqual("045");
     expect(limitTo(number, '-3')).toEqual("045");
+    expect(limitTo(arrayLike, -3)).toEqual(['f', 'g', 'h']);
+    expect(limitTo(arrayLike, '-3')).toEqual(['f', 'g', 'h']);
   });
 
   it('should return the last X items until index Y when X and Y are negative', function() {
@@ -51,6 +73,8 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, '-3', -3)).toEqual(['c', 'd', 'e']);
     expect(limitTo(str, -3, -3)).toEqual("uvw");
     expect(limitTo(str, '-3', '-3')).toEqual("uvw");
+    expect(limitTo(arrayLike, -3, '-3')).toEqual(['c', 'd', 'e']);
+    expect(limitTo(arrayLike, '-3', -3)).toEqual(['c', 'd', 'e']);
   });
 
   it('should return the last X items until index Y when X is negative and Y is positive', function() {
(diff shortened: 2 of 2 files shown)


test/ng/filter/limitToSpec.js
```diff
@@ -4,15 +4,35 @@ describe('Filter: limitTo', function() {
...
+      6: 'g',
+      7: 'h',
+      get length() {
+        return Object.keys(this).length - 1;
+      }
+    };
     limitTo = $filter('limitTo');
   }));
 
+  function isInObject(obj) {
```
campersau: This is never used.

src/ng/filter/limitTo.js
```diff
@@ -108,19 +108,25 @@ function limitToFilter() {
...
       } else {
-        return input.slice(Math.max(0, begin + limit), begin);
+        return sliceFn(input, Math.max(0, begin + limit), begin);
       }
     }
   };
 }
+
+function sliceFn(input, idx, length) {
+  if (isString(input)) return input.slice(idx, length);
```
gkalpak: I would check if `input` has a `slice` method (or property) so we support Array-like objects with custon `slice` methods.

src/ng/filter/limitTo.js
```diff
@@ -108,19 +108,25 @@ function limitToFilter() {
...
+
+  return Array.prototype.slice.call(input, idx, length);
```
gkalpak: You can also call the "IIFE global" `slice` method: `slice.call(...)`
(It is initialized as `[].slice`.)

test/ng/filter/limitToSpec.js
```diff
@@ -71,6 +105,11 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, 'undefined')).toEqual(items);
     expect(limitTo(items, null)).toEqual(items);
     expect(limitTo(items, undefined)).toEqual(items);
+    expect(limitTo(arrayLike, 'bogus')).toEqual(arrayLike);
+    expect(limitTo(arrayLike, 'null')).toEqual(arrayLike);
+    expect(limitTo(arrayLike, 'undefined')).toEqual(arrayLike);
+    expect(limitTo(arrayLike, null)).toEqual(arrayLike);
+    expect(limitTo(arrayLike, undefined)).toEqual(arrayLike);
```
gkalpak: I would rather use `toBe` in this test as it captures the intent better. (I know it was `toEqual` even before this PR.)

test/ng/filter/limitToSpec.js
```diff
@@ -97,12 +136,16 @@ describe('Filter: limitTo', function() {
     expect(limitTo(str, '3', 'undefined')).toEqual(limitTo(str, '3'));
     expect(limitTo(str, '-3', null)).toEqual(limitTo(str, '-3', 0));
     expect(limitTo(str, 3, undefined)).toEqual(limitTo(str, 3));
+    expect(limitTo(arrayLike, 3, 'bogus')).toEqual(limitTo(items, 3, 0));
+    expect(limitTo(arrayLike, -3, 'null')).toEqual(limitTo(items, -3));
+    expect(limitTo(arrayLike, '3', 'undefined')).toEqual(limitTo(items, '3', 0));
+    expect(limitTo(arrayLike, '-3', null)).toEqual(limitTo(items, '-3'));
+    expect(limitTo(arrayLike, 3, undefined)).toEqual(limitTo(items, 3, 0));
```
gkalpak: _items_  -->  _arrayLike_

test/ng/filter/limitToSpec.js
```diff
@@ -97,12 +136,16 @@ describe('Filter: limitTo', function() {
...
   });
 
-  it('should return input if not String or Array or Number', function() {
+  it('should return input if not String or array-like or Number', function() {
```
gkalpak: Remove _"String or"_.

test/ng/filter/limitToSpec.js
```diff
@@ -97,12 +136,16 @@ describe('Filter: limitTo', function() {
...
     expect(limitTo(null, 1)).toEqual(null);
     expect(limitTo(undefined, 1)).toEqual(undefined);
-    expect(limitTo({}, 1)).toEqual({});
```
gkalpak: Why did you remove this ?

test/ng/filter/limitToSpec.js
```diff
@@ -111,8 +154,13 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, '9')).toEqual(items);
     expect(limitTo(items, -9)).toEqual(items);
     expect(limitTo(items, '-9')).toEqual(items);
+    expect(limitTo(arrayLike, 9)).toEqual(items);
+    expect(limitTo(arrayLike, '9')).toEqual(items);
+    expect(limitTo(arrayLike, -9)).toEqual(items);
+    expect(limitTo(arrayLike, '-9')).toEqual(items);
 
-    expect(limitTo(items, 9)).not.toBe(items);
```
gkalpak: Why did you remove this ?

test/ng/filter/limitToSpec.js
```diff
@@ -111,8 +154,13 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, '9')).toEqual(items);
     expect(limitTo(items, -9)).toEqual(items);
     expect(limitTo(items, '-9')).toEqual(items);
+    expect(limitTo(arrayLike, 9)).toEqual(items);
+    expect(limitTo(arrayLike, '9')).toEqual(items);
+    expect(limitTo(arrayLike, -9)).toEqual(items);
+    expect(limitTo(arrayLike, '-9')).toEqual(items);
```
gkalpak: _items_  -->  _arrayLike_

test/ng/filter/limitToSpec.js
```diff
@@ -111,8 +154,13 @@ describe('Filter: limitTo', function() {
...
 
-    expect(limitTo(items, 9)).not.toBe(items);
+    expect(limitTo(arrayLike, 9)).not.toBe(items);
+    expect(limitTo(arrayLike, 9)).not.toBe(items);
```
gkalpak: _items_  -->  _arrayLike_

test/ng/filter/limitToSpec.js
```diff
@@ -129,6 +177,10 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, 'Infinity')).toEqual(items);
     expect(limitTo(items, -Infinity)).toEqual(items);
     expect(limitTo(items, '-Infinity')).toEqual(items);
+    expect(limitTo(arrayLike, Infinity)).toEqual(items);
+    expect(limitTo(arrayLike, 'Infinity')).toEqual(items);
+    expect(limitTo(arrayLike, -Infinity)).toEqual(items);
+    expect(limitTo(arrayLike, '-Infinity')).toEqual(items);
```
gkalpak: _items_  -->  _arrayLike_

gkalpak:
Generally looks good - I left somme comments.
I think it would be nice to test against a couple more Array-likes (e.g. a NodeList, a jqLite/jQuery collection).

src/ng/filter/limitTo.js
wesleycho: This separation is specifically because `String.prototype.slice` returns a string, while `Array.prototype.slice` returns an array.

wesleycho:
Made adjustments as per comments.

I do have a question though - I notice in the `orderBy` filter, it only checks against a similar object I have constructed here, and not against typical alternative array-likes. Does it make sense to test those here given that?

src/ng/filter/limitTo.js
gkalpak: OK, let's stick with the current separation, so we are always returnign either a string or an array.

test/ng/filter/limitToSpec.js
```diff
@@ -44,38 +64,52 @@ describe('Filter: limitTo', function() {
...
+    expect(limitTo(arrayLike, 'null')).toBe(arrayLike);
+    expect(limitTo(arrayLike, 'undefined')).toBe(arrayLike);
+    expect(limitTo(arrayLike, null)).toBe(arrayLike);
+    expect(limitTo(arrayLike, undefined)).toBe(arrayLike);
   });
 
   it('should return an empty string when X = 0', function() {
     expect(limitTo(str, 0)).toEqual("");
     expect(limitTo(str, '0')).toEqual("");
+    expect(limitTo({}, 1)).toEqual({});
```
gkalpak: Why this ?

test/ng/filter/limitToSpec.js
```diff
@@ -111,8 +149,13 @@ describe('Filter: limitTo', function() {
     expect(limitTo(items, '9')).toEqual(items);
     expect(limitTo(items, -9)).toEqual(items);
     expect(limitTo(items, '-9')).toEqual(items);
+    expect(limitTo(arrayLike, 9)).toEqual(items);
+    expect(limitTo(arrayLike, '9')).toEqual(items);
+    expect(limitTo(arrayLike, -9)).toEqual(items);
+    expect(limitTo(arrayLike, '-9')).toEqual(items);
 
     expect(limitTo(items, 9)).not.toBe(items);
+    expect(limitTo(arrayLike, 9)).not.toBe(items);
```
gkalpak: Shouldn't this be `arrayLike` (instead of `items`) ?

gkalpak:
`$filter` has a couple of array-likes (see [filterSpec.js#L428-L443][1]), but I don't think consistency is important here.

I don't mean add all array-likes in all usecases, just add a simple testcase that tests basic limiting on 2-3 array-like types. But I don't feel strongly about it, so feel free to ignore :smile: 

[1]: https://github.com/angular/angular.js/blob/cfc8b41771e6f2f5bcf692400e960144009b1f18/test/ng/filter/filterSpec.js#L428-L443

wesleycho:
Updated as per comments (sloppiest PR I've done in a while...) - I also added a simple test case for basic array likes at the very end.

src/ng/filter/limitTo.js
wesleycho: One alternative implementation I can do is instead of the `isString(input)` check, I can do
```js
return input.slice ? input.slice.call(input, idx, length) : slice.call(input, idx, length);
```
Thoughts?

wesleycho: Nevermind, looks like this fails on falsy types.

gkalpak: I think it's best to keep things as is, so we don't break the guarantee that `limitTo` returns an Array or String.
(I don't see any real benefit in breaking that contract.)

wesleycho:
Bah, I did a mistaken force push to try to restart the CI - will push again when I'm at work tomorrow, some changes got wiped out.

wesleycho:
This should be up to date now.

src/ng/filter/limitTo.js
```diff
@@ -108,19 +108,25 @@ function limitToFilter() {
...
+        return sliceFn(input, limit, input.length);
       } else {
-        return input.slice(Math.max(0, begin + limit), begin);
+        return sliceFn(input, Math.max(0, begin + limit), begin);
       }
     }
   };
 }
+
+function sliceFn(input, idx, length) {
```
gkalpak: Nitpick: Renaming the arguments to `input, begin, end` sounds more intuitive.

test/ng/filter/limitToSpec.js
```diff
@@ -97,12 +130,16 @@ describe('Filter: limitTo', function() {
...
+    expect(limitTo(arrayLike, '3', 'undefined')).toEqual(limitTo(arrayLike, '3', 0));
+    expect(limitTo(arrayLike, '-3', null)).toEqual(limitTo(arrayLike, '-3'));
+    expect(limitTo(arrayLike, 3, undefined)).toEqual(limitTo(arrayLike, 3, 0));
   });
 
-  it('should return input if not String or Array or Number', function() {
+  it('should return input if not array-like or Number', function() {
     expect(limitTo(null, 1)).toEqual(null);
     expect(limitTo(undefined, 1)).toEqual(undefined);
-    expect(limitTo({}, 1)).toEqual({});
```
gkalpak: Why that ?

test/ng/filter/limitToSpec.js
```diff
@@ -153,19 +201,38 @@ describe('Filter: limitTo', function() {
...
+    function getArguments() {
+      return arguments;
+    }
+    var argsObj = getArguments({name: 'Misko'}, {name: 'Igor'}, {name: 'Brad'});
+
+    var nodeList = jqLite("<p><span>Misko</span><span>Igor</span><span>Brad</span></p>")[0].childNodes;
+
+    expect(limitTo(argsObj, 2).length).toBe(2);
+    expect(limitTo('abc', 1).length).toBe(1);
+    expect(limitTo(nodeList, 3).length).toBe(3);
```
gkalpak: This is not testing anything. You need a `limit` that is less than the actual length.

gkalpak:
Now it's time for the "real stuff": We need to update the docs :smiley:

wesleycho:
Added documentation update, might need tweaking on wording (or maybe removing the bit about strings or numbers).

gkalpak:
@wesleycho, I don't see any docs :worried:

wesleycho:
That's because I'm a moron :( - pushed

src/ng/filter/limitTo.js
```diff
@@ -9,9 +9,10 @@
  * Creates a new array or string containing only a specified number of elements. The elements
  * are taken from either the beginning or the end of the source array, string or number, as specified by
  * the value and sign (positive or negative) of `limit`. If a number is used as input, it is
- * converted to a string.
+ * converted to a string. Array-like values (e.g. NodeLists, jQuery objects, TypedArrays, Strings, etc)
+ * are also supported.
  *
- * @param {Array|string|number} input Source array, string or number to be limited.
+ * @param {Array} input Array or array-like to be limited.
```
petebacondarwin: So the param type should be `{Array|string|number|object}` right?

gkalpak: In `orderBy`, we use `ArrayLike` (which is totally made-up, but pretty descriptive).

petebacondarwin: OK. That works for me - does `orderBy` convert numbers to strings too?

gkalpak: `orderBy` only accepts `{Array|ArrayLike}` (and calls `Array.prototype.map` on its input).

gkalpak:
I just realized, @googlebot didn't bother verifying the CLA fos us :unamused:
Can we trigger it manually ?

petebacondarwin:
Not sure but @wesleycho is definitely a yes already :-)

gkalpak:
I'll add the label manually, because my PR-merging script doesn't proceed otherwise :)

PR #14695, opened 2016-05-31
feat($route): implement `resolveRedirectTo`

### LINKED ISSUE #5150: Make $routeProvider redirectTo injectable
I would like to add dependency injection support for the ```redirectTo``` function, but I can't see a way to implement this without breaking the existing API.  

I frequently need services available when determining where to redirect, and as a workaround, I end up defining routes with empty templates purely for handling redirect logic, and calling ```$location.path()``` from within a resolve function on that route.  It seems messy and would be much nicer to just keep this logic contained inside ```redirectTo```.

Would this be welcome as a PR and if so, any suggestions on implementation so as to avoid breaking existing code that expects normal function parameters on this method?

gkalpak:
Feature

It is not possible to have a redirection gunction that is injectable and asynchronous (i.e. returns a promise).
See #5150.

`resolveRedirectTo` can be used to provide an injectable, possibly asynchronous redirection function.

Yes (see below).

**Other information**:
`resolveRedirectTo` can be set to a function that will (eventually) return the new URL to redirect
to. The function supports dependency injection and should return the new URL either as a string or
as a promise that will be resolved to a string.

If the `resolveRedirectTo` function returns `undefined` or returns a promise that gets resolved to
`undefined`, no redirection will take place and the current route will be processed normally.

If the `resolveRedirectTo` function throws an error or the returned promise gets rejected, no
further processing will take place (e.g. no template fetched, no `resolve` functions run, no
controller instantiated) and a `$routeChangeError` will be broadcasted.

`redirectTo` takes precedence over `resolveRedirectTo`, so specifying both on the same route
definition, will cause the latter to be ignored.

Fixes #5150

BREAKING CHANGE:

Previously, if `redirectTo` was a function that threw an Error, execution was aborted without firing
a `$routeChangeError` event.
Now, if a `redirectTo` function throws an Error, a `$routeChangeError` event will be fired.

### DIFF
--- a/test/ngRoute/routeSpec.js
+++ b/test/ngRoute/routeSpec.js
@@ -1045,7 +1045,7 @@ describe('$route', function() {
       });
 
 
-      it('should broadcast allow custom redirectTo function to be used', function() {
+      it('should broadcast `$routeChangeError` when redirectTo throws', function() {
         var error = new Error('Test');
 
         module(function($exceptionHandlerProvider, $routeProvider) {
@@ -1224,6 +1224,23 @@ describe('$route', function() {
       });
 
 
+      it('should support returning a promise', function() {
+        module(function($routeProvider) {
+          $routeProvider.
+            when('/foo', {resolveRedirectTo: function($q) { return $q.resolve('/bar'); }}).
+            when('/bar', {template: 'Bar'});
+        });
+
+        inject(function() {
+          $location.path('/foo');
+          $rootScope.$digest();
+
+          expect($location.path()).toBe('/bar');
+          expect($route.current.template).toBe('Bar');
+        });
+      });
+
+
       it('should support dependency injection', function() {
         module(function($provide, $routeProvider) {
           $provide.value('nextRoute', '/bar');
@@ -1265,23 +1282,6 @@ describe('$route', function() {
           $rootScope.$digest();
 
           expect($location.path()).toBe('/passed');
-        });
-      });
-
-
-      it('should support returning a promise', function() {
-        module(function($routeProvider) {
-          $routeProvider.
-            when('/foo', {resolveRedirectTo: function($q) { return $q.resolve('/bar'); }}).
-            when('/bar', {template: 'Bar'});
-        });
-
-        inject(function() {
-          $location.path('/foo');
-          $rootScope.$digest();
-
-          expect($location.path()).toBe('/bar');
-          expect($route.current.template).toBe('Bar');
         });
       });
 
@@ -1427,7 +1427,7 @@ describe('$route', function() {
       });
 
 
-      it('should ignore previous redirection if newer tansition happened', function() {
+      it('should ignore previous redirection if newer transition happened', function() {
         var spies = createSpies();
         var called = false;
         var deferred;
--- a/test/ngRoute/routeSpec.js
+++ b/test/ngRoute/routeSpec.js
@@ -901,223 +901,632 @@ describe('$route', function() {
 
 
   describe('redirection', function() {
-    it('should support redirection via redirectTo property by updating $location', function() {
-      module(function($routeProvider) {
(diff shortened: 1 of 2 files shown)


gkalpak:
We could avoid the BC, but the previous behavior feels like an oversight.

test/ngRoute/routeSpec.js
```diff
@@ -901,223 +901,632 @@ describe('$route', function() {
...
-        expect(path).toEqual('/foo/id3');
-        expect(search).toEqual({ subid: 'sid1', appended: 'true' });
-        return '/custom';
-      }
+      it('should support dependency injection', function() {
+        module(function($provide, $routeProvider) {
+          $provide.value('nextRoute', '/bar');
 
-      module(function($routeProvider) {
-        $routeProvider.when('/bar/:id/:subid/:subsubid', {templateUrl: 'bar.html'});
```
petebacondarwin: I assume that you removed this line because it is not used in the test, right?

test/ngRoute/routeSpec.js
```diff
@@ -901,223 +901,632 @@ describe('$route', function() {
...
+          expect($location.path()).toEqual('/custom');
+        });
       });
-    });
 
 
-    it('should properly process route params which are both eager and optional', function() {
-      module(function($routeProvider) {
-        $routeProvider.when('/foo/:param1*?/:param2', {templateUrl: 'foo.html'});
+      it('should broadcast allow custom redirectTo function to be used', function() {
```
petebacondarwin: This spec description doesn't make sense to me. 😱

test/ngRoute/routeSpec.js
```diff
@@ -901,223 +901,632 @@ describe('$route', function() {
...
+        });
       });
 
-      inject(function($location, $rootScope, $route) {
-        $location.path('/foo/id1/subid2/');
-        $rootScope.$digest();
 
-        expect($location.path()).toEqual('/foo/id1/subid2');
-        expect($route.current.templateUrl).toEqual('foo.html');
+      it('should redirect to the returned url', function() {
```
petebacondarwin: Isn't this really just a special case of the `should support returning a promise` test further down?
Perhaps it would be clearer to keep them together?

test/ngRoute/routeSpec.js
```diff
@@ -901,223 +901,632 @@ describe('$route', function() {
...
-        var element = $compile('<div><ng-view></ng-view></div>')($rootScope);
-        $location.path('/redirect/to/undefined');
-        $rootScope.$digest();
-        expect(controller).toHaveBeenCalled();
-        expect(templateFn).toHaveBeenCalled();
-        expect($location.path()).toEqual('/redirect/to/undefined');
-        dealoc(element);
+
+
+      it('should ignore previous redirection if newer tansition happened', function() {
```
petebacondarwin: typo `tansition`

petebacondarwin:
A few nit-picks but otherwise LGTM for 1.6

test/ngRoute/routeSpec.js
gkalpak: Me neither :scream:

test/ngRoute/routeSpec.js
gkalpak: Right!

test/ngRoute/routeSpec.js
gkalpak: The promise-returning test relies on DI being supported. The DI-being-supported test relies on this one. Tough decisions :stuck_out_tongue: 
I'll move the promise-returning next to this one :smiley:

test/ngRoute/routeSpec.js
gkalpak: Changed to:
> should broadcast `$routeChangeError` when redirectTo throws

gkalpak:
Thx for the review!
I'll wait for Travis and merge.

PR #14699, opened 2016-05-31
fix(ngRoute): Don't deep-copy the route object, just plain copy it instead.

### LINKED ISSUE #14478: $routeProvider deep-copies its route objects before using them, which is probably not necessary
Bug, I guess ?

$routeProvider deep-copies its route object in the beginning of the when('...', route) call, which is a heavy way of solving #8181 and #9731.

Handle routes without needing a deep-copy. 

I'm trying to plumb the Closure goog.html types in Angular as an alternative to the $sce types. These rely on a private marker that is passed to each instance by builders to ensure other classes can't fake them, and the deep copy breaks that link when passing an instance of that type in templateUrl.  

Recent ones ? That was initially fixed in https://github.com/angular/angular.js/commit/b4770582f84f26c8ff7f2320a36a6b0ceff6e6cc . 

I think this issue can be worked around by replacing templateUrl : safeTypeInstance with templateUrl : function(){return safeTypeInstance;}, but that's ugly. Martin Probst asked us to write an issue, saying "deep copying the entire object seems overkill if all you wanted was to (shallow) copy the fields in the angular.extend call.".

rjamet:
Bug fix for #14478

Angular runs angular.copy on its routes, which isn't necessary and breaks some things

No more angular.copy on these, which appears to not break anything.

No, it shouldn't. 

That's Angular's guts, the previous behavior wasn't documented, and the new one seems obvious enough. 

**Other information**:

This seems to fix #14478. No need for a deep copy, just a plain one is enough as ngRoute does not modify the fields in the route object. 

I'm not that familiar with ngRoute, and the fix seems too easy: if something's wrong, I'll be glad to add tests for that :) As mentionned in the issue, #8181 and #9731 seems to be relevant, but this PR shouldn't reopen the issues.

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -160,7 +160,7 @@ function $RouteProvider() {
    */
   this.when = function(path, route) {
     //copy original route object to preserve params inherited from proto chain
-    var routeCopy = angular.copy(route);
+    var routeCopy = route;
     if (angular.isUndefined(routeCopy.reloadOnSearch)) {
       routeCopy.reloadOnSearch = true;
     }
--- a/test/ngRoute/routeSpec.js
+++ b/test/ngRoute/routeSpec.js
@@ -1297,6 +1297,39 @@ describe('$route', function() {
       });
     });
 
+
+    it('should not deep-copy or alter route fields', function() {
+      // Object.seal is used to block future modifications of the object: we do not want
+      // to alter the objects passed as parameters.
+      var canaryObject = Object.seal({});
+
+       // No point faking a full template, use the syntax from above
+      var canaryTemplate = function() {
+        return "<p>hi</p>";
+      };
+      canaryTemplate.someField = canaryObject;
+      Object.seal(canaryTemplate);  // sealed as well, just in case
+
+      module(function($routeProvider) {
+        $routeProvider.when('/foo',
+            {
+              template: canaryTemplate,
+              myObject: canaryObject,
+            });
+      });
+
+      inject(function($route, $location, $rootScope) {
+        $location.path('/foo');
+        $rootScope.$digest();
+
+        // Copies of the canaries won't be their originals.
+        expect($route.current.myObject).toBe(canaryObject);
+        expect($route.current.template).toBe(canaryTemplate);
+        expect($route.current.template.someField).toBe(canaryObject);
+      });
+    });
+
+
     describe('reload', function() {
       var $location;
       var $log;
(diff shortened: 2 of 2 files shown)


gkalpak:
The question is why was the original `extend` used. We need to find out, to make sure this change isn't breaking anything.

rjamet:
The commit introducing the code is this: https://github.com/angular/angular.js/commit/b4770582f84f26c8ff7f2320a36a6b0ceff6e6cc, where they switched extending an object with route (without the inherited components of route) to extending routeCopy. Their test still pass, and my way will keep the original prototype chain of the route fields.

gkalpak:
I'm talking about the commit that introduced the previous implementation, using `extend()`. I haven't looked for it yet, though.

Narretz:
In it's current form, it seems to come from here: https://github.com/angular/angular.js/commit/15ecc6f3668885ebc5c7130dd34e00059ddf79ae 

Note the comment in the changed version that talks about merging routes

gkalpak:
It seems it has been using `.extend` since forever (70e401e), but for no apparent reason.
In any case, this is a breaking change - should be documented as such.

I agree that taking into account inherited properties is a more common usecase than extending a route, but ideally we should provide a way to support both.

src/ngRoute/route.js
```diff
@@ -160,7 +160,7 @@ function $RouteProvider() {
    */
   this.when = function(path, route) {
     //copy original route object to preserve params inherited from proto chain
-    var routeCopy = angular.copy(route);
+    var routeCopy = route;
```
IgorMinar: I believe that the intention of the copying the object was to avoid the possibility of the user mutating the passed in object after the router partially processed it. This could get the router into an unknown state.

By copying we know for sure that the state we read has not changed.

The problem this causes is that if the `templateUrl` is set to be a TrustedResourceUrl instance, the instance is not copied correctly and the `templateUrl` fails to be properly unwrapped by sce later.

Since both `template` and `templateUrl` properties are both (wrapped) immutable strings, we could use `shallowCopy` instead to protect the router from internal state corruption while not corrupting the sce wrappers.

Another option is to teach angular.copy how to copy sce wrappers safely, but given that there could be custom implementations of these wrappers (e.g. at google we have our own), making the the mechanism generic might be tricky.

`shallowCopy` seems to give us all of the benefits with little down-sides, so I'd prefer that one.

gkalpak: I think the intention of using `copy` instead of `extend` was to preserve inherited properties (talk about using a sledgehammer to crack a nut :smiley:). `shallowCopy` (which we need to copy over to `angular-route` btw) would also shallow-copy inherited properties, so we are covered.

Issue #12645, opened 2015-08-21
Document the options parameter of $animate functions


djdmbrwsk:
Unless I missed it, there is no documentation explaining the `options` parameter that is optional in a bunch of the `$animate` service's functions. Ex:

```js
$animate.enter(element, parent, [after], [options]);
```

I just spent a bit digging through the source to finally find you can pass in an object like this:

```js
{
  addClass: 'some-new-class',
  removeClass: ' some-old-class'
}
```

Narretz:
@djdmbrwsk Would you like to submit a PR which adds the info to the docs?

PR #14700, opened 2016-06-01
docs($animate): detail options object

wesleycho:
- Document options object used in various $animate methods

This should address #12645.

### DIFF
--- a/src/ng/animate.js
+++ b/src/ng/animate.js
@@ -420,7 +420,13 @@ var $AnimateProvider = ['$provide', function($provide) {
        * @param {DOMElement} parent the parent element which will append the element as
        *   a child (so long as the after element is not present)
        * @param {DOMElement=} after the sibling element after which the element will be appended
-       * @param {object=} options an optional collection of options/styles that will be applied to the element
+       * @param {object=} options an optional collection of options/styles that will be applied to the element.
+       *   The object can have the following properties:
+       *
+       *   - **addClass** - `{string}` - space-separated CSS classes to add to element
+       *   - **from** - `{Object}` - CSS properties & values at the beginning of animation. Must have matching `to`
+       *   - **removeClass** - `{string}` - space-separated CSS classes to remove from element
+       *   - **to** - `{Object}` - CSS properties & values at end of animation. Must have matching `from`
        *
        * @return {Promise} the animation callback promise
        */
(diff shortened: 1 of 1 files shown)


src/ng/animate.js
```diff
@@ -420,7 +420,14 @@ var $AnimateProvider = ['$provide', function($provide) {
...
        * @param {DOMElement=} after the sibling element after which the element will be appended
-       * @param {object=} options an optional collection of options/styles that will be applied to the element
+       * @param {object=} options an optional collection of options/styles that will be applied to the element.
+       *   The object has the following properties:
+       *
+       *   - **addClass** - `{string}` - space-separated CSS classes to add to element
+       *   - **from** - `{Object}` - CSS properties & values from beginning of animation
+       *   - **removeClass** - `{string}` - space-separated CSS classes to remove from element
+       *   - **to** - `{Object}` - CSS properties & values at end of animation
+       *
```
gkalpak: Could you also document the `options` argument of other methods (e.g. `move()`) ?

gkalpak:
This should indeed fix #12645. Feel free to add `Fixes #12645` in the commit message.

src/ng/animate.js
```diff
@@ -420,7 +420,13 @@ var $AnimateProvider = ['$provide', function($provide) {
        * @param {DOMElement} parent the parent element which will append the element as
        *   a child (so long as the after element is not present)
        * @param {DOMElement=} after the sibling element after which the element will be appended
-       * @param {object=} options an optional collection of options/styles that will be applied to the element
+       * @param {object=} options an optional collection of options/styles that will be applied to the element.
+       *   The object has the following properties:
+       *
+       *   - **addClass** - `{string}` - space-separated CSS classes to add to element
+       *   - **from** - `{Object}` - CSS properties & values from beginning of animation
```
Narretz: "from the beginning" -> "at the beginning"

src/ng/animate.js
```diff
@@ -420,7 +420,13 @@ var $AnimateProvider = ['$provide', function($provide) {
        * @param {DOMElement} parent the parent element which will append the element as
        *   a child (so long as the after element is not present)
        * @param {DOMElement=} after the sibling element after which the element will be appended
-       * @param {object=} options an optional collection of options/styles that will be applied to the element
+       * @param {object=} options an optional collection of options/styles that will be applied to the element.
+       *   The object has the following properties:
```
Narretz: It can have the following properties, but it doesn't have to. The only requirement is that "from" must have a matching "to"

PR #14705, opened 2016-06-02
fix(select): remove workaround for Chrome bug

Narretz:
Removes a no-longer need workaround

In the Chrome issue (https://bugs.chromium.org/p/chromium/issues/detail?id=381459)
I noticed that the bug can't be reproduced in Chrome 45 and the issue was
closed subsequently. I've also tested again with Chrome 51 and it still works.

Related: #6828

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -2,15 +2,6 @@
 
 var noopNgModelController = { $setViewValue: noop, $render: noop };
 
-function chromeHack(optionElement) {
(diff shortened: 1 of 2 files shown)


gkalpak:
LGTM as long ad we have a test for the expected behavior.

Narretz:
Difficult, as the bug doesn't present itself anymore and we changed ngOptions massively in the meantime. :<

Narretz:
Ok, not 'that* difficult :>

PR #14710, opened 2016-06-03
fix(ngSanitize): call attribute setter in linky for all links

### LINKED ISSUE #14707: attributes function parameter for linky only works for first link

Bug.

The attributes function is only called back for the first link.

As you can see in this plunk, the setAttributes function is only called back once, even though there are two links in the text.

http://plnkr.co/edit/d5zWDOcF46wZEQjdLUYf?p=preview

There should be 2 callbacks, one for each link.

So that each link can have different attributes depending on whether or not it's an external or internal link.

All browsers and the latest version of Angular (1.5.5 I believe).

It appears that the attributes function is overwritten by its return value in the first call of addLink which causes it not to be called back in further calls (since it's now an object).

Narretz:
bug fix

Fixes #14707

Fixes #14707

### DIFF
--- a/src/ngSanitize/filter/linky.js
+++ b/src/ngSanitize/filter/linky.js
@@ -141,6 +141,11 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
     if (text == null || text === '') return text;
     if (!isString(text)) throw linkyMinErr('notstring', 'Expected string but received: {0}', text);
 
+    var attributesFn =
+      angular.isFunction(attributes) ? attributes :
+      angular.isObject(attributes) ? function getAttributesObject() {return attributes;} :
+      function getEmptyAttributesObject() {return {};};
+
     var match;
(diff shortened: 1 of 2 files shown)


src/ngSanitize/filter/linky.js
```diff
@@ -141,6 +141,11 @@ angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
     if (text == null || text === '') return text;
     if (!isString(text)) throw linkyMinErr('notstring', 'Expected string but received: {0}', text);
 
+    var attributesFn =
+      isFunction(attributes) ? attributes :
+      isObject(attributes) ? function getAttributesObject() {return attributes} :
```
gkalpak: Shouldn't these be `angular.isObject` and `angular.isFunction` ?

gkalpak:
LGTM (except for the helper function `isXyz() --> angular.isXyz()`)

Issue #11406, opened 2015-03-23
Interpolation should call toString on objects


ewinslow:
The string value is what is going to be used in the rendering of the information, but this doesn't seem to be respected.

http://plnkr.co/edit/ZLmiavC9lSlKj1b6cQJk?p=preview

For a real-world use-case: We'd like to return goog.Uri instances from our controller and have angular appropriately format them in the template, but it doesn't work -- it seems like the uri instances are compared by identity instead of string value (even though only string value will affect what the user sees), which I think was causing infinite loops. In the plunkr above I cannot get the right information to show up without manually stringifying first.

Right now our workaround is to call `toString()` or `+ ""` manually within the controller, but this is just visual noise... it also is incentive to pass around strings instead of js objects, which I think would obscure the intent of the code.

wesleycho:
I don't think Angular should do that - you would get some weird behavior happening if it did that on objects.  For example, here is some of the undesired behavior one might get if this was supported:
<pre>
var foo = {};
foo.toString() // yields '[object Object]'
{} + '' // yields 0
</pre>
These two examples would be strong arguments against supporting this IMO.  I would rather Angular fail silently in those two cases.

For your case, you could always create a custom filter to format it as desired.

ewinslow:
Right now angular shows "{}" for the above example. That's not even weirder than calling toString()?

Current behavior is silly. Let's improve it. Throw a useful error, or call "toString", or concat with a string, but don't just show curlies or enter into an infinite loop...

ewinslow:
Also: `'' + x` gives a consistent result regardless of the type of x, so far as I can tell (i.e. calls `toString()`). The one rub is that for `null`, `false`, and `undefined` it gives "null", "false", and "undefined" respectively, when you might want that to be empty string instead. Those can be trivially special-cased if need be.

lgalfaso:
Hi, there is always a constant struggle between
* Failing silently, and developers spending hours trying to figure out why something is not working
* Failing every single time something is out of order, and end up very inflexible
* Being too flexible by picking a behavior, and not making everybody happy as this will not be the behavior some are expecting

In most cases, the answer is a balance. We needs to be flexible and at the same time let developers know when there is something that is not supported/will not work. On top, semver rules should be followed for breaking changes.

With this in mind, I find the behavior to be acceptable. Now if the consensus is that calling `toString`, then this cannot happen in a minor release, and 1.5 would be the first opportunity.

ewinslow:
[...]

Issue #7317, opened 2014-04-30
$interpolate : add a way to hook string representation


jpolo:
$interpolate has only two behavior for stringify : IF string THEN output without change OR  perform json conversion.
It would be helpful to add a detection for hookable toString behavior for objects that are always represented by string or number (= where json representation has no value at all).

Use case:  when using custom Int64, URI, ..., objects, everybody would expect that {{myUri}} would display a string representation instead of a json representation or a quoted string.

Idea of implementations : 
 - detect that valueOf or toString is overriden ?
 - add a callback in $interpolate module ?
 - apply toJson only for plain objects (i.e. where Object.getPrototypeOf(o) === Object.prototype) ?

PR #14715, opened 2016-06-05
fix($interpolate): use custom toString() function & fix(ngBind): use same json conversion as $interpolate

### LINKED ISSUE #11716: Inconsistency in ng-bind and interpolate
Here is example:
http://plnkr.co/edit/C95zazjXx5jfDrMqIE67?p=preview
Same data gives diffrerent output in interpolate and ng-bind
```
ng-bind: [object Object]
interpolate {"a":1,"b":2,"c":3}
```
I expected that `ng-bind` will render the same result that interpolate does.

Narretz:
bug fix

ngBind has different binding logic for objects than interpolation

Same logic as interpolate

Yes, the visible output is now different, i.e. consistent with interpolate. I don't think anyone relied on ngBind producing this output.
However, this PR is a step away from the proposal in #11406, which wants $interpolate to call toString() instead of toJson. See also PR #8350, and issue #7317

- [x] Docs have been updated (we should probably document how this works in the guide)

**Other information**:

Fixes #11716

BREAKING CHANGE:

`ngBind` now uses the same logic as $interpolate (i.e. {{myString}}) when
binding, which means objects are now transformed with JSON.stringify. Before, it
would use the object's toString() function.

The following example shows the different output:

``` js
$scope.myObject = {a: 1, b: 2};
```

``` html
<span ng-bind="myObject"></span>
```

Before:

``` html
<span ng-bind="myObject">[object Object]</span>
```

After:

``` html
<span ng-bind="myObject">{"a":1,"b":2}</span>
```

### DIFF
--- a/docs/content/guide/interpolation.ngdoc
+++ b/docs/content/guide/interpolation.ngdoc
@@ -26,6 +26,15 @@ normal {@link ng.$rootScope.Scope#$digest digest} cycle.
 
 Note that the interpolateDirective has a priority of 100 and sets up the watch in the preLink function.
 
+### How the string representation is computed
+
+If the interpolated value is not a `String`, it is computed as follows:
+- `undefined` and `null` are converted to `''`
+- if the value is an object that is not a `Number`, `Date` or `Array`, $interpolate looks for
+a custom `toString()` function on the object, and uses that. Custom means that
+`myObject.toString !== `Object.prototype.toString`.
+- if the above doesn't apply, `JSON.stringify` is used.
+
 ### Binding to boolean attributes
 
 Attributes such as `disabled` are called `boolean` attributes, because their presence means `true` and
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -86,6 +86,7 @@
   getBlockNodes: true,
   hasOwnProperty: true,
   createMap: true,
+  stringify: true,
 
   NODE_TYPE_ELEMENT: true,
   NODE_TYPE_ATTRIBUTE: true,
@@ -1901,6 +1902,27 @@ function getBlockNodes(nodes) {
  */
 function createMap() {
   return Object.create(null);
+}
+
+function stringify(value) {
+  if (value == null) { // null || undefined
+    return '';
+  }
+  switch (typeof value) {
+    case 'string':
+      break;
+    case 'number':
+      value = '' + value;
+      break;
+    default:
+      if (hasCustomToString(value) && !isArray(value) && !isDate(value)) {
+        value = value.toString();
+      } else {
+        value = toJson(value);
+      }
+  }
+
+  return value;
 }
 
 var NODE_TYPE_ELEMENT = 1;
--- a/src/ng/directive/ngBind.js
+++ b/src/ng/directive/ngBind.js
@@ -60,7 +60,7 @@ var ngBindDirective = ['$compile', function($compile) {
         $compile.$$addBindingInfo(element, attr.ngBind);
         element = element[0];
         scope.$watch(attr.ngBind, function ngBindWatchAction(value) {
-          element.textContent = isUndefined(value) ? '' : value;
+          element.textContent = stringify(value);
         });
       };
     }
--- a/src/ngMessageFormat/messageFormatCommon.js
+++ b/src/ngMessageFormat/messageFormatCommon.js
@@ -8,15 +8,7 @@
 /* global isFunction: false */
 /* global noop: false */
 /* global toJson: false */
-
-function stringify(value) {
-  if (value == null /* null/undefined */) { return ''; }
-  switch (typeof value) {
-    case 'string':     return value;
-    case 'number':     return '' + value;
-    default:           return toJson(value);
-  }
-}
+/* global $$stringify: false */
 
 // Convert an index into the string into line/column for use in error messages
 // As such, this doesn't have to be efficient.
--- a/test/ng/directive/ngBindSpec.js
+++ b/test/ng/directive/ngBindSpec.js
@@ -46,6 +46,43 @@ describe('ngBind*', function() {
       expect(element.text()).toEqual('-0false');
     }));
 
+    they('should jsonify $prop', [[{a: 1}, '{"a":1}'], [true, 'true'], [false, 'false']], function(prop) {
+      inject(function($rootScope, $compile) {
+        $rootScope.value = prop[0];
+        element = $compile('<div ng-bind="value"></div>')($rootScope);
+        $rootScope.$digest();
+        expect(element.text()).toEqual(prop[1]);
+      });
+    });
+
+    it('should use custom toString when present', inject(function($rootScope, $compile) {
+      $rootScope.value = {
+        toString: function() {
+          return 'foo';
+        }
+      };
+      element = $compile('<div ng-bind="value"></div>')($rootScope);
+      $rootScope.$digest();
+      expect(element.text()).toEqual('foo');
+    }));
+
+    it('should NOT use toString on array objects', inject(function($rootScope, $compile) {
+      $rootScope.value = [];
+      element = $compile('<div ng-bind="value"></div>')($rootScope);
+      $rootScope.$digest();
+      expect(element.text()).toEqual('[]');
+    }));
+
+
+    it('should NOT use toString on Date objects', inject(function($rootScope, $compile) {
+      $rootScope.value = new Date(2014, 10, 10, 0, 0, 0);
+      element = $compile('<div ng-bind="value"></div>')($rootScope);
+      $rootScope.$digest();
+      expect(element.text()).toBe(JSON.stringify($rootScope.value));
+      expect(element.text()).not.toEqual($rootScope.value.toString());
+    }));
+
+
     it('should one-time bind if the expression starts with two colons', inject(function($rootScope, $compile) {
       element = $compile('<div ng-bind="::a"></div>')($rootScope);
       $rootScope.a = 'lucas';
--- a/test/ng/interpolateSpec.js
+++ b/test/ng/interpolateSpec.js
@@ -35,6 +35,29 @@ describe('$interpolate', function() {
     expect($interpolate('{{ false }}')({})).toEqual('false');
   }));
 
+  it('should use custom toString when present', inject(function($interpolate, $rootScope) {
+    var context = {
+      a: {
+        toString: function() {
+          return 'foo';
+        }
+      }
+    };
+
+    expect($interpolate('{{ a }}')(context)).toEqual('foo');
+  }));
+
+  it('should NOT use toString on array objects', inject(function($interpolate) {
+    expect($interpolate('{{a}}')({ a: [] })).toEqual('[]');
+  }));
+
+
+  it('should NOT use toString on Date objects', inject(function($interpolate) {
+    var date = new Date(2014, 10, 10);
+    expect($interpolate('{{a}}')({ a: date })).toBe(JSON.stringify(date));
+    expect($interpolate('{{a}}')({ a: date })).not.toEqual(date.toString());
+  }));
+
 
   it('should return interpolation function', inject(function($interpolate, $rootScope) {
     var interpolateFn = $interpolate('Hello {{name}}!');
--- a/test/ngMessageFormat/messageFormatSpec.js
+++ b/test/ngMessageFormat/messageFormatSpec.js
@@ -311,6 +311,30 @@ describe('$$ngMessageFormat', function() {
     }));
 
 
+    it('should use custom toString when present', inject(function($interpolate, $rootScope) {
+       var context = {
+        a: {
+          toString: function() {
+            return 'foo';
+          }
+        }
+      };
+
+      expect($interpolate('{{ a }}')(context)).toEqual('foo');
+    }));
+
+    it('should NOT use toString on array objects', inject(function($interpolate) {
+      expect($interpolate('{{a}}')({ a: [] })).toEqual('[]');
+    }));
+
+
+    it('should NOT use toString on Date objects', inject(function($interpolate) {
+      var date = new Date(2014, 10, 10);
+      expect($interpolate('{{a}}')({ a: date })).toBe(JSON.stringify(date));
+      expect($interpolate('{{a}}')({ a: date })).not.toEqual(date.toString());
+    }));
+
+
     it('should return interpolation function', inject(function($interpolate, $rootScope) {
       var interpolateFn = $interpolate('Hello {{name}}!');
 
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -96,6 +96,7 @@
     "createMap": false,
     "VALIDITY_STATE_PROPERTY": false,
     "reloadWithDebugInfo": false,
+    "stringify": false,
 
     "NODE_TYPE_ELEMENT": false,
     "NODE_TYPE_ATTRIBUTE": false,
--- a/src/AngularPublic.js
+++ b/src/AngularPublic.js
@@ -156,7 +156,8 @@ function publishExternalAPI(angular) {
     '$$minErr': minErr,
     '$$csp': csp,
     '$$encodeUriSegment': encodeUriSegment,
-    '$$encodeUriQuery': encodeUriQuery
+    '$$encodeUriQuery': encodeUriQuery,
+    '$$stringify': stringify
   });
 
   angularModule = setupModuleLoader(window);
--- a/src/ng/interpolate.js
+++ b/src/ng/interpolate.js
@@ -111,23 +111,6 @@ function $InterpolateProvider() {
         replace(escapedEndRegexp, endSymbol);
     }
 
-    function stringify(value) {
-      if (value == null) { // null || undefined
-        return '';
-      }
-      switch (typeof value) {
-        case 'string':
-          break;
-        case 'number':
-          value = '' + value;
-          break;
-        default:
-          value = toJson(value);
-      }
-
-      return value;
-    }
-
     //TODO: this is the same as the constantWatchDelegate in parse.js
     function constantWatchDelegate(scope, listener, objectEquality, constantInterp) {
       var unwatch;
--- a/src/ngMessageFormat/messageFormatService.js
+++ b/src/ngMessageFormat/messageFormatService.js
@@ -10,7 +10,6 @@
 /* global noop: true */
 /* global toJson: true */
 /* global MessageFormatParser: false */
-/* global stringify: false */
 
 /**
  * @ngdoc module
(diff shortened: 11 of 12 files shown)


src/ng/directive/ngBind.js
```diff
@@ -60,7 +60,7 @@ var ngBindDirective = ['$compile', function($compile) {
         $compile.$$addBindingInfo(element, attr.ngBind);
         element = element[0];
         scope.$watch(attr.ngBind, function ngBindWatchAction(value) {
-          element.textContent = isUndefined(value) ? '' : value;
+          element.textContent = isUndefined(value) ? '' : stringify(value);
```
campersau: This can be simplified to:
```js
element.textContent = stringify(value);
```

gkalpak:
I think it makes sense to make `ngBind` and `$interpolate` consistent in terms of object stringification, but moving from `toString` to `toJson` will be a step backwards for oobjects with custom `toString` methods.

I think a good compromise would be to use `toString` if it different than `Object.prototype.toString` and `toJson` otherwise (either only on `ngBind` or on `$interpolate` too - but that's a bigger breaking change).

petebacondarwin:
I agree with @gkalpak that we should use `toString` if a custom one is available. Given that this is a BC anyway. Let's change `$interpolate` to do this too and only merge to 1.6

Narretz:
I've updated the PR to include the commit that calls custom toString() on objects. PTAL.

I've noticed a problem with our own date input tests. In the original PR, Date objects als had toString() called when interpolating them. This change made our tests fail that allow min and max attributes of date inputs to be date objects: https://github.com/angular/angular.js/blob/f3377da6a748007c11fde090890ee58fae4cefa5/test/ng/directive/inputSpec.js#L1896. We expect a jsonified string like: `'"2014-11-09T23:00:00.000Z"'`. Date.toString() however produces a locale aware string that looks like this:  `Mon Nov 10 2014 00:00:00 GMT+0100 (Mitteleuropäische Zeit), `

As a fix, I've updated the createDateParser function to convert the passed string into an object if manual iso date parsing fails. This will produce a valid Date from the result of Date.toString(), and NaN from everything that can't be parsed (NaN is also returned if nothing works).

There's also the issue that ngMessageFormat duplicates the stringify function, and at the moment all our unit tests will use this function instead of the one defined in Angular.js. This is a bit weird, but at least it guarantees that you notice when you have made changes to the core implementation.

src/Angular.js
```diff
@@ -1929,6 +1930,28 @@ function createMap() {
...
+    return '';
+  }
+  switch (typeof value) {
+    case 'string':
+      break;
+    case 'number':
+      value = '' + value;
+      break;
+    default:
+      if (isFunction(value.toString) && value.toString !== Object.prototype.toString && !isArray(value)) {
```
gkalpak: I think we already have `toString` as an "IIFE global".

src/ng/directive/input.js
```diff
@@ -1290,6 +1290,9 @@ function createDateParser(regexp, mapping) {
           }
         });
         return new Date(map.yyyy, map.MM - 1, map.dd, map.HH, map.mm, map.ss || 0, map.sss * 1000 || 0);
+      } else {
+        // If we fail to parse the string with the ISO regex, it might be because it's the result of Date.toString()
+        return Date.parse(iso);
```
gkalpak: This can lead to unexpected results. E.g. `Date.parse('0')` is the same as `Date.parse('2000')` which is the same as `2000-01-01T00:00:00` in local time - which is probably not the intended value :smiley:

If this is only to facilitate our new `toString()` stringification?

src/ng/directive/input.js
```diff
@@ -1358,17 +1361,21 @@ function createDateInputType(type, regexp, parseDate, format) {
...
+      } else {
+        return isDefined(val) && !isDate(val) ? parseDate(val) || undefined : val;
+      }
     }
   };
 }
 
+function isValidDate(value) {
+  // Invalid Date: getTime() returns NaN
+  return value && !(value.getTime && value.getTime() !== value.getTime());
```
gkalpak: Unrelated to this PR: 
That's a weird validation function (e.g. `isValidDate({})` -- true).
I would expect `value && value.getTime && value.getTime() === value.getTime()`.

src/ng/directive/input.js
Narretz: Yes, this is only the make it work with Date.toString(). Maybe we can only use Date.parse() when the string has a minimum length of x?

src/Angular.js
Narretz: We do have a hasCustomToString fn

gkalpak: Even better :smiley:

test/ng/directive/ngBindSpec.js
```diff
@@ -46,6 +46,15 @@ describe('ngBind*', function() {
...
 
+    they('should jsonify $prop', [[{a: 1}, '{"a":1}'], [true, 'true'], [false, 'false']], function(prop) {
+      inject(function($rootScope, $compile) {
+        $rootScope.bind = prop[0];
+        element = $compile('<div ng-bind="bind"></div>')($rootScope);
+        $rootScope.$digest();
+        expect(element.text()).toEqual(prop[1]);
+      });
+    });
+
```
gkalpak: I would add a test with a non-default `toString` here too.

test/ng/interpolateSpec.js
```diff
@@ -35,6 +35,22 @@ describe('$interpolate', function() {
     expect($interpolate('{{ false }}')({})).toEqual('false');
   }));
 
+  it('should use custom toString when present', inject(function($interpolate, $rootScope) {
+     var context = {
```
gkalpak: Wrong indentation :stuck_out_tongue:

src/ng/directive/input.js
gkalpak: If I understand correctly, this change was made in order to fix the tests. I don't think it's the right fix.

We wanted to test that JSONified dates would be parsed correctly. In order to test that, we relied on the fact that `$interpolate` would convert date objects to JSON. Now that `$interpolate` does not convert them to JSON (but calls `toString()` instead), we should change the _tests_, so that JSONified dates are still passed to `max`/`min`.

With the current fix, we have a test that is supposed to test parsing JSONified dates, but it tests stringified dates instead, right ?

--
Part of the big BC will be that `max`/`min` won't be able to handle interpolated date objects. One will need to either use `ngMax/Min` or change `{{ myDateObj }}` to `{{ myDateObj.toISOString() }}` or `{{ myDateObj | json }}` (or manually JSONify the date object in the controller).
I don't think it's too bad, though.

Another idea would be to check if `new Date(iso).toString() === iso`. If this is true, then `iso` must be the `toString()` output of a Date equal to `new Date(iso)`.

--
BTW, `Date.parse()` returns the number of milliseconds. Don't we need a Date object to be returned ?

Narretz: I'm not 100% sure, but maybe you are on to something here.
The tests say "should parse ISO-based date strings as a valid min date value", but they specifically use Date objects. The commit that introduced that - https://github.com/angular/angular.js/commit/088545c1856ce1c3ec3416965dff65077a6e0523 - also said it closes #6755, but it looks to me that it should say Closes #6754, which is precisely about using Min/Max with Date objects: https://github.com/angular/angular.js/issues/6754
For #6755, there's another PR: https://github.com/angular/angular.js/commit/b3502835039178296b730b7526e5666b66ba9156

So I assume the intention was to allow JSONified Date objects; however this change was never documented.
I'd say that we should make it work and document it. I like your comparson approach: `new Date(iso).toString() === iso`

I assumed `Date.parse()` returns a Date object -_-. Yes, we probably want new Date().

petebacondarwin: I think that the 6754 and 6755 closes were just the wrong way round.

Changing the failing tests to use `toISOString()` prevents the hack in `input.js`.

The question is whether we need additional tests that specifically check that we can use real `Date` objects as well?

petebacondarwin: The following test passes, without the hack:

```
    it('should validate ng-min values that are Date objects correctly', function() {
      $rootScope.min = new Date(2013, 1, 1, 1, 2, 0);
      var inputElm = helper.compileInput('<input type="datetime-local" ng-model="value" name="alias" ng-min="min" />');

      helper.changeInputValueTo('2010-01-01T12:34:00');
      expect(inputElm).toBeInvalid();

      $rootScope.min = new Date(2014, 1, 1, 1, 2, 0);
      $rootScope.$digest();

      expect(inputElm).toBeInvalid();

      $rootScope.min = new Date(2009, 1, 1, 1, 2, 0);
      $rootScope.$digest();

      expect(inputElm).toBeValid();
    });
```

petebacondarwin:
Perhaps we should move `stringify` into its own file, which is included in both `angular.js` and `angular-message-format.js`? That way we have only one version in the source code (and for testing).

petebacondarwin:
Actually there already IS a `stringify.js` file that is shared with the `angular-loader.js` file. So we should make use of that, right?

Narretz:
Yes, but we need to share between ng and ngMessageFormat and it looks like ngMessageFormat only needs the stringify and not the other functions

src/ng/directive/input.js
Narretz: #6754 was specifically about `min` and `max` using date objects, not the `ng-` variants. Because you are missing out on native validation if you use the `ng` variants. So I don't think the test ensures that we support what the commit claimed to support.

src/ng/directive/ngBind.js
petebacondarwin: Can you include this change too @Narretz ?

petebacondarwin:
One nitpic but otherwise LGTM

Narretz:
What's that nitpick @petebacondarwin ?

Narretz:
PR has been updated, tests pass, PTAL

test/ng/directive/ngBindSpec.js
```diff
@@ -78,7 +78,7 @@ describe('ngBind*', function() {
       $rootScope.bind = new Date(2014, 10, 10, 0, 0, 0);
       element = $compile('<div ng-bind="bind"></div>')($rootScope);
       $rootScope.$digest();
-      expect(element.text()).toEqual('"2014-11-09T23:00:00.000Z"');
+      expect(element.text()).toBe('"' + $rootScope.bind.toJSON() + '"');
```
petebacondarwin: Unrelated to this change but, is it me, or does it seem weird to use a property called `bind` in this test?

test/ng/interpolateSpec.js
```diff
@@ -53,7 +53,8 @@ describe('$interpolate', function() {
 
 
   it('should NOT use toString on Date objects', inject(function($interpolate) {
-    expect($interpolate('{{a}}')({ a: new Date(2014, 10, 10, 0, 0, 0) })).toEqual('"2014-11-09T23:00:00.000Z"');
+    var date = new Date(2014, 10, 10);
+    expect($interpolate('{{a}}')({ a: date })).toBe('"' + date.toJSON() + '"');
```
petebacondarwin: Perhaps, keeping more inline with the test description we should also have

```
expect($interpolate('{{a}}')({ a: date })).not.toEqual('"' + date.toString() + '"');
```

test/ngMessageFormat/messageFormatSpec.js
```diff
@@ -329,7 +329,8 @@ describe('$$ngMessageFormat', function() {
 
 
     it('should NOT use toString on Date objects', inject(function($interpolate) {
-      expect($interpolate('{{a}}')({ a: new Date(2014, 10, 10, 0, 0, 0) })).toEqual('"2014-11-09T23:00:00.000Z"');
+      var date = new Date(2014, 10, 10);
+      expect($interpolate('{{a}}')({ a: date })).toBe('"' + date.toJSON() + '"');
```
petebacondarwin: Similarly here?

petebacondarwin:
The nitpick was https://github.com/angular/angular.js/pull/14715#discussion_r68054276 but it looks like it is dealt with.

LGTM

petebacondarwin:
Now that this is not a breaking change, right? Can we backport it to 1.5 too?

Narretz:
I think respecting a custom toString fn could be a breaking change for some.

I've also marked the ngBind change as breaking, but I guess this is more of a fix that now has the same consequences as using toString with interpolate.

docs/content/guide/interpolation.ngdoc
```diff
@@ -26,6 +26,15 @@ normal {@link ng.$rootScope.Scope#$digest digest} cycle.
 
 Note that the interpolateDirective has a priority of 100 and sets up the watch in the preLink function.
 
+### How the string representation is computed
+
+If something other than a String is interpolated, the string is computed as follows:
+- If the object is not a `Number`, `Date` or `Array`, $interpolate looks for a custom `toString()`
+function on the object, and uses that. Custom means that
+`myObject.toString() !== `Object.prototype.toString()`.
```
gkalpak: This is not accurate and may confuse people. You mean `myObject.toString !== Object.prototype.toString`.

docs/content/guide/interpolation.ngdoc
```diff
@@ -26,6 +26,15 @@ normal {@link ng.$rootScope.Scope#$digest digest} cycle.
...
+
+- If the above doesn't apply, then `JSON.stringify` is used.
```
gkalpak: There is no mention about `null`/`undefined` being converted to the empty string.

src/Angular.js
```diff
@@ -1903,6 +1904,27 @@ function createMap() {
...
+    return '';
+  }
+  switch (typeof value) {
+    case 'string':
+      break;
+    case 'number':
+      value = '' + value;
+      break;
+    default:
+      if (hasCustomToString(value) && !isArray(value) && !isDate(value)) {
```
gkalpak: I am not sure I am a big fan of not `toString`ifying Dates.

src/ngMessageFormat/messageFormatCommon.js
```diff
@@ -8,15 +8,7 @@
 /* global isFunction: false */
 /* global noop: false */
 /* global toJson: false */
```
gkalpak: I think this is not necessary any more (but it may have gone away in master anyway).

test/ng/directive/ngBindSpec.js
```diff
@@ -46,6 +46,42 @@ describe('ngBind*', function() {
...
+      $rootScope.$digest();
+      expect(element.text()).toEqual('[]');
+    }));
+
+
+    it('should NOT use toString on Date objects', inject(function($rootScope, $compile) {
+      $rootScope.bind = new Date(2014, 10, 10, 0, 0, 0);
+      element = $compile('<div ng-bind="bind"></div>')($rootScope);
+      $rootScope.$digest();
+      expect(element.text()).toBe('"' + $rootScope.bind.toJSON() + '"');
```
gkalpak: Why not `JSON.stringify($rootScope.bind)`?

test/ng/interpolateSpec.js
```diff
@@ -35,6 +35,28 @@ describe('$interpolate', function() {
...
+  }));
+
+  it('should NOT use toString on array objects', inject(function($interpolate) {
+    expect($interpolate('{{a}}')({ a: [] })).toEqual('[]');
+  }));
+
+
+  it('should NOT use toString on Date objects', inject(function($interpolate) {
+    var date = new Date(2014, 10, 10);
+    expect($interpolate('{{a}}')({ a: date })).toBe('"' + date.toJSON() + '"');
```
gkalpak: Same as above.

gkalpak:
Both changes (to `ngBind` and `$interpolate`) are breaking changes imo.
Currently, the `ngBind` is a much bigger change, since many people might be affected by calling `toJson` on Date objects instead of it's `toString` method.

(TBH, I think it is marginally better to call `toString` on Dates (instead of `toJson`), although that would be a significant change for `$interpolate`.)

docs/content/guide/interpolation.ngdoc
Narretz: Thanks that's right

src/Angular.js
Narretz: They were jsonified before and that output is more universal than toString, which produces an American date format while jsonified is ISO.

gkalpak: They were JSONified in `$interpolate`, but `toString`ified in `ngBind`. Since we are making the two consistent, we need to break one of them.

But you convicned me: a universal format is better than forcing American English for all users.
(For some reason, I was under the impression `toString()` returned a localized string...)

Narretz: I thought that too (localized string) and might have written that somewhere.

Narretz:
Updated again, should be good to go now

docs/content/guide/interpolation.ngdoc
```diff
@@ -26,6 +26,14 @@ normal {@link ng.$rootScope.Scope#$digest digest} cycle.
 
 Note that the interpolateDirective has a priority of 100 and sets up the watch in the preLink function.
 
+### How the string representation is computed
+
+If the interpolated value is not a `String`, it is computed as follows:
+- `undefined` and `null` are converted to `''`
+- if the value is an object that is not a `Number`, `Date` or `Array`, $interpolate looks for
+a custom `toString()` function on the object, and uses that. Custom means that
+`myObject.toString !== `Object.prototype.toString`. If that doesn't apply, `JSON.stringify` is used.
```
gkalpak: > If that doesn't apply, `JSON.stringify` is used.

Nitpick: I would place that on it's own bullet.

gkalpak:
LGTM

BTW, in the commit message for the `ngBind` change, the proposed migration instructions (using a `toString` filter) is not 100% correct, as it will not handle `undefined`/`null` values. Besides, I believe a better/easier migration would be using `value.toString()` (which takes advantage of Angular's "forgiving-ness" for `undefined`/`null` expressions.

Narretz:
@gkalpak Okay, so ngBind previously set `undefined` to empty string, and `null` was handled by setting textContent, which also sets it to '' (I think). So in the migration note I can special case both of them.

gkalpak:
You jave to special case them if you propose using a filter. But I think a filter is not necessary in this case (you can just call `.toString` (that's all the filter does anyway)).

shmdhussain:
What is the status of this PR, I believe it is merged into master branch... but I am getting he inconsistent results when interpolating the object, sometime getting full object some time getting `[object Object]`

for example

```
{{ {id:123}  }} //sometime outputs '{id:123}'
{{ {id:123}  }} //sometime outputs [object Object]
```

gkalpak:
This is merged and included in all version since 1.6.0-rc.0 onwards.
Even before this PR, it shouldn't cause inconsistent behavior. Your problem is caused by something else.

shmdhussain:
See this screenshot, sometimes I am getting the stringified version of Object sometime getting `[object Object]`
Screenshot 2020-02-24 at 17 59 50

PR #14717, opened 2016-06-06
Docs componentrouter install & deprecation

Narretz:


docs/content/guide/component-router.ngdoc
```diff
@@ -5,6 +5,11 @@
 
 # Component Router
 
+<div class="alert alert-danger">
+**Deprecation Notice:** The Component Router (ngComponentRouter module) has been deprecated and will not receive further updates.
```
petebacondarwin: Perhaps we should be saying... 

> This implementation of the component router has been deprecated ... 

So that it doesn't make it sound like there will never be a component router

micaelmbagira:
Could you explain this one ? Is `ngComponentRouter` merged with `ngRoute` now ?

Narretz:
@micaelmbagira No that means the component router won't be developed any further, not in ng1 and not in ng2. The whole specification wasn't sufficient for the framework, and unfortunately we noticed this rather late. So ng2 is getting a new router, which will maybe ported to ng1.

petebacondarwin:
To be fair the "new" router that @Narretz talk about will still be called the Component Router. It is just that it will have some significant API changes and be a complete reimplementation so the current guide is not much use right now.

nightcrawler91:
@Narretz  May I suggest you add some of the context you just talked about to the docs as well? I was very surprised when I saw this being deprecated with no apparent explanation until I read this, and I think others would benefit from that as well.

ulitiy:
And the concept was much cleaner than router/ui-router so it's really disappointing to see it's being deprecated. Have started a couple of projects for my clients using this new promising technology  that was deprecated in no time.

petebacondarwin:
The concept is not being deprecated :-) just this particular implementation.
There is a fancy new router in the works - but it is being developed for Angular 2 primarily and we need to look into how we can backport it to Angular 1.

petebacondarwin:
See http://angularjs.blogspot.co.uk/2016/06/improvements-coming-for-routing-in.html for info about the "new" version

petebacondarwin:
We need to look at how easy it will be to port this to Angular 1

PR #14718, opened 2016-06-06
refactor(ngMock): extract browserTrigger from ngScenario

Narretz:


petebacondarwin:
We need to fix the jshint failures. Also, should this only go in master or 1.5.x as well?

Narretz:
I'd say master only. If someone uses ngScenario with browserTrigger in 1.5, and and doesn't use ngMock at the same time, it would be a BC.

petebacondarwin:
I agree we should remove ngScenario in 1.6. It has been deprecated for some time and I even tried to get some people who claim to use it to move it to its own repo to no avail.

PR #14723, opened 2016-06-06
docs(guide/Component Router): improving comms

StephenFluin:
docs update


Narretz:
I agree that the current wording isn't very optimistic, but we shouldn't get people's hopes up about backporting the new router before a decision is made on this.

petebacondarwin:
Landed. Thanks @PeEllAvaj

Issue #14637, opened 2016-05-19
$resource: enumerable methods properties and `toJSON`


jrvidal:
**Do you want to request a *feature* or report a *bug*?**
Feature.

**What is the current behavior?**
If I want to apply some transformations to a `Resource` instance, I could encounter some nasty surprises:

```javascript
Cat.get({id: catId}).$promise.then(cat => {
  this.catDetails = doSomethingWithPairs(_.pairs(cat));

  this.otherCatDetails = doSomethingElse(_.pairs(_.omit(cat, ["name", "$promise", "$resolved"])));
});
```

`doSomethingWithPairs` receives an array of pairs that contains `["$promise", /*...*/]` and `["$resolved", /*...*/]`. `doSomethingElse` contains entries for all instance methods (!) such as `$get`, `$query`, etc.

**What is the expected behavior?**

This is a weird interaction with Underscore defaults<a name="lodash-back"></a>[<sup>[1]</sup>](#lodash) that is not for Angular to fix. However, I do think that some things that could be improved:

* Make `$promise` and `$resolved` non-enumerable properties of `Resource` instances.
* Make all `Resource` instance methods non-enumerable (a la ES6 classes).
* Document the behavior of `Resource.prototype.toJSON` method.

The latter is the workaround I have used to get a "normal" object from a `Resource` instance, although I would feel more confident if it were documented.

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
Related issues:
- #14460
- #6266

----

<a name="lodash"></a>[[1]](#lodash-back) Lodash only includes _own_, _enumerable_ properties for `omit`.

gkalpak:
:+1: for documenting `Resource.prototype.toJSON`. PRs welcome :wink:

I would also consider making Angular-stuff non-enumerable in 1.6/1.7 (since it's a breaking change), as long as all supported browsers behave consistently.
If anyone wants to take a crack at it, please do !

PR #14725, opened 2016-06-07
docs($resource): document Resource#toJSON

jrvidal:
Document the `toJSON` method of Resource instances as part of its public API.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -265,6 +265,14 @@ function shallowClearAndCopy(src, dst) {
  *   - `$cancelRequest`: If there is a cancellable, pending request related to the instance or
  *      collection, calling this method will abort the request.
  *
+ *   The Resource instances have these additional methods:
+ *
+ *   - `toJSON`: It returns a simple object without any of the extra properties added as part of
+ *     the Resource API. This object can be serialized through {@link angular.toJson} safely
+ *     without attaching Angular-specific fields. Notice that `JSON.stringify` (and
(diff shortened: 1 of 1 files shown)


gkalpak:
Thx for submit this! Could you expand a little bit on when `toJSON` is used and what it does? Maybe add a link to [MDN][1].

jrvidal:
> Could you expand a little bit on when toJSON is used and what it does? Maybe add a link to MDN.

Will do 👍 

I have just realized that `toJSON` [does not remove](https://github.com/jrvidal/angular.js/blob/df23708b9385062562168a87a12767a627da1c0e/src/ngResource/resource.js#L656-L661) `$cancelRequest`, so strictly speaking, it doesn't return a "simple object without the extra methods". Do you think it is pertinent to modify `toJSON` to do that, and if so, is it OK to do it in this PR?

gkalpak:
`toJSON` is not a general purpose function - it's sole purpose is facilitating proper conversion to JSON. As such, it makes sense to do the least amount of work necessary. Functions are not serialized to JSON, so `$cancelRequest` is not deleted on purpose.

jrvidal:
Gotcha, I will update the docs accordingly.

jrvidal:
Notice however how this behavior means that the workaround I mentioned in #14637, where I tried to use a `toJSON` call as a way of getting "just the data", is not valid. In any case, documenting this method improves the situation so at least we know where we stand.

jrvidal:
I rebased my commit with a better description.

PR #14726, opened 2016-06-07
feat(ngMessagesInclude): don't break on empty (or whitespace-only) templates

### LINKED ISSUE #12941: ng-messages-include raises "Error: [jqLite:nosel]" when including empty file
As shown in http://plnkr.co/edit/VFrt4rXZaK194zwJg0P2?p=preview, the ng-messages-include directive attempts to call `angular.element()` with the content of the file (empty or spaces) and, because jqLite interprets every string that does not start with "<" as a selector, it will raise an error. 

The error will be raised if the included file is not empty but starts with a text. I am not sure if angular-messages should handle the case where the included file contains text outside a `<div ng-message="..."></div>`, but I believe it should handle empty files gracefully.

gkalpak:
Feature (or bug fix - depends)

Using `ngMessagesInclude` with an empty template, throws an error. This happens, because when trying to `$compile` the template, jqLite thinks we are trying to call it with a selector.
See #12941.

Using `ngMessagesInclude` with an empty template (or a template that contains only whitespace), will not throw.

Only if you relied on `ngMessagesInclude` throwing an error on empty templates.

**Other information**:
Fixes #12941.

### DIFF
--- a/src/ngMessages/messages.js
+++ b/src/ngMessages/messages.js
@@ -264,377 +264,377 @@ var jqLite = angular.element;
  */
 angular.module('ngMessages', [])
 
-   /**
-    * @ngdoc directive
-    * @module ngMessages
-    * @name ngMessages
-    * @restrict AE
-    *
-    * @description
-    * `ngMessages` is a directive that is designed to show and hide messages based on the state
-    * of a key/value object that it listens on. The directive itself complements error message
-    * reporting with the `ngModel` $error object (which stores a key/value state of validation errors).
-    *
-    * `ngMessages` manages the state of internal messages within its container element. The internal
-    * messages use the `ngMessage` directive and will be inserted/removed from the page depending
-    * on if they're present within the key/value object. By default, only one message will be displayed
-    * at a time and this depends on the prioritization of the messages within the template. (This can
-    * be changed by using the `ng-messages-multiple` or `multiple` attribute on the directive container.)
-    *
-    * A remote template can also be used to promote message reusability and messages can also be
-    * overridden.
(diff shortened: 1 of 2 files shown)


src/ngMessages/messages.js
```diff
@@ -552,22 +552,32 @@ angular.module('ngMessages', [])
...
-            var comment = $compile.$$createComment ?
-                $compile.$$createComment('ngMessagesInclude', src) :
-                $document[0].createComment(' ngMessagesInclude: ' + src + ' ');
-            var anchor = jqLite(comment);
-            element.after(anchor);
-
-            // we don't want to pollute the DOM anymore by keeping an empty directive element
-            element.remove();
-          });
+          if (isString(html) && !html.trim()) {
```
Narretz: trim is doing an isString check, so we should be able to skip it here

Narretz:
LGTM

src/ngMessages/messages.js
petebacondarwin: The `isString` is needed

petebacondarwin:
LGTM

src/ngMessages/messages.js
Narretz: Test passes without it ...

gkalpak: The `isString()` is needed in case the request returns some non-string data. This is normally not the case, but can happen (especially in tests).

Narretz: Yes that's why I was confused. If it's not tested it's not really specified. But in that case it's not big deal

PR #14737, opened 2016-06-08
fix($compile): don't add replaced attributes twice to $attrs

### LINKED ISSUE #8159: Directive with 'replace: true' and template duplicates classes in the attrs parameter of the link fn
When I create a directive with a template that has a class attribute and `replace: true`. The attrs parameter in the link function duplicates the classes. They are also duplicated when I look at the element in the HTML. 

This appears to only happen when the tag that the directive is replacing also has a class attribute.

Here is some example code, along with the output I see in the console.

```html
<test class="btn-block"></test>
```

```javascript
directive("test", function () {
        return {
            restrict: 'E',
            replace: 'true',
            link: function(scope, element, attrs) {
                console.log(attrs);
            },
            template: '<div class="center-block"></div>'
        };
    })
```

screen shot 2014-07-11 at 3 57 28 pm

Narretz:
bugfix

In replace directives, attribute values from the template are added twice to the replaced element when both the replaced element and the template element have the attribute. This does not affect the DOM, as it normalizes duplicate values. The values are however duplicated in the $attrs object that is available to directive link functions.

No

- [-] Docs have been added / updated (for bug fixes / features)

**Other information**:
@gkalpak I fixed this while working on another attribute-related PR.

Fixes #8159
Closes #5597

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2761,19 +2761,19 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       // copy the new attributes on the old attrs object
       forEach(src, function(value, key) {
-        // Check if we already set this attribute in the loop above
-        if (!dst[key]) {
+        // Check if we already set this attribute in the loop above.
+        // `dst` will never contain hasOwnProperty as DOM parser won't let it.
+        // You will get an "InvalidCharacterError: DOM Exception 5" error if you
+        // have an attribute like "has-own-property" or "data-has-own-property", etc.
+        if (!dst.hasOwnProperty(key)) {
 
           if (key === 'class') {
             safeAddClass($element, value);
-            dst['class'] = (dst['class']   ? dst['class'] + ' ' : '') + value;
+            dst['class'] = value;
           } else if (key === 'style') {
             $element.attr('style', $element.attr('style') + ';' + value);
-            dst['style'] = (dst['style'] ? dst['style'] + ';' : '') + value;
-            // `dst` will never contain hasOwnProperty as DOM parser won't let it.
-            // You will get an "InvalidCharacterError: DOM Exception 5" error if you
-            // have an attribute like "has-own-property" or "data-has-own-property", etc.
-          } else if (key.charAt(0) !== '$' && !dst.hasOwnProperty(key)) {
+            dst['style'] = value;
+          } else if (key.charAt(0) !== '$') {
             dst[key] = value;
             dstAttr[key] = srcAttr[key];
           }
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2765,18 +2765,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         // `dst` will never contain hasOwnProperty as DOM parser won't let it.
         // You will get an "InvalidCharacterError: DOM Exception 5" error if you
         // have an attribute like "has-own-property" or "data-has-own-property", etc.
-        if (!dst.hasOwnProperty(key)) {
-
-          if (key === 'class') {
-            safeAddClass($element, value);
-            dst['class'] = value;
-          } else if (key === 'style') {
-            $element.attr('style', $element.attr('style') + ';' + value);
-            dst['style'] = value;
-          } else if (key.charAt(0) !== '$') {
-            dst[key] = value;
-            dstAttr[key] = srcAttr[key];
-          }
+        if (!dst.hasOwnProperty(key) && key.charAt(0) !== '$') {
+          dst.$set(key, value, true, srcAttr[key]);
         }
       });
     }
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2766,7 +2766,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         // You will get an "InvalidCharacterError: DOM Exception 5" error if you
         // have an attribute like "has-own-property" or "data-has-own-property", etc.
         if (!dst.hasOwnProperty(key) && key.charAt(0) !== '$') {
-          dst.$set(key, value, true, srcAttr[key]);
+          dst[key] = value;
+
+          if (key !== 'class' && key !== 'style') {
+            dstAttr[key] = srcAttr[key];
+          }
         }
       });
     }
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2761,18 +2761,22 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       // copy the new attributes on the old attrs object
       forEach(src, function(value, key) {
-        if (key === 'class') {
-          safeAddClass($element, value);
-          dst['class'] = (dst['class'] ? dst['class'] + ' ' : '') + value;
-        } else if (key === 'style') {
-          $element.attr('style', $element.attr('style') + ';' + value);
-          dst['style'] = (dst['style'] ? dst['style'] + ';' : '') + value;
-          // `dst` will never contain hasOwnProperty as DOM parser won't let it.
-          // You will get an "InvalidCharacterError: DOM Exception 5" error if you
-          // have an attribute like "has-own-property" or "data-has-own-property", etc.
-        } else if (key.charAt(0) !== '$' && !dst.hasOwnProperty(key)) {
-          dst[key] = value;
-          dstAttr[key] = srcAttr[key];
+        // Check if we already set this attribute in the loop above
+        if (!dst[key]) {
+
+          if (key === 'class') {
+            safeAddClass($element, value);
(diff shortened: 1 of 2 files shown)


test/ng/compileSpec.js
```diff
@@ -980,6 +980,29 @@ describe('$compile', function() {
...
+            directive('logAttrs', function() {
+              return {
+                link: function($scope, $element, $attrs) {
+                  attrs = $attrs;
+                }
+              }
+            })
+          });
+
+          inject(function($compile, $rootScope) {
```
gkalpak: Aren't `$compile` and `$rootScope` already available inside the top-level `describe's scope ?
Why is everyone injecting it ? :stuck_out_tongue:

src/ng/compile.js
```diff
@@ -2761,18 +2761,22 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-          // have an attribute like "has-own-property" or "data-has-own-property", etc.
-        } else if (key.charAt(0) !== '$' && !dst.hasOwnProperty(key)) {
-          dst[key] = value;
-          dstAttr[key] = srcAttr[key];
+        // Check if we already set this attribute in the loop above
+        if (!dst[key]) {
+
+          if (key === 'class') {
+            safeAddClass($element, value);
+            dst['class'] = (dst['class']   ? dst['class'] + ' ' : '') + value;
```
gkalpak: We have already checked that `dst[key]` is falsy, so no need for this check.

src/ng/compile.js
```diff
@@ -2761,18 +2761,22 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
+          } else if (key === 'style') {
+            $element.attr('style', $element.attr('style') + ';' + value);
+            dst['style'] = (dst['style'] ? dst['style'] + ';' : '') + value;
+            // `dst` will never contain hasOwnProperty as DOM parser won't let it.
+            // You will get an "InvalidCharacterError: DOM Exception 5" error if you
+            // have an attribute like "has-own-property" or "data-has-own-property", etc.
+          } else if (key.charAt(0) !== '$' && !dst.hasOwnProperty(key)) {
```
gkalpak: We could replace the `!dst[key]` check above with `!dst.hasOwnProperty(key)` and remove the check from here.

test/ng/compileSpec.js
```diff
@@ -980,6 +980,29 @@ describe('$compile', function() {
...
+            element = $compile(
+              '<div><div log-attrs replace class="myLog"></div><div>')($rootScope);
+            var div = element.find('div');
+            expect(div.attr('class')).toBe('myLog log');
+            expect(attrs.class).toBe('myLog log');
+          });
+        });
```
gkalpak: I would add a test for `style` as well, since it is a different code path.

src/ng/compile.js
```diff
@@ -2761,18 +2761,22 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-          $element.attr('style', $element.attr('style') + ';' + value);
-          dst['style'] = (dst['style'] ? dst['style'] + ';' : '') + value;
-          // `dst` will never contain hasOwnProperty as DOM parser won't let it.
-          // You will get an "InvalidCharacterError: DOM Exception 5" error if you
-          // have an attribute like "has-own-property" or "data-has-own-property", etc.
-        } else if (key.charAt(0) !== '$' && !dst.hasOwnProperty(key)) {
-          dst[key] = value;
-          dstAttr[key] = srcAttr[key];
+        // Check if we already set this attribute in the loop above
+        if (!dst[key]) {
```
gkalpak: Without thinking much about it, using `!dst.hasOwnProperty(key)` feels safer (and should be quicker).

test/ng/compileSpec.js
Narretz: I just tried removing it - injection seems necessary

test/ng/compileSpec.js
Narretz: Really? Both are handled by the `if (!dst[key]) {` before the src loop.

Narretz:
I just thought about this a bit more, and with this change we now have two different strategies of updating the element attributes for class & style. Before this change, classes and styles from "src" (new element) would always be added with safeAddClass() and $element.attr('style') respectively. Now we only add them with dst.$set().

Narretz:
The whole mergeTemplateAttributes fn is a crazy patchwork of commits. 
History:
https://github.com/angular/angular.js/commit/8af4fde18246ac1587b471a549e70d5d858bf0ee#diff-0242b0d33ee2848e02ae5bf32cc2b645R511
https://github.com/angular/angular.js/commit/f49eaf8bf2df5f4e0e82d6c89e849a4f82c8d414
https://github.com/angular/angular.js/commit/fb99b539b4d851773b43f1564f7032adb157c0db
https://github.com/angular/angular.js/commit/e1254b266dfa2d4e3756e4317152dbdbcabe44be#diff-a732922b631efed1b9f33a24082ae0dbR1591
https://github.com/angular/angular.js/commit/1ab6e908b15470d59b52eb0ead20c755c66ec3b8

Narretz:
Oh, look, you can replace the whole special-casing for the src attributes!

test/ng/compileSpec.js
gkalpak: :(

test/ng/compileSpec.js
gkalpak: There are different `if/else` branhes for each. Most importantly, the one uses `safeAddClass(...)` and the other `$element.attr('style', ...)`. We need to make sure that whatever these code paths do (which are skipped when `dst[key]` is trythy, has already been done in the previous loop (where we use `dst.$set(...)`).

src/ng/compile.js
```diff
@@ -2761,18 +2761,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-          // have an attribute like "has-own-property" or "data-has-own-property", etc.
-        } else if (key.charAt(0) !== '$' && !dst.hasOwnProperty(key)) {
-          dst[key] = value;
-          dstAttr[key] = srcAttr[key];
+        // Check if we already set this attribute in the loop above.
+        // `dst` will never contain hasOwnProperty as DOM parser won't let it.
+        // You will get an "InvalidCharacterError: DOM Exception 5" error if you
+        // have an attribute like "has-own-property" or "data-has-own-property", etc.
+        if (!dst.hasOwnProperty(key) && key.charAt(0) !== '$') {
+          dst.$set(key, value, true, srcAttr[key]);
```
gkalpak: Interesting. This will also do some more validation/sanitizing (e.g. for `src` attributes etc). I guess it's a good thing, but maybe also a breaking change.

I wasn't able to find where in the previous code were these new `dst` attributes (the ones copied from `src`) added to the actual element :disappointed: - we might duplicating some work (not that it's a big deal).

Narretz: Oh ... interesting. The attributes are previously added with the `replaceWith` function, which basically overwrites the dst element with everything from the src element. mergeTemplateAttributes is then really just for merging, not plain adding. So using dst.$set actually does work we already did in replaceWith(), and since replaceWith adds all template attributes that don't need merging, we should be able to remove `$element.attr('style', value)` and `safeAddClass($element, value)`.

petebacondarwin:
Holy Cow! Now that is a neat bit of refactoring.
I haven't tested it but since you have only added tests then I guess it is good to go!

src/ng/compile.js
```diff
@@ -2761,18 +2761,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-        } else if (key.charAt(0) !== '$' && !dst.hasOwnProperty(key)) {
+        // Check if we already set this attribute in the loop above.
+        // `dst` will never contain hasOwnProperty as DOM parser won't let it.
+        // You will get an "InvalidCharacterError: DOM Exception 5" error if you
+        // have an attribute like "has-own-property" or "data-has-own-property", etc.
+        if (!dst.hasOwnProperty(key) && key.charAt(0) !== '$') {
           dst[key] = value;
-          dstAttr[key] = srcAttr[key];
+
+          if (key !== 'class' && key !== 'style') {
```
gkalpak: I think this is a breaking (and unnecessary) change :smiley:
Why not set `dstAttr[key]` for `class` and `style` ?

petebacondarwin: @gkalpak - I don't believe this is a BC. The previous behaviour was not to set `distAttr` for `class` and `style`, so doing so would actually be a BC.

gkalpak: Wow, you are right, it didn't. Any idea why?
It seems like an oversight (bug?)...

Anyway, this is consistent, so no harm :smiley:

PR #14738, opened 2016-06-08
 fix($compile): don't trim white-space in attributes

Narretz:
Bug fix

$compile trims white-space from attributes when making them available to directives.

Yes. Previosuly stray white-space was trimmed, and therefore didn't impact application logic. I suspect this could have a moderate impact on existing apps - it depends on how common it is that attributes are read and used.
I think however that not trimming white-space lets Angular work more like it is expected from HTML (which allows almost anything as an attribute value)

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2753,7 +2753,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       forEach(dst, function(value, key) {
         if (key.charAt(0) !== '$') {
           if (src[key] && src[key] !== value) {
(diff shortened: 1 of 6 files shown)


gkalpak:
Even looking back at[ where the trimming originated][1], I couldn't find any compelling reason for trimming. Since there are usecases for which not trimming is necessary/makes things much easier, I think it makes sense to stop trimming.

LGTM

[1]: https://github.com/angular/angular.js/commit/8af4fde18246ac1587b471a549e70d5d858bf0ee

petebacondarwin:
LGTM

Issue #9140, opened 2014-09-18
Loader isn't working when adding ng-resource (and others)


mustela:
I'm using the last angular version (1.2.25), and using [angular-loader](https://github.com/angular/bower-angular-loader), but it seems that isn't working when trying to load ngresource, or ng-sanitize (and others). 

I've tested it using angular-seed (using aync index) to see if it was working there, but having the same results:

```
[Error] TypeError: 'undefined' is not a function (evaluating 'angular.$$minErr('$resource')') (angular-resource.js, line 8)
[Error] Error: [$injector:modulerr] Failed to instantiate module myApp due to:
[$injector:modulerr] Failed to instantiate module ngResource due to:
[$injector:nomod] http://errors.angularjs.org/1.2.25/$injector/nomod?p0=ngResource
http://localhost:8000/app/:22:327
http://localhost:8000/app/:23:323
http://localhost:8000/app/bower_components/angular/angular.js:3878:35
forEach@http://localhost:8000/app/bower_components/angular/angular.js:325:22
loadModules@http://localhost:8000/app/bower_components/angular/angular.js:3872:12
http://localhost:8000/app/bower_components/angular/angular.js:3879:51
forEach@http://localhost:8000/app/bower_components/angular/angular.js:325:22
loadModules@http://localhost:8000/app/bower_components/angular/angular.js:3872:12
createInjector@http://localhost:8000/app/bower_components/angular/angular.js:3812:22
doBootstrap@http://localhost:8000/app/bower_components/angular/angular.js:1444:34
bootstrap@http://localhost:8000/app/bower_components/angular/angular.js:1459:23
http://localhost:8000/app/:52:24
o@http://localhost:8000/app/:36:684
http://localhost:8000/app/:36:167
[...]

PR #14746, opened 2016-06-09
docs(misc/downloading): update content - include missing modules

gkalpak:
Docs update

No

**Other information**:
I deliberately didn't mention the `ngMessageFormat` module, since it seems to not be developed/maintained any more. Do we still want to mention it ?

### DIFF
--- a/docs/content/misc/downloading.ngdoc
+++ b/docs/content/misc/downloading.ngdoc
@@ -2,22 +2,22 @@
 @name Downloading
 @description
 
-# Including angular scripts from the Google CDN
-The quickest way to get started is to point your html `<script>` tag to a Google CDN URL.
+# Including Angular scripts from the Google CDN
+The quickest way to get started is to point your html `<script>` tag to a
+[Google CDN](https://developers.google.com/speed/libraries/#angularjs) URL.
 This way, you don't have to download anything or maintain a local copy.
 
-There are two types of angular script URLs you can point to, one for development and one for
+There are two types of Angular script URLs you can point to, one for development and one for
 production:
 
-* __angular.js__ — This is the human-readable, non-minified version, suitable for web
-development.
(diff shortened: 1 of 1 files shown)


docs/content/misc/downloading.ngdoc
```diff
@@ -29,58 +29,94 @@ example points to the minified version 1.5.6:
...
+
+* __`angular-csp.css`__ — You only need this file if you are using
+  [CSP (Content Security Policy)](https://developer.mozilla.org/en/Security/CSP). See
+  {@link directive:ngCsp here} for more info.
+
+* __`angular-mocks.js`__ — This file contains an implementation of mocks that makes testing angular
+  apps even easier. Your unit/integration test harness should load this file after `angular.js` is
+  loaded.
+
+* __`angular-loader.js`__ — Module loader for Angular modules. If you are loading multiple
```
Narretz: Actually, you need to load angular.js before any other module files, as many of them rely on angular.* helpers.

Narretz:
I don't see why we can't "advertise" ngMessageFormat. It works even though it's not complete.

docs/content/misc/downloading.ngdoc
gkalpak: I left it as it was previously (probably outdated as the rest of the content :smiley:).

TBH, I've never used the loader. So, does it mean you need to load the loader (synchronously), then angular.js (synchronously), then other modules (asynchronously and in any order) ?

Narretz: Yes, I think so. See https://github.com/angular/angular.js/issues/9140#issuecomment-164430751

gkalpak: I see. I'll add a note, but `angular-seed` needs fixing too.
(Although I think its `index-async` is already broken for soe reason. It's been sitting on my todo list for a while to take a deeper look - I guess it's time :smiley:)

PR #14750, opened 2016-06-09
fix($routeProvider): do not deep-copy route definition objects

### LINKED ISSUE #14478: $routeProvider deep-copies its route objects before using them, which is probably not necessary

gkalpak:
Bug fix

When registering a route, route definition objects are deep-copied, which breaks specific custom implementations of `$sce` (used to trust a `templateUrl` as RESOURCE_URL).

When registering a route, route definition objects are shallow-copied, which doesn't break custom implementations of `$sce`.

No

**Other information**:
Deep-copying route definition objects can break specific custom implementations of `$sce` (used to trust a `templateUrl` as RESOURCE_URL. The purpose of copying route definition objects was to guard against the user's modifying the route definition object after route registration, while still capturing inherited properties.
As suggested by @IgorMinar in https://github.com/angular/angular.js/pull/14699#discussion_r66480539, we can achieve both _and_ support custom `$sce` implementations, by shallow-copying instead.

This is an alternative implementation for #14699, which avoids the breaking change.

Fixes #14478
Closes #14699

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -1,5 +1,7 @@
 'use strict';
 
+/* global shallowCopy: false */
+
 // There are necessary for `shallowCopy()` (included via `src/shallowCopy.js`)
 var isArray = angular.isArray;
 var isObject = angular.isObject;
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -1,5 +1,9 @@
 'use strict';
 
+// There are necessary for `shallowCopy()` (included via `src/shallowCopy.js`)
+var isArray = angular.isArray;
+var isObject = angular.isObject;
+
 /**
  * @ngdoc module
  * @name ngRoute
@@ -37,17 +41,6 @@ var ngRouteModule = angular.module('ngRoute', ['ng']).
 function $RouteProvider() {
   function inherit(parent, extra) {
     return angular.extend(Object.create(parent), extra);
-  }
-
-  function createShallowCopy(src) {
-    if (!angular.isObject(src)) return src;
-
-    var dst = {};
-    for (var key in src) {
-      dst[key] = src[key];
-    }
-
-    return dst;
   }
 
   var routes = {};
@@ -171,7 +164,7 @@ function $RouteProvider() {
    */
   this.when = function(path, route) {
     //copy original route object to preserve params inherited from proto chain
-    var routeCopy = createShallowCopy(route);
+    var routeCopy = shallowCopy(route);
     if (angular.isUndefined(routeCopy.reloadOnSearch)) {
       routeCopy.reloadOnSearch = true;
     }
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -39,6 +39,17 @@ function $RouteProvider() {
     return angular.extend(Object.create(parent), extra);
   }
 
+  function createShallowCopy(src) {
+    if (!angular.isObject(src)) return src;
+
+    var dst = {};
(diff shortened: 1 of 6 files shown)


test/ngRoute/routeSpec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-describe('$route', function() {
+fdescribe('$route', function() {
```
IgorMinar: fdescribe

src/ngRoute/route.js
```diff
@@ -160,7 +171,7 @@ function $RouteProvider() {
    */
   this.when = function(path, route) {
     //copy original route object to preserve params inherited from proto chain
-    var routeCopy = angular.copy(route);
+    var routeCopy = createShallowCopy(route);
```
IgorMinar: can't you use this one: https://github.com/angular/angular.js/blob/3ba29c464da9d73de0ea7c7716dc1ecc60bc5287/src/Angular.js#L941 ?

gkalpak: Since we are in a external module, we need to copy it. I started by copying that, but then removed the stuff that didn't seem relevant.

IgorMinar: ok. sgtm.

test/ngRoute/routeSpec.js
```diff
@@ -900,6 +900,87 @@ describe('$route', function() {
...
+    function() {
+      var $routeProvider;
+
+      module(function(_$routeProvider_) {
+        $routeProvider = _$routeProvider_;
+      });
+
+      inject(function($location, $rootScope, $route, $sce) {
+        var sceWrappedUrl = $sce.trustAsResourceUrl('foo.html');
+        $routeProvider.when('/foo', {templateUrl: sceWrappedUrl});
```
Narretz: why is the `$routeProvider` registration in the inject and not the module block?

Narretz:
LGTM. If you needed another one. :>

src/ngRoute/route.js
petebacondarwin: Isn't this another case where we should split the shared function into its own file so that we can reuse it in multiple modules?

gkalpak: We could, but I wouldn't.

We don't need the exact same function. The one from core can shallow-copy arrays as well, ignores `$$` prefixed properties and supports passing your own `dst` object, which properties are shallow-copied onto. We we don't need any of that.

test/ngRoute/routeSpec.js
gkalpak: Because I couldn't access `$sce` during config :worried:

petebacondarwin:
LGTM

PR #14752, opened 2016-06-11
docs(ngOpen): add note about browser compatibility

wesleycho:
- Add note about browser compatibility and recommend ng-show/ng-hide

This should address #7337.


Narretz:
Thanks for the PR. We should also add a note about incomplete two-way binding in browsers that support details / summary, so if that's something you'd like to document: https://github.com/angular/angular.js/issues/3867

Issue #7218, opened 2014-04-23
html5 progress tag in IE11 does not recognize "value" attribute when bound


axlekb:
Here's a fiddle that shows the issue.  Works in Chrome/Firefox/Safari, but not IE11.

http://jsfiddle.net/AXNGq/1/

PR #14753, opened 2016-06-11
docs(guide): add note about value attribute

wesleycho:
- Add note about value attribute for the progress tag with IE

This should address #7218.


Narretz:
Can you please add the note to this section, too? https://docs.angularjs.org/guide/interpolation#-ngattr-for-binding-to-arbitrary-attributes
And while you are at it, add the note about "placeholder in <textarea> in Internet Explorer 10/11" to the IE compat section?

wesleycho:
Done

Issue #8562, opened 2014-08-11
Document default successCallback, errorCallback, notfiyCallback for promise.then(...)


fiveop:
The documentation for the `then` method of promises constructed using the `$q` service does not describe the behaviour of the default callbacks.

After I figure out what the defaults are, I will draft a patch.

PR #14754, opened 2016-06-11
docs(q): add note about optional arguments

wesleycho:
- Add note about optional arguments in the promise.then api

This should address #8562.


Narretz:
Optional arguments are usually marked with `[...]`. Can you add these to the initial description of .then under Methods?

wesleycho:
Done

PR #14756, opened 2016-06-13
docs($exceptionHandler): code snippet syntax is incorrect

geoffreybauduin:
Updated the documentation on exception handler, there was an end bracket missing.

Supposedly not a breaking change, it's only documentation update.


lght:
I signed it!

PR #14757, opened 2016-06-13
feat($q): implement $q.race

### LINKED ISSUE #12929: $q.race() method to match ES6
Hi,

Any chance of a `$q.race()` method being implemented? $q is compatible with the ES6 promise spec apart from the race method.

Thanks

yihangho:
Implement $q.race

$q.race is non-existent (#12929)

Nope.

**Other information**:

Implement $q.race. $q.race takes in an array or hash of promises and
returns a promsie that is resolved as soon as the first promise is
resolved.

Closes #12929

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -636,6 +636,34 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
     return deferred.promise;
   }
 
+  /**
+   * @ngdoc method
+   * @name $q#race
+   * @kind function
+   *
+   * @description
+   * Returns a promise that is resolved similarly as the first promise to resolve.
+   *
+   * @param {Array.<Promise>|Object.<Promise>} promises An array or hash of promises.
+   * @returns {Promise} Returns a promise that will be resolved as soon as the first promise in
+   *   `promises` is resolved. The resolved value will be the same as the value of the first
+   *   resolved promise.
+   */
+
+  function race(promises) {
+    var deferred = new Deferred();
+
+    forEach(promises, function(promise) {
+      when(promise).then(function(value) {
+        deferred.resolve(value);
+      }, function(reason) {
+        deferred.reject(reason);
+      });
+    });
+
+    return deferred.promise;
+  }
+
   var $Q = function Q(resolver) {
     if (!isFunction(resolver)) {
       throw $qMinErr('norslvr', "Expected resolverFn, got '{0}'", resolver);
@@ -665,6 +693,7 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
   $Q.when = when;
   $Q.resolve = resolve;
   $Q.all = all;
+  $Q.race = race;
 
   return $Q;
 }
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -651,14 +651,10 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
    */
 
   function race(promises) {
-    var deferred = new Deferred();
+    var deferred = defer();
 
     forEach(promises, function(promise) {
-      when(promise).then(function(value) {
-        deferred.resolve(value);
-      }, function(reason) {
-        deferred.reject(reason);
-      });
+      when(promise).then(deferred.resolve, deferred.reject);
     });
 
     return deferred.promise;
--- a/test/ng/qSpec.js
+++ b/test/ng/qSpec.js
@@ -1986,7 +1986,7 @@ describe('q', function() {
     });
   });
 
-  describe('race', function() {
+  describe('race (array)', function() {
     it('should do nothing if given an empty array', function() {
       q.race([]).then(success(), error());
       expect(mockNextTick.queue.length).toBe(0);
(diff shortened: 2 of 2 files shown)


test/ng/qSpec.js
```diff
@@ -1986,6 +1986,42 @@ describe('q', function() {
...
+      expect(logStr()).toBe('');
+      syncReject(deferred1, 'hi');
+      expect(logStr()).toBe('error(hi)->reject(hi)');
+      syncResolve(deferred2, 'cau');
+      expect(logStr()).toBe('error(hi)->reject(hi)');
+      syncReject(deferred, 'hola');
+      expect(logStr()).toBe('error(hi)->reject(hi)');
+    });
+  });
+
```
petebacondarwin: We should probably also test passing an object as the param, e.g. `$q.race({a: promise1, b: promise2})`.

src/ng/q.js
```diff
@@ -636,6 +636,34 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
...
+   * @returns {Promise} Returns a promise that will be resolved as soon as the first promise in
+   *   `promises` is resolved. The resolved value will be the same as the value of the first
+   *   resolved promise.
+   */
+
+  function race(promises) {
+    var deferred = new Deferred();
+
+    forEach(promises, function(promise) {
+      when(promise).then(function(value) {
```
petebacondarwin: Is there a situation where `forEach` can throw? If there is then we should catch it and reject the `deferred`.

petebacondarwin:
Other than a few minor questions, this LGTM

test/ng/qSpec.js
yihangho: I've added tests for object params.

src/ng/q.js
yihangho: Hmm, interesting. I am not too sure, but looking at the source for [`forEach`](https://github.com/angular/angular.js/blob/915e7a109c8bb512d6e31e639d898095be326c29/src/Angular.js#L237-L277), I don't think it will throw.

test/ng/qSpec.js
gkalpak: Do we want to support this (considering it is not supported in native Promises)?

src/ng/q.js
```diff
@@ -636,6 +636,34 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
...
+  /**
+   * @ngdoc method
+   * @name $q#race
+   * @kind function
+   *
+   * @description
+   * Returns a promise that is resolved similarly as the first promise to resolve.
+   *
+   * @param {Array.<Promise>|Object.<Promise>} promises An array or hash of promises.
+   * @returns {Promise} Returns a promise that will be resolved as soon as the first promise in
```
gkalpak: This description is a little missleading. It gives the impression we are looking for the first promise to actually resolve (not reject).

src/ng/q.js
gkalpak: Indeed the code inside the `forEach` callback (`when(...).then(...)`) won't throw.

gkalpak: BTW, this can be abbreviated as:

```js
when(promise).then(deferred.resolve, deferred.reject);
```

src/ng/q.js
yihangho: How about something like this:

```
Returns a promise that will be settled as soon as the first promise in `promises` is settled. The value or reason will be the same as that of the first settled promise.
```

Note that the term `settled` is also used in the [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects).

src/ng/q.js
yihangho: > Indeed the code inside the `forEach` callback (`when(...).then(...)`) won't throw.

I think what @petebacondarwin meant is more on whether `forEach` will throw on its own (i.e., due to bad input), no? But as mentioned, looking at how `forEach` is implemented, I don't think that will happen. Furthermore, since `when(...).then(...)` is also safe, I guess this should be safe. :smile:

> BTW, this can be abbreviated as:
>
> `when(promise).then(deferred.resolve, deferred.reject);`

Good point. Thanks for pointing this out!

src/ng/q.js
```diff
@@ -636,6 +636,30 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
...
+   *   `promises` is resolved. The resolved value will be the same as the value of the first
+   *   resolved promise.
+   */
+
+  function race(promises) {
+    var deferred = defer();
+
+    forEach(promises, function(promise) {
+      when(promise).then(deferred.resolve, deferred.reject);
```
yihangho: @gkalpak Does this look good to you? (Notice I changed the initialization of `deferred` from `new Deferred()` to `defer()`. The reason is the the `resolve` and `reject` methods are bounded if the object is coming from `defer()`.)

test/ng/qSpec.js
yihangho: Hmm, I am not too sure too. I decided to also take care of objects as this is also the case for `$q.all`.

src/ng/q.js
gkalpak: I just realized that we are using `resolved` inconsistently in the docs (either as `fulfilled` or `resolved` in the sense of the spec) :unamused: (TBH, I myself thought `resolved` means `fulfilled` until now...)

I would suggest being a little more explicit:
> Returns a promise that will be settled (i.e. fulfilled or rejected)...

Don't forget to also update the description above :innocent:

src/ng/q.js
gkalpak: I didn't realize you were using `new Deferred()` before. I think it's fine either way. Leaving it up to you!

petebacondarwin:
LGTM - merging

petebacondarwin:
tweaked the docs description to match that on MDN

thanks @yihangho

PR #14758, opened 2016-06-13
docs(guide/filters): clarify when filters are executed

Narretz:


docs/content/guide/filter.ngdoc
```diff
@@ -31,6 +31,16 @@ Filters may have arguments. The syntax for this is
 E.g. the markup `{{ 1234 | number:2 }}` formats the number 1234 with 2 decimal points using the
 {@link ng.filter:number `number`} filter. The resulting value is `1,234.00`.
 
+### When filters are executed
+
+In templates, filters are only executed when their inputs have changed. This is more performant than executing
```
gkalpak: I would clarify that this only applies to stateless filters (the default kind). Filters marked as stateful are called on every digest.

docs/content/guide/filter.ngdoc
```diff
@@ -31,6 +31,16 @@ Filters may have arguments. The syntax for this is
...
+a filter on each {@link ng.$rootScope.Scope#$digest `$digest`} as is the case with {@link guide/expression expressions}.
+However, this is only true for filters that take [primitive values](https://developer.mozilla.org/docs/Glossary/Primitive)
+as inputs. Filters that receive [Objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Objects)
+as input are executed on each `$digest`, as it would be too costly to track if they have changed.
+
+This means custom filters should only rely on their inputs and not on external state (e.g. other Angular services).
```
gkalpak: I think should is too strong here. Only stateless filters "should" do that, right ?

Narretz: We are suggesting that all filters are stateless. The $stateful flag was really only a concession that was made after there were too many complaints. So I wouldn't draw too much attention to them.

docs/content/guide/filter.ngdoc
Narretz: Yes, I should mention this for completeness

docs/content/guide/filter.ngdoc
gkalpak: I don't want to draw attention to it either, but saying that "custom filters should only rely on their inputs" can be confusing for those implementing a stateful filter. If we can make this recommendation more accurate while not drawing much attention to statefull filters, that would be great :smiley:

gkalpak:
LGTM

PR #14760, opened 2016-06-13
fix(ngAnimate): allow removal of class that is scheduled to be added …

### LINKED ISSUE #14582: ngAnimate & ngClass - Does not remove a class that is still animating
Bug

http://plnkr.co/edit/wsVFlxSJFCslfV9Hossg?p=preview
When a class is added via ngClass then removed too quickly we end up in a state where still have the class on the element.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
http://plnkr.co/edit/wsVFlxSJFCslfV9Hossg?p=preview
*This is a race condition.  This plunker is failing over 90% of the time though,

The falsy ngClass is removed

Plnkr was failing in 1.4 -> 1.5.5 

Definitely related to ngAnimate as the issue goes away when its not used.
Likely related to the intermediate animate states.  In the plunker above we would see .background-add on the element at the time we try to remove .background
This appears to only happen when the new ngClass state is also adding classes.

Narretz:
bug fix

See #14582

- [-] Docs have been added / updated (for bug fixes / features)

**Other information**:

…with requestAnimationFrame

The following case can happen when ngClass updates an element's classes in very quick order in the following way:
- First animation adds class "a"
- A digest passes, but "a" is not yet added to the element
- Second animation adds class "b"
- No digest passes, and "a" is still not added to the element,
  because requestAnimationFrame hasn't been flushed yet
- Third animation removes class "a"
- the third animation gets merged into the second animation

Before this change:
- Because the element doesn't have class "a" yet, ngAnimate
  resolves that it cannot remove class "a". However,
  the first animation is still running, and finally adds "a"

After this change:
- ngAnimate reacts to the temporary class "add-a", which indicates
  that "a" is about to be added and decides that "a" can be removed
  after all.

This is a very rare case where setting the element's class
is not fast enough, and subsequent animations operate on incorrect assumptions.

"In the wild", this is caused by rapidly updating ngClass,
which uses inidvidual addClass and removeClass calls when both operations happen in a single digest.

Fixes #14582

### DIFF
--- a/src/ngAnimate/shared.js
+++ b/src/ngAnimate/shared.js
@@ -285,10 +285,10 @@ function resolveElementClasses(existing, toAdd, toRemove) {
     var prop, allow;
     if (val === ADD_CLASS) {
       prop = 'addClass';
-      allow = !existing[klass];
+      allow = !existing[klass] || existing[klass + REMOVE_CLASS_SUFFIX];
     } else if (val === REMOVE_CLASS) {
       prop = 'removeClass';
-      allow = existing[klass];
+      allow = existing[klass] || existing[klass + ADD_CLASS_SUFFIX];
     }
     if (allow) {
       if (classes[prop].length) {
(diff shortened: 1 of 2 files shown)


gkalpak:
LGTM

Issue #14761, opened 2016-06-13
API Site down 6/13: Angular dependency issue?


olozzalap:
API site fails to load Angular currently, defaults to unstyled html (see screenshot). I tested on Chrome, Firefox and Safari on multiple different pages.
![angularjs](https://cloud.githubusercontent.com/assets/4325756/16016139/4afd6d56-3167-11e6-901c-4aaefd907702.jpg)

llanzo:
I can verify that I've been having the same issue. ( on https://docs.angularjs.org/api )

kbuechner:
having the same issue 

![image](https://cloud.githubusercontent.com/assets/13348260/16016790/b222f7e6-316a-11e6-916f-12e3c1530d5a.png)

emilycoco:
+1

jgalb:
+1

joaquinnunez:
+1

mullermw:
+1

caitp:
Don't enter "+1" guys, you can just "like" the issue now, and it's less spammy that way =)

/CC @jeffbcross maybe you can poke someone to see what's wrong with the google CDN?

chaseoli:
+1

Narretz:
Docs should be back up. You may need to do a hard reload for it to take effect. Thanks for noticing.

olozzalap:
Looks good!

petebacondarwin:
Hi Everyone! Sorry about the docs outage.

We made a change to our tooling so that we would have a better build experience while working offline. See https://github.com/angular/angular.js/commit/6a134605f1904098022f4e25ade67d0a91a52621

The aim of this commit was to prevent requests to remote servers for get the release tags and check the CDN versions when the building machine was not connected to the internet. In this cases we would still be able to build but we would not get the correct version information for the "production" docs app.

This commit was merged into master at 14:57 UTC and master builds started using it on our CI server.

Unfortunately our CI server seems to think it is offline at random points during its builds, even though it is permanently connected to the internet. We are not sure why - perhaps its bandwidth or latency is a bit flakey.

So although the builds to master succeeded, resulting in updates to our docs website, the CDN information was missing, which caused the production version of the site to fail to load the angular libraries.

@gkalpak reverted the offending commit at 17:44pm but the master build takes over 25 minutes to run.
@jeffbcross manually reverted the update to the docs website which brought it back online around 18:00 UTC.

Now that the commit has been reverted, future builds should not cause the docs to fail in the same way.

[...]

PR #14769, opened 2016-06-14
Cdn check

petebacondarwin:

chores

- failure to collect the correct CDN version information can cause the docs website to crash
- builds always attempt to make requests to servers, even if you are offline

- builds fail if the CDN version information is missing or if the production version of the app is not working
- you can set your local builds to run without making requests to the server

Nope

**Other information**:

Related to #14761

### DIFF
--- a/docs/app/e2e/app.scenario.js
+++ b/docs/app/e2e/app.scenario.js
@@ -39,7 +39,7 @@ describe('docs.angularjs.org', function () {
 
 
     it('should change the page content when clicking a link to a service', function () {
-      browser.get('build/docs/index.html');
+      browser.get('build/docs/index-production.html');
 
       var ngBindLink = element(by.css('.definition-table td a[href="api/ng/directive/ngClick"]'));
       ngBindLink.click();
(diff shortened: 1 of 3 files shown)


Narretz:
OFFLINE is not 100% correct, it's more like NO_REMOTE_REQUESTS

docs/app/e2e/app.scenario.js
```diff
@@ -39,7 +39,7 @@ describe('docs.angularjs.org', function () {
 
 
     it('should change the page content when clicking a link to a service', function () {
-      browser.get('build/docs/index.html');
+      browser.get('build/docs/index-production.html');
```
gkalpak: So, by using `-production` we are not testing against our local build, but only against the CDN version, right?
This makes sense, since it is a separate app that just happens to live on the same repo and the app will use a CDN version (not our local build).

lib/versions/version-info.js
```diff
@@ -138,6 +141,10 @@ var getCdnVersion = function() {
     .reverse()
     .reduce(function(cdnVersion, version) {
       if (!cdnVersion) {
+        if (OFFLINE) {
+          // We are offline so just use the most recent version
+          cdnVersion = version;
```
gkalpak: Shouldn't we return here?

lib/versions/version-info.js
```diff
@@ -204,3 +211,13 @@ exports.gitRepoInfo = gitRepoInfo = getGitRepoInfo();
 exports.previousVersions = previousVersions = getPreviousVersions();
 exports.cdnVersion = cdnVersion = getCdnVersion();
 exports.currentVersion = getTaggedVersion() || getSnapshotVersion();
+
+if (OFFLINE) {
+  console.log('==============================================================================================');
+  console.log('Running Offline:');
+  console.log(' - this due to the "ANGULAR_1_OFFLINE_BUILD" environment variable being defined.');
```
gkalpak: _this due_  -->  _this is due_ (?)

petebacondarwin:
OK @Narretz, I can change that constant. Any other concerns?

Narretz:
I'm still testing.

lib/versions/version-info.js
Narretz: Yes, otherwise we are making one request

docs/app/e2e/app.scenario.js
petebacondarwin: Yes, so there are other tests in this folder that hit the normal `index.html`.
Also all the main e2e tests from the examples are run against local angular.

lib/versions/version-info.js
petebacondarwin: OK!

lib/versions/version-info.js
petebacondarwin: Right

lib/versions/version-info.js
```diff
@@ -204,3 +213,13 @@ exports.gitRepoInfo = gitRepoInfo = getGitRepoInfo();
 exports.previousVersions = previousVersions = getPreviousVersions();
 exports.cdnVersion = cdnVersion = getCdnVersion();
 exports.currentVersion = getTaggedVersion() || getSnapshotVersion();
+
+if (NO_REMOTE_REQUESTS) {
+  console.log('==============================================================================================');
+  console.log('Running Offline:');
```
Narretz: "Running without getting remote tag and cdn version data"?

petebacondarwin: "Running with no remote requests for version data"?

Narretz:
LGTM. Please be prepared for a follow-up that tries to make the "online" version work on Windows (because it doesn't like `-o dev/null`)

PR #14772, opened 2016-06-14
fix(input): update $viewValue when cleared

### LINKED ISSUE #11193: AngularJS binding does not update when IE's clear button is clicked
IE has an "X" in each text input that will clear the input. However, when clicking this button, while it clears the textbox, it does not update the Angular model that the input is bound to.

````
<input type="text" ng-model="name" />
````

See http://jsfiddle.net/p5x1zwr9/ for an example of the behavior.

See http://youtu.be/LFaEwliTzpQ for a video of the behavior.

I am using IE 11.

This actually worked in AngularJS versions prior to 1.3.6 so this may be a new Angular bug.

### PR DESCRIPTION
**Pull request by wesleycho**
- Fix when user clicks clear button in an input element in IE, $viewValue not being correctly updated

This should fix #11193.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1126,7 +1126,11 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     });
   }
 
-  var timeout;
+  var timeout, oldVal;
+  var viewValueUpdated = false, msieInput = msie >= 10 && msie <= 11;
+  if (msieInput) {
+    oldVal = element.val();
+  }
 
   var listener = function(ev) {
     if (timeout) {
@@ -1152,10 +1156,18 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     }
   };
 
+  function ieListener(ev) {
+    var val = element.val();
+    if (val === oldVal && !viewValueUpdated) return;
+    oldVal = val;
+    viewValueUpdated = false;
+    listener(ev);
+  }
+
   // if the browser does support "input" event, we are fine - except on IE9 which doesn't fire the
   // input event on backspace, delete or cut
   if ($sniffer.hasEvent('input')) {
-    element.on('input', listener);
+    element.on('input', msie ? ieListener : listener);
   } else {
     var deferListener = function(ev, input, origValue) {
       if (!timeout) {
@@ -1212,6 +1224,10 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     // Workaround for Firefox validation #12102.
     var value = ctrl.$isEmpty(ctrl.$viewValue) ? '' : ctrl.$viewValue;
     if (element.val() !== value) {
+      // Workaround for IE 10 & 11 input updates #11193
+      if (msieInput) {
+        viewValueUpdated = true;
+      }
       element.val(value);
     }
   };
--- a/src/ng/sniffer.js
+++ b/src/ng/sniffer.js
@@ -73,7 +73,7 @@ function $SnifferProvider() {
         // when cut operation is performed.
         // IE10+ implements 'input' event but it erroneously fires under various situations,
         // e.g. when placeholder changes, or a form is focused.
-        if (event === 'input' && msie <= 11) return false;
+        if (event === 'input' && msie <= 9) return false;
 
         if (isUndefined(eventSupport[event])) {
           var divElm = document.createElement('div');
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Narretz** [comment]
Cool! I believe writing a test is difficult because we can't click the clear button?

**Comment by wesleycho** [comment]
Yeah - fussing around with this a bit to get the right condition currently, but I think the right fix might be attainable here.

**Comment by wesleycho** [comment]
This is a little tough - so falsy model changes will convert to the empty string, so this potentially breaks some behavior. Any ideas on good workarounds for that situation?

File: src/ng/directive/input.js [hunk 1]
```diff
@@ -1126,7 +1126,11 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     });
   }
 
-  var timeout;
+  var timeout, oldVal;
+  var viewValueUpdated = true, msieInput = msie >= 10 && msie <= 11;
+  if (msieInput && inputType === 'text') {
```
jbedard: Why only text inputs? Won't types such as number have the same issue?

wesleycho: Looks like you're correct - I assumed it would have the spinner, but apparently not.

src/ng/directive/input.js
```diff
@@ -1152,10 +1158,18 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     }
   };
 
+  function ieListener(ev) {
+    var val = element.val();
+    if (val === oldVal && !viewValueUpdated) return;
```
jbedard: Is it possible that the validity changed but the value did not (similar to the `value === '' && ctrl.$$hasNativeValidators` check in `listener`)? Although I'm not seeing how that would have been handled previously either (in the `!hasEvent('input')` case)...

src/ng/directive/input.js
```diff
@@ -1126,7 +1128,11 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     });
   }
 
-  var timeout;
+  var timeout, oldVal;
+  var viewValueUpdated = true, msieInput = msie >= 10 && msie <= 11;
+  if (msieInput && attr.type in IE_INPUTS_WITH_CLEARING) {
```
jbedard: Why do we even care about the type for setting the initial `oldVal` though? I'd think we'd always want to setup the `oldVal` if `msieInput` is true...

src/ng/directive/input.js
```diff
@@ -1126,7 +1128,11 @@ function baseInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     });
   }
 
-  var timeout;
+  var timeout, oldVal;
+  var viewValueUpdated = false, msieInput = msie >= 10 && msie <= 11;
+  if (msieInput && attr.type in IE_INPUTS_WITH_CLEARING) {
```
jbedard: Why do we even care about the type for setting the initial oldVal though? I'd think we'd always want to setup the `oldVal` if `msieInput` is true...

wesleycho: Wouldn't that trigger a potentially unnecessary DOM access though? That is likely more expensive than this check I think. I originally had it just accessing the value, I'd be fine changing it back.

jbedard: If we keep it I think it should be part of the `msieInput` condition, not this setup condition.

But I'd think `element.val()` is cheap enough it doesn't matter.

wesleycho:
There are some complicated scenarios I'm having trouble figuring out a good solution to.

Here is an incomplete table:
<table>
  <thead>
    <tr>
      <th>Initial view value</th>
      <th>Initial model value</th>
      <th>View value after view change</th>
      <th>Model value after model change</th>
      <th>Erroneous IE input events</th>
      <th>Run $parsers/$validators</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>''</td>
      <td>undefined</td>
      <td>''</td>
      <td>N/A</td>
      <td>false</td>
      <td>true</td>
    </tr>
    <tr>
      <td>''</td>
      <td>null</td>
      <td>''</td>
      <td>N/A</td>
      <td>false</td>
      <td>true</td>
    </tr>
    <tr>
      <td>''</td>
      <td>null</td>
      <td>N/A</td>
      <td>''</td>
      <td>false</td>
      <td>true</td>
    </tr>
    <tr>
      <td>''</td>
      <td>undefined</td>
      <td>N/A</td>
      <td>N/A</td>
      <td>true</td>
      <td>false</td>
    </tr>
  </tbody>
</table>

gkalpak:
Wouldn't it be easier to just add an extra listener on IEs:

```js
element.on('??mousedown/mouseup/click??', function(event) {
  deferListener(event, this, this.value);
});
```

Narretz:
@wesleycho Do you still want to work on this?

wesleycho:
Don't have the bandwidth anymore :( .

blemaire:
What's the status on this issue (not doubt i'm not the only one encountering this)
Thanks

PR #14775, opened 2016-06-14
fix(ngTransclude): correctly support fallback content

### LINKED ISSUE #14765: Input with ng-model as default transclude keeps validators alive

Report a bug

When using multi slot transclusion, if an input element with ng-model is being used as the default\fallback, when replace with trascluded content, validators from the default input are still present and are preventing the form from being valid.

I think it is explained well above.

The validators should be destroyed in the $destroy function of the NgModelController

Explained aboe

angular 1.5.6
chrome 51.0.2704.84
Mac osx El Capitan
I just started using the multi slot transclusion and since it is new, i don't believe this ever worked before.

for now i used ng-if on the default input as a workaround and checked for transcluded slot existance.

### LINKED ISSUE #14768: Combination of transclude, templateUrl, require produces "Error: [$compile:ctreq] Controller 'foo', required by directive 'blubb', can't be found!"
Dear angular guys,
we encountered following unexpected error message while using a combination of transclude with fallback content, templateUrl, and require:

    angular.js:13642 Error: [$compile:ctreq] Controller 'foo', required by directive 'blubb', can't be     found!
    http://errors.angularjs.org/1.5.6/$compile/ctreq?p0=foo&p1=blubb
        at angular.js:68
        at getControllers (angular.js:9237)
        at getControllers (angular.js:9244)
        at nodeLinkFn (angular.js:9160)
        at compositeLinkFn (angular.js:8459)
        at nodeLinkFn (angular.js:9151)
        at angular.js:9496
        at processQueue (angular.js:16104)
        at angular.js:16120
        at Scope.$eval (angular.js:17378)

The error occurs in following situations:
- A directive "bar" is defined with transclusion and an embedded fallback content
- The embedded fallback-content contains a directive "barr" which template is loaded via templateUrl
- The template of "barr" contains a directive "blubb" which requires a top controller "foo"
- Directive "bar" is used and own content is provided so the transclusion fallback content should be discarded.

The error does not show up, if no content on "bar" is provided and fallback content is used. Either it does not show up if template is provided via "template"-property and not "templateUrl"-property.

It looks like the error has no functional impact and just produces noise in the javascript console.

**Reproducable:** always
**Browsers:** Chromium 50, Firefox 46 
**Operating system:** Linux
**AngularJs:**  1.5.6, 1.4.0, 1.3.0

**Steps to reproduce**:

You can run this: https://plnkr.co/edit/0r3IFyL8og0Wo4sGIPBW

Or use following code:

    <!DOCTYPE html>
    
    <html>

    <head>
      <script data-require="angularjs@1.5.6" data-semver="1.5.6" src="https://code.angularjs.org/1.5.6/angular.js"></script>
      <link rel="stylesheet" href="style.css" />
    </head>
    
    <body ng-app="myApp">
      <div foo>
        
        <bar>Hi there!</bar>
        
        <bar></bar>
      </div>
    
      <script>
        angular.module('myApp', [])
          .directive('foo', function() {
            return {
              restrict: 'A',
              controller: function(){}
            };
          })
          .directive('bar', function($templateCache) {
                return {
                  transclude: true,
                  template: '<ng-transclude><barr></barr></ng-transclude>'
                };
          })
          .directive('barr', function($templateCache) {
            $templateCache.put("tmpl/barr.html", '<div blubb>Hello World!</div>');
            return {
              templateUrl: 'tmpl/barr.html'
            };
          })
          .directive('blubb', function() {
            return {
              restrict: 'A',
              require: ['^foo'],
              link: function() {}
            };
          })
      </script>
    </body>
    
    </html>

**update 1:**
Our expected behaviour is that if the "fallback content" from a transclude is not used there should be no lookup of required controllers from any directives from inside this discarded "fallback content".

dcherman:
If the transclude function did not return content, then the transcluded
scope that was created needs to be cleaned up in order to avoid a slight
amount of unnecessary overhead since the additional scope is no longer
needed.

If the transcluded content did return content, the the fallback content
should never have been linked in the first place as it was intended to be
immediately removed.

Fixes #14768
Fixes #14765

### DIFF
--- a/src/ng/directive/ngTransclude.js
+++ b/src/ng/directive/ngTransclude.js
@@ -159,35 +159,45 @@
  * </example>
  */
 var ngTranscludeMinErr = minErr('ngTransclude');
-var ngTranscludeDirective = ngDirective({
-  restrict: 'EAC',
-  link: function($scope, $element, $attrs, controller, $transclude) {
-
-    if ($attrs.ngTransclude === $attrs.$attr.ngTransclude) {
-      // If the attribute is of the form: `ng-transclude="ng-transclude"`
-      // then treat it like the default
-      $attrs.ngTransclude = '';
-    }
-
-    function ngTranscludeCloneAttachFn(clone) {
-      if (clone.length) {
-        $element.empty();
-        $element.append(clone);
+var ngTranscludeDirective = ['$compile', function($compile) {
+  return {
+    restrict: 'EAC',
+    terminal: true,
+    link: function($scope, $element, $attrs, controller, $transclude) {
+      if ($attrs.ngTransclude === $attrs.$attr.ngTransclude) {
+        // If the attribute is of the form: `ng-transclude="ng-transclude"`
+        // then treat it like the default
+        $attrs.ngTransclude = '';
       }
+
+      function ngTranscludeCloneAttachFn(clone, transcludedScope) {
+        if (clone.length) {
+          $element.empty();
+          $element.append(clone);
+        } else {
+          // Since this is the fallback content rather than the transcluded content,
+          // we compile against the scope we were linked against rather than the transcluded
+          // scope since this is the directive's own content
+          $compile($element.contents())($scope);
+
+          // There is nothing linked against the transcluded scope since no content was available,
+          // so it should be safe to clean up the generated scope.
+          transcludedScope.$destroy();
+        }
+      }
+
+      if (!$transclude) {
+        throw ngTranscludeMinErr('orphan',
+        'Illegal use of ngTransclude directive in the template! ' +
+        'No parent directive that requires a transclusion found. ' +
+        'Element: {0}',
+        startingTag($element));
+      }
+
+      // If there is no slot name defined or the slot name is not optional
+      // then transclude the slot
+      var slotName = $attrs.ngTransclude || $attrs.ngTranscludeSlot;
+      $transclude(ngTranscludeCloneAttachFn, null, slotName);
     }
-
-    if (!$transclude) {
-      throw ngTranscludeMinErr('orphan',
-       'Illegal use of ngTransclude directive in the template! ' +
-       'No parent directive that requires a transclusion found. ' +
-       'Element: {0}',
-       startingTag($element));
-    }
-
-    // If there is no slot name defined or the slot name is not optional
-    // then transclude the slot
-    var slotName = $attrs.ngTransclude || $attrs.ngTranscludeSlot;
-    $transclude(ngTranscludeCloneAttachFn, null, slotName);
-  }
-});
-
+  };
+}];
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -7937,11 +7937,23 @@ describe('$compile', function() {
 
       it('should clear contents of the ng-translude element before appending transcluded content' +
         ' if transcluded content exists', function() {
+        var contentsDidLink = false;
+
         module(function() {
+          directive('inner', function() {
+            return {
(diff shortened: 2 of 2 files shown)


src/ng/directive/ngTransclude.js
```diff
@@ -159,35 +159,45 @@
...
-    }
+      function ngTranscludeCloneAttachFn(clone, transcludedScope) {
+        if (clone.length) {
+          $element.empty();
+          $element.append(clone);
+        } else {
+          // Since this is the fallback content rather than the transcluded content,
+          // we compile against the scope we were linked against rather than the transcluded
+          // scope since this is the directive's own content
+          $compile($element.contents())($scope);
```
dcherman: Should probably check to see if any contents exist prior to calling `$compile`.

Narretz:
Could you please create new tests for this instead of modifying the old ones?

src/ng/directive/ngTransclude.js
```diff
@@ -159,35 +159,45 @@
...
 
-    function ngTranscludeCloneAttachFn(clone) {
-      if (clone.length) {
-        $element.empty();
-        $element.append(clone);
+          // There is nothing linked against the transcluded scope since no content was available,
+          // so it should be safe to clean up the generated scope.
+          transcludedScope.$destroy();
```
gkalpak: So, this is essentially fixing a leak, right ? Cool

src/ng/directive/ngTransclude.js
```diff
@@ -159,35 +159,45 @@
  * </example>
  */
 var ngTranscludeMinErr = minErr('ngTransclude');
-var ngTranscludeDirective = ngDirective({
-  restrict: 'EAC',
-  link: function($scope, $element, $attrs, controller, $transclude) {
+var ngTranscludeDirective = ['$compile', function($compile) {
+  return {
+    restrict: 'EAC',
+    terminal: true,
```
gkalpak: Does the fact that we will compile the fallback content later (if necessary) introduce a timing difference with the previous implementation? I.e. is the `$transclude` function called synchronously or will the uncompiled content stay visible for longer?

src/ng/directive/ngTransclude.js
petebacondarwin: This is not semantically necessary as the compile bails out if there is nothing to compile.

petebacondarwin: although it could improve performance slightly

petebacondarwin:
I moved the tests into their own specs.

src/ng/directive/ngTransclude.js
petebacondarwin: OK, so we could add a compile function that extracts the contents and removes if from the document so that it is hidden straight away

dcherman: Since linking is a sync process (at least as far as ngTransclude is concerned), I'm pretty sure you shouldn't be able to observe the uncompiled content.

The timing of the pre and post linking functions may have changed in relation to to cousin or uncle directives (yay DOM family!), however I believe that the timing should remain unchanged in relation to ancestor directives which is the more important part.

gkalpak: I'm mainly concerned about the timing of rendering (so users won't see flashing `{{ ... }}`s).
From what I understand from @dcherman's comment (because I was too lazy to read the source code :stuck_out_tongue:), this hasn't changed. So we're good :+1:

petebacondarwin: There was a problem in this PR with failing to link fallback content if there was a no content provided for an optional named slot. My #14787 fixes that

PR #14778, opened 2016-06-15
docs($resource): Fix the @-prefixed parameter examples to match the p…

hawicz:
**What kind of change does this PR introduce? docs update

n/a

n/a

no

**Other information**:

…roperty names of the corresponding data objects.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -210,7 +210,7 @@ function shallowClearAndCopy(src, dst) {
  *   as  methods with the `$` prefix. This allows you to easily perform CRUD operations (create,
  *   read, update, delete) on server-side data like this:
  *   ```js
- *   var User = $resource('/user/:userId', {userId:'@id'});
+ *   var User = $resource('/user/:userId', {userId:'@userId'});
  *   var user = User.get({userId:123}, function() {
  *     user.abc = true;
  *     user.$save();
@@ -327,7 +327,7 @@ function shallowClearAndCopy(src, dst) {
  * operations (create, read, update, delete) on server-side data.
 
    ```js
-     var User = $resource('/user/:userId', {userId:'@id'});
+     var User = $resource('/user/:userId', {userId:'@userId'});
      User.get({userId:123}, function(user) {
        user.abc = true;
        user.$save();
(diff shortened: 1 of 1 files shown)


petebacondarwin:
I'm pretty sure we have had this discussion before, similar to #11544.
The `@id` refers to the property on the resource object, not the param name in the URL.

>If the parameter value is prefixed with @ then the value for that parameter will be extracted from the corresponding property on the data object (provided when calling an action method). For example, if the defaultParam object is {someParam: '@someProp'} then the value of someParam will be data.someProp.

hawicz:
yes, and for some of the examples on the page @id is correct, but the ones that have a userId param also refer to an object whose property is called userId, not id.

hawicz:
Take the first mention of userId on that page, in the "Returns" section:

    var User = $resource('/user/:userId', {userId:'@id'});
    var user = User.get({userId:123}, function() {
      user.abc = true;
      user.$save();
    });

how would the value 123 from the object passed to get() end up mapped into "/user/123" here?

petebacondarwin:
In this case the object that is retrieved from the REST service has a property `user.id`.
Now, when we call save, the URL that is used in the POST request is `/user/123` because the `user.id == 123`.
You could test this by changing the `user.id` to something else and looking at what URL is requested in the POST when `$save()` is called

gkalpak:
I think the confusion stems from the fact that non-get requests (e.g. `User.save({...})`) receive a `data` object (from which the parameters can be extracted using the `{userId: '@id'}` mapping), while get requests (e.g. `User.get({...})`) can receive a `params` object (which has to reference params usign their actual name (i.e. `userId`).

A similar explanation can be found [here][1] as well.

[1]: https://github.com/angular/angular.js/pull/9243/files#r17963997

hawicz:
So, then perhaps the description of the @-parameters in the section about paramDefaults should say something like "...(provided when calling an action method, except for get() which uses the attributes in its passed object directly)..." ?

petebacondarwin:
We do need to improve the docs on this as it keeps coming up.

gkalpak:
I clarified this a bit in 7ab00f4.

petebacondarwin:
Thanks @gkalpak

PR #14779, opened 2016-06-15
docs(step_15): create new step about accessibility

felixzapata:
Docs update to add a new step about accessibility

The current tutorial is not accessible. With this new step you can learn [how to fix the issues](https://github.com/angular/angular-phonecat/pull/341).

Introduce the accessibility into the tutorial.

No, it does'nt. Only affects to the tutorial.

**Other information**:
My opinion is this step should be a process inside of the tutorial. Not a separate step. This way the user can understand that is optional but it should be mandatory.

### DIFF
--- a/docs/content/tutorial/step_15.ngdoc
+++ b/docs/content/tutorial/step_15.ngdoc
@@ -5,12 +5,10 @@
 
 <ul doc-tutorial-nav="15"></ul>
 
-In this step, we will learn how to fix the accessibility issues in our application.
+In this step, we will learn how to fix the accessibility issues in the application.
 
 * We will fix some issues in the templates.
 * We will use the [protractor-accessibility-plugin](protractor-a11y-plugin) to review the accessibility of our application when we run our e2e tests.
-* We will explain how to add the {@link ngAria ngAria} module to use WAI-ARIA easily 
-if you need it in your project.
 
 <div doc-tutorial-reset="15"></div>
 
@@ -27,7 +25,7 @@ to consume the content in a way that suits their needs.
 * [Notes On Client-Rendered Accessibility](https://www.smashingmagazine.com/2015/05/client-rendered-accessibility/)
 * [Apps For All: Coding Accessible Web Applications](https://shop.smashingmagazine.com/products/apps-for-all)
 
-## WAI-ARIA
+## About WAI-ARIA
 
 WAI-ARIA (_the Accessible Rich Internet Applications Suite_) is a standard set of attributes used
 to fill accessibility support gaps in custom HTML elements and SVG.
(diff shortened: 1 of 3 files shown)


gkalpak:
fyi @marcysutton

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,327 @@
...
+In this step, we will learn how to fix the accessibility issues in our application.
+
+* We will use the {@link ngAria ngAria} module to enable WAI-ARIA.
+* We will fix some issues in the templates.
+* We will use the [protractor-accessibility-plugin](protractor-a11y-plugin) to review the accessibility of our application when we run our e2e tests.
+
+<div doc-tutorial-reset="15"></div>
+
+## Why accessibility matters
+
```
gkalpak: Could you add a paragraph on what is "accessibility" and "aria" and how it affects people. Just a brief summary and then list the links for more info.

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,327 @@
...
+To fix this, we add two directives to inform the user how many items are know in the list after search or order them. This directive will fill an element in our page to tell the user how many elements are after the action.
+
+<br />
+**`app/phone-list/phone-list-template.html`:**
+
+```html
+  <div>
+    <label for="search">Search:</label> 
+    <input id="search" ng-model="$ctrl.query" ng-model-options="{ debounce: 900 }">
```
gkalpak: Why this large debounce delay? Is it necessary? Does the delay have to be that long?

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,327 @@
...
+  **Warning:** If a new version of Angular has been released since you last ran `npm install`, then
+  you may have a problem with the `bower install` due to a conflict between the versions of
+  angular.js that need to be installed. If you run into this issue, simply delete your
+  `app/bower_components` directory and then run `npm install`.
+</div>
+
+
+## How to use the `ngAria` module
+
+The `ngAria` module provides support for common ARIA attributes that convey state or semantic information about the application for users of assistive technologies, such as screen readers.
```
gkalpak: We should also have a paragragh on how our specific app will be benefited just by including `ngAria` (to give more emphasis on what `ngAria` does under the hood for us).

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,327 @@
...
+  }
+]);
+
+```
+
+## Using the keyboard
+
+The current detail of a phone prevents a user to use a keyboard to navigate for over each thumbnail image. It is very important due to accessibility that all the actions and information are available via keyboard too. For example, lightboxes, carousels, slides, etc.
+
+_Tip_: Try to navigate over the application using only the _TAB_ key. You will discover that some actions can't be accessed nor the focus on some elements.
```
gkalpak: So, how are we solving this?

gkalpak:
I left a few comments. Overall looks good :+1:

A couple more necessary changes:
- Limit lines to 100 chars max.
- Don't mention anything protractor-related until the "Testing" section (except for the step summary at the top).
- I would rather not introduce new features at this point. Custom `directives`, `$scope.$watch` etc, are not explained before. Are they necessary?

docs/content/tutorial/step_15.ngdoc
felixzapata: My first attempt was not to use the debounce, but without it, there problems to inform the user about the number of results. The information that the user obtains is not correct. It is neccesary an small "gap" to get the results, and update the aria region.

I can try to reduce this number to the mininum required.

felixzapata:
For this example, I think is better the directives rather to use a component. These directives are needed to intercept the repeater and return the values to the aria regions. Maybe this can do it in another way.

Another option would be to remove the section "How to use the directives" and add a link to the [Directives webpage](http://localhost:8000/build/docs/guide/directive).

docs/content/tutorial/step_15.ngdoc
felixzapata: I've reduced the debounce to 300.

docs/content/tutorial/step_15.ngdoc
felixzapata: I've added how to fix this problem (adding a button element). For this reason, and after check again the project, the `ngAria` module is not necessary (I added it, to activate the `ng-click` inside the `img` element).

docs/content/tutorial/step_15.ngdoc
felixzapata: I've added some definitions.

felixzapata:
@gkalpak I've made some updates. I will ping you, when I complete all of them (limit of lines, explanation of protractor accessibility plugin).

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,337 @@
...
+
+A website or application accessible allows the users with different preferences and abilitities, to consume the content in a way that suits their needs.
+
+* [Creating an accessible web - Access iQ](https://www.youtube.com/watch?v=47n0wEcm6JU)
+* [Web Accessibility](https://www.youtube.com/watch?v=bEM9Fn9aOG8)
+* [Google Accessibility course](https://webaccessibility.withgoogle.com/course)
+* [Accessibility in AngularJS and Beyond](http://marcysutton.com/slides/fluent2015/)
+* [Introduction to Web Accessibility](https://www.w3.org/WAI/intro/accessibility.php)
+* [Apps For All: Coding Accessible Web Applications](https://shop.smashingmagazine.com/products/apps-for-all)
+
```
marcysutton: Here's one more article that was written in response to Angular a11y: https://www.smashingmagazine.com/2015/05/client-rendered-accessibility/

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,337 @@
...
+## WAI-ARIA
+
+WAI-ARIA (_the Accessible Rich Internet Applications Suite_) gives you the ability to reclassify and otherwise augment the perceived meaning (or semantics) of your HTML.
+
+ Currently certain functionality used in Web sites is not available to some users with disabilities, especially people who rely on screen readers and people who cannot use a mouse. WAI-ARIA addresses these accessibility challenges, for example, by defining new ways for functionality to be provided to assistive technology. With WAI-ARIA, developers can make advanced Web applications accessible and usable to people with disabilities.
```
marcysutton: I would mention that ARIA is a standard set of attributes used to fill accessibility support gaps in custom HTML elements and SVG.

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,337 @@
...
+
+  <div class="aria-status" aria-live="assertive" aria-atomic="true"></div>
+  
+  ...
+  
+  <div class="aria-status-sort" aria-live="assertive" aria-atomic="true"></div>
+  
+```
+
+The new element is a _region live_. It means that every time this region is updated, it will inform the user about its content.
```
marcysutton: *live region

docs/content/tutorial/step_15.ngdoc
```diff
@@ -0,0 +1,337 @@
...
+    <li ng-repeat="img in $ctrl.phone.images">
+      <button type="button" aria-label="view enlarge version of the image" ng-click="$ctrl.setImage(img)"><img ng-src="{{img}}" alt="" /></button>
+    </li>
+  </ul>
+  ...
+```
+
+To fix this issue, we add a `button` element, wrapping the image to allow the user to access the content via keyboard.
+
+Why we are using a `button` element instead of an `a` element? It is more correct to use buttons elements to call actions like show content, hidden content or similar.
```
marcysutton: Because it does not navigate to a new page or resource.

felixzapata:
@gkalpak the limit of 100 characters is only for the code examples or all the document?

gkalpak:
@felixzapata, the limit is for the whole document (where ever possible, e.g. if a URL is too long it's OK).
Is this ready for another review?

felixzapata:
@gkalpak I have to review about the lines and I have to think what to do with the directives. Have you seen my previous comment about that? I mean, remove the section "How to use the directives" and add a link to the Directives webpage, instead of copy the code of the directives.

felixzapata:
@gkalpak is ready for another review:

- I've removed the information about directives. Now the tutorial has a link to the directives guide.
- I've removed the information about how to instal the ngAria module becuase, finally, is not used in this step.

When the step is approved, I will squash the commits.

felixzapata:
@gkalpak it is ready for another review.

felixzapata:
@gkalpak hi, any news about this pull request?

felixzapata:
I signed it!

gkalpak:
Sorry it is taking so long, @felixzapata. It is still on my todo list. I'll get to it after 1.5.9 and 1.6.0-rc.0 are released.

felixzapata:
ok @gkalpak

PR #14780, opened 2016-06-15
chore(version-info.js): ignore curl file write errors

Narretz:
On Windows, curl can obviously not write to `dev/null`, and throws code 23, which prevents us from checking that the CDN server responded. However, we don't need the file anyway, we just need the status code, so we can ignore the file write error.

This needs to be tested on Linux / Mac however!

### DIFF
--- a/lib/versions/version-info.js
+++ b/lib/versions/version-info.js
@@ -151,7 +151,9 @@ var getCdnVersion = function() {
           var cdnResult = shell.exec('curl http://ajax.googleapis.com/ajax/libs/angularjs/' + version + '/angular.min.js ' +
                     '--head --write-out "%{http_code}" -o /dev/null -silent',
                                       {silent: true});
-          if (cdnResult.code === 0) {
+          // Ignore if Curl can't write the file (code 23), e.g. on Windows, as we still get
+          // the needed output from --write-out
+          if (cdnResult.code === 0 || cdnResult.code === 23) {
             var statusCode = cdnResult.output.trim();
             if (statusCode === '200') {
               cdnVersion = version;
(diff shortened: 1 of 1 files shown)


lib/versions/version-info.js
```diff
@@ -151,7 +151,9 @@ var getCdnVersion = function() {
           var cdnResult = shell.exec('curl http://ajax.googleapis.com/ajax/libs/angularjs/' + version + '/angular.min.js ' +
                     '--head --write-out "%{http_code}" -o /dev/null -silent',
                                       {silent: true});
-          if (cdnResult.code === 0) {
+          // Ignore if Curl can't write the file (code 23), e.g. on Windows, as we still get
+          // the needed output from --write-out
+          if (cdnResult.code === 0 || cdnResult.code === 23) {
             var statusCode = cdnResult.output.trim();
```
petebacondarwin: Really we only care about whether requests to this URL are 404 or not, right?

Why don't we ditch the curl altogether and just use the `http` library?

https://nodejs.org/dist/latest-v4.x/docs/api/http.html#http_http_request_options_callback

lib/versions/version-info.js
```diff
@@ -220,6 +222,8 @@ if (NO_REMOTE_REQUESTS) {
...
   console.log('==============================================================================================');
-}
-console.log('CDN version:', cdnVersion.raw);
-console.log('Current version:', exports.currentVersion.raw);
\ No newline at end of file
+  console.log('Current local version', exports.currentVersion.raw);
+}  else {
+  console.log('CDN version:', cdnVersion.raw);
+  console.log('Current version:', exports.currentVersion.raw);
+}
```
petebacondarwin: We still have a `cdnVersion` that is going to be used in the docs, so I think it is worth knowing what we are using.

Issue #14781, opened 2016-06-15
Different grouping in select 1.4.10 vs. 1.5.0 vs. 1.5.4 (CodePens attached)


stschaefer:
Hi guys,

after updating angular from 1.4.9 to 1.5.6 I experience a different grouping behaviour in the select when using ng-options. 

Please take a look on my codepens:

http://codepen.io/hoody/pen/gMMPBg (Angular 1.4.10: expected behaviour)
http://codepen.io/hoody/pen/EyyPeJ (Angular 1.5.0: new optgroup container around option)
http://codepen.io/hoody/pen/ZOOQjq (Angular 1.5.6: new optgroup container around option AND empty optgroup before listing of optgroups)

Is this a known bug? Will it be fixed? Do you have a workaround for me? :)

Greetings,
Stefan

Narretz:
Thanks for the report, very thorough.
Quick Answer: Not a known bug, yes, no. :D

Narretz:
Sorry, I was a bit too quick. This is not a bug, but a known documented breaking change: https://code.angularjs.org/snapshot/docs/guide/migration#directives-ngoptions-

stschaefer:
Nevertheless, thanks for your quick reply. With the help of the link I could fix the issue. :+1: 

Greetings,
Stefan

Narretz:
@gkalpak There's still a bug here :>

gkalpak:
Sorry, I was a bit too quick. :stuck_out_tongue:

Narretz:
Fix for the duplication bug landed in https://github.com/angular/angular.js/commit/6bc81ae6ef96436b27a9aa6ee3769e394118ecc9

PR #14784, opened 2016-06-15
fix(ngOptions): don't duplicate groups with falsy values

Narretz:
bugfix

see #14781

- ~~[ ]Docs have been added / updated (for bug fixes / features)~~

This happened when the options were updated, because ngOptions would fail
to remove optgroups with falsy values

Related #14781

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -629,7 +629,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
 
           for (var i = options.items.length - 1; i >= 0; i--) {
             var option = options.items[i];
-            if (option.group) {
+            if (isDefined(option.group)) {
               jqLiteRemove(option.element.parentNode);
             } else {
               jqLiteRemove(option.element);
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngOptions.js
```diff
@@ -661,7 +661,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
               listFragment.appendChild(groupElement);
 
               // Update the label on the group element
-              groupElement.label = option.group;
+              groupElement.label = option.group === null ? 'null' : option.group;
```
gkalpak: Is this needed for some specific browser? In Chrome that I tried, `null` seems to be converted to `"null"` anyway.

gkalpak:
LGTM

src/ng/directive/ngOptions.js
Narretz: Yes for Safari :|

gkalpak: Who else? Safari is officially the new IE :stuck_out_tongue:

PR #14786, opened 2016-06-15
perf(ngOptions): don't render options before the first digest

Narretz:
perf

see commit message

see commit message

No

- ~~[ ] Tests for the changes have been added (for bug fixes / features)~~

**Other information**:

Previously, ngOptions would collect and render the options during linking
and then again immediately after the next digest passed, in order to have
the options ready when the select directive tries to render ngModel.
Since we are completely removing the options when they are updated, this means
we are essentially creating each ngOptions select twice after it is compiled.

This fix changes it so that ngOptions creates a noop version of the internal API
so that select can correctly try to render ngModel

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -587,8 +587,14 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
       selectElement.empty();
 
       // We need to do this here to ensure that the options object is defined
-      // when we first hit it in writeNgOptionsValue
-      updateOptions();
+      // when we first hit it in writeNgOptionsValue updateOptions();
+      // The watchCollection will then initalize the actual options and render them
+      options = {
+        getViewValueFromOption: noop,
+        getOptionFromViewValue: noop,
+        selectValueMap: {},
+        items: []
+      };
 
       // We will re-render the option elements if the option values or labels change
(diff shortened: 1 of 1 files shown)


Narretz:
I wanted to change this before, but Pete you had some reservations - maybe this time it'll be okay.

gkalpak:
How confident are we that this is not a BC?

Won't this for example cause the model value to change for one digest?
What if someone is watching for changes?

src/ng/directive/ngOptions.js
```diff
@@ -587,8 +587,14 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
       selectElement.empty();
 
       // We need to do this here to ensure that the options object is defined
-      // when we first hit it in writeNgOptionsValue
-      updateOptions();
+      // when we first hit it in writeNgOptionsValue updateOptions();
```
petebacondarwin: ~~`updateOptions();`~~

petebacondarwin:
Can you actually show that this has a measurable impact on performance?
Without it I would rather leave things as they are

Narretz:
I don't see a measurable impact. Probably because the rendering with documentFragment is very fast now. I mainly found it weird that we render the whole select twice after it has been created (since we aren't reusing options anymore).

PR #14787, opened 2016-06-15
fix(ngTransclude): 'cos things are never simple

petebacondarwin:

fix

broken

fixed (see below)

nope

**Other information**:

### DIFF
--- a/src/ng/directive/ngTransclude.js
+++ b/src/ng/directive/ngTransclude.js
@@ -163,41 +163,56 @@ var ngTranscludeDirective = ['$compile', function($compile) {
   return {
     restrict: 'EAC',
     terminal: true,
-    link: function($scope, $element, $attrs, controller, $transclude) {
-      if ($attrs.ngTransclude === $attrs.$attr.ngTransclude) {
-        // If the attribute is of the form: `ng-transclude="ng-transclude"`
-        // then treat it like the default
-        $attrs.ngTransclude = '';
-      }
+    compile: function ngTranscludeCompile(tElement) {
 
-      function ngTranscludeCloneAttachFn(clone, transcludedScope) {
-        if (clone.length) {
-          $element.empty();
-          $element.append(clone);
-        } else {
-          // Since this is the fallback content rather than the transcluded content,
-          // we compile against the scope we were linked against rather than the transcluded
-          // scope since this is the directive's own content
-          $compile($element.contents())($scope);
+      // Remove and cache any original content to act as a fallback
+      var fallbackLinkFn = $compile(tElement.contents());
+      tElement.empty();
 
-          // There is nothing linked against the transcluded scope since no content was available,
-          // so it should be safe to clean up the generated scope.
-          transcludedScope.$destroy();
+      return function ngTranscludePostLink($scope, $element, $attrs, controller, $transclude) {
+
+        if (!$transclude) {
+          throw ngTranscludeMinErr('orphan',
+          'Illegal use of ngTransclude directive in the template! ' +
+          'No parent directive that requires a transclusion found. ' +
+          'Element: {0}',
+          startingTag($element));
         }
-      }
 
-      if (!$transclude) {
-        throw ngTranscludeMinErr('orphan',
-        'Illegal use of ngTransclude directive in the template! ' +
-        'No parent directive that requires a transclusion found. ' +
-        'Element: {0}',
-        startingTag($element));
-      }
 
-      // If there is no slot name defined or the slot name is not optional
-      // then transclude the slot
-      var slotName = $attrs.ngTransclude || $attrs.ngTranscludeSlot;
-      $transclude(ngTranscludeCloneAttachFn, null, slotName);
+        // If the attribute is of the form: `ng-transclude="ng-transclude"` then treat it like the default
+        if ($attrs.ngTransclude === $attrs.$attr.ngTransclude) {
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngTransclude.js
```diff
@@ -163,41 +163,54 @@ var ngTranscludeDirective = ['$compile', function($compile) {
...
+          // then treat it like the default
+          $attrs.ngTransclude = '';
+        }
 
-          // There is nothing linked against the transcluded scope since no content was available,
-          // so it should be safe to clean up the generated scope.
-          transcludedScope.$destroy();
+        function ngTranscludeCloneAttachFn(clone, transcludedScope) {
+          if (clone.length) {
+            $element.append(clone);
```
gkalpak: Should we also release any fallback contents for GC (e.g. `fallbackContents = null`) ?

test/ng/compileSpec.js
```diff
@@ -9824,37 +9863,6 @@ describe('$compile', function() {
         expect(element.children().eq(2).text()).toEqual('dorothy');
       });
     });
-
-    it('should not overwrite the contents of an `ng-transclude` element, if the matching optional slot is not filled', function() {
```
gkalpak: Why remove this ?

src/ng/directive/ngTransclude.js
```diff
@@ -163,41 +163,54 @@ var ngTranscludeDirective = ['$compile', function($compile) {
   return {
     restrict: 'EAC',
     terminal: true,
-    link: function($scope, $element, $attrs, controller, $transclude) {
-      if ($attrs.ngTransclude === $attrs.$attr.ngTransclude) {
-        // If the attribute is of the form: `ng-transclude="ng-transclude"`
-        // then treat it like the default
-        $attrs.ngTransclude = '';
-      }
+    compile: function ngTranscludeCompile(tElement, tAttrs) {
```
dcherman: `tAttrs` is unused

test/ng/compileSpec.js
petebacondarwin: because it is actually an exact duplicate of the previous (or following) test :-) It must have sneaked in there as part of a rebase in the past

src/ng/directive/ngTransclude.js
petebacondarwin: Ahah! I was using it for debugging. Thanks

src/ng/directive/ngTransclude.js
petebacondarwin: Sounds reasonable

petebacondarwin: Actually this is more complicated. We need to compile this content into a linking function, which we will use to link and clone the fallback content, in the case that this appears inside an ng-repeat.

petebacondarwin:
@dcherman and @gkalpak PTAL

src/ng/directive/ngTransclude.js
```diff
@@ -163,41 +163,56 @@ var ngTranscludeDirective = ['$compile', function($compile) {
...
-    link: function($scope, $element, $attrs, controller, $transclude) {
-      if ($attrs.ngTransclude === $attrs.$attr.ngTransclude) {
-        // If the attribute is of the form: `ng-transclude="ng-transclude"`
-        // then treat it like the default
-        $attrs.ngTransclude = '';
-      }
+    compile: function ngTranscludeCompile(tElement) {
+
+      // Remove and cache any original content to act as a fallback
+      var fallbackLinkFn = $compile(tElement.contents());
```
dcherman: How common do you feel it is to use fallback content and have the user not provide their own?  Wondering if it's worth calling `$compile` the first time the fallback content is needed rather than at compile time, but I have no idea how common this usage is to determine whether or not that'd be worth it.

src/ng/directive/ngTransclude.js
```diff
@@ -163,41 +163,56 @@ var ngTranscludeDirective = ['$compile', function($compile) {
...
+            // so it should be safe to clean up the generated scope.
+            transcludedScope.$destroy();
+          }
+        }
+
+        function useFallbackContent() {
+          // Since this is the fallback content rather than the transcluded content,
+          // we link against the scope of this directive rather than the transcluded scope
+          fallbackLinkFn($scope, function(clone) {
+            $element.append(clone);
```
gkalpak: Doesn't this mean, that if the fallback content requires a parent controller, it won't be able to find it (since we append it to the DOM after it has been linked)?

src/ng/directive/ngTransclude.js
petebacondarwin: This is faster than the situation we had before before your PR, since then we were compiling AND linking the fallback content, even if it was never used.
We could optimize to lazy compile the content, but I would like to see some numbers that show it makes any real difference before doing so.

src/ng/directive/ngTransclude.js
petebacondarwin: Perhaps a suitable unit test?

dcherman: The append is being done in the cloneAttachFn which occurs prior to link, so it should be able to find any required controllers.

gkalpak: Doh!

gkalpak:
LGTM

petebacondarwin:
Thanks @gkalpak - merging

PR #14793, opened 2016-06-16
style(input[number]): use parseFloat with one argument

romkor:
style(input[number]): use `parseFloat` with one argument


romkor:
I signed it!

PR #14794, opened 2016-06-16
fix(*): do not break when modules are used with `angular-loader`

gkalpak:
Bug fix.

When using `angular-loader`, you should be able to load the other scripts in any order. Yet, some of the external modules were assuming that `window.angular` would be available as soon as the script was loaded and executed. If the external module file happened to get loaded before the core angular file, things would fall apart.

When using `angular-loader`, you are be able to load the other scripts in any order.

No


test/ngSanitize/sanitizeSpec.js
```diff
@@ -309,7 +308,6 @@ describe('HTML', function() {
 
   describe('htmlSanitizerWriter', function() {
     /* global htmlSanitizeWriter: false */
-    if (angular.isUndefined(window.htmlSanitizeWriter)) return;
```
petebacondarwin: +1

test/ngSanitize/sanitizeSpec.js
```diff
@@ -19,7 +19,6 @@ describe('HTML', function() {
 
   describe('htmlParser', function() {
     /* global htmlParser */
-    if (angular.isUndefined(window.htmlParser)) return;
```
petebacondarwin: +1

Great that we can remove this kind of thing

src/ngSanitize/sanitize.js
```diff
@@ -187,327 +195,339 @@ function $SanitizeProvider() {
...
+    }
   }
 }
 
+function sanitizeText(chars) {
+  var buf = [];
+  var writer = htmlSanitizeWriter(buf, noop);
+  writer.chars(chars);
+  return buf.join('');
+}
```
petebacondarwin: why is this function outside the closure when the previous function is inside?

src/ngAnimate/shared.js
```diff
@@ -38,7 +24,7 @@ var CSS_PREFIX = '', TRANSITION_PROP, TRANSITIONEND_EVENT, ANIMATION_PROP, ANIMA
 // Also, the only modern browser that uses vendor prefixes for transitions/keyframes is webkit
 // therefore there is no reason to test anymore for other vendor prefixes:
 // http://caniuse.com/#search=transition
-if (isUndefined(window.ontransitionend) && isDefined(window.onwebkittransitionend)) {
+if ((window.ontransitionend === void 0) && (window.onwebkittransitionend !== void 0)) {
```
petebacondarwin: Is this change necessary?
`isUndefined` and `isDefined` are in the closure from `module.js`, right?

src/ngAnimate/shared.js
```diff
@@ -47,7 +33,7 @@ if (isUndefined(window.ontransitionend) && isDefined(window.onwebkittransitionen
   TRANSITIONEND_EVENT = 'transitionend';
 }
 
-if (isUndefined(window.onanimationend) && isDefined(window.onwebkitanimationend)) {
+if ((window.onanimationend === void 0) && (window.onwebkitanimationend !== void 0)) {
```
petebacondarwin: same here

src/ngAnimate/shared.js
```diff
@@ -69,10 +55,6 @@ var ANIMATION_DURATION_PROP = ANIMATION_PROP + DURATION_KEY;
 var TRANSITION_DELAY_PROP = TRANSITION_PROP + DELAY_KEY;
 var TRANSITION_DURATION_PROP = TRANSITION_PROP + DURATION_KEY;
 
-var isPromiseLike = function(p) {
-  return p && p.then ? true : false;
-};
-
```
petebacondarwin: How did you spot this?

karma-modules.conf.js
```diff
@@ -7,7 +7,7 @@ module.exports = function(config) {
   sharedConfig(config, {testName: 'AngularJS: modules', logFile: 'karma-modules.log'});
 
   config.set({
-    files: angularFiles.mergeFilesFor('karmaModules', 'angularSrcModules'),
+    files: angularFiles.mergeFilesFor('karmaModules'),
```
petebacondarwin: +1

Narretz:
We should have tests to ensure this doesn't break later.

src/ngSanitize/sanitize.js
gkalpak: It is used by the linky filter. So it needs to be accessible outside of the closure.

petebacondarwin:
@Narretz I agree. Are we now testing all the modules in a test run where the core angular source files are not loaded?

src/ngAnimate/shared.js
gkalpak: They are initialized in the config phase. Here they are accessed before that. If we move the initialization of the constant inside a provider or config block, then we need to ensure that everything that depends on them doesn't run before they are available.

I opted for the shortest viable fix :smile:

src/ngAnimate/shared.js
gkalpak: I checked all global entries from `.jshistrc` to see in which files they are used and if they are still needed. I removed a few that weren't used any more.
This is only used in test (maybe it used to be used here as well), so I moved it there.

gkalpak:
When we test modules, we incude the bundled `angular.js` file and all module source files.

What @Narretz probably means, is that we should have a test were we use `angular-loader` and force the module files to be loaded before the `angular.js` file and make sure they work.
We don't have such a test (our tests on angular-loader are pretty minimal - just testing the exposed API), that's why these issues sneaked into modules.

I'm not sure how we could force the module files to be loaded before the core `angular.js` file though.
(It might not be worth the trouble, since bundling all JS files seems to be a much more widespread practice than using `angular-loader` nowadays.)

petebacondarwin:
Could we set-up an e2e test that uses the loader?

gkalpak:
I suppose we could (althoug we would probably need some extra dependencies, such as a script loader - angular-seed is using https://github.com/ded/script.js).

The thing is, if the module files aren't loaded (as in downloaded from the server) before the core `angular.js` file, then the tests are useless for this scenario that we want to test.

Maybe we can use the script loader to guarantee the order.

gkalpak:
BTW, this would fix #9140.

petebacondarwin:
What if we create an e2e test that simply loads the files in the "wrong" order:

```html
<script src="angular-loader.js"></script>
<script src="angular-resource.js"></script>
<script src="angular-animate.js"></script>
<script src="angular.js"></script>
```

I think that the browser ensures that these files will run in the order in which they appear in the DOM, right?

gkalpak:
Obviously :smiley:
I'll put together an e2e test.

gkalpak:
I added a commit. The test does indeed break on our current master.

petebacondarwin:
Great! LGTM - please merge

PR #14795, opened 2016-06-17
allow json callback customisation

petebacondarwin:
feat

hard-coded jsonp callback handling

overridable `$jsonpCallbacks` service that offers a place where customisation can be applied

No

**Other information**:

See the commits


test/ng/httpBackendSpec.js
```diff
@@ -310,25 +330,28 @@ describe('$httpBackend', function() {
...
-      callbacks[url[2]]('some-data');
+      $jsonpCallbacks[url[2]]('some-data');
       browserTrigger(script, "load");
 
       expect(callback).toHaveBeenCalledOnce();
     });
 
 
     it('should clean up the callback and remove the script', function() {
+      spyOn($jsonpCallbacks, 'wasCalled').and.callThrough();
```
gkalpak: Is this needed ?

src/ng/jsonpCallbacks.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+/**
+ * @ngdoc service
+ * @name $jsonpCallbacks
+ * @description
```
gkalpak: I'm not sure we are consistent with all services, but could add a `@requires $window`.

src/ng/jsonpCallbacks.js
```diff
@@ -0,0 +1,83 @@
...
+ * Override this service if you wish to customise where the callbacks are stored and
+ * how they vary compared to the requested url.
+ */
+var $jsonpCallbacksProvider = function() {
+  this.$get = ['$window', function($window) {
+    var counter = 0;
+    $window.angular.callbacks = {};
+    var callbackMap = {};
+
+    function createCalback(callbackId) {
```
gkalpak: _createCalback_  -->  _createCallback_

src/ng/jsonpCallbacks.js
```diff
@@ -0,0 +1,83 @@
...
+       * @ngdoc method
+       * @name $jsonpCallbacks#getResponse
+       * @param {string} callbackPath the path to the callback that was sent in the JSONP request
+       * @returns {*} the data received from the response via the registered callback
+       * @description
+       * {@link $httpBackend} calls this method to get hold of the data that was provided to the callback
+       * in the JSONP response.
+       */
+      getResponse: function(callbackPath) {
+        var callback = callbackMap[callbackPath];
```
gkalpak: Nitpick: `return callbackMap[callbackPath].data` would be more consistent with `wasCalled`.

src/ng/jsonpCallbacks.js
```diff
@@ -0,0 +1,83 @@
...
+       * @ngdoc method
+       * @name $jsonpCallbacks#wasCalled
+       * @param {string} callbackPath the path to the callback that was sent in the JSONP request
+       * @returns {boolean} whether the callback has been called, as a result of the JSONP response
+       * @description
+       * {@link $httpBackend} calls this method to find out whether the JSONP response actually called the
+       * callback that was passed in the request.
+       */
+      wasCalled: function(callbackPath) {
+        return callbackMap[callbackPath].called;
```
gkalpak: Since we are deleting the entries, do we need to safeguard against the `callbackPath` property not existing on `callbackMap` ?

test/ng/jsonpCallbacksSpec.js
```diff
@@ -0,0 +1,73 @@
+'use strict';
+
+describe('$jsonpCallbacks', function() {
+  var $mockWindow;
```
gkalpak: Not used.

gkalpak:
A few minor comments  - LGTM otherwise.
IE doesn't seem to share my opinion though :stuck_out_tongue:

petebacondarwin:
I fixed up IE... the problem was that occasionally, in the test, it was actually triggering the script callback at some point in the future, after the test had completed. So I put in a check that the status of the response is a 200 before trying to get hold of the response.

test/ng/jsonpCallbacksSpec.js
petebacondarwin: Thanks

src/ng/jsonpCallbacks.js
petebacondarwin: According to the way that I have written it in `$httpBackend` this should never happen.
If it does then we have a programming error somewhere, which should trigger an error so that we can know to fix it.

src/ng/jsonpCallbacks.js
petebacondarwin: Done

src/ng/jsonpCallbacks.js
petebacondarwin: Thanks

src/ng/jsonpCallbacks.js
petebacondarwin: Done

test/ng/httpBackendSpec.js
petebacondarwin: Nope

petebacondarwin:
@gkalpak thanks for the review. PTAL

src/ng/httpBackend.js
```diff
@@ -59,16 +59,11 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
...
-        callbacks[callbackId].called = true;
-      };
-
-      var jsonpDone = jsonpReq(url.replace('JSON_CALLBACK', 'angular.callbacks.' + callbackId),
-          callbackId, function(status, text) {
-        completeRequest(callback, status, callbacks[callbackId].data, "", text);
-        callbacks[callbackId] = noop;
+      var callbackPath = callbacks.createCallback(url);
+      var jsonpDone = jsonpReq(url, callbackPath, function(status, text) {
+        var response = (status === 200) && callbacks.getResponse(callbackPath);
```
gkalpak: Minor: I would add a note that `status` is set in `jsonpReq()` and it can be either `-1`, `404` or `200`.
The first time I saw this, I thought: "What if it is in the (200, 300) range?"

gkalpak:
LGTM

petebacondarwin:
Thanks. I'll comment on the `status` codes and merge

PR #14799, opened 2016-06-19
feat($compile): attach the containing controller to transcluded scope

petebacondarwin:
This is a POC strawman...
- What are the memory-leak implications of this approach?
- Would it be sufficient to cover the kind of cases that people want to achieve with #14386?


petebacondarwin:
I see that this PR doesn't support the ability to "leak" locals such as `$index` which would be useful if we were using `ngRepeat` to stamp out multiple clones of the transclusion inside the component's template.

PR #14811, opened 2016-06-21
Pr 14656

petebacondarwin:

**Other information**:

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -316,6 +316,39 @@
  *   they are waiting for their template to load asynchronously and their own compilation and linking has been
  *   suspended until that occurs.
  *
+ * ** $doCheck example **
+ *
+ * This example show how you might use `$doCheck` to customise the equality check of component inputs.
+ *
+ * <example name="doCheckExample" module="do-check-module">
+ *   <file name="index.html">
+ *     <div ng-init="items = []">
+ *       <button ng-click="items.push(items.length)">Add Item</button>
+ *       <button ng-click="items = []">Reset Items</button>
+ *       <pre>{{ items }}</pre>
+ *       <test items="items"></test>
+ *     </div>
+ *   </file>
+ *   <file name="app.js">
+ *      angular.module('do-check-module', [])
+ *        .component('test', {
+ *          bindings: { items: '<' },
+ *          template:
+ *            '<pre>{{ $ctrl.log | json }}</pre>',
+ *          controller: function() {
+ *            this.log = [];
+ *
+ *            this.$doCheck = function() {
+ *              if (this.items_ref !== this.items) { this.log.push('doCheck: items changed'); }
+ *              if (!angular.equals(this.items_clone, this.items)) { this.log.push('doCheck: items mutated'); }
+ *
+ *              this.items_clone = angular.copy(this.items);
+ *              this.items_ref = this.items;
+ *            };
+ *          }
+ *        });
+ *   </file>
+ * </example>
  *
  * #### `require`
  * Require another directive and inject its controller as the fourth argument to the linking function. The
@@ -2506,9 +2539,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             }
           }
           if (isFunction(controllerInstance.$doCheck)) {
-            controllerInstance.$doCheck();
-          }
-          if (isFunction(controllerInstance.$doCheck)) {
+            controllerScope.$watch(function() { controllerInstance.$doCheck(); });
             controllerInstance.$doCheck();
           }
           if (isFunction(controllerInstance.$onDestroy)) {
@@ -3275,11 +3306,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         }
       });
 
-      if (isFunction(destination.$doCheck)) {
-        var doCheckWatch = scope.$watch(triggerDoCheckHook);
-        removeWatchCollection.push(doCheckWatch);
-      }
-
       function recordChanges(key, currentValue, previousValue) {
         if (isFunction(destination.$onChanges) && currentValue !== previousValue) {
           // If we have not already scheduled the top level onChangesQueue handler then do so now
@@ -3305,10 +3331,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         destination.$onChanges(changes);
         // Now clear the changes so that we schedule onChanges when more changes arrive
         changes = undefined;
-      }
-
-      function triggerDoCheckHook() {
-        destination.$doCheck();
       }
 
       return {
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -300,6 +300,12 @@
  *   `changesObj` is a hash whose keys are the names of the bound properties that have changed, and the values are an
  *   object of the form `{ currentValue, previousValue, isFirstChange() }`. Use this hook to trigger updates within a
  *   component such as cloning the bound value to prevent accidental mutation of the outer value.
+ * * `$doCheck()` - Called on each turn of the digest cycle. Provides an opportunity to detect and act on
+ *   changes. Any actions that you wish to take in response to the changes that you detect must be
+ *   invoked from this hook; implementing this has no effect on when `$onChanges` is called. For example, this hook
+ *   could be useful if you wish to perform a deep equality check, or to check a Date object, changes to which would not
+ *   be detected by Angular's change detector and thus not trigger `$onChanges`. This hook is invoked with no arguments;
+ *   if detecting changes, you must store the previous value(s) for comparison to the current values.
  * * `$onDestroy()` - Called on a controller when its containing scope is destroyed. Use this hook for releasing
  *   external resources, watches and event handlers. Note that components have their `$onDestroy()` hooks called in
  *   the same order as the `$scope.$broadcast` events are triggered, which is top down. This means that parent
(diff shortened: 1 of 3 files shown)


src/ng/compile.js
```diff
@@ -2499,6 +2538,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               $exceptionHandler(e);
             }
           }
+          if (isFunction(controllerInstance.$doCheck)) {
+            controllerScope.$watch(function() { controllerInstance.$doCheck(); });
```
petebacondarwin: we don't need to manually track the remove watch function because we are after the point when we might swap out the controller for another instance. This watch will naturally be removed when the controller scope is destroyed.

src/ng/compile.js
```diff
@@ -310,6 +316,39 @@
  *   they are waiting for their template to load asynchronously and their own compilation and linking has been
  *   suspended until that occurs.
  *
+ * ** $doCheck example **
+ *
+ * This example show how you might use `$doCheck` to customise the equality check of component inputs.
```
gkalpak: _show_  -->  _shows_

src/ng/compile.js
```diff
@@ -310,6 +316,39 @@
...
+ *          controller: function() {
+ *            this.log = [];
+ *
+ *            this.$doCheck = function() {
+ *              if (this.items_ref !== this.items) { this.log.push('doCheck: items changed'); }
+ *              if (!angular.equals(this.items_clone, this.items)) { this.log.push('doCheck: items mutated'); }
+ *
+ *              this.items_clone = angular.copy(this.items);
+ *              this.items_ref = this.items;
+ *            };
```
gkalpak: How about a simpler (and more lightweight example). We should better not encourage people to use `angular.copy` and `angular.equals` on potentially large arrays of potentially deep or large objects :smiley:

What happened to checking an internal field of a Date object? :smiley:

test/ng/compileSpec.js
```diff
@@ -3828,6 +3828,96 @@ describe('$compile', function() {
...
+          ]);
+        });
+      });
+
+      it('should work if $doCheck is provided in the constructor', function() {
+        var log = [];
+
+        function TestController() {
+          this.$doCheck = function() { log.push('$doCheck'); };
+        	this.$onChanges = function() { log.push('$onChanges'); };
```
gkalpak: <sub>Wrong indentation</sub>

test/ng/compileSpec.js
```diff
@@ -3828,6 +3828,96 @@ describe('$compile', function() {
...
+          $rootScope.$apply('val = 2');
+          expect(log).toEqual([
+            '$doCheck',
+            '$onChanges',
+            '$doCheck'
+          ]);
+        });
+      });
+
+      it('should work if $doCheck is provided in the constructor', function() {
```
gkalpak: I like that we have a test that verifies our incompatibility wiht ng2 :stuck_out_tongue:

src/ng/compile.js
petebacondarwin: This was the lightest thing I could come up with but sure I could give the Date object mutation a go too. (Or are you just ribbing me?)

test/ng/compileSpec.js
petebacondarwin: More ribbing? :-P
It is a common pattern in ng1, where classes are not the be-all-and-end-all to put methods on the instance :-)

src/ng/compile.js
petebacondarwin: thanks

gkalpak:
One thing to note:

In Angular 2, there is no "digest loop" (running until the values have stabilized, which means at least twice per digest cycle) and the change detection is "single pass". Since the change detection works differently in Angular 1, there can't be a 1-to-1 mapping of the behavior of `DoCheck` between the two.

In particular, the ng2 behavior could be mapped to two strategies in ng1:

1. We run `$doCheck()` on every digest loop iteration (i.e. multiple times per digest cycle).
2. We run `$doCheck()` as a post-digest task (i.e. once after the digest cycle is over).
    (This would be similar to how `$onChanges()` is implemented, btw.)

--
This PR uses strategy (1), so currently the order of execution when there is a change is: `$doCheck --> $onChanges --> $doCheck`. In ng2, it will be `$onChanges --> $doCheck` (see [demo][1]).
It is not 100% consistent, but it is as consistent as we can get (within reason).

--
What I am sceptical about, is that with the current ng1 implementation it will be possible to do changes that will be picked up in the current digest cycle and apps will work fine. In ng2, this won't be the case.
You will either get exceptions (in dev mode) or the UI won't reflect the changes initiated through `ngDoCheck()` (see [other demo][2] - comment `enableProdMode();` in `app/main.ts` to change mode).

In that sense, it might be more consistent if we implemented strategy (2) without invoking another digest.

<sub>I am more like thinking load than proposing anything tbh...</sub>

[1]: http://plnkr.co/edit/2HBe7itRRvbtcAZ1w8LW?p=preview
[2]: http://plnkr.co/edit/DACyex0mvGtjFTC4x4tQ?p=preview

petebacondarwin:
Although it is worth noting that in A2 it is not recommended to use both `ngDoCheck` and `ngOnChanges` together.

test/ng/compileSpec.js
gkalpak: No ribbing intended :smiley:
It's a good thing that we have this test. It's not ideal that we are incompatible with ng2, but it's much better that being incompatible with ng1 users and common patterns :wink:

src/ng/compile.js
gkalpak: I was thinking something like:

```html
<button ng-click="myDate.setMonth(myDate.getMonth() + 1)">Postpone for a month</button>
```

```js
this.$doCheck = function() {
  var newValue = this.myDate.valueOf();
  if (newValue !== this.oldValue) { this.log.push('doCheck: we have a new date'); }
  this.oldValue = newValue;
};
```

gkalpak: Another usecase I have in mind for `$doCheck()` is for implementing a poor man's immutability-based change detection:

```js
// In parent:
this.items = [];
this.items._version = 0;

this.addItem = function addItem(item) {
  this.items.push(item);
  this.items._version++;
};

// In child:
this.$doCheck = function $doCheck() {
  if (this.items._version !== this.oldVersion) {
    // `items` has been mutated
  }
  this.oldVersion = this.items._version;
};
```

(Although I am not sure this is a vaid usecase :grin:)

gkalpak:
To be clear, my concern is not about `$onChanges`, it's about the view not being updated with changes initiated from `ngDoCheck` (in ng2).

petebacondarwin:
As an aside here are two demos of A2 ngDoChanges:

Date http://plnkr.co/edit/1d1rlm2efCP31MphBvsf?p=preview
Arrays http://plnkr.co/edit/ZssO7amzxBIg3WL5W1at?p=preview

petebacondarwin:
@gkalpak can you put a unit test (or demo) together to illustrate your concern in A1?

petebacondarwin:
Oh, I just disabled `devMode` and see what you now mean...

petebacondarwin:
But even if we put the changes in `$$postDigest` we will still have a problem right? SInce in dev mode it errors and in ng1 it would just do nothing (or wait till the next digest)

mazzur:
@gkalpak as far as I understood the main constraint is that we should not be able to trigger changes outside the component in its doCheck hook when using A2 while changes inside the component work just as it's implemented in this request. Please correct me if I'm wrong.

mazzur:
> But even if we put the changes in $$postDigest we will still have a problem right? Since in dev mode it errors and in ng1 it would just do nothing (or wait till the next digest)
We can trigger digest on the component scope following the $doCheck call in $$postDigest so that only values from inside the component would be applied right away as it seems to be in A2.

petebacondarwin:
>We can trigger digest on the component scope following the $doCheck call in $$postDigest so that only values from inside the component would be applied right away as it seems to be in A2.

Partial digesting is a worrying topic...

petebacondarwin:
@mazzur I think you are correct in your assumption that we should not trigger changes outside - but I don't think that we can honour this in Angular 1 without doing dastardly stuff, like partial digests

test/ng/compileSpec.js
```diff
@@ -3872,6 +3872,51 @@ describe('$compile', function() {
           ]);
         });
       });
+
+      it('should work if $doCheck is provided in the constructor', function() {
```
zbjornson: This test looks identical to the one I added above?

gkalpak: The test above assigns the methods on the prototype. This test defines them on the instance from inside the constructor.
(Note that while both work in ng1, in ng2 only the prototype will work.)

zbjornson: Aha okay, thanks for explaining :)

petebacondarwin:
So I think that we should go with this implementation but document the following divergences from Angular 2:

* in Angular 2 you cannot define the `ngDoCheck` method inside the constructor, it must be on the prototype
* in Angular 2 you must not use the `ngDoCheck` method to trigger a change outside of the component's template.
* in Angular 1 you may get many more calls to `$doCheck` than you would to `ngDoCheck` in Angular 2 due to the differences in the change detection mechanism

gkalpak:
Except that if we document the 1st deviation (not being on the prototype), we should document it for all lifecycle hooks.

PR #14816, opened 2016-06-22
docs(ngReadonly): added note about input type checkbox and radio

vasekch:

Current docs does not mention, that ngReadonly does work with input type checkbox|radio.

Added note to docs string.

No, text change in comment only.


vasekch:
I signed it!

Narretz:
This doesn't only apply to checkbox and radio - see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly

I'd would rather add a link to this section in the docs, so that we don't have to keep that up to date.

vasekch:
@Narretz Ok, agreed, thanks for processing through.

PR #14821, opened 2016-06-24
feat($resource): deep-merge custom actions with default actions.

gabrielmaldi:

Feature.

If you want to specify a timeout, transform, interceptor, or any other property for an action, you also have to include the other properties of the default action (method, isArray):

```
var Resource = $resource('/foo', {}, { save: { method: 'POST', timeout: 10000 }});
```

Perform a deep-merge between the provided custom actions and the default actions, so you can specify just the properties you want to add or override:

```
var Resource = $resource('/foo', {}, { save: { timeout: 10000 }});
```

Theoretically yes, but should have no impact in practice.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -132,7 +132,7 @@ function shallowClearAndCopy(src, dst) {
  *   Note that the parameter will be ignored, when calling a "GET" action method (i.e. an action
  *   method that does not accept a request body)
  *
- * @param {Object.<Object>=} actions Hash with declaration of custom actions that should extend
+ * @param {Object.<Object>=} actions Hash with declaration of custom actions that should be merged with
  *   the default set of resource actions. The declaration should be created in the format of {@link
  *   ng.$http#usage $http.config}:
  *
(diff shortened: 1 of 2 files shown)


gkalpak:
TBH, I am not a fan of this change. It would save a few characters for a few usecases and add a few characters for a few other usecases. But it would increase the complexity/reduce the declarativeness in all usecases: In order to know what some method of a $resource class/instance does, one would have to look up the default action methods (which might or might not be overriden).

I don't think that the few usecases that would benefit from this PR, justify the extra cost.

--
I will leave it open to see what others think.

gabrielmaldi:
I get your point. What I think is that most devs would expect actions to work this way, which is what happened to me: I didn't understand why setting a timeout on `save` would change it from `POST` to `GET`.

Narretz:
I'm kind of on the fence about this. It does make sense to expect that setting the options preserve the defaults, especially in cases like "save" where it is unlikely that the verb is changed. Then again, it doesn't seem to be a huge problem, as I've never seen this mentioned before.

gabrielmaldi:
Also, if, for instance, you add a custom header to `query`, or specify a timeout, you'd lose `isArray: true` and would have to spend some time figuring out what happened. I agree that it's not a huge problem, but I think that this change may someday save someone some time.

gkalpak:
I would not introduce a breaking change so that I _"may someday save someone some time"_ :stuck_out_tongue: 
(Espacially, since this would mean that I may someday cause someone else to lose some time.)

Narretz:
I agree with @gkalpak that this might cause more problems than it solves. I've updated the documentation to clarify what happens when the custom action has the same key as the default action.

PR #14824, opened 2016-06-25
fix($jsonpCallbacks): do not overwrite callbacks added by other apps

gkalpak:
Bug fix.

When bootstrapping multiple apps on the same page, one could overwrite JSONP callbacks defined by others.

The callback paths are usnique across apps.

No.

**Other information**:
/cc @petebacondarwin

### DIFF
--- a/src/ng/jsonpCallbacks.js
+++ b/src/ng/jsonpCallbacks.js
@@ -11,9 +11,9 @@
  */
 var $jsonpCallbacksProvider = function() {
   this.$get = ['$window', function($window) {
-    var counter = 0;
-    $window.angular.callbacks = {};
     var callbackMap = {};
+    var callbacks = $window.angular.callbacks = $window.angular.callbacks || {};
+    callbacks.$$counter = callbacks.$$counter || 0;
 
     function createCallback(callbackId) {
       var callback = function(data) {
@@ -35,10 +35,10 @@ var $jsonpCallbacksProvider = function() {
        * to pass to the server, which will be used to call the callback with its payload in the JSONP response.
        */
       createCallback: function(url) {
-        var callbackId = '_' + (counter++).toString(36);
+        var callbackId = '_' + (callbacks.$$counter++).toString(36);
         var callbackPath = 'angular.callbacks.' + callbackId;
         var callback = createCallback(callbackId);
-        callbackMap[callbackPath] = $window.angular.callbacks[callbackId] = callback;
+        callbackMap[callbackPath] = callbacks[callbackId] = callback;
         return callbackPath;
       },
       /**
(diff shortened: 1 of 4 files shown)


src/ng/jsonpCallbacks.js
```diff
@@ -11,9 +11,9 @@
  */
 var $jsonpCallbacksProvider = function() {
   this.$get = ['$window', function($window) {
-    var counter = 0;
-    $window.angular.callbacks = {};
     var callbackMap = {};
+    var callbacks = $window.angular.callbacks = $window.angular.callbacks || {};
```
Narretz: Duplicated?

gkalpak: Nope :smiley:

This is basically doing:
```js
foo = foo || {};   // foo <--- $window.angular.callbacks
var bar = foo;   // bar  <--- callbacks
```

The idea is that if `$window.angular.callbacks` has already been initialized (e.g. by another app running on the same page or by a different instance of `$jsonpCallbacks` created by a different injector), it should not be re-initialized/overwritten.

petebacondarwin:
LGTM

petebacondarwin:
I wonder what would happen if someone overrode `$window` in a test without providing `$window.angular`? I guess that would massively break just about everything?

gkalpak:
> Not related specifically to this PR: I wonder what would happen if someone overrode $window in a test without providing $window.angular? I guess that would massively break just about everything?

I am pretty sure it would :smiley: If nothing else, we are using `angular.element.cleanData()` in mocks iirc.

--
BTW, I just found out (accidentally while reviewing an unrelated PR) that `angular.callbacks` is intialized in [AngularPublic.js][1]. So, some of this is not necessary - I'll push a small fixup.

[1]: https://github.com/angular/angular.js/blob/f6c3b357451a2125550811946084fc25ae4eb0ce/src/AngularPublic.js#L153

test/ng/jsonpCallbacksSpec.js
```diff
@@ -2,11 +2,6 @@
 
 describe('$jsonpCallbacks', function() {
 
-  beforeEach(module(function($provide) {
-    // mock out the $window object
-    $provide.value('$window', { angular: {} });
-  }));
```
gkalpak: There is already code in [angular-mocks][1] that cleans up `angular.callbacks`. This isn't necessary.

[1]: https://github.com/angular/angular.js/blob/f6c3b357451a2125550811946084fc25ae4eb0ce/src/ngMock/angular-mocks.js#L2954-L2957

gkalpak:
@petebacondarwin , @Narretz , PTAL.

petebacondarwin:
So you removed your test... is that because it cannot actually test what you wanted to test?

gkalpak:
I didn't remove my test. I remove a `beforeEach` block (that you had added :stuck_out_tongue:).
It wasn't necessary (afaict), but we can leave it by changing the mock window to `{angular: {callbacks: {$$counter: 0}}}`.

petebacondarwin:
OK, so I didn't scroll down far enough. I am sure it wasn't there when I looked a minute ago...

petebacondarwin:
OK, so still LGTM

PR #14825, opened 2016-06-25
docs(guide/interpolation): add known issue (do not change the content of interpolated strings)

### LINKED ISSUE #12813: Can't update DOM-node attr while using {{}} expression
There is a kind of race condition between dom-node attr internal recalculation when we are using {{}}-expression and dom-node attr external modifying. This topic is based on https://github.com/angular/angular.js/issues/12763 issue and also related to https://github.com/angular/angular.js/issues/9109 issue. Also I'd like to thank @gkalpak for he clarified a problem.

----------

Here is a little repro on this issue: http://jsfiddle.net/dhilt/8tceyw2w/9/

I have an element with style-attr contains {{}}-expression:

```
<div id="myElement" style="padding-left: 15px; color:{{toggle ? 'red' : 'black' }}">
```

and I want to play with some style property which isn't presented in the markup:

```
document.getElementById('myElement').style.paddingTop = '20px'; 
```

Let's run the [demo](http://jsfiddle.net/dhilt/8tceyw2w/9/)!

 **1.** Initial state. Before 1st digest ends.

Run the demo. Get the alert before out from $watch.
We see that both of padding-left and padding-top properties are set to 15 and 20 px.
And all of them are applied properly; we can see it despite the UI data is still unbound cause of alert-interruption.

 **2.** Initial state. The end of initial digest.

Close the alert. Let the 1st digest cycle end.
See that padding-top is no longer applied to the element.
Something is broken and initial template's value of style-attr has a destructive priority over further updates on it.

 **3.** Recalculate style expression via $scope.toggle changing.

Click on checkbox, start the new digest cycle.
We see that there is no problem with inline padding-left property.
But padding-top property is completely removed from element's style. Just an empty string.
And `myElementStyle.paddingTop = '20px'` within $watch doesn't work.

----------
.
Thanks and good luck)

gkalpak:
Docs update.

No.

**Other information**:
Closes #12813

### DIFF
--- a/docs/content/guide/interpolation.ngdoc
+++ b/docs/content/guide/interpolation.ngdoc
@@ -106,6 +106,19 @@ can be used with `ngAttr` instead. The following is a list of known problematic 
 - **type** in `<button>` in Internet Explorer 11 (see [issue 14117](https://github.com/angular/angular.js/issues/5025))
 - **value** in `<progress>` in Internet Explorer = 11 (see [issue 7218](https://github.com/angular/angular.js/issues/7218))
 
+## Known Issues
+
+### Dynamically changing an interpolated value
+
+You should avoid dynamically changing the content of an interpolated string (e.g. attribute value
+or text node). Your changes are likely to be overwriten, when the original string gets evaluated.
+This restriction applies to both directly changing the content via JavaScript or indirectly using a
+directive.
+
+For example, you should not use interpolation in the value of the `style` attribute (e.g.
+`style="color: {{ 'orange' }}; font-weight: {{ 'bold' }};"`) **and** at the same time use a
+directive that changes the content of that attributte, such as `ngStyle`.
+
 ### Embedding interpolation markup inside expressions
 
 <div class="alert alert-danger">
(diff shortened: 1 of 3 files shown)


Narretz:
So are known issues things we don't want to fix? To me it sounds like we are aware and maybe gonna fix it.
In the guide I would give the section a more descriptive name or group the section about embedded interpolation together with this one under Known Issues

Narretz:
To clarify, known issue sounds like something that can maybe be fixed, but this here is difficult to fix because it's how the framework works. And we don't recommend a workaround we simply say "don't do this"

gkalpak:
For me, known issues are things that are indeed not working as expected, but are either too difficult to fix or would require a considerable breaking change or for whatever reason there is no plan to fix them any time soon.

I am fine documenting this particular one as a limitation (something that is not supported and should not be used).

Narretz:
I think we are not making a definitive decision here wheter this is considered a bug or simply expected behavior. So let's leave it as is - only thing is, can you
>>> give the section a more descriptive name and group the section about embedded interpolation together with this one under "Known Issues"

?

PR #14833, opened 2016-06-27
feat(ngAria): add support for ignoring a specific element

### LINKED ISSUE #14602: Allow for ngAria to ignore specific elements
It would be great if there was a flag/directive to apply to elements to exclude them from being altered by ngAria. 

We include a non-native element with ng-model (angular-ui/ui-select) which gets a tabindex="0" automatically added by ngAria. Since the actual input is encapsulated within the ng-model div, you'd have to tab twice to get to the element.

### LINKED ISSUE #14672: ngAria should check for interactive child elements before binding role=button and tabindex=0
I've seen this bug in the wild on a large web application using ngAria: `ng-click` tries to help accessibility by binding `role="button"` and `tabindex="0"`, unless opted out. The problem is if that DIV with an `ng-click` wraps another interactive element such as `<button>` or `<a>`, the extra `tabindex` and `role` cause problems.

**Current behavior**

 `tabindex="0"` on everything causes keyboard hell with often-redundant tab stops; and the `role=button` wrapping other content confuses screen reader users. 

Here is a Plunker showing the wrapping interactive elements, prompted from #14665: https://plnkr.co/edit/SPxQ0X2RQvvfvAmDVTNA?p=preview

If we can do it fast enough, ngAria should test an `ng-click` target's child elements against the nodeBlacklist and return false on the first instance, leaving off `tabindex="0"` and `role="button"`. The DIV with `ng-click` will then stay a mouse-only element to prevent degrading accessibility in the rest of that DOM subtree.

ngAria currently causes more problems than it solves with "fixes" like the one `ng-click` tried to solve. Addressing this problem would improve the utility of the library.

Visible in the latest version.

gkalpak:
Feature

It is not possible to force `ngAria` to "ignore" a specific element, i.e. not add `aria-*` attributes, attach `roles`, bind keyboard events etc. The only work-around is to provide those attributes yourself (in which case `ngAria` won't overwrite them), but this is tedious and doesn't work for all usecases.

It is possible to force `ngAria` to "ignore" a specific element, by adding the `ng-aria-disable` attribute.

No.

**Other information**:
Fixes #14602
Fixes #14672

### DIFF
--- a/src/ngAria/aria.js
+++ b/src/ngAria/aria.js
@@ -14,8 +14,8 @@
  *
  * For ngAria to do its magic, simply include the module `ngAria` as a dependency. The following
  * directives are supported:
- * `ngModel`, `ngChecked`, `ngReadonly`, `ngRequired`, `ngValue`, `ngDisabled`, `ngShow`, `ngHide`, `ngClick`,
- * `ngDblClick`, and `ngMessages`.
+ * `ngModel`, `ngChecked`, `ngReadonly`, `ngRequired`, `ngValue`, `ngDisabled`, `ngShow`, `ngHide`,
+ * `ngClick`, `ngDblClick`, and `ngMessages`.
  *
  * Below is a more detailed breakdown of the attributes handled by ngAria:
  *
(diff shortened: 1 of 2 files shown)


gkalpak:
/cc @marcysutton 

Notes:

- The current implementation will only ignore the specific element, not the whole subtree. I thought about adding support for ignoring the whole subtree, but:
  1. The implementation would be more involved (e.g. we might need to traverse up the DOM looking for an attribute or we could have a directive and require its controller from all `ngAria` directives).
  2. Unless there is a strong (valid) usecase, we should not promote (or make it easy) to disable `ngAria` in large parts of the DOM.

- A possible future enhancement is to disable specific features. E.g. `ng-aria-disable="ariaHidden, bindRoleForClick, tabindex"`.

marcysutton:
I like this idea; it would also help to address the config being global in https://github.com/angular/angular.js/issues/11174. 

It makes sense that disabling would only work for that specific node...crawling DOM subtrees would be too expensive. I like the future enhancement of disabling certain features, that was my one question after reviewing the PR.

Narretz (approved):

gkalpak:
Rebased. (AFAICT, nothing has been added to `ngAria` that is not accounted for.)

Splaktar:
Just in case anyone (like me) is looking for when this feature was released, it was first available in `1.7.1`.

PR #8841, opened 2014-08-29
fix($resource): provide HTTP status to the success

SekibOmazic:
provide status code to the success callback

fixes #8341

PR #14836, opened 2016-06-28
feat($resource): pass `status`/`statusText` to success callbacks

### LINKED ISSUE #8341: Feature request: provide HTTP status to the success callback of $resource
When a $resource request fails, you get the response object passed to the error callback, and you can retrieve that HTTP status from that. When the request succeeds, however, the success callback is given only the payload and the headers.

It's often important to be able to distinguish between different success cases using the HTTP response code. It's not uncommon to have POST requests that either create and return a new resource (HTTP 201) or find and return an existing one that will fill the present need (HTTP 200). Or, say, the difference between a 201 ("your resource was created, and here it is") vs. a 202 ("we'll try to create this resource; check back later").

gkalpak:
Feature.

Only the data/instance and the headers function are passed to success callbacks.

The `status` and `statusText` are also passed to the success callbacks.

No. Yes. I don't know. Could be...in tests...if someone verifies what arguments their callbacks are called with...

**Other information**:
Fixes #8341
Closes #8841

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -238,9 +238,9 @@ function shallowClearAndCopy(src, dst) {
  *   - non-GET instance actions:  `instance.$action([parameters], [success], [error])`
  *
  *
- *   Success callback is called with (value, responseHeaders) arguments, where the value is
- *   the populated resource instance or collection object. The error callback is called
- *   with (httpResponse) argument.
+ *   Success callback is called with (value (Object|Array), responseHeaders (Function),
+ *   status (number), statusText (string)) arguments, where the value is the populated resource
+ *   instance or collection object. The error callback is called with (httpResponse) argument.
  *
  *   Class actions return empty instance (with additional properties below).
  *   Instance actions return promise of the action.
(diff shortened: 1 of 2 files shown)


gkalpak:
Another idea (but a slightly bigger breaking change) would be to pass the whole `response` object as a 2nd argument and let the user get what they want from there.

Considering the potential BC (mainly in tests), I suggest we land this on 1.6.x only. Should I add a notice (i.e. is it a real BC)?

Narretz:
What things other than status & statusText could be realistically required from the response? On one hand, the more generic version is better because the API won't get ugly when you have to add it later anyway, but on the other hand, if no other things but status & statusText are used, then we should stick to them.

gkalpak:
I can't think of anything else - just future-proofing :smile: 
I agree - let's avoid a bigger breaking change for now. If it turns out people need more control, we can always make the change at later version.

Narretz:
I've merged it in into master and I'm happy to merge it into 1.5.x, as breaking in tests isn't a BC to me.

gkalpak:
I am fine merging it into 1.5.x too.

PR #14839, opened 2016-06-28
fix($resource): fulfill promise with the correct value on error

### LINKED ISSUE #14837: bug($resource): promise not resolved with value returned by `responseErrorInterceptor`
Bug.

Since 71cf28c (landed in master only), when there is either a `responseError` interceptor, the promise returned by a `$resource` instance action (or exposed as `instance.$promise`) is always resolved with the original response, instead of being fulfilled (resolved or rejected) with the value returned by the interceptor.

[Demo][1] (running with `1.5.7` - change versions to `snapshot` to see the issue)
- `1.5.7`: The promise is rejected with the value returned by the `responseError` interceptor.
- `snapshot`: The promise is resolved (instead of rejected) with the original response (instead of the value returned by the `responseError` interceptor).

Behavior on `snapshot` (eventual 1.6.x) should be the same as on `1.5.x`.

Fix a regression.

Works on `1.5.x`. Broken on `snapshot` (eventual 1.6.x).

[1]: https://codepen.io/ExpertSystem/pen/mEWpEL?editors=0012

gkalpak:
Bug fix.

See #14837.

Restores the 1.5.x behavior.

No, since it fixes (yet unreleased) regression.

**Other information**:
Fixes #14837

The first commit (which only contains a couple of tests) could be backported to 1.5.x, in order to verify that the behavior is correct on that branch.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -787,7 +787,8 @@ angular.module('ngResource', ['ng']).
               } :
               undefined);
             if (hasError && !hasResponseErrorInterceptor) {
-              // Avoid PUR, but still reject the returned promise
+              // Avoid `Possibly Unhandled Rejection` error,
+              // but still fulfill the returned promise with a rejection
               promise.catch(noop);
             }
 
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -695,6 +695,8 @@ angular.module('ngResource', ['ng']).
               defaultResponseInterceptor;
             var responseErrorInterceptor = action.interceptor && action.interceptor.responseError ||
               undefined;
+            var hasError = !!error;
+            var hasResponseErrorInterceptor = !!responseErrorInterceptor;
             var timeoutDeferred;
             var numericTimeoutPromise;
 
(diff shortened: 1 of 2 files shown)


src/ngResource/resource.js
```diff
@@ -776,13 +778,18 @@ angular.module('ngResource', ['ng']).
                 (success || noop)(value, response.headers);
                 return value;
               },
-              responseErrorInterceptor || error ?
+              (hasError || hasResponseErrorInterceptor) ?
```
Narretz: Can we use a real `if` here? The whole block is pretty hard to read (for me)

src/ngResource/resource.js
```diff
@@ -776,13 +778,18 @@ angular.module('ngResource', ['ng']).
...
-              }
-              : undefined);
+                if (hasError) error(response);
+                return hasResponseErrorInterceptor ?
+                    responseErrorInterceptor(response) :
+                    $q.reject(response);
+              } :
+              undefined);
+            if (hasError && !hasResponseErrorInterceptor) {
+              // Avoid PUR, but still reject the returned promise
```
Narretz: PUR?

gkalpak: You know. **P**ossibly **U**nhandled **R**ejection :stuck_out_tongue:

petebacondarwin: false economy :-P

gkalpak: Fixed.

PR #14845, opened 2016-06-29
fix(copy): fix typed subarrays handling

### LINKED ISSUE #14842: angular.copy doesn't copy typed subarrays
Regression bug:

It copies the whole buffer, not a slice.
Here is the demo: http://plnkr.co/edit/nU76OITTchFc48jFab2a?p=preview

When copying typed subarray expected behavior is to return the value of the same length. 

I use a big byte buffer for the whole server API response and I make slices (via subarrays) in it for response object fields when deserializing. Then I need to copy object with fields to use at another place. With current behavior new copy is invalid: all byte fields have wrong contents.

I can reproduce the bug in v1.5.7. It worked fine in previous angular version I used: v1.4.12

I think the bug was introduced in this commit:
https://github.com/angular/angular.js/commit/7b51243be597900b1f765495dadfea5fccd2228e
#13641

I think it doesn't work as expected because of this code:

```javascript
  function copyType(source) {
    switch (toString.call(source)) {
      case '[object Int8Array]':
      case '[object Int16Array]':
      case '[object Int32Array]':
      case '[object Float32Array]':
      case '[object Float64Array]':
      case '[object Uint8Array]':
      case '[object Uint8ClampedArray]':
      case '[object Uint16Array]':
      case '[object Uint32Array]':
        return new source.constructor(copyElement(source.buffer));
```

it always copies the whole buffer, even if the view has shorter byte length. We can check that `source` is subarray by checking e.g. `byteOffset` value e.g.

In v1.4.12 the code was like this:

```javascript
    if (isArray(source)) {
      destination = [];
      needsRecurse = true;
    } else if (isTypedArray(source)) {
      destination = new source.constructor(source);
```

zhukov:

Bug fix

It copies the whole buffer, not a slice.
Here is the demo: http://plnkr.co/edit/nU76OITTchFc48jFab2a?p=preview
#14842

When copying typed subarray the new value has the same length as the initial typed array.
This is not a feature change.

No

**Other information**:

Previously, it would return copy of the whole original typed array, not its slice.
Now buffer offset and length are also saved.

Closes #14842

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -901,7 +901,7 @@ function copy(source, destination) {
       case '[object Uint8ClampedArray]':
       case '[object Uint16Array]':
       case '[object Uint32Array]':
-        return new source.constructor(copyElement(source.buffer));
+        return new source.constructor(copyElement(source.buffer), source.byteOffset, source.length);
 
       case '[object ArrayBuffer]':
         //Support: IE10
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -240,13 +240,13 @@ describe('angular', function() {
       }
     });
 
-    it("should handle Uint8Array subarray", function() {
-      if (typeof Uint8Array !== 'undefined') {
-        var arr = new Uint8Array(4);
+    it("should handle Uint16Array subarray", function() {
+      if (typeof Uint16Array !== 'undefined') {
+        var arr = new Uint16Array(4);
         arr[1] = 1;
         var src = arr.subarray(1, 2);
         var dst = copy(src);
-        expect(dst instanceof Uint8Array).toBeTruthy();
+        expect(dst instanceof Uint16Array).toBeTruthy();
         expect(dst.length).toEqual(1);
         expect(dst[0]).toEqual(1);
         expect(dst).not.toBe(src);
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -240,6 +240,20 @@ describe('angular', function() {
       }
     });
 
+    it("should handle Uint8Array subarray", function() {
+      if (typeof Uint8Array !== 'undefined') {
+        var arr = new Uint8Array(4);
+        arr[1] = 1;
+        var src = arr.subarray(1, 2);
+        var dst = copy(src);
+        expect(dst instanceof Uint8Array).toBeTruthy();
+        expect(dst.length).toEqual(1);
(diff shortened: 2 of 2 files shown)


src/Angular.js
```diff
@@ -901,7 +901,7 @@ function copy(source, destination) {
       case '[object Uint8ClampedArray]':
       case '[object Uint16Array]':
       case '[object Uint32Array]':
-        return new source.constructor(copyElement(source.buffer));
+        return new source.constructor(copyElement(source.buffer), source.byteOffset, source.length);
```
gkalpak: Is `length` the same as `byteLength` (for non-uint8 arrays)?

zhukov: @gkalpak nope, `source.byteLength = source.length * source.BYTES_PER_ELEMENT`.
I've made some sample code to demo why `source.length` and not `source.byteLength`:
https://plnkr.co/edit/afWbk8OboJopB6BgSri7

test/AngularSpec.js
```diff
@@ -240,6 +240,20 @@ describe('angular', function() {
       }
     });
 
+    it("should handle Uint8Array subarray", function() {
```
gkalpak: Can you use a different type, to make the distinction between `source.length` and `source.byteLength` more obvious?

gkalpak:
Other than one minor comment, LGTM :+1:

test/AngularSpec.js
zhukov: Do you mean a different test? Or which type?

zhukov: @gkalpak Do you mean a different test? Or which type?

gkalpak: In this test, use some type that doesn't have `BYTES_PER_ELEMENT === 1` (as does `Uint8Array`), so that we verify that everything works correctly. E.g. use `Uint16Array`.

zhukov: @gkalpak actually it was already tested by existing cases like `should copy a Int16Array with no destination`. And this is the reason I changed byteLength to length in the initial PR.
Changed the new test anyway, it will be more obvious indeed.

zhukov:
Hm.. no idea, what should I fix to make it pass. Works fine on my branch.

gkalpak:
@zhukov , it is probably a flake. I restarted the job. Thx for the changes :+1:

PR #14848, opened 2016-06-30
perf($compile): wrap try/catch of collect comment directives into a function to avoid V8 deopt

drpicox:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)** its a performance optimization. It turns out that try/catch statements prevents the optimization of the function collectDirectives, which has a great weight in the initial document parse.

It should be around a 5% faster in the angular initial document parse, and also in directives compilation.

There is no change in the functionality.

No

**Other information**:
See: https://github.com/GoogleChrome/devtools-docs/issues/53

I did all performance checking over existing code from my customers, in their code this optimization makes it go a **5% faster in the initial parse time and compile time** in average (I just performed 200 evaluation comparison with execution times around 50 - 100ms of each evaluation). I did the performance test over chrome, safari and firefox, all of them with similar speedup results.
I am not sure which environment do you use to check performance (I think that it is probably related to benchpress). ¿Do you have a guide about how to set up and do testing?

I found the problem while doing performance optimization in a customer code. Just clicking around I found this message lots of times:
screen shot 2016-06-29 at 21 44 27

screen shot 2016-06-29 at 21 46 24

It should solve it, because in production environments usually there are no comments, so, this function is likely to be never executed.

I did no new documentation neither new tests because nothing changes but performance.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2052,24 +2052,30 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           addTextInterpolateDirective(directives, node.nodeValue);
           break;
         case NODE_TYPE_COMMENT: /* Comment */
-          try {
-            match = COMMENT_DIRECTIVE_REGEXP.exec(node.nodeValue);
-            if (match) {
-              nName = directiveNormalize(match[1]);
-              if (addDirective(directives, nName, 'M', maxPriority, ignoreDirective)) {
-                attrs[nName] = trim(match[2]);
-              }
-            }
-          } catch (e) {
-            // turns out that under some circumstances IE9 throws errors when one attempts to read
-            // comment's node value.
-            // Just ignore it and continue. (Can't seem to reproduce in test case.)
-          }
+          collectCommentDirectives();
           break;
       }
 
       directives.sort(byPriority);
       return directives;
+
+      function collectCommentDirectives() {
+        // function created because of performance, try/catch disables
+        // the optimization of the whole function #14848
+        try {
+          match = COMMENT_DIRECTIVE_REGEXP.exec(node.nodeValue);
+          if (match) {
+            nName = directiveNormalize(match[1]);
+            if (addDirective(directives, nName, 'M', maxPriority, ignoreDirective)) {
+              attrs[nName] = trim(match[2]);
+            }
+          }
+        } catch (e) {
+          // turns out that under some circumstances IE9 throws errors when one attempts to read
+          // comment's node value.
+          // Just ignore it and continue. (Can't seem to reproduce in test case.)
+        }
+      }
     }
 
     /**
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2052,29 +2052,29 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           addTextInterpolateDirective(directives, node.nodeValue);
           break;
         case NODE_TYPE_COMMENT: /* Comment */
-          collectCommentDirectives();
+          collectCommentDirectives(node, directives, attrs, maxPriority, ignoreDirective);
           break;
       }
 
       directives.sort(byPriority);
       return directives;
+    }
 
-      function collectCommentDirectives() {
-        // function created because of performance, try/catch disables
-        // the optimization of the whole function #14848
-        try {
-          match = COMMENT_DIRECTIVE_REGEXP.exec(node.nodeValue);
-          if (match) {
-            nName = directiveNormalize(match[1]);
-            if (addDirective(directives, nName, 'M', maxPriority, ignoreDirective)) {
-              attrs[nName] = trim(match[2]);
-            }
+    function collectCommentDirectives(node, directives, attrs, maxPriority, ignoreDirective) {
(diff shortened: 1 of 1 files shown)


Narretz:
Great PR, thanks for this. Could you add a comment to the function that indicates that this is for perf reasons?
We have benchpress, but it's very coarse atm, so I don't know if it'll show usable numbers. If you cloned and build angular, you can run the included benchmarks like this: 

>>> There is a configured grunt task for building the benchmarks,
`grunt bp_build`, which places the runnable benchmarks in "/build/benchmarks/".
The existing `grunt webserver` task can be used to serve the built benchmarks at `localhost:8000/build/benchmarks/<benchmark-name>`

Funnily enough, the try-catch was introduced because there was a not reproducible error in IE9 ...

drpicox:
Ok! I'll add the comment right now.
Thanks for howto about benchpress, I hope I can try it later.

src/ng/compile.js
```diff
@@ -2052,24 +2052,30 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
-            // Just ignore it and continue. (Can't seem to reproduce in test case.)
-          }
+          collectCommentDirectives();
           break;
       }
 
       directives.sort(byPriority);
       return directives;
+
+      function collectCommentDirectives() {
```
dcherman: Rather than extracting the entire case to a function, you may want to consider extracting just the bit that might throw (accessing `nodeValue`) and putting that function as a sibling to the `collectDirectives` function.

Reason being is that since `collectDirectives` is invoked once for every element which is often in the thousands (1558 on a quick sample I took from one app that I work on), we can avoid creating thousands of additional functions that need to be GCed.

Narretz: Isn't this something that an interpreter can optimize as it knows that this function is always used the same, only with a different context?

jbedard: The interpreter still has to create a new closure and function instance for each and every one though (must ensure two of these `!=` each other etc.).

I'd definitely vote to avoid recreating the same function per node...

drpicox: Ok,

I think that comments from @dcherman and @jbedard are not completely true. Although, may be I have some reasonable doubts, but I expose the following:

- functions are statically parsed in the beginning of the execution: it does not matter if they are nested or whatever. I'm quite sure that it is being doing by this way because even the old Modula II did that,

- There is no new Function(bodyString) is executed neither eval("function() {...}"): so, I do not see where the compiler creates the new function instance for each and every one though

- Closures are a old concepts that can be optimized easily: In this case the compiler can analyse statically the code, it can see that collectCommentDirectives has no other use outside collectDirectives, it does not define any new variable, it does not has arguments, so there is no reason to not reuse the stack from collectDirectives. It looks a lot like C nested functions, so it is possible to the compiler to apply the same optimization techniques. I'm not sure if compilers are doing it, but it is a very simple optimisation, it can be applied in many different situations, so it is very likely that it might be already implemented

- this function is only called when comments are presented: in the worst case that there are no optimizations, the closure creation will only happen when comments are parsed. It turns out that in production environments in which people is concerned by performance, there should be no comments at all (they are a waste of bandwidth and parse time)

In addition, the strong point of having this implemented by a nested function is to make it easier to understand, and easier to maintain: the implementation is close to where it is called, it is the same code that was before, it uses the same variables and structure that the parent function. And basically this function, is only a protection, has no other kind of meaning.

Any way,

- I'll try to do some performance benchmarks about this case with synthetic cases (it is a pity that jsperf is offline) just to make sure that there is no a performance problem (in my tests in makes no difference to have this function in this level or in another level with clausure variables passed by value, but it is almost not called)

- If it is just me that thinks that it is easier to understand and maintain implemented by this way, I can change it, no problem

dcherman: So I'll retract my statement, and you can revert the commit if you'd like.

http://webcache.googleusercontent.com/search?q=cache:9UdL1nWH_xsJ:ariya.ofilabs.com/2012/07/lazy-parsing-in-javascript-engines.html+&cd=1&hl=en&ct=clnk&gl=us

All browsers at least back to IE9 seem to defer parsing of functions until necessary, and memory allocations did not seem to occur in the tests that I've done in Chrome, FF, or IE11.

About pre-mature optimization; the compile and linking phases are some of the hottest areas in Angular, so I think it's more important to consider the implications of any given change here versus tinkering with something like `$http` where optimization is way less likely to bear measurable fruit.  Worst case scenario is you learn something about how browsers work internally as we did here, and that's never a bad thing.  In this case, I suggested the change initially because it's semantically identical to the original PR without any additional complexity or significant restructuring, but is more obviously correct to someone that was not aware of engine optimizations (guilty as charged here).

If you want to look into something similar, I've been meaning to make a similar change in `$digest` to eliminate the deopts there and profile to see if it speeds up the digest cycle at all.  I suspect that

petebacondarwin:
LGTM

drpicox:
So, it is ready for merge then.

petebacondarwin:
Yep - will do on Monday if no one gets to it before me

gkalpak:
Too low hanging a fruit - sorry :stuck_out_tongue:

PR #14852, opened 2016-06-30
docs($templateCache): let's get rid of ng-include not to create bad example

stevermeister:

### DIFF
--- a/src/ng/cacheFactory.js
+++ b/src/ng/cacheFactory.js
@@ -382,13 +382,7 @@ function $CacheFactoryProvider() {
  *   $templateCache.put('templateId.html', 'This is the content of the template');
  * });
  * ```
- *
- * To retrieve the template later, simply use it in your HTML:
- * ```html
- * <div ng-include=" 'templateId.html' "></div>
- * ```
- *
- * or get it via Javascript:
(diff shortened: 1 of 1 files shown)


Narretz:
Can you explain why this is a bad example?

stevermeister:
@Narretz 
"ng-include" first thing that you need to remove from your project if you aim to have good component approach. 

http://teropa.info/blog/2015/10/18/refactoring-angular-apps-to-components.html#replace-ng-include-with-component-directive

petebacondarwin:
How about we rework the order and emphasis here, rather than just deleting an outdated but perfectly valid example.

For instance, we could have three examples:

* use the template from a component directive
* access the template programmatically
* use the template via ng-include (perhaps with a comment that the recommended approach is to use components rather than including templates for most scenarios)

stevermeister:
@petebacondarwin 
good point, will add it

Narretz:
@stevermeister I'm gonna close this PR, but please feel free to reopen it when you have amended it based on Pete's ideas.

stevermeister:
ok, created new one with component example - https://github.com/angular/angular.js/pull/15120

PR #14856, opened 2016-07-01
Extend with customized reduce function

jl2922:

feature

'extend' will use new value to override original value

by using 'extendWith', user can specify the new value assigned with a customized function of both the destination value and source value.
The test case provides an example where concat is used as an customized function for array like properties. Without the if statement, extendWith will behave the same as extend.

no

**Other information**:

### DIFF
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -36,6 +36,7 @@
     "nextUid": false,
     "setHashKey": false,
     "extend": false,
+    "extendWith": false,
     "toInt": false,
     "inherit": false,
     "merge": false,
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -27,6 +27,7 @@
   nextUid: true,
   setHashKey: true,
   extend: true,
+  extendWith: true,
   toInt: true,
   inherit: true,
   merge: true,
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -324,7 +324,7 @@ function setHashKey(obj, h) {
 }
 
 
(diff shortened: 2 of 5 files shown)


jl2922:
I am a googler, LDAP: junhao

Narretz:
Thanks for the PR. However, we usually only add angular.* helper functions to the core if the core actually benefits from it. This looks like a function for a very specific use case. So I don't think it should be added to core.

gkalpak:
I agree that this doesn't belong in core.

<sub>(But IF we decided to get it in, we should probably implement all version (i.e. `extend`, `merge`, `extendWith`) with `reducers`. There is no reason for `baseExtend` to take both a `deep` **and** a `reducer` argument.)</sub>

Narretz:
Closing as we have two votes against and no response from @jl2922

jl2922:
Sorry for the late reply. I don't quite understand why you say it is a very specific use case. I feel it is common that instead of always using the source value to overwrite the destination value, it gives user an option to use value computed from both the source and destination, isn't it:)

This functionality will make the code written in Angular more organized and easy to understand than everyone giving up angular.extend and writing their own extend function when facing with this need.

Both extendWith and mergeWith like functions are already in lodash and are about to be added in underscore. My team is using AngularJS, so I feel it would be best that this could be in our own product. 

And if we agree upon the fact that deep and customized reducer can benefit users, I don't see any reason why combining them is not a general use case and why we don't try to extend the baseExtend but to create another baseExtendWith that contains all the codes in baseExtend.

PR #14859, opened 2016-07-03
Performance: disable css and comment CM directives by default

drpicox:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)** Performance/Feature.

**What is the current behavior? (You can also link to an open issue here)**: By default it compiles css and comment directives, and it has a downside in performance.

**What is the new behavior (if this is a feature change)?**: It stops compiling comment and css directives by default, but it can be configured to turn them on again.

**Does this PR introduce a breaking change?** YES.

**BREAKING CHANGE**: directives inside class attributes or comments will stop working, by default the compiler will ignore them.

To migrate: we recommend stop using CM attributes and use EA instead,

Before:

```
    restrict: 'EACM'

    <div class="my-directive">...</div>
```

After:

```
    restrict: 'EA',

    <my-directive>...</my-directive>
```

Css and comment attributes can be compiled again through $compileProvider but loosing performance, ex:

```
    app.config(function($compileProvider) {
      $compileProvider.commentDirectivesEnabled(true);
      $compileProvider.cssDirectivesEnabled(true);
    });
```

It is possible to activate only of both:

```
    app.config(function($compileProvider) {
      $compileProvider.cssDirectivesEnabled(true);
    });
```

**Other information**:

It is related to https://github.com/angular/angular.js/pull/14850 and https://github.com/angular/angular.js/pull/14850#issuecomment-229651776 

I did this PR thinking in 1.6.x or future releases, but also thinking in test better the #14850 . When the behavior changes by default it affects to all tests, so, it is a good opportunity to review case by case and see if everything works as expected.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1386,7 +1386,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     return TTL;
   };
 
-  var ignoreMDirectives = false;
+  var ignoreMDirectives = true;
   /**
    * @ngdoc method
    * @name $compileProvider#commentDirectivesEnabled
@@ -1395,16 +1395,16 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    * It indicates to the compiler if the current project
    * does require to use comment directives.
    *
-   * If comment directives are disabled
+   * By default comment directives are disabled,
    * it allows to the compiler to make optimizations
    * and skip comments parsing.
    *
-   * Default value is ‘true’
+   * Default value is false
    *
    * Example:
    *
    * ```
-   * $compileProvider.commentDirectivesEnabled(false);
+   * $compileProvider.commentDirectivesEnabled(true);
    * ```
    *
    * @param {boolean} false if the compiler may ignore comment directives
@@ -1414,7 +1414,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
   };
 
 
-  var ignoreCDirectives = false;
+  var ignoreCDirectives = true;
   /**
    * @ngdoc method
    * @name $compileProvider#commentDirectivesEnabled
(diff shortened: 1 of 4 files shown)


Narretz:
I think class directives are more common than you'd think. So I'm not sure if it's the right thing to disable them by default, even if it is just for 1.6. I also don't really understand how it makes it easier to test if they are disabled by default.

drpicox:
About test better is to test better https://github.com/angular/angular.js/pull/14850 , I explain:

- In the #14850 there is a flag that turns off comment and class directives
- I have tested it with few tests, but it will affect to an application as global
- In this PR I have disabled them by default, so all tests runs like an app would run disabling those directives
- When I ran tests, many of them has failed (as I have anticipated)
- So, to verify that #14850 I have:
  - revised test by test manually
  - activate the css or comment directives in which tests are failing
  - verify that it works
  - verify that it really requires css or comment directives to make the test

So thanks to this PR I have a strong security that #14850 will work as expected.


About css directives, the reason why css and comment directives can be enabled or disabled independently is because that I knew cases that css directives are required and comment are not.
In #14850 I have listed some of the cases. But, from my point of view it is not really common to use CSS. 
To decide if it is ok to enable them or not by default I would want to consider:
- Speed improvement of disabling css directives in a common app
- Percentage of apps already using css directives

Maybe it should better make the change in 1.7.x and give people to learn about performance optimizations and stop using css directives (like I have stopped using element.scope()).

But if it is not merged, it's ok, it worked well to test #14850.

Narretz:
I think it's simpler just to have a series of tests that expect CSS class and comment directives to be enabled by default, i.e. in the series of tests in #14850, add an entry that doesn't specify and set the enabled status, but simply assumes that both are enabled by default

drpicox:
It make sense. I didn't wanted to change orders and start messing around.

Narretz:
Ok, can we then continue work on PR #14850 - with the changes I suggestd. I think it also needs a rebase, because the comment compilation fn change has already been merged

PR #14864, opened 2016-07-04
fix(select): don't register options when select has no ngModel

Narretz:
bug fix

Options are added even if select has no ngModel. ngValue replaces the value attribute of the element with a hashed value, and this might break tests (it does in angular material)

Options are not added to select

?
I can't think of a valid use case where you'd want registerning options without ngModel integration. If you don't have model binding, then you don't need to store the options internally. You can actually use ngOptions without ngModel to render a list of options - but you don't need to register the options for that.

The registerOption function is also not documented, so this is an implementation detail from the perspective of the user

**Other information**:

When option elements use ngValue, value or interpolated text simply to set
the option value, i.e. when the parent select doesn't have an ngModel,
there is no necessity in registering the options with the select controller.
This was previously no problem, as the ngModelController is set to a noop controller,
so that all further interactions are aborted ($render etc)

However, ngValue sets a hashed value inside the option value (to support arbitrary object types).
This can cause issues with tests that expect unhashed values.

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -519,11 +519,16 @@ var selectDirective = function() {
 
   function selectPreLink(scope, element, attr, ctrls) {
 
-      // if ngModel is not defined, we don't need to do anything
-      var ngModelCtrl = ctrls[1];
-      if (!ngModelCtrl) return;
-
       var selectCtrl = ctrls[0];
+      var ngModelCtrl = ctrls[1];
+
+      // if ngModel is not defined, we don't need to do anything but set the registerOption
+      // function to noop, so options don't get added internally
+      if (!ngModelCtrl) {
+        selectCtrl.registerOption = noop;
+        return;
+      }
+
 
       selectCtrl.ngModelCtrl = ngModelCtrl;
 
--- a/test/ng/directive/selectSpec.js
+++ b/test/ng/directive/selectSpec.js
@@ -136,6 +136,30 @@ describe('select', function() {
 
   });
 
+  it('should not add options to the select if ngModel is not present', inject(function($rootScope) {
+    var scope = $rootScope;
+    scope.d = 'd';
+    scope.e = 'e';
+    scope.f = 'f';
+
+    compile('<select>' +
+      '<option ng-value="\'a\'">alabel</option>' +
+      '<option value="b">blabel</option>' +
+      '<option >c</option>' +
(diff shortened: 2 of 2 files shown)


gkalpak:
WRT _"You can actually use ngOptions without ngModel to render a list of options"_. Isn't `ngModel` required by `ngOptions`?

gkalpak:
> When option elements use ngValue, value or interpolated text simply to set the option value, i.e. when the parent select doesn't have an ngModel, there is no necessity in registering the options with the select controller.

Afaict, this PR breaks these usecases. I think we should keep the current behaviors (e.g. set the `value` based on the interpolated text), but revert the special `ngValue` handling.

Narretz:
Sorry, I don't follow. Why does this PR break these use cases? Do you mean simply using ngValue, interpolated value or text to set the value? The registerOption function should (and I am pretty sure does) only include code that integrates the options with the select controller, so listening on changes, removal etc. and then updating the interal representation, calling render, setViewValue etc,

You are right, ngOptions must be used with ngModel. :sweat: Using it without could be interesting for the material guys actually, as ng-options has been requested.

gkalpak:
> Why does this PR break these use cases?

Sorry, I was exaggerating a bit :smile: I meant it breaks the case with `interpolateTextFn`.

Narretz:
You mean because interpolateTextFn sets the value when the text changes? Since HTML5 uses the text value of an option as its value if there's no value attribute, I thought this would be okay. But if someone specifically listens on the value attribute of an option that has interpolated text, yes that would be a BC. But I think we can live that (the changes are in 1.6 only anyway)

Narretz:
We actually used to not add options to the selectCtrl if there was no ngModel: https://github.com/angular/angular.js/blob/bba600414784ef969f601556662560d0e9b24658/src/ng/directive/select.js#L754

gkalpak:
> We actually used to not add options to the selectCtrl if there was no ngModel

But then we used to handle the `interpolateTextFn` case from the [optionDirective][1].

Anyway, since the select's value will be set correctly, I am fine dropping this as long as we mention the BC.

[1]: https://github.com/angular/angular.js/blob/bba600414784ef969f601556662560d0e9b24658/src/ng/directive/select.js#L759-L765

Narretz:
Even that might be technically an implementation detail as the option directive is not documented at all.

PR #14867, opened 2016-07-04
fix($resource): append extra params even if `Object.prototype` has properties with the same name

### LINKED ISSUE #14866: $resource v1.5.6 on Firefox-only: does not pass "watch" query string parameter for GET URLs
Bug

$resource should pass parameters for GET on query string. On Firefox only, it appears to skip parameter named "watch" (specific to only that parameter name).

Plnkr at:  http://plnkr.co/edit/E7KwZfBaYQPqDFZzTTsC?p=preview

Note I did not stub $http; check the Network console to see the missing parameter "watch" in FireFox only. Parameters "day" and "slot" work fine. Specific URL examples below.

Expected behavior: Firefox impl of $resource should include "watch=..." on query string, like it does for other names in the query string data.

The fact that it's missing seems like a bug; it's only present in Firefox.

v1.5.6 (Plnkr above)

! I didn't stub $http in the Plnkr above; check the Console Network log to see that, e.g., 

  Good/Chrome:  
      http://run.plnkr.co/r/get-data?day=2016-07-04&slot=1&watch=1

  Bad/FireFox (missing watch=1)  
      http://run.plnkr.co/r/get-data?day=2016-07-04&slot=1  
      !!Missing watch=1

gkalpak:
Bug fix

If `Object.prototype` has a property named `xyz`, then an extra `xyz` param passed to a `$resource` action call, **does not** get passed to `$http` (in `config.params`).
See #14866.

If `Object.prototype` has a property named `xyz`, then an extra `xyz` param passed to a `$resource` action call, **does** get passed to `$http` (in `config.params`).

Possibly. If someone relied on using a `params` object with own, enumerable properties that have the same names with properties on `Object.prototype` and not have these properties passed to `$http` (and included in the URL as query params), then this would break them. Highly unlikely though.

**Other information**:
Fixes #14866

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -535,7 +535,7 @@ angular.module('ngResource', ['ng']).
             encodedVal,
             protocolAndDomain = '';
 
-          var urlParams = self.urlParams = {};
+          var urlParams = self.urlParams = Object.create(null);
           forEach(url.split(/\W/), function(param) {
             if (param === 'hasOwnProperty') {
               throw $resourceMinErr('badname', "hasOwnProperty is not a valid parameter name.");
@@ -591,7 +591,7 @@ angular.module('ngResource', ['ng']).
 
           // set params - delegate param encoding to $http
           forEach(params, function(value, key) {
-            if (!self.urlParams.hasOwnProperty(key)) {
+            if (!self.urlParams[key]) {
               config.params = config.params || {};
               config.params[key] = value;
             }
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -591,7 +591,7 @@ angular.module('ngResource', ['ng']).
 
           // set params - delegate param encoding to $http
           forEach(params, function(value, key) {
-            if (!self.urlParams[key]) {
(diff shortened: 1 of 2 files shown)


Narretz:
I think we usually did createMap() (which does Object.create(null)) in these cases. Could we apply this here too?

gkalpak:
> I think we usually did createMap() (which does Object.create(null)) in these cases

That was my initial intention, but then somehow I forgot :confused: 
Fixed!

lgalfaso:
I would change the commit log to `Add as params all owned properties` and think it is best to squash this into a single commit. Otherwise LGTM

gkalpak:
Should we just merge it in 1.6.x only and mark it as a BC. (I don't expect this to actually break any apps or tests, but it wouldn't be the first time something breaks unexpectedly :smiley:)

Narretz:
Yes, let's merge just in 1.6 with a BC notice. It should be pretty easy to work around this in an affected app (not to mention that this happens rarely)

gkalpak:
Landed on 1.6.x. Thx for the reviews :smile:

Issue #6726, opened 2014-03-17
ng-model does not set initial value and position of input[type=range]


fschwiet:
Basically, a range input is set up with an initial value using ng-model.  The actual value of the HTML element, and the position of the range slider, are incorrect.  I would expect the slider position and html input value to be in sync with the model variable.

  Here is a repro: http://jsfiddle.net/fschwiet/HVp2J/

  Verified on chrome and firefox running on Windows 8.

  I see there are other issues opened for input[type=range], its not clear they are related.

kelle62819:
Yes, still an issue.

mpromonet:
I asked something similar in stackoverflow  http://stackoverflow.com/questions/26453634/how-to-initialize-the-value-of-an-inputrange-using-angularjs-when-value-is-ove . The given answer works with 1 slider, but not with more.  Using 1.4.0-beta.build.3791 this bug is still present.

bastianwegge:
+1

Yimiprod:
+1

mrbinky3000:
+1 for this issue

Tooa:
Facing the same problem: +1

larpo1:
yep - same problem +1

onemenny:
this is basic html5 - whats the status on this, is anyone working on this?

onemenny:
taking under consideration this: https://github.com/danielcrisp/angular-rangeslider/issues/51
where Error: ngModel:numfmt range slider is thrown
i adjusted @ardf69 solution to this: 
angular.module('common').directive('rangeSlider', [function () {
    return {
        replace: true,
        restrict: 'E',
        require: 'ngModel',
        template: '<input type="range"/>',
        link: function (scope, element, attrs, ngModel) {
            var ngRangeMin;
            var ngRangeMax;
            var ngRangeStep;
            var value;

            function init() {
                if (!angular.isDefined(attrs.ngRangeMin)) {
                    ngRangeMin = 0;
                } else {
                    scope.$watch(attrs.ngRangeMin, function (newValue, oldValue, scope) {
                        if (angular.isDefined(newValue)) {
                            ngRangeMin = newValue;
                            setValue();
                        }
                    });
                }
                if (!angular.isDefined(attrs.ngRangeMax)) {
                    ngRangeMax = 100;
                } else {
                    scope.$watch(attrs.ngRangeMax, function (newValue, oldValue, scope) {
                        if (angular.isDefined(newValue)) {
                            ngRangeMax = newValue;
[...]

PR #14870, opened 2016-07-05
feat(input): add support for input[type=range]

### LINKED ISSUE #5892: Add support for the input type=range
Originally filed here: https://github.com/angular/angular.js/issues/1189

(This was incorrectly closed and despite requests to reopen, has remained closed so I'm reposting as a new ticket. For future reference, closing old tickets due to lack of activity is probably not a good idea.)

With multiple inputs to the same model <input type="number" /> doesn't play ball. It works as an input but updating it to a value set from another input always fails. Even if the source is also constrained as numeric (type="range" in chrome). This is due to the model being stored as a string.

Demo: http://jsfiddle.net/TXazw/5/Edit
Edit any input except the number and everything updates except the number input. Edit the number input and everything works fine.

A workaround is included. Uncomment parseFloat line and all inputs work correctly.

Should values not be automatically parsed when applied to a number type input? At the very least type="range" should be outputting a number rather than a string.

Discussed in the group:
https://groups.google.com/forum/?fromgroups#!topic/angular/Ecjx2fo8Qvk

Narretz:
feature

**Other information**:

the input[type=range] behavior is the same of an input[type=number]
with min=0, max=100 and step=1 as defaults

Closes #5892
Closes #9715

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1031,6 +1031,113 @@ var inputType = {
    */
   'radio': radioInputType,
 
+  /**
+   * @ngdoc input
+   * @name input[range]
+   *
+   * @description
+   * Native range input with validation and transformation.
+   *
+   * The model for the range input must always be a `Number`.
+   *
+   * IE9 and other browsers that do not support the `range` type fall back
+   * to a text input. Model binding, validation and number parsing are nevertheless supported.
+   *
+   * Browsers that support range (latest Chrome, Safari, Firefox, Edge) treat `input[range]`
+   * in a way that never allows the input to hold an invalid value. That means:
+   * - any non-numerical value is set to `(max + min) / 2`.
+   * - any numerical value that is less than the current min val, or greater than the current max val
+   * is set to the min / max val respectively.
+   *
+   * This has the following consequences for Angular:
+   *
+   * Since the element value should always reflect the current model value, a range input
+   * will set the bound ngModel expression to the value that the browser has set for the
+   * input element. For example, in the following input `<input type="range" ng-model="model.value">`,
+   * if the application sets `model.value = null`, the browser will set the input to `'50'`.
+   * Angular will then set the model to `50`, to prevent input and model value being out of sync.
+   *
+   * That means the model for range will immediately be set to `50` after `ngModel` has been
+   * initialized. It also means a range input can never have the required error.
+   *
+   * This does not only affect changes to the model value, but also to the values of the `min` and
+   * `max` attributes. When these change in a way that will cause the browser to modify the input value,
+   * Angular will also update the model value.
+   *
+   * Automatic value adjustment also means that a range input element can never have the `required`,
+   * `min`, or `max` errors, except when using `ngMax` and `ngMin`, which are not affected by automatic
+   * value adjustment, because they do not set the `min` and `max` attributes.
+   *
+   * @param {string}  ngModel Assignable angular expression to data-bind to.
+   * @param {string=} name Property name of the form under which the control is published.
+   * @param {string=} min Sets the `min` validation to ensure that the value entered is greater
+   *                  than `min`. Can be interpolated.
+   * @param {string=} max Sets the `max` validation to ensure that the value entered is less than `max`.
+   *                  Can be interpolated.
+   * @param {string=} ngMin Takes an expression. Sets the `min` validation to ensure that the value
+   *                  entered is greater than `min`. Does not set the `min` attribute and therefore
+   *                  adds no native HTML5 validation. It also means the browser won't adjust the
+   *                  element value in case `min` is greater than the current value.
+   * @param {string=} ngMax Takes an expression. Sets the `max` validation to ensure that the value
+   *                  entered is less than `max`. Does not set the `max` attribute and therefore
+   *                  adds no native HTML5 validation. It also means the browser won't adjust the
+   *                  element value in case `max` is less than the current value.
+   * @param {string=} ngChange Angular expression to be executed when the ngModel value changes due
+   *                  to user interaction with the input element.
+   *
+   * @example
+      <example name="range-input-directive" module="rangeExample">
+        <file name="index.html">
+          <script>
+            angular.module('rangeExample', [])
+              .controller('ExampleController', ['$scope', function($scope) {
+                $scope.value = 75;
+                $scope.min = 10;
+                $scope.max = 90;
+              }]);
+          </script>
+          <form name="myForm" ng-controller="ExampleController">
+
+            Model as range: <input type="range" name="range" ng-model="value" min="{{min}}"  max="{{max}}">
+            <hr>
+            Model as number: <input type="number" ng-model="value"><br>
+            Min: <input type="number" ng-model="min"><br>
+            Max: <input type="number" ng-model="min"><br>
+            value = <code>{{value}}</code><br/>
+            myForm.range.$valid = <code>{{myForm.range.$valid}}</code><br/>
+            myForm.range.$error = <code>{{myForm.range.$error}}</code>
+          </form>
+        </file>
+      </example>
+
+   * ## Range Input with ngMin & ngMax attributes
+
+   * @example
+      <example name="range-input-directive-ng" module="rangeExample">
+        <file name="index.html">
+          <script>
+            angular.module('rangeExample', [])
+              .controller('ExampleController', ['$scope', function($scope) {
+                $scope.value = 75;
+                $scope.min = 10;
+                $scope.max = 90;
+              }]);
+          </script>
+          <form name="myForm" ng-controller="ExampleController">
+            Model as range: <input type="range" name="range" ng-model="value" ng-min="min" ng-max="max">
+            <hr>
+            Model as number: <input type="number" ng-model="value"><br>
+            Min: <input type="number" ng-model="min"><br>
+            Max: <input type="number" ng-model="min"><br>
+            value = <code>{{value}}</code><br/>
+            myForm.range.$valid = <code>{{myForm.range.$valid}}</code><br/>
+            myForm.range.$error = <code>{{myForm.range.$error}}</code>
+          </form>
+        </file>
+      </example>
+
+   */
+  'range': rangeInputType,
 
   /**
    * @ngdoc input
@@ -1382,10 +1489,7 @@ function badInputChecker(scope, element, attr, ctrl) {
   }
 }
 
-function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
-  badInputChecker(scope, element, attr, ctrl);
-  baseInputType(scope, element, attr, ctrl, $sniffer, $browser);
-
+function numberFormatterParser(ctrl) {
   ctrl.$$parserName = 'number';
   ctrl.$parsers.push(function(value) {
     if (ctrl.$isEmpty(value))      return null;
@@ -1402,6 +1506,12 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     }
     return value;
   });
+}
+
+function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
+  badInputChecker(scope, element, attr, ctrl);
+  numberFormatterParser(ctrl);
+  baseInputType(scope, element, attr, ctrl, $sniffer, $browser);
 
   if (isDefined(attr.min) || attr.ngMin) {
     var minVal;
@@ -1434,6 +1544,110 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
       ctrl.$validate();
     });
   }
+}
+
+function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
+  badInputChecker(scope, element, attr, ctrl);
+  numberFormatterParser(ctrl);
+  baseInputType(scope, element, attr, ctrl, $sniffer, $browser);
+
+  var minVal = 0,
+      maxVal = 100,
+      supportsRange = ctrl.$$hasNativeValidators && element[0].type === 'range',
+      validity = element[0].validity;
+
+  var originalRender = ctrl.$render;
+
+  ctrl.$render = supportsRange && isDefined(validity.rangeUnderflow) && isDefined(validity.rangeOverflow) ?
+    //Browsers that implement range will set these values automatically, but reading the adjusted values after
+    //$render would cause the min / max validators to be applied with the wrong value
+    function rangeRender() {
+      originalRender();
+      ctrl.$setViewValue(element.val());
+    } :
+    originalRender;
+
+  function minChange(val) {
+    if (isDefined(val) && !isNumber(val)) {
+      val = parseFloat(val);
+    }
+    minVal = isNumber(val) && !isNaN(val) ? val : undefined;
+    // ignore changes before model is initialized
+    if (isNumber(ctrl.$modelValue) && isNaN(ctrl.$modelValue)) {
+      return;
+    }
+
+    if (supportsRange && minAttrType === 'min') {
+      var elVal = element.val();
+      // IE11 doesn't set the el val correctly if the minVal is greater than the element value
+      if (minVal > elVal) {
+        element.val(minVal);
+        elVal = minVal;
+      }
+      ctrl.$setViewValue(elVal);
+    } else {
+      // TODO(matsko): implement validateLater to reduce number of validations
+      ctrl.$validate();
+    }
+  }
+
+  var minAttrType = isDefined(attr.ngMin) ? 'ngMin' : isDefined(attr.min) ? 'min' : false;
+  if (minAttrType) {
+    ctrl.$validators.min = isDefined(attr.min) && supportsRange ?
+      function noopMinValidator(value) {
+        // Since all browsers set the input to a valid value, we don't need to check validity
+        return true;
+      } :
+      // ngMin doesn't set the min attr, so the browser doesn't adjust the input value as setting min would
+      function minValidator(modelValue, viewValue) {
+        return ctrl.$isEmpty(viewValue) || isUndefined(minVal) || viewValue >= minVal;
+      };
+
+    // Assign minVal when the directive is linked. This won't run the validators as the model isn't ready yet
+    minChange(attr.min);
+    attr.$observe('min', minChange);
+  }
+
+  function maxChange(val) {
+    if (isDefined(val) && !isNumber(val)) {
+      val = parseFloat(val);
+    }
+    maxVal = isNumber(val) && !isNaN(val) ? val : undefined;
+    // ignore changes before model is initialized
+    if (isNumber(ctrl.$modelValue) && isNaN(ctrl.$modelValue)) {
+      return;
+    }
+
+    if (supportsRange && maxAttrType === 'max') {
+      var elVal = element.val();
+      // IE11 doesn't set the el val correctly if the maxVal is less than the element value
+      if (maxVal < elVal) {
+        element.val(maxVal);
+        elVal = minVal;
+      }
+      ctrl.$setViewValue(elVal);
+    } else {
+      // TODO(matsko): implement validateLater to reduce number of validations
+      ctrl.$validate();
+    }
+  }
+  var maxAttrType = isDefined(attr.max) ? 'max' : attr.ngMax ? 'ngMax' : false;
+  if (maxAttrType) {
+    ctrl.$validators.max = isDefined(attr.max) && supportsRange ?
+      function noopMaxValidator() {
+        // Since all browsers set the input to a valid value, we don't need to check validity
+        return true;
+      } :
+      // ngMax doesn't set the max attr, so the browser doesn't adjust the input value as setting max would
+      function maxValidator(modelValue, viewValue) {
+        return ctrl.$isEmpty(viewValue) || isUndefined(maxVal) || viewValue <= maxVal;
+      };
+
+    // Assign maxVal when the directive is linked. This won't run the validators as the model isn't ready yet
+    maxChange(attr.max);
+    attr.$observe('max', maxChange);
+  }
+
 }
 
 function urlInputType(scope, element, attr, ctrl, $sniffer, $browser) {
--- a/test/ng/directive/inputSpec.js
+++ b/test/ng/directive/inputSpec.js
@@ -2819,6 +2819,430 @@ describe('input', function() {
     });
   });
 
+  describe('range', function() {
+
+    var scope;
+
+    var rangeTestEl = angular.element('<input type="range">');
+    var supportsRange = rangeTestEl[0].type === 'range';
+    beforeEach(function() {
+      scope = $rootScope;
+    });
+
+    if (supportsRange) {
+      // This behavior only applies to browsers that implement the range input, which do not
+      // allow to set a non-number value and will set the value of the input to 50 even when you
+      // change it directly on the element.
+      // Other browsers fall back to text inputs, where setting a model value of 50 does not make
+      // sense if the input value is a string. These browsers will mark the input as invalid instead.
+
+      it('should render as 50 if null', function() {
+        var inputElm = helper.compileInput('<input type="range" ng-model="age" />');
+
+        helper.changeInputValueTo('25');
+        expect(scope.age).toBe(25);
+
+        scope.$apply('age = null');
+
+        expect(inputElm.val()).toEqual('50');
+      });
+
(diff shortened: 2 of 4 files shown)


Narretz:
The biggest issue is that browser that implement range set the element value to a valid value if you try to set an invalid value. Two cases: you try to set a non-number: input value is set to 50. You try to set a value that exceeds max or is lower than min: value is set to max / min value.

This is problematic for two reasons:
- When we try to set the input value from the scope, as it creates a mismatch between input and scope value.
- null / undefined models are set to 50, which means the input can never have the required error set, even though from an application perspective the model is still null / undefined

In the PR, I have changed it so that we update the model after the browser has changed the value.
This obviously runs parsers / validators again.
It also causes some unexpected effects when you bind the same model to range and number for example. When you try to clear the number input, the range will always set the model (and input) back to 50.

As an alternative, I though of creating a validator that adds a misMatch error the the control that is set whenever the browsers updates the value and it is now distinct from the model.

gkalpak:
BTW, it doesn't set it to 50, it sets it to `(max - min) / 2`. So, basically, you can't set it to an undefined value. I am still not sure it is a good idea to add something like this is core - too many quirks :worried:

src/ng/directive/input.js
```diff
@@ -1436,6 +1481,23 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
 }
 
+function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
+  numberInputType(scope, element, attr, ctrl, $sniffer, $browser);
+
+  var originalRender = ctrl.$render;
+
+  ctrl.$render = function() {
+    if (ctrl.$isEmpty(ctrl.$viewValue)) {
+      ctrl.$viewValue = '50';
```
gkalpak: This shouldn't be hard-coded to 50. It should be `(max - min) / 2`.

test/ng/directive/inputSpec.js
```diff
@@ -2819,6 +2819,313 @@ describe('input', function() {
     });
   });
 
+  describe('range', function() {
+
+    var scope;
+
+    beforeEach(function() {
+      scope = $rootScope;
```
gkalpak: :smile:

test/ng/directive/inputSpec.js
```diff
@@ -2819,6 +2819,313 @@ describe('input', function() {
...
+      expect(inputElm.val()).toBe('100');
+
+      try {
+        // to allow non-number values, we have to change type so that
+        // the browser which have number validation will not interfere with
+        // this test. IE8 won't allow it hence the catch.
+        inputElm[0].setAttribute('type', 'text');
+      } catch (e) {}
+
+      helper.changeInputValueTo('100X');
```
gkalpak: Since this is not possible to happen, why do we need to test this?

src/ng/directive/input.js
gkalpak: BTW, i case it wasn't clear, `50` is the result of `(max - min) / 2` for the default min (`0`) and max (`100`) values.

test/ng/directive/inputSpec.js
```diff
@@ -2819,6 +2819,313 @@ describe('input', function() {
...
+          expect(scope.form.alias.$error.min).toBeFalsy();
+        });
+      }
+
+      it('should validate if "range" is not implemented', function() {
+        var inputElm = helper.compileInput('<input type="range" ng-model="value" name="alias" min="10" />');
+
+        // To test how browsers that don't implement range behave, we have to change the type, so
+        // that special layout and validation rules don't apply
+        inputElm[0].setAttribute('type', 'text');
```
gkalpak: I think this might introduce some funny behavior that is different from the actuall behavior (not to mention you said IE8 will throw - not that we care :stuck_out_tongue: ).
I suggest we check if the type is not supported (e.g. `inputElm[0].type === 'text'`) and only then continue with the test.

gkalpak:
I think we need to better handle the situation where `min`/`max` change dynamically. The browser will adjust the element's value to the new `min`/`max` values if necessary and we don't account for that.

test/ng/directive/inputSpec.js
Narretz: This was copied from another test so we can remove the IE8 line.
Your suggestion is good although i wouldn't put it into the spec itself. Rather make it part of the sniffer?

test/ng/directive/inputSpec.js
Narretz: It can happen in IE9. Browsers that don't implement the range should still be able to validate the input correctly  (we mimic the validation rules if no validity is present )

src/ng/directive/input.js
Narretz: Yep, the PR is currently missing the min max handling.

test/ng/directive/inputSpec.js
gkalpak: We already use this approach in ngModel (iirc).
The sniffer typically has some "flags" set, do you suggest we test all available input types on start-up and set appropriate flags on sniffer. I wouldn't be very keen on this. A helper function sounds better. Something like this:

```js
function isOfSupportedType(input) {
  return input.getAttribute('type') === input.type;
}
```

gkalpak: We use it in `baseInputType`. We basically check: `input.type === attrs.type`

Narretz:
That's true. I noticed that but haven't included it yet. So what do you in general think about updating the model when the browser changes the values?

gkalpak:
Yeah, I think we should read the viewValue every time the `min`, `max` values change (and if necessary run parsers/validators).
(Not sure how `step` affects the element's value.)

Narretz:
When a step is set, the browser rounds non-matching values to the nearest step ...

gkalpak:
So, how does affect `ngModel`? When the model is set to an "off-step" value, the browser will round the value to the nearest step and the change will propagate to the modelValue?

src/ng/directive/input.js
Narretz: the correct formula is actually `min + ((max - min) / 2)`

gkalpak: Sorry, I meant `(max + min) / 2` (which is equivalent to your formula).
I will correct it to avoid confusion.

gkalpak:
It is a problematic situation either way. For example if we leave the original (off-step) modelValue, the field will be invalid, but a valid value will be shown to the user. And in order to set the modelValue to the corrected, nearest step value, they will have to move the knob to a different value and back, so we can detect a viewValue change and parse that.

Things like that make me think that supporting range is going to cause more headaches than it will solve problems. Users are probably better of using a non-native slider/range input - e.g. as an independant module or from a UI framework (such as ngMaterial). :disappointed:

Narretz:
But for simple cases, binding to range is almost as simple as binding to number. So I'm not sure we should bail out of this simply because there are some edge cases.

petebacondarwin:
So from reading the HTML5 specs, I think it is fair to say that range is “special"

In the sense that it can ​*never*​ hold an invalid value

That means that we must always update the ngModel $viewValue to a valid value however it is updated (via user interaction or programmatically).

If we agree that this is the case, and document it clearly, then I think we could go ahead with this directive...

I think that we should also, when reacting to model changes, simply compute the appropriate valid value that would work for the control and update the model explicitly ourselves

So in the case of a model that is out of “step”, if we have this range

```
<input type="range" min="0" max="9" step="2" ng-model="obj.val">
```

and we try to set `obj.val` to `2.6`, then we force the model to `2` as part of the formatting.

src/ng/directive/input.js
```diff
@@ -1530,6 +1530,11 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
 
     // TODO(matsko): implement validateLater to reduce number of validations
     if (minAttrType === 'min') {
+      var elVal = element.val();
+      // IE11 doesn't set the el val correctly if the maxVal is less than the element value
```
roryokane: This comment doesn’t match the surrounding code. Bad copy-paste from [line 1572](https://github.com/angular/angular.js/blob/0dbe342417f8e0357f72314e2375f62ddaaec6ab/src/ng/directive/input.js#L1572)?

karma-shared.conf.js
```diff
@@ -17,7 +17,7 @@ module.exports = function(config, specificOptions) {
       testName: specificOptions.testName || 'AngularJS',
       startConnect: true,
       options: {
-        // We need selenium version +2.46 for Firefox 39 and the last selenium version for OS X is 2.45.
+        // We need selenium version +2.46 for 'Firefox' 39 and the last selenium version for OS X is 2.45.
```
gkalpak: ???

Is this comment event relevant anymore.?

src/ng/directive/input.js
```diff
@@ -1031,6 +1031,112 @@ var inputType = {
...
+   * - any non-numerical value is set to `(max + min) / 2`.
+   * - any numerical value that is less than the current min val, or greater than the current max val
+   * is set to the min / max val respectively.
+   *
+   * This has the following consequences for Angular:
+   *
+   * Since the element value should always reflect the current model value, a range input
+   * will set the bound ngModel expression to the value that the browser has set for the
+   * input element. For example, in the following input `<input type="range" ng-model="model.value">`,
+   * if I set `model.value = null`, the browser will set the input to `'50'`. Angular will then set
```
gkalpak: _I set_  -->  _the user sets_ (?)

src/ng/directive/input.js
```diff
@@ -1031,6 +1031,112 @@ var inputType = {
...
+   * @param {string=} min Sets the `min` validation to ensure that the value entered is greater
+   *                  than `min`. Can be interpolated.
+   * @param {string=} max Sets the `max` validation to ensure that the value entered is less than `max`.
+   *                  Can be interpolated.
+   * @param {string=} ngMin Takes an expression. Sets the `min` validation to ensure that the value
+   *                  entered is greater than `min`. Does not set the `min` attribute and therefore
+   *                  adds no native HTML5 validation. It also means the browser won't adjust the
+   *                  element value in case `min` is greater than the current value.
+   * @param {string=} ngMax Takes an expression. Sets the `max` validation to ensure that the value
+   *                  entered is less than `max`. Does not set the `min` attribute and therefore
```
gkalpak: _min_  -->  _max_

src/ng/directive/input.js
```diff
@@ -1031,6 +1031,112 @@ var inputType = {
...
+               $scope.value = 75;
+               $scope.min = 10;
+               $scope.max = 90;
+             }]);
+         </script>
+         <form name="myForm" ng-controller="ExampleController">
+
+           Model as range: <input type="range" name="range" ng-model="value"
+                          min="{{min}}"  max="{{max}}">
+          <hr>
```
gkalpak: Wrong indentation here and below.

src/ng/directive/input.js
```diff
@@ -1031,6 +1031,112 @@ var inputType = {
...
+           angular.module('rangeExample', [])
+             .controller('ExampleController', ['$scope', function($scope) {
+               $scope.value = 75;
+               $scope.min = 10;
+               $scope.max = 90;
+             }]);
+         </script>
+         <form name="myForm" ng-controller="ExampleController">
+           Model as range: <input type="range" name="range" ng-model="value"
+                          ng-min="10" ng-max="90">
```
gkalpak: _10_  -->  _min_
_90_  -->  _max_

src/ng/directive/input.js
```diff
@@ -1436,6 +1542,124 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
 }
 
+function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
+  badInputChecker(scope, element, attr, ctrl);
+  baseInputType(scope, element, attr, ctrl, $sniffer, $browser);
+
+  var minVal = 0,
+      maxVal = 100,
+      supportsRange = ctrl.$$hasNativeValidators && element[0].type === 'range',
+      validity = element[0].validity;
```
gkalpak: It might be better to use `VALIDITY_STATE_PROPERTY`.

src/ng/directive/input.js
```diff
@@ -1436,6 +1542,124 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+        element.val(minVal);
+      }
+      ctrl.$setViewValue(element.val());
+    } else {
+      // TODO(matsko): implement validateLater to reduce number of validations
+      ctrl.$validate();
+    }
+  }
+
+  var minAttrType = isDefined(attr.min) ? 'min' : attr.ngMin ? 'ngMin' : false;
```
gkalpak: Don't `ngMin`/`ngMax` set `attr.min/max` anyway?

src/ng/directive/input.js
```diff
@@ -1436,6 +1542,124 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+  if (minAttrType) {
+    ctrl.$validators.min = isDefined(attr.min) && supportsRange ?
+      function noopMinValidator(value) {
+        // Since all browsers set the input to a valid value, we don't need to check validity
+        return true;
+      } :
+      // ngMin doesn't set the min attr, so the browser doesn't adjust the input value as setting min would
+      function minValidator(modelValue, viewValue) {
+        return ctrl.$isEmpty(viewValue) || isUndefined(minVal) || viewValue >= minVal;
```
gkalpak: Why calling `$isEmpty` with the `viewValue`?

src/ng/directive/input.js
```diff
@@ -1436,6 +1542,124 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+      return;
+    }
+
+    if (supportsRange && minAttrType === 'min') {
+      var elVal = element.val();
+      // IE11 doesn't set the el val correctly if the minVal is greater than the element value
+      if (minVal > elVal) {
+        element.val(minVal);
+      }
+      ctrl.$setViewValue(element.val());
```
gkalpak: Can't we use `elVal` here?

test/ng/directive/inputSpec.js
```diff
@@ -2819,6 +2819,436 @@ describe('input', function() {
...
+      helper.changeInputValueTo('75');
+      expect(scope.age).toBe(75);
+    });
+
+    if (supportsRange) {
+      // This behavior only applies to browsers that implement the range input, which do not
+      // allow to set a non-number value and will set the value of the input to 50 even when you
+      // change it directly on the element.
+      // Other browsers fall back to text inputs, where setting a model value of 50 does not make
+      // sense if the input value is a string. These browser will mark the input as invalid instead.
```
gkalpak: _browser_  -->  _browsers_ :innocent:

test/ng/directive/inputSpec.js
```diff
@@ -2819,6 +2819,436 @@ describe('input', function() {
...
+
+      it('should render as 50 if null', function() {
+        var inputElm = helper.compileInput('<input type="range" ng-model="age" />');
+
+        scope.$apply('age = null');
+
+        expect(inputElm.val()).toEqual('50');
```
gkalpak: Nitpick: I would change the element's value first, in order to verify that setting the model to `null` indeed sets the element's value to `'50'` (instead of leaving in unchanged for example).

src/ng/directive/input.js
gkalpak: Or not...

test/ng/directive/inputSpec.js
```diff
@@ -2819,6 +2819,436 @@ describe('input', function() {
...
+      expect(inputElm).toBeValid();
+
+      helper.changeInputValueTo('this-will-fail-because-of-the-badInput-flag');
+
+      expect(scope.age).toBeUndefined();
+      expect(inputElm).toBeInvalid();
+    });
+
+
+    it('should validate with undefined viewValue when $validate() called', function() {
```
gkalpak: _called_  -->  _is called_

gkalpak: Tbh, I am not sure what this test is supposed to test.

test/ng/directive/inputSpec.js
```diff
@@ -2819,6 +2819,436 @@ describe('input', function() {
...
+      expect(inputElm).toBeValid();
+      expect(scope.form.alias.$error.number).toBeUndefined();
+    });
+
+
+    it('should throw if the model value is not a number', function() {
+      expect(function() {
+        scope.value = 'one';
+        var inputElm = helper.compileInput('<input type="range" ng-model="value" />');
+      }).toThrowMinErr('ngModel', 'numfmt', "Expected `one` to be a number");
```
gkalpak: Inconsistent use of quotes. (@mgol, I have your back :stuck_out_tongue: )

mgol: Yeah, let's use single anyway, especially that we're not validating it yet.

gkalpak:
I just found out that you can make an `input[range]` invalid. Changing the `step` at runtime (while having an off-step value) makes it invalid: it sets the `validity.stepMismatch` flag to true (but keeps the previous value).

I think we need some tests with `step`.

karma-shared.conf.js
Narretz: I don't think so. The change is accidental. Sublime CTRL+D and ' in the wrong file

src/ng/directive/input.js
Narretz: *the application sets*. For me, stuff that's set on the scope is only the user's action if it comes from an input.

src/ng/directive/input.js
Narretz: They do. But iirc at this point (linking) it is not yet set, so I can distinguish it based on attr.min / ngMin

src/ng/directive/input.js
Narretz: Because we are validating the $viewValue. I think dates use the modelValue, but this is something we should change, imo.

src/ng/directive/input.js
Narretz: I guess that works

test/ng/directive/inputSpec.js
Narretz: I think that's copied from input[number] with number substituted for range.

Narretz:
I left step out, because I wanted to get it to work with min/max first. I'd like to work on step for number/range after this is merged

src/ng/directive/input.js
gkalpak: I wouldn't rely on default directive priorities for that. Let's be on the safe side with:

```js
isDefined(attr.ngMin) ? 'ngMin' : isDefined(attr.min) ? 'min' : false;
```

<sub>Note that `attr.ngMin`/`min` could be numbers, and `0` would be falsy, so we need `isDefined` in both cases.</sub>

src/ng/directive/input.js
gkalpak: I think we almost always call `$isEmpty` with the `modelValue` (regardless of what value we are actually validating). Imo it makes sense to (a) be consistent what we call `$isEmpty` with and (b) always call it with `modelValue` (which can be `$modelValue` or `$$rawModelValue`), in order to account for custom parsers.

For example, an `input[range]`'s viewValue can never be empty, but I might want to have a custom parser that converts `'0'` to `null` and I want this to be considered empty.

--
<sub>It is debatable whether we should be validating the `$viewValue` or the `$modelValue`. I think there are valid usecases for both.</sub>

gkalpak:
> I'd like to work on step for number/range after this is merged

I have mixed feelings about this :smiley: I would rather not release `input[range]` with `step` support.

--
Btw, here is a browser inconsistency (probably a Chrome bug) I found out:

According to (my understanding of) [the spec][1], browsers should automatically adjust the `value` whenevr `step` changes, in order to ensure that there is no step mismatch. When changing the `step` to a string value (e.g. `input.step = '2'`), both Chrome and Firefox follow the spec. When using a number (instead of a string), although both browsers convert `step` to its string representation, only Firefox adjust the `value` and makes the element valid again. Chrome leaves the previous `value` and sets the `validity.stepMismatch` error. E.g.:

```html
<input type="range" step="10" />
```
```js
console.log(input.step)   // --> '10'
console.log(input.value)   //  --> '50'
console.log(input.validity.valid)   // --> true

input.step = '15';
console.log(input.step)   // --> '15'
console.log(input.value)   //  --> '45'
console.log(input.validity.valid)   // --> true

input.step = 20;
console.log(input.step)   // --> '20'
console.log(input.value)   //  --> Firefox: '40'  |  Chrome: '45'
console.log(input.validity.valid)   // --> Firefox: true  |  Chrome: false
```

[1]: https://www.w3.org/TR/html5/forms.html#range-state-(type=range)

src/ng/directive/input.js
Narretz: Yes, there are definitely both approaches in the code base. But for example the most recent change that touches this, are the 'ng-empty', and 'ng-not-empty' classes, which are set based on the result of `$ctrl.isEmpty(viewValue)`. https://github.com/angular/angular.js/blob/master/src/ng/directive/ngModel.js#L323
I'd rather make all built-in validators use isEmpty(viewValue), because technically all validators validate what the user has entered, even if we parse it a Date or Number.

Narretz:
Okay, step support should be in the same release as the basic support. But it doesn't have to be in the same commit.

src/ng/directive/input.js
```diff
@@ -1594,15 +1592,16 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
-      ctrl.$setViewValue(element.val());
+      ctrl.$setViewValue(elVal);
     } else {
       // TODO(matsko): implement validateLater to reduce number of validations
       ctrl.$validate();
     }
   }
 
-  var minAttrType = isDefined(attr.min) ? 'min' : attr.ngMin ? 'ngMin' : false;
+  var minAttrType = isDefined(attr.ngMin) ? 'ngMin' : isDefined(attr.min) ? 'min' : false;
```
gkalpak: The same change is necessary for `max` below.

src/ng/directive/input.js
gkalpak: I haven't thought that through tbh, but I see merit in both cases. Fwiw, the initial intent seems to be to call it against the `$modelValue`, as indicated by the default implementation (which checks for `undefined`, `null`, `NaN` - values that can hardly appeat as `$viewValue`s).

Imo, the way someone decides to "interpret" the value - i.e. through custom parsers - should be taken into account if possible. But, I understand there are "technical" difficulties with this approach, so checking against the `$viewValue` might be preferrable. 

Most importantly, we should be consistent (if we find ourselves needing to call `$isEmpty` with both values, it is probably an indication that we need to rethink the API - maybe have two separate methods) and communicate it well through docs.

gkalpak:
LGTM (the travis failures are not related to this PR and have been fixed on master).

src/ng/directive/input.js
Narretz: FWIW, the required validator calls $isEmpty with the viewValue: https://github.com/angular/angular.js/blob/cd2f6d9d3bcdbcaa3d09dfd9cc75a5058d83b44e/src/ng/directive/validators.js#L70

Though I agree that it's a problem in general that validators sometimes validate the viewValue and sometimes the modelValue.

Narretz:
Thanks for the review, @gkalpak ! And thanks again @cironunes for the initial implementation.

mpromonet:
Maybe I misunderstand, but even if it changes the behaviour, this PR doesnot fix the issue #6726.   
Using a simple sample from [How to initialize the value of an input[range] using AngularJS when value is over 100](http://stackoverflow.com/q/26453634/3102264) : 
```
    <html ng-app="App">
    <head>
    	<script src="https://code.angularjs.org/snapshot/angular.min.js"></script>
    	<script>
    		angular.module('App', ['App.controllers']);
    		angular.module('App.controllers', []).controller('AppController', function($scope) {
    			$scope.model = {'min':50, 'max':150, 'value':150};
    		});		
    	</script>
    </head>
    <body ng-controller="AppController" >
    	{{model.min}}<input type="range" min="{{model.min}}" max="{{model.max}}"   ng-model="model.value" value="{{model.value}}"  />{{model.max}}<br/>
    	value:{{model.value}}
    </body>
    </html>
```
Before the commit 9130166767c4792c5d32d08a918fc7becf32c9a6 it gives : 

> value = 150 but the slider is at 100.

After it gives :
 
> value = 100 but the slider is at 100.

The 2 views of value are now consistent, but the value and the slider should be at 150 as it is in the range [50,150].

Narretz:
@mpromonet Or it's a bug ... (it's a bug). Thanks for testing and reporting.

PR #14871, opened 2016-07-06
docs(guide/Directives): remove confusing definition object message

jdai8:

Docs update

- [x ] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format

**Other information**:

Prior to this point, the doc does not mention returning a postLink function, and all the examples use the definition object form. New readers could misinterpret "returning a function" as returning the factory function mentioned in the previous section.

Since this page is only a "gentle introduction" to directives, using the definition object is a best practice, and the message is already echoed in the docs for [$compile](https://docs.angularjs.org/api/ng/service/$compile), I think we should just remove the message altogether.


jdai8:
I signed it!

PR #14878, opened 2016-07-07
chore(package.json): update jasmine-reporters to 2.2.0

Narretz:
Currently, you cannot use fit or fdescribe with protractor due to a bug in our version of jasmine-reporters

### DIFF
--- a/npm-shrinkwrap.clean.json
+++ b/npm-shrinkwrap.clean.json
@@ -7,10 +7,10 @@
           "version": "3.3.6"
         },
         "express": {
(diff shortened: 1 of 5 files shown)


gkalpak:
LGTM

mhutton86:
@Narretz, can you provide a bit more information:
* what are the recreation steps?
* what's the actual error you're seeing?
* what exact version of angular are you running?
* can you provide a gist / plunker?
* is there anything else that needs to be done for this issue?

PR #14890, opened 2016-07-10
feat($http): support sending XSRF token to whitelisted origins

### LINKED ISSUE #7862: Adding XSRF header when request is NOT same origin, but is TRUSTED origin (subdomain)
So this may (probably) just be that I'm not understanding XSRF prevention correctly, so feel free to tell me I'm an idiot. But currently in the code (https://github.com/angular/angular.js/blob/dd1d189ee785a37fe1d9bddf3818152db6aa210a/src/ng/http.js#L883-L893), angular is checking if the requested URL is same origin with `urlIsSameOrigin(config.url)`. If it's not same origin, angular does not add the XSRF header token (`X-XSRF-TOKEN`).

Now, that's all good. Except that I'm requesting `api.example.com` from `example.com`, which fails the same origin check. Should this be allowed? Or does that somehow completely break the XSRF prevention?

gkalpak:
Feature.

The XSRF token header is not set on requests to cross-origin requests. This is usually desired, but there are cases where you want to whitelist specific origins.
See #7862 for context.

It is possible to whitelist specific origins, so requests to URLs on these origins will also have the XSRF token header set.

No.

**Other information**:
Fixes #7862

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -161,6 +161,7 @@
     "urlResolve": false,
     "urlIsSameOrigin": false,
     "urlIsSameOriginAsBaseUrl": false,
+    "urlIsAllowedOriginChecker": false,
 
     /* ng/controller.js */
     "identifierForController": false,
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -161,7 +161,7 @@
     "urlResolve": false,
     "urlIsSameOrigin": false,
     "urlIsSameOriginAsBaseUrl": false,
-    "urlIsAllowedOriginChecker": false,
+    "urlIsAllowedOriginFactory": false,
 
     /* ng/controller.js */
     "identifierForController": false,
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -34,7 +34,7 @@ function $HttpParamSerializerProvider() {
    * * `{'foo': {'bar':'baz'}}` results in `foo=%7B%22bar%22%3A%22baz%22%7D` (stringified and encoded representation of an object)
    *
    * Note that serializer will sort the request parameters alphabetically.
-   * */
+   */
 
   this.$get = function() {
     return function ngParamSerializer(params) {
@@ -101,7 +101,7 @@ function $HttpParamSerializerJQLikeProvider() {
    * });
    * ```
    *
-   * */
+   */
   this.$get = function() {
     return function jQueryLikeParamSerializer(params) {
       if (!params) return '';
@@ -261,7 +261,7 @@ function isSuccess(status) {
  *
  * @description
  * Use `$httpProvider` to change the default behavior of the {@link ng.$http $http} service.
- * */
+ */
 function $HttpProvider() {
   /**
    * @ngdoc property
@@ -315,7 +315,7 @@ function $HttpProvider() {
    * - **`defaults.xsrfHeaderName`** - {string} - Name of HTTP header to populate with the
    * XSRF token. Defaults value is `'X-XSRF-TOKEN'`.
    *
-   **/
+   */
   var defaults = this.defaults = {
     // transform incoming response data
     transformResponse: [defaultHttpResponseTransform],
@@ -362,7 +362,7 @@ function $HttpProvider() {
    *
    * @returns {boolean|Object} If a value is specified, returns the $httpProvider for chaining.
    *    otherwise, returns the current configured value.
-   **/
+   */
   this.useApplyAsync = function(value) {
     if (isDefined(value)) {
       useApplyAsync = !!value;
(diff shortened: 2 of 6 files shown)


test/ng/httpSpec.js
```diff
@@ -9,22 +9,17 @@ describe('$http', function() {
...
-  });
+  }));
 
   beforeEach(module({
+    $$cookieReader: function() { return mockedCookies; },
     customParamSerializer: customParamSerializer
   }));
-  beforeEach(module(function($exceptionHandlerProvider) {
-    $exceptionHandlerProvider.mode('log');
-  }));
```
gkalpak: Just refactoring `beforeEach` blocks merging unnecessarily separate blocks.

test/ng/httpSpec.js
```diff
@@ -35,7 +30,6 @@ describe('$http', function() {
       throw 'Unhandled exceptions trapped in $exceptionHandler!';
     }
 
-    $rootScope.$digest();
```
gkalpak: Not necessary. `$httpBackend.verifyNoOutstandingExpectation()` will do a $digest first anyway (unless called with `false`, which is not the case here).

test/ng/httpSpec.js
```diff
@@ -784,18 +778,6 @@ describe('$http', function() {
...
-      it('should not set XSRF cookie for cross-domain requests', inject(function($browser) {
-        mockedCookies['XSRF-TOKEN'] =  'secret';
-        $browser.url('http://host.com/base');
-        $httpBackend.expect('GET', 'http://www.test.com/url', undefined, function(headers) {
-          return isUndefined(headers['X-XSRF-TOKEN']);
-        }).respond('');
-
-        $http({url: 'http://www.test.com/url', method: 'GET', headers: {}});
-        $httpBackend.flush();
-      }));
```
gkalpak: The XSRF tests where moved to a "dedicated" `describe('XSRF')` block.

Narretz:
LGTM but needs a rebase

gkalpak:
Rebased. PTAL

Narretz (changes requested):
Only a few cosmetic changes requested, otherwise LGTM

src/.eslintrc.json
```diff
@@ -161,6 +161,7 @@
     "urlResolve": false,
     "urlIsSameOrigin": false,
     "urlIsSameOriginAsBaseUrl": false,
+    "urlIsAllowedOriginChecker": false,
```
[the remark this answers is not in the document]
gkalpak: But this one is not like the other functions :grin:
This one returns a function that checks (i.e. returns what one would call `isUrlAllowedOrigin()`).

So, I think naming it `isUrlAllowedOrigin` would be confusing. Open to other naming suggestions.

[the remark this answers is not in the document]
gkalpak: How about `urlIsAllowedOriginFactory()`?

Narretz (approved):

Narretz:
`Factory` is good - I noticed that this wasn't the actual function during the review, but then forgot to change my comment 😅

src/ng/http.js
```diff
@@ -383,9 +383,49 @@ function $HttpProvider() {
...
-   **/
+   */
   var interceptorFactories = this.interceptors = [];
 
+  /**
+   * @ngdoc property
+   * @name $httpProvider#xsrfWhitelistedOrigins
+   * @description
+   *
+   * Array containing URLs whose origins are trusted to receive the XSRF token. See the
```
petebacondarwin: Array containing "origins" that are trusted?
Or if you explicitly accept URLs (but ignore the path etc) then change the name to `xsrfWhitelistedUrls`.

src/ng/http.js
```diff
@@ -383,9 +383,49 @@ function $HttpProvider() {
...
+   * </div>
+   *
+   * @example
+   *
+   * ```js
+   * // App served from `https://example.com/`.
+   * angular.
+   *   module('xsrfWhitelistedOriginsExample', []).
+   *   config(['$httpProvider', function($httpProvider) {
+   *     $httpProvider.xsrfWhitelistedOrigins.push('https://api.example.com/');
```
petebacondarwin: Note that the `/` at the end of the URL is part of the path and not the origin.

src/ng/urlUtils.js
```diff
@@ -109,37 +110,66 @@ function urlIsSameOriginAsBaseUrl(requestUrl) {
...
+   * based on a list of whitelisted-origin URLs. The current location's origin is implicitly
+   * trusted.
+   *
+   * @param {string|Object} requestUrl - The URL to be checked (provided as a string that will be
+   *     resolved or a parsed URL object).
+   *
+   * @returns {boolean} - Whether the specified URL is of an allowed origin.
+   */
+  return function urlIsAllowedOrigin(requestUrl) {
+    var parsedUrl = isString(requestUrl) ? urlResolve(requestUrl) : requestUrl;
```
petebacondarwin: Not related to this PR but perhaps we should change `urlResolve` to be a `noop` if the `url` is already parsed (i.e. not a string)? This would save some of this boilerplate here and further down.

test/ng/httpSpec.js
```diff
@@ -2266,6 +2208,139 @@ describe('$http', function() {
   });
 
 
+  describe('XSRF', function() {
+    var $http;
+    var $httpBackend;
+
+    beforeEach(module(function($httpProvider) {
+      $httpProvider.xsrfWhitelistedOrigins.push('https://whitelisted.example.com/');
```
petebacondarwin: Note that the trailing slash is ignored. Should we have tests for URLs with different paths - to demonstrate that they are ignored?

petebacondarwin (approved):

src/ng/http.js
gkalpak: I do want to explicitly accept URLs (but ignore the path etc), but I think `xsrfWhitelistedUrls` will be misleading in that case. It would sound like you can whitelist specific URLs.

How about I change the description to _"...containing origins that are trusted..."_ and keep the name as is (and also mention below that path etc will be ignored)?

gkalpak:
@petebd, PTAL.

petebacondarwin (approved):

src/ng/http.js
petebacondarwin: Changing the description works for me.

Issue #12731, opened 2015-09-02
select multi: $render is called twice when model reference changes


tepez:
This is a followup on a previous [issue](https://github.com/angular/angular.js/issues/11329). I've just found this issue on my own directives and I think it'd be better fixed in angular as well.

The `$render` function of a multi-select element is called twice whenever the model value changes by reference.
See plnker: http://plnkr.co/edit/1npwJi5tCS7gGLafYfRp?p=preview

After a model changes, $render is called first from [ngModel](https://github.com/angular/angular.js/blob/4175860af1ea51ddc525a73f586506fa14f35816/src/ng/directive/ngModel.js#L840).
It will be be called again from the viewValue [watch](https://github.com/angular/angular.js/blob/master/src/ng/directive/select.js#L368) of the select directive.

This can be fixed by keeping the last rendered value instead of the last view value in select:

```
// https://github.com/angular/angular.js/blob/master/src/ng/directive/select.js#L368

var lastRendered;

ngModelCtrl.$render = function() {
  selectCtrl.writeValue(ngModelCtrl.$viewValue);
  if (attr.multiple) {
    lastRendered = shallowCopy(ngModelCtrl.$viewValue);
  }
};

...

if (attr.multiple) {

  ...
  // we have to do it on each watch since ngModel watches reference, but
  // we need to work of an array, so we need to see if anything was inserted/removed
  var lastViewRef = NaN;
  scope.$watch(function selectMultipleWatch() {
    if (lastViewRef === ngModelCtrl.$viewValue && !equals(lastRendered, ngModelCtrl.$viewValue)) {
      ngModelCtrl.$render();
    }
    lastViewRef = ngModelCtrl.$viewValue;
  });

}
```

To make the [spec](https://github.com/angular/angular.js/blob/master/test/ng/directive/selectSpec.js#L926) cover this case. Just add an additional $digest after every $apply (before checking the number of $render calls).

petebacondarwin:
This is a performance issue rather than a consistency issue, right?

petebacondarwin:
Really we should be fixing this at the ngModel level, which at the moment doesn't handle "complex" model values very well.

tepez:
Yes, it's a performance and not a consistency issue.

PR #14894, opened 2016-07-11
fix(ngOptions): remove selected attribute from unselected options

### LINKED ISSUE #14419: "Selected" attributes for <select> options elements not being updated properly
**Request Type:** bug

**Current Behavior:**
When a select element is created with both ng-options and ng-model attributes to track the selected option, the "selected" element property is added to the selected attribute (per this issue: #8366). However, when a different option is selected, the "selected" attribute is not removed, resulting in all of the options containing a selected="selected" attribute (if all options have been selected). This is important due to accessibility guidelines stipulate the use of the selected attribute on options to indicate what option has been selected. [Aria Guidelines] (http://www.w3.org/TR/WCAG20-TECHS/H91).

**Reproduce?**
Any select box that has a ng-model tracking the value within an ng-options attribute for a select element. This is a modification of the Angular docs example for ng-options that illustrates nicely the select attribute not being added and removed properly: https://plnkr.co/edit/PjbWbo

**Expected behavior?**
Only one selected="selected" attributes should be present within the options of a select element. When a selection is changed the attribute should be removed and added to the newly selected option.

As per mentioned above, having one option with a selected attribute allows for proper accessibility functionality.

**Versions**
This was tested in Angular 1.4.1 and 1.5.3 in Chrome 49, Canary 52, Safari 9 and Firefox 42

I found this was addressed in commits 79538af and 3d393dc

### LINKED ISSUE #14892: Select ng-options is dont reset with '<input type="reset"> '

Bug

When I have a select with ng-options with an array with objects, the behavior is weird when I reset the form by clicking in a  `<input type="reset" value="clear">`

The problem occurs because when you select an option angular put a attribute `selected="selected"` on the option and never clears the attribute in the others options.

Chosse different options in the select and then press the reset button.

You can watch the error here: https://plnkr.co/edit/Epq5fVrS0I6p2csrntli?p=preview

select-dom

The standard reset form button works properly

The standard reset form button doesn't works properly when ng-options is used with a array with objects.

1.5.5, 1.5.7

The partial solution of this issue is commenting the line of the angular.js 1.5.5

28012  `// option.element.setAttribute('selected', 'selected')`;

Narretz:
bug fix

See #14892 nad #14419

No

**Other information**:

When the select model changes, we add the "selected" attribute to the selected option, so that
screen readers know which option is selected.

Previously, we failed to remove the attribute when the model changed to match a different option, or
the unknown / empty option.

When using "track by", the behavior would also show when a user selected an option, and then the
model was changed, because track by watches the tracked expression, and calls the $render function
on change.

This fix reads the current select value, finds the matching option and removes the "selected"
attribute.

Fixes #14892
Fixes #14419
Related #12731

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -448,9 +448,10 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
       var removeEmptyOption = function() {
         if (!providedEmptyOption) {
           emptyOption.remove();
+        } else {
+          emptyOption.removeAttr('selected');
         }
       };
-
 
       var renderUnknownOption = function() {
         selectElement.prepend(unknownOption);
@@ -467,7 +468,12 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
       if (!multiple) {
 
         selectCtrl.writeValue = function writeNgOptionsValue(value) {
+          var selectedOption = options.selectValueMap[selectElement.val()];
           var option = options.getOptionFromViewValue(value);
+
+          // Make sure to remove the selected attribute from the previously selected option
+          // Otherwise, screen readers might get confused
+          if (selectedOption) selectedOption.element.removeAttribute('selected');
 
           if (option) {
             // Don't update the option when it is already selected.
@@ -509,6 +515,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
 
         // If we are using `track by` then we must watch the tracked value on the model
         // since ngModel only watches for object identity change
+        // FIXME: When a user selects an option, this watch will fire needlessly
         if (ngOptions.trackBy) {
           scope.$watch(
             function() { return ngOptions.getTrackByValue(ngModelCtrl.$viewValue); },
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -120,6 +120,47 @@ describe('ngOptions', function() {
             return { pass: errors.length === 0, message: message };
           }
         };
+      },
+      toBeMarkedAsSelected: function() {
+        // Selected is special because the element property and attribute reflect each other's state.
+        // IE9 will wrongly report hasAttribute('selected') === true when the property is
+        // undefined or null, and the dev tools show that no attribute is set
+        return {
+          compare: function(actual) {
+            var errors = [];
+            if (actual.selected === null || typeof actual.selected === 'undefined' || actual.selected === false) {
+              errors.push('Expected option property "selected" to be truthy');
+            }
+
+            if (msie !== 9 && actual.hasAttribute('selected') === false) {
+              errors.push('Expected option to have attribute "selected"');
+            }
+
+            var result = {
+              pass: errors.length === 0,
+              message: errors.join('\n')
+            };
+
+            return result;
+          },
+          negativeCompare: function(actual) {
+            var errors = [];
+            if (actual.selected) {
+              errors.push('Expected option property "selected" to be falsy');
+            }
+
+            if (msie !== 9 && actual.hasAttribute('selected')) {
+              errors.push('Expected option not to have attribute "selected"');
+            }
+
+            var result = {
+              pass: errors.length === 0,
+              message: errors.join('\n')
+            };
+
+            return result;
+          }
+        };
       }
     });
   });
@@ -743,6 +784,41 @@ describe('ngOptions', function() {
     expect(options.eq(3)).toEqualOption('c');
   });
 
+
+  it('should remove the "selected" attribute from the previous option when the model changes', function() {
+    scope.values = [{id: 10, label: 'ten'}, {id:20, label: 'twenty'}];
+
+    createSelect({
+      'ng-model': 'selected',
+      'ng-options': 'item.label for item in values'
+    }, true);
+
+    var options = element.find('option');
+    expect(options[0]).toBeMarkedAsSelected();
+    expect(options[1]).not.toBeMarkedAsSelected();
+    expect(options[2]).not.toBeMarkedAsSelected();
+
+    scope.selected = scope.values[0];
+    scope.$digest();
+
+    expect(options[0]).not.toBeMarkedAsSelected();
+    expect(options[1]).toBeMarkedAsSelected();
+    expect(options[2]).not.toBeMarkedAsSelected();
+
+    scope.selected = scope.values[1];
+    scope.$digest();
+
+    expect(options[0]).not.toBeMarkedAsSelected();
+    expect(options[1]).not.toBeMarkedAsSelected();
+    expect(options[2]).toBeMarkedAsSelected();
+
+    scope.selected = 'no match';
+    scope.$digest();
+
+    expect(options[0]).toBeMarkedAsSelected();
+    expect(options[1]).not.toBeMarkedAsSelected();
+    expect(options[2]).not.toBeMarkedAsSelected();
+  });
 
   describe('disableWhen expression', function() {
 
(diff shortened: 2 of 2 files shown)


test/ng/directive/ngOptionsSpec.js
```diff
@@ -743,6 +743,37 @@ describe('ngOptions', function() {
...
 
+  it('should remove the "selected" attribute from the previous option when the model changes', function() {
+    scope.values = [{id: 10, label: 'ten'}, {id:20, label: 'twenty'}];
+
+    createSelect({
+      'ng-model': 'selected',
+      'ng-options': 'item.label for item in values'
+    }, true);
+
+    var resetButton = angular.element('<input type="reset" />');
```
gkalpak: Isn't the reset button irrelevant. If so, I would avoid complicating the testcase.

test/ng/directive/ngOptionsSpec.js
```diff
@@ -743,6 +743,37 @@ describe('ngOptions', function() {
...
+    formElement.append(resetButton);
+
+    scope.selected = scope.values[0];
+    scope.$digest();
+
+    var options = element.find('option');
+    expect(options[1].getAttribute('selected')).toBe('selected');
```
gkalpak: I would feel more comfortable, if you checked all options for the `selected` attribute (here and below).

test/ng/directive/ngOptionsSpec.js
```diff
@@ -1395,6 +1426,28 @@ describe('ngOptions', function() {
         });
       }).not.toThrow();
     });
+
+    it('should remove the "selected" attribute when the model changes', function() {
```
gkalpak: _when the model changes_  -->  _when the view changes_ (?)

test/ng/directive/ngOptionsSpec.js
```diff
@@ -1395,6 +1426,28 @@ describe('ngOptions', function() {
...
+
+      var options = element.find('option');
+      browserTrigger(options[2], 'click');
+
+      expect(scope.selected).toEqual(scope.arr[1]);
+
+      scope.selected = {};
+      scope.$digest();
+
+      options[2].selected = false;
```
gkalpak: Why is this necessary?

Narretz: Sorry, left over from a test inside the test

test/ng/directive/ngOptionsSpec.js
Narretz: Technically, the problem appears when you update the model (which then updates the view)

test/ng/directive/ngOptionsSpec.js
```diff
@@ -743,6 +743,37 @@ describe('ngOptions', function() {
     expect(options.eq(3)).toEqualOption('c');
   });
 
+  it('should remove the "selected" attribute from the previous option when the model changes', function() {
+    scope.values = [{id: 10, label: 'ten'}, {id:20, label: 'twenty'}];
+
+    createSelect({
+      'ng-model': 'selected',
+      'ng-options': 'item.label for item in values'
```
fovotz: I think is missing the "track by id", sorry if I wrong

Narretz: There's a different test for track by

test/ng/directive/ngOptionsSpec.js
Narretz: Good catch, because "selected" isn't removed from the empty option

gkalpak: It's based on personal experience. I tried a fix for this myself and I missed the empty option the first time :smile:

Narretz: Ah, sorry I should have assigned myself at the beginning, so we didn't have to duplicate work

gkalpak: I can't blame, since I did the same :stuck_out_tongue:

Narretz:
Sorry for the potential ton of update emails you got, @gkalpak But now it's ready for another review

test/ng/directive/ngOptionsSpec.js
```diff
@@ -744,6 +778,42 @@ describe('ngOptions', function() {
...
+    scope.$digest();
+
+    expect(options[0]).not.toBeMarkedAsSelected();
+    expect(options[1]).not.toBeMarkedAsSelected();
+    expect(options[2]).toBeMarkedAsSelected();
+
+    scope.selected = 'no match';
+    scope.$digest();
+
+    expect(options[0].selected).toBe(true);
```
gkalpak: Is this necessary? Isn't `.toBeMarkedAsSelected()` enough?

src/ng/directive/ngOptions.js
```diff
@@ -467,9 +467,15 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
...
         selectCtrl.writeValue = function writeNgOptionsValue(value) {
+          var selectedOption = options.selectValueMap[selectElement.val()];
           var option = options.getOptionFromViewValue(value);
 
+          // Make sure to remove the selected attribute from the previously selected option
+          // Otherwise, screen readers might get confused
+          if (selectedOption) selectedOption.element.removeAttribute('selected');
+
           if (option) {
+            if (providedEmptyOption) emptyOption.removeAttr('selected');
```
gkalpak: I think ti would be safer to move this inside the `removeEmptyOption()` function, s we don't have to manually call it every time.

For example, there is currently a call for `removeEmptyOption()` inside `selectCtrl.readValue()`. Do we need to also remove the `selected` attribute? Probably not, but who knows :stuck_out_tongue:

gkalpak:
A couple of minor comments/questions. Otherwise LGTM :+1:

src/ng/directive/ngOptions.js
Narretz: I guess not, because I think when the empty option is removed and then re-added, we need it to be selected. But moving it into the fn it saves precious CPU cycle!

Narretz: I removed this line completely and all the tests pass ... not good

src/ng/directive/ngOptions.js
```diff
@@ -446,12 +446,12 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
       };
 
       var removeEmptyOption = function() {
+        emptyOption.removeAttr('selected');
```
gkalpak: Minor (as it isn't expected to happen often in an app's lifecycle), but this could be moved to an `else` branch (since we only care about the attribute if we aren't removing the element).

test/ng/directive/ngOptionsSpec.js
```diff
@@ -120,6 +120,55 @@ describe('ngOptions', function() {
...
+            var errors = [];
+            if (actual.selected === null || typeof actual.selected === 'undefined' || actual.selected === false) {
+              errors.push('Expected option property "selected" to be truthy');
+            }
+
+            if (msie !== 9 && actual.hasAttribute('selected') === false) {
+              errors.push('Expected option to have attribute "selected"');
+            }
+
+            var message = function() {
```
gkalpak: This function doesn't seem necessary. (You can just `message: errors.join('\n')`.)

gkalpak:
Do the tests fail now without the `emptyOption.removeAttr('selected')` line?

Narretz:
Yep, the tests fail now, when I remove the remoeAttr code. Thanks for the review!

dustin-page:
Thanks for working on this. However, I noticed it is still an issue in AngularJS 1.5.11. When I use ngOptions with track by the selected attributes for the unselected options are not removed.

ng-options-multiple-selected-attributes

Here is a link to Plnkr example:
http://plnkr.co/edit/tpl:fdTouZ7cjNtUDbeZTYnY?p=preview

gkalpak:
It happens on latest master too. @dustin-page can you please open a new issue. This has probably something to do with how you use `track by`.

dromerop:
using 1.6.8 and got same error when using track by: changing the selected element adds the selected attribute to the current item but doesn't remove it from the old one.

PR #14902, opened 2016-07-12
refactor($animate): avoid unnecessary lookup

gkalpak:
Refactoring. (Theoretically a perf improvement, but it is so insignificant, that it's not worth mentioning as such :stuck_out_tongue: )

In `$animate.enabled()`, we look up in `disabledElementsLookup` even if we never use the value.

In `$animate.enabled()`, we look up in `disabledElementsLookup` only if we need the value.

No.

**Other information**:
No :confused:


petebacondarwin:
LGTM

PR #14903, opened 2016-07-12
fix($sanitize): blacklist the `usemap` attribute

gkalpak:
Backport of 234053f.

BREAKING CHANGE:

The `$sanitize` service will now remove instances of the `usemap` attribute from any elements passed to it.

This attribute is used to reference another element by `name` or `id`. Since the `name` and `id` attributes are already blacklisted, a sanitized `usemap` attribute could only reference unsanitized content, which is a security risk.

### DIFF
--- a/karma-shared.conf.js
+++ b/karma-shared.conf.js
@@ -45,8 +45,8 @@ module.exports = function(config, specificOptions) {
       'SL_Safari': {
         base: 'SauceLabs',
         browserName: 'safari',
-        platform: 'OS X 10.9',
-        version: '7'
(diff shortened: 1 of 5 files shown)


petebacondarwin:
LGTM

attritionorg:
Is this fix related to http://pivotal.io/security/cve-2016-0926?

gkalpak:
@attritionorg , I doubt it, since this is a backport of 234053f, which has been fixed since January 22, 2016.

gkalpak:
Merged as ac0d528 and dd4ce50.

PR #14906, opened 2016-07-13
fix($resource): allow params in `hostname` (except for IPv6 addresses)

### LINKED ISSUE #14542: ngResource does not replace parameters in the host
Bug

When you use a parameter (i.e. :domain) in the hostname, it is not replaced correctly.  Instead, it is kept as the parameter name with the colon intact.

Here is the broken plunkr using 1.4.5: http://plnkr.co/edit/9m0b2er2P6Gb3Jx9Bppo?p=preview
Here is the working version using 1.4.3: http://plnkr.co/edit/RsI3TgKwcjGEXcTMKoQR?p=preview

It should replace any parameters in the domain, ignoring IPv6 hosts and ports correctly.

This behavior was standard in ngResource 1.4.3.  As of 1.4.5, it changed in order to support IPv6 URLs (see #12512).  The fix implemented in that issue was too aggressive and effectively ignores hostnames/domains from parameter replacement.

I make use of this behavior heavily in order to support a multi-tenanted application where the the sub-domain decides the tenant.  I am not the only one either: http://stackoverflow.com/questions/32463584/passing-domain-name-as-parameter-with-resource-in-angularjs#

1.4.5 is where the issue first appeared (I didn't see a 1.4.4 of ngResource), and 1.4.3 worked correctly as described above.

I think if we can simply add a regex to catch IPv6 addresses and for all other cases keep the previous behavior, including the same handling of port numbers (if the param contains only numbers, ignore it). I'm not sure of IPv6 syntax, but in the examples given in the linked issue, something as simple as looking for square brackets around the hostname may catch it.

gkalpak:
Bug fix (regression).

Params inside the `hostname` part of a URL are ignored.
See #14542.

Params inside the `hostname` part of a URL are not ignored (except if the `hostname` is a `[...]` enclosed IPv6 address).

Hardly.

**Other information**:
Support for IPv6 addresses (in b643f0d) was too aggressive and broke support for params in the
`hostname` part of a URL.
This commit restores support for params in the `hostname`, as long as it is not an IPv6 address.

Fixes #14542

### DIFF
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -321,11 +321,51 @@ describe('basic usage', function() {
   });
 
   it('should support IPv6 URLs', function() {
-    var R = $resource('http://[2620:0:861:ed1a::1]/:ed1a/', {}, {}, {stripTrailingSlashes: false});
-    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/foo/').respond({});
-    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/').respond({});
-    R.get({ed1a: 'foo'});
-    R.get({});
+    var keepSlashes = {stripTrailingSlashes: false};
+
+    var R1 = $resource('http://[2620:0:861:ed1a::1]', null, null, keepSlashes);
+    var R2 = $resource('http://[2620:0:861:ed1a::1]/', null, null, keepSlashes);
+    var R3 = $resource('http://[2620:0:861:ed1a::1]/:ed1a', null, null, keepSlashes);
+    var R4 = $resource('http://[2620:0:861:ed1a::1]/:ed1a/', null, null, keepSlashes);
+
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]').respond(null);        // R1
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/').respond(null);       // R2
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/foo').respond(null);    // R3
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/').respond(null);       // R3
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/foo/').respond(null);   // R4
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/').respond(null);       // R4
+
+    R1.get({ed1a: 'foo'});
+    R2.get({ed1a: 'foo'});
+    R3.get({ed1a: 'foo'});
+    R3.get({});
+    R4.get({ed1a: 'foo'});
+    R4.get({});
+  });
+
+  it('should support params in the `hostname` part of the URL', function() {
+    var keepSlashes = {stripTrailingSlashes: false};
+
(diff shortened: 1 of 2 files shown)


Narretz:
Didn't Pete comment on this?
Anyway, my regex foo isn't very strong ...
So inside Ipv6 urls, you cannot make param replacements. But can you put ipv6 urls in params that replace the host?

gkalpak:
@Narretz: Pete asked for a clarification, but didn't comment further iirc :worried:

--
> But can you put ipv6 urls in params that replace the host?

Good question. You can't actually, because we are encoding the leading/trailing `[`/`]`, so the URL breaks.

There is also a more general bug (that could be a feature, but I don't think it is), that `:param` inside of parameter values are also replaced:

```js
$resource('http://:foo/:bar').get({foo: ':bar.com', bar: 'baz'});
//  --> GET http://baz.com/baz
```

This shouldn't be a problem usually, but it may affect IPv6 URLs as param values.

--
I will take another look at how we can address these issues.

petebacondarwin:
Any update on this one @gkalpak ?

gkalpak:
Not yet. Will work on it this week.

gkalpak:
A couple of thoughts after having looked into it again:

Currently, we have the following limitations:

1. There can be no param replacements inside a static IPv6 URL. E.g. `$resource('http://[2620:0:861:ed1a::1]').get({ed1a: 'foo'})` will call `http://[2620:0:861:ed1a::1]`.  
    _AFAICT, this is a good thing, since you don't want accidental replacements and IPv6 URLs are full of `:xyz` (which look like params)._

2. While you can set the hostname via a param (e.g. `$resource('http://:hostname')`), you can't set it to an IPv6 URL, because the quare brackets (`[...]`) will be URI-encoded.  
    _This is minor imo, but we might be able to address it._

3. A `:xyz` inside a param **value** may be treated as a param itself and replaced (depending on the relative order of params). E.g. `$resource('http://:foo/:bar').get({foo: ':bar.com', bar: 'baz'})` will call `http://baz.com/baz`.  
   _This could also be a feature, and "fixing" it might break apps relying on it. On the other hand it was never documented (not intended probably) and is not hard to do manually. It doesn't currently affect IPv6 URLs, due to (2), but if (2) were to be addressed, this would become more of a concern.

test/ngResource/resourceSpec.js
```diff
@@ -321,11 +321,51 @@ describe('basic usage', function() {
...
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/').respond(null);       // R3
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/foo/').respond(null);   // R4
+    $httpBackend.expect('GET', 'http://[2620:0:861:ed1a::1]/').respond(null);       // R4
+
+    R1.get({ed1a: 'foo'});
+    R2.get({ed1a: 'foo'});
+    R3.get({ed1a: 'foo'});
+    R3.get({});
+    R4.get({ed1a: 'foo'});
+    R4.get({});
```
petebacondarwin: And similarly for the other new test?

petebacondarwin:
I think this is good enough for now. Being able to insert non-IPv6 hosts via parameters is better than we had before. If someone then asks for that support we can look at it again.
The whole thing about inserting parameters into parameters has not been a problem so far, so I don't think we should try to fix that either, right now.

So let's get this PR merged. Thanks @gkalpak

PR #14910, opened 2016-07-14
Pr 14893

petebacondarwin:

fix

`ng-view` directives inside async loaded templates do not display the correct view if the location change occurred before $route was instantiated.

Now we track all location changes before $route is instantiated, which are then used to initialize the current route.

No (I don't think)

**Other information**:


petebacondarwin:
This actually creates more potential problems than it creates and #14893 is simpler.

PR #14912, opened 2016-07-15
Update Protractor to 4.0.0 and use FF 47 in e2e tests on Travis

Narretz:


gkalpak:
According to the [changelog][1], we should not be using `getInnerHtml()` with v4.x. There seem to be a couple of places where we still use them (e.g. [here][2] and [there][3]).
They still seem to work, but if it is easy to get rid of them, I would go for it (not necessary though).

Can you also make sure that this doesn't break with 482a3ae (which involves quite a bit of scrolling)?
Other than that LGTM :+1:  

[1]: https://github.com/angular/protractor/blob/master/CHANGELOG.md#400
[2]: https://github.com/angular/angular.js/blob/83bc2479675fd1e0e023c5be1dbcbeb2e2932c6d/src/ng/sce.js#L632-L645
[3]: https://github.com/angular/angular.js/blob/83bc2479675fd1e0e023c5be1dbcbeb2e2932c6d/src/ngSanitize/sanitize.js#L108-L138

Narretz:
@gkalpak You don't know by any chance what the migration for getInnerHTML ist? Simply accessing the property?

gkalpak:
From what I understand `element(...).getAttribute('innerHTML')` might work (since it will return the value of the property, if no attribute is found.
If not, I am afraid you have to use `browser.executeScript()`.

PR #14914, opened 2016-07-15
feat($animate): add support for `customFilter`

### LINKED ISSUE #14891: ngAnimate: Set class name filter to include also children elements?
Here's an example: http://plnkr.co/edit/jQUwh3KsZ1uR3lr3TrLK?p=preview  
I want to set `classNameFilter` setting to apply to the children elements as it applies to the parent which contains class name in filter. So, how to achieve that as I haven't seen such option (so I guess this is a feature request)?  
Thanks in advance!

gkalpak:
Feature.

This commit adds a new `customFilter()` function on `$animateProvider` (similar to `classNameFilter()`), which can be used to filter animations (i.e. decide whether they are allowed or not), based on the return value of a custom filter function.
This allows to easily create arbitrarily complex rules for filtering animations, such as allowing specific events only, or enabling animations on specific subtrees of the DOM etc.

No.

**Other information**:
Fixes #14891.

Needs docs:
- [x] Complete the API docs.
- [x] Update the Developer Guide.

### DIFF
--- a/docs/content/guide/animations.ngdoc
+++ b/docs/content/guide/animations.ngdoc
@@ -6,20 +6,26 @@
 
 # Animations
 
-AngularJS provides animation hooks for common directives such as `ngRepeat`, `ngSwitch`, and `ngView`, as well as custom directives
-via the `$animate` service. These animation hooks are set in place to trigger animations during the life cycle of various directives and when
-triggered, will attempt to perform a CSS Transition, CSS Keyframe Animation or a JavaScript callback Animation (depending on if an animation is
-placed on the given directive). Animations can be placed using vanilla CSS by following the naming conventions set in place by AngularJS
-or with JavaScript code when it's defined as a factory.
+AngularJS provides animation hooks for common directives such as
+{@link ng.directive:ngRepeat ngRepeat}, {@link ng.directive:ngSwitch ngSwitch}, and
+{@link ngRoute.directive:ngView ngView}, as well as custom directives via the `$animate` service.
+These animation hooks are set in place to trigger animations during the life cycle of various
+directives and when triggered, will attempt to perform a CSS Transition, CSS Keyframe Animation or a
+JavaScript callback Animation (depending on whether an animation is placed on the given directive).
+Animations can be placed using vanilla CSS by following the naming conventions set in place by
+AngularJS or with JavaScript code, defined as a factory.
 
 <div class="alert alert-info">
-  Note that we have used non-prefixed CSS transition properties in our examples as the major browsers now support non-prefixed
-  properties. If you intend to support older browsers or certain mobile browsers then you will need to include prefixed
-  versions of the transition properties.  Take a look at http://caniuse.com/#feat=css-transitions for what browsers require prefixes,
-  and https://github.com/postcss/autoprefixer for a tool that can automatically generate the prefixes for you.
+  Note that we have used non-prefixed CSS transition properties in our examples as the major
+  browsers now support non-prefixed properties. If you intend to support older browsers or certain
+  mobile browsers then you will need to include prefixed versions of the transition properties. Take
+  a look at http://caniuse.com/#feat=css-transitions for what browsers require prefixes, and
+  https://github.com/postcss/autoprefixer for a tool that can automatically generate the prefixes
+  for you.
 </div>
 
-Animations are not available unless you include the {@link ngAnimate `ngAnimate` module} as a dependency within your application.
+Animations are not available unless you include the {@link ngAnimate `ngAnimate` module} as a
+dependency of your application.
 
 Below is a quick example of animations being enabled for `ngShow` and `ngHide`:
 
@@ -59,8 +65,9 @@ You may also want to setup a separate CSS file for defining CSS-based animations
 
 ## How they work
 
-Animations in AngularJS are completely based on CSS classes. As long as you have a CSS class attached to a HTML element within
-your website, you can apply animations to it. Lets say for example that we have an HTML template with a repeater in it like so:
+Animations in AngularJS are completely based on CSS classes. As long as you have a CSS class
+attached to a HTML element within your application, you can apply animations to it. Lets say for
+example that we have an HTML template with a repeater like so:
 
 ```html
 <div ng-repeat="item in items" class="repeated-item">
@@ -68,22 +75,21 @@ your website, you can apply animations to it. Lets say for example that we have 
 </div>
 ```
 
-As you can see, the `.repeated-item` class is present on the element that will be repeated and this class will be
-used as a reference within our application's CSS and/or JavaScript animation code to tell AngularJS to perform an animation.
+As you can see, the `repeated-item` class is present on the element that will be repeated and this
+class will be used as a reference within our application's CSS and/or JavaScript animation code to
+tell AngularJS to perform an animation.
 
-As ngRepeat does its thing, each time a new item is added into the list, ngRepeat will add
-a `ng-enter` class name to the element that is being added. When removed it will apply a `ng-leave` class name and when moved around
-it will apply a `ng-move` class name.
+As `ngRepeat` does its thing, each time a new item is added into the list, `ngRepeat` will add an
+`ng-enter` class to the element that is being added. When removed it will apply an `ng-leave` class
+and when moved around it will apply an `ng-move` class.
 
-Taking a look at the following CSS code, we can see some transition and keyframe animation code set for each of those events that
-occur when ngRepeat triggers them:
+Taking a look at the following CSS code, we can see some transition and keyframe animation code set
+up for each of those events that occur when `ngRepeat` triggers them:
 
 ```css
 /*
-  We're using CSS transitions for when
-  the enter and move events are triggered
-  for the element that has the .repeated-item
-  class
+  We are using CSS transitions for when the enter and move events
+  are triggered for the element that has the `repeated-item` class
 */
 .repeated-item.ng-enter, .repeated-item.ng-move {
   transition: all 0.5s linear;
@@ -91,10 +97,8 @@ occur when ngRepeat triggers them:
 }
 
 /*
- The ng-enter-active and ng-move-active
- are where the transition destination properties
- are set so that the animation knows what to
- animate.
+  `.ng-enter-active` and `.ng-move-active` are where the transition destination
+  properties are set so that the animation knows what to animate
 */
 .repeated-item.ng-enter.ng-enter-active,
 .repeated-item.ng-move.ng-move-active {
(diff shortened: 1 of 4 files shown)


test/ngAnimate/animateSpec.js
```diff
@@ -292,6 +292,129 @@ describe("animations", function() {
...
 
+    describe('customFilter()', function() {
+      it('should clear the `customFilter` if no function is passed',
+        module(function($animateProvider) {
+          expect($animateProvider.customFilter()).toBeNull();
+
+          $animateProvider.customFilter(angular.noop);
+          expect($animateProvider.customFilter()).toEqual(jasmine.any(Function));
+
+          $animateProvider.customFilter({});
```
petebacondarwin: perhaps we should throw instead of silently ignoring?

petebacondarwin:
One minor question - otherwise LGTM

test/ngAnimate/animateSpec.js
gkalpak: I kept this consistent with `classNameFilter` (although it is preferrable to throw obviously).
Throwing would be a BC for `classNameFilter`. Should I fix this one to throw and make `classNameFilter` throw for 1.6 (or leave it as is)?

petebacondarwin: I would change both as you suggest.

Narretz:
I like it. I would add a note though that this going to be called for every attempted animation, so developers should keep the function lean.

test/ngAnimate/animateSpec.js
gkalpak: I missed that this needed to land in 1.6.0 (due to the BC) :(
I will make only `customFilter` throw for 1.6.x/1.7.x and change `classNameFilter` to also throw for 1.7.x only.

gkalpak: Actually, I just realized that the current implementation allows to "clear" the filter/className by passing a non-Function/non-RegExp argument. So, if we want to throw, we need to still allow some special value (e.g. `null`/`undefined`) for unsetting the filter/className.

gkalpak:
I added docs for `customFilter()` (and large-ish docs commit that face-lifts the Animations Guide - sorry).
I didn't make any change wrt throwing on "invalid" argument types, because of https://github.com/angular/angular.js/pull/14914#r94752256.

Feel free to TAL ☺️

src/ng/animate.js
```diff
@@ -233,6 +234,45 @@ var $AnimateProvider = ['$provide', /** @this */ function($provide) {
 
   /**
    * @ngdoc method
+   * @name $animateProvider#customFilter
+   *
+   * @description
```
Narretz: I would add two things:
- How does this relate to the classFilter, .i.e. what takes precedence, or do both run
- Note that this runs for every `animatable` action, not just for animations that are actually defined, and that the user should keep the function fast

You've added this to the guide, but I think it should be here too

docs/content/guide/animations.ngdoc
```diff
@@ -6,20 +6,25 @@
 
 # Animations
 
-AngularJS provides animation hooks for common directives such as `ngRepeat`, `ngSwitch`, and `ngView`, as well as custom directives
-via the `$animate` service. These animation hooks are set in place to trigger animations during the life cycle of various directives and when
-triggered, will attempt to perform a CSS Transition, CSS Keyframe Animation or a JavaScript callback Animation (depending on if an animation is
-placed on the given directive). Animations can be placed using vanilla CSS by following the naming conventions set in place by AngularJS
-or with JavaScript code when it's defined as a factory.
+AngularJS provides animation hooks for common directives such as `ngRepeat`, `ngSwitch`, and
```
Narretz: While you are at it, can you add links to the directives here?

docs/content/guide/animations.ngdoc
```diff
@@ -255,17 +259,19 @@ myModule.directive('my-directive', ['$animate', function($animate) {
...
+after the page has been loaded. Animations are disabled, so that UI and content are instantly
+visible. Otherwise, with many animations on the page, the loading process may become too visually
+overwhelming, and the performance may suffer.
 
-Internally, `ngAnimate` waits until all template downloads that are started right after bootstrap have finished.
-Then, it waits for the currently running {@link ng.$rootScope.Scope#$digest} and the one after that to finish.
-This ensures that the whole app has been compiled fully before animations are attempted.
+Internally, `ngAnimate` waits until all template downloads that are started right after bootstrap
+have finished. Then, it waits for the currently running {@link ng.$rootScope.Scope#$digest $digest}
+and the one more after that to finish. This ensures that the whole app has been compiled fully
```
Narretz: "and one more after that, to finish."

docs/content/guide/animations.ngdoc
```diff
@@ -294,42 +324,44 @@ app.config(function($animateProvider) {
...
 This function can be used to enable / disable animations in two different ways:
 
-With a single `boolean` argument, it enables / disables animations globally: `$animate.enabled(false)`
-disables all animations in your app.
+With a single `boolean` argument, it enables / disables animations globally:
+`$animate.enabled(false)` disables all animations in your app.
 
 When the first argument is a native DOM or jqLite/jQuery element, the function enables / disables
 animations on this element *and all its children*: `$animate.enabled(myElement, false)`. This is the
 most flexible way to change the animation state. For example, even if you have used it to disable
```
Narretz: Is "This is the most flexible way to change the animation state." still true with the new customFilter? I don't think so, as $animate.enabled() does not work when the classNameFilter or customFilter are active. So let's remove this sentence

docs/content/guide/animations.ngdoc
```diff
@@ -379,66 +412,71 @@ In that case, you can add styles to the CSS that make sure the element stays hid
...
+higher priority than any styles that are not related to `ngAnimate`.
 
-You can also use one of the two other {@link guide/animations#how-to-selectively-enable-disable-and-skip-animations strategies to disable animations}.
+You can also use one of the other
+{@link guide/animations#how-to-selectively-enable-disable-and-skip-animations
+strategies to disable animations}.
 
 
-### Enable animations for elements outside of the Angular application DOM tree: {@link ng.$animate#pin $animate.pin()}
+### Enable animations outside of the application DOM tree: {@link ng.$animate#pin $animate.pin()}
```
Narretz: This should be `##`

Narretz (changes requested):

src/ng/animate.js
gkalpak: > How does this relate to the classFilter, .i.e. what takes precedence, or do both run

👍 

> Note that this runs for every animatable action, not just for animations that are actually defined, and that the user should keep the function fast

You mean in addition to the alert-box below?

gkalpak:
Comments addressed. PTAL

src/ng/animate.js
Narretz: If possible, I thought we should make it explicit (in the alert box) that this doesn't only affect the actual animations, but every "action" (class change, element entering). I think this difference is easily forgotten, and people might think "Oh well I don't have many animations anyway".

Narretz (changes requested):

gkalpak:
Note to self: Update links after #15654.

src/ng/animate.js
gkalpak: I see what you mean now. Good point. I will update the API docs and guide.
BTW, I think we should re-order the conditions, because now `customFilter` and `classNameFilter` are executed even if animations are globally disabled or the document is hidden etc.

Narretz:
This looks (almost) done, wanna put it into 1.6.5?

Narretz:
@gkalpak Looks like this needs a rebase. After that, I will do another review - or do you have some other changes to make?

gkalpak:
I have squashed and rebased this and added two new commits:
- One for further clarifying when the `customFilter` function is called.
- One for skipping the `customFilter` and `classNameFilter` checks, if animations are globally disabled.

@Narretz PTAL

docs/content/guide/animations.ngdoc
```diff
@@ -59,191 +65,190 @@ You may also want to setup a separate CSS file for defining CSS-based animations
...
     addClass: function(element, className, done) {},
     removeClass: function(element, className, done) {}
   }
 });
 ```
 
 With these generated CSS class names present on the element at the time, AngularJS automatically
-figures out whether to perform a CSS and/or JavaScript animation. If both CSS and JavaScript animation
-code is present, and match the CSS class name on the element, then AngularJS will run both animations at the same time.
+figures out whether to perform a CSS and/or JavaScript animation. If both CSS and JavaScript
```
Narretz: That actually hasn't been true since 1.4 😱  https://code.angularjs.org/snapshot/docs/guide/migration#animation-nganimate-

docs/content/guide/animations.ngdoc
```diff
@@ -59,191 +65,190 @@ You may also want to setup a separate CSS file for defining CSS-based animations
...
+## Class and `ngClass` animation hooks
 
-AngularJS also pays attention to CSS class changes on elements by triggering the **add** and **remove** hooks.
-This means that if a CSS class is added to or removed from an element then an animation can be executed in between,
-before the CSS class addition or removal is finalized. (Keep in mind that AngularJS will only be
-able to capture class changes if an **expression** or the **ng-class** directive is used on the element.)
+AngularJS also pays attention to CSS class changes on elements by triggering the **add** and
+**remove** hooks. This means that if a CSS class is added to or removed from an element then an
+animation can be executed in between, before the CSS class addition or removal is finalized.
+(Keep in mind that AngularJS will only be able to capture class changes if an **expression** or the
```
Narretz: I think **interpolated expression** is clearer here

Narretz (changes requested):
Just two smaller docs changes needed, otherwise LGTM

docs/content/guide/animations.ngdoc
gkalpak: So, what happens if both are detected. Who wins?

Narretz: JS animations - because you can run CSS in JS: http://plnkr.co/edit/rdRAZTiXHNAFTEOHyZVo?p=preview

Although I'm surprised this isn't explicitly said in the migrate docs. But it is here: https://code.angularjs.org/snapshot/docs/api/ngAnimate

Unfortuante spreading out of content ... moving this in guide would be a good passion project for rainy afternoons. :p

gkalpak: Changed and added a link to that `api/ngAnimate` section.
There is indeed useful info there that would be better in the guide 😞

docs/content/guide/animations.ngdoc
gkalpak: Fixed.

Narretz (approved):

PR #14936, opened 2016-07-20
chore(travis): update sauce-connect

gkalpak:


Narretz:
Any particular reason for this update? Just for testing?

gkalpak:
I saw that message from SauceConnect that there is a newer version available and I thought it might make it less flaky. (It doesn't seem to be the case - I had to restart a couple of jobs to get this to pass :smile:)

PR #14942, opened 2016-07-22
fix($parse): correctly escape unsafe identifier characters 

gkalpak:
Bug fix.

While it is possible to use arbitrary identifier characters in Angular expressions, the RegExp that detect unsafe identifiers (in order to escape them in the generated JS code), was failing to detect them in some cases.

No.

**Other information**:
This PR also includes some minor clean-up changes and fixes the `setLiteral()` tests to run in CSP mode too.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -398,8 +398,7 @@ AST.prototype = {
 
   filterChain: function() {
     var left = this.expression();
-    var token;
-    while ((token = this.expect('|'))) {
+    while (this.expect('|')) {
       left = this.filter(left);
     }
     return left;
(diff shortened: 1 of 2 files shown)


lgalfaso:
LGTM

src/ng/parse.js
```diff
@@ -1260,7 +1261,7 @@ ASTCompiler.prototype = {
   },
 
   nonComputedMember: function(left, right) {
-    var SAFE_IDENTIFIER = /[$_a-zA-Z][$_a-zA-Z0-9]*/;
+    var SAFE_IDENTIFIER = /^[$_a-zA-Z][$_a-zA-Z0-9]$/;
```
jbedard: Don't we need to keep the `*`? Otherwise almost everything will use the `['...']` format...

gkalpak: You are right. I must have accidentally removed it :innocent:

PR #14951, opened 2016-07-26
fix($parse): block assigning to fields of a constructor prototype

gkalpak:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
Bug fix.

**What is the current behavior? (You can also link to an open issue here)**
See #14939.

**Does this PR introduce a breaking change?**
No.

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**Other information**:
1. **1st commit:** Add a few more tests for constructor access.
2. **2nd commit:** Fix + tests.
3. **3rd commit:** Add missing error page.

PR #14957, opened 2016-07-27
perf($parse): improve performance of assignment expressions

gkalpak:
Minor perf improvement.

No.

**Other information**:
There was a ~5% improvement in the `parsed-expressions-bp/assignment` benchmark (which only contains assignment expressions). In real-world applications, the time spent in assignment expressions will be a tiny fragment of the overall processing time (so this change is hardly expected to make any noticeable difference).

Related to #14951.

### DIFF
--- a/benchmarks/parsed-expressions-bp/app.js
+++ b/benchmarks/parsed-expressions-bp/app.js
@@ -17,11 +17,10 @@ app.directive('bmPeWatch', function() {
   return {
     restrict: 'A',
     compile: function($element, $attrs) {
-      $element.text( $attrs.bmPeWatch );
+      $element.text($attrs.bmPeWatch);
       return function($scope, $element, $attrs) {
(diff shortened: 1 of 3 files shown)


gkalpak:
Btw, this essentially voids the prototype check if any of the banned prototypes changes dynamically (i.e. after we have cached it). Considering these checks are not security-related and that dynamically changing prototypes of core constructors is likely to break other things as well, I don't think it should be a concern.

/fyi @petebacondarwin

petebacondarwin:
I note that we were already caching `Function.prototype.call`, `Function.prototype.apply` and `Function.prototype.bind`. So I am confident that this is OK. LGTM.
Can we get another LGTM from someone, perhaps @IgorMinar or @lgalfaso?

lgalfaso:
LGTM

PR #14964, opened 2016-07-29
Make select and select+ngOptions consistent

Narretz:
Half refactoring, half fix.

It makes the concepts of unknown / empty option consistent.

When the model does not match any option, the select generates an unknown option with the value
'? hashedModelVal ?' (select) or simply '?' (ngOptions) and inserts it into the select. This
option is then auto-selected by the browser. Once the user selects an option / the application sets
a matching model, the unknown option is removed from the select. It is therefore not user-selectable.

Alternatively, the application can provide an option with the value='', the so-called empty option.
The empty option takes over the role of the unknown option, in that it is selected when there's no
matching model, but it is not removed when a option and model match, and it can also be selected by
the user. The empty option has the value='' and sets the model to  if selected.

Previously, these concepts where a bit mushy, especially in ngOptions.

This fix now delegates the bulk of the work to the selectCtrl, as the concept of unknown and empty
option are the same for both directives. Only the generation of the empty option is different.

The commit also removes a test for an unsupported use-case: multiple empty options. The docs are
clear that only a single empty option is allowed. We support dynamically generated empty options,
but not multiple ones.

Possibly for users that used multiple empty options in their selects. This is undocumented behavior however.

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -416,15 +416,14 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
 
       // The emptyOption allows the application developer to provide their own custom "empty"
       // option when the viewValue does not match any of the option values.
-      var emptyOption;
       for (var i = 0, children = selectElement.children(), ii = children.length; i < ii; i++) {
         if (children[i].value === '') {
-          emptyOption = children.eq(i);
+          selectCtrl.emptyOption = children.eq(i);
           break;
         }
       }
 
-      var providedEmptyOption = !!emptyOption;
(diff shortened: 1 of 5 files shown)


petebacondarwin:
1st commit LGTM

petebacondarwin:
2nd commit LGTM although I think that rather than removing the test we should change it to demonstrate the expected behaviour in the case that a second empty option is added

petebacondarwin:
2nd commit LGTM

Narretz:
Landed as 2785ad72599ca5f9558a116baecd83a5bebe3292 and c75698df55f5a026bcd7fcecbb9d4ff0bc3ebc3e

PR #14970, opened 2016-07-30
feat(input[range]): support step

Narretz:
feat

Input range doesn't handle the step attribute wrt to validation and model updates

No

**TODO**
- [ ] Add tests for step + min + max
- [x] Add docs
- [x] Find out exactly what is broken in Chrome

**Other information**:

Wrt the Chrome bug mentioned in https://github.com/angular/angular.js/pull/14870#issuecomment-235829620, it looks like setting a number isn't the problem, but that Chrome has a bug when you change the step value multiple times. Not sure yet, though.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -679,7 +679,17 @@ var inputType = {
    * @param {string} ngModel Assignable angular expression to data-bind to.
    * @param {string=} name Property name of the form under which the control is published.
    * @param {string=} min Sets the `min` validation error key if the value entered is less than `min`.
+   *    Can be interpolated.
    * @param {string=} max Sets the `max` validation error key if the value entered is greater than `max`.
+   *    Can be interpolated.
+   * @param {string=} ngMin Like `min`, sets the `min` validation error key if the value entered is less than `ngMin`,
+   *    but does not trigger HTML5 native validation. Takes an expression.
+   * @param {string=} ngMax Like `max`, sets the `max` validation error key if the value entered is greater than `ngMax`,
+   *    but does not trigger HTML5 native validation. Takes an expression.
+   * @param {string=} step Sets the `step` validation error key if the value entered does not fit the `step` constraint.
+   *    Can be interpolated.
+   * @param {string=} ngStep Like `step`, sets the `max` validation error key if the value entered does not fit the `ngStep` constraint,
+   *    but does not trigger HTML5 native validation. Takes an expression.
    * @param {string=} required Sets `required` validation error key if the value is not entered.
    * @param {string=} ngRequired Adds `required` attribute and `required` validation constraint to
    *    the element when the ngRequired expression evaluates to true. Use `ngRequired` instead of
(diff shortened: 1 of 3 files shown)


Narretz:
@gkalpak I can't replicate the bug you described in https://github.com/angular/angular.js/pull/14870#issuecomment-235829620. Here's my repro: http://plnkr.co/edit/cBc2YAySJzD8y0NukAEQ?p=preview
What happens instead is that Chrome sets the value to "60", which is also weird, as "40" (like Firefox sets it) is much closer to "45" (the previous value).

There's also another test case (fn angTestCase()) that shows how Chrome stops adjusting the element value based on the step when the value has been modified previously. 

I think I'll raise both bugs in the Webkit tracker, but I'm totally prepared that this is following the spec in some obscure way ...

gkalpak:
I can't reproduce it on Chrome v52 either. (I was on v51 when I reproduced it. It got probably fixed,  introducing another bug :wink:)

Narretz:
Reported as https://bugs.chromium.org/p/chromium/issues/detail?id=633075 (stops adjusting) and https://bugs.chromium.org/p/chromium/issues/detail?id=633074 (adjusts to wrong value)

src/ng/directive/input.js
```diff
@@ -1570,6 +1574,64 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+      // the max value, which might be interpolated
+      element.attr('step', attr.step);
+    }
+
+    // This initalizes the step value, so that non-support browsers validate with the correct
+    // values during the initial $render
+    stepChange(attr.step);
+    attr.$observe('step', stepChange);
+  }
+
```
petebacondarwin: There is a lot of similar (almost duplicate) code here. In the interests of keeping the code size down is it possible to move this into a helper function? Perhaps:

```js
setupHandlers(type, attrType, noopValidationFn, validatorFn, changeFn) {
  if (attrType) {
    ctrl.$validators[type] = attributeType === type && supportsRange ? noopValidatorFn : validatorFn;
    if (attrType === type) {
      element.attr(type, attr[type]);
    }
    changeFn(attr[type]);
    attr.$observe(type, changeFn);
  }
}
```

petebacondarwin:
One refactoring comment. Other than that LGTM

Narretz:
@petebacondarwin Sorry Pete, the current state of the PR is outdated :o I just pushed the updates. 
It also included step support for input number.

The main difference for range to before is that we don't support ngStep, because the browser would still use the default step of 1, and we don't test the step change updates anymore, because not all browsers implement them.

src/ng/directive/input.js
```diff
@@ -1642,6 +1690,21 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     }
   }
 
+  function stepChange(val) {
+    stepVal = parseNumberAttrVal(val);
```
petebacondarwin: Perhaps do this after the check for NaN below to save unnecessary computation?

petebacondarwin: Oh - this is a side-effect? OK

petebacondarwin:
I feel sad that we don't support `ngStep` for native range inputs. It feels confusing for the developer to have it but only for non-native ranges.

Narretz:
That's why I didn't document it at all. Because it's not really useful.

Narretz:
Landed as e1da4bed8e291003d485a8ad346ab80bed8ae2e3 and 9a8b8aaa9641bfc85d245c4355fa5481285b593f

PR #14972, opened 2016-07-31
fix($httpBackend): complete the request on timeout

### LINKED ISSUE #14969: $http promise does not settle when the connection attempt times out on current Safari
When using $http to connect to a server which does no react to the TCP connection attempts at all on a current Safari browser, the promise never gets settled. On other browsers it gets rejected after some time as expected.

It can be reproduced using angular 1.5.8 and the following code on Safari 9.1.2, OS X 10.11.6:
```
$http({
  method: "GET",
  url: "http://github.com:12345/"
}).finally(function (result) {
  console.log("settled");
});
```

The cause seems to be that the current Safari version invokes the ontimeout handler on the XHR object in this case even though no timeout has been set on it. (https://xhr.spec.whatwg.org/#the-timeout-attribute). Attaching the ontimeout handler in addition to onerror and onabort in $httpBackend fixes the issue.

gkalpak:
Bug fix.

`$httpBackend` fails to detect a request timeout, when `.timeout` has been specified on the XHR (or any timeout in Safari 9).
See #14969.

`$httpBackend` detects request timeouts.

No.

**Other information**:
When using the [timeout attribute](https://xhr.spec.whatwg.org/#the-timeout-attribute) and an XHR request times out, browsers trigger the `timeout` event (and execute the XHR's `ontimeout` callback). Additionally, Safari 9 handles timed-out requests in the same way, even if no `timeout` has been explicitly set on the XHR.
In the above cases, `$httpBackend` would fail to capture the XHR's completing (with an error), so the corresponding `$http` promise would never get fulfilled.

Fixes #14969.

### DIFF
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -108,6 +108,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
 
       xhr.onerror = requestError;
       xhr.onabort = requestError;
+      xhr.ontimeout = requestError;
 
       forEach(eventHandlers, function(value, key) {
           xhr.addEventListener(key, value);
--- a/test/ng/httpBackendSpec.js
+++ b/test/ng/httpBackendSpec.js
@@ -155,10 +155,11 @@ describe('$httpBackend', function() {
   });
 
   it('should not try to read response data when request is aborted', function() {
-    callback.and.callFake(function(status, response, headers) {
+    callback.and.callFake(function(status, response, headers, statusText) {
       expect(status).toBe(-1);
       expect(response).toBe(null);
       expect(headers).toBe(null);
+      expect(statusText).toBe('');
     });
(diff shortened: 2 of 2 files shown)


Narretz:
So all other browser call the onError handler when no onTimeout is specified?

gkalpak:
Almost :smiley: When no `timeout` (not `ontimeout`) is specified Chrome calls `onerror`. Safari calls `ontimeout` even if not `timeout` is set on the XHR instance.

(I didn't test other browsers, but since nobody has complained thus far, I assume most of them follow the behavior of Chrome.)

Narretz:
Ok, I've tested in FF, IE11 and Edge and they all behave like Chrome (i.e. they call ontimeout only when a timeout is set). I was a bit worried about calling event handlers twice, but I don't think that can happen.
I would mention in the commit message that we don't actually use the `ontimeout` handler when it's set in http. We use $browser.defer for some reason, and then abort the xhr.

Otherwise, LGTM

Issue #14959, opened 2016-07-27
Back button does not work when using html5Mode without <base> tag


onlywei:
***Note*: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
This is a little difficult to explain so bear with me. To get to this point, a number of workarounds of existing limitations were used:

1. `$locationProvider.html5Mode({enabled: true, requireBase: false})` is turned on. Base tag is avoided in order for `<svg>` icons with `xlink:href` to be used.
2. Since there is no base tag, in order to actually leave the SPA and go to another page on the same server, we have to manually call `location.reload()` when a URL that does not belong to our routers configured routes is detected.
3. After `location.reload()`, the new page loads as expected, but the back button does not work. It changes the URL, but the browser never asks the server for a page refresh.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar**

http://plnkr.co/edit/LyCi3ikziOttlI7lZfCv?p=preview

To see the problem, open the plnkr demo, Read through the code a bit, and then click the link to bring you to the html5mode page. In that page, you should see another link that will bring you back to first page. Then, click the back button in your browser. It should not work!

**What is the expected behavior?**
I expect the browser to completely forget that Angular ever did anything to its URL change behavior when I load a new page via `location.reload()`. When I load the 2nd destination page, it may be an Angular page, or it may not be. In both cases the back button should take me back to the original html5mode angular page.

**What is the motivation / use case for changing the behavior?**

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**
1.5.6, 1.5.7 - did not test older versions

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
I have noticed that if I inject `$location` into the page that I expect to press the back button in, it works! But, I don't want to have to inject `$location` into that page for no reason other than to fix this.

gkalpak:
[...]

PR #14976, opened 2016-08-01
feat(location): add configurable attr to force html5mode link rewriting

onlywei:
Feature

When `rewriteLinks` is disabled in Location's `html5mode`, all links are treated as if they should reload from the server.

Also see this discussion: https://github.com/angular/angular.js/issues/14959

This allows the user to add an `ng-noserver` attribute to the `<a>` tag, like this:
`<a href="somewhere" ng-noserver>click me</a>` to explicitly mark this link as one that should **not** reload from server, i.e. to use angular javascript routing instead.

No

**Other information**:

See this discussion: https://github.com/angular/angular.js/issues/14959

When using html5mode, sometimes it is necessary to not use the <base> tag
in order to support SVG icons. An example of this is in the discussion
linked above.

When we do this, sometimes we also want to disable "rewriteLinks" so that
we can still leave the Angular application via navigation.

This feature allows the user to explicitly mark an <a> tag as a link that
should not refresh from server, even if rewriteLinks is disabled.

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -750,8 +750,12 @@ function $LocationProvider() {
    *     whether or not a <base> tag is required to be present. If `enabled` and `requireBase` are
    *     true, and a base tag is not present, an error will be thrown when `$location` is injected.
    *     See the {@link guide/$location $location guide for more information}
-   *   - **rewriteLinks** - `{boolean}` - (default: `true`) When html5Mode is enabled,
-   *     enables/disables url rewriting for relative links.
+   *   - **rewriteLinks** - `{boolean|String}` - (default: `true`) When html5Mode is enabled,
+   *     enables/disables url rewriting for relative links. If set to a string, url rewriting will
+   *     only happen on links with an attribute that matches the given string. For example, if set
+   *     to "internal-link", then the URL will only be rewritten for `<a internal-link>` links. Note
+   *     that attribute name (de)normalization is not happening here. So "internal-link" is
+   *     different from "internalLink".
    *
    * @returns {Object} html5Mode object if used as getter or itself (chaining) if used as setter
    */
@@ -769,7 +773,7 @@ function $LocationProvider() {
         html5Mode.requireBase = mode.requireBase;
       }
 
-      if (isBoolean(mode.rewriteLinks)) {
+      if (isBoolean(mode.rewriteLinks) || isString(mode.rewriteLinks)) {
         html5Mode.rewriteLinks = mode.rewriteLinks;
       }
 
@@ -866,10 +870,11 @@ function $LocationProvider() {
     }
 
     $rootElement.on('click', function(event) {
+      var rewriteLinks = html5Mode.rewriteLinks;
       // TODO(vojta): rewrite link when opening in new tab/window (in legacy browser)
       // currently we open nice url link and redirect then
 
-      if (!html5Mode.rewriteLinks || event.ctrlKey || event.metaKey || event.shiftKey || event.which === 2 || event.button === 2) return;
+      if (!rewriteLinks || event.ctrlKey || event.metaKey || event.shiftKey || event.which === 2 || event.button === 2) return;
 
       var elm = jqLite(event.target);
 
@@ -878,6 +883,8 @@ function $LocationProvider() {
         // ignore rewriting if no A tag (reached root element, or no parent - removed from document)
         if (elm[0] === $rootElement[0] || !(elm = elm.parent())[0]) return;
       }
+
+      if (isString(rewriteLinks) && isUndefined(elm.attr(rewriteLinks))) return;
 
       var absHref = elm.prop('href');
       // get the actual href attribute - see
--- a/test/ng/locationSpec.js
+++ b/test/ng/locationSpec.js
@@ -1588,8 +1588,36 @@ describe('$location', function() {
 
 
     it('should not rewrite links when rewriting links is disabled', function() {
-      configureTestLink({linkHref: 'link?a#b', html5Mode: {enabled: true, rewriteLinks:false}, supportHist: true});
+      configureTestLink({linkHref: 'link?a#b'});
       initService({html5Mode:{enabled: true, rewriteLinks:false},supportHistory:true});
+      inject(
+        initBrowser({ url: 'http://host.com/base/index.html', basePath: '/base/index.html' }),
+        setupRewriteChecks(),
+        function($browser) {
+          browserTrigger(link, 'click');
+          expectNoRewrite($browser);
+        }
+      );
+    });
+
+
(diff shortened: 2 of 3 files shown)


src/ng/location.js
```diff
@@ -877,6 +879,8 @@ function $LocationProvider() {
         if (elm[0] === $rootElement[0] || !(elm = elm.parent())[0]) return;
       }
 
+      if (!html5Mode.rewriteLinks && isUndefined(elm.attr('ng-noserver'))) return;
```
gkalpak: This will not recognize prefixed forms (such as `x-ng-...` or `data-ng-...`), but I am not sure this is an issue.

src/ng/location.js
```diff
@@ -749,7 +749,9 @@ function $LocationProvider() {
    *     true, and a base tag is not present, an error will be thrown when `$location` is injected.
    *     See the {@link guide/$location $location guide for more information}
    *   - **rewriteLinks** - `{boolean}` - (default: `true`) When html5Mode is enabled,
-   *     enables/disables url rewriting for relative links.
+   *     enables/disables url rewriting for relative links. Note that if this is disabled, you can
+   *     still force a link to be rewritten by adding the "ng-noserver" attr to the `<a>` element:
+   *     `<a href="somewhere" ng-noserver>click me</a>`.
```
gkalpak: Another idea would be allowing the users to set `rewriteLinks` to a string, indicating the attribute that forces rewriting. E.g.:

- `rewriteLinks: true`: Always rewrite links.
- `rewriteLinks: false`: Never rewrite links.
- `rewriteLinks: 'some-attr'`: Only rewrite links that have the `some-attr` attribute.

gkalpak:
I think `noserver` is not a very intuitive name. Also URL-rewriting doesn't guarantee that no request will be done to the server afaict.

onlywei:
Thanks @gkalpak, I was hoping to get this kind of useful feedback. I've updated the code. Please review at your leisure.

src/ng/location.js
```diff
@@ -748,8 +748,10 @@ function $LocationProvider() {
    *     whether or not a <base> tag is required to be present. If `enabled` and `requireBase` are
    *     true, and a base tag is not present, an error will be thrown when `$location` is injected.
    *     See the {@link guide/$location $location guide for more information}
-   *   - **rewriteLinks** - `{boolean}` - (default: `true`) When html5Mode is enabled,
-   *     enables/disables url rewriting for relative links.
+   *   - **rewriteLinks** - `{boolean|String}` - (default: `true`) When html5Mode is enabled,
+   *     enables/disables url rewriting for relative links. If set to a string, url rewriting will
+   *     only happen on links with an attribute that matches the given string. For example, if set
+   *     to "internalLink", then the URL will only be rewritten for `<a internal-link>` links.
```
gkalpak: There is no attribute name (de)normalization going on. In order to match `<a internal-link>` you need to set `rewriteLinks` to `'internal-link'` (not `'internalLink'`).

test/ng/locationSpec.js
```diff
@@ -1601,6 +1601,25 @@ describe('$location', function() {
     });
 
 
+    it('should rewrite links when the specified rewriteLinks attr is detected', function() {
```
gkalpak: We need another test (or entend this test if possible), that verifies that the URL is not rewritten if the link doesn't have the correct attribute. (I.e. I think this test would pass even before the PR.)

test/ng/locationSpec.js
```diff
@@ -1601,6 +1601,25 @@ describe('$location', function() {
...
+      configureTestLink({
+        linkHref: 'link?a#b',
+        attrs: 'force-rewrite',
+        html5Mode: {enabled: true, rewriteLinks:false},
+        supportHist: true
+      });
+      initService({html5Mode:{enabled: true, rewriteLinks:'forceRewrite'},supportHistory:true});
```
gkalpak: Unless I am wrong, this test should fail (due to https://github.com/angular/angular.js/pull/14976/files#r73798407), but there are currently no failures on Travis (at least not one related to this test :wink:). Which makes me wonder, does this test even run? Needs investigation...

gkalpak:
Btw, docs need to be updated in other places (e.g. the `Using $location` section of the developer guide).
Adding it here as a reminder, but feel free to defer that until the implementation has finalized.

test/ng/locationSpec.js
onlywei: You're right, it looks like none of these tests are being run, I wonder why...

test/ng/locationSpec.js
onlywei: Done

src/ng/location.js
onlywei: Done

test/ng/locationSpec.js
onlywei: It wasn't the tests, there is a configuration normalizer that forces configurations to be boolean, that I just updated.

test/ng/locationSpec.js
```diff
@@ -2357,19 +2394,32 @@ describe('$location', function() {
...
 
 
-      it('should only overwrite existing properties if values are boolean', function() {
+      it('should only overwrite existing properties if values are the correct type', function() {
         module(function($locationProvider) {
           $locationProvider.html5Mode({
             enabled: 'duh',
             requireBase: 'probably',
-            rewriteLinks: 'nope'
+            rewriteLinks: false
```
gkalpak: This change doesn't test what it should. You need to supply an unsupported type and verify that it didn't get copied to the actual config.

test/ng/locationSpec.js
```diff
@@ -1601,6 +1601,43 @@ describe('$location', function() {
     });
 
 
+    it('should rewrite links when the specified rewriteLinks attr is detected', function() {
+      configureTestLink({
+        linkHref: 'link?a#b',
+        attrs: 'force-rewrite',
+        html5Mode: {enabled: true, rewriteLinks:false},
+        supportHist: true
```
gkalpak: I don't know if you copied this from somewhere else, but `configureTestLink` doesn't seem to support `html5Mode` or `supportHist` :confused:

test/ng/locationSpec.js
onlywei: Done

test/ng/locationSpec.js
onlywei: I copied it from here: https://github.com/angular/angular.js/blob/master/test/ng/locationSpec.js#L1591

gkalpak:
LGTM

gkalpak:
I'll let @petebacondarwin do the final LGTM, since he is the `$location` master :smiley:

onlywei:
@petebacondarwin Ping. Please review when convenient.

petebacondarwin:
LGTM

PR #14980, opened 2016-08-03

nikhilknoldus:

**docs** Document Update.

Application stops, when this error occur we get official [link](https://docs.angularjs.org/error/ng/areq?p0=SourceCtrl&p1=not) but this does not solve the error.

## 

No

**Other information**:

Encountered this error many times, the provided details are very encrypted (don't seem practical), i have added a line that is 99% practical and main cause for this error. These 3 points that i have added are seriously core reasons for this error.

### DIFF
--- a/docs/content/error/ng/areq.ngdoc
+++ b/docs/content/error/ng/areq.ngdoc
@@ -6,3 +6,5 @@
 AngularJS often asserts that certain values will be present and truthy using a
 helper function. If the assertion fails, this error is thrown. To fix this problem,
 make sure that the value the assertion expects is defined and truthy.
+Check and make sure that provided controller name is exact, this may be a syntax error
+as well and you have included app-name and check that the current controller is included in root html.
--- a/docs/content/error/ng/areq.ngdoc
+++ b/docs/content/error/ng/areq.ngdoc
@@ -6,5 +6,5 @@
 AngularJS often asserts that certain values will be present and truthy using a
 helper function. If the assertion fails, this error is thrown. To fix this problem,
 make sure that the value the assertion expects is defined and truthy.
-Check and make sure that provided controller name is exact, this may be a syntax error
-as well and you have included app-name and check that the current controller is included in root html.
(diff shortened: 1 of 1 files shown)


nikhilknoldus:
I signed it.

nikhilknoldus:
CI failed due to your server timeout, not mine. Please review.

Narretz:
The error is too specific. `areq` is not only thrown for controllers, but for other parts of the code, too.

nikhilknoldus:
@Narretz your view is right but the doc's lines are not worth at all, written so complicatedly, number of angular developers i know who open this bug line from browser console, read it and say what's this? They don't infer anything from this detail. The comment which i have written matching major error criteria and the earlier doesn't.

Narretz:
But your change only mentions controller, which is is also confusing. If someone gets the error without having ngController, then this is not helpful at all.
I'm okay with including some note about syntax errors, but it should be less specific. If you have example code that shows a common example how this error can happen, then we can include it.

nikhilknoldus:
@Narretz Martin your view is again right in this case, but as i mentioned, syntax error and earlier error doesn't infer it, and yes my commit is focusing on controller as you said, i can modify it for larger criteria. And for the example thing: my project is quite big (for an international bank), i do have lots of controllers too and in that case watching this error horribly blank our(js developers) mind when something like this happens: Examples:
1-    angular.module('someApp').controller(-'SomeCntrl', SomeCtrl);
*notice the hyphen before. it may be any symbol, char etc. That happens by mistake obviously.

and this is only 1 scenario there are much that we face alot.

I have updated my commit as well.

Thanks !!

nikhilknoldus:
Hi @Narretz any update regarding same.

Narretz:
I've taken another look and I think you are right that this error is confusing when it is thrown with for controllers. I've opened a PR that throws a better error here: https://github.com/angular/angular.js/pull/15015

nikhilknoldus:
Hey thanks Martin for considering my PR.

PR #14994, opened 2016-08-06
fix($parse): treat falsy values as defined in assignment expressions

### LINKED ISSUE #14990: fix($parse): treat falsy values as defined in assignment expressions
Bug.

When Angular `$parse`s an assignment expression - e.g. `'foo.toFixed = "bar"'` - it treats falsy values as if they don't exist in the context and overwrites them. I.e.:

```js
// Correct (because `foo` is truthy)
scope.foo = 1;
scope.$eval('foo.toFixed = "bar"');   // --> Error, trying to create property `toFixed` on 1 (expected)
console.log(scope.foo);   // --> `1`

// Incorrect (because `foo` is falsy
scope.foo = 0;
scope.$eval('foo.toFixed = "bar"');
console.log(scope.foo);   // --> `{toFixed: 'bar'}` (unexpected)
```

Falsy values should be treated as defined (the same as truthy values).

Doing the right thing :smiley:

All browsers.
I tested on Angular v1.5.8 and snapshot.

As suggested by @lgalfaso, the calls to `self.not(...)` should be replaced with `self.notNull(...)` at appropriate places (such as [here][1] and [there][2]).

[1]: https://github.com/angular/angular.js/blob/v1.5.8/src/ng/parse.js#L1026
[2]: https://github.com/angular/angular.js/blob/v1.5.8/src/ng/parse.js#L1052

m-amr:
bug,  solving Issue https://github.com/angular/angular.js/issues/14990

Closes #14990

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -944,7 +944,7 @@ ASTCompiler.prototype = {
           self.if_(self.stage === 'inputs' || 's', function() {
             if (create && create !== 1) {
               self.if_(
-                self.not(self.nonComputedMember('s', ast.name)),
+                self.isNull(self.nonComputedMember('s', ast.name)),
                 self.lazyAssign(self.nonComputedMember('s', ast.name), '{}'));
             }
             self.assign(intoId, self.nonComputedMember('s', ast.name));
@@ -973,7 +973,7 @@ ASTCompiler.prototype = {
             }
           } else {
             if (create && create !== 1) {
-              self.if_(self.not(self.nonComputedMember(left, ast.property.name)), self.lazyAssign(self.nonComputedMember(left, ast.property.name), '{}'));
+              self.if_(self.isNull(self.nonComputedMember(left, ast.property.name)), self.lazyAssign(self.nonComputedMember(left, ast.property.name), '{}'));
             }
             expression = self.nonComputedMember(left, ast.property.name);
             self.assign(intoId, expression);
@@ -1153,6 +1153,10 @@ ASTCompiler.prototype = {
 
   not: function(expression) {
     return '!(' + expression + ')';
+  },
+
+  isNull: function(expression) {
+    return expression + '==null';
   },
 
   notNull: function(expression) {
@@ -1546,7 +1550,7 @@ ASTInterpreter.prototype = {
   identifier: function(name, context, create, expression) {
     return function(scope, locals, assign, inputs) {
       var base = locals && (name in locals) ? locals : scope;
-      if (create && create !== 1 && base && !(base[name])) {
+      if (create && create !== 1 && base && base[name] == null) {
         base[name] = {};
       }
       var value = base ? base[name] : undefined;
@@ -1583,7 +1587,7 @@ ASTInterpreter.prototype = {
     return function(scope, locals, assign, inputs) {
       var lhs = left(scope, locals, assign, inputs);
       if (create && create !== 1) {
-        if (lhs && !(lhs[right])) {
+        if (lhs && lhs[right] == null) {
           lhs[right] = {};
         }
       }
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -3155,6 +3155,72 @@ describe('parser', function() {
           expect(isFunction(s.toString)).toBe(true);
           expect(l.toString).toBe(1);
         }));
+
+        it('should overwrite undefined / null scope properties when assigning', inject(function($parse) {
+          var scope;
+
+          scope = {};
+          $parse('a.b = 1')(scope);
+          $parse('c["d"] = 2')(scope);
+          expect(scope).toEqual({a: {b: 1}, c: {d: 2}});
+
+          scope = {a: {}};
+          $parse('a.b.c = 1')(scope);
+          $parse('a.c["d"] = 2')(scope);
+          expect(scope).toEqual({a: {b: {c: 1}, c: {d: 2}}});
+
+          scope = {a: undefined, c: undefined};
+          $parse('a.b = 1')(scope);
+          $parse('c["d"] = 2')(scope);
+          expect(scope).toEqual({a: {b: 1}, c: {d: 2}});
+
+          scope = {a: {b: undefined, c: undefined}};
+          $parse('a.b.c = 1')(scope);
+          $parse('a.c["d"] = 2')(scope);
+          expect(scope).toEqual({a: {b: {c: 1}, c: {d: 2}}});
+
+          scope = {a: null, c: null};
+          $parse('a.b = 1')(scope);
+          $parse('c["d"] = 2')(scope);
+          expect(scope).toEqual({a: {b: 1}, c: {d: 2}});
+
+          scope = {a: {b: null, c: null}};
+          $parse('a.b.c = 1')(scope);
+          $parse('a.c["d"] = 2')(scope);
+          expect(scope).toEqual({a: {b: {c: 1}, c: {d: 2}}});
+        }));
+
+        they('should not overwrite $prop scope properties when assigning', [0, false, '', NaN],
+          function(falsyValue) {
+            inject(function($parse) {
+              var scope;
+
+              scope = {a: falsyValue, c: falsyValue};
+              tryParseAndIgnoreException('a.b = 1');
+              tryParseAndIgnoreException('c["d"] = 2');
+              expect(scope).toEqual({a: falsyValue, c: falsyValue});
+
+              scope = {a: {b: falsyValue, c: falsyValue}};
+              tryParseAndIgnoreException('a.b.c = 1');
+              tryParseAndIgnoreException('a.c["d"] = 2');
+              expect(scope).toEqual({a: {b: falsyValue, c: falsyValue}});
+
+              // Helpers
+              //
+              // Normally assigning property on a primitive should throw exception in strict mode
+              // and silently fail in non-strict mode, IE seems to always have the non-strict-mode behavior,
+              // so if we try to use 'expect(function() {$parse('a.b=1')({a:false});).toThrow()' for testing
+              // the test will fail in case of IE because it will not throw exception, and if we just use
+              // '$parse('a.b=1')({a:false})' the test will fail because it will throw exception in case of Chrome
+              // so we use tryParseAndIgnoreException helper to catch the exception silently for all cases.
+              //
+              function tryParseAndIgnoreException(expression) {
+                try {
+                    $parse(expression)(scope);
+                } catch (error) {/* ignore exception */}
+              }
+            });
+          });
       });
 
       describe('literal', function() {


src/ng/parse.js
```diff
@@ -1274,6 +1274,10 @@ ASTCompiler.prototype = {
     return '!(' + expression + ')';
   },
 
+  null: function(expression) {
```
lgalfaso: Please rename to isNull

lgalfaso:
Looks like variations of the same issue are present at
https://github.com/angular/angular.js/blob/master/src/ng/parse.js#L1069
and
https://github.com/angular/angular.js/blob/master/src/ng/parse.js#L1080

with the expressions `foo.bar.baz = 1` and `foo.bar['baz'] = 1` where in both cases `foo = {bar: ''}`

It would be best to fix the three cases in one go.

m-amr:
@lgalfaso 

I have a problem with the unit test.

     it('should not assign a property to locals properties with falsy value', inject(function($parse) {
          var local = {a:0};
          expect(function() {
            $parse("a.b =1")(local);
          }).toThrow();
      });

I get this error 'Expected function to throw an exception' 
but when i call $parse("a.b =1")(local) alone i get
	TypeError: Cannot create property 'b' on number '0'

I have no idea why toThrow() doesn't detect that TypeError was thrown from parse function ?

lgalfaso:
There are at least 2 parts:

The `CSP: true` version also needs to be fixed, these are
https://github.com/angular/angular.js/blob/master/src/ng/parse.js#L1706
https://github.com/angular/angular.js/blob/master/src/ng/parse.js#L1731
https://github.com/angular/angular.js/blob/master/src/ng/parse.js#L1750

There might be discrepancies on how this is handled in different browsers.

m-amr:
@lgalfaso 
After the fix that i had done.
Here are the results of the three cases 
Case 1:
       
        $scope.a = 0;
        $scope.$eval("a.b=2");
I get TypeError: Cannot create property 'b' on number '0'

Case 2:

        $scope.a = 0;
        $scope.$eval("a['b']=2");
I get TypeError: Cannot create property 'b' on number '0'

Case 3:

        $scope.a = 0;
        $scope.$eval("a={'b':2}");
This shouldn't throw any error because assigning 'a' with another value.

Is that what mean ?

lgalfaso:
For some reasons that are not important, the following 3 expressions follow 3 slightly different code paths

```js
foo.bar = 1;
foo.bar.baz = 1;
foo.bar['baz'] = 1;
```

In this case we want to fix the three of them as they all have the same issue. This is, the following cases should behave in the same way

```js
$parse('foo.bar = 1')({foo: 0});
$parse('foo.bar.baz = 1')({foo: {bar: 0}});
$parse('foo.bar["baz"] = 1')({foo: {bar: 0}});
```

m-amr:
After I handled these two cases

      $parse('foo.bar.baz = 1')({foo: {bar: 0}});
      $parse('foo.bar["baz"] = 1')({foo: {bar: 0}});

Some old cases failed

         it('should prevent assigning only in the context of an actual constructor', function() {
          // foo.constructor is not a constructor.
          expect(function() {
            delete scope.foo;
            scope.$eval('foo.constructor[0] = ""', {foo: {constructor: ''}});
          }).not.toThrow();

and

       it('should prevent assigning only in the context of an actual prototype', function() {
          // foo.constructor.prototype is not a constructor prototype.
          expect(function() {
            delete scope.foo;
            scope.$eval('foo.constructor.prototype[0] = ""', {foo: {constructor: {prototype: ''}}});
          }).not.toThrow();

So there is these Cases are against the the change to handle 

        $parse('foo.bar.baz = 1')({foo: {bar: 0}});
        $parse('foo.bar["baz"] = 1')({foo: {bar: 0}});

so what is your suggestion ?

test/ng/parseSpec.js
```diff
@@ -3900,6 +3900,34 @@ describe('parser', function() {
           expect(isFunction(s.toString)).toBe(true);
           expect(l.toString).toBe(1);
         }));
+
+        it('should assign a property to locals properties with null or undefined value', inject(function($parse) {
```
gkalpak: The description is misleading. There is no locals object in the test. Use scope instead.

test/ng/parseSpec.js
```diff
@@ -3900,6 +3900,34 @@ describe('parser', function() {
...
+          $parse("a.b =1")(local);
+          expect(local.a).toEqual({b:1});
+        }));
+
+        it('should not assign a property to locals properties with falsy value', inject(function($parse) {
+          var local = {a:0};
+
+          expect(function() {
+            $parse("a.b =1")(local);
+          }).toThrow();
```
gkalpak: I would also check that `local.a` is still the original value after each `.toThrow()`.

m-amr:
@gkalpak 
Could you please check this.
I have no idea why toThrow function didn't detect any exception throw from parse function
when I write 

     
          var scope = {a:0};
          $parse("a.b =1")(scope);
I get this error.
TypeError: Cannot create property 'b' on number '0'
	    at fn (eval at compile (/Apache Server Root/repositories/angular.js/src/ng/parse.js:906:15), <anonymous>:4:248)
	    at Object.<anonymous> (/Apache Server Root/repositories/angular.js/test/ng/parseSpec.js:3916:27)
	    at Object.invoke (/Apache Server Root/repositories/angular.js/src/auto/injector.js:879:19)
	    at Object.workFn (/Apache Server Root/repositories/angular.js/src/ngMock/angular-mocks.js:3103:20)
	Error: Declaration Location
	    at window.inject.angular.mock.inject (/Apache Server Root/repositories/angular.js/src/ngMock/angular-mocks.js:3065:25)
	    at Suite.<anonymous> (/Apache Server Root/repositories/angular.js/test/ng/parseSpec.js:3914:81)
	    at Suite.<anonymous> (/Apache Server Root/repositories/angular.js/test/ng/parseSpec.js:3867:7)
	    at /Apache Server Root/repositories/angular.js/test/ng/parseSpec.js:1838:5
Chrome 52.0.2743 (Mac OS X 10.11.5): Executed 5535 of 5535 (2 FAILED) (22.486 secs / 21.484 secs)

but when i use 

          var scope = {a:0};
          expect(function() {
            $parse("a.b =1")(scope);
          }).toThrow();

I get 
Chrome 52.0.2743 (Mac OS X 10.11.5) parser csp: true locals should not assign a property to scope properties with falsy value FAILED
	Expected function to throw an exception.
	    at Object.<anonymous> (/Apache Server Root/repositories/angular.js/test/ng/parseSpec.js:3919:14)
	    at Object.invoke (/Apache Server Root/repositories/angular.js/src/auto/injector.js:879:19)
	    at Object.workFn (/Apache Server Root/repositories/angular.js/src/ngMock/angular-mocks.js:3103:20)

I have no Idea why the test didn't detect the thrown exception, but when i used    
$parse("a.b =1")(scope); it throw exception.

Does angular handle these exception in a different way ? or I am missing something.

gkalpak:
@m-amr, I see you pushed another commit. Were you able to solve the issue you had or should I take a look?

m-amr:
@gkalpak, No I still have the same problem, Could you please check it.

m-amr:
I have solved The Issue but I have a problem related to Internet Explorer, after i had added some unit tests for example

      var scope = {a:falsyValue};
               expect(function() {
                 $parse('a.b=1')(scope);
         }).toThrow();

I works well with Chrome and Firefox, but in case of IE it doesn't throw an exception, and I didn't found any test for the true values which will throw the same exception.

@gkalpak Do you have any suggestion for this problem ? should i ignore adding tests ?

jbedard:
Might be good to get this into 1.6, @gkalpak?

@m-amr do we need to test that `false.prop = 1` throws? How about just testing that `false` isn't replaced with `{}` like it previously did?  Or if we really need it to throw maybe we have to throw ourselves instead of relying on the browser?

gkalpak:
Hm...I was thinking of this as a fix (that would go into 1.5.x as well as 1.6.x). Do you think it should be treated as a breaking change?

jbedard:
I was thinking it would be a BC. People could be relying on the current behaviour, and the change could make existing apps start throwing if I understand it correctly?

m-amr:
@jbedard 
 We must test that false.prop = 1 throws, because if we didn't test it, and just test that false didn't replaced with {}. parse function will throw exception incase of chrome, firefox and the test will fail.
      
        var scope = {a:false};
        $parse('a.b=1')(scope); // will throw exception incase of chrome and test will fail.

I think that the right solution is throw custom exception and doesn't depend on the browser to throw it, as @jbedard  mentioned.
@gkalpak  what it your opinion about that ?

gkalpak:
Normally trying to assing a property on a primitive should throw in strict mode and silently fail in non-strict mode. IE seems to always have the non-strict-mode behavior. I don't think we should be fighting the browser here: If the browser throws: fine. if not: not our concern.

This shouldn't make any difference to end-users anyway and for developers I assume nobody is debugging or running their tests in IE only, so the error will come up in other browsers that do the right thing.

(AFAICT, there would be a slight overhead if we decided to throw ourselves, as we should check the actual type of the left-hand side at runtime. So, while I don't feel too strongly about explicitly throwing, I'd rather not.)

--
Regarding the tests, the following should pass on all browsers and verify what we really care about (i.e. that `$parse` did not overwrite the falsy value with an object):

```js
it('should overwrite `undefined`/`null` scope properties when assigning',
  inject(function($parse) {
    var scope;

    scope = {};
    $parse('a.b = 1')(scope);
    $parse('c["d"] = 2')(scope);
    expect(scope).toEqual({a: {b: 1}, c: {d: 2}});

    scope = {a: {}};
    $parse('a.b.c = 1')(scope);
    $parse('a.c["d"] = 2')(scope);
    expect(scope).toEqual({a: {b: {c: 1}, c: {d: 2}}});

    scope = {a: undefined, c: undefined};
    $parse('a.b = 1')(scope);
    $parse('c["d"] = 2')(scope);
    expect(scope).toEqual({a: {b: 1}, c: {d: 2}});

    scope = {a: {b: undefined, c: undefined}};
    $parse('a.b.c = 1')(scope);
    $parse('a.c["d"] = 2')(scope);
    expect(scope).toEqual({a: {b: {c: 1}, c: {d: 2}}});

    scope = {a: null, c: null};
    $parse('a.b = 1')(scope);
    $parse('c["d"] = 2')(scope);
    expect(scope).toEqual({a: {b: 1}, c: {d: 2}});

    scope = {a: {b: null, c: null}};
    $parse('a.b.c = 1')(scope);
    $parse('a.c["d"] = 2')(scope);
    expect(scope).toEqual({a: {b: {c: 1}, c: {d: 2}}});
  })
);

they('should not overwrite `$prop` scope properties when assigning', [false, 0, '', NaN],
  function(falsyValue) {
    inject(function($parse) {
      var scope;

      scope = {a: falsyValue, c: falsyValue};
      tryParse('a.b = 1');
      tryParse('c["d"] = 2');
      expect(scope).toEqual({a: falsyValue, c: falsyValue});

      scope = {a: {b: falsyValue, c: falsyValue}};
      tryParse('a.b.c = 1');
      tryParse('a.c["d"] = 2');
      expect(scope).toEqual({a: {b: falsyValue, c: falsyValue}});

      // Helpers
      function tryParse(expression) {
        try {
          $parse(expression)(scope);
        } catch (err) { /* Ignored */ }
      }
    });
  }
);
```

(I have slightly tweaked the tests to cover more cases, so please use these, @m-amr.)

gkalpak:
> I was thinking it would be a BC. People could be relying on the current behaviour, and the change could make existing apps start throwing if I understand it correctly?

@jbedard, indeed, they could start throwing due to relying on an undocumented bug. I hate this gray area between a bug fix and a breaking change 😠

m-amr:
@gkalpak  
Thanks for your feedback,  
I had added all the test cases that you had mentioned and every thing work, 
could give it a last look ?

test/ng/parseSpec.js
```diff
@@ -3155,6 +3155,72 @@ describe('parser', function() {
...
+              // and silently fail in non-strict mode, IE seems to always have the non-strict-mode behavior,
+              // so if we try to use 'expect(function() {$parse('a.b=1')({a:false});).toThrow()' for testing
+              // the test will fail in case of IE because it will not throw exception, and if we just use
+              // '$parse('a.b=1')({a:false})' the test will fail because it will throw exception in case of Chrome
+              // so we use tryParseAndIgnoreException helper to catch the exception silently for all cases.
+              //
+              function tryParseAndIgnoreException(expression) {
+                try {
+                    $parse(expression)(scope);
+                } catch (error) {/** ignore exception*/}
```
gkalpak: Nit: Use `/* ... */` (and consistent whitespace).

gkalpak:
LGTM
I'll let @lgalfaso sign it off :smiley:

test/ng/parseSpec.js
m-amr: Done.

lgalfaso:
LGTM
landed as 4f44e018948c45bfb07f0170de4f703d22778d71

PR #14996, opened 2016-08-06
Fix input range init

### LINKED ISSUE #14982: Input range doesn't respect interpolated min / max on initialization
See https://github.com/angular/angular.js/pull/14870#issuecomment-237301382

Narretz:

**Other information**:

This PR fixes 1 bigger and 2 smaller issues:
- correctly initialize with interpolated min / max values

The interpolation directive only sets the actual element attribute value after
a digest passed. That means previously the min/max values on input range were
not set when the first $render happened, so the browser would not adjust
the input value according to min/max. This meant the range input and model would
not be initialzed as expected.

With this change, input range will set the actual element attribute value during its own
linking phase, as it is already available on the attrs argument passed to the link fn.

Fixes #14982
- correctly sets the range value when min > max. Chrome and IE don't set this correctly in all cases.
- removes interpolation from an ng-maxlength attribute in a test (interpolation in an attribute that expects an expression is discouraged)

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1640,7 +1640,8 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
       // IE11 doesn't set the el val correctly if the maxVal is less than the element value
       if (maxVal < elVal) {
         element.val(maxVal);
-        elVal = minVal;
+        // IE11 and Chrome don't set the value to the minVal when max < min
+        elVal = maxVal < minVal ? minVal : maxVal;
       }
       ctrl.$setViewValue(elVal);
     } else {
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1061,8 +1061,11 @@ var inputType = {
    * Angular will also update the model value.
    *
    * Automatic value adjustment also means that a range input element can never have the `required`,
-   * `min`, or `max` errors, except when using `ngMax` and `ngMin`, which are not affected by automatic
-   * value adjustment, because they do not set the `min` and `max` attributes.
+   * `min`, or `max` errors.
+   *
+   * Note that `input[range]` is not compatible with`ngMax` and `ngMin`, because they do not set the
+   * `min` and `max` attributes, which means that the browser won't automatically adjust the input
+   * value based on their values, and will always assume min = 0 and max = 100.
    *
    * @param {string}  ngModel Assignable angular expression to data-bind to.
    * @param {string=} name Property name of the form under which the control is published.
@@ -1070,14 +1073,6 @@ var inputType = {
    *                  than `min`. Can be interpolated.
    * @param {string=} max Sets the `max` validation to ensure that the value entered is less than `max`.
    *                  Can be interpolated.
-   * @param {string=} ngMin Takes an expression. Sets the `min` validation to ensure that the value
-   *                  entered is greater than `min`. Does not set the `min` attribute and therefore
-   *                  adds no native HTML5 validation. It also means the browser won't adjust the
-   *                  element value in case `min` is greater than the current value.
-   * @param {string=} ngMax Takes an expression. Sets the `max` validation to ensure that the value
-   *                  entered is less than `max`. Does not set the `max` attribute and therefore
-   *                  adds no native HTML5 validation. It also means the browser won't adjust the
-   *                  element value in case `max` is less than the current value.
    * @param {string=} ngChange Angular expression to be executed when the ngModel value changes due
    *                  to user interaction with the input element.
    *
@@ -1551,8 +1546,8 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
       minVal = supportsRange ? 0 : undefined,
       maxVal = supportsRange ? 100 : undefined,
       validity = element[0].validity,
-      minAttrType = isDefined(attr.ngMin) ? 'ngMin' : isDefined(attr.min) ? 'min' : false,
-      maxAttrType = isDefined(attr.ngMax) ? 'ngMax' : isDefined(attr.max) ? 'max' : false;
+      hasMinAttr = isDefined(attr.min),
+      hasMaxAttr = isDefined(attr.max);
 
   var originalRender = ctrl.$render;
 
@@ -1565,20 +1560,20 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
     } :
     originalRender;
 
-  if (minAttrType) {
-    ctrl.$validators.min = minAttrType === 'min' && supportsRange ?
+  if (hasMinAttr) {
+    ctrl.$validators.min = supportsRange ?
       // Since all browsers set the input to a valid value, we don't need to check validity
       function noopMinValidator() { return true; } :
-      // ngMin doesn't set the min attr, so the browser doesn't adjust the input value as setting min would
+      // non-support browsers validate the range
       function minValidator(modelValue, viewValue) {
         return ctrl.$isEmpty(viewValue) || isUndefined(minVal) || viewValue >= minVal;
       };
 
-    setInitialValueAndObserver(minAttrType, 'min', minChange);
+    setInitialValueAndObserver('min', minChange);
   }
 
-  if (maxAttrType) {
-    ctrl.$validators.max = maxAttrType === 'max' && supportsRange ?
+  if (hasMaxAttr) {
+    ctrl.$validators.max = supportsRange ?
       // Since all browsers set the input to a valid value, we don't need to check validity
       function noopMaxValidator() { return true; } :
       // ngMax doesn't set the max attr, so the browser doesn't adjust the input value as setting max would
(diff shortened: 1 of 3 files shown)


src/ng/directive/input.js
```diff
@@ -1647,7 +1647,8 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
       // IE11 doesn't set the el val correctly if the maxVal is greater than the element value
```
gkalpak: I just noticed this comment should say "less than" (instead of "greater than"), right?

src/ng/directive/input.js
```diff
@@ -1647,7 +1647,8 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
       if (maxVal < elVal) {
         element.val(maxVal);
-        elVal = minVal;
+        // IE11 and Chrome don't set the value to the minVal when max < min
+        elVal = maxVal < minVal ? minVal : maxVal;
```
gkalpak: Shouldn't this be reflected to `element.val()` then? I would expect something like:

```js
if (maxVal < elVal) {
  elVal = Math.max(minVal, maxVal);
  element.val(elVal);
}
```

src/ng/directive/input.js
```diff
@@ -1587,36 +1632,19 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+    maxVal = isNumber(val) && !isNaN(val) ? val : 100;
     // ignore changes before model is initialized
     if (isNumber(ctrl.$modelValue) && isNaN(ctrl.$modelValue)) {
       return;
     }
 
     if (supportsRange && maxAttrType === 'max') {
       var elVal = element.val();
-      // IE11 doesn't set the el val correctly if the maxVal is less than the element value
+      // IE11 doesn't set the el val correctly if the maxVal is greater than the element value
```
gkalpak: So change seems invalid.

src/ng/directive/input.js
```diff
@@ -1563,11 +1565,54 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+      // ngMax doesn't set the max attr, so the browser doesn't adjust the input value as setting max would
+      function maxValidator(modelValue, viewValue) {
+        return ctrl.$isEmpty(viewValue) || isUndefined(maxVal) || viewValue <= maxVal;
+      };
+
+    setInitialValueAndObserver(maxAttrType, 'max', maxChange);
+  }
+
+  function setInitialValueAndObserver(attrType, attrName, changeFn) {
+   var initialAttrValue;
```
gkalpak: Wrong indentation.

src/ng/directive/input.js
```diff
@@ -1563,11 +1565,54 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+    changeFn(initialAttrValue);
+    attr.$observe(attrName, changeFn);
+  }
+
   function minChange(val) {
     if (isDefined(val) && !isNumber(val)) {
       val = parseFloat(val);
     }
-    minVal = isNumber(val) && !isNaN(val) ? val : undefined;
+    minVal = isNumber(val) && !isNaN(val) ? val : 0;
```
gkalpak: Why this change? Doesn't it break our "convention" that an invalid or undefined value results to no validation?
With this `minVal` can never be undefined.

Or is this trying to emulate the default browser behavior for browsers that don't natively support `input[range]`?

test/ng/directive/inputSpec.js
```diff
@@ -3245,22 +3270,21 @@ describe('input', function() {
...
-        it('should keep the initial default value when min and max are specified', function() {
+        it('should set the correct initial value when min and max are specified', function() {
           scope.max = 80;
           scope.min = 40;
           var inputElm = helper.compileInput('<input type="range" ng-model="value" name="alias" max="{{max}}" min="{{min}}" />');
 
-          expect(inputElm.val()).toBe('50');
-          expect(scope.value).toBe(50);
+          expect(inputElm.val()).toBe('60');
+          expect(scope.value).toBe(60);
```
gkalpak: So, we even had tests that verified the buggy behavior, nice :stuck_out_tongue:

src/ng/directive/input.js
Narretz: Good point. I assume that browsers that don't force valid values should not assume min = 0 and max = 100.

test/ng/directive/inputSpec.js
Narretz: Yeah, I don't know what I was thinking ...

gkalpak: I LGTM'd that, so I don't know what I was thinking either :stuck_out_tongue:

Narretz:
@gkalpak Travis finally green, you may review this again :>

src/ng/directive/input.js
```diff
@@ -1563,6 +1565,49 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+    if (attrType === attrName) {
+      initialAttrValue = attr[attrName];
+      // Set the actual element attribute so that the browser can adjust the value based on
+      // the max value, which might be interpolated
+      element.attr(attrName, initialAttrValue);
+    }
+
+    // This initalizes the min/max value, so that non-support browsers validate with the correct
+    // values during the initial $render
+    changeFn(initialAttrValue);
```
gkalpak: `initialAttrValue` will be undefined for `ngMin`/`ngMax`. Is this intended?

It is not clear why this function would do different things for `ngXyz` vs `xyz` attributes. If this is intended, you could add a short comment explaining what is going on.

test/ng/directive/inputSpec.js
```diff
@@ -3051,7 +3108,7 @@ describe('input', function() {
     describe('ngMin', function() {
 
       it('should validate', function() {
-        var inputElm = helper.compileInput('<input type="range" ng-model="value" name="alias" ng-min="50" />');
+        var inputElm = helper.compileInput('<input type="range" ng-model="value" name="alias" ng-min="20" />');
```
gkalpak: Ooc, does this change make any difference?

test/ng/directive/inputSpec.js
```diff
@@ -3164,6 +3240,34 @@ describe('input', function() {
           expect(scope.form.alias.$error.max).toBeFalsy();
         });
 
+        it('should not assume a max val of 100 if the max attribute interpolates to a non-number', function() {
```
gkalpak: I don't see how this test illustrates this description. (But I am not sure if it is the description or the test or the implementation that needs to change :smiley:)

test/ng/directive/inputSpec.js
Narretz: Nope. I'll revert it.

src/ng/directive/input.js
Narretz: At this point we really only need to set the element attribute if we have min / max and the browser supports range. I've removed the call to changeFn as its use isn't tested currently. 
The actual issue is that all validators that are based on observation if interpolation or ngX attributes only set their values after a digest. That means when the input renders these validators' valuea aren't defined. When the observers fire we will validate again with the correct values but this causes the model to be unset if the value is invalid. Imo this should still count as initialization so that shouldn't happen. The other thing is that we validate too often (once for every observer). 
I'll open an issue for part 1 of this.

src/ng/directive/input.js
```diff
@@ -1593,19 +1589,16 @@ function rangeInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
-      // Set the actual element attribute so that the browser can adjust the value based on
-      // the max value, which might be interpolated
-      element.attr(attrName, initialAttrValue);
+  function setInitialValueAndObserver(actualAttrName, htmlAttrName, changeFn) {
+    // e.g. max === max
+    if (actualAttrName === htmlAttrName) {
+      // interpolated attributes set the attribute value only after a digest, but we need the
+      // attribute value when the input is first rendered, so that the browser can adjust the
+      // input value based on the min/max value
+      element.attr(htmlAttrName, attr[htmlAttrName]);
```
gkalpak: So, if I understand correctly, when using `min`/`max` (with or without interpolation) the browser will set an initial value (if none is set) to `(max + min) / 2`. But if you use `ngMin`/`Max` the browser will ignore them and still assume the default `min`/`max` values (0 and 100 resp.) and set the initial value of the input to 50.
Is that correct?

Additionally, when using `ngMin`/`Max`, the browser will ignore their values and only allow values between the default 0 and 100.
This doesn't sound very useful... :confused:

Narretz: Yes, you are correct. If you put it like this, it indeed doesn't sound very useful. It's all because ngMin/Max et. al. doesn't set the actual attribute on the element. I assume it's done for a reason, but I haven't investigated.

Narretz: What if we simply document that you shouldn't use ngMin/ngMax with input range? Or not document them at all?

gkalpak: Iirc, `ngMin/Max` will set `attrs.min/max`, but not the element's attribute or property. The reason is that we want to be able to have in-Angular validation, but not validation from the browser. This makes sense for other inputs, but not for `input[range]`, because it cannot have invalid values (in supporting browsers at least).

So, in a sense, the browser forces its restrictions on the user (they can't enter a value that is outside of the specified range). Because of that, I agree that `ngMin/Max` don't have anything to offer for `input[range]`.

I would add a little warning in the docs, though, saying that `ngMin/Max` will hav no effect on `input[range]`, because I expect many people will try to use them :smile:

Narretz: Okay, then I'll remove ngMin/Max handling and docs, and add a note to the docs instead.

gkalpak:
LGTM

icorne:
Will this fix be back-ported to older angular versions?

petebacondarwin:
@icorne - which versions did you have in mind? We have no plans to backport it further back than 1.5.x

icorne:
I'm having trouble figuring out where this was merged into. I'm working on 1.5.x, so that's what I'm interested in. I've tried upgrading from 1.5.5 to 1.5.9 but I still get the issue.

petebacondarwin:
Ahah! Right. OK, it didn't make it into 1.5.9, since that was primarily a security release. It will be in 1.5.10.

gkalpak:
This is already included in 1.5.9 (2e7121b), but support for `input[range]` is opt-in on 1.5.x in order to avoid a breaking change (see 07b8761 and [the docs][1]).

[1]: https://code.angularjs.org/1.5.9/docs/api/ng/input/input%5Brange%5D

petebacondarwin:
Oops! I am so confused with all the rebasing we did last week that I forgot that we actually did put this in 1.5.9 :-) Thanks @gkalpak

PR #14998, opened 2016-08-06
WIP - feat($injector): add support for non-string IDs (and other minor stuff)

gkalpak:
Feature.

Only strings can be used as service identifiers. See #10347.

Any value can be used as a service identifier.

No (afaict).

**Other information**:
Among other things, non-string identifiers enable:
- Private services (without relying on naming conventions).
- Collision avoidance (without relying on custom prefixes and namespacing).
- Better toolability (e.g. autocomplete support for service identifiers).
- Better compression (i.e. strings can't be minified, non-string values could).

Identifiers for directives and filters are still restricted to string values, since they need to be referenced in HTML (text).

For services with string IDs, the corresponding provider ID is constructed by appending `Provider`
to the service ID. For services with non-string IDs, the corresponding provider has the exact same
ID (it is the context that determines if a service or a provider should be injected).
<sub>(Thi might be confusing, but I couldn't come up with a better alternative.)</sub>

The main implementation is the 4th commit. The first 3 commits are minor enhancements/fixes/cleaup either necessry or related to the main commit and could be independently backported to v1.5.x (if we wanted to).

Fixes #10347


gkalpak:
Still missing docs (I would like to get some feedback, before spending time on that).

--
Considerations:

1. Using non-string values as identifiers for directives or filter is not supported. Nevertheless, passing non-string values silently fails (or worse, registers things it shouldn't), because of the feature that allows to pass an object contains multiple directives/filters.  
   Although this was the case before as well, people are more likely to run into this if other registration methods (`.provider`, `.service`, `.factory` etc) do support non-string values.  
   <sub>We might want to fix this, independently of this PR though.</sub>

2. Both `$injector.invoke()` and `$injector.instantiate` (and indirectly `$controller`) support passing a `locals` object, where dependencies are retrieved from (if present) instead of getting them through DI. There is currently no way of passing dependencies with non-string identifiers as `locals`, which is inconvenient and "asymmetric".

jbedard:
FYI I created a library that does something similar, while also adding Angular style annotations for declaring modules/services/components. I regret making those annotations the same as Angular (and might rename them someday), but the injection-by-type is quite nice. It does require the reflect-metadata typescript library though.

http://github.com/jbedard/ng-facade/

Issue #14999, opened 2016-08-07
feat($location/$route): support changing location without side effects/route reloading


gkalpak:
There seem to be several issues that requests/discuss/implement the ability to make changes to the current location (URL/history API), without Angular "reacting" to the change. Below is a list of such issues I gathered from a very quick search that seem (more or less) related:
## PRs:
#5860: feat(ngRoute): add method $route.silent(path) for change location path without triggering route controller
## Issues:
#1699: Option on $location to allow hash/path change w/o reloading the route
#3789: Directly modifying the URL path & search without any side effects
#7925: Something along the lines of `reloadOnPathParams` in `ngRoute`
#11075: Changing the location outside of $location causes infdig and/or resets the original url
#12719: Do not force a reload when navigating out of the app base

---

I see two main categories/features:
1. Change the location to something that is either outside of the Angular app (e.g. at a different base) or would map to a different route (assuming client-side routing is used).
2. Change the location to something that would map to the same route (but with different route params).
## 

While I think it would be possible to support both usecases, I haven't seen any (reasonably common) usecase for (1). Furthermore, if we implemented (1), it would make it too easy for people to unknowingly break their client-side routing/deep-linking. Based on that and the fact that it would further complicate what is one of the most complex services, I am not too keen on adding support for (1) - but can be convinced other wise :smiley:

In contrary, having the ability to update the route params (and having the change reflect back to the URL, so deep-linking continues to work as expected), but without having to reload the same route, re-compile the same template, re-instantiate the same controller, re-resolve all `resolve` values etc, sounds pretty reasonable and common.

It is essentially a variation of the `reloadOnSearch` configuration option, but for path segments. I.e. if the path changes, but the new path maps to the same route (with different route params) and that route is configured to not reload on path change, then we just emit a `$routeUpdate` event (the same we do for query param/hash changes on routes configured with `reloadOnSearch: false`) and not load the same route all over.

For routes configured to reload on path changes (probably the default), the behavior will be the same as it is today.
## 

Also note that the Angular 2 router will support a similar feature (although there can't be a 1-to-1 match with `ngRoute`, since the former is hierarchical.

gkalpak:
Usecase (2) addressed by #15002 and we do not intend to support usecase (1) atm.
Closing.

PR #5860, opened 2014-01-17
feat(ngRoute): add method $route.silent(path) for change location path without triggering route controller

kulikov:

Issue #1699, opened 2012-12-12
Option on $location to allow hash/path change w/o reloading the route


cgross:
We have a paradigm in our app that user's create new things on the same page that they view things.  So our route is like /thing/:id.  When creating a new thing they go to /thing/new.  Once the thing has been successfully saved we want to change the route to /thing/1234 (or whatever its new id is).  Now the partial doesnt need to be reloaded because the data is all the same.  We just want the path to be updated so a user can now bookmark the correct link, etc.

Having an option on $location (not on the route definition) to enable/disable route loading would work but I'm sure there are other ways to implement the feature.

pankajkumar:
+1

macinnir:
+1

marcocom:
+1 what about a completely seperate method?  leaving path() as is, and just adding a $location.hash method that simply changes the browser's url without controller reload.  this allows us to control back-navigation of the browser and is a valid use-case scenario in rich applications.

pencilcheck:
+1

Currently using the snippet here in this page: http://joelsaupe.com/programming/angularjs-change-path-without-reloading/

benbeadle:
+1

zuzusik:
+1

briansanchez:
+1

vegetableman:
:+1:

van-nguyen:
A workaround I found was to add a child state with an absolute url.

    $stateProvider.state('state.child', {
        url: '^/some-other-url'
    });

Then simply do $state.go('.child') and that should change the url to the one you want without reloading and keep everyone happy.

AreYouSiries:
+1

manikantag:
+1

tsl0922:
+1

MarkoCen:
+1

swkruger:
+1

MartelBenjamin:
+1

codler:
+1

petebacondarwin:
@btford how is this dealt with (if at all) in the `componentRouter`?

iki:
Thanks @balsarori for [a great solution](https://github.com/angular/angular.js/issues/1699#issuecomment-45048054). Works great!

Elaborated on top of it to allow updating $route path parameters without reloading the controller:
```js
'use strict';

angular.module('params', [
  'ng',
  'ngRoute',
])
.factory('params', function($route, $location) {
  var service = {};

  /**
   * @ngdoc method
   * @name params#replace
   *
   * @description
[...]

PR #15002, opened 2016-08-07
feat($route): add support for the `reloadOnUrl` configuration option

### LINKED ISSUE #7925: Something along the lines of reloadOnPathParams in ngRoute
It would be nice to have an option on a route to only reload the view when path params change, much like `reloadOnSearch` only reloads the view when `$location.search()` changes.

E.g.: If route `/foo/:id` has `reloadOnPathParams = false` set, then moving from `/foo/id1` to `/foo/id2` would only broadcast a `$routeUpdate` event, and not reload the view and re-instanciate the controller.

gkalpak:
Feature.

When the path changes and the new URL maps to the current route, the route is reloaded.
See #14999 (and some of the issues mentioned there). 

The user can determine that specific routes should not get reloaded, when the path changes and the new URL maps to the current route.

No.

**Other information**:
Enables users to specify that a particular route should not be reloaded after a URL change
(including a change in `$location.path()`), if the new URL maps to the same route.
The default behavior is still to always load the matched route when any part of the URL changes.

Related to #1699, #5860, #14999 (potentially closing the first two).

Fixes #7925.

### DIFF
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -196,7 +196,9 @@ function $RouteProvider() {
    *      If the option is set to `false` and the URL in the browser changes, then a `$routeUpdate`
    *      event is broadcasted on the root scope (without reloading the route).
    *
-   *      **Note:** This option has no effect if `reloadOnUrl` is set to false.
+   *      <div class="alert alert-warning">
+   *        **Note:** This option has no effect if `reloadOnUrl` is set to `false`.
+   *      </div>
    *
    *    - `[caseInsensitiveMatch=false]` - `{boolean=}` - match routes without being case sensitive
    *
@@ -556,9 +558,9 @@ function $RouteProvider() {
      * @name $route#$routeUpdate
      * @eventType broadcast on root scope
      * @description
-     * Any of the `reloadOnSearch` and `reloadOnUrl` properties has been set to false and we are
-     * reusing the same instance of the route (including template, controller instance, resolved
-     * dependencies etc).
+     * Broadcasted if the same instance of a route (including template, controller instance,
+     * resolved dependencies, etc.) is being reused. This can happen if either `reloadOnSearch` or
+     * `reloadOnUrl` has been set to `false`.
      *
      * @param {Object} angularEvent Synthetic event object
      * @param {Route} current Current/previous route information.
@@ -666,21 +668,7 @@ function $RouteProvider() {
       var lastRoute = $route.current;
 
       preparedRoute = parseRoute();
-      preparedRouteIsUpdateOnly =
-          // IF this is not a forced reload
-          !forceReload
-          // AND both `lastRoute`/`preparedRoute` are defined
-          && preparedRoute && lastRoute
-          // AND they map to the same Route Definition Object
-          && (preparedRoute.$$route === lastRoute.$$route)
-          // AND `reloadOnUrl` is disabled
-          && (!preparedRoute.reloadOnUrl
-              // OR `reloadOnSearch` is disabled
-              || (!preparedRoute.reloadOnSearch
-                  // AND both routes have the same path params
-                  && angular.equals(preparedRoute.pathParams, lastRoute.pathParams)
-              )
-          );
+      preparedRouteIsUpdateOnly = isNavigationUpdateOnly(preparedRoute, lastRoute);
 
       if (!preparedRouteIsUpdateOnly && (lastRoute || preparedRoute)) {
         if ($rootScope.$broadcast('$routeChangeStart', preparedRoute, lastRoute).defaultPrevented) {
@@ -858,6 +846,29 @@ function $RouteProvider() {
       });
       // No route matched; fallback to "otherwise" route
       return match || routes[null] && inherit(routes[null], {params: {}, pathParams:{}});
+    }
+
+    /**
+     * @param {Object} newRoute - The new route configuration (as returned by `parseRoute()`).
+     * @param {Object} oldRoute - The previous route configuration (as returned by `parseRoute()`).
+     * @returns {boolean} Whether this is an "update-only" navigation, i.e. the URL maps to the same
+     *                    route and it can be reused (based on the config and the type of change).
+     */
+    function isNavigationUpdateOnly(newRoute, oldRoute) {
+      // IF this is not a forced reload
+      return !forceReload
+          // AND both `newRoute`/`oldRoute` are defined
+          && newRoute && oldRoute
+          // AND they map to the same Route Definition Object
+          && (newRoute.$$route === oldRoute.$$route)
+          // AND `reloadOnUrl` is disabled
+          && (!newRoute.reloadOnUrl
+              // OR `reloadOnSearch` is disabled
+              || (!newRoute.reloadOnSearch
+                  // AND both routes have the same path params
+                  && angular.equals(newRoute.pathParams, oldRoute.pathParams)
+              )
+          );
     }
 
     /**
--- a/src/ngRoute/route.js
+++ b/src/ngRoute/route.js
@@ -183,11 +183,20 @@ function $RouteProvider() {
    *      `redirectTo` takes precedence over `resolveRedirectTo`, so specifying both on the same
    *      route definition, will cause the latter to be ignored.
    *
+   *    - `[reloadOnUrl=true]` - `{boolean=}` - reload route when any part of the URL changes
+   *      (inluding the path) even if the new URL maps to the same route.
+   *
+   *      If the option is set to `false` and the URL in the browser changes, but the new URL maps
+   *      to the same route, then a `$routeUpdate` event is broadcasted on the root scope (without
+   *      reloading the route).
+   *
    *    - `[reloadOnSearch=true]` - `{boolean=}` - reload route when only `$location.search()`
    *      or `$location.hash()` changes.
    *
-   *      If the option is set to `false` and url in the browser changes, then
-   *      `$routeUpdate` event is broadcasted on the root scope.
+   *      If the option is set to `false` and the URL in the browser changes, then a `$routeUpdate`
+   *      event is broadcasted on the root scope (without reloading the route).
+   *
+   *      **Note:** This option has no effect if `reloadOnUrl` is set to false.
    *
    *    - `[caseInsensitiveMatch=false]` - `{boolean=}` - match routes without being case sensitive
    *
@@ -202,6 +211,9 @@ function $RouteProvider() {
   this.when = function(path, route) {
     //copy original route object to preserve params inherited from proto chain
     var routeCopy = shallowCopy(route);
+    if (angular.isUndefined(routeCopy.reloadOnUrl)) {
+      routeCopy.reloadOnUrl = true;
+    }
     if (angular.isUndefined(routeCopy.reloadOnSearch)) {
       routeCopy.reloadOnSearch = true;
     }
@@ -544,8 +556,9 @@ function $RouteProvider() {
      * @name $route#$routeUpdate
      * @eventType broadcast on root scope
      * @description
-     * The `reloadOnSearch` property has been set to false, and we are reusing the same
-     * instance of the Controller.
+     * Any of the `reloadOnSearch` and `reloadOnUrl` properties has been set to false and we are
+     * reusing the same instance of the route (including template, controller instance, resolved
+     * dependencies etc).
      *
      * @param {Object} angularEvent Synthetic event object
      * @param {Route} current Current/previous route information.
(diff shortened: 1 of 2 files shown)


gkalpak:
I am not super happy with the API (`reloadOnUrl` is a little cryptic - as is `reloadOnSearch`). I would prefer something like:

```js
reloadOnChange: {
  hash: true/false,
  search: true/false,
  url: true/false
}
```

But at this point I figured it is better to (a) keep consistent with the existing API and (b) not introduce multiple way to do the same thing. But we might want to consider deprecating the existing way and introduce a different API in a future version.

gkalpak:
Btw, in the future we might want to make this more flexible by allowing the user to specify a function (or service) for determining whether to reload or not. This would allow to have a default value for the route, but overwrite it in specific cases.

Mischi:
Both features where the original reason for @davidreher and me to choose the now deprecated ngComponentRouter. We would love to see see these features added to ngRoute!

gkalpak:
Also, considering that Angular 2's latest router re-uses components by default (afaik), it might be a good idea to re-use components by default as well (for 1.6 or 1.7).

Narretz:
So this PR is basically a sub-set of the old problem where you wan't to change the location but not let Angular do anything (like change / reload a route).
With it, you can now "change the location to something that would map to the same route (but with different route params)"
The problem I see is that a static route definition doesn't really make the following use cases simpler (#1699, #3789):

> So our route is like /thing/:id. When creating a new thing they go to /thing/new. Once the thing has been successfully saved we want to change the route to /thing/1234

> Or when someone enters the wrong title for a particular ID. See Stackoverflow, when you change the title but enter just the ID, it will correct the title portion of the URL to be correct.

If the route never reloads on url change, if you do /thing/5678, then it also doesn't change the route, and you'd have to manually do work in the routeUpdate event (e.g. fetching data), instead of relying on resolves.

So I think we either need a $location function that doesn't trigger route change (don't really like that, though) or a dynamic way to decide if a route should be reloaded.

gkalpak:
> The problem I see is that a static route definition doesn't really make the following use cases simpler (#1699, #3789)

This doesn't indeed cover #3789 (in a convenient way), but that sounds better handled with a redirection tbh.

Regarding #1699, I think this PR covers their exact usecase: You are on `/thing/new`, you create and save a new item (so at that point you have all the data you need for `/thing/123`), you change the URL to `/thing/123` and the route doesn't get reloaded (including the resolves). If you need to also re-resolve the data, then you can handle that inside the controller (which by some is considered a better practice anyway).

--
WRT dynamically deciding whether a route should be reloaded or not:

1. Having the controller/component decide is not a good idea imo. This is something that needs to be decided on a route level, so should be in Route Definition Object.

2. Adding support for this by allowing a function or service as value of `reloadOnSearch` sounds like an interesting idea. It is still configurable on a per-route basis, but can be dynamic and the controller can have a saying via some DI'ed service.

Taking (2) one step further, we could introduce a new `reloadStrategy` property that accepts a service that implements the `ReloadStrategy` interface. This could replace both `reloadOnSearch` and `reloadOnUrl`. We could for example have built-in strategies for not reloading when the hash/search/path changes, but users could specify their own.

--
> So I think we either need a $location function that doesn't trigger route change

In general, I believe `$location` should be kept out of this (unless we are talking about a url change that should not update `$location`'s internal `path`/`search`/`hash` properties/state, which would be asking for serious trouble and abuse - but is a different discussion). If we are talking about `ngRoute`-specific behavior/features, it should be entriely included in `ngRoute`.

miwillhite:
Any updates on this?

Narretz:
@miwillhite No, not at the moment.
To get some traction for this, you could test the patch and see if it would work for your use-case, or otherwise give some feedback on the behavior / implementation.

Btw, does ui-router have this functionality?

src/ngRoute/route.js
```diff
@@ -653,9 +666,21 @@ function $RouteProvider() {
       var lastRoute = $route.current;
 
       preparedRoute = parseRoute();
-      preparedRouteIsUpdateOnly = preparedRoute && lastRoute && preparedRoute.$$route === lastRoute.$$route
-          && angular.equals(preparedRoute.pathParams, lastRoute.pathParams)
-          && !preparedRoute.reloadOnSearch && !forceReload;
+      preparedRouteIsUpdateOnly =
```
Narretz: This would be even more readable if it was in a separate fnction

test/ngRoute/routeSpec.js
```diff
@@ -1677,62 +1678,380 @@ describe('$route', function() {
...
+        $location.path('/path/bar');
         $rootScope.$digest();
         expect(routeChange).not.toHaveBeenCalled();
-        expect(routeUpdate).toHaveBeenCalled();
+        expect($routeParams).toEqual({param: 'bar'});
       });
     });
 
 
+    describe('reload', function() {
```
Narretz: Can this describe be a bit more explicit? There's already a top-level describe with this name. How about `with reload()`

test/ngRoute/routeSpec.js
```diff
@@ -1677,62 +1678,380 @@ describe('$route', function() {
   });
 
 
-  describe('reloadOnSearch', function() {
-    it('should reload a route when reloadOnSearch is enabled and .search() changes', function() {
-      var reloaded = jasmine.createSpy('route reload');
+  describe('reloadOnUrl', function() {
+    it('should reload a route when `reloadOnUrl` is enabled and `.url()` changes', function() {
```
Narretz: The wording is inconsistent in the changes - mostly it's `enabled` but further below it's `true` (I prefer the latter).

Narretz (changes requested):
A few smaller things. 
The avanced API outlined in the PR would have been nice, but I assume that most people use uiRouter anyway at this point.

src/ngRoute/route.js
gkalpak: Fixed.

src/ngRoute/route.js
```diff
@@ -544,8 +556,9 @@ function $RouteProvider() {
      * @name $route#$routeUpdate
      * @eventType broadcast on root scope
      * @description
-     * The `reloadOnSearch` property has been set to false, and we are reusing the same
-     * instance of the Controller.
+     * Any of the `reloadOnSearch` and `reloadOnUrl` properties has been set to false and we are
```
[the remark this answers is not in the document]
gkalpak: Rephrased.

src/ngRoute/route.js
```diff
@@ -183,11 +183,20 @@ function $RouteProvider() {
...
+   *
    *    - `[reloadOnSearch=true]` - `{boolean=}` - reload route when only `$location.search()`
    *      or `$location.hash()` changes.
    *
-   *      If the option is set to `false` and url in the browser changes, then
-   *      `$routeUpdate` event is broadcasted on the root scope.
+   *      If the option is set to `false` and the URL in the browser changes, then a `$routeUpdate`
+   *      event is broadcasted on the root scope (without reloading the route).
+   *
+   *      **Note:** This option has no effect if `reloadOnUrl` is set to false.
```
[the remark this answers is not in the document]
gkalpak: Wrapped.

test/ngRoute/routeSpec.js
gkalpak: Made consistent.

test/ngRoute/routeSpec.js
gkalpak: I don't see a top-level `describe` with the same name 😕.

But I agree the descriptions are not great (I copied them from the existing `reloadOnSearch` tests 😳).
I've updated both to ``with `$route.reload()` `` (since they are testing how the configs work in conjuction with `$route.reload()`).

gkalpak:
Thx, @Narretz!
Comments adressed. PTAL

test/ngRoute/routeSpec.js
Narretz: Yeah, I didn't see the reload() block was in reloadOnSearch

Narretz:
I think you forgot to push the changes @gkalpak 👀

gkalpak:
Ooops 😳 Definitely pushed now :smiley:

Narretz:
LGTM - maybe @petebacondarwin should take another look?

Narretz (approved):

test/ngRoute/routeSpec.js
```diff
@@ -65,8 +65,8 @@ describe('$route', function() {
       $httpBackend.when('GET', 'Chapter.html').respond('chapter');
       $httpBackend.when('GET', 'test.html').respond('test');
       $httpBackend.when('GET', 'foo.html').respond('foo');
-      $httpBackend.when('GET', 'baz.html').respond('baz');
       $httpBackend.when('GET', 'bar.html').respond('bar');
+      $httpBackend.when('GET', 'baz.html').respond('baz');
```
petebacondarwin: Does this change have any effect?

test/ngRoute/routeSpec.js
```diff
@@ -1677,95 +1678,413 @@ describe('$route', function() {
...
+
+        module(function($routeProvider) {
+          $routeProvider.
+            when('/path/:param', {reloadOnUrl: false}).
+            otherwise({});
+        });
+
+        inject(function($location, $rootScope, $routeParams) {
+          $rootScope.$on('$routeChangeStart', routeChange);
+          $rootScope.$on('$routeChangeSuccess', routeChange);
```
petebacondarwin: NIT: if we care to test that both these are being triggered then they should have different spies. Otherwise we don't know that only one of the events is being triggered twice rather than each being triggered only once.

test/ngRoute/routeSpec.js
```diff
@@ -1677,95 +1678,413 @@ describe('$route', function() {
...
+        $log.reset();
+
+        $route.reload();
+        $rootScope.$digest();
+        expect($location.path()).toBe('/path/foo');
+        expect(routeChangeStart).toHaveBeenCalledOnce();
+        expect(routeChangeSuccess).toHaveBeenCalledOnce();
+        expect($log.debug.logs).toEqual([['initialized']]);
+
+        $log.reset();
```
petebacondarwin: Not needed

test/ngRoute/routeSpec.js
```diff
@@ -1677,95 +1678,413 @@ describe('$route', function() {
...
+          expect($routeParams).toEqual({param: 'bar'});
+
+          $route.reload();
+          $rootScope.$digest();
+          expect(routeChangeStart).toHaveBeenCalledOnce();
+          expect(routeChangeSuccess).toHaveBeenCalledOnce();
+          expect($log.debug.logs).toEqual([['initialized']]);
+          expect($routeParams).toEqual({param: 'bar'});
+
+          $log.reset();
```
petebacondarwin: Not needed

test/ngRoute/routeSpec.js
```diff
@@ -1677,95 +1678,413 @@ describe('$route', function() {
...
+          $location.hash('baz');
+          $rootScope.$digest();
+          expect(routeChange).not.toHaveBeenCalled();
+          expect(routeUpdate).toHaveBeenCalled();
+        });
+      }
+    );
+
+
+    it('should reload when `reloadOnSearch` is true and url differs only in route path param',
```
petebacondarwin: 'should reload when `reloadOnSearch` is ~~true~~ false and ...' ?

petebacondarwin (approved):
Few minor nits but OLGTM

test/ngRoute/routeSpec.js
gkalpak: No afaict :grin:
<sub>(Except for ordering definitions in proper foo-bar-baz order, which is as great an effect as a change can have 😇)</sub>

test/ngRoute/routeSpec.js
gkalpak: We have other tests to verify that these are called correctly on route change.
Here we mainly want to verify that the `routeChangeStart/Success` combo is emitted and `routeUpdate` isn't.

gkalpak:
Fixed up and rebased.

test/ngRoute/routeSpec.js
gkalpak: Actually, these calls are needed, because there is an `afterEach()` call that asserts the log is empty:

https://github.com/angular/angular.js/blob/84d80be2b4624b9fcfac81204ff047eba108d804/test/helpers/testabilityPatch.js#L50

petebacondarwin: Weird because I thought there are other tests that do not call `$log.reset()` at the end but do cause logging...

By calling reset we are preventing the `afterEach` from checking that other logs were not posted. In any case this is a bit counter-intuitive. Perhaps we should have a helper that checks the log and then removes that item?

PR #15015, opened 2016-08-11
feat($controller): throw error when requested controller is not regis…

Narretz:
docs

when a requested controller is not found (not registered), then a pretty non-descriptive ng:areq error is thrown

a new error is thrown

**Other information**:

…tered

Previously, it would throw the ng:areq error, which is less
specific and just informs that the requested controller did
not yield a function. Given how commonly controllers are used
in Angular, it makes sense to have a specific error.

The ng:areq error is still thrown when the registered controller
is not a function.

Closes #14980

### DIFF
--- a/docs/content/error/$controller/ctrlreg.ngdoc
+++ b/docs/content/error/$controller/ctrlreg.ngdoc
@@ -1,0 +1,24 @@
+@ngdoc error
+@name $controller:ctrlreg
+@fullName A controller with this name is not registered.
+@description
+
+This error occurs when the {@link ng.$controller `$controller()`} service is called
+with a string that does not match any of the registered controllers. The controller service may have
+been invoked directly, or indirectly through the {@link ng.ngController `ngController`} directive,
+or when inside a {@link angular.Module#component component} / {@link angular.Module#directive directive}
+definition (when using string notation for the controller property).
+
+Sources for this error can be:
+
+1. You have a typo in the {@link ng.ngController `ngController`} directive,
+in a {@link angular.Module#component component} / {@link angular.Module#directive directive}
+definition's controller property, or in the call to {@link ng.$controller `$controller()`}.
+2. You have not registered the controller (neither via {@link angular.Module#controller `Module.controller`}
+nor {@link ng.$controllerProvider#register `$controllerProvider.register()`}.
+3. You have a typo in the *registered* controller name.
+4. You want to use controllers defined on the `window`, but have turned off
+{@link ng.$controllerProvider#allowGlobals `allowGlobals()`}.
+
+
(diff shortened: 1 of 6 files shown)


nikhilknoldus:
Hey @Narretz thanks for including #14980

docs/content/error/$controller/ctrlreg.ngdoc
```diff
@@ -0,0 +1,24 @@
+@ngdoc error
+@name $controller:ctrlreg
+@fullName A controller with this name is not registered.
+@description
+
+This error occurs when the {@link ng.$controller `$controller()`} service is called
+with a string that does not match any of the registered controllers. The controller service may have
+been invoked directly, or indirectly through the {@link ng.ngController `ngController`} directive,
```
gkalpak: I would add "for example through...", because the list is non-exhaustive.
(E.g. a route definitions for `ngRoute` or `ui.router` and possibly `ngMaterial` instantiate controllers as well under the hood.)

docs/content/error/$controller/ctrlreg.ngdoc
```diff
@@ -0,0 +1,24 @@
...
+
+Sources for this error can be:
+
+1. You have a typo in the {@link ng.ngController `ngController`} directive,
+in a {@link angular.Module#component component} / {@link angular.Module#directive directive}
+definition's controller property, or in the call to {@link ng.$controller `$controller()`}.
+2. You have not registered the controller (neither via {@link angular.Module#controller `Module.controller`}
+nor {@link ng.$controllerProvider#register `$controllerProvider.register()`}.
+3. You have a typo in the *registered* controller name.
+4. You want to use controllers defined on the `window`, but have turned off
```
gkalpak: I would mention that this  feature is deprecated/not recommened.
E.g. something like `[...] defined on `window` (a feature which is deprecated and turned off by default), but [...]`.

test/ng/controllerSpec.js
```diff
@@ -161,6 +161,12 @@ describe('$controller', function() {
     }).toThrow();
   }));
 
+  it('should throw ctrlreg when the controller name does not match a registered controller', inject(function($compile, $rootScope) {
```
gkalpak: Unused injections.

gkalpak:
A couple of minor comments.
LGTM otherwise :+1:

docs/content/error/$controller/ctrlreg.ngdoc
Narretz: I think then it's better to remove this point completely.

Narretz:
I've made the changes you noted, and also removed the test for ngController. As the $controller service can be used by many different services, we would have to test all of them, which is tedious duplication.

gkalpak:
LGTM

PR #15011, opened 2016-08-10
chore(eslint): enable `quotes: ["error", "single"]`

mgol:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
Chore

**What is the current behavior? (You can also link to an open issue here)**
Quote type is not validated; most of the codebase uses single quotes but random places use double quotes.

**What is the new behavior (if this is a feature change)?**
String delimiters are enforced to be single quotes everywhere except the e2e tests generated by dgeni.

**Does this PR introduce a breaking change?**
No

**Please check if the PR fulfills these requirements**
- [z] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [x] ~~Tests for the changes have been added (for bug fixes / features)~~
- [x] ~~Docs have been added / updated (for bug fixes / features)~~

**Other information**:
Most of this PR has been auto-generated via `eslint . --fix`; examples had to be fixed manually, though as we can't autofix JSDoc.
## 

The quotes rule had to be disabled for e2e tests generated from ngdoc
because dgeni templates use double quotes as string delimiters.

Since we can't have guarantees that dgeni template wrappers will follow
the same JS code style the Angular 1 repo uses, we should find a way
to enforce our ESLint setup only for the parts in this repo, perhaps
via prepending a generated `/* eslint-enable OUR_RULES */` pragma.

PR #15016, opened 2016-08-11
chore(build): fix version placeholder matching

gkalpak:
Bug fix.

``` js
angular.version === {
  ...
  major: 'NG_VERSION_MAJOR',
  minor: 'NG_VERSION_MINOR',
  dot: 'NG_VERSION_DOT',
  ...
}
```

``` js
angular.version === {
  ...
  major: 1,
  minor: 5,
  dot: 9,
  ...
}
```

No.

**Other information**:
During the `build` task, the version placeholders will be replaced with the actual values using a
RegExp, which expects the placeholders to be surrounded by double quotes. By replacing the quotes
from double to single in #15011, the RegExp was not able to match the placeholders.

Btw, this has broken `ngMaterial` (since they polyfill `$animateCss` based on the value of `angular.version.minor`).

### DIFF
--- a/lib/grunt/utils.js
+++ b/lib/grunt/utils.js
@@ -122,12 +122,12 @@ module.exports = {
 
   process: function(src, NG_VERSION, strict) {
     var processed = src
-      .replace(/"NG_VERSION_FULL"/g, NG_VERSION.full)
-      .replace(/"NG_VERSION_MAJOR"/, NG_VERSION.major)
-      .replace(/"NG_VERSION_MINOR"/, NG_VERSION.minor)
-      .replace(/"NG_VERSION_DOT"/, NG_VERSION.patch)
-      .replace(/"NG_VERSION_CDN"/, NG_VERSION.cdn)
-      .replace(/"NG_VERSION_CODENAME"/, NG_VERSION.codeName);
+      .replace(/(['"])NG_VERSION_FULL\1/g, NG_VERSION.full)
+      .replace(/(['"])NG_VERSION_MAJOR\1/, NG_VERSION.major)
+      .replace(/(['"])NG_VERSION_MINOR\1/, NG_VERSION.minor)
+      .replace(/(['"])NG_VERSION_DOT\1/, NG_VERSION.patch)
+      .replace(/(['"])NG_VERSION_CDN\1/, NG_VERSION.cdn)
+      .replace(/(['"])NG_VERSION_CODENAME\1/, NG_VERSION.codeName);
     if (strict !== false) processed = this.singleStrict(processed, '\n\n', true);
     return processed;
   },
--- a/test/e2e/tests/version.spec.js
+++ b/test/e2e/tests/version.spec.js
@@ -1,0 +1,51 @@
+'use strict';
+
+describe('angular.version', function() {
+  var version;
+
+  beforeEach(function() {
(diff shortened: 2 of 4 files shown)


gkalpak:
For reference, here are the RegExps that match and replace the version placeholders: [lib/grunt/utils.js#L125-L130][1]

[1]: https://github.com/angular/angular.js/blob/859348c7f61ff5f93b9f81eb7f46842bd018d8e3/lib/grunt/utils.js#L125-L130

test/e2e/tests/version.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+  it('should contain property `codeName` (string)', function() {
+    expect(version.then(get('codeName'))).toEqual(jasmine.any(String));
+  });
+
+  it('should have `full` === `"major.minor.dot"`', function() {
+    expect(version.then(validate)).toBe(true);
+
+    function validate(ver) {
+      return ver.full.indexOf([ver.major, ver.minor, ver.dot].join('.')) === 0;
```
Narretz: Why not just `ver.full === [ver.major, ver.minor, ver.dot].join('.')`?

gkalpak: That is not always true. E.g. `full` can be `1.5.0-rc.2` or `1.5.9-build.4949` or `1.5.9-local+sha.859348c` (this last one is what we have in the tests actually).

Narretz: Makes sense. Can you add a comment with that info?

mgol:
How about we don't change what you changed here and keep our style but update the regexes so that they accept both string delimiter types? E.g. we'd change:
```js
.replace(/"NG_VERSION_FULL"/g, NG_VERSION.full)
```
to:
```js
.replace(/(?:'|")NG_VERSION_FULL(?:'|")/g, NG_VERSION.full)
```
etc.?

lib/grunt/utils.js
```diff
@@ -122,12 +122,12 @@ module.exports = {
...
-      .replace(/"NG_VERSION_MAJOR"/, NG_VERSION.major)
-      .replace(/"NG_VERSION_MINOR"/, NG_VERSION.minor)
-      .replace(/"NG_VERSION_DOT"/, NG_VERSION.patch)
-      .replace(/"NG_VERSION_CDN"/, NG_VERSION.cdn)
-      .replace(/"NG_VERSION_CODENAME"/, NG_VERSION.codeName);
+      .replace(/(['"])NG_VERSION_FULL\1/g, NG_VERSION.full)
+      .replace(/(['"])NG_VERSION_MAJOR\1/, NG_VERSION.major)
+      .replace(/(['"])NG_VERSION_MINOR\1/, NG_VERSION.minor)
+      .replace(/(['"])NG_VERSION_DOT\1/, NG_VERSION.patch)
+      .replace(/(['"])NG_VERSION_CDN\1/, NG_VERSION.cdn)
```
gkalpak: I could find this anywhere inside the codebase. It is probably not needed, but left it just in case.

gkalpak:
I changed the RegExps to accept both types of quotes. @mgol, PTAL.
(I didn't account for escaped quotes, since we don't need them anywhere if we can use `'` and `"` interchangeably.)

mgol:
LGTM

PR #15021, opened 2016-08-12
fix($sniffer): don't use history.pushState for sandboxed Chrome apps

Parent5446:

Bug fix

When using Angular in a [sandboxed Chrome app](https://developer.chrome.com/apps/manifest/sandbox), errors occur when attempting to call `history.pushState()`, which is not available.

Similar to #11932 and building on the code in #13945, check if inside a sandboxed app, and substitute `history.pushState()` with a no-op.

No

**Other information**:

Check in $sniffer if operating in a sandboxed Chrome app, which does not
have access to chrome.app.runtime like other apps, but also does not
have access to history.pushState. If inside a sandboxed Chrome app, do
not try and use history.pushState, else an error will occur. See #11932
and #13945 for previous work.

### DIFF
--- a/src/ng/sniffer.js
+++ b/src/ng/sniffer.js
@@ -21,8 +21,13 @@ function $SnifferProvider() {
   this.$get = ['$window', '$document', function($window, $document) {
     var eventSupport = {},
         // Chrome Packaged Apps are not allowed to access `history.pushState`. They can be detected by
-        // the presence of `chrome.app.runtime` (see https://developer.chrome.com/apps/api_index)
-        isChromePackagedApp = $window.chrome && $window.chrome.app && $window.chrome.app.runtime,
+        // the presence of `chrome.app.runtime` (see https://developer.chrome.com/apps/api_index).
+        // For sandboxed apps, check for an extension runtime ID, but no access to other Chrome
+        // runtime APIs. See https://developer.chrome.com/apps/manifest/sandbox
+        isChromePackagedApp =
+            $window.chrome &&
+            ($window.chrome.app && $window.chrome.app.runtime ||
+                !$window.chrome.app && $window.chrome.runtime && $window.chrome.runtime.id),
         hasHistoryPushState = !isChromePackagedApp && $window.history && $window.history.pushState,
         android =
           toInt((/android (\d+)/.exec(lowercase(($window.navigator || {}).userAgent)) || [])[1]),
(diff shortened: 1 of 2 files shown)


Parent5446:
:neutral_face: build says Karma tests failed, but I don't see any failed tests.

src/ng/sniffer.js
```diff
@@ -21,8 +21,13 @@ function $SnifferProvider() {
...
         // Chrome Packaged Apps are not allowed to access `history.pushState`. They can be detected by
-        // the presence of `chrome.app.runtime` (see https://developer.chrome.com/apps/api_index)
-        isChromePackagedApp = $window.chrome && $window.chrome.app && $window.chrome.app.runtime,
+        // the presence of `chrome.app.runtime` (see https://developer.chrome.com/apps/api_index).
+        // For sandboxed apps, check for an extension runtime ID, but no access to other Chrome
+        // runtime APIs. See https://developer.chrome.com/apps/manifest/sandbox
+        isChromePackagedApp =
+            $window.chrome &&
+            ($window.chrome.app && $window.chrome.app.runtime ||
+                !$window.chrome.app && $window.chrome.runtime && $window.chrome.runtime.id),
```
gkalpak: I don't think this check can distinguish sandboxed apps from extensions. We need a check that can reliably detect sandboxed apps _only_.

Parent5446: It should be able to. Extensions should always have a $window.chrome.app object. It may not have much in the object (there's no runtime child like with apps), but it will be there nonetheless, whereas with sandbox apps the object does not exist at all, because the APIs are not provided.

I'll try and investigate a better way of detecting this, but this seems to be the only way I can find that does not use APIs that would require an extension to have non-generic permissions for that API.

gkalpak: ah, I didn't pay much attention to the `!$window.chrome.app` check. It should be good then.
I wish there were a more straight-forward way to detect chrome packaged apps :disappointed:

gkalpak:
I basically LGTM, but it would be awesome if there was a more straight-forward way for detecting Chrome Packaged Apps. I'll let you look into it for a few days if you feel like it :smile: - feel free to ping me when you're done investigating.

src/ng/sniffer.js
Parent5446: Sorry for the delay. I've waited for advice, but it appears there really is no other way of checking at the moment.

PR #15025, opened 2016-08-15
fix($compile): correctly merge consecutive text nodes on IE11

### LINKED ISSUE #14924: Interpolations in CJK and Cyrillic languages broken in IE11
**Bug**
Certain interpolations are being broken into multiple spans. For example in a template we have
\<span>{{ctrl.variableName}}\</span>.

In Chrome, Firefox, Safari, Edge this always becomes \<span>Text associated with variable\</span>

but in IE11 in CJK languages this is becoming \<span>{{\</span>\<span>ctrl.variableName}}\</span> which displays on our page as the variable.

There are several places this bug is displayed in production for our site but I will include steps to reproduce for just one of them.

**To reproduce:**

1) Go to https://console.cloud.google.com in IE11
2) Click the hamburger menu in the top left -> click API Manager
3) Click Credentials in the left side nav (if you're prompted to create a project click "Create")
4) Click "Create credentials" -> "API Key" -> "Server Key" -> "Create"
5) Go back to the Credentials Page and click on the name of your new API Key
6) Click the vertical "..." in the top right -> Preferences -> Language and regional formats
7) Change the language to a CJK language (Japanese for example) and click Save
8) The page will reload. Observe the {{ctrl.key.currentKey}} and other variables displayed in the page. 

Here is a screenshot of the issue: 
angularbug

** Details **
This is a bug specific to IE11 (and earlier IE). I'm using IE version 11.0.10240.16603 on Windows 10 Enterprise. We are running angular 1.5 but this also occurred in 1.4

gkalpak:
Bug fix. (Actually, IE11 bug work-around.)

Under certain circumstances, IE11 will break up text nodes, thus breaking interpolation. In some cases Angular fails to fix this (by merging consecutive text nodes) or compiles the merged nodes twice.
See also #14924.

Angular does a better job merging consecutive text nodes and doesn't compile merged nodes twice.

Yes.

**Other information**:
As explained in #11781 and #14924, IE11 can (under certain circumstances) break up a text node into multiple consecutive ones, breaking interpolated expressions (e.g. `{{'foo'}}` would become `{{` + `'foo'}}`). To work-around this IE11 bug, #11796 introduced extra logic to merge consecutive text nodes (on IE11 only), which relies on the text nodes' having the same `parentNode`.

This approach works fine in the common case, where `compileNodes` is called with a live NodeList object, because removing a text node from its parent will automatically update the latter's `.childNodes` NodeList. It falls short though, when calling `compileNodes` with either a jqLite/jQuery collection or an Array. In fails in two ways:
1. If the text nodes do not have a parent at the moment of compiling, there will be no merging.
   (This happens for example on directives with `transclude: {...}`.)
2. If the text nodes do have a parent, just removing a text node from its parent does **not** remove it from the collection/array, which means that the merged text nodes will still get compiled and linked (and possibly be displayed in the view). E.g. `['{{text1}}', '{{text2}}', '{{text3}}']` will become `['{{text1}}{{text2}}{{text3}}', '{{text2}}', '{{text3}}']`.
## 

This commit works around the above problems by:
1. Merging consecutive text nodes in the provided list, even if they have no parent.
2. When merging a text node, explicitly remove it from the list (unless it is a live, auto-updating list).

This can nonetheless have undesirable (albeit rare) side effects by overzealously merging text nodes that are not meant to be merged (see the "BREAKING CHANGE" section below).

Fixes #14924

BREAKING CHANGE:

**Note:** Everything described below affects **IE11 only**.

Previously, consecutive text nodes would not get merged if they had no parent. They will now, which might have unexpectd side effects in the following cases:
1. Passing an array or jqLite/jQuery collection of parent-less text nodes to `$compile` directly:
   
   ``` js
   // Assuming:
   var textNodes = [
     document.createTextNode('{{'),
     document.createTextNode('"foo"'),
     document.createTextNode('}}')
   ];
   var compiledNodes = $compile(textNodes)($rootScope);
   
   // Before:
   console.log(compiledNodes.length);   // 3
   console.log(compiledNodes.text());   // {{'foo'}}
   
   // After:
   console.log(compiledNodes.length);   // 1
   console.log(compiledNodes.text());   // foo
   
   // To get the old behavior, compile each node separately:
   var textNodes = [
     document.createTextNode('{{'),
     document.createTextNode('"foo"'),
     document.createTextNode('}}')
   ];
   var compiledNodes = angular.element(textNodes.map(function (node) {
     return $compile(node)($rootScope)[0];
   }));
   ```
2. Using multi-slot transclusion with non-consecutive, default-content text nodes (that form interpolated expressions when merged):
   
   ``` js
   // Assuming the following component:
   .compoent('someThing', {
    template: '<ng-transclude><!-- Default content goes here --></ng-transclude>'
    transclude: {
      ignored: 'veryImportantContent'
    }
   })
   ```
   
   ``` html
   <!-- And assuming the following view: -->
   <some-thing>
    {{
    <very-important-content>Nooot</very-important-content>
    'foo'}}
   </some-thing>
   
   <!-- Before: -->
   <some-thing>
    <ng-transclude>
      {{       <-- Two separate
      'foo'}}  <-- text nodes
    </ng-transclude>
   </some-thing>
   
   <!-- After: -->
   <some-thing>
    <ng-transclude>
      foo  <-- The text nodes were merged into `{{'foo'}}`, which was then interpolated
    </ng-transclude>
   </some-thing>
   
   <!-- To (visually) get the old behavior, wrap top-level text nodes on -->
   <!-- multi-slot transclusion directives into `<span>` elements; e.g.: -->
   <some-thing>
    <span>{{</span>
    <very-important-content>Nooot</very-important-content>
    <span>'foo'}}</span>
   </some-thing>
   
   <!-- Result: -->
   <some-thing>
    <ng-transclude>
      <span>{{</span>       <-- Two separate
      <span>'foo'}}</span>  <-- nodes
    </ng-transclude>
   </some-thing>
   ```

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1842,12 +1842,40 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     function compileNodes(nodeList, transcludeFn, $rootElement, maxPriority, ignoreDirective,
                             previousCompileContext) {
       var linkFns = [],
-          attrs, directives, nodeLinkFn, childNodes, childLinkFn, linkFnFound, nodeLinkFnFound;
+          // `nodeList` can be either an element's `.childNodes` (live NodeList)
+          // or a jqLite/jQuery collection or an array
+          notLiveList = isArray(nodeList) || (nodeList instanceof jqLite),
+          attrs, directives, node, nodeLinkFn, childNodes, childLinkFn, linkFnFound, nodeLinkFnFound;
+
 
       for (var i = 0; i < nodeList.length; i++) {
         attrs = new Attributes();
+        node = nodeList[i];
 
-        // we must always refer to nodeList[i] since the nodes can be replaced underneath us.
+        // Workaround for #11781 and #14924
+        if ((msie === 11) && (node.nodeType === NODE_TYPE_TEXT)) {
+          var parent = node.parentNode;
+          var sibling;
+
+          while (true) {
+            sibling = parent ? node.nextSibling : nodeList[i + 1];
+            if (!sibling || sibling.nodeType !== NODE_TYPE_TEXT) {
+              break;
+            }
+
+            node.nodeValue = node.nodeValue + sibling.nodeValue;
+
+            if (sibling.parentNode) {
+              sibling.parentNode.removeChild(sibling);
+            }
+            if (notLiveList && sibling === nodeList[i + 1]) {
+              nodeList.splice(i + 1, 1);
+            }
+          }
+        }
+
+        // We must always refer to `nodeList[i]` hereafter,
+        // since the nodes can be replaced underneath us.
         directives = collectDirectives(nodeList[i], [], attrs, i === 0 ? maxPriority : undefined,
                                         ignoreDirective);
 
@@ -2046,13 +2074,6 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
           break;
         case NODE_TYPE_TEXT: /* Text Node */
-          if (msie === 11) {
-            // Workaround for #11781
-            while (node.parentNode && node.nextSibling && node.nextSibling.nodeType === NODE_TYPE_TEXT) {
-              node.nodeValue = node.nodeValue + node.nextSibling.nodeValue;
-              node.parentNode.removeChild(node.nextSibling);
-            }
-          }
           addTextInterpolateDirective(directives, node.nodeValue);
           break;
         case NODE_TYPE_COMMENT: /* Comment */
@@ -2325,9 +2346,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
             var slots = createMap();
 
-            $template = jqLite(jqLiteClone(compileNode)).contents();
-
-            if (isObject(directiveValue)) {
+            if (!isObject(directiveValue)) {
+              $template = jqLite(jqLiteClone(compileNode)).contents();
+            } else {
 
               // We have transclusion slots,
               // collect them up, compile them and store their transclusion functions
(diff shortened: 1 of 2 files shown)


lgalfaso:
I am ok-ish with the PR. Anyhow, please move this fix to a new function that is called at the beginning of `compileNodes`. Is there a chance that `nodeList` is a line `NodeList` without a parent?

gkalpak:
> I am ok-ish with the PR.

Yeah, I am not too happy either :disappointed:

> Anyhow, please move this fix to a new function that is called at the beginning of compileNodes.

Done. Or did you mean traverse the whole `nodeList` before entering the loop?

> Is there a chance that nodeList is a line NodeList without a parent?

Not afaict. The only kind of live `NodeList` that Angular itself is passing is some element's `childNodes`, so by definition, if they lose their parent, they are removed from the live list. I tried other types of live lists (e.g. `document.getElementsByClassName`) and elements are also removed from them if they are removed from their parents.

lgalfaso:
> Or did you mean traverse the whole nodeList before entering the loop?

Yep.

This a side, LGTM for 1.6 and 1.5 (for the last one, confirm with Pete)

gkalpak:
>> Or did you mean traverse the whole nodeList before entering the loop?

> Yep.

My only concern with that is that we are going to miss cases where a compile function adds a text node after the compiled element. It's a very rare usecase (I doubt anyone does that), so I am fine moving that before the loop if you think it's better.

Considering that the merging of consecutive text nodes was not backported to 1.5x (because it is kind of a breaking change), I assume this is for 1.6.x only as well. @petebacondarwin, thoughts?

Narretz:
Let's merge in 1.6. We should get immediate feedback from the google app, right?

endamaco:
I am experiencing exactly the same on Angular 1.6.3 .
My case is with these two spans one after the other and having the second one not binded:
`<span>&#xe080;</span>
<span>{{::model.myVar}}</span>`

Narretz:
@endamaco You should open a new issue with a reproduction.

PR #15036, opened 2016-08-17
docs: services are created using a factory function

akshar-raaj:
Earlier the docs was saying that controllers are created using a factory
function. Instead it should say that services are created using a
factory function.


gkalpak:
This change is invalid. The previous section was indeed talking about `controllers` not `services`.
(Fwiw, I made some minor improvements to the `Conceptual Overview` section in 88f3517.)

PR #15039, opened 2016-08-17
Fix sce context

Narretz:
rebased and slightly tweaked #14019


test/ng/compileSpec.js
```diff
@@ -10152,6 +10151,32 @@ describe('$compile', function() {
       expect(element.attr('src')).toEqual('http://example.com/image2.png');
     }));
 
+    // Older IEs seem to reject the video tag with "Error: Not implemented"
+    // if (!msie || msie > 9) {
```
gkalpak: This seems necessary. Tests are failing on IE9 without it.

src/ng/sce.js
```diff
@@ -537,7 +537,7 @@ function $SceDelegateProvider() {
  * | `$sce.HTML`         | For HTML that's safe to source into the application.  The {@link ng.directive:ngBindHtml ngBindHtml} directive uses this context for bindings. If an unsafe value is encountered and the {@link ngSanitize $sanitize} module is present this will sanitize the value instead of throwing an error. |
  * | `$sce.CSS`          | For CSS that's safe to source into the application.  Currently unused.  Feel free to use it in your own directives. |
  * | `$sce.URL`          | For URLs that are safe to follow as links.  Currently unused (`<a href=` and `<img src=` sanitize their urls and don't constitute an SCE context. |
```
gkalpak: > `<a href=` and `<img src=` sanitize their urls and don't constitute an SCE context

Is the same true for `<audio/video/source src=`?
/cc @rjamet

test/ng/compileSpec.js
Narretz: yes, I was just checking if all of the elemnts fail

src/ng/sce.js
Narretz: I guess that depends on what "sanitize" means. The important part seems to be that in the src attribute, no script execution is possible, that's why we can remove the url context. But Idk if that means sanitization

test/ng/compileSpec.js
```diff
@@ -10152,6 +10151,33 @@ describe('$compile', function() {
       expect(element.attr('src')).toEqual('http://example.com/image2.png');
     }));
 
+    // IE9 rejects the video / audio tag with "Error: Not implemented" and the source tag with
+    // "Unable to get value of the property 'childNodes': object is null or undefined"
```
mgol: Why is that so? IE 9 supports `video` & `audio` elements.

Narretz: But not if we try to do it like we do in the test. It's possible that we need different HTML for it to work: http://stackoverflow.com/questions/7358689/audio-tag-not-working-in-ie9

mgol: The OP answered that adding a DOCTYPE helped. Do our test cases not have a DOCTYPE? If that's the case we should definitely add them, we should make sure all our tests are running in standards mode, not quirks mode that we don't even officially support AFAIK.

All IE versions older than 10 in quirks mode invoke a separate engine simulating IE 5.5 and, as a consequence, missing features from newer versions (including `<video>` and `<audio>`, obviously).

mgol: I've looked at a couple of pages used for e2e tests and they all have a DOCTYPE. The template used by Karma also specifies a DOCTYPE. So this shouldn't be a problem.

Narretz: Maybe it's because of this? https://github.com/videojs/video.js/issues/290#issue-9996576

Narretz:
Can I merge this in the meantime or do you want this resolved first @mgol?

mgol:
@Narretz Go ahead & merge, we can look into it later since it requires further investigation, apparently.

src/ng/sce.js
rjamet: These three won't be sanitized, but as far as we know there's no browser that both supports these tags, and runs javascript: URLs there.

rjamet:
Hey, sorry, I left for a few days right when you started the conversation :/ This PR is still fine. As you spotted, there won't be a $sce context anymore for these three. I think having no sanitization there is fine since no browser will execute script there, but if you want consistency you'd just need to add them where sanitization happens: https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L1631 .

Narretz:
@rjamet Ok, I will then re-add track and merge the PR

PR #15044, opened 2016-08-19
feat($q): Add notify functionality to ES6-style constructor

philip98:
feature

The ES6-style constructor $q() doesn't support the notify() function.

The resolver function passed to the ES6-style constructor $q() now takes three parameters: resolve, reject and notify.

No

**Other information**:

The resolver function passed to $q() may now take a third argument: notify, which wraps the deferred.notify() function.

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -679,7 +679,7 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
     function rejectFn(reason) {
       deferred.reject(reason);
     }
-    
+
     function notifyFn(value) {
       deferred.notify(value);
     }
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -32,7 +32,8 @@
  *
  *   function asyncGreet(name) {
  *     // perform some asynchronous operation, resolve or reject the promise when appropriate.
- *     return $q(function(resolve, reject) {
+ *     return $q(function(resolve, reject, notify) {
+ *       notify('Determining whether it\'s ok to greet.');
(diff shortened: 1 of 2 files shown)


philip98:
I signed it!

Narretz:
Hi,
yes, this would be useful, but we cannot merge it without tests.

gkalpak:
I am not sure we want to support this, since neither the native implementation nor the Promises/A+ spec support `notify`.

Narretz:
Hm, didn't think about that. We do support notify in `$q.deferred()` though.

gkalpak:
I think it's because the [Promises/A proposal][1] specified (optional) progress handlers (aka `notify`), but it was dropped in the [Promises/A+ spec][2]. Or something like that :smiley:

In any case, I think we should not encourage using that feature, because it is not supported in the ES2015 spec and has been deprecated (or is not supported) by other popular Promises/A+ compliant libraries.
(And there were reasons, e.g. as explained [here][3] and [there][4].)

[1]: http://wiki.commonjs.org/wiki/Promises/A
[2]: https://promisesaplus.com/
[3]: http://lists.w3.org/Archives/Public/www-dom/2013JulSep/0116.html
[4]: http://lists.w3.org/Archives/Public/www-dom/2013JulSep/0122.html

philip98:
Ok, didn't know about the drawbacks of progress notification in Promises. It's just that - at my company - we are refactoring an Angular app that has these notify callbacks cluttered all over it, and I was looking for a way to clean up all those `$q.defer()`'s and `return deferred.promise;`'s

PR #15051, opened 2016-08-23
camelCase directive in html referenced as kebab-case, The kebab-case shall be camel-case

rohitbind:

**Other information**:

In line no 71 and 72 `camelCase` is directive name and in HTML the it is said that it should be kebab-case where as it should be 'camel-case'.

I felt that beginners who is trying to learn from tutorial can get confused.

### DIFF
--- a/docs/content/tutorial/step_03.ngdoc
+++ b/docs/content/tutorial/step_03.ngdoc
@@ -69,7 +69,7 @@ To create a component, we use the {@link angular.Module#component .component()} 
 Definition Object (CDO for short).
 
 Remember that (since components are also directives) the name of the component is in `camelCase`,
-but we will use `kebab-case`, when referring to it in our HTML.
+but we will use `camel-case`, when referring to it in our HTML.
 
 In its simplest form, the CDO will just contain a template and a controller. (We can actually omit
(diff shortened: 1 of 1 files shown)


Narretz:
Hm, I see the problem. We should use an example name for the component then, instead of camelCase, kebab-case, because the new example can also be confusing.

rohitbind:
Hi Narretz,
travis failed what could be the reason, I just edited the doc. Is there is any problem from my side?

gkalpak:
Thx for the PR, but this change is invalid. The used case is indeed called kebab-case.

gkalpak:
I think it's best to add links to definitions of camel/kebab case.

rohitbind:
I am not getting if I create a component with name 'camelCase' can I call that component in html by using 'kebab-case'?

Narretz:
The names are not the real component names. It's just to illustrate that when defining the component, you do it like this: myCustomComponent. And when using it in the template do: my-custom-component

gkalpak:
cb31067 hopefully improves the explanation of casing.

rohitbind:
Thanks for accepting the suggestion :+1: :)

PR #15055, opened 2016-08-24
feat($compile):make the required controllers available during the instantiation of the controller

kuitos:
feature

the required controllers will not be available during the instantiation of the controller when `bindToContrller` is truthy, but they are guaranteed to be available just before the $onInit method is executed

the required controllers is available during the instantiation of the controller now when `bindToContrller` is truthy

No

**Other information**:

---

This change is [Reviewable](https://reviewable.io/reviews/angular/angular.js/15055)

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -376,11 +376,6 @@ between each other. This can be achieved in a component by providing an
 object mapping for the `require` property. The object keys specify the property names under which
 the required controllers (object values) will be bound to the requiring component's controller.
 
-<div class="alert alert-warning">
-Note that the required controllers will not be available during the instantiation of the controller,
-but they are guaranteed to be available just before the `$onInit` method is executed!
-</div>
-
 Here is a tab pane example built from components:
 
 <example module="docsTabsExample" name="component-tabs-pane">
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -423,8 +423,7 @@
  * controllers.
  *
  * If the `require` property is an object and `bindToController` is truthy, then the required controllers are
- * bound to the controller using the keys of the `require` property. This binding occurs after all the controllers
- * have been constructed but before `$onInit` is called.
+ * bound to the controller using the keys of the `require` property.
  * If the name of the required controller is the same as the local name (the key), the name can be
  * omitted. For example, `{parentDir: '^^'}` is equivalent to `{parentDir: '^^parentDir'}`.
  * See the {@link $compileProvider#component} helper for an example of how this can be used.
@@ -2625,11 +2624,18 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
         }
 
+        // get the sorted controller directive names to ensure the controllers construction order
+        var sortedControllerDirectiveNames = getSortedControllerDirectiveNames();
+
         // Initialize bindToController bindings
-        for (var name in elementControllers) {
+        forEach(sortedControllerDirectiveNames, function(name) {
           var controllerDirective = controllerDirectives[name];
           var controller = elementControllers[name];
           var bindings = controllerDirective.$$bindings.bindToController;
+
+          // bind required controllers before directive bindings initialize
+          // thus controller can receive required controllers during constructing
+          bindRequiredControllers(name, controllerDirective);
 
           if (controller.identifier && bindings) {
             controller.bindingInfo =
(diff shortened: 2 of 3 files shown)


gkalpak:
The current behavior is intended, as it is the only way to support all usecases. The proposed approach fails with "symbiotic" directives (i.e. dirA requiring dirB and dirB requiring dirA). Additionally, the priority-based order of controller instantiation is violated (making the behavior less straight-forward).

kuitos:
> The current behavior is intended, as it is the only way to support all usecases. 

uh... which usecase that the new behavior not support? I don't have any ideas can u show some code?

> The proposed approach fails with "symbiotic" directives (i.e. dirA requiring dirB and dirB requiring dirA). 

u are right the new behavior will cause the call stack overflow exception with "symbiotic" directives,but it isn't a fatal problem,I can set a counter to detect whether u have defined a "symbiotic" directive.

gkalpak:
>> The current behavior is intended, as it is the only way to support all usecases.

> uh... which usecase that the new behavior not support? I don't have any ideas can u show some code?

Symbiotic directives :smile:

--
>> The proposed approach fails with "symbiotic" directives (i.e. dirA requiring dirB and dirB requiring dirA).

> u are right the new behavior will cause the call stack overflow exception with "symbiotic" directives,but it isn't a fatal problem,I can set a counter to detect whether u have defined a "symbiotic" directive.

We can do a lot of (complicated) things. But what is the point? We will complicate both the code and the behavior, but I don't see any gain that would justify that.

Is there any usecase that isn't supported with current implementation?

kuitos:
@gkalpak 
yes the new behavior is complicated but it is a real demand, sorry about my late code, here is my usecase:

I use es6 class to define my contrller constructor of directive:
```js
class Contrller() {

  get ngModel() {
    return this._ngModel;
  }

  set ngModel(value) {
    this._ngModel = value;
    this.ngModelCtrl.$setViewValue(value);
  }
}
``` 

and the directive ddo:
```js
import Controller from './Controller';
const ddo = {
  require: {
    ngModelCtrl: 'ngModel'
  },
  controller: Controller
  bindings: {
    ngModel: '<'
  }
};
```

I use the class getter/setter feature of es6 to detect changes which have the same effect with `$scope.$watch`.

But the problem is, the `initializeDirectiveBindings` method will initialize the `ngModel` property and trigger the setter of ngModel, then exception come out.

dcherman:
I'd recommend against what you're doing there anyway imo.  `NgModel` already has utilities to track state internally that you should be using in `$render` and `$viewChangeListeners`.  If you truly need a watch on the model value (and not the view value), then add a `$watch` to `ngModelCtrl.$modelValue`.

petebacondarwin:
I'm with @dcherman here. I don't really understand what you are trying to achieve by intercepting the ngModel value being changed.

kuitos:
Alright, maybe the example cause some confusion, the `ngModel` just represent a binding prop from component in my original intention. What if I modify the example like this:

```js
import Controller from './Controller';
const ddo = {
  require: {
    anotherComponentCtrl: 'anotherComponent'
  },
  controller: Controller
  bindings: {
    state: '<'
  }
};
```

```js
class Controller() {

  get state() {
    return this._state;
  }

  set state(value) {
    this._state = value;
    this.anotherComponentCtrl.setState(value);
  }
}
```

It is a very common requirement between linked components.
I want to keep my controller purely as a pure ViewModel that why I Did not inject `$scope` and use `$watch` to detect changes.

petebacondarwin:
This is the reason for the `$onChanges` hook:

```
class Controller() {
  $onChanges(changes) {
    if (changes.state) {
      this.anotherComponentCtrl.setState(this.state);
    }
  }
}
```

kuitos:
@petebacondarwin 
I am aware of the `$onChanges` hook for component since angular 1.5.3, but was it designed for the use case above? I don't think so. It could solve the problem but It doesn't designed for them.

Another use case, what if I call the required controller in constructor?

```js
const ddo = {
  controller: Controller,
  require: {
    anotherComponentCtrl: 'anotherComponent'
  }
  bindings: {
    state: '<'
  }
};
```

```js
class Controller {
  constructor() {
    this.anotherComponentCtrl.setState(this.state);
  }
}
```

Yes we have the `$onInit` to take the place of class constructor, then the problem gone. 

Actually what made me confused was, we could not receive the required controller but the state of bindings is available at the same time. Is the inconsistent behavior reasonable? 

In my opinion, we should keep the bevaviors consistent between prop bindings and required controllers, both are available or not.

petebacondarwin:
The idea is not to use the constructor for any real work. Use `$onInit` to do such things, since this is the only point where you know everything is wired up and ready to go. Moreover in a future version of Angular 1, we intend to stop "pre-binding" properties before the constructor is called, which makes using the constructor even less useful.

kuitos:
I agree with your words.
The "pre-binding" props before constructor behaviour should be abandon after `$onInit` was born I think, it makes the class constructor with a bad semantics which should not contains a context in this.
I think you need to tag the issue before I close it.😀

petebacondarwin:
Thanks @kuitos - I will close this now.

PR #15057, opened 2016-08-24
feat($compile): throw error when directive name or factory fn is invalid

### LINKED ISSUE #15056: Syntax error in directive does not throw exception
I am reporting a bug.

If a directive is missing a comma in a particular position, no exception is thrown into the console and the directive does not work.  Here is an example, where the missing comma is at the end of line 8 of the JavaScript section:
Working directive: https://jsfiddle.net/derfsubterfuge/y9xd3mdb/
Broken directive: https://jsfiddle.net/derfsubterfuge/2nfy6b4c/1/

The example is using Angular 1.5.8.  I have attempted this on Fedora 24 for both Chrome 52.0.2743.116 (64-bit) and Firefox 48.0 (64-bit).

I would expect a syntax error to be thrown in the console and not swallowed.

Narretz:
feat

when the directive name or factory fn is undefined, no error is thrown

better errors are thrown

No. Although I'm not sure if isFunction will catch all exotic ways to define a function.

**Other information**:

Closes: #15056

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1087,6 +1087,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    * @returns {ng.$compileProvider} Self for chaining.
    */
   this.directive = function registerDirective(name, directiveFactory) {
+    assertArg(name, 'name');
     assertNotHasOwnProperty(name, 'directive');
     if (isString(name)) {
       assertValidDirectiveName(name);
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -214,6 +214,7 @@ describe('$compile', function() {
       });
       inject(function($compile) {});
     });
+
     it('should throw an exception if a directive name has leading or trailing whitespace', function() {
       module(function() {
         function assertLeadingOrTrailingWhitespaceInDirectiveName(name) {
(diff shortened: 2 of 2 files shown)


gkalpak:
If we do this for directives, why not do the same for controllers, filters, services etc? I feel we are going to far trying to be helpful here.

Narretz:
Is there a thing as too helpful? :D
The difference between directives and controllers, filters, services is that the latter will complain when they are actually used. Directives are just attributes in DOM elements, so if there's no match than just nothing happens.
I think it's a change we can make - it's very little code.

gkalpak:
> Is there a thing as too helpful? :D

Absolutely :stuck_out_tongue:

> The difference between directives and controllers, filters, services is that if the latter haven't been registered, they will complain when they are actually used.

Point taken :smile:
I don't find this necessary, but since it's such a small change I am OK merging this (as long as this not a considered a breaking change).

Narretz:
I don't think it's a breaking change, as I can't see a use case for adding an "empty" directive. And functions should be detected regardless of which shape they are. I guess I was a bit worried because we had that issue with the constructor fns in compile, but this is something different.

Narretz:
I've relaxed the check for the factory, as we can't guarantee 100% that our isFunction detects the most exotic ways of defining a function.

PR #15064, opened 2016-08-29
refactor($q): separate Deferred out of Promise implementation

jbedard:
refactor / cleanup, maybe performance gains when creating promises

This makes `Deferred` a simple wrapper of `Promise`. `Promise`s can now be constructed, chained etc without creating any `Deferred` objects.

Best viewed [ignoring whitespace](https://github.com/angular/angular.js/pull/15064/files/?w=1)

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -299,13 +299,17 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
    * @returns {Deferred} Returns a new instance of deferred.
    */
   function defer() {
-    var d = new Deferred();
-    //Necessary to support unbound execution :/
-    d.resolve = simpleBind(d, d.resolve);
-    d.reject = simpleBind(d, d.reject);
-    d.notify = simpleBind(d, d.notify);
-    return d;
+    return new Deferred();
   }
+
+  function Deferred() {
+    var promise = this.promise = new Promise();
+    //Non prototype methods necessary to support unbound execution :/
+    this.resolve = function(val) { resolvePromise(promise, val); };
+    this.reject = function(reason) { rejectPromise(promise, reason); };
+    this.notify = function(progress) { notifyPromise(promise, progress); };
+  }
+
 
   function Promise() {
     this.$$state = { status: 0 };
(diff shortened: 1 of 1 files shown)


petebacondarwin:
I was thinking of something similar to this myself recently!

src/ng/q.js
```diff
@@ -316,13 +320,13 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
...
 
       this.$$state.pending = this.$$state.pending || [];
       this.$$state.pending.push([result, onFulfilled, onRejected, progressBack]);
       if (this.$$state.status > 0) scheduleProcessQueue(this.$$state);
 
-      return result.promise;
+      return result;
     },
 
     'catch': function(callback) {
```
aciccarello: This was here before but why are `catch` and `finally` defined with strings and `then` is not?

jbedard: They are javascript keywords

aciccarello: I should have realized that with catch. Didn't realize finally was also a keyword.

src/ng/q.js
```diff
@@ -299,14 +299,18 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
    * @returns {Deferred} Returns a new instance of deferred.
    */
   var defer = function() {
-    var d = new Deferred();
-    //Necessary to support unbound execution :/
-    d.resolve = simpleBind(d, d.resolve);
```
lgalfaso: I think this is a small breaking change. Before, it was possible to intercept all calls to `Deferred.prototype.resolve` by changing the prototype of `Deferred` (that you can do by getting the prototype from an instance).

I think this is a small edge case, but should be clarified and the commit should make this clear.

jbedard:
I've rebased this.

gkalpak:
So, why are spared creating a `Deferred` object when constructing/chaining a `Promise`, which is not a bad thing. But we do lose the (indirect) ability to hook into `Deferred.prototype` (as @lgalfaso has mentioned elsewhere). So this is a potential breaking change (in practical terms).

I wonder to what extend was this hidden feature utilized. Maybe @lgalfaso has a better idea, since he said this has been explicitly discussed in the past.

jbedard:
I can revert that part if really wanted, just seemed like something we can simplify along with everything else in this PR.

Note that what @lgalfaso is referring to is doing `Object.getPrototypeOf($q.defer()).resolve = ...`. I don't think this is really a "supported" use case...

lgalfaso:
Just to be sure about my position. I think that this change is good and that it simplifies some of the existing code at `$q`. That said, I would only merge it in 1.6. The main reason is not only that this is a breaking change, but that there is no work around.

gkalpak:
(Adding it to the `1.6.0-rc.1` milestone, in order to consider it.)

petebacondarwin:
LGTM

jbedard:
Rebased again.

petebacondarwin:
Great. Can you add some information about the BC to the commit message and then we can merge.

jbedard:
Adding a BC message.

gkalpak (approved):

PR #15071, opened 2016-08-30
chore(npm): use require.resolve when possible to avoid hard coded mod…

Narretz:
Cleaned up and rebased version of https://github.com/angular/angular.js/pull/13320.

I've tested the affected commands locally, and they all went through.


lib/grunt/plugins.js
```diff
@@ -39,7 +39,7 @@ module.exports = function(grunt) {
 
 
   grunt.registerTask('docs', 'create angular docs', function() {
-    var gruntProc = shelljs.exec('"node_modules/.bin/gulp" --gulpfile docs/gulpfile.js');
+    var gruntProc = shelljs.exec('node "' + require.resolve('gulp/bin/gulp') + '" --gulpfile docs/gulpfile.js');
```
Narretz: I had to add `node` here, otherwise Windows wouldn't execute the command

mgol: I'd rather not depend on the internal `gulp` package structure. How about creating the npm script `gulp` in package.json:
```json
{
  "scripts": {
    "gulp": "gulp
  }
}
```
and then `npm run gulp -- --gulpfile docs/gulpfile.js` should work everywhere.

lib/grunt/utils.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
       done();
       return;
     }
-    var p = spawn('node', ['node_modules/protractor/bin/webdriver-manager', 'update']);
+    var p = spawn('node', [require.resolve('protractor/bin/webdriver-manager'), 'update']);
```
mgol: `webdriver-manager` is a separate package (and Protractor [just uses it](https://github.com/angular/protractor/blob/79184ff13e9370e0803b70000f67b601934866ec/bin/webdriver-manager)) so we could switch as well.

Narretz: True. So simply `require.resolve('webdriver-manager')`?

mgol: That would only resolve what's exported, not the binary.

Have you tried specifying the binary as `'./node_modules/.bin/webdriver-manager'` instead of `node` and using [cross-spawn](https://www.npmjs.com/package/cross-spawn) for Windows compatibility?

Narretz: I didn't want to complicate the PR actually, but if this is future-proof then let's do it

Narretz:
@mgol I've updated the PR with your suggestions, can you take a look?

mgol:
LGTM.

Issue #15068, opened 2016-08-30
ngRepeat does not trigger move animation


pondermatic:
**Overview of the Issue**
There appears to be a *bug* with ngRepeat and move animations.

**What is the current behavior?**
When an item is inserted or deleted, the move animation is **not** triggered for the items after the inserted or deleted item.

**Steps to reproduce:**
Go to http://plnkr.co/edit/MPpxQUc8OyrRAgDgmDLp?p=preview
Click on an *Insert* button or *Remove* text.

**What is the expected behavior?**
The move animation should be triggered for the items after the inserted or deleted items.

**Versions**
Angular: 1.2.30, 1.3.20, 1.4.12, 1.5.8, snapshot
Browser: Chrome 52
OS: Windows 10

**What is the motivation / use case for changing the behavior?**
I would like to add an animated border or background color change to items that are added to, deleted from or have moved in a list. The [documentation](https://docs.angularjs.org/api/ng/directive/ngRepeat#animations) for ngRepeat's *move* animation says:

> when an adjacent item is filtered out causing a reorder or when the item contents are reordered

**Possible related issue**
#5160

I have implemented a fix for this issue and will be submitting a pull request today.
Here is a plunker with the revised Angular code.
http://plnkr.co/edit/ru77WOctvnEinWdTBMMQ?p=preview

gkalpak:
Possibly related to #14984 as well.

PR #15072, opened 2016-08-30
fix(ngRepeat): trigger move animation

pondermatic:
bug fix

#15068


src/ng/directive/ngRepeat.js
```diff
@@ -340,9 +339,9 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
     scope.$odd = !(scope.$even = (index & 1) === 0);
   };
 
-  var getBlockStart = function(block) {
-    return block.clone[0];
-  };
+  // var getBlockStart = function(block) {
+  //   return block.clone[0];
+  // };
```
austinoneil: Could you please remove this commented-out code?

pondermatic: I had to comment out this unused function to pass the eslint test.

pondermatic:
The Travis build failed because of a socket timeout to saucelabs.com for the Safari tests. How can I start a Travis-CI test again?

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,129 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
             collectionKeys = collection;
             trackByIdFn = trackByIdExpFn || trackByIdArrayFn;
           } else {
             trackByIdFn = trackByIdExpFn || trackByIdObjFn;
             // if object, extract keys, in enumeration order, unsorted
-            collectionKeys = [];
-            for (var itemKey in collection) {
-              if (hasOwnProperty.call(collection, itemKey) && itemKey.charAt(0) !== '$') {
-                collectionKeys.push(itemKey);
+            for (collectionKey in collection) {
```
Narretz: Why this change? Key of the current item makes more sense to me than key of collection.

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,129 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
+                  keyIdentifier, nextBlockMap[nextKey].key, nextLength);
+
+                delete nextBlockMap[nextKey].key;
+                delete nextBlockMap[nextKey].value;
               });
             }
           }
+
+          // leave
+          // This must go after enter and move because leave prevents getting element's parent.
```
Narretz: Can you explain this further? The leave animation section seems to have changed quite a bit from the original. Is this necessary for this PR?

src/ng/directive/ngRepeat.js
pondermatic: You are probably right. I went back and forth on several names and don't remember why I settled on collectionKey. Possibly for consistency in how I named other keys. Let me know if you want me to change it.

Narretz:
Hi @pondermatic, thanks for this PR. This is a much requested feature, and I'd be happy if we can get this in. Since ngRepeat is a performance critical part of the code, we'll have to review it thoroughly, though (and run some benchmarks).
I've looked at the changes, and I have the feeling that some of them aren't really related to the PR. could you maybe comment on these sections that are essential to the PR, or even create a commit that only includes the changes need to fix this?

src/ng/directive/ngRepeat.js
pondermatic: The two main changes are 1) when the leave animations are triggered, and 2) no need to keep track of deleted elements because we aren't doing any more processing where we might need to ignore the element.

pondermatic:
The diff is pretty ugly because I changed several variable names. I'll work on a version that preserves as many names as possible.
I would appreciate performance testing. I timed the unit tests in Windows Chrome between the previous version and this version of ngRepeat and didn't see a noticeable difference. But it's critical that this directive run as fast as possible.

Narretz:
I've made some tests with the orderBy benchmark (after some fixes), rendering 500 ngRepeated spans from scratch, and found that the new version is about 15% slower than the current master. I'm not sure how much truth is in these numbers, but the difference was consistent.

pondermatic:
I've made a version that minimizes the changes, but the diff is still hard to follow. It'll probably be better to explain the key changes.

I'm going to do some profiling and see if I can speed things up.

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,129 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
-                block.clone = clone;
-                nextBlockMap[block.id] = block;
-                updateScope(block.scope, index, valueIdentifier, value, keyIdentifier, key, collectionLength);
+                nextBlockMap[nextKey].clone = clone;
+                updateScope(scope, nextBlockMap[nextKey].index,
+                  valueIdentifier, nextBlockMap[nextKey].value,
+                  keyIdentifier, nextBlockMap[nextKey].key, nextLength);
+
+                delete nextBlockMap[nextKey].key;
+                delete nextBlockMap[nextKey].value;
```
Narretz: This is new, right? Is this necessary? The tests do not fail without this.
 If we are doing this for every item, this might actually impact performance.

pondermatic: The previous version also assigned the clone property and called updateScope(). What's new is deleting the key and value from the map. The previous version did not store key and value in the map.

pondermatic:
I've been running the orderby-bp benchmark, updated by @Narretz today 07849779ba365f371a8caa3b58e23f677cfdc5ad.
- number of samples to collect and average: 20 (default)
- number of ngRepeats: 5000 (default)
- Loop 25x
- Windows Chrome v53
- One incognito window with five tabs, each tab with a different test
- added up mean test time from setup and $apply in all five tests
- each pass opened in new incognito window

c54f7a93e09d165d8e6742da7c8b7df2391f0c91 before pondermatic
pass 1 = 1058.66 ms
pass 2 = 1067.40 ms

263be6bbdfe67a877c50430fcb4acd9045ff8448 after pondermatic
pass 1 = 1110.78 ms
pass 2 = 1127.65 ms

My version is 5.29% slower. :(

When I removed the lines that delete the collection item's key and value from nextBlockMap, pass 1 was faster but pass 2 was much slower. Either way, the last collection is retained in memory which is not good.

Then I put the key/value in its own array so that deleting is unnecessary. I also tried an array of just the keys, which should use less memory than 263be6bbdfe67a877c50430fcb4acd9045ff8448, especially if the values are large objects. The mean of those benchmark passes was 1111.31.

I plan on committing revised code with this memory tweak and another tweak.

src/ng/directive/ngRepeat.js
```diff
@@ -340,15 +339,10 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
     scope.$odd = !(scope.$even = (index & 1) === 0);
   };
 
-  var getBlockStart = function(block) {
-    return block.clone[0];
-  };
-
```
pondermatic: getBlockStart function is no longer used.

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,124 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
+            lastBlockOrder = [],
+            lastKey,
+            nextBlockMap = createMap(),
+            nextBlockOrder = [],
+            nextKey,
+            nextLength,
+            previousNode = $element[0],  // node that cloned nodes should be inserted after
+                                         // initialized to the comment node anchor
+            trackById,
+            trackByIdFn;
```
pondermatic: variables are defined in alphabetical order (mostly)

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,124 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
 
           if (aliasAs) {
             $scope[aliasAs] = collection;
           }
 
-          if (isArrayLike(collection)) {
-            collectionKeys = collection;
+          // get collectionKeys
+          collectionIsLikeArray = isArrayLike(collection);
```
pondermatic: It is faster to check the collectionIsLikeArray boolean than to compare value and type of collectionKeys and collection inside a loop.

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,124 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
           }
+          nextLength = collectionIsLikeArray ? collection.length : collectionKeys.length;
 
-          collectionLength = collectionKeys.length;
-          nextBlockOrder = new Array(collectionLength);
-
-          // locate existing items
-          for (index = 0; index < collectionLength; index++) {
-            key = (collection === collectionKeys) ? index : collectionKeys[index];
+          // setup nextBlockMap
```
pondermatic: This loop has a similar purpose to the "locate existing items" loop in the previous version.

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,124 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
-            } else {
-              // new never before seen block
-              nextBlockOrder[index] = {id: trackById, scope: undefined, clone: undefined};
-              nextBlockMap[trackById] = true;
+                'Duplicates in a repeater are not allowed. Use \'track by\' expression to specify unique keys. Repeater: {0}, Duplicate key: {1}, Duplicate value: {2}',
+                expression, trackById, value);
             }
-          }
 
-          // remove leftover items
```
pondermatic: This loop removes elements and now occurs after the "enter and move" loop.

src/ng/directive/ngRepeat.js
```diff
@@ -425,126 +419,124 @@ var ngRepeatDirective = ['$parse', '$animate', '$compile', function($parse, $ani
...
-              nextNode = previousNode;
+            if (lastBlockMap[nextKey]) {
+              // we have already seen this object and need to reuse the associated scope/element
+              block = lastBlockMap[nextKey];
 
-              // skip nodes that are already pending removal via leave animation
-              do {
-                nextNode = nextNode.nextSibling;
-              } while (nextNode && nextNode[NG_REMOVED]);
+              // move
```
pondermatic: The logic of when to move and where to move is a key change in this pull request.

pondermatic:
Commit 7a981722316779a8c14c5a1669d974db54b847ef ran the orderby-bp benchmark in 1037.09 ms for pass 1 and 1079.71 ms for pass 2. ngRepeat is now as fast as or a little bit faster than it was before.

Narretz:
Made some more perf tests and this looks to be basically as fast as before, nice!
@matsko do you think you have time to give this a quick look?

Narretz:
Hi @pondermatic, I'm very close to merging this half a year later ... while doing some tests, I found a small bug and fixed it here: https://github.com/Narretz/angular.js/commit/7dbf428765d3093cd0419c8dd5b4e7bd52cbb290 

Would you like to review and include this into your PR?

pondermatic:
Hi @Narretz. Thank you for catching that bug! Your fix and unit test work perfectly.

src/ng/directive/ngRepeat.js
Narretz: Indentation of the comment here is wrong

test/ng/directive/ngRepeatSpec.js
```diff
@@ -1566,6 +1572,107 @@ describe('ngRepeat animations', function() {
...
+
+      item = $animate.queue.shift();
+      expect(item.event).toBe('leave');
+      expect(item.element.text()).toBe('1');
+
+      item = $animate.queue.shift();
+      expect(item.event).toBe('leave');
+      expect(item.element.text()).toBe('2');
+    })
+  );
```
Narretz: Please put a newline between the two `it`s

test/ng/directive/ngRepeatSpec.js
```diff
@@ -1566,6 +1572,107 @@ describe('ngRepeat animations', function() {
...
+
+        expect(element.text()).toBe('ABCD'); // the original order should be preserved
+        $animate.flush();
+        $timeout.flush(1500); // 1s * 1.5 closing buffer
+        expect(element.text()).toBe('ABCD');
+
+        $rootScope.items = [items[0], items[1], items[3]];
+        $rootScope.$digest();
+
+        // The leaving item should maintain it's position until it is removed
```
Narretz: it's -> its (I know, I added this test ;))

Narretz (changes requested):
LGTM, except a few nitpicks. I want someone else from the team to look at this before merging as ngRepeat is an important directive.

Narretz:
Here's a plunker with ngRepeat and the fixed version to play around with: http://plnkr.co/edit/fu2SDoNfxCrUZPA0v0zm?p=preview

src/ng/directive/ngRepeat.js
mgol: @Narretz Is it? Looks good to me.

test/ng/directive/ngRepeatSpec.js
```diff
@@ -1488,7 +1488,13 @@ describe('ngRepeat animations', function() {
     $rootScope.items = ['1','3'];
     $rootScope.$digest();
 
-    item = $animate.queue.shift();
+    while ($animate.queue.length) {
+      item = $animate.queue.shift();
+      if (item.event === 'leave') {
+        break;
+      }
+    }
```
petebacondarwin: A comment to explain why this is needed would be helpful.

test/ng/directive/ngRepeatSpec.js
```diff
@@ -1566,6 +1572,107 @@ describe('ngRepeat animations', function() {
...
       expect(item.event).toBe('move');
       expect(item.element.text()).toBe('3');
+
+      item = $animate.queue.shift();
+      expect(item.event).toBe('move');
+      expect(item.element.text()).toBe('1');
+    })
+  );
+
+  it('should fire off the move animation for filtered items',
```
petebacondarwin: This description could be more clear. Perhaps:

```
"should fire the move animation for items that shifted due to removal of previous items"
```

Or something like that.

petebacondarwin (commented):
should the tests be checking the `previousNode` value that is passed to the move animation?

petebacondarwin:
Generally looks like a good refactoring. Thanks @pondermatic

test/ng/directive/ngRepeatSpec.js
pondermatic: I am going to revert this code. The ngRepeat tests assume the animation queue is in a certain order. During refactoring of ngRepeat, the order was different in this test so I altered the test. My latest version of ngRepeat, c969ae23c48eab32989435d8ed4f9767e89c5ad4, passes the original version of this test.

test/ng/directive/ngRepeatSpec.js
pondermatic: How about this description?
`should fire off the move animation for items that change position when other items are filtered out`

test/ng/directive/ngRepeatSpec.js
pondermatic: Should tests assume that animation events are queued in a certain order? If not, then perhaps the queue should be tested like this:
```javascript
// Test each queue item's event type
while ($animate.queue.length) {
  item = $animate.queue.shift();
  switch (item.element.text()) {
    case '2':
      expect(item.event).toBe('leave');
      break;
    case '3':
      expect(item.event).toBe('move');
      break;
    default:
      expect(item).toBeUndefined();
  }
}
```

Maybe we should add  a Jasmine matcher in the "ngRepeat Animations" block that compares item.event and item.text() values between two arrays? Should this be new GitHub issue?

test/ng/directive/ngRepeatSpec.js
petebacondarwin: +1

test/ng/directive/ngRepeatSpec.js
pondermatic: Apparently, the animation queue order IS different than it used to be.
Options:
1. use a while loop to search for the leave event, as in commit 2331c42d15ad294ba839a499d3e2109816b3e6f8
2. shift the queue array a second time, assuming that the leave event is the second item in the queue
3. add a Jasmine matcher that searches the queue for a item with a specific element text and event

As far as I can tell, tests should not assume the order of events in the queue. So I propose the following matcher be added to the 'ngRepeat animations' group of tests:
```javascript
beforeEach(function() {
  jasmine.addMatchers({
    toContainQueueItem: function() {
      return {
        compare: generateCompare(false),
        negativeCompare: generateCompare(true)
      };
      function generateCompare(isNot) {
        /**
         * Matcher that checks that the expected element text is in the actual Array of
         * animation queue items and that the event matches.
         * @param {array} actual
         * @param {string} expectedElementText
         * @param {string} expectedEvent optional if isNot = true
         * @returns {{pass: boolean, message: message}}
         */
        return function(actual, expectedElementText, expectedEvent) {
          if (expectedEvent === undefined) {
            expectedEvent = '';
          }
          var actualLength = actual.length;
          var i;
          var item;
          var message = valueFn(
            'Expected animation queue to ' + (isNot ? 'not ' : '') + 'contain an item where the element\'s text is "'
            + expectedElementText + '"' + (isNot ? '' : ' and the event is "' + expectedEvent + '"')
          );
          var pass = isNot;

          for (i = 0; i < actualLength; i++) {
            item = actual[i];
            if (item.element.text() === expectedElementText) {
              pass = item.event === expectedEvent;
              break;
            }
          }

          return {'pass': pass, 'message': message};
        };
      }
    }
  });
});
```
The tests should use the matcher like this:
```javascript
expect($animate.queue).not.toContainQueueItem('1');
expect($animate.queue).toContainQueueItem('2', 'leave');
expect($animate.queue).toContainQueueItem('3', 'move');
$animate.queue = [];
```
What does everyone think?

petebacondarwin (approved):

petebacondarwin (changes requested):
LGTM - except for the failing tests :-)

pondermatic:
I can't believe I pushed that change without running tests. :)

test/ng/directive/ngRepeatSpec.js
```diff
@@ -1424,15 +1424,58 @@ describe('ngRepeat animations', function() {
     };
   }));
 
+  beforeEach(function() {
```
pondermatic: Tests should not rely on the order of items in the animation queue. This Jasmine matcher looks for a queued item with a given element text and matches against that item's event.

petebacondarwin (approved):

Narretz:
I decided to do some more tests, and it looks like we created a performance problem with the new behavior when ngAnimate is used. See this updated plnkr: http://plnkr.co/edit/fu2SDoNfxCrUZPA0v0zm?p=preview

The second repeat renders 10000 elements (I know that you generally should avoid these huge lists)

-  Now remove the second item in the array. Only Chrome handles this gracefully. Firefox and Edge have a noticable delay. become unresponsive and might even display "page not responding" warnings
- This behavior occurs even when you remove the animation styles, because ngAnimate will still check each item for animatable styles. Only when you disable animations or remove ngAnimate, perf returns to normal
- The bottleneck is not the ngRepeat computation or the dom operation
- ngAnimate caches the transitions / animations, so the problem mainly lies with the application / rendering of the styles to the elements (essentially the browser engine).
- with 1000 items, the performance is okay

@matsko had done some interesting changes with regard to caching animation styles here: https://github.com/angular/angular.js/pull/14166 With ngAnimate enabled and no actual animation styles, the whole process is aborted earlier with this PR, but there is no noticable performance gain. 

So the question is, are we okay with the perf dropoff with large repeats? There are many ways to disable animations on perf critical parts of the page, but this might still bite people whose apps will suddenly perform worse.

I suggest we should at least look into making ngAnimate faster when no actual css animations are present.

Issue #15018, opened 2016-08-12
Trailing space issue in Angular Locale


TudorSfatosu:
**_Note_: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a _feature_ or report a _bug_?**
Reporting a bug

**What is the current behavior?**
A trailing space is added to RO and IT angular locale, maybe in similar locales as well.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**
DEMO: http://jsfiddle.net/4gyjtu8g/245/
![one](https://cloud.githubusercontent.com/assets/6681480/17619465/785a5a26-607d-11e6-958d-eebd4540df75.png)

**What is the expected behavior?**
There should be no trailing space, like the EN angular locale

**What is the motivation / use case for changing the behavior?**
To fix a general bug

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**
http://ajax.googleapis.com/ajax/libs/angularjs/1.3.0/angular.min.js
https://cdnjs.cloudflare.com/ajax/libs/angular-i18n/1.5.8/angular-locale_ro.js

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
![two](https://cloud.githubusercontent.com/assets/6681480/17619560/f16ddfb4-607d-11e6-91ac-77afce803daf.png)

gkalpak:
This is kind of expected and is not specific to the `ro` locale. All locales that included a space before or after the currency symbol are affected in the same way.

According to the `ro` locale file, the currency-formatted number should be suffix with ` <CURRENCY_SYMBOL>` (a space, followed by the currency symbol).
By default, the currency symbol is `RON`: `currencyFilter(10)  -->  '10.00 RON'`
When you are replace the default currency symbol with the empty string, the result will still contain the space: `currencyFilter(10, '')  -->  '10.00 '`

You won't see the same problem with the default `en` locale, because in that locale there is no space between the number and the currency symbol:
`currencyFilter(10)  -->  '$10'` (no space before `10`)
`currencyFilter(10, '')  -->  '10'` (no space before `10`)

--
[...]

PR #15080, opened 2016-09-01
feat(filters): add trim filter

austinoneil:

feature

There is no trim filter.  This feature was in response to [Issue #15018.](https://github.com/angular/angular.js/issues/15018)

There is a built-in trim filter.

no

- [yes ] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ yes] Tests for the changes have been added (for bug fixes / features)
- [ yes] Docs have been added / updated (for bug fixes / features)

**Other information**:

add trim filter. Motivation: issue #15018 requests PRs.

### DIFF
--- a/src/ng/filter.js
+++ b/src/ng/filter.js
@@ -8,6 +8,7 @@
  lowercaseFilter: true,
  numberFilter: true,
  orderByFilter: true,
+ trimFilter: true,
  uppercaseFilter: true,
  */
 
@@ -156,6 +157,7 @@ function $FilterProvider($provide) {
     lowercaseFilter: false,
     numberFilter: false,
     orderByFilter: false,
+	trimFilter: false,
     uppercaseFilter: false
   */
 
@@ -168,4 +170,5 @@ function $FilterProvider($provide) {
   register('number', numberFilter);
   register('orderBy', orderByFilter);
   register('uppercase', uppercaseFilter);
+  register('trim', trimFilter);
 }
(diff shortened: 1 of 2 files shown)


petebacondarwin:
@austinoneil thanks for sending this in. Trim is an easy filter to create but we don't want to add more of this sort of stuff to the core library. Instead it should be a 3rd party module that people can install if needed. See http://ngmodules.org/

petebacondarwin:
I think that @gkalpak was not suggesting that we add it to the core but that one could make one's own trim filter to clean up his example.

austinoneil:
For future reference, what does the "PRs plz!" tag mean?

petebacondarwin:
@austinoneil your assumption was correct that `PRs plz!` suggests that we would like people to put together a PR for an issue. In this case the request, if I understand it correctly, was to fix the `currency` filter to do the trimming automatically. Perhaps @gkalpak  and @wesleycho could clarify?

wesleycho:
I don't think this PR is relevant to the discussion in #15018 - that discussion was specifically about the currency filter and a particular string there being trimmed due to certain locales. One can create such a filter like this in a user's app, but this type of filter shouldn't be in core.

Instead, what I suggested was that perhaps within the currency filter the relevant string be trimmed so that the issue in particular locales would not have an extra space at the end of the string.

gkalpak:
@petebacondarwin is right. I did not suggest we created a `trim` filter in core. In order to adrress #15018 (in core), we need to "incorporate" some trimming into `currencyFilter`. That is what the `PRs plz!` label refers to.

Introducing a `trim` filter was suggested as one of the possible workarounds for people to avoid this issues **until** this is properly fixed in core (by fixing `currencyFilter`) - not as something to be introduced in core.

Sorry, for the confusion, @austinoneil :disappointed: (I'll try to clarify this a bit more in #15018).

austinoneil:
yeah, I made [a new pull request.](https://github.com/angular/angular.js/pull/15085)

PR #15085, opened 2016-09-02
fix(currencyFilter): trim currency string

austinoneil:

Bug fix

For certain locales, when the `currencySymbol` is set to an empty string, there is trailing white space. [See issue #15018 ](https://github.com/angular/angular.js/issues/15018)

Any currency string will be trimmed.

yes

**Other information**:

Trim any beginning or trailing white space in a currency string.
Motivation: global fix for issue 15018

### DIFF
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -69,10 +69,11 @@ function currencyFilter($locale) {
     }
 
     // if null or undefined pass it through
-    return (amount == null)
+    var currencyString = (amount == null)
         ? amount
         : formatNumber(amount, formats.PATTERNS[1], formats.GROUP_SEP, formats.DECIMAL_SEP, fractionSize).
             replace(/\u00A4/g, currencySymbol);
+	return currencyString.trim();
   };
 }
 
--- a/src/ng/filter/filters.js
+++ b/src/ng/filter/filters.js
@@ -27,12 +27,6 @@ var ZERO_CHAR = '0';
            .controller('ExampleController', ['$scope', function($scope) {
              $scope.amount = 1234.56;
            }]);
-		 angular.module('otherLocaleExample', []) {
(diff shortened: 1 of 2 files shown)


austinoneil:
Correction: this is a breaking change.

austinoneil:
Does anybody know how I would change locale for an automated test?

Narretz:
You mean in a unit test?

austinoneil:
Yes

Narretz:
Why did you remove the fix from the commit?

austinoneil:
@Narretz The fix is still there. I was just trying things to get a unit test. Check the "files changed" tab.

Narretz:
Ah I see, sorry. The test code you changed is a docs end 2 end test,  not a unit test though. The filter spec files are in the test folder.

austinoneil:
@Narretz Thanks!  Got the unit test in.

gkalpak:
This does not fix #15018. In fact it just fixes certain cases (i.e. locales which put the currency symbol _after_ the number), but is not a generic fix (see https://github.com/angular/angular.js/issues/15018#issuecomment-239480197 for more details).

I am tempted to merge this even if it fixes only some of the cases (as a proper fix is quite more involved), but that would be "unfair" to certain locales :/

austinoneil:
@gkalpak Would appending a `.replace(\-/s*\g, '-')` fix the potential issue with having a space between the negative sign and the currency amount?  I don't see any locale delimiting their currency with anything that would make the fix problematic.

I'll try to add some unit tests and fix based off those when I get home.

gkalpak:
No, I don't think it would work for all cases (although it would take care of several more than the current implementation). Take into account that people might be overwriting the locals `pattern` object. E.g. in finance/accounting-related apps people prefer to display negative values wrapped in `(...)` instead of prefixing them with `-`.

I have a more general fix. I will submit a PR.

PR #15087, opened 2016-09-02
fix(ngLocale): set correct first day of week for french people

fl0s:
Update locale file

**Other information**:

For french people, first day of week is monday


Narretz:
Hi, Angular.js uses the Google Closure I18N library to generate its own I18N files (the ngLocale module). This means that any changes to these files would be lost the next time that we import the library. Since the Closure library i18n data is itself auto-generated from the data of the Common Locale Data Repository (CLDR) project, errors in the data should be reported there. See also the Closure guide to i18n changes.

PR #15089, opened 2016-09-02
docs(tutorial/8 - Templating Links

Ruanito:
What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)
docs update

What is the current behavior? (You can also link to an open issue here)

What is the new behavior (if this is a feature change)?

Does this PR introduce a breaking change?

Please check if the PR fulfills these requirements


Ruanito:
I signed it!

gkalpak:
**tl;dr**
The links are correct (withou the `!`) for step 8.

Te longer version:
The tutorial uses version 1.5.x, on which the default hash-prefix is `''` (the empty string).
So, unless we configure `$location` differently, the links should not contain anything between `#` and the path. 
In step 9, we set the hash-prefix to `'!'` (note: this will be the default in 1.6.x, but not in 1.5.x), so from that point onwards, all links must have `!` between the `#` and the path. (The links are indeed updated in step 9 of the tutorial).

PR #15098, opened 2016-09-06
fix($compile): avoid calling `$onChanges()` twice for `NaN` initial values

gkalpak:
Bug fix.

When the initial value of a binding is `NaN`, `$onChanges()` will be called twice (because we fail to detect that `NaN` equals `NaN`. For all other values, `$onChanges()` is called once.

`$onChanges()` is always called once, even if the initial value of a binding is `NaN`.

No

**Other information**:
Discussed in #15095.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3500,7 +3500,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       });
 
       function recordChanges(key, currentValue, previousValue) {
-        if (isFunction(destination.$onChanges) && currentValue !== previousValue) {
+        if (isFunction(destination.$onChanges) && currentValue !== previousValue &&
+            // eslint-disable-next-line no-self-compare
+            (currentValue === currentValue || previousValue === previousValue)) {
           // If we have not already scheduled the top level onChangesQueue handler then do so now
           if (!onChangesQueue) {
             scope.$$postDigest(flushOnChangesQueue);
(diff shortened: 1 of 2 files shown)


src/ng/compile.js
```diff
@@ -3500,7 +3500,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       });
 
       function recordChanges(key, currentValue, previousValue) {
-        if (isFunction(destination.$onChanges) && currentValue !== previousValue) {
+        if (isFunction(destination.$onChanges) && currentValue !== previousValue &&
+            // eslint-disable-next-line no-self-compare
+            (currentValue === currentValue || previousValue === previousValue)) {
```
petebacondarwin: But this will not trigger a call to `$onChanges` if we go from a real number to `NaN`, no?

petebacondarwin: But I think a test of that situation would be useful

petebacondarwin: Landed but forgot to add this extra test.

gkalpak: I'll add the extra test. It is good excuse to fix [an indentation issue][1] :stuck_out_tongue: 

[1]: https://github.com/angular/angular.js/commit/7d7efbf545c8c07713eb45301660dcfca4121445#diff-348c2f3781ed66a24894c2046a52c628R4400

gkalpak: For reference: c0cbe54

petebacondarwin: You wouldn't believe how much time I spent fixing up the indentation in my conflict editor :-)

PR #15112, opened 2016-09-09
feat($compileProvider): use options as controller in .component if is function or class

drpicox:
New feature.

When creating a component you must specify the controller unless you are ok using an empty controller.

If you use a class or function as Component config object it becomes the controller unless it specifies a controller.

Yes.

Small breaking change with probably low affectation (if any): 
if the config object is a function or class and no controller is defined,
now it will execute it and use its instance as controller instead of `function() {}`

**Other information**:

The reason of this  feature is to enable to use classes as components with the little effort possible and closer to Angular 2.

> #### Without this PR:

Now if you want to create a component with a controller instantiated from a class you need to do something like:

``` javascript
class EditableFieldController {
  constructor() {
    this.editMode = false;
  }

  handleModeChange() {
    if (this.editMode) {
      this.onUpdate({value: this.fieldValue});
      this.fieldValueCopy = this.fieldValue;
    }
    this.editMode = !this.editMode;
  }

  reset() {
    this.fieldValue = this.fieldValueCopy;
  }

  $onInit() {
    this.fieldValueCopy = this.fieldValue;

    if (!this.fieldType) {
      this.fieldType = 'text';
    }
  }
}

export const EditableFieldComponent = {
  templateUrl: 'heroDetail.html',
  bindings: {
    fieldValue: '<',
    fieldType: '@?',
    onUpdate: '&'
  },
  controller: EditableFieldController,
};
```

Currently, there is a **trick** (at least if you are using Typescript) that allows you to do the following:

``` javascript
export class EditableFieldComponent {
  static controller = EditableFieldComponent; // *** this is the trick
  static templateUrl = 'heroDetail.html';
  static bindings = {
    fieldValue: '<',
    fieldType: '@?',
    onUpdate: '&'
  };

  editMode = false;

  handleModeChange() {
    if (this.editMode) {
      this.onUpdate({value: this.fieldValue});
      this.fieldValueCopy = this.fieldValue;
    }
    this.editMode = !this.editMode;
  }

  reset() {
    this.fieldValue = this.fieldValueCopy;
  }

  $onInit() {
    this.fieldValueCopy = this.fieldValue;

    if (!this.fieldType) {
      this.fieldType = 'text';
    }
  }
}
```

> #### With this PR:

This PR sets as default the Component config object as controller
if it is a function or class. So you do not need to use _the trick_.

So in ES2015 you can do:

``` javascript
export class EditableFieldComponent {
  static get templateUrl() {
    return 'heroDetail.html';
  }
  static get bindings() {
    return {
      fieldValue: '<',
      fieldType: '@?',
      onUpdate: '&'
    };
  }

  constructor() {
    this.editMode = false;
  }

  handleModeChange() {
    if (this.editMode) {
      this.onUpdate({value: this.fieldValue});
      this.fieldValueCopy = this.fieldValue;
    }
    this.editMode = !this.editMode;
  }

  reset() {
    this.fieldValue = this.fieldValueCopy;
  }

  $onInit() {
    this.fieldValueCopy = this.fieldValue;

    if (!this.fieldType) {
      this.fieldType = 'text';
    }
  }
}
```

And in esnext, babel, traceur and typescript you can do:

``` javascript
export class EditableFieldComponent {
  static templateUrl = 'heroDetail.html';
  static bindings = {
    fieldValue: '<',
    fieldType: '@?',
    onUpdate: '&'
  };

  editMode = false;

  handleModeChange() {
    if (this.editMode) {
      this.onUpdate({value: this.fieldValue});
      this.fieldValueCopy = this.fieldValue;
    }
    this.editMode = !this.editMode;
  }

  reset() {
    this.fieldValue = this.fieldValueCopy;
  }

  $onInit() {
    this.fieldValueCopy = this.fieldValue;

    if (!this.fieldType) {
      this.fieldType = 'text';
    }
  }
}
```

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -73,7 +73,7 @@ It's also possible to add components via {@link $compileProvider#component} in a
 | bindings          | No                   | Yes (binds to controller)      |
 | bindToController  | Yes (default: false) | No (use bindings instead)       |
 | compile function  | Yes                  | No       |
-| controller        | Yes                  | Yes (default `function() {}`)     |
+| controller        | Yes                  | Yes (default `function() {}` or _the Component config object_ if it is a function or class )     |
 | controllerAs      | Yes (default: false) | Yes (default: `$ctrl`)       |
 | link functions    | Yes                  | No       |
 | multiElement      | Yes                  | No       |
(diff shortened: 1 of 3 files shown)


gkalpak:
Without having looked at the implementation, I do not like this feature. It adds very little imo (it just saves one line as you mentioned yourself (_the "trick"_)), while complicating things a bit, adding yet another way to do the same thing (which is something we've been trying to avoid) and introducing a breaking change.

> The reason of this feature is to enable to use classes as components with the little effort possible

I don't see how this reduces the effort (unless you mean removing that one line, which totally not worth it imo). Also note that _the "trick"_ is not only possible with TypeScript, but with ES2015 as well (with `static get controller() { return EditFieldComponent; }` - [demo][1]).

> The reason of this feature is to enable to use classes as components [...] closer to Angular 2

I would argue that...

```js
{
  template: '...',
  bindings: {...}
  controller: class MyComponent {
    constructor() {...}
    $onInit() {...}
    ...
  }
}
```
...is actually closer to Angular 2 than...
```js
class MyComponent {
  static get template() { return '...'; }
  static get bindings() { return {...}; }

  constructor() {...}
  $onInit() {...}
  ...
}
```

--
So, it's :-1: from me for this one.
But totally :+1: for all the great ideas and work that you've been putting in to this and other features lately-ish. Really great stuff!

[1]: https://codepen.io/ExpertSystem/pen/qaOdLE?editors=0011

drpicox:
Ok! 
That one was really easy (just one line).

The idea was to use one class declaration and export that class declaration, nothing else, and avoid have them sepparated in two files like https://github.com/toddmotto/angular-styleguide#stateful-components .

But I'm totally ok with you. I do not like the idea to mix everything in one single structure.
So, thanks for all all you guys 👍

PR #15113, opened 2016-09-09
fix(ngAria): Prevent aria-invalid from being added to hidden inputs

edclements:
This fixes a error found using the Google Accessibility Developer Tools audit.
Input fields of type hidden shouldn't have aria attributes.
https://www.w3.org/TR/html-aria/#allowed-aria-roles-states-and-properties-1

### DIFF
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -421,7 +421,7 @@ describe('$aria', function() {
       expect(element.attr('aria-invalid')).toBe('userSetValue');
     });
 
-    it('should not attach if input is hidden', function() {
+    it('should not attach if input is type="hidden"', function() {
       compileElement('<input type="hidden" ng-model="txtInput">');
       expect(element.attr('aria-invalid')).toBeUndefined();
     });
--- a/test/ngAria/ariaSpec.js
+++ b/test/ngAria/ariaSpec.js
@@ -420,6 +420,11 @@ describe('$aria', function() {
       scope.$apply('txtInput=\'LTten\'');
       expect(element.attr('aria-invalid')).toBe('userSetValue');
     });
+
+    it('should not attach if input is hidden', function() {
+      compileElement('<input type="hidden" ng-model="txtInput">');
+      expect(element.attr('aria-invalid')).toBeUndefined();
+    });
   });
 
   describe('aria-invalid when disabled', function() {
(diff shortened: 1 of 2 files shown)


edclements:
I signed it!

gkalpak:
> Input fields of type hidden shouldn't have aria attributes.

We are already excluding all input fields as part of the `!isNodeOneOf(...)` check.
I al going to close this, since everything seems to be working as expected, but feel free to continue the discussion below.

BTW, the section of the spec you have linked to does not refer to what your PR is about. It refers to elements with a `hidden` attribute (e.g. `<div hidden>`).

edclements:
The isNodeOneOf check doesn't apply for aria-invalid as allowBlacklistEls is being passed in as true. The test I wrote will fail without the code change.

The relevant row in the spec reads as follows:
input type = hidden	No corresponding role	No role or aria* attributes

gkalpak:
Sorry, I read this wrong. I am reopening and will review soon.
In the meantime, make sure you have signed the CLA (or we won't be able to merge this).

test/ngAria/ariaSpec.js
```diff
@@ -420,6 +420,11 @@ describe('$aria', function() {
       scope.$apply('txtInput=\'LTten\'');
       expect(element.attr('aria-invalid')).toBe('userSetValue');
     });
+
+    it('should not attach if input is hidden', function() {
```
gkalpak: Nitpick: It would be more clear to say: _...if input is **type=**"hidden"_

gkalpak:
LGTM (but we still need a signed CLA to merge).
@edclements, feel free to ping me when/if you have sorted it out.

edclements:
Thanks for taking a look @gkalpak. I signed the CLA as a company: BookingBug.

gkalpak:
@IgorMinar (or someone else), can you verify the CLA?

IgorMinar:
@gkalpak BookingBug is good! 👍

gkalpak:
Actually, I am going to mark this as a breaking change and merge it into 1.6 only. It might break tests (unlikely but possible).

gkalpak:
We decided not to add more breaking changes into 1.6 at this point, so moving this to 1.7.

Narretz:
LGTM2 and it can be merged because master is on 1.7

gkalpak:
Actually, I just realized that -while the [the spec][1] only mentions `input[type=hidden]` - this PR affects all `[type=hidden]` elements. @edclements can you update it so that it ignores `input` element's onlyand add a corresponding test?

[1]: https://www.w3.org/TR/html-aria#input-hidden

Narretz:
@edclements it's been a while but are you still interested / available to make the changes required for this PR? See https://github.com/angular/angular.js/pull/15113#issuecomment-276305376

Narretz:
I've created an updated PR here: https://github.com/angular/angular.js/pull/16367 No need to update this PR anymore. Thanks @edclements for getting this on the road.

PR #15114, opened 2016-09-09
Replace arrays and `HashMap` with native `Map` (with a fallback shim implementation)

gkalpak:
Perf improvement (from incorporating #13209) and several refactorings.

Same old, same old.

See #13209. On top of that, the internal `HashMap`/`$$HashMap` is replaced with native `Map`/`NgMapShim`, which should improve performance when native implementations are available (i.e. in most cases) and probably slightly decrease speed when falling back to the shim (tbd).

No.

**Other information**:
This incorporates and builds on top of #13209.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -89,7 +89,6 @@
   hasOwnProperty,
   createMap,
   stringify,
-  ES6Map,
 
   NODE_TYPE_ELEMENT,
   NODE_TYPE_ATTRIBUTE,
@@ -750,61 +749,6 @@ function arrayRemove(array, value) {
   return index;
 }
 
-// A minimal ES6 Map implementation.
-// Should be bug/feature equivelent to the native implementations of supported browsers.
-// See https://kangax.github.io/compat-table/es6/#test-Map
-var nanPlaceholder = Object.create(null);
-function ES6MapShim() {
-  this._keys = [];
-  this._values = [];
-  this._lastKey = NaN;
-  this._lastIndex = -1;
-}
-ES6MapShim.prototype = {
-  _idx: function(key) {
-    if (key === this._lastKey) {
-      return this._lastIndex;
-    }
-    return (this._lastIndex = (this._keys.indexOf(this._lastKey = key)));
-  },
-  _transformKey: function(key) {
-    return isNumberNaN(key) ? nanPlaceholder : key;
-  },
-  get: function(key) {
-    var idx = this._idx(this._transformKey(key));
-    if (idx !== -1) {
-      return this._values[idx];
-    }
-  },
-  set: function(key, value) {
-    key = this._transformKey(key);
-    var idx = this._idx(key);
-    if (idx === -1) {
-      idx = this._lastIndex = this._keys.length;
-    }
-    this._keys[idx] = key;
-    this._values[idx] = value;
-
-    // Support: IE11
-    // Do not `return this` to simulate the partial IE11 implementation
-  },
-  delete: function(key) {
-    var idx = this._idx(this._transformKey(key));
-    if (idx === -1) {
-      return false;
-    }
-    this._keys.splice(idx, 1);
-    this._values.splice(idx, 1);
-    this._lastKey = NaN;
-    this._lastIndex = -1;
-    return true;
-  }
-};
-
-var ES6Map = isFunction(window.Map) && toString.call(window.Map.prototype) === '[object Map]'
-  ? window.Map
-  : ES6MapShim;
-
 /**
  * @ngdoc function
  * @name angular.copy
@@ -898,7 +842,7 @@ function copy(source, destination) {
   return copyElement(source);
 
   function copyRecurse(source, destination) {
-    (stack || (stack = new ES6Map())).set(source, destination);
+    (stack || (stack = new NgMap())).set(source, destination);
 
     var h = destination.$$hashKey;
     var key;
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -749,7 +749,9 @@ function arrayRemove(array, value) {
   return index;
 }
 
-
+// A minimal ES6 Map implementation.
+// Should be bug/feature equivelent to the native implementations of supported browsers.
+// See https://kangax.github.io/compat-table/es6/#test-Map
 function ES6MapShim() {
   this._keys = [];
   this._values = [];
@@ -777,29 +779,15 @@ ES6MapShim.prototype = {
     }
     this._keys[idx] = key;
     this._values[idx] = value;
-    return this;
+
+    // Support: IE11
+    // Do not `return this` to simulate the partial IE11 implementation
   }
 };
 
-function testES6Map(Map) {
-  var m, o = {};
-  return isFunction(Map) && (m = new Map())
-
-    // Required functions
-    && isFunction(m.get) && isFunction(m.set)
-
-    // Number keys, must not call toString
-    && m.get(1) === undefined
-    && m.set(1, o) === m
-    && m.get(1) === o
-    && m.get('1') === undefined
-
-    // Object keys, must use instance as key and not call toString
-    && m.set(o, 2) === m && m.get(o) === 2
-    && m.get({}) === undefined;
-}
-
-var ES6Map = testES6Map(window.Map) ? window.Map : ES6MapShim;
+var ES6Map = isFunction(window.Map) && toString.call(window.Map.prototype) === '[object Map]'
+  ? window.Map
+  : ES6MapShim;
 
 /**
  * @ngdoc function
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -753,18 +753,27 @@ function arrayRemove(array, value) {
 function ES6MapShim() {
   this._keys = [];
   this._values = [];
+  this._lastKey = NaN;
+  this._lastIndex = -1;
 }
 ES6MapShim.prototype = {
+  _idx: function(key) {
+    if (key === this._lastKey) {
+      return this._lastIndex;
+    }
+    return (this._lastIndex = (this._keys.indexOf(this._lastKey = key)));
+  },
   get: function(key) {
-    var idx = this._keys.indexOf(key);
+    var idx = this._idx(key);
     if (idx !== -1) {
       return this._values[idx];
     }
   },
   set: function(key, value) {
-    var idx = this._keys.indexOf(key);
+    var idx = this._idx(key);
     if (idx === -1) {
-      idx = this._keys.length;
+      idx = this._lastIndex = this._keys.length;
+      this._lastKey = key;
     }
     this._keys[idx] = key;
     this._values[idx] = value;
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -749,6 +749,49 @@ function arrayRemove(array, value) {
   return index;
 }
 
+
+function ES6MapShim() {
+  this._keys = [];
+  this._values = [];
+}
+ES6MapShim.prototype = {
+  get: function(key) {
+    var idx = this._keys.indexOf(key);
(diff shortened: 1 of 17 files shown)


gkalpak:
I haven't checked how these changes affect performance for various cases.

Things that are affected (and the effects are likely to be noticeable):

- `copy()` (and anything using `copy()` internally).
- `ngAnimate`

Things that are affected (and the effects are unlikely to be noticeable):

- `$injector`
- `select`
- `ngMocks`

gkalpak:
Safari 8 is failing. Buggy implementation of `window.Map`?

src/Angular.js
```diff
@@ -765,23 +767,37 @@ ES6MapShim.prototype = {
...
     if (idx !== -1) {
       return this._values[idx];
     }
   },
   set: function(key, value) {
+    key = this._transformKey(key);
     var idx = this._idx(key);
     if (idx === -1) {
       idx = this._lastIndex = this._keys.length;
-      this._lastKey = key;
```
jbedard: Won't deleting this line break something? Every assignment of `lastIndex` must be beside an assignment to `lastKey` (and the reverse)...

src/Angular.js
```diff
@@ -765,23 +767,37 @@ ES6MapShim.prototype = {
...
     // Do not `return this` to simulate the partial IE11 implementation
+  },
+  delete: function(key) {
+    var idx = this._idx(this._transformKey(key));
+    if (idx === -1) {
+      return false;
+    }
+    this._keys.splice(idx, 1);
+    this._values.splice(idx, 1);
+    this._lastKey = NaN;
```
jbedard: Only have to reset the `last*` if `idx <= lastIdx`, I think? Not sure if it's worth any extra logic here though...

jbedard:
Only think I can see is safari not support constructor arguments: https://kangax.github.io/compat-table/es6/#test-Map

src/Angular.js
gkalpak: `this._lastKey` has been set to `key` in the previous call to `_idx()`.
We need to re-assign `this._lastIndex` here, because the currently set `-1` won't be true any more (since we are adding the item), but `this._lastKey` stays the same.

src/Angular.js
gkalpak: Not true. Actually true, but `this._lastIndex` will be set to `idx` inside the `_idx()` call, so `idx === this._lastIdx` will always be true.

src/Angular.js
jbedard: Yeah, looks like you're right. That isn't really obvious just looking at this though. Worth putting it back? Or more likely just a comment?

gkalpak:
Actually, it could also be a buggy implementation of something unrelated to `Map`, e.g. `splice`.
I'll take a look on SauceLabs (since I don't have local access to Safari 8 :disappointed:).

src/AngularPublic.js
```diff
@@ -262,7 +262,7 @@ function publishExternalAPI(angular) {
         $window: $WindowProvider,
         $$rAF: $$RAFProvider,
         $$jqLite: $$jqLiteProvider,
-        $$HashMap: $$HashMapProvider,
+        $$Map: $$MapProvider,
```
jbedard: Do we even need this? If we're going to rename it (breaking change), can we just delete it?

gkalpak: This is a private service, so no breaking change (theoretically). It is used by `ngAnimate` (which doesn't have access to the closure).

src/Angular.js
gkalpak: It might not be obvious if you are trying to figure out why it being deleted, but if you look at the code (without knowing it was ever there), I think it is clearer without the comment.
I am fine adding if you still think it helps, though.

src/Angular.js
jbedard: Right...

gkalpak:
I couldn't find anything suspicious with Safari 8 + Map. I restarted the job (just in case).

thorn0:
It's Mobile Safari, not the desktop 8.0 one. It says `Mobile Safari 8.0.0 (iOS 10.10.5)`. The tests for #15119 (a docs-only PR!) are also failing in this browser. It might be something like this https://github.com/jquery/jquery/issues/2145

mgol:
> It says `Mobile Safari 8.0.0 (iOS 10.10.5)`

I assume they mean OS X 10.10.5, not iOS 10.10.5... Sauce Labs should make it less confusing.

mgol:
...but that would mean they're running it on an iOS simulator on OS X and that, in turn, eliminates the bug you linked to as it only manifested itself on real devices.

thorn0:
I'm just saying it may be a similar randomly appearing bug, not this specific one. Yes, it's a simulator, and it's even not totally clear which version of iOS it simulates.

gkalpak:
Doh! RTFLog. So, it's **Mobile** Safari. Let's see...
(Thx for the head-up guys :smile: )

thorn0:
Travis was restarted for #15119, the tests for it are passing now, and it seems that the log from the previous run has been lost. At least, I can't find it any more. However, that log could of interest for those who will investigate this Safari heisenbug, and I remember the most important part. 

Two tests failed in the same way: `TypeError: null is not an object` on the line 5961 of [jQuery 2.2.4](https://unpkg.com/jquery@2.2.4). And the next line of the stack traces was 6217 from where `getWidthOrHeight` was called. If you look closer at this code and this error, it becomes obvious that this must be a browser bug.

gkalpak:
Yeah, I restarted the test on #15119. It seems unrelated to this one. It was probably a flake, but this one fails consistently on the same test.

gkalpak:
The `Map` implementation in `Mobile Safari 8` is really bad (see also https://github.com/google/traceur-compiler/issues/1810#issuecomment-91179335). Although we might be able to work around this specific failure, I am afraid other incomplete or buggy implementations will cause Angular to fail unexpectedly (and randomly).

Maybe we need stricter checks for `Map` (e.g. [es6-shim][1] has many more).

[1]: https://github.com/paulmillr/es6-shim/blob/3e46993a95b1867665166e0290e5cbaaf2ccb2f6/es6-shim.js#L3343-L3354

mgol:
@gkalpak The comment only mentions that the `Map` constructor doesn't accept parameters (which we already know) and that its iterator is non-compliant (but we don't use it anyway, do we?). Any other problems?

gkalpak:
Our specific problem (which I am not exaclty sure what it is) is indeed not mentioned in the comment, but it is a discussion about the incompleteness of the implementation of Map in that version of Safari.

So, although we don't need any of the _known_ missing or buggy features, we are affected by some other bug, which results in not being able to retrieve values associated with `jqLite` objects as keys. Unfortunately, since it's iterators do not have a `.next()` methos, nor does it support `for...of` or `Array.from()` and its `forEach()` methods iterates over the values, I couldn't find a way to inspect the contents of the Map closer.

--
If anyone is interested, in order to test this, I changed the code to:

- Always use `NgMapShim` (even if `window.Map` was available).
- Instantiate a `window.Map` within every `NgMapShim`.
- For every operation (`get`/`set`/`delete`) that was performed with the `NgMapShim` instance, performa the same operation on the native `Map`, compare the results and report any differences.

There were a bunch of differences (in `GET` operations only), where the native `Map` retunred `undefined`, while the shim returned something - in all cases the sizes of the two maps where the same (and although I haven't kept proof, the values were there as well - can't tell anything about the keys).

[Here][1] is a Travis log showing (some of) the above (not sure how long Travis keeps old logs around).

[1]: https://travis-ci.org/angular/angular.js/builds/159011401

gkalpak:
BTW, I updated the commit to use the shim if `(new Map()).keys().next` is not a function and the tests are green. But I am still not sure how safe it is to rely on native implementations :disappointed:

jbedard:
Originally I had a [test method](https://github.com/angular/angular.js/pull/13209/commits/d8f593c10017cd7ac2db220a7b1d8bbf7bb49d19#diff-1d54c5f722aebc473dbe96f836ddf974R769) before [switching](https://github.com/angular/angular.js/pull/13209/commits/6a102ef45333cdcb86ec8cace5fc9584e617b4c9) to essentially `isNative(window.Map)`.  Maybe we do need that to detect these these buggy implementations? The tests will have to catch errors like this one, whatever this one is...

jbedard:
Why specifically `(new Map()).keys().next`? I assume that makes IE11 not pass either?

gkalpak:
> Why specifically (new Map()).keys().next? I assume that makes IE11 not pass either?

True :disappointed: There is nothing specific about it, except that was a way to detect Safari 8 (and probably other browsers with incomplete implementations - but definitely not all).

I am fine doing the detection in another, more targeted way, but my main concern remains:
Are we confident enough that our tests would have caught any issues with incomplete/buggy `Map` implementations?
(I don't think I am. For example, note there are dozens of inconsistencies between our shim and the native `Map` on Mobile Safari 8 (see [that log][1] is still available), but only one test fails when using the native one. Who says there are more failures, which our tests didn't happen to catch?)

[1]: https://travis-ci.org/angular/angular.js/builds/159011401

jbedard:
I'd prefer finding ways of detecting specific bugs, instead of just detecting Safar8-like implementations. But then it might get too ugly, having a test that creates `jqLite("<input class='a b c'>")` etc...

That said, if this allows us to replace `HashMap` it might be worth it. Unlike originally where this was only for `copy`. I think it might be worth it...

jbedard:
Looking through the [es6-shim checks](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L3190) I don't see anything looking for a bug like this :|

thorn0:
It'd be interesting to run [test262](https://github.com/tc39/test262) in this simulator, but what about real iphone and ipads, not simulated? Is Angular actually tested on them?

gkalpak:
I ran the `ngBind` version of [largetable benchmark][1] (on Chrome) with `$$HashMap`, `window.Map` and `NgMapShim`. This benchmark has a lot of `ngRepeat`ed nodes, which use `$animate.enter()`, which adds/removes the `ng-enter/leave[-active]` classes, which in turn uses the Map-based queue under the hood. This is probably the most common way (along with `copy()`ing deeply watched objects) that Angular will be exercising the `Map`.

It didn't see any consclusive difference between the three (which probably means that the time spent interacting with `Map` is not significant in the big picture). For reference, here are the values:

|                   | `$$HashMap` | `window.Map` | `NgMapShim` |
|----------------|-------------------:|-------------------:|--------------------:|
| **destroy** |           178 ms |           173 ms |           188 ms |
| **create**   |          949 ms |            954 ms |           924 ms |
| **$apply**  |             11 ms |              10 ms |             12 ms |

--
@jbedard, how did you meassure the performance difference for `copy()`?

[1]: https://github.com/angular/angular.js/tree/d14c7f3c31deb098bf8f1c50ea6d00af758dbdcb/benchmarks/largetable-bp

jbedard:
https://github.com/angular/angular.js/commit/f7bf118ecaaab16ed8b3d16febb0ee6bb1d6749c is the copy benchmark I normally use

gkalpak:
@jbedard, was that benchmark included into angular/angular.js at some point or are you using it locally?

jbedard:
That benchmark is not checked in and I just use it locally. At some point I committed it to a local branch to show someone the benchmark I was using, so github does know about the commit but it was never merged.

gkalpak:
We have decided to:

- Replace the current `HashMap` implementation with the `NgMapShim` implementation that is closer to the native `Map` API.
- Look again in the future at whether to make use of native `Map`, once it has stabilised in more browsers.

gkalpak:
Closing in favor of #15483.

PR #15119, opened 2016-09-09
docs($compile): remove meaningless sentence

### LINKED ISSUE #15109: $compile: "the root of the template always gets a new scope..." What is it about?
In the [docs for $compile](https://docs.angularjs.org/api/ng/service/$compile), we can read the following:

> The scope property can be true, an object or a falsy value: ...
> true: A new child scope that prototypically inherits from its parent will be created for the directive's element. If multiple directives on the same element request a new scope, only one new scope is created. _The new scope rule does not apply for the root of the template since the root of the template always gets a new scope._

I can't understand the last sentence. Does it mean that even if you set `scope` to `false` in the DDO, your directive will create a new scope if it has a template? Then, it's not true: see [this plunk](http://plnkr.co/edit/61rLxTRbF3LgIHlZ8PDc?p=preview). Does this sentence mean something different?

thorn0:
Closes #15109

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -267,14 +267,13 @@
  * and other directives used in the directive's template will also be excluded from execution.
  *
  * #### `scope`
- * The scope property can be `true`, an object or a falsy value:
+ * The scope property can be `false`, `true`, or an object:
  *
- * * **falsy:** No scope will be created for the directive. The directive will use its parent's scope.
+ * * **`false` (default):** No scope will be created for the directive. The directive will use its parent's scope.
  *
  * * **`true`:** A new child scope that prototypically inherits from its parent will be created for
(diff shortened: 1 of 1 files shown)


src/ng/compile.js
```diff
@@ -269,12 +269,11 @@
  * #### `scope`
  * The scope property can be `true`, an object or a falsy value:
  *
- * * **falsy:** No scope will be created for the directive. The directive will use its parent's scope.
+ * * **falsy (default):** No scope will be created for the directive. The directive will use its parent's scope.
```
gkalpak: If you are adding `(default)` then change it to `false` (because `falsy` is not a concrete value :smiley:).

thorn0: done

PR #15120, opened 2016-09-10
docs($templateCache): describe your change...

stevermeister:

Docs

### DIFF
--- a/src/ng/cacheFactory.js
+++ b/src/ng/cacheFactory.js
@@ -385,15 +385,20 @@ function $CacheFactoryProvider() {
  * });
  * ```
  *
- * To retrieve the template later, simply use it in your HTML:
+ * To retrieve the template later use $templateCache.get:
+ * ```js
+ * myApp.component('myComponent', {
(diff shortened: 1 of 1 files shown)


gkalpak:
I agree that `ngInclude` is not a great example, but I don't like the new example any better.
In fact, it is much better to use `ngInclude` than manually compile and append some HTML inside your component (except for very specific usecases, which are not suitable for a beginner-friendly example).

Maybe an example showing the cached template used as a component's template would be more suitable.

And I would still leave the JS example below, because it is not a common usecase.

gkalpak:
I just found out this is related to the discussion on #14852.
I believe that what I said above is what @petebacondarwin meant in https://github.com/angular/angular.js/pull/14852#issuecomment-229886472.

E.g. the "component example" should, be something like:

```js
.component('myComponent', {
  template: 'my/cached/template.html',
  ...
})
```

stevermeister:
ok, one more try - https://github.com/angular/angular.js/pull/15153/files
fixed according to your comments
(I know that it would be better to update this one, but I've created this one with `patch`)

PR #15123, opened 2016-09-12
fix($compile): do not overwrite values set in `$onInit()` for `<`-bound literals

### LINKED ISSUE #15118: One-way bindings of arrays behave inconsistently when a default value is set in the constructor

Bug.

One-way bindings of arrays behave inconsistently when a default value is set in the constructor.

[Minimal JSFiddle example](https://jsfiddle.net/donmccurdy/ga9w15ab/).

`childCtrl.list2` is set but `childCtrl.list1` never is.  The `changes` argument to `$onChanges(changes)` contains both values, however.

Perhaps I made a terrible mistake by trying to set default values (`list1=[]` and `list2=[]`) in my child controller's constructor. Probably that's my fault, and `list1` being empty is expected. But the problem did take me longer to track down because some previous code (like `list2`) had worked just fine.

I'd have expected `list1` and `list2` to behave the same way here. If setting a default in the constructor is bad and should never be done, then maybe no change is needed.

v1.5.8, v1.5.9

Thanks!

gkalpak:
Bug fix.

See #15118.

No.

**Other information**:
Fixes #15118.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3506,18 +3506,21 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             if (optional && !attrs[attrName]) break;
 
             parentGet = $parse(attrs[attrName]);
+            var deepWatch = parentGet.literal;
 
             var initialValue = destination[scopeName] = parentGet(scope);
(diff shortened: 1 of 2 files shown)


Narretz:
LGTM, but does this mean a literal in a component binding gets double checked every time there's a change?

PR #15132, opened 2016-09-14
fix($scope): fixes multiple root element error when there is a whites…

### LINKED ISSUE #15108: Directive template with comment throws multiple root element error because jqLite thinks whitespace after comment is a node

Bug

If one tries to load a directive with a top-level comment angular throws an error 
Error: [$compile:tplrt] Template for directive 'myDirective' must have exactly one root element

This happens because jqLite thinks that the whitespace between the comment node and html node is a text node. In the `removeComments()` function before template invocation, the text template is converted into jqLite object, where the array contains three objects `[comment, text, div]`. Only the comment node gets removed and the presence of two remaining nodes throws this error. 

I guess this should be fixed in jqLite but for now I have suggested a simple fix (see below after the next question)

Using 1.5.8 in nwjs 0.16.1 on Windows 10 

Change the test in `removeComments()`
```
if (node.nodeType === NODE_TYPE_COMMENT)
```
to
```
if (node.nodeType === NODE_TYPE_COMMENT || (node.nodeType === NODE_TYPE_TEXT && node.nodeValue.trim() === '' ))
```

renovatorruler:
Bug fix

If one tries to load a directive with a top-level comment angular throws an error 
Error: [$compile:tplrt] Template for directive 'myDirective' must have exactly one root element
https://github.com/angular/angular.js/issues/15108

No

**Other information**:

fixes multiple root element error when there is a whitespace after a comment

Added new conditional for NODE_TYPE_TEXT inside removeComments method of $compile
Added corresponding unit tests.

Closes #15108

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3679,8 +3679,9 @@ function removeComments(jqNodes) {
 
   while (i--) {
     var node = jqNodes[i];
-    if (node.nodeType === NODE_TYPE_COMMENT) {
-      splice.call(jqNodes, i, 1);
+    if (node.nodeType === NODE_TYPE_COMMENT ||
+       (node.nodeType === NODE_TYPE_TEXT && node.nodeValue.trim() === '')) {
+         splice.call(jqNodes, i, 1);
     }
   }
   return jqNodes;
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -1364,6 +1364,22 @@ describe('$compile', function() {
           });
         });
 
+        it('should ignore whitespace betwee comment and root node when replacing with a template', function() {
+          module(function() {
+            directive('replaceWithWhitespace', valueFn({
+              replace: true,
+              template: '<!-- ignored comment --> <p>Hello, world!</p> <!-- ignored comment-->'
+            }));
+          });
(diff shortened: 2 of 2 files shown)


Narretz:
This issue doesn't only happen with replace directives, but with any directive that has a comment node with whitespace around it. I think these two tests should be added: 

```
            // comments are ok
            $templateCache.put('template.html', '<!-- oh hi --><div></div> \n');
            $compile('<p template></p>');
            $rootScope.$apply();
            expect($exceptionHandler.errors).toEqual([]);

            // white space around comments is ok
            $templateCache.put('template.html', '  <!-- oh hi -->  <div></div>  <!-- oh hi -->\n');
            $compile('<p template></p>');
            $rootScope.$apply();
            expect($exceptionHandler.errors).toEqual([]);
```

for the test `it('should fail if replacing and template doesn\'t have a single root element'`

prashantpawar:
I will add your tests, but @Narretz the it block you suggested has replace:true in the setup.

prashantpawar:
@Narretz Added those tests

Narretz:
@prashantpawar Sorry, you are right, it's all replace directives. It looks like these tests are duplicated in a few places.

PR #15140, opened 2016-09-15
fix(ngTransclude): use fallback content if only whitespace is provided

### LINKED ISSUE #15077: Transcluded content is not trimmed
When specifying an `ng-transclude` element, the documentation says you can [provide a fallback](https://docs.angularjs.org/api/ng/directive/ngTransclude#transclude-fallback-content) that will render if no content is transcluded. However, if there is **only** whitespace in the transclusion slot it will replace the fallback with this whitespace.

I prefer to markup my components as follows:

```
<ns:legend-item
	ng-repeat="item in $ctrl.legendItems"
	ns:color="{{ item.color }}"
	ns:label="{{ item.name }}">
</ns:legend-item>
```

Doing it this way means it always passes some whitespace as the transcluded content and my fallback is removed. To fix I have to change the markup to:

```
<ns:legend-item
	ng-repeat="item in $ctrl.legendItems"
	ns:color="{{ item.color }}"
	ns:label="{{ item.name }}"></ns:legend-item>
```

It doesn't seem right that the position of the closing tag changes how the component renders. 

I'm using Angular 1.5.5.

petebacondarwin:

bug

Fallback content is used even if only whitespace is provided

Fallback content is not used if only whitespace is provided

Yes

**Other information**:

If the transcluded content is only whitespace then we should use the
fallback content instead. This allows more flexibility in formatting
your HTML.

Closes #15077

BREAKING CHANGE:

Previously whitespace only transclusion would be treated as the transclusion
being "not empty", which meant that fallback content was not used in that
case.

Now if you only provide whitespace as the transclusion content, it will be
assumed to be empty and the fallback content will be used instead.

If you really do want whitespace then you can force it to be used by adding
a comment to the whitespace.

### DIFF
--- a/src/ng/directive/ngTransclude.js
+++ b/src/ng/directive/ngTransclude.js
@@ -13,8 +13,8 @@
  *
  * If the transcluded content is not empty (i.e. contains one or more DOM nodes, including whitespace text nodes), any existing
  * content of this element will be removed before the transcluded content is inserted.
- * If the transcluded content is empty, the existing content is left intact. This lets you provide fallback content in the case
- * that no transcluded content is provided.
+ * If the transcluded content is empty (or only whitespace), the existing content is left intact. This lets you provide fallback
+ * content in the case that no transcluded content is provided.
  *
  * @element ANY
  *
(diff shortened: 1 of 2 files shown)


Narretz:
LGTM :shipit:

PR #15141, opened 2016-09-15
refactor(compileSpec): make tests consistent

Narretz:
refactor

Tests are duplicated and have non-optimal expects

Tests use they and have been moved to one place

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -1201,41 +1201,52 @@ describe('$compile', function() {
           });
         });
 
-        it('should fail if replacing and template doesn\'t have a single root element', function() {
-          module(function() {
-            directive('noRootElem', function() {
+        describe('replace and not exactly one root element', function() {
+          var templateVar;
+
+          beforeEach(module(function() {
+            directive('template', function() {
               return {
                 replace: true,
-                template: 'dada'
+                template: function() {
+                  return templateVar;
+                }
               };
             });
-            directive('multiRootElem', function() {
-              return {
-                replace: true,
-                template: '<div></div><div></div>'
-              };
-            });
-            directive('singleRootWithWhiteSpace', function() {
-              return {
-                replace: true,
-                template: '  <div></div> \n'
-              };
-            });
+          }));
+
+          they('should throw if: $prop',
+            {
+              'no root element': 'dada',
+              'multiple root elements': '<div></div><div></div>'
+            }, function(directiveTemplate) {
+
+              inject(function($compile) {
+                templateVar = directiveTemplate;
+                expect(function() {
+                  $compile('<p template></p>');
+                }).toThrowMinErr('$compile', 'tplrt',
+                  'Template for directive \'template\' must have exactly one root element.'
+                );
+              });
           });
 
-          inject(function($compile) {
-            expect(function() {
-              $compile('<p no-root-elem></p>');
-            }).toThrowMinErr('$compile', 'tplrt', 'Template for directive \'noRootElem\' must have exactly one root element. ');
+          they('should not throw if the root element is accompanied by: $prop',
+            {
+              'whitespace': '  <div>Hello World!</div> \n',
+              'comments': '<!-- oh hi --><div>Hello World!</div> \n',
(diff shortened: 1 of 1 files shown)


test/ng/compileSpec.js
```diff
@@ -1201,41 +1201,50 @@ describe('$compile', function() {
...
 
-          inject(function($compile) {
-            expect(function() {
-              $compile('<p no-root-elem></p>');
-            }).toThrowMinErr('$compile', 'tplrt', 'Template for directive \'noRootElem\' must have exactly one root element. ');
+          they('should throw if: $prop',
+            {
+              'no root element': 'dada',
+              'multiple root elements': '<div></div><div></div>'
+            }, function($prop) {
```
gkalpak: Calling this `$prop` too is confusing (since it is not the same `$prop` as in the description).
Maybe `template` or something.

test/ng/compileSpec.js
```diff
@@ -1201,41 +1201,50 @@ describe('$compile', function() {
...
 
-            expect(function() {
-              $compile('<p multi-root-elem></p>');
-            }).toThrowMinErr('$compile', 'tplrt', 'Template for directive \'multiRootElem\' must have exactly one root element. ');
+              inject(function($compile, $templateCache, $rootScope) {
```
gkalpak: Unused depedencies?

test/ng/compileSpec.js
```diff
@@ -1201,41 +1201,50 @@ describe('$compile', function() {
...
-              $compile('<p single-root-with-white-space></p>');
-            }).not.toThrow();
+          they('should not throw if the root element is accompanied by: $prop',
+            {
+              'whitespace': '  <div>Hello World!</div> \n',
+              'comments': '<!-- oh hi --><div>Hello World!</div> \n',
+              'comments + whitespace': '  <!-- oh hi -->  <div>Hello World!</div>  <!-- oh hi -->\n'
+            }, function($prop) {
+
+              inject(function($compile, $templateCache, $rootScope) {
```
gkalpak: Unused dependency?

test/ng/compileSpec.js
```diff
@@ -1201,41 +1201,50 @@ describe('$compile', function() {
...
+                templateVar = $prop;
+                var element;
+                expect(function() {
+                  element = $compile('<p template></p>')($rootScope);
+                }).not.toThrowMinErr('$compile', 'tplrt', 'Template for directive \'template\' must have exactly one root element.');
```
gkalpak: `.not.toThrow()` is preferrable, since right now a small change in the message (or a typo) could render the test useless.

test/ng/compileSpec.js
```diff
@@ -2075,55 +2052,53 @@ describe('$compile', function() {
           }
         ));
 
+        describe('replace and not exactly one root element', function() {
```
gkalpak: Does it have to be the same description as above? What is the difference?

gkalpak (changes requested):

test/ng/compileSpec.js
Narretz: Ok, but I think we should match against the error key (tplrt), because that is unlikely to change and we want to make sure that exact error is thrown.

test/ng/compileSpec.js
Narretz: The one above is for template, this one is for templateUrl

test/ng/compileSpec.js
gkalpak: I am talking about negative assertions only (i.e. with `.not`). Unless there is a reason (e.g. there is another error thrown that we can't prevent), I prefer to have as little info as possible (in order to make the assertion as strong as possible).

For example: `.not.toThrow()` is stronger than `.not.toThrowMinErr('$compile')` or `.not.toThrowMinErr('$compile', 'tplrt')`.

Narretz: Ok, got it.

Narretz:
Can you check / approve this one @gkalpak ?

gkalpak (approved):
LGTM

PR #15142, opened 2016-09-15
docs(ngCsp): update unsafe-eval and unsafe-inline descriptions

davidcigital:
Update description of: 
- Angular conflict with CSP 
- 'unsafe-eval' and 'unsafe-inline' to point out that these are keywords that disables certain default rules (and thereby weakens the CSP). 
- auto-detection of 'unsafe-eval' not allowed

The way it was presented previously was slightly misleading as it indicated that these were rules forbidding certain things, when in fact they're keywords in the CSP that disable the very rules that were described. I hope this update clarifies this better.

### DIFF
--- a/src/ng/directive/ngCsp.js
+++ b/src/ng/directive/ngCsp.js
@@ -17,21 +17,21 @@
  *
  * The following default rules in CSP affect Angular:
  *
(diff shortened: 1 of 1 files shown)


davidcigital:
I signed it!

PR #15143, opened 2016-09-15
feat(http): JSONP requests now require trusted resource URLs

### LINKED ISSUE #11328: Inconsistent handling of $sce trustedUrl with $http

### LINKED ISSUE #11352: $http's JSONP should require trusted resource URLs

petebacondarwin:
JSONP requests allow full access to the browser and the JavaScript context. So allowing a malicious attacker to make a JSONP request to an evil server could have bad results.

By requiring that JSONP urls are trusted we make it easier for developers to see that their app is only able to make JSONP requests to urls that they have audited.

There are two ways to trust:
- whitelisting the url
- explicitly marking a url as trusted by calling `$sce.trustAs`

The first commit in this PR ensures that all JSONP requests use a URL that is validated against the `$sce` ResourceUrl whitelist/blacklist checking. This commit creates a Breaking Change for app developers as they must now add all of their JSONP endpoints to the whitelist before their app will work. Although as significant breaking change, it is fairly simple to fix: just search for all calls to `$http.jsonp` and identify the url. Of course if the url is being generated dynamically then this is harder to fix but also indicates that perhaps there is a security vulnerability.

The second commit allows developers to provide explicitly trusted urls as a result of calling `$sce.trustAsResourceUrl`. This second approach has a few quirks:
- the `$http` service needs to append parameters to the url, which means that the trusted url must be unwrapped, modified and then re-wrapped as trusted before passing it to `$httpBackend`.
- it may be possible (??) that there might be an attack vector if an attacker is able to access a different endpoint by changing the parameters. But perhaps since the domain and path cannot be modified by the param serializer this is not an issue?

_Is it acceptable (from a security point of view) to allow parameters to be added to a trusted url? If so then we could refactor the checks to happen in the `$http` service instead of `$httpBackend` which would remove the need for the re-wrapping of the built url._

Closes #11352
Closes #11328

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -379,8 +379,8 @@ function $HttpProvider() {
    **/
   var interceptorFactories = this.interceptors = [];
 
-  this.$get = ['$browser', '$httpBackend', '$$cookieReader', '$cacheFactory', '$rootScope', '$q', '$injector',
-      function($browser, $httpBackend, $$cookieReader, $cacheFactory, $rootScope, $q, $injector) {
+  this.$get = ['$browser', '$httpBackend', '$$cookieReader', '$cacheFactory', '$rootScope', '$q', '$injector', '$sce',
+      function($browser, $httpBackend, $$cookieReader, $cacheFactory, $rootScope, $q, $injector, $sce) {
 
     var defaultCache = $cacheFactory('$http');
 
@@ -954,7 +954,7 @@ function $HttpProvider() {
         throw minErr('$http')('badreq', 'Http request configuration must be an object.  Received: {0}', requestConfig);
       }
 
-      if (!isString(requestConfig.url)) {
+      if (!isString($sce.valueOf(requestConfig.url))) {
         throw minErr('$http')('badreq', 'Http request configuration url must be a string.  Received: {0}', requestConfig.url);
       }
 
@@ -1255,11 +1255,13 @@ function $HttpProvider() {
           cache,
           cachedResp,
           reqHeaders = config.headers,
-          url = buildUrl(config.url, config.paramSerializer(config.params));
+          needsTrustedUrl = (lowercase(config.method) === 'jsonp'),
+          url = needsTrustedUrl ? $sce.getTrustedResourceUrl(config.url) : config.url;
+
+      url = buildUrl(url, config.paramSerializer(config.params));
 
       $http.pendingRequests.push(config);
       promise.then(removePendingReq, removePendingReq);
-
 
       if ((config.cache || defaults.cache) && config.cache !== false &&
           (config.method === 'GET' || config.method === 'JSONP')) {
@@ -1299,7 +1301,7 @@ function $HttpProvider() {
           reqHeaders[(config.xsrfHeaderName || defaults.xsrfHeaderName)] = xsrfValue;
         }
 
-        $httpBackend(config.method, url, reqData, done, reqHeaders, config.timeout,
+        $httpBackend(config.method, needsTrustedUrl ? $sce.trustAsResourceUrl(url) : url, reqData, done, reqHeaders, config.timeout,
             config.withCredentials, config.responseType,
             createApplyHandlers(config.eventHandlers),
             createApplyHandlers(config.uploadEventHandlers));
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -881,6 +881,12 @@ function $HttpProvider() {
 </file>
 <file name="script.js">
   angular.module('httpExample', [])
+    .config(['$sceDelegateProvider', function($sceDelegateProvider) {
+      $sceDelegateProvider.resourceUrlWhitelist([
+        'self',
+        'https://angularjs.org/**'
+      ]);
+    }])
     .controller('FetchController', ['$scope', '$http', '$templateCache',
       function($scope, $http, $templateCache) {
         $scope.method = 'GET';
--- a/src/ng/httpBackend.js
+++ b/src/ng/httpBackend.js
@@ -49,17 +49,19 @@ function $xhrFactoryProvider() {
  * $httpBackend} which can be trained with responses.
  */
 function $HttpBackendProvider() {
-  this.$get = ['$browser', '$jsonpCallbacks', '$document', '$xhrFactory', function($browser, $jsonpCallbacks, $document, $xhrFactory) {
-    return createHttpBackend($browser, $xhrFactory, $browser.defer, $jsonpCallbacks, $document[0]);
+  this.$get = ['$sce', '$browser', '$jsonpCallbacks', '$document', '$xhrFactory', function($sce, $browser, $jsonpCallbacks, $document, $xhrFactory) {
+    return createHttpBackend($sce, $browser, $xhrFactory, $browser.defer, $jsonpCallbacks, $document[0]);
   }];
 }
 
-function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDocument) {
+function createHttpBackend($sce, $browser, createXhr, $browserDefer, callbacks, rawDocument) {
   // TODO(vojta): fix the signature
   return function(method, url, post, callback, headers, timeout, withCredentials, responseType, eventHandlers, uploadEventHandlers) {
-    url = url || $browser.url();
 
     if (lowercase(method) === 'jsonp') {
+      // This is a pretty sensitive operation where we're allowing a script to have full access to
+      // our DOM and JS space.  So we require that the URL satisfies SCE.RESOURCE_URL.
+      url = $sce.getTrustedResourceUrl(url) || $browser.url();
       var callbackPath = callbacks.createCallback(url);
       var jsonpDone = jsonpReq(url, callbackPath, function(status, text) {
         // jsonpReq only ever sets status to 200 (OK), 404 (ERROR) or -1 (WAITING)
@@ -69,6 +71,7 @@ function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDoc
       });
     } else {
 
+      url = url || $browser.url();
       var xhr = createXhr(method, url);
 
       xhr.open(method, url, true);
--- a/test/ng/httpBackendSpec.js
+++ b/test/ng/httpBackendSpec.js
@@ -3,11 +3,16 @@
 
 describe('$httpBackend', function() {
 
-  var $backend, $browser, $jsonpCallbacks,
+  var $sce, $backend, $browser, $jsonpCallbacks,
       xhr, fakeDocument, callback;
 
-  beforeEach(inject(function($injector) {
+  beforeEach(module(function($sceDelegateProvider) {
+    // Setup a special whitelisted url that we can use in testing JSONP requests
+    $sceDelegateProvider.resourceUrlWhitelist(['http://special.whitelisted.resource.com/**']);
+  }));
 
+  beforeEach(inject(function($injector) {
+    $sce = $injector.get('$sce');
     $browser = $injector.get('$browser');
 
     fakeDocument = {
@@ -48,7 +53,7 @@ describe('$httpBackend', function() {
       }
     };
 
-    $backend = createHttpBackend($browser, createMockXhr, $browser.defer, $jsonpCallbacks, fakeDocument);
+    $backend = createHttpBackend($sce, $browser, createMockXhr, $browser.defer, $jsonpCallbacks, fakeDocument);
     callback = jasmine.createSpy('done');
   }));
 
@@ -273,7 +278,7 @@ describe('$httpBackend', function() {
 
   it('should call $xhrFactory with method and url', function() {
     var mockXhrFactory = jasmine.createSpy('mockXhrFactory').and.callFake(createMockXhr);
-    $backend = createHttpBackend($browser, mockXhrFactory, $browser.defer, $jsonpCallbacks, fakeDocument);
+    $backend = createHttpBackend($sce, $browser, mockXhrFactory, $browser.defer, $jsonpCallbacks, fakeDocument);
     $backend('GET', '/some-url', 'some-data', noop);
     expect(mockXhrFactory).toHaveBeenCalledWith('GET', '/some-url');
   });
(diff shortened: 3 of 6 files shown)


src/ng/http.js
```diff
@@ -1293,7 +1297,7 @@ function $HttpProvider() {
           reqHeaders[(config.xsrfHeaderName || defaults.xsrfHeaderName)] = xsrfValue;
         }
 
-        $httpBackend(config.method, url, reqData, done, reqHeaders, config.timeout,
+        $httpBackend(config.method, (url === $sce.valueOf(origUrl) ? origUrl : url), reqData, done, reqHeaders, config.timeout,
```
petebacondarwin: So this line is weird since we it is saying: "if the newly built url is not the same as the original one then just use the original one. But this defeats the purpose of being able to provide parameters, etc.
But if we pass the untrusted URL through it should not be allowed by SCE inside the httpBackend, which now expects trusted URLs for JSONP.
And if we "retrust" the built URL for JSONP requests we are undermining the security unless we can ensure that the `buildUrl` function cannot modify the URL in a way that allows an attack.

Given all of this, it might be easier to say that `$http` can only take strings as URLs, which is what is suggested by @klhuillier in https://github.com/angular/angular.js/issues/11328#issuecomment-81778364. That way the only way to get JSONP requests to work is if the request URL is in the SCE resource URL whitelist.

Narretz: So the `url` parameter for $http only takes strings. Is `toString()` meant when talking about non strings?

src/ng/httpBackend.js
```diff
@@ -49,14 +49,19 @@ function $xhrFactoryProvider() {
...
 
-function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDocument) {
+function createHttpBackend($sce, $browser, createXhr, $browserDefer, callbacks, rawDocument) {
   // TODO(vojta): fix the signature
   return function(method, url, post, callback, headers, timeout, withCredentials, responseType, eventHandlers, uploadEventHandlers) {
+    if (lowercase(method) == 'jsonp') {
+      // This is a pretty sensitive operation where we're allowing a script to have full access to
+      // our DOM and JS space.  So we require that the URL satisfies SCE.RESOURCE_URL.
+      url = $sce.getTrustedResourceUrl(url);
+    }
```
gkalpak: I would move this block below the next line (inside the other `jsonp` block).

src/ng/http.js
```diff
@@ -1249,7 +1249,11 @@ function $HttpProvider() {
           cache,
           cachedResp,
           reqHeaders = config.headers,
-          url = buildUrl(config.url, config.paramSerializer(config.params));
+          // origUrl could be a $sce trusted URL which used a wrapper class
+          origUrl = config.url,
+          url = $sce.valueOf(origUrl);
```
gkalpak: Isn't there a check near the top of `$http` that will fail if `requestConfig.url` is not a string?
That check also needs to be tweaked to accept `$sce`-wrapped values.

(I haven't looked at the failed tests, but I hope some failed because of this :smiley:)

src/ng/http.js
```diff
@@ -1249,7 +1249,11 @@ function $HttpProvider() {
...
+
+      url = buildUrl(url, config.paramSerializer(config.params));
```
gkalpak: I think it is simpler if all changes inside `$http` are done inside `buildUrl`. Something like this:

```diff
 function buildUrl(url, serializedParams) {
   if (serializedParams.length > 0) {
+    url = $sce.valueOf(url); 
     url += ((url.indexOf('?') === -1) ? '?' : '&') + serializedParams;
   }
   return url;
 }
```

src/ng/http.js
gkalpak: > So this line is weird since it is saying: "if the newly built url is not the same as the original one then just use the original one." But using the original one defeats the purpose of being able to provide parameters, etc.

Not quite :smiley: It says: "If the newly built URL _is_ the sameas the original one (possibly unwrapped) - in other words, if there were no parameters to be added - then use the original one (which is just a possibly wrapped version of the newly built one)."

It is true that if you want to pass parameters for JSONP requests, then an $sce-trusted URL won't help you. The resulting URL (inclusing the parameters) must be approved based on the white-/blacklist for resource URLs.
It is not an ideal solution, but I don't think we can ensure that a URL that is safe before `buildUrl()` will be safe after. 

In any case, this behavior should be documented (if we decide to go with this implementation).

test/ng/httpBackendSpec.js
```diff
@@ -405,6 +413,18 @@ describe('$httpBackend', function() {
...
+    it('should throw error if the url is not a trusted resource', function() {
+      expect(function() {
+        $backend('JSONP', 'http://example.org/path?cb=JSON_CALLBACK', null, callback);
+      }).toThrowMinErr('$sce', 'insecurl');
+    });
+
+    it('should not throw error if the url is an explicitly trusted resource', function() {
+      expect(function() {
+        $backend('JSONP', $sce.trustAsResourceUrl('http://example.org/path?cb=JSON_CALLBACK'), null, callback);
+      }).not.toThrowMinErr('$sce', 'insecurl');
```
gkalpak: I think generic `.not.toThrow()` expectations are preferrable, because a slight change in the error name (or a typo) render the test useless.

gkalpak (changes requested):

src/ng/http.js
petebacondarwin: There is! I just found it. All the tests in this PR are only testing the `$httpBackend` so none of the changes in this file are tested.

I am working on improving that :-)

src/ng/http.js
petebacondarwin: I agree

src/ng/http.js
petebacondarwin: If we must rely on the whitelist for parameterised requests then there is no benefit in allowing explicitly trusted urls for those requests at all.

The trouble is that many JSONP requests are going to be parameterised, right? So the majority of requests will not benefit from being able to explicitly trust their url.

So perhaps we should only allow string urls (i.e. you cannot explicitly trust them), which is a massive breaking change for JSONP requests.

src/ng/httpBackend.js
petebacondarwin: OK!

test/ng/httpBackendSpec.js
petebacondarwin: OK

petebacondarwin (commented):
Thanks @gkalpak

src/ng/httpBackend.js
```diff
@@ -49,17 +49,19 @@ function $xhrFactoryProvider() {
...
-function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDocument) {
+function createHttpBackend($sce, $browser, createXhr, $browserDefer, callbacks, rawDocument) {
   // TODO(vojta): fix the signature
   return function(method, url, post, callback, headers, timeout, withCredentials, responseType, eventHandlers, uploadEventHandlers) {
-    url = url || $browser.url();
 
     if (lowercase(method) === 'jsonp') {
+      // This is a pretty sensitive operation where we're allowing a script to have full access to
+      // our DOM and JS space.  So we require that the URL satisfies SCE.RESOURCE_URL.
+      url = $sce.getTrustedResourceUrl(url) || $browser.url();
```
gkalpak: So, we are basically only allowing JSONP requests to whitelisted URLs (no support for trusting a URL "on-the-fly").
I am fine with this, except that we should document it :smiley:

src/ng/httpBackend.js
```diff
@@ -49,17 +49,19 @@ function $xhrFactoryProvider() {
...
+  this.$get = ['$sce', '$browser', '$jsonpCallbacks', '$document', '$xhrFactory', function($sce, $browser, $jsonpCallbacks, $document, $xhrFactory) {
+    return createHttpBackend($sce, $browser, $xhrFactory, $browser.defer, $jsonpCallbacks, $document[0]);
   }];
 }
 
-function createHttpBackend($browser, createXhr, $browserDefer, callbacks, rawDocument) {
+function createHttpBackend($sce, $browser, createXhr, $browserDefer, callbacks, rawDocument) {
   // TODO(vojta): fix the signature
   return function(method, url, post, callback, headers, timeout, withCredentials, responseType, eventHandlers, uploadEventHandlers) {
-    url = url || $browser.url();
```
gkalpak: Why this change? Why are we implicitly allowing JSONP requests to the current URL?
I don't expect this to be a security concern, but (a) it is most probably a programming error and (b) we are duplicating the `|| browser.url()` part unnecessarily. Did I miss something?

gkalpak (changes requested):

src/ng/httpBackend.js
petebacondarwin: Right - documentation ....

src/ng/httpBackend.js
petebacondarwin: I originally moved the trust check inside the second `if` clause as you suggested but it broke stuff as we need to convert a `null` url to the current browser url after the trust check. So now I basically moved this line to lines 64 and 74 below.

gkalpak: Couldn't we leave this `url = url || browser.url()` here and just add `url = $sce.getTrustedResourceUrl(url)` in the JSONP case.

src/ng/http.js
gkalpak: > The trouble is that many JSONP requests are going to be parameterised, right?

There are two types of "parameterized". If the user trusts the URL (including the parameters - e.g. `$http.jsonp($sce.trustAsResourceUrl('url?with=params', ...)`), then it could be fine.
What they can't do, is pass the parameters as `config.params` (e.g. `$http.jsonp('url', {params: {...}})`).

--
> So perhaps we should only allow string urls (i.e. you cannot explicitly trust them), which is a massive breaking change for JSONP requests.

This particular thing is not a breaking change, since we are already restricting to string URLs, right?

src/ng/httpBackend.js
petebacondarwin: The `url = url || ... ` has to go after the `getTrustedResource` so you are back to earlier when you suggested that we move the check inside the if statement.

Anyway, I am hoping that we can accept that the param changes are not a security issue, in which case we can refactor the whole thing out and move the checks out from the `$httpBackend`, which means that we don't need to do the re-wrapping etc

mprobst:
So if I understand this correctly, the question is how to vet the safety of a JSONP URL, and safe construction when appending parameters to it.

As far as I understand, Angular 1 always appends query parameters with ?=..., so as long as there is any original safe URL, it'll be safe to append them for the local page context (AFAIK). If your attack scenario is just whether the `<script src=...>` URL is safe, I think that'd be sufficient.

However many JSONP endpoints take a function name to call with the result as a parameter (e.g. `?callback=myFunction`). If an attacker can inject a parameter that overrides this callback parameter, they can probably inject code into the page (e.g. `?callback=myFunction&arg1=val1&callback=evilCode`).

To defend against that, I think indeed you'll have to get a trusted value for both the base URL and all of the query parameter names and their values. That might turn the API into something unusable though, `$http.jsonp($sce.trustAsResourceUrl('/myurl'), {params: {a: $sce.trustAsResourceUrl(aval), b: $sce.trustAsResourceUrl(bval)}});` doesn't exactly roll off the tongue. And you'd still risk attackers controlling the parameter names.

I'm not sure if this is worth securing?

/CC @rjamet

src/ng/http.js
```diff
@@ -1249,12 +1255,14 @@ function $HttpProvider() {
           cache,
           cachedResp,
           reqHeaders = config.headers,
-          url = buildUrl(config.url, config.paramSerializer(config.params));
+          needsTrustedUrl = (lowercase(config.method) === 'jsonp'),
```
rjamet: needsTrustedUrl is a little misleading, since you have a $sce.trustAsUrl function for another context :/ needsTrustedResourceUrl ?

petebacondarwin: Right thanks!

petebacondarwin:
@mprobst are you questioning whether we should attempt to secure the params or whether it is worth trying to secure JSONP at all in this way?

rjamet:
For the `?callback=myFunction&arg1=val1&callback=evilCode` case, maybe removing the rightmost duplicate parameters after appending them would be enough?

I'm not fond of having to trust parameters either: this doesn't really make sense with the ResourceUrl semantics, a param value isn't a resource url, and getTrustedResourceUrl with its same-origin whitelist will see everything as relative URLs :/ It just sounds bad. I'm not sure of the right approach, though. Being permissive and building the request on top of a trusted URL from strings isn't great, but it might be the best option.

mprobst:
@petebacondarwin I think the reasonable thing to do is requiring the base URL to be a trusted resource URL, and then just accepting the fact that there's a risk factor in the callback keys and values. It's better than nothing, and at least I cannot come up with something more appealing.

I think @rjamet is of the same opinion, if I understand him correctly.

src/ng/http.js
```diff
@@ -1293,7 +1301,7 @@ function $HttpProvider() {
           reqHeaders[(config.xsrfHeaderName || defaults.xsrfHeaderName)] = xsrfValue;
         }
 
-        $httpBackend(config.method, url, reqData, done, reqHeaders, config.timeout,
+        $httpBackend(config.method, needsTrustedUrl ? $sce.trustAsResourceUrl(url) : url, reqData, done, reqHeaders, config.timeout,
```
rjamet: Slight nit, this is the first time AngularJS calls a trustAs function in the core library. That's not wrong, but it's a weird pattern: how about a parameter that will disable the check below instead? 

(selfish reason for that: I maintain a custom $sce that uses the Closure's implementation of safe types and disables trustAs calls, and this is going to need ad-hoc patching :/)

rjamet:
@mprobst @petebacondarwin There's an alternativ : since the jsonp callback is a parameter, could these be $sce.getTrustedJs / trustAsJs ? You have the plumbing already in place. This type hasn't been used before, but since parameters there might be run as-is... But this seems like yet another hurdle that any dev wanting to use jsonp will jump through without thinking :/ WDYT ?

src/ng/http.js
petebacondarwin: If we are going with not securing the params then I can do a tidy refactor on this PR that removes the need for this `trustAs` call

petebacondarwin:
Closing in favour of #16161 - PTAL @mprobst @rjamet

Issue #15144, opened 2016-09-15
Consider base URL when deciding if a URL is allowed by 'self' URL policy


adob:
Angular SCE currently rejects perfectly valid relative URLs in some situations when a <base> tag is used. For example, if the page contains `<base href="http://www.example.com">`, Angular will start to reject relative URLs like 'foo.template.html'. Full example at https://plnkr.co/edit/3sFwq1d2d7wiWVBx4OHn

The root cause is that the urlIsSameOrigin() function in urlUtils.js matches against location.href instead of, or in addition to, document.baseURI.

Angular should permit URLs if protocol/host/port matches the base URL being used.

Affects AngularJS v1.5.8.

gkalpak:
Sounds reasonable, although I am not aware of any security implications.

/cc @rjamet, @mprobst

rjamet:
Sounds good to me too, a base href is a sign that the destination is trusted, that's in line with the same-origin whitelist.

(Small caveat, base href doesn't seem to require a RESOURCE_URL to be set in bindings, and I'm not entirely sure whether it affects things that can run scripts, like script src. If it does affect these, it definitely should be a RESOURCE_URL context. I'll write a PR.)

gkalpak:
Maybe it is indeed a good idea to also sanitize `href` for `<base>` elements, e.g. [here][1].

[1]: https://github.com/angular/angular.js/blob/cf241c425b20569e4c7909bfbd379f5c07694288/src/ng/compile.js#L1662-L1665

rjamet:
That's implicit in the resource url contract: the default whitelist won't allow different protocols than the current page, and that's also why a resource_url is valid in URL context. 

I don't really know how weird things like base href="data:text/html" behave (at least this hasn't executed javascript: URLs in a while), but you'd have to whitelist them first for it to work. I'll run some tests before the end of the week and write a PR at least for the context thing.

PR #15145, opened 2016-09-15
fix($sce): Consider document base URL for 'self' URL policy.

adob:
This PR fixes the issue in bug #15144. If the page specifies a base URL using `<base>` tag, Angular may start to reject relative resource URL if the base URL domain does not match the loading domain. This PR adjusts matchUrl() to also consider the base URL (`document.baseURI`) in addition to `location.href` when matching URLs against the 'self' URl policy.

This PR does not introduce any breaking changes.

Security note: SCE policy can be circumvented by rogue `<base>` tags on the page. However, I believe this is not too much of a concern. If an attacker can inject `<base>` tags, then it's very likely they can inject arbitrary HTML, which is a way worse vulnerability.


src/ng/urlUtils.js
```diff
@@ -94,3 +95,22 @@ function urlIsSameOrigin(requestUrl) {
...
+ * Parse a request URL and determine whether it is same-origin as the document base URL.
+ *
+ * Note: The base URL is usually the same the document location (`location.href`) but can
+ * overriden by using the `<base>` tag.
+ *
+ * @param {string|object} requestUrl The url of the request as a string that will be resolved
+ * or a parsed URL object.
+ * @returns {boolean} Whether the URL is same-origin as the document base URL.
+ */
+function urlIsSameOriginAsBaseUrl(requestUrl) {
```
gkalpak: Nit: I would rather not duplicate the code. Maybe have a base function that `urlIsSameOrigin`/`urlIsSameOriginAsBase` can delegate two (and pass the `originUrl`/`baseUrl` as parameters).

src/ng/urlUtils.js
```diff
@@ -94,3 +95,22 @@ function urlIsSameOrigin(requestUrl) {
...
+  if (!baseUrl) {
```
gkalpak: What if `base.href` changes dynamically ? I think we need to account for that too.

test/e2e/tests/base_tag.spec.js
```diff
@@ -0,0 +1,17 @@
...
+    var url = browser.getLocationAbsUrl();
+    var urlIsTrusted = browser.executeScript('return isTrustedUrl(arguments[0])', url);
+    expect(urlIsTrusted).toBe(true);  // sanity check
+
+    var urlIsTrusted = browser.executeScript('return isTrustedUrl("//evil.com/")');
+    expect(urlIsTrusted).toBe(false);  // sanity check
+
+    urlIsTrusted = browser.executeScript('return isTrustedUrl("/relative")');
+    expect(urlIsTrusted).toBe(true);
+  });
```
gkalpak: Let's add a test for absolute URL that matches `base.href`.

gkalpak (changes requested):
Looks good in general. I have left a couple of minor (or not so minor) comments.

I think we need to account for `base.href` chaning dynamically.
And there should probably be some unit tests too.

src/ng/urlUtils.js
adob: Done.

src/ng/urlUtils.js
adob: That is a good point. We can do this relatively painlessly by working directly with the anchor tag, since it keeps in sync with the base URL.

test/e2e/tests/base_tag.spec.js
adob: Done.

adob:
Please take another look. I refactored the code to use the the anchor tag more directly. Since it's always in always in sync with the base URL, the code now keeps up with updates to the base tag.

adob:
It turned out the code was not working due to a bug in Firefox so I had to refactor again. When the base changed, the .href property on the anchor tag would get updated, but .host would retain its old value. I think the current implementation is very reasonable, though.

adob:
Hey gkalpak, can you take another look? Let me know if there are any outstanding issues.

src/ng/urlUtils.js
```diff
@@ -91,6 +91,59 @@ function urlResolve(url) {
...
   var parsed = (isString(requestUrl)) ? urlResolve(requestUrl) : requestUrl;
-  return (parsed.protocol === originUrl.protocol &&
-          parsed.host === originUrl.host);
+  return urlsAreSameOrigin(parsed, originUrl);
+}
+
+/**
+ * Parse a request URL and determine whether it is same-origin as the current document base URL.
+ *
+ * Note: The base URL is usually the same the document location (`location.href`) but can
```
gkalpak: _can_  -->  _can be_

src/ng/urlUtils.js
```diff
@@ -91,6 +91,59 @@ function urlResolve(url) {
...
+ * overriden by using the `<base>` tag.
+ *
+ * @param {string|object} requestUrl The url of the request as a string that will be resolved
+ * or a parsed URL object.
+ * @returns {boolean} Whether the URL is same-origin as the document base URL.
+ */
+function urlIsSameOriginAsBaseUrl(requestUrl) {
+  var parsed = (isString(requestUrl)) ? urlResolve(requestUrl) : requestUrl;
```
gkalpak: Why not move this line into `urlsAreSameOrigin` as well?

src/ng/urlUtils.js
```diff
@@ -91,6 +91,59 @@ function urlResolve(url) {
...
+ * Determines if two URLs share the same origin.
+ *
+ * @param {object} url1 First URL to compare. Must be a normalized URL in the form of a
+ *     dictionary object returned by `urlResolve()`.
+ * @param {object} url2 Second URL to compare. Must be a normalized URL in the form of a
+ *     dictionary object returned by `urlResolve()`.
+ * @return {boolean} True if both URLs have the same origin, and false otherwise.
+ */
+function urlsAreSameOrigin(url1, url2) {
+  // IE sometimes includes a port in the 'host' property, even if it is the default 80 port so
```
gkalpak: Why is that a problem? Both URLs should be consistent (either contain the port or don't).
Under what circumstances can the URLs differ in that regard?

test/e2e/tests/base_tag.spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+  it('allows absolute URLs from the base origin', function() {
+    checkUrl('http://www.example.com/path/to/file.html', true);
+  });
+
+  it('tracks changes to the base URL', function() {
+    browser.executeScript(
+        'document.getElementsByTagName("base")[0].href = "http://xxx.example.com/";');
+    checkUrl('http://xxx.example.com/path/to/file.html', true);
```
gkalpak: Can you also add `checkUrl('http://www.example.com/path/to/file.html', false);` (just in case :smiley:)?

test/e2e/tests/base_tag.spec.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+describe('SCE URL policy when base tags are present', function() {
+  function checkUrl(url, allowed) {
+    var urlIsTrusted = browser.executeScript('return isTrustedUrl(arguments[0])', url);
+    expect(urlIsTrusted).toBe(allowed);
+  }
+
+  beforeAll(function() {
+    loadFixture('base_tag');
```
gkalpak: Can you change all names to kebab-case (to be consistent with the rest of the tests)?

gkalpak (changes requested):

test/e2e/tests/base_tag.spec.js
chainoy: Fix

adob:
closed/re-opened pull request to get Travis CI to rebuild. Looks like there was some error connection to SauceLabs

gkalpak:
Thx for the changes, @adob!
I just realized there are no unit tests (e.g. [here][1] and [there][2]).

Other than that LGTM 👍 

[1]: https://github.com/adob/angular.js/blob/df5fdb91b6fc8b9bbe145d2713dbdea902db37b9/test/ng/urlUtilsSpec.js
[2]: https://github.com/adob/angular.js/blob/df5fdb91b6fc8b9bbe145d2713dbdea902db37b9/test/ng/sceSpecs.js

adob:
I added some tests for urlIsSameOriginAsBaseUrl() in test/ng/urlUtilsSpec.js. I'm not quite sure how to test non-default base URL behavior in the unit test. We generally need to change the global document base URL to fully test things, but it seems sub-optimal to do so from a unit test.

I originally put tests in an end-to-end test where we have more control over the page to avoid this kind of issue. So I think all the the tests are there, just kind of spread out.

adob:
gkalpak, can you advise on how to test inside the unit tests? Or just LGTM if you think it's okay as is :-)

gkalpak:
@adob, I was planning to look into it, but I haven't found the time yet. We are focusing on getting the 1.6.0 release out, so we prioritize related tasks and issues/PRs. Once 1.6.0 is out, I will come back to this. 

<sub>(I am not even sure if it possible to add these tests in a reasonable way. I'll have to look into it. Feel free to leave the PR in its current state and I'll take it from there.)</sub>

Thx for all the work so far btw 👍

adob:
OK. Thank you

adob:
gkalpak, can you take a look at this? Is there anything I should add?

gkalpak:
I would still like to see a test in `sceSpec.js`. After, leaving it up to @rjamet to give the final LGTM.

adob:
Added unit tests to test/ng/sceSpecs.js. Please take a look.

src/ng/urlUtils.js
gkalpak: _the same the document location_ --> _the same as ..._

test/ng/sceSpecs.js
```diff
@@ -464,6 +464,38 @@ describe('SCE', function() {
...
+        var baseElem = window.document.createElement('BASE');
+        var cfg = {whitelist: ['self'], blacklist: []};
+        baseElem.setAttribute('href', window.location.protocol + '//foo.example.com/path/');
+        beforeAll(function() {
+          window.document.head.appendChild(baseElem);
+        });
+        afterAll(function() {
+          window.document.head.removeChild(baseElem);
+        });
+        function expectAllowed($sce, url) {
```
gkalpak: This helper (and `expectBlocked` below) seem redundant in this case. Let's remove them and inline the code.

test/ng/sceSpecs.js
```diff
@@ -464,6 +464,38 @@ describe('SCE', function() {
             '$sce', 'insecurl', 'Blocked loading resource from url not allowed by $sceDelegate policy.  URL: foo');
         }
       ));
+
+      describe('when the document base URL has changed', function() {
+        var baseElem = window.document.createElement('BASE');
+        var cfg = {whitelist: ['self'], blacklist: []};
+        baseElem.setAttribute('href', window.location.protocol + '//foo.example.com/path/');
+        beforeAll(function() {
```
gkalpak: If there is no particular reason, I would prefer re-creating/adding/removing the `<base>` element for **each** test.

gkalpak (changes requested):

src/ng/urlUtils.js
adob: Done

test/ng/sceSpecs.js
adob: Done.

test/ng/sceSpecs.js
adob: Done.

gkalpak (approved):
LGTM

gkalpak:
cc @rjamet (to give the final LGTM as per discussion in #15537)

rjamet:
LGTM too !

gkalpak:
FYI:
Reverted on v1.5.x (f73a651) and v1.6.x (f418ffd), while investigating security implications of this without #15597 (which is possibly a breaking change, thus not suitable for these branches).

rjamet:
(cc @xtofian and @adob for more pairs of eyes: bindings on base href don't have a context currently in AngularJS. This PR adds baseURI's origin to the URLs allowed by the 'self' whitelist for RESOURCE_URLs, and #15597 bumps the context of base href bindings to RESOURCE_URL. Do you think this PR is okay even without the added context on base href, or did I miss anything?)

I assumed that was okay, but it's more complex than I thought, you're right to rollback :/ sorry I didn't figure that out before. 

As far as i can tell, there's only three relevant cases: an app on `a.com` has a static base href to `b.com` (1), a dev-controlled binding on base href to `b.com` (2), or a user-controlled one (3). That's assuming the $sce RESOURCE_URL whitelist contains 'self'.

1. This PR makes it explicit that the other origin is trusted, so any relative link will work without $sce.trustAsResourceUrl (which is the intuitive meaning of 'self' IMO). If the dev added checks before on bindings when doing these conversions, they're still valid, same for  `$sce` blacklist entries. The second PR doesn't change anything. I think that's @adob's use case.

2. Same general idea, except the $sce requirement in the second PR will signify to the dev that this is something security-sensitive (those are the breaking cases). But fundamentally, trust on b.com is already fairly established: before the first PR, none of the RESOURCE_URL relative links work without explicit security whitelisting or bypasses. 

3. The situation is already dire for the app: the attacker has partial control over any relative link on the page, $sce or not: `<script src="./whatever">` in templates is enough to have at least a weird race or maybe XSS. Since there's no binding on that, there'll be no $sce involved on that script, so this PR does't change that. It only makes the $sce accept them in case of bindings, but IMO that's not worse than before, and the second PR solves the whole mess assuming the whitelist/trustAs is used correctly.

The only caveat I can think of there is if the app lives on `a.com`, baseURI is `b.com/something/`, with a whitelist containing `'self'` and `'b.com/something/**'`, then the whole `b.com` origin would become trusted after this. That also happens currently with apps living on `a.com/myapp/...` who trust the whole `a.com` origin by default, so it's arguably WAI. Finally, that's only after bindings happen: attacker control over a RESOURCE_URL binding is bad news in the first place regardless of baseURI issues.

On the plus side, that will remove the need for a lot of trustAsResourceUrl calls. They're seen as the only way to solve `insecurl` errors, and used in stuff like `iframe src="{{variable | noCheckTrustAsFilter}}"`, which bypasses whitelists entirely. I'd better have same-origin checks than none at all.

I lean towards saying it's fine. The more concerning stuff with this PR without the other only happens when binding on base href, which is a bad idea in the first place. Otherwise, I don't see regressions, only quirkish behavior in unusual cases.

rjamet:
Quick tests for the order of things when setting base href in various way: https://plnkr.co/edit/R54ihAo4q9gIlV6QPlye?p=preview for base href with no interpolation
https://plnkr.co/edit/kSIixu57Xd4Nxk1IP29L?p=preview for base href with interpolation

More quirks :/

gkalpak:
I can think of some very extreme scenarios, like:

1. An app uses the `self` policy, has interpolation in `base[href]` and a vulnerability that allows a malicious user to take over `base[href]`.
  (This is already disastrous from a security perspective, but let's see how Angular affects security in such a situation.)
2. The app is supposed to use absolute URLs in `RESOURCE_URL`-sensitive contexts.
3. Due to a typo, a relative URL is used (which essentially allows the malicious user to control the destination).

- **Before both PRs:** The app is "safe", since the destination URL is not trusted as `RESOURCE_URL`.
- **After both PRs:** The app is "safe" (or at least the devs have the tools to properly audit it), because `base[href]` is `RESOURCE_URL` context.
- **Without the 2nd PR:** Angular will happily trust the `base[href]`-based URL, without requiring a `RESOURCE_URL` context for it. (AFAIK, this i what SCE is trying to prevent.)

I understand this is a very contrived example, but still :smiley:

##
So, here is where we are currently:

- **v1.5.x** does not contain either PR. (This is likely to stay this way, since v1.6.x is the new stable version and v1.5.x is not actively maintained.)
- **v1..6.x** does not contain either PR. (We could add features, but we should avoid breaking changes, since v1.6.x is the current stable version.)
- **master** (which will become v1.7+) contains the first PR and will also contain #15597 (once merged). (This is OK, because we can make breaking changes on master.)

I think it is safest to keep things that way (and not merge this PR into v1.6.x without #15597).
Unless @rjamet strongly believes this PR is safe on its own.... :smiley:

rjamet:
I agree with your analysis. We could argue on whether these PRs make the app less safe than it was before, but overall, I'm siding with you: both PRs at the same time would be the safest.

PR #15149, opened 2016-09-16
fix($compile): set attribute value even if `ngAttr*` contains no interpolation

### LINKED ISSUE #15133: $interpolate 'allOrNothing' ignores empty case
bug

`$interpolate` service when running with `allOrNothing`, does not pass the value if there's nothing to interpolate (this affects ngAtrr*, which does not write the corresponding attribute when it has a simple value)

http://plnkr.co/edit/1QhnA9xb5BgyzWHpeL7O?p=preview

should see the poster image on video... 

Although there is no real need for using ngAttr/ngSrc/ngHref without interpolation, it's sometimes useful to do so when exploring initial html template or learning angular. Having to use `ng-attr-*="{{'value'}}"` instead of `ng-attr-*="value"` is not intuitive.

angular 1.5.8
win 7
chrome 53, IE 11, firefox 47

following https://github.com/angular/angular.js/issues/15130#issuecomment-246769914

gkalpak:
Bug fix

Previoulsy, when the value of an `ngAttrXyz` attribute did not contain any interpolation, then the `xyz` attribute was never set.
See #15133.

No.

**Other information**:
BTW, this commit adds a negligible overhead (since we have to set up a one-time watcher for
example), but it is justifiable for someone that is using `ngAttrXyz` (instead of `xyz` directly).

Fixes #15133.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3251,15 +3251,15 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
     }
 
 
-    function addAttrInterpolateDirective(node, directives, value, name, allOrNothing) {
+    function addAttrInterpolateDirective(node, directives, value, name, isNgAttr) {
       var trustedContext = getTrustedContext(node, name);
-      allOrNothing = ALL_OR_NOTHING_ATTRS[name] || allOrNothing;
+      var mustHaveExpression = !isNgAttr;
+      var allOrNothing = ALL_OR_NOTHING_ATTRS[name] || isNgAttr;
 
-      var interpolateFn = $interpolate(value, true, trustedContext, allOrNothing);
+      var interpolateFn = $interpolate(value, mustHaveExpression, trustedContext, allOrNothing);
 
       // no interpolation found -> ignore
       if (!interpolateFn) return;
-
 
       if (name === 'multiple' && nodeName_(node) === 'select') {
         throw $compileMinErr('selmulti',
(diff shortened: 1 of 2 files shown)


Narretz:
LGTM,
although I can't really see that with this change there's any possibility of adding an attribute interpolate directive that does not have an embedded expression. If you are using interpolation markers, then everything inside is treated as an expression, and only if you use interpolation markers a interpolation directive is added. That means you technically don't have to require mustHaveExpression, because it'll always have one, no?

gkalpak:
Not really (if I understand correctly what you are asking :grin:).

We call `addAttrInterpolateDirective(...)` for _every_ attribute and it is `addAttrInterpolateDirective`'s job to find out whether there is an interpolation and an actual directive needs to be added.

In order to do this, it passed the string value to `$interpolate` and checks whether interpolate return something. (With `mustHaveExpression` true, `$interpolate` will only return "something" - i.e. an interpolation function - if there is something to interpolate.)

In case of constant values in `ngAttrXyz`, `$interpolate` would return `undefined` (because of `mustHaveExpression`) and no directive would be added, which means the `xyz` attribute would not be set.

--
With this change though, since `mustHaveExpression` will be false for `ngAttrXyz`, `$interpolate` will return something (even if there is not expression in the string value) and a directive will be set up, which will in turn set the `xyz` property.

(And because `$interpolate` will return a special interpolation function it has for constants, the created watcher will be unregistered after the first time it is triggered, so there is no extra cost - except for the initial directive+watcher setup cost, of course.)

PR #15157, opened 2016-09-19
feat($http): remove deprecated callback methods: `success()/error()`

gkalpak:
Feature removal (refactor but with a breaking change).

The promises returned by `$http` have additions, non-standard methods: `success()`, `error()`. They have beed deprecated since v1.4.4 (#12112).
Although there was a switch to disable them (`$httpProvider.useLegacyPromiseExtensions()`), they were enabled by default.

`success()` and `error()` are gone. Use the standard `then()`/`catch()` promise methods instead.

Yes.

**Other information**:
- [ ] Add a note to the v1.5.x docs that the methods/switch will be removed in v1.6.0.


Narretz:
Were all docs updated when this feature was deprecated? Otherwise we'd have to check that.

gkalpak:
Good point. I saw there were docs updates in #12112, so I assumed all docs were updated.
It turns out there was one last survivor in `$sce`'s docs :smiley:

Narretz:
LGTM.
Minor note: in the commit message BC notes, you advise using .catch for error handling, but you don't mention the second argument of then (the error handler fn). Is this on purpose?
Did you talk about the consequences in the meeting? We rarely removed deprecated features. What's the consensus on the impact of this change?

gkalpak:
> In the commit message BC notes, you advise using `.catch` for error handling, but you don't mention the second argument of `then` (the error handler fn). Is this on purpose?

Yes and no. I don't specifically advise using `.catch` to replace `.error`. I have a generic statement that you should use `then`/`catch` now and I have one example, where `.error` is replaced by `.catch` (because in that particular example it is more straight-forward).

Pedantically speaking, `.then(onSuccess, onError)` is semantically closer to `.success(onSuccess).error(onError)`. But only because `.success`/`.error` were bad APIs :smiley: What people usually want is `.then(onSuccess).catch(onError)` (which is **not** equivalent to `.success(...).error(...)`).

So, while I didn't mean to explicitly advice people replacing `.success(...).error(...)` with `.then(...).catch(...)`, I do think it is good practice (i.e. better than `.then(..., ...)` - unless you know what you're doing).

But I don't feel strongly about it. If you have concerns about the implications of that, I would be happy to discuss :smiley:

--
Did you talk about the consequences in the meeting? We rarely removed deprecated features. What's the consensus on the impact of this change?

We didn't talk much about it. We just agreed it is time for this long-deprecated API to go.
The impact would be the same as with most breaking changes afaict: People will need to migrate their apps to not use the removed feature :smiley:

<sub>BTW, it is not true that we haven't removed other deprecated stuff before (don't judge by `replace: true` - it is a special case 😛 ).</sub>

--
I'll leave this unmerged for the time being and let's revisit on the next meeting to make sure we didn't miss anything.

PR #15160, opened 2016-09-19
fix($resource): pass `options` when binding default params

gkalpak:
Bug fix.

`Resource.bind()`, which creates an new Resource class with the same properties except for the
additional bound default parameters, was not passing the original Resource class' `options`,
resulting in different behavior.

The `options` are passed when creating the new Resource class.

No.

**Other information**:
This feature is undocumented. It might be a better idea to remove it altogether (although it does sounds useful). In any case, if we're keeping it, let's make it bug-free :smiley:

This PR also includes a couple of `$resource` refactorings that don't change the behavior.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -825,7 +825,8 @@ angular.module('ngResource', ['ng']).
         });
 
         Resource.bind = function(additionalParamDefaults) {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
LGTM as soon as Travis is happy

PR #15161, opened 2016-09-20
feat($httpBackend): JSONP requests now require trusted resource 

petebacondarwin:
This PR replaces #15143 which is too complicated.


src/ng/http.js
```diff
@@ -1088,7 +1094,12 @@ function $HttpProvider() {
         }
 
         // send request
-        return sendReq(config, reqData).then(transformResponse, transformResponse);
+        try {
+          var promise = sendReq(config, reqData);
```
mprobst: nit: I find it a bit more readable to define `var`s in the block scope that they are used in, even if they are function scoped. That is, move the declaration of `var promise` up one line?

mprobst (approved):
LGTM

src/ng/http.js
petebacondarwin: Sure thing

petebacondarwin:
If it is possible to add a parameter with both user generated key and value then it would be possible to completely control the browser when the JSONP response returns, assuming that the server does not do any sanitizing of its callback parameter value.

Given that every JSONP endpoint could define its own way of specifying the callback, i.e. it is not necessarily a param with key `callback` then it is very hard to generally constrain the param at the client.

It seems to me that in JSONP requests it is more likely that an app developer would inadvertently allow user generated content in the params than it is to allow user generated content in the URL itself (which is what we are targeting in this PR).

Perhaps we could change the `$http` JSONP to require that the callback parameter key is specified in the method call. We could then delete any user provided parameters that have this key before we attach our own callback parameter internally? This would remove the need for the clunky `JSONP_CALLBACK` pattern match and lock out the param attack vector.

An example API would look like:

```
var promise = $http({
  method: 'JSONP',
  url: $sce.trustAsResourceUrl('https://trusted.com/path'),
  callbackKey: 'callback'
});
```

Without the `callbackKey` the call would fail. This param could be specified in the `$http.defaults.jsonp.callbackKey='callback'` to save it needing to be added to every request.

petebacondarwin:
Of course this puts the onus on the app developer to remember to specify the correct callback param key, since if it doesn't match then the real callback param would still be vulnerable.

mprobst:
That sounds reasonable, and arguably like a nicer API than JSONP_CALLBACK.
But wouldn't this cause quite some migration pain?

petebacondarwin:
If we setup the default callback param key to be `callback` by default, which is very much the defacto standard for JSONP, then there would be no breaking change... All the current JSONP calls that look like:

```
$http.jsonp('some/trusted/url', { callback: 'JSONP_CALLBACK' });
```
or
```
$http.jsonp('some/trusted/url?callback=JSONP_CALLBACK');
```

would simply have that callback deleted before the service adds it back in itself.

We could even log a warning message if we find a param with the callback key name...

petebacondarwin:
The problem comes in the case where the callback key is not `callback`:

```
$http.jsonp('some/trusted/url', { cb: 'JSONP_CALLBACK' });
```

In this case, we would not automatically delete the `cb` parameter, since by default we would be looking for a `callback` param.

We could parse the params ones that have a value of `JSONP_CALLBACK` and log a warning again in that case, that some migration needs to be done.

There might be some rare case where `callback` is a valid param name for the JSONP call, while the expected callback param key is something else, (e.g. `cb`). In this case we would delete a valid parameter causing the request to fail in some manner.

src/ng/http.js
```diff
@@ -1249,12 +1260,19 @@ function $HttpProvider() {
           cache,
           cachedResp,
           reqHeaders = config.headers,
-          url = buildUrl(config.url, config.paramSerializer(config.params));
+          url = config.url;
+
+      if (lowercase(config.method) === 'jsonp') {
+        // This is a pretty sensitive operation where we're allowing a script to have full access to
+        // our DOM and JS space.  So we require that the URL satisfies SCE.RESOURCE_URL.
+        url = $sce.getTrustedResourceUrl(url);
```
gkalpak: So, if I understand correctly, `config.url` must be a string for all non-JSONP requests (i.e. it can't be an `$sce`-wrapped value - not that it needs to). And for JSONP requests it can be either an `$sce`-trusted resourceUrl or matched by a rule in the resourceUrlWhiteList (but not in the blackList). Right?

It might be more staight-forward if we allowed wrapped values for other methods too. E.g.:

```js
if (... === 'jsonp') {
  ...
} else {
  url = $sce.valueOf(url);
}
```

test/ng/httpSpec.js
```diff
@@ -1010,16 +1016,41 @@ describe('$http', function() {
...
+      });
+
+      it('should not throw error if the url is an explicitly trusted resource', function() {
+        expect(function() {
+          $httpBackend.expect('JSONP', 'http://example.org/path?cb=JSON_CALLBACK').respond('');
+          $http({ method: 'JSONP', url: $sce.trustAsResourceUrl('http://example.org/path?cb=JSON_CALLBACK'), callback: callback});
+        }).not.toThrow();
+      });
+
+      it('jsonp() should allow trusted url', inject(['$sce', function($sce) {
```
gkalpak: This description is misleading.

gkalpak (commented):
Basically LGTM. It's obviously not a full-proof solution, but it is better that what we have now.

src/ng/http.js
rjamet: +1: accepting safe types where they make sense but aren't technically needed is a good thing.

petebacondarwin: Sadly there is no `$sce.valueOf` method but I will implement this suggestion.

rjamet: There is: https://github.com/angular/angular.js/blob/master/src/ng/sce.js#L389.

(it's just seems undocumented, weirdly ?)

rjamet:
LGTM too, both for the current and the discussed modified version. It's still JSONP, so it's still going to result in bugs, but at least the framework does what it can that way. Thanks a lot !

petebacondarwin:
I've added the callback parameter checking feature in 308f3bf.
@mprobst, @rjamet, @gkalpak and @IgorMinar - PTAL

src/ng/http.js
```diff
@@ -1386,5 +1447,23 @@ function $HttpProvider() {
...
+    function sanitizeJsonpCallbackParam(url, key) {
+      if (/[&?][^=]+=JSON_CALLBACK/.test(url)) {
+        // Throw if the url already contains a reference to JSON_CALLBACK
+        throw $httpMinErr('badjsonp', 'Illegal use of JSON_CALLBACK in url, "{0}"', url);
+      }
+
+      var callbackParamRegex = new RegExp('([&?]' + key + '=)');
+      if (callbackParamRegex.test(url)) {
+        // Throw if the callback param was already provided
+        throw $httpMinErr('badjsonp', 'Illegal use of callback param, "{0}", in url, "{1}"', key, url);
```
Narretz: Is it necessary to throw here? Couldn't we just remove the superfluous param?

Narretz:
Some change suggestions to the commit message:

feat($http): specify the JSONP callback via the `callbackParam` config value

The query parameter that will be used to transmit the JSONP callback to the
server is now specified via the `callbackParam` config value  **of the $httpProvider**, instead of
using the `JSON_CALLBACK` placeholder. **instead of adding a query param to the url that has the `JSON_CALLBACK` placeholder value**?

* Any use of `JSON_CALLBACK` in a JSONP request URL will cause an error.
* Any request that provides a parameter with the same name as that given
by the `callbackParam` config property will cause an error.

This is to prevent malicious attack via the response from an app inadvertently
allowing untrusted data to be used to generate the callback parameter.

BREAKING CHANGE

You can no longer use the `JSON_CALLBACK` placeholder in your JSONP requests.
Instead you must provide the name of the query parameter that will pass the
callback via the `callbackParam` property of the config object, or app-wide via
the `$http.defaults.callbackParam` property, which is `callback` by default.

Before this change:

```
$http.json**p**('trusted/url?callback=JSON_CALLBACK');
$http.json**p**('other/trusted/url', {params:cb:'JSON_CALLBACK'});
```

After this change:

```
$http.json**p**('trusted/url');
$http.json**p**('other/trusted/url', {callbackParam:'cb'});
```

Narretz (changes requested):

Narretz:
The Travis failure is real, and is caused by the jsonp e2e tests in the docs

src/ng/http.js
```diff
@@ -309,7 +313,9 @@ function $HttpProvider() {
     xsrfCookieName: 'XSRF-TOKEN',
     xsrfHeaderName: 'X-XSRF-TOKEN',
 
-    paramSerializer: '$httpParamSerializer'
+    paramSerializer: '$httpParamSerializer',
+
+    callbackParam: 'callback'
```
petebacondarwin: OK

petebacondarwin:
@Narretz I fixed the e2e test that was failing.

src/ng/http.js
petebacondarwin: Done

petebacondarwin:
@Narretz regarding the changes to the commit message. I agree with adding the code comments to the examples.

In the earlier description though:

>The query parameter that will be used to transmit the JSONP callback to the
server is now specified via the callbackParam config value of the $httpProvider, ~~instead of
using the JSON_CALLBACK placeholder.~~ **instead of adding a query param to the url that has the JSON_CALLBACK placeholder value?**

I mostly disagree - because it was possible before that the callback was passed in some other form, such as part of the URL (e.g. `https://some-domain.org/end/point/JSON_CALLBACK`). I know that this probably never happened but it was possible and we supported it. Now we will indeed only  support a parameter. If you have to deal with the breaking change then you will already understand what the purpose and use of `JSON_CALLBACK` was so I don't think that the way I described it was confusing.

test/ng/httpSpec.js
```diff
@@ -289,27 +289,48 @@ describe('$http', function() {
...
-      }).toThrowMinErr('$http','badreq', 'Http request configuration url must be a string.  Received: false');
+        $http({});
+      }).toThrowMinErr('$http','badreq', 'Http request configuration url must be a string or a $sce trusted object.  Received: undefined');
     });
 
+    it('should accept a $sce trusted object for the request configuration url', function() {
+      expect(function() {
+        $httpBackend.expect('GET', '/url').respond('');
+        $http({url: $sce.trustAsResourceUrl('/url')});
+      }).not.toThrowMinErr('$http','badreq', 'Http request configuration url must be a string.  Received: false');
```
gkalpak: This assertion is too weak (because a negation of a too specific assertion).
Better use the stricter assertion possible. E.g. in order of preference:

- `.not.toThrow()`
- `.not.toThrow('$http')`
- `.not.toThrow('$http', 'badreq')`

docs/content/error/$http/badjsonp.ngdoc
```diff
@@ -0,0 +1,18 @@
+@ngdoc error
+@name $http:badjsonp
+@fullName Bad JSONP Request Configuration
+@description
+
+This error occurs when the URL generated from the configuration object contains a parameter with the same name as the configured `jsonpCallbackParam`
+property; or when it contains a parameter whose value is `JSON_CALLBACK`.
+
+`$http` JSON requests need to attach a callback query parameter to the URL. The name of this parameter is specified in the configuration
```
gkalpak: _JSON_  --.>  _JSONP_

src/ng/http.js
```diff
@@ -286,6 +286,10 @@ function $HttpProvider() {
    *  If specified as string, it is interpreted as a function registered with the {@link auto.$injector $injector}.
    *  Defaults to {@link ng.$httpParamSerializer $httpParamSerializer}.
    *
+   * - **`defaults.jsonpCallbackParam`** - `{string} - the name of the query parameter that passes the callback in a JSONP
+   * request. The value of this parameter will be replaced with the expression generated by the {@link jsonpCallbacks}
```
gkalpak: Does this link work? Shouldn't it be `$jsonpCallbacks` or something?

src/ng/http.js
```diff
@@ -1146,11 +1168,34 @@ function $HttpProvider() {
      *
      * @description
      * Shortcut method to perform `JSONP` request.
+     *
+     * Note that, since JSONP requests are sensitive because the response is given full acces to the browser,
```
gkalpak: _acces_  -->  _access_

src/ng/http.js
```diff
@@ -1146,11 +1168,34 @@ function $HttpProvider() {
...
+     * the url must be declared, via {@link $sce} as a trusted resource URL.
+     * You can trust a URL by adding it to the whitelist via
+     * {@link $sceDelegateProvider#resourceUrlWhitelist  `$sceDelegateProvider.resourceUrlWhitelist`} or
+     * by explicitly trusted the URL via {@link $sce#trustAsResourceUrl `$sce.trustAsResourceUrl(url)`}.
```
gkalpak: _trusted_  -->  _trusting_

src/ng/http.js
```diff
@@ -1146,11 +1168,34 @@ function $HttpProvider() {
...
+     *
+     * JSONP requests must specify a callback to be used in the response from the server. This callback
+     * is passed as as a query parameter in the request. You must specify the name of this parameter by
```
gkalpak: _as as_  -->  _as_

src/ng/http.js
```diff
@@ -1146,11 +1168,34 @@ function $HttpProvider() {
...
+     * setting the `jsonpCallbackParam` property on the request config object.
+     *
+     * ```
+     * $http.jsonp('some/trusted/url', {jsonpCallbackParam: 'callback'})
+     * ```
+     *
+     * You can also specify a global callback parameter key in `$http.defaults.jsonpCallbackParam`.
+     * By default this is set to `callback`.
```
gkalpak: (\`callback\`  -->  \`'callback'\`)

gkalpak: (\`callback\`  -->  \`'callback'\`)

src/ng/http.js
```diff
@@ -1386,5 +1447,23 @@ function $HttpProvider() {
...
       return url;
     }
+
+    function sanitizeJsonpCallbackParam(url, key) {
+      if (/[&?][^=]+=JSON_CALLBACK/.test(url)) {
+        // Throw if the url already contains a reference to JSON_CALLBACK
+        throw $httpMinErr('badjsonp', 'Illegal use of JSON_CALLBACK in url, "{0}"', url);
+      }
+
+      var callbackParamRegex = new RegExp('([&?]' + key + '=)');
```
gkalpak: Why the parenthesis?

test/ng/httpSpec.js
petebacondarwin: Yes, OK, OK. In fact even better is to remove the `not.toThrow` altogether!

src/ng/http.js
petebacondarwin: Because I was originally using it in a replace. Good spot.

src/ng/http.js
petebacondarwin: The aim of throwing is to a) make it more obvious where something has been missed in a migration and b) to indicate that a malicious attacker is trying to inject their own callback.

But perhaps this is a bit harsh?

petebacondarwin: I think we should go with the harsh BC of throwing and always rollback to just ignoring in a patch release if we get lots of complaints.

test/ng/httpSpec.js
petebacondarwin: changed

src/ng/http.js
petebacondarwin: fixed

test/ng/httpSpec.js
petebacondarwin: fixed

docs/content/error/$http/badjsonp.ngdoc
petebacondarwin: fixed

src/ng/http.js
petebacondarwin: fixed

src/ng/http.js
petebacondarwin: fixed

src/ng/http.js
petebacondarwin: fixed

src/ng/http.js
petebacondarwin: fixed

src/ng/http.js
petebacondarwin: fixed

src/ng/http.js
petebacondarwin: fixed

src/ng/http.js
```diff
@@ -1146,11 +1168,34 @@ function $HttpProvider() {
...
+     *
+     * JSONP requests must specify a callback to be used in the response from the server. This callback
+     * is passed as a query parameter in the request. You must specify the name of this parameter by
+     * setting the `jsonpCallbackParam` property on the request config object.
+     *
+     * ```
+     * $http.jsonp('some/trusted/url', {jsonpCallbackParam: 'callback'})
+     * ```
+     *
+     * You can also specify a global callback parameter key in `$http.defaults.jsonpCallbackParam`.
```
gkalpak: _global_ sounds a little confusing to me. How about _default_?

gkalpak (commented):
LGTM (other than a couple of nitpicks and the occasional >100chars-long lines 😛)

docs/content/error/$http/badjsonp.ngdoc
```diff
@@ -0,0 +1,18 @@
+@ngdoc error
+@name $http:badjsonp
+@fullName Bad JSONP Request Configuration
+@description
+
+This error occurs when the URL generated from the configuration object contains a parameter with the same name as the configured `jsonpCallbackParam`
```
petebacondarwin: K

src/ng/http.js
```diff
@@ -286,6 +286,10 @@ function $HttpProvider() {
    *  If specified as string, it is interpreted as a function registered with the {@link auto.$injector $injector}.
    *  Defaults to {@link ng.$httpParamSerializer $httpParamSerializer}.
    *
+   * - **`defaults.jsonpCallbackParam`** - `{string} - the name of the query parameter that passes the callback in a JSONP
```
petebacondarwin: K

src/ng/http.js
```diff
@@ -1088,7 +1103,11 @@ function $HttpProvider() {
         }
 
         // send request
-        return sendReq(config, reqData).then(transformResponse, transformResponse);
+        try {
+          return sendReq(config, reqData).then(transformResponse, transformResponse);
+        } catch (e) {
+          return $q.reject(e);
```
petebacondarwin: I'll wait. I have reviewed #15213 - I expect we can land that today

src/ng/http.js
petebacondarwin: default is OK.

src/ng/http.js
petebacondarwin: Actually this is needed. We are not trying to catch exceptions from the `then` handlers, we are catching exceptions thrown synchronously inside `sendReq`. E.g. `url = $sce.getTrustedResourceUrl(url);` can throw.

gkalpak: This is inside `serverRequest`, which is passed as an "onFulfilled" callback to `promise`: `promise.then(serverRequest)`.
So, even synchronous errors in `sendReq()` will be converted to rejections.

petebacondarwin: Oh OK. I'll remove it after 15213 then

gkalpak (approved):
LGTM

PR #15171, opened 2016-09-21
Revert input[range] change in 1.5.x

petebacondarwin:


Narretz:
I think these are all the commits. I think it's also possible to revert multiple commits in a single commit, but it's also okay as is.

petebacondarwin:
@Narretz I didn't know you could do that. There were conflicts I had to solve in order to revert these commits. It was not straight forward. This is why I wanted a review of the changes.

Narretz:
LGTM.
A few of the changes - docs, refactors, and the spec change in the validatorsSpec.js file could be backported to 1.5.x without a BC.

petebacondarwin:
@Narretz thanks for the review. It is cleaner to revert commits as they were and then reapply any bits that didn't need it as new commits afterwards.

I'll merge this and then could you look into reapplying those bits?

petebacondarwin:
Landed on v1.5.x as  6a167e8e, aa604910, 5b633d81 and cb4603b8

PR #15172, opened 2016-09-21
Update component-router.ngdoc

ppham27:

docs update

The current example in the docs is broken. The user is never prompted to discard changes.

The new behavior fixes this

No.

**Other information**:

this.cancel should not reassign ctrl.editName. The reassignment causes this.$routerCanDeactivate to always return true, and the user is never prompted to Discard changes. 

Should the user press "Cancel" and not discard changes, the editName is preserved. Should the user press "Yes" and discard the changes, the ctrl.crisis.name will remain the same, which is exactly the behavior that we want.

### DIFF
--- a/docs/content/guide/component-router.ngdoc
+++ b/docs/content/guide/component-router.ngdoc
@@ -376,7 +376,6 @@ You can see the complete application running below.
       };
 
       this.cancel = function() {
-        ctrl.editName = ctrl.crisis.name;
         ctrl.gotoCrises();
       };
 


ppham27:
I signed it!

gkalpak:
Thx for taking the time to submit a PR 👍
I am afraid we are going to pass this time, as this is the expected behavior.

Pressing "Cancel" means "I want to explicitly cancel my edits and return to the list". So we don't want to prompt the user for unsaved changes. The prompt should only appear if the user navigates away of the view by other means than pressing any of the buttons (e.g. clicking on a navigation menu item or clicking the bowser's "Back" button).

<sub>This is obviously not be ideal from a UX perspective, but it's purpose is to showcase the different possibilities :grin:</sub>

PR #15173, opened 2016-09-21
docs(guide/Filters): Fix incorrect wording

3noch:
**What kind of change does this PR introduce?** Docs

**What is the current behavior?** Incorrect wording

**What is the new behavior (if this is a feature change)?** Correct wording

**Does this PR introduce a breaking change?** No

**Other information**:

I don't think "Idempotence" is a property that filters need to fulfill: https://en.wikipedia.org/wiki/Idempotence. Idempotence means that `f(f(x)) == f(x)`, i.e. that applying a filter to the result of itself gives the same value as only applying it once. Many common filters are not idempotent. For example, applying reverse twice is not the same as applying it once.

### DIFF
--- a/docs/content/guide/filter.ngdoc
+++ b/docs/content/guide/filter.ngdoc
@@ -109,8 +109,9 @@ as the first argument. Any filter arguments are passed in as additional argument
 function.
 
 The filter function should be a [pure function](http://en.wikipedia.org/wiki/Pure_function), which
-means that it should be stateless and idempotent, and not rely for example on other Angular services.
-Angular relies on this contract and will by default execute a filter only when the inputs to the function change.
+means that it should always return the same result given the same input arguments and should not affect
+external state, for example, other Angular services. Angular relies on this contract and will by default
+execute a filter only when the inputs to the function change.
 {@link guide/filter#stateful-filters Stateful filters} are possible, but less performant.
 
 <div class="alert alert-warning">


docs/content/guide/filter.ngdoc
```diff
@@ -109,7 +109,7 @@ as the first argument. Any filter arguments are passed in as additional argument
 function.
 
 The filter function should be a [pure function](http://en.wikipedia.org/wiki/Pure_function), which
-means that it should be stateless and idempotent, and not rely for example on other Angular services.
+means that it should be deterministic and not rely, for example, on other Angular services.
```
gkalpak: Why remove `idempotent`? It is a very important characteristic and is not synonym to deterministic.
Also, "stateless" (even if not 100% accurate) is important (and again not covered by deterministic).

3noch: `idempotent` is not actually correct for filters.

3noch: `deterministic` actually does cover the idea of `stateless` since all stateless functions are inherently deterministic. However, it's also quite possible to have stateful functions that are also deterministic as long as the state is entirely self-contained. This is the reason I changed it.

3noch: Also note that in the context, this is giving explanation of "pure function" in the sentence. Pure functions are most certainly not idempotent by necessity (`f(x) = x + 1` is pure but not idempotent), and they can even be stateful in careful cases.

gkalpak: > idempotent is not actually correct for filters.

Here, it is used in the broader, "computer science" meaning:
_"In computer science, the term idempotent is used more comprehensively to describe an operation that will produce the same results if executed once or multiple times."_

But I agree it can be confusing, given the stricter mathematical (and pure functional programming) meanings.

--
> `deterministic` actually does cover the idea of `stateless` since all stateless functions are inherently deterministic.

This is not necessarily true (depending on the strictness of your definition). Basically, `stateless` isn't accurate either.

--
Since we already have a link to formal definition of "pure function", why don't we keep it simple here and change it to something like:

```diff
 The filter function should be a [pure function](http://en.wikipedia.org/wiki/Pure_function), which
-means that it should be stateless and idempotent, and not rely for example on other Angular services.
+means that it should always return the same result given the same input arguments and should not affect external state, for example, other Angular services.
```

3noch: Excellent. I can live with that. I'll rework the commit.

3noch: ✅

gkalpak:
LGTM

PR #15181, opened 2016-09-23
Align jqLite's attr method with jQuery

mgol:
Refactors & breaking API changes.

Boolean attribute getters take properties into account and setters modify them. All falsy values for boolean attribute setters remove the attribute and set the property to false. The `null` value in a setter (for all attributes, not only boolean ones) sets the attribute value to the string `"null"` instead of removing it.

Boolean attribute getters/setters don't touch properties. The `null` value in a setter (for all attributes, not only boolean ones) removes the attribute. The `false` value removes it only for boolean attributes. Other defined values for boolean attributes set the value to the lowercased attribute name.

Yes.

- [ ] ~~Docs have been added / updated (for bug fixes / features)~~ - only in a form of commit messages with their `BREAKING CHANGES` sections; is anything else needed?

**Other information**:

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -654,11 +654,13 @@ forEach({
         return element.getAttribute(name) != null ? lowercasedName : undefined;
       }
     } else if (isDefined(value)) {
-      element.setAttribute(name, value);
+      if (value === null) {
+        element.removeAttribute(name);
+      } else {
+        element.setAttribute(name, value);
+      }
     } else if (element.getAttribute) {
-      // the extra argument "2" is to get the right thing for a.href in IE, see jQuery code
-      // some elements (e.g. Document) don't have get attribute, so return undefined
-      var ret = element.getAttribute(name, 2);
+      var ret = element.getAttribute(name);
       // normalize non-existing attributes to undefined (as jQuery)
       return ret === null ? undefined : ret;
     }
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -645,7 +645,7 @@ forEach({
     var lowercasedName = lowercase(name);
     if (BOOLEAN_ATTR[lowercasedName]) {
       if (isDefined(value)) {
-        if (value) {
+        if (value !== false && value !== null) {
           element.setAttribute(name, name);
         } else {
           element.removeAttribute(name);
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -646,17 +646,12 @@ forEach({
     if (BOOLEAN_ATTR[lowercasedName]) {
       if (isDefined(value)) {
         if (value) {
-          element[name] = true;
-          element.setAttribute(name, lowercasedName);
+          element.setAttribute(name, name);
         } else {
-          element[name] = false;
-          element.removeAttribute(lowercasedName);
+          element.removeAttribute(name);
         }
       } else {
-        return (element[name] ||
-                 (element.attributes.getNamedItem(name) || noop).specified)
-               ? lowercasedName
-               : undefined;
+        return element.getAttribute(name) != null ? lowercasedName : undefined;
       }
     } else if (isDefined(value)) {
       element.setAttribute(name, value);
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -638,30 +638,32 @@ forEach({
   },
 
   attr: function(element, name, value) {
+    var ret;
     var nodeType = element.nodeType;
     if (nodeType === NODE_TYPE_TEXT || nodeType === NODE_TYPE_ATTRIBUTE || nodeType === NODE_TYPE_COMMENT) {
       return;
     }
+
     var lowercasedName = lowercase(name);
-    if (BOOLEAN_ATTR[lowercasedName]) {
-      if (isDefined(value)) {
-        if (value !== false && value !== null) {
-          element.setAttribute(name, name);
-        } else {
-          element.removeAttribute(name);
-        }
-      } else {
-        return element.getAttribute(name) != null ? lowercasedName : undefined;
-      }
-    } else if (isDefined(value)) {
-      if (value === null) {
+    var isBooleanAttr = BOOLEAN_ATTR[lowercasedName];
+
+    if (isDefined(value)) {
+      // setter
+
+      if (value === null || (value === false && isBooleanAttr)) {
         element.removeAttribute(name);
       } else {
-        element.setAttribute(name, value);
+        element.setAttribute(name, isBooleanAttr ? lowercasedName : value);
       }
     } else if (element.getAttribute) {
-      var ret = element.getAttribute(name);
-      // normalize non-existing attributes to undefined (as jQuery)
+      // getter
+
+      ret = element.getAttribute(name);
+
+      if (isBooleanAttr && ret !== null) {
+        ret = lowercasedName;
+      }
+      // Normalize non-existing attributes to undefined (as jQuery).
       return ret === null ? undefined : ret;
     }
   },
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -78,7 +78,7 @@
  * - [`prop()`](http://api.jquery.com/prop/)
  * - [`ready()`](http://api.jquery.com/ready/)
  * - [`remove()`](http://api.jquery.com/remove/)
- * - [`removeAttr()`](http://api.jquery.com/removeAttr/)
+ * - [`removeAttr()`](http://api.jquery.com/removeAttr/) - Does not support multiple attributes
  * - [`removeClass()`](http://api.jquery.com/removeClass/) - Does not support a function as first argument
  * - [`removeData()`](http://api.jquery.com/removeData/)
  * - [`replaceWith()`](http://api.jquery.com/replaceWith/)
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -1060,11 +1060,11 @@ forEach({
     }
     return isDefined(value) ? value : this;
   };
-
-  // bind legacy bind/unbind to on/off
-  JQLite.prototype.bind = JQLite.prototype.on;
-  JQLite.prototype.unbind = JQLite.prototype.off;
 });
+
+// bind legacy bind/unbind to on/off
+JQLite.prototype.bind = JQLite.prototype.on;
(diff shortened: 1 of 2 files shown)


Narretz:
There are a few failures when running the tests with jquery 2.2, which is probably expected. How do we handle this?

mgol:
@Narretz I fixed the test: I'm just skipping that one in jQuery 2.x as this was a jQuery 3.0 change.

BTW, most other changes here are aligning even with older jQuery versions, this is not just about 3.0 changes.

mgol:
Tests pass, PR ready for review.

test/jqLiteSpec.js
```diff
@@ -635,6 +635,56 @@ describe('jqLite', function() {
       expect(select.attr('multiple')).toBe('multiple');
     });
 
+    it('should set respective properties when getting/setting boolean attributes', function() {
```
gkalpak: As far as I understand, it is actually the browser that does that, not jqLite/`.attr()`. 
We should either remove the test (if we don't consider it to be jqLite's job, or make sure jqLite always sets the properties (instead of relying on the browsers to do it).

gkalpak: Also, `getting` seems not related here.

test/jqLiteSpec.js
```diff
@@ -635,6 +635,56 @@ describe('jqLite', function() {
...
+
+      div[0].multiple = true;
+      expect(div.attr('multiple')).toBe(undefined);
+
+      div.attr('multiple', 'multiple');
+      div[0].multiple = false;
+      expect(div.attr('multiple')).toBe('multiple');
+    });
+
+    it('should not set properties when settting respective boolean attributes', function() {
```
gkalpak: Too much `t` in `setttings` :smiley:

src/jqLite.js
```diff
@@ -638,33 +638,32 @@ forEach({
   },
 
   attr: function(element, name, value) {
+    var ret;
```
gkalpak: <sub>Nit: I prefer declaring vars in the block in which they are used.</sub>

src/jqLite.js
```diff
@@ -638,33 +638,32 @@ forEach({
...
-          element[name] = false;
-          element.removeAttribute(lowercasedName);
-        }
+    var isBooleanAttr = BOOLEAN_ATTR[lowercasedName];
+
+    if (isDefined(value)) {
+      // setter
+
+      if (value === null || (value === false && isBooleanAttr)) {
+        element.removeAttribute(name);
```
gkalpak: Previously, for boolean attributes, we used the `lowercasedName`. Does it make a difference?

src/jqLite.js
```diff
@@ -638,33 +638,32 @@ forEach({
...
-        return (element[name] ||
-                 (element.attributes.getNamedItem(name) || noop).specified)
-               ? lowercasedName
-               : undefined;
+        element.setAttribute(name, isBooleanAttr ? lowercasedName : value);
+      }
+    } else {
+      // getter
+
+      ret = element.getAttribute(name);
```
gkalpak: Is there any chance `getAttribute(name)` could return `undefined` (on any browser)?

src/jqLite.js
```diff
@@ -638,33 +638,32 @@ forEach({
...
+
+      if (isBooleanAttr && ret !== null) {
+        ret = lowercasedName;
       }
-    } else if (isDefined(value)) {
-      element.setAttribute(name, value);
-    } else if (element.getAttribute) {
```
gkalpak: Previously, there was this `element.getAttribute` check, which is now removed.
Was it necessary for IE<9 browsers only? It would be useful to have a comment (e.g. in the PR) explaining why it is not necessary any more.

src/jqLite.js
```diff
@@ -638,33 +638,32 @@ forEach({
...
-      // the extra argument "2" is to get the right thing for a.href in IE, see jQuery code
-      // some elements (e.g. Document) don't have get attribute, so return undefined
-      var ret = element.getAttribute(name, 2);
```
gkalpak: Same as above: It would be useful to have a comment (e.g. in the PR) explaining why this is not necessary any more (assuming it isn't 😛).

gkalpak (commented):
I have left a couple of (mostly minor) comments. 

One thing that I believe is not pointed out explicitly enough in the BC notices is the fact that boolean attribute getters do **not** rely on the value of the corresponding property (as was previously the case).

From the commit message:

> Previously, all boolean attributes were reflected into properties in a setter and from a property in a getter, even on elements that don't treat those attributes in a special way. Now it's only happening for the elements where browsers do the conversion automatically.

This sounds to me like boolean attribute getters will still take the property into account, but only on browsers that do the conversion automatically (which is not the case).

test/jqLiteSpec.js
mgol: IMO it often does make sense to test things like that; in a way that we should have had a test like that in `1.5.x` and it should remain passing after we remove our own reflection. In this way we'd know if a buggy browser broke it and we need to patch it ourselves.

But maybe here we wouldn't do the patching either way so I'd be fine either way.

src/jqLite.js
mgol: It shouldn't. Also, jQuery removes `name`, not `lowercasedName`.

src/jqLite.js
mgol: Good catch. It's needed for `window` doesn't have `getAttribute`. In such cases [jQuery falls back from `attr` to `prop`](https://github.com/jquery/jquery/blob/3.1.1/src/attributes/attr.js#L37-L39); should we do the same here?

src/jqLite.js
mgol: Good catch. It's needed for `window` doesn't have `getAttribute`. In such cases [jQuery falls back from `attr` to `prop`](https://github.com/jquery/jquery/blob/3.1.1/src/attributes/attr.js#L37-L39); should we do the same here?

src/jqLite.js
mgol: I don't think so but if that happened then what do you propose? jQuery only checks for `null` to convert it to `undefined`, if a browser returned `undefined` we'd already have what we want.

test/jqLiteSpec.js
gkalpak: I was under the impression that we wouldn't do the patching even if a browser/version failed to do the right thing. If the plan is to do the patching in that case, then I would rather be on the afe side and set the properties anyway. (But if jQuery doesn't do it, I think it is better for jqLite not to do it either.)

I don't feel strongly about keeping the test around, but a different description that makes it clear we expect _the browser_ to do it would probably make things a little easier fr our future selves, if something breaks :smiley:

src/jqLite.js
gkalpak: SGTM

src/jqLite.js
gkalpak: [This line][1] was using a `!= null` check, so I assumed `undefined` is a possible return value. If that is the case, then we shouldn't set `ret = lowercasedName` when `ret === undefined`, but afaict `getAttribute()` shouldn't ever return `undefined` (in which case we are fine).

[1]: https://github.com/angular/angular.js/pull/15181/commits/6dc5e761bdee39baa9718316acc0c764f5130d25#diff-8e0d9e5c18fe4f1050484080525d4367R654

test/jqLiteSpec.js
mgol: OK, I removed the test.

src/jqLite.js
mgol: Done.

mgol: It was still failing in a setter which jQuery doesn't so I fixed that as well. I decided to not fall back to prop for now, though, but just return as this feature is not very well tested & documented in jQuery.

src/jqLite.js
mgol: Done.

src/jqLite.js
```diff
@@ -56,7 +56,7 @@
  * - [`after()`](http://api.jquery.com/after/)
  * - [`append()`](http://api.jquery.com/append/)
  * - [`attr()`](http://api.jquery.com/attr/) - Does not support functions as parameters
- * - [`bind()`](http://api.jquery.com/bind/) - Does not support namespaces, selectors or eventData
+ * - (_deprecated_, use `on()`) [`bind()`](http://api.jquery.com/bind/) - Does not support namespaces, selectors or eventData
```
mgol: Is that a good way to mark methods as deprecated? Thoughts?

mgol: That's how it looks like:
screen shot 2016-10-05 at 12 42 29

mgol: Changed.

gkalpak: We should alse:

1. Document deprecations with the release in which they will be removed (usually the next minor release).
2. Place an issue in the appropriate milestone to remind us to do so.

test/jqLiteSpec.js
```diff
@@ -2256,4 +2334,19 @@ describe('jqLite', function() {
       expect(onLoadCallback).toHaveBeenCalledOnce();
     });
   });
+
+
+  describe('bind()/unbind()', function() {
```
gkalpak: <sub>Super-nit: Other descriptions don't have parenthesis :grin:</sub>

src/jqLite.js
petebacondarwin: I prefer small functions so that there is not much difference between the two approaches :-P

test/jqLiteSpec.js
petebacondarwin: Don't worry about it

petebacondarwin:
LGTM

test/jqLiteSpec.js
mgol: Already changed locally! :D

gkalpak (approved):
LGTM
<sub>(Don't forget the issue for removing `bind/unbind' :smiley:)</sub>

gkalpak:
(Also, feel free to tweak the commit message as per https://github.com/angular/angular.js/pull/15181#pullrequestreview-2284088 :smiley:.)

petebacondarwin:
@mgol - feel free to tweak as suggested by @gkalpak  and then merge. Thanks for your work on this.

mgol:
Removal issue: #15221.

PR #15189, opened 2016-09-26
fix(ngAnimate): allow animations when $rootElement is a comment (ng-v…

Tzook:
Bug fix.

Animations don't work when the root element of the app has ng-view on it.

Animations work now.

No.

**Other information**:
https://jsbin.com/napenerajo/1/edit?html,js,output
See the jsbin to better understand the issue.

If the root of the app has ng-view, the $rootElement is then becoming its comment, and this lets ngAnimate think that animations should be skipped.

### DIFF
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -614,6 +614,18 @@ var $$AnimateQueueProvider = ['$animateProvider', /** @this */ function($animate
       return getDomNode(nodeOrElmA) === getDomNode(nodeOrElmB);
     }
 
+    function getRootElement() {
+      var rootElement;
+      var rootElementNode = getDomNode($rootElement);
+      // if the root is a comment, like in ng-view case, we want its sibling and not the comment itself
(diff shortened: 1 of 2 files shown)


gkalpak:
Having `ng-view` on the root element is not a good idea imo. The root element is special; for example `ngStrictDi` is looked for in the root element. It is also common (although not necessary) to put `ngCsp` and `ngJq` on the root element. 

Therefore, putting a structural directive (such as `ngView`) on the root element is quite risky. (And just because work now, it doesn't mean they will work  in the future as well.)

With `ngView` on an element it is possible to have multiple instances (copies) of the element or remove the element altogether (leaving a comment placeholder). These behavior are not recommended for the root element of the application.

--
Besides, regarding the implementation in this PR, there is no guarantee that the comment's sibling is the element we are looking for. For example, there might be animation in both the leaving `ngView` and the entering `ngView`. So, this only partially fixes the issue.

--
I am going to close this, but feel free to continue the discussion below. If there is a compelling usecase I would be happy to reopen this.

PR #15203, opened 2016-09-30
docs(ngPattern): add observe explain

TomKarachristos:
docs update

**Other information**:
The '$observe' is applied only in the 'pattern' not in 'ng-pattern'. Doc doesn't say that. We lost a day searching because of this, until I see the source code.

### DIFF
--- a/src/ng/directive/validators.js
+++ b/src/ng/directive/validators.js
@@ -113,6 +113,10 @@ var requiredDirective = function() {
  *   </li>
  * </ol>
  * </div>
+ * <div class="alert alert-info">
+ * **Note:** The $observe method is used for tracking changes only in the 'pattern'.
+ * </div>
+ *
  *
  * @example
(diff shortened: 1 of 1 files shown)


gkalpak:
I feel like this is an implementation detail that shouldn't concern developers. So, I am going to close this.
If you have run into an edge case or pitfall that needs to be documented or otherwise clarified in the docs, I would be happy to discuss, so please share more details below.

TomKarachristos:
We must add the info that if you change the value of the attribute 'ng-pattern' it will not work but if you change the value of the attribute ,pattern' will work. Many people in my work had this problem.

gkalpak:
I am not sure what you mean. As mentioned in the docs, `ngPattern` accepts an Angular expression (in contrast to `pattern` which accepts a - possibly interpolated - string). If you update the value of the `ngPattern` expression, it works as expected.

Could you post an example of what doesn't work with `ngPattern`?

Issue #14734, opened 2016-06-08
form.$error.required and form.$invalid validators are not syncing


Dean-Christian-Armada:
I was wondering if this is an angular bug. The $invalid is undefined while my $error.required is true when a custom $validators is declared on a directive.

Check this example: http://plnkr.co/edit/NuDGB64IetpcsaVB03T7?p=preview

gkalpak:
This seems like an Angular bug indeed. BTW, it has been browkn since `v1.3.0-rc.1`!

Dean-Christian-Armada:
Oh ok, so that's why thanks for the response

gkalpak:
After further investigation, I'm not sure if this is Angular's fault or your directive's.
There is definitely a bug in your directive, but I'm not sure if Angular should treat it differently - it's a gray, undocumented area :smiley:

Here's what happens:
In your custom validator, when `modelValue` is falsy (e.g. `undefined`, as happens initially), you return `check` without ever initializing it. You declare `var check` inside the if block, but since `var` is function scoped (not block scoped), the declaration gets hoisted to the top of the function block. Thus, when you `return check`, `check` is declared but not initialized (i.e. `undefined`).
This is obviously unintentional. You should change the validator to something like:

```js
function (modelValue, viewValue) {
  var check = true;
  if (modelValue) {
    ...
  }
  return check;
}
```

Now, Angular normally expects a boolean (true or false) to be returned from a validator. Returning `undefined` triggers a special behavior though: It designates that the validation is pending and sets both `$invalid` and `$valid` to `undefined`. This behavior was probably intended (and is internally used) for async validators.
Returning `undefined` from a sync validator, tricks Angular into setting the valididty state as `$pending` (i.e. as happens with async validators).

--
Feature or Bug ?

I lean towards the latter. @Narretz,  thoughts ?

BTW, the fix would be to change the first line [here][1], so that result is always a boolean, regardless of what the validator returns: `var result = !!validator(modelValue, viewValue);`
(But it might be a breaking change...)

[1]: https://github.com/angular/angular.js/blob/f3377da6a748007c11fde090890ee58fae4cefa5/src/ng/directive/ngModel.js#L622-L624

Dean-Christian-Armada:
Ok, I understand your point but since the $validator is 'available' as it is declared.

ngModel.$validators.available = function (modelValue, viewValue) {

}

[...]

PR #15208, opened 2016-10-03
Patch 14734 boolean sync validators

BobChao87:

Fix/clarification of how synchronous validators work.

#14734 All values except `false` literal and `undefined` are treated as a pass. `undefined` is treated as pending, which makes no sense for synchronous operation.

All falsy value returns to a synchronous validator are treated as fails, including `undefined`.

Yes, see commit message.

- [ x ] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ x ] Tests for the changes have been added (for bug fixes / features)

**Other information**:

I believe a change to docs in unnecessary. We still expect `true` and `false` to be the primary returns and don't want to encourage otherwise. This is mostly clarifying a niche, undocumented, edge-case to behave more intuitively.

### DIFF
--- a/test/ng/directive/ngModelSpec.js
+++ b/test/ng/directive/ngModelSpec.js
@@ -847,36 +847,36 @@ describe('ngModel', function() {
         expect(ctrl.$valid).toBe(true);
       });
 
-      it('should treat all responses as boolean for synchronous validators', function() {
-        var curry = function(v) {
-          return function() {
-            return v;
-          };
-        };
+      fit('should treat all responses as boolean for synchronous validators', function() {
+        // var curry = function(v) {
+        //   return function() {
+        //     return v;
+        //   };
+        // };
 
-        var tester = function(v, e) {
+        var expectValid = function(v, e) {
           ctrl.$modelValue = undefined;
-          ctrl.$validators.a = curry(v);
+          ctrl.$validators.a = valueFn(v);
 
           ctrl.$validate();
           expect(ctrl.$valid).toBe(e);
         };
 
         // False tests
-        tester(false, false);
-        tester(undefined, false);
-        tester(null, false);
-        tester(0, false);
-        tester(NaN, false);
-        tester('', false);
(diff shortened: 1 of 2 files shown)


test/ng/directive/ngModelSpec.js
```diff
@@ -847,6 +847,38 @@ describe('ngModel', function() {
         expect(ctrl.$valid).toBe(true);
       });
 
+      it('should treat all responses as boolean for synchronous validators', function() {
+        var curry = function(v) {
```
gkalpak: You don't need a new function for this. You can use the internal helper: `valueFn()`

test/ng/directive/ngModelSpec.js
```diff
@@ -847,6 +847,38 @@ describe('ngModel', function() {
...
+          return function() {
+            return v;
+          };
+        };
+
+        var tester = function(v, e) {
```
gkalpak: Nit: I would prefer a more descriptive name, such as `expectValid()`.

gkalpak (commented):
Apart from a couple of nitpicks, LGTM.

<sub>I would add a little more details to the commit message (regarding what was happening before) - can be done while merging.</sub>

test/ng/directive/ngModelSpec.js
BobChao87: Interesting, I'll make the switch. I was actually hoping to talk about the `curry()` anyways, because it's now used three times in this file (hence why I used it in the first place). I'll submit a separate PR to fix the others.

gkalpak: Good catch! Feel free to change all other instances too.

BobChao87:
Hopefully those changes to the commit message are in the direction you wanted.

gkalpak (approved):
LGTM

petebacondarwin:
Landed thanks.

PR #15213, opened 2016-10-04
fix($q): treat thrown errors as regular rejections

### LINKED ISSUE #3174: Chained promises broken when testing due to $exceptionHandler interference
The Promises specification has the following to say about chaining promises:

> `then` must return a promise

> promise2 = promise1.then(onFulfilled, onRejected);

> If either `onFulfilled` or `onRejected` returns a value that is not a promise, promise2 must be fulfilled with that value.

> If either `onFulfilled` or `onRejected` throws an exception, promise2 must be rejected with the thrown exception as the reason.

However, when using Karma to test a module, I noticed that the following inside the method under test failed:

    return $http.post(endpoint, data).then(function(response) {
      //return some data from the response
    }, function(error) {
      var throwable;
      //grab some data from the response to create a meaningful rejection reason
      throw throwable;
    });

When invoking the method as follows, with a mocked HTTP backend to force a failure:

    service.do(args).then(function(unused) {
      expect(false).toBe(true);//In the absence of a fail() method, use a contradiction
    }, function(error) {
      expect(error.code).toBe(FAILURE_VAL);
    });

Instead of the expectation being met in the `onRejected` function, the exception propagates up the stack and fails the test. This is because apparently during tests, the default policy of `$exceptionHandler` is *rethrow*. 

I don't understand what business $exceptionHandler has at all being invoked by $q, **especially** since this behavior violates the Promises/A specification. Exceptions being thrown is a normal and expected part of promise execution.

### LINKED ISSUE #14745: $q: Control flow duplication (forking) when uncaught error is handled in .then()
Triggered when:
```js
$q.resolve().then(function() {
  throw new Error();
}).catch(function(e) {
  // ok, but the catch-all handler also receives `e` (incorrect)
});
```

Correctly handled when:
```js
$q.resolve().then(function() {
  try {
    throw new Error();
  }
  catch(e) {
    return $q.reject(e);
  }
}).catch(function(e) {
  // ok, and the catch-all handler has not been activated (correct)
});
```

Relevant code: https://github.com/angular/angular.js/blob/c9dffde1cb167660120753181cb6d01dc1d1b3d0/src/ng/q.js#L366

1. line 365: the Error is correctly passed up the promise chain: `deferred.reject(e);`
2. line 366: the Error is also passed to the `exceptionHandler(e);`, which effectively forks the control flow, if the Promise chain in 1. was not ignored.

Normally this is not a ciritical issue, since `exceptionHandler()` is usually used to just print errors to the console, however correctly the error would make it to the `exceptionHandler()` only if it was not ignored by the `deferred.reject(e);` chain (which, I'm not sure if it's even feasible to implement).

Possibly related; #7992, #13653

gkalpak:
Refactor/Fix (depending on your point of view).

Errors thrown in a promise's `onFulfilled` or `onRejected` handlers are treated in a
slightly different manner than regular rejections:  
They are passed to the `$exceptionHandler()` (in addition to being converted to rejections).
See also #3174 and #14745.

The distinction is removed, by skipping the call to `$exceptionHandler()`, thus treating
thrown errors as regular rejections.

Yes.

**Other information**:
Previously, errors thrown in a promise's `onFulfilled` or `onRejected` handlers were treated in a
slightly different manner than regular rejections:
They were passed to the `$exceptionHandler()` (in addition to being converted to rejections).

The reasoning for this behavior was that an uncaught error is different than a regular rejection, as
it can be caused by a programming error, for example. In practice, this turned out to be confusing
or undesirable for users, since neither native promises nor any other popular promise library
distinguishes thrown errors from regular rejections.
(Note: While this behavior does not go against the Promises/A+ spec, it is not prescribed either.)

This commit removes the distinction, by skipping the call to `$exceptionHandler()`, thus treating
thrown errors as regular rejections.

**Note:**
Unless explicitly turned off, possibly unhandled rejections will still be caught and passed to the
`$exceptionHandler()`, so errors thrown due to programming errors and not otherwise handled (with a
subsequent `onRejected` handler) will not go unnoticed.

BREAKING CHANGE:

Previously, throwing an error from a promise's `onFulfilled` or `onRejection` handlers, would result
in passing the error to the `$exceptionHandler()` (in addition to rejecting the promise with the
error as reason).

Now, a thrown error is treated exactly the same as a regular rejection. This applies to all
services/controllers/filters etc that rely on `$q` (including built-in services, such as `$http` and
`$route`). For example, `$http`'s `transformRequest/Response` functions or a route's `redirectTo`
function as well as functions specified in a route's `resolve` object, will no longer result in a
call to `$exceptionHandler()` if they throw an error. Other than that, everything will continue to
behave in the same way; i.e. the promises will be rejected, route transition will be cancelled,
`$routeChangeError` events will be broadcasted etc.

Fixes #3174
Fixes #14745

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3052,8 +3052,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       $compileNode.empty();
 
-      $templateRequest(templateUrl).
-        then(function(content) {
+      $templateRequest(templateUrl)
+        .then(function(content) {
           var compileNode, tempTemplateAttrs, $template, childBoundTranscludeFn;
 
           content = denormalizeTemplate(content);
@@ -3131,13 +3131,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               childBoundTranscludeFn);
           }
           linkQueue = null;
-        }).
-        catch(function(error) {
+        }).catch(function(error) {
           if (error instanceof Error) {
             $exceptionHandler(error);
           }
-        }).
-        catch(noop);
+        }).catch(noop);
 
       return function delayedNodeLinkFn(ignoreChildLinkFn, scope, node, rootElement, boundTranscludeFn) {
         var childBoundTranscludeFn = boundTranscludeFn;
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3052,8 +3052,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       $compileNode.empty();
 
-      $templateRequest(templateUrl)
-        .then(function(content) {
+      $templateRequest(templateUrl).
+        then(function(content) {
           var compileNode, tempTemplateAttrs, $template, childBoundTranscludeFn;
 
           content = denormalizeTemplate(content);
@@ -3131,7 +3131,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               childBoundTranscludeFn);
           }
           linkQueue = null;
-        }).catch(noop);
+        }).
+        catch(function(error) {
+          if (error instanceof Error) {
+            $exceptionHandler(error);
+          }
+        }).
+        catch(noop);
 
       return function delayedNodeLinkFn(ignoreChildLinkFn, scope, node, rootElement, boundTranscludeFn) {
         var childBoundTranscludeFn = boundTranscludeFn;
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -1854,17 +1854,18 @@ describe('$compile', function() {
         ));
 
 
-        it('should throw an error and clear element content if the template fails to load', inject(
-            function($compile, $httpBackend, $rootScope) {
-              $httpBackend.expect('GET', 'hello.html').respond(404, 'Not Found!');
-              element = $compile('<div><b class="hello">content</b></div>')($rootScope);
+        it('should throw an error and clear element content if the template fails to load',
+          inject(function($compile, $exceptionHandler, $httpBackend, $rootScope) {
+            $httpBackend.expect('GET', 'hello.html').respond(404, 'Not Found!');
+            element = $compile('<div><b class="hello">content</b></div>')($rootScope);
 
-              expect(function() {
-                $httpBackend.flush();
-              }).toThrowMinErr('$compile', 'tpload', 'Failed to load template: hello.html');
-              expect(sortedHtml(element)).toBe('<div><b class="hello"></b></div>');
-            }
-        ));
+            $httpBackend.flush();
+
+            expect(sortedHtml(element)).toBe('<div><b class="hello"></b></div>');
+            expect($exceptionHandler.errors[0].message).toMatch(
+                /^\[\$compile:tpload] Failed to load template: hello\.html/);
+          })
+        );
 
 
         it('should prevent multiple templates per element', function() {
@@ -1878,13 +1879,15 @@ describe('$compile', function() {
               templateUrl: 'template.html'
             }));
           });
-          inject(function($compile, $httpBackend) {
+          inject(function($compile, $exceptionHandler, $httpBackend) {
             $httpBackend.whenGET('template.html').respond('<p>template.html</p>');
-            expect(function() {
-              $compile('<div><div class="sync async"></div></div>');
-              $httpBackend.flush();
-            }).toThrowMinErr('$compile', 'multidir', 'Multiple directives [async, sync] asking for template on: ' +
-                '<div class="sync async">');
+
+            $compile('<div><div class="sync async"></div></div>');
+            $httpBackend.flush();
+
+            expect($exceptionHandler.errors[0].message).toMatch(new RegExp(
+                '^\\[\\$compile:multidir] Multiple directives \\[async, sync] asking for ' +
+                'template on: <div class="sync async">'));
           });
         });
 
@@ -2667,14 +2670,15 @@ describe('$compile', function() {
         );
 
         it('should not allow more than one isolate/new scope creation per element regardless of `templateUrl`',
-          inject(function($httpBackend) {
+          inject(function($exceptionHandler, $httpBackend) {
             $httpBackend.expect('GET', 'tiscope.html').respond('<div>Hello, world !</div>');
 
-            expect(function() {
-              compile('<div class="tiscope-a; scope-b"></div>');
-              $httpBackend.flush();
-            }).toThrowMinErr('$compile', 'multidir', 'Multiple directives [scopeB, tiscopeA] ' +
-                'asking for new/isolated scope on: <div class="tiscope-a; scope-b ng-scope">');
+            compile('<div class="tiscope-a; scope-b"></div>');
+            $httpBackend.flush();
+
+            expect($exceptionHandler.errors[0].message).toMatch(new RegExp(
+                '^\\[\\$compile:multidir] Multiple directives \\[scopeB, tiscopeA] ' +
+                'asking for new/isolated scope on: <div class="tiscope-a; scope-b ng-scope">'));
           })
         );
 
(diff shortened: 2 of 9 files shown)


gkalpak:
Should we document that the `$exceptionHandler` is called on 1.5.x or earlier (as suggested in https://github.com/angular/angular.js/issues/3174#issuecomment-250259596)?
Sounds like low level stuff.

src/ng/compile.js
```diff
@@ -3052,8 +3052,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
       $compileNode.empty();
 
-      $templateRequest(templateUrl)
-        .then(function(content) {
+      $templateRequest(templateUrl).
```
petebacondarwin: So you're a "dots at the end of the line" kind of guy eh?
This is not in-keeping with the rest of the codebase

test/ng/compileSpec.js
```diff
@@ -9706,12 +9711,15 @@ describe('$compile', function() {
...
+
               $compile('<div template first></div>');
-              expect(function() {
-                $httpBackend.flush();
-              }).toThrowMinErr('$compile', 'multidir', /Multiple directives \[first, second\] asking for transclusion on: <p .+/);
+              $httpBackend.flush();
+
+              expect($exceptionHandler.errors[0].message).toMatch(new RegExp(
+                  '^\\[\\$compile:multidir] Multiple directives \\[first, second] asking for ' +
+                  'transclusion on: <p '));
```
petebacondarwin: I wonder if we should have a helper function for checking the exception handler message?

test/ng/templateRequestSpec.js
```diff
@@ -115,19 +115,17 @@ describe('$templateRequest', function() {
...
-
-    $templateRequest('tpl.html');
-
-    $rootScope.$digest();
+    inject(function($exceptionHandler, $httpBackend, $rootScope, $templateRequest) {
+      $httpBackend.expectGET('tpl.html').respond(404, '', {}, 'Not found');
 
-    expect(function() {
-      $rootScope.$digest();
+      $templateRequest('tpl.html').catch(noop);
```
petebacondarwin: Perhaps this `noop` should be a spy; and we should check that it gets called?

test/ng/templateRequestSpec.js
```diff
@@ -115,19 +115,17 @@ describe('$templateRequest', function() {
   }));
 
   it('should throw an error when the template is not found',
```
petebacondarwin: This description is no longer accurate.

petebacondarwin (changes requested):
A few minor things but generally LGTM

src/ng/compile.js
gkalpak: Knee-jerk reaction, sorry 😞 I'll amend.

<sub>(Although, there are some occurrences and I myself have sneaked in a [couple of those][1] lately 😛)</sub>

[1]: https://github.com/angular/angular.js/blob/823295fee058f169523e84f94be13962a9513d3d/src/ngRoute/route.js#L679-L685

petebacondarwin: I did a search of the `src` folder for `/\w+\.$/` and didn't find any - although there were lots of matches for sentences in the docs and I might have missed the code ones.

src/ng/compile.js
```diff
@@ -3131,7 +3131,13 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
           }
           linkQueue = null;
-        }).catch(noop);
+        }).
+        catch(function(error) {
+          if (error instanceof Error) {
+            $exceptionHandler(error);
+          }
+        }).
+        catch(noop);
```
dcherman: This catch shouldn't be needed.  The one above it is already catching all errors and is not re-throwing or returning a rejected promise, so this should be unreachable.

test/ng/compileSpec.js
gkalpak: It would be helpful indeed. I'll add one.

src/ng/compile.js
gkalpak: Depends on what the `$exceptionHandler` does. Throwing from the `$exceptionHandler` (as happens in tests) results in a Possibly Unhandled Rejection error.

dcherman: Ah ok, never dawned on me that would be a valid thing to do.

src/ng/compile.js
gkalpak: I was a little more lucky (without even taking the `test/` directory into account, because of too any occurrences) :smiley: :

https://github.com/angular/angular.js/blob/823295f/compare-master-to-stable.js#L75-L81
https://github.com/angular/angular.js/blob/823295f/compare-master-to-stable.js#L93-L104
https://github.com/angular/angular.js/blob/823295f/compare-master-to-stable.js#L110-L124
https://github.com/angular/angular.js/blob/823295f/compare-master-to-stable.js#L132-L142
https://github.com/angular/angular.js/blob/823295f/compare-master-to-stable.js#L154-L155
https://github.com/angular/angular.js/blob/823295f/compare-master-to-stable.js#L160-L164
https://github.com/angular/angular.js/blob/823295f/docs/app/test/errorSpec.js#L6-L7
https://github.com/angular/angular.js/blob/823295f/i18n/e2e/i18n-e2e.js#L73-L82
https://github.com/angular/angular.js/blob/823295f/i18n/e2e/i18n-e2e.js#L171-L172
https://github.com/angular/angular.js/blob/823295f/i18n/src/closureI18nExtractor.js#L93-L100
https://github.com/angular/angular.js/blob/823295f/lib/grunt/utils.js#L206-L207
https://github.com/angular/angular.js/blob/823295f/src/Angular.js#L1406-L1409
https://github.com/angular/angular.js/blob/823295f/src/Angular.js#L1425-L1433
https://github.com/angular/angular.js/blob/823295f/src/AngularPublic.js#L173-L174
https://github.com/angular/angular.js/blob/823295f/src/AngularPublic.js#L218-L223
https://github.com/angular/angular.js/blob/823295f/src/jqLite.js#L153-L154
https://github.com/angular/angular.js/blob/823295f/src/jqLite.js#L412-L413
https://github.com/angular/angular.js/blob/823295f/src/ng/cacheFactory.js#L66-L67
https://github.com/angular/angular.js/blob/823295f/src/ng/compile.js#L3292-L3293
https://github.com/angular/angular.js/blob/823295f/src/ng/exceptionHandler.js#L25-L26
https://github.com/angular/angular.js/blob/823295f/src/ng/http.js#L893-L894
https://github.com/angular/angular.js/blob/823295f/src/ng/interpolate.js#L111-L112
https://github.com/angular/angular.js/blob/823295f/src/ng/sce.js#L42-L44
https://github.com/angular/angular.js/blob/823295f/src/ng/directive/ngClass.js#L300-L301
https://github.com/angular/angular.js/blob/823295f/src/ng/directive/ngClass.js#L359-L362
https://github.com/angular/angular.js/blob/823295f/src/ng/directive/ngClass.js#L407-L410
https://github.com/angular/angular.js/blob/823295f/src/ng/directive/ngCloak.js#L46-L49
https://github.com/angular/angular.js/blob/823295f/src/ng/directive/ngModel.js#L94-L95
https://github.com/angular/angular.js/blob/823295f/src/ng/directive/ngModel.js#L160-L161
https://github.com/angular/angular.js/blob/823295f/src/ng/filter/filters.js#L74-L75
https://github.com/angular/angular.js/blob/823295f/src/ng/filter/filters.js#L563-L570
https://github.com/angular/angular.js/blob/823295f/src/ngAria/aria.js#L56-L57
https://github.com/angular/angular.js/blob/823295f/src/ngCookies/cookies.js#L19-L20
https://github.com/angular/angular.js/blob/823295f/src/ngCookies/cookieStore.js#L3-L4
https://github.com/angular/angular.js/blob/823295f/src/ngResource/resource.js#L431-L432
https://github.com/angular/angular.js/blob/823295f/src/ngResource/resource.js#L477-L478
https://github.com/angular/angular.js/blob/823295f/src/ngResource/resource.js#L489-L490
https://github.com/angular/angular.js/blob/823295f/src/ngRoute/route.js#L28-L30
https://github.com/angular/angular.js/blob/823295f/src/ngRoute/route.js#L680-L685
https://github.com/angular/angular.js/blob/823295f/src/ngRoute/route.js#L727-L728
https://github.com/angular/angular.js/blob/823295f/src/ngRoute/route.js#L753-L760
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/sanitize.js#L110-L116
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/sanitize.js#L122-L123
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/sanitize.js#L131-L132
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/sanitize.js#L423-L433
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/filter/linky.js#L91-L92
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/filter/linky.js#L98-L99
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/filter/linky.js#L107-L108
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/filter/linky.js#L115-L117
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/filter/linky.js#L122-L124
https://github.com/angular/angular.js/blob/823295f/src/ngSanitize/filter/linky.js#L752-L755

petebacondarwin: Oh well...

petebacondarwin (approved):

Issue #13542, opened 2015-12-15
ngMock contains unannotated functions


bisubus:
ngMock contains unannotated functions, i.e.

https://github.com/angular/angular.js/blob/d67e999dfbdf47b79fdb3830a04f4f4010a98b98/src/ngMock/angular-mocks.js#L2527

https://github.com/angular/angular.js/blob/d67e999dfbdf47b79fdb3830a04f4f4010a98b98/src/ngMock/angular-mocks.js#L2652

This prevents `angular-mocks.js` from being minified.

The problem was conditional (`angular.mock.module({ mockedService: ... })` syntax was incompatible with minification before [this commit](https://github.com/angular/angular.js/commit/d67e999dfbdf47b79fdb3830a04f4f4010a98b98)), now it throws

> [$injector:modulerr] Failed to instantiate module function (e){l.$providerInjector=e} due to:
> Error: [$injector:unpr] Unknown provider: e
> http://errors.angularjs.org/1.5.0-beta.2/$injector/unpr?p0=e

unconditionally.

Narretz:
Probably happens because ngMock isn't usually minified, because it should only run in tests, where minification isn't really needed. But we should probably fix it anyway.

gkalpak:
Indeed I don't think it is intended to be minified. (In fact we are not producing a `angular-mocks.min.js` as part of our build.)
Why minify it ?

szimek:
@gkalpak I just modified our Gulp file to prevent minification of `angular-mocks.js` as a workaround for this issue. Previously, we just added `angular-mocks.js` to a list of all external JS files (but only if `ENV === "test"`) that later all ended up as a single `vendor.js` file, which is loaded in `index.html`. It was just much easier to load it like this, especially for e2e tests.

gkalpak:
OK, from a quick search, this has been brought up #3534 and #4448 before.

The last comment from @IgorMinar sums it up nicely, I think:

> [...] We have no way to test it right now, so if we break it in the future, we'll fix it when someone complains.

So, @szimek, did you just complain ? :smile:

szimek:
@gkalpak I just wanted to explain why people minify it :smile:  I found a [workaround](http://reasoncodeexample.com/2015/10/03/loading-angular-mocks-js-as-a-mock-module-angular-integration-tests/) for it, so it's fine for me as it is, but I suspect there are a lot more people who just add `angular-mocks.js` to their standard pipeline that minifies everything and will complain later on :smile:

[...]

PR #15225, opened 2016-10-07
Module info

petebacondarwin:

feature

No way to find information about the currently loaded modules

Two new features:
- a method on `Module` (`info()`) that lets developers add arbitrary info about their modules.
- a property on `$injector` (`modules`) that lets developers access the module objects that were loaded into the injector at bootstrap.

This second item could also be useful to @ocombe for his ocLazyLoad project.

No

**Other information**:

Replaces the previous #12465 PR

### DIFF
--- a/src/AngularPublic.js
+++ b/src/AngularPublic.js
@@ -265,5 +265,6 @@ function publishExternalAPI(angular) {
         $$cookieReader: $$CookieReaderProvider
       });
     }
-  ]);
+  ])
+  .info({ angularVersion: '"NG_VERSION_FULL"' });
 }
--- a/src/loader.js
+++ b/src/loader.js
@@ -79,6 +79,9 @@ function setupModuleLoader(window) {
      * @returns {angular.Module} new module with the {@link angular.Module} api.
      */
     return function module(name, requires, configFn) {
+
+      var info = {};
+
       var assertNotHasOwnProperty = function(name, context) {
         if (name === 'hasOwnProperty') {
           throw ngMinErr('badname', 'hasOwnProperty is not a valid {0} name', context);
@@ -113,6 +116,45 @@ function setupModuleLoader(window) {
           _invokeQueue: invokeQueue,
           _configBlocks: configBlocks,
           _runBlocks: runBlocks,
+
+          /**
+           * @ngdoc method
+           * @name angular.Module#info
+           * @module ng
+           *
+           * @param {Object=} info Information about the module
+           * @returns {Object|Module} The current info object for this module if called as a getter,
+           *                          or `this` if called as a setter.
+           *
+           * @description
+           * Read and write custom information about this module.
+           * For example you could put the version of the module in here.
+           *
+           * ```js
+           * angular.module('myModule', []).info({ version: '1.0.0' });
+           * ```
+           *
+           * The version could then be read back out by accessing the module elsewhere:
+           *
+           * ```
+           * var version = angular.module('myModule').info().version;
+           * ```
+           *
+           * You can also retrieve this information during runtime via the
+           * {@link $injector#modules `$injector.modules`} property:
+           *
+           * ```js
+           * var version = $injector.modules['myModule'].info().version;
+           * ```
+           */
+          info: function(value) {
+            if (isDefined(value)) {
+              if (!isObject(value)) throw ngMinErr('aobj', 'Argument \'{0}\' must be an object', 'value');
+              info = value;
+              return this;
+            }
+            return info;
+          },
 
           /**
            * @ngdoc property
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -2,6 +2,37 @@
 
 /* globals support: false */
 
+describe('injector.modules', function() {
+    it('should expose the loaded module info on the instance injector', function() {
+      var test1 = angular.module('test1', ['test2']).info({ version: '1.1' });
+      var test2 = angular.module('test2', []).info({ version: '1.2' });
+      module('test1');
+      inject(['$injector', function($injector) {
+        expect(Object.keys($injector.modules)).toEqual(['ng', 'ngLocale', 'ngMock', 'test1', 'test2']);
+        expect($injector.modules['test1'].info()).toEqual({ version: '1.1' });
+        expect($injector.modules['test2'].info()).toEqual({ version: '1.2' });
+      }]);
+    });
+
+    it('should expose the loaded module info on the provider injector', function() {
+      var providerInjector;
+      var test1 = angular.module('test1', ['test2']).info({ version: '1.1' });
+      var test2 = angular.module('test2', [])
+        .info({ version: '1.2' })
+        .provider('test', ['$injector', function($injector) {
+          providerInjector = $injector;
+          return { $get: function() {} };
+        }]);
+      module('test1');
+      // needed to ensure that the provider blocks are executed
+      inject();
+
+      expect(Object.keys(providerInjector.modules)).toEqual(['ng', 'ngLocale', 'ngMock', 'test1', 'test2']);
+      expect(providerInjector.modules['test1'].info()).toEqual({ version: '1.1' });
+      expect(providerInjector.modules['test2'].info()).toEqual({ version: '1.2' });
+    });
+});
+
 describe('injector', function() {
   var providers;
   var injector;
--- a/test/loaderSpec.js
+++ b/test/loaderSpec.js
@@ -156,4 +156,35 @@ describe('module loader', function() {
   it('should expose `$$minErr` on the `angular` object', function() {
     expect(window.angular.$$minErr).toEqual(jasmine.any(Function));
   });
+
+  describe('Module', function() {
+    describe('info()', function() {
+      var theModule;
+
+      beforeEach(function() {
+        theModule = angular.module('theModule', []);
+      });
+
+      it('should default to an empty object', function() {
+        expect(theModule.info()).toEqual({});
+      });
+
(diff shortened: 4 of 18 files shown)


ocombe:
It's happening !!!!

OMG

Is this real life?

src/loader.js
```diff
@@ -115,6 +118,39 @@ function setupModuleLoader(window) {
...
+           *
+           * ```js
+           * angular.module('myModule', []).info({ version: '1.0.0' });
+           * ```
+           *
+           * You can retrieve this information during runtime via the
+           * {@link $injector#modules `$injector.modules`} property:
+           *
+           * ```js
+           * var version = $injector.modules['myModule'].info().version;
```
gkalpak: Wouldn't `angular.module('myModule').info()` also work?

src/loader.js
```diff
@@ -115,6 +118,39 @@ function setupModuleLoader(window) {
...
+           * @ngdoc method
+           * @name angular.Module#info
+           * @module ng
+           *
+           * @param {Object=} info Information about the module
+           * @returns {Object|Module} The current info object for this module if called as a getter,
+           *                          or `this` if called as a setter.
+           *
+           * @description
+           * Additional info about this module
```
gkalpak: This makes it sound as if passing an object would extend the existing info (not overwrite it).
I think we should be pretty explicit about that.

test/loaderSpec.js
```diff
@@ -156,4 +156,29 @@ describe('module loader', function() {
...
+        theModule = angular.module('theModule', []);
+      });
+
+      it('should default to an empty object', function() {
+        expect(theModule.info()).toEqual({});
+      });
+
+      it('should store the object passed as a param', function() {
+        theModule.info({ version: '1.2' });
+        expect(theModule.info()).toEqual({ version: '1.2' });
```
gkalpak: I would add a test to verify it overwrites the previous info object.

src/loader.js
petebacondarwin: OK

src/loader.js
petebacondarwin: Yes, but in that case you do know, necessarily if the module was
a) loaded in the first place
b) added to the injector
If it was not loaded then your line of code would throw an exception

test/loaderSpec.js
petebacondarwin: OK

src/loader.js
gkalpak: I mean use it in place of `$injector.modules['myModule'].info()` - i.e. at "runtime" - so it should have been added to the injector.
I was mainly curious basically. `$injector.modules['myModule']` works for me :smiley:

--
Hm...since we brought it up, `$injector` is kind of ambiguous, since it may refer to the `providerInjector` (which doesn't know about `modules`) or the `instanceInjector` (which does).

I doubt anyone is using the `providerInjector`, but still it might be a good idea to add `modules` on the `providerInjector` as well (for symmetry).

petebacondarwin: I don't follow. You can't use `angular.module('myModule').info()` since you don't know if that module has been loaded or included in the current app.
In fact, you don't even know what the current app is since there might be more than one bootstrapped on a page.

Also, `$injector` is not ambiguous. It is the only injector that you can inject into components and services, etc.

petebacondarwin: But maybe I'm missing the point...

dcherman: I think he's referring to the `$injector` that you can inject into `config` function which is not the same injector as is injected into a service, component, etc.

gkalpak: Yes, I am referring to the `providerInjector` that can be injected into providers/config blocks. E.g.:

```js
.provider('foo', ['$injector', function fooProvider(providerInjector) {
  this.$get = ['$injector', function foo(instanceInjector) {
    console.log(providerInjector === instanceInjector);   // false
  }
})
```

(Again, I don't expect many people to be actually using it, but you never know :smiley:)

--
Regarding `angular.module('someModule').info()`:
Assume you are the author of a library or addon that relies on another library (e.g. `myAwesomeMaterialToolbar`). You can request that your addon is loaded after `ngMaterial`, so you can detect its version and do different things. E.g.:

```js
// my-awesome-material-toolbar.js
var mdVersion = angular.module('ngMaterial').info().version;
var deps = (mdVersion >= '1.1.1') ? [] : ['myShimForAFeatureAddedInNgMaterial1.1.1'];

angular.module('myAwesomeMaterialToolbar', deps);
```

petebacondarwin: I will fix the providerInjector/instanceInjector thing.

Regarding the load time dependencies use case. For sure you can do what you suggest with the functionality provided in this PR, although you are having to ensure that the developer loads the script files in the correct order and is not using some async loading helper that might mess things up.

It would be safer (IMO) to do configuration of your awesome library in providers, etc... Although I accept it is difficult to optionally add in a directive, you could make the directive a noop if not wanted.

gkalpak: Works for me 👍

petebacondarwin:
PTAL @gkalpak

src/loader.js
```diff
@@ -115,6 +118,45 @@ function setupModuleLoader(window) {
...
+           * You can also retrieve this information during runtime via the
+           * {@link $injector#modules `$injector.modules`} property:
+           *
+           * ```js
+           * var version = $injector.modules['myModule'].info().version;
+           * ```
+           */
+          info: function(value) {
+            if (isDefined(value)) {
+              if (!isObject(value)) throw ngMinErr('aobj', 'Argument \'{0}\' must be an object', 'value');
```
gkalpak: This should probably be `value` instead of `'value'` (i.e. without the quotes), right?

test/auto/injectorSpec.js
```diff
@@ -2,6 +2,37 @@
 
 /* globals support: false */
 
+describe('injector.modules', function() {
+    it('should expose the loaded module info on the instance injector', function() {
+      var test1 = angular.module('test1', ['test2']).info({ version: '1.1' });
+      var test2 = angular.module('test2', []).info({ version: '1.2' });
+      module('test1');
+      inject(['$injector', function($injector) {
```
gkalpak: OOC, is the array necessary?

gkalpak (changes requested):

src/loader.js
petebacondarwin: No. I don't think so. The idea is that the error can be general and be used with other parameters names

test/auto/injectorSpec.js
petebacondarwin: Didn't we decide to make all our tests minification proof?

petebacondarwin: I'm sure there was an issue about it

src/loader.js
gkalpak: Right. Then why do you need the `{0}`? Why don't you just write `'Argument \'value\' must be an object'` directly? I guess in order to make it easier to recognize the part that is interchangable.
👍

test/auto/injectorSpec.js
gkalpak: I don't remember anything like that (why would we)?
Maybe we have decided to make `ngMocks` minification-safe (because it makes it easier for some people to incorporate it in their workflow), but not our tests.

petebacondarwin: Ah, yes you are right @gkalpak (as always) - I was thinking of this: https://github.com/angular/angular.js/issues/13542

gkalpak (approved):

gkalpak:
LGTM <sub>(with or without the array notation :smiley:)</sub>

petebacondarwin:
Thanks @gkalpak - I just thought we should check that there is something similar in A2 before landing this.

test/auto/injectorSpec.js
mgol: They don't have to be minification-proof but shouldn't we aim at making them pass in strict DI mode?

gkalpak: There is not such thing as "strict DI" in tests :)
(Also none of the other tests are, so it is obviously not a problem.)

petebacondarwin:
So in Angular there is the `NgModule` concept, which provides a similar place to store useful metadata about your modules/app. Let's merge this.

src/AngularPublic.js
```diff
@@ -265,5 +265,6 @@ function publishExternalAPI(angular) {
         $$cookieReader: $$CookieReaderProvider
       });
     }
-  ]);
+  ])
+  .info({ angularVersion: '"NG_VERSION_FULL"' });
```
gkalpak: Why not call it just `version` (since it's already namespaced by the module name).
Especially in other modules (`ngAnimate`, `ngAria`, etc).

gkalpak (commented):
Have you considered "version-stamping" `ngLocale`?

src/AngularPublic.js
petebacondarwin: I guess I was leaning towards the idea that you would stamp your module with the version of Angular that it was designed for, rather than the version of the module itself...

mgol: To me `angularVersion` seems more intuitive; otherwise I'd feel we're talking about the module version as Pete said.

petebacondarwin:
I will add version stamping to the locale modules too.

petebacondarwin:
It turns out that adding the version stamp to the locale files is a little more involved with changes to grunt required :-(

src/AngularPublic.js
gkalpak: I tend to think it as the module version indeed. But it doesn't make any real difference. `angularVersion` works for me.

gkalpak:
Sounds like it's not worth it. It still LGTM without vervion-stamping locale files :smiley:

PR #15228, opened 2016-10-08
ngClass[Odd/Even] overhaul

gkalpak:
Fix, Test, Refactor, Perf

Broken, inefficient.

Fixed, efficient.

No (to the best of my knowledge).

**Other information**:
Supercedes (and partly based on) #14404. Mad props to @drpicox for the initial idea and implementation.

This PR has been broken up in smaller, focused commits for easier reviewing. Better to review commit by commit (and probably in order). Most commits are (more or less) trivial - except for the last one.

Refactoring commits:
- 8822006: refactor(ngClass): remove unnecessary dependency on `$animate`
- 0409589: refactor(ngClass): remove redundant `$observe`r
- 2871a11: refactor(ngClass): simplify conditions
- a7e69f7: refactor(ngClass): move helper functions outside the closure
- 37cec7a: refactor(ngClass): exit `arrayDifference()` early if an input is empty

Test commits:
- 2705085: test(ngClass): add some tests about one-time bindings and objects inside arrays

Fix commits:
- 0ba4be6: fix(ngClassOdd/Even): keep track of classes even if odd/even do not match index

Perf commits:
- 89268af: perf(ngClass): only access the element's `data` once
- 5708424: perf(ngClass): avoid deep-watching (if possible) and unnecessary copies

Closes #14404

### DIFF
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -15,14 +15,12 @@ function classDirective(name, selector) {
       link: function(scope, element, attr) {
         var oldVal;
 
-        scope.$watch(attr[name], ngClassWatchAction, true);
-
         if (name !== 'ngClass') {
           scope.$watch('$index', function($index, old$index) {
             /* eslint-disable no-bitwise */
             var mod = $index & 1;
             if (mod !== (old$index & 1)) {
-              var classes = arrayClasses(scope.$eval(attr[name]));
+              var classes = arrayClasses(oldVal);
               if (mod === selector) {
                 addClasses(classes);
               } else {
@@ -32,6 +30,8 @@ function classDirective(name, selector) {
             /* eslint-enable */
           });
         }
+
+        scope.$watch(attr[name], ngClassWatchAction, true);
 
         function addClasses(classes) {
           var newClasses = digestClassCounts(classes, 1);
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -97,41 +97,42 @@ function classDirective(name, selector) {
         }
       }
     };
-
-    function arrayDifference(tokens1, tokens2) {
-      var values = [];
-
-      outer:
-      for (var i = 0; i < tokens1.length; i++) {
-        var token = tokens1[i];
-        for (var j = 0; j < tokens2.length; j++) {
-          if (token === tokens2[j]) continue outer;
-        }
-        values.push(token);
-      }
-      return values;
-    }
-
-    function arrayClasses(classVal) {
-      var classes = [];
-      if (isArray(classVal)) {
-        forEach(classVal, function(v) {
-          classes = classes.concat(arrayClasses(v));
-        });
-        return classes;
-      } else if (isString(classVal)) {
-        return classVal.split(' ');
-      } else if (isObject(classVal)) {
-        forEach(classVal, function(v, k) {
-          if (v) {
-            classes = classes.concat(k.split(' '));
-          }
-        });
-        return classes;
-      }
-      return classVal;
-    }
   }];
+
+  // Helpers
+  function arrayDifference(tokens1, tokens2) {
+    var values = [];
+
+    outer:
+    for (var i = 0; i < tokens1.length; i++) {
+      var token = tokens1[i];
+      for (var j = 0; j < tokens2.length; j++) {
+        if (token === tokens2[j]) continue outer;
+      }
+      values.push(token);
+    }
+    return values;
+  }
+
+  function arrayClasses(classVal) {
+    var classes = [];
+    if (isArray(classVal)) {
+      forEach(classVal, function(v) {
+        classes = classes.concat(arrayClasses(v));
+      });
+      return classes;
+    } else if (isString(classVal)) {
+      return classVal.split(' ');
+    } else if (isObject(classVal)) {
+      forEach(classVal, function(v, k) {
+        if (v) {
+          classes = classes.concat(k.split(' '));
+        }
+      });
+      return classes;
+    }
+    return classVal;
+  }
 }
 
 /**
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -17,11 +17,6 @@ function classDirective(name, selector) {
 
         scope.$watch(attr[name], ngClassWatchAction, true);
 
-        attr.$observe('class', function(value) {
-          ngClassWatchAction(scope.$eval(attr[name]));
-        });
-
-
         if (name !== 'ngClass') {
           scope.$watch('$index', function($index, old$index) {
             /* eslint-disable no-bitwise */
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -8,6 +8,7 @@
 
 function classDirective(name, selector) {
   name = 'ngClass' + name;
+  var indexWatchExpression;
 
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngClass.js
```diff
@@ -69,12 +70,9 @@ function classDirective(name, selector) {
...
           toRemove = digestClassCounts(toRemove, -1);
-          if (toAdd && toAdd.length) {
-            $animate.addClass(element, toAdd);
-          }
-          if (toRemove && toRemove.length) {
-            $animate.removeClass(element, toRemove);
-          }
+
+          attr.$addClass(toAdd);
+          attr.$removeClass(toRemove);
```
petebacondarwin: I never knew about these helpers!!

test/ng/directive/ngClassSpec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-describe('ngClass', function() {
+fdescribe('ngClass', function() {
```
petebacondarwin: Naught naughty

test/ng/directive/ngClassSpec.js
```diff
@@ -311,37 +311,79 @@ fdescribe('ngClass', function() {
...
+      expect(e1).toHaveClass('three');
+      expect(e1).not.toHaveClass('four');
+      expect(e1).toHaveClass('five');
+      expect(e2).toHaveClass('one');
+      expect(e2).toHaveClass('two');
+      expect(e2).toHaveClass('three');
+      expect(e2).not.toHaveClass('four');
+      expect(e2).toHaveClass('five');
+
+      $rootScope.$apply('two = "too"');
```
petebacondarwin: although this is very clever "two" and "too" are very similar and actually make the test harder to follow.

test/ng/directive/ngClassSpec.js
```diff
@@ -477,6 +477,62 @@ fdescribe('ngClass', function() {
...
+      $rootScope.$apply('className = "foo"');
+      expect(element).toHaveClass('foo');
+
+      $rootScope.$apply('className = "bar"');
+      expect(element).toHaveClass('foo');
+      expect(element).not.toHaveClass('bar');
+    })
+  );
+
+  it('should do value remove the watcher when static map one-time binding',
```
petebacondarwin: the test description here seems like a hybrid of the two tests above :-)

test/ng/directive/ngClassSpec.js
```diff
@@ -477,6 +477,62 @@ fdescribe('ngClass', function() {
...
+      $rootScope.$digest();
+      expect(element).toHaveClass('foo');
+    })
+  );
+
+  it('should track changes of mutating object inside an array',
+    inject(function($rootScope, $compile) {
+      $rootScope.classVar = [{orange: true}];
+      element = $compile('<div ng-class="classVar"></div>')($rootScope);
+      $rootScope.$digest();
```
petebacondarwin: Is it worth adding an expectation that it should have the class "orange" here?

test/ng/directive/ngClassSpec.js
```diff
@@ -462,6 +462,47 @@ fdescribe('ngClass', function() {
     expect(e2.hasClass('odd')).toBeFalsy();
   }));
 
+
+  it('should keep track of old classes even if odd/even does not match `$index` atm',
```
petebacondarwin: This description makes hard to understand what is actually being tested.

Are you saying that ngClass is not removing old classes when both the $index and the expression value changes?

petebacondarwin (commented):
> When using large objects as keys (e.g.: `{loaded: $ctrl.data}`).

I think you mean "When using large objects as values...", right?

test/ng/directive/ngClassSpec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-fdescribe('ngClass', function() {
+describe('ngClass', function() {
```
petebacondarwin: ahah!

petebacondarwin:
A few minor comments but generally looks good to me.

test/ng/directive/ngClassSpec.js
gkalpak: I already have a fixup commit at the end 😛

test/ng/directive/ngClassSpec.js
gkalpak: The cleverness credits go to the original test :smiley:
I can change it.

test/ng/directive/ngClassSpec.js
gkalpak: It might be because I was trying to solve it differently at first. Here is what was happening before this change:

1. Assume `ng-class-odd="foo"`, `$index = 0`, `foo = 'foo'`. The element has `className === 'foo'`.
2. If the `$index` and `foo` change simultaneously, say `$index = 1`, `foo = 'bar'`, then:
    1. The `attr[name]` watch kicks in first and it checks if `scope.$index & 1 === selector` (i.e. it uses the current value of `$index` to determine if it needs to update the classes). It decides it doesn't need to do anything, so at this point the element still has `className === 'foo'`.
    2. The `$index` watch kicks in. Since it determines that `$index & 1 !== old$index & 1`, it decides to remove the classes. But it tries to `scope.$eval(attr[name])` which now evaluates to `'bar'`. So the element will still have `className === 'foo'`.

Basically, the problem is that both watch-actions check against the current value of each other to determine what to do, which fails when both change in the same digest.
I.e. the `attr[name]` watch uses `scope.$index` (while it should use the previous index) and the `$index` watch uses `scope.$eval(attr[name])` (while it should use the previous value - the "old classes"). 

--
The commit fxes it by:

1. Moving the `$index` watch _before_ the `attr[name]` watch.
2. Using `oldVal` from the `$index` watch to add/remove.

This way, when the `$index & 1` changes, the `$index` watch will correctly add/remove the classes and then the `attr[name]` watch will update the classes (i.e. add the new ones) if necessary.

--
I hope this makes sense :smiley:

How about:

> It should add/remove the correct classes when the expression and `$index` change simultaneously

gkalpak:
Thx for the review @petebacondarwin, @jbedard. I have added some fixup/squash commits addressing the comments (except for the parse interceptor for `$index & 1`).

PTAL

petebacondarwin:
LGTM

src/ng/directive/ngClass.js
```diff
@@ -8,129 +8,175 @@
...
+          oldClassString = newClassString;
         }
       }
     };
+  }];
 
-    function arrayDifference(tokens1, tokens2) {
-      var values = [];
+  // Helpers
+  function arrayDifference(tokens1, tokens2) {
```
witoldszpur: Maybe this will be better ?

```function arrayDifference(tokens1, tokens2) {
      var tokens1Length = tokens1.length, tokens2Length = tokens2.length;
      
      if (!tokens1 || !tokens1Length) return [];
      if (!tokens2 || !tokens2Length) return tokens1;

      var values = [];

      outer:
        for (var i = 0; i < tokens1Length; i++) {
          var token = tokens1[i];
          for (var j = 0; j < tokens2Length; j++) {
            if (token === tokens2[j]) continue outer;
          }
          values.push(token);
        }

      return values;
    }```

gkalpak:
Closing in favor of #15246 (which is just a clean-up version of this).
<sub>All comments and discussions here apply to #15246 as well.</sub>

Thx for the reviews everyone!

PR #15229, opened 2016-10-08
feat(input): add opt-in support for `input[range]`

gkalpak:
Feature

`input[range]` is not supported.

Support for `input[range]` can be opted into, by using the `ng-range-input` attribute.

No.

**Other information**:
This is a POC. Docs need to be updated.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1027,6 +1027,121 @@ var inputType = {
    */
   'radio': radioInputType,
 
+  /**
+   * @ngdoc input
+   * @name input[range]
+   *
+   * @description
+   * Native range input with validation and transformation.
+   *
+   * The model for the range input must always be a `Number`.
+   *
+   * IE9 and other browsers that do not support the `range` type fall back
+   * to a text input without any default values for `min`, `max` and `step`. Model binding,
+   * validation and number parsing are nevertheless supported.
+   *
+   * Browsers that support range (latest Chrome, Safari, Firefox, Edge) treat `input[range]`
+   * in a way that never allows the input to hold an invalid value. That means:
+   * - any non-numerical value is set to `(max + min) / 2`.
+   * - any numerical value that is less than the current min val, or greater than the current max val
+   * is set to the min / max val respectively.
+   * - additionally, the current `step` is respected, so the nearest value that satisfies a step
+   * is used.
+   *
+   * See the [HTML Spec on input[type=range]](https://www.w3.org/TR/html5/forms.html#range-state-(type=range))
+   * for more info.
+   *
+   * This has the following consequences for Angular:
+   *
+   * Since the element value should always reflect the current model value, a range input
+   * will set the bound ngModel expression to the value that the browser has set for the
+   * input element. For example, in the following input `<input type="range" ng-model="model.value">`,
+   * if the application sets `model.value = null`, the browser will set the input to `'50'`.
+   * Angular will then set the model to `50`, to prevent input and model value being out of sync.
+   *
+   * That means the model for range will immediately be set to `50` after `ngModel` has been
+   * initialized. It also means a range input can never have the required error.
+   *
+   * This does not only affect changes to the model value, but also to the values of the `min`,
+   * `max`, and `step` attributes. When these change in a way that will cause the browser to modify
+   * the input value, Angular will also update the model value.
+   *
+   * Automatic value adjustment also means that a range input element can never have the `required`,
+   * `min`, or `max` errors.
+   *
+   * However, `step` is currently only fully implemented by Firefox. Other browsers have problems
+   * when the step value changes dynamically - they do not adjust the element value correctly, but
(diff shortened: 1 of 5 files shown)


petebacondarwin:
My initial comment is that we need to check that other things that were changed in these commits are not impacted, such as ngMax, ngMin, ngStep(?), etc.

gkalpak:
With the exception of [this line][1] (which should not affect other inputs afaict), there has been no change in any input type other than `range`.
(There are a couple of minor refactorings, but no actual change.)

[1]: https://github.com/angular/angular.js/pull/15229/commits/0ad4228d8ac731ea5f11be6c2ee2f4d9b1662824#diff-ae1963e24d6328c3aa019c0c1b9f4432R887

src/ng/directive/input.js
```diff
@@ -1398,6 +1510,19 @@ function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
...
+  if (isDefined(val) && !isNumber(val)) {
+    val = parseFloat(val);
+  }
+  return !isNumberNaN(val) ? val : undefined;
+}
+
+function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
+  badInputChecker(scope, element, attr, ctrl);
+  numberFormatterParser(ctrl);
+  baseInputType(scope, element, attr, ctrl, $sniffer, $browser);
```
petebacondarwin: It probably doesn't make any difference but we should probably switch this line with the line above to ensure the order of the code is the same as previously.

src/ng/directive/ngModel.js
```diff
@@ -883,7 +883,8 @@ var NgModelController = ['$scope', '$exceptionHandler', '$attrs', '$element', '$
         ctrl.$viewValue = ctrl.$$lastCommittedViewValue = viewValue;
         ctrl.$render();
 
-        ctrl.$$runValidators(modelValue, viewValue, noop);
+        // It is possible that model and view value have been updated during render
+        ctrl.$$runValidators(ctrl.$modelValue, ctrl.$viewValue, noop);
```
petebacondarwin: So this is not a breaking change because if $render didn't change the value then it will be the same as before, right? I think it is only the range input that is likely to do this...

test/ng/directive/inputSpec.js
```diff
@@ -3017,10 +3629,10 @@ describe('input', function() {
         ['scheme://example.com/path?query#~`!@#$%^&*-_=+|\\;:\'",.?/()[]{}<>', true],
 
         // Validating miscellaneous
-        ['scheme://☺.�.⌘.➡/䨹', true],
-        ['scheme://مثال.إخ�بار', true],
+        ['scheme://☺.?⌘.➡/䨹', true],
+        ['scheme://مثال.إخ??ار', true],
         ['scheme://例子.测试', true],
-        ['scheme://उदाहरण.�रीक्षा', true],
+        ['scheme://उदाहरण.???ीक्षा', true],
```
petebacondarwin: Hmm, I wonder if this change is right? It looks like previously the strings were UTF-16?

test/ng/directive/validatorsSpec.js
```diff
@@ -351,7 +351,7 @@ describe('validators', function() {
 
 
     it('should accept values of any length when maxlength is non-numeric', function() {
-      var inputElm = helper.compileInput('<input type="text" ng-model="value" ng-maxlength="{{maxlength}}" />');
+      var inputElm = helper.compileInput('<input type="text" ng-model="value" ng-maxlength="maxlength" />');
```
petebacondarwin: I guess that this was a bad test in the first place is not really part of this PR? Perhaps, if so, then when you are squashing all this, we move this change into its own commit?

petebacondarwin:
OK, I think this is about right.
Let's check with @Narretz and then if @tbosch thinks it is usefult we can fix up the docs and merge into 1.5

src/ng/directive/ngModel.js
gkalpak: Among the built-in `$render` function, yes. But the truth is it could be a breaking change for user defined `$render` functions (theoretically).

gkalpak: Although even in that case, it could be considered a fix :grin:

Narretz:
I think it's pretty unorthodox of us to provide an opt-in for a regular input in core, especially since it becomes the default in 1.6. We are also forcing Googlers to update to 1.6 sooner if we leave it out of 1.5 ;) But the implementation is clean, so we might as well. Although I would like to see the refactor commits that affect other inputs in a separate commit.

gkalpak:
I have tidied this up and fixed docs. (I also changed `ng-range-input` to `ng-input-range` as it seemed more consistent).

petebacondarwin:
Landed in 1.5.x as

* 091152ad6fe708540c88819a396bdff8ed6c9beb
* d2834628b2ccaace8a3831fd5d2dca5b6f47590f
* 8830ee8fa9cbf38d272e2437587ef4838dc80783

PR #15230, opened 2016-10-08
docs($controller): deprecate the use of $controllerProvider#allowGlobals

jbedard:
docs

puts the warning for a future one

Anything else that needs to be done to prepare for removing this in 1.7?

### DIFF
--- a/src/ng/controller.js
+++ b/src/ng/controller.js
@@ -59,6 +59,9 @@ function $ControllerProvider() {
    * @ngdoc method
    * @name $controllerProvider#allowGlobals
    * @description If called, allows `$controller` to find controller constructors on `window`
+   *
+   * @deprecated
+   * This method of finding controllers has been deprecated. This will be removed in 1.7.
    */
   this.allowGlobals = function() {
     globals = true;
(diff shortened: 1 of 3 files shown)


petebacondarwin:
We are going to remove the feature in 1.7?
We should document at which version it will be removed.

Narretz:
I agree. The version when this is removed should go into the deprecated tag. @petebacondarwin can we render this nicely with dgeni?

petebacondarwin:
We should be able to tweak the dgeni templates. I'll take a look tomorrow

src/ng/controller.js
```diff
@@ -59,6 +59,9 @@ function $ControllerProvider() {
    * @ngdoc method
    * @name $controllerProvider#allowGlobals
    * @description If called, allows `$controller` to find controller constructors on `window`
+   *
+   * @deprecated
+   * This method of finding controllers has been deprecated. This will be removed in a future version.
```
petebacondarwin: This will be removed in 1.7

jbedard:
@petebacondarwin updated with a specific version

src/ng/controller.js
```diff
@@ -59,6 +59,9 @@ function $ControllerProvider() {
    * @ngdoc method
    * @name $controllerProvider#allowGlobals
    * @description If called, allows `$controller` to find controller constructors on `window`
+   *
+   * @deprecated
+   * This method of finding controllers has been deprecated. This will be removed in 1.7
```
gkalpak: Missing trailing `.` 😇

jbedard: Updated

PR #15235, opened 2016-10-09
fix($location): prevent infinite digest with IDN urls in Edge

### LINKED ISSUE #15217: Inject $location service cause $rootScope:infdig error on Edge 14.14393 (latest) and Angular 1.5.8 when html5mode is on and the page has an Internationalized Domain Name domain
Hello, 

Yes, as the title define it "Inject $location service cause $rootScope:infdig error on Edge 14.14393 (latest) and Angular 1.5.8 when html5mode is on and the page have an IDN domain" quite specific and quite complicate to find the issue too.

**Before go to the detail** 
It happen only with the latest version of Edge (14.14393) and the latest version of Angular (1.5.8), if we use ng 1.5.8 and a previous version of Edge the problem doesn't happen, the same if we use ng 1.5.5 and the latest version of Edge. So, I am not totally sure if the problem could be on Edge or Angular, but for sure would be possible to have a workaround to avoid it.

**How to reproduce the error**
Run the following code in a IDN domain and browse the page using Edge (14.4393) 

http://jsfiddle.net/Lvc0u55v/10528/

Web console output:
```
Watchers fired in the last 5 iterations: []
http://errors.angularjs.org/1.5.8/$rootScope/infdig?p0=10&p1=%5B%5D
   at Scope.prototype.$digest (http://währungsrechner.dev/assets/javascript/all.73777e.js:27780:13)
   at Scope.prototype.$apply (http://währungsrechner.dev/assets/javascript/all.73777e.js:28008:13)
   at done (http://währungsrechner.dev/assets/javascript/all.73777e.js:22049:36)
   at completeRequest (http://währungsrechner.dev/assets/javascript/all.73777e.js:22251:7)
   at requestLoaded (http://währungsrechner.dev/assets/javascript/all.73777e.js:22184:9) undefined
all.73777e.js (24138,11)

SCRIPT5022: [$rootScope:infdig] 10 $digest() iterations reached. Aborting!
Watchers fired in the last 5 iterations: []
http://errors.angularjs.org/1.5.8/$rootScope/infdig?p0=10&p1=%5B%5D
all.73777e.js (28011,13)
```

If the domain is not an idn domain (doesn't matter if on the cloud or locally) the error doesn't happen.
If the html5mode is not activated, the error doesn't happen.

**Posible fix meanwhile**
Downgrade Angular to 1.5.5.
If possible doesn't use html5mode.
If possible doesn't use $location service, use window.location instead.

**Some links**
Edge 14.14393 (latest): https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/ (the version of BrowserStack doesn't throw error because is not the latest)
IDN domain: https://en.wikipedia.org/wiki/Internationalized_domain_name

Since usually Edge is updated automatically to the latest version, this could be a serious bug. Have html5mode activated and use $location service could be a common practice.

Thanks,
J.

Narretz:
Fix

Internationalized Domain Name Urls (IDN), e.g. with an Umlaut, in Edge lead to an infinite digest on page load.

No Infdig

No

Fixes #15217

Note:
We don't test on Edge yet, but I've tested this locally with the https://github.com/nickmccurdy/karma-edge-launcher

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -49,7 +49,7 @@ function parseAppUrl(relativeUrl, locationObj) {
 }
 
 function startsWith(haystack, needle) {
-  return haystack.lastIndexOf(needle, 0) === 0;
+  return haystack.indexOf(needle) === 0;
 }
 
 /**
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -48,8 +48,8 @@ function parseAppUrl(relativeUrl, locationObj) {
   }
 }
 
-function startsWith(haystack, needle) {
-  return haystack.indexOf(needle) === 0;
+function startsWith(str, search) {
+  return str.slice(0, search.length) === search;
 }
 
 /**
--- a/test/ng/locationSpec.js
+++ b/test/ng/locationSpec.js
@@ -2450,10 +2450,11 @@ describe('$location', function() {
 
 
   describe('LocationHtml5Url', function() {
-    var locationUrl, locationIndexUrl;
+    var locationUrl, locationUmlautUrl, locationIndexUrl;
 
     beforeEach(function() {
       locationUrl = new LocationHtml5Url('http://server/pre/', 'http://server/pre/', 'http://server/pre/path');
+      locationUmlautUrl = new LocationHtml5Url('http://särver/pre/', 'http://särver/pre/', 'http://särver/pre/path');
       locationIndexUrl = new LocationHtml5Url('http://server/pre/index.html', 'http://server/pre/', 'http://server/pre/path');
     });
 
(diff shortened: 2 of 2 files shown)


jbedard:
See [my comment](https://github.com/angular/angular.js/issues/15217#issuecomment-252509972) on the bug. I'm not sure if that really matters, depends where this method is used.

src/ng/location.js
```diff
@@ -49,7 +49,7 @@ function parseAppUrl(relativeUrl, locationObj) {
 }
 
 function startsWith(haystack, needle) {
-  return haystack.lastIndexOf(needle, 0) === 0;
+  return haystack.indexOf(needle, 0) === 0;
```
jbedard: Can drop the `0` param.

petebacondarwin:
[core-js](https://github.com/zloirock/core-js/blob/v2.4.1/modules/es6.string.starts-with.js#L16) has an interesting patch for startsWith:

```
str.slice(index, index + search.length) === search;
```

what do you think?

src/ng/location.js
```diff
@@ -49,7 +49,7 @@ function parseAppUrl(relativeUrl, locationObj) {
 }
 
 function startsWith(haystack, needle) {
-  return haystack.lastIndexOf(needle, 0) === 0;
+  return haystack.indexOf(needle) === 0;
```
gkalpak: Maybe it is worth adding a comment to prevent someone from "optimizing" it back in the future.

gkalpak (commented):
If were concerned about perfromance, we could "sniff" if the implementation of `lastIndexOf` is broken and fall back to `indexOf` or `slice` (I am not sure which on is faster).

But this function isn't on a performance-critical path, so it might not be worth "super-optimizing".

petebacondarwin:
I know that micro-performance is not a good way to measure but....

Testing the following strings against `abcdefghijklmnopqrstuvwxyz` yields:

```
RegExp#test: abc x 2,574,830 ops/sec ±2.78% (76 runs sampled)
String#indexOf abc x 11,061,777 ops/sec ±2.00% (76 runs sampled)
String#slice abc x 21,384,362 ops/sec ±1.78% (78 runs sampled)
RegExp#test: xyz x 3,043,606 ops/sec ±1.78% (74 runs sampled)
String#indexOf xyz x 11,240,707 ops/sec ±1.94% (78 runs sampled)
String#slice xyz x 21,867,749 ops/sec ±1.92% (79 runs sampled)
RegExp#test: 123 x 3,072,821 ops/sec ±1.79% (79 runs sampled)
String#indexOf 123 x 12,216,397 ops/sec ±1.99% (75 runs sampled)
String#slice 123 x 22,642,025 ops/sec ±1.53% (81 runs sampled)
```

with string slice winning out easily in all cases

gkalpak:
> I know that micro-performance is not a good way to measure

(No buts 😛)

Narretz:
I can put in the slice version as startsWith, without considering any existing startsWith fn. Btw, the bug is reported here: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9271625/ I'll add a comment to the line

petebacondarwin:
LGTM

Issue #15221, opened 2016-10-06
Remove jqLite#bind & jqLite#unbind


mgol:
***Note*: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**
A refactor (removing a feature).

**What is the current behavior?**
`jqLite#bind` & `jqLite#unbind` exist and are aliases to `jqLite#on` & `jqLite#off` respectively.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**
N/A

**What is the expected behavior?**
`jqLite#bind` & `jqLite#unbind` should be removed.

**What is the motivation / use case for changing the behavior?**
Those methods will be deprecated in 1.6.0 so we should remove them in 1.7.0.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**
N/A

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
N/A

dcherman:
@mgol Why plan for removal at all?  It seems like a potentially big breaking change in order to save 2 lines of code that don't introduce significant complexity, and it doesn't align with jQuery's goals of deprecation without removal for event shorthands/aliases (as you definitely know)

gkalpak:
We have recently decided that deprecated features should generally stay in the code for one minor release and be removed in the next minor (breaking) release. And that we should create issues with an appropriate milestone as a reminder.

This is not a hard rule; we can discuss it when the time comes for 1.7.0.

--
Admittedly, `bind`/`unbind` are kind of special, since they are just alises. On the other hand, migrating is pretty easy, so I don't see much benefit in allowing people use a method that has been deprecated for a long time. (Just my 2 cents.)

In any case, thx for raising this @dcherman so we can take it into account when the time comes :smiley:

mgol:
Thanks @dcherman for the comment.

@gkalpak It's true it's easy to change the code but this applies to your code only. jQuery has a large ecosystem of plugins, not all of which are regularly updated, especially as long as jQuery doesn't remove a particular deprecated API.

[...]

PR #15237, opened 2016-10-09
feat(jqLite): implement jqLite(f) as alias to jqLite(document).ready(f) (also: deprecate the latter in the second commit)

mgol:
Feature

`jqLite(callback)` is not supported.

`jqLite(callback)` is an alias to `jqLite(document).ready(callback)`. The latter is now deprecated.

No.

**Other information**:

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -56,7 +56,7 @@
  * - [`after()`](http://api.jquery.com/after/)
  * - [`append()`](http://api.jquery.com/append/)
  * - [`attr()`](http://api.jquery.com/attr/) - Does not support functions as parameters
- * - [`bind()`](http://api.jquery.com/bind/) (_deprecated_ - to be removed in 1.7.0, use [`on()`](http://api.jquery.com/on/)) - Does not support namespaces, selectors or eventData
+ * - [`bind()`](http://api.jquery.com/bind/) (_deprecated_, use [`on()`](http://api.jquery.com/on/)) - Does not support namespaces, selectors or eventData
  * - [`children()`](http://api.jquery.com/children/) - Does not support selectors
  * - [`clone()`](http://api.jquery.com/clone/)
  * - [`contents()`](http://api.jquery.com/contents/)
@@ -85,7 +85,7 @@
  * - [`text()`](http://api.jquery.com/text/)
  * - [`toggleClass()`](http://api.jquery.com/toggleClass/) - Does not support a function as first argument
  * - [`triggerHandler()`](http://api.jquery.com/triggerHandler/) - Passes a dummy event object to handlers
- * - [`unbind()`](http://api.jquery.com/unbind/) (_deprecated_ - to be removed in 1.7.0, use [`off()`](http://api.jquery.com/off/)) - Does not support namespaces or event object as parameter
+ * - [`unbind()`](http://api.jquery.com/unbind/) (_deprecated_, use [`off()`](http://api.jquery.com/off/)) - Does not support namespaces or event object as parameter
  * - [`val()`](http://api.jquery.com/val/)
  * - [`wrap()`](http://api.jquery.com/wrap/)
  *
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -76,7 +76,7 @@
  * - [`parent()`](http://api.jquery.com/parent/) - Does not support selectors
  * - [`prepend()`](http://api.jquery.com/prepend/)
  * - [`prop()`](http://api.jquery.com/prop/)
- * - [`ready()`](http://api.jquery.com/ready/)
+ * - [`ready()`](http://api.jquery.com/ready/) (_deprecated_, use `angular.element(callback)` instead of `angular.element(document).ready(callback)`)
  * - [`remove()`](http://api.jquery.com/remove/)
  * - [`removeAttr()`](http://api.jquery.com/removeAttr/) - Does not support multiple attributes
  * - [`removeClass()`](http://api.jquery.com/removeClass/) - Does not support a function as first argument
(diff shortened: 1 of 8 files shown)


src/jqLite.js
```diff
@@ -519,29 +521,34 @@ function jqLiteDocumentLoaded(action, win) {
   }
 }
 
+function jqLiteReady(fn) {
+  var fired = false;
+
+  function trigger() {
+    if (fired) return;
+    fired = true;
```
dcherman: I know the old code didn't do this, but shouldn't we just remove the event handlers rather than setting a flag?

mgol: Good idea. I jut copied it without analysing the contents too much. :D

petebacondarwin: Since we are wiring up both `DOMContentLoaded` and `load` events, is it not possible that they will both be triggered even if we remove the event handlers from inside the `trigger` function?
In other words, if the browser queues up both handlers before either handler has run, then removing the handlers inside one of the handlers would not prevent the other from running.

petebacondarwin (approved):
LGTM - we can do the refactor (if it works) of removing the event handlers in another PR later.

src/jqLite.js
```diff
@@ -76,7 +76,7 @@
  * - [`parent()`](http://api.jquery.com/parent/) - Does not support selectors
  * - [`prepend()`](http://api.jquery.com/prepend/)
  * - [`prop()`](http://api.jquery.com/prop/)
- * - [`ready()`](http://api.jquery.com/ready/)
+ * - [`ready()`](http://api.jquery.com/ready/) (_deprecated_, use `angular.element(callback)` instead of `angular.element(document).ready(callback)`)
```
petebacondarwin: document removal in 1.7?

src/jqLite.js
dcherman: Doubt it.  I'd imagine it would be considered a browser bug if any browsers behaved like that, otherwise you also couldn't add a handler for `load` in `DOMContentReady` which is almost certainly expected to work.

There's also precedent since this is how jQuery does it, so it's pretty battle tested.

src/jqLite.js
```diff
@@ -288,6 +288,8 @@ function JQLite(element) {
 
   if (argIsString) {
     jqLiteAddNodes(this, jqLiteParseHTML(element));
+  } else if (isFunction(element)) {
```
gkalpak: Couldn't this go above the `instanceof JQLite` check?

src/jqLite.js
```diff
@@ -519,29 +521,34 @@ function jqLiteDocumentLoaded(action, win) {
...
+  if (window.document.readyState === 'complete') {
+    window.setTimeout(trigger);
+  } else {
+    // We can not use jqLite since we are not done loading and jQuery could be loaded later.
+
+    // Works for modern browsers and IE9
+    window.document.addEventListener('DOMContentLoaded', trigger);
+
+    // Fallback to window.onload for others
+    window.addEventListener('load', trigger);
```
gkalpak: By using `addEventListener` (here and above) doesn't it mean the people can't use `.triggerHandler()`?
It could break apps (and even more likely tests).

src/jqLite.js
mgol: I didn't want to delay the most common path and jQuery has this check below the selector code. Maybe with JITs it's not a problem.

src/jqLite.js
mgol: It does but we no longer have an underlying element and [that's what jQuery uses](https://github.com/jquery/jquery/blob/3.1.1/src/core/ready.js#L88-L92) so this is aligning with jQuery.

src/jqLite.js
mgol: To be discussed; see #15221.

src/jqLite.js
gkalpak: It's probably not a problem, but it can't do much damage either (how many times would one call `jqLite(fn)`?).
So if jQuery has it this way, let's leave it as is.

src/jqLite.js
gkalpak: OK. Should we document it as breaking change then?

gkalpak: Actually, it shouldn't be possible to exercise the difference from a test, since you can't stop the `DOMContentLoaded`/`load` events. So, it's fine, I guess.

gkalpak (approved):
LGTM

src/jqLite.js
mgol: OK, jQuery does it as it's done here, I read the code incorrectly.

src/jqLite.js
mgol: I added a "refactor" in a separate commit.

src/jqLite.js
mgol: OK, I'll assume there's nothing to do here. :D

gkalpak: Yeah, I meant let's not consider it a breaking change :smiley:

gkalpak:
Still LGTM

src/jqLite.js
```diff
@@ -519,29 +521,32 @@ function jqLiteDocumentLoaded(action, win) {
...
+function jqLiteReady(fn) {
+  function trigger() {
+    window.document.removeEventListener('DOMContentLoaded', trigger);
+    window.removeEventListener('load', trigger);
+    fn();
+  }
+
+  // check if document is already loaded
+  if (window.document.readyState === 'complete') {
+    window.setTimeout(trigger);
```
jbedard: Should this just be `window.setTimeout(fn)` since we aren't adding the listeners?

src/jqLite.js
mgol: Conclusion: no removal note.

src/jqLite.js
mgol: Good point, fixed.

PR #15238, opened 2016-10-10
refactor(data): camelCase keys in jqLite#data

mgol:
A behavior change.

`jqLite#data` uses keys unchanged.

`jqLite#data` camelCases keys in its getters/setters. This aligns jqLite with jQuery.

Yes.

**Other information**:

This is WIP; this might slightly decrease performance so I wanted to discuss it first.

Also, finishing this would require making our `camelCase` function more strict so I'd like to discuss it first but I think we should do that anyway; the `-moz-` hack doesn't really belong here.

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -136,22 +136,31 @@ JQLite._data = function(node) {
 function jqNextId() { return ++jqId; }
 
 
-var SPECIAL_CHARS_REGEXP = /([:\-_]+(.))/g;
-var MOZ_HACK_REGEXP = /^moz([A-Z])/;
+var DASH_LOWERCASE_REGEXP = /-([a-z])/g;
+var MS_HACK_REGEXP = /^-ms-/;
 var MOUSE_EVENT_MAP = { mouseleave: 'mouseout', mouseenter: 'mouseover' };
 var jqLiteMinErr = minErr('jqLite');
 
 /**
- * Converts snake_case to camelCase.
- * Also there is special case for Moz prefix starting with upper case letter.
+ * Converts kebab-case to camelCase.
+ * There is also a special case for the ms prefix starting with a lowercase letter.
  * @param name Name to normalize
  */
-function camelCase(name) {
-  return name.
-    replace(SPECIAL_CHARS_REGEXP, function(_, separator, letter, offset) {
-      return offset ? letter.toUpperCase() : letter;
-    }).
-    replace(MOZ_HACK_REGEXP, 'Moz$1');
+function cssKebabToCamel(name) {
+    return kebabToCamel(name.replace(MS_HACK_REGEXP, 'ms-'));
+}
+
+function fnCamelCaseReplace(all, letter) {
+  return letter.toUpperCase();
+}
+
+/**
+ * Converts kebab-case to camelCase.
+ * @param name Name to normalize
+ */
+function kebabToCamel(name) {
+  return name
+    .replace(DASH_LOWERCASE_REGEXP, fnCamelCaseReplace);
 }
 
 var SINGLE_TAG_REGEXP = /^<([\w-]+)\s*\/?>(?:<\/\1>|)$/;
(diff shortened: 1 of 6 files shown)


test/jqLiteSpec.js
```diff
@@ -594,6 +594,39 @@ describe('jqLite', function() {
         }).not.toThrow();
       });
     });
+
+    fdescribe('camelCasing keys', function() {
+      // jQuery 2.x has different behavior; skip the tests.
+      // if (isJQuery2x()) return;
```
petebacondarwin: commented out?

test/jqLiteSpec.js
```diff
@@ -594,6 +594,39 @@ describe('jqLite', function() {
...
+
+      it('should camelCase the key in a setter', function() {
+        var element = jqLite(a);
+
+        element.data('a-B-c-d-42--e', 'z-x');
+        expect(element.data()).toEqual({'a-BCD-42-E': 'z-x'});
```
petebacondarwin: is this actually how we want the camelcasing behaviour to work when there are already capital letters and double hyphens in the key? Or is it because of our current implementation of `camelCase()`?

petebacondarwin (commented):
If performance is a concern then we ought to run some benchmarking...
Can we make `camelCase()` more efficient in the common case?

mgol:
The first (new) commit is the commit from PR #15249, it's a prerequisite to make the second commit pass the tests.

mgol:
I updated the commit message, adding a breaking change info. I'm not very happy about the migration steps there, any better ideas?

src/jqLite.js
```diff
@@ -391,16 +401,18 @@ function jqLiteData(element, key, value) {
...
       if (massGetter) {  // data()
         return data;
       } else {
         if (isSimpleGetter) { // data('key')
           // don't force creation of expandoStore if it doesn't exist yet
-          return data && data[key];
+          return data && data[kebabToCamel(key)];
         } else { // mass-setter: data({key1: val1, key2: val2})
-          extend(data, key);
+          for (prop in key) {
```
gkalpak: This is different than `extend`:

1. `extend` only copies own properties.
2. `extend` ignores `$$hashKey`.

If the chenges are intentional (e.g. to align with jQuery), we should clearly document them in the breaking change notice.

gkalpak (changes requested):

gkalpak:
The migration steps for that usecase work for me. I would also add an example of how this affects interacting with the data retrieved with a mass-getter. E.g.:

```js
// Before:
elem.data('foo-bar', 42);
elem.data()['foo-bar'];

// After:
elem.data('foo-bar', 42);
elem.data()['fooBar'];
```

petebacondarwin:
Both these commits should be scoped as `fix`. We are "fixing" the behaviour to match that of JQuery; but also with this the changes would not appear in the CHANGELOG automatically.

petebacondarwin:
I am not clear about what the migration step in 99b2dde is saying. Why are there more items in the "before" version? It would be helpful if the migration described what the "before" was trying to achieve and then how the "after" achieves it.

test/jqLiteSpec.js
mgol: This is exactly as jQuery does it. Only a single dash + a lowercase letter get converted.

mgol: This is also how the HTML5 spec converts data attribute names to keys in dataset. That's where jQuery took it from.

It's also nice in that it's fully 1-1, the conversion doesn't lose any information.

mgol:
> I am not clear about what the migration step in 99b2dde is saying. Why are there more items in the "before" version?

I meant that either one of them may be replaced by either one of the cases in the "After" section. Can we have multiple before/after pairs?

petebacondarwin:
OK, so it would be good if you said this explicitly in the commit message.

mgol:
> Both these commits should be scoped as fix

I fixed the first commit in #15249 and the second here & rebased. PTAL.

mgol:
@gkalpak PTAL as well, I added sth like what you suggested.

test/jqLiteSpec.js
petebacondarwin: OK

petebacondarwin:
Landed as fc0c11db845d53061430b7f05e773dcb3fb5b860 and 73050cdda04675bfa6705dc841ddbbb6919eb048

PR #15243, opened 2016-10-10
chore(benchpress): add a ngClass benchmark

drpicox:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**: benchmark
Adds a benchmark to measure the impact of changes in ngClass.

**What is the current behavior? (You can also link to an open issue here)**: https://github.com/angular/angular.js/pull/15228#commitcomment-19347288 

**What is the new behavior (if this is a feature change)?** no change

**Does this PR introduce a breaking change?** no

**Other information**:
It adds a benchmark in `benchmarks/ng-class-bp`
The benchmark was designed to measure the changes introduced by: https://github.com/angular/angular.js/pull/15228

It has the following steps:
- _setup_: cleanups preview, and create structures, not worth to measure
- _create_: creation of a table/list with the selected number of items and applied classes
- _$apply_: an apply without changes (compute digest time)
- _update_: cost of updating classes (changes ng-classes values)
- _unclass_: cost of removing all classes (falses all ngClass values)
- _class_: cost of apply classes (applies all classes given already constructed elements)
- _destroy_: cost of removing everything from dom (usually not worth to measure)

There are three versions to paint:
- **optimized table**: a table with optimized ngClass values `ng-class="'success' && todo.completed"`
- **unoptimized table**: a table with object ngClass syntaxis: `ng-class="{success: todo.completed}"`
- **list**: list of todos with an object ngClass and all cases: `ng-class="{completed: todo.completed, important: todo.important, urgent: todo.urgent}"`
- **single optimized**: it computes a single ngClass using a complex object but optimizing the access `ng-class="{success: !!$ctrl.todos, danger: !$ctrl.todos}"`
- **single**: it computes a single ngClass using a complex object `ng-class="{success: $ctrl.todos, danger: !$ctrl.todos}"`

I ran them over the previous version of ngClass and the version of #15228 , the results are:
- create, update, unclass, class:
  - variability in create, update, unclass and class seems to have more or less the same performance
  - the variability seems bigger in the #15228 implementation
- $apply:
  - optimized table:
    - #15228 is around a 25% slower than previous version
    - in my computer previous version took 0.75ms for 1000 rows, and for  #15228 around 1ms
    - I think that this implementation is not often used
  - unoptimized table and list:
    - #15228 is faster than previous version, around 5x and 10x faster
    - I ran the benchmark many times just to be sure
    - previous version has high variability from 7ms ~ 13ms
    -  #15228 maintains the speed around 1ms
  - single optimized:
    - #15228 seems to be slower but there is high variability
    - times are around 0.02ms ~ 0.08ms
  - single:
    - #15228 is really really faster (around more than 100x faster)
    - times of #15228 are close to single optimized
    - previous version takes around 1.5ms

I also modified #15228 to use compile instead of link (commented in https://github.com/angular/angular.js/pull/15228#commitcomment-19360023 ). Results are:
- create:
  - it improves the performance around 5%~10% (faster compile time)
- other cases:
  - no appreciable changes

So, as a conclusion, it seems that the **#15228  optimizes automatically code in the same fashion that an experienced angular programmers would do** manually. 
In the other hand, given that 1 frame is around 16ms, we should give 8ms to the browser to update, it leaves 8ms for the application to work. Experiments show that digest loop times moves from values like 8 or 10ms to just 1ms in large lists, or 0.01ms in a single big object condition.

### DIFF
--- a/benchmarks/ng-class-bp/app.js
+++ b/benchmarks/ng-class-bp/app.js
@@ -1,0 +1,108 @@
+'use strict';
+
+var app = angular.module('ngClassBenchmark', []);
+
+app.controller('DataController', function DataController($scope) {
+
+  this.init = function() {
+    this.numberOfTodos = 1000;
+    this.implementation = 'tableOptimized';
+    this.completedPeriodicity = 3;
+    this.importantPeriodicity = 13;
+    this.urgentPeriodicity = 29;
+
+    this.createTodos(100);
+    this.setTodosValuesWithSeed(0);
+  };
+
+  this.clearTodos = function() {
+    this.todos = null;
+  };
+
+  this.createTodos = function(count) {
+    var i;
+    this.todos = [];
+    for (i = 0; i < count; i++) {
+      this.todos.push({
+        id: i + 1,
+        completed: false,
+        important: false,
+        urgent: false
+      });
+    }
+  };
+
+  this.setTodosValuesWithSeed = function(offset) {
+    var i, todo;
+    for (i = 0; i < this.todos.length; i++) {
+      todo = this.todos[i];
+      todo.completed = 0 === (i + offset) % this.completedPeriodicity;
+      todo.important = 0 === (i + offset) % this.importantPeriodicity;
+      todo.urgent = 0 === (i + offset) % this.urgentPeriodicity;
+    }
+  };
+
+  this.init();
+
+
+  benchmarkSteps.push({
+    name: 'setup',
+    fn: function() {
+      $scope.$apply();
+      this.clearTodos();
+      this.createTodos(this.numberOfTodos);
+    }.bind(this)
+  });
+
+  benchmarkSteps.push({
+    name: 'create',
+    fn: function() {
+      // initialize data for first time that will construct the DOM
+      this.setTodosValuesWithSeed(0);
+      $scope.$apply();
+    }.bind(this)
+  });
+
+  benchmarkSteps.push({
+    name: '$apply',
+    fn: function() {
+      $scope.$apply();
+    }
+  });
+
+  benchmarkSteps.push({
+    name: 'update',
+    fn: function() {
+      // move everything but completed
+      this.setTodosValuesWithSeed(3);
+      $scope.$apply();
+    }.bind(this)
+  });
+
+  benchmarkSteps.push({
+    name: 'unclass',
+    fn: function() {
+      // remove all classes
+      this.setTodosValuesWithSeed(NaN);
+      $scope.$apply();
+    }.bind(this)
+  });
+
+  benchmarkSteps.push({
+    name: 'class',
+    fn: function() {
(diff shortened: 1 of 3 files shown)


benchmarks/ng-class-bp/app.js
```diff
@@ -0,0 +1,99 @@
...
+
+app.controller('DataController', function DataController($scope) {
+
+  this.numberOfTodos = 1000;
+  this.implementation = 'tableOptimized';
+  this.completedPeriodicity = 3;
+  this.importantPeriodicity = 13;
+  this.urgentPeriodicity = 29;
+  this.todos = buildTodos(100);
+  configureTodos(this.todos, 0, this);
```
gkalpak: Are these two lines necessary (since we have the `setup` step)?

benchmarks/ng-class-bp/main.html
```diff
@@ -0,0 +1,136 @@
...
+    <h3>Parameters</h3>
+    <p>
+      <label>Number of todos</label><br>
+      <input type="number" ng-model="benchmark.numberOfTodos">
+    </p>
+    <p>
+      Implementation:<br>
+      <label>
+        <input type="radio" ng-model="benchmark.implementation" value="tableOptimized">
+        Table optimized <code>ng-class="'success' && todo.completed"</code>
```
gkalpak: You probably meant `todo.completed && 'success'` :smiley:

gkalpak (commented):
Nice! It is obviously not exhaustive (e.g. no array or one-time bindings or alternating `$index`es), but it covers most common usecases (and is of course much better than what we (didn't) have before) 👍

So, if I understand correctly, #15228 is indeed a major improvements in most cases (but a slight de-opt in some). I think we can live with that, given the benefits.

💯 on putting this together and doing the comparisons!

benchmarks/ng-class-bp/main.html
drpicox: Yes, indeed, I was too quick commiting :angel:

benchmarks/ng-class-bp/app.js
drpicox: They are helpful.
These lines helped me to make sure that the benchmark does what it means to do, and they are also helpful because now it shows a small preview of the selection that you have done: list, table, single value.
And I do not rebuild the list of 100 todos at the end of the tests to induce a refresh: the list/table is empty and it is restored doing refresh.

drpicox:
I like the result, I usually write guides or check the code in critical zones making sure that my colleagues are using optimized ngClass values. With #15228 seems that it will work well from time 0.

I have rewritted the code a little bit, and fixed the "spelling" problems.

Another question:
- is there any existing tool to execute benchpress from command line? (like karma does with tests)
It would be nice to have benchmarks running automatically and follow the chart of performance changes through the time.

PR #15244, opened 2016-10-11
fix($resource): Delete $cancelRequest from toJSON()

hereblur:
Bug fix

result of ngResource.toJSON still include $cancelRequest

remove delete $cancelRequest from ngResource.toJSON's result.

No

- [ Y ] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ N ] Tests for the changes have been added (for bug fixes / features)
- [ N ] Docs have been added / updated (for bug fixes / features)

**Other information**:

### DIFF
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -29,9 +29,6 @@ describe('basic usage', function() {
         }
       }
 
-    },
-    {
-      cancellable: true
     });
     callback = jasmine.createSpy('callback');
   }));
@@ -740,7 +737,7 @@ describe('basic usage', function() {
     expect(person2).toEqual(jasmine.any(Person));
   });
 
-  it('should not include $promise, $resolved when resource is toJson\'ed', function() {
+  it('should not include $promise and $resolved when resource is toJson\'ed', function() {
     $httpBackend.expect('GET', '/CreditCard/123').respond({id: 123, number: '9876'});
     var cc = CreditCard.get({id: 123});
     $httpBackend.flush();
@@ -757,14 +754,22 @@ describe('basic usage', function() {
   });
 
   it('should not include $cancelRequest when resource is toJson\'ed', function() {
-    $httpBackend.expect('GET', '/CreditCard/123').respond({id: 123, number: '9876'});
-    var cc = CreditCard.get({id: 123});
+    $httpBackend.whenGET('/CreditCard').respond({});
+
+    var CreditCard = $resource('/CreditCard', {}, {
+      get: {
+        method: 'GET',
+        cancellable: true
+      }
+    });
+
+    var creditCard = CreditCard.get();
+
+    expect(creditCard.$cancelRequest).toBeDefined();
+
     $httpBackend.flush();
 
-    expect(cc.$cancelRequest).toBeDefined();
-
-    var json = cc.toJSON();
-
+    var json = creditCard.toJSON();
     expect(json.$cancelRequest).not.toBeDefined();
   });
 
--- a/test/ngResource/resourceSpec.js
+++ b/test/ngResource/resourceSpec.js
@@ -737,7 +737,7 @@ describe('basic usage', function() {
     expect(person2).toEqual(jasmine.any(Person));
   });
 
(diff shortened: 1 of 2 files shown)


gkalpak:
This for the PR. This was an intentional omission. Since `$cancelRequest` is always a function (and `JSON.stringify()` ignores functions), it shouldn't make any difference.

I am going to close this, but if you have a special usecase that breaks because of `$cancelRequest()`, please share below and I'll be happy to investigate.

hereblur:
In my opinion, from the method name `toJSON()` we should be expected for a valid and ready to use of a  JSON object.

In many cases, people need the literal object from server without extra keys.
I got wrong number when I do `Object.keys( result ).length`.

I think one should not have to do `JSON.parse(JSON.stringifyr(result))` to get it works.

Thank you.

gkalpak:
`toJSON` is a special method recognized by `JSON.stringify()` and is only intended to be used implicitly by that - i.e. not directly by the user ([MDN docs][1]).

That said, I guess it doesn't hurt to remove one more property ourselves (instead of relying on `JSON.stringify()`), so why not?

Reopening, but we need some tests for this (if we are making the change, we should make sure there will be no regressions in the future).

gkalpak:
Hm...I just realized `toJSON` is publicly documented and the docs state that the returned object does not contain any of the extra properties. So, we should definitely fix that. 

Thx @hereblur bringing it up.

hereblur:
I signed it!

hereblur:
@gkalpak Cool!, Thank you.
And I just pushed the test.

test/ngResource/resourceSpec.js
```diff
@@ -750,6 +750,7 @@ describe('basic usage', function() {
     var json = JSON.parse(angular.toJson(cc));
     expect(json.$promise).not.toBeDefined();
     expect(json.$resolved).not.toBeDefined();
+    expect(json.$cancelRequest).not.toBeDefined();
```
gkalpak: Can you also add an expectation that is was there before? (Similar to what we have for `$promise`/`$resolved`.)

gkalpak (changes requested):

test/ngResource/resourceSpec.js
gkalpak: I just realized that this test uses `JSON.parse(angular.toJson(cc))`, which means it would pass even without this PR.
Can you add a new test (similar to this one) that uses `cc.toJSON()` directly? (Leave this one as is - it doesn't hurt.)

gkalpak (changes requested):

test/ngResource/resourceSpec.js
```diff
@@ -737,7 +740,7 @@ describe('basic usage', function() {
     expect(person2).toEqual(jasmine.any(Person));
   });
 
-  it('should not include $promise and $resolved when resource is toJson\'ed', function() {
+  it('should not include $promise, $resolved when resource is toJson\'ed', function() {
```
gkalpak: Revert this change.

test/ngResource/resourceSpec.js
```diff
@@ -29,6 +29,9 @@ describe('basic usage', function() {
         }
       }
 
+    },
+    {
+      cancellable: true
```
gkalpak: Nit: It probably doesn't matter, but I would rather keep changes local, instead of affecting all tests. I.e. create a new, cancellable resource for your test.

gkalpak (changes requested):

test/ngResource/resourceSpec.js
```diff
@@ -753,6 +753,26 @@ describe('basic usage', function() {
...
+        method: 'GET',
+        cancellable: true
+      }
+    });
+
+    var creditCard = CreditCard.get();
+
+    expect(creditCard.$cancelRequest).toBeDefined();
+
+    $httpBackend.flush();
```
gkalpak: Is this necessary?

hereblur: Yes, we need to make sure It was there before remove it. same as the test next to it (for `$promise` and `$resolved`).

Issue #15126, opened 2016-09-12
Account for remaining jQuery 3 changes in jqLite


mgol:
***Note*: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**
Feature.

**What is the current behavior?**
jqLite behaves more like jQuery 2 than jQuery 3.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**
N/A

**What is the expected behavior?**
jqLite should be aligned with jQuery 3. Some of the changes have been already applied in PR #15104 but there are others (related to attributes, ready handling etc.)

**What is the motivation / use case for changing the behavior?**
jqLite is a slim implementation of subset of the jQuery APIs. In the parts it implements it should align with the latest jQuery as much as possible.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**
N/A

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
N/A

PR #15249, opened 2016-10-12
refactor(*): separate jqLite/compile/sce camelCasing logic

### LINKED ISSUE #7744: ngStyle ignores properties with underscore only when jQuery is installed
Default expected behaviour:
 - `ng-style="{background_color:'blue'}"` translates to `style="background-color: blue"`
 - `ng-style="{backgroundColor:'blue'}"` translates to `style="background-color: blue"`
 -  in Plunker(no jQuery): http://plnkr.co/edit/lCBUkbKeqLxtkTpHx0bo?p=preview

When jQuery is installed, the first declaration via undescope is ignored, the output is the following:
 - `ng-style="{background_color:'blue'}"` translates to nothing
 - `ng-style="{backgroundColor:'blue'}"` translates to `style="background-color: blue"`
 - in Plunker(with jQuery): http://plnkr.co/edit/G13qq3ReaMprVnI8p7sv?p=preview

Is there a workaround that would enable properties with underscore to be processed?

mgol:
Refactor

jqLite's camelCasing logic converts not only dashes but also underscores & colons; it also collapses multiple dashes etc.

jqLite's camelCasing logic is separated from the one used in compile/sce and more strict, making jqLite more similar to the jQuery behavior. This is also needed to make the new planned data key camelCasing work like in jQuery 3.

Yes.

**Other information**:

jqLite needs camelCase for it's css method; it should only convert one dash
followed by a lowercase letter to an uppercase one; it shouldn't touch
underscores, colons or collapse multiple dashes into one. This is behavior
of jQuery 3 as well.

This commit separates jqLite camelCasing from the $compile one (and $sce but
that's an internal-only use). The $compile version should behave as before.

Also, jqLite's css camelCasing logic was put in a separate function and
refactored: now the properties starting from an uppercase letter are used by
default (i.e. Webkit, not webkit) and the only exception is for the -ms- prefix
that is converted to ms, not Ms. This makes the logic clearer as we're just
always changing a dash followed by a lowercase letter by an uppercase one; this
is also how it works in jQuery.

Ref #15126
Fix #7744

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -136,22 +136,31 @@ JQLite._data = function(node) {
 function jqNextId() { return ++jqId; }
 
 
-var SPECIAL_CHARS_REGEXP = /([:\-_]+(.))/g;
-var MOZ_HACK_REGEXP = /^moz([A-Z])/;
+var DASH_LOWERCASE_REGEXP = /-([a-z])/g;
+var MS_HACK_REGEXP = /^-ms-/;
 var MOUSE_EVENT_MAP = { mouseleave: 'mouseout', mouseenter: 'mouseover' };
 var jqLiteMinErr = minErr('jqLite');
 
 /**
- * Converts snake_case to camelCase.
- * Also there is special case for Moz prefix starting with upper case letter.
+ * Converts kebab-case to camelCase.
+ * There is also a special case for the ms prefix starting with a lowercase letter.
  * @param name Name to normalize
  */
-function camelCase(name) {
-  return name.
-    replace(SPECIAL_CHARS_REGEXP, function(_, separator, letter, offset) {
-      return offset ? letter.toUpperCase() : letter;
-    }).
-    replace(MOZ_HACK_REGEXP, 'Moz$1');
+function cssKebabToCamel(name) {
+    return kebabToCamel(name.replace(MS_HACK_REGEXP, 'ms-'));
+}
+
+function fnCamelCaseReplace(all, letter) {
+  return letter.toUpperCase();
+}
+
+/**
+ * Converts kebab-case to camelCase.
+ * @param name Name to normalize
+ */
+function kebabToCamel(name) {
+  return name
+    .replace(DASH_LOWERCASE_REGEXP, fnCamelCaseReplace);
 }
 
 var SINGLE_TAG_REGEXP = /^<([\w-]+)\s*\/?>(?:<\/\1>|)$/;
@@ -633,7 +642,7 @@ forEach({
   hasClass: jqLiteHasClass,
 
   css: function(element, name, value) {
-    name = camelCase(name);
+    name = cssKebabToCamel(name);
 
     if (isDefined(value)) {
       element.style[name] = value;
--- a/src/ng/sce.js
+++ b/src/ng/sce.js
@@ -27,6 +27,13 @@ var SCE_CONTEXTS = {
 
 // Helper functions follow.
 
+var UNDERSCORE_LOWERCASE_REGEXP = /_([a-z])/g;
+
+function snakeToCamel(name) {
+  return name
+    .replace(UNDERSCORE_LOWERCASE_REGEXP, fnCamelCaseReplace);
+}
+
 function adjustMatcher(matcher) {
   if (matcher === 'self') {
     return matcher;
@@ -1054,13 +1061,13 @@ function $SceProvider() {
 
     forEach(SCE_CONTEXTS, function(enumValue, name) {
       var lName = lowercase(name);
-      sce[camelCase('parse_as_' + lName)] = function(expr) {
+      sce[snakeToCamel('parse_as_' + lName)] = function(expr) {
         return parse(enumValue, expr);
       };
-      sce[camelCase('get_trusted_' + lName)] = function(value) {
+      sce[snakeToCamel('get_trusted_' + lName)] = function(value) {
         return getTrusted(enumValue, value);
       };
-      sce[camelCase('trust_as_' + lName)] = function(value) {
+      sce[snakeToCamel('trust_as_' + lName)] = function(value) {
         return trustAs(enumValue, value);
       };
     });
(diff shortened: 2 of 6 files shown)


src/jqLite.js
```diff
@@ -136,22 +136,31 @@ JQLite._data = function(node) {
 function jqNextId() { return ++jqId; }
 
 
-var SPECIAL_CHARS_REGEXP = /([:\-_]+(.))/g;
-var MOZ_HACK_REGEXP = /^moz([A-Z])/;
+var DASH_LOWERCASE_REGEXP = /-([a-z])/g;
```
gkalpak: I wonder if should match `[A-Z]` as well.

src/ng/sce.js
```diff
@@ -27,6 +27,13 @@ var SCE_CONTEXTS = {
 
 // Helper functions follow.
 
+var UNDERSCORE_LOWERCASE_REGEXP = /_([a-z])/g;
```
gkalpak: Here too, I wonder what might break by only matching lowercase letters.

test/jqLiteSpec.js
```diff
@@ -2267,25 +2366,35 @@ describe('jqLite', function() {
...
+      expect(kebabToCamel('foo-bar')).toBe('fooBar');
+      expect(kebabToCamel('foo-bar-baz')).toBe('fooBarBaz');
+      expect(kebabToCamel('foo:bar_baz')).toBe('foo:bar_baz');
+    });
+
+    it('should covert leading dashes followed by a lowercase letter', function() {
+      expect(kebabToCamel('-foo-bar')).toBe('FooBar');
     });
 
+    it('should not convert slashes followed by a non-letter', function() {
```
gkalpak: _slashes_ --> _dashes_ (?)

mgol:
I updated the commit message, adding a breaking change info.

src/jqLite.js
mgol: We shouldn't. That's the whole point of this change!

src/ng/sce.js
mgol: Nothing. ;) This is a purely internal function, only used to iterate once through a few constant strings during the initialization.

src/jqLite.js
mgol: And I even wrote some tests to ensure that's not getting converted. ;)

test/jqLiteSpec.js
```diff
@@ -1055,6 +1055,105 @@ describe('jqLite', function() {
...
+      var jqA = jqLite(a);
+
+      jqA.css('foo', 'foo');
+      jqA.css('fooBar', 'bar');
+
+      expect(a.style.foo).toBe('foo');
+      expect(a.style.fooBar).toBe('bar');
+    });
+
+    it('should covert dash-separated strings to camelCase', function() {
```
petebacondarwin: covert -> convert

test/jqLiteSpec.js
```diff
@@ -1055,6 +1055,105 @@ describe('jqLite', function() {
...
+      jqA.css('foo-bar', 'foo');
+      jqA.css('foo-bar-baz', 'bar');
+      jqA.css('foo:bar_baz', 'baz');
+
+      expect(a.style.fooBar).toBe('foo');
+      expect(a.style.fooBarBaz).toBe('bar');
+      expect(a.style['foo:bar_baz']).toBe('baz');
+    });
+
+    it('should covert leading dashes followed by a lowercase letter', function() {
```
petebacondarwin: covert -> convert

test/jqLiteSpec.js
```diff
@@ -1055,6 +1055,105 @@ describe('jqLite', function() {
...
+
+    it('should not convert slashes followed by a non-letter', function() {
+      // jQuery 2.x had different behavior; skip the test.
+      if (isJQuery2x()) return;
+
+      var jqA = jqLite(a);
+
+      jqA.css('foo-42- -a-B', 'foo');
+
+      expect(a.style['foo-42- A-B']).toBe('foo');
```
petebacondarwin: This is a crazy case. Do we really care that this is what it does?

test/jqLiteSpec.js
```diff
@@ -1055,6 +1055,105 @@ describe('jqLite', function() {
...
+    });
+
+    it('should covert the -ms- prefix to ms instead of Ms', function() {
```
petebacondarwin: covert -> convert

test/jqLiteSpec.js
```diff
@@ -2267,25 +2366,35 @@ describe('jqLite', function() {
...
     it('should covert dash-separated strings to camelCase', function() {
-      expect(camelCase('foo-bar')).toBe('fooBar');
-      expect(camelCase('foo-bar-baz')).toBe('fooBarBaz');
-      expect(camelCase('foo:bar_baz')).toBe('fooBarBaz');
+      expect(kebabToCamel('foo-bar')).toBe('fooBar');
+      expect(kebabToCamel('foo-bar-baz')).toBe('fooBarBaz');
+      expect(kebabToCamel('foo:bar_baz')).toBe('foo:bar_baz');
+    });
+
+    it('should covert leading dashes followed by a lowercase letter', function() {
```
petebacondarwin: covert -> convert

test/jqLiteSpec.js
mgol: I don't expect keys exactly like that to be used in practice but this string tests the fact that only `-[a-z]` gets converted pretty well. I can split it into separate cases if you prefer but I thought one string should suffice.

mgol:
Typos fixed.

mgol:
@petebacondarwin Should it be `fix` instead of `refactor` as well? This is aligning with jQuery amongst other things.

mgol:
I tweaked the commit message (see the first commit).

@petebacondarwin does that look OK?

petebacondarwin:
Better!

petebacondarwin:
Landed via #15238

PR #15263, opened 2016-10-13
docs($rootScope.Scope): grammar

thorn0:


src/ng/rootScope.js
```diff
@@ -437,7 +437,7 @@ function $RootScopeProvider() {
        * If any one expression in the collection changes the `listener` is executed.
        *
        * - The items in the `watchExpressions` array are observed via standard $watch operation and are examined on every
-       *   call to $digest() to see if any items changes.
+       *   call to $digest() to see if any items change.
```
gkalpak: This makes it sound as if the act of examining itself may cause the change. I think it is more clear to say: _"if any item(s) changed"_

thorn0: What if we put it this way?

*The items in the `watchExpressions` array are observed via standard `$watch` operation. Their return values are examined for changes on every call to `$digest`.*

gkalpak: Works for me. I'll change it to what you suggested while merging.

gkalpak:
@googlebot, could you please do your CLA thing?

gkalpak:
@thorn0, could you comment on the PR. Maybe that would trigger @googlebot.

thorn0:
What's wrong with @googlebot?

PR #15264, opened 2016-10-13
fix(input): fix `step` validation for `input[number]`/`input[range]`

### LINKED ISSUE #15257: [1.5.9] NgModel set value undefined in input number with decimal
No set undifined only for 0.1, 0.2, 0.4, 0.8, 0.16, 0.32....

In 4999 - 5060 bower-angular builds

https://plnkr.co/edit/CMvhwlRFNSMGVKN0RAVH?p=preview

image

gkalpak:
Bug fix.

`step` validation (for `input[number]`/`input[range]`) fails on some occasions due to limitations of Floating Point Arithmetic (as described in #15257) and does not fully account for a `stepBase` as descussed in https://github.com/angular/angular.js/commit/9a8b8aa#commitcomment-19108436. 

It works as expected!
BTW, according to [the spec](https://www.w3.org/TR/html5/forms.html#concept-input-min-zero), when there is no `min` attribute (to determine the step base), but there is a value attribute, then the step base is set to the value of that. In our case this doesn't apply very well, because the value attribute is not compatible with `ngModel` and is essentially ignored.

No

**Other information**:
- [ ] Needs to be backported to 1.5.x for `[ng-input-range]`.

Fixes https://github.com/angular/angular.js/commit/9a8b8aa#commitcomment-19108436
Fixes #15257

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1529,13 +1529,62 @@ function parseNumberAttrVal(val) {
   return !isNumberNaN(val) ? val : undefined;
 }
 
+function isNumberInteger(num) {
+  // See http://stackoverflow.com/questions/14636536/how-to-check-if-a-variable-is-an-integer-in-javascript#14794066
+  // (minus the assumption that `num` is a number)
+
+  // eslint-disable-next-line no-bitwise
+  return (num | 0) === num;
+}
+
+function countDecimals(num) {
+  var numString = num.toString();
+  var decimalSymbolIndex = numString.indexOf('.');
+
+  if (decimalSymbolIndex === -1) {
+    if (-1 < num && num < 1) {
+      // It may be in the exponentional notation format (`1e-X`)
+      var match = /e-(\d+)$/.exec(numString);
+
+      if (match) {
+        return Number(match[1]);
+      }
+    }
+
+    return 0;
+  }
+
+  return numString.length - decimalSymbolIndex - 1;
+}
+
+function isValidForStep(viewValue, stepBase, step) {
+  // At this point `stepBase` and `step` are expected to be non-NaN values
+  // and `viewValue` is expected to be a valid stringified number.
+  var value = Number(viewValue);
+
+  // Due to limitations is Floating Point Arithmetic (e.g. `0.3 - 0.2 !== 0.1` or
+  // `0.5 % 0.1 !== 0`), we need to make sure all numbers are integers before proceeding.
+  if (!isNumberInteger(value) || !isNumberInteger(stepBase) || !isNumberInteger(step)) {
+    var decimalCount = Math.max(countDecimals(value), countDecimals(stepBase), countDecimals(step));
+    var multiplier = Math.pow(10, decimalCount);
+
+    value = value * multiplier;
+    stepBase = stepBase * multiplier;
+    step = step * multiplier;
+  }
+
+  return (value - stepBase) % step === 0;
+}
+
 function numberInputType(scope, element, attr, ctrl, $sniffer, $browser) {
   badInputChecker(scope, element, attr, ctrl);
   numberFormatterParser(ctrl);
   baseInputType(scope, element, attr, ctrl, $sniffer, $browser);
 
+  var minVal;
+  var maxVal;
+
   if (isDefined(attr.min) || attr.ngMin) {
-    var minVal;
     ctrl.$validators.min = function(value) {
       return ctrl.$isEmpty(value) || isUndefined(minVal) || value >= minVal;
     };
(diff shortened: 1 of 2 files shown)


test/ng/directive/inputSpec.js
```diff
@@ -2748,7 +2749,8 @@ describe('input', function() {
           expect($rootScope.value).toBe(30);
 
           // 0.3 - 0.2 === 0.09999999999999998
-          $rootScope.$apply('min = 0.2; step = 0.09999999999999998; value = 0.3');
+          $rootScope.$apply('min = 0.2; step = (0.3 - 0.2)');
+          helper.changeInputValueTo('0.3');
```
gkalpak: Not changing the scope value (e.g. `$rootScope.$apply('value = '0.3')`), because `ngModel` does not overwrite programmatically changed scope values unless the validity of the input changes. Depending on the order in which the validators are run and the previous state, this may lead to the scope value remaining `0.3` instead of being set to undefined.

test/ng/directive/inputSpec.js
```diff
@@ -2621,154 +2621,171 @@ describe('input', function() {
...
+          expect($rootScope.value).toBeUndefined();
+        });
 
-        // unparsable string is ignored
-        $rootScope.$apply('step = "abc"');
-        expect(inputElm).toBeValid();
-        expect($rootScope.value).toBe(10);
-        expect(inputElm.val()).toBe('10');
-        expect($rootScope.form.alias.$error.step).toBeFalsy();
+        it('should correctly validate even in cases where `(x*y % x !== 0)`', function() {
```
Narretz: Isn't this case an implementation detail? Can we give a better description?

src/ng/directive/input.js
```diff
@@ -1529,13 +1529,62 @@ function parseNumberAttrVal(val) {
...
+
+  return numString.length - decimalSymbolIndex - 1;
+}
+
+function isValidForStep(viewValue, stepBase, step) {
+  // At this point `stepBase` and `step` are expected to be non-NaN values
+  // and `viewValue` is expected to be a valid stringified number.
+  var value = Number(viewValue);
+
+  // Due to limitations is Floating Point Arithmetic (e.g. `0.3 - 0.2 !== 0.1` or
```
Narretz: is -> in/of

src/ng/directive/input.js
```diff
@@ -1529,13 +1529,62 @@ function parseNumberAttrVal(val) {
...
+  // `0.5 % 0.1 !== 0`), we need to make sure all numbers are integers before proceeding.
```
Narretz: need to convert all numbers to integers?

test/ng/directive/inputSpec.js
```diff
@@ -2621,154 +2621,171 @@ describe('input', function() {
...
-      it('should validate', function() {
-        $rootScope.step = 10;
-        $rootScope.value = 20;
-        var inputElm = helper.compileInput('<input type="number" ng-model="value" name="alias" step="{{step}}" />');
 
-        expect(inputElm.val()).toBe('20');
-        expect(inputElm).toBeValid();
-        expect($rootScope.value).toBe(20);
-        expect($rootScope.form.alias.$error.step).toBeFalsy();
+    forEach({
```
gkalpak: I want to repeat the `describe` block. `they` is for repeating `it` blocks.

test/ng/directive/inputSpec.js
gkalpak: It short of is, but `value % step === 0` is the obvious way to implement it and it is the exact implementation that leads into errors. I mean we wouldn't have this issues if we had implemented it differently, but we want to explicitly test that the obvious implementation falls short.

That said, I am always open to suggestions :wink:

gkalpak:
@Narretz, PTAL.

test/ng/directive/inputSpec.js
Narretz: Maybe "in cases where the JS floating point arithmetic fails"?

test/ng/directive/inputSpec.js
gkalpak: It will be in `describe`'s...description, so it is visible.

test/ng/directive/inputSpec.js
gkalpak: Works for me.

test/ng/directive/inputSpec.js
gkalpak: Changed.

test/ng/directive/inputSpec.js
gkalpak: Changed.

Narretz:
LGTM

PR #15269, opened 2016-10-14
chore(doc-gen, docs-app): create plnkr examples with correct Angular …

### LINKED ISSUE #15267: docs: mismatch between Angular version in embedded examples and plnkr examples on master
See http://stackoverflow.com/questions/40043549/ng-switch-when-separator-not-working-in-angularjs/40044632#40044632

Maybe this should be fixed in dgeni-packages/examples.

The plnkr opener requests the manifest.json, which contains the files that should appear in the plnkr. And the files are generated by dgeni based on the deployments. So maybe we need an extra plnkr deployment?

Narretz:
docs-gen / app fix

The plnkr examples include the latest stable Angular version instead of the snapshot version when opened from snapshot docs.

Fixes the plnkr examples generated from snapshot docs

- ~[ ] Tests for the changes have been added (for bug fixes / features)~

**Other information**:

…version

When the docs are based on the snapshot, the plnkr examples must use the snapshot files
from code.angularjs.org

Fixes https://github.com/angular/angular.js/issues/15267

### DIFF
--- a/docs/app/src/examples.js
+++ b/docs/app/src/examples.js
@@ -181,9 +181,8 @@ angular.module('examples', [])
           filePromises.push($http.get(exampleFolder + '/' + filename, { transformResponse: [] })
             .then(function(response) {
 
-              // The manifests provide the production index file but Plunkr wants
-              // a straight index.html
-              if (filename === 'index-production.html') {
+              // Plunkr needs an index file that's simply named index.html
+              if (filename === 'index-plnkr.html') {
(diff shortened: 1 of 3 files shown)


petebacondarwin:
I didn't look at the deployment file in detail but appears to LGTM. Let's suck it and see!

Narretz:
👍 The deployment file is basically copied from the index.template.html, and only the cdn variable has been adjusted. So it should be good.

PR #15272, opened 2016-10-15
docs(ngModel.NgModelController): describe your change...

juboba:

Bug fix: The model should be cleared out only if we don't allow rollback.

Currently, the model is always being cleared out to ''

The new behaviour matches the docs.

No

**Other information**:

If there's no else, then the same will happen in both inputs (the model value is overwritten).

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -458,8 +458,9 @@ NgModelController.prototype = {
    *           e.preventDefault();
    *           if (rollback) {
    *             $scope.myForm[value].$rollbackViewValue();
+   *           } else {
+   *             $scope.model[value] = '';
    *           }
-   *           $scope.model[value] = '';
    *         }
    *       };
    *     }]);
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -451,16 +451,15 @@ NgModelController.prototype = {
    *     angular.module('cancel-update-example', [])
    *
    *     .controller('CancelUpdateController', ['$scope', function($scope) {
-   *       $scope.model = {};
+   *       $scope.model = {value1: '', value2: ''};
    *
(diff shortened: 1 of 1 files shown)


juboba:
I signed it!

gkalpak:
Although the example is broken, this fix is not correct. As described above the example, what it tries to showcase is setting the model (i.e. `scope.value1`) to some value programmatically when there is a pending update.

So, in both cases, it must set the model to an empty string and the difference is that when your don't cancel the pending update, it will overwrite the empty string you set once the field is blurred.

In order to see that, the model value has to already be the empty string, which is not the case in this example. That's why it seems/is broken.

So, the proper fix is initializing both models to an empty string inside the controller:

```js
$scope.model = {value1: '', value2: ''};
```

--
@juboba, would you like to update your PR with the above change?

juboba:
@gkalpak yes, I'd like to update my PR, but I'm checking in this in plnkr:

https://plnkr.co/edit/ZtqpkoOhHWdJQwOTF9mY?p=preview

You can see the value is being reset to the previous one, but the model is then set to '', so the viewValue is also changed to ''

gkalpak:
If you remove the `alert` (which takes the focus and causes the field to blur prematurly) it works as expectd for me.

Narretz:
@juboba Are you planning on completing the changes for this PR?

juboba:
Hi Martin,

I'm confused with Georgios Kalpakas' reply. Do you agree that the expected
result is obtained when making his changes?

"Injustice anywhere is a threat to justice everywhere." -- Martin Luther
King, Jr

Narretz:
Yes, I think for the purpose of this example, @gkalpak's explanation is correct. It's a very specific scenario that shows that you need $rollBackViewValue in some cases when you have non-immediate scope updates.

juboba:
I updated my commit according to @gkalpak's explanation.

PR #15274, opened 2016-10-16
fix($parse): allow watching object/array literals

jbedard:
This moves the object/array-literal watching logic from $compile to $parse so it can by used by watchers anywhere (not only bindings).

By moving this to `$parse` we can potentially improve it more. Maybe pushing the `equals` check into `expressionInputDirtyCheck` for literals, then we can avoid creating the object and avoid calling `equals` on the full object each digest?

bug fix

watching object/array literals throws infdig if any inputs are non-simple values

watching object/array literals does what compile previously did to solve this

no

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3522,7 +3522,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               }
               recordChanges(scopeName, newValue, oldValue);
               destination[scopeName] = newValue;
-            }, deepWatch);
+            });
 
             removeWatchCollection.push(removeWatch);
             break;
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1826,7 +1826,7 @@ function $ParseProvider() {
             oldInputValueOf = newInputValue && getValueOf(newInputValue);
           }
           return lastResult;
-        }, listener, objectEquality, prettyPrintExpression);
+        }, listener, objectEquality || parsedExpression.literal, prettyPrintExpression);
       }
 
       var oldInputValueOfValues = [];
@@ -1852,7 +1852,7 @@ function $ParseProvider() {
         }
 
         return lastResult;
-      }, listener, objectEquality, prettyPrintExpression);
+      }, listener, objectEquality || parsedExpression.literal, prettyPrintExpression);
     }
 
     function oneTimeWatchDelegate(scope, listener, objectEquality, parsedExpression, prettyPrintExpression) {
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -3117,6 +3117,29 @@ describe('parser', function() {
           scope.$digest();
           expect(objB.value).toBe(scope.input);
         }));
+
+        it('should support watching literals', inject(function($parse) {
+          var lastVal = NaN;
+          var callCount = 0;
+          var listener = function(val) { callCount++; lastVal = val; };
+
+          scope.$watch('{val: val}', listener);
+
+          scope.$apply('val = 1');
+          expect(callCount).toBe(1);
+          expect(lastVal).toEqual({val: 1});
+
+          scope.$apply('val = []');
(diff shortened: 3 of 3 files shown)


gkalpak:
This sounds more like a feature than a fix.

src/ng/compile.js
```diff
@@ -3508,21 +3508,20 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
...
             parentGet = $parse(attrs[attrName]);
-            var deepWatch = parentGet.literal;
 
             var initialValue = destination[scopeName] = parentGet(scope);
             initialChanges[scopeName] = new SimpleChange(_UNINITIALIZED_VALUE, destination[scopeName]);
 
             removeWatch = scope.$watch(parentGet, function parentValueWatchAction(newValue, oldValue) {
               if (oldValue === newValue) {
-                if (oldValue === initialValue || (deepWatch && equals(oldValue, initialValue))) {
+                if (equals(oldValue, initialValue)) {
```
gkalpak: This is different that what we had before. `equals(oldValue, initialValue)` might be true even if `oldValue !== initialValue` for non-literals. In that case, we don't want to return.

This line shouldn't change afaict.

gkalpak:
I wonder how much people would abuse this, without realizing it is inefficient :)

gkalpak (changes requested):

src/ng/compile.js
jbedard: I was hoping to remove all "deepWatch" logic from here. I can just revert it, but any other ideas?

jbedard:
I think this is a fix because today `$watch("{foo: bar}")` works sometimes but not others. Basically if `bar` is a simple value it works, but as soon as `bar` becomes an object of any type it breaks.

That said... I think I would prefer a BC where we stop using `equals` for literals and instead use `$watchGroup`.  So `$watch("{a: x, b: y}")` would essentially be `$watchGroup(["x", "y"])` instead of a deep watch like today.  This would improve performance when those objects are large and also stop invoking the watcher just because a child property (that we might not care about) changed.

src/ng/compile.js
gkalpak: I don't have any better idea 😞

The thing is we need to call `$onChanges` synchronously (while the watchAction will only fire asynchronously), so we call `$onChanges` manually. When the watchAction does fire for the first time we need to determine if the value has changed - i.e. if `initialValue` is equal to `oldValue` in the `$watch` sense. That's why we need to differentiate between literals (compared using `equals`) and non-literals (compared using `===`).

jbedard: I've undone this and added tests in a separate PR: #15300

src/ng/compile.js
```diff
@@ -3522,7 +3522,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               }
               recordChanges(scopeName, newValue, oldValue);
               destination[scopeName] = newValue;
-            }, deepWatch);
```
gkalpak: It probably makes sense to rename `deepWatch` now that isn't not used for deep-watching :smiley:

gkalpak:
It still feels more like a feature to me. It is ony by chance that "shallow" literals started working at some point. :smiley:

LGTM (but it would be good for someone else to take a look as well)

jbedard:
I think I'd actually prefer to close this in favour of only doing #15301 in 1.7.

This PR makes literals consistent by expanding the deep-equals functionality of one-way-bindings to all watched literals. Where #15301 makes literals consistent by removing the deep-equals and instead (simple) watching only the variables within the literal. So we probably shouldn't expand the use of deep-equals if we plan to land #15301 and remove it...

gkalpak (approved):

gkalpak:
Works for me. Feel free to go ahead and close this in favor of #15301.

jbedard:
Closing then... also note that I think the first commit in #15301 could go into 1.6, but I'll comment on that there...

PR #15276, opened 2016-10-17
refactor($parse): remove unused expression arguments

jbedard:
Not sure if I'm missing some...?


gkalpak:
Not sure if you are missing some, but the ones you got lgtm (as long as travis is happy).

PR #15281, opened 2016-10-17
chore(docs-app): allow switching to newer versions

Narretz:
Feature for the docs app

You cannot switch to a newer version from an older version in the docs app.

For docs app versions that contain this commit, switchting is possible. The docs app
will request the new versions-data.json file from the snapshot folder and fill the version switcher with
this data.

No

- ~[] Tests for the changes have been added (for bug fixes / features)~

**Other information**:

With this patch, the docs app will request available versions data from the snapshot,
which are now stored in a separate versions-data.json file.

This only affects docs apps that contain this commit, so older doc apps
will still only display the versions up to their own version.

### DIFF
--- a/docs/app/src/versions.js
+++ b/docs/app/src/versions.js
@@ -2,23 +2,35 @@
 
 angular.module('versions', [])
 
-.controller('DocsVersionsCtrl', ['$scope', '$location', '$window', 'NG_VERSIONS', function($scope, $location, $window, NG_VERSIONS) {
+.controller('DocsVersionsCtrl', [
+  '$scope',
+  '$location',
+  '$window',
+  '$http',
+  '$q',
+  'NG_VERSIONS', function($scope, $location, $window, $http, $q, NG_VERSIONS) {
+
   $scope.docs_version  = NG_VERSIONS[0];
   $scope.docs_versions = NG_VERSIONS;
 
-  for (var i = 0, minor = NaN; i < NG_VERSIONS.length; i++) {
-    var version = NG_VERSIONS[i];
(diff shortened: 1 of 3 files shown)


docs/app/src/versions.js
```diff
@@ -2,36 +2,70 @@
...
+    if ($scope.docs_version.isSnapshot) {
+      reject();
+    } else {
+      resolve();
     }
-    version.isLatest = true;
-    minor = version.minor;
-  }
+  }).then(function() {
+    return $http.get('../snapshot/versions_data.json');
```
Narretz: Hmm. So for https://docs.angularjs.org (which is the snapshot) we cannot use this relative path. At the moment, it doesn't matter, because we only get the snapshot versions-data when we are not on the snapshot. 

For all other versions the root is https://code.angularjs.org/<version>/docs. In that case, this should be either `../../snapshot/versions_data.json` or `/snapshot/versions_data.json`. 

However, since we want to point docs.angularjs.org to the latest stable, an absolute url is probably the best idea. At the moment, this doesn't work because of CORS, though.

Narretz: Yes. It will also fail silently if the snapshot json data can't be requested for some reason. It's strictly an optional enhancement.

gkalpak:
:+1: This is a much needed improvement.

petebacondarwin:
@Narretz and I had a chat about this. We came up with new idea that doesn't require a change to the docs app and will work with older versions too.
It involves rewrite rules on the code.angularjs.org server...

PR #15287, opened 2016-10-17
refactor($sniffer): remove $sniffer.vendorPrefix

### LINKED ISSUE #13690: Remove $sniffer.vendorPrefix (was: Rethink the CSS prefixing strategy)
Currently, Angular tries to detect the CSS prefix the browser supports and then uses the saved one. This strategy is not ideal as currently some browsers are supporting more than one vendor prefix. The best example is Microsoft Edge that added `-webkit-` prefixes to be more Web-compatible; Firefox is doing a similar thing (but AFAIK only on mobile so far). Some of the `-webkit-`-prefixed things are now even getting into specs to sanction that they're now required for Web compatibility.

In some cases Edge even supports only the `-webkit-`-prefixed property; one example is `-webkit-appearance`.

That's why I think we shouldn't try to detect the vendor prefix & save it; we should just do the detection for each CSS property we're interested in using. That's similar to what jQuery does and it works fine but here, in Angular, it's even easier to cache it as we only need a few CSS properties.

Such a strategy change will make Angular more future-proof.

mgol:
removal

`$sniffer.vendorPrefix` is a vendor prefix of the current browser in the form used for properties in `element.style`, e.g. `Moz` or `Webkit`. It's an incorrect prefix in case of IE/Edge - it should be `ms`, not `Ms` there as this prefix is only available in the lowercase form.

The API doesn't exist.

No.

**Other information**:

Previously, Angular tried to detect the CSS prefix the browser supports and
then use the saved one. This strategy is not ideal as currently some browsers
are supporting more than one vendor prefix. The best example is Microsoft Edge
that added -webkit- prefixes to be more Web-compatible; Firefox is doing
a similar thing on mobile. Some of the -webkit--prefixed things are now even
getting into specs to sanction that they're now required for Web compatibility.

In some cases Edge even supports only the -webkit--prefixed property; one
example is -webkit-appearance.

$sniffer.vendorPrefix is no longer used in Angular core outside of $sniffer
itself; taking that and the above problems into account, it's better to just
remove it. The only remaining use case was an internal use in detection of
support for transitions/animations but we can directly check the webkit prefix
there manually; no other prefix matters for them anyway.

$sniffer is undocumented API so this removal is not a breaking change. However,
if you've previously been using it in your code, just paste the following
to get the same function:

```
var vendorPrefix = (function() {
  var prefix, prop, match;
  var vendorRegex = /^(Moz|webkit|ms)(?=[A-Z])/;
  for (prop in document.createElement('div').style) {
    if ((match = vendorRegex.exec(prop))) {
      prefix = match[0];
      break;
    }
  }
  return prefix;
})();
```

The vendorPrefix variable will contain what $sniffer.vendorPrefix used to.

Note that we advise to not check for vendor prefixes this way; if you have to
do it, it's better to check it separately for each CSS property used for the
reasons described at the beginning. If you use jQuery, you don't have to do
anything; it automatically adds vendor prefixes to CSS prefixes for you in
the .css() method.

Fixes #13690

### DIFF
--- a/src/ng/sniffer.js
+++ b/src/ng/sniffer.js
@@ -34,33 +34,15 @@ function $SnifferProvider() {
           toInt((/android (\d+)/.exec(lowercase(($window.navigator || {}).userAgent)) || [])[1]),
         boxee = /Boxee/i.test(($window.navigator || {}).userAgent),
         document = $document[0] || {},
-        vendorPrefix,
-        vendorRegex = /^(Moz|webkit|ms)(?=[A-Z])/,
         bodyStyle = document.body && document.body.style,
         transitions = false,
-        animations = false,
-        match;
+        animations = false;
 
     if (bodyStyle) {
-      for (var prop in bodyStyle) {
-        if ((match = vendorRegex.exec(prop))) {
-          vendorPrefix = match[0];
-          vendorPrefix = vendorPrefix[0].toUpperCase() + vendorPrefix.substr(1);
-          break;
-        }
-      }
-
-      if (!vendorPrefix) {
-        vendorPrefix = ('WebkitOpacity' in bodyStyle) && 'webkit';
-      }
-
-      transitions = !!(('transition' in bodyStyle) || (vendorPrefix + 'Transition' in bodyStyle));
-      animations  = !!(('animation' in bodyStyle) || (vendorPrefix + 'Animation' in bodyStyle));
-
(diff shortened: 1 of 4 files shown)


petebacondarwin:
LGTM

mgol:
The tests were failing in old WebKit, I fixed the PR. PTAL.

petebacondarwin:
LGTM

PR #15288, opened 2016-10-18
fix(input[radio]): use strict comparison when evaluating checked-ness

### LINKED ISSUE #15283: Confusing Type Coercion with ng-value
When a model value is `null` or `undefined`, a radio button `<input>` with `ng-value="false"` won't be shown as checked.

When the model value is `""` or `0`, a radio button `<input>` with `ng-value="false"` **will** be shown as checked.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
http://plnkr.co/edit/oY0YipxdTJtjjZDDCDDu?p=preview

I understand the intent here (need for a "blank" value), but if type coercion is going to happen, shouldn't we either go all in, or not do it at all? This "halfway" type coercion is confusing. **There should be some sort of option for `ng-value` that lets you turn on/off strict type coercion**.

Narretz:
Fix

When using input[radio], the checked status is determined by doing
a non-strict comparison between the value of the input and the ngModel.$viewValue. This can
lead to inconsistencies when using values such as 0 and false.

Yes

**Other information**:

Closes #15283

BREAKING CHANGE:

When using input[radio], the checked status is now determined by doing
a strict comparison between the value of the input and the ngModel.$viewValue.
Previously, this was a non-strict comparison (==).

This means in the following examples the radio is no longer checked:

```
  <!-- this.selected = 0 -->
  <input type="radio" ng-model="$ctrl.selected" value="0" >

  <!-- this.selected = 0; this.value = false; -->
  <input type="radio" ng-model="$ctrl.selected" ng-value="$ctrl.value" >
```

The migration strategy is to convert values that matched with non-strict
conversion so that they will match with strict conversion.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1761,9 +1761,7 @@ function radioInputType(scope, element, attr, ctrl) {
     if (doTrim) {
       value = trim(value);
     }
-    // Strict comparison would cause a BC
-    // eslint-disable-next-line eqeqeq
-    element[0].checked = (value == ctrl.$viewValue);
+    element[0].checked = (value === ctrl.$viewValue);
   };
 
   attr.$observe('value', ctrl.$render);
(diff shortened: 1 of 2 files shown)


mgol:
Milestone moved back to `1.6.0-rc.1`.

Issue #15268, opened 2016-10-14
After initial values, $onChanges is called with an incorrect `this`


jonathantyates:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
$onChanges is being called with an incorrect this after the initial value changes.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**
http://plnkr.co/edit/WbpIZLsDfmqjmMvXa3Hp?p=preview

**What is the expected behavior?**
They should be called with the same this.

**What is the motivation / use case for changing the behavior?**
To give it more consistent expected behavior.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**
I am only aware of 1.5.9 being broken, though previous iterations might be too.
Chrome is affected.

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
None.

Narretz:
Your plnkr throws an error (class constructors must be invoked with |new|), I guess this is not intended?

jonathantyates:
I am not getting this error in Chrome 53.  What browser are you using?

Narretz:
FF 49. It's because you are using a class as a Service. This is currently not well supported, because we cannot detect this in Firefox, so we are not calling new on it.

jonathantyates:
Since I don't have to worry about FF, here are some Chrome solutions:

1 ) If you change it to a one way binding '<',  the scoping is corrected.
2) If you set a flag inside of $onChanges (this.changes_ = changes), you can use it inside of $doCheck when it is called later in the digest cycle.

So my current plan is just to swap to one-way bindings.  Do you have any other recommendations for using ES6 classes?  Also, why would '<' work while '@' wouldn't?

jonathantyates:
Also, being able to reference the same instance in $onChanges and $doCheck even though they have different attributes makes no sense.

gkalpak:
I need to investigate more the exact roots of the problem, but it is related to the combination of pre-assigning bindings and classes. E.g. if you use a function (as the controller), it works as expected.

In 1.5.9 and 1.6+ you will be able to disable bindings pre-assignment, which also gets rid of the problem. ([Demo][1])

--
[...]

PR #15298, opened 2016-10-19
fix($compile): Cleanup attribute-binding observers

jonathantyates:
Bug fix

https://github.com/angular/angular.js/issues/15268

Cleans up attributes observer on change.

No.

**Other information**:

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -4378,6 +4378,46 @@ describe('$compile', function() {
             });
           });
 
+          it('should reinitialize watchers on attribute change', function() {
+            var constructorSpy = jasmine.createSpy('constructor');
+
+            class TestDirective {
+              constructor() {
+                this.constructorProp = true;
+                this.constructorSpy = constructorSpy;
+              }
+
+              $onChanges(changes) {
+                this.constructorSpy(this.constructorProp);
+              }
+            }
+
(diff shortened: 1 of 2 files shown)


gkalpak (changes requested):
The implementation looks good. The testcase doesn't seem to test the new change, though.

Off the top of my head, you can exercise the appropriate path by having a controller that returns an object from its constructor. E.g. something like:

``` js
function TestController() {
  return {
    $onChanges: ...
  };
}

TestController.prototype.$onChanges = ...
```

In thie above case, both `$onChanges` functions should continue to get called without the fix. With your fix, only the returned one should continue to get called onve the controller is initialized.

test/ng/compileSpec.js
```diff
@@ -4378,6 +4378,38 @@ describe('$compile', function() {
             });
           });
 
+          it('should reinitialize watchers on attribute change', function() {
+            var constructorSpy = jasmine.createSpy('constructor');
+            var prototypeSpy = jasmine.createSpy('prototype');
+
+            function TestDirective() {
```
gkalpak: Nit: Rename to `TestController`.

gkalpak:
LGTM 👍 <sub>(I will rename the controller while merging.)</sub>

✨ Thx for finding the bug and fixing it too! ✨

PR #15301, opened 2016-10-20
fix+perf($parse): allow watching array/object literal values, disable deep watch for one-way bindings

jbedard:
I think this is a better alternative to #15274, but has a breaking change (which didn't break any tests!).

Bug fix, perf

Watching literals doesn't always work, literals passed to components are deep watched to work around this.

Watching literals now does a watch of the inputs.

YES - see below or second commit

BREAKING CHANGE (65ff861):
Previously when a literal value was passed into a directive/component via one-way binding it would be watched with a deep watcher.

For example, for `<my-component input="[a]">`, a new instance of the array would be passed into the directive/component (and trigger $onChanges) not only if `a` changed but also if any sub property of `a` changed such as `a.b` or `a.b.c.d.e` etc.

This also means a new but equal value for `a` would NOT trigger such a change.

Now literal values use a non-deep watch similar to other directive/component one-way bindings. Changes are only trigger when the value itself changes.

Avoiding deep watchers for array/object literals will improve watcher performance of all literals passed as one-way bindings, especially those containing references to large/complex objects.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3508,14 +3508,14 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             if (optional && !attrs[attrName]) break;
 
             parentGet = $parse(attrs[attrName]);
-            var deepWatch = parentGet.literal;
+            var isLiteral = parentGet.literal;
 
             var initialValue = destination[scopeName] = parentGet(scope);
             initialChanges[scopeName] = new SimpleChange(_UNINITIALIZED_VALUE, destination[scopeName]);
 
             removeWatch = scope.$watch(parentGet, function parentValueWatchAction(newValue, oldValue) {
               if (oldValue === newValue) {
-                if (oldValue === initialValue || (deepWatch && equals(oldValue, initialValue))) {
+                if (oldValue === initialValue || (isLiteral && equals(oldValue, initialValue))) {
                   return;
                 }
                 oldValue = initialValue;
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3522,7 +3522,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               }
               recordChanges(scopeName, newValue, oldValue);
               destination[scopeName] = newValue;
-            }, deepWatch);
+            });
 
             removeWatchCollection.push(removeWatch);
             break;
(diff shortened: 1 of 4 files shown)


gkalpak:
Seems like a good thing (it has to wait for 1.7.0 though).
May be worth documenting - may be not.

src/ng/compile.js
```diff
@@ -3522,7 +3522,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               }
               recordChanges(scopeName, newValue, oldValue);
               destination[scopeName] = newValue;
-            }, deepWatch);
```
thorn0: Deep watching never happens, but we still have a variable named  `deepWatch`. Isn't it a bit misleading?

petebacondarwin: OK so let's change `deepWatch` to `isLiteral`.

petebacondarwin (approved):
I think this looks good. I would like to see a test in `compileSpec.js` that demonstrates that we are no longer deep watching - just for the record.

petebacondarwin (changes requested):
So I guess I shouldn't approve this since I have asked for two changes:

* rename `deepWatch` local variable to `isLiteral`
* add test to `compileSpec.js`

jbedard:
Renamed the var and added some tests. I think those should both be squashed with the second commit.

Note that c175619d0f0acb82d24343f538e6ada12d99aa13 (first commit) can actually go into 1.6 since it only adds support for literal-watching to `$parse` (without any use of `equals` like #15274 did).  Then 65ff861f99e2db0489c23fce5813e0bc76b27c44 has the breaking change - dropping `equals` in one-way bindings.

test/ng/compileSpec.js
```diff
@@ -4259,6 +4259,78 @@ describe('$compile', function() {
           });
 
 
+          it('should trigger `$onChanges` for literal expressions when expression input value changes (simple value)', function() {
```
gkalpak: Can you also add tests verifying that `$onChanges` is not called when a property of the input value changes?

gkalpak (commented):
I wonder how this change affects watching object literals with computed property names.

petebacondarwin (approved):
Great. Thanks @jbedard 
I will squash and merge. The first commit should probably be a `feat`, right? And I agree that it can go into 1.6

PR #15302, opened 2016-10-20
docs(orderBy): clarify behavior of default comparator

arw-travela:
Doc update to clarify behavior of the default comparator starting in v1.3.7. See https://github.com/angular/angular.js/issues/15293

Issue #15293, opened 2016-10-18
orderBy: incorrect sorting of string arrays with null values


arw-travela:
Bug, seen in angular 1.5.7, chrome 53.0.2785.143 (64-bit) (Windows) and IE11, Also Chrome on OSX.

Using $filter('orderBy')(...) with a collection that contains strings, but some entries are null, results in the entries with null being sorted as if their entries contained the string `null` instead.

Expected behavior is that entries with null in the sort column would be treated as blank.

Internally replacing a nonexistent value with the string `null` for the purposes of comparison results in an unexpected sort order.

We have confirmed that in angular 1.2.27 arrays with null entries behaves with the expected behavior, so this is an undocumented regression.

The code in https://github.com/angular/angular.js/blob/master/src/ng/filter/orderBy.js#L654-L656 that sets `value = 'null'` results in the unexpected behavior.  The immediate fix isn't obvious, since that change was made to address a specific issue with unspecified predicates in 48e1f56.

gkalpak:
I don't think 48e1f56 introduced this behavior. `null` was converted to `'null'` before (see [this line][1]).

The change seems to have been introduced in a097aa9 and is documented as a breaking change in the changelog ([v1.3.7][2]).
<sub>(Fixing a regression with a breaking change in a patch release might not be ideal, but it is what it is.)</sub>

--
So, this is indeed a breaking change, but it is documented and has not happened in 1.5.x (which is the current stable branch). I can think of the following action items:

1. Add this breaking change in the [migration guide][3]. (Since it happened on a patch release, it hasn't been documented on the migration guide - only in the changelog.)

2. Point out the special handling for `null` values in the [`orderBy` docs][4].

3. Discuss changing the behavior of the default comparator wrt `null` values. (Since this would be a breaking change, it can't happen before 1.7.)

PRs are mostly welcome for (1) and (2). For (3) could you open a feature request to get the discussion started, @arw-travela

[1]: https://github.com/angular/angular.js/commit/48e1f5605edd32a63318fd78f5165c7d1f1a20f9#diff-131b4811903f9fb834e9994a4386dc05L232
[2]: https://github.com/angular/angular.js/blob/master/CHANGELOG.md#137-leaky-obstruction-2014-12-15
[3]: https://docs.angularjs.org/guide/migration
[4]: https://docs.angularjs.org/api/ng/filter/orderBy#the-default-comparator

arw-travela:
[...]

PR #15304, opened 2016-10-20
docs(orderBy): Clarify behavior of default comparator

arw-travela:
Documentation updates to clarify default sort behavior. See
https://github.com/angular/angular.js/issues/15293 for more information

(These changes are slightly different than the ones in #15302, since I started from master rather than 1.3.x docs)

### DIFF
--- a/docs/content/guide/migration.ngdoc
+++ b/docs/content/guide/migration.ngdoc
@@ -1229,8 +1229,7 @@ or simply use:
 ### Filters (`orderBy`)
 
 - due to [a097aa95](https://github.com/angular/angular.js/commit/a097aa95b7c78beab6d1b7d521c25f7d9d7843d9),
-  the default orderBy comparator sorts null values (strings, objects, etc.) as if they were the string `'null'`:
-  a null string will be sorted as if it started with the letter N.
+  the default orderBy comparator now treats null values as the string 'null'
 
 
 
--- a/docs/content/guide/migration.ngdoc
+++ b/docs/content/guide/migration.ngdoc
@@ -1226,6 +1226,16 @@ or simply use:
   As communicated before, IE8 is no longer supported.
 
 
+### Filters (`orderBy`)
+
+- due to [a097aa95](https://github.com/angular/angular.js/commit/a097aa95b7c78beab6d1b7d521c25f7d9d7843d9),
+  the default orderBy comparator sorts null values (strings, objects, etc.) as if they were the string `'null'`:
+  a null string will be sorted as if it started with the letter N.  To sort nonexistent items to the last (or first,
+  if sort order is reversed) positions in the sort results, the sort predicate should use the empty string instead
+  of null. Similarly, predicate values that are `undefined` will be sorted as if they were strings starting with
+  the letter U.
+
+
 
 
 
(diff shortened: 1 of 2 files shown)


docs/content/guide/migration.ngdoc
```diff
@@ -1226,6 +1226,16 @@ or simply use:
   As communicated before, IE8 is no longer supported.
 
 
+### Filters (`orderBy`)
+
+- due to [a097aa95](https://github.com/angular/angular.js/commit/a097aa95b7c78beab6d1b7d521c25f7d9d7843d9),
+  the default orderBy comparator sorts null values (strings, objects, etc.) as if they were the string `'null'`:
```
gkalpak: Null values can't be strings, objects etc. A null value is just `null` :)

docs/content/guide/migration.ngdoc
```diff
@@ -1226,6 +1226,16 @@ or simply use:
...
+  a null string will be sorted as if it started with the letter N.  To sort nonexistent items to the last (or first,
+  if sort order is reversed) positions in the sort results, the sort predicate should use the empty string instead
+  of null. Similarly, predicate values that are `undefined` will be sorted as if they were strings starting with
```
gkalpak: I don't think this applies to `undefined`. The change only affects `null` afaict.

src/ng/filter/orderBy.js
```diff
@@ -86,6 +86,9 @@
  *
  * **Note:** If you notice numbers not being sorted as expected, make sure they are actually being
  *           saved as numbers and not strings.
+ * **Note:** If null or undefined values are not being sorted as expected, convert them to the empty
```
gkalpak: What this note should mention is that `null` is treated as the string `'null'` while sorting. There doesn't need to be a special mention about `undefined`, since it follows the algorithm described above.

I would also refrain from suggesting any work arounds, because there are different ones available, each with its pros and cons (e.g. use an extra predicate to separate null from non-null value, convert  `null` to an empty string or `undefined`, use a custom comparator etc).

gkalpak (changes requested):

docs/content/guide/migration.ngdoc
gkalpak: Hm...actually things are a little more complicated: https://github.com/angular/angular.js/issues/15293#issuecomment-255190763

For this breaking change note, you need to mention the change introduced in a097aa95.
For the docs, you need to mention the current situation, as introduced in 48e1f56.

src/ng/filter/orderBy.js
arw-travela: I've changed as suggested.

src/ng/filter/orderBy.js
```diff
@@ -86,6 +86,8 @@
  *
  * **Note:** If you notice numbers not being sorted as expected, make sure they are actually being
  *           saved as numbers and not strings.
+ * **Note:** If null values are not being sorted as expected it may be due to rule number
+ *           1 above which sorts `null` as the string `'null'`.
```
gkalpak: No, it is not related to rule number 1. Normally, the type of `null` is `object`. But we sort it as if it's type were `string`.

I would write something like:

> **Note:** Normally the type of `null` is `object`. When applying the above algorithm, `null` values are converted to the string `'null'` (and thus have a type of `string`).

gkalpak (changes requested):

arw-travela:
I've made additional wording updates.  I really do want to try and keep something in the wording indicating the effect (unexpected sort order), not just the fact that nulls are converted to strings.

PR #15306, opened 2016-10-21
chore(gruntfile): check the node version before starting

petebacondarwin:

**Other information**:

We specify the node version that is required to run the build in the `.nvmrc`
file. So let's check that the current node version satisfies this and report
a helpful message if it is not.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,6 +9,14 @@ var versionInfo = require('./lib/versions/version-info');
 var path = require('path');
 var e2e = require('./test/e2e/tools');
 
+var semver = require('semver');
+var fs = require('fs');
+
+var useNodeVersion = fs.readFileSync('.nvmrc', 'utf8');
+if (!semver.satisfies(process.version, useNodeVersion)) {
+  throw new Error('Invalid node version; please use node v' + useNodeVersion);
+}
+
 module.exports = function(grunt) {
   //grunt plugins
   require('load-grunt-tasks')(grunt);


gkalpak:
How about printing a warning instead? It might be too restrictive and the version might not matter for certain tasks.

mgol:
I like it. I have sth like that in build systems of most of my non-trivial projects.

@gkalpak A warning will just be ignored in lots of output... If we want to make it a soft warning, I'd use [inquirer](https://www.npmjs.com/package/inquirer) to ask the user if they really want to run the task with an unsupported version.

mgol:
@petebacondarwin Note, though, that when we switch to Node 6 we might want to try using ES6 in our Node.js code. In that case this check would not even run in older Nodes due to parse errors.

The workaround for that would be to have `Gruntfile.js` have just this check & then require another file that would contain the "true" Gruntfile written in ES6.

petebacondarwin:
@gkalpak - generally I just run `nvm use` whenever I change folder to a new project, which just fixes up the version for me with very little overhead.

mgol:
@petebacondarwin I have a script that does it for me automatically every time I switch to a different directory: https://github.com/creationix/nvm#calling-nvm-use-automatically-in-a-directory-with-a-nvmrc-file

I don't even have to remember about `nvm use`. :)

gkalpak:
So, are you saying that you don't care about the poor souls that can't or won't use `nvm use` or zsh hooks? Fair enough :smiley:

LGTM

thorn0:
@petebacondarwin FYI `nvm` doesn't work on Windows.

petebacondarwin:
This version - https://github.com/coreybutler/nvm-windows - works for me on Windows...

gkalpak:
Me too :smiley:

PR #15310, opened 2016-10-24
Japan Yen do never use fractions

davidmartinezmarin:

**Other information**:

As seen in hundreds of threads in Internet, this issue is kind of universal when working in a website having the JP locale and currency. Japanese currency never use Fractions and they are really picky about it.


murciano:
Sorry about the commit message, I saw the guidelines after the commit...

gkalpak:

If the patch gets accepted there, you can ping us, so we pull the latest version of Google Closure I18N.

PR #15317, opened 2016-10-26
docs(guide/Conceptual Overview): describe your change...

stevanlauriault:

**Other information**:

No definition found for Table yahoo.finance.xchange

### DIFF
--- a/docs/content/guide/concepts.ngdoc
+++ b/docs/content/guide/concepts.ngdoc
@@ -384,3 +384,5 @@ built-in service provided by Angular for accessing a server backend. `$http` is 
 [`XMLHttpRequest`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest)
 and [JSONP](http://en.wikipedia.org/wiki/JSONP) transports.
 
+*This no longer works.  No definition found for Table yahoo.finance.xchange.
+


stevanlauriault:
I signed it.

Narretz:
I don't see this message, neither on the [snapshot docs](https://code.angularjs.org/snapshot/docs/guide/concepts) nor in the plnkr.

gkalpak:
I see it in the plnkr. I think it might be related to temporary issue with `datatables.org`.
Let's wait and see if the problem persists. (But if it does, we should find an alternative - just adding a note that the example does not work is not good enough imo.)

stevanlauriault:
Under Accessing the backend, shouldn't {{invoice.total(c) | currency:c}} in
index.html have an output?

Narretz:
That's right, there should be output which is blocked because the url cannot be loaded. Did you mean that originally? It helps to be specific from the start in these cases.

Narretz:
It looks like the Yahoo Finance API is gone or at least broken: http://stackoverflow.com/questions/38355075/has-yahoo-finance-web-service-disappeared-api-changed-down-temporarily

There's also no official page from Yahoo for it, so we probably have to use a different service.

Narretz:
Actually ... the API still works, but our jsonp doesn't work on plnkr in the snapshot / master build.
Reason:

- when the plnkr is generated from the snapshot build, it doesn't use the snapshot angular files, but the stable version (1.5.8 currently). Since we changed how the JSON_CALLBACK is replaced in the url in the snapshot, the example breaks because the url definition and the 1.5.8 implementation are not compatible.

Conclusion: 
- make it so that the docs for snapshot use the corresponding snapshot files in the plnkr examples. 
- Additionally, wee need to add the finance api to the $sceDelegateProvider.resourceUrlWhitelist to make it snapshot compatible (to make it work both in docs app and plnkr)

````
                          docs  |   plnkr
1.5.8                     y     |   n
snapshot                  n     |   n
snapshot (whitelisted)    ?     |   y
```

Narretz:
The fix for JSONP is here: https://github.com/angular/angular.js/pull/15336

And the issue to track the mismatch of versions with plnkr is tracked here: https://github.com/angular/angular.js/issues/15267

PR #15324, opened 2016-10-28
Add CDNJS version badge in readme

x09326:
This will add the badge to show its version on CDNJS and also link to its page on CDNJS!


x09326:
I signed it!

gkalpak:
Since this is not something the team is actively involved in/in control of, I would rather not have this badge listed. I'll leave it open for a while to see what others think.

Narretz:
Also not a big fan. The angular versions are picked up by various CDNs, and the only one we are involved with is the google cdn.

petebacondarwin:
We also use unpkg for our examples.
Given that we only show the build status and not the raft of other possible badges that are available then I agree we should not land this.

PR #15326, opened 2016-10-28
docs(CONTRIBUTING.md): add note about scope wildcard

jsoref:

docs update

no hint that `*` is a valid scope -- it hints you could omit the `(...)` bit, but most commits of late don't

documentation of `*`

no

**Other information**:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -249,6 +249,8 @@ Must be one of the following:
 The scope could be anything specifying place of the commit change. For example `$location`,
 `$browser`, `$compile`, `$rootScope`, `ngHref`, `ngClick`, `ngView`, etc...
 
+You can use `*` when the change affects more than a single scope.
+
 ### Subject
 The subject contains succinct description of the change:
 


CONTRIBUTING.md
```diff
@@ -249,6 +249,8 @@ Must be one of the following:
 The scope could be anything specifying place of the commit change. For example `$location`,
 `$browser`, `$compile`, `$rootScope`, `ngHref`, `ngClick`, `ngView`, etc...
 
+You can use `*` if there isn't a more fitting scope.
```
gkalpak: According to my (unofficial) understanding, I would say `*` is for when the change affects more than a single scope.

Narretz:
Wasn't there a goodle doc where these guidelines originate from? Here: https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#

It doesn't mention the asterisks or leaving out the scope, maybe it's not the latest version?

jsoref:
@Narretz: @stevemao talked about that url in https://github.com/angular/angular.js/pull/12032#issuecomment-120232324 when `CONTRIBUTING.md` was modified to mention optional scope. @petebacondarwin claimed to have updated the Google Doc. Unfortunately, I can't see the revision history for that file, so I can't see what happened, afaict, the optional scope bit is not present in that document at this time.

Further, If there's a document like that, CONTRIBUTING.md should reference it too...

stevemao:
> You can use `*` if there isn't a more fitting scope.

I thought this was true... I don't think this is in any doc yet.

PR #15336, opened 2016-10-29
docs(guide/Conceptual Overview): fix external api example

Narretz:
docs

Example is broken in the snapshot because of changes to JSONP

**Other information**:

The example will still be broken when opened in a plnkr, because they use the stable version instead of the snapshot.

### DIFF
--- a/docs/content/guide/concepts.ngdoc
+++ b/docs/content/guide/concepts.ngdoc
@@ -141,7 +141,7 @@ different currencies and also pay the invoice.
           <b>Total:</b>
           <span ng-repeat="c in invoice.currencies">
             {{invoice.total(c) | currency:c}}
-          </span>
+          </span><br>
           <button class="btn" ng-click="invoice.pay()">Pay</button>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


docs/content/guide/concepts.ngdoc
```diff
@@ -318,13 +318,21 @@ The following example shows how this is done with Angular:
...
+          'https://query.yahooapis.com/v1/public/**'
+        ]);
       }]);
   </file>
   <file name="finance3.js">
     angular.module('finance3', [])
       .factory('currencyConverter', ['$http', function($http) {
         var YAHOO_FINANCE_URL_PATTERN =
-              '//query.yahooapis.com/v1/public/yql?q=select * from ' +
+              'https://query.yahooapis.com/v1/public/yql?q=select * from ' +
```
gkalpak: Is this necessary? This won't work if the plnkr is loaded ovr `http:` iirc.

gkalpak (commented):
BTW, I think we could just use `.get()` instead of `.jsonp()`.

Narretz:
True, the api sets `Access-Control-Allow-Origin:"*"`. I'll make that change.

docs/content/guide/concepts.ngdoc
Narretz: The plnkrs are loaded over https. I had to add it, otherwise it didn't match.

gkalpak:
LGTM

PR #15338, opened 2016-10-30
docs(README.md): add Pug name

Nifled:
Added a '/Pug' instead of rename Jade as a whole, due to the fact that Jade is now Pug.


Nifled:
I signed it!

PR #15340, opened 2016-10-31
docs(CONTRIBUTING.md): Updated commit types.

butla:
docs update

The commit types went out of sync with https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type
What prompted me to do this fix is that there wasn't 100% clear that test fixes (e.g. for flaky tests) should come in the "test" commits.

Commit types are the same as in Angular repo.

No.

**Other information**:
Expanded "test" to also mean test fixes, added "build" and "ci".
This is to mirror the documentation in Angular (without JS) repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -241,7 +241,7 @@ Must be one of the following:
   semi-colons, etc)
 * **refactor**: A code change that neither fixes a bug nor adds a feature
 * **perf**: A code change that improves performance
-* **test**: Adding missing tests
+* **test**: Adding missing or correcting existing tests
 * **chore**: Changes to the build process or auxiliary tools and libraries such as documentation
   generation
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -241,7 +241,9 @@ Must be one of the following:
   semi-colons, etc)
 * **refactor**: A code change that neither fixes a bug nor adds a feature
 * **perf**: A code change that improves performance
-* **test**: Adding missing tests
+* **test**: Adding missing tests or correcting existing tests
+* **build**: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
```
gkalpak: Now `build` sounds almost identical to `chore`. If we decide to bring this over from Angular 2, we should update other types accordingly.

butla: You're right, I'll update chore

Narretz:
Are build and ci really necessary?

gkalpak:
AFAICT, the idea is to only have the same commit message guidelines across (angular) projects (see #15341). I am not sure I like the idea, though, because different projects (even ones as similar as Angular 1 vs 2) have different needs. 

In the name of consistency, we would either end up having a ton of types that are irrelevant to one project or being restricted to too generic types on another.

butla:
I've added some more explanations to #15341.
If you're not happy with the changes I can scrap everything except for the mention that "test" is also for fixing tests.

butla:
@gkalpak So... any update on the issue?

Narretz:
Yes, please. We don't plan to use the additional types currently.

butla:
@Narretz Fixed.

PR #15344, opened 2016-11-01
perf(ngOptions): avoid calls to element.value

klieber:

Performance

In some cases IE11/Edge calls to element.value are very slow when the element.value has not been set. Normally, these calls are usualy 0-3 ms but in these cases it can take 200-300 ms. This can easily add 3 or more seconds to the load time on a view that has 10 or more select tags using ngOptions.

The line this pull request is changing not only suffers from the performance issue described above but it also appears to be broken. The code is checking that option.value does not equal element.value but then sets element.value to option.selectValue. I don't believe option.value is actually defined anywhere and likely it was always intended to be option.selectValue. This means that check would always be true and since this code has been this way for quite a while and is causing a performance issue I've just removed the check. This way a call to element.value is never made prior to it's value being set.

No

**Other information**:

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -591,7 +591,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
           element.label = option.label;
           element.textContent = option.label;
         }
-        if (option.value !== element.value) element.value = option.selectValue;
+        element.value = option.selectValue;
       }
 
       function updateOptions() {


petebacondarwin:
Thanks @klieber

src/ng/directive/ngOptions.js
```diff
@@ -591,7 +591,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
           element.label = option.label;
           element.textContent = option.label;
         }
-        if (option.value !== element.value) element.value = option.selectValue;
+        element.value = option.selectValue;
```
gkalpak: Interestingly, `option.value` will always be undefined, so this was a redundant.
I wonder what the intenton was, though.

klieber: @gkalpak I noticed the same thing.  That's how I new this check would be safe to remove.  My guess is it was intended to be `option.selectValue !== element.value` and the thought was that would prevent setting `element.value` if it didn't need to be.  Strange thing about that though is that this function is only ever called from the `addOptionElement` function and it creates a new element cloned from `optionTemplate` everytime . So, I don't think `element.value` will ever actually have a value prior to this line anyway.  Makes me wonder if the check a few lines above `option.label !== element.label` is necessary.

klieber:
@petebacondarwin do you know which release will this be included in?

gkalpak:
I will be included in the next 1.5.x/1.6.x releases (i.e. 1.5.9 and 1.6.0-rc.1).

klieber:
Thanks!

PR #15345, opened 2016-11-01
perf(*): don't trigger digests after enter/leave of structural directives

Narretz:
perf

Structural animations such as ngIf trigger an additional digest after $animate.leave completes

No

### DIFF
--- a/src/ng/directive/ngIf.js
+++ b/src/ng/directive/ngIf.js
@@ -115,8 +115,8 @@ var ngIfDirective = ['$animate', '$compile', function($animate, $compile) {
             }
             if (block) {
               previousElements = getBlockNodes(block.clone);
-              $animate.leave(previousElements).then(function() {
-                previousElements = null;
+              $animate.leave(previousElements).done(function(response) {
+                if (response !== false) previousElements = null;
               });
               block = null;
             }
--- a/test/ng/directive/ngSwitchSpec.js
+++ b/test/ng/directive/ngSwitchSpec.js
@@ -54,12 +54,14 @@ describe('ngSwitch', function() {
     $rootScope.name = 'shyam';
     $rootScope.$apply();
     expect(element.text()).toEqual('first:shyam, first too:shyam');
+
     $rootScope.select = 2;
     $rootScope.$apply();
     expect(element.text()).toEqual('second:shyam');
     $rootScope.name = 'misko';
     $rootScope.$apply();
     expect(element.text()).toEqual('second:misko');
+
     $rootScope.select = true;
     $rootScope.$apply();
     expect(element.text()).toEqual('true:misko');
@@ -301,7 +303,66 @@ describe('ngSwitch', function() {
   }));
 
 
+  it('should not trigger a digest after an element is removed', inject(function($$rAF, $compile, $rootScope, $timeout) {
+    var spy = spyOn($rootScope, '$digest').and.callThrough();
+
+    $rootScope.select = 1;
+    element = $compile(
+      '<div ng-switch="select">' +
+        '<div ng-switch-when="1">first</div>' +
+        '<div ng-switch-when="2">second</div>' +
+      '</div>')($rootScope);
+    $rootScope.$apply();
+
+    expect(element.text()).toEqual('first');
+
+    $rootScope.select = 2;
+    $rootScope.$apply();
+    spy.calls.reset();
+    expect(element.text()).toEqual('second');
+    // If ngSwitch re-introduces code that triggers a digest after an element is removed (in an
+    // animation .then callback), flushing the queue ensures the callback will be called, and the test
+    // fails
+    $$rAF.flush();
+
+    expect(spy).not.toHaveBeenCalled();
+    // A digest may have been triggered asynchronously, so check the queue
+    $timeout.verifyNoPendingTasks();
+  }));
+
+
+  it('should handle changes to the switch value in a digest loop with multiple value matches',
+    inject(function($compile, $rootScope) {
+      var scope = $rootScope.$new();
+      scope.value = 'foo';
+
+      scope.$watch('value', function() {
+        if (scope.value === 'bar') {
+          scope.$evalAsync(function() {
+            scope.value = 'baz';
+          });
+        }
+      });
+
+      element = $compile(
+        '<div ng-switch="value">' +
+          '<div ng-switch-when="foo">FOO 1</div>' +
+          '<div ng-switch-when="foo">FOO 2</div>' +
+          '<div ng-switch-when="bar">BAR</div>' +
+          '<div ng-switch-when="baz">BAZ</div>' +
+        '</div>')(scope);
+
+      scope.$apply();
+      expect(element.text()).toBe('FOO 1FOO 2');
+
+      scope.$apply('value = "bar"');
+      expect(element.text()).toBe('BAZ');
+    })
+  );
+
+
   describe('ngSwitchWhen separator', function() {
+
     it('should be possible to define a separator', inject(function($rootScope, $compile) {
       element = $compile(
         '<div ng-switch="mode">' +
@@ -315,9 +376,11 @@ describe('ngSwitch', function() {
       expect(element.children().length).toBe(2);
       expect(element.text()).toBe('Block1|Block2|');
       $rootScope.$apply('mode = "b"');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block1|');
       $rootScope.$apply('mode = "c"');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block3|');
     }));
@@ -336,9 +399,11 @@ describe('ngSwitch', function() {
       expect(element.children().length).toBe(2);
       expect(element.text()).toBe('Block1|Block2|');
       $rootScope.$apply('mode = ""');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block1|');
       $rootScope.$apply('mode = "c"');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block3|');
     }));
@@ -357,9 +422,11 @@ describe('ngSwitch', function() {
       expect(element.children().length).toBe(2);
       expect(element.text()).toBe('Block1|Block2|');
       $rootScope.$apply('mode = "b"');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block1|');
       $rootScope.$apply('mode = "c"');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block3|');
     }));
@@ -378,9 +445,11 @@ describe('ngSwitch', function() {
       expect(element.children().length).toBe(2);
       expect(element.text()).toBe('Block1|Block2|');
       $rootScope.$apply('mode = "b|a"');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block1|');
       $rootScope.$apply('mode = "c"');
+
       expect(element.children().length).toBe(1);
       expect(element.text()).toBe('Block3|');
     }));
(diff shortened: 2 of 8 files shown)


dcherman:
:clap:

Nitpicky, but `unnessecary` --> `unnecessary` in the commit message.  :heart:

gkalpak (changes requested):
An alternative, more radical approach (involving a BC) would be using `$$q` instead of `$q` for the `AnimateRunner` and let people manually call `$apply()` if their callback needs it. Most of the time, the animation callbacks trigger DOM-related operations/cleanup (and not touch on business logic or data), so a digest should not be necessary.

Unfortunately, such a BC has to wait for 1.7 now 😞

Narretz:
@gkalpak Using $$q and letting people do $apply manually was implemented before and was changed specifically to use digest-integrated promises.  I don't know if this was actually an issue for people, but I assume that it at least makes the behavior consistent with $http, and $q issued promises.

src/ng/directive/ngIf.js
```diff
@@ -115,7 +115,7 @@ var ngIfDirective = ['$animate', '$compile', function($animate, $compile) {
             }
             if (block) {
               previousElements = getBlockNodes(block.clone);
-              $animate.leave(previousElements).then(function() {
+              $animate.leave(previousElements).done(function() {
```
Narretz: Doh ... you are right, I'll change that.

Narretz:
@gkalpak This is ready for another review :)

src/ng/animate.js
```diff
@@ -416,8 +416,8 @@ var $AnimateProvider = ['$provide', /** @this */ function($provide) {
        * @param {Promise} animationPromise The animation promise that is returned when an animation is started.
        */
       cancel: function(runner) {
-        if (runner.end) {
-          runner.end();
+        if (runner.cancel) {
+          runner.cancel();
```
gkalpak: I think this is a good change (i.e. the right thing to do), but isn't it a breaking change.
I couldn't find the behavior documented anywhere btw.

gkalpak (commented):
LGTM with the exception of one comment/question.

src/ng/animate.js
Narretz: If it's not documented that it's not a change, so also no Breaking Change :>
Seriously though, I thought this change would only align ng.$animate with ngAnimate.$animate, but it turns out cancelling an actual animation resolves the promise. 
The reason why I change this code is because otherwise .done isn't called with `false`, so some switch tests fail. The problem is that these ngSwitch tests are "synthetic" in the sense that they never complete the "animations" they start because the rAFs aren't flushed. I think if we did that, then we didn't have to do this change. However, we would have to insert raF flushes in about 8 ngSwitch tests where they actually don't need to be there from considering the test's scope (these are not animation tests).

gkalpak: > I thought this change would only align `ng.$animate` with `ngAnimate.$animate`

Isn't it the exact same `$animate` that is being used. Only underlying pieces (e.g. `$$animateQueue`) are different with `ngAnimate`. So, whether we change it or not, `$animate` will be consistent with and without `ngAnimate`.

--
> The reason why I change this code is because otherwise .done isn't called with false, so some switch tests fail.

It sounds more like a test issue, not an implementation issue.

--
> I'd be okay with introducing promise rejection on cancellation as a breaking change for 1.7 on both ng + ngAnimate, but that would mean this perf improvemtn has to wait

Sorry, I didn't get why. Apart from some (bad) tests failing, what is the problem with using `.done()` (but keeping the `runner.end()` call above)?

Narretz: If $animate.cancel doesn't reject the promise, then .done callback handlers aren't called with false, which means in the ngSwitch done handler, I cannot distinguish between failed and successful animations, which leads to "cannot read function .end of undefined" error I had at the beginning. I then changed the code in ngSwitch to read the stored runners from the end to work around this problem, but then you said, I should check for successful animations inside done, which is how I found this problem with cancel in the first place

gkalpak: I still don't understand why it worked before but not now. Previously, `.then(cb)` was used. This is implemented in `$animate` like this (code changed for brevity):

```js
then: resolveHandler => {
  return $q((resolve, reject) => this.done(status => status === false ? reject() : resolve())).
            then(resolveHandler)
}
```

So, afaict we should be able to safely replace `.then(cb)` with `.done(status => status !== false && cb())` keeping the exact same behavior minus the extra digest (and asynchronicity) of `$q`.

Is the fact that the callback is now synchronously called causing the problem?

Narretz: No, the callbacks are never called synchronously. They are always called after at least one animation frame passes.  If we use then, there's the digest needed.
It turns out it's simply a problem with the test setup. I'll post details later.

Narretz: So, the change with calling runner.cancel on cancel is not correct (at least not in this context), it just ensures the tests pass.
The problem is that in the ngSwitch tests, the animations are never completed with $$rAF flush. 
Howevr, previously unfinished leave animations are cancelled when we switch to a new element. This is a unique behavior of ngSwitch. Now cancel calls end, which executes everything inside the _doneCallbacks *without* waiting for an rAF (this might be an issue of its own). This wasn't a problem when using `then`, because the promise is resolved on the next tick. However, when using `done`, the _doneCallbacks are called synchronously, which means when there are two unfinished leave animations, the previousLeaveAnimations array is length 1 after the first animation is removed, but the second animation tries to remove the second element (now element 1).

with then and without flush
- switch 1: nothing
- switch 2: put both switch 1 leave animations into previousAnimation
  (they are not completed / removed, because not AF is flushed)
- switch true:
-- both 1+2 are identified as previousLeaveAnimations and are cancelled, which
calls the doneCallbacks (without extra rAF) 
-- however, since we are using promises they are only invoked after the next tick
-- switch 2 leave animation is put into previousAnimation
  (it are not completed / removed, because no AF is flushed / digest is called)

with done and without flush
- switch 1: nothing
- switch 2: put both switch 1 leave animations into previousAnimation
  (they are not completed / removed, because not AF is flushed)
- switch true:
-- both 1+2 are identified as previousLeaveAnimations and are cancelled, which
calls the doneCallbacks (without extra rAF) 
-- since we are not using promises they are immediately invoked, which means the first stored
previousLeaveAnimation is removed from the array, which means the second stored animation tries
to access index 1 which doesn' exist anymore

with done && animate.cancel =>
- done callbacks are called with false, which means the stored previousAnimations are never removed

gkalpak: Nice detective work :smiley:

From what I understand though, the same failures that the test are demonstrating can happen in real apps, so it is a bug that we should fix. I.e. it is possible that the `ngSwitch` expression changes more than once as part of the same digest loop (which means before `$$rAF` has been called to perform the leave animations). 

In that case, we will hit the same problem of trying to access a value by index, while something else had modified the array by splicing elements.

What I think we should do, is pass the value itself (e.g. the animation runner) and look it up in the `previousLeaveAnimations`, instead of lokking it up by index. I.e. change [ngSwitch.js#L155-L159][1] to:

```diff
-var spliceFactory = function(array, index) {
+var spliceFactory = function(array, item) {
   return function(response) {
-    if (response !== false) array.splice(index, 1);
+    if (response !== false) arrayRemove(array, item);
   };
 };
```

And then change how we call it in [ngSwitch.js#L173][2]:

```diff
-runner.done(spliceFactory(previousLeaveAnimations, i));
+runner.done(spliceFactory(previousLeaveAnimations, runner));
```

This will have a very slight overhead, but considering that the `previousLeaveAnimations` array will be tiny (most of the time just one signle element), it will be negligible.

<sub>I wonder if a similar issue affects other structural directives, e.g. `ngInclude`...</sub>

[1]: https://github.com/Narretz/angular.js/blob/6eae31c8b906e17a2773825be86a32f33bcd5482/src/ng/directive/ngSwitch.js#L155-L159
[2]: https://github.com/Narretz/angular.js/blob/6eae31c8b906e17a2773825be86a32f33bcd5482/src/ng/directive/ngSwitch.js#L173

Narretz: Hm, we'd have to create a scenario where this can happen. I mean, if the expression changes multiple times during the digest then the watchActionFn is still only called with the result, so there should be no difference. Can you create a real-world example where this will happen?
And I don't think ngInclude, ngView, ngIf are affected because they only ever switch one element with another element.

gkalpak: @Narretz, here is an example that shows that when we replace `.then(fn)` with `.done(status => status && fn())`, bad things can happen: [POC][1]. If you remove the `$animate` decorator, everything works.

[1]: https://codepen.io/ExpertSystem/pen/dOoedG?editors=0011

Narretz: Cool, thanks! I see the error. Your fix doesn't work, though, unless I'm missing something. The problem isn't so much in the array splicer, but in the cancel loop. Because cancel triggers a synchronous call to the doneCallbacks the splicer is called on the first element, removing it, and so the next loop tries to access and index that's not there anymore.
Reading the array from the end would work.

gkalpak: I don't see why my fix would fail. It doesn't rely on the index but on the item itself (looking it up via `previousLeaveAnimations.indexOf(...)`). So, it is not affected by structural changes to the array.

Removing in reverse order should also work. (Why did I say I didn't like that?)

Narretz: Because the error isn't coming from the splicer but from the cancel loop: http://plnkr.co/edit/MPUiYm4mTr9nNwg8SRS3?p=preview (Apologies if I'm being dense, it's very late again :D)

gkalpak: Right. I think we have two problems (and my suggestion only takes care of one of them). So, in addition we should also change the cancellation code as follows:

```js
// Before:
for (i = 0, ii = previousLeaveAnimations.length; i < ii; ++i) {
  $animate.cancel(previousLeaveAnimations[i]);
}
previousLeaveAnimations.length = 0;

// After:
while (previousLeaveAnimations.length) {
  $animate.cancel(previousLeaveAnimations.shift());
}
```

--
In any case, I think it is fine to traverse the array backwards as an alternative (if it doesn't create other problems).

Narretz: But do we need the splice change? I don't say any errors without it.

test/ng/directive/ngSwitchSpec.js
```diff
@@ -333,6 +333,36 @@ describe('ngSwitch', function() {
   }));
 
 
+  it('should handle changes to the switch value in a digest loop with multiple value matches',
+    inject(function($$rAF, $compile, $rootScope, $timeout) {
```
gkalpak: Are `$$rAF` and `$timeout` needed?

test/ng/directive/ngSwitchSpec.js
```diff
@@ -54,14 +54,18 @@ describe('ngSwitch', function() {
     $rootScope.name = 'shyam';
     $rootScope.$apply();
     expect(element.text()).toEqual('first:shyam, first too:shyam');
+
     $rootScope.select = 2;
     $rootScope.$apply();
+    $$rAF.flush();
```
gkalpak: Now that 696b7a4 has been added, are these extra `$$rAF` calls necessary?

test/ng/directive/ngSwitchSpec.js
```diff
@@ -292,16 +296,80 @@ describe('ngSwitch', function() {
...
+
+    expect(spy).not.toHaveBeenCalled();
+    // A digest may have been triggered asynchronously, so check the queue
+    $timeout.verifyNoPendingTasks();
+  }));
+
+
+  it('should handle changes to the switch value in a digest loop with multiple value matches',
+    inject(function($$rAF, $compile, $rootScope, $timeout) {
```
gkalpak: Are `$$rAF` and `$timeout` necessary?

gkalpak (commented):
A couple of minor comments. LGTM otherwise.

Narretz:
Good catch, no, these flushes are not necessary. The only one necessary is in the test that ensures that no digest is triggered after the leave animation

petebacondarwin:
OK, are we happy to merge this?

gkalpak (approved):

gkalpak:
I am :smiley:

petebacondarwin:
Please do: 1.5.x and master?

PR #15346, opened 2016-11-01
fix(security): do not auto-bootstrap when loaded from an extension.

mprobst:
Extension URIs (`resource://...`) bypass Content-Security-Policy in Chrome and
Firefox and can always be loaded. Now if a site already has a XSS bug, and uses
CSP to protect itself, but the user has an extension installed that uses
Angular, an attacked can load Angular from the extension, and Angular's
auto-bootstrapping can be used to bypass the victim site's CSP protection.

Notes:
- `isAutoBootstrapAllowed` must be initialized on load, so that `currentScript`
  is set correctly.
- The tests are a bit indirect as reproducing the actual scenario is too
  complicated to reproduce (requires signing an extension etc). I have confirmed
  this to be working manually.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -97,9 +97,6 @@
   NODE_TYPE_DOCUMENT,
   NODE_TYPE_DOCUMENT_FRAGMENT
 */
-/* global
-  console
-*/
 
 ////////////////////////////////////
 
@@ -1626,8 +1623,8 @@ function angularInit(element, bootstrap) {
   });
   if (appElement) {
     if (!isAutoBootstrapAllowed) {
-      console.error('Angular: disabling automatic bootstrap. <script> protocol indicates an ' +
-          'extension, document.location.href does not match.');
+      window.console.error('Angular: disabling automatic bootstrap. <script> protocol indicates ' +
+          'an extension, document.location.href does not match.');
       return;
     }
     config.strictDi = getNgAttribute(appElement, 'strict-di') !== null;
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -97,6 +97,9 @@
   NODE_TYPE_DOCUMENT,
   NODE_TYPE_DOCUMENT_FRAGMENT
 */
+/* global
+  console
+*/
 
 ////////////////////////////////////
 
@@ -1444,6 +1447,26 @@ function getNgAttribute(element, ngAttr) {
   return null;
 }
 
+function allowAutoBootstrap(document) {
+  if (!document.currentScript) {
+    return true;
+  }
+  var src = document.currentScript.getAttribute('src');
+  var link = document.createElement('a');
+  link.href = src;
+  var scriptProtocol = link.protocol;
+  var docLoadProtocol = document.location.protocol;
+  if ((scriptProtocol === 'resource:' ||
+       scriptProtocol === 'chrome-extension:') &&
+      docLoadProtocol !== scriptProtocol) {
+    return false;
+  }
+  return true;
+}
+
+// Cached as it has to run during loading so that document.currentScript is available.
+var isAutoBootstrapAllowed = allowAutoBootstrap(window.document);
+
 /**
  * @ngdoc directive
  * @name ngApp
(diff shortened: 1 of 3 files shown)


koto:
LGTM

src/Angular.js
```diff
@@ -1602,6 +1622,11 @@ function angularInit(element, bootstrap) {
     }
   });
   if (appElement) {
+    if (!isAutoBootstrapAllowed) {
```
gkalpak: This check could be moved at the top (unless we expect anyone to load angular via `resource:`/`chrome-extension:` in a non-extention page and bootstrap it manually.

mprobst: I didn't do this intentionally - otherwise you'll get an error message even if you intend to manually bootstrap by calling `angular.bootstrap`. Which I guess is a legit use case, if somebody does inject Angular in the page but also controls it to call `angular.bootstrap` (not a security issue because they can already execute JS in the page in that case).

src/Angular.js
```diff
@@ -1602,6 +1622,11 @@ function angularInit(element, bootstrap) {
...
+      console.error('Angular: disabling automatic bootstrap. <script> protocol indicates an ' +
+          'extension, document.href does not match.');
```
mprobst: Done.

src/Angular.js
```diff
@@ -97,6 +97,9 @@
   NODE_TYPE_DOCUMENT,
   NODE_TYPE_DOCUMENT_FRAGMENT
 */
+/* global
+  console
```
gkalpak: You can avoid this by accessing `console` off `window` (not sure what is better).

src/Angular.js
gkalpak: Yeah, I thought it might be intentional. I don't expect anyone to be loading angular over `resource:` on non-extension pages, but you never know :smiley:

Hm...so, if it is a supported usecase (loading angular.js over `resource:` on non-extension pages), isn't this chage breaking? In that case, we should have a breaking change notice.

src/Angular.js
```diff
@@ -1444,6 +1444,26 @@ function getNgAttribute(element, ngAttr) {
...
+    return true;
+  }
+  var src = document.currentScript.getAttribute('src');
+  var link = document.createElement('a');
+  link.href = src;
+  var scriptProtocol = link.protocol;
+  var docLoadProtocol = document.location.protocol;
+  if ((scriptProtocol === 'resource:' ||
+       scriptProtocol === 'chrome-extension:') &&
+      docLoadProtocol !== scriptProtocol) {
```
petebacondarwin: So we do allow auto bootstrap if the protocol of the document matches that of the script?
Why is that?

mprobst: Because that's the intended use case - an extension that has a page (within the extension) that loads angular, and contains `ng-app`. We could disable that, but if we support auto bootstrap, I think we should support it within extensions, shouldn't we?

src/Angular.js
```diff
@@ -1602,6 +1622,11 @@ function angularInit(element, bootstrap) {
     }
   });
   if (appElement) {
+    if (!isAutoBootstrapAllowed) {
+      window.console.error('Angular: disabling automatic bootstrap. <script> protocol indicates ' +
+          'an extension, document.location.href does not match.');
```
petebacondarwin: I feel that this message could be clearer or that it should link to a webpage/issue that explains what is happening?

mprobst: Do you have a suggestion? This should really only happen when somebody tries to exploit the security hole, so I'm not too worried about usability here. I don't think we need docs or a website.

petebacondarwin: But it shouldn't block merging this.

src/Angular.js
mmccallum: > So I don't think we need to document anything. Most users will never notice this.

This should be documented, assuming that most users won't notice it means that when someone does notice it, there is no help available on how to go about making it work.

petebacondarwin: @mmccallum - feel free to send in a PR.

mprobst: @mmccallum FWIW, when auto bootstrap gets disabled, Angular prints out a pretty clear message in the console.

PR #15352, opened 2016-11-02
feat($compile): set preAssignBindingsEnabled to false by default

### LINKED ISSUE #15350: Flip the $compileProvider.preAssignBindingsEnabled flag to false by default

A breaking change request.

`$compileProvider.preAssignBindingsEnabled` is set to `true` by default. This means bindings are pre-assigned in component constructors.

`$compileProvider.preAssignBindingsEnabled` should be `false` by default.

The `true` value of the flag causes bindings to be pre-assigned in component constructors which introduces unneeded (now that we have the `$onInit` lifecycle hook) complexity (we can't use regular `new MyConstructor()` way of creating controllers) and doesn't work with ES6 classes.

All.

mgol:
A breaking change.

#15350

No.

Yes

- [x] Docs have been added / updated (for bug fixes / features) (docs were already there)

**Other information**:

Fixes #15350

BREAKING CHANGE: Previously, $compileProvider.preAssignBindingsEnabled was
set to true by default. This means bindings were pre-assigned in component
constructors. In Angular 1.5+ the place to put the initialization logic
relying on bindings being present is the controller $onInit method.

To migrate follow the example below:

Before:

```js
angular.module('myApp', [])
  .component('myComponent', {
    controller: 'MyController',
    template: '<div>{{ $ctrl.a }} + {{ $ctrl.b }} = {{ $ctrl.sum }}</div>',
    bindings: {
      a: '<',
      b: '<'
    }
  })
  .controller('MyController', function() {
    this.sum = this.a + this.b;
  });
```

After:
```js
angular.module('myApp', [])
  .component('myComponent', {
    controller: 'MyController',
    template: '<div>{{ $ctrl.a }} + {{ $ctrl.b }} = {{ $ctrl.sum }}</div>',
    bindings: {
      a: '<',
      b: '<'
    }
  })
  .controller('MyController', function() {
    this.$onInit = function () {
      this.sum = this.a + this.b;
    };
  });
```

If you need to support both Angular 1.4 and 1.6, e.g. you're writing a library,
you need to check for Angular version and apply proper logic. Follow the
example below:

Before:
```js
angular.module('myApp', [])
  .directive('myDirective', function() {
    return {
      controller: 'MyController',
      template: '<div>{{ $ctrl.a }} + {{ $ctrl.b }} = {{ $ctrl.sum }}</div>',
      scope: {
        a: '=',
        b: '='
      },
      bindToController: true,
      controllerAs: '$ctrl'
    };
  })
  .controller('MyController', function($scope) {
    this.sum = this.a + this.b;
  });
```

After:
```js
angular.module('myApp', [])
  .directive('myDirective', function() {
    return {
      controller: 'MyController',
      template: '<div>{{ $ctrl.a }} + {{ $ctrl.b }} = {{ $ctrl.sum }}</div>',
      scope: {
        a: '=',
        b: '='
      },
      bindToController: true,
      controllerAs: '$ctrl'
    };
  })
  .controller('MyController', function($scope) {
    this.$onInit = function() {
      this.sum = this.a + this.b;
    };
    if (angular.version.major === 1 && angular.version.minor <= 4) {
      this.$onInit();
    }
  });
```

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1379,7 +1379,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    *
    * The default value is true in Angular 1.5.x but will switch to false in Angular 1.6.x.
    */
-  var preAssignBindingsEnabled = true;
+  var preAssignBindingsEnabled = false;
   this.preAssignBindingsEnabled = function(enabled) {
     if (isDefined(enabled)) {
       preAssignBindingsEnabled = enabled;
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -171,7 +171,9 @@ describe('$compile', function() {
 
     it('should allow preAssignBindingsEnabled to be configured', function() {
       module(function($compileProvider) {
-        expect($compileProvider.preAssignBindingsEnabled()).toBe(true); // the default
+        expect($compileProvider.preAssignBindingsEnabled()).toBe(false); // the default
+        $compileProvider.preAssignBindingsEnabled(true);
+        expect($compileProvider.preAssignBindingsEnabled()).toBe(true);
         $compileProvider.preAssignBindingsEnabled(false);
         expect($compileProvider.preAssignBindingsEnabled()).toBe(false);
       });


petebacondarwin:
LGTM - in the BC message 2nd example you have 

```
  .config(function($compileProvider) {
    $compileProvider.preAssignBindingsEnabled(false);
  })
```

This should not be there right? It would affect all apps that use the library.
By putting logic in the `$onInit` function we are good whatever the setting of `preAssignBindingsEnabled` is.

mgol:
@petebacondarwin That was a leftover from my tests in a plunker. Fixed.

gkalpak:
Regarding the commit message:

1. I think it is better for the examples to be as concise as possible. E.g.

    ```js
    // Instead of:
    angular.module('myApp', [])
      .component('myComponent', {
        controller: 'MyController',
        template: '<div>{{ $ctrl.a }} + {{ $ctrl.b }} = {{ $ctrl.sum }}</div>',
        bindings: {
          a: '<',
          b: '<'
        }
      })
      .controller('MyController', function() {
        this.sum = this.a + this.b;
      });

    // I would prefer:
    .component('myComponent', {
      bindings: {value: '<'},
      controller: function() {
        this.doubleValue = this.value * 2;
      }
    })
    ```

Also, the section about supporting 1.4 and 1.6 feels a little out of place imo. It is not really related to this change (changing the default to true), but rather to us making it configurable in the first place. The exact same is true for 1.5.9+, because you don't know whether the app will have pre-assignment enabled or not. 

It is useful info and should probably be included somewhere in the docs/migration guide, but not in the commit message imo.

mgol:
@gkalpak I like the simplification in general; do you think it's not necessary to include a template?

I agree the library part would be better placed in a commit adding the `preAssignBindingsEnabled` flag; it's too late for that but I guess we can add it in a migration guide.

mgol:
@gkalpak @petebacondarwin Commit message updated.

gkalpak:
LGTM

petebacondarwin:
LGTM2 - please merge - can you create an issue for the migration doc that we mention this 1.4-1.6 library thing?

mgol:
Sure, will do.

PR #15358, opened 2016-11-03
chore(docs-gen): create plnkr examples with the correct version

Narretz:
Currently, plnkrs are opened with the cdnVersion of the built docs app, which means
that snapshot plunkers might not work because they use the cdn (stable version).

**Other information**:

Should we update this info for the other deployments too?

### DIFF
--- a/docs/config/services/deployments/production.js
+++ b/docs/config/services/deployments/production.js
@@ -1,16 +1,30 @@
 'use strict';
 
 var versionInfo = require('../../../../lib/versions/version-info');
-var cdnUrl = '//ajax.googleapis.com/ajax/libs/angularjs/' + versionInfo.cdnVersion;
+
+var googleCdnUrl = '//ajax.googleapis.com/ajax/libs/angularjs/';
+var angularCodeUrl = '//code.angularjs.org/';
+
+var cdnUrl = googleCdnUrl + versionInfo.cdnVersion;
+
+// The plnkr examples must use the code.angularjs.org repo for the snapshot,
+// and the cdn for the tagged version and, if the build is not tagged, the currentVersion.
+//
+// The currentVersion may not be available on the cdn (e.g. if built locally, or hasn't been pushed
+// yet). This will lead to a 404, but this is preferable to loading a version with which the example
+// might not work (possibly in subtle ways).
+var examplesCdnUrl = versionInfo.isSnapshot ?
+  (angularCodeUrl + 'snapshot') :
+  (googleCdnUrl + (versionInfo.version || versionInfo.currentVersion));
 
 module.exports = function productionDeployment(getVersion) {
   return {
     name: 'production',
     examples: {
       commonFiles: {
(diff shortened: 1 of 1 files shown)


docs/config/services/deployments/production.js
```diff
@@ -1,16 +1,30 @@
...
+
+// The plnkr examples must use the code.angularjs.org repo for the snapshot,
+// and the cdn for the tagged version and, if the build is not tagged, the currentVersion.
+//
+// The currentVersion may not be available on the cdn (e.g. if built locally, or hasn't been pushed
+// yet). This will lead to a 404, but this is preferable to loading a version with which the example
+// might not work (possibly in subtle ways).
+var examplesCdnUrl = versionInfo.isSnapshhot ?
+  (angularCodeUrl + 'snapshot') :
+  (googleCdnUrl + (versionInfo.version || versionInfo.currentVersion));
```
petebacondarwin: I guess if the `versionInfo.version` is not available then we are doing a local non-snapshot build. In which case the plunkers are going to be useless. Perhaps we should default to the snapshot though in that case?

petebacondarwin:
One comment, otherwise LGTM

docs/config/services/deployments/production.js
Narretz: I thought about setting the cdnVersion. Snapshot is also possible, but my idea was that in the rare case of no .version + testing in plnkr (basically only local), it's better to have the immediate feedback of a 404 script instead of a (probably) working app, that might not be 100% compatible with the angular version it runs on.

docs/config/services/deployments/production.js
```diff
@@ -1,16 +1,30 @@
...
+
+var cdnUrl = googleCdnUrl + versionInfo.cdnVersion;
+
+// The plnkr examples must use the code.angularjs.org repo for the snapshot,
+// and the cdn for the tagged version and, if the build is not tagged, the currentVersion.
+//
+// The currentVersion may not be available on the cdn (e.g. if built locally, or hasn't been pushed
+// yet). This will lead to a 404, but this is preferable to loading a version with which the example
+// might not work (possibly in subtle ways).
+var examplesCdnUrl = versionInfo.isSnapshhot ?
```
gkalpak: _isSnapshhot_ --> _isSnapshot_

gkalpak (dismissed):

docs/config/services/deployments/production.js
Narretz: Oh snap

petebacondarwin:
Other than the `snapshhot` fix it LGTM

gkalpak:
@Narretz , shouldn't this be backported to v1.5.x?

Narretz:
I think I didn't backport since the change only affects the snapshot anyway. At least on angularjs.org. But we can backport for consistency

gkalpak:
You are right, it is not necessary to backport it (theoretically). I'll backport it anyway for consistency :grin:

PR #15359, opened 2016-11-03
fix($location): don't remove multiple slashes from urls

petebacondarwin:

**Other information**:

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -30,20 +30,25 @@ function parseAbsoluteUrl(absoluteUrl, locationObj) {
   locationObj.$$port = toInt(parsedUrl.port) || DEFAULT_PORTS[parsedUrl.protocol] || null;
 }
 
+function countLeadingSlashes(url) {
+  for (var i = 0; url.charAt(i) === '/'; i++) { /* noop */}
+  return i;
+}
 
-function parseAppUrl(relativeUrl, locationObj) {
-  var prefixed = (relativeUrl.charAt(0) !== '/');
-  if (prefixed) {
-    relativeUrl = '/' + relativeUrl;
-  }
-  var match = urlResolve(relativeUrl);
-  locationObj.$$path = decodeURIComponent(prefixed && match.pathname.charAt(0) === '/' ?
-      match.pathname.substring(1) : match.pathname);
+function parseAppUrl(url, locationObj) {
+  var slashCount = countLeadingSlashes(url);
+  var slashes = url.substr(0, slashCount);
+
+  // We need to pass a url with a single leading slash to the `urlResolve` function
+  // so we strip off all the slashes and put just one special one back on
+  var match = urlResolve('/' + url.substr(slashCount));
+  // Then we put the slashes back on (removing the special one)
+  locationObj.$$path = decodeURIComponent(slashes + match.pathname.substr(1));
   locationObj.$$search = parseKeyValue(match.search);
   locationObj.$$hash = decodeURIComponent(match.hash);
 
-  // make sure path starts with '/';
-  if (locationObj.$$path && locationObj.$$path.charAt(0) !== '/') {
+  // fix up if we were passed a url with none or multiple leading slashes
+  if (locationObj.$$path && slashCount !== 1) {
     locationObj.$$path = '/' + locationObj.$$path;
   }
 }
--- a/test/ng/locationSpec.js
+++ b/test/ng/locationSpec.js
@@ -2453,9 +2453,9 @@ describe('$location', function() {
     var locationUrl, locationUmlautUrl, locationIndexUrl;
 
     beforeEach(function() {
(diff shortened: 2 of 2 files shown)


test/ng/locationSpec.js
```diff
@@ -2453,9 +2453,9 @@ describe('$location', function() {
     var locationUrl, locationUmlautUrl, locationIndexUrl;
 
     beforeEach(function() {
-      locationUrl = new LocationHtml5Url('http://server/pre/', 'http://server/pre/', 'http://server/pre/path');
-      locationUmlautUrl = new LocationHtml5Url('http://särver/pre/', 'http://särver/pre/', 'http://särver/pre/path');
-      locationIndexUrl = new LocationHtml5Url('http://server/pre/index.html', 'http://server/pre/', 'http://server/pre/path');
+      locationUrl = new LocationHtml5Url('http://server/pre/', 'http://server/pre/', '#!');
+      locationUmlautUrl = new LocationHtml5Url('http://särver/pre/', 'http://särver/pre/', '#!');
```
mprobst: Maybe consider adding tests for `http://server//pre/`? It's not a problem in this change I think, but it might be nice to get slightly more complete test coverage.

test/ng/locationSpec.js
```diff
@@ -2480,6 +2480,10 @@ describe('$location', function() {
       expect(parseLinkAndReturn(locationUrl, 'someIgnoredAbsoluteHref', '#test')).toEqual('http://server/pre/otherPath#test');
     });
 
+    it('should cope with double slashes in the path', function() {
+      expect(parseLinkAndReturn(locationUrl, 'http://server/pre///other/path')).toEqual('http://server/pre///other/path');
```
mprobst: I wonder if a more integration-style test that checks how the app URL is used to resolve a template would be useful? With this change, we still end up with multiple slashes in the URL, so we could still accidentally drop a path segment if we later parse it using `<a href>` again (not sure we do, but protecting from it would be nice).

mprobst (approved):
Looks good.

test/ng/locationSpec.js
petebacondarwin: BTW the changes in this beforeEach were only because the setup was actually wrong! The third parameter is supposed to be the hashbang prefix.

test/ng/locationSpec.js
petebacondarwin: I'll look into it...

Krzysztof also suggested that we need to deal with multiple leading backslashes too...

src/ng/location.js
```diff
@@ -30,20 +30,25 @@ function parseAbsoluteUrl(absoluteUrl, locationObj) {
...
+  var match = urlResolve('/' + url.substr(slashCount));
+  // Then we put the slashes back on (removing the special one)
+  locationObj.$$path = decodeURIComponent(slashes + match.pathname.substr(1));
   locationObj.$$search = parseKeyValue(match.search);
   locationObj.$$hash = decodeURIComponent(match.hash);
 
-  // make sure path starts with '/';
-  if (locationObj.$$path && locationObj.$$path.charAt(0) !== '/') {
+  // fix up if we were passed a url with none or multiple leading slashes
+  if (locationObj.$$path && slashCount !== 1) {
```
gkalpak: Do we need the extra leading `/` for `slashCount > 1`?

test/ng/locationSpec.js
koto: It's not just multiple backslashes. Basically `\s*[\\/]{2,}` would be a problem (but since in parseAppUrl `/` is prepended, the whitespace vector goes away, so I think changing:

```for (var i = 0; url.charAt(i) === '/'; i++)```
to 
```for (var i = 0; url.charAt(i) === '/' || url.charAt(i) === '\\' ; i++)```

would be enough to fix the bug. 

That said, I don't know if the multiple slashes are worth preserving in the beginning of a path for Angular apps. Existing apps served on those would be broken without this patch, so there is no backwards-compatibility problem and it's hard to tell if accepting "//" would not cause further errors when resolving relative URLs for the app later on. Prohibiting that overall seems like a good, conservative choice.

src/ng/location.js
petebacondarwin: Yes, because it gets stripped off at some other point in the code

gkalpak: But don't we add it back 5 lines above with `locationObj.$$path = decodeURIComponent(slashes + match.pathname.substr(1))`?

petebacondarwin: It's complicated. Try it! Tests will fail :-)

test/ng/locationSpec.js
petebacondarwin: OK, I agree that we may as well lock this down, rather than potentially passing the issue onto some other part of the code. We can always relax this error and do something similar to this PR, later, if people complain.

petebacondarwin:
Closing to put a PR that errors instead of trying to stumble along.

PR #15361, opened 2016-11-04
docs($cookiesProvider): describe your change...

coreystinson2:
Provided an example on how to use the $cookiesProvider to set default behaviour.
Many similar services support overriding the `defaults` object with a new one, where this service only supports changing individual properties one at a time.


coreystinson2:
I signed it!

PR #15365, opened 2016-11-04
Location path 2

petebacondarwin:
fix

see the commit message

see the commit message

The previous behaviour would have led to a broken app, so it is highly
unlikely that any app relied on this behaviour. So this doesn't constitute
a breaking change.

**Other information**:

Alternative solution #15359

### DIFF
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -30,13 +30,18 @@ function parseAbsoluteUrl(absoluteUrl, locationObj) {
   locationObj.$$port = toInt(parsedUrl.port) || DEFAULT_PORTS[parsedUrl.protocol] || null;
 }
 
+var DOUBLE_SLASH_REGEX = /^\s*[\\/]{2,}/;
+function parseAppUrl(url, locationObj) {
 
-function parseAppUrl(relativeUrl, locationObj) {
-  var prefixed = (relativeUrl.charAt(0) !== '/');
-  if (prefixed) {
-    relativeUrl = '/' + relativeUrl;
+  if (DOUBLE_SLASH_REGEX.test(url)) {
+    throw $locationMinErr('badpath', 'Invalid url "{0}".', url);
   }
-  var match = urlResolve(relativeUrl);
+
+  var prefixed = (url.charAt(0) !== '/');
+  if (prefixed) {
+    url = '/' + url;
+  }
+  var match = urlResolve(url);
   locationObj.$$path = decodeURIComponent(prefixed && match.pathname.charAt(0) === '/' ?
       match.pathname.substring(1) : match.pathname);
   locationObj.$$search = parseKeyValue(match.search);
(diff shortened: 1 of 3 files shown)


src/ng/location.js
```diff
@@ -144,9 +149,10 @@ function LocationHtml5Url(appBase, appBaseNoFile, basePrefix) {
     var appUrl, prevAppUrl;
     var rewrittenUrl;
 
+
     if (isDefined(appUrl = stripBaseUrl(appBase, url))) {
       prevAppUrl = appUrl;
-      if (isDefined(appUrl = stripBaseUrl(basePrefix, appUrl))) {
+      if (basePrefix && isDefined(appUrl = stripBaseUrl(basePrefix, appUrl))) {
```
petebacondarwin: interestingly I think this was actually the cause of the bug in the first place...

mgol: There are no tests for this change, are there?

test/ng/locationSpec.js
```diff
@@ -2480,6 +2480,15 @@ describe('$location', function() {
       expect(parseLinkAndReturn(locationUrl, 'someIgnoredAbsoluteHref', '#test')).toEqual('http://server/pre/otherPath#test');
     });
 
+    it('should complain if the path starts with double slashes', function() {
+      expect(function() {
+        parseLinkAndReturn(locationUrl, 'http://server/pre///other/path');
+      }).toThrowMinErr('$location', 'badpath');
+
+      expect(function() {
+        parseLinkAndReturn(locationUrl, 'http://server/pre/\\\\other/path');
```
koto: Could you add a test case with mixed slashes (`pre//\\//other/path`) ?

src/ng/location.js
```diff
@@ -30,13 +30,18 @@ function parseAbsoluteUrl(absoluteUrl, locationObj) {
   locationObj.$$port = toInt(parsedUrl.port) || DEFAULT_PORTS[parsedUrl.protocol] || null;
 }
 
+var DOUBLE_SLASH_REGEX = /\s*[\\/]{2,}/;
+function parseAppUrl(url, locationObj) {
 
-function parseAppUrl(relativeUrl, locationObj) {
-  var prefixed = (relativeUrl.charAt(0) !== '/');
+  if (url.search(DOUBLE_SLASH_REGEX) === 0) {
```
gkalpak: <sub>You could also add `^` at the beginning of `DOUBLE_SLASH_REGEX` and use  `DOUBLE_SLASH_REGEX.test(url)`?</sub>

gkalpak (approved):

test/ng/locationSpec.js
petebacondarwin: will do

src/ng/location.js
petebacondarwin: No new tests, specifically, but it doesn't break any of the current tests
And there are the new tests with the double slashes.
I couldn't think of a specific test for this change. Any ideas?

petebacondarwin:
@koto and @mprobst - any further comments before we merge this?

koto:
Looks good to me, security-wise.

mprobst:
LGTM

ivanrey:
This fix has problems with internal routing of an app. Sometimes double slash in fragments might be important to indicate an specific state.

I believe this fix should not take into account the fragment part of the URL since it's not sent to the server and breaks some applications.

gkalpak:
@ivanrey, afaict this is supposed to only check for the beginning of the client-side URL and I think it didn't work correctly before either. Do you have a specific usecase/demo you can share, that used to work before and broke as a result of this fix?

PR #15381, opened 2016-11-09
chore(docs): improve version picker

petebacondarwin:

chore

Versions in the version picker are hard-coded into each build
This means that newer versions are never shown in older versions of the docs

Now we get the version list from the npm registry as needed

Nope

**Other information**:

### DIFF
--- a/docs/app/src/versions.js
+++ b/docs/app/src/versions.js
@@ -1,37 +1,131 @@
 'use strict';
+/* global console */
 
 angular.module('versions', [])
 
-.controller('DocsVersionsCtrl', ['$scope', '$location', '$window', 'NG_VERSIONS', function($scope, $location, $window, NG_VERSIONS) {
-  $scope.docs_version  = NG_VERSIONS[0];
-  $scope.docs_versions = NG_VERSIONS;
+.factory('getVersions', ['$http', '$q', function($http, $q) {
 
-  for (var i = 0, minor = NaN; i < NG_VERSIONS.length; i++) {
-    var version = NG_VERSIONS[i];
-    if (version.isSnapshot) {
-      version.isLatest = true;
-      continue;
-    }
-    // NaN will give false here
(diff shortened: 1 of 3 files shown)


Narretz:
I don't particularly like it.
- external provider for cross-origin introduces another point of failure out of our control
- the whole version getting is fully asnyc, which is bad usability (and bad for people with bad connections). Not sure if it can be cached, either.
- It adds comparably a lot of code

Since this will only expose the complete list on any versions that include this commit, why don't we just generate a version list (maybe json) on code.angularjs.org and get that, additionally to the data in the constant NG_VERSIONS? We'd still have to update the nginx config so it's reachable from docs.angularjs.org. This would eliminate the cross-origin proviider, and the need for so much code.

petebacondarwin:
I agree about the CORS provider, and using a file that is stored on our own server with CORS enabled would be much better.

The increase in code size is a fair point and if we generate our own files then we can prepare them so that we don't need to do any semver parsing etc.

I am not convinced that asynchronicity is an issue. The requests are made in the background and do not block the user from accessing the docs app and navigating between API pages. It seems unlikely that the user will load the page and immediately try to use the version dropdown before it has been filled with data. In fact putting the data into script files that must be loaded before we can bootstrap Angular is probably going to impact the usability more, since it adds a delay before the entire app can be used.

I'll try again.

Narretz:
We don't know how many people change the docs version instantly. But I wouldn't underestimate the number of people  who still use older versions, and just go to docs.angularjs.org and change their version. This was probably even more of an issue when we were defaulting to snapshot instead of latest stable. It's true about the payload, but you didn't remove the NG_VERSIONS creation, so I thought why not use it.

Some more feedback would be good, but if we go full async then we should put the current version instantly into the select, and provide a loading indicator when the select has been opened and the versions haven't arrived yet.

petebacondarwin:
Closing in favour of #15385

PR #15383, opened 2016-11-10
docs(angular.isArray): Document that isArray is an alias

graingert:
**What kind of change does this PR introduce?** docs update

no

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -601,7 +601,7 @@ function isDate(value) {
  * @kind function
  *
  * @description
- * Determines if a reference is an `Array`.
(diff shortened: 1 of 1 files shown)


Narretz:
This is technically an implementation detail, why do you want to document this?

graingert:
@Narretz so users can make an informed decision about whether to use Array.isArray or angular.isArray

gkalpak:
<sub>For the record, I agree this is an implementation detail (was different in the past and may change in the future - not likely but possible). We don't document other helper/function implementations to allow informed decisions. Just saying...</sub>

PR #15385, opened 2016-11-10
chore(docs): improve version picker

petebacondarwin:

chore

Versions in the version picker are hard-coded into each build
This means that newer versions are never shown in older versions of the docs

Now the version information has split into the current and the list of all versions.
In the production deployment the version list is now read from the snapshot build, so we always
get the latest list.

No

**Other information**:

Replaces #15381

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -346,7 +346,7 @@ module.exports = function(grunt) {
 
   grunt.registerTask('minify', ['bower', 'clean', 'build', 'minall']);
   grunt.registerTask('webserver', ['connect:devserver']);
-  grunt.registerTask('package', ['bower', 'validate-angular-files', 'clean', 'buildall', 'minall', 'collect-errors', 'docs', 'copy', 'write', 'compress']);
+  grunt.registerTask('package', ['bower', 'validate-angular-files', 'clean', 'buildall', 'minall', 'collect-errors', 'write', 'docs', 'copy', 'compress']);
   grunt.registerTask('ci-checks', ['ddescribe-iit', 'merge-conflict', 'eslint']);
   grunt.registerTask('default', ['package']);
 };
--- a/docs/config/processors/versions-data.js
+++ b/docs/config/processors/versions-data.js
@@ -1,6 +1,7 @@
 'use strict';
 
-var _ = require('lodash');
+var exec = require('shelljs').exec;
+var semver = require('semver');
 
 /**
  * @dgProcessor generateVersionDocProcessor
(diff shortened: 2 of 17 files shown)


docs/config/services/deployments/production.js
```diff
@@ -39,7 +39,8 @@ module.exports = function productionDeployment(getVersion) {
       'components/lunr.js-' + getVersion('lunr.js') + '/lunr.min.js',
       'components/google-code-prettify-' + getVersion('google-code-prettify') + '/src/prettify.js',
       'components/google-code-prettify-' + getVersion('google-code-prettify') + '/src/lang-css.js',
-      'js/versions-data.js',
+      'js/version-data.js',
+      'https://code.angularjs.org/snapshot/docs/js/versions-data.js',
```
petebacondarwin: note that in production we read the list from snapshot
otherwise we just use the local

petebacondarwin: This will require some CORS setup on code.angularjs.org I think

petebacondarwin: actually no it won't because we are loading via a script not XHR

petebacondarwin:
We could change the docs apps for most releases on code.angularjs.org to use this same code.

docs/config/processors/versions-data.js
```diff
@@ -12,23 +15,93 @@ module.exports = function generateVersionDocProcessor(gitData) {
   return {
     $runAfter: ['generatePagesDataProcessor'],
     $runBefore: ['rendering-docs'],
+    blacklist: ['1.3.4-build.3588'],
     $process: function(docs) {
 
-      var versionDoc = {
+      var blacklist = this.blacklist;
```
Narretz: What's the meaning of the blacklist? Can you please add a comment?

Narretz (changes requested):
A few minor things that shold be clarified, otherwise I really like the solution.

Gruntfile.js
```diff
@@ -346,7 +346,7 @@ module.exports = function(grunt) {
 
   grunt.registerTask('minify', ['bower', 'clean', 'build', 'minall']);
   grunt.registerTask('webserver', ['connect:devserver']);
-  grunt.registerTask('package', ['bower', 'validate-angular-files', 'clean', 'buildall', 'minall', 'collect-errors', 'docs', 'copy', 'write', 'compress']);
+  grunt.registerTask('package', ['bower', 'validate-angular-files', 'clean', 'buildall', 'minall', 'collect-errors', 'write', 'docs', 'copy', 'compress']);
```
petebacondarwin: This is because the version.json file is generated by the `write` grunt task. So we need it to run before the `docs` task, which now relies upon it.

docs/config/processors/versions-data.js
petebacondarwin: OK. The main reason was that we appear to have npm published a random rogue release, 1.3.4-build.3588, which does not have an associated folder on code.angularjs.org

docs/config/processors/versions-data.js
```diff
@@ -12,23 +15,93 @@ module.exports = function generateVersionDocProcessor(gitData) {
...
+              if (version) {
+                var key = version.major + '.' + version.minor;
+                var latest = latestMap[key];
+                if (!latest || version.compare(latest) > 0) {
+                  latestMap[key] = version;
+                }
+              }
+              return version;
+          })
+          .filter(function(version) {
```
gkalpak: AFAICT, filtering after having populated `latestMap`, will allow `0.x` versions to appear in the "Latest" group of the dropdown. The filtering should be done before populating ``latestMap`.

gkalpak (changes requested):

gkalpak:
Not directly related to this PR, but close enough:
I miss showing the build number in the version picker for snapshot releases.

docs/config/processors/versions-data.js
Narretz: I can't reproduce this locally. It also looks like we don't include any 0.x versions at all.

Narretz:
The build number is shown on the api docs home page. Why do you want to have it in the picker?

gkalpak:
Convenience :) It makes it easy to see when refreshing the page gets you a newer build.
It's probably only useful to me, not to our users :)

docs/config/processors/versions-data.js
gkalpak: I didn't try it locally, but from looking at the code I assumed that 0.x versions would be shown in the "Latest" group. Are you sure they are not there (only in the "Latest" group though).

petebacondarwin:
I think we should display the build number of the docs somewhere on the page. Perhaps near the bottom?

Narretz:
The build number is displayed in the footer. :D

gkalpak:
Right. But you have to scroll all the way down to see the footer :smiley:
Anyway, it is probably not worth making it more convenient for me/us, if it is something that most users don't care about. The info is there if someone needs it (even if they have to scroll :smiley:).

gkalpak (approved):

gkalpak:
LGTM as long as Travis is happy.

petebacondarwin:
Travis is not happy but it is an east fix

petebacondarwin:
Travis is now happy - after a copied the all-versions.js file onto the snapshot :-)

petebacondarwin:
See https://github.com/angular/angular.js/commit/48362788303a70d9e4ab36e672ec68a4ce5645c3

PR #15389, opened 2016-11-14
Ng model options extend

petebacondarwin:

feat

`ngModelOptions` cannot be inherited from ancestor `ngModelOptions`

`ngModelOptions` can inherit if specified via `$inherit` values

A small internal-ish change to the programmatic API for reading options.

**Other information**:


test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -0,0 +1,768 @@
...
+
+      it('should be retrieved from an ancestor element containing an `ngModelOptions` directive', function() {
+        var doc = $compile(
+            '<form name="test" ' +
+                'ng-model-options="{ debounce: 10000, updateOn: \'blur\' }" >' +
+              '<input type="text" ng-model="name" name="alias" />' +
+            '</form>')($rootScope);
+        $rootScope.$digest();
+
+        var inputElm = doc.find('input').eq(0);
```
gkalpak: Nit: Isn't `.eq(0)` redundant?

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,378 @@
+'use strict';
+
+/* global $DefaultModelOptionsProvider: true */
+
+/**
+ * @ngdoc provider
+ * @name $defaultModelOptionsProvider
+ * @description
+ *
+ * Here, you can change the default settings from which {@link ngModelOptions}
```
petebacondarwin: Actually I don't want them to change defaults - ever. That would put us back in the same position as before where application developers can break library components that expect defaults to be a certain value.

Instead if one wants to change stuff globally then they must put an `ngModelOptions` directive on the same element as the `ng-app` attribute. Then use `"*": "$inherit"`

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,378 @@
...
+     * @name $defaultModelOptions
+     * @type ModelOptions
+     * @description
+     *
+     * This service provides the application wide default {@link ModelOptions} options that
+     * will be used by {@link ngModel} directives if no {@link ngModelOptions} directive is
+     * specified.
+     */
+    $get: function() {
+      return new ModelOptions(defaultOptions, defaultOptions);
```
petebacondarwin: Are you suggesting we create a global variable!?!
I agree.

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,378 @@
...
+  };
+}
+
+
+/**
+ * @ngdoc directive
+ * @name ngModelOptions
+ *
+ * @description
+ * This directive allows you to modify the behaviour of ngModel and input directives within your
```
petebacondarwin: OK OK

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,378 @@
...
+ * The `updateOn` and `debounce` properties allow you to specify a custom list of events that will
+ * trigger a model update and/or a debouncing delay so that the actual update only takes place when
+ * a timer expires; this timer will be reset after another change takes place.
+ *
+ * Given the nature of `ngModelOptions`, the value displayed inside input fields in the view might
+ * be different from the value in the actual model. This means that if you update the model you
+ * should also invoke {@link ngModel.NgModelController `$rollbackViewValue`} on the relevant input field in
+ * order to make sure it is synchronized with the model and that any debounced action is canceled.
+ *
+ * The easiest way to reference the control's {@link ngModel.NgModelController `$rollbackViewValue`}
```
petebacondarwin: Not sure. This is just cut-n-paste from the previous implementation

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ * @ngdoc type
+ * @name ModelOptions
+ * @description
+ * A container for the options set by the {@link ngModelOptions} directive
+ * and the {@link $defaultModelOptions} service.
+ */
+function ModelOptions(defaultOptions, options, parentOptions) {
+  this._defaultOptions = defaultOptions;
```
petebacondarwin: OK

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,378 @@
...
+    var options = this._options;
+    var inheritAll = false;
+
+    // Inherit options from the parent if specified by the value `"$inherit"`
+    forEach(options, function(option, key) {
+      if (option === '$inherit') {
+        if (key === '*') {
+          inheritAll = true;
+        } else {
+          options[key] = isDefined(parentOptions[key]) ? parentOptions[key] : defaultOptions[key];
```
petebacondarwin: OK

test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -0,0 +1,768 @@
+'use strict';
+
+/* globals generateInputCompilerHelper: false */
+describe('ngModelOptions', function() {
+
+  describe('$defaultModelOptions', function() {
```
petebacondarwin: Actually they are not :-P
I just need to fix the docs - see my other comment

test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -0,0 +1,768 @@
...
+      it('if there is no `ngModelOptions` directive', inject(function($defaultModelOptions) {
+        var inputElm = helper.compileInput(
+            '<input type="text" ng-model="name" name="alias" />');
+
+        var inputOptions = $rootScope.form.alias.$options;
+        expect(inputOptions.getOption('updateOn')).toEqual($defaultModelOptions.getOption('updateOn'));
+        expect(inputOptions.getOption('updateOnDefault')).toEqual($defaultModelOptions.getOption('updateOnDefault'));
+        expect(inputOptions.getOption('debounce')).toEqual($defaultModelOptions.getOption('debounce'));
+        expect(inputOptions.getOption('getterSetter')).toEqual($defaultModelOptions.getOption('getterSetter'));
+        expect(inputOptions.getOption('allowInvalid')).toEqual($defaultModelOptions.getOption('allowInvalid'));
```
petebacondarwin: Right

test/ng/directive/ngModelOptionsSpec.js
petebacondarwin: I think that `find` does not return the object that we want (i.e. some collection thing) so we must convert it to get an instance.

src/ng/directive/ngModelOptions.js
gkalpak: It's fine by me. But then why do we need a public service for that? People will always be able to overwrite that.

test/ng/directive/ngModelOptionsSpec.js
petebacondarwin: I see that it is not used elsewhere... I'll try removing it

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,378 @@
...
+  /**
+   * @ngdoc method
+   * @name ModelOptions#getOption
+   * @param {string} name the name of the option to retrieve
+   * @returns {*} the value of the option
+   * @description
+   * Returns the value of the given option
+   */
+  getOption: function(name) {
+    return isDefined(this._options[name]) ? this._options[name] : this._defaultOptions[name];
```
gkalpak: It wouldn't make any difference anyway. Even with the current implementation we would return `this._defaultOptions[key]` which would be undefined if we didn't specify the option in `$defaultModelOptions`. 

But I agree it is a good idea to have all options in `$defaultModelOptions`.

test/ng/directive/ngModelOptionsSpec.js
gkalpak: `eq` returns a new collection with a subset of the items (typically just one):

```js
  eq: function(index) {
      return (index >= 0) ? jqLite(this[index]) : jqLite(this[this.length + index]);
  }
```

Here there is only one `input` element  anyway, so it is redundant.

src/ng/directive/ngModelOptions.js
petebacondarwin: Good point. I'll change that

gkalpak (changes requested):

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,358 @@
...
+    if (inheritAll) {
+      delete this.$$options['*'];
+      defaults(this.$$options, this.$$parentOptions);
+    }
+  },
+
+  $$processOptions: function() {
+    var options = this.$$options;
+    // updateOn and updateOnDefault
+    if (isDefined(options.updateOn) && options.updateOn.trim()) {
```
petebacondarwin: Right! Now this is a good review point.  I knew there must be something that needed fixing. I'll take a look right now.

src/ng/directive/ngModelOptions.js
```diff
@@ -0,0 +1,345 @@
...
+
+    // Inherit options from the parent if specified by the value `"$inherit"`
+    forEach(options, /* @this */ function(option, key) {
+      if (option === '$inherit') {
+        if (key === '*') {
+          inheritAll = true;
+        } else {
+          options[key] = this.$$options[key];
+        }
+      } else if (key === 'updateOn') {
```
gkalpak: There are still some issues with this implementation:

- If `{updateOn: '$inherit'}` it will inherit the `updateOn`, but not the `updateOnDefault` (which might be inherited by `$defaultModelOptions`).

gkalpak (changes requested):

src/ng/directive/ngModelOptions.js
petebacondarwin: Great. I am really working your reviewing skills today!

petebacondarwin:
OK, trying again. See if you can find another failing scenario @gkalpak

test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -180,6 +180,31 @@ describe('ngModelOptions', function() {
       });
 
 
+      it('should `updateOnDefault` as well if we have `updateOn: "$inherit"`', function() {
+        var container = $compile(
+                  '<div ng-model-options="{updateOn: \'keyup\'}">' +
+                    '<input ng-model-options="{\'updateOn\': \'$inherit\'}">' +
```
gkalpak: Unnecessary quotes around `updateOn` (here and below).

gkalpak (approved):
Fourth time's a charm :smiley:
LGTM

PR #15391, opened 2016-11-15
docs(ngModel.NgModelController): correct the mistake of demos

jasonxia23:
The demo of `$rollbackViewValue()` is kind of different from the doc.

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -459,7 +459,6 @@ NgModelController.prototype = {
    *           if (rollback) {
    *             $scope.myForm[value].$rollbackViewValue();
    *           }
-   *           $scope.model[value] = '';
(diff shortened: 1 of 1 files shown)


gkalpak:
Thx for the PR, but the example works as intended. Why do you think it is not?

jasonxia23:
After type the `esc`, example of `$rollbackViewValue()` will set back to empty rather than the previous model value.

jasonxia23:
`I signed it!`

gkalpak:
That is exactly what it is supposed to do (as explained in tge example). Hitting ESC should empty the fields (not reset them to their last committed value).

What the example demonstrates, is the correct way of programmatically updating debounced inputs; i.e. roll back any pending view value updates first (via `$rollbackViewValue()`).

PR #15392, opened 2016-11-15
docs($compile): add double compilation known issue

Narretz:


src/ng/compile.js
```diff
@@ -943,6 +943,103 @@
...
+  })
+  ```
+ * At first glance, it looks like removing the original `addInput` attribute is all there is needed
+ * to make this example work.
+ * However, if the directive element or its children have other directives attached, they will be compiled and
+ * linked again, because the compiler doesn't keep track of which directives have been assigned to which
+ * elements.
+ *
+ * This can cause unpredictable behavior, e.g. `ngModel` $formatters and $parsers will be
+ * attached again to the ngModelController. It can also degrade performance, as
```
gkalpak: >  `ngModel` $formatters and $parsers will be attached again to the ngModelController

Why? `ngModel` seems to only be compiled once.

src/ng/compile.js
```diff
@@ -943,6 +943,103 @@
...
+ * watchers for text interpolation are added twice to the scope.
+ *
+ * Double compilation should therefore avoided. In the above example, the better way is to only
```
gkalpak: _avoided_ --> _be avoided_

gkalpak: _the better way_ --> _a better way_ (?)

src/ng/compile.js
```diff
@@ -943,6 +943,103 @@
...
+ * compile the new element:
+ * ```
+  angular.module('app').directive('addInput', function($compile) {
+    return {
+      link: function(scope, element, attrs) {
+        var newEl = angular.element('<input ng-model="$ctrl.value">');
+        $compile(newEl)(scope); // Only compile the new element
+        element.append(newEl);
```
gkalpak: I think usually it is preferrable to first append the element and then compile it, in order to have proper context (e.g. requiring parent directive controllers).

src/ng/compile.js
```diff
@@ -943,6 +943,103 @@
...
+ * compile again.
+ * ```html
+ * <input ng-model="$ctrl.value" add-options>
+ * ```
+ *
+  ```
+  angular.module('app').directive('addOptions', function($compile) {
+    return {
+      link: function(scope, element, attrs) {
+        attrs.$set('addInput', null) // To stop infinite compile loop
```
gkalpak: _addInput_ --> _addOptions_

src/ng/compile.js
```diff
@@ -943,6 +943,103 @@
...
+        attrs.$set('ngModelOptions', '{debounce: 1000}');
+        $compile(element)(scope); // Double compilation
+      }
+    }
+  });
+  ```
+ *
+ * In that case, it is necessary to intercept the *initial* compilation of the element:
+ *
+ * 1. give your directive the `terminal` property and a higher priority than directives
```
gkalpak: Capitalize these sentences.

src/ng/compile.js
```diff
@@ -943,6 +943,103 @@
...
+ * that should not be compiled twice. In the example, the compiler will only compile directives
+ * which have a priority of 100 or higher.
+ * 2. inside this directive's compile function, remove the original directive attribute from the element,
+ * and add any other directive attributes. Removing the attribute is necessary, because otherwise the
```
gkalpak: You don't need to remove the attribute, since you pass the priority to `$compile`.

gkalpak (changes requested):
We could also put that in the compiler guide (with a short note here in the API docs that double-compilation is not supported).

src/ng/compile.js
Narretz: True, I was actually thinking of an issue with replace. I will change the example.

Narretz:
I've updated the example and moved it to the compiler guide, and added a short paragraph with a link to the compile api docs

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,110 @@ restrict: 'E',
...
+A common scenario where this happens is a directive that calls `$compile` in a directive link
+function on the directive element. In the following example, a directive adds a mouseover behavior
+to a button with `ngClick` on it:
+
+```
+angular.module('app').directive('addMouseover', function($compile) {
+  return {
+    link: function(scope, element, attrs) {
+      var newEl = angular.element('<span ng-show="showHint"> My Hint</span>');
+      element.on('mouseenter mouseout', function() {
```
gkalpak: The opposite of `mouseenter` is `mouseout` :grin:

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,110 @@ restrict: 'E',
...
+
+Double compilation occurs when an already compiled part of the DOM gets compiled again. This is an
+undesired effect and can lead to misbehaving directives, performance issues, and memory
+leaks.
+A common scenario where this happens is a directive that calls `$compile` in a directive link
+function on the directive element. In the following example, a directive adds a mouseover behavior
+to a button with `ngClick` on it:
+
+```
+angular.module('app').directive('addMouseover', function($compile) {
```
gkalpak: We should have a big, fat comment or a danger alert that this is what you should NOT do. People tend to get confused when the docs show bad examples to be avoided.

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,110 @@ restrict: 'E',
...
+  return {
+    link: function(scope, element, attrs) {
+      var newEl = angular.element('<span ng-show="showHint"> My Hint</span>');
+      element.on('mouseenter mouseout', function() {
+        scope.$apply('showHint = !showHint');
+      });
+
+      attrs.$set('addMouseover', null);
```
gkalpak: No need to remove this any more.

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,110 @@ restrict: 'E',
...
+      element.append(newEl);
+      $compile(newEl)(scope); // Only compile the new element
+    }
+  }
+})
+```
+
+Another scenario is adding a directive programmatically to a compiled element and then executing
+compile again.
```
gkalpak: Again, lets make it clear this is a BAD thing.

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,110 @@ restrict: 'E',
...
+```
+
+In that case, it is necessary to intercept the *initial* compilation of the element:
+
+  1. Give your directive the `terminal` property and a higher priority than directives
+  that should not be compiled twice. In the example, the compiler will only compile directives
+  which have a priority of 100 or higher.
+  2. Inside this directive's compile function, remove the original directive attribute from the element,
+  and add any other directive attributes. Removing the attribute is necessary, because otherwise the
+  compilation would result in an infinite loop.
```
gkalpak: You don't need to remove the attribute, since you pass the priority to $compile.

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,110 @@ restrict: 'E',
...
+  3. Compile the element but restrict the maximum priority, so that any already compiled directives
+  are not compiled twice.
+  4. In the link function, link the compiled element with the element's scope
+
+```
+angular.module('app').directive('addOptions', function($compile) {
+  return {
+    priority: 100, // ngModel has priority 1
+    terminal: true,
+    template: '<input ng-model="$ctrl.value">',
```
gkalpak: You don't want to have a template.

gkalpak (changes requested):

docs/content/guide/compiler.ngdoc
Narretz: Sorry, I don't understand. Do you mean it should be `mouseleave`?

docs/content/guide/compiler.ngdoc
Narretz: I agree, but we should have a general solution to this.

Narretz: I tried putting a div around it, but then the code blocks aren't rendered correctly. Seems to be an issue with the marked library: https://github.com/chjj/marked/issues/596

Narretz:
I've updated the PR @gkalpak PTAL

docs/content/guide/compiler.ngdoc
gkalpak: This is still `mouseout` 😞

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,106 @@ restrict: 'E',
...
+
+In that case, it is necessary to intercept the *initial* compilation of the element:
+
+  1. Give your directive the `terminal` property and a higher priority than directives
+  that should not be compiled twice. In the example, the compiler will only compile directives
+  which have a priority of 100 or higher.
+  2. Inside this directive's compile function, add any other directive attributes to the template.
+  3. Compile the element, but restrict the maximum priority, so that any already compiled directives
+  (including the `addOptions` directive) are not compiled again.
+  4. In the link function, link the compiled element with the element's scope
```
gkalpak: Missing period at the end.

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,106 @@ restrict: 'E',
...
+### Double Compilation, and how to avoid it
+
+Double compilation occurs when an already compiled part of the DOM gets compiled again. This is an
+undesired effect and can lead to misbehaving directives, performance issues, and memory
+leaks.
+A common scenario where this happens is a directive that calls `$compile` in a directive link
+function on the directive element. In the following **faulty example**, a directive adds a mouseover behavior
+to a button with `ngClick` on it:
+
+```
```
gkalpak: Should this be ```` ```js ```` (here and below) or does marked autodetect JS?

docs/content/guide/compiler.ngdoc
```diff
@@ -380,3 +380,106 @@ restrict: 'E',
...
+angular.module('app').directive('addMouseover', function($compile) {
```
gkalpak: I would still prefer a comment above this line; something like:

```js
// BAD - Don't do this
```

And an equivalent comment in the recommended alternatives; e.g.:

```js
// OK - You can do this
```

But I can leave without if you think it's too much. Up to you.

gkalpak (approved):
A couple of typos.
LGTM otherwise 👍

docs/content/guide/compiler.ngdoc
Narretz: I always thought it auto-detects

Narretz:
I think the part about `mouseout/leave` is broken on Github's UI. You can see that it's mouseleave in the file at the latest fixup commit. It just shows wrongly in the "Files Changed" view. I can also not comment on it from the comment view.

docs/content/guide/compiler.ngdoc
Narretz: I'll leave it as is, and we'll see how soon someone stumbles over this

gkalpak:
There were two occurrences of `mouseout`. Yo have replaced the second one with `mouseleave`, but there is still one more (on line 397).

Narretz:
Omg ...

gkalpak (approved):
LGTM (as soon as travis is happy)

PR #15394, opened 2016-11-15
chore(docs): deprecation notices for methods and properties

### LINKED ISSUE #15351: @deprecated tags are not reflected in event/methods/properties descriptions

A bug.

@deprecated tags are not reflected in event/methods/properties descriptions 

See https://code.angularjs.org/1.6.0-rc.0/docs/api/ng/provider/$controllerProvider#allowGlobals
The `allowGlobals` method is deprecated in 1.6 and will be removed in 1.7. However, this info is not visible on the page even though [it's present in the ngdoc](https://github.com/angular/angular.js/blob/828f8a6/src/ng/controller.js#L63-L64).

The deprecatuon info should be displayed in event/methods/properties descriptions.

The whole point of deprecation is to make people know they should stop using the API.

All

petebacondarwin:

**Other information**:

Closes #15351


petebacondarwin:
I'm aware that Travis is not happy. I will fix this later today and maybe even break the changes into smaller commits for easier reviewing :-)

petebacondarwin:
OK, smaller commits. Travis is happy.
What more could you want of a PR?

src/ngCookies/cookieStore.js
```diff
@@ -5,6 +5,8 @@ angular.module('ngCookies').
  * @ngdoc service
  * @name $cookieStore
  * @deprecated
+ * Please use the {@link ngCookies.$cookies `$cookies`} service instead.
```
mgol: Hasn't this been deprecated for some time now? Perhaps we should schedule a removal timeline?

mgol: It'd be also nice to mention the version in which this was deprecated.

docs/app/assets/css/docs.css
```diff
@@ -662,6 +662,12 @@ ul.events > li {
   max-width: 100%;
 }
 
+.deprecated-title {
```
Narretz: Shouldn't this technically be part of the deprecation commit?

Narretz:
Good stuff. Having the templates in the repo allows us to move faste with changes, too.
A few notes:
-  I think using "warning" is a bit too weak for deprecation. It should be danger.
- The API deprecation template is different. We should ideally use the same as for props and methods (can do in a follow-up)
- In a follow-up, we should also make the deprecation notices in `@description` blocks consistent. Most noticable in the compile docs

mgol:
> Can we have "sub-annoations" for `@deprecated`? then we could insert `@since_version`, `@remove_version`, and `@description`, and make this consistent across the board

That would be wonderful.

docs/config/templates/ngdoc/api/module.template.html
```diff
@@ -5,41 +5,33 @@
...
-{% else %}
+{% if doc.name != 'ng' and doc.name != 'auto' %}
+  <h2>Installation</h2>
 
-  <p>First, get the file:</p>
+  <p>First include {$ doc.packageFile | code $} in your HTML:</p>
+
+  {% code %}
+      <script src="angular.js">
+      <script src="{$ doc.packageFile $}">
```
gkalpak: Missing closing `<script>` tags.

docs/config/templates/ngdoc/api/module.template.html
```diff
@@ -5,41 +5,33 @@
...
-      <a href="https://www.npmjs.com/">NPM</a> e.g.
-      {% code %}npm install {$ doc.packageName $}@X.Y.Z{% endcode %}
+      <a href="https://developers.google.com/speed/libraries/devguide#angularjs">Google CDN</a><br>
+      e.g. {$ ("//ajax.googleapis.com/ajax/libs/angularjs/X.Y.Z/" + doc.packageFile) | code $}
     </li>
     <li>
-      <a href="http://bower.io">Bower</a> e.g.
-      {% code %}bower install {$ doc.packageName $}#X.Y.Z{% endcode %}
+      <a href="http://bower.io">Bower</a><br>
+      e.g. {% code %}bower install {$ doc.packageName $}@X.Y.Z{% endcode %}
```
gkalpak: Shouldn't this be `#` instead of `@`?

docs/config/templates/ngdoc/api/module.template.html
```diff
@@ -69,7 +61,6 @@ <h2 id="known-issues">Known Issues</h2>
 {% endif %}
 
 
-{% if doc.componentGroups.length %}
```
gkalpak: Why was this removed?

gkalpak (approved):

docs/config/templates/ngdoc/api/module.template.html
petebacondarwin: this is what is in dgeni-packages - so it looks like a breakage - but then it is re-fixed in a subsequent commit.

petebacondarwin:
> Can we have "sub-annoations" for @deprecated? then we could insert @since_version, @remove_version, and @description, and make this consistent across the board

The content of an annotation is completely configurable. But it can't really use sub-annotations, because the parser would see them as no top level annotations.
But you can do something like:

```
@deprecated since-version="..." remove-version="..." what is left is description
```

Then we would just parse for the attributes at the start of the description and pull them out.

Is that what you want?

petebacondarwin:
Fixed up.

src/ngCookies/cookieStore.js
mgol: Sure. :)

Narretz:
@petebacondarwin  Yes, this custom annoation content like you describe is what I'd like. I think `@param` does something like that? It has some custom `transforms` in the tag-defs. I was lead a bit in a wrong direction by looking a `@method` which seems to have "real" sub-annoations, but then again that's jsdoc, which is different.

petebacondarwin:
OK, so I have updated the `@deprecation` tag to take `sinceVersion` and `removeVersion` options.

I also moved the docs.css file changes to the correct commit :-)

@Narretz PTAL

mgol:
There are some linting violations but it's most likely due to us not checking docs properly. We can fix it later.

The new `@deprecated` syntax is not yet used, is it, @petebacondarwin?

mgol:
The deprecation notice has disappeared from the `ngComponentRouter` page.

PR #15395, opened 2016-11-15
fix(ngMock/$controller): respect `$compileProvider.preAssignBindingsEnabled()`

### LINKED ISSUE #15387: preAssignBindingsEnabled behavior is not reflected in ngMock $controller

Not sure how this should be classified. I suppose it’s a bug, but it concerns new stuff in RC.

When using ngMock’s decorated `$controller` service in unit tests, even if `preAssignBindingsEnabled` is false, `$controller(exp, locals, bindings)` will try adding bindings old-style, prior to instantiation.

Demonstration with Jasmine test:

http://plnkr.co/edit/IsNy8WjASeKtp6KIpXWt?p=preview

When `preAssignBindingsEnabled` is `false`, I would expect ngMock to honor that.

If ngMock’s decoration doesn’t take into account `preAssignBindingsEnabled`, one could pretty easily have test results that pass which actually wouldn’t work live. This seems important because if people are converting existing code to work with the new constraints, tests will not currently reveal places where they continue to rely on attribute bindings being available in constructors.

6.0.0-rc0

If I change the test to use native `class`, it passes. I’m guessing this is because there’s a try-catch in there somewhere that falls back on binding afterwards when forced too. But outside of mocks, when `preAssignBindingsEnabled` is false, all controller constructors are treated as if they had been defined as ES6 classes, i.e., that they cannot be bound in advance without zany hijinks (e.g. `class OneOffChild extends ActualCtrl {}, Object.assign(OneOffChild.prototype, bindings), new OneOffChild(...injections)`).

gkalpak:
Bug fix.

`ngMock`'s `$controller` service always pre-assigns bindings (if available), regardless of the value of `$compileProvider.preAssignBindingsEnabled()`. This may cause false positives or negatives in tests.
See #15387.

`ngMock`'s `$controller` service does not pre-assign bindings (if available), unless `$compileProvider.preAssignBindingsEnabled()` returns true.

No.

**Other information**:
Fixes #15387

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2184,6 +2184,10 @@ angular.mock.$RootElementProvider = function() {
  * A decorator for {@link ng.$controller} with additional `bindings` parameter, useful when testing
  * controllers of directives that use {@link $compile#-bindtocontroller- `bindToController`}.
  *
+ * Depending on the value of
+ * {@link ng.$compileProvider#preAssignBindingsEnabled `preAssignBindingsEnabled()`}, the properties
+ * will be bound before or after invoking the constructor.
+ *
  *
  * ## Example
  *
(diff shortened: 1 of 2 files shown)


petebacondarwin:
LGTM as long as Travis is happy.
(Although I am not that keen on using `forEach([true, false], function(enabled) {` style tests. I would rather just duplicate. But not enough to force a change here.)

gkalpak:
It turned out `forEach()` is not available in modules test.Since I had to change the tests anyway, I took the opportunity to split them as well. I'll wait for Travis to be happy and merge.
(cc @petebacondarwin)

petebacondarwin:
Travis seems happy. Feel free to merge.

arkaitzgarro:
This change ended up on `1.5.10` release. Is this correct? Because I can see here was assigned to `1.6.0-rc.1` milestone. Our unit tests are failing now if we use `angular-mocks 1.5.10`: `TypeError: undefined is not a constructor (evaluating 'compileProvider.preAssignBindingsEnabled()')`

gkalpak:
Yes, this was intentional. This is complementary to the new `$compileProvider.preAssignBindingsEnabled()` .
Make sure you use matching versions of core angular (`angular.js`) and external modules (`angular-mocks.js`, `angular-animate.js` etc).

PR #15401, opened 2016-11-16
Ng model options fix

petebacondarwin:

**Other information**:

### DIFF
--- a/test/ng/directive/ngModelOptionsSpec.js
+++ b/test/ng/directive/ngModelOptionsSpec.js
@@ -469,25 +469,51 @@ describe('ngModelOptions', function() {
         var inputElm = helper.compileInput(
             '<input type="text" ng-model="name" name="alias" ' +
               'ng-model-options="{' +
-                'updateOn: \'default blur\', ' +
+                'updateOn: \'default blur mouseup\', ' +
                 'debounce: { default: 10000, blur: 5000 }' +
               '}"' +
             '/>');
 
         helper.changeInputValueTo('a');
-        expect($rootScope.checkbox).toBeUndefined();
+        expect($rootScope.name).toBeUndefined();
         $timeout.flush(6000);
-        expect($rootScope.checkbox).toBeUndefined();
+        expect($rootScope.name).toBeUndefined();
         $timeout.flush(4000);
         expect($rootScope.name).toEqual('a');
+
         helper.changeInputValueTo('b');
         browserTrigger(inputElm, 'blur');
         $timeout.flush(4000);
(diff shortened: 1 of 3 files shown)


test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -480,15 +480,37 @@ describe('ngModelOptions', function() {
         expect($rootScope.checkbox).toBeUndefined();
```
gkalpak: Not directly related to this PR, but `checkbox` should be `name` (here and above).

test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -480,15 +480,37 @@ describe('ngModelOptions', function() {
...
+        helper.changeInputValueTo('c');
+        browserTrigger(helper.inputElm, 'mouseup');
+        // counter-intuitively `default` in `debounce` is a catch-all
+        expect($rootScope.name).toEqual('b');
+        $timeout.flush(10000);
+        expect($rootScope.name).toEqual('c');
       });
 
 
+      it('should trigger immediately for the default event if not listed in the debounce list',
```
gkalpak: This is not specific to the `default` event. The bug affected any event that was not listed in `debounce` as long as `debounce` was an object and didn't contain the catch-all `default`. I wouldn't mind adding one more assertion to the test, like this:

```diff
          var inputElm = helper.compileInput(
               '<input type="text" ng-model="name" name="alias" ' +
                 'ng-model-options="{' +
-                  'updateOn: \'default blur\', ' +
+                  'updateOn: \'default blur foo\', ' +
                   'debounce: { blur: 5000 }' +
                 '}"' +
               '/>');
   
           helper.changeInputValueTo('a');
           expect($rootScope.name).toEqual('a');
+          helper.changeInputValueTo('b');
+          browserTrigger(inputElm, 'foo');
+          expect($rootScope.name).toEqual('b');
        });
```

gkalpak (approved):
LGTM
(My two comments are non-blocking, but good-to-have :smiley:)

test/ng/directive/ngModelOptionsSpec.js
petebacondarwin: The trouble with the test as you suggest is that the `default` event is triggered by `helper.changeInputValueTo('b')` in any case. And so this would pass even if the `foo` event had a weird truthy timeout.

test/ng/directive/ngModelOptionsSpec.js
petebacondarwin: good spot

PR #15405, opened 2016-11-17
protect from ev.elapsedTime to be undefined (IE)

shaiatias:
(fix): protect from elapsedTime to be undefined

in ie, some plugins emit events without elapsedTime, so i set the elapsedTime to maxDuration in order to end the animation immediately


gkalpak:
Would it be possible to add a unit test for this? We also need a signed CLA before we can merge it.

Narretz:

PR #15407, opened 2016-11-17
chore(*): cleanup msie handling; add support comments

mgol:
Chore

N/A

N/A

No

**Other information**:

1. The conditions checking the msie variable value have been simplified.
There is e.g. no point to check if `msie <= 11` since there IE 12 won't ever
exist.
2. Edge UA-sniffing has been added to tests (only!) where appropriate
3. Support comments for IE/Edge have been added.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -169,6 +169,7 @@ var
     angularModule,
     uid               = 0;
 
+// Support: IE 9-11 only
 /**
  * documentMode is an IE-only property
  * http://msdn.microsoft.com/en-us/library/ie/cc196988(v=vs.85).aspx
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1280,6 +1280,7 @@ function fromJson(json) {
 
 var ALL_COLONS = /:/g;
 function timezoneToOffset(timezone, fallback) {
+  // Support: IE 9-11 only, Edge 13-14+
   // IE/Edge do not "understand" colon (`:`) in timezone
   timezone = timezone.replace(ALL_COLONS, '');
   var requestedTimezoneOffset = Date.parse('Jan 01, 1970 00:00:00 ' + timezone) / 60000;
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -469,6 +469,10 @@ describe('$compile', function() {
       }));
 
       // NOTE: This test may be redundant.
+      // Support: Edge 14+
+      // An `<svg>` element inside a `<foreignObject>` element on MS Edge has no
+      // size, causing the included `<circle>` element to also have no size and thus fails an
+      // assertion (relying on the element having a non-zero size).
       if (!isEdge) {
         it('should handle custom svg containers that transclude to foreignObject' +
            ' that transclude to custom svg containers that transclude to custom elements', inject(function() {
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -11163,8 +11163,8 @@ describe('$compile', function() {
   });
 
   // Support: IE 9-10 only
-  // IE <=11 don't support srcdoc
-  if (!msie || msie >= 11) {
+  // IEs <11 don't support srcdoc
+  if (!msie || msie === 11) {
     describe('iframe[srcdoc]', function() {
       it('should NOT set iframe contents for untrusted values', inject(function($compile, $rootScope, $sce) {
         element = $compile('<iframe srcdoc="{{html}}"></iframe>')($rootScope);
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -1127,7 +1127,8 @@ describe('$compile', function() {
           expect(element).toHaveClass('class_2');
         }));
 
-        if (!msie || msie > 11) {
+        // Support: IE 9-11 only
+        if (!msie) {
           // style interpolation not working on IE (including IE11).
           it('should handle interpolated css style from replacing directive', inject(
             function($compile, $rootScope) {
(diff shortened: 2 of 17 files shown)


test/ng/directive/booleanAttrsSpec.js
```diff
@@ -283,8 +284,9 @@ describe('ngHref', function() {
     expect(element.attr('href')).toEqual(undefined);
   }));
 
-  if (msie) {
-    // IE11/10/Edge fail when setting a href to a URL containing a % that isn't a valid escape sequence
+  // Support: IE 9-11 only, Edge 12-14+
+  if (msie || /\bEdge\/[\d\.]+\b/) {
```
Narretz: Wouldn't it make more sense to put Edge in a variable, same as `msie`?

test/ng/rootScopeSpec.js
```diff
@@ -13,9 +13,7 @@ describe('Scope', function() {
 
 
     it('should expose the constructor', inject(function($rootScope) {
-      if (msie < 11) return;
-      // eslint-disable-next-line no-proto
-      expect($rootScope.__proto__).toBe($rootScope.constructor.prototype);
+      expect(Object.getPrototypeOf($rootScope)).toBe($rootScope.constructor.prototype);
```
Narretz: Why is the `msie < 11` check gone?

Narretz (changes requested):

test/ng/directive/booleanAttrsSpec.js
jbedard: Doesn't this need a `.test(...)`? Right now it will be true 100% of the time...

test/ng/rootScopeSpec.js
mgol: It was there only because of the Annex B (i.e. compatibility "don't use" stuff) `__proto__` that wasn't present in IE <11. The standard ES5+ way to get the prototype is `Object.getPrototypeOf(object)` which works in IE9+.

test/ng/directive/booleanAttrsSpec.js
mgol: @jbedard Nice catch. This means, though, that we can just remove the `if` and run the test everywhere.

@Narretz A variable would make it easier to reuse and we should avoid UA-sniffing (it's more acceptable in tests but still); it's not robust, it can catch more browsers than intended and cause compatibility concerns. This is actually a good example as the `if` was unnecessary, I'll remove it.

IE sniffing is different as:

1. Browsers are trying to pretend they have nothing to do with IE; differently to e.g. Chrome which is imitated by lots of browsers.
2. We don't do it via the lying user agent string but the non-standard `document.documentMode` that other browsers will never implement.
3. There are lots of differences between IE & modern browsers and the gap will only get larger. Edge gets updated and doesn't lag so much so it won't apply to it.

IE sniffing is, therefore, one of the few safe browser sniffing available.

mgol: Ah, no, this particular condition was necessary, the tests are now failing. Fixing...

mgol:
@Narretz PR updated; PTAL.

src/Angular.js
```diff
@@ -1279,6 +1280,7 @@ function fromJson(json) {
 
 var ALL_COLONS = /:/g;
 function timezoneToOffset(timezone, fallback) {
+  // Support: IE 9-11 only, Edge 13-14+
```
gkalpak: What is the difference between `Edge 13-14+` and `Edge 13+`?

test/ng/compileSpec.js
```diff
@@ -11154,6 +11162,8 @@ describe('$compile', function() {
     }));
   });
 
+  // Support: IE 9-10 only
+  // IE <=11 don't support srcdoc
```
gkalpak: `<=` --> `<`

test/ng/compileSpec.js
```diff
@@ -11154,6 +11162,8 @@ describe('$compile', function() {
...
   if (!msie || msie >= 11) {
```
gkalpak: This could be `!msie || msie === 11` (for consistency).

test/ng/rootScopeSpec.js
```diff
@@ -125,6 +123,8 @@ describe('Scope', function() {
       function Listener() {
         expect(this).toBeUndefined();
       }
+      // Support: IE 9 only
+      // IE 9 doesn't support strict mode so its `this` will always be defined.
       if (msie < 10) return;
```
gkalpak: This could be `msie === 9`.

test/ng/locationSpec.js
```diff
@@ -57,13 +60,14 @@ describe('$location', function() {
...
         setAttribute: angular.noop
       };
-    }));
+    });
 
-    afterEach(inject(function($sniffer) {
-      if (msie) return;
+    afterEach(function() {
+      // Support: non-Windows browsers
+      if (msie || /\bEdge\/[\d\.]+\b/.test(window.navigator.userAgent)) return;
```
gkalpak: It would be nice to have this as a reusable flag in tests.

src/Angular.js
mgol: In the former we've checked that it, indeed, is broken in both v13 & v14. In the latter we've just checked v13 and the v14 state is unknown.

test/ng/locationSpec.js
mgol: I guess it doesn't hurt when it's used just in tests... (I would definitely *not* want sth like that in source). But would you like to have it in tests globally or just in `locationSpec`? If globally then we should define similar variables for other browsers.

I'm a little afraid of making it too easy to use UA-sniffing, though.

gkalpak: No, it doesn't hurt. It is just about the mental overhead of "deciphering" `/\bEdge\/[\d\.]+\b/.test(window.navigator.userAgent)` vs `isEdge`.

> I'm a little afraid of making it too easy to use UA-sniffing, though.

Fair enough :smiley:

gkalpak (approved):
LGTM

mgol:
@Narretz PTAL, your review is blocking this. :)

Narretz (approved):

PR #15412, opened 2016-11-18
feat(ngModelOptions): support catch-all ("*") debounce settings

### LINKED ISSUE #15411: feat(ngModelOptions): support for easily debouncing `default` only
Feature.

The delay specified in an `ngModelOptions`' `debounce.default` property will be used as a "catch-all" value, debouncing all `updateOn` events that don't explicitly specifiy a debounce delay.

[Demo][1]

`debounce.default` should only apply to the `default` event.
For a "catch-all" delay, we could use the `'*'` property, which is similar to how it is used for defining inheritance in `ngModelOptions`. E.g.

```js
{
  updateOn: 'default blur keyup keydown',
  debounce: {default: 100, blur: 0, '*': 333}
}
```

The current behavior can be confusing, since `updateOn` supports a special event, `default` which means the default input event for that input/browser. With the proposed change, it will be still possible to provide a "catch-all" delay (via `'*'`) and `debounce.default` will only affect the `default` event, which is more intuitive. Additionally, it will be easier to apply debouncing on the `default` event only, without having to explicitly set delays for all other supported events.

Probably 1.3+.

[1]: https://codepen.io/ExpertSystem/pen/rWWEjQ?editors=1000

petebacondarwin:

feat

the `default` key acts as a catch all for debounce events but this prevents one from being able to set a different debounce value for `default` than for other unspecified events.

the new `*` key acts as a catch all for unspecified events

No!
If you do not use `*` as a key then the behaviour is unchanged.

**Other information**:

This change enables the use of a `"*"` key in the `debounce` setting of
`ngModelOptions`, which will set the debounce value for all events that
do not have an explicit value.

Closes #15411

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -812,7 +812,10 @@ NgModelController.prototype = {
 
     if (isNumber(debounceDelay[trigger])) {
       debounceDelay = debounceDelay[trigger];
+    } else if (isNumber(debounceDelay['*'])) {
+      debounceDelay = debounceDelay['*'];
     } else if (isNumber(debounceDelay['default'])) {
+      // for backwards compatibility we fallback on the `default` trigger
       debounceDelay = debounceDelay['default'];
     }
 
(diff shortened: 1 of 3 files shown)


petebacondarwin:
Closing as we would rather have BC and not allow falling back to `default`

PR #15424, opened 2016-11-23
fix($rootScope): when adding/removing watchers during $digest

### LINKED ISSUE #15422: digest is skipping watches when a watch addes watches in its watch function.

Bug

because the $digest function works from length to 0, counting down and the $watch() uses unshift() to prepend the new watch one digest loop will skip watches.. (and executes them in a next one and that can be in a loop hitting the max 10)

as an example i have 100 watches, the current counter (the length variable) is on 75
that 75 watch is then adding 25 watches those are prepended so now i have 125 watches and the new once are 0-25. The length counter is then decremented to 74, But on the 74 position is not the value that was there before but suddenly what was 49, so 50-74 are all skipped.

This is problematic for us because in those 50-74 we have the same kind of watches that all add a few watches. So if that are 10+ then we will get the "digest cycle aborted because of 10 nested calls"
Which is in our eyes not really the case, it should do them as one digest. (or 2 doing all the added watches one time after the first one)

http://plnkr.co/edit/ojUQkuVvLLWwxxoksdJK?p=preview

see this plnkr i just expect that this runs with 1 or maybe 2 digest cycles, but now it aborts because it hits 10

The current behavior with the length counter just counting down in a loop, is not really correct it should be more like a concurrent iterator. even if there are more added it should never skip.
It should always really get the next one and the next one that was really in line.

Personally i could live with that the watch doesn't prepend but just append, that would mean that all the watches that are generated by watches are appended and skipped in that loop, but those are then executed all the next time. IF those then add again watches and that would hit the 10 digest cycles that would be a problem in the code.. That is a developers problem. (you shouldn't not constantly add watches) . But we don't do that we only do it once. That we expect are done in 1 digest cycle.

Angular 1.5.8 (all browsers)

gkalpak:
Bug fix.

Adding a watcher during a `$digest` (i.e. from within a watcher), results in the next watcher getting skipped. Similarly, removing a watcher during a `$digest` could result in the current watcher being run twice (if the removed watcher had not run yet in the current `$digest`).

Adding/Removing watchers during a `$digest` works as expected.

No.

**Other information**:
This commit fixes both cases by keeping track of the current watcher index during a digest and properly updating it when adding/removing watchers.

There might be a tiny extra overhead due to looking up `$$digestWatchIndex` on `scope.$$watchers` during the `$digest`. 

Fixes #15422

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -413,15 +413,21 @@ function $RootScopeProvider() {
 
         if (!array) {
           array = scope.$$watchers = [];
+          array.$$digestWatchIndex = -1;
         }
         // we use unshift since we use a while loop in $digest for speed.
         // the while loop reads in reverse order.
         array.unshift(watcher);
+        array.$$digestWatchIndex++;
         incrementWatchersCount(this, 1);
 
         return function deregisterWatch() {
-          if (arrayRemove(array, watcher) >= 0) {
+          var index = arrayRemove(array, watcher);
+          if (index >= 0) {
             incrementWatchersCount(scope, -1);
+            if (index < array.$$digestWatchIndex) {
+              array.$$digestWatchIndex--;
+            }
           }
           lastDirtyWatch = null;
         };
@@ -754,7 +760,6 @@ function $RootScopeProvider() {
       $digest: function() {
         var watch, value, last, fn, get,
             watchers,
-            length,
             dirty, ttl = TTL,
             next, current, target = this,
             watchLog = [],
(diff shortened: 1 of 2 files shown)


jcompagner:
looks good to me, i will test (patch my 158 install) with that code tomorrow to see how it works but at first sight that should fix it

one thing: "should not skip watchers when removing a watcher during digest'"

thats not really the case, for removing its not really skipping right? its more about not executing the same watch again? i think this check:

expect(log).toEqual([1, 2, 1, 2]);

would be without your fix:

expect(log).toEqual([1, 1, 2, 1, 2]);

right?

jcompagner:
i patched my 158 install and i don't need the evalAsync patch that i now had in my code
it seems to work fine (also going through the application i don't notice anything weird)

Narretz:
We should definitely test this for performance.

jcompagner:
the only thing that is really changed for digest() is that instead of a local variable we now have a scoped variable on a object. Anybody know what that lookup/assignment cost? Is that really measurable ?

i think the code in add/remove watch shouldn't be to much to worry about because that shouldn't really be called in high volume (like the digest cycle)

gkalpak:
Yeah, the only change that matters (performance-wise) is the switch from a local variable to a property lookup. I doubt it will have any noticeable impact (compared to all other operations typically run alongside).

I meassured the duration of the digest (without any changes) in a finished instance of the `largetable/ngBind` benchmark (>40.000 watchers) and I didn't see any difference. In both cases (before and after this change) the duration was ~13-14ms.

petebacondarwin:
LGTM

PR #15427, opened 2016-11-23
feat(security): explicitly whitelist URL schemes for bootstrap.

mprobst:
Many browsers have some extension URL scheme. It is unclear how many of
those have the security issue of allowing parser-inserted loads of
extension URLs.

To be conservative, this code whitelists the URL schemes that are known
to be subject to CSP, i.e. the ones that are expected and safe.

Note: there is no change in tests as behavior does not change for any known URL.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1455,12 +1455,20 @@ function allowAutoBootstrap(document) {
   link.href = src;
   var scriptProtocol = link.protocol;
   var docLoadProtocol = document.location.protocol;
-  if ((scriptProtocol === 'resource:' ||
-       scriptProtocol === 'chrome-extension:') &&
-      docLoadProtocol !== scriptProtocol) {
-    return false;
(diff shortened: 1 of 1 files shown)


mprobst:
@Rob--W thanks for the suggestion by the way, whitelisting is indeed much better.

@rjamet, @koto FYI.

This change is motivated by the discussion in https://github.com/mozilla/addons-linter/issues/1000.

petebacondarwin (approved):
Lgtm

src/Angular.js
```diff
@@ -1455,12 +1455,20 @@ function allowAutoBootstrap(document) {
...
+  switch(scriptProtocol) {
+    case 'http:':
+    case 'https:':
+    case 'ftp:':
+    case 'blob:':
+    case 'file:':
+    case 'data:':
+      return true;
+    default:
+      false;
```
dcherman: Did you mean `return false`?

PR #15433, opened 2016-11-24
feat(minErr):  set max depth for angular error JSON stringify

### LINKED ISSUE #15402: Feature Request: set max depth for angular error JSON stringify
**I'm submitting a ...**
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
```

**Current behavior**
Currently, when a (very) deep object is thrown as part of an error. The error message attempts to stringify the object, this freezes browsers for a significant period of time, and then produces an error message, many thousands of lines long. See [screenshot](https://i.imgsafe.org/cefa793a60.png)

**Expected behavior**
If there were a default max depth to stringify to, a produced error message would only stringify to this depth from the original object.

**Minimal reproduction of the problem with instructions**
An object (`parent`) references many child objects in a key called `children`, each child references its parent object in a key called `parent`. Should an error be thrown on a child object (say a `Duplicates in a repeater are not allowed` error) the entire structure is stringified in the error section `Duplicate value: `.

Speedier error messages, that allow for quicker debugging.

**Please tell us about your environment:**
OSX 10.12.1, Sublime Text 3, BreezeJS

* **Angular version:** 1.5.6

* **Browser:** all
 
* **Language:** all (as far as im aware)

m-amr:
feature

https://github.com/angular/angular.js/issues/15402

**Other information**:

closes #15402


src/minErr.js
```diff
@@ -57,7 +58,7 @@ function minErr(module, ErrorConstructor) {
 
     for (i = SKIP_INDEXES, paramPrefix = '?'; i < templateArgs.length; i++, paramPrefix = '&') {
       message += paramPrefix + 'p' + (i - SKIP_INDEXES) + '=' +
-        encodeURIComponent(toDebugString(templateArgs[i]));
+        encodeURIComponent(toDebugString(templateArgs[i]), MAX_DEPTH);
```
gkalpak: `MAX_DEPTH` should be passed to `toDebugString`, not `encodeURIComponent`.

src/Angular.js
```diff
@@ -815,9 +815,11 @@ function arrayRemove(array, value) {
     </file>
   </example>
  */
-function copy(source, destination) {
+function copy(source, destination, maxDepth) {
   var stackSource = [];
   var stackDest = [];
+  var currentDepth = 0;
+  maxDepth = (maxDepth && isNumber(maxDepth) && maxDepth > 0) ? maxDepth : -1;
```
gkalpak: I would simplify this: `if (!isNumber(maxDepth)) maxDepth = NaN;`

src/Angular.js
```diff
@@ -840,43 +842,48 @@ function copy(source, destination) {
...
-    return copyRecurse(source, destination);
+    return copyRecurse(source, destination, currentDepth);
   }
 
-  return copyElement(source);
+  return copyElement(source, currentDepth);
+
+  function copyRecurse(source, destination, currentDepth) {
+    currentDepth++;
+    if (maxDepth !== -1 && currentDepth > maxDepth) {
```
gkalpak: By setting `maxDepth` to `NaN` when not specified, this condition could be: `if (currentDepth > maxDepth)`

src/stringify.js
```diff
@@ -2,10 +2,16 @@
...
 
-function serializeObject(obj) {
+function serializeObject(obj, maxDepth) {
   var seen = [];
+  var tempObj = {};
+  if (maxDepth && isNumber(maxDepth) && maxDepth > 0) {
+    copy(obj, tempObj, maxDepth)
+  } else {
+    tempObj = obj;
+  }
```
gkalpak: This could be simplified to:

```js
if (isNumber(maxDepth)) obj = copy(obj, {}, maxDepth);
```

src/Angular.js
```diff
@@ -840,43 +842,48 @@ function copy(source, destination) {
...
+      return typeof source;
```
gkalpak: I would prefer some less ambiguous value that also makes it clear there was more stuff that has been truncated. Also, `typeof` can be misleading or at least not so useful, because almost anything that is not a primitive will return `object`.

If we wanted to keep it simple, maybe something like `'...'`. If we wanted to get more fancy, trying to get the name of the value's constructor and returning something like `constructorName + '(...)'` would be nice, but probably not worth it.

src/minErr.js
```diff
@@ -34,6 +34,7 @@ function minErr(module, ErrorConstructor) {
   ErrorConstructor = ErrorConstructor || Error;
   return function() {
     var SKIP_INDEXES = 2;
+    var MAX_DEPTH = 1;
```
gkalpak: There should be a way to customize this (I am not sure what it should be) and the default should be a little higher imo (maybe 3 or 4).

gkalpak (changes requested):
On a high level it looks good. Still missing:

1. Tests.
2. Making the max depth configurable.

src/Angular.js
petebacondarwin: we are already using `...` for cyclic references so I think that we should do the same for this.

petebacondarwin:
Wouldn't it be simpler and more performant just to provide a specialized `replacer` function to the `JSON.stringify` method, which tracked depth? Or is that not possible?

gkalpak:
I couldn't think of a way to track depth from `jsonReplacer` (in a reasonable way) :(

petebacondarwin:
Yeah, fair enough.
This project https://github.com/Canop/JSON.prune ended up having to rewrite JSON.stringify completely!

src/minErr.js
m-amr: I think to make max depth configurable we can,

    var exampleMinErr = minErr('example');
    throw exampleMinErr('one', 'This {0} is {1}', foo, bar);

1. Refactor exampleMinErr to be 

    throw exampleMinErr({code:'one', message: 'This {0} is {1}', messageArgs:[foo, boo]});

and we can keep both styles working.


2.  Include the depth as a special string in the message like  {$maxDepth=3}   
'This {0} is {1} {$maxDepth=3}' so we get the depth
  
 throw exampleMinErr('one', 'This {0} is {1} {$maxDepth=3}', foo, bar);

this Depth will be applied to all the message.

Or Include a depth for each placeholder like 'This {0:d8} is {1:d8} }'

    throw exampleMinErr('one', 'This {0:d8} is {1:d8}', foo, bar);
d8 means that the depth for stringifying this object is 8.

I think the second solution is easier for implementation and have no backward compatibility Issue,  Any Suggestion about that ?

m-amr:
@gkalpak 
 I have added some unit tests and we can configure the max depth by passing it in the placeholder

``` javascript
var exampleMinErr = minErr('example');
throw exampleMinErr('one', 'This {0,maxDepth=2}', foo);
```

Could please give this PR another look ?

test/AngularSpec.js
```diff
@@ -602,6 +602,21 @@ describe('angular', function() {
       expect(copy(new Number(NaN)).valueOf()).toBeNaN();
       /* eslint-enable */
     });
+
+    it('should copy source until reaching a given max depth', function() {
+      var destinationA = {};
+      var destinationB = {};
+
+      destinationA = copy({a1:1, b1:{b2:{b3:1}}, c1:[1,{c2:1}], d1:{d2:1}}, destinationB, 1);
+      expect(destinationA).toEqual({a1:1, b1:'...', c1:'...', d1:'...'});
```
gkalpak: This is not necessary, since `destinationA === destinationB` (and testing that is out-of-scope for this test).

test/AngularSpec.js
```diff
@@ -602,6 +602,21 @@ describe('angular', function() {
...
+      expect(destinationB).toEqual({a1:1, b1:'...', c1:'...', d1:'...'});
+
+      destinationA = {};
+      destinationB = {};
+      destinationA = copy({a1:1, b1:{b2:{b3:1}}, c1:[1,{c2:1}], d1:{d2:1}}, destinationB, 2);
+      expect(destinationA).toEqual({a1:1, b1:{b2:'...'}, c1:[1,'...'], d1:{d2:1}});
+      expect(destinationB).toEqual({a1:1, b1:{b2:'...'}, c1:[1,'...'], d1:{d2:1}});
+    });
```
gkalpak: Add a few more tests to cover more cases. E.g.:

- `maxDepth: 4`
- `maxDepth: 0`
- `maxDepth: -1`
- `maxDepth: NaN`
- `maxDepth: true`
- `maxDepth: null`
- no `maxDepth`

test/minErrSpec.js
```diff
@@ -68,6 +68,15 @@ describe('minErr', function() {
     expect(myError.message).toMatch(/a is {"b":{"a":"..."}}/);
   });
 
+  it('should handle arguments that are objects with max depth', function() {
+    var a = { b: {c:{d:1}} };
```
gkalpak: Nit: Inconsistent spacing. (Recommended: `{b: {c: {d: 1}}}`)

src/minErr.js
```diff
@@ -34,19 +34,32 @@ function minErr(module, ErrorConstructor) {
...
+      argsDebuggedStringIndexes = [],
       code = templateArgs[0],
       message = '[' + (module ? module + ':' : '') + code + '] ',
       template = templateArgs[1],
       paramPrefix, i;
 
-    message += template.replace(/\{\d+\}/g, function(match) {
-      var index = +match.slice(1, -1),
-        shiftedIndex = index + SKIP_INDEXES;
+    function getArgumentAndConvertToDebugString(index, maxDepth) {
```
gkalpak: It doesn't make much sense to make `maxDepth` configurable per argument (or even per call).
What we need is for the user to be able to configure it globally. `minErr` is private, so it is Angular that calls `minErr`, not the user and having to configure `maxDepth` for each argument is a burden, not a feature.

The first things that come to mind for making `maxDepth` globally configurable are:

1. Adding a new helper function on `angular`. E.g. `angular.setObjectMaxDepthInErrors()` (the name could be better - but you get the idea :smiley:)
2. Adding a new service/provider for configuring `maxDepth` (but it might miss some early errors, before bootstrapping the app).
3. Adding it as a property in the app's bootstrap config (for manual bootstrapping). This would be the best if it weren't for automatically bootstrapped apps (via `ngApp`). Now we need a `ng-object-max-depth-in-errors` directive (similar to `ng-strict-di`), which is not cool.

I think I prefer (1). Seems to be the simplest to implement and having the less shortcomings.

gkalpak (changes requested):
I left some comments for things that need to change imo. In addition to them, we also need test for `serializeObject`/`toDebugString`.

##
Note to self:
If we update `minErr`, we also need to update [ng-closure-runner/minErr][1] (and release and use new version in `angular.js`). 

[1]: https://github.com/angular/ng-closure-runner/blob/master/src/minErr.js

src/minErr.js
gkalpak: I don't think it'll be useful in this case. See https://github.com/angular/angular.js/pull/15433/files#r97893306.

src/minErr.js
m-amr: @gkalpak  Thanks for your Feedback, 
I prefer Option(1) too but,
I would like to ask if we use option (1) how we can add tests for it ?
I mean how we can change the object max depth during the tests ?
Is there any case already implemented like this ?

gkalpak: Usually, `anguar.xyz` helpers are assigned to a `xyz` function that is available inside the Angular closure (and globally in tests, because there is no closure). See `copy` for example.

I expect the code to look something like this:

```js
var errorMaxDepth = ...;
function setErrorMaxDepth(newErrorMaxDepth) {
  errorMaxDepth = newErrorMaxDepth;
}

function minErr(...) {
  // Use `errorMaxDepth`...
}
```

##
Now in tests, you have direct access to both `errorMaxDepth` and `setErrorMaxDepth`:

```js
var originalErrorMaxDepth = errorMaxDepth;

afterEach(function() {
  errorMaxDepth = originalErrorMaxDepth;
});

it('...', function() {
  setErrorMaxDepth(...);
  expect(testError(...).message).toMatch(...);
});
```

m-amr: @gkalpak Thanks for helping, 
I forget to add configureMinErr in test/.eslintrc.json so i couldn't use inside the tests,

I have fixed all things that you left in your comments.
Please tell me if there is something else missing.

src/Angular.js
```diff
@@ -125,6 +127,30 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
 
+var OBJECT_MAX_DEPTH_IN_ERROR_MESSAGE = 5;
+
+/**
+ * @ngdoc function
+ * @name angular.configureMinErr
+ * @module ng
+ * @kind function
+ *
+ * @description Use this function to configure minErr
```
gkalpak: Note to self: More details needed.

src/Angular.js
```diff
@@ -125,6 +127,30 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
+ *
+ * @param {Object=} config an object for defining configuration for the minErr. The
```
gkalpak: `minErr` is private/undocumented, so most users won't know what it is. Change to something more descriptive.

src/.eslintrc.json
```diff
@@ -15,6 +15,8 @@
     "splice": false,
     "push": false,
     "toString": false,
+    "OBJECT_MAX_DEPTH_IN_ERROR_MESSAGE":false,
```
gkalpak: Nit: Add space after `:` (here and elsewhere).

gkalpak (changes requested):

src/minErr.js
```diff
@@ -36,17 +36,27 @@ function minErr(module, ErrorConstructor) {
     var SKIP_INDEXES = 2;
 
     var templateArgs = arguments,
+      argsDebuggedStringIndexes = [],
       code = templateArgs[0],
       message = '[' + (module ? module + ':' : '') + code + '] ',
       template = templateArgs[1],
       paramPrefix, i;
 
+    function getTemplateArgAndConvertToDebugStringFormat(index) {
```
m-amr: I tried to Improve it because i found that toDebug method is called 2 times for every argument times 
One inside the template replace callback

    message += template.replace(/\{\d+\}/g, function(match) {
        var index = +match.slice(1, -1),
          shiftedIndex = index + SKIP_INDEXES;
  
        if (shiftedIndex < templateArgs.length) {
         return toDebugString(templateArgs[shiftedIndex]);
        }

and during encoding the url

    for (i = SKIP_INDEXES, paramPrefix = '?'; i < templateArgs.length; i++, paramPrefix = '&') {
        message += paramPrefix + 'p' + (i - SKIP_INDEXES) + '=' +
         encodeURIComponent(toDebugString(templateArgs[i]));
      } 

So toDebugString is doing a lot of work (copying and stringifying), i thought if it would be better if i cache the debugged string since i will use it again during encoding the url.

you are right 
"this is an exceptional situation (we are throwing an error), so performance is not a concern."

Do you want me to undo these changes ?

gkalpak: Oh, I see. It doesn't matter much, because most of the time the arguments will be strings (which are really cheap to call `toDebugString()` on), but I agree there is unnecessary overhead.
Yet your implementation seems too complicated (especially given that we are going to use every argument at least once), how about we simplify it like this:

```
-    var templateArgs = arguments,
-        code = templateArgs[0],
-        message = '[' + (module ? module + ':' : '') + code + '] ',
-        template = templateArgs[1],
+    var code = arguments[0],
+        message = '[' + (module ? module + ':' : '') + code + '] ',
+        template = arguments[1],
+        templateArgs = sliceArgs(arguments).map(function(arg) { return toDebugString(arg, objectMaxDepthInErrorMessage); }),
```

m-amr: @gkalpak 
I did all that changes that you had requested.

src/Angular.js
```diff
@@ -125,6 +127,30 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
 
+var objectMaxDepthInErrorMessage = 5;
+
+/**
+ * @ngdoc function
+ * @name angular.configureErrorHandling
+ * @module ng
+ * @kind function
+ *
+ * @description Use this function to configure minErr
```
gkalpak: Update the description.

src/Angular.js
```diff
@@ -125,6 +127,30 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
+ *
+ * @param {Object=} config an object for defining configuration for error handling. The
```
gkalpak: _an object_ --> _An object_

src/minErr.js
```diff
@@ -35,18 +35,20 @@ function minErr(module, ErrorConstructor) {
...
 
-    var templateArgs = arguments,
-      code = templateArgs[0],
+    var code = arguments[0],
       message = '[' + (module ? module + ':' : '') + code + '] ',
-      template = templateArgs[1],
-      paramPrefix, i;
+      template = arguments[1],
+      paramPrefix, i,
+      templateArgs = slice.call(arguments, SKIP_INDEXES).map(function(arg) {
```
gkalpak: Why not `sliceArgs`?

src/Angular.js
```diff
@@ -125,6 +127,30 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
+ *     following keys are supported:
+ *
+ * * `objectMaxDepth` - used to stringify objects in error messages until reaching the max depth.
```
gkalpak: Mention the default value (`Default: 5`).

src/Angular.js
```diff
@@ -125,6 +127,30 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
+ *
+ */
+function configureErrorHandling(config) {
+  if (isObject(config)) {
+    if (isNumber(config.objectMaxDepth) && config.objectMaxDepth > 0) {
```
gkalpak: It is now impossible to unset it (which means that there will always be copying in `toDebugString`). I would change this to something like:

```js
objectMaxDepthInErrorMessage = isValidObjectMaxDepth(config.objectMaxDepth) ? config.objectMaxDepth : NaN;

function isValidObjectMaxDepth(maxDepth) {
  return isNumber(maxDepth) && maxDepth > 0;
}
```

<sub>And you can reuse `isValidObjectMaxDepth()` inside `copy()` and `serializeObject()` for DRY-ness and consistency.
</sub>

src/stringify.js
```diff
@@ -2,9 +2,15 @@
...
 /* global toDebugString: true */
 
-function serializeObject(obj) {
+function serializeObject(obj, maxDepth) {
   var seen = [];
 
+  // There is no direct way to stringify object until reaching a specific depth
+  // and a very deep object can cause a performance issue, so we copy the object
+  // based on this specific depth and then stringify it.
+  if (maxDepth > 0) {
```
gkalpak: Unlikely to happen, but in order to avoid unnecessry copying (e.g. if `maxDepth === '1'), let's be consistent with this condition is all places (e.g. `copy()`, `serializeObject()`, `configureErrorHandling()`).

test/minErrSpec.js
```diff
@@ -9,6 +9,11 @@ describe('minErr', function() {
   var emptyTestError = minErr(),
     testError = minErr('test');
 
+  var originalObjectMaxDepthInErrorMessage = objectMaxDepthInErrorMessage;
+  afterEach(function() {
+    configureErrorHandling({objectMaxDepth: originalObjectMaxDepthInErrorMessage});
```
gkalpak: I probably said you could it like this (via `configureErrorHandling`), but you can actually set `objectMaxDepthInErrorMessage` directly).

test/minErrSpec.js
```diff
@@ -68,6 +73,27 @@ describe('minErr', function() {
...
+  });
+
+  they('should handle arguments that are objects and ignore max depth when objectMaxDepth = $prop',
+      [NaN, null, undefined, true, false, -1, 0], function(maxDepth) {
+        var a = {b: {c: {d: 1}}};
+
+        configureErrorHandling({objectMaxDepth: maxDepth});
+        var myError = testError('26', 'a is {0}', a);
+        expect(myError.message).toMatch(/a is {"b":{"c":{"d":1}}}/);
+  });
```
gkalpak: Add a test about the default `maxDepth` value.

gkalpak (changes requested):
Almost done :smiley:

test/minErrSpec.js
m-amr: I have added this lines

      configureErrorHandling({objectMaxDepth: originalObjectMaxDepthInErrorMessage});
      myError = testError('26', 'a when objectMaxDepth is default=5 is {0}', a);
      expect(myError.message).toMatch(/a when objectMaxDepth is default=5 is {"b":{"c":{"d":     {"e":{"f":"..."}}}}}/);

to the previous case.

m-amr:
Thanks for your feedback 😄 
I have done all the changes that you had requested.
Please tell me if there is something else missing.

src/Angular.js
```diff
@@ -125,6 +128,39 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
+ * * `objectMaxDepth` - used to stringify objects in error messages until reaching the max depth.
+ * if it is not supplied the default max depth which is 5 will be used instead, or if it is invalid
+ * (less than `1` or not a `Number`) no max depth will be used.
+ *
+ */
+function configureErrorHandling(config) {
+  if (isObject(config)) {
+    objectMaxDepthInErrorMessage = isValidObjectMaxDepth(config.objectMaxDepth) ? config.objectMaxDepth : NaN;
+  }
+}
```
gkalpak: With the prospect of #15590 being solved via this method, I think we need to revise the API to make it easily support what we need. Basically, we need it to support the following:

1. Getting the current values.
2. Setting specific values (so users don't have to specify all of them every time).

##
I suggest something along these lines:

```js
/**
 * @ngdoc function
 * @name angular.errorHandlingConfig
 * @module ng
 * @kind function
 *
 * @description
 * Configure several aspects of error handling in AngularJS if used as a setter or return the
 * current configuration if used as a setter. The following options are supported:
 *
 * - **objectMaxDepth**: The maximum depth to which objects are traversed when stringified for error
 *                       messages.
 * - **someOtherOption**: Blah, blah, blah.
 * - ...
 *
 * Omitted or undefined options will leave the corresponding configuration values unchanged.
 *
 * @param {Object=} config - The configuration object. MAy only contain the options that need to be
 *     updated. Supported keys:
 *     - `objectMaxDepth` **{Number}** - The max depth for stringifying objects. Setting to a
 *       non-positive or non-numeric value, removes the max depth limit.
 *       Default: 5
 *     - `someOtherOption` **{SomeType}** - Blah, blah, blah.
 *       Default: someDefaultValue
 *     - ...
 */
function errorHandlingConfig(config) {
  if (config) {
    return {
      objectMaxDepth: objectMaxDepthInErrorMessage
    };
  } else {
    if (isDefined(config.objectMaxDepth)) {
      objectMaxDepthInErrorMessage = isValidObjectMaxDepth(config.objectMaxDepth) ? config.objectMaxDepth : NaN;
    }
    if (isDefined(config.someOtherOption)) {
      /* someOtherOption = config.someOtherOption */
    }
    ...
  }
}
```

WDYT?

m-amr: Did you mean !config in ??

``` javascript
if (config) {
    return {
      objectMaxDepth: objectMaxDepthInErrorMessage
    };
  } 
```

you want to return configuration in case of config arg is not supplied ?

m-amr: And i would like to ask why we need to reset max depth in case of invalid argument ?

``` javascript
function configureErrorHandling(config) {
  if (isObject(config)) {
    objectMaxDepthInErrorMessage = isValidObjectMaxDepth(config.objectMaxDepth) ? config.objectMaxDepth : NaN;
  }
}
```

I suggest another constant called DEFAULT_OBJECT_MAX_DEPTH_IN_ERROR_MESSAGE
and use this value in case of invalid objectMaxDepth

``` javascript
function configureErrorHandling(config) {
  if (isObject(config)) {
    objectMaxDepthInErrorMessage = isValidObjectMaxDepth(config.objectMaxDepth) ? config.objectMaxDepth :DEFAULT_OBJECT_MAX_DEPTH_IN_ERROR_MESSAGE;
  }
}
```

And we can supply large maxDepth like 1000 if we want to stringify all the object ?

gkalpak: I don't see much benefit in that. Setting a large `maxDepth`, will also unnecessarily copy potentially deep objects without reason. I prefer to keep it simple. Also, setting a replacing an invalid value with a default value o not consistent with other places (we never do that iirc).

m-amr: @gkalpak 
I have changed the implementation to
 
``` javascript
function errorHandlingConfig(config) {
  if (!isObject(config)) {
    return {
      objectMaxDepth: objectMaxDepthInErrorMessage
    };
  } else {
    if (isDefined(config.objectMaxDepth)) {
      objectMaxDepthInErrorMessage = isValidObjectMaxDepth(config.objectMaxDepth) ? config.objectMaxDepth : NaN;
    }
  }
}
``` 
so it will work as a setter when config (object type ) is supplied.
please tell me if there anything missing.

m-amr:
@gkalpak
I prefer your suggestion about using errorHandlingConfig is a getter and setter,
I have changed the implementation to

function errorHandlingConfig(config) {
  if (!isObject(config)) {
    return {
      objectMaxDepth: objectMaxDepthInErrorMessage
    };
  } else {
    if (isDefined(config.objectMaxDepth)) {
      objectMaxDepthInErrorMessage = isValidObjectMaxDepth(config.objectMaxDepth) ? config.objectMaxDepth : NaN;
    }
  }
}
so it will work as a setter when config (object type ) is supplied.
please tell me if there anything missing.

src/Angular.js
```diff
@@ -125,6 +128,49 @@ var VALIDITY_STATE_PROPERTY = 'validity';
...
+ * Configure several aspects of error handling in AngularJS if used as a setter or return the
+ * current configuration if used as a getter. The following options are supported:
+ *
+ * - **objectMaxDepth**: The maximum depth to which objects are traversed when stringified for error messages.
+ *
+ * Omitted or undefined options will leave the corresponding configuration values unchanged.
+ *
+ * @param {Object=} config - The configuration object. May only contain the options that need to be
+ *     updated. Supported keys:
+ * -     `objectMaxDepth`  **{Number}** - The max depth for stringifying objects. Setting to a
```
gkalpak: Why is the `dash` at the beginning of the line? Is it intentional?

petebacondarwin: I imagine it might have been cut-and-pasted from a diff?

m-amr: When the dash is at the beginning of the line it will be converted to dot

screen shot 2017-02-03 at 12 32 39 am copy

which will be consistent with angular.bootsrtrap config docs

screen shot 2017-02-03 at 12 40 53 am copy


But when it is not at the beginning

screen shot 2017-02-03 at 12 37 37 am copy

Do you want to change it ?

petebacondarwin: What if you change line 148 to remove the indentation too?

m-amr: I have change it, it is now identical to the docs of 'angular.bootstrap' config in the identation
please tell me if it needs to change ?

m-amr:
@petebacondarwin 
@gkalpak 
I have change it, it is now identical to the docs of 'angular.bootstrap' config in the identation
please tell me if it needs to change ?

src/Angular.js
```diff
@@ -807,6 +855,7 @@ function arrayRemove(array, value) {
  *                   Can be any type, including primitives, `null`, and `undefined`.
  * @param {(Object|Array)=} destination Destination into which the source is copied. If
  *     provided, must be of the same type as `source`.
+ * @param {Number=} maxDepth All properties of the source will be copied until reaching the max depth.
```
gkalpak: We discussed this internally and we decided that it is better not to make this parameter public at this point.
So, please remove any mentions from the `copy` docs.

src/Angular.js
```diff
@@ -125,6 +128,50 @@ var VALIDITY_STATE_PROPERTY = 'validity';
 
 var hasOwnProperty = Object.prototype.hasOwnProperty;
 
+var objectMaxDepthInErrorMessage = 5;
```
gkalpak: Please change this to be an object with one property per config option. E.g.:

```
var minErrConfig = {
  objectMaxDepth: 5
};
```

gkalpak (changes requested):
We discussed this during the team meeting. I suggested two minor changes and then this willbe good to go.

m-amr:
@gkalpak 
I did all the changes that you had requested.

1. Changed
``` javascript
var objectMaxDepthInErrorMessage = 5;
``` 
To be
``` javascript
var minErrConfig = {
  objectMaxDepth: 5
};
``` 
2. Removed maxDepth from `copy` docs

gkalpak (approved):

gkalpak:
Regarding my [earlier comment][1], i turns out that we don't need to change [ng-closure-runner/minErr][2], because the change in this PR are not applicable there. 


[1]: https://github.com/angular/angular.js/pull/15433#pullrequestreview-18514243
[2]: https://github.com/angular/ng-closure-runner/blob/master/src/minErr.js

PR #15435, opened 2016-11-25
chore(*): use Yarn for dependency management

petebacondarwin:

**Other information**:


Narretz:
Node 6, here we come!
Can we have an explicite mention of the upgrade to node 6 in one of the commit messages? Is it implied by the change to yarn?

scripts/validate-commit-msg.js
```diff
@@ -16,7 +16,7 @@ var util = require('util');
```
gkalpak: Is this file used any more (with `init-repo.sh` gone)? We could as well get rid of it altogether.

src/ng/directive/input.js
```diff
@@ -22,10 +22,10 @@ var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+(?:[+-
 //   8. Query
 //   9. Fragment
 //                 1111111111111111 222   333333    44444        555555555555555555555555    666     77777777     8888888     999
```
gkalpak: Nit: The last `5` should be a space.

scripts/jenkins/set-node-version.sh
```diff
@@ -3,5 +3,7 @@
 # Install nvm for this shell
 source ~/.nvm/nvm.sh
 
-# Use node.js at 4.2.x
-nvm install 4.4
+# Use version of node.js found in .nvmrc
+nvm install
+
+npm install -g yarn
```
gkalpak: Nit: This file now does more than its name suggests. Maybe rename it to something like `setup-node-env`?

gkalpak (commented):
:tada: BTW, we also need to update `docs/content/misc/contribute.ngdoc`.

petebacondarwin:
I'll add better commit messages. I am pretty sure that yarn requires node 6+.

Gruntfile.js
```diff
@@ -322,13 +318,6 @@ module.exports = function(grunt) {
     }
   });
 
-  // global beforeEach task
-  if (!process.env.TRAVIS) {
```
mgol: Shouldn't we run `yarn` here?

changelog.js
```diff
@@ -1,208 +0,0 @@
-#!/usr/bin/env node
-
```
mgol: Does this removal have anything to do with switching to Yarn? Looks like a general cleanup.

src/ngComponentRouter/Router.js
```diff
@@ -10,11 +10,11 @@
  *
  * @installation
  *
- * Currently, the **Component Router** module must be installed via `npm`, it is not yet available
+ * Currently, the **Component Router** module must be installed via `npm`/`yarn`, it is not yet available
```
mgol: Unrelated, but this "yet" is a little reaching... :) I'd remove the word.

package.json
```diff
@@ -9,8 +9,8 @@
     "url": "https://github.com/angular/angular.js.git"
   },
   "engines": {
-    "node": "<5",
-    "npm": "~2.5"
+    "node": "^6.9.0",
```
mgol: Maybe `^6.9.1`? 6.9.1 was released a day after 6.9.0 and contained critical fixes.

changelog.js
petebacondarwin: Yes, you're right, I will move it into separate commits

src/ngComponentRouter/Router.js
petebacondarwin: I'll do so in a separate commit :-)

package.json
petebacondarwin: OK

Gruntfile.js
```diff
@@ -322,13 +318,6 @@ module.exports = function(grunt) {
...
-    grunt.task.run('shell:npm-install');
```
gkalpak: FWIW, I think this was kind of useful...
<sub>(I can leave without it though.)</sub>

gkalpak (commented):
Added a few more nits :smiley:

Gruntfile.js
petebacondarwin: OK, that sounds right.

scripts/jenkins/release.sh
```diff
@@ -38,8 +38,7 @@ function build {
   source ./init-node.sh
   cd ../..
 
-  yarn
-  node_modules/.bin/grunt  ci-checks package --no-color
+  grunt  ci-checks package --no-color
```
gkalpak: Two spaces between 'grunt' and 'ci' 😱

scripts/jenkins/build.sh
```diff
@@ -21,22 +21,21 @@ rm -f angular.js.size
 
 
 # BUILD #
-yarn
-node_modules/.bin/grunt ci-checks package --no-color
+grunt ci-checks package --no-color
```
gkalpak: Is grunt available globally?

Gruntfile.js
```diff
@@ -292,10 +292,9 @@ module.exports = function(grunt) {
     },
 
     shell: {
-      'npm-install': {
-        command: 'node scripts/npm/check-node-modules.js'
+      'install-node-dependencies': {
+        command: 'yarn'
```
gkalpak: Won't this take more time that simply comparing two files (s was happening before). I am not very keen on sacrificing speed for _every_ task for this 😕

scripts/travis/build.sh
```diff
@@ -6,19 +6,19 @@ export BROWSER_STACK_ACCESS_KEY=`echo $BROWSER_STACK_ACCESS_KEY | rev`
...
   else
     BROWSERS="SL_Chrome,SL_Firefox,SL_Safari_8,SL_Safari_9,SL_IE_9,SL_IE_10,SL_IE_11,SL_iOS"
   fi
 
-node_modules/.bin/grunt test:promises-aplus
-node_modules/.bin/grunt test:unit --browsers="$BROWSERS" --reporters=dots
-node_modules/.bin/grunt tests:docs --browsers="$BROWSERS" --reporters=dots
+grunt test:promises-aplus
+grunt test:unit --browsers="$BROWSERS" --reporters=dots
+grunt tests:docs --browsers="$BROWSERS" --reporters=dots
```
gkalpak: Bad indentation 😛

Gruntfile.js
mgol: It shouldn't be a problem. The following was run on the angular.js repo with npm & yarn cache cleared:
```
$ time yarn --ignore-scripts
yarn --ignore-scripts  28.22s user 21.07s system 125% cpu 39.178 total
$ time yarn --ignore-scripts
yarn --ignore-scripts  6.49s user 2.68s system 120% cpu 7.583 total
$ time yarn --ignore-scripts
yarn --ignore-scripts  0.87s user 0.07s system 70% cpu 1.321 total
$ time yarn --ignore-scripts
yarn --ignore-scripts  0.86s user 0.07s system 70% cpu 1.313 total
```

i18n/generate.sh
```diff
@@ -3,14 +3,13 @@
 set -e
 
 BASE_DIR=`dirname $0`
-cd $BASE_DIR
```
mgol: This reminded me that scripts like this one are dangerous unless you add `set -u` at the beginning. Otherwise if suddenly a variable like that gets unset and later used, you may get unexpected results.

scripts/jenkins/init-node.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Install nvm for this shell
+source ~/.nvm/nvm.sh
+
+# Use version of node.js found in .nvmrc
+nvm install
+
+npm install -g yarn
```
mgol: We should install a version matching what we have in package.json's `engines.yarn` field.

mgol (changes requested):

scripts/jenkins/init-node.sh
mgol: [jq](https://stedolan.github.io/jq/) is a good tool for that. You can play with it at https://jqplay.org/

Gruntfile.js
gkalpak: Right :+1:

When trying this myself, I stopped after the 2nd time :smiley:

mgol: Could you try? It'd be good to make sure it's that quick not only on *NIX-es.

gkalpak: I did try of course. I didn't take your word for it 😛
It "stabilizes" at 2secs on my machine, which is OK (for me at least).

mgol: I wonder why it's not that fast on the second try, though. I'd expect it to behave identically in all situations where cache is populated and dependencies are correct so something looks wrong here.

mgol: Let's see: https://github.com/yarnpkg/yarn/issues/2054

docs/content/misc/contribute.ngdoc
```diff
@@ -62,7 +69,7 @@ cd angular.js
 git remote add upstream "https://github.com/angular/angular.js.git"
 
 # Install node.js dependencies:
-npm install
+yarn
```
petebacondarwin: Hmm, the Yarn docs generally recommend just `yarn` but OK.

docs/content/misc/contribute.ngdoc
```diff
@@ -62,7 +69,7 @@ cd angular.js
...
 
 # Install bower components:
 bower install
```
petebacondarwin: removed.

scripts/jenkins/init-node.sh
mgol: Some environments, like Heroku, use `engines` to decide what version to use. I guess it won't happen in our situation but we should still specify the version somewhere and it doesn't hurt to do it in `engines` and read from there; if someone wanted to do sth on one of these engines-reading tools, they could use these settings.

As for humans, we could always add a check for an installed version and require it to be not older than whatever's in `engines.yarn`.

.nvmrc
```diff
@@ -1,2 +1 @@
-4
-
+6.9.1
```
gkalpak: Isn't that too restrictive?

Gruntfile.js
```diff
@@ -322,13 +333,10 @@ module.exports = function(grunt) {
     }
   });
 
-  // global beforeEach task
```
gkalpak: You could leave this.

docs/content/misc/contribute.ngdoc
```diff
@@ -24,20 +24,27 @@ Before you can build AngularJS, you must install and configure the following dep
...
+  Installing Git](https://help.github.com/articles/set-up-git) is a good source of information.
 
-* [Node.js v4.x](http://nodejs.org): We use Node to generate the documentation, run a
-development web server, run tests, and generate distributable files. Depending on your system, you can install Node either from source or as a
-pre-packaged bundle. (Currently our build does not work properly on Node v5 or greater - please use v4.x.)
+* [Node.js v6.x (LTS)](http://nodejs.org): We use Node to generate the documentation, run a
+  development web server, run tests, and generate distributable files. Depending on your system,
+  you can install Node either from source or as a pre-packaged bundle.
+
+  We recommend using [nvm](https://github.com/creationix/nvm) to manage and install Node.js,
```
gkalpak: I've been using https://github.com/coreybutler/nvm-windows for some time now and it works fine. I think it is worth mentioning it too.

docs/content/misc/contribute.ngdoc
```diff
@@ -24,20 +24,27 @@ Before you can build AngularJS, you must install and configure the following dep
...
+  which makes it easy to change the version of Node.js per project.
+
+* [Yarn](https://yarnpkg.com): We use Yarn to install our Node.js module dependencies (rather than using npm).
```
gkalpak: It might be worth mentioning the installation command (`npm install -g yarn`).

package.json
```diff
@@ -9,16 +9,13 @@
     "url": "https://github.com/angular/angular.js.git"
   },
   "engines": {
-    "node": "<5",
-    "npm": "~2.5"
+    "node": "6.9.1",
```
gkalpak: Why restrict this to `6.9.1`?

scripts/travis/build.sh
```diff
@@ -14,9 +14,9 @@ elif [ "$JOB" == "unit" ]; then
     BROWSERS="SL_Chrome,SL_Firefox,SL_Safari_8,SL_Safari_9,SL_IE_9,SL_IE_10,SL_IE_11,SL_iOS"
   fi
 
-  grunt test:promises-aplus
-  grunt test:unit --browsers="$BROWSERS" --reporters=dots
-  grunt tests:docs --browsers="$BROWSERS" --reporters=dots
+grunt test:promises-aplus
+grunt test:unit --browsers="$BROWSERS" --reporters=dots
+grunt tests:docs --browsers="$BROWSERS" --reporters=dots
```
gkalpak: Bad indentation (this one never gets old 😛 )

gkalpak (approved):
A couple of minor comments/questions.
LGTM

gkalpak:
BTW, I have tested this on Windows (10) and Ubuntu (16.04) with `yarn && grunt test` and everything works fine.

docs/content/misc/contribute.ngdoc
Daniel15: @gkalpak - It's better to link to https://yarnpkg.com/en/docs/install. Installing Yarn via npm is not the recommended installation method and should only be used in cases where a proper package is not available for your environment.

scripts/jenkins/init-node.sh
Daniel15: Please avoid installing Yarn via npm if possible. This is significantly slower than all other methods of installing Yarn, and often results in weird issues around Yarn's dependencies. We (Yarn team) do not recommend it

If running on Debian or Ubuntu, you should use Yarn's Debian repository. Otherwise, use the installation script (`curl -o- -L https://yarnpkg.com/install.sh | bash`) or just extract the tarball manually. Both methods (Debian package repository and installation script) allow you to specify the exact version of Yarn to install.

.nvmrc
petebacondarwin: 6.9 is the LTS version of node. There are actually only two versions node of 6.9: 6.9.0 and 6.9.1 The former has issues that need to be fixed in 6.9.1. I don't see any reason why we should let people use versions below 6.9.

petebacondarwin: Moreover `nvm` is not capable of parsing `^6.9.1` so we would be stuck with `6.9` or worse `6`, which could allow versions that we don't want such as 6.8.x and 6.9.0.

Gruntfile.js
petebacondarwin: OK, although I found it slightly confusing due to its similarity to testing terminology - like it was something special. How about I modify it to something more humane?

src/ng/directive/input.js
```diff
@@ -21,11 +21,11 @@ var ISO_DATE_REGEXP = /^\d{4,}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+(?:[+-
 //   7. Path
 //   8. Query
 //   9. Fragment
-//                 1111111111111111 222   333333    44444        555555555555555555555555    666     77777777     8888888     999
-var URL_REGEXP = /^[a-z][a-z\d.+-]*:\/*(?:[^:@]+(?::[^@]+)?@)?(?:[^\s:/?#]+|\[[a-f\d:]+\])(?::\d+)?(?:\/[^?#]*)?(?:\?[^#]*)?(?:#.*)?$/i;
+//                 1111111111111111 222   333333    44444        55555555555555555555555     666     77777777     8888888     999
```
Daniel15: The lack of named capture groups in JavaScript regular expressions is really annoying :P

Daniel15:
Please avoid installing Yarn via npm if possible. This is not recommended. It's significantly slower than all other methods of installing Yarn, and often results in weird issues around Yarn's dependencies.

If running on Debian or Ubuntu, you should use Yarn's Debian repository. Otherwise, use the installation script (`curl -o- -L https://yarnpkg.com/install.sh | bash`) or just extract the tarball manually. Both methods (Debian package repository and installation script) allow you to specify the exact version of Yarn to install.

.nvmrc
gkalpak: But it means that we always need to have (and switch to) 6.9.1 locally, even when 6.9.2+ comes out 😞

scripts/travis/build.sh
petebacondarwin: Just keeping you interested :-P

petebacondarwin:
@Daniel15 - thanks for the feedback about installing Yarn. Do you have any input on how we could lock down the version of Yarn that we are using to install our dependencies?

Daniel15:
It depends on how you're installing Yarn. If you use the Debian repo, you
can pass the version number to apt-get (eg. `apt-get install
yarn=0.17.8-1`). If you use the install script, it can take a version as a
command line argument (eg. `./install.sh --version 0.17.8`).

Locking down the Yarn version used on CI servers is a good idea.

scripts/travis/before_build.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+yarn global add grunt-cli
```
mgol: Let's lock the version. ;-)

mgol (changes requested):

petebacondarwin:
@Daniel15 - thanks. I think we are going to go with:

```
curl -o- -L https://raw.githubusercontent.com/yarnpkg/yarn/v0.17.9/scripts/install-latest.sh | bash -s -- --version 0.17.9
```

which ensures that the install script is also locked to the version being installed.

Daniel15:
@petebacondarwin - The version of `install-latest.sh` in the tag is missing some features, such as GPG verification and the `--version` flag. I don't think we cherry-picked any of the changes to the install script into the stable branch, as we expect everyone to use https://yarnpkg.com/install.sh rather than directly referencing the script on GitHub. Eventually the script will move out of the Yarn repo into the releases repo, where the Debian repo is also kept.


If you like, you could just have your own small script that downloads and extracts Yarn, based off the install-latest.sh script. Or, just grab https://yarnpkg.com/install.sh and check it into your repo.

mgol:
@Daniel15 Pete & me have tried using the install script but it doesn't work... I've tried downloading it & evaluating:

```shell
$ curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.17.9
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    94  100    94    0     0    205      0 --:--:-- --:--:-- --:--:--   206
100  6074  100  6074    0     0   7515      0 --:--:-- --:--:-- --:--:-- 43078
Installing Yarn!
bash: line 164: yarn: command not found
```

as well as downloading first & then using:

```shell
$ curl -O -L https://yarnpkg.com/install.sh
$ chmod +x install.sh
$ ./install.sh --version 0.17.9
Installing Yarn!
./install.sh: line 164: yarn: command not found
```

Daniel15:
@mgol - Looks like there's a bug in the script when `~/.yarn` exists but there's no Yarn version installed in it. I just filed an issue for it: https://github.com/yarnpkg/yarn/issues/2061

scripts/travis/before_build.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-yarn global add grunt-cli
+yarn global add grunt-cli@1.2.0
```
gkalpak: AFAICT, this will use the latest 1.x release. Is that intended or should we use the [`--exact` option][1]?

[1]: https://yarnpkg.com/en/docs/cli/add#toc-yarn-add-exact-e

scripts/travis/build.sh
```diff
@@ -14,9 +14,9 @@ elif [ "$JOB" == "unit" ]; then
     BROWSERS="SL_Chrome,SL_Firefox,SL_Safari_8,SL_Safari_9,SL_IE_9,SL_IE_10,SL_IE_11,SL_iOS"
   fi
 
-grunt test:promises-aplus
-grunt test:unit --browsers="$BROWSERS" --reporters=dots
-grunt tests:docs --browsers="$BROWSERS" --reporters=dots
+  grunt test:promises-aplus
+  grunt test:unit --browsers="$BROWSERS" --reporters=dots
+  grunt tests:docs --browsers="$BROWSERS" --reporters=dots
```
gkalpak: Good indentation! 😛

scripts/travis/before_build.sh
mgol: This is weird, npm in this case would install the provided version. Also, Yarn docs are contradicting each other in the very same page:

> `yarn add package-name@1.2.3` installs a specific version of a package from the registry.

vs:

> For example, `yarn add foo@1.2.3` would accept version `1.9.1`, but `yarn add foo@1.2.3 --exact` would only accept version `1.2.3`.

@Daniel15 could you clarify which one is right?

mgol: No, it's CLI, we should lock everything as much as possible.

gkalpak: Trying this locally, it seems that it will install the specific version (i.e. `--exact` makes no difference).
Probably a docs issue.

scripts/travis/before_build.sh
petebacondarwin: locked!

mgol (approved):

Gruntfile.js
```diff
@@ -10,26 +10,63 @@ var path = require('path');
...
+var exec = require('shelljs').exec;
+var pkg = require(__dirname + '/package.json');
 
-var useNodeVersion = fs.readFileSync('.nvmrc', 'utf8');
-if (!semver.satisfies(process.version, useNodeVersion)) {
-  throw new Error('Invalid node version; please use node v' + useNodeVersion);
+// Node.js version checks
+if (!semver.satisfies(process.version, pkg.engines.node)) {
+  reportOrFail('Invalid node version (' + process.version + '). ' +
+               'Please use a version that satisfies ' + pkg.engines.node);
```
gkalpak: Nit: Missing trailing period :grin:
Nit: I would make the message a little "softer". For example, the build will (hopefully) work fine with v7+; calling it an "invalid version" sounds "strong".

Gruntfile.js
```diff
@@ -10,26 +10,63 @@ var path = require('path');
...
 }
 
+// Yarn version checks
+var expectedYarnVersion = pkg.engines.yarn;
+var currentYarnVersion = exec('yarn --version', {silent: true}).stdout.trim();
+if (!semver.satisfies(currentYarnVersion, expectedYarnVersion)) {
+  reportOrFail('Invalid yarn version (' + currentYarnVersion + '). ' +
+               'Please use a version that satisfies ' + expectedYarnVersion);
```
gkalpak: Nit: Missing trailing period.

Gruntfile.js
```diff
@@ -10,26 +10,63 @@ var path = require('path');
...
+var expectedGruntVersion = pkg.engines.grunt;
+var currentGruntVersions = exec('grunt --version', {silent: true}).stdout;
+var match = /^grunt-cli v(.+)$/m.exec(currentGruntVersions);
+if (!match) {
+  reportOrFail('Unable to compute the current grunt-cli version. We found:\n' +
+               currentGruntVersions);
+} else {
+  if (!semver.satisfies(match[1], expectedGruntVersion)) {
+  reportOrFail('Invalid grunt-cli version (' + match[1] + '). ' +
+               'Please use a version that satisfies ' + expectedGruntVersion);
```
gkalpak: Nit: Missing trailing period.

gkalpak (approved):
LGTM

Gruntfile.js
petebacondarwin: You're too kind! If you feel strongly about it I will let you fix up the messaging in a future commit.

PR #15438, opened 2016-11-26
chore(docs): use correct script-URL for plnkr on snapshot

### LINKED ISSUE #15437: All the sample plunker from API Reference page examples aren't working
**Bug**

All the plunker which are provided on [API Reference](https://docs.angularjs.org/api) page aren't working. Suppose take an example of [ngRepeat](https://docs.angularjs.org/api/ng/directive/ngRepeat) API, when click on **Edit Plunkr link** (provided on top of example code window)  it opens a plunker but that code isn't working because of wrong angular.js version reference.

**Below is the link taken from plunkr**

<script src="//ajax.googleapis.com/ajax/libs/angularjs/1.5.10-build.5166/angular.min.js"></script>

If you look at script tag inside a plunker it has version like this `1.5.10-build.5166`, which should be ideally something like `1.5.8`(stable version).

gkalpak:
Bug fix (for docs app).

The URLs used for scripts when opening an example from the snapshot version in plnkr is incorrect and the code does not work.
See #15437.

The correct script URLs are used and the examples work.

No.

**Other information**:
Fixes #15437

### DIFF
--- a/docs/config/services/deployments/production.js
+++ b/docs/config/services/deployments/production.js
@@ -15,7 +15,7 @@ var cdnUrl = googleCdnUrl + versionInfo.cdnVersion;
 // might not work (possibly in subtle ways).
 var examplesCdnUrl = versionInfo.currentVersion.isSnapshot ?
   (angularCodeUrl + 'snapshot') :
-  (googleCdnUrl + (versionInfo.version || versionInfo.currentVersion));
+  (googleCdnUrl + (versionInfo.currentVersion.version || versionInfo.currentVersion));
 
 module.exports = function productionDeployment(getVersion) {
   return {
--- a/docs/config/services/deployments/production.js
+++ b/docs/config/services/deployments/production.js
@@ -13,7 +13,7 @@ var cdnUrl = googleCdnUrl + versionInfo.cdnVersion;
 // The currentVersion may not be available on the cdn (e.g. if built locally, or hasn't been pushed
 // yet). This will lead to a 404, but this is preferable to loading a version with which the example
 // might not work (possibly in subtle ways).
-var examplesCdnUrl = versionInfo.isSnapshot ?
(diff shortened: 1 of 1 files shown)


docs/config/services/deployments/production.js
```diff
@@ -13,7 +13,7 @@ var cdnUrl = googleCdnUrl + versionInfo.cdnVersion;
 // The currentVersion may not be available on the cdn (e.g. if built locally, or hasn't been pushed
 // yet). This will lead to a 404, but this is preferable to loading a version with which the example
 // might not work (possibly in subtle ways).
-var examplesCdnUrl = versionInfo.isSnapshot ?
+var examplesCdnUrl = versionInfo.currentVersion.isSnapshot ?
   (angularCodeUrl + 'snapshot') :
   (googleCdnUrl + (versionInfo.version || versionInfo.currentVersion));
```
Narretz: Isn't this also wrong then? (there is no versionInfo.version)

Narretz:
LGTM. We can handle the comment later

docs/config/services/deployments/production.js
gkalpak: You are right. I couldn't find where `version` may have been defined, so I left it untouched in order to avoid breaking anything. But it is clear now that (a) `versionInfo` can never have a `version` property and (b) it is referring to the `version` property of the `SemVer` object (returned from `semver.parse(...)`) of `versionInfo.currentVersion`.

gkalpak: BTW, it doesn't matter in practice, because `Semver.prototype.toString()` just returns the `version` property, but I will change it to `versionInfo.currentVersion.version` just in case.

Narretz: @gkalpak That's right. Maybe we can simplify this later. For now, let's get the fix merged, so the examples start working again.

PR #15439, opened 2016-11-26
chore(*): silence ng-closure-runner warnings during `grunt minall`

gkalpak:
Chore/Refactor.

During the `grunt minall` task, `ng-closure-runner` logs some warnings due to:
1. Throwing errors that do not seem to be `minErr` instances.
2. A `@this` annotation in a non-JSDoc comment. 

No warnings during `grunt minall`.

No.

### DIFF
--- a/src/minErr.js
+++ b/src/minErr.js
@@ -79,7 +79,6 @@ function isValidObjectMaxDepth(maxDepth) {
  *   error from returned function, for cases when a particular type of error is useful.
  * @returns {function(code:string, template:string, ...templateArgs): Error} minErr instance
  */
-
 function minErr(module, ErrorConstructor) {
   ErrorConstructor = ErrorConstructor || Error;
   return function() {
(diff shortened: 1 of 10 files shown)


src/minErr.js
```diff
@@ -63,3 +62,38 @@ function minErr(module, ErrorConstructor) {
...
+ *                                  // but avoids `ng-closuer-runner` warnings.
+ * }
+ * ```
+ *
+ * @param {string} ignoredCode - Ignored, but necessary for `ng-closure-runner`.
+ * @param {string} ignoredTemplate - Ignored, but necessary for `ng-closure-runner`.
+ * @param {*} error - The error object that will be returned.
+ * @returns {*} - The passed in error.
+ */
+function noMinErr(ignoredCode, ignoredTemplate, err) {
```
gkalpak: I am not thrilled about having to pass the first two string arguments.
Alternatively, we could teach `ng-closure-runner` to ignore `noMinErr` instances (i.e. not treat them as `minErr` instances, but don't log a warning either).

gkalpak:
For reference:

**Before**
```
Running "minall" task
...
build/angular.js:5038: WARNING - Throw expression is not a minErr instance.
          throw err;
          ^
build/angular.js:8763: WARNING - Throw expression is not a minErr instance.
            throw errors;
            ^
build/angular.js:12860: WARNING - Throw expression is not a minErr instance.
            throw e;
            ^
build/angular.js:14531: WARNING - Throw expression is not a minErr instance.
        throw e;
        ^
build/angular.js:18772: WARNING - Throw expression is not a minErr instance.
            throw e;
            ^
build/angular.js:30307: WARNING - Parse error. Non-JSDoc comment has annotations. Did you mean to start it with '/**'?
    forEach(options, /* @this */ function(option, key) {
                     ^
0 error(s), 6 warning(s)
>> build/angular.js minified into build/angular.min.js
...
```

**After**
```
Running "minall" task
...
>> build/angular.js minified into build/angular.min.js
...
```

petebacondarwin:
Did you check whether this ends up creating `no` error pages in the docs?

gkalpak:
No 😕

gkalpak:
AFAICT, it does not. It does create a `no: {'': ''}` entry in `build/errors.json`, but as long as there is no corresponding error doc, no entry is created in `nav-data.json`.

petebacondarwin:
That is true but you now get this in the build:

```
warn:    No error doc exists for min error: no:
```

gkalpak:
Irony 😒 
This is still better than what we get today (and I am sure we can silence that warning 😈).
The main question is: Do we like this approach?

petebacondarwin:
Given that we are then creating workarounds for workarounds, I start to feel that it would be better to fix the minerr extraction tool instead.

gkalpak:
As discussed, it doesn't make sense to fix it in `ng-closure-runner` at this point.
I moved the jsdoc comment fix to #16575.

Issue #14352, opened 2016-03-31
angular.copy doesn't copy File object correctly


sbespaly:
**_Note_: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

angular.copy doesn't copy File object (part of file API) correctly. There was the same issue with Blob, but it was fixed.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**

var file = new File([], "filename");
var file2 = angular.copy(file);
//variable 'file2' is not valid, and can not be used with file API anymore.

**What is the expected behavior?**

copy function should create valid copy of File object, like it was made for Blob

**What is the motivation / use case for changing the behavior?**

File object and Blob are very common objects, but for some reason copy function doesn't copy File object, but do this for Blob's. This lead to unclear behavior of copy function.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**

1.5.3 release and latest sources on GitHub, 
All browsers affected
Don't have implementation in previous versions

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
add copy code for File object in function 
function copyType(source) in file Angular.js line:886

linuxjackto7:
Deadpool (2016) 720P Download ：https://movied.org/deadpool-2016_801438.html

lgalfaso:
I would like to know what the expectations of supporting this are. If the expectations are that if should be able to take a `File` that originates from `FileList` and this from `<input type="file">`, and that the new `File` will be able to read the content of the `<input>`, then it looks like this bug is hiding a large part of the work that is needed.

Note: `angular.copy` should not be the solution for all things that may need to be copied. There are  libraries that specialize in handling many of the browsers quirks and are better suited as a complete solution to clone objects.

thorn0:
Aren't files from `<input type="file">` read-only? Why copy them?

sbespaly:
[...]

PR #15440, opened 2016-11-27
docs($rootScope): add note about watching File objects

wesleycho:
- Add note recommending against watching `File` objects with deep watchers

This should address #14352

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -297,6 +297,8 @@ function $RootScopeProvider() {
        *   according to the {@link angular.equals} function. To save the value of the object for
        *   later comparison, the {@link angular.copy} function is used. This therefore means that
        *   watching complex objects will have adverse memory and performance implications.
+       * - This should not be used to watch for changes in objects that are
+       *   or contain `File` objects due to limitations with `angular.copy`.
        * - The watch `listener` may change the model, which may trigger other `listener`s to fire.
        *   This is achieved by rerunning the watchers until no changes are detected. The rerun
        *   iteration limit is 10 to prevent an infinite loop deadlock.


src/ng/rootScope.js
```diff
@@ -296,7 +296,7 @@ function $RootScopeProvider() {
        * - When `objectEquality == true`, inequality of the `watchExpression` is determined
        *   according to the {@link angular.equals} function. To save the value of the object for
        *   later comparison, the {@link angular.copy} function is used. This therefore means that
-       *   watching complex objects will have adverse memory and performance implications.
+       *   watching complex objects will have adverse memory and performance implications. This should not be used to watch for objects that are or contain `File` objects.
```
Narretz: I think this should go into its own bullet point, with an explanation why it's not recommended.
Or if it is not recommended because file objects use a lot of memory, then this should also be pointed out.

Narretz (changes requested):

wesleycho:
Updated as per request

src/ng/rootScope.js
```diff
@@ -297,6 +297,8 @@ function $RootScopeProvider() {
        *   according to the {@link angular.equals} function. To save the value of the object for
        *   later comparison, the {@link angular.copy} function is used. This therefore means that
        *   watching complex objects will have adverse memory and performance implications.
+       * - This should not be used to watch for changes in objects that are
+       *   or contain `File` objects due to limitations with `angular.copy`.
```
Narretz: Can you please add a link to angular.copy?

Narretz (changes requested):

Narretz:
Thanks @wesleycho, I've added the link myself. I didn't close the other issue because we should probably add docs to angular.copy itself, maybe a knownIssue for all types that are currently unsupported

Issue #14316, opened 2016-03-25
[Docs] Why is lowercase/uppercase deprecation not in changelog?


adgoncal:
Doc entries for angular.lowercase and angular.uppercase were removed in angular 1.5.0-rc.2 (See #11387, https://github.com/angular/angular.js/commit/6a92e9111f8eddc9cebaa1ee01c8453c891f23bd).

Why is this not in the changelog?

gkalpak:
Docs changes are usually not included in the changelog. On the other hand, this was the first time we removed something from the docs, but not from the code (afaik). It makes sense to add a deprecation warning in the changelog.

@adgoncal, would like to submit a PR for that ?

PR #15441, opened 2016-11-27
docs(changelog): add deprecation notice

wesleycho:
- Add deprecation notice for `angular.lowercase` and `angular.uppercase`

This should address #14316

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2028,6 +2028,7 @@ This version of AngularJS is problematic due to a issue during its release. Plea
 
 - The `ngTouch` module's `ngClick` directive has been deprecated and disabled by default. See the breaking
 changes section for more information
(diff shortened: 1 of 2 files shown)


Narretz:
So we did something strange with this. We deprecated the methods and removed their docs, which I think only adds to the confusion. Imo, ww should also re-add the docs, but give the methods the `@deprecated` tag

petebacondarwin:
Can we add the deprecation and removal versions to the deprecation notice?

Narretz:
I can do that. to be removed in 1.7?

petebacondarwin:
Yep - although I think we will need to leave the functions in-place internally.

mgol:
@petebacondarwin as long as they're not exposed, it doesn't matter. Once we no longer need them, we can remove them then even in a patch release.

Narretz:
Yes, we use them internally. So we should remove them from the public angular object in 1.7

Issue #13709, opened 2016-01-07
$q.when should be deprecated


basslagter:
Since Angular 1.4.1 there is a new (more consistent) method $q.resolve that does the exact same thing. I think it might be a good idea to deprecate the $q.when method and remove it in a future release.

Narretz:
We are not really removing features in the 1.x branch, so deprecation is a bit meaningless, but we can update the docs to recommend using $q.resolve instead of $q.when.

tonyfinn:
The docs currently state: 

resolve(value, [successCallback], [errorCallback], [progressCallback]);

Alias of when to maintain naming consistency with ES6.


This seems to be a reversal of what's currently recommended? https://docs.angularjs.org/api/ng/service/$q

aciccarello:
The `.resolve` alias doesn't really change the size of angular in any significant way so removing `.when` wouldn't change much. I do agree that the docs should recommend `resolve`.

gkalpak:
If someone wants to change all instances of `$q.when()` in the docs with `$q.resolve()` and submit a PR, I'd be happy to take a look.

PR #15442, opened 2016-11-27
chore($q): change when to resolve

wesleycho:
- Change `$q.when` usage to `$q.resolve`

Note that this doesn't address #13709 fully, but it modifies all meaningful areas where `$q.when` is being used.


gkalpak:
LGTM

PR #15445, opened 2016-11-28
Remove angular.uppercase/lowercase

Narretz:
fix/refactor

we expose lowercase/uppercase on the public angular object

yes

**Other information**:

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -2072,13 +2072,13 @@ function MockXhr() {
     var header = this.$$respHeaders[name];
     if (header) return header;
 
-    name = angular.lowercase(name);
+    name = angular.$$lowercase(name);
     header = this.$$respHeaders[name];
     if (header) return header;
 
     header = undefined;
     angular.forEach(this.$$respHeaders, function(headerVal, headerName) {
-      if (!header && angular.lowercase(headerName) === name) header = headerVal;
+      if (!header && angular.$$lowercase(headerName) === name) header = headerVal;
     });
     return header;
   };
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -173,15 +173,7 @@ function isValidObjectMaxDepth(maxDepth) {
 }
 
 /**
- * @ngdoc function
- * @name angular.lowercase
- * @module ng
- * @kind function
- *
- * @deprecated
- * sinceVersion="1.5.0"
- * removeVersion="1.7.0"
- * Use [String.prototype.toLowerCase](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase) instead.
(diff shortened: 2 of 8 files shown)


src/ngMock/angular-mocks.js
```diff
@@ -2072,13 +2072,13 @@ function MockXhr() {
     var header = this.$$respHeaders[name];
     if (header) return header;
 
-    name = angular.lowercase(name);
+    name = angular.$$lowercase(name);
```
mgol: Why keep the service, even as a private one? `String.prototype` methods should be enough.

mgol (dismissed):

src/ngMock/angular-mocks.js
Narretz: I just noticed that angular.lowercase / uppercase is overwritten with a manual version that handles Turkish `i`: https://github.com/angular/angular.js/blob/19bc52127f8393ffccac563b21267adb75c19275/src/Angular.js#L210-L232

So in core, we use a "safe" version. But we don't want anyone else to use it? I guess we don't want anyone to use the special version when usually the prototype functions are enough.

mgol: jQuery doesn't have any logic like that in its lowercasing and we haven't heard any complaints related to Turkish keyboards. Given jQuery's popularity I'd assume this code is no longer needed, especially that ES5 mandates no such differences so I doubt any relatively recent browser would still have whatever issue existed (if it ever has).

petebacondarwin:
LGTM - keeping the private methods for the time being.

Narretz:
K. FWIW, the "problem" still exists in modern browsers: http://plnkr.co/edit/cQHWgcqTLfZZ2LZXQQqG?p=preview 
I can't really test toLocaleLowerCase correctly because the browser must be turkish ...

mgol:
@Narretz I don't see a problem in your fiddle. `toLowerCase` should always work in the same way so the Turkish lower "i" won't be the same as lowercased regular uppercase "I"; that seems expected.

Narretz:
Hm. So if we use regular lowercase, then theoretically there could be a case where we lowercase some developer / user provided value, which is then sent to a server and arrives "incorrectly" lower cased when compared to the expected lower case string on the server. I believe we do this to http headers, for example.

mgol:
The `manualLowercase` function won't help in such a case either as it only lowercases Latin letters (from `A` to `Z`). This is, BTW, different from the native `String#toLowerCase`:
http://plnkr.co/edit/dZr9bnG8nIboMjwpgMpW?p=preview

So if there is any browser that violates ES5 and with Turkish locale has `'i' !== 'I'.toLowerCase()` then we'll use a lowercasing function that doesn't convert lots of letters that we convert otherwise. Wouldn't it cause issues?

PR #15451, opened 2016-11-30

Karthikvenkat86:
Docs Update

Improving the error doc with more appropriate error message.
Improving the error message for ng:areq with more details and examples.

Added more examples

No.

**Other information**:

Improving the error message for ng:areq with more details and examples.

### DIFF
--- a/docs/content/error/ng/areq.ngdoc
+++ b/docs/content/error/ng/areq.ngdoc
@@ -9,3 +9,32 @@ assertion expects is defined and matches the type mentioned in the error.
 
 If the type is `undefined`, make sure any newly added controllers/directives/services are properly
 defined and included in the script(s) loaded by your page.
+
+You can get this error if you accidentally miss to include your AngularJS Controller / Directive / Services file inside your ng-app HTML page.
+
+```
+<html ng-app>
+  <head>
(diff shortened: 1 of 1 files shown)


docs/content/error/ng/areq.ngdoc
```diff
@@ -9,3 +9,32 @@ assertion expects is defined and matches the type mentioned in the error.
 
 If the type is `undefined`, make sure any newly added controllers/directives/services are properly
 defined and included in the script(s) loaded by your page.
+
+You can get this error if you accidentally miss to include your AngularJS Controller / Directive / Services file inside your ng-app HTML page.
```
Karthikvenkat86: Yeah Its exactly same as above text.

But, I was going through some of the accessibility programs and realizes that examples and more contextual meanings reach more than plain text.

 If not needed we can wipe out the new line as you said.  But, I feel examples will improve the meaning of the page.

Narretz:
I think a single example is not very useful, as this error can occur in different contexts. If you have only one example, people might be mislead into thinking this is the problem, while it could have another origin.

gkalpak:
I agree. Closing, since adding such a specific example for such a generic message will be confusing in a lot of cases.

Issue #15454, opened 2016-11-30
ng-options provided empty option


johan-miller:
***Note*: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using ng-if on an empty default option a javascript error is thrown.
`TypeError: element.removeAttribute is not a function
    at removeAttr (https://code.angularjs.org/1.5.9/angular.js:3343:13)
    at JQLite.(anonymous function) [as removeAttr] (https://code.angularjs.org/1.5.9/angular.js:3479:9)
    at removeEmptyOption (https://code.angularjs.org/1.5.9/angular.js:29457:23)
    at SelectController.writeNgOptionsValue [as writeValue] (https://code.angularjs.org/1.5.9/angular.js:29491:15)
    at NgModelController.ngModelCtrl.$render (https://code.angularjs.org/1.5.9/angular.js:31826:20)
    at ngModelWatch (https://code.angularjs.org/1.5.9/angular.js:28422:14)
    at Scope.$digest (https://code.angularjs.org/1.5.9/angular.js:17747:34)
    at Scope.$apply (https://code.angularjs.org/1.5.9/angular.js:18021:24)
    at bootstrapApply (https://code.angularjs.org/1.5.9/angular.js:1806:15)
    at Object.invoke (https://code.angularjs.org/1.5.9/angular.js:4762:19)`

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**
https://plnkr.co/edit/29iYtxHrrp5beeCmwsuW

**What is the expected behavior?**
No error is thrown.

**What is the motivation / use case for changing the behavior?**
It has worked in previous versions.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**
Same code works fine in 1.5.8. Seems to be broken since version 1.5.9. Also in the latest snapshot.

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
I have seen suggestions to use ng-show to achieve the same functionality. This does fix the javascript error but also introduces another problem with IE. IE still displays the empty option even when when ng-show="false".

[...]

PR #15456, opened 2016-11-30
fix(ngOptions): ignore comment nodes when removing 'selected' attribute

Narretz:
fix (regression)

When an empty option has ngIf on it in an ngOptions select, and the ngIf expression is false, then
an error is thrown, because we try to remove the "selected" attribute from the ngIf comment node.

No

**Other information**:
~The logic is different in master, and the bug should not be happening there.~ The same bug exists on master


petebacondarwin:
This should fix #15454

petebacondarwin:
LGTM

petebacondarwin:
@Narretz - Please do land and forward port to 1.6

PR #15458, opened 2016-12-01
test($compile): work around Chrome issue with reported size for `<foreignObject>`

### LINKED ISSUE #15333: Two svg compile tests fail in Chrome 54

gkalpak:
Fixes two (potential) test failures on Chrome 53-57+.

[These tests][1] can fail on Chrome, because the **reported** size of the inner divs is subject to global display settings (e.g. font size) and browser settings (e.g. default zoom level).
See also #15333.

The assertions are now more robust by comparing against the size of the equivalent, hand-written SVG instead of fixed widths/heights.

No.

**Other information**:
Fixes #15333

[1]: https://github.com/angular/angular.js/blob/f1db7d735b475a7954023cc63f2b3f0ef685ea7e/test/ng/compileSpec.js#L445-L469

### DIFF
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -442,30 +442,70 @@ describe('$compile', function() {
     }));
 
     if (supportsForeignObject()) {
+      // Supports: Chrome 53-57+
+      // Since Chrome 53-57+, the reported size of `<foreignObject>` elements and their descendants
+      // is affected by global display settings (e.g. font size) and browser settings (e.g. default
+      // zoom level). In order to avoid false negatives, we compare against the size of the
+      // equivalent, hand-written SVG instead of fixed widths/heights.
+      var HAND_WRITTEN_SVG =
+        '<svg width="400" height="400">' +
+          '<foreignObject width="100" height="100">' +
+            '<div style="position:absolute;width:20px;height:20px">test</div>' +
+          '</foreignObject>' +
+        '</svg>';
+
       it('should handle foreignObject', inject(function() {
-        element = jqLite('<div><svg-container>' +
-            '<foreignObject width="100" height="100"><div class="test" style="position:absolute;width:20px;height:20px">test</div></foreignObject>' +
-            '</svg-container></div>');
+        element = jqLite(
+          '<div>' +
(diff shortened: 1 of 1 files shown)


test/ng/compileSpec.js
```diff
@@ -442,30 +442,70 @@ describe('$compile', function() {
...
-        var bounds = testElem[0].getBoundingClientRect();
-        expect(bounds.width === 20 && bounds.height === 20).toBe(true);
+        var referenceElem = element.find('div')[0];
+        var testElem = element.find('div')[1];
+        var referenceBounds = referenceElem.getBoundingClientRect();
+        var testBounds = testElem.getBoundingClientRect();
+
+        expect(isHTMLElement(testElem)).toBe(true);
+        expect(referenceBounds.width).toBeGreaterThan(0);
+        expect(referenceBounds.height).toBeGreaterThan(0);
```
petebacondarwin: Is there any reason to believe that this might not be greater than zero?
And even if it was zero, do we care? If that is what the browser does with the handwritten HTML then so be it?
Right?

petebacondarwin:
A minor comment but LGTM, with or without changes.

test/ng/compileSpec.js
gkalpak: It is just a sanity check that we are doing the right thing and the hand-written HTML produces something. If this changes in the future (e.g. the hand-written HTML does not produce the expected result) **and** we break `$compile`, the tests might pass although the code will be broken.
It is a somewhat far fetched scenario, so I can remove it if you think it is more confusing than helpful.

Narretz:
These tests also fail on Edge 14, but this patch fixes it (tested locally)

petebacondarwin:
Doesn't this need to be backported @gkalpak?

gkalpak:
Done! Thx :)
(I tried to backport to `v1.5.x` and didn't notice it failed.)

PR #15459, opened 2016-12-01
fix(ngOptions): ignore comment nodes when removing 'selected' attribute

Narretz:
fix (regression)

When an empty option has ngIf on it in an ngOptions select, and the ngIf expression is false, then
an error is thrown, because we try to remove the "selected" attribute from the ngIf comment node.

No

**Other information**:

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -420,6 +420,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
       // option when the viewValue does not match any of the option values.
       for (var i = 0, children = selectElement.children(), ii = children.length; i < ii; i++) {
         if (children[i].value === '') {
+          selectCtrl.hasEmptyOption = true;
           selectCtrl.emptyOption = children.eq(i);
           break;
         }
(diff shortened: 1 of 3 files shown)


src/ng/directive/select.js
```diff
@@ -70,7 +70,9 @@ var SelectController =
   };
 
   self.unselectEmptyOption = function() {
-    if (self.emptyOption) {
+    // Empty options may be comment nodes if they have been removed by
+    // transclude directives such as `ngIf`
+    if (self.emptyOption && self.emptyOption[0].nodeType === NODE_TYPE_ELEMENT) {
```
gkalpak: Isn't there always going to be a comment there, even when `ngIf` is true?
Shouldn't we look for the first non-comment element in `self.emptyOption`?

Narretz: That's right. That means the whole adding / removing "selected" attribute is currently unsound when ngIf / a transclusion directive is used. 
and it's not as easy as looking for the first non-comment element, because self.emptyOption is sometimes *only* the comment

Narretz:
I've updated the PR with an actual solution. Basically, we now require that emptyOption is always an actual option element. If the emptyOption is a comment after compilation, we use a modified version of the registerOption function on the select controller to add the empty option.

src/ng/directive/ngOptions.js
```diff
@@ -556,9 +565,35 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
...
-        selectCtrl.emptyOption.removeClass('ng-scope');
+        if (selectCtrl.emptyOption[0].nodeType === NODE_TYPE_COMMENT) {
+          // This means the empty option has currently no actual DOM node, probably because
+          // it has been modified by a transclusion directive.
+          selectCtrl.hasEmptyOption = false;
+
+          // Redefine the registerOption function, which will catch
+          // options that are added by ngIf etc. (rendering of the node is async because of
+          // lazy transclusion)
+          selectCtrl.registerOption = function(optionScope, optionEl) {
```
petebacondarwin: Is it worth just adding this behaviour to the normal `registerOption` rather than monkeypatching?

Narretz: Theoretically it already does, because without ngOptions, the option directive adds the empty option (via selectCtrl.registerOption -> selectCtrl.addOption). However, it does more work that we not actually need.
We also noop .registerOption in the ngOptions directive, so that we don't accidentially add options to the selectCtrl. So this fn is only relevant if there's a dynamic empty option.

Maybe we can refactor this later. It feels like there's always room to refactor select and ngOptions ...

PR #15462, opened 2016-12-01
fix(copy): angular copy should throw exception when source and destin…

### LINKED ISSUE #15444: angular copy should throw exception when source and destination are not the same type.

bug

In the following cases

1 - angular copy a source (Array Type) and destination (Object Type), then the app crashes
TypeError: destination.push is not a function
    at copyRecurse (angular.js:911).......

2 -  angular copy a source (Object Type) and destination ( Array Type) then source will equal empty array []

3 -  angular copy a source (String Type) and destination ( Object Type) then 

    var result = angular.copy('abc', {}); // output {"0":"a","1":"b","2":"c"}

Plunker for the case no 1
https://plnkr.co/edit/BwmL7DHUF9xxo4G58orM?p=preview

Plunker for case no 2,3
https://plnkr.co/edit/vmn7wv8sI33PvSxVtM1q?p=preview 

I suggest if the source and the destination are not the same type it should throw an exception
      
Prevent the app for crashing for the first case, and make sure that copy method is working as expected for all the cases. 
since that Javascript is not static typed language we can't make sure that the developer will always enter a source and destination of the same type

1.5.8

In case of destination in not undefined check if source and destination are not the same type then throw exception
    
     function copy(source, destination) {
  
      var stackSource = [];
      var stackDest = [];
  
      if (destination) {
        // Add this case
        if(typeof source !== typeof destination){
           throw ngMinErr('cpi', 'Can\'t copy! Source({0} type) and destination({1} type) are not the same type', typeof source, typeof destination);
      }
    
    if (isTypedArray(destination) || isArrayBuffer(destination)) {
      throw ngMinErr('cpta', 'Can\'t copy! TypedArray destination cannot be mutated.');
    }
    if (source === destination) {
      throw ngMinErr('cpi', 'Can\'t copy! Source and destination are identical.');
    }

m-amr:

bug

https://github.com/angular/angular.js/issues/15444

when angular copy source and destination are not the same type, it throw an exception

**Other information**:

…ation are not the same type

Closes #15444

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -820,6 +820,10 @@ function copy(source, destination) {
   var stackDest = [];
 
   if (destination) {
+    if (toString.call(destination) !== toString.call(source)) {
+      throw ngMinErr('cpi', 'Can\'t copy! Source({0} type) and destination({1} type) aren\'t the same type.',
+          toString.call(source), toString.call(destination));
+    }
     if (isTypedArray(destination) || isArrayBuffer(destination)) {
       throw ngMinErr('cpta', 'Can\'t copy! TypedArray destination cannot be mutated.');
     }


m-amr:
This will require removing some test cases like

     it('should clear destination arrays correctly when source is non-array', function() {
      expect(copy(null, [1,2,3])).toEqual([]);
      expect(copy(undefined, [1,2,3])).toEqual([]);
      expect(copy({0: 1, 1: 2}, [1,2,3])).toEqual([1,2]);
      expect(copy(new Date(), [1,2,3])).toEqual([]);
      expect(copy(/a/, [1,2,3])).toEqual([]);
      expect(copy(true, [1,2,3])).toEqual([]);
    });

    it('should clear destination objects correctly when source is non-array', function() {
      expect(copy(null, {0:1,1:2,2:3})).toEqual({});
      expect(copy(undefined, {0:1,1:2,2:3})).toEqual({});
      expect(copy(new Date(), {0:1,1:2,2:3})).toEqual({});
      expect(copy(/a/, {0:1,1:2,2:3})).toEqual({});
      expect(copy(true, {0:1,1:2,2:3})).toEqual({});
    });

since that these cases will never happened because the source and destination must be the same type. 
Is there is a reason to keep these cases ? the are useless now

gkalpak:
I am putting this in the Ice Box, as I don't think this change is appropriate (see https://github.com/angular/angular.js/issues/15444#issuecomment-263214601).
I could get cnvinced otherwise thougg (if there is a compelling usecase).

Issue #14015, opened 2016-02-12
ng-hide and ng-show showing at the same time for a short period of time in IE11


PhilipWallin:
When including ng-animate there seem to be a brief period of time where a ng-show and ng-hide element are showing at the same time, even if no animations are used.

Example (the problem is only visible if using IE11):
http://plnkr.co/edit/vqVGcf9cjFQPAbq0Igyq?p=preview

In the example you can see the flickering of the two elements showing at the same time when checking and unchecking the checkbox which trigger ng-show/ng-hide.

Narretz:
This sounds familiar. The problem is that I can never reliably reproduce these in IE11. It has happened maybe once in 30 times for me. I've also seen it once on Firefox, so it probably has something to do with how fast the browser flushed the requestAnimationFrame queue.

Narretz:
Can you please test with the latest snapshot: //code.angularjs.org/snapshot/angular-animate.js

PhilipWallin:
In my IE11 I get the flash of both showing every time I click the checkbox. Using the linked angular-animate didn't seem to change this behavior.

Narretz:
That's weird. What IE and OS version do you have? And is your computer generally slow or under load?

PhilipWallin:
We have quite high-end computers (I tried on three of them), my initial guess was actually that it was fast enough to show something which is only visible for a short amount of time. I'm running Windows 8.1 with IE version 11.0.9600.18202. One of the computers I tried it on was Windows 7, showing the same problem.

PhilipWallin:
I'm currently applying this css to avoid the problem:
.ng-hide.ng-hide-animate {
    display: none !important;
}

This could of course cause problems for us in the future if we want to animate some ng-hide and not animate some.

Narretz:
Hm, I didn't think about the computer being too fast. That could it be it. I also have a slightly newer version of IE11 (11.0.9600.18204), but I doubt that this is the reason.

PhilipWallin:
I tried this now in IE11 on my virtual Windows 7, and experience the problem maybe once every 2-10 try. This could be because my virtual machine is running slower, but this is only a guess.

plestik:
[...]

PR #15463, opened 2016-12-02
fix($animateCss): avoid flicker caused by temporary classes

Narretz:
bug fix

See #14015

No

**Other information**:

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -692,6 +692,13 @@ var $AnimateCssProvider = ['$animateProvider', /** @this */ function($animatePro
 
           runner = new $$AnimateRunner(runnerHost);
 
+          if (options.tempClasses) {
+            // Since animateCss waits for an animation frame to start the actual
+            // animation, the temp class styles might be visible for a very short time.
+            // Removing and re-adding before the actual start solves this problem.
+            $$jqLite.removeClass(element, options.tempClasses);
+          }
+
           waitUntilQuiet(start);
 
           // we don't have access to pause/resume the animation
@@ -833,6 +840,11 @@ var $AnimateCssProvider = ['$animateProvider', /** @this */ function($animatePro
         if (!node.parentNode) {
           close();
           return;
+        }
+
+        if (options.tempClasses) {
+          $$jqLite.addClass(element, options.tempClasses);
+          options.tempClasses = null;
         }
 
         // even though we only pause keyframe animations here the pause flag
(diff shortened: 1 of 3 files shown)


gkalpak:
I wish we could come up with a more straightforward fix. Adding, removing, re-adding temporary classes doesn't feel good (both from a maintenance and a robustness perspective).

AFAICT, `tempClasses` is an undocumented, private feature that is used for `ngShow`/`ngHide` only. From a quick look, it doesn't seem to be necessary to apply the classes in the `beforeStart()` function. What would happen if we only applied `tempClasses` once the animation actually started?

Narretz:
tempClasses is undocumented, but supposed to be documented, see https://github.com/angular/angular.js/issues/9549
If tempClasses is removed, a bunch of tests fail that check if the tempClasses are on the element :D
There's definitely a specific reason why ngShow uses tempClasses - it must be in the tests or code somewhere. But maybe it's obsolete, I don't know.
Still, tempClasses should be able to trigger an animation same as any other class, so if we keep them we'd need to apply them before animation detection is done.

gkalpak:
> `tempClasses` is undocumented, but supposed to be documented

The important thing is that it is not documented right now :grin:
So, we still have the chance to fix any issues and document the new, correct behavior.

--
> There's definitely a specific reason why `ngShow` uses `tempClasses` - it must be in the tests or code somewhere. But maybe it's obsolete, I don't know.

The reason why `ngShow`/`ngHide` use `tempClasses` is that we don't want to apply the `display: none` style on `.ng-hide` while animating, but we can't use the "standard" `ng-animate` class either (see 39d0b36). So we need the temp `ng-hide-animate` class in order to be able to write `.ng-hide:not(.ng-hide-animate) { ... }`

To be clear, I am not proposing to ditch `tempClasses` altogether. I am just saying that we should postpone adding them until they are necessary (instead of adding/removing them repeatedly during an animation). I.e. only add them [here][1].

--
> Still, `tempClasses` should be able to trigger an animation same as any other class, so if we keep them we'd need to apply them before animation detection is done.

I am not sure I agree. (OK, actually I am sure I disagree 😛)
I don't see why they should be able to trigger an animation. As far as I understand, the purpose of `tempClasses` is to offer more expressiveness/flexibility in CSS selectors during the actual animation. There is no indication that they are intended to be able to trigger an animation (and I can't think of a good reason why they should).

BTW, the only relevant mention I could find is the [`ngShow` docs][2] (but not in `ngHide`! :smiley: - I'll fix that):

> Note that the selector that needs to be used is actually `.ng-hide:not(.ng-hide-animate)` to cope with extra animation classes that can be added.

[1]: https://github.com/angular/angular.js/pull/15463/files#diff-7ef231a7873228717749111b56b1eb84R846
[2]: https://docs.angularjs.org/api/ng/directive/ngShow

gkalpak:
A PR is worth a thousand words :smiley:
Here is what I had in mind: #15472

Narretz:
This PR doesn't work unfortunately in all cases.

PR #15464, opened 2016-12-02
feat($compile): support expansion of special `ngBindon` attributes

### LINKED ISSUE #15455: Feature: Support unidirectional dataflow syntactic sugar
Similar to Angular 2's "banana in a box" syntactic sugar to support adding a one-way binding and an expression to set that one-way binding all in one shot, how would you feel about supporting something like:

`<my-component ng-banana-foo="bar>` which would expand internally to `<my-component foo="bar" on-foo-changed="bar=$event">`?

While the proposed syntax isn't quite as nice as the Angular2 one, I thought it might be important to not use the same syntax in order to avoid confusion between the two frameworks.  With the push towards components and unidirectional dataflow since 1.5, this would be a really nice enhancement to reduce some of the verbosity associated with actually following those ideas.

dcherman:

feat

Supporting unidirectional data-flow is rather verbose

Supporting unidirectional data-flow is less verbose

Nope

In order to reduce some of the verbosity associated with conforming to the
ideas behind unidirectional data-flow, we can introduce a special
attribute syntax that will be automatically expanded in order to simulate
two-way data binding.

For example,

`<my-component ng-bindon-foo="bar">` will be expanded into
`<my-component foo="bar" foo-changed="bar = $event">`

Fixes #15455

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1803,7 +1803,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             : function denormalizeTemplate(template) {
               return template.replace(/\{\{/g, startSymbol).replace(/}}/g, endSymbol);
         },
-        NG_ATTR_BINDING = /^ngAttr[A-Z]/;
+        NG_SPECIAL_ATTR = /^ng(Attr|Bindon)[A-Z]/;
     var MULTI_ELEMENT_DIR_RE = /^(.+)Start$/;
 
     compile.$$addBindingInfo = debugInfoEnabled ? function $$addBindingInfo($element, binding) {
@@ -2139,8 +2139,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
               directiveNormalize(nodeName), 'E', maxPriority, ignoreDirective);
 
           // iterate over the attributes
-          for (var attr, name, nName, ngAttrName, value, isNgAttr, nAttrs = node.attributes,
-                   j = 0, jj = nAttrs && nAttrs.length; j < jj; j++) {
+          for (var attr, name, nName, ngAttrName, value, isNgAttr, isSpecialAttr, isNgBindon,
+            nAttrs = node.attributes, j = 0, jj = nAttrs && nAttrs.length; j < jj; j++) {
             var attrStartName = false;
             var attrEndName = false;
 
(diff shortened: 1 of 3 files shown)


dcherman:
This PR is intended mainly to serve as a proof of concept and start a discussion around whether or not we like this implementation and also, the actual name of the attribute prefix we'd like to use.

Will work on documentation and additional tests shortly assuming we like it.

src/ng/compile.js
```diff
@@ -1803,7 +1803,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             : function denormalizeTemplate(template) {
               return template.replace(/\{\{/g, startSymbol).replace(/}}/g, endSymbol);
         },
-        NG_ATTR_BINDING = /^ngAttr[A-Z]/;
+        NG_ATTR_BINDING = /^ng(Attr|Autobind)[A-Z]/;
```
gkalpak: The name of the variable needs to change.

test/ng/compileSpec.js
```diff
@@ -11441,6 +11441,40 @@ describe('$compile', function() {
     });
   });
 
+  describe('ngAutobind attributes', function() {
+    it('should expand `ng-autobind-foo` to a one-way binding and an output expression', function() {
```
gkalpak: The `one-way` part is not accurate.

test/ng/compileSpec.js
```diff
@@ -11441,6 +11441,40 @@ describe('$compile', function() {
...
+      var testController;
+
+      module(function($compileProvider) {
+        $compileProvider.component('test', {
+          bindings: {
+            foo: '<',
+            fooChange: '&'
+          },
+          controller: function($attrs) {
+            testController = this;
```
gkalpak: We usually use `element.controller('test')` for accessing the controller. It is a good idea to keep the consistency.

gkalpak (commented):
I like `ng-bindon-` better, because it matches ng2's `bindon-` pretty closely (but maybe it is too close).

dcherman:
Ah, I forgot that ng2 even had `bindon` since I was used to the shorthand syntax.  That explains the suggestion.

Anyway, everything should be fixed up.

dcherman:
Any update here?  Was hopeful that this would make it to 1.6, but would love it if it could be considered for the next release (and maybe a patch to 1.5?  :heart: )

gkalpak:
This can still land on 1.6. Since it is not a breaking change, it didn't have to make it into 1.6.0.
Let's discuss this next week.

dcherman:
Any update here?

gkalpak:
Let's discuss this next week. <sub>(This weeks meeting was too busy :grin:)</sub>

dcherman:
For anyone that's been waiting for this, I've written a polyfill that can be used to detect and add support for this feature.  It does depend on implementation details and some assumptions since there's no other way to do it, and it does have the cost of extra iterations of the attributes object.

https://gist.github.com/dcherman/57621dd904cf598f12b81ae6e0f4b6b2

PR #15465, opened 2016-12-02
docs(guide/Templates): make directive name camelCase for consistency

ojab:
ng-controller in the comment above is written in camelcase, so make ng-click also camelcase.


ojab:
signed CLA

PR #15468, opened 2016-12-03
fix(select): do not break when removing the element (e.g. via `ngIf`)

### LINKED ISSUE #15466: 1.6.0-rc.2: ng-if on select component with an ng-model produces "Cannot read property '$$phase' of null" error

Report a bug.

Current behavior is that using ng-if on select input with an ng model produces the following console error:
```
    angular.js:14290TypeError: Cannot read property '$$phase' of null
    at Object.$$debounceViewValueCommit (angular.js:28509)
    at Object.$setViewValue (angular.js:28490)
    at Array.<anonymous> (angular.js:31800)
    at Scope.$digest (angular.js:17815)
    at Scope.$apply (angular.js:18030)
    at HTMLButtonElement.<anonymous> (angular.js:26642)
    at defaultHandlerWrapper (angular.js:3579)
    at HTMLButtonElement.eventHandler (angular.js:3567)
```

**Repro**
View the following plunkers with the dev console open and click the "Toggle ng-if on select" button:

1.6 rc 2 behavior (broken): https://plnkr.co/edit/EhBpLWbKrHTadmSsRjKZ?p=preview
Produces error noted above.

1.5.9 behavior (no error): https://plnkr.co/edit/PH4sJl65DlseChiMEcUy?p=preview
No error.

I would expect ng-if on a select input to not produce a console error.

Appears to be a bug unless I haven't processed [#13286](https://github.com/angular/angular.js/pull/13286) or the various select input changes properly.

This seems to be an issue in all browsers, is present in the latest 1.6x version, and is not present in the latest 1.5.x version.

It looks like this is likely related to the change from
```$rootScope.$$phase``` to ```this.$$scope.$root.$$phase``` [located here](https://github.com/jbedard/angular.js/blob/2bc598096260f6593ea31d13a3d409f2e20bbb4c/src/ng/directive/ngModel.js#L824) in the [#13286](https://github.com/angular/angular.js/pull/13286) PR.

### PR DESCRIPTION
**Pull request by gkalpak**
Bug fix.

When removing `<option>` elements, we schedule some operations on the `ngModelCtrl`. For performance reasons, these operation are scheduled to be run as a post-digest operation. If the `<select>` element happens to be removed as well (and its scope destroyed), an error may be thrown due to certain values not being available on the `$scope`.
See #15466.

Nothing breaks.

No.

**Other information**:
Fixes #15466.

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -76,11 +76,9 @@ var SelectController =
     }
   };
 
-  var scopeDestroyed = false;
   $scope.$on('$destroy', function() {
     // disable unknown option so that we don't do work when the whole select is being destroyed
     self.renderUnknownOption = noop;
-    scopeDestroyed = true;
   });
 
   // Read the value of the select control, the implementation of this changes depending
@@ -184,7 +182,7 @@ var SelectController =
     updateScheduled = true;
 
     $scope.$$postDigest(function() {
-      if (scopeDestroyed) return;
+      if ($scope.$$destroyed) return;
 
       updateScheduled = false;
       self.ngModelCtrl.$setViewValue(self.readValue());
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -76,9 +76,11 @@ var SelectController =
     }
   };
 
+  var scopeDestroyed = false;
   $scope.$on('$destroy', function() {
     // disable unknown option so that we don't do work when the whole select is being destroyed
     self.renderUnknownOption = noop;
+    scopeDestroyed = true;
   });
 
   // Read the value of the select control, the implementation of this changes depending
@@ -182,6 +184,8 @@ var SelectController =
     updateScheduled = true;
 
     $scope.$$postDigest(function() {
+      if (scopeDestroyed) return;
+
       updateScheduled = false;
       self.ngModelCtrl.$setViewValue(self.readValue());
       if (renderAfter) self.ngModelCtrl.$render();
--- a/test/ng/directive/selectSpec.js
+++ b/test/ng/directive/selectSpec.js
@@ -7,7 +7,7 @@ describe('select', function() {
     formElement = jqLite('<form name="form">' + html + '</form>');
     element = formElement.find('select');
     $compile(formElement)(scope);
-    scope.$apply();
+    scope.$digest();
   }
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by Narretz** [comment]
Does the app actually break when the select is removed, or does it just throw? I mean, is there some cleanup that is not performed? I think the commit message should reflect that more precisely.

File: src/ng/directive/select.js [hunk 1]
```diff
@@ -76,9 +76,11 @@ var SelectController =
     }
   };
 
+  var scopeDestroyed = false;
```
gkalpak: Yes, I could. But since we were already having a `$destroy` callback I opted for using that instead of the `$$destroy` as it seems unnecssary. Can go either way.

gkalpak:
@Narretz, not sure what you mean. An error is thrown in the middle of a digest. Whether that breaks an app depends on what was going on during that digest, what watchers were skipped due to the error etc.

gkalpak:
BTW, the particular error in #15466 is caused by [this line][1]: When a previously selected option element is destroyed, we schedule to re-read the select's value asynchronously. Since destroying the select itself also destroys its option children, it triggers this check (which then fails because the scope has been destroyed already.
But similar problems would arise whenever we call `scheduleViewValueUpdate()` and the select is destroyed before the `$$postDigest` phase, so guarding against it inside `scheduleViewValueUpdate`'s `$postDigest` callback seems appropriate.

[1]: https://github.com/gkalpak/angular.js/commit/47c15fbcc10f118170813021e8e605ffd263ad84#diff-0c5bee20e5ea71c419c05616636e4959R244

Narretz:
@gkalpak In the middle of a digest? It throws in $$postDigest, no?

gkalpak:
Yeah, sorry, I meant `$$postDigest`. So, whether the app breaks, depends on what tasks are still pending in the `postDigestQueue` :grin:

gkalpak:
Aha, I see your point. I just realized it only breaks if the `$exceptionHandler` rethrows. I will change the wording to "not throw".

src/ng/directive/select.js
petebacondarwin: If `$$destroyed` would work then I would prefer that. It is one less thing to store in memory and keep track of in maintenance of the code.

petebacondarwin:
LGTM

PR #15472, opened 2016-12-07
fix($animateCss): avoid flicker caused by temporary classes

### LINKED ISSUE #14015: ng-hide and ng-show showing at the same time for a short period of time in IE11

gkalpak:
Bug fix.

Adding the temporary classes immediately, while forcing a reflow before starting the animation, could cause flickering. Currently, only `ngHide`/`ngShow` animations are affected (but theoretically any future animation relying on `tempClasses` would be).
See #14015.

The temporary classes are not applied until the animation actually starts, which avoids the flickering.

No. (Since `tempClasses` was an undocumented, private feature (currently only used for `ngHide`/`ngShow`), no other built-in or custom animations will be affected).

**Other information**:
This is an alternative approach to #15463.
Fixes #14015.

### DIFF
--- a/src/ngAnimate/animateCss.js
+++ b/src/ngAnimate/animateCss.js
@@ -835,6 +835,11 @@ var $AnimateCssProvider = ['$animateProvider', /** @this */ function($animatePro
           return;
         }
 
+        if (options.tempClasses) {
+          $$jqLite.addClass(element, options.tempClasses);
+          options.tempClasses = null;
+        }
+
         // even though we only pause keyframe animations here the pause flag
         // will still happen when transitions are used. Only the transition will
         // not be paused since that is not possible. If the animation ends when
--- a/src/ngAnimate/animateJs.js
+++ b/src/ngAnimate/animateJs.js
@@ -90,6 +90,11 @@ var $$AnimateJsProvider = ['$animateProvider', /** @this */ function($animatePro
             return runner;
           }
 
+          if (options.tempClasses) {
+            $$jqLite.addClass(element, options.tempClasses);
+            options.tempClasses = null;
+          }
+
           runner = new $$AnimateRunner();
           var closeActiveAnimations;
           var chain = [];
--- a/src/ngAnimate/animation.js
+++ b/src/ngAnimate/animation.js
@@ -134,7 +134,6 @@ var $$AnimationProvider = ['$animateProvider', /** @this */ function($animatePro
       var tempClasses = options.tempClasses;
       if (tempClasses) {
         classes += ' ' + tempClasses;
-        options.tempClasses = null;
       }
 
       var prepareClassName;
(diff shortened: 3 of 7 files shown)


Narretz:
I think that's good.
Ideally, not every drive would have to add the temp classes, but since there are no custom drivers (to my knowledge), this is irrelevant.

Just to clarify, before this change was it possible to use a temporary class to trigger a css animation? I.e. a temporary class applies a transition that makes the animator detect an animation.

gkalpak:
> Just to clarify, before this change was it possible to use a temporary class to trigger a css animation? I.e. a temporary class applies a transition that makes the animator detect an animation.

It was and it still is. This PR only affects adding the classes in `beforeStart()`, at which point the animations are determined. The `tempClasses` are used [elsewhere][1], so this PR does not affect that:
[Animaton works with this PR][2])

[1]: https://github.com/angular/angular.js/pull/15472/files#diff-4a45cb4e0ccbe72be86d77bb8d713c93R136
[2]: https://codepen.io/ExpertSystem/pen/mRyddm?editors=0101

Narretz:
Actually, the current ngHide example from the docs doesn't work correctly for the "show" (ng-hide-remove) action with this patch, see https://plnkr.co/edit/tyteqvfPbhcwrBKXpn5c?p=preview

When you comment in the patched angular-animate (copied from you pen), the show action doesn't animate anymore.

gkalpak:
😞 It's always possible that I've missed something. If #15463 works, I'm happy to go with that.
<sub>I've had too much `ngAnimate` lately to investigate further 😩 </sub>

Narretz:
My PR actually has the same problem ... the thing is that the plnkr example I posted puts the actual transition css on the temporary class, which is not something we advocate. I'm not sure where I got this plnkr / css from, maybe an issue or something?
So if we are okay that this way of defining animations is wrong, then we could use this PR.

gkalpak:
The question is: Can we fix the issue without breaking the usecase?

Narretz:
I don't think so. The rAF call that causes this is elementary to the way ngAnimate works. Removing it or working around it would probably mean a big rewrite.

gkalpak:
In that case, I vote for fixing it with a breaking change (i.e. that animation styles on temporary classes won't work) for 1.7.0.

Narretz:
Bad news: I've re-tested the plnkr with our patches, and even with the regular ngShow / ngHide example from the docs, they don't work for showing: https://plnkr.co/edit/tyteqvfPbhcwrBKXpn5c?p=preview (the current ngAnimate file is active, but use the patches and it doesn't work).
I thought this didn't work because the transition was set on the temporary class, but this is not the case.

Narretz:
Here's the original repro to test again: https://plnkr.co/edit/l70kaJ?p=preview (flickers in IE11)

gkalpak:
Yeah, I realized our both our fixes are problematic. I started looking into the issue again 😞

gkalpak:
I give up 😱 Unfortunately, I couldn't get to the bottom of this in a reasonable amount of time 😞
It only happens on IE for me (which is not great for debugging :grin:) and there are work-arounds available:
1. If you don't mind destroying and recreating the view, you can use `ng-switch`.
2. If you want to keep the view around (but hidden), you can use `ng-class="{'ng-hide': checked}"` 

In any case, I am going to close this PR, because it breaks a valid usecase, so is not an appropriate fix anyway.

PR #15473, opened 2016-12-07
Clarify where ngDisabled should be used

samuraijane:
Update documentation.

n/a

n/a

No

**Other information**:

My understanding is that ngDisabled is for form elements only so I propose that the documentation reflect this more clearly.

### DIFF
--- a/src/ng/directive/attrs.js
+++ b/src/ng/directive/attrs.js
@@ -159,7 +159,7 @@
  *
  * @description
  *
- * This directive sets the `disabled` attribute on the element if the
+ * This directive sets the `disabled` attribute on a form element if the
  * {@link guide/expression expression} inside `ngDisabled` evaluates to truthy.
  *
  * A special directive is necessary because we cannot use interpolation inside the `disabled`


src/ng/directive/attrs.js
```diff
@@ -159,7 +159,7 @@
  *
  * @description
  *
- * This directive sets the `disabled` attribute on the element if the
+ * This directive sets the `disabled` attribute on a form element if the
```
gkalpak: I think this can be more confusing than helpful. ("form element" often refers to the `<form>` element.)
How about:

> This directive sets the `disabled` attribute on the element (typically a form control) if the...

Narretz: I think there are only these element that natively respond to disabled: option, input, textarea, button, select. Why not list them all?

samuraijane:
<docs>(<ngDisabled>): <Clarify use of ngDisabled for form elements>
<BLANK LINE>
<body>
<Modify documentation to clarify that ngDisabled is only for form elements.>
<footer>

Narretz:
@samuraijane No need to close it, you can update the PR.

samuraijane:
@Narretz Thank you. Still learning how to do this.

src/ng/directive/attrs.js
gkalpak: [There are more][1] (e.g. optgroup, fieldset). Plus there is nothing stopping people from using it on elements that don't natively respond to `disabed`, such as custom controls. I wouldn't make too strong a statement (i.e. "it can be used on _these_ elements only). Mentioning the common usecase is fine though (either under the general term "form controls" or by explicit listing). 

[1]: https://www.w3.org/TR/html5/disabled-elements.html

Narretz: Yeah, I meant list all the elements that are supported *natively*, but not write that they can be used only on these elements. I'm fine with 
> (typically a form control, e.g. `input`, `button`, `select` etc.)

Narretz:
@samuraijane are you going to update this PR? If not, I'm gonna close it, and you can reopen it once you have time to update it.

PR #15475, opened 2016-12-08
fix(jqLite): silently ignore `after()` if element has no parent

### LINKED ISSUE #15331: `parent` is null in JQLite code
Sentry is reporting an error in our app in Angular's source code for JQLite:

image

The line in question is https://github.com/angular/angular.js/blob/master/src/jqLite.js#L986

It appears under some circumstances `parent` can be `null`, in which case this error pops up. It's probably a good idea to guard against unexpected values of `parent` so this error doesn't occur.

This was seen happening on Angular 1.5.8, on Firefox v26 on Windows XP.

gkalpak:
Bug fix.

The element was always assumed to have a parent and an error was thrown when that was not the case.

jqLite is now consistent with jQuery, which silently ignores a call on elements that do not have a parent.

No. (This change only affects already broken code, that would otherwise throw an error.)

**Other information**:
Fixes #15331
Closes #15367

### DIFF
--- a/test/jqLiteSpec.js
+++ b/test/jqLiteSpec.js
@@ -2187,6 +2187,7 @@ describe('jqLite', function() {
     it('should not throw when the element has no parent', function() {
       var span = jqLite('<span></span>');
       expect(function() { span.after('abc'); }).not.toThrow();
+      expect(span).toJqEqual(jqLite('<span></span>'));
     });
   });
 
--- a/test/jqLiteSpec.js
+++ b/test/jqLiteSpec.js
@@ -2187,7 +2187,8 @@ describe('jqLite', function() {
     it('should not throw when the element has no parent', function() {
       var span = jqLite('<span></span>');
       expect(function() { span.after('abc'); }).not.toThrow();
-      expect(span).toJqEqual(jqLite('<span></span>'));
+      expect(span.length).toBe(1);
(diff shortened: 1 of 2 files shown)


test/jqLiteSpec.js
```diff
@@ -2182,6 +2182,12 @@ describe('jqLite', function() {
       span.after('abc');
       expect(root.html().toLowerCase()).toEqual('<span></span>abc');
     });
+
+
+    it('should not throw when the element has no parent', function() {
+      var span = jqLite('<span></span>');
+      expect(function() { span.after('abc'); }).not.toThrow();
```
Narretz: Can you add an expect that the span is still the span?

Narretz (changes requested):

test/jqLiteSpec.js
Narretz: The test is now failing :o is it because toJqEqual is wrapped in jqlite?

gkalpak: 😱 It passed locally (on Chrome). IEs seems to have a hard time.  I'll take a look.

gkalpak: Checking DOM Nodes for deep equality isn't an intended usecase apparenty :smiley:
I changed the expectations.

Narretz (approved):

PR #15478, opened 2016-12-08
fix(ngOptions): do not unset the `selected` property unless necessary

### LINKED ISSUE #15477: multiple select always jumps back to lowest marked element
Bug

- multiple select: In successively selecting one of the last and one of the first elements, the selector jumps back to the lowest selected element.
I think angular background magic bred this bug.

- example: https://plnkr.co/edit/8Cw3duixywJui0TYELKF?p=preview
It's only a minimal example, i build this in different cases but i can't post the original source code.
**special look at the ```  $interval(function() {}, 100); ```**

* the select works normally without jumping.

 - angularjs 1.5.6
- Firefox / IE

sorry for my bad english, I'm from Germany.

with best regards
Sogmog

gkalpak:
Bug fix.

When updating the value of a `select[multiple]` element, all options are first set to `selected = false` and then the selected ones are set to `true`. By setting an already selected option to `selected = false` and then `true` again - essentially unselecting and reselecting it - causes some browsers (including Firefox, IE and under some circumstances Chrome) to unexpectedly scroll to the last selected option.

The `selected` property of the `<option>` elements is only set if its current value is different than the new one and even then it is set to its final value at once (i.e. without first setting it to `false`), thus avoiding the undesirable behavior.

No.

**Other information**:
Fixes #15477

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -505,11 +505,11 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
 
       } else {
 
-        selectCtrl.writeValue = function writeNgOptionsMultiple(value) {
+        selectCtrl.writeValue = function writeNgOptionsMultiple(values) {
           // Only set `<option>.selected` if necessary, in order to prevent some browsers from
           // scrolling to `<option>` elements that are outside the `<select>` element's viewport.
 
-          var selectedOptions = value && value.map(getAndUpdateSelectedOption) || [];
+          var selectedOptions = values && values.map(getAndUpdateSelectedOption) || [];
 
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngOptions.js
```diff
@@ -506,16 +506,16 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
       } else {
 
         selectCtrl.writeValue = function writeNgOptionsMultiple(value) {
```
gkalpak: Done

Narretz:
LGTM

Narretz:
I restarted the unit test job, there were some mobile safari tests for animateCss, they looked unrelated

PR #15483, opened 2016-12-09
refactor(*): replace `HashMap` with `NgMap`

gkalpak:
Refactoring.

Same old, same old!

`HashMap` is replaced with `NgMap`, which for the time being points to `NgMapShim`, an API-compatible `Map` implementation (for the features Angular needs). This will allow us to easily switch to native implementations once they become more stable.

No.

**Other information**:
See #13209 and #15114 for related discussions.

### DIFF
--- a/src/apis.js
+++ b/src/apis.js
@@ -1,6 +1,5 @@
 'use strict';
 
-
 /**
  * Computes a hash of an 'obj'.
  * Hash of a:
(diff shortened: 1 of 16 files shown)


gkalpak:
BTW, this is the same code as in #15114, minus #13209. The `ES6Map` part of #13209 is irrelevant now. The `perf(copy)` part of #13209 (which is still relevant) can be rebased and merged once this has been merged.

src/apis.js
```diff
@@ -33,49 +32,69 @@ function hashKey(obj, nextUidFn) {
...
-var $$HashMapProvider = [/** @this */function() {
+// For now, always use `NgMapShim`, even if `window.Map` is available. Some native implementations
+// are still buggy (often in subtle ways) and can cause hard-to-debug failures. When native `Map`
+// implementations get more stable, we can reconsider switching to `window.Map` (when available).
+var NgMap = NgMapShim;
+
+var $$MapProvider = [/** @this */function() {
   this.$get = [function() {
-    return HashMap;
+    return NgMap;
```
dcherman: Can we feature detect native `Map` implementation and just use that if present?  It should be significantly faster than this implementation since lookups are not O(n), and all modern browsers support it at this point.  This fallback is really just for like...IE9 and IE10.

dcherman:
So I know we were debating this in that other issue, but I still think an implementation closer to `$$HashMap` is more ideal with property lookups than using dual arrays for perf reasons.

It would not be a breaking change since due to how it worked, the old HashMap implementation could never support frozen objects anyway since it did a property assignment on them.  This mainly affects IE9/IE10 though since in newer browsers, we should just use the native implementation.

src/apis.js
gkalpak: There is some discussion in #15114. It is not as much about specific bugs as it is about our lack of confidence that there aren't any (or that our tests would have caught them).

gkalpak:
According to my understanding, the discussion on #13209 was not conclusive on whether our implementation of `HashMap` is faster than our implementation of `NgMap`. Anyone feels like benchmarking? :smiley:

Furthermore, this benchmarking would only make sense for `copy()`, because based on https://github.com/angular/angular.js/pull/15114#issuecomment-246338413 other uses of `NgMap` don't seem to have a noticeable impact on performance. (And this PR does not affect `copy()`, so performance-wise it should make no difference overall.)

The main benefit of using `NgMap` imo is that it is easy to swap it for `window.Map` when we feel we're ready. Of course, we could change the underlying implementation of `NgMapShim` to use a `HashMap` if we find out it offers better performance, but keeping a `window.Map`-compatible interface is a good thing imo.

I suggest we merge this (which is just a refactoring - not a `perf` change) and debate performance on #13209 (which is a `perf` change).

src/apis.js
dcherman: With that logic, the native Map will never be used as there is no concrete way to judge whether or not they're "ready" - that's why I was suggesting we put some kind of comment in here that "when these things are fixed, re-evaluate"

dcherman:
JSPerf is down at the moment so I can't do any basic tests, but do you have any stats on what the average size of a Map is versus the max size?  If N is low enough in your use cases, it's possible that it's fast enough, but I suspect that it'd breakdown with larger maps.  Will try to test once I can make some benchmarks.

No matter what we do, I am 100% in agreement that we should refactor any APIs to use a `Map` compatible interface.  As long as it's present in the injector as well rather than being a closure variable, applications could choose to override the private `$$Map` factory with the native one if they deem it important so long as they understand that they're tinkering with private details that are subject to change at any time.

gkalpak:
> As long as it's present in the injector as well rather than being a closure variable, applications could choose to override the private `$$Map` factory with the native one if they deem it important so long as they understand that they're tinkering with private details that are subject to change at any time.

I assume "people" is "you" 😛 The problem is that not all cases retrieve `NgMap` via DI. Off the top of my head, `NgMap` is used in the following places:

- `ngAnimate`
- `ngOptions`
- `createInjector()`

(In the future, it may also be used by `copy()`.)

`ngAnimate` already uses DI and `ngOptions` can be changed to use it too.
`createInjector()` obviously can't and I don't think `copy()` (which is probably the most critical performance-wise) can either.

dcherman:
> I assume "people" is "you" 

Yep.  I tinker with a lot of internals for fun/performance sometimes, but always with the understanding that nothing I do is supported.  Almost have an entire implementation of Observable bindings working that only participate in the digest cycle when necessary - including interpolation in templates and a new `$` binding in components :smiling_imp:

Anyway, I will try to run some basic perf tests when JSPerf is back up.

test/auto/injectorSpec.js
```diff
@@ -46,14 +46,13 @@ describe('injector', function() {
...
 
-//          s1
-//        /  | \
-//       /  s2  \
-//      /  / | \ \
-//     /s3 < s4 > s5
-//    //
-//   s6
-
+    //          s1
```
jbedard: This looks unrelated?

gkalpak: Wrong indentation is always relevant! 😛

test/helpers/testabilityPatch.js
```diff
@@ -54,17 +54,14 @@ beforeEach(function() {
...
-  if (node) {
-    node.$$hashKey = null;
-  }
-  var bod = window.document.body;
-  if (bod) {
-    bod.$$hashKey = null;
-  }
-  window.document.$$hashKey = null;
+  // These Nodes are persisted across tests.
+  // They used to be assigned a `$$hashKey` when animated, but not any more.
```
Narretz: I find this could be more explicit. If you read this in a year , would you know hat this means?

gkalpak: You wouldn't? 😱 
Updated the comment just in case 😛

gkalpak:
PTAL

Narretz:
LGTM

PR #15493, opened 2016-12-11
fix(ngModelOptions): work correctly when on the template of `replace` directives

### LINKED ISSUE #15492: ng-model-options in a template of component makes error in angular 1.6.0
I have this following code written in typescript

```

let app = angular.module('app', []);

app.component('testComponent', {
    controllerAs: 'ctrl',
    controller: class SampleComponent {
        public id: number = 10;
    },
    template: '<div><test-directive ng-model="ctrl.id"></test-directive> {{ctrl.id}} </div>'
});

class TestDirective implements ng.IDirective {

    public terminal = true;
    public scope = false;
    public replace = true;
    public required = 'ngModel';
    public template = `<input type="text" ng-model-options="{ updateOn : 'change' , allowInvalid : true }" />`;

    public constructor() { }
}

app.directive('testDirective', () => new TestDirective());
```

I'm using this component as well in angular 1.5.9 and doesn't have any problem an issue, but when I updated to angular 1.6.0 it makes error, consider this error

TypeError: Cannot read property 'getOption' of undefined

what's wrong ?

[sample link](https://github.com/ymoradi/ng61ngModelOptionsProblem)

gkalpak:
Bug fix.

In order for `ngModel` and `ngModelOptions` to work correctly together, the latter's pre-linking function should be run before the former's pre-linking function. This is typically what happens, except when `ngModel` is used on an element which also has a `replace` directive, whose template includes `ngModelOptions`. In that case, the order is reversed.
See #15492 for more details.

The initialization logic of `ngModelOptions` is moved from its pre-linking function to its controller's `$onInit()` lifecycle hook. This way, it is always guaranteed to be run before `ngModel`'s pre-linking function, even in the `replace` scenario mentioned above. 

Maybe.
There is no breaking change wrt 1.5.x (where the initialization of `ngModelOptions` used to take place in the controller itself), but wrt to 1.6.0 the following usecase will no longer be possible:
A sibling directive to `ngModelOptions`, let's call it `foo` could be modifying the model options definition (e.g. `ng-model-options="modelOptionsDefinition"`) in its `$onInit()` lifecycle hook. If the `foo` directive had a lower priority that `ngModelOptions`, then it will no longer work after this PR, since its `$onInit()` method will be called _after_ `ngModelOptions` is initialized. 
(This can be worked around by changing the priority of `foo` to be higher than `ngModelOptions` or by overwriting `ngModelOptions#$options` from `foo#$onInit()`.

**Other information**:
Fixes #15492.

### DIFF
--- a/src/ng/directive/ngModelOptions.js
+++ b/src/ng/directive/ngModelOptions.js
@@ -331,19 +331,28 @@ defaultModelOptions = new ModelOptions({
  *
  */
 var ngModelOptionsDirective = function() {
+  NgModelOptionsController.$inject = ['$attrs', '$scope'];
+  function NgModelOptionsController($attrs, $scope) {
+    this.$$attrs = $attrs;
+    this.$$scope = $scope;
+ }
+  NgModelOptionsController.prototype = {
+    $onInit: function() {
+      var parentCtrl = this.parentOptionsCtrl;
+      var parentOptions = parentCtrl ? parentCtrl.$options : defaultModelOptions;
+      var modelOptionsDefinition = this.$$scope.$eval(this.$$attrs.ngModelOptions);
+
+      this.$options = parentOptions.createChild(modelOptionsDefinition);
+    }
+  };
+
   return {
     restrict: 'A',
     // ngModelOptions needs to run before ngModel and input directives
     priority: 10,
-    require: ['ngModelOptions', '?^^ngModelOptions'],
-    controller: function NgModelOptionsController() {},
-    link: {
-      pre: function ngModelOptionsPreLinkFn(scope, element, attrs, ctrls) {
-        var optionsCtrl = ctrls[0];
-        var parentOptions = ctrls[1] ? ctrls[1].$options : defaultModelOptions;
-        optionsCtrl.$options = parentOptions.createChild(scope.$eval(attrs.ngModelOptions));
-      }
-    }
+    require: {parentOptionsCtrl: '?^^ngModelOptions'},
+    bindToController: true,
+    controller: NgModelOptionsController
   };
 };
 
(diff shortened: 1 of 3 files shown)


petebacondarwin:
On the surface I like the look of this! I need to spend some time tomorrow checking it.

src/ng/directive/ngModelOptions.js
```diff
@@ -331,19 +331,28 @@ defaultModelOptions = new ModelOptions({
...
-    require: ['ngModelOptions', '?^^ngModelOptions'],
-    controller: function NgModelOptionsController() {},
-    link: {
-      pre: function ngModelOptionsPreLinkFn(scope, element, attrs, ctrls) {
-        var optionsCtrl = ctrls[0];
-        var parentOptions = ctrls[1] ? ctrls[1].$options : defaultModelOptions;
-        optionsCtrl.$options = parentOptions.createChild(scope.$eval(attrs.ngModelOptions));
-      }
-    }
+    require: {parentOptionsCtrl: '?^^ngModelOptions'},
```
petebacondarwin: I think the reason why I moved it to the pre-link function is because this feature of requiring directly onto the controller was either not available or very new.

src/ng/directive/ngModelOptions.js
```diff
@@ -331,19 +331,28 @@ defaultModelOptions = new ModelOptions({
...
  */
 var ngModelOptionsDirective = function() {
+  NgModelOptionsController.$inject = ['$attrs', '$scope'];
+  function NgModelOptionsController($attrs, $scope) {
+    this.$$attrs = $attrs;
+    this.$$scope = $scope;
+ }
+  NgModelOptionsController.prototype = {
+    $onInit: function() {
+      var parentCtrl = this.parentOptionsCtrl;
```
petebacondarwin: given that we can name the binding anything we like why not just call it `parentCtrl` and then there would be very little motivation for this line any more.

petebacondarwin (approved):
One minor naming suggestion but LGTM.
Good fix @gkalpak

petebacondarwin (approved):

PR #15494, opened 2016-12-11
fix($rootScope): don't allow explicit digest calls to affect $evalAsync

### LINKED ISSUE #15127: Manually calling $digest on a scope prevents $rootScope.$digest from executing for pending $evalAsyncs
### Issue

This is a bug report for an intermittent issue we are seeing where due to specific timing of asynchronous callbacks executing causes $rootScope.$digest from being called when it is expected that Angular should otherwise do so.

I have a [plunkr](http://plnkr.co/edit/MdS08w) that simulates the scenario to induce the failure every time.

The current behavior is that a controller "A" scope is not digested upon conclusion of a $.then when another controller "B" scope's $digest is called manually.

This [plunkr](http://plnkr.co/edit/MdS08w) demonstrates the issue.  Please follow these steps:
1.  Click "Test (Bugged)"
2.  Observe that ctrl.result changes to "Started" but does not end with "Finished". Nor does ctrl.shared.foo display the same value as ctrl2.shared.foo.

The expected behavior is that ctrl.result ends displaying "Finished" and ctrl.shared.foo matches what is shown for ctrl2.shared.foo.

What is being demonstrated is the $digest for ctrl2 causes the digest for ctrl to be skipped.  Because ctrl is using $q the expectation is that its scope will be digested at the conclusion of the $q.then function (presumably as a result of a $rootScope.$digest being performed by Angular).
- After clicking "Reset", click the "Test (Workaround)" to see the test using a hack we are currently using in some of our code to workaround the issue. Note how the values displayed for ctrl are as expected.
  This workaround (undesirably) inspects $scope.$$asyncQueue.length and if there are any values, it avoids the manual call to $scope.$digest. This permits Angular's deferred call to process the queue and rootScope.digest to later run, and everything is digested as expected.

It appears that all browsers and all "recent" versions of Angular exhibit the issue.
### What is going on?

In our application, here is what's going on:
- Controller A and B are servicing completely separate parts of the UI/DOM. Their scopes are in different branches.
- Controller A performs an asynchronous operation (outside of angular) which it wraps in a $q.when. When the data becomes available, the corresponding $q.then is called where we load up the controller's model and then we expect Angular to digest the changes and update the DOM.
- Controller B likewise is performing an asynchronous operation (via a service actually, but that's irrelevant to honing in on the issue).  However, it is _not_ using $q and instead performs a manual call to $scope.$digest() on its own scope branch due to performance reasons. The motivation for this is an optimization to prevent digesting the entire app for something that is happening extremely frequently and affects just a tiny portion of the UI.
- Normally these two completely unrelated things work just fine, and we get what we are after with regard to Controller B's performance impact.
- However, on some occasions typically when the browser gets busy for a bit, Controller B's $digest stomps on Controller A's digest.

Tracing through Angular here is what I've come up with:
1. Controller A's async result comes in and corresponding callback invocation (which involves $q.then) is in the browser's dispatch queue. At some point just before this callback being invoked Controller B's async result callback is pushed to the queue.  This is the key to seeing the failure -- you need both callbacks to be in the browser's queue.
2. Controller A's callback is then called as expected. Angular adds the .then call to the $$asyncQueue and defers a call to flush the $$asyncQueue and do a **$rootScope.$digest**.
3. Controller B's callback is then called. Controller B's scope.$digest is called. It sees that the $$asyncQueue has stuff to do, so it does it -- **without** a $rootScope.$digest.
4. The deferred call from step 2 executes, sees that asyncQueue is empty, and so it doesn't do anything.

It seems like scope.$digest shouldn't be processing the asyncQueue like this unless the scope being digested is the rootScope.

dcherman:

fix

#15127

If an `$evalAsync` call occurs prior to a `$digest` being invoked on a scope other than the `$rootScope`, the target of the operation is changed to the `$rootScope` in order to allow the watchers that may have reacted to the `$evalAsync` operation to occur

Nope

**Other information**:

Fixes #15127

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -764,7 +764,7 @@ function $RootScopeProvider() {
         var watch, value, last, fn, get,
             watchers,
             dirty, ttl = TTL,
-            next, current, target = this,
+            next, current, target = asyncQueue.length ? $rootScope : this,
             watchLog = [],
             logIdx, asyncTask;
 
--- a/test/ng/rootScopeSpec.js
+++ b/test/ng/rootScopeSpec.js
@@ -1498,6 +1498,25 @@ describe('Scope', function() {
         $browser.defer.flush(100000);
         expect(log).toEqual(['eval-ed 1!', 'eval-ed 2!']);
       });
+
+      it('should not have execution affected by an explicit $digest call', function() {
+        var scope1 = $rootScope.$new();
+        var scope2 = $rootScope.$new();
+
+        scope1.$watch('value', function(value) {
+          scope1.result = value;
+        });
+
+        scope1.$evalAsync(function() {
+          scope1.value = 'bar';
+        });
+
+        scope2.$digest();
+
+        $browser.defer.flush(0);
+
+        expect(scope1.result).toBe('bar');
+      });
     });
   });
 


dcherman:
@gkalpak So after thinking this over for a bit, I don't think having internal `$evalAsync` calls here that affect external services/scopes not be evaluated is a problem.  That would be no different than a `$watch` affecting something similar, but not being reflected; the developer is explicitly telling us via `$digest` that only that subtree will be modified.

This fix does address the case of either `$evalAsync` (or the same via `$http`) being invoked prior to the `$digest` being called which is what the original issue was about.

gkalpak:
I personally am fine with such a change (as I am not a strong believer in the local `$digest()` optimization :smiley:). But this will make the optimization void whenever an async task is queued - keep in mind that the following "things" use `$evalAsync`:

- `$q` (and everything that uses `$q` under the hood, e.g. `$timeout`, `$http`, some animation features, async `ngModel` validators etc)
- `$interval` 
- `$watchCollection`
- `ngFocus/ngBlur`
- (Other internal stuff that are less likely to affect apps.)

Yet, imo voiding an optimization is preferrable to unexpectedly broken behavior (because some bindings did not propagate).

##
I am a little on the fence on whether this is a breaking change though, as it does break certain assumptions that used to hold. (Although it is not very likely to actually breaks apps, I wonder how tests could be affected by this.)

##
**tl;dr**

Looks OK for 1.7.x - Still not sure about 1.6.x

petebacondarwin:
I am not that keen on this change.
I am sure that we have always encouraged people to use `$apply` not `$digest`, and those that choose not to have to accept that they are not going to get guaranteed behaviour across the app.

gkalpak:
The problem with the current implementation is that a global `$digest` might not be run, because some other unrelated scope happened to also `$digest` locally. For example:

```js
// Imagine two totally unrelated scopes (e.g. two directives scopes coming from
// third party libraries and located in distinct and unrelated parts of the page.
var scope1 = $rootScope.$new();
var scope2 = $rootScope.$new();

// `scope1` calls `$evalAsync()` (and expects the updated values to propagate to the view
scope1.$evalAsync(someExpr);

// Before the next tick (when `$rootScope.$digest()` will be called)
// `scope2` calls `$digest()` to update some value in its own view
scope2.$digest();

// Now, because of that last call, `someExpr` will be evaluated, the `asyncQueue` will be
// emptied and as a result, `$rootScope.$digest()` won't run.
```

petebacondarwin:
Yes, and I am saying that people shouldn't be doing `scope2.$digest()` :-P
Unfortunately I appreciate that they do for performance reasons.

So how about we do not drain the async queue if the call to `$digest` was not on the root scope? This is subtly different to the current proposed solution.

dcherman:
@petebacondarwin That would be a lot worse than the current behavior.  You would no longer be able to use anything that relies on `$evalAsync` whether it be direct usage, an already resolved promise using `.then`, anything that uses `$watchGroup`, etc...

gkalpak:
How about #15501?

dcherman:
That's definitely an alternative, but I'm unclear on why that's better since it causes additional loops of the digest cycle whereas this keeps it as minimal as possible.

Is it to address situations where someone is mutating global (or just outside state) inside a digest cycle that was initated with `$digest`?

gkalpak:
I am not sure it is better. It is just an alternative, I guess :smiley:

Initially I thought it was a good idea, because this keeps the execution order closer to what I would expect and it might also be less expensive if the local digest doesn't stabilize immediately. But today it doesn't seem any better than your version (both having pros and cons).

gkalpak:
After thinking it some more, I think this one is better. So, I am again back to: 

Looks OK for 1.7.x - Still not sure about 1.6.x

PR #15501, opened 2016-12-12
fix($rootScope): correctly propagate async changes after local `$digest()`

### LINKED ISSUE #15127: Manually calling $digest on a scope prevents $rootScope.$digest from executing for pending $evalAsyncs

gkalpak:
Bug fix.

If an async task is scheduled (via `$evalAsync()`) on a scope and `$digest()` happens to be called on another, unrelated scope (local digest), then the `asyncQueue` will be drained but `$rootScope.$digest()` will not be called (potentially preventing the changes to propagate correctly through the app).
See #15127.

Angular keeps track on whether `$digest()` has been called on `$rootScope` or not and calls `$rootScope.$digest()` if necessary (even if the `asyncQueue` has been drained by a local `$digest()`).

No.

**Other information**:
This is a more targeted alternative to #15494.
Fixes #15127.

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -799,6 +799,10 @@ function $RootScopeProvider() {
           }
           asyncQueue.length = 0;
 
+          if (this === $rootScope) {
+            hasRootScopeDigestRun = true;
+          }
+
           traverseScopesLoop:
           do { // "traverse the scopes" loop
             if ((watchers = current.$$watchers)) {
@@ -1014,8 +1018,9 @@ function $RootScopeProvider() {
         // if we are outside of an $digest loop and this is the first time we are scheduling async
         // task also schedule async auto-flush
         if (!$rootScope.$$phase && !asyncQueue.length) {
+          hasRootScopeDigestRun = false;
           $browser.defer(function() {
-            if (asyncQueue.length) {
+            if (!hasRootScopeDigestRun || asyncQueue.length) {
               $rootScope.$digest();
             }
           });
@@ -1333,6 +1338,7 @@ function $RootScopeProvider() {
     var postDigestQueue = $rootScope.$$postDigestQueue = [];
     var applyAsyncQueue = $rootScope.$$applyAsyncQueue = [];
 
+    var hasRootScopeDigestRun = false;
     var postDigestQueuePosition = 0;
 
     return $rootScope;
--- a/test/ng/rootScopeSpec.js
+++ b/test/ng/rootScopeSpec.js
@@ -1498,6 +1498,26 @@ describe('Scope', function() {
         $browser.defer.flush(100000);
         expect(log).toEqual(['eval-ed 1!', 'eval-ed 2!']);
       });
+
+
+      it('should not have execution affected by a local `$digest` call', function() {
+        var scope1 = $rootScope.$new();
+        var scope2 = $rootScope.$new();
+
+        scope1.$watch('value', function(value) {
+          scope1.result = value;
+        });
+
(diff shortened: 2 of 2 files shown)


gkalpak:
Thinking it over, I think #15494 (even if not perfect) is a better alternative.

PR #15506, opened 2016-12-13
fix(input): fix `step` validation for `input[type=number/range]`

### LINKED ISSUE #15504: Broken ngStep for input[type="number"] directive

It's a *bug*

For values like `1.16`, `20.1` etc the `input[type=number]` field with `ng-step="0.01"` is marked as invalid.

Steps to reproduce:

1. Update to AngularJS version `1.6.0`
2. Define an `input[type=number]` with `ng-step="0.01"`
3. For values like `1.16`, `20.1` the field will be marked as `$invalid`

See a small demo: https://plnkr.co/edit/LW2BsaA7FfEVu9nB6RoG?p=preview

I believe it's a problem with floating points arithmetic and this little method https://github.com/angular/angular.js/blob/master/src/ng/directive/input.js#L1563

```javascript
    // ...

    value = value * multiplier;
    stepBase = stepBase * multiplier;
    step = step * multiplier;
  }

  return (value - stepBase) % step === 0;
```

When for instance `value = 1.16` and `multiplier = 100` `value * multiplier` gives us `115.99999999999999` so the return expression returns false even when the given value is valid.

gkalpak:
Bug fix.

The validation will incorrectly fail in certain cases (e.g. `step: 0.01`, `value: 1.16 or 20.1`), due to Floating Point Arithmetic limitations. The previous fix for FPA limitations (081d06ff) tried to solve the issue by converting the numbers to integers, before doing the actual calculation, but it failed to account for cases where the conversion itself returned non-integer values (again due to FPA limitations).
See also #15504.

The validation works correctly by ensuring that the values used in the final calculation are always integers.

No.

**Other information**:
Fixes #15504.

### DIFF
--- a/src/ng/directive/input.js
+++ b/src/ng/directive/input.js
@@ -1565,15 +1565,27 @@ function isValidForStep(viewValue, stepBase, step) {
   // and `viewValue` is expected to be a valid stringified number.
   var value = Number(viewValue);
 
+  var isNonIntegerValue = !isNumberInteger(value);
+  var isNonIntegerStepBase = !isNumberInteger(stepBase);
+  var isNonIntegerStep = !isNumberInteger(step);
+
   // Due to limitations in Floating Point Arithmetic (e.g. `0.3 - 0.2 !== 0.1` or
   // `0.5 % 0.1 !== 0`), we need to convert all numbers to integers.
-  if (!isNumberInteger(value) || !isNumberInteger(stepBase) || !isNumberInteger(step)) {
-    var decimalCount = Math.max(countDecimals(value), countDecimals(stepBase), countDecimals(step));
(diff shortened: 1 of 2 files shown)


Narretz:
Looks like IE9 doesn't like the fix:
```
IE 9.0.0 (Windows 7 0.0.0) input range step should correctly validate even in cases where the JS floating point arithmetic fails FAILED

	Expected to have ng-valid and not have ng-invalid but had ng-untouched ng-not-empty ng-dirty ng-valid-number ng-invalid ng-inv
```

gkalpak:
For the first time in a long time, IE9 was right and I was wrong :smiley:
PTAL

Narretz:
LGTM

PR #15507, opened 2016-12-13
docs($sanitize): Fix not working example

fengthedroid:
doc update

before:
screen shot 2016-12-13 at 3 25 06 pm

after:
screen shot 2016-12-13 at 3 25 33 pm


fengthedroid:
I signed it!

gkalpak:
Thx for submitting the PR. This change is not valid though. One column is supposed to show the "raw" HTML snippet (thus the escaped `<`/`>`) and the other column contains the actual result of that HTML snippet.

PR #15513, opened 2016-12-15
revert: fix($resource): allow params in `hostname` (except for IPv6 addresses)

matsko:
This reverts commit 7f45b5fee79e2cb87d65bdd015d455304cec1ee4.


Narretz:
Can you please add to the commit message why this has been reverted (i.e what use it breaks)? Doesn't have to be google specific ;)

mgol:
It'd be good to have a test that verifies the thing 7f45b5f broke is no longer broken.

gkalpak:
We'll have more info once the breakage has been investigated (probably next week). For now, the fix was reverted (since it isn't critical), so we are unblocked for the 1.5.10 release.

Issue #14124, opened 2016-02-24
ngAnimate with ngRepeat in directive using  templateUrl issue


xavadu:
Hello,

Sorry but i can't reproduce this error on plunkr yet.

The issue is that ngRepeat animation fail for the first time when is applied on a directive that use a templateUrl, it start to happen in 1.5.0 upgrade, on 1.4.1 it was working fine. 

The css animation never get the state of _.ng-enter-active_ and it keep on the _.ng-enter_ state.

some PoC:

```
.animate-repeat.ng-enter,
.animate-repeat.ng-move {
    transition: 0.5s ease-out all;  
    animation-fill-mode: both;
}

.animate-repeat.ng-enter {
  background-color: red;
  transform: translate3d(0, -5%, 0);
}

.animate-repeat.ng-enter.ng-enter-active {
  background-color: blue;
  transform: none;
}

.animate-repeat.ng-enter-stagger {
  transition-delay: 0.1s;
  animation-delay: 0.1s;
  transition-duration: 0s;
}

<my-directive ng-repeat="colors in sky" class="animate-repeat"><my-directive>
```

if _sky_ is equal to null, not defined or empty array, etc.. it happen, if _sky_ is previously populate it doesn't happen. On our case _sky_ is populate by an $http request after some user interaction in a form. 

The situation is that myDirective keep with background red and never get the background blue. Later a re-submit, it works, because _sky_ had data previously and wasn't null or empty. 

The interesting is that it occur only if my-directive use a templateUrl, if the template is inline all works fine. 

Is quite complicate to explain, let me know if you need a better description of some part. 

It become really annoying when instead a a background color, you are going from opacity 0 to opacity 1. So, the ngRepeat list don't appear.

dcherman:
In all likelihood, this is a duplicate of https://github.com/angular/angular.js/issues/14074

Narretz:
It does sound a bit different than the other issue, so a plnkr would really help. Especially the part where it works the second time.

dcherman:
@Narretz Yup, would definitely help.  The reason it might work (and be the same issue) is that once transclusion occurs, the original async directive is replaced with a derived sync one, so the second time the ngRepeat unrolls after the `$templateRequest` completes, it wouldn't exhibit the bug.

xavadu:
Hello, 

[...]

Issue #14074, opened 2016-02-18
$animateCss used on a directive that uses a templateUrl breaks


CodySchaaf:
$animateCss ends up operating on an element that does not end up in the view after the directive is rendered.

As you can see in this plunker if you uncomment the template (and comment the templateUrl) the element is correctly styled and appears in the view. However when the directive is using tamplateUrl the element is not styled.

https://plnkr.co/edit/yO1AvvupoveRi3IWevVL?p=preview

This was not the case in 1.4.x, but is a new issue when upgrading to 1.5.

While the directive is using templateUrl you can also check the console and click reveal in elements pane to see that no element is revealed. 

This issue means that you can not do any clean up in the done function or any further styling (my issue is I want to scroll the element into view once it is expanded since the element would be too short to scroll into view before. You can also see that '*-active' and '*-add-active' styles are not removed.

CodySchaaf:
I also see this error in my own project  `Cannot read property '$$ngAnimateParentKey' of null` which comes from this line 

```js
return (c.$$ngAnimateParentKey || (c.$$ngAnimateParentKey = ++K)) + "-" + a.getAttribute("class") + "-" + b
```

Narretz:
Thanks for reporting. Can you please create a plnkr that repdroduces the error you mentioned?

CodySchaaf:
There is a plnkr included in the original story, here it is again. Let me know if this one doesn't work.

https://plnkr.co/edit/yO1AvvupoveRi3IWevVL?p=preview

Narretz:
@CodySchaaf You said you have an error in your project. I was specifically asking for a repro for that error, because your posted plnkr doesn't throw an error.

Narretz:
Also, as far as I can see, the behavior in 1.4.9 and 1.5.0 is the same: no animation with templateUrl.

Narretz:
So I doubt this worked in 1.4.x, because this inherently happens because with templateUrl, it takes at least a few milliseconds until the template is available and the direcitve is linked. But the animation runs before that happens, so there's nothing to animate. This is a known limitation. You can fix this by putting the template in the cache: https://plnkr.co/edit/o7SyQNY5oMZxAKVh7Eeg?p=preview

CodySchaaf:
Oh that makes sense. Sorry. 

[...]

PR #15514, opened 2016-12-16
fix(ngAnimate): correctly animate transcluded clones with `templateUrl`

### LINKED ISSUE #15510: Enter/leave animation not working for for first item in ng-repeat
If an element has both ng-repeat and a directive that uses templateUrl `<div my-dir ng-repeat="...">`, the enter animation doesn't work for the first item, even if the template is in $templateCache.

[Demo](http://plnkr.co/edit/j86chAryADuQsSM8ctte?p=preview)

If you don't pre-warm the $templateCache, the leave animation doesn't work either. You can comment out the run block in the demo to see this.

If the directive is on a child element, the problem goes away, and all animations work. And if fetching the template is fast enough, enter and leave animations both work from a cold cache.

    <div ng-repeat="x in y"> <div my-dir></div> </div>

I tested this in Angular 1.5.8 and 1.6.0, with Chrome 54, Safari 10, and Firefox 50.

@gkalpak suggested I make this ticket in #12612.

gkalpak:
Bug fix.

`$animate` decides whether an animation should be cancelled based on some assumption that don't hold in specific cases (e.g. when animating transcluded clones with `templateUrl` directives on them for the first time). As a result, the entering elements will not be animated in such cases. This affects commonly used, structural built-in directives (`ngIf`, `ngRepeat`, `ngSwitch` etc).
See also #15510.

All entering animations work as expected (from the first time).

No.

**Other information**:
BTW, I believe this is a side effect of lazily compiling transcluded content.

In addition to the last `fix` commit that fixes the bug, this PR also includes several `refactor` commits that simplify `$animate` internally, mainly by removing unnecessary wrapping/unwrapping of DOM nodes in `jqLite`/`jQuery` collections. This changes may also positively affect performance, especially in animation-heavy apps (~~although I haven't benchmarked it - yet~~ it does indeed :wink:).
(It is probably easier to review one commit at a time.)

Party addresses #14074 and #14124.

Fixes #15510.

### DIFF
--- a/benchmarks/animation-bp/app.js
+++ b/benchmarks/animation-bp/app.js
@@ -1,0 +1,44 @@
+'use strict';
+
+angular
+  .module('animationBenchmark', ['ngAnimate'], config)
+  .controller('BenchmarkController', BenchmarkController);
+
+// Functions - Definitions
+function config($compileProvider) {
+  $compileProvider
+    .commentDirectivesEnabled(false)
+    .cssClassDirectivesEnabled(false)
+    .debugInfoEnabled(false);
+}
+
+function BenchmarkController($scope) {
+  var self = this;
+  var itemCount = 1000;
+  var items = (new Array(itemCount + 1)).join('.').split('');
+
+  benchmarkSteps.push({
+    name: 'create',
+    fn: function() {
+      $scope.$apply(function() {
+        self.items = items;
+      });
+    }
+  });
+
+  benchmarkSteps.push({
+    name: '$digest',
+    fn: function() {
+      $scope.$root.$digest();
+    }
+  });
+
+  benchmarkSteps.push({
+    name: 'destroy',
+    fn: function() {
+      $scope.$apply(function() {
+        self.items = [];
+      });
+    }
+  });
+}
--- a/benchmarks/animation-bp/main.html
+++ b/benchmarks/animation-bp/main.html
@@ -1,0 +1,28 @@
+<style>
+  [ng-cloak] { display: none !important; }
+  .animation-container .ng-enter,
+  .animation-container .ng-leave {
+    transition: all 0.1s;
+  }
+
+  .animation-container .ng-enter,
+  .animation-container .ng-leave.ng-leave-active {
+    opacity: 0;
+  }
+
+  .animation-container .ng-enter.ng-enter-active,
+  .animation-container .ng-leave {
+    opacity: 1;
+  }
+</style>
+<div ng-app="animationBenchmark" ng-cloak ng-controller="BenchmarkController as bm">
+  <div class="container-fluid">
+    <h2>Large collection of elements animated in and out with ngAnimate</h2>
+
+    <div class="animation-container">
+      <div ng-repeat="i in bm.items track by $index">
+        Just a plain ol' element
+      </div>
+    </div>
+  </div>
+</div>
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -302,13 +302,13 @@ var $$AnimateQueueProvider = ['$animateProvider', /** @this */ function($animate
 
     return $animate;
 
-    function queueAnimation(element, event, initialOptions) {
+    function queueAnimation(originalElement, event, initialOptions) {
       // we always make a copy of the options since
       // there should never be any side effects on
       // the input data when running `$animateCss`.
       var options = copy(initialOptions);
 
-      element = stripCommentsFromElement(element);
+      var element = stripCommentsFromElement(originalElement);
       var node = getDomNode(element);
       var parentNode = node && node.parentNode;
 
@@ -479,6 +479,15 @@ var $$AnimateQueueProvider = ['$animateProvider', /** @this */ function($animate
       markElementAnimationState(node, PRE_DIGEST_STATE, newAnimation);
 
       $rootScope.$$postDigest(function() {
+        // It is possible that the DOM nodes inside `originalElement` have been replaced. This can
+        // happen if the animated element is a transcluded clone and also has a `templateUrl`
+        // directive on it. Therefore, we must recreate `element` in order to interact with the
+        // actual DOM nodes.
+        // Note: We still need to use the old `node` for certain things, such as looking up in
+        //       HashMaps where it was used as the key.
+
+        element = stripCommentsFromElement(originalElement);
+
         var animationDetails = activeAnimationsLookup.get(node);
         var animationCancelled = !animationDetails;
         animationDetails = animationDetails || {};
--- a/src/ngAnimate/animateQueue.js
+++ b/src/ngAnimate/animateQueue.js
@@ -181,10 +181,7 @@ var $$AnimateQueueProvider = ['$animateProvider', /** @this */ function($animate
       return this === arg || !!(this.compareDocumentPosition(arg) & 16);
     };
 
-    function findCallbacks(parent, element, event) {
-      var targetNode = getDomNode(element);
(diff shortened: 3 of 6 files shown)


gkalpak:
I added one more commit with a (very basic) animation benchmark. It simply `$animate.enter`s and immediatelly `$animate.leave`s 1000 elements - via `ngRepeat`. With the changes proposed in this commit, the benchmark shows a 5%-7% reduction in total execution time (which is not bad at all for a bunch of refatoring commits 😛 - I am tempted to change them to `perf`).

<sub>(Take the results with a gran of salt, as the benchmark scenario is a little artificial.)</sub>

test/ngAnimate/animateSpec.js
```diff
@@ -459,6 +459,26 @@ describe('animations', function() {
...
+      module(function($compileProvider) {
+        $compileProvider.directive('foo', valueFn({templateUrl: 'foo.html'}));
+      });
+
+      inject(function($animate, $compile, $rootScope, $templateCache) {
+        parent.append(jqLite('<foo ng-if="showFoo"></foo>'));
+        $templateCache.put('foo.html', '<div>FOO</div>');
+        $compile(parent)($rootScope);
+
+        expect(capturedAnimation).toBe(null);
```
mgol: Everywhere in other places it's only checked if `capturedAnimation` is falsy. We should be consistent.

benchmarks/animation-bp/app.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+angular.
+  module('animationBenchmark', ['ngAnimate'], config).
```
mgol: I dislike this dot placement a lot. :P We use the leading dot in most places and I'd like to ESLint-enforce it one day...

benchmarks/animation-bp/main.html
```diff
@@ -0,0 +1,28 @@
+<style>
+  [ng-cloak] { display: none !important; }
+  .animation-container .ng-enter,
+  .animation-container .ng-leave {
+    transition: all 0.1s
```
mgol: A missing semi-colon;

mgol (changes requested):

benchmarks/animation-bp/main.html
gkalpak: 😱

src/ngAnimate/animateQueue.js
```diff
@@ -610,65 +610,61 @@ var $$AnimateQueueProvider = ['$animateProvider', /** @this */ function($animate
...
-    function areAnimationsAllowed(element, parentElement, event) {
-      var bodyElement = jqLite($document[0].body);
-      var bodyElementDetected = isMatchingElement(element, bodyElement) || element[0].nodeName === 'HTML';
-      var rootElementDetected = isMatchingElement(element, $rootElement);
+    function areAnimationsAllowed(node, parentNode, event) {
+      var bodyNode = $document[0].body;
+      var rootNode = getDomNode($rootElement);
+
+      var bodyNodeDetected = (node === bodyNode) || node.nodeName === 'HTML';
+      var rootNodeDetected = (node === rootNode);
```
gkalpak: A couple of reasons (at least):

1. I find them more readable. I think it is a remnant from my early Java days (and [this ancient document][1]).
2. It is also visually closer to `isMatchingElement(...)` 😛

<sub>(Hey, I changed the trailing dots, I am not removing the parens.)</sub>

[1]: http://www.oracle.com/technetwork/java/codeconventions-150003.pdf

test/ngAnimate/animateSpec.js
gkalpak: So, you want me to change all other places to `toBe(null)`? Fair enough...

benchmarks/animation-bp/app.js
mgol: > Come on! I speeded up your animations 5%-7%. And all you care about is where I put a tiny little dot?

No, I like this PR a lot! I'm just providing feedback for some small issues. :)

Good job overall, though, perhaps I should have spelled that out. 👏

gkalpak:
Addressed (most of) @mgol's comments.

test/ngAnimate/animateSpec.js
mgol: I was thinking the other way round but maybe your idea has more sense. ;)

In some cases it's initially `undefined`, though, not `null`, this file is not very consistent in that regard. Perhaps that's material for a separate PR (or at least a commit) but I just want us to be consistent. :)

test/ngAnimate/animateSpec.js
```diff
@@ -168,7 +168,7 @@ describe('animations', function() {
       inject(function($animate, $rootScope) {
         $animate.enter(element, parent);
         $rootScope.$digest();
-        expect(capturedAnimation).toBeFalsy();
+        expect(capturedAnimation).toBe(null);
```
mgol: I'd put those changes in a separate commit as they don't really have a lot in common with the original one.

test/ngAnimate/animateSpec.js
mgol: Ah, it's `undefined` only in one test:
https://github.com/gkalpak/angular.js/blob/acd9e7b757cdb9f04ae7a12ff21fdc8d27f314b5/test/ngAnimate/animateSpec.js#L21-L46

I guess the changes are fine then.

gkalpak: I (blindly) changed all occurrences of `expect(capturedAnimation).toBeFalsy()` to `expect(capturedAnimation).toBe(null)` and the tests were passing :confused:

benchmarks/animation-bp/app.js
gkalpak: <sub>Just in case it wasn't clear, I was just kidding (in an attempt to distract everyone from the fact that I was busted trying to sneak in those trailing dots 😛 </sub>

gkalpak:
Rearranged the commits (and fixed a test failure).

src/ngAnimate/animateQueue.js
Narretz: Looks like we need an eslint rule for that :D But I prefer the parens too

mgol: I like not having parents here but as long as ESLint will shout at me if I forget them I'd be satisfied as well. ;)

src/ngAnimate/animateQueue.js
```diff
@@ -479,6 +479,15 @@ var $$AnimateQueueProvider = ['$animateProvider', /** @this */ function($animate
...
 
       $rootScope.$$postDigest(function() {
+        // It is possible that the DOM nodes inside `originalElement` have been replaced. This can
+        // happen if the animated element is a transcluded clone and also has a `templateUrl`
+        // directive on it. Therefore, we must recreate `element` in order to interact with the
+        // actual DOM nodes.
+        // Note: We still need to use the old `node` for certain things, such as looking up in
+        //       HashMaps where it was used as the key.
+
+        element = stripCommentsFromElement(originalElement);
```
Narretz: I take it this has no performance implications? This now applies to all animations.

gkalpak: I haven't benchmarked this change on its own. I assume it has an unnoticeable impact on performance (considering all the performance-heavy-ish stuff that is going on during animations). But overall, this PR offers a non-negligible performance gain (see https://github.com/angular/angular.js/pull/15514#issuecomment-267764426).

gkalpak:
@mgol, @Narretz: PTAL

Narretz:
LGTM
@mgol can you check this again?

mgol (approved):

Issue #14631, opened 2016-05-18
$q: Unhandled rejections should not be stringified


dcherman:
**_Note_: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

Unhandled rejections are converted to a string prior to being passed to `$exceptionHandler`.  Also, errors aren't correctly stringified either.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**

http://jsfiddle.net/rogqp47y/28/

**What is the expected behavior?**

The value that the promise was rejected with should be passed to `$exceptionHandler`

**What is the motivation / use case for changing the behavior?**

An unhandled rejection should be treated no differently than an unhandled exception - if I throw an uncaught error, I expect to catch that in the global $exceptionHandler, not a string representing the error.  The default implementation of `$exceptionHandler` that logs to the console should handle the stringification.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**

Git / Nightly (Future 1.6)

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

lgalfaso:
I do not think that there is a clear cut (this is that `$exceptionHandler` should receive the value from an unhandled rejected promise). The reason being that there is no reason to believe that this value will be an `Error`. This is, if the value is just passed to `$exceptionHandler`, there is no context to know that this came from an unhandled promise. That said, I am open to change https://github.com/angular/angular.js/blob/master/src/ng/q.js#L383 to `exceptionHandler(extends(new UnhandledPromiseError(), {reason: toCheck.value}), errorMessage);` (or something equivalent) `UnhandledPromiseError` being something that extends `Error`.

gkalpak:
I agree that it should be clear why this is an error (i.e. `possibly unhandled rejection`).
I was thinking we might add an independent API about notifying for unhandled rejections. Something similar to the [unhandledrejection event][1].

[1]: https://developer.mozilla.org/en-US/docs/Web/Events/unhandledrejection

dcherman:
[...]

PR #15527, opened 2016-12-19
Add traceback to unhandled promise rejections, Fixes: #14631

graingert:
Feature

Console is filled with meaningless:  `Possibly unhandled rejection: {}` errors

Console is filled with juicy traceback info.

I don't think so

**Other information**:

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -381,7 +381,11 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
       if (!toCheck.pur) {
         toCheck.pur = true;
         var errorMessage = 'Possibly unhandled rejection: ' + toDebugString(toCheck.value);
-        exceptionHandler(errorMessage);
+        if (toCheck.value instanceof Error) {
+          exceptionHandler(toCheck.value, errorMessage);
+        } else {
+          exceptionHandler(errorMessage);
+        }
       }
     }
   }
--- a/test/ng/qSpec.js
+++ b/test/ng/qSpec.js
@@ -181,15 +181,13 @@ describe('q', function() {
   };
 
 
-  function exceptionHandler(reason) {
-    exceptionHandlerCalls.push(reason);
+  function exceptionHandler(exception, reason) {
+    if (typeof reason === 'undefined') {
+      exceptionHandlerCalls.push({ reason: exception });
+    } else {
+      exceptionHandlerCalls.push({ reason: reason, exception: exception });
+    }
   }
-
-
-  function exceptionHandlerStr() {
-    return exceptionHandlerCalls.join('; ');
-  }
-
 
   beforeEach(function() {
     q = qFactory(mockNextTick.nextTick, exceptionHandler, true);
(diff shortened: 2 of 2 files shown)


graingert:
@gkalpak uhh, are the tests for the unhandled rejection even run? I added these to intentionally fail...

gkalpak:
Not sure what you mean. The tests should run (and I am quite sure they do).
(The tests need some fixes, because they are not currently testing what they should be testing, but I that is unrelated.)

graingert:
@gkalpak https://github.com/angular/angular.js/pull/15527/files#diff-0cfb390955e411a4a5437fa337d52481R2181 should fail on the fixture.type === 'exception' because it doesn't stringify to 'foo'.

graingert:
@gkalpak are you in IRC?

gkalpak:
@graingert, if you want it to fail, then you better [not reject the promise with `foo`] 😛
<sub>(I am rarely (read "never") on IRC these days, but you can find me on gitter.)</sub>

[1]: https://github.com/angular/angular.js/pull/15527/files#diff-0cfb390955e411a4a5437fa337d52481R2179

graingert:
@gkalpak oh I am a silly... Fixing...

gkalpak:
BTW, it would made our life a tiny bit easier if you followed our [commit message guidelines][1] :wink:

graingert:
@gkalpak will do once I've got the tests working. Also your URL is out of date: you want https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format

src/ng/q.js
```diff
@@ -381,7 +381,11 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
       if (!toCheck.pur) {
         toCheck.pur = true;
         var errorMessage = 'Possibly unhandled rejection: ' + toDebugString(toCheck.value);
```
gkalpak: Here is another idea (either for `toDebugString()` or just here):
If `toCheck.value` has `toString()` method that is not the default `Object.prototype.toString()`, then use that for serializing it. Something similar to [stringify][1].

[1]: https://github.com/angular/angular.js/blob/465d1734559ca4a7f4aa24387060f88fcc53ecb1/src/Angular.js#L1995

gkalpak:
Thx! (My URL was fine btw - just intentionally pointing to the supersection of yours 😛)

graingert:
@gkalpak ah it didn't go to a section for me at all.

graingert:
@gkalpak and we're green!

gkalpak:
@graingert, awesome! Now on to that commit message 😛

test/ng/qSpec.js
```diff
@@ -2167,45 +2165,69 @@ describe('q', function() {
...
-      var defer = q.defer();
-      defer.reject('foo');
-      mockNextTick.flush();
-      expect(exceptionHandlerStr()).toBe('Possibly unhandled rejection: foo');
-    });
+    var exceptionEg = new Error('Fail');
+    var exceptionStr = toDebugString(exceptionEg);
+    var fixtures = [
+      {
+        type: 'exception',
```
gkalpak: Change "exception" to "Error object".

test/ng/qSpec.js
```diff
@@ -2167,45 +2165,69 @@ describe('q', function() {
...
+        expected: {
+          reason: 'Possibly unhandled rejection: foo'
+        }
+      }
+    ];
+    forEach(fixtures, function(fixture) {
+      var type = fixture.type;
+      var value = fixture.value;
+      var expected = fixture.expected;
+      it('should log an unhandled' + type + ' rejected promise', function() {
```
gkalpak: Instead of repeating `type` in all `it` blocks, create a `describe block for each fixture and put `type` in its description.

test/ng/qSpec.js
```diff
@@ -2167,45 +2165,69 @@ describe('q', function() {
...
+        value: 'foo',
+        expected: {
+          reason: 'Possibly unhandled rejection: foo'
+        }
+      }
+    ];
+    forEach(fixtures, function(fixture) {
+      var type = fixture.type;
+      var value = fixture.value;
+      var expected = fixture.expected;
```
gkalpak: Extra points for putting some space before `it` blocks :smiley:

test/ng/qSpec.js
```diff
@@ -2167,45 +2165,69 @@ describe('q', function() {
...
+          reason: 'Possibly unhandled rejection: ' + exceptionStr
+        }
+      },
+      {
+        type: 'plain value',
+        value: 'foo',
+        expected: {
+          reason: 'Possibly unhandled rejection: foo'
+        }
+      }
```
gkalpak: A couple more fixtures would be in order imo. E.g.: 

- One with a non-`Error` object.
- One with an object that "extends" `Error`.

gkalpak (changes requested):
I left some comments, but the general idea LGTM.

test/ng/qSpec.js
graingert: Cool

test/ng/qSpec.js
graingert: Yeah makes sense

test/ng/qSpec.js
graingert: Ah makes sense

graingert:
@gkalpak done

test/ng/qSpec.js
```diff
@@ -2167,45 +2165,97 @@ describe('q', function() {
...
+      },
+      {
+        type: 'Non-Error object',
+        value: nonErrorObj,
+        expected: {
+          reason: 'Possibly unhandled rejection: ' + nonErrorObjStr
+        }
+      },
+      {
+        type: 'plain value',
```
gkalpak: Now everyone is capitalized except for `plain value`. Just because it is nothing more than a plain value, doesn't mean you have to treat it like a 2nd-class citizen 😛

test/ng/qSpec.js
gkalpak: Or `describe` blocks...

test/ng/qSpec.js
```diff
@@ -2167,45 +2165,97 @@ describe('q', function() {
...
+        expected: {
+          reason: 'Possibly unhandled rejection: foo'
+        }
+      }
+    ];
+    forEach(fixtures, function(fixture) {
+      var type = fixture.type;
+      var value = fixture.value;
+      var expected = fixture.expected;
+      describe(type, function() {
```
gkalpak: Nit: Using something like `'with ' + type` will make the description read better (in case of an error for example).

gkalpak (approved):
A couple of nitpicks, but otherwise LGTM.

test/ng/qSpec.js
graingert: I went with lower case, and more accurate names

graingert:
@gkalpak done

gkalpak (approved):
LGTM, but let's wait for some feedback from others as well.
<sub>(In the mean time, feel free to fix up that commit message :grin:)</sub>

graingert:
@gkalpak I did fix the commit message... What's wrong with it?

gkalpak:
Actual:

> fix($q): Add traceback to unhandled promise rejections, Fixes: #14631

Expected something like:

> fix($q): add traceback to unhandled promise rejections
>
> Fixes #14631

mgol:
@graingert Thanks for the PR, this is awesome. 👏 

I don't have any remarks apart from the incorrect commit message format.

graingert:
@gkalpak commit message fixed

src/ng/q.js
```diff
@@ -381,7 +381,11 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
...
-        exceptionHandler(errorMessage);
+        if (toCheck.value instanceof Error) {
+          exceptionHandler(toCheck.value, errorMessage);
```
mgol: Is this the correct order? It looks weird when the error message is *after* the stack trace.

mgol (changes requested):

src/ng/q.js
mgol: Hmm, OK. I mostly wondered because this:
screen shot 2016-12-21 at 15 48 43
looks a little weird.

mgol: There is nothing in the `$exceptionHandler` code that would make the order important; that's all of it:
```js
function $ExceptionHandlerProvider() {
  this.$get = ['$log', function($log) {
    return function(exception, cause) {
      $log.error.apply($log, arguments);
    };
  }];
}
```

I wonder what's the reason for the current documentation. @gkalpak, @petebacondarwin does any one of you know?

graingert: what about just doing
```js
exceptionHandler(toCheck.value, 'Possibly unhandled rejection')
```

mgol: That sounds sensible but I'd still swap the order of operations. Note that currently we are trying to show the `'Possibly unhandled rejection'` message first and the error afterwards; we just put it all in one string argument so the stack trace gets lost. So I'd like:
```js
exceptionHandler('Possibly unhandled rejection', toCheck.value);
```
but that would require updating the docs to no longer require the cause to be the second parameter (no code changes required as it already works as I mentioned before).

I'd just want to know if there's any reason for the current signature so let's see what others say. :)

graingert: @mgol my custom $exceptionHandlers are configured to pass the first argument as the Error value, as per the documentation.

I presume others follow the same documentation also.

graingert: @mgol swapping the types of the parameters of $exceptionHandler is another issue for another PR, and would be a breaking change.

mgol: OK, fair enough. In that case, maybe it's better to leave it as it is for now; your proposal to just write:
```js
exceptionHandler(toCheck.value, 'Possibly unhandled rejection');
```
sounds nice but we pass other data types through `toDebugString` and we'd lose it in that case. If we want to tweak that, it would be better to tackle that separately after this PR lands.

mgol (approved):

src/ng/q.js
```diff
@@ -381,7 +381,11 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
       if (!toCheck.pur) {
         toCheck.pur = true;
         var errorMessage = 'Possibly unhandled rejection: ' + toDebugString(toCheck.value);
-        exceptionHandler(errorMessage);
+        if (toCheck.value instanceof Error) {
```
dcherman: Though I don't think it necessarily a concern here, do we care that this will not identify errors that originate from an iframe?  Don't know how common of a use case that is.

mgol: I doubt it's common but if it turns out to be an issue we can revisit.

Ideally we would convert to a string (via `toDebugString`) only certain types of values and log others unchanged. The problem is that then we'd have to pass `Possibly unhandled rejection` and `toCheck.value` as separate arguments to `$exceptionHandler` and if we do it in a documented way (error first, (optional) reason second) the message would get swapped from e.g.:
```
'Possibly unhandled rejection: MY_THING'
```
to:
```
'MY_THING' 'Possibly unhandled rejection'
```
which is less pleasant.

We can discuss it but let's do it in a separate issue if desired.

mgol:
Landed, thanks!

mgol:
@graingert One nit (I fixed it myself): there should be no colon after `Fixes` in the commit description.

PR #15528, opened 2016-12-19
docs($exceptionHandler): replace Exception with the JS standard: Error.

graingert:
JS doesn't have Exceptions.

doc fix.

docs mention 'exceptions'. Is this Java or C#?

Use Errors. This is the ECMAScript term for stack traversing failures.

Nope

**Other information**:

### DIFF
--- a/src/ng/exceptionHandler.js
+++ b/src/ng/exceptionHandler.js
@@ -7,7 +7,7 @@
  * @this
  *
  * @description
- * Any uncaught exception in angular expressions is delegated to this service.
+ * Any uncaught Error in angular expressions is delegated to this service.
  * The default implementation simply delegates to `$log.error` which logs it into
  * the browser console.
  *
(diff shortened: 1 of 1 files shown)


gkalpak:
`Error` is just a type of exception. I don't think it makes much of a difference, but "pedantically" speaking "exception" is more appropriate. `$exceptionHandler` does not only handle `Error` objects, but any uncaught exception.

Also, the spec seems to favor "exception" throughout. E.g. quoting [the section][1] on **Error Objects**:

> Instances of Error objects are thrown as exceptions when runtime errors occur. The Error objects may also serve as base objects for user-defined exception classes.

##
Based on that (and the fact that the term "exception" is more consistent with the name of the "$exceptionHandler" service itself), I would say the current wording is preferable :smiley:

[1]: http://www.ecma-international.org/ecma-262/5.1/#sec-15.11

PR #15537, opened 2016-12-21
fix(sce/urlUtils): Use baseURI for same-origin checks, and bump base#href to RESOURCE_URL 

rjamet:
When a base tag is set in the page, relative links will be resolved to its href
attribute. However, isSameOrigin currently checks against the location, so relative
links wouldn't be considered same-origin for security purposes. This commit changes
to baseURI: it should behave the same when baseURI is left to the default value, but
that preserves the $sce's implicit trust in relative URIs when baseURI is changed (which
was #15144). 

And while I'm at it, bump base#href to RESOURCE_URL, since it affects security behavior.

Bug fix for #15144 + fix the wrong base#href context.

Base#href is just URL context, isUrlSameOrigin looks at the location for same-origin checks.

Yes, if you dynamically set base#href or rely on your old origin being trusted after changing baseURI. I don't think that represents many apps, but it's hard to tell?

======

"End of the week" turned out to be optimistic, I'm sorry about the delay :/ 

I was wary of the way baseURI is handled, which differs a lot depending on browsers. Try setting base href="data:text/", and a href="html,foo", you'll see that chrome sets the baseURI but sets the href to the empty string, while Firefox keeps the href but doesn't set the baseURI. It's all very strange and I was afraid the checks wouldn't follow the behavior of the browser, but it's a red herring in this case. Angular just shouldn't let base href be controlled by an attacker (= RESOURCE_URL).

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3256,6 +3256,9 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
       // maction[xlink:href] can source SVG.  It's not limited to <maction>.
       } else if (attrNormalizedName === 'xlinkHref' ||
           (tag === 'form' && attrNormalizedName === 'action') ||
+          // base href is just plain weird and scary. If relative URLs don't go where they are
+          // expected to, all bets are off.
+          (tag === 'base' && attrNormalizedName === 'href') ||
           // links can be stylesheets or imports, which can run script in the current origin
           (tag === 'link' && attrNormalizedName === 'href')
       ) {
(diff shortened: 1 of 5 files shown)


rjamet:
Something's wrong with IE9, I think that's an unrelated test that accepts example.com because setting the base#href to null didn't reset the baseURI in earlier tests. I'll try to fix it over the holidays.

gkalpak:
@rjamet, isn't this similar to https://github.com/angular/angular.js/pull/15145?

rjamet:
It's the same, I didn't see that... @adob's PR looks better than mine too (but he doesn't do the base change, which I should probably have split off mine). How do you prefer we move forward?

gkalpak:
If you think @adob's PR is better and since it was submitted first, we can merge that in and then you can add the base change on top of that (i.e. in a new PR). How does that sound?

gkalpak:
This has been "split" into #15145 and #15597. Closing...

PR #15538, opened 2016-12-21
fix($injector): do not use inherited $inject

### LINKED ISSUE #15536: $inject with ES6 classes can inject the wrong dependencies

Bug

When angular annotates a base class, it uses the inherited annotations when resolving the dependencies of derived classes.

http://plnkr.co/edit/55dt9Xx8z5qvHZQ9eDdQ?p=preview

Derived classes should not inherit their parent's $inject property, as they do not necessarily share dependencies.

This behavior introduces a potentially hard to debug issue, and is probably not what the user expects to happen.

This is still an issue when running against the snapshot as of 2016-12-21

ChrisMondok:
Ensure that a function has its own $inject property, and not an inherited
property, when annotating. This ensures that derived functions / classes do not
get the dependencies of their super class.

Fixes #15536

Bug fix

#15536 

Yes

**Other information**:

BREAKING CHANGE:

Prior to this PR, derived classes with no dependencies could extend base classes with dependencies, as long as the derived classes did not have constructors. As of this PR, the derived class would be seen as having no dependencies, and the base class's constructor would be called with no arguments.

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -100,7 +100,8 @@ function annotate(fn, strictDi, name) {
       last;
 
   if (typeof fn === 'function') {
-    if (!($inject = fn.$inject)) {
+    $inject = fn.hasOwnProperty('$inject') ? fn.$inject : undefined;
+    if (!$inject) {
       $inject = [];
       if (fn.length) {
         if (strictDi) {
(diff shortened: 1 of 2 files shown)


dcherman:
How would this play with derived classes that do not define their own constructor? Wouldn't that fail as it would reject the use of the parent's dependencies?

https://plnkr.co/edit/v4S9ViiJIaq9jZwFU1jO?p=preview

ChrisMondok:
I'm closing this PR: while the issue still exists, it's pretty clear that this (at least, not exactly this), is not the solution.

Issue #15533, opened 2016-12-20
angular.copy and angular.equals don't play well with MobX


thorn0:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

In theory, it should be easy to start working with [MobX](https://github.com/mobxjs/mobx). It will be fun they said. Just replace plain objects and arrays with observables and the existing code shouldn't notice much difference. Just a couple of tweaks might be needed. Unfortunately, it's not the case with Angular 1.x apps. Watchers get broken when they are passed observable arrays.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (template: http://plnkr.co/edit/tpl:yBpEi4).**

`$watch` ignores changes in observable arrays:
http://plnkr.co/edit/eNBaj5HBvvGdd6PAos8g?p=preview

`$watch` uses `angular.copy` and `angular.equals` internally, and it's these functions that don't work with MobX:

```js
var observableArray = mobx.observable([1, 2, 3]);
var copy = angular.copy(observableArray);
console.log(angular.equals(observableArray, copy)); // true
observableArray.push(100);
console.log(angular.equals(observableArray, copy)); // true!
// `copy` is completely unusable:
copy[1]; // TypeError
copy.length; // TypeError
```

**What is the expected behavior?**

Would be great to have a way to teach `angular.copy` and `angular.equals` new object types.

**What is the motivation / use case for changing the behavior?**

I decided to play with MobX in my app to see whether it can make the state management simpler. As the first step, I was going to just replace my model/state objects with observables to see what changes are needed.

**Which versions of Angular, and which browser / OS are affected by this issue? Did this work in previous versions of Angular? Please also test with the latest stable and snapshot (https://code.angularjs.org/snapshot/) versions.**

Angular 1.6, 1.5

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

This issue may sound like #10096, but it's not about specifying an alternative `copy` and `equals` implementation on every call to `$watch`. A way to override this logic globally for all the watchers, directive bindings, etc. is needed.

gkalpak:
This has nothing to do with MobX (in the sense that it is in no-way MobX-specific). So, the title is unnecessarily specific imo.

This is basically a general feature request about supporting `equals` and `copy` for custom/arbitrary objects (which in turn would enable such objects to be used in bindings/watchers).

##
[...]

PR #15541, opened 2016-12-22
feat(isArray): make angular.isArray support Array subclasses

thorn0:

feature

Angular doesn't recognise objects prototypically inherited from `Array.prototype` as arrays.
See #15533

In particular, this change gets Angular to play nicely with MobX observable arrays. 

Yes. `angular.isArray` is used throughout Angular. In particular, after this change, `angular.copy` will use the array-cloning logic for objects prototypically inherited from `Array.prototype` whereas now it uses the object-cloning logic for them. This, in turn, affects deep watchers and other parts of the framework that use `angular.copy`.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -219,8 +219,7 @@ function isArrayLike(obj) {
 
   // NodeList objects (with `item` method) and
   // other objects with suitable length characteristics are array-like
-  return isNumber(length) &&
-    (length >= 0 && ((length - 1) in obj || obj instanceof Array) || typeof obj.item === 'function');
+  return isNumber(length) && (length >= 0 && (length - 1) in obj || typeof obj.item === 'function');
 
 }
 
@@ -635,12 +634,14 @@ function isDate(value) {
  * @kind function
  *
  * @description
- * Determines if a reference is an `Array`. Alias of Array.isArray.
+ * Determines if a reference is an `Array`.
  *
  * @param {*} value Reference to check.
  * @returns {boolean} True if `value` is an `Array`.
  */
-var isArray = Array.isArray;
+function isArray(arr) {
+  return arr instanceof Array || Array.isArray(arr);
+}
 
 /**
  * @description
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -1251,6 +1251,16 @@ describe('angular', function() {
     });
   });
 
+  describe('isArray', function() {
+
+    it('should return true if passed an object prototypically inherited from Array.prototype', function() {
+      function FooArray() {}
+      FooArray.prototype = [];
+      expect(isArray(new FooArray())).toBe(true);
+    });
+
+  });
+
   describe('isArrayLike', function() {
 
     it('should return false if passed a number', function() {


thorn0:
Please consider merging this into the 1.5 branch too.

gkalpak:
> Please consider merging this into the 1.5 branch too.

1.5? I am even skeptical about 1.6. This is a BC, I'm afraid.

src/Angular.js
```diff
@@ -232,8 +232,7 @@ function isArrayLike(obj) {
 
   // NodeList objects (with `item` method) and
   // other objects with suitable length characteristics are array-like
-  return isNumber(length) &&
-    (length >= 0 && ((length - 1) in obj || obj instanceof Array) || typeof obj.item === 'function');
+  return isNumber(length) && (length >= 0 && (length - 1) in obj || typeof obj.item === 'function');
```
gkalpak: We are relaxing the condition a bit, but it is OK imo.

gkalpak (approved):
LGTM (for 1.7.x)
Could you expand the BC notice a little to explain what other things might be affected (apart from direct `angular.isArray(...)`)? E.g. watchers, `copy`ing etc.

src/Angular.js
```diff
@@ -634,12 +633,14 @@ function isDate(value) {
...
  * @description
- * Determines if a reference is an `Array`. Alias of Array.isArray.
+ * Determines if a reference is an `Array`.
  *
  * @param {*} value Reference to check.
  * @returns {boolean} True if `value` is an `Array`.
  */
-var isArray = Array.isArray;
+function isArray(arr) {
+  return arr instanceof Array || Array.isArray(arr);
```
bathos: `Array.isArray(arr)` is unreachable here, since by definition any array exotic object is `instanceof Array`.

gkalpak:
BTW, in anyone wanted to use an evil and totally not recommended hack for 1.5/1.6, they could surround the `angular.js` script with code that modified `Array.isArray`, let Angular store it locally, then restore the original method ([bad bad demo][1]).

[1]: https://codepen.io/ExpertSystem/pen/gLyKNV?editors=1012

src/Angular.js
gkalpak: Not if it comes from a different context (e.g. another Window object). E.g.:

```js
var win = window.open();
var arr = new win.Array();

console.log(arr instanceof Array);   // false
console.log(Array.isArray(arr));   // true
```

src/Angular.js
thorn0: `isArray` is used in a couple of lines above. So if `obj instanceof Array`, the function hits return on that line.

thorn0:
> Could you expand the BC notice

Here in PR? Or in the commit message?

> BTW, in anyone wanted to use an evil and totally not recommended hack

I'm just patching angular.js locally :)

src/Angular.js
gkalpak: I know. But previous there was also the restriction that `isNumber(length)` and `length >= 0`, which is gone now.

What I am saying is that `isArrayLike` (and anything depending on it) is slightly affected by the change in `isArray`.

gkalpak:
> I'm just patching angular.js locally :)

I know :wink:

src/Angular.js
bathos: good call

Narretz:
@thorn0 the BC change message still needs to be updated as @gkalpak has requested for this to land in 1.7.0

thorn0:
@Narretz done

Narretz:
Cool! Can you please add that this change also affects `angular.merge`, `angular.forEach`, and `angular.equals`?

thorn0:
@Narretz done. Travis' error messages are irrelevant.

Narretz:
LGTM. Is that okay for you @gkalpak ?

gkalpak:
I would like to see a more explicit mention of how the implementation changed/what _"objects that weren't previously recognized as arrays [...] are now recognized as such"_. Just a brief notice that objects that prototypally inherit from `Array` are now recognized as arrays.

gkalpak:
Tweaked the commit message, added some more tests and landed.
Thx for working on this, @thorn0 👍

PR #15542, opened 2016-12-22
Fix issue: 15536

### LINKED ISSUE #15536: $inject with ES6 classes can inject the wrong dependencies

mhevery:
This PR only demonstrates a fix, it still needs for someone to add tests.

Closes #15536

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -100,7 +100,7 @@ function annotate(fn, strictDi, name) {
       last;
 
   if (typeof fn === 'function') {
-    if (!($inject = fn.$inject)) {
+    if (!($inject = fn.hasOwnProperty('$inject'))) {
       $inject = [];
       if (fn.length) {
         if (strictDi) {


src/auto/injector.js
```diff
@@ -100,7 +100,7 @@ function annotate(fn, strictDi, name) {
       last;
 
   if (typeof fn === 'function') {
-    if (!($inject = fn.$inject)) {
+    if (!($inject = fn.hasOwnProperty('$inject'))) {
```
gkalpak: I assume you meant something like https://github.com/angular/angular.js/pull/15538/files#diff-ced481bc839279a343e9f8bb8dcf6e98R103.

gkalpak:
Duplicate of #15538 (afaict).
<sub>(In case I missed the point please reopen.)</sub>

PR #15555, opened 2016-12-29
fix($compile): support transcluding multi-element directives

### LINKED ISSUE #15554: fix($compile): support transcluding multi-element directives

Bug.

Transcluding multi-element directives (e.g. `ng-repeat-start`/`ng-repeat-end`) is not supported on elements with multi-slot transclusion and a `$compile:uterdir` error is thrown.
Originally reported by @jamie-pate in https://github.com/angular/angular.js/issues/14653#issuecomment-269564856.

[Reproduction][1].

Multi-element directives should be transcluded correctly on elements with multi-slot transclusion and without throwing.

Affects all Angular versions since a least 1.5.0 (including 1.6.x/master).

[1]: https://plnkr.co/edit/bMcETDXdd0EIpNiZG3wq?p=preview

gkalpak:
Bug fix.

Transcluding multi-element directives (e.g. `foo-start`/`foo-end`) is not supported on elements with multi-slot transclusion (a `uterdir` error is thrown).

Transcluded nodes are put into a DocumentFragment, where they can be traversed via `.nextSibling`. This way, transcluding multi-element directives on elements with multi-slot transclusion works correctly.

No.

**Other information**:
Fixes #15554.

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -2575,7 +2575,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
               // We have transclusion slots,
               // collect them up, compile them and store their transclusion functions
-              $template = [];
+              $template = window.document.createDocumentFragment();
 
               var slotMap = createMap();
               var filledSlots = createMap();
@@ -2603,10 +2603,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
                 var slotName = slotMap[directiveNormalize(nodeName_(node))];
                 if (slotName) {
                   filledSlots[slotName] = true;
-                  slots[slotName] = slots[slotName] || [];
-                  slots[slotName].push(node);
+                  slots[slotName] = slots[slotName] || window.document.createDocumentFragment();
+                  slots[slotName].appendChild(node);
                 } else {
-                  $template.push(node);
+                  $template.appendChild(node);
                 }
               });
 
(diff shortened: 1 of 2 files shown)


src/ng/compile.js
```diff
@@ -2480,7 +2480,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
 
               // We have transclusion slots,
               // collect them up, compile them and store their transclusion functions
-              $template = [];
+              $template = window.document.createDocumentFragment();
```
jbedard: Should this use $window or $document?

gkalpak: Other places inside this file use `window` directly (e.g. [here][1], [here][2] and [here][3]). Normally, I would prefer `$window` (or `$document`). But it doesn't bother me too much here because:

1. This file is dealing with pretty "low-level" stuff.
2. Switching to `$window` would be a breaking change (since every `$window` or `$document` mock should then provide the necessary APIs).

[1]: https://github.com/gkalpak/angular.js/blob/5a2db25b5d5d331f414894f3e13c096c8ec85779/src/ng/compile.js#L1495
[2]: https://github.com/gkalpak/angular.js/blob/5a2db25b5d5d331f414894f3e13c096c8ec85779/src/ng/compile.js#L1840
[3]: https://github.com/gkalpak/angular.js/blob/5a2db25b5d5d331f414894f3e13c096c8ec85779/src/ng/compile.js#L3389

jbedard:
@gkalpak what's the status of this? Seems like a reasonable fix...

gkalpak:
Waiting for approval :grin:

Narretz:
LGTM with one question - does this affect  performance? Creating a document fragment is more expensive than creating an array. But I assume it's okay since it only happens for slot transclusion.

gkalpak:
Yes, it olny affects directives with transclusion and yes it should be a little slower (not much though according to my [totally inadequate jsperf][1]) and I don't expect it to have any noticeable impact.

[1]: https://jsperf.com/angularjs-array-vs-documentfragment/1

Narretz:
We'll find out :D

Issue #12571, opened 2015-08-13
$location causes infinite digest


yeerkkiller1:
Steps to reproduce:

1) Use ng-include
2) Run `window.history.pushState('', '', '?/');` inside the link function of a directive
3) Use that directive inside an ng-repeat
4) Set $locationProvider to use html5 mode

The issue can be seen in a fairly simple form here: http://jsbin.com/kapedo/edit?html,console

This is a massive issue, I am not trying to use $location at all. I only set it to html5 mode so it doesn't change my url.

This has the same consequence as https://github.com/angular/angular.js/issues/1417, but the source is probably different. $location seems to have a lot of bugs related to its storing of the url and assuming no one else is going to change it. Using $location is not an option, window.history.pushState is being called by a library I have no control over.

At the very least an opt-out for $location should exist, as it is already in AngularJS and any bugs in it are going to mess up the url, even for people who are not using $location.

PR #15561, opened 2016-12-30
fix($location): correctly handle external URL change during `$digest`

### LINKED ISSUE #11075: Bug: Changing the location outside of $location causes infdig and/or resets the original url
Hi,

As said in the title, if the location is changed from outside the $location service (but inside a digest cycle, and especially _inside a $q.then callback_), then Angular will break and there is two possible outcome, apparently random (maybe some kind of race condition?) :
- Either Angular break with an infinite digest loop (and an empty watcher list)
- Or Angular will reset the URL to the old one by going into [this code branch](https://github.com/angular/angular.js/blob/master/src/ng/location.js#L946)

Here is a minimal showcase to reproduce this issue :

``` html
<script src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.3.13/angular.js"></script>

<script>
    angular.module( 'testApp', [ ] ).controller( 'testController', function ( $location, $q, $scope, $window ) {
        $scope.triggerThing = function ( ) {
            $q.when( ).then( function ( ) {
                $window.history.pushState( { }, null, '#/testfoo' );
                $scope.$broadcast( 'test' );
            } );
        };
    } );
</script>

<div ng-app="testApp">
    <div ng-controller="testController">
        <a ng-click="triggerThing()">Trigger</a>
    </div>
</div>
```

### LINKED ISSUE #15556: Internet Explorer infinite digest with location change
Bug

**Current behavior**

http://plnkr.co/edit/7SldRRsrYFO32hiG9abV?p=preview
Infinite digest error in Internet Explorer console after button click
**[$rootScope:infdig] 10 $digest() iterations reached. Aborting!
Watchers fired in the last 5 iterations: []**

**Expected behavior**

Location changes without error

**Minimal reproduction of the problem with instructions**
http://plnkr.co/edit/7SldRRsrYFO32hiG9abV?p=preview with Internet Explorer
Check error console after button click.
Note: this happens when the location change occurs during an $evalAsync 

We have a large app with different redirect scenarios that may or may not be local to our app.  We wish to set location.href.  

* **Angular version:**  1.5/1.6

* **Browser:** [ IE 11 ] 

In internet explorer the hashchange event fires asynchronously.  this causes the $locationWatch to think that the $location url was updated as it doesn't match $browser url.   It seems like the location logic should know whether a url change occured via $location setters

**Relevant Code:
https://github.com/angular/angular.js/blob/master/src/ng/location.js#L970

gkalpak:
Bug fix(es).

1. `$browser.$$checkUrlChange()` (which was run before each `$digest`) will only detect an external change (i.e. not via `$location`) to the browser URL. External changes to `history.state` will not be detected and propagated to `$location`.

   This would not be a problem if changes were followed by a `popstate` or `hashchange` event (which would call `cacheStateAndFireUrlChange()`). But since `history.pushState()/replaceState()` do not fire any events, calling these methods manually will result in `$location` getting out-of-sync with the actual
history state.

   This is not detected in tests, because the mocked `window.history` will incorrectly trigger `popstate` when calling `pushState()/replaceState()`, which "covers" the bug.

2. When the URL is changed directly (e.g. via `location.href`) during a `$digest` (e.g. via `scope.$evalAsync()` or `promise.then()`) the change is not handled correctly, unless a `popstate` or `hashchange` event is fired synchronously.

   This is an issue when calling `history.pushState()/replaceState()` in all browsers, since these methods do not emit any event. This is also an issue when setting `location.href` in IE11, where (unlike other browsers) no `popstate` event is fired at all for hash-only changes ([known bug][1]) and the `hashchange` event is fired asynchronously (which is too late).

1. The first bug is fixed by ensuring `cacheState()` is always called, before looking for and propagating a URL/state change.

2. The second bug(s) is fixed by:
   1. Keeping track of `$location` setter methods being called and only processing a URL change if it originated from such a call. If there is a URL difference but no setter method has been called, this means that the browser URL/history has been updated directly and the change hasn't yet been propagated to `$location` (e.g. due to no event being fired synchronously or at all).
   2. Checking for URL/state changes at the end of the `$digest`, in order to detect changes via `history` methods (that took place during the `$digest`).

<sub>Who k</sub>**NO**<sub>ws?</sub>
(This PR fixes some uncommon usecases, but I am not 100% sure it was not possible to somehow "exploit" the bugs to support a usecase that is now broken.)

**Other information**:
Fixes #11075, #12571 and #15556.

[1]: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/3740423/

### DIFF
--- a/src/ng/browser.js
+++ b/src/ng/browser.js
@@ -96,7 +96,6 @@ function Browser(window, document, $log, $sniffer) {
       };
 
   cacheState();
-  lastHistoryState = cachedState;
 
   /**
    * @name $browser#url
@@ -150,8 +149,6 @@ function Browser(window, document, $log, $sniffer) {
       if ($sniffer.history && (!sameBase || !sameState)) {
         history[replace ? 'replaceState' : 'pushState'](state, '', url);
         cacheState();
-        // Do the assignment again so that those two variables are referentially identical.
-        lastHistoryState = cachedState;
       } else {
         if (!sameBase) {
           pendingLocation = url;
@@ -200,8 +197,7 @@ function Browser(window, document, $log, $sniffer) {
 
   function cacheStateAndFireUrlChange() {
     pendingLocation = null;
-    cacheState();
-    fireUrlChange();
+    fireStateOrUrlChange();
   }
 
   // This variable should be used *only* inside the cacheState function.
@@ -215,11 +211,16 @@ function Browser(window, document, $log, $sniffer) {
     if (equals(cachedState, lastCachedState)) {
       cachedState = lastCachedState;
     }
+
     lastCachedState = cachedState;
+    lastHistoryState = cachedState;
   }
 
-  function fireUrlChange() {
-    if (lastBrowserUrl === self.url() && lastHistoryState === cachedState) {
+  function fireStateOrUrlChange() {
+    var prevLastHistoryState = lastHistoryState;
+    cacheState();
+
+    if (lastBrowserUrl === self.url() && prevLastHistoryState === cachedState) {
       return;
     }
 
@@ -285,7 +286,7 @@ function Browser(window, document, $log, $sniffer) {
    * Needs to be exported to be able to check for changes that have been done in sync,
    * as hashchange/popstate events fire in async.
    */
-  self.$$checkUrlChange = fireUrlChange;
+  self.$$checkUrlChange = fireStateOrUrlChange;
 
   //////////////////////////////////////////////////////////////
   // Misc API
--- a/src/ng/location.js
+++ b/src/ng/location.js
@@ -137,6 +137,8 @@ function LocationHtml5Url(appBase, appBaseNoFile, basePrefix) {
 
     this.$$url = encodePath(this.$$path) + (search ? '?' + search : '') + hash;
     this.$$absUrl = appBaseNoFile + this.$$url.substr(1); // first char is always '/'
+
+    this.$$urlUpdatedByLocation = true;
   };
 
   this.$$parseLinkUrl = function(url, relHref) {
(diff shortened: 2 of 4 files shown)


petebacondarwin (approved):
This sounds like a reasonable change.
The fact that the tests were artificially hiding the real behaviour was worrying (probably my fault).
Are we concerned that we now need a digest for the location to be updated - or was it the case that this was just broken completely previously, so waiting for a digest is a step forward?

gkalpak:
Wow, that was quick :smiley:

> Are we concerned that we now need a digest for the location to be updated - or was it the case that this was just broken completely previously, so waiting for a digest is a step forward?

The latter. `$location` will still be updated at the same point in time for cases that worked previously.
Previously, there was a "family" of external changes that would either go unnoticed (e.g. changes to `history.state` only) or cause an infinite digest error (e.g. changes to `location.href` in IE11 or `history.pushState(...)` in all browsers). Now, such changes will not cause infinite digest errors and will get noticed (but it is possible that they will immediately trigger another `$digest`).

I am a (tiny) little concerned that people might somehow rely on `history.state` changes being unnoticed and using that to support their obscure usecases :smiley:

petebacondarwin:
I don't that is a breaking change. It was not documented to do that and it is very reasonable that the previous functionality is wrong.

gkalpak:
I've merged this into master and v1.6.x. I thought it might be too much for v1.5.x (considering it is a non-trivial change/fix), but could be convinced otherwise :smiley:

dlongley:
@gkalpak,

> I am a (tiny) little concerned that people might somehow rely on history.state changes being unnoticed and using that to support their obscure usecases...

You were right! :)

I was relying on this behavior to ensure that I could hit the `back` button after opening modals that should not trigger route changes. I was previously adding a new history state and checking for its presence to determine whether or not the back button should close a modal or actually go back in history. It worked great in Angular 1.5.

With this patch now (in Angular 1.6) route changes are triggered after you push the custom history state -- and if you try and cancel them via `preventDefault`, the `$browser` code mangles the history. I'm hoping to discover a workaround to restore the functionality.

petebacondarwin:
@dlongley - sorry about this. If you don't find a workaround please post again and we can see what we can do to fix it.

PR #15571, opened 2017-01-03
fix(angularInit): allow auto-bootstraping from inline script

### LINKED ISSUE #15567: Disabling automatic bootstrap. <script> protocol indicates an extension, document.location.href does not match

Bug

On certain browsers (Tested: Safari 9, PhantomJS 2.1), Angular refuses to bootstrap. It generates the error: `Angular: Disabling automatic bootstrap. <script> protocol indicates an extension, document.location.href does not match.`

I find it very hard to debug this issue and reproduce it. What might help: we're using `Webpack` to compile everything down to a single <script> tag (WITHOUT `src=` or anything else)

I expect angular to auto-bootstrap.

It's very weird that different browsers generate different results. We're already in production and only now we're getting reports of people being unable to open our website.

WORKS in macOS Sierra with Safari 10.x, Firefox, Chrome.
BROKEN in macOS with Safari 9.x, PhantomJS, SINCE 1.5.9

I've traced the code to https://github.com/angular/angular.js/blob/master/src/Angular.js#L1509 .

Maybe I need to specify a src= tag, or something?

I'm aware that this report may miss some critical info. Please let me know what other information you need!

gkalpak:
Bug fix.

Some browsers (e.g. Safari 9.x, PhantomJS) do not set `link.origin/protocol` correctly, when setting `link.href` to `null`, which prevents auto-bootstraping Angular from scripts without a `src` attribute (i.e. inline scripts).
See also #15567.

Auto-bootstraping Angular from inline scripts is allowed.

No.

**Other information**:
Inline scripts are on the same origin as the loading page, so auto-bootstraping
should be allowed.

Fixes #15567.

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1479,12 +1479,16 @@ function getNgAttribute(element, ngAttr) {
 }
 
 function allowAutoBootstrap(document) {
-  if (!document.currentScript) {
+  var script = document.currentScript;
+  var src = script && script.getAttribute('src');
+
+  if (!src) {
     return true;
   }
-  var src = document.currentScript.getAttribute('src');
+
   var link = document.createElement('a');
   link.href = src;
+
   if (document.location.origin === link.origin) {
     // Same-origin resources are always allowed, even for non-whitelisted schemes.
     return true;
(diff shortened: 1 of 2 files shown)


gkalpak:
cc @mprobst

mprobst (approved):

petebacondarwin (approved):

PR #15584, opened 2017-01-05
feat(ngOptions): add $value variable for easier use of trackBy+selectAs

icholy:

Feature.

Currently it's a pain to use `trackBy` and `selectAs` together.

It introduces a `$value` variable which simplifies how the `trackBy` expression can be written.

No.

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -108,7 +108,8 @@ var ngOptionsMinErr = minErr('ngOptions');
  * expression evaluates to `items[0].subItem.id` (which is undefined). As a result, the model value
  * is not matched against any `<option>` and the `<select>` appears as having no selected value.
  *
- * here is the fixed version of the broken example above.
+ * The solution is to use `$value` variable which provides uniform access to each `item` and
+ * `ngModel` value.  Here is the fixed version of the broken example above.
  *
  * ```html
  * <select ng-options="item.subItem as item.label for item in items track by $value.id" ng-model="selected"></select>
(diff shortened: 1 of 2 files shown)


Narretz:
So this is basically a shortcut for `(val.id || val.subItem.id)`? From your changes to the docs, that's not entirely clear, so I would like a better description of this feature.
It's also not enough to change the existing tests, there should be new tests that make explicit what this feature does.

icholy:
> So this is basically a shortcut for (val.id || val.subItem.id)?

In the simple case, yes.

> But in which cases exactly do you need this?

Here's a contrived example (note the id's don't match up).

```js
$scope.items = [{
  id: 0,
  label: 'aLabel',
  subItem: { name: 'aSubItem', id: 1 }
}, {
  id: 1,
  label: 'bLabel',
  subItem: { name: 'bSubItem', id: 0 }
}];

$scope.selected = $scope.items[0].subItem;
```

If you use `(val.id || val.subItem.id)` here, it will select the wrong object.

demo: https://plnkr.co/edit/8MQXJpRqEdtwF1hcCM3w?p=preview

icholy:
@Narretz I noticed you put this in the Ice Box milestone and I don't want to invest more time into this if it's not going anywhere. I think that I presented a valid use case in my last comment, but if that's not the case, let me know so I don't waste my time.

icholy:
@Narretz I've add a bit more to the documentation, but I'm not sure about the tests. Here's what it says above the tests I changed.

```
  /**
   * This behavior is broken and should probably be cleaned up later as track by and select as
   * aren't compatible.
   */
```

This is what the `$value` variable aims to fix. So it doesn't make sense to still have these tests using the hacky workaround.

PR #15597, opened 2017-01-11
fix($compile) : Add base#href to the list of RESOURCE_URL context attrs

rjamet:
This will put the same security constraints to bindings on base#ref as on iframes or script srcs, since it changes the behavior of relative URLs across the page. Currently, they don't have a $sce context. Also, it's more generally a good idea in Angular itself, as since #15144 and its fix #15145 we'll consider that baseURI as a trusted origin.

Yup: something like `<base href="{{myUrl}}"/>` will send myUrl to the $sce's RESOURCE_URL checks. By default, it will break if myUrl isn't same-origin. Also, concatenation in trusted contexts is disabled: "/{{myPathComponent}}/something" won't work at all.

This follows the discussion in #15537 and the initial issue in #15144, fixed in #15145.


gkalpak (approved):
LGTM, but the commit message needs a proper breaking change notice.

gkalpak:
@rjamet, what is your take on https://github.com/angular/angular.js/pull/15145#issuecomment-272115379? Is it OK to release #15145 without this change, or should they rather go together (or not at all)?

petebacondarwin:
For anyone wondering this question was answered in https://github.com/angular/angular.js/pull/15145#issuecomment-273113931

> both PRs at the same time would be the safest.

petebacondarwin (approved):
LGTM

PR #15627, opened 2017-01-19
fix (ngResource) dynamic params escaping fix

vteremasov:
Bug fix

It doesn't work correctly according to [docs](https://docs.angularjs.org/api/ngResource/service/$resource) if url params are dynamic. So, for instance, this is gonna work: 

```javascript
// In SomeResourceName factory:
$resouce('/path/\\.json',  ...).save({});
```

but almost the same code works incorrectly: 

```javascript
// In SomeResourceName factory:
$resouce('/path/:json',  {json: '\\.json'}, ...).save({});
```
it sends request to ` :path/%5C.json`

Might be

**Other information**:

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -593,7 +593,7 @@ angular.module('ngResource', ['ng']).
           // then replace collapse `/.` if found in the last URL path segment before the query
           // E.g. `http://url.com/id./format?q=x` becomes `http://url.com/id.format?q=x`
           url = url.replace(/\/\.(?=\w+($|\?))/, '.');
-          // replace escaped `/\.` with `/.`
+          // replace escaped `/\.` with `/.` (`%5C` is url encoded backslash `\`)
           config.url = protocolAndIpv6 + url.replace(/\/(\\|%5C)\./, '/.');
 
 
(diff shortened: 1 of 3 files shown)


src/ngResource/resource.js
```diff
@@ -594,7 +594,7 @@ angular.module('ngResource', ['ng']).
           // E.g. `http://url.com/id./format?q=x` becomes `http://url.com/id.format?q=x`
           url = url.replace(/\/\.(?=\w+($|\?))/, '.');
           // replace escaped `/\.` with `/.`
-          config.url = protocolAndIpv6 + url.replace(/\/\\\./, '/.');
+          config.url = protocolAndIpv6 + url.replace(/\/(\\|%5C)\./, '/.');
```
gkalpak: Please, add a comment about how `%5C` came to be.

gkalpak (commented):
Implementation-wise LGTM.

##
I am a little concerned about whether this is a breaking change, though. The docs are not clear on whether `\.` in param values is expected to be escaped and there is the possibility that people rely on the current behavior.

src/ngResource/resource.js
vteremasov: Done

gkalpak:
Regarding the breaking change, thinking about it again the only usecase that would fail is the following (which is very unlikely):

```js
$resource('/foo/:bar', {bar: '\\.baz'})
// Request URL before: /foo/%5C.baz
// Request URL after:  /foo/.baz
```

And considering that `$resource('/foo/:bar', {bar: '.baz'})` would collapse the URL (i.e. `/foo.baz`), it is an indication that the `/.` or `/\.` rules apply to parameters values as well.

So, this is a fix 🎉

gkalpak (approved):

PR #15632, opened 2017-01-20
fix(select): keep ngModel when selected option is recreated with ngRe…

### LINKED ISSUE #15630: Angular 1.6 incorrectly resets `<select>'s `ng-model` value

Bug

Angular 1.6 resets `<select>`s `ng-model` value when `<option>`s list changes (see "Other information" section).
Angular 1.5 preserves it.

Angular 1.6: https://plnkr.co/edit/wHc4rzy7x9PAEqQLfVQT?p=preview
Angular 1.5.11: https://plnkr.co/edit/e0LVMEtGP2j4JxNf8lsp?p=preview

`ng-model` preserves it's value.

It's not correct.

Angular 1.6.x is affected.
Angular 1.5.x is not.

The thing is Angular 1.6 tracks options by it's model and not by the `options`s value.

In the snippets above options array contains 2 objects: `[{val: "1"}, {val: "2"}]`.
`ng-repeat` iterates through them and uses `val` property as both option value and title.
`vm.selected` is used as select's `ng-model` value and is initially set to "2".

After pressing `Change options` button I replace options array with this one: `[{val: "2"}, {val: "3"}]` and Angular 1.6 resets `vm.selected` value **but it shouldn't** because option with value `2` is still in the list.

Angular 1.5 behaves correctly in this case and `vm.selected` value is preserved.

If `vm.options` contains primitive values bug doesn't appear ([snippet](https://plnkr.co/edit/tSd6MGChGB4s0eX1UIVS?p=preview)).

Narretz:
…peat

Fixes #15630

### DIFF
--- a/test/ng/directive/selectSpec.js
+++ b/test/ng/directive/selectSpec.js
@@ -2339,11 +2339,15 @@ describe('select', function() {
             scope.options = [{ name: 'B'}, { name: 'C'}, { name: 'D'}];
           });
 
+          var previouslySelectedOptionElement = optionElements[1];
           optionElements = element.find('option');
           expect(optionElements.length).toEqual(3);
+
           expect(optionElements[0].value).toBe('B');
           expect(optionElements[0]).toBeMarkedAsSelected();
           expect(scope.obj.value).toBe('B');
+          // Ensure the assumption that the element is re-created is true
+          expect(previouslySelectedOptionElement).not.toBe(optionElements[0]);
       });
 
     });
--- a/test/ng/directive/selectSpec.js
+++ b/test/ng/directive/selectSpec.js
@@ -2316,6 +2316,34 @@ describe('select', function() {
 
       });
 
+      it('should keep the ngModel value when the selected option is recreated by ngRepeat', function() {
+
+          scope.options = ['A', 'B', 'C'];
+          scope.obj = {
+            value: 'B'
+          };
+
+          compile(
+            '<select ng-model="obj.value">' +
+              '<option ng-repeat="option in options" value="{{option}}">{{option}}</option>' +
+            '</select>'
+          );
+
+          var optionElements = element.find('option');
+          expect(optionElements.length).toEqual(3);
+          expect(optionElements[0].value).toBe('A');
+          expect(optionElements[1]).toBeMarkedAsSelected();
+
+          scope.$apply(function() {
+            scope.options = ['B', 'C', 'D'];
+          });
+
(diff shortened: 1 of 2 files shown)


Narretz:
Looks like my test passes without the patch 😨 Let's see

Narretz:
Test has been fixed

test/ng/directive/selectSpec.js
```diff
@@ -2316,6 +2316,36 @@ describe('select', function() {
...
+          );
+
+          var optionElements = element.find('option');
+          expect(optionElements.length).toEqual(3);
+          expect(optionElements[0].value).toBe('A');
+          expect(optionElements[1]).toBeMarkedAsSelected();
+          expect(scope.obj.value).toBe('B');
+
+          scope.$apply(function() {
+            // Only when new objects are used, ngRepeat re-creates the element from scratch
```
gkalpak: Just to make it more explicit, I would add an expactation that `optionElements[1] (before) !== optionElements[0] (after)`.

gkalpak (approved):
LGTM (with one optional suggestion).

Narretz:
Updated, PTAL

test/ng/directive/selectSpec.js
```diff
@@ -2316,6 +2316,40 @@ describe('select', function() {
...
+            scope.options = [{ name: 'B'}, { name: 'C'}, { name: 'D'}];
+          });
+
+          var previouslySelectedOptionElement = optionElements[1];
+          optionElements = element.find('option');
+          expect(optionElements.length).toEqual(3);
+
```
gkalpak: Misplaced newline?

gkalpak (approved):

odedniv:
This is a really bad breakage, why isn't this in 1.6 already? Is it waiting for something I can help with?

Narretz:
Hi @odedniv this is in 1.6 as https://github.com/angular/angular.js/commit/131af8272d269a541d04cb522c264a91e0ec8b6a, released with 1.6.2
Not sure why you think it's not in 1.6 yet

odedniv:
@Narretz right, wasn't looking properly, but I'm still having issues with `<select ng-model>` resetting value that I don't have in 1.5. Any ideas?

Narretz:
@odedniv I don't know of any other issues and since my crystal ball is still in repair, I can't say what might be going on with your selects. Please open a new issue with a minimal reproduction of the problem. Thanks!

PR #15658, opened 2017-01-31
docs(guide/Components): add replace option

pmcelhaney:
Adds "replace" in the comparison to directives. The replace option is deprecated, but it still exists in directives, so it's worth pointing out as a difference between directives and components.

Docs

n/a

n/a

No

- [n/a] Tests for the changes have been added (for bug fixes / features)

**Other information**:

I wanted to link to this question on Stack Overflow, but I'm not sure whether to use Markdown or some other syntax. https://stackoverflow.com/questions/24194972/why-is-replace-deprecated-in-angularjs

### DIFF
--- a/docs/content/guide/component.ngdoc
+++ b/docs/content/guide/component.ngdoc
@@ -78,6 +78,7 @@ It's also possible to add components via {@link $compileProvider#component} in a
 | link functions    | Yes                  | No       |
 | multiElement      | Yes                  | No       |
 | priority          | Yes                  | No       |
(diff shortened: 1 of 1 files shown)


gkalpak:
This was a deliberate omission (to discourage people from using `replace: true` even on directives). But maybe it is more confusing to omit it altogether and may lead people making incorrect assumptions. Since we haven't removed it from the `$compile` docs or other places, I think adding it here clears things up.

Can you sign the CLA?

pmcelhaney:
Thanks, I signed the SLA.

PR #15674, opened 2017-02-03
feat($resource): add support for `request` and `requestError` interceptors

gkalpak:
Just a rebased version of #13273.

### DIFF
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -187,8 +187,9 @@ function shallowClearAndCopy(src, dst) {
  *     [requestType](https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest#responseType).
  *   - **`interceptor`** - `{Object=}` - The interceptor object has four optional methods -
  *     `request`, `requestError`, `response`, and `responseError`. See
- *     {@link ng.$http $http interceptors} for details. In addition, the
- *     resource instance or array object is accessible by the `resource` property of the
+ *     {@link ng.$http $http interceptors} for details. Note that `request`/`requestError`
+ *     interceptors are applied before calling `$http`, thus before any global `$http` interceptors.
+ *     The resource instance or array object is accessible by the `resource` property of the
  *     `http response` object passed to response interceptors.
  *     Keep in mind that the associated promise will be resolved with the value returned by the
  *     response interceptor, if one is specified. The default response interceptor returns
--- a/src/ngResource/resource.js
+++ b/src/ngResource/resource.js
@@ -185,11 +185,11 @@ function shallowClearAndCopy(src, dst) {
  *     for more information.
  *   - **`responseType`** - `{string}` - see
  *     [requestType](https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest#responseType).
- *   - **`interceptor`** - `{Object=}` - The interceptor object has two optional methods -
- *     `response` and `responseError`. Both `response` and `responseError` interceptors get called
- *     with `http response` object. See {@link ng.$http $http interceptors}. In addition, the
+ *   - **`interceptor`** - `{Object=}` - The interceptor object has four optional methods -
+ *     `request`, `requestError`, `response`, and `responseError`. See
+ *     {@link ng.$http $http interceptors} for details. In addition, the
  *     resource instance or array object is accessible by the `resource` property of the
- *     `http response` object.
+ *     `http response` object passed to response interceptors.
  *     Keep in mind that the associated promise will be resolved with the value returned by the
  *     response interceptor, if one is specified. The default response interceptor returns
  *     `response.resource` (i.e. the resource instance or array).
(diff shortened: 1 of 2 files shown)


test/ngResource/resourceSpec.js
```diff
@@ -1129,6 +1130,187 @@ describe('basic usage', function() {
...
+
+        $httpBackend.expect('GET', '/DebitCard').respond([{id: 1}]);
+
+        var resource = CreditCard.query();
+        resource.$promise.then(successSpy, failureSpy);
+
+        $httpBackend.flush();
+        expect(successSpy).toHaveBeenCalledOnce();
+        expect(failureSpy).not.toHaveBeenCalled();
+      });
```
petebacondarwin: Do you need a `$httpBackend.verifyNoOutstandingRequest();` here to check that the request is actually made or will the `expect(...)` cause an error if the spec completes without this being requested?

src/ngResource/resource.js
```diff
@@ -743,7 +746,14 @@ angular.module('ngResource', ['ng']).
               extend({}, extractParams(data, action.params || {}), params),
               action.url);
 
-            var promise = $http(httpConfig).then(function(response) {
+            // Start the promise chain
+            var promise = $q.
+              resolve(httpConfig).
+              then(requestInterceptor).
+              catch(requestErrorInterceptor).
+              then($http);
```
petebacondarwin: Interested to know why you "reimplemented" request interceptors here rather than modifying the `httpConfig` to add the interceptors to the request `httpConfig` (or a clone of it) and rely upon `$http` to deal with the intercepting.

petebacondarwin (changes requested):
One question and one minor suggestion.

src/ngResource/resource.js
gkalpak: Currently `$http` does not support per-request interceptors. Are you suggesting add that as a new feature to `$http` (and then use it to implement `$resource` request interceptors?

test/ngResource/resourceSpec.js
gkalpak: There is a `verifyNoOutstandingExpectation()` [after each test][1] to verify that the request was made.
But I think you are right, we could be more specific in our expectations (e.g. verify call args).

[1]: https://github.com/gkalpak/angular.js/blob/0416afe05aa55980a36c300554bc59fbae30d0ab/test/ngResource/resourceSpec.js#L39

gkalpak:
PTAL.

src/ngResource/resource.js
Narretz: This would actually be my preferred solution, as the resource interceptors do not work 100% like httpInterceptors anyway (see https://github.com/angular/angular.js/issues/11409. Note that even though #13273 is linked in this issue as a solution, adding request/Error interceptors does not actually solve the issue); and because http transforms do not support changing all variables (headers cannot be changed, see https://github.com/angular/angular.js/issues/12095)
So you would be able to set `interceptors` in the http config and `httpInterceptors` in ngResource.

Obviously this would be a much bigger change which would make http much more complex.

gkalpak: > This would actually be my preferred solution, as the resource interceptors do not work 100% like httpInterceptors anyway

`$resource` response interceptors are special, so indeed do not work exactly as `$http` interceptors. But request interceptors work pretty much the same afaict.

> So you would be able to set interceptors in the http config and httpInterceptors in ngResource.
> Obviously this would be a much bigger change which would make http much more complex.

I agree with both observations :wink:

Narretz: If we go with requestInterceptors (fine with me), then we should take this opportunity to document responseInterceptors of resource and the differences to http behavior, preferrably in the same or a follow up PR.

gkalpak: What differences do you have in mind? The only deifference that I know of is the extra `response.resource` property and this is already documented.

test/ngResource/resourceSpec.js
petebacondarwin: Did I ever mention that I don't like `before/afterEach` :-P

petebacondarwin (approved):
LGTM

src/ngResource/resource.js
petebacondarwin: I didn't realise that we couldn't do per-request interceptors in `$http`!

Narretz: @gkalpak I'll have to check, but since https://github.com/angular/angular.js/issues/11409 is still open I assumed the docs are not clear, or a workaround for this is not covered by the docs.

Narretz: Just saw there's an extra issue for the inconsistency in response intereceptors: https://github.com/angular/angular.js/issues/9334

test/ngResource/resourceSpec.js
gkalpak: Yup 😇

src/ngResource/resource.js
gkalpak: #9334, is about inconsistency between `response`/`responseError` interceptors in `$resource` itself (not about inconsistency against `$http` interceptors). #9334 has valid points (that are probably still worth doing something about for 1.7), but is not related to `request`/`requestError` interceptors (which are doing the "right thing"&trade; afaict).

Narretz: Yeah, it's not directly related. I just thought it could be a good time to clarify any docs in general. But it's not a must.

PR #15676, opened 2017-02-03
fix(select): keep original selection when using shift to add options …

### LINKED ISSUE #15675: not able to select multiple options on IE with ng-model
Reporting Bug: Not able select multiple options sequentially (shift+down/up arrow alone) when binded with ng-model on IE alone (tested versions IE edge, 10, 9)

Current behavior: Not allowing more than two options to be selected

JSFiddle sample code: http://jsfiddle.net/JB3Un/597/
https://plnkr.co/edit/?p=preview
This issue can be reproduced with the angular documentation sample as well for multi select.

Expected behavior: Should be able to select more than two options sequentially and should hold all the sequence for both shift+ up/down arrow

OS: windows 10
Browser: IE edge, 10, 9
AngularJS: 1.6.2

Narretz:
…in IE/Edge

In IE9-11 + Edge, the selected options were previously incorrect under the following
circumstances:
- at least two options are selected
- shift+click or shift+down/up is used to add to the selection (any number of options)

In these cases, only the last of the previously selected options and the newly selected
options would be selected.

The problems seems to be that the render engine gets confused when an option that
already has selected = true gets selected = true set again.

Note that this is not testable via unit test because it's not possible to simulate
click / keyboard events on option elements (the events are delegated to the select element
change event).

Fixes #15675

Bugfix

See #15675

No

- ~~[ ] Tests for the changes have been added (for bug fixes / features)~~

**Other information**:

### DIFF
--- a/src/ng/directive/select.js
+++ b/src/ng/directive/select.js
@@ -4,6 +4,18 @@
 
 var noopNgModelController = { $setViewValue: noop, $render: noop };
 
+function setOptionSelectedStatus(optionEl, value) {
+  optionEl.prop('selected', value); // needed for IE
+  /**
+   * When unselecting an option, setting the property to null / false should be enough
+   * However, screenreaders might react to the selected attribute instead, see
+   * https://github.com/angular/angular.js/issues/14419
+   * Note: "selected" is a boolean attr and will be removed when the "value" arg in attr() is false
+   * or null
+   */
+  optionEl.attr('selected', value);
+}
+
 /**
  * @ngdoc type
  * @name  select.SelectController
@@ -44,14 +56,14 @@ var SelectController =
     var unknownVal = self.generateUnknownOptionValue(val);
     self.unknownOption.val(unknownVal);
     $element.prepend(self.unknownOption);
-    setOptionAsSelected(self.unknownOption);
+    setOptionSelectedStatus(self.unknownOption, true);
     $element.val(unknownVal);
   };
 
   self.updateUnknownOption = function(val) {
     var unknownVal = self.generateUnknownOptionValue(val);
     self.unknownOption.val(unknownVal);
-    setOptionAsSelected(self.unknownOption);
+    setOptionSelectedStatus(self.unknownOption, true);
     $element.val(unknownVal);
   };
 
@@ -66,7 +78,7 @@ var SelectController =
   self.selectEmptyOption = function() {
     if (self.emptyOption) {
       $element.val('');
-      setOptionAsSelected(self.emptyOption);
+      setOptionSelectedStatus(self.emptyOption, true);
     }
   };
 
@@ -102,7 +114,7 @@ var SelectController =
     // Make sure to remove the selected attribute from the previously selected option
     // Otherwise, screen readers might get confused
     var currentlySelectedOption = $element[0].options[$element[0].selectedIndex];
-    if (currentlySelectedOption) currentlySelectedOption.removeAttribute('selected');
+    if (currentlySelectedOption) setOptionSelectedStatus(jqLite(currentlySelectedOption), false);
 
     if (self.hasOption(value)) {
       self.removeUnknownOption();
@@ -112,7 +124,7 @@ var SelectController =
 
       // Set selected attribute and property on selected option for screen readers
       var selectedOption = $element[0].options[$element[0].selectedIndex];
-      setOptionAsSelected(jqLite(selectedOption));
+      setOptionSelectedStatus(jqLite(selectedOption), true);
     } else {
       if (value == null && self.emptyOption) {
         self.removeUnknownOption();
@@ -292,11 +304,6 @@ var SelectController =
       }
     });
   };
-
-  function setOptionAsSelected(optionEl) {
-    optionEl.prop('selected', true); // needed for IE
-    optionEl.attr('selected', true);
-  }
 }];
 
 /**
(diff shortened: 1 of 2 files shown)


src/ng/directive/select.js
```diff
@@ -607,8 +607,17 @@ var selectDirective = function() {
...
-            option.selected = !!value && (includes(value, option.value) ||
-                                          includes(value, selectCtrl.selectValueMap[option.value]));
+            var shouldBeSelected = !!value && (includes(value, option.value) ||
+                                                includes(value, selectCtrl.selectValueMap[option.value]));
+            var currentlySelected = option.selected;
+
+            // IE and Edge will de-select selected options when you set the selected property again, e.g.
+            // when you add to the selection via shift+click/UP/DOWN
+            // Therefore, only set it if necessary
+            if ((shouldBeSelected && !currentlySelected) || (!shouldBeSelected && currentlySelected)) {
```
jbedard: `shouldBeSelected !== currentlySelected`?

Narretz: Oh, I had the feeling this could be simpler :D

src/ng/directive/select.js
```diff
@@ -607,8 +607,17 @@ var selectDirective = function() {
         // Write value now needs to set the selected property of each matching option
         selectCtrl.writeValue = function writeMultipleValue(value) {
           forEach(element.find('option'), function(option) {
-            option.selected = !!value && (includes(value, option.value) ||
-                                          includes(value, selectCtrl.selectValueMap[option.value]));
+            var shouldBeSelected = !!value && (includes(value, option.value) ||
+                                                includes(value, selectCtrl.selectValueMap[option.value]));
```
gkalpak: Nit: Align with `includes` above.

src/ng/directive/select.js
gkalpak: If you swithc to `!==` you need to make sure both values are booleans.
(I am not sure about `option.selected`. Theoretically it should always be boolean, but you never know especially with older browsers.)

gkalpak (approved):
Couldn't this be tested by directly calling `selectCtrl.writeValue`?

Narretz:
I tried to test it directly with the selectCtrl API, but this doesn't replicate the bug. It must have something to do with the ui of the select box in IE / Edge.

gkalpak:
Maybe we could just test that we don't re-set the `selected` property. (I just realized this seems equivalent to #15478, but using `<option>` instead of `ngOptions`).

Maybe something like [this][1] would work (although it borderline testing "implementation details").

[1]: https://github.com/angular/angular.js/pull/15478/files#diff-2cda0d7a6d8ec47da3560fbe59a634e3R2847

Narretz:
Hm. I agree that this is an implementation detail. I'll let you decide if I should add this test or not.

src/ng/directive/select.js
```diff
@@ -607,8 +607,24 @@ var selectDirective = function() {
...
+            // IE and Edge, adding options to the selection via shift+click/UP/DOWN,
+            // will de-select already selected options if "selected" on those options was set
+            // more than once (i.e. when the options were already selected)
+            // So we only modify the selected property if neccessary.
+            // Note: this behavior cannot be replicated via unit tests because it only shows in the
+            // actual user interface.
+            if ((shouldBeSelected && !currentlySelected)) {
+              setOptionAsSelected(jqLite(option));
+            } else if ((!shouldBeSelected && currentlySelected)) {
+              option.selected  = shouldBeSelected;
```
gkalpak: I wonder why we are using `option.selected` here, but `option.prop('selected', true)` in `setOptionAsSelected()`.
Probably doesn't matter, so we should use the same for consistency. In fact, I think it is better to make `setOptionSelected()` a helper that can both select and unselect an option and use that whenever needed.

gkalpak: (BTW, there is a double space before the `=` 😱

gkalpak (changes requested):
"No test" is fine by me.

src/ng/directive/select.js
```diff
@@ -4,9 +4,16 @@
 
 var noopNgModelController = { $setViewValue: noop, $render: noop };
 
-function setOptionAsSelected(optionEl) {
-  optionEl.prop('selected', true); // needed for IE
-  optionEl.attr('selected', true);
+function setOptionSelectedStatus(optionEl, value) {
+  optionEl.prop('selected', value); // needed for IE
+  if (value) {
+    optionEl.attr('selected', value);
```
gkalpak: FWIW: `.attr('selected', null)` also removes the attribute. So, this could be:

```js
optionEl.prop('selected', value);
optionEl.attr('selected', value || null);
```

gkalpak (approved):

src/ng/directive/select.js
```diff
@@ -4,6 +4,16 @@
...
 var noopNgModelController = { $setViewValue: noop, $render: noop };
 
+function setOptionSelectedStatus(optionEl, value) {
+  optionEl.prop('selected', value); // needed for IE
+  /**
+   * When unselecting an option, setting the property to null / false should be enough
+   * However, screenreaders might react to the selected attribute instead, see
+   * https://github.com/angular/angular.js/issues/14419
+   */
+  optionEl.attr('selected', value || null);
```
gkalpak: I said it works, but do we have tests that the property is actually removed? 😛

Narretz: Oh, you little rascal! 
We have the toBeMarkedAsSelected() custom matcher which complains when the selected attr is in the wrong state. And jqlite also removes the attribute when you pass `false` and the element is boolean. So I'm going with that.

PR #15682, opened 2017-02-06
refactor($rootScope): asyncQueue is pre-$parsed, no need to call $eval

thorn0:

minor refactoring

no

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -786,12 +786,13 @@ function $RootScopeProvider() {
           current = target;
 
           // It's safe for asyncQueuePosition to be a local variable here because this loop can't
-          // be reentered recursively. Calling $digest from a function passed to $applyAsync would
+          // be reentered recursively. Calling $digest from a function passed to $evalAsync would
           // lead to a '$digest already in progress' error.
           for (var asyncQueuePosition = 0; asyncQueuePosition < asyncQueue.length; asyncQueuePosition++) {
             try {
               asyncTask = asyncQueue[asyncQueuePosition];
-              asyncTask.scope.$eval(asyncTask.expression, asyncTask.locals);
+              fn = asyncTask.fn;
+              fn(asyncTask.scope, asyncTask.locals);
             } catch (e) {
               $exceptionHandler(e);
             }
@@ -1025,7 +1026,7 @@ function $RootScopeProvider() {
           });
         }
 
-        asyncQueue.push({scope: this, expression: $parse(expr), locals: locals});
+        asyncQueue.push({scope: this, fn: $parse(expr), locals: locals});
       },
 
       $$postDigest: function(fn) {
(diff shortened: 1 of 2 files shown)


src/ng/rootScope.js
```diff
@@ -786,12 +786,12 @@ function $RootScopeProvider() {
...
 
           // It's safe for asyncQueuePosition to be a local variable here because this loop can't
-          // be reentered recursively. Calling $digest from a function passed to $applyAsync would
+          // be reentered recursively. Calling $digest from a function passed to $evalAsync would
           // lead to a '$digest already in progress' error.
           for (var asyncQueuePosition = 0; asyncQueuePosition < asyncQueue.length; asyncQueuePosition++) {
             try {
               asyncTask = asyncQueue[asyncQueuePosition];
-              asyncTask.scope.$eval(asyncTask.expression, asyncTask.locals);
+              asyncTask.expression(asyncTask.scope, asyncTask.locals);
```
gkalpak: Won't this affect the `this` inside expression?

gkalpak (changes requested):
One little thing. LGTM otherwise.

src/ng/rootScope.js
thorn0: If you mean string expressions like `this.foo`, no, it won't affect them:
```js
angular.injector(['ng']).get('$parse')('this.foo').call({foo: 1}, {foo: 2}); // 2
```
If you mean functions passed as expressions to `$evalAsync`, it will affect them, but does this matter?

gkalpak: Yes, I mean functions. I think it does (matter). For example, having a function that incorrectly tries to access `this.someProp`, would previously throw (because `this === undefined`), but will now silently fail. 

I would rather not have the current behavior changed.

thorn0:
@gkalpak ping

gkalpak (approved):

src/ng/rootScope.js
```diff
@@ -786,12 +786,13 @@ function $RootScopeProvider() {
...
 
           // It's safe for asyncQueuePosition to be a local variable here because this loop can't
-          // be reentered recursively. Calling $digest from a function passed to $applyAsync would
+          // be reentered recursively. Calling $digest from a function passed to $evalAsync would
           // lead to a '$digest already in progress' error.
           for (var asyncQueuePosition = 0; asyncQueuePosition < asyncQueue.length; asyncQueuePosition++) {
             try {
               asyncTask = asyncQueue[asyncQueuePosition];
-              asyncTask.scope.$eval(asyncTask.expression, asyncTask.locals);
+              fn = asyncTask.fn;
```
gkalpak: As @jbedard correctly pointed out, it is a good idea to have some test about the context (or lack thereof), to guard against future "optimizations".

gkalpak (changes requested):

src/ng/rootScope.js
thorn0: done

Issue #11433, opened 2015-03-26
Testing the malformed JSON server responses with ngMock does not work anymore since 1.3.0-rc.5


kayhadrin:
# Problem definition

``` gherkin
Given I have an angular app using $http to do ajax requests
When my app makes an ajax call
And the server response is:
   - status=200
   - content-type=application/json
   - data= invalid JSON string
Then I see that the $http error handler is called
```
# Context

This issue surfaced after commit 7b6c1d08aceba6704a40302f373400aed9ed0e0b where several people already exposed their concerns about the side-effects of that change.

Here's a test case that shows how it used to work before 1.3.0-rc.5: http://codepen.io/kayhadrin/pen/Joxxrd

Log output:

```
pen.js:9 angular.version: =  1.3.0-rc.4
pen.js:46 test: good response
pen.js:52 mock response: arguments =  ["GET", "/test1", undefined, Object]
pen.js:54 mock response: ret =  [200, "{"result": "some message"}"]
pen.js:19 received(/test1): success:  {"result":"some message"}
pen.js:62 test: bad response 1: http error status
pen.js:68 mock response: arguments =  ["GET", "/test2", undefined, Object]
pen.js:74 mock response: ret =  [500, "{"error": "some message"}", Object]
pen.js:24 received(/test2): failure:  Object {data: Object, status: 500, headers: function, config: Object, statusText: ""}config: Objectdata: Objecterror: "some message"__proto__: Objectheaders: function (name) {status: 500statusText: ""__proto__: Object
pen.js:82 test: bad response 2: invalid json string
pen.js:88 mock response: arguments =  ["GET", "/test3", undefined, Object]
pen.js:94 mock response: ret =  [200, "dummy data", Object]
pen.js:21 received(/test3): failure: invalid response data:  "dummy data"
pen.js:104 $rootScope.$apply() still works
```

And this one shows how it breaks test cases after: http://codepen.io/kayhadrin/pen/RNvvGo

Log output:

```
angular.version: =  1.3.15
pen.js:46 test: good response
pen.js:52 mock response: arguments =  ["GET", "/test1", undefined, Object]
pen.js:54 mock response: ret =  [200, "{"result": "some message"}"]
pen.js:19 received(/test1): success:  {"result":"some message"}
pen.js:62 test: bad response 1: http error status
pen.js:68 mock response: arguments =  ["GET", "/test2", undefined, Object]
pen.js:74 mock response: ret =  [500, "{"error": "some message"}", Object]
[...]

PR #15689, opened 2017-02-08
test($http): ensure json deserialization errors are forwarded to erro…

Narretz:
…r handler

Since https://github.com/angular/angular.js/commit/e13eeabd7e34a78becec06cfbe72c23f2dcb85f9,
errors thrown from onFulfilled and onRejected handlers are passed to the regular http
error handlers. Before this, JSON deserialization errors lead to hard application errors, and could
not be handled by application code. This behavior was introduced in https://github.com/angular/angular.js/commit/7b6c1d08aceba6704a40302f373400aed9ed0e0b, and originally, a malformed JSON string was forwarded
as the data to the http success response handler.

This commit adds a specifc test case, even though the behavior is unlikely to break in the future without
a change in the $q rejection handling.

Related #11433

### DIFF
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1368,6 +1368,20 @@ describe('$http', function() {
               expect(callback.calls.argsFor(1)[0].data).toEqual('{"is": "not"} ["json"]');
             }
           );
+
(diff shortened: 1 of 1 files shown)


gkalpak (approved):
LGTM

Issue #15692, opened 2017-02-08
angular.copy loses own non-enumerable properties


zuzusik:
**Do you want to request a *feature* or report a *bug*?**

Report a **bug**.

**What is the current behavior?**

angular.copy [uses as destination object created from prototype](https://github.com/angular/angular.js/blob/a0641ea/src/Angular.js#L984), which it then [populates with own object properties](https://github.com/angular/angular.js/blob/a0641ea/src/Angular.js#L947). 
This way it loses own not enumerable properties which can cause bugs.

**Demo** : https://jsfiddle.net/zuzusik/7ye3m006/

**What is the expected behavior?**

Angular doesn't lose non-enumerable properties. I assume this can be achieved by iterating through `Object.getOwnPropertyNames` instead of using `for (key in source)` loop and checking for `hasOwnProperty`.

**What is the motivation / use case for changing the behavior?**

This causes bug with JSData v3: https://jsfiddle.net/zuzusik/6evanz3g/
Details: https://github.com/js-data/js-data/issues/442

**Which versions of AngularJS**

Angular 1.6.2, any browser.

Narretz:
This has been documented as a "known issue": https://github.com/angular/angular.js/commit/14519f84f54b0100fcadb334220c711cfa8242b1 and changing this would impact existing apps. See https://github.com/angular/angular.js/issues/14853 for more discussion.

PR #15693, opened 2017-02-08
fix(angular.copy): copy own non-enumerable properties

zuzusik:

Bug fix

https://github.com/angular/angular.js/issues/15692

not sure if this can be called a breaking change

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -941,6 +941,17 @@ function copy(source, destination, maxDepth) {
       for (key in source) {
         destination[key] = copyElement(source[key], maxDepth);
       }
+      // don't use Object.getOwnPropertyNames with RegExp to avoid copying RegExp lastIndex property
+    } else if (isObject(source) && typeof Object.getOwnPropertyNames === 'function' && !isRegExp(source)) {
+      Object.getOwnPropertyNames(source).forEach(function(key) {
+        var elementCopy = copyElement(source[key], maxDepth);
+
+        if (source.propertyIsEnumerable(key)) {
+          destination[key] = elementCopy;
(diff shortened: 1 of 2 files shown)


src/Angular.js
```diff
@@ -941,6 +941,11 @@ function copy(source, destination, maxDepth) {
       for (key in source) {
         destination[key] = copyElement(source[key], maxDepth);
       }
+      // don't use Object.getOwnPropertyNames with RegExp to avoid copying RegExp lastIndex property
+    } else if (source && typeof Object.getOwnPropertyNames === 'function' && !isRegExp(source)) {
```
zuzusik: Not sure if it is reasonable to do `isRegExp` check here as I assume it can affect performance.

Not doing this check leads to copying own non-enumerable `lastIndex` property of `RegExp` object to destination.

src/Angular.js
```diff
@@ -941,6 +941,11 @@ function copy(source, destination, maxDepth) {
       for (key in source) {
         destination[key] = copyElement(source[key], maxDepth);
       }
+      // don't use Object.getOwnPropertyNames with RegExp to avoid copying RegExp lastIndex property
+    } else if (isObject(source) && typeof Object.getOwnPropertyNames === 'function' && !isRegExp(source)) {
```
zuzusik: Not sure if it is reasonable to do `isRegExp` check here as I assume it can affect performance.

Not doing this check leads to copying own non-enumerable `lastIndex` property of `RegExp` object to destination.

Narretz:
See https://github.com/angular/angular.js/issues/15692#issuecomment-312871701

PR #15697, opened 2017-02-09
fix(angular.copy) support for Map and Set

mrohr:
bug fix

the copied Map or Set will fail to call functions with an error such as:
TypeError: Method Map.prototype.size called on incompatible receiver #<Map>

angular.copy will now property create new Map or Set objects with the same values as the source.

No

- [ x] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ x] Tests for the changes have been added (for bug fixes / features)
- [ x] Docs have been added / updated (for bug fixes / features)

**Other information**:

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1032,17 +1032,23 @@ function copy(source, destination, maxDepth) {
       case '[object Blob]':
         return new source.constructor([source], {type: source.type});
       case '[object Map]':
-        var copied = new source.constructor();
-        source.forEach(function(value, key){
-          copied.set(key, value);
+        // If we're in this case we know the environment supports Map
+        /* eslint-disable no-undef */
+        var copiedMap = new Map();
+        /* eslint-enable */
+        source.forEach(function(value, key) {
+          copiedMap.set(key, value);
         });
-        return copied;
+        return copiedMap;
       case '[object Set]':
-        var copied = new source.constructor();
-        source.forEach(function(value){
-          copied.add(value);
-        })
-        return copied;
+        // If we're in this case we know the environment supports Set
+        /* eslint-disable no-undef */
+        var copiedSet = new Set();
+        /* eslint-enable */
+        source.forEach(function(value) {
+          copiedSet.add(value);
+        });
+        return copiedSet;
     }
 
     if (isFunction(source.cloneNode)) {
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1031,6 +1031,9 @@ function copy(source, destination, maxDepth) {
 
       case '[object Blob]':
         return new source.constructor([source], {type: source.type});
+      case '[object Map]':
+      case '[object Set]':
+        return new source.constructor(source);
     }
 
     if (isFunction(source.cloneNode)) {
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1032,8 +1032,17 @@ function copy(source, destination, maxDepth) {
       case '[object Blob]':
         return new source.constructor([source], {type: source.type});
       case '[object Map]':
+        var copied = new source.constructor();
+        source.forEach(function(value, key){
+          copied.set(key, value);
+        });
+        return copied;
       case '[object Set]':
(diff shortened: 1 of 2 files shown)


mrohr:
signed!

mrohr:
updated email for CLA

frederikprijck:
@mrohr I did some digging in the reason why the tests was failing IE11, it looks like `Object.prototype.toString.call(source)` (https://github.com/mrohr/angular.js/blob/d5dbb5973b4588e75afee9d81a0f008f69dc033d/src/Angular.js#L997) evaluate to `[object Object]` in IE11, resulting in the usage of `Object.create(Object.getPrototypeOf(source)` to copy instead of your code. 
This throws the exception you're getting when accessing `size` on the destination object: `Map.prototype.size: 'this' is not a Map object`.

I created a fiddle to demonstrate the issue outside angularjs: https://jsfiddle.net/0mstLj6q/7/

Opening this plunkr in IE11 (Windows 10) shows:

image

src/Angular.js
```diff
@@ -1031,6 +1031,24 @@ function copy(source, destination, maxDepth) {
 
       case '[object Blob]':
         return new source.constructor([source], {type: source.type});
+      case '[object Map]':
+        // If we're in this case we know the environment supports Map
+        /* eslint-disable no-undef */
+        var copiedMap = new Map();
```
gkalpak: We access globals through the `window` in order to support non-browser environments (such as jsdom).

src/Angular.js
```diff
@@ -1031,6 +1031,24 @@ function copy(source, destination, maxDepth) {
...
+        /* eslint-enable */
+        source.forEach(function(value, key) {
```
gkalpak: Wouldn't `return new Map(source)` work as well?

gkalpak (changes requested):

src/Angular.js
frederikprijck: If I understand this correctly, IE11 does not support Map and Set constructor with arguments ? https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Map

jbedard: This also needs to copy the keys/values, not just assign them, probably by invoking `copyRecurse` so we also get the `maxDepth` behavior.

src/Angular.js
mrohr: I had originally used new source.constructor() here, tried changing it to Map/Set to deal /w the IE11 issue that is failing the build.  My next update I will change it back

mrohr:
@frederikprijck Thanks for your help, I'm guessing this feature is pretty much stuck, unless we set up the tests to not run in browsers that don't print map/set objects .toString() as '[object Map]'/'[object Set]'?

frederikprijck:
Or are there alternative approaches available to detect Map and Set in IE11 (or all browsers)?

I just did some debugging and this validates to true in IE11:

````
source.constructor === Map;
Object.getPrototypeOf(source) === Map.prototype;
```
aswell as
````
source.constructor === Set;
Object.getPrototypeOf(source) === Set.prototype;
```

gkalpak:
There are a couple of problems with this feature (which are not specific to the feature itself, rather general problems with similar features):

1. In order to be able to reliably use the feature throughout AngularJS, other internal helpers need to support it too (e.g. `equals()`). Supporting a feature in `copy()` only, is useful when using `copy()` as a general helper utility, but this is not its purpose and adds extra size and maintenance overhead.

2. It is really difficult to ensure that the feature works reliably and consistently across browsers and browser versions, because it relies on JS features that are not supported by all browsers, are fairly new additions (which means browser implementations are buggy and may change frequently) and are only partially supported by some browsers (which makes it even more difficult to rely upon or polyfill).

For these reasons, I am going to put it in the Ice Box for now (but I am open to being convinced otherwise - now or in the future :smiley:).
In any case, thx for working on this @mrohr 👍

PR #15699, opened 2017-02-10
fix($sanitize): prevent clobbered elements from freezing the browser

petebacondarwin:

bug fix

browser can be frozen with bad HTML

an exception is thrown instead

No

**Other information**:

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -18,6 +18,7 @@ var forEach;
 var isDefined;
 var lowercase;
 var noop;
+var nodeContains;
 var htmlParser;
 var htmlSanitizeWriter;
 
@@ -218,6 +219,11 @@ function $SanitizeProvider() {
   htmlParser = htmlParserImpl;
   htmlSanitizeWriter = htmlSanitizeWriterImpl;
 
+  nodeContains = window.Node.prototype.contains || /** @this */ function(arg) {
+    // eslint-disable-next-line no-bitwise
+    return !!(this.compareDocumentPosition(arg) & 16);
+  };
+
   // Regular Expressions for parsing tags and attributes
   var SURROGATE_PAIR_REGEXP = /[\uD800-\uDBFF][\uDC00-\uDFFF]/g,
     // Match everything outside of normal chars and " (quote character)
@@ -381,12 +387,12 @@ function $SanitizeProvider() {
         if (node.nodeType === 1) {
           handler.end(node.nodeName.toLowerCase());
         }
-        nextNode = node.nextSibling;
+        nextNode = getNonDescendant('nextSibling', node);
         if (!nextNode) {
           while (nextNode == null) {
-            node = node.parentNode;
+            node = getNonDescendant('parentNode', node);
             if (node === inertBodyElement) break;
-            nextNode = node.nextSibling;
+            nextNode = getNonDescendant('nextSibling', node);
             if (node.nodeType === 1) {
               handler.end(node.nodeName.toLowerCase());
             }
@@ -518,8 +524,17 @@ function $SanitizeProvider() {
         stripCustomNsAttrs(nextNode);
       }
 
-      node = node.nextSibling;
+      node = getNonDescendant('nextSibling', node);
     }
+  }
+
+  function getNonDescendant(propName, node) {
+    // An element is clobbered if its `propName` property points to one of its descendants
+    var nextNode = node[propName];
+    if (nextNode && nodeContains.call(node, nextNode)) {
+      throw $sanitizeMinErr('elclob', 'Failed to sanitize html because the element is clobbered: {0}', node.outerHTML || node.outerText);
+    }
+    return nextNode;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/ngSanitize/sanitize.js
```diff
@@ -218,6 +219,11 @@ function $SanitizeProvider() {
   htmlParser = htmlParserImpl;
   htmlSanitizeWriter = htmlSanitizeWriterImpl;
 
+  nodeContains = window.Node.prototype.contains || /** @this */ function(arg) {
+    // eslint-disable-next-line no-bitwise
+    return !!(this.compareDocumentPosition(arg) & 16);
```
koto: `compareDocumentPosition` can also be clobbered. That will throw an exception - is that OK for your case i.e. is the exception caught and safely resolved?

src/ngSanitize/sanitize.js
```diff
@@ -381,12 +387,12 @@ function $SanitizeProvider() {
...
-        nextNode = node.nextSibling;
+        nextNode = getNonDescendant('nextSibling', node);
         if (!nextNode) {
           while (nextNode == null) {
-            node = node.parentNode;
+            node = getNonDescendant('parentNode', node);
             if (node === inertBodyElement) break;
-            nextNode = node.nextSibling;
+            nextNode = getNonDescendant('nextSibling', node);
             if (node.nodeType === 1) {
```
koto: nodeType can be clobberred.

src/ngSanitize/sanitize.js
```diff
@@ -381,12 +387,12 @@ function $SanitizeProvider() {
         if (node.nodeType === 1) {
           handler.end(node.nodeName.toLowerCase());
```
koto: nodeName can be clobbered.

src/ngSanitize/sanitize.js
gkalpak: It is OK afaict. What we are trying to avoid is having an infinite loop, while traversing the DOM tree, which will eventually crash the browser.

src/ngSanitize/sanitize.js
gkalpak: What we are trying to avoid (afaict) is having an infinite loop, while traversing the DOM tree, which will eventually crash the browser.

I couldn't find a way to achive this via `nodeType` or `nodeName`. Did I miss something?

src/ngSanitize/sanitize.js
gkalpak: See https://github.com/angular/angular.js/pull/15699#discussion_r100998529.

src/ngSanitize/sanitize.js
koto: If an infinite loop is the only concern, disregard my comments. JS-based sanitizers can usually fail by making them throw which breaks applications not catching an exception. Commonly, DOM clobbering is used to do that. If exceptions in ngSanitize do not result in apps breaking, it's fine.

gkalpak: > If an infinite loop is the only concern, disregard my comments.

I meant that is what we are address in this PR :smiley:
But afaict, we are only sanitizing HTML from `ngBindHtml`'s watch-action. Throwing inside a watch-action, will catch the error, pass it to the `$exceptionHandler` and move on to the next watcher.

But if someone uses it differently, it is possible that it breaks the app (as any thrown exception could). Do you think `$sanitize` should be handling its own errors, @koto ?

koto: If the errors are triggered by processing untrusted HTML (which is likely the most common code), I think the safe default would be to return an empty string instead of throwing and requiring application to catch. But tat this point it's not really a security issue, but should be similar to how other core services behave (i.e. is it expected for them to throw and it should be handled by apps).

src/ngSanitize/sanitize.js
```diff
@@ -518,8 +524,17 @@ function $SanitizeProvider() {
...
 
-      node = node.nextSibling;
+      node = getNonDescendant('nextSibling', node);
+    }
+  }
+
+  function getNonDescendant(propName, node) {
+    // An element is clobbered if its `propName` property points to one of its descendants
+    var nextNode = node[propName];
+    if (nextNode && nodeContains.call(node, nextNode)) {
```
petebacondarwin: are we bothered about generic descendents or only inputs that are members of a form?
if the latter then could we do:

```
if (nextNode && nextNode.form === node)
```

gkalpak: I don't know of another usecase that would be affected other than `<form>` elements. (`document` is affected too, but we can't access `document` inside the htmlParsers afaict.)
I can't be 100% sure though. DOM has many dark corners :smiley:

gkalpak: Hm...it seems that this won't work for `img` elements:

```html
<form name="foo>
  <input name="bar" />
  <img name="baz" />
</form>
```

```js
document.foo.bar        //--> `<input ... />`
document.foo.bar.form   //--> `<form ... >`

document.foo.baz        //--> `<img ... />`
document.foo.baz.form   //--> undefined
```

petebacondarwin: OK, let's leave it as it is then

petebacondarwin:
We merged this as-is. If there turns out to be performance issues then we will revisit and consider a different algorithm.

PR #15717, opened 2017-02-16
Test on Safari 10

Narretz:


Narretz:
This is the test that fails: https://github.com/angular/angular.js/blob/4d43ee3327fe5fbeb6bc3ac3e91426e574d8c25f/test/ngMock/angular-mocksSpec.js#L949-L955

In Safari 10.0.1, this test fails if the test file uses "use strict". Maybe this was fixed in 10.0.2/3

Here's a plunkr to test: http://plnkr.co/edit/GD0CeGnzuYa1NHIEcj65?p=preview

Narretz:
Safari 10 fails here:
```
  ngMock

    jasmine module and inject

      inject

        error stack trace when called outside of spec context

          on browsers supporting stack traces

            ✗ should update thrown Error stack trace with inject call location

	Expected 'test/ngMock/angular-mocksSpec.js:1011:28

	WorkFn@src/ngMock/angular-mocks.js:3159:26

	test/ngMock/angular-mocksSpec.js:1020:45

	attemptSync@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1950:28

	run@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1938:20
```

Narretz:
Added on master

PR #15724, opened 2017-02-19
fix($http): JSON parse failure

### LINKED ISSUE #15695: JSON parse failure in defaultHttpResponseTransform
```
function defaultHttpResponseTransform(data, headers) {
  if (isString(data)) {
    // Strip json vulnerability protection prefix and trim whitespace
    var tempData = data.replace(JSON_PROTECTION_PREFIX, '').trim();

    if (tempData) {
      var contentType = headers('Content-Type');
      **if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData))** {
        data = fromJson(tempData);
      }
    }
  }

  return data;
}
```

In defaultHttpResponseTransform() above, the response data is parsed as json, if either content type is application/json or is JSON like. In this case the data is unable to parsed as JSON in fromJson(), the exception is thrown at JSON.parse, which causes request to be rejected with empty object.
This makes it hard to debug response issues in production especially if response is dynamically generated, as no response data can be captured in this case.
My suggestion would be to reject with original response data in such cases and let client handle the failure.

chirag64:
Fixes #15695

Bug fix

#15695 

No

**Other information**:

### DIFF
--- a/docs/content/error/$http/baddata.ngdoc
+++ b/docs/content/error/$http/baddata.ngdoc
@@ -3,14 +3,12 @@
 @fullName Bad JSON Data
 @description
 
-The default @{link ng$http#default-transformations `transformResponse`} will try to parse the
+The default @{link ng.$http#default-transformations `transformResponse`} will try to parse the
 response as JSON if the `Content-Type` header is `application/json` or the response looks like a
 valid JSON-stringified object or array.
 This error occurs when that data is not a valid JSON object.
 
 The error message should provide additional context such as the actual response.
 
-To resolve this error, make sure you pass a valid JSON data object to `transformResponse`.
-
-For more information, see the {@link ng$http#default-transformations `transformResponse`} service
-API documentation.
+To resolve this error, make sure you pass valid JSON data to `transformResponse` or use an
+appropriate `Content-Type` header for non-JSON data.
--- a/docs/content/error/$http/baddata.ngdoc
+++ b/docs/content/error/$http/baddata.ngdoc
@@ -1,0 +1,13 @@
+@ngdoc error
+@name $http:baddata
+@fullName Bad JSON Data
+@description
+
+This error occurs when the data parameter passed to the {@link ng.$http `defaultHttpResponseTransform`} service is not a valid JSON object.
+`defaultHttpResponseTransform` expects the first parameter as a valid JSON object if the second parameter of headers specifies a Content-Type of JSON.
+
+The error message should provide additional context such as the actual value of the parameter that was received.
+
+To resolve this error, make sure you pass a valid JSON data object to `defaultHttpResponseTransform`.
+
+For more information, see the {@link ng.$http `defaultHttpResponseTransform`} service API documentation.
--- a/docs/content/error/$http/baddata.ngdoc
+++ b/docs/content/error/$http/baddata.ngdoc
@@ -3,11 +3,14 @@
 @fullName Bad JSON Data
 @description
 
-This error occurs when the data parameter passed to the {@link ng.$http `defaultHttpResponseTransform`} service is not a valid JSON object.
-`defaultHttpResponseTransform` expects the first parameter as a valid JSON object if the second parameter of headers specifies a Content-Type of JSON.
+The default @{link ng$http#default-transformations `transformResponse`} will try to parse the
+response as JSON if the `Content-Type` header is `application/json` or the response looks like a
+valid JSON-stringified object or array.
+This error occurs when that data is not a valid JSON object.
 
-The error message should provide additional context such as the actual value of the parameter that was received.
+The error message should provide additional context such as the actual response.
 
-To resolve this error, make sure you pass a valid JSON data object to `defaultHttpResponseTransform`.
+To resolve this error, make sure you pass a valid JSON data object to `transformResponse`.
 
-For more information, see the {@link ng.$http `defaultHttpResponseTransform`} service API documentation.
+For more information, see the {@link ng$http#default-transformations `transformResponse`} service
+API documentation.
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -142,7 +142,7 @@ function defaultHttpResponseTransform(data, headers) {
           data = fromJson(tempData);
         } catch (e) {
           throw $httpMinErr('baddata', 'Data must be a valid JSON object. Received: "{0}". ' +
-          'Error occurred: "{1}"', data, e);
+          'Parse error: "{1}"', data, e);
         }
       }
     }
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -138,7 +138,11 @@ function defaultHttpResponseTransform(data, headers) {
     if (tempData) {
       var contentType = headers('Content-Type');
       if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData)) {
-        data = fromJson(tempData);
+        try {
+          data = fromJson(tempData);
+        } catch (e) {
+          throw minErr('$http')('baddata', 'Data must be a valid JSON object. Received: {0}', data);
+        }
       }
     }
   }
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -141,7 +141,8 @@ function defaultHttpResponseTransform(data, headers) {
         try {
           data = fromJson(tempData);
         } catch (e) {
-          throw minErr('$http')('baddata', 'Data must be a valid JSON object. Received: {0}', data);
+          throw $httpMinErr('baddata', 'Data must be a valid JSON object. Received: "{0}". ' +
+          'Error occurred: "{1}"', data, e);
         }
       }
     }
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1442,6 +1442,17 @@ describe('$http', function() {
           expect(callback.calls.argsFor(1)[0]).toBe(null);
           expect(callback.calls.argsFor(2)[0]).toBe('');
         });
+
+        it('should return JSON data with error message if JSON is invalid', function() {
+          var errCallback = jasmine.createSpy('error');
+          $httpBackend.expect('GET', '/url').respond('{abcd}', {'Content-Type': 'application/json'});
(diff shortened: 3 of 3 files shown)


chirag64:
I signed it!

docs/content/error/$http/baddata.ngdoc
```diff
@@ -0,0 +1,13 @@
+@ngdoc error
+@name $http:baddata
+@fullName Bad JSON Data
+@description
+
+This error occurs when the data parameter passed to the {@link ng.$http `defaultHttpResponseTransform`} service is not a valid JSON object.
```
gkalpak: Please limit lines to 100 chars.

gkalpak: The link should point tot  a specific section (e.g. `ng.$httpProvider.defaults`, `ng.$http.defaults`, or `ng$http#default-transformations`).

BTW, I just noticed that `$httpProvider.defaults.transformRequest` and `$httpProvider.defaults.transformResponse` are not documented. They should be.

docs/content/error/$http/baddata.ngdoc
```diff
@@ -0,0 +1,13 @@
...
+`defaultHttpResponseTransform` expects the first parameter as a valid JSON object if the second parameter of headers specifies a Content-Type of JSON.
```
gkalpak: Don't mention `defaultHttpResponseTransform`, because it is private and the users don't know what it is.
Just say that _"The default `transformResponse` will try to parse the response as JSON if the `Content-Type` header is `application/json` or the response looks like a valid JSON-stringified object or array"_.

docs/content/error/$http/baddata.ngdoc
```diff
@@ -0,0 +1,13 @@
...
+
+The error message should provide additional context such as the actual value of the parameter that was received.
```
gkalpak: _actual value of the parameter that was received_ --> _actual response_

src/ng/http.js
```diff
@@ -138,7 +138,11 @@ function defaultHttpResponseTransform(data, headers) {
     if (tempData) {
       var contentType = headers('Content-Type');
       if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData)) {
-        data = fromJson(tempData);
+        try {
+          data = fromJson(tempData);
+        } catch (e) {
+          throw minErr('$http')('baddata', 'Data must be a valid JSON object. Received: {0}', data);
```
gkalpak: We already have an `$httpMinErr` instance. Use that.

gkalpak: Also, please, embed the actual error message into the `minErr`.

gkalpak (changes requested):

gkalpak:
For tests, you can "train" `$httpBackend` with an invalid response and call `$http` (which will indirectly invoke `defaultHttpResponseTransform`).

Narretz:
@chirag64 There's already a test in httpSpec that should show you how to test this, see my comment here: https://github.com/angular/angular.js/issues/15695#issuecomment-279353290

In short, you don't access the http transform function directly. You call http, and define a mock json answer that will be parsed.

docs/content/error/$http/baddata.ngdoc
```diff
@@ -0,0 +1,16 @@
+@ngdoc error
+@name $http:baddata
+@fullName Bad JSON Data
+@description
+
+The default @{link ng$http#default-transformations `transformResponse`} will try to parse the
```
gkalpak: Shouldn't this be `ng.$http#...` (with a dot after `ng`)?

docs/content/error/$http/baddata.ngdoc
```diff
@@ -0,0 +1,16 @@
...
+response as JSON if the `Content-Type` header is `application/json` or the response looks like a
+valid JSON-stringified object or array.
+This error occurs when that data is not a valid JSON object.
+
+The error message should provide additional context such as the actual response.
+
+To resolve this error, make sure you pass a valid JSON data object to `transformResponse`.
+
+For more information, see the {@link ng$http#default-transformations `transformResponse`} service
```
gkalpak: This is not necessary, since we already link to `default-transformations` above imo.

docs/content/error/$http/baddata.ngdoc
```diff
@@ -0,0 +1,16 @@
...
+@description
+
+The default @{link ng$http#default-transformations `transformResponse`} will try to parse the
+response as JSON if the `Content-Type` header is `application/json` or the response looks like a
+valid JSON-stringified object or array.
+This error occurs when that data is not a valid JSON object.
+
+The error message should provide additional context such as the actual response.
+
+To resolve this error, make sure you pass a valid JSON data object to `transformResponse`.
```
gkalpak: _a valid JSON data object_ --> _valid JSON data_

Also, the correct resolution depends on the source of the problem. In some cases, one might need to send correct headers (to not include `application/json` for example).

test/ng/httpSpec.js
```diff
@@ -1442,6 +1442,17 @@ describe('$http', function() {
           expect(callback.calls.argsFor(1)[0]).toBe(null);
           expect(callback.calls.argsFor(2)[0]).toBe('');
         });
+
+        it('should return JSON data with error message if JSON is invalid', function() {
+          var errCallback = jasmine.createSpy('error');
+          $httpBackend.expect('GET', '/url').respond('{abcd}', {'Content-Type': 'application/json'});
+            $http({method: 'GET', url: '/url'}).then(callback).catch(errCallback);
```
gkalpak: Wrong indentation.

Nit: It is simpler to use `$http.get('/url')`.

src/ng/http.js
```diff
@@ -138,7 +138,12 @@ function defaultHttpResponseTransform(data, headers) {
     if (tempData) {
       var contentType = headers('Content-Type');
       if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData)) {
-        data = fromJson(tempData);
+        try {
+          data = fromJson(tempData);
+        } catch (e) {
+          throw $httpMinErr('baddata', 'Data must be a valid JSON object. Received: "{0}". ' +
+          'Error occurred: "{1}"', data, e);
```
gkalpak: Maybe: _Error occurred_ --> _Original error_ or _Parse error_

gkalpak (changes requested):

docs/content/error/$http/baddata.ngdoc
chirag64: How about

To resolve this error, make sure you pass valid JSON data to `transformResponse` or use the appropriate `Content-Type` instead of the specified `application/json` if passing non-JSON data.

gkalpak: How about:

> To resolve this error, make sure you pass valid JSON data to `transformResponse` or use an appropriate `Content-Type` header for non-JSON data.

Narretz:
One thing we need to decide is if the original response data should be available in the thrown error, for recovery attempts. I think it's useful, but we don't really have the option for this. the data is only available in serialized form in the error (or is it), so we'd have to create a new API or wrap the error with a different response object.

gkalpak:
@Narretz, not sure what you mean 😕 The original response is usually just text so we can include it in the error message without further serialization.

chirag64:
@gkalpak Please let me know if you want me to make any changes on top of the last commit that I made.

Narretz:
@gkalpak My concern is that you cannot get the original response string out the minErr without some extra parsing, because the data passed to the minErr is stringified into the message template. So with this solution it's harder to get the actual response string in case you want to something with the response only.

gkalpak:
I see. I am that concerned about that. If they just want to log it (e.g. to the server), the extra context won't hurt. And if they want to try and fix the issue or handle it in some other way, then they having their own `tranformResponse` sounds a better approach anyway.

But I don't feel too strongly about it.

The problem is that I don't think it is possible to somehow pass multiple values (e.g. our error and the response as searate entities). Unless you have a clever idea :smiley:

Narretz:
Ok, I think it's reasonable to implement your own transformer if you need recovery or something special. 

@chirag64 there's a failing test that expects something else from a failed JSON transform at https://github.com/angular/angular.js/blob/3dc0096dc4469808bd221610d71febe3f3e52d66/test/ng/httpSpec.js#L1372-L1383. Please 

* remove this test 
* move your added test at this position (in the section about default response transform).

chirag64:
@Narretz done 👍🏼

test/ng/httpSpec.js
```diff
@@ -1369,17 +1369,15 @@ describe('$http', function() {
...
           );
 
-          it('should forward json deserialization errors to the http error handler',
-              function() {
+          it('should return JSON data with error message if JSON is invalid', function() {
             var errCallback = jasmine.createSpy('error');
-
-            $httpBackend.expect('GET', '/url').respond('abcd', {'Content-Type': 'application/json'});
-            $http({method: 'GET', url: '/url'}).then(callback).catch(errCallback);
+            $httpBackend.expect('GET', '/url').respond('{abcd}', {'Content-Type': 'application/json'});
```
gkalpak: Why this change?

chirag64: I've replaced the `it` function of `forwarding json deserialization errors` with mine. The above highlighted statements are pretty much similar to the old code except that we're using an object `'{abcd}'` within the string instead of a direct string `'abcd'`. Are you referring to that string or something else?

Narretz: That's fine with me. @gkalpak wdyt?

gkalpak: @chirag64, yes, I am referring to the string. Why change `abcd` to `{abcd}`?
<sub>I am mostly curious - I don't think it makes any difference.</sub>

chirag64: I wanted to make it look clear that we're passing an invalid JSON instead of passing a string.

gkalpak (approved):

gkalpak:
LGTM. Leaving it to @Narretz to give the final LGTM.

Narretz:
LGTM, please merge @gkalpak

PR #15736, opened 2017-02-21
feat($compile): Lower the security context of SVG's a and image xlink:href

rjamet:
They should go through regular sanitization, RESOURCE_URL is overkill there.
This does not change the context for the rest of xlink:href attributes.

Minor breaking change. 

xlink:href are always RESOURCE_URL context, so they go through the $sce whitelist, but aren't sanitized. That context is intended for links to things that are potentially executed as scripts in your origin, which isn't the case here.

image and a xlink:href are now sanitized as link URLs (like a href), and won't complain on different-origin links anymore.

This is breaking, but only in marginal (and probably unsafe) cases. If you whitelisted something for a xlink:href binding on one of those two, you'll need to change the aHrefSanitizationWhitelist instead. In the overwhelming majority of cases, everything that works with the default settings will keep working, and those defaults aren't usually modified.

**Other information**:
I'm almost sure there are issues around ng-href in svg lying around, and probably general svg bindings security issues too :/ ng-href security context currently doesn't match the xlink:href ones. It can't be fixed without breaking changes though, so that's for another PR. Some simple tests: https://plnkr.co/edit/PbR1FH2v4luScSYdU16t?p=preview

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1674,7 +1674,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             (nodeName === 'img' && key === 'src') ||
             (nodeName === 'image' && key === 'xlinkHref')) {
           // sanitize a[href] and img[src] values
-          this[key] = value = $$sanitizeUri(value, key === 'src');
+          this[key] = value =
+              $$sanitizeUri(value, nodeName === 'img' || nodeName === 'image');
         } else if (nodeName === 'img' && key === 'srcset' && isDefined(value)) {
           // sanitize img[srcset] values
           var result = '';
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1671,7 +1671,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         nodeName = nodeName_(this.$$element);
 
         if ((nodeName === 'a' && (key === 'href' || key === 'xlinkHref')) ||
-            (nodeName === 'img' && key === 'src')) {
+            (nodeName === 'img' && key === 'src') ||
+            (nodeName === 'image' && key === 'xlinkHref')) {
           // sanitize a[href] and img[src] values
           this[key] = value = $$sanitizeUri(value, key === 'src');
         } else if (nodeName === 'img' && key === 'srcset' && isDefined(value)) {
@@ -3254,8 +3255,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         if (['img', 'video', 'audio', 'source', 'track'].indexOf(tag) === -1) {
           return $sce.RESOURCE_URL;
         }
-      // maction[xlink:href] can source SVG.  It's not limited to <maction>.
-      } else if (attrNormalizedName === 'xlinkHref' ||
+      } else if (
+          // Some xlink:href are okay, most aren't
+          (attrNormalizedName === 'xlinkHref' && (tag !== 'image' && tag !== 'a')) ||
+          // Formaction
           (tag === 'form' && attrNormalizedName === 'action') ||
           // If relative URLs can go where they are not expected to, then
           // all sorts of trust issues can arise.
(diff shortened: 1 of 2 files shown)


src/ng/compile.js
```diff
@@ -3254,8 +3255,10 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         if (['img', 'video', 'audio', 'source', 'track'].indexOf(tag) === -1) {
           return $sce.RESOURCE_URL;
         }
-      // maction[xlink:href] can source SVG.  It's not limited to <maction>.
-      } else if (attrNormalizedName === 'xlinkHref' ||
+      } else if (
+          // Some xlink:href are okay, most aren't
+         (attrNormalizedName === 'xlinkHref' && (tag !== 'image' && tag !== 'a')) ||
```
gkalpak: Wrong indentation 😱

gkalpak (approved):
LGTM

src/ng/compile.js
rjamet: Awww, I'm bad at counting spaces :/ Fixed !

gkalpak:
@rjamet: We discussed this and it does indeed qualify as a breaking change (even if the broken cases are dubious/uncommon). Can you update the commit message with a BC notice (and potentially change the scope from `fix` to `feature`)?
<sub>Unless you feel strongly about convincing us otherwise 😛 </sub>

rjamet:
Done. I don't feel strongly about anything here: this change is to be more permissive, but I can wait for the next breaking change window without worries.

gkalpak:
👍 
Did you update the commit message? I still see the old ones.

rjamet:
How can I do so? Rewriting history locally then force-pushing to my branch?

mgol:
> How can I do so? Rewriting history locally then force-pushing to my branch?

Yup.

rjamet:
Really done so this time, then. Thanks for the help :)

src/ng/compile.js
```diff
@@ -1671,7 +1671,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
         nodeName = nodeName_(this.$$element);
 
         if ((nodeName === 'a' && (key === 'href' || key === 'xlinkHref')) ||
-            (nodeName === 'img' && key === 'src')) {
+            (nodeName === 'img' && key === 'src') ||
+            (nodeName === 'image' && key === 'xlinkHref')) {
           // sanitize a[href] and img[src] values
           this[key] = value = $$sanitizeUri(value, key === 'src');
```
gkalpak: Shouldn't `key ===src` be updated to also account for `<image>` elements. Shouldn't `image[xlink:href]` be sanitized based on `imgSrcSanitizationWhitelist` rather than `aHrefSanitizationWhitelist`?

gkalpak: Ping @rjamet

rjamet: Sorry for the delay :/

For the two whitelists: yes, good catch, fixed! I swapped testing for 'src' with testing for the right tagnames, and I added a small test for that.

A few remarks:
 - The whitelists purpose aren't super clear: I interpret them as navigational / media URLs, but maybe someone has a different opinion?  
- Come to think of it, we could also add video/audio/picture src to media sanitization? That would require adding more data: types to the default media whitelist and break a few users, but it seems like an oversight on my part. I don't believe that they might cause XSS in the current state, but it's not like <video src="javascript:foo"> has any meaning either. 
- I'm advocating for a single whitelist in https://github.com/angular/angular.js/pull/14250 anyways :)

src/ng/compile.js
```diff
@@ -1674,7 +1674,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
             (nodeName === 'img' && key === 'src') ||
             (nodeName === 'image' && key === 'xlinkHref')) {
           // sanitize a[href] and img[src] values
-          this[key] = value = $$sanitizeUri(value, key === 'src');
+          this[key] = value =
+              $$sanitizeUri(value, nodeName === 'img' || nodeName === 'image');
```
gkalpak: Nit: There is no need to wrap this line (as it does not exceed 100 chars).

gkalpak (approved):

PR #15744, opened 2017-02-23
docs(orderBy): Clarify that orderBy does not work on JavaScript objects. 

maecapozzi:
This PR specifies that orderBy does not work on JavaScript objects. This is unclear in the documentation, and users would benefit from clarification here.

Docs update. 

Currently, `orderBy` does not work on objects. I think that the documentation would benefit from making this abundantly clear! 

N/A

N/A

**Other information**:

### DIFF
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -13,7 +13,7 @@
  * `[{id: 'bar'}, {id: 'foo'}]`.
  *
  * The `collection` can be an Array or array-like object (e.g. NodeList, jQuery object, TypedArray,
- * String, etc).
+ * String, etc). This will not work on a JavaScript object. 
  *
  * The `expression` can be a single predicate, or a list of predicates each serving as a tie-breaker
  * for the preceding one. The `expression` is evaluated against each item and the output is used
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -13,7 +13,7 @@
  * `[{id: 'bar'}, {id: 'foo'}]`.
  *
  * The `collection` can be an Array or array-like object (e.g. NodeList, jQuery object, TypedArray,
- * String, etc). This will not work on a JavaScript object. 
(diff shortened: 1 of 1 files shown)


maecapozzi:
`I signed it!`

gkalpak:
> This PR specifies that orderBy does not work on JavaScript objects. This is unclear in the documentation

How is this not clear enough:
_"The `collection` can be an Array or array-like object (e.g. NodeList, jQuery object, TypedArray, String, etc)."_

##
Also, "JavaScript objects" is a general term (and it's not accurrate to say it doesn't work witht them). Most things in JavaScripts (including Arrays and array-like objects) are JavaScript Obejcts.

maecapozzi:
Great point, @gkalpak. Let me specify what I mean and why I am suggesting this change.

Perhaps a better adjustment would be: 

**"Please note that `orderBy` does not work on Object Literals."** 

or 

**"an Object Literal does not qualify as an array-like object."**  

While white is already there is definitely clear if you are reading closely, if you are quickly glancing at documentation I think that specifically pointing out that `orderBy` doesn't work on Object Literals will help developers accidentally trying to use orderBy on an Object Literal. 

 The reason I am suggesting this change is because I had a team member run into an issue with orderBy when he assumed that an Object Literal is not an array-like object. Clearly, he wasn't reading closely enough, but it would have helped him if the documentation specified that `orderBy` doesn't work on Object Literals.   

Does that make my reasoning behind this PR clearer?

gkalpak:
I am really not sure whether this is going to be more helpful than confusing. (And there is only that much you can do for people not reading the docs carefully 😛)

BTW, there is nothing about object literals that makes them inherently "non-array-like-y". For example the following would work just fine 😛 

```html
<div ng-repeat="value in {0: {name: 'B'}, 1: {name: 'A'}, length: 2} | orderBy:'name'">
  {{ value.name }}
</div>
```

maecapozzi:
Ah, ok. All valid points! Thank you so much for taking the time to review this PR and enjoy the rest of your day. Closing now. :)

PR #15749, opened 2017-02-24
Autobootstrap fixes

petebacondarwin:

**Other information**:

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1532,6 +1532,12 @@ function getNgAttribute(element, ngAttr) {
 
 function allowAutoBootstrap(document) {
   var script = document.currentScript;
+
+  // If the `currentScript` property has been clobbered just return false, since this indicates a probable attack
+  if (!(script instanceof window.HTMLScriptElement || script instanceof window.SVGScriptElement)) {
+    return false;
+  }
+
   var src = script && script.getAttribute('src') || script.getAttribute('href') || script.getAttribute('xlink:href');
 
   if (!src) {
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1533,6 +1533,11 @@ function getNgAttribute(element, ngAttr) {
 function allowAutoBootstrap(document) {
   var script = document.currentScript;
 
+  if (!script) {
+    // IE does not have `document.currentScript`
+    return true;
+  }
+
   // If the `currentScript` property has been clobbered just return false, since this indicates a probable attack
   if (!(script instanceof window.HTMLScriptElement || script instanceof window.SVGScriptElement)) {
     return false;
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -1543,33 +1543,35 @@ function allowAutoBootstrap(document) {
     return false;
   }
 
-  var src = script && script.getAttribute('src') || script.getAttribute('href') || script.getAttribute('xlink:href');
+  var srcs = [script.getAttribute('src'), script.getAttribute('href'), script.getAttribute('xlink:href')];
 
-  if (!src) {
-    return true;
-  }
-
-  var link = document.createElement('a');
-  link.href = src;
-
-  if (document.location.origin === link.origin) {
-    // Same-origin resources are always allowed, even for non-whitelisted schemes.
-    return true;
-  }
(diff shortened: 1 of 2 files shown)


test/AngularSpec.js
```diff
@@ -1708,83 +1708,78 @@ describe('angular', function() {
...
 
-      var src = extensionScheme + '://something';
-      // Fake a minimal document object (the actual document.currentScript is readonly).
-      var fakeDoc = {
-        currentScript: { getAttribute: function() { return src; } },
-        location: {protocol: extensionScheme + ':', origin: extensionScheme + '://something'},
-        createElement: document.createElement.bind(document)
-      };
-      expect(allowAutoBootstrap(fakeDoc)).toBe(true);
+        function createFakeDoc(attrs, protocol, origin, currentScript) {
```
mgol: 4 parameters are not very readable. Could you get the protocol from the origin? `currentScript` & `attrs` could be unified, just check if `currentScript` is a plain object and create a true `script` that extends it if it is. Then you'd only have 2 params.

Otherwise a dictionary would be more readable (i.e. `function createFakeDoc({attrs, protocol, origin, currentScript})` but in ES5).

test/AngularSpec.js
```diff
@@ -1708,83 +1708,94 @@ describe('angular', function() {
...
 
-      src = extensionScheme + '://something-else';
-      expect(allowAutoBootstrap(fakeDoc)).toBe(false);
-    });
+          protocol = protocol || 'http:';
+          origin = origin || 'http://some-value';
```
gkalpak: Maybe change to `origin = origin || protocol + '//some-vaue';` and save some boilerplate in tests.

gkalpak (changes requested):
I think we should also have more tests featuring cases where more than one of `src`/`href`/`xlink:href` are present.

src/Angular.js
```diff
@@ -1532,7 +1532,13 @@ function getNgAttribute(element, ngAttr) {
 
 function allowAutoBootstrap(document) {
   var script = document.currentScript;
-  var src = script && script.getAttribute('src');
+
+  // If the `currentScript` property has been clobbered just return false, since this indicates a probable attack
+  if (!(script instanceof window.HTMLScriptElement || script instanceof window.SVGScriptElement)) {
```
gkalpak: This fails on IE, which does not support `document.currentScript`. This should probably be:

```js
if (!script) {
  return true;
} else if (!(script instanceof window.HTMLScriptElement || script instanceof window.SVGScriptElement)) {
  return false;
}

var src = ...
```

src/Angular.js
```diff
@@ -1532,7 +1532,13 @@ function getNgAttribute(element, ngAttr) {
...
+    return false;
+  }
+
+  var src = script && script.getAttribute('src') || script.getAttribute('href') || script.getAttribute('xlink:href');
```
petebacondarwin: Good catch let's just check them all.

src/Angular.js
petebacondarwin: Hmm, OK. I am concerned that one could clobber the `currentScript` to become null?

test/AngularSpec.js
petebacondarwin: I have lowered it to three params. I don't want to start with a dictionary because for the majority of the tests you only provide the first param.

test/AngularSpec.js
petebacondarwin: I have refactored to remove the `origin` param altogether

src/Angular.js
gkalpak: 1. This was the case before anyway (i.e. if `script` was falsy, we returned `true`).
2. If they can clobber `document.currentScript` to be null, I guess we are doomed. DOOOMED 😱

mgol (approved):

gkalpak (approved):

Narretz:
Landes as ebe90051eda8a3328e5993cca1663e28d03113d0 a649758655843275cc477fb638f8e55f72a4eaa6 c357b1aba6f990820c6789ff43727e4754e7713a

petebacondarwin:
Weird - thought it would auto-close as I thought I fast forward merged into master.

PR #15761, opened 2017-02-28
fix(ngTouch): remove ngClick override and $touchProvider.ngClickOverr…

### LINKED ISSUE #15755: Remove ngClick override from ngTouch

Narretz:
…ideEnabled()

Closes #15755

BREAKING CHANGE:

The `ngClick` directive from the ngTouch module has been removed, and with the also the
corresponding $touchProvider.ngClickOverrideEnabled() method.

If you have included ngTouch in your application with a version of 1.5.0 or higher, and have not
changed the value of $touchProvider.ngClickOverrideEnabled()`, then there are no migration steps.

The `ngClick` override directive had been deprecated and by default disabled since v1.5.0,
because of buggy behavior in edge cases, and a general trend to avoid special touch based
overrides of click events. If you still need a touch override, consider using
[Fastclick](https://github.com/ftlabs/fastclick).

Note that the `$touch` service still offers the `ngClickOverrideEnabled()` function for
libraries that want to detect if an app uses the click override. Starting with 1.7.0, it
will always return `false`.

Other info:

This one is a bit tricky, as we have deprecated the ngClick directive, but at the same time introduced two new APIs: $touchProvider.ngClickOverrideEnabled and $touch.ngClickOverrideEnabled().
I have removed the former, because it's directly tied to the directive. I think that's fine, as someone who uses the ngClick override must migrate to a different solution anyway, and will see an error when he updates and the provider fn is missing. I've kept the actual provider for the unlikely case that we introduce new config for the ngTouch module.
I've also kept the $touch.ngClickOverrideEnabled() method, as this one is used by material to check if they should include their own touch stuff. I think if we removed this then they wouldn't be able to determine if the ngClick override is present (because ngTouch module but no $touch service could also mean a pre 1.5.0 version of ngTouch)


src/ngTouch/touch.js
```diff
@@ -114,14 +54,13 @@ function $TouchProvider($provide, $compileProvider) {
       /**
        * @ngdoc method
        * @name  $touch#ngClickOverrideEnabled
-       *
-       * @returns {*} current value of `ngClickOverrideEnabled` set in the {@link ngTouch.$touchProvider $touchProvider},
-       * i.e. if {@link ngTouch.ngClick ngTouch's ngClick} directive is enabled.
+       *       *
```
mgol: A redundant line? Also, an extra `*`.

src/ngTouch/touch.js
```diff
@@ -35,77 +33,19 @@ function nodeName_(element) {
...
+function $TouchProvider() {
 
   /**
   * @ngdoc service
   * @name $touch
   * @kind object
   *
   * @description
-  * Provides the {@link ngTouch.$touch#ngClickOverrideEnabled `ngClickOverrideEnabled`} method.
+  * Provides the legacy {@link ngTouch.$touch#ngClickOverrideEnabled `ngClickOverrideEnabled`} method.
```
mgol: Can we deprecate it at the same time?

mgol (changes requested):

mgol:
> I think if we removed this then they wouldn't be able to determine if the ngClick override is present (because ngTouch module but no $touch service could also mean a pre 1.5.0 version of ngTouch)

This is unclear to me - if there was no `$touch` service, injecting it would result in an error. Is the service missing or just a method? If it's just a method then just checking for its existence could be treated as it returning `false`. Obviously we can still go through a deprecation cycle but I don't see why we shouldn't be able to remove it eventually.

> This one is a bit tricky, as we have deprecated the ngClick directive, but at the same time introduced two new APIs: $touchProvider.ngClickOverrideEnabled and $touch.ngClickOverrideEnabled().
> I have removed the former, because it's directly tied to the directive. I think that's fine, as someone who uses the ngClick override must migrate to a different solution anyway, and will see an error when he updates and the provider fn is missing.

Why does a person using the override have to migrate to a different solution? This doesn't break the app and in current mobile browsers the click delay is disabled in many situations anyway so it may very well be redundant. If we're keeping the method on `$touch` for now I think we may as well keep the setter on the provider and just document it's now a noop (and deprecate at the same time).

src/ngTouch/touch.js
gkalpak: Is there any reason to keep this, if we are removing `ngClick`?

mgol: @gkalpak It's all explained in the first post of the pull request. :-)

gkalpak: Oh, we are supposed to read that, right? Doh!

Narretz: So, it looks like material never implemented a check with ngClickOverrideEnabled, it simply is not compatible with ngTouch: https://github.com/angular/material/blob/24370e7627389a873a95777f8bebc8bb43c842ce/src/core/core.js#L23

I think that means we can remove $touch.ngClickOverrideEnabled
I doubt that anyone else uses it (famous last words)

Narretz:
> This is unclear to me - if there was no $touch service, injecting it would result in an error. Is the service missing or just a method? If it's just a method then just checking for its existence could be treated as it returning false. Obviously we can still go through a deprecation cycle but I don't see why we shouldn't be able to remove it eventually.

Good point. It turns out material never used the $touch service. FWIW, you can use $injector.has() or try-catch to find out if a service is available or not.

> Why does a person using the override have to migrate to a different solution? This doesn't break the app and in current mobile browsers the click delay is disabled in many situations anyway so it may very well be redundant. If we're keeping the method on $touch for now I think we may as well keep the setter on the provider and just document it's now a noop (and deprecate at the same time).

They don't have to, but I assume they *want* to. If they are on 1.6, they have to opt-in to the ngClick override, soif they did that, they want this behavior, which means they want it in the future, too.
But since material doesn't actually use $touch.ngClickOverrideEnabled we can remove the function.

petebacondarwin:
I think we should remove the `$touch.ngClickOverrideEnabled` getter too. Now that we have landed the `module.info()` method then library providers can work out what version of the `ngTouch` module is loaded.

Narretz:
I've removed the fn on the service, PTAL

mgol:
How long do we want to keep the noop `$touch` service?

Narretz:
I've kept it in case we want to put some stuff on it in the future. At the same time I think this is unlikely and what we should do instead is to deprecate ngTouch completely.

mgol (approved):

gkalpak:
Why deprecate `ngTouch` completely? Aren't the `ngSwipes` still useful?

Narretz:
My main point is that we have a few open PRs and bugy about ngSwipe that we will probably never merge / fix because we don't have the resource to test and make sure they work on all platforms. There are other touch gestures libraries that can provide this stuff, we should really focus on the core. There's also not touch stuff in Angular, it's either material or something 3rd party.

mgol:
I guess it makes sense to not keep this zombie stuff here then.

Narretz:
With Zombie stuff you mean the $touchProvider?

mgol:
The whole ngTouch, if we're not really maintaining it.

mgol:
In jQuery we have a separate org - https://github.com/jquery-archive/ - where we keep projects that we no longer maintain. Were we to remove ngTouch completely one day we could put it somewhere for others to pick up if they want.

petebacondarwin:
That is where `ngScenario` should go too :-)

Narretz:
I think I have a PR for removing ngScenario lying around.

And in that case, I will remove the $touchProvider completely.

petebacondarwin:
We were trying to get someone in the community to take ownership of ngScenario in https://github.com/angular/ngScenario but no one stepped up. Perhaps if we remove it in 1.7/8 then it will trigger more interest?

mgol:
@Narretz Regarding the `BREAKING CHANGE` commit message - in 99% of cases instead of using FastClick today it's better to set the proper viewport or a proper touch-manipulation value on html (I don't remember which one). I feel it's better to direct people to those solutions as they avoid all the problems related to the hacks solutions like FastClick have to apply.

petebacondarwin:
I agree with @mgol that the commit message should point to documentation (not necessarily ours) on best practices for this sort of thing rather than just pointing to fastclick. Otherwise LGTM (for master only)

Narretz:
I've updated the commit message (see last commit)

mgol:
@Narretz Re commit msg:

> and with the also the

A typo.

> - Since iOS 8, Safari removes the delay on so-called "slow taps".

This is incomplete. I wouldn't even mention iOS 8 but only 9.3+ which have conditions very similar to Chrome/Firefox/Edge mobile: https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_9_1.html#//apple_ref/doc/uid/TP40014305-CH10-SW8

mgol (changes requested):
@Narretz Re commit msg:

> and with the also the

A typo.

> - Since iOS 8, Safari removes the delay on so-called "slow taps".

This is incomplete. I wouldn't even mention iOS 8 due to its low market share (https://developer.apple.com/support/app-store/) but only 9.3+ which have conditions very similar to Chrome/Firefox/Edge mobile:
https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_9_1.html#//apple_ref/doc/uid/TP40014305-CH10-SW8

Narretz:
@mgol Updated again, PTAL

mgol (approved):

PR #15762, opened 2017-02-28
fix($controller): remove the option to instantiate controllers from c…

### LINKED ISSUE #15349: Remove the $controllerProvider.allowGlobals flag

A breaking change request.

`$controllerProvider.allowGlobals` is set to `false` by default. Setting it to `true` allows `$controller` to find controller constructors on `window` which is discouraged.

`$controllerProvider.allowGlobals` flag should be removed.

Relying on global controller constructors is an anti-pattern and can mask bugs as you can't always control all the global elements on the page. We've been discouraging this pattern for some time.

All.

Narretz:
…onstructors on `window`

This also removes the likewise deprecated `$controllerProvider.allowGlobals()` method.

Closes #15349

BREAKING CHANGE:

The option to instantiate controllers from constructors on the global `window` object
has been removed. Likewise, the deprecated `$controllerProvider.allowGlobals()`
method that could enable this behavior, has been removed.

This behavior had been deprecated since AngularJS v1.3.0, because polluting the global scope
is bad. To migrate, remove the call to $controllerProvider.allowGlobals() in the config, and
register your controller via the Module API or the $controllerProvider, e.g.

```
angular.module('myModule', []).controller('myController', function() {...});

angular.module('myModule', []).config(function($controllerProvider) {
  $controllerProvider.register('myController', function() {...});
});

```

### DIFF
--- a/docs/content/tutorial/step_02.ngdoc
+++ b/docs/content/tutorial/step_02.ngdoc
@@ -150,37 +150,21 @@ To learn more about AngularJS scopes, see the {@link ng.$rootScope.Scope Angular
 
 # Testing
 
-The "AngularJS way" of separating controller from the view, makes it easy to test code as it is being
-developed. If our controller were available on the global namespace, we could simply instantiate it
-with a mock scope object:
+## Testing Controllers
 
-<br />
-```js
-describe('PhoneListController', function() {
+The "AngularJS way" of separating the controller from the view makes it easy to test code as it is being
+developed. In the section "Model and Controller" we have registered our controller via a constructor
+function on the `phonecatApp` module.
 
-  it('should create a `phones` model with 3 phones', function() {
-    var scope = {};
-    var ctrl = new PhoneListController(scope);
+In tests, we use an AngularJS service, `$controller`, which will retrieve a controller by name. It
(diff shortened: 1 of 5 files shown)


mgol (approved):

gkalpak (approved):
There is also [a section in the phonecat tutorial][1] that refers to global controllers. It is not exactly clear what it refers to (does it refer to `allowGlobals` or a hypothetical non-AngularJS app that defines controllers as globals), but it is weird enough and maybe could be removed along with `allowGlobals()`.

Either way LGTM (as long as Travis si happy).

[1]: https://github.com/Narretz/angular.js/blob/678d5451c86b8d5390f7b200f287e2a4b61f0162/docs/content/tutorial/step_02.ngdoc#L151-L206

Narretz:
Thanks for the reviews - also feel free to re-start stalled Travis tests ;)

We should probably remove the section about global controllers from the tutorial.

Narretz:
I've removed the section about global controllers from the tutorial, PTAL

mgol (approved):
The removed tutorial section was also using manual construction that bypasses the DI so it was really bad anyway.

PR #15765, opened 2017-03-01
docs(guide/Migrating from Previous Versions): add info for 1.4 ngPatt…

### LINKED ISSUE #15758: The ng-pattern validator change from 1.3 to 1.4 is not mentioned in the migration guide
ng-pattern directive in version 1.3.16 validates modelValue 

```
ctrl.$validators.pattern = function(value) {
    return ctrl.$isEmpty(value) || isUndefined(regexp) || regexp.test(value);
 };
```
ng-pattern directive in version 1.4.14 validates viewValue

```
ctrl.$validators.pattern = function(modelValue, viewValue) {
    // HTML5 pattern constraint validates the input value, so we validate the viewValue
     return ctrl.$isEmpty(viewValue) || isUndefined(regexp) || regexp.test(viewValue);
 };
```
 it's a breaking change which is not mentioned in the migration guide from 1.3 to 1.4

diegomrsantos:
…ern and pattern breaking change

Introduced in commit 0e00108

Closes #15758

### DIFF
--- a/docs/content/guide/migration.ngdoc
+++ b/docs/content/guide/migration.ngdoc
@@ -1687,12 +1687,12 @@ ngModelCtrl.$formatters.push(function(value) {
 
 Due to [0e001084](https://github.com/angular/angular.js/commit/0e001084ffff8674efad289d37cb16cc4e46b50a),
 The `ngPattern` and `pattern` directives will validate the regex
-against the `viewValue` of `ngModel`, i.e. the value of the model
-before the $parsers are applied. Previously, the modelValue
+against the `$viewValue` of `ngModel`, i.e. the value of the model
+before the $parsers are applied. Previously, the `$modelValue`
 (the result of the $parsers) was validated.
 
 This fixes issues where `input[date]` and `input[number]` cannot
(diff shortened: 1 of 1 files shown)


diegomrsantos:
I signed it!

docs/content/guide/migration.ngdoc
```diff
@@ -1683,6 +1683,79 @@ ngModelCtrl.$formatters.push(function(value) {
 });
 ```
 
+#### ngPattern and pattern
+
+Due to [0e001084](https://github.com/angular/angular.js/commit/0e001084ffff8674efad289d37cb16cc4e46b50a),
+The `ngPattern` and `pattern` directives will validate the regex
+against the `viewValue` of `ngModel`, i.e. the value of the model
```
Narretz: You should prefix viewValue, modelValue with `$` as this is how they are stored in the model

Narretz (changes requested):

Narretz:
Can you add a note to the commit message that the content in the migration guide is taken from the commit message of 0e00108?

Narretz (approved):

PR #15767, opened 2017-03-01
fix($log): don't parse error stacks manually outside of IE/Edge

### LINKED ISSUE #15590: No sourcemap-enabled stack trace in loadModules $injectorMinErr (at least not in Chrome)

Bug

**Current behavior + steps to reproduce**

1. Use webpack or some other concatenation/minification of scripts that benefits of sourcemaps during debugging
2. (to be sure that minification errors are thrown) Enable strictDi: true in angular.bootstrap (or use ng-strict-di)
3. Attempt to include an angular module that has a reference or other problem. In this case, there is a problem with an undefined variable in config.js
4. Notice that the stack trace thrown by loadModules() is referencing to your bundled/minified javascript and not using sourcemaps (latest Chrome):

```
Error: [$injector:modulerr] Failed to instantiate module app due to:
ReferenceError: env is not defined
    at http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:87540:36
    at Object.invoke (http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:13342:20)
    at runInvokeQueue (http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:13225:36)
    at http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:13234:12
    at forEach (http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:8852:21)
    at loadModules (http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:13215:6)
    at createInjector (http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:13137:20)
    at doBootstrap (http://localhost:8080/bundle.app.cf5a8012b28c1225c1ed.js:10333:21)
```
5. Also notice that it is possible to click the arrow in the Chrome dev console and see a sourcemaps-enabled stack-trace, but *without the entry of config.js where the actual ReferenceError is* (it starts from the parent entry to the actual error)

On step 4, the stack trace should reference the actual source files, so that it is easy to understand where the problem lies in the source code, namely config.js:

```
ReferenceError: env is not defined
    at config.js:1268
    at Object.invoke (angular.js:4847)
    at runInvokeQueue (angular.js:4730)
    at angular.js:4739
    at forEach (angular.js:357)
    at loadModules (angular.js:4720)
    at createInjector (angular.js:4642)
    at doBootstrap (angular.js:1838)
...
```

Easier debugging of module inclusion, especially important when migrating to webpack or other bundler that requires sourcemaps to even have a change debugging in the development-environment.

Angular 1.6.1 (latest stable)
Chrome 55
Mac OSX El Capitan

Have not tested with latest snapshot.

A workaround is to replace the following line in angular.js:
        throw $injectorMinErr('modulerr', 'Failed to instantiate module {0} due to:\n{1}',
                  module, e.stack || e.message || e);
with:
        throw e;

This yields an understandable stacktrace in the console (shown above).

Note, according to https://bugs.chromium.org/p/chromium/issues/detail?id=376409 "there is no way to fix it for Error.stack property because this property is generated by V8 which know nothing about source map". Despite this, e.stack is preferred as the ErrorConstructor argument over e. 

This is evident when running
        console.log(e, e.stack);
just before the error is throw. First an error with a sourcemap-enabled stacktrace is displayed in the console, then a stacktrace without sourcemaps. 

However, simply replacing "e.stack || e.message || e" with "e" leads to a stack trace with sourcemaps enable but without config.js mentioned:

Uncaught Error: [$injector:modulerr] Failed to instantiate module app due to:
{}
http://errors.angularjs.org/1.6.1/$injector/modulerr?p0=app&p1=%7B%7D
    at angular.js:68
    at angular.js:4764
    at forEach (angular.js:357)
    at loadModules (angular.js:4720)
    at createInjector (angular.js:4642)
    at doBootstrap (angular.js:1838)
    at bootstrap (angular.js:1859)

mgol:
IE/Edge display errors in such a way that it requires the user to click in
4 places to see the stack trace. There is no way to feature-detect it so
there's a chance of the user agent sniffing to go wrong but since it's only
about logging, this shouldn't break apps. Other browsers display errors in
a sensible way and some of them map stack traces along source maps if available
so it makes sense to let browsers display it as they want.

Fixes #15590

Bug fix.

Stack traces don't have source maps applied in any browser. The stack trace is formatted by Angular and printed as a string.

Stack traces have source maps applied in Chrome/Safari. The received error is printed as-is instead of printing a transformed stack trace.

Maybe? The format of logged exceptions is not specified in the docs but there is a chance someone depends on it always being a string.

**Other information**:

### DIFF
--- a/src/ng/log.js
+++ b/src/ng/log.js
@@ -67,6 +67,15 @@ function $LogProvider() {
   };
 
   this.$get = ['$window', function($window) {
+    // Support: IE 9-11, Edge 12-14+
+    // IE/Edge display errors in such a way that it requires the user to click in 4 places
+    // to see the stack trace. There is no way to feature-detect it so there's a chance
+    // of the user agent sniffing to go wrong but since it's only about logging, this shouldn't
+    // break apps. Other browsers display errors in a sensible way and some of them map stack
+    // traces along source maps if available so it makes sense to let browsers display it
+    // as they want.
+    var formatStackTrace = msie || /\bEdge\//.test($window.navigator && $window.navigator.userAgent);
+
     return {
       /**
        * @ngdoc method
@@ -124,14 +133,7 @@ function $LogProvider() {
 
     function formatError(arg) {
       if (arg instanceof Error) {
-        // Support: IE 9-11, Edge 12-14+
-        // IE/Edge display errors in such a way that it requires the user to click in 4 places
-        // to see the stack trace. There is no way to feature-detect it so there's a chance
-        // of the user agent sniffing to go wrong but since it's only about logging, this shouldn't
-        // break apps. Other browsers display errors in a sensible way and some of them map stack
-        // traces along source maps if available so it makes sense to let browsers display it
-        // as they want.
-        if (arg.stack && (msie || /\bEdge\//.test($window.navigator.userAgent))) {
+        if (arg.stack && formatStackTrace) {
           arg = (arg.message && arg.stack.indexOf(arg.message) === -1)
               ? 'Error: ' + arg.message + '\n' + arg.stack
               : arg.stack;
--- a/src/ng/log.js
+++ b/src/ng/log.js
@@ -124,7 +124,14 @@ function $LogProvider() {
 
     function formatError(arg) {
       if (arg instanceof Error) {
-        if (arg.stack) {
+        // Support: IE 9-11, Edge 12-14+
+        // IE/Edge display errors in such a way that it requires the user to click in 4 places
+        // to see the stack trace. There is no way to feature-detect it so there's a chance
+        // of the user agent sniffing to go wrong but since it's only about logging, this shouldn't
+        // break apps. Other browsers display errors in a sensible way and some of them map stack
+        // traces along source maps if available so it makes sense to let browsers display it
+        // as they want.
+        if (arg.stack && (msie || /\bEdge\//.test($window.navigator.userAgent))) {
           arg = (arg.message && arg.stack.indexOf(arg.message) === -1)
               ? 'Error: ' + arg.message + '\n' + arg.stack
               : arg.stack;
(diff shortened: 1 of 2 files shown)


src/ng/log.js
```diff
@@ -124,7 +124,14 @@ function $LogProvider() {
...
       if (arg instanceof Error) {
-        if (arg.stack) {
+        // Support: IE 9-11, Edge 12-14+
+        // IE/Edge display errors in such a way that it requires the user to click in 4 places
+        // to see the stack trace. There is no way to feature-detect it so there's a chance
+        // of the user agent sniffing to go wrong but since it's only about logging, this shouldn't
+        // break apps. Other browsers display errors in a sensible way and some of them map stack
+        // traces along source maps if available so it makes sense to let browsers display it
+        // as they want.
+        if (arg.stack && (msie || /\bEdge\//.test($window.navigator.userAgent))) {
```
gkalpak: AFAICT, this is adding a new requirement that mocking `$window` on tests (directly or indirectly) relying on `$log.error()` must include `navigator.userAgent`. This might break some tests.

gkalpak (commented):
A concern about a possible BC (for tests).
LGTM otherwise.

src/ng/log.js
mgol: This could be fixed by changing that into `$window.navigator && $window.navigator.userAgent`. WDYT?

gkalpak: Works for me. 

Also, I think it is a good idea to do this whole check once when instanciating the `$log` service:

```js
var formatStackTrace = msie || ($window.navigator && /\bEdge\//.test($window.navigator.userAgent));
```

mgol: PR updated.

petebacondarwin (approved):
LGTM

gkalpak (approved):

PR #15771, opened 2017-03-02
docs(guide/Providers): Remove redundant apology

jason-larigakis-hs:
There is no longer a service called "service", so this is confusing.


jason-larigakis-hs:
@googlebot: I signed it!

Issue #15801, opened 2017-03-12
ngOptions slow performance in IE due to option rerendering


pbr1111:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Currently, when we create a `<select>` with ngOptions, the `updateOptions()` method is called twice, causing a complete repainting of the options. The first time is called manually and the second time is called from the getWatchables watch. This is the code:

```
// We need to do this here to ensure that the options object is defined
// when we first hit it in writeNgOptionsValue
updateOptions();

// We will re-render the option elements if the option values or labels change
scope.$watchCollection(ngOptions.getWatchables, updateOptions);
```

This could be a bug, or not. In my opinion it doesn't matter if we reuse the elements, but with the current implementation it's a performance flaw. 

The problem appears with the [perf(ngOptions): use documentFragment to populate select](https://github.com/angular/angular.js/commit/97b3e003bc60f979fddecd16bc50304b5bef7e53) commit when the behavior of updateOptions() was changed. 

The main problem with this commit occurs when `options` are already defined (the options have changed), all current options are removed from the DOM one by one. In slow browsers (like IE), removing a child is an expensive operation. With this commit you have tried to solve a performance problem, creating a worse one.

The real performance problem is solved with this commit  [perf(ngOptions): avoid calls to `element.value`](https://github.com/angular/angular.js/commit/3b7f29ff63e8bf02327a1430dcc2a4c83915a206). I tried to revert the commit [Revert 'perf(ngOptions): use documentFragment to populate select' commit](https://github.com/pbr1111/angular.js/commit/7e27b1e50278506aeafbfb382ebcdcdb9ae788fc) and the performance in IE is now similar to Chrome. 

I create a plunker to show the difference: http://plnkr.co/edit/T83r9GXG6NfMZqSR5jXZ (to change between Angular current implementation and the revert of the commit [Revert 'perf(ngOptions): use documentFragment to populate select' commit](https://github.com/pbr1111/angular.js/commit/7e27b1e50278506aeafbfb382ebcdcdb9ae788fc) change the commented `<scripts src="">`). To be able to calculate the render time of `<select>` I've made a wrap inside a custom directive. The time it takes to render is displayed in the browser console. The `<select>` have 10000 items to render (IE will not be able to load the `<select>` with the current implementation). Change between the angular.js scripts and see the difference.

I'll show you another bug introduced in this commit. The new updateOptions method starts with: 
```
[...]

PR #15812, opened 2017-03-14
perf(ngOptions): prevent initial options repaiting

pbr1111:
Bug fix.

Refers to this issue: [ngOptions slow performance in IE due to option rerendering](https://github.com/angular/angular.js/issues/15801)

Now the options are loaded once when we create the `<select ng-options>` directive. Also, if we add an `<option>` element with an ng-if directive, it is not deleted now.

No. 

**Other information**:
None.

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -428,6 +428,9 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
         }
       }
 
+      // The empty option will be compiled and rendered before we first generate the options
+      selectElement.empty();
+
       var providedEmptyOption = !!selectCtrl.emptyOption;
 
       var unknownOption = jqLite(optionTemplate.cloneNode(false));
@@ -449,6 +452,9 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
       if (!multiple) {
 
         selectCtrl.writeValue = function writeNgOptionsValue(value) {
+          if (!options)
+            return;
+
           var selectedOption = options.selectValueMap[selectElement.val()];
           var option = options.getOptionFromViewValue(value);
 
@@ -508,6 +514,9 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
       } else {
 
         selectCtrl.writeValue = function writeNgOptionsMultiple(values) {
+          if (!options)
+            return;
+
           // Only set `<option>.selected` if necessary, in order to prevent some browsers from
           // scrolling to `<option>` elements that are outside the `<select>` element's viewport.
 
@@ -552,12 +561,11 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
 
       if (providedEmptyOption) {
 
-        // we need to remove it before calling selectElement.empty() because otherwise IE will
-        // remove the label from the element. wtf?
-        selectCtrl.emptyOption.remove();
-
         // compile the element since there might be bindings in it
         $compile(selectCtrl.emptyOption)(scope);
+
+        // Ensure that the empty option is always there if it was explicitly provided
+        selectElement.prepend(selectCtrl.emptyOption);
 
         if (selectCtrl.emptyOption[0].nodeType === NODE_TYPE_COMMENT) {
           // This means the empty option has currently no actual DOM node, probably because
@@ -590,12 +598,6 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
 
       }
 
-      selectElement.empty();
-
-      // We need to do this here to ensure that the options object is defined
-      // when we first hit it in writeNgOptionsValue
-      updateOptions();
-
       // We will re-render the option elements if the option values or labels change
       scope.$watchCollection(ngOptions.getWatchables, updateOptions);
 
@@ -624,7 +626,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
         // the option list in listbox style, i.e. the select is [multiple], or specifies a [size].
         // See https://github.com/angular/angular.js/issues/11314 for more info.
         // This is unfortunately untestable with unit / e2e tests
-        if (option.label !== element.label) {
+        if (isDefined(option.label)) {
           element.label = option.label;
           element.textContent = option.label;
         }
@@ -654,11 +656,6 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
         options = ngOptions.getOptions();
 
         var groupElementMap = {};
-
-        // Ensure that the empty option is always there if it was explicitly provided
-        if (providedEmptyOption) {
-          selectElement.prepend(selectCtrl.emptyOption);
-        }
 
         options.items.forEach(function addOption(option) {
           var groupElement;
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -2641,6 +2641,25 @@ describe('ngOptions', function() {
       dealoc(element);
     }));
 
+
+    it('should remove all comments inside the select element except the initial falsy ngIf empty option', function() {
+      scope.values = [
+        {name:'black'},
+        {name:'white'},
+        {name:'red'}
+      ];
+      scope.isBlank = false;
+
+      createSingleSelect('<!-- test comment -->'
+                          + '<option ng-if="isBlank" value="">blank</option>'
+                          + '<!-- test comment 2 -->');
+      scope.$apply('isBlank = true');
+
+      expect(element.find('option')[0].value).toBe('');
+      expect(element.find('option')[0].textContent).toBe('blank');
+      expect(element.find('option')[1].textContent).toBe('black');
+    });
+
   });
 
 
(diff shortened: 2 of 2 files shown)


Narretz:
Hi,
thanks for the Pr. As you can see in the Travis logs, there are a lot of test failures. Most of these come from the fact that there's a different writeSelectValue for select multiple, so you'll to make the change to this fn too.

Please also note that we definitely need to support an empty option with ngIf - there are some tests for this, and I assume some of them will fail in the current state.

You can run the unit tests locally with `grunt autotest`. This runs the test every time you change a file.

pbr1111:
Hi, let's see if I can fix it. Thanks for the info!

pbr1111:
I just uploaded the fixed code. Let's see if it pass the tests.

pbr1111:
The code just passed all the tests. This solution avoids double repainting of `<options>` and also makes ng-if work correctly in the empty option.

Narretz:
Thx @pbr1111, I will review this next week. Was there any reason for closing / reopening the issue so often?

pbr1111:
I close and reopen because eslint tests fails several times and I've made a rebase of my branch each time it fails to upload the new commit with the errors solved.

Narretz:
If you push new / changed code, it'll run Travis again, no need to close / reopen the issue.

src/ng/directive/ngOptions.js
```diff
@@ -624,7 +629,7 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
         // the option list in listbox style, i.e. the select is [multiple], or specifies a [size].
         // See https://github.com/angular/angular.js/issues/11314 for more info.
         // This is unfortunately untestable with unit / e2e tests
-        if (option.label !== element.label) {
+        if (option.label !== undefined) {
```
gkalpak: Why this change?

src/ng/directive/ngOptions.js
```diff
@@ -420,12 +420,19 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
...
       // The emptyOption allows the application developer to provide their own custom "empty"
       // option when the viewValue does not match any of the option values.
-      for (var i = 0, children = selectElement.children(), ii = children.length; i < ii; i++) {
-        if (children[i].value === '') {
-          selectCtrl.hasEmptyOption = true;
-          selectCtrl.emptyOption = children.eq(i);
-          break;
-        }
+      var optionsToRemove = [];
+      for (var i = 0, children = selectElement[0].childNodes, ii = children.length; i < ii; i++) {
```
gkalpak: Is the move from jqLite/jQuery APIs to native APIs part of the perf fix or is it just a "generally good idea" type of change?

pbr1111: jqLite/jQuery children() method doesn't return the nodes of nodeType equal to NODE_TYPE_COMMENT. We can use jqLite/jQuery contents() method instead, but I prefer to use native API.

src/ng/directive/ngOptions.js
pbr1111: element.label always will be undefined (because we create a new empty option element each time), then it's useless to read element.label value if we know that its value is undefined.

src/ng/directive/ngOptions.js
```diff
@@ -420,12 +420,19 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
...
+          if (!selectCtrl.hasEmptyOption
+              && (children[i].value === '' || children[i].nodeType === NODE_TYPE_COMMENT)) {
```
gkalpak: So, any comment will be interpreted as empty option? This doesn't seem right.

pbr1111: True. I'm doing changes to my commit. I will simplify this whole part and leave it as close to the previous code.

pbr1111:
To modify the commit, is there any way to do it without having to rebase my branch?

pbr1111:
I just modified the previous commit. Now the select element is emptied after it founds (or not) an empty option (the code is the same as before, it works because the options haven't yet been compiled in postLink, my previous solution was totally incorrect). The empty option element is readded if it's provided.

Narretz:
@pbr1111 You can just push a new commit and you / we will squash before merging.

> it works because the options haven't yet been compiled in postLink, my previous solution was totally incorrect

You should include a test that works with the correct version but fails with the incorrect one - it sounds like we are missing a test there

pbr1111:
I added some tests to check both cases. The second test https://github.com/angular/angular.js/pull/15812/commits/2ec5c07a95d1f6a90151048b49e6579e7cf36539 should fail with the current implementation, but it doesn't fail with https://github.com/angular/angular.js/pull/15812/commits/eb3661c6e4d66c76c3cbaed099b21719308dea6a commit.

src/ng/directive/ngOptions.js
```diff
@@ -449,36 +452,38 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
       if (!multiple) {
 
         selectCtrl.writeValue = function writeNgOptionsValue(value) {
-          var selectedOption = options.selectValueMap[selectElement.val()];
-          var option = options.getOptionFromViewValue(value);
+          if (options) {
```
Narretz: Can you change to `if (!options) return`? This will reduce the whitespace changes.

test/ng/directive/ngOptionsSpec.js
```diff
@@ -2641,6 +2658,23 @@ describe('ngOptions', function() {
...
+
+    it('should ignore the comments within the select', function() {
+      scope.values = [
+        {name:'black'},
+        {name:'white'},
+        {name:'red'}
+      ];
+      scope.isBlank = true;
+
+      createSingleSelect('<!-- test comment --><option ng-if="isBlank" value="">blank</option><!-- test comment 2 -->');
```
Narretz: Line is too long, please split at around 100 chars

test/ng/directive/ngOptionsSpec.js
```diff
@@ -2641,6 +2658,23 @@ describe('ngOptions', function() {
       dealoc(element);
     }));
 
+
+    it('should ignore the comments within the select', function() {
```
Narretz: There are no expectations regarding the comments you added. Should this test  that all comments that are not part of the empty option are removed?

test/ng/directive/ngOptionsSpec.js
```diff
@@ -779,6 +779,23 @@ describe('ngOptions', function() {
     expect(options[2]).not.toBeMarkedAsSelected();
   });
 
+
+  it('should render the options only one time', function() {
```
Narretz: the **initial** options

test/ng/directive/ngOptionsSpec.js
```diff
@@ -779,6 +779,23 @@ describe('ngOptions', function() {
...
+    scope.values = ['black', 'white', 'red'];
+    createSelect({
+      'ng-model': 'value'
+    });
+
+    element.attr('ng-options', 'value for value in values');
```
Narretz: you should pass the ngoptions expression like you did with ng-model. Or use createSingleSelect which has useful defaults.

src/ng/directive/ngOptions.js
Narretz: You should use `isDefined`here

src/ng/directive/ngOptions.js
```diff
@@ -428,6 +428,9 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
         }
       }
 
+      // Remove all the elements inside the select.
+      selectElement.empty();
```
Narretz: Please add something like:
"The empty option will be compiled and rendered before we first generate the options"

Narretz (changes requested):
Looking good so far! Some minor clarifications and changes nedded.  I would like for @petebacondarwin too also take a look because I remember he once wasn't a fan of doing `if (!options) return`

test/ng/directive/ngOptionsSpec.js
```diff
@@ -779,6 +779,23 @@ describe('ngOptions', function() {
...
+    var repaint = 0;
+    element[0].addEventListener('DOMSubtreeModified', function(ev) {
+      repaint++;
+      expect(repaint).toBeLessThan(2);
```
Narretz: Firefox 47 failed this expect. Probably because `DOMSubtreeModified`behaves differently in it. Maybe we should use MutationObserver instead for Browsers that support it?

Narretz (changes requested):

test/ng/directive/ngOptionsSpec.js
```diff
@@ -779,6 +779,32 @@ describe('ngOptions', function() {
...
 
+
+  it('should render the initial options only one time', function() {
+    scope.values = ['black', 'white', 'red'];
+    // Insert a select element into the DOM without ng-options to suscribe to DOM change events
+    createSelect({
+      'ng-model': 'value'
+    });
+
+    // Add ngOptions as attribute before recompiling the select element
```
Narretz: Thanks for clarifying why you did it like this ;)
However, recompiling is generally a bad idea, and I wouldn't put it in a test. The better option is to add the listener via a custom directive.

Narretz (changes requested):

Narretz:
Unfortunately, Firefox still reports 2 paints ... :(

PR #15824, opened 2017-03-16
chore(github): update the issue template

Narretz:
Updates the issue template.
- make it more like the Angular one
- include extra information into comments, so the actual output is cleaner and you can find the relevant info faster when triaging
- make it easier to differentiate between bugs and features
- remove section that are only relevant for features ... this is often filled out with redundant info
- make sections shorter and hopefully clearer

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,27 +1,36 @@
-***Note*: for support questions, please use one of these channels: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports.**
+<!--
+IF YOU DON'T FILL OUT THE FOLLOWING INFORMATION WE MIGHT CLOSE YOUR ISSUE WITHOUT INVESTIGATION
+-->
 
-**Do you want to request a *feature* or report a *bug*?**
+<!--
+- For *SUPPORT QUESTIONS*, use one of the
+[support channels](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question).
+- Before submitting, please **SEARCH GITHUB** for a similar issue or PR. -->
 
+**I'm submitting a ...**
+<!-- (check one with "x") -->
+- [ ] bug report
+- [ ] feature request
+- [ ] other (Please do not submit support requests here (see above))
 
+**Current behavior:**
+<!-- Describe how the bug manifests / how the current features are insufficient. -->
 
-**What is the current behavior?**
+**Expected / new behavior:**
+<!-- Describe what the behavior would be without the bug / how the feature would improve AngularJS -->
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,27 +1,40 @@
...
+IF YOU DON'T FILL OUT THE FOLLOWING INFORMATION WE MIGHT CLOSE YOUR ISSUE WITHOUT INVESTIGATING
+-->
+
+<!--
+- For *SUPPORT QUESTIONS*, use one of the
+[support channels](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#question).
+- Before submitting, please **SEARCH GITHUB** for a similar issue or PR. -->
+
+**I'm submitting a ...**  (check one with "x")
+```
```
gkalpak: I really hate that on the angular ISSUE_TEMPLATE. Why put this inside a code-block?

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,27 +1,40 @@
...
+https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:yBpEi4).
+-->
+
+**Angular version:** 1.x.y
+<!-- Check whether this is still an issue in the most recent stable or in the snapshot AngularJS version (https://code.angularjs.org/snapshot/) -->
+
+**Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
+<!-- All browsers where this could be reproduced -->
+
+**Environment:**
```
gkalpak: I am not sure this is very relevant for AngularJS. I would rather remove it from the template and "manually" request the extra info if necessary.

gkalpak (commented):
A couple of nits. 💯 otherwise :smiley:

.github/ISSUE_TEMPLATE.md
Narretz: I think it's because when it's inline you have an actual checkbox - which is good - but Github will show the issue as having "1 of 3" in the issue list. That might be a problem when we / someone adds more checkboxes, because then you cannot distinguish between the issue type and some sort of to-do list. Basically what we want here is a radio group ...

.github/ISSUE_TEMPLATE.md
Narretz: Sure - I'll add a note about the OS to the browser section

.github/ISSUE_TEMPLATE.md
gkalpak: I don't feel too strongly about it. I slightly prefer the checkboxes over code-blocks, but I am happy with this PR either way 👍

PR #15827, opened 2017-03-17
docs(filter): clarify the comparator parameter

fcostarodrigo:
The previous description does not make it clear how this parameter is used with other parameters
and its purpose.

This is a doc update.

N/A

N/A

This PR does not introduce a breaking change.

**Other information**:

### DIFF
--- a/src/ng/filter/filter.js
+++ b/src/ng/filter/filter.js
@@ -45,9 +45,9 @@
  *     The final result is an array of those elements that the predicate returned true for.
  *
  * @param {function(actual, expected)|true|false} [comparator] Comparator which is used in
- *     determining if matches returned by the argument expression should be considered a match based
- *     on the the expected value (from the filter expression) and actual value (from the object in
- *     the array).
+ *     determining if values retrieved using expression (when it is not a function) should be
+ *     considered a match based on the the expected value (from the filter expression) and actual
+ *     value (from the object in the array).
  *
  *   Can be one of:
  *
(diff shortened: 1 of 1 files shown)


fcostarodrigo:
I signed it!

src/ng/filter/filter.js
```diff
@@ -44,9 +44,9 @@
  *
  *     The final result is an array of those elements that the predicate returned true for.
  *
- * @param {function(actual, expected)|true|false} [comparator] Comparator which is used in
- *     determining if the expected value (from the filter expression) and actual value (from
- *     the object in the array) should be considered a match.
+ * @param {function(actual, expected)|true|false} [comparator] A predicate to further refine the
+ *     search results of an object or string expression. It should be used to determine if the
+ *     actual value from the array is equal to the expected value from the filter expression.
```
gkalpak: TBH, I feel this change is more confusing. E.g. `comparator` is **not** used to _"further refine the search results"_.  And _"should be used"_ sounds more like the user should do something (vs AngularJS using the function under the hood). And also `actual` is not necessarily a _"value from the array"_, but can be a property or sub-property of an item in the collection/array.

gkalpak (changes requested):

src/ng/filter/filter.js
```diff
@@ -45,8 +45,9 @@
  *     The final result is an array of those elements that the predicate returned true for.
  *
  * @param {function(actual, expected)|true|false} [comparator] Comparator which is used in
- *     determining if the expected value (from the filter expression) and actual value (from
- *     the object in the array) should be considered a match.
+ *     determining if matches returned by the argument expression should be considered a match based
```
gkalpak: The matches are not _returned_ by the `expression` argument. How about something like:

> determining if values retrieved using `expression` should be...

PR #15836, opened 2017-03-21
Allow NgModelController to be copied

jbedard:
fix

`copy` throws when copying a `NgModelController` instance. This will occur when deep watching such as when doing `<e ref="[myNgModel]">` (#15833)

The internal `NgModelController` `$$scope` is hidden

no, unless someone is doing something crazy that depends on `$$scope` being enumerable?

**Other information**:
I've split the commit up into 3.  I think the refactor + (modified) test should go into 1.7. I don't think the actual fix should go into 1.7.  The specific test case (watched a literal containing an ng-model) will no longer fail in 1.7. If we want to prevent other cases like manually deep-watching or manually `copy`ing an ng-model I think we should do it differently in 1.7...

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -890,8 +890,8 @@ function setupModelWatcher(ctrl) {
   //    -> scope value did not change since the last digest as
   //       ng-change executes in apply phase
   // 4. view should be changed back to 'a'
-  ctrl.$$scope.$watch(function ngModelWatch() {
-    var modelValue = ctrl.$$ngModelGet(ctrl.$$scope);
+  ctrl.$$scope.$watch(function ngModelWatch(scope) {
+    var modelValue = ctrl.$$ngModelGet(scope);
 
     // if scope model value and ngModel value are out of sync
     // TODO(perf): why not move this to the action fn?
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -281,7 +281,9 @@ function NgModelController($scope, $exceptionHandler, $attr, $element, $parse, $
 
   this.$$currentValidationRunId = 0;
 
-  this.$$scope = $scope;
+  // https://github.com/angular/angular.js/issues/15833
+  // Prevent `$$scope` from being iterated over by `copy` when NgModelController is deep watched
+  Object.defineProperty(this, '$$scope', {value: $scope});
   this.$$attr = $attr;
   this.$$element = $element;
   this.$$animate = $animate;
--- a/test/ng/compileSpec.js
+++ b/test/ng/compileSpec.js
@@ -5858,6 +5858,29 @@ describe('$compile', function() {
               expect(componentScope.owRef).toEqual({name: 'lucas', item: {name: 'martin'}});
             }));
 
+            // https://github.com/angular/angular.js/issues/15833
+            it('should work with ng-model inputs', function() {
+              var componentScope;
+
+              module(function($compileProvider) {
+                $compileProvider.directive('undi', function() {
+                  return {
+                    restrict: 'A',
+                    scope: {
+                      undi: '<'
+                    },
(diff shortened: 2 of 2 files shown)


src/ng/directive/ngModel.js
```diff
@@ -281,7 +281,10 @@ function NgModelController($scope, $exceptionHandler, $attr, $element, $parse, $
 
   this.$$currentValidationRunId = 0;
 
-  this.$$scope = $scope;
+  //https://github.com/angular/angular.js/issues/15833
+  //Prevent `$$scope` from being iterated over by `copy` when NgModelController is deep watched
+  //TODO: remove this or do not merge into 1.7
```
gkalpak: We shouldn't merge into 1.7, so you can remove this.

test/ng/compileSpec.js
```diff
@@ -5858,6 +5858,33 @@ describe('$compile', function() {
...
+              });
+
+              inject(function($compile, $rootScope) {
+                element = $compile('<form name="f" undi="[f.i]"><input name="i" ng-model="a"/></form>')($rootScope);
+                $rootScope.$apply();
+                expect($rootScope.f.i).toBeDefined();
+                expect(componentScope.undi).toBeDefined();
+                //TODO: the `.not` should be removed in 1.7 due to
+                //https://github.com/angular/angular.js/commit/fd4f0111188b62773b99ab6eab38b4d2b5d8d727
+                expect(componentScope.undi).not.toBe($rootScope.f.i);
```
gkalpak: I think this line is redundant (it doesn't add anything to the test). Could we use `toEqual()` (which would work the same in  1.6.x/1.7.x)?

jbedard: If we do keep this line I'd prefer `.toBe` in 1.7 to verify that it is the exact same instance, so I figured `.not.toBe` would be better for now, if we keep this line.

Only the `$compile` line is actually necessary to reproduce the issue. Maybe I should drop everything else?

src/ng/directive/ngModel.js
jbedard: I've dropped this line.

test/ng/compileSpec.js
gkalpak: MY reasoning is that there are other tests that verify the `toBe()` behavior (for 1.7), so I wouldn't be too worried about it here. But dropping everything not necessary for the test works for me too.

jbedard: Ok I've dropped everything except the verification that the binding actually got set...

gkalpak (approved):

Narretz:
LGTM, please squash and merge

test/ng/compileSpec.js
```diff
@@ -5858,6 +5858,29 @@ describe('$compile', function() {
               expect(componentScope.owRef).toEqual({name: 'lucas', item: {name: 'martin'}});
             }));
 
+            //https://github.com/angular/angular.js/issues/15833
```
mgol: Could we start every comment with a space?

src/ng/directive/ngModel.js
```diff
@@ -281,7 +281,9 @@ function NgModelController($scope, $exceptionHandler, $attr, $element, $parse, $
 
   this.$$currentValidationRunId = 0;
 
-  this.$$scope = $scope;
+  //https://github.com/angular/angular.js/issues/15833
+  //Prevent `$$scope` from being iterated over by `copy` when NgModelController is deep watched
```
mgol: Can we have spaces after `//` in above two lines?

test/ng/compileSpec.js
nachodiaz: why? order?

src/ng/directive/ngModel.js
nachodiaz: Yes, my friend ...

mgol (approved):

PR #15840, opened 2017-03-22
refactor($parse): move duplicate $parse interpreter/compiler logic into common location

jbedard:
... and deleted a couple other unused things.


gkalpak (approved):

Narretz:
Please add "and deleted a couple other unused things." to the commit message body, otherwise LGTM

PR #15842, opened 2017-03-22
refactor($compile): reuse shared simpleCompare method

jbedard:
I think this could be used in some other places as well once it's available...

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -69,6 +69,7 @@
     "arrayRemove": false,
     "copy": false,
     "shallowCopy": false,
+    "simpleCompare": false,
     "equals": false,
     "csp": false,
     "concat": false,
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -62,6 +62,7 @@
   includes,
   arrayRemove,
   copy,
+  simpleCompare,
   equals,
   csp,
   jq,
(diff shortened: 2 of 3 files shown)


gkalpak (changes requested):
In general LGTM. But if it is going to be used in other places, I would rather have it be done as part of this PR instead on a separate one. From a quick search, the two places I could find are:

- [equals][1]
- [ngModel][2]

(There are also some linting issues.)

[1]: https://github.com/jbedard/angular.js/blob/8a5996f6feebf54142700bbc15cd107f0f5e04cd/src/Angular.js#L1095
[2]: https://github.com/jbedard/angular.js/blob/8a5996f6feebf54142700bbc15cd107f0f5e04cd/src/ng/directive/ngModel.js#L898-L901

jbedard:
The `equals` and the `ngModel` watch cases are performance sensitive (invoked per digest) so I didn't want to change them.

The one I do want to change in a separate commit is [this block](https://github.com/angular/angular.js/blob/816236221700e2844ab6a52b44b36e135a98d293/src/ng/parse.js#L1792-L1816) which (I think) is equivalent to `simpleCompare` if the 3rd param is true...

gkalpak (approved):
The function is pretty simple, so I wouldn't expect it to decrease performance (I hope it would be inlined by the VM anyway). But I don't feel strongly about it.

LGTM

Narretz:
LGTM

PR #15844, opened 2017-03-22
Warn for missing / obsolete error docs

Narretz:
Chore

We don't know when error docs are missing or obsolete

No

**Other information**:
I have removed an obsolete animation error and moved the tpload error to the correct namespace. In the $templateRequest file, it looks like the error is created in the $compile namespace, but it actually ends up as the $templateRequest namespace, where it should be.


src/ng/templateRequest.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var $templateRequestMinErr = minErr('$compile');
+var $templateRequestMinErr = minErr('$templateRequest');
```
gkalpak: And no test broke? Seems like we're missing one :smiley:

gkalpak (approved):
I'm no dgeni guru, but afaik LGTM :smiley:
<sub>(It would be nice to have a test for `templateRequestMinErr`, but not a blocker.)</sub>

💯

src/ng/templateRequest.js
Narretz: Actually a lot of tests broke. 😨 I didn't think of implications of the change. I think we said that changing the error name/ namespace is a breaking change so we can't do that in 1.6.

Narretz:
I think this is actually an issue with min Err / error extraction task / script: - The min error for $compile:tpload gets extracted as $templateRequest:tpload to errors.json which means the erro match script gets thrown off

gkalpak:
We probably shouldn't use misleading namespaces (var name !== namespace).
We could fix it in 1.7.x

PR #15846, opened 2017-03-22
fix(jqLite): make jqLite invoke jqLite.cleanData as a method

mgol:
The previous implementation of jqLite didn't use cleanData from the jqLite
object but instead used a cached version which maede it impossible to
monkey-patch jqLite.cleanData similarly to how you can do it in jQuery.

The cleanData method is not meant to be called directly by userland code;
its purpose is mainly to be able to be monkey-patched; therefore, the previous
implementation didn't make a lot of sense.

This commit enables one of the tests so far run only with jQuery to run with
jqLite as well.

Ref #8486
Ref #8695

Fix.

`jqLite.cleanData` is not used directly so it can't be monkey-patched like its jQuery equivalent.

The opposite.

I don't think so.

**Other information**:
This is a followup to #8695.

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -303,11 +303,10 @@ function jqLiteClone(element) {
 }
 
 function jqLiteDealoc(element, onlyDescendants) {
-  if (!onlyDescendants && element.nodeType === NODE_TYPE_ELEMENT) jqLite.cleanData([element]);
+  if (!onlyDescendants && jqLiteAcceptsData(element)) jqLite.cleanData([element]);
 
   if (element.querySelectorAll) {
-    var descendants = element.querySelectorAll('*');
-    jqLite.cleanData(descendants);
+    jqLite.cleanData(element.querySelectorAll('*'));
   }
 }
 
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -201,12 +201,6 @@ function jqLiteHasData(node) {
   return false;
 }
 
-function jqLiteCleanData(nodes) {
-  for (var i = 0, ii = nodes.length; i < ii; i++) {
-    jqLiteRemoveData(nodes[i]);
-  }
-}
-
 function jqLiteBuildFragment(html, context) {
   var tmp, tag, wrap,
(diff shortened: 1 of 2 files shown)


src/jqLite.js
```diff
@@ -309,13 +303,11 @@ function jqLiteClone(element) {
 }
 
 function jqLiteDealoc(element, onlyDescendants) {
-  if (!onlyDescendants) jqLiteRemoveData(element);
+  if (!onlyDescendants && element.nodeType === NODE_TYPE_ELEMENT) jqLite.cleanData([element]);
```
gkalpak: Are we sure this doesn't break some usecases or tests?

gkalpak (commented):
Generally LGTM. My only concern is whether this might break some (obscure) tests.

src/jqLite.js
mgol: I've updated the logic.

src/jqLite.js
```diff
@@ -309,13 +303,10 @@ function jqLiteClone(element) {
 }
 
 function jqLiteDealoc(element, onlyDescendants) {
-  if (!onlyDescendants) jqLiteRemoveData(element);
+  if (!onlyDescendants && jqLiteAcceptsData(element)) jqLite.cleanData([element]);
```
mgol: @gkalpak Let me elaborate why it shouldn't break anything now. The only situation where we would fire `jqLiteRemoveData(element)` before and not fire `jqLite.cleanData([element])` here now is when `jqLiteAcceptsData(element)` is `false`. But if you then look at `jqLiteRemoveData` source, it changes anything only if `element.ng339` is defined. This property is only set to something non-`undefined` inside `jqLiteExpandoStore` and only if the second passed argument is truthy. That, in turn, happens in two places only: one is inside `jqLiteData`, the other one is in `jqLite#on`. In both cases this setter is guarded by the `jqLiteAcceptsData(element)` being truthy which contradicts our assumptions.

Therefore, there should be no observable differences to external code here.

gkalpak (approved):

PR #15853, opened 2017-03-23
docs(error/numfmt): Add priority 1 to string-to-number directive example

jmwolfe:
This directive does not work with default priority. Took forever to find out what the real issue is.

doc update

does not work in some situations (I cannot articulate exactly what conditions, but another page used priority and it made this directive work for me. I was using this in a template, with an extended JSON object assigned to ng-model after fetching from a database..

probably works in all situations

No, not that I am aware.

- [X ] The commit message follows our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format

**Other information**:

My first PR for a doc change and I didn't know about the commit comment format. Assuming the github prompting handles it for you, since there is only one line on the comment field, not entirely clear. I found the solution here: http://stackoverflow.com/questions/25657130/how-to-convert-string-to-number-or-date-in-angularjs-expression, so I can't take any claim to finding it, just for submitting to update the doc for future angular-nauts.

### DIFF
--- a/docs/content/error/ngModel/numfmt.ngdoc
+++ b/docs/content/error/ngModel/numfmt.ngdoc
@@ -41,6 +41,7 @@ directive to convert it into the format the `input[number]` directive expects.
 
   .directive('stringToNumber', function() {
     return {
+      priority: 1,
       require: 'ngModel',
       link: function(scope, element, attrs, ngModel) {
         ngModel.$parsers.push(function(value) {


jmwolfe:
I signed the CLA

Narretz:
Thanks for the PR. I feel that without an explanation, this is too confusing. Obviously the directive can be influenced by other directives that also add parsers. But if the example works on its own right now without the priority change, then I'd rather leave it as is.

gkalpak:
I agree there is no point in making this change. It will not work in all situations when priority can be an issue. There can always be another directive with a higher priority that causes this to break. This is out-of-scope for this example imo.

jmwolfe:
Well, I'd argue this is a fairly common, basic need as Javascript is not
strongly typed. Discussion abounds about Angulars lack of handling
autocasting when using <input type="number">.  Many people will attempt to
use this directive and it will not work. I can see that the priority
setting by itself is not so useful. Would a pull request adding a bit of
text at the bottom to try adding the priority: attribute if it fails to
function be more acceptable?

I am not currently using many third party directives (angular-ui and
bootstrap, though) in my app and this example didn't work without the
priority setting.

gkalpak:
I not arguing that the priority is not necessary in your setup. What I am saying is that what priority works for a usecase depends solely on the directives used. This is also true for every directive that (implicitly or explicitly) interacts with other directives. While using `priority: 1` works in your setup, a higher (or lower) priority will be needed in someone else's setup.

Adding a note that you might need to adjust the priority sounds like a better idea.

Narretz:
Closing due to inactivity. Feel free to reopen this / open a new PR with the comment about the priority.

PR #15854, opened 2017-03-23
fix($rootScope): provide correct value of one-time bindings in watchGroup

jbedard:
I don't recall how or why I came across this issue, and it's confusing me, but I found this stashed and the tests do make sense...?

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -490,9 +490,8 @@ function $RootScopeProvider() {
         }
 
         forEach(watchExpressions, function(expr, i) {
-          var unwatchFn = self.$watch(expr, function watchGroupSubAction(value, oldValue) {
+          var unwatchFn = self.$watch(expr, function watchGroupSubAction(value) {
             newValues[i] = value;
-            oldValues[i] = oldValue;
             if (!changeReactionScheduled) {
               changeReactionScheduled = true;
               self.$evalAsync(watchGroupAction);
@@ -504,11 +503,17 @@ function $RootScopeProvider() {
         function watchGroupAction() {
           changeReactionScheduled = false;
 
-          if (firstRun) {
-            firstRun = false;
-            listener(newValues, newValues, self);
-          } else {
-            listener(newValues, oldValues, self);
+          try {
+            if (firstRun) {
+              firstRun = false;
+              listener(newValues, newValues, self);
+            } else {
+              listener(newValues, oldValues, self);
+            }
+          } finally {
+            for (var i = 0; i < watchExpressions.length; i++) {
+              oldValues[i] = newValues[i];
+            }
           }
         }
 
--- a/test/ng/rootScopeSpec.js
+++ b/test/ng/rootScopeSpec.js
@@ -1097,6 +1097,94 @@ describe('Scope', function() {
       expect(log).toEqual('');
     });
 
+    it('should remove all watchers once one-time/constant bindings are stable', function() {
+      //empty
+      scope.$watchGroup([], noop);
+      //single one-time
+      scope.$watchGroup(['::a'], noop);
+      //multi one-time
+      scope.$watchGroup(['::a', '::b'], noop);
+      //single constant
+      scope.$watchGroup(['1'], noop);
+      //multi constant
+      scope.$watchGroup(['1', '2'], noop);
+
+      expect(scope.$$watchersCount).not.toBe(0);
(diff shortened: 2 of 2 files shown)


test/ng/rootScopeSpec.js
```diff
@@ -1097,6 +1097,94 @@ describe('Scope', function() {
       expect(log).toEqual('');
     });
 
+    it('should remove all watchers once one-time/constant bindings are stable', function() {
```
jbedard: This seems unrelated, but there are no other watcher-count tests so I thought I'd leave it in...

test/ng/rootScopeSpec.js
```diff
@@ -1097,6 +1097,94 @@ describe('Scope', function() {
...
+      expect(newValues).toEqual([1, 2, 3, 4]);
+      expect(oldValues).toEqual([1, 2, 2, 4]);
+
+      $rootScope.$apply('b = 4');
+      expect(newValues).toEqual([1, 2, 4, 4]);
+      expect(oldValues).toEqual([1, 2, 3, 4]);
+    }));
+  });
+
+  describe('$watchGroup'/*with logging $exceptionHandler*/, function() {
```
Narretz: Why is part of this commented out?

Narretz (commented):
This looks useful. I wonder if there are performance implications with the new try ... catch block? I vaguely remember an issue where a try / catch block was not correctly optimized in V8, I think. 
But I assume it's not performance critical as it is in the watchGroupAction fn

jbedard:
The try/catch only wraps the listener, not anything invoked per digest, so I wouldn't be too worried about it. The method is also quite simply and mainly just calls another method, so if that one method isn't optimized I don't think it's a big deal.

test/ng/rootScopeSpec.js
jbedard: idk, I guess I thought it was too long. I'll uncomment it...

jbedard: Updated

Narretz (approved):
Ok, LGTM then. Needs one other review.

src/ng/rootScope.js
```diff
@@ -504,11 +503,17 @@ function $RootScopeProvider() {
         function watchGroupAction() {
           changeReactionScheduled = false;
 
-          if (firstRun) {
-            firstRun = false;
-            listener(newValues, newValues, self);
-          } else {
-            listener(newValues, oldValues, self);
+          try {
```
gkalpak: FWIW, we could avoid the `try/finally` block, by using one extra array. But my intuition is that it won't make any difference.

test/ng/rootScopeSpec.js
```diff
@@ -1097,6 +1097,94 @@ describe('Scope', function() {
...
+      scope.$watchGroup(['1'], noop);
+      //multi constant
+      scope.$watchGroup(['1', '2'], noop);
+
+      expect(scope.$$watchersCount).not.toBe(0);
+      scope.$apply('a = b = 1');
+      expect(scope.$$watchersCount).toBe(0);
+    });
+
+    it('should maintain correct new/old values with one time bindings', inject(function($rootScope, $exceptionHandler) {
```
gkalpak: `$exceptionHandler` is not used. And you could probably use `scope` instead of `$rootScope` to avoid `inject()` altogether.

test/ng/rootScopeSpec.js
```diff
@@ -1097,6 +1097,94 @@ describe('Scope', function() {
...
+    }));
+  });
+
+  describe('$watchGroup with logging $exceptionHandler', function() {
+    it('should maintain correct new/old values even when listener throws', function() {
+      module(function($exceptionHandlerProvider) {
+        $exceptionHandlerProvider.mode('log');
+      });
+
+      inject(function($rootScope, $exceptionHandler) {
```
gkalpak: Consider using `scope` instead of `$rootScope` for consistency :grin:

gkalpak (approved):
A couple of minor clean-up comments.
LGTM otherwise.

jbedard:
Updated the tests. Will merge later today...

test/ng/rootScopeSpec.js
jbedard: I've done this in the other tests which already have a `beforeEach` that sets `scope`, but this `describe` block is new... worth adding a `beforeEach` when there's only one test?

PR #15858, opened 2017-03-24
fix($parse): standardize one-time literal vs non-literal and interceptors

jbedard:
Previously literal one-time bindings did not use the expression `inputs`, causing infinite digest issues with literal values. This often forces the use of deepEquals when watching one-time literals.

`ng-class` is one example of deepEquals which is no longer required (and had to be updated to keep all tests passing).

This one-time/literal behavior is now also consistently propagated through interceptors (and had to be updated to keep all tests passing).

----

The ngClass/interceptor part was not originally planned but had to be changed to keep tests passing. Maybe this can be split into smaller commits? But it's mainly just deleting...!

### DIFF
--- a/src/ng/directive/ngClass.js
+++ b/src/ng/directive/ngClass.js
@@ -14,13 +14,6 @@ function classDirective(name, selector) {
     return {
       restrict: 'AC',
       link: function(scope, element, attr) {
-        var expression = attr[name].trim();
-        var isOneTime = (expression.charAt(0) === ':') && (expression.charAt(1) === ':');
-
-        var watchInterceptor = isOneTime ? toFlatValue : toClassString;
-        var watchExpression = $parse(expression, watchInterceptor);
-        var watchAction = isOneTime ? ngClassOneTimeWatchAction : ngClassWatchAction;
-
         var classCounts = element.data('$classCounts');
         var oldModulo = true;
         var oldClassString;
@@ -43,7 +36,7 @@ function classDirective(name, selector) {
           scope.$watch(indexWatchExpression, ngClassIndexWatchAction);
         }
 
-        scope.$watch(watchExpression, watchAction, isOneTime);
+        scope.$watch($parse(attr[name], toClassString), ngClassWatchAction);
 
         function addClasses(classString) {
           classString = digestClassCounts(split(classString), 1);
@@ -85,9 +78,9 @@ function classDirective(name, selector) {
         }
 
         function ngClassIndexWatchAction(newModulo) {
-          // This watch-action should run before the `ngClass[OneTime]WatchAction()`, thus it
+          // This watch-action should run before the `ngClassWatchAction()`, thus it
           // adds/removes `oldClassString`. If the `ngClass` expression has changed as well, the
-          // `ngClass[OneTime]WatchAction()` will update the classes.
+          // `ngClassWatchAction()` will update the classes.
           if (newModulo === selector) {
             addClasses(oldClassString);
           } else {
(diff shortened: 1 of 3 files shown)


src/ng/parse.js
```diff
@@ -1875,41 +1876,22 @@ function $ParseProvider() {
...
-      function isAllDefined(value) {
-        var allDefined = true;
-        forEach(value, function(val) {
-          if (!isDefined(val)) allDefined = false;
-        });
-        return allDefined;
-      }
+    function isAllDefined(value) {
+      var allDefined = true;
+      forEach(value, function(val) {
```
gkalpak: Nit: Not directly related to this PR, but I think it makes sense to exit early once an undefined value is found.

jbedard: I think `value` can be an object or array which `forEach` currently handles or us. Did you have something in mind?

Since this is currently only used for object/array literal cases, we could split this into separate object/array helper methods (one using for-in one using array.every?)...

src/ng/directive/ngClass.js
```diff
@@ -97,15 +90,9 @@ function classDirective(name, selector) {
           oldModulo = newModulo;
         }
 
-        function ngClassOneTimeWatchAction(newClassValue) {
+        function ngClassWatchAction(newClassValue) {
           var newClassString = toClassString(newClassValue);
```
jbedard: I think you're correct, updated...

src/ng/parse.js
gkalpak: I agree it is not worth it. It is just for one-time literals and it will only matter if one has huge literals which don't settle quickly (a very unlikely usecase).

src/ng/directive/ngClass.js
jbedard: Unfortunately we were wrong. In the case of one-time bindings the pre-interceptor value is returned up until that value passes the one-time test: https://github.com/angular/angular.js/blob/2b0c0505e222e28449829e6f4f49826b1d1ce59e/src/ng/parse.js#L1933

I've updated it to do this `toClassString` conditionally and added a comment.

src/ng/directive/ngClass.js
```diff
@@ -97,15 +90,13 @@ function classDirective(name, selector) {
           oldModulo = newModulo;
         }
 
-        function ngClassOneTimeWatchAction(newClassValue) {
-          var newClassString = toClassString(newClassValue);
-
-          if (newClassString !== oldClassString) {
-            ngClassWatchAction(newClassString);
+        function ngClassWatchAction(newClassString) {
+          //When using a one-time binding the newClassString will return
```
gkalpak: You realy don't like spaces after `//`, do you 😛

src/ng/directive/ngClass.js
```diff
@@ -97,15 +90,13 @@ function classDirective(name, selector) {
...
+          //the pre-interceptor value until the one-time is complete
+          if (newClassString && typeof newClassString !== "string") {
```
gkalpak: Why not just `!isString(newClassString)`?
Also, I don't think it makes much difference over always applying `toClassString()` (which just does `isArray()` and `isObject()`).

Narretz (changes requested):

src/ng/parse.js
```diff
@@ -1859,6 +1859,7 @@ function $ParseProvider() {
     }
 
     function oneTimeWatchDelegate(scope, listener, objectEquality, parsedExpression, prettyPrintExpression) {
+      var doneCriteria = parsedExpression.literal ? isAllDefined : isDefined;
```
jbedard: Good idea, done.

src/ng/directive/ngClass.js
jbedard: Bad (or just different!) habit I guess :)

src/ng/directive/ngClass.js
jbedard: I've updated to `isString`. I kind of like an `if` block just so it's obvious that the comment applies to the full block, nothing more nothing less...

jbedard:
I think this is ready, PTAL.

gkalpak (approved):

Narretz (approved):

PR #15872, opened 2017-03-30
provide isError and use it in place of instanceof Error

### LINKED ISSUE #15868: provide isError and use it in place of instanceof Error
**I'm submitting a ...**

- [x] bug report
- [x] feature request

**Current behavior:**

angular sometimes stringifies unhandled rejections, even if they're errors.

**Expected / new behavior:**

Angular should pass them on to the correct excetion handler

Difficult to reproduce

**Angular version:** 1.x.y
1.6.3

**Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Chrome

**Anything else:**
Provide angular.isError:
```js
function isError(V) {
  var tagName = Object.prototype.toString.call(V);
  return (
    tagName === '[object Error]' ||
    tagName === '[object Exception]' ||
    V instanceof Error
  );
}
```
MIT from https://github.com/yefremov/iserror/blob/master/index.js

and use it wherever instanceof Error was used.

graingert:

feature/bug

Angular should pass them on to the correct excetion handler, and there's an angular.isError function now.

No

**Other information**:

re-implimented https://github.com/yefremov/iserror/blob/master/index.js to avoid license and make code faster.

Fixes #15868

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -45,6 +45,7 @@
   isNumber,
   isNumberNaN,
   isDate,
+  isError,
   isArray,
   isFunction,
   isRegExp,
@@ -672,6 +673,24 @@ function isDate(value) {
  * @returns {boolean} True if `value` is an `Array`.
  */
 var isArray = Array.isArray;
+
+/**
+ * @description
+ * Determines if a reference is an `Error`.
+ * Loosely based on https://www.npmjs.com/package/iserror
+ *
+ * @param {*} value Reference to check.
+ * @returns {boolean} True if `value` is an `Error`.
+ */
+function isError(value) {
+  var tag = toString.call(value);
+  switch (tag) {
+    case '[object Error]': return true;
+    case '[object Exception]': return true;
+    case '[object DOMException]': return true;
+    default: return value instanceof Error;
+  }
+}
 
 /**
  * @ngdoc function
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -1880,6 +1880,43 @@ describe('angular', function() {
     });
   });
 
+  describe('isError', function() {
+    function testErrorFromDifferentContext(createError) {
+      var iframe = document.createElement('iframe');
+      document.body.appendChild(iframe);
+      try {
+        var error = createError(iframe.contentWindow);
+        expect(isError(error)).toBe(true);
+      } finally {
+        iframe.parentElement.removeChild(iframe);
+      }
+    }
+
+    it('should not assume objects are errors', function() {
+      var fakeError = { message: 'A fake error', stack: 'no stack here'};
+      expect(isError(fakeError)).toBe(false);
+    });
+
+    it('should detect simple error instances', function() {
+      expect(isError(new Error())).toBe(true);
+    });
+
+    it('should detect errors from another context', function() {
+      testErrorFromDifferentContext(function(win) {
+        return new win.Error();
+      });
+    });
+
+    it('should detect DOMException errors from another context', function() {
+      testErrorFromDifferentContext(function(win) {
+        try {
+          win.document.querySelectorAll('');
+        } catch (e) {
+          return e;
+        }
+      });
+    });
+  });
 
   describe('isRegExp', function() {
     it('should return true for RegExp object', function() {
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -50,6 +50,7 @@
     "isNumber": false,
     "isNumberNaN": false,
     "isDate": false,
+    "isError": false,
     "isArray": false,
     "isFunction": false,
     "isRegExp": false,
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -3098,7 +3098,7 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
           }
           linkQueue = null;
         }).catch(function(error) {
-          if (error instanceof Error) {
+          if (isError(error)) {
             $exceptionHandler(error);
           }
         });
--- a/src/ng/log.js
+++ b/src/ng/log.js
@@ -132,7 +132,7 @@ function $LogProvider() {
     };
 
     function formatError(arg) {
-      if (arg instanceof Error) {
+      if (isError(arg)) {
         if (arg.stack && formatStackTrace) {
           arg = (arg.message && arg.stack.indexOf(arg.message) === -1)
               ? 'Error: ' + arg.message + '\n' + arg.stack
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -381,7 +381,7 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
       if (!toCheck.pur) {
         toCheck.pur = true;
         var errorMessage = 'Possibly unhandled rejection: ' + toDebugString(toCheck.value);
-        if (toCheck.value instanceof Error) {
+        if (isError(toCheck.value)) {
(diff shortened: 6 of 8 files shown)


src/Angular.js
```diff
@@ -675,6 +676,28 @@ var isArray = Array.isArray;
 
 /**
  * @ngdoc function
+ * @name angular.isError
```
gkalpak: I don't think we want to make this public.

src/Angular.js
```diff
@@ -675,6 +676,28 @@ var isArray = Array.isArray;
...
+ * @kind function
+ *
+ * @description
+ * Determines if a reference is an `Error`.
+ *
+ * @param {*} value Reference to check.
+ * @returns {boolean} True if `value` is an `Error`.
+ */
+function isError(value) {
+  var tag = ({}).toString.call(value);
```
gkalpak: We already have a closure-global `toString = Object.prototype.toString`. You can simply `toString.call(value)`.

graingert:
@gkalpak what would be a good way to construct a test that boots up an iframe? And where should I put it?

```js
function testError(createError) {
    var iframe = document.createElement('iframe');
    document.body.appendChild(iframe);
    try {
        var error = createError(contentWindow);
        expect(error instanceof Error).to.be.false();
        expect(isError(error)).to.be.true();
    } finally {
        iframe.parentElement.removeChild(iframe);
    }
}

testError(function (win) {
    return new win.Error();
});

testError(function (win) {
    try {
        win.document.querySelectorAll('');
    catch(e) {
        return e;
    }
});
```

test/minErrSpec.js
```diff
@@ -17,6 +17,7 @@ describe('minErr', function() {
   it('should return an Error factory', function() {
     var myError = testError('test', 'Oops');
     expect(myError instanceof Error).toBe(true);
+    expect(isError(myError)).toBe(true);
```
gkalpak: This seems irrelevant here. Please remove.

gkalpak (commented):
A couple of minor comments. Looks OK otherwise (except for the missing tests).
The tests can go in https://github.com/graingert/angular.js/blob/af80f4765e2a9e795f64103b6e15770cf6eda050/test/AngularSpec.js.

graingert:
What's wrong with publishing it?

gkalpak:
> What's wrong with publishing it?

We have to maintain it in a backwards compatible way. If it is not public we can change/improve/remove it without any concerns.

graingert:
How do you not publish it?

gkalpak:
> How do you not publish it?

Remove the `@ngdoc` anotation (and others that are not relevant/necessary: `@name`, `@module ng`, `@kind`).
(It won't be available anyway, so we don't want to have it appear on the docs either.)

src/ngMock/angular-mocks.js
```diff
@@ -984,7 +984,7 @@ angular.mock.dump = function(object) {
     } else if (angular.isObject(object)) {
       if (angular.isFunction(object.$eval) && angular.isFunction(object.$apply)) {
         out = serializeScope(object);
-      } else if (object instanceof Error) {
+      } else if (angular.isError(object)) {
```
graingert: a publishes isError is required here.

test/AngularSpec.js
```diff
@@ -1861,6 +1861,44 @@ describe('angular', function() {
...
+    it('should not assume objects are errors', function() {
+      expect(isError({ message: 'A fake error', stack: 'no stack here'}))
+        .toBe(false);
+    });
+
+    it('should detect simple error instances', function() {
+      expect(isError(new Error())).toBe(true);
+    });
+
+    function testError(createError) {
```
gkalpak: Move this either at the top or at the bottom of the `describe` block (or even create a nested `describe` for "other-context" tests) and give it a more descriptive name (e.g. `testErrorFromDifferent context`.

Maybe having a separate `describe` block with `beforeAll`/`afterAll` blocks for creating/destorying the iframe is a good idea.

src/ngMock/angular-mocks.js
gkalpak: We could make it "private API" (e.g. `angular.$$isError`), but I don't think it is worth it. This function (`angular.mock.dump`) is very rarely used. It is also very rare to have new window contexts in unit tests and it is even more rare to have them throw errors. Plus, `DOMException`s get decently stringified by `JSON.stringify` (in contrast to plain `Error`s which become `"{}"`). All that considered, I think this one occurrence can remain `object instanceof Error` :smiley:

gkalpak (changes requested):

test/AngularSpec.js
graingert: try { } finally { } is more fine grained in this usecase

gkalpak: What do you mean?

graingert: it's specific to the actual call to the testErrorFromDifferentContext function. Like the lifetime of the iframe is shorter than a test, no point leaving it hanging around until afterAll/beforeAll is called.

gkalpak: I am fine either way, so feel free to leave things as they are :smiley:

graingert: ok

graingert:
@gkalpak ok, I'm pretty I've resolved anything.  Interestingly enough if seems as though Safari thinks DOMExceptions from frames are instances of Errors

test/AngularSpec.js
```diff
@@ -1861,6 +1861,43 @@ describe('angular', function() {
...
+        var error = createError(iframe.contentWindow);
+        expect(isError(error)).toBe(true);
+      } finally {
+        iframe.parentElement.removeChild(iframe);
+      }
+    }
+
+    it('should not assume objects are errors', function() {
+      expect(isError({ message: 'A fake error', stack: 'no stack here'}))
+        .toBe(false);
```
gkalpak: Nit: Unnecessarily wrapped line.

gkalpak (approved):
LGTM

gkalpak:
(The commit message needs some love (as it doesn't follow [our guidelines][1]).)

petebacondarwin (approved):

graingert:
I'd like to add DOMError too

gkalpak:
I am going to merge this as is, but feel free to add `DOMError` in another PR :grin:

graingert:
<del>I'd like to add DOMError too</del>

scratch that, DOMError only has a 'name' property and never has a stack, so it shouldn't be added imhop.

gkalpak:
Actually, before I merge it can you take care of the following:
1. Fix https://github.com/angular/angular.js/pull/15872#discussion_r113686119.
2. Fix the commit message to follow the commit message guidelines.
3. Add a comment above the function pointing to the original project.

graingert:
> Fix the commit message to follow the commit message guidelines.

what message should I use?

gkalpak:
Something that follows [the guidelines][1]. E.g. something like:

> fix(*): correctly detect Error instances from different contexts 
>
> 12345678901234567890123456789012345678901234567890123456789012345678901234567890
> Errors thrown from different contexts (such as an iframe or webworker) are not detected as Error
> instances and handled accordingly.
> This commit fixes it by introducing an `isError()` helper, that is able to correctly detect such
> instances.
> 
> Fixes #15868 

graingert:
@gkalpak any updates on this?

PR #15877, opened 2017-04-01
feat($rootScope): Implement stopPropatagion for $broadcast events

StrangelyTyped:
Feature

$scope.$broadcast events cannot be cancelled and do not have a stopPropagation() function/method

This change allows any scope to prevent it's children from receiving that
event by calling stopPropagation on the event object. Other listeners on the scope which called
stopPropagation will continue to receive the event, as will siblings of that scope and any
children of those siblings.

No

**Other information**:

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -1151,8 +1151,8 @@ function $RootScopeProvider() {
        *   - `currentScope` - `{Scope}`: the scope that is currently handling the event. Once the
        *     event propagates through the scope hierarchy, this property is set to null.
        *   - `name` - `{string}`: name of the event.
-       *   - `stopPropagation` - `{function=}`: calling `stopPropagation` function will cancel
-       *     further event propagation (available only for events that were `$emit`-ed).
+       *   - `stopPropagation` - `{function}`: calling `stopPropagation` function will cancel
+       *     further event propagation.
        *   - `preventDefault` - `{function}`: calling `preventDefault` sets `defaultPrevented` flag
        *     to true.
        *   - `defaultPrevented` - `{boolean}`: true if `preventDefault` was called.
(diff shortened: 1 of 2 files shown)


StrangelyTyped:
I signed it!

StrangelyTyped:
I'm aware (though didn't spot it until after I'd already done mine) that there has been a previous PR for this functionality [here](https://github.com/angular/angular.js/pull/12672).

I don't care which of these is preferred - mine or just-boris' - but I would like to see one of them accepted for the following reasons:

* This change brings $broadcast and it's dispatched event object into line with $emit and it's event.
* $broadcast-ed events and $emit-ed events are both dispatched through $on, but at present only one has a stopPropagation method, meaning $on needs to be aware of whether or not the event came from above via $broadcast or below via $emit.
* Any arguments for needing stopPropagation on $emit-ed events also applies to $broadcast.
* Given the angular movement towards component-heirarchy-based applications it is likely to become more common to need to intercept broadcast events to handle them at a higher level and either prevent the default behaviour or broadcast a modified event instead.
* The established 'workaround' for lack of stopPropagation is to use preventDefault, but this only works if the author has control over both event handlers to check preventDefault which is not always the case.

src/ng/rootScope.js
```diff
@@ -1322,8 +1329,8 @@ function $RootScopeProvider() {
           // Insanity Warning: scope depth-first traversal
           // yes, this code is a bit crazy, but it works and we have tests to prove it!
           // this piece should be kept in sync with the traversal in $digest
-          // (though it differs due to having the extra check for $$listenerCount)
-          if (!(next = ((current.$$listenerCount[name] && current.$$childHead) ||
+          // (though it differs due to having the extra check for $$listenerCount and stopPropagation)
+          if (!(next = ((current.$$listenerCount[name] && !stopPropagation && current.$$childHead) ||
               (current !== target && current.$$nextSibling)))) {
```
Narretz: PR #12672 has an additional check here, why not this PR?

Narretz:
Personally, I'm kind of indifferent about this change, with the following caveat:

The scope event system in AngularJS is kind of bolted on the scope system and should not be a cornerstone of application design. Componentized app architecture should imo rely on the bindings as clearer interfaces - events rely on scopes, component bindings abstract them. So  I don't think component archtitecture creates a big need for this - otherwise it would have been requested before and more loudly.

src/ng/rootScope.js
StrangelyTyped: Actually it doesn't, it has an assignment here, since it's not part of the condition and sin e the condition is already confusing enough i put mine on line 1310

3imed-jaberi (approved):

PR #15881, opened 2017-04-02
fix(angular-loader): do not depend on "closure" globals that may not be available

### LINKED ISSUE #15880: fix(angular-loader): do not depend on "closure" globals that are not available in `angular-loader`
**I'm submitting a ...**
- [x] bug report
- [ ] feature request

**Current behavior:**
Commits such as cf43ccdf9 and a0641ea47 (possibly others) have changed `stringify.js` in ways that do not take into account the fact that it is also included in `angular-loader` where it does not have access to all th necessary "closure" globals available in `angular.js`.
Fortunately, this bug only appears when there are errors to be thrown.

**Expected / new behavior:**
Files that are included in `angular-loader` should not break by trying to access properties/functions that are not available in `angular-loader`.

**Angular version:** At least 1.3.3+

**Browser:** all

gkalpak:
Bug fix.

See #15880.

Code that is distributed as part of both `angular.js` and `angular-loader.js` does not depend on "closure" globals that may not be available in `angular-loader`.

No.

**Other information**:
Fixes #15880.

### DIFF
--- a/src/loader.prefix
+++ b/src/loader.prefix
@@ -7,5 +7,24 @@
 (function() {
     function isFunction(value) {return typeof value === 'function';}
     function isDefined(value) {return typeof value !== 'undefined';}
+    function isNumber(value) {return typeof value === 'number';}
     function isObject(value) {return value !== null && typeof value === 'object';}
+    function isScope(obj) {return obj && obj.$evalAsync && obj.$watch;}
+    function isWindow(obj) {return obj && obj.window === obj;}
+    function sliceArgs(args, startIndex) {return Array.prototype.slice.call(args, startIndex || 0);}
+    function toJsonReplacer(key, value) {
+      var val = value;
+
+      if (typeof key === 'string' && key.charAt(0) === '$' && key.charAt(1) === '$') {
+        val = undefined;
+      } else if (isWindow(value)) {
+        val = '$WINDOW';
+      } else if (value &&  window.document === value) {
+        val = '$DOCUMENT';
+      } else if (isScope(value)) {
+        val = '$SCOPE';
+      }
+
+      return val;
+    }
 
--- a/src/loader.prefix
+++ b/src/loader.prefix
@@ -5,11 +5,16 @@
  */
 'use strict';
 (function() {
+    // NOTE:
+    // These functions are copied here from `src/Angular.js`, because they are needed inside the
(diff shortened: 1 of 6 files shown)


src/stringify.js
```diff
@@ -9,7 +13,10 @@ function serializeObject(obj, maxDepth) {
   // and a very deep object can cause a performance issue, so we copy the object
   // based on this specific depth and then stringify it.
   if (isValidObjectMaxDepth(maxDepth)) {
-    obj = copy(obj, null, maxDepth);
+    if (!copyFn) {
```
Narretz: Shouldn't `angular.copy` be always defined here?

src/loader.prefix
```diff
@@ -7,5 +7,24 @@
 (function() {
     function isFunction(value) {return typeof value === 'function';}
     function isDefined(value) {return typeof value !== 'undefined';}
+    function isNumber(value) {return typeof value === 'number';}
     function isObject(value) {return value !== null && typeof value === 'object';}
+    function isScope(obj) {return obj && obj.$evalAsync && obj.$watch;}
+    function isWindow(obj) {return obj && obj.window === obj;}
+    function sliceArgs(args, startIndex) {return Array.prototype.slice.call(args, startIndex || 0);}
+    function toJsonReplacer(key, value) {
```
Narretz: Should we add a note where this and other functions are used? This makes it easier later to check if they are still necessary after refactorings etc.

Narretz (approved):
A few comments, otherwise LGTM

src/loader.prefix
gkalpak: You mean add a comment here? Or in the places where these are used?

Narretz: In this file - a note at the top is probably enough, saying something of the effect that these functions are defined here because they must be available in angular-loader without angular.js loaded, and that they are copied from the Angular.js file

gkalpak:
Rebased, addressed comments and [fixed a bug][1]. Will merge as soon as Travis is happy too.

[1]: https://github.com/angular/angular.js/pull/15881/commits/c7e48c05b3b6178e61a95ababc1dc2d13c2eeef4#diff-c2636daee36d92d1958c5fcc6cbd0554R17

Narretz:
LGTM. Looks like nobody uses the loader ...

Issue #11387, opened 2015-03-21
Why does Angular need its own angular.uppercase and lowercase methods?


thorn0:
What's wrong with the standard `toUpperCase` and `toLowerCase` methods?
Looks like there was some issue with the Turkish locale, but does it still exist in the browsers supported by Angular? I [asked](http://stackoverflow.com/questions/28792027/in-which-exactly-js-engines-are-tolowercase-touppercase-locale-sensitive) about it on StackOverflow, but didn't get an answer.

realityking:
The reason is documented in a source comment: https://github.com/angular/angular.js/blob/e5d1d6587d1f38898bca673a829a0bc0ffaaf2c8/src/Angular.js#L161-L163

It's also documented on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase

Lastly, I think the StackOverflow answers were pretty accurate.

thorn0:
Neither the comment nor the SO answers are informative.
The question is "In which exactly JS engines are toLowerCase() & toUpperCase() locale-sensitive?", right? How do those answers answer it? How does the source comment answer it? Moreover, the comment and the answers contradict each other. In theory, toLowerCase() & toUpperCase() shouldn't be locale-sensitive. That's what one of the answers says. But the comment states something very opposite:
> String#toLowerCase and String#toUpperCase don't produce correct results in browsers with Turkish locale
Again, in which exactly browsers does this happen?

wesleycho:
Searching on Google, this seems like a general problem due to the Turkish language itself - this article seems to explain it: http://www.i18nguy.com/unicode/turkish-i18n.html

If I had to guess, this is a problem with every browser.

thorn0:
The general problem exists, that's true. But who said that it (still) exists also in JavaScript? Who faced it during say last 10 years? Nobody knows for sure, just rumours and guesses. What a great engineering approach.

wesleycho:
I doubt it was pulled from thin air - that is not how the Angular team functions.

thorn0:
The hacks and workarounds for old IE8 are being removed from the code of Angular now. But this thing looks like a workaround for a bug in some even older browser.

pkozlowski-opensource:
@thorn0 why not testing on your end (you seem to be from Turkey) and sending a PR to remove those work-arounds if no longer necessary? I'm pretty sure that everyone would be more than happy to remove unnecessary code.

[...]

PR #15890, opened 2017-04-05
chore(*): remove manualLowercase & manualUppercase functions

mgol:
The `manualLowercase` & `manualUppercase` functions were inspired by Google Caja
code. Caja is written in Java, though, where problems with `toLowerCase`
working differently in Turkish locale are well known[1]. In JavaScript
`String#toLowerCase` is defined in the ECMAScript spec and all implementations
are required to lowercase I in the same way, regardless of the current locale.
Differences may (and do) happen only in `String#toLocaleLowerCase`.

Other libraries doing string normalization, like jQuery or DOMPurify don't
apply special lowercasing logic in a Turkish environment.

Therefore, the `manualLowercase` & `manualUppercase` logic is dead code in
AngularJS and can be removed.

Also, the `manualLowercase` & `manualUppercase` functions are incomplete; they
only lowercase ASCII characters which is different to native
`String#toLowerCase`. Since those functions are used in many places in the
library, they would break a lot of code. For example, the lowercase filter would
not lowercase Ω to ω but leave it as Ω.

[1] https://garygregory.wordpress.com/2015/11/03/java-lowercase-conversion-turkey/

Ref #11387

Dead code removal.

The `manualLowercase` & `manualUppercase` functions are defined & used if `'I'.toLowerCase() !== 'i'`.

`String#toLowerCase` is used everywhere.

No.

**Other information**:

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -28,8 +28,6 @@
     "REGEX_STRING_REGEXP" : false,
     "lowercase": false,
     "uppercase": false,
-    "manualLowercase": false,
-    "manualUppercase": false,
     "isArrayLike": false,
     "forEach": false,
     "forEachSorted": false,
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -21,8 +21,6 @@
 
   lowercase,
   uppercase,
-  manualLowercase,
-  manualUppercase,
   nodeName_,
   isArrayLike,
   forEach,
@@ -146,31 +144,6 @@ var lowercase = function(string) {return isString(string) ? string.toLowerCase()
  * @returns {string} Uppercased string.
  */
 var uppercase = function(string) {return isString(string) ? string.toUpperCase() : string;};
-
-
-var manualLowercase = function(s) {
-  /* eslint-disable no-bitwise */
-  return isString(s)
-      ? s.replace(/[A-Z]/g, function(ch) {return String.fromCharCode(ch.charCodeAt(0) | 32);})
-      : s;
-  /* eslint-enable */
-};
-var manualUppercase = function(s) {
-  /* eslint-disable no-bitwise */
-  return isString(s)
-      ? s.replace(/[a-z]/g, function(ch) {return String.fromCharCode(ch.charCodeAt(0) & ~32);})
-      : s;
-  /* eslint-enable */
-};
-
-
-// String#toLowerCase and String#toUpperCase don't produce correct results in browsers with Turkish
-// locale, for this reason we need to detect this case and redefine lowercase/uppercase methods
-// with correct but slower alternatives. See https://github.com/angular/angular.js/issues/11387
-if ('i' !== 'I'.toLowerCase()) {
-  lowercase = manualLowercase;
-  uppercase = manualUppercase;
-}
 
 
 var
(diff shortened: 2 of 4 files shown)


Narretz (approved):
LGTM. Finally resolved :)

Narretz:
ios8 fails with something completely unrelated: 
>>> Mobile Safari 8.0.0 (iOS 10.10.5) ngAnimate $$animateCssDriver when active anchored animations should continue the anchor animation by seeding the to styles based on where the final anchor element will be positioned FAILED

mgol:
I need a little more time for this, removing milestone.

mgol:
I'll look into it tomorrow. This should effectively be removing dead code so it may even land in a patch release, though. We could aim for 1.7.0 just in case, to be safe, as it's easier to add it back later than remove. But it shouldn't really matter, I think.

mgol:
I summarized it a little and wrote a summary to Mario from Cure53 & Pete. The gist of it:

ngSanitize has a test suite at https://github.com/angular/angular.js/tree/master/test/ngSanitize, in particular, https://github.com/angular/angular.js/blob/master/test/ngSanitize/sanitizeSpec.js tests the HTML parser.

The thing is, our custom toLowerCase & toUpperCase functions are only used when the lowercase "i" is not what you get when running toLowerCase() on an uppercase "I":
https://github.com/angular/angular.js/blob/v1.6.9/src/Angular.js#L182-L188
Otherwise we use regular toLowerCase & toUpperCase. Our custom implementations are very different from the native methods, though, as they only lowercase & uppercase the A-Z & a-z ranges so if they were used in some scenarios they could lead to bugs.

The ECMAScript spec doesn't allow "I".toLowerCase() to give different results for different locales:
https://www.ecma-international.org/ecma-262/8.0/#sec-string.prototype.tolowercase
The toLocaleLowerCase method is the one that may differ and the Turkish language is even directly mentioned in the spec:
https://www.ecma-international.org/ecma-262/8.0/#sec-string.prototype.tolocalelowercase
"There will only be a difference in the few cases (such as Turkish) where the rules for that language conflict with the regular Unicode case mappings."

This looks like code backported from Java where such a difference exists and is a well-known language bug: http://mattryall.net/blog/2009/02/the-infamous-turkish-locale-bug. All search queries I did for this issue in JavaScript either led to articles about Java or to questions about this Angular.js implementation.

I've run tests in a macOS VM with OS language & browser languages set to Turkish and there was no difference in behavior there in Safari, Chrome & Firefox compared to the English locale when it comes to "i" & toLowerCase on "I".

AFAIK Angular 2+ doesn't use any custom toLowerCase implementation for the Turkish language.

In light of all that, do you think we can remove the supposedly unused fallback functions from Angular.js, Mario?

There was an issue (from a Turkish user) asking about the necessity of those custom implementations and Misko replied in:
https://github.com/angular/angular.js/issues/11387#issuecomment-287970029
that it was added "request of google security team" but I'm not sure if that's a remark about Angular.js or the Java implementation it was backported from.
Pete, could we perhaps consult a Google Security person just to make sure?

cc @petebacondarwin @cure53

gkalpak (approved):

petebacondarwin (approved):

PR #15906, opened 2017-04-10
docs($http): correct and clarify default transforms

### LINKED ISSUE #15897: $http is parsing a request to JSON ignoring the content-type
**I'm submitting a ...**
- [X] bug report
- [ ] feature request

**Current behavior:**
The transformResponse is being executed based on a text/plain content, instead of the content-type headers, causing errors sometimes.

**Expected / new behavior:**
The transformResponse has to be called based on the content-type headers.

$http.get is trying parse to json a text/plain response (the content-type header from the server is "text/plain") when the text starts with "[" and ends with "]" characters.

Looking the documentation for [baddata](https://docs.angularjs.org/error/$http/baddata) it says that it parses the body when the content-type header is application/json, and "the response looks like a valid JSON-stringified object or array".

And when I change the default transformResponse function to:
`
$http.get('/', {
      transformResponse: [function (data) {
          return data;
      }]
    })
`
To do nothing when it's called, the request works.

The $http request ignores the content-type response header and try to parse my response based in the body of the response?

**Angular version:** 1.6.2

**Browser:** all

**Anything else:**
I'm not sure how it detects that the body "looks like" a json to parse it automatically. But if the content-type of the response is "text/plain" it doesn't have to parse it, even if the body is a json object (not my case).

Narretz:
- baddata error described incorrect http behavior, and workarounds
- httpProvider defaults were missing transformResponse / transformRequest
- http was not clear about JSON detection strategy

Closes #15897


src/ng/http.js
```diff
@@ -280,15 +274,38 @@ function $HttpProvider() {
...
+   *
+   *   - If the `data` property of the request configuration object contains an object, serialize it
+   *     into JSON format.
+   *
+   * - **`defaults.transformResponse`** -
+   * `{Array<function(data, headersGetter, status)>|function((data, headersGetter, status))}` -
+   * An array of functions (or a single function) which are applied to the response data. By default,
+   * this is an array which applies one response transformation function that does two things:
+   *
+   *  - If XSRF prefix is detected, strip it
```
gkalpak: I don't think "XSRF prefix" is accurate. The prefix won't prevent an XSRF afaict, it will just prevent a potential XSRF vulnerability to be combined with other vulnerabilities.

_EDIT:_ I see you've copied it from elsewhere.  I'm not 100% sure, so I don't feel strongly about it :grin:

gkalpak (commented):
:100:

src/ng/http.js
Narretz: I was wondering about this too .. let's keep it for now

src/ng/http.js
```diff
@@ -280,15 +274,38 @@ function $HttpProvider() {
...
+   * - **`defaults.transformRequest`** -
+   * `{Array<function(data, headersGetter)>|function(data, headersGetter)}` -
+   * An array of functions (or a single function) which are applied to the request data.
+   * By default, this is an array with one request transformation function:
+   *
+   *   - If the `data` property of the request configuration object contains an object, serialize it
+   *     into JSON format.
+   *
+   * - **`defaults.transformResponse`** -
+   * `{Array<function(data, headersGetter, status)>|function((data, headersGetter, status))}` -
```
Narretz: Probably a typo

Issue #15907, opened 2017-04-12
linky filter does not handle certain punctuation well


ppham27:
**I'm submitting a ...**

- [x] bug report
- [ ] feature request
- [ ] other (Please do not submit support requests here (see above))

**Current behavior:**

Links like `(www.example.com):` include `):` as part of the link when it is punctuation.

**Expected / new behavior:**

`):` should not be part of the link.

**Minimal reproduction of the problem with instructions:**

Type `(www.example.com):`  in the example https://docs.angularjs.org/api/ngSanitize/filter/linky

**Angular version:** 1.6.4

**Browser:** [all]

**Anything else:**

gkalpak:
Closing as duplicate of #7213.

gkalpak:
I just realized this something different than #7213. Yet, the main problem is discussed in #7213, which is that `linky` filter currently disallows certain characters that is should allow (per the spec), but aligning it with the spec would break people's expectations wrt to links in plain text.

There is also #13699, which discusses a more general problem, which is that the RegExp used in `linky` is pretty broken anyway :grin:

I'll keep this closed. Let's move the discussion on #7213 and/or #13699.

Issue #13699, opened 2016-01-07
Linky filter ignores invalid URL characters


dchacke:
I tried to parse links in a string and I noticed that the links got way too long because they only ever stopped when a space was encountered. In my case, they should have stopped earlier at an escaped quote (`\"`), but I suppose there should be other special characters that properly delimit a URL from the rest of the string.

- **Reproducible**: Sometimes (with text that contains links and delimits them with characters that are not URL safe)
- **Browsers**: Chrome 47.0.2526.106 (64-bit), Safari version 9.0 (11601.1.56)
- **Operating system**: OS X El Capitan 10.11
- **Angular version**: 1.4.7
- **Potential location of issue**: https://github.com/angular/angular.js/blob/master/src/ngSanitize/filter/linky.js#L134
- **Example/how to reproduce**: https://regex101.com/r/zV5zJ4/6

This example includes my specific case where I tried to parse some stringified JSON and the recognized link is too long. I also put some special characters in there that should not be allowed, which I base on http://stackoverflow.com/a/1547940/1371131.

Forgive me if my reproduction of the regex is wrong. I haven't used regexes much, nor do I know much about the exact regex format URLs need to follow, so take it with a grain of salt. The issue I encountered seemed weird though.

Narretz:
Thanks for the detailed explanation. The linky regex is currently pretty weak. The thing is that I'm not sure if `\` is really disallowed. The same SO question has an answer where it is said, that `\` is "unwise" but not forbidden. In any case, I don't really want to complicate the linky regex too much .... I wonder what @gkalpak's opinion is on this, he has recently dealt with the mail regex.

gkalpak:
Yes, the `LINKY_URL_REGEXP` seems pretty broken. Not sure how much of it can be fixed without breaking existing apps. I'll look into some more...

PR #15908, opened 2017-04-12
fix(linky): better handle punctuation

ppham27:
Fixes issue #15907

Bug fix

Colons aren't handled properly.

Colons are considered punctuation.

No.

**Other information**:

### DIFF
--- a/src/ngSanitize/filter/linky.js
+++ b/src/ngSanitize/filter/linky.js
@@ -129,7 +129,7 @@
  */
 angular.module('ngSanitize').filter('linky', ['$sanitize', function($sanitize) {
   var LINKY_URL_REGEXP =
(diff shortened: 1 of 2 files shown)


gkalpak:
Can you change the commit message to include `Fixes #7213`?

gkalpak:
Actually my previous comment is invalid. #7213 is not directly related to this. But there are more fundamental issues with the RegExp used in `linky` (see https://github.com/angular/angular.js/issues/15907#issuecomment-293542109 for more details), which this PR does not address. E.g. it will still fail with `www.example.com):This is bad`, matching `www.example.com):This` as the URL).

I am going to close this PR, but I would be more than happy to work with you on improving this RegExp (which will probably involve a breaking change and a lot of tests :grin:).
Let's move the discussion to #13699.

PR #15911, opened 2017-04-12
refactor(*): remove workarounds for IE <9, log all parameters in IE 9

mgol:
1. Remove remanining workarounds for IE <9
2. Log all parameters in IE 9, not just the first two.
3. Update IE/Edge-related support comments.
4. ES6 classes now require Edge 14 or newer to work.

A refactor.

Some code targeting IE <=8 still exists in the code base, support comments don't account for Edge 15.

The opposite. Also, `$log` is now able to log more than 2 parameters on IE and all browsers use one code path for passing parameters to a proper logging function.

No, unless we consider breaking support for classes in Edge 13 (current version is 15) a breaking change.

**Other information**:

### DIFF
--- a/test/ng/logSpec.js
+++ b/test/ng/logSpec.js
@@ -101,24 +101,29 @@ describe('$log', function() {
       })
     );
 
-    it('should not attempt to log the second argument in IE if it is not specified', inject(
-      function() {
-        log = function(arg1, arg2) { logger += 'log;' + arg2; };
-        warn = function(arg1, arg2) { logger += 'warn;' + arg2; };
-        info = function(arg1, arg2) { logger += 'info;' + arg2; };
-        error = function(arg1, arg2) { logger += 'error;' + arg2; };
-        debug = function(arg1, arg2) { logger += 'debug;' + arg2; };
-      },
-      removeApplyFunctionForIE,
-      function($log) {
-        $log.log();
-        $log.warn();
-        $log.info();
-        $log.error();
-        $log.debug();
-        expect(logger).toEqual('log;warn;info;error;debug;');
-      })
-    );
+    // Support: Safari 9.1 only, iOS 9.3 only
+    // For some reason Safari thinks there is always 1 parameter passed here.
+    if (!/\b9\.\d(\.\d+)* safari/i.test(window.navigator.userAgent) &&
+      !/\biphone os 9_/i.test(window.navigator.userAgent)) {
+      it('should not attempt to log the second argument in IE if it is not specified', inject(
+        function() {
+          log = function(arg1, arg2) { logger += 'log,' + arguments.length + ';'; };
+          warn = function(arg1, arg2) { logger += 'warn,' + arguments.length + ';'; };
+          info = function(arg1, arg2) { logger += 'info,' + arguments.length + ';'; };
+          error = function(arg1, arg2) { logger += 'error,' + arguments.length + ';'; };
+          debug = function(arg1, arg2) { logger += 'debug,' + arguments.length + ';'; };
+        },
+        removeApplyFunctionForIE,
+        function($log) {
+          $log.log();
+          $log.warn();
+          $log.info();
+          $log.error();
+          $log.debug();
+          expect(logger).toEqual('log,0;warn,0;info,0;error,0;debug,0;');
+        })
+      );
+    }
   });
 
   describe('$log.debug', function() {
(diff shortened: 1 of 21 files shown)


Narretz:
I think this should be split into 3 commits:
- IE8 removal refactor
- IE9 log change
- Edge class change

I think it's acceptable that we only support Edge 15 for classes, especially since current FF is still choking on them, too.

mgol:
I can separate IE 9 log changes & Edge class fallback removal to teo
separate small commits. The other part is not just about removing IE 8-
stuff but also updating support comments related to IE/Edge. Would you like
to separate that as well? Those two have some overlap.
-- 
Michał Gołębiowski

test/ng/logSpec.js
```diff
@@ -101,24 +101,29 @@ describe('$log', function() {
...
-        $log.log();
-        $log.warn();
-        $log.info();
-        $log.error();
-        $log.debug();
-        expect(logger).toEqual('log;warn;info;error;debug;');
-      })
-    );
+    // Support: Safari 9.1 only, iOS 9.3 only
+    // For some reason Safar thinks there is always 1 parameter passed here.
```
gkalpak: _Safar_ --> _Safari_

test/ng/logSpec.js
```diff
@@ -101,24 +101,29 @@ describe('$log', function() {
...
+    if (!/\b9\.\d(\.\d+)* safari/i.test(window.navigator.userAgent) &&
+      !/\biphone os 9_/i.test(window.navigator.userAgent)) {
+      it('should not attempt to log the second argument in IE if it is not specified', inject(
```
gkalpak: _in IE_ --> _in ..._

test/ng/logSpec.js
```diff
@@ -101,24 +101,29 @@ describe('$log', function() {
...
+        function() {
+          log = function(arg1, arg2) { logger += 'log,' + arguments.length + ';'; };
+          warn = function(arg1, arg2) { logger += 'warn,' + arguments.length + ';'; };
+          info = function(arg1, arg2) { logger += 'info,' + arguments.length + ';'; };
+          error = function(arg1, arg2) { logger += 'error,' + arguments.length + ';'; };
+          debug = function(arg1, arg2) { logger += 'debug,' + arguments.length + ';'; };
+        },
+        removeApplyFunctionForIE,
```
gkalpak: _ForIE_ 😞

gkalpak (commented):
A couple of nits. Looks good otherwise.
(:+1: for breaking up into separate commits.)

test/ng/logSpec.js
mgol: What do you mean?

gkalpak: Nothing. I misinterpreted the change.
Please ignore :grin:

test/ng/logSpec.js
mgol: This test is a little weird. It's supposed to simulate the IE logging behavior (i.e. missing `apply` on `console` methods). Normally we'd just run those tests in IE without patching them but to be able to mock those methods we have to replace them with mocks that are regular functions so, in turn, we have to delete the methods.

Is there any way to reduce the confusion here?

mgol:
PR updated & split into 3 commits.

test/ng/logSpec.js
gkalpak: After taking a look at the whole `describe` block, I think it's fine.
I think we could now have a test that all arguments are logged (not just the first two as used to happen).

gkalpak (approved):
One minor suggestion, but LGTM otherwise.

Narretz (approved):
LGTM. How did you figure out which IE workarounds can be removed? Because many only say "In IE", without specifying a version

mgol:
> How did you figure out which IE workarounds can be removed?

I tested them all manually in various IE/Edge versions.

test/ng/logSpec.js
mgol: @gkalpak I think we may even want to wrap the whole file in this block removing `apply` to simulate IE to have two test blocks - one with apply and another one without it; just to make sure there is now no observable difference in logging behavior with regular parameters in IE & non-IE.

I could do this cleanup in a separate PR, I don't want to drag this one further.

petebacondarwin (approved):
Seems secure...

Issue #14881, opened 2016-07-07
Possible improvement to AngularJS orderBy.js


OpenSiteMobile:
As best I understand, the orderBy.js file was improved in Angular version 1.5.7 to deal with "tiebreakers" using the getComparisonObject function:

"NOTE: We are adding an extra `tieBreaker` value based on the element's index."

But I also noticed that the doComparison function which "compares" the values, uses the "compare" function defined under the orderByFilter function. Since the orderByFilter function can use a developer supplied comparison function (which would probably not be aware of the comparison tiebreaker code), would it not be better to have the doComparison function written as this:

```
    function doComparison(v1, v2) {
      for (var i = 0, ii = predicates.length; i < ii; i++) {
        var result = compare(v1.predicateValues[i], v2.predicateValues[i]);
        if (result) {
          return result * predicates[i].descending * descending;
        }
      }

      return defaultCompare(v1.tieBreaker, v2.tieBreaker) * descending;
    }
```

...so that the tiebreaker feature is always available regardless of whether or not the developer knew about it.

Current behavior:

Would ingnore the tieBreaker code from getComparisonObject function if the developer supplies his/her own comparison function.

Expected behavior:

Although the use case where this makes a difference may be small, adding the above change would always force the "tieBreaker" code to work where applicable. The downside might be upexpected results from developers not aware of the new "tieBreaker" comparison feature when upgrading to this version of AngularJS.

Angular version 1.5.7 is the first version I noticed the code for "tieBreaker" and should be browser independent, ref. file:

https://github.com/angular/angular.js/blob/master/src/ng/filter/orderBy.js

gkalpak:
> As best I understand, the orderBy.js file was improved in Angular version 1.5.7 to deal with "tiebreakers"

Yes and no.
Yes it was improved (adding support for custom comparators, improving docs, cleaning up the code a bit).
No, the tie-breaker feature was not introduced in 1.5.7. It has been there for a while. We just documented it more explicitly, because now that a custom comparator can be specified, the user needs to be aware of it.

--
> Since the orderByFilter function can use a developer supplied comparison function (which would probably not be aware of the comparison tiebreaker code)

That's why we documented it, so that the developer is aware :smiley:

--
> would it not be better to have the doComparison function written as this: [...]

Not really, because the developer might want to supply custom tie-breaker logic.

--
> Current behavior: Would ingnore the tieBreaker code from getComparisonObject function if the developer supplies his/her own comparison function.

[...]

Issue #14481, opened 2016-04-21
Spacing in ng-repeat using key, value pair


LeftPinkie:
I'm using AngularJS v1.2.15, Firefox 45.

I am experiencing a bug with ng-repeat using key/value pair.

The error message returned is `ngRepeat:iidexp`

These will generate the error (space next to either of the parenthesis):
```
<div ng-repeat='( group, products) in FeaturedProducts'>
...
</div>

<div ng-repeat='(group, products ) in FeaturedProducts'>
...
</div>

<div ng-repeat='( group, products ) in FeaturedProducts'>
...
</div>
```

These will work (no space next to either of the parenthesis):
```
<div ng-repeat='(group, products) in FeaturedProducts'>
...
</div>

<div ng-repeat='(group , products) in FeaturedProducts'>
...
</div>
```

My object is:
```
FeaturedProducts => {
  "Test 1": [
    {
      "id": "1A",
      "title": "blah 1A"
    },
    {
      "id": "1B",
      "title": "blah 1B"
    }
  ],
  "Test 2": [
    {
      "id": "2A",
      "title": "blah 2A"
    }
  ]
}
```

It's not really a show-stopper, just an annoyance.  Any idea why having a space next the parenthesis would cause the error?

Narretz:
1.2.15 is a very old version of Angular and 1.2.x only receives criticial bug fixes. You should at least use 1.2.29 or upgrade to 1.5.x. The bug sounds vaguely familiar and I think it has been fixed in newer releases.

LeftPinkie:
Thanks for the follow-up. It's annoying more than a bug. I am new to Angular & just took over a biz critical site with no dev/test environment so I am a little hesitant to upgrade anything at this time.

Narretz:
Ok I see. The reason is that the content of ngRepeat is  parsed separately from other expressions. So the regex is simply a bit too strict in this case.

Narretz:
This works at least since 1.3.19: http://plnkr.co/edit/k5IE189fWGt2h0j0Zsn2?p=preview
So if you don't have to support IE8, you should consider upgrading from 1.2.x.

PR #15914, opened 2017-04-13
Update(orderBy): Guaranteed stable sort

BobChao87:

Feature

#14881 

Falls back on built-in comparison function if user-provided comparison function fails to break a tie between elements being ordered.

Yes

**Other information**:

BREAKING CHANGE: Using `orderBy` with a custom `comparator` function while
simultaneously setting `reverse` to `true` results in a potentially different
list order than previously.

Specifically, when the custom `comparator` fails to sort two objects (compares
them as being "equal"), the objects would previously have remained in their
original order in the array relative to each other. They will now switch places
as the `reverse` will be more accurately respected.

To migrate, make sure that your custom `comparator` properly differentiates
between all cases that your code base cares about and that "equal" cases are
not order dependent. We expect that due to the nature of this change, since the
entries are considered "equal", that very few projects will be affected by this
change in all but a minor visual manor that still respects the expected
operation being performed.

### DIFF
--- a/src/ng/filter/orderBy.js
+++ b/src/ng/filter/orderBy.js
@@ -49,9 +49,8 @@
  * second, or `1` otherwise.
  *
  * In order to ensure that the sorting will be deterministic across platforms, if none of the
- * specified predicates can distinguish between two items, `orderBy` will automatically introduce a
- * dummy predicate that returns the item's index as `value`.
- * (If you are using a custom comparator, make sure it can handle this predicate as well.)
+ * specified predicates nor the custom comparator can distinguish between two items, `orderBy` will
+ * automatically introduce a dummy predicate that returns the item's index as `value`.
  *
  * Finally, in an attempt to simplify things, if a predicate returns an object as the extracted
  * value for an item, `orderBy` will try to convert that object to a primitive value, before passing
@@ -599,7 +598,7 @@ function orderByFilter($parse) {
         }
       }
 
-      return compare(v1.tieBreaker, v2.tieBreaker) * descending;
+      return (compare(v1.tieBreaker, v2.tieBreaker) || defaultCompare(v1.tieBreaker, v2.tieBreaker)) * descending;
     }
   };
 
--- a/test/ng/filter/orderBySpec.js
+++ b/test/ng/filter/orderBySpec.js
@@ -457,6 +457,19 @@ describe('Filter: orderBy', function() {
 
         expect(orderBy(items, expr, reverse, comparator)).toEqual(sorted);
       });
+
+      it('should use the default comparator to break ties on a provided comparator', function() {
+        // Some list that won't be sorted "naturally", i.e. should sort to ['a', 'B', 'c']
+        var items = ['c', 'a', 'B'];
+        var expr = null;
+        function comparator() {
+          return 0;
+        }
+        var reversed = ['B', 'a', 'c'];
+
+        expect(orderBy(items, expr, false, comparator)).toEqual(items);
+        expect(orderBy(items, expr, true, comparator)).toEqual(reversed);
+      });
     });
 
     describe('(object as `value`)', function() {


test/ng/filter/orderBySpec.js
```diff
@@ -457,6 +457,21 @@ describe('Filter: orderBy', function() {
...
+        // Some list that won't be sorted "naturally", i.e. should sort to ['a', 'B', 'c']
+        var items = ['c', 'a', 'B'];
+        var expr = null;
+        function comparator() {
+          return 0;
+        }
+        var reversed = ['B', 'a', 'c'];
+        var sorted = ['a', 'B', 'c'];
+
+        expect(orderBy(items)).toEqual(sorted);
```
gkalpak: This doesn't seem related.

src/ng/filter/orderBy.js
```diff
@@ -115,7 +115,7 @@
  *
  * @param {boolean=} reverse - If `true`, reverse the sorting order.
  * @param {(Function)=} comparator - The comparator function used to determine the relative order of
- *    value pairs. If omitted, the built-in comparator will be used.
+ *    value pairs. If omitted or finds two entries to be equal, the built-in comparator will be used.
```
gkalpak: I think thisis more confusing than helpful. If they are interested thay read the details below.

gkalpak (changes requested):
A couple of minor comments only.
I am not sure about the BC tbh. I don't really consider it to be one. It would only break usecases where all of the following conditions are true:

- The app is using a custom comparator.
- The custom comparator fails to differentiate between two values.
- The app expects both ascending and descending sorting to yield the same results (which depend on the browser).

BobChao87:
Fair enough about the breaking change. I wasn't sure either, but figured it would be better to write it up than gloss over it. 

I'll get right to work on those changes.

src/ng/filter/orderBy.js
```diff
@@ -599,7 +602,7 @@ function orderByFilter($parse) {
         }
       }
 
-      return compare(v1.tieBreaker, v2.tieBreaker) * descending;
+      return (compare(v1.tieBreaker, v2.tieBreaker) || defaultCompare(v1.tieBreaker, v2.tieBreaker)) * descending;
```
lgalfaso: I do not fully understand this change, this should be just
`return defaultCompare(v1.tieBreaker, v2.tieBreaker) * descending;`

There is no reason for an external function to understand the internal `tieBreaker` structure.

BobChao87: This is not intended to defer to the `defaultCompare` function over the user-provided function, but rather to provide a reasonable fallback should a user-provided function fail to the break the tie. See [#14481 (comment)](https://github.com/angular/angular.js/issues/14881#issuecomment-231164555)

gkalpak: @lgalfaso, for better or worse, this tiebreaker structure is documented and has been used with custom comparator 😕 Do you think we should change that?
I thought it is useful to give the custom comparator a chance to break the tie in its own way (in the rare case where this will be needed).

lgalfaso:
Was not aware that this was added as part of the public API. Once this was
added, then I think it is best to keep it.

BobChao87:
Was there anything else pending or can we go ahead and mark this as ready to be merged? I think we've covered all the concerns that have been raised.

gkalpak (approved):

src/ng/filter/orderBy.js
```diff
@@ -460,6 +460,9 @@
  * way. (When specifying a custom comparator, you also need to pass a value for the `reverse`
  * argument - passing `false` retains the default sorting order, i.e. ascending.)
  *
+ * If your comparator fails to differentiate between two items, they will fall back to the built-in
+ * comparator function, this produces a guarantee of stable sorting.
+ *
```
gkalpak: Actually, this is the wrong place for this comment. It should be moved to [L55][1] and be changed to something like:

> If a custom comparator cannot still can't distinguish between two items, then they will be sorted based on their index using the built-in comparator.

[1]: https://github.com/BobChao87/angular.js/blob/8f236af514d749f194c0a1ef8fa7f692386997eb/src/ng/filter/orderBy.js#L55

gkalpak (changes requested):
Can you also change the commit message to adhere to our guidelines. E.g.:

> fix(orderByFilter): guarantee stable sort
>
> If the user-provided comparator fails to break a tie, fall back to the built-in comparator.
>
> Fixes #14881

BobChao87:
Apologies about the commit message, apparently it's been a while since I read the contributing guidelines and I use a similar scheme on another project.

src/ng/filter/orderBy.js
```diff
@@ -49,9 +49,8 @@
  * second, or `1` otherwise.
  *
  * In order to ensure that the sorting will be deterministic across platforms, if none of the
- * specified predicates can distinguish between two items, `orderBy` will automatically introduce a
- * dummy predicate that returns the item's index as `value`.
- * (If you are using a custom comparator, make sure it can handle this predicate as well.)
```
gkalpak: This line should not be removed. It is still necessary for the comparator to handle the tie-break value. E.g. if it has an array of strings, it should not expect that all values will be strings, because there is the possibility of getting passed a number (the index).

src/ng/filter/orderBy.js
```diff
@@ -49,9 +49,8 @@
...
+ * specified predicates nor the custom comparator can distinguish between two items, `orderBy` will
```
gkalpak: This change is confusing imo. The custom comparator is not something additional to the predicates. They are used together (the predicates provide the values and the comparator compares them).

gkalpak (approved):
A couple of nit. I will fix up while merging.

PR #15919, opened 2017-04-17
docs(guide/Migrating from Previous Versions): Removes duplicate #13c252

TheHalcyonSavant:
Removes duplicate #13c252 **IE11 only** migration text.


TheHalcyonSavant:
I've just signed the CLA.

PR #15922, opened 2017-04-18
Fixes to select and ngOptions - "required" validation and behavior with empty / unknown option

Narretz:
bugfixes

- (a) when the unknown option is selected and the select is "required", the required error is not set
- (b) ngOptions: when the model is unmatched and an empty option exists, the empty option is always selected, even if the model is not null / undefined.

- (a) the error is set
- (b) the unknown option is selected if the model is not null / undefined

Possibly? For (a), the following situation is possible: The model is programatically set to a value that is no longer in select options, and the select is required. With this patch, the user cannot keep the unmatched value, because of the required error, which considers this value invalid (just realized this, so it's not in the commit message).

For (b), it's unlikely that anyone relied on the fact that the empty option is selected instead of the unmatched option - the fixd behavior was also present in regular select since the beginning.

**Other information**:

### DIFF
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -704,7 +704,6 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
             ngModelCtrl.$render();
           }
         }
-
       }
   }
 
--- a/src/ng/directive/ngOptions.js
+++ b/src/ng/directive/ngOptions.js
@@ -449,12 +449,12 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
       if (!multiple) {
 
         selectCtrl.writeValue = function writeNgOptionsValue(value) {
-          var selectedOption = options.selectValueMap[selectElement.val()];
+          var selectedOption = selectElement[0].options[selectElement[0].selectedIndex];
           var option = options.getOptionFromViewValue(value);
 
           // Make sure to remove the selected attribute from the previously selected option
           // Otherwise, screen readers might get confused
-          if (selectedOption) selectedOption.element.removeAttribute('selected');
+          if (selectedOption) selectedOption.removeAttribute('selected');
 
           if (option) {
             // Don't update the option when it is already selected.
@@ -464,7 +464,6 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile, 
 
             if (selectElement[0].value !== option.selectValue) {
               selectCtrl.removeUnknownOption();
-              selectCtrl.unselectEmptyOption();
 
               selectElement[0].value = option.selectValue;
               option.element.selected = true;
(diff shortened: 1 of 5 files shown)


src/ng/directive/ngOptions.js
```diff
@@ -449,12 +449,12 @@ var ngOptionsDirective = ['$compile', '$document', '$parse', function($compile,
       if (!multiple) {
 
         selectCtrl.writeValue = function writeNgOptionsValue(value) {
-          var selectedOption = options.selectValueMap[selectElement.val()];
+          var selectedOption = selectElement[0].options[selectElement[0].selectedIndex];
```
mgol: Why do you need to skip jqLite/jQuery here?

petebacondarwin:
I am not sure that (a) is the desired behaviour here. For me required error implies that the input is missing a value. In this scenario it has a value but it is an invalid one. It feels like saying that a number input box has the required error if its value is higher than the maximum allowed.

Narretz:
@petebacondarwin Hmm ... I implemented this based on the assumptions / opinions found here: https://github.com/angular/angular.js/issues/13172 I know, it's been a while :D

I agree that "unknown" is different than "empty" - the latter specifically means the select is "unselected", so setting required error seems solid. For unknown, I guess it depends on the app logic if it fulfills the required condition. The thing is that selects basically only have required validation, because the app is usually responsible for giving the select list.

I don't insist on this change, but I think we should provide a way to manipulate this behavior. The custom isEmpty method can be implemented by a 3rd party directive, but for that we need to expose / document properties of the select controller, mainly ctrl.unknownOption ... and maybe some others. This would obviously put us in a position where we expose more API that we have to maintain :o But it would also allow people to create a different error, e.g. "unmatched value" that is different than  "required".

src/ng/directive/select.js
```diff
@@ -650,6 +655,21 @@ var selectDirective = function() {
...
+
+        // A single select is also considered empty when the unknown option is present, or when
+        // the empty option is selected
+        ngModelCtrl.$isEmpty = function(value) {
+
+          return originalIsEmpty(value) ||
+            !!selectCtrl.unknownOption.parent().length ||
+            // When the empty option is selected from the scope, then the viewValue can be different
+            // than '', and the originalIsEmpty fn will not catch this case
+            (selectCtrl.hasEmptyOption &&
```
Narretz: Note that this condition only affects empty option required validation, not unknown option - commit message needs to be updated.

src/ng/directive/ngOptions.js
Narretz: We don't. I just didn't think to use jQuery here. How do you get the selected option with jquery?

mgol: Hmm, I was just comparing it to the previous code which used `options.selectValueMap`. Could you explain the change?

petebacondarwin:
OK, I think exposing some more public API to empower developers is a reasonable goal here.

src/ng/directive/ngOptions.js
Narretz: Ah, I see. Regular select was using the same technique - getting the option directly from the DOM, so I made it consistent. I don't think there's a huge difference, it just felt right to look at the actual DOM

Narretz:
I've removed the change about required and the unknown option, and expanded on the tests around this behavior. PR is ready to review.
Select API changes will follow in a different PR.

petebacondarwin:
I'll review tomorrow morning

petebacondarwin (commented):
Nice work @Narretz - LGTM

petebacondarwin (approved):

petebacondarwin:
Please merge when you have a moment @Narretz

PR #15927, opened 2017-04-20
chore: unit test latest + latest-1 version of FF, Chrome, Edge

Narretz:
At the moment, we use fixed versions of these browsers to test - especially for FF and Chrome it's useful to test the latest, as they release faster.
For the future, we could also look into using  `latest-1` additionally.

### DIFF
--- a/karma-shared.conf.js
+++ b/karma-shared.conf.js
@@ -45,12 +45,22 @@ module.exports = function(config, specificOptions) {
       'SL_Chrome': {
         base: 'SauceLabs',
         browserName: 'chrome',
-        version: '59'
+        version: 'latest'
+      },
+      'SL_Chrome-1': {
+        base: 'SauceLabs',
+        browserName: 'chrome',
+        version: 'latest-1'
       },
       'SL_Firefox': {
         base: 'SauceLabs',
         browserName: 'firefox',
-        version: '54'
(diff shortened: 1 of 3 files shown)


gkalpak:
I am not sure how I feel about this (especially for Edge since it is tied to the Windows version).
But for Chrome and Firefox we are probably fine (although I would feel more confident with `latest-1`).

Narretz:
`latest-1` sounds good - at the moment, there's a failing test in FF 52.0.0, which I can't reproduce in my local FF 52.0.2. Or does Saucelabs not report the patch version?
In any case, Saucelabs is kind of slow updating to the latest version, so we're probably better of using latest-1

Narretz:
With jasmine 2.6.4, all browsers fail to complete: https://travis-ci.org/angular/angular.js/jobs/243851080#L619
Locally, the csp tests in AngularSpec.js are to blame: https://github.com/angular/angular.js/blob/e58bcfab99450d2c4777d87a7ddfea19b9f38b5c/test/AngularSpec.js#L1026

When excluded, the tests pass. Probably another jasmine bug.

Btw, I don't even know why I updated jasmine. I think an old version was incompatible with Chrome 58?

petebacondarwin (approved):
Don't forget to squash

PR #15931, opened 2017-04-21
chore: completely remove the ng scenario runner

### LINKED ISSUE #9405: chore(scenarioRunner): do not publish updates to bower/npm
Suggested by @IgorMinar.

Since this has been in "maintenance mode" for a while, it's probably good to start winding down on it and encouraging users to switch to protractor.

Narretz:
Removes the deprecated scenario runner (protractor ancestor)

Closes https://github.com/angular/angular.js/issues/9405

TODO:
- [x] remove / adjust doc info about the scenario runner
- [x] document browserTrigger as part of ngMock (where it has been moved to) 
- [ ] move scenario runner to stand-alone repo (community) before removing it from core? - removing it means no more updates to npm / bower, which could hurt people's build process

**Other information**:


src/ngMock/browserTrigger.js
```diff
@@ -2,13 +2,23 @@
...
    *
    * @param {Object} element Either a wrapped jQuery/jqLite node or a DOMElement
-   * @param {string} eventType Optional event type
+   * @param {string=} eventType Optional event type. If none is specified, the function tries
+   *                            to determine the right event type for the element, e.g. `change` for
+   *                            `input[text]`.
    * @param {Object=} eventData An optional object which contains additional event data (such as x,y
-   * coordinates, keys, etc...) that are passed into the event when triggered
+   *                            coordinates, keys, etc...) that are passed into the event when
+   *                            triggered.
```
mgol: Do changes in this file belong in this PR?

Narretz: It's just a little improvement on the formatting, doesn't hurt, right? 😅 Or do you mean browserTrigger docs in general? We moved bT before in preparation of the scenario runner removal, but never documented it,

src/ng/browser.js
```diff
@@ -67,7 +67,7 @@ function Browser(window, document, $log, $sniffer) {
 
   /**
    * @private
-   * Note: this method is used only by scenario runner
+   * Note: this method is used only by protractor / scenario runner
```
gkalpak: Is protractor really using this? I think it uses testability.

src/ngMock/browserTrigger.js
gkalpak: Shouldn't we be more specific about what properties are supported in `eventData` what they are used for?

src/ng/location.js
```diff
@@ -938,8 +938,6 @@ function $LocationProvider() {
           // update location manually
           if ($location.absUrl() !== $browser.url()) {
             $rootScope.$apply();
-            // hack to work around FF6 bug 684208 when scenario runner clicks on links
-            $window.angular['ff-684208-preventDefault'] = true;
```
gkalpak: This is still used in `browserTrigger`: https://github.com/angular/angular.js/pull/15931/files#diff-6fcc18d824a1325cf54ef306cd2fb7a5R124

src/ng/browser.js
Narretz: Testability.whenStable calls this function

gkalpak: I know, but this comment does not belong here. For example, we have discussed in the past that we should make Testability public, which means that anyone could use `whenStable()`. And if we do that, it is very unlikely that someone will remember to update this comment.

If we want to have such a comment, it should be on `whenStable()` imo.

Narretz:
- removed the comment in $browser
- removed the workaround for Firefox in browserTrigger.
- updated eventData with more info. It's still not perfect, because we use different event creation paths for different browsers / events, but it's better than nothing (since we had no docs before).

src/ngMock/browserTrigger.js
```diff
@@ -9,16 +9,37 @@
...
+   *
+   *  - `keys`: an array of possible modifier keys (ctrl, alt, shift, meta) for the event.
+   *
+   *  - `x`: x-coordinates for [MouseEvent](https://developer.mozilla.org/docs/Web/API/MouseEvent)
+   *    and [TouchEvent](https://developer.mozilla.org/docs/Web/API/TouchEvent).
+   *
+   *  - `y`: y-coordinates for [MouseEvent](https://developer.mozilla.org/docs/Web/API/MouseEvent)
+   *    and [TouchEvent](https://developer.mozilla.org/docs/Web/API/TouchEvent).
+   *
+   *  - `relatedTarget`: the related target for
```
gkalpak: Maybe also link to [MouseEvent.relatedTarget][1].

[1]: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget

src/ngMock/browserTrigger.js
```diff
@@ -9,16 +9,37 @@
...
+   *  - `elapsedTime`: the elapsedTime for
+   *    [TransitionEvent](https://developer.mozilla.org/docs/Web/API/TransitionEvent)
+   *    and [AnimationEvent](https://developer.mozilla.org/docs/Web/API/AnimationEvent).
+   *
+   *  - `bubbles`: [Event.bubbles](https://developer.mozilla.org/docs/Web/API/Event/bubbles).
+   *    Not applicable to all events.
+   *
+   *  - `cancelable`: [Event.cancelable](https://developer.mozilla.org/docs/Web/API/Event/cancelable).
+   *    Not applicable to all events.
+
```
gkalpak: Missing `*` :smiley:

gkalpak: Maybe also mention `keyCode`, `charCode` and `which` for `key*` events.

gkalpak (approved):
I love the docs&nbsp;:heart:
LGTM (with a couple of nits/suggestions) 👍

PR #15947, opened 2017-04-28
perf(ngStyleDirective): use $watchCollection

jbedard:
This re-commits https://github.com/angular/angular.js/commit/4c8d8ad5083d9dd17c0b8480339d5f95943f1b71 which was [reverted](https://github.com/angular/angular.js/commit/36fd167e1d6668ab497970f491a6d95344e97e41) (with [tests added](https://github.com/angular/angular.js/commit/da75d138b144d196fc2b397c7a86a74528c151a5)).  This is possible now because of 189461f9bf6fda18ddbd16c42f2e959cf939c3da, although note that change is currently causing #15905 so maybe we should hold off on merging this one for now?

FYI @gdi2290

### DIFF
--- a/src/ng/directive/ngStyle.js
+++ b/src/ng/directive/ngStyle.js
@@ -52,10 +52,10 @@
    </example>
  */
 var ngStyleDirective = ngDirective(function(scope, element, attr) {
-  scope.$watch(attr.ngStyle, function ngStyleWatchAction(newStyles, oldStyles) {
+  scope.$watchCollection(attr.ngStyle, function ngStyleWatchAction(newStyles, oldStyles) {
     if (oldStyles && (newStyles !== oldStyles)) {
       forEach(oldStyles, function(val, style) { element.css(style, '');});
     }
     if (newStyles) element.css(newStyles);
-  }, true);
+  });
 });
--- a/test/ng/directive/ngStyleSpec.js
+++ b/test/ng/directive/ngStyleSpec.js
@@ -38,6 +38,18 @@ describe('ngStyle', function() {
   }));
 
 
+  it('should support binding for object literals', inject(function($rootScope, $compile) {
+    element = $compile('<div ng-style="{height: heightStr}"></div>')($rootScope);
+    $rootScope.$digest();
+    expect(parseInt(element.css('height') + 0, 10)).toEqual(0); // height could be '' or '0px'
+    $rootScope.$apply('heightStr = "40px"');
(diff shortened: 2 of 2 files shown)


gkalpak:
:+1: for waiting until #15905 is resolved.

PatrickJS (approved):

Narretz:
Interestingly, there's no test for normal literal bindings:

```
  it('should support binding for object literals', inject(function($rootScope, $compile) {
    element = $compile('<div ng-style="{height: heightStr}"></div>')($rootScope);
    $rootScope.$digest();
    expect(parseInt(element.css('height') + 0, 10)).toEqual(0); // height could be '' or '0px'
    $rootScope.$apply('heightStr = "40px"');
    expect(element.css('height')).toBe('40px');

    $rootScope.$apply('heightStr = "100px"');
    expect(element.css('height')).toBe('100px');
  }));
```

jbedard:
@Narretz I've added a commit with that test. Also changed to 1.7 because this depends on 189461f9bf6fda18ddbd16c42f2e959cf939c3da which is going to be reverted in 1.6 (#15958).

jbedard:
I think this can go into 1.7 now @gkalpak?

jbedard:
Note: this will break cases where people are relying on a value being an object and automatically invoking `.toString()` such as `{color: obj}` where `obj.toString()` is stateful.  This can be fixed by changing the expression to `{color: obj.toString()}`. Should add a breaking change notice for this.

gkalpak (approved):
As discussed "offline", this is good-to-go for 1.7 (with the breaking change notice) 👍

jbedard:
I've updated the commit message with a "BREAKING CHANGE" message and changed the "Closes #" to this PR. Look good?

Narretz:
Should there be a test for the case that is described in the BC notice?

jbedard:
A test verifying that styles don't get re-applied when an object sub-property changes? Could...

Narretz:
Just because we are setting an expectation for this behavior

jbedard:
I've added a test. The other option is verifying that `ng-style="x"` invokes `$watchCollection(x)`...?

Narretz:
LGTM

gdi2290:
nice

gkalpak:
A tiny change with potentially big gains :smiley:
(It also took some serious poking around the `$parse` service - kudos to @jbedard 💯)

PR #15949, opened 2017-04-29
test($rootScope): add $watch/$watchCollection tests

jbedard:
These are various cases that I felt were missing tests, mainly around constant/literal/one-time $watch/$watchCollection.

### DIFF
--- a/test/ng/rootScopeSpec.js
+++ b/test/ng/rootScopeSpec.js
@@ -146,16 +146,6 @@ describe('Scope', function() {
       expect(spy).toHaveBeenCalled();
     }));
 
-    it('should not keep constant expressions on watch queue', inject(function($rootScope) {
-      $rootScope.$watch('1 + 1', function() {});
-      expect($rootScope.$$watchers.length).toEqual(1);
-      expect($rootScope.$$watchersCount).toEqual(1);
-      $rootScope.$digest();
-
-      expect($rootScope.$$watchers.length).toEqual(0);
-      expect($rootScope.$$watchersCount).toEqual(0);
-    }));
-
     it('should decrement the watcherCount when destroying a child scope', inject(function($rootScope) {
       var child1 = $rootScope.$new(),
         child2 = $rootScope.$new(),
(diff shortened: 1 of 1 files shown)


test/ng/rootScopeSpec.js
```diff
@@ -1108,6 +1418,8 @@ describe('Scope', function() {
       scope.$watchGroup(['1'], noop);
       //multi constant
       scope.$watchGroup(['1', '2'], noop);
+      //multi one-time/constant
+      scope.$watchGroup(['::a', '1'], noop);
```
gkalpak: Wrong indentation.

test/ng/rootScopeSpec.js
```diff
@@ -212,10 +212,52 @@ describe('Scope', function() {
       expect($rootScope.$$watchersCount).toBe(2);
     }));
 
-    it('should not keep constant literals on the watch queue', inject(function($rootScope) {
+    it('should remove $watch of constant literals after initial digest', inject(function($rootScope) {
```
gkalpak: Wrong indentation (here and below).

gkalpak (approved):
Awesome (except for the indentation issues 😛)
:100:

test/ng/rootScopeSpec.js
jbedard: I don't understand, isn't it the same as before and the same as all the others around it?

test/ng/rootScopeSpec.js
```diff
@@ -212,10 +212,52 @@ describe('Scope', function() {
...
-      expect($rootScope.$$watchers.length).toEqual(2);
+      $rootScope.$watch('1', function() {});
+      $rootScope.$watch('"foo"', function() {});
+      expect($rootScope.$$watchers.length).not.toEqual(0);
+      $rootScope.$digest();
+
+      expect($rootScope.$$watchers.length).toEqual(0);
+    }));
+
+    it('should remove $watchCollection of constant literals after initial digest', inject(function($rootScope) {
```
Narretz: There is a `$watchCollection` describe in this file, I think the watchCollection specs should go there (also for all the other watchCollection tests you've added)

Narretz (changes requested):
Tests look good (based on my limited understanding :D) but I would like the specs to be put into the relevant describes.

test/ng/rootScopeSpec.js
gkalpak: Sorry, it turned out I had a Chrome extension that was messing the GitHub diffs. 😞
Indentation is fine.

test/ng/rootScopeSpec.js
jbedard: That `describe` block has a `beforeEach` that does `$watchCollection("var", ...)` so I converted it to `describe("$watchCollection var", ...)`.

Maybe I should add a parent `describe("$watchCollection", ...)` that wraps the var/literal/constant describes and can also contain this test?

I'll do that now...

jbedard: Actually this specific `$watchCollection` test is beside the equivalent `$watch` test.

What's better, putting similar watch/watchCollection beside each other or separating them into watch/watchCollection groups?  I feel like this specific test is better grouped along with other constant related tests instead of with $watchCollection...

jbedard:
I've wrapped all the `describe('$watchCollection X'` blocks into one parent `describe('$watchCollection'`, and removed a couple unnecessary vars/inject-params. See the second commit. I think the commits should be squashed before merging though.

@Narretz can you clarify [your comment](https://github.com/angular/angular.js/pull/15949#discussion_r114509149)? I'd like to get this merged so I can add more... 😃

test/ng/rootScopeSpec.js
Narretz: Ok, how about wrapping the constant related watch and watchCollection tests with another describe? Btw, what about watchGroup :D

jbedard: I've wrapped the constant-cleanup and onetime-cleanup tests into describe blocks. Also added some watchGroup tests.

Narretz (approved):
Thanks @jbedard!

Narretz:
There are some lint errors, otherwise LGTM

jbedard:
Fixed the lint issues. I'll squash and merge later today if no one else gets to it.

PR #15953, opened 2017-05-02
fix($interval): change $interval ngMock service to handle second argu…

### LINKED ISSUE #15952: $interval ngMock has infinite loop when delay argument is equivalent to zero
**I'm submitting a ...**

- [x] bug report
- [ ] feature request

**Current behavior:**

When writing a spec if an interval is set with a delay argument equivalent to zero, then using the flush method will generate an infinite loop.

**Expected behavior:**

When running `setInterval(..., 0);` in the browser it is executed almost every tick, for mock the same behavior is expected.

https://plnkr.co/edit/uzZeMC8ruoEOT1l2MijI?p=preview

1. Open the `app.spec.js` file
2. Uncomment the line with the flush method

Result: the browser's tab will crush due infinite loop.

**Angular version:** 1.6.4

**Browser:** all

**Anything else:**

```
PhantomJS 2.1.1 (Windows 8 0.0.0) ERROR
  Disconnected, because no message in 10000 ms.
```

luciomartinez:
…ment equivalent to zero

Enable test interval calls with delay set to zero without infinite loop.
If a spec has `$interval.flush(ms)` and the code to test has
`$interval(callback, 0)` then instead of running an infinite while loop
it would move forward `ms` executing the $interval's callback.

Fixes #15952

Bug fix

https://github.com/angular/angular.js/issues/15952

No.

**Other information**:

### DIFF
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -504,8 +504,8 @@ angular.mock.$IntervalProvider = function() {
       }
 
       repeatFns.push({
-        nextTime:(now + delay),
-        delay: delay,
+        nextTime: (now + (delay || 0)),
+        delay: delay || 1,
         fn: tick,
         id: nextRepeatId,
         deferred: deferred
@@ -555,10 +555,16 @@ angular.mock.$IntervalProvider = function() {
      * @return {number} The amount of time moved forward.
      */
     $interval.flush = function(millis) {
+      var before = now;
       now += millis;
       while (repeatFns.length && repeatFns[0].nextTime <= now) {
         var task = repeatFns[0];
         task.fn();
+        if (task.nextTime === before) {
+          // this can only happen the first time
+          // a zero-delay interval gets triggered
+          task.nextTime++;
+        }
         task.nextTime += task.delay;
         repeatFns.sort(function(a, b) { return a.nextTime - b.nextTime;});
       }
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -351,6 +351,75 @@ describe('ngMock', function() {
     }));
 
 
+    it('should allow you to NOT specify the delay time', inject(function($interval) {
+      var counterA = 0;
+      var counterB = 0;
+
+      $interval(function() { counterA++; });
+      $interval(function() { counterB++; }, 0);
+
+      $interval.flush(1000);
+      expect(counterA).toBe(1000);
+      expect(counterB).toBe(1000);
+      $interval.flush(1000);
+      expect(counterA).toBe(2000);
+      expect(counterB).toBe(2000);
+    }));
+
+
+    it('should run tasks in correct relative order', inject(function($interval) {
+      var counterA = 0;
+      var counterB = 0;
+      $interval(function() { counterA++; }, 0);
+      $interval(function() { counterB++; }, 1000);
+
+      $interval.flush(1000);
+      expect(counterA).toBe(1000);
+      expect(counterB).toBe(1);
+      $interval.flush(999);
+      expect(counterA).toBe(1999);
+      expect(counterB).toBe(1);
+      $interval.flush(1);
+      expect(counterA).toBe(2000);
+      expect(counterB).toBe(2);
+    }));
+
+
+    it('should NOT trigger zero-delay interval when flush has ran before', inject(function($interval) {
+      var counterA = 0;
+      var counterB = 0;
+
+      $interval.flush(100);
+
+      $interval(function() { counterA++; });
+      $interval(function() { counterB++; }, 0);
+
+      expect(counterA).toBe(0);
+      expect(counterB).toBe(0);
+
+      $interval.flush(100);
+
+      expect(counterA).toBe(100);
+      expect(counterB).toBe(100);
+    }));
+
+
+    it('should trigger zero-delay interval only once on flush zero', inject(function($interval) {
+      var counterA = 0;
+      var counterB = 0;
+
+      $interval(function() { counterA++; });
+      $interval(function() { counterB++; }, 0);
+
+      $interval.flush(0);
+      expect(counterA).toBe(1);
+      expect(counterB).toBe(1);
+      $interval.flush(0);
+      expect(counterA).toBe(1);
+      expect(counterB).toBe(1);
+    }));
+
+
     it('should allow you to specify a number of iterations', inject(function($interval) {
       var counter = 0;
       $interval(function() {counter++;}, 1000, 2);


src/ngMock/angular-mocks.js
```diff
@@ -556,9 +556,13 @@ angular.mock.$IntervalProvider = function() {
      */
     $interval.flush = function(millis) {
       now += millis;
-      while (repeatFns.length && repeatFns[0].nextTime <= now) {
+      var firstTask = repeatFns[0];
```
gkalpak: Unused

src/ngMock/angular-mocks.js
```diff
@@ -556,9 +556,13 @@ angular.mock.$IntervalProvider = function() {
...
+      while (repeatFns.length && (repeatFns[0].nextTime <= now || !repeatFns[0].nextTime)) {
         var task = repeatFns[0];
         task.fn();
+        if (!task.delay) {
+          task.delay = task.nextTime = 1;
+        }
```
gkalpak: Wrong indentation.

src/ngMock/angular-mocks.js
```diff
@@ -556,9 +556,13 @@ angular.mock.$IntervalProvider = function() {
      */
     $interval.flush = function(millis) {
       now += millis;
-      while (repeatFns.length && repeatFns[0].nextTime <= now) {
+      var firstTask = repeatFns[0];
+      while (repeatFns.length && (repeatFns[0].nextTime <= now || !repeatFns[0].nextTime)) {
         var task = repeatFns[0];
         task.fn();
+        if (!task.delay) {
+          task.delay = task.nextTime = 1;
```
gkalpak: Why set `nextTime` to 0 (since we are also incrementing it below)?

test/ngMock/angular-mocksSpec.js
```diff
@@ -351,6 +351,18 @@ describe('ngMock', function() {
...
+    it('should allow you to NOT specify the delay time', inject(function($interval) {
+      var counter = 0;
+      $interval(function() {counter++;});
+      $interval(function() {counter++;}, 0);
+
+      $interval.flush(1000);
+      expect(counter).toBe(2000);
+      $interval.flush(1000);
+      expect(counter).toBe(4000);
+    }));
```
gkalpak: We need a test with other, non-zero-delay timeouts, to verify that the relative orders are correct.

test/ngMock/angular-mocksSpec.js
```diff
@@ -351,6 +351,18 @@ describe('ngMock', function() {
     }));
 
 
+    it('should allow you to NOT specify the delay time', inject(function($interval) {
+      var counter = 0;
+      $interval(function() {counter++;});
+      $interval(function() {counter++;}, 0);
+
+      $interval.flush(1000);
+      expect(counter).toBe(2000);
```
gkalpak: I would expect this to be 1000. Can you explain why it isn't?

gkalpak (changes requested):

src/ngMock/angular-mocks.js
luciomartinez: If `delay` is zero then `nextTime` will also be zero. Setting to 1 means the next round it'll be 2 since in fact will be the round number 2. Thanks to first set to 1 and then increasing it means that passing a value of 1 to `ms` will run only once and not twice the function. I hope this makes sense :-)

src/ngMock/angular-mocks.js
luciomartinez: Sorry, I set two spaces as indentation. Is it wrong or I'm missing something else?

test/ngMock/angular-mocksSpec.js
luciomartinez: When calling interval it checks the second argument to be a number or equivalent to zero, hence undefined gives the same results. Since above two statements are increasing counter on a flush of 1000 both functions are run and counter finish with a value of 2000.

test/ngMock/angular-mocksSpec.js
luciomartinez: There are specs with non-zero-delay timeouts but maybe you mean mixing them. What do you reckon? Thanks

src/ngMock/angular-mocks.js
gkalpak: Sorry, it turned out I had a Chrome extension that was messing the GitHub diffs. Indentation is fine.

test/ngMock/angular-mocksSpec.js
gkalpak: Duh! Both callbacks increment the same counter.
Still it might be worth having two separate counters to be safer.

test/ngMock/angular-mocksSpec.js
gkalpak: Yes, I meant mixing zero and non-zero intervals in the same test.

test/ngMock/angular-mocksSpec.js
luciomartinez: Good idea! Will do

test/ngMock/angular-mocksSpec.js
luciomartinez: Brilliant!

src/ngMock/angular-mocks.js
luciomartinez: I'm wondering if it'd be helpful to write it down as a comment. What is your though?

src/ngMock/angular-mocks.js
luciomartinez: No worries 👍

src/ngMock/angular-mocks.js
gkalpak: A short comment wouldn't hurt for sure :smiley:
(Feel free to ping me when this is ready for another review.)

luciomartinez: Hey thanks! I was afraid of bothering you 😄 
It should be ready now, please let me know if I'm missing something else.

src/ngMock/angular-mocks.js
```diff
@@ -556,9 +556,16 @@ angular.mock.$IntervalProvider = function() {
...
-      while (repeatFns.length && repeatFns[0].nextTime <= now) {
+      while (repeatFns.length && (repeatFns[0].nextTime <= now || !repeatFns[0].nextTime)) {
         var task = repeatFns[0];
         task.fn();
+        // when a delay of zero is given
+        // it has to be replaced with one
+        // to avoid running the while loop
+        // one more time than the specified
+        if (!task.delay) {
+          task.delay = task.nextTime = 1;
```
gkalpak: Actually, I don't think this will work if this `0` delay timer was not added in the beginning (i.e. if there was a flush before adding it). E.g.:

```js
var counter = 0;
$interval.flush(1000);
$interval(function() { counter++; });
$interval.flush(1000);

expect(counter).toBe(1000);   // <-- Will probably be 2000
```

Also, due to how `repeatFns.nextTime` is set up initially (`nextTime: now + delay`), using leaving the delay unspecified and using `0` would behave differently.

I think we should change both `$interval` and `$interval.flush` to something like:

```js
$interval = function(fn, delay, ...) {
  ...
  repeatFns.push({
    nextTime: (now + (delay || 0)),
    delay: delay || 1,
    ...
  });
```

```js
$interval.flush = function(millis) {
  var before = now;
  now += millis;
  while (repeatFns.length && repeatFns[0].nextTime <= now) {
    ...
    if (task.nextTime === before) {
      // This can only happen the first time a zero-delay interval get triggered.
      task.nextTime++;
    }
    task.nextTime += task.delay;
    ...
```

Then we need to have more tests that verify the correct behavior of zero-delay intervals (with both `delay: 0` and `delay: undefined`) for at least the following cases: 

- Interval being added when `now` > 0.
- Calling `flush(0)` (should only trigger zero-delay intervals and only trigger them once).
- Calling `flush(0)` for a second time (should not trigger any intervals).

gkalpak (changes requested):

src/ngMock/angular-mocks.js
luciomartinez: Nice! I wasn't sure about changing `$interval` but I like this implementation. Thanks!

src/ngMock/angular-mocks.js
```diff
@@ -555,10 +555,16 @@ angular.mock.$IntervalProvider = function() {
...
      */
     $interval.flush = function(millis) {
+      var before = now;
       now += millis;
       while (repeatFns.length && repeatFns[0].nextTime <= now) {
         var task = repeatFns[0];
         task.fn();
+        if (task.nextTime === before) {
+          // this can only happen the first time
+          // a zero-delay interval get triggered
```
gkalpak: _get_ --> _gets_

test/ngMock/angular-mocksSpec.js
```diff
@@ -351,6 +351,70 @@ describe('ngMock', function() {
...
+      var counterB = 0;
+
+      $interval.flush(100);
+
+      $interval(function() { counterA++; });
+      $interval(function() { counterB++; }, 0);
+
+      expect(counterA).toBe(0);
+      expect(counterB).toBe(0);
+    }));
```
gkalpak: We need another `$interval.flush(X)` and anssertions that the callbacks where run only X.

gkalpak (changes requested):

luciomartinez:
@gkalpak updated changes. Thanks!

gkalpak (approved):

PR #15967, opened 2017-05-06
test(*): run class-related tests everywhere; fix eval syntax

mgol:

Test fixes.

There was a syntax error in class-related tests; the test wasn't failing only because it was disabled everywhere outside of Chrome and Chrome <59 incorrectly accepted it.

1. Wrap an evaled class definition in parens; previously they weren't.
2. The classes support test was modified to check not only if a class definition parses but also if it stringifies correctly which is required by AngularJS. This restriction disables class-related tests in current Firefox (53) but will work in v55 or newer.

No.

**Other information**:

### DIFF
--- a/test/.eslintrc.json
+++ b/test/.eslintrc.json
@@ -176,6 +176,7 @@
     "browserSupportsCssAnimations": false,
     "browserTrigger": false,
     "jqLiteCacheSize": false,
-    "createAsync": false
+    "createAsync": false,
+    "support": false
   }
 }
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -1,7 +1,5 @@
 'use strict';
 
-/* globals support: false */
-
 describe('injector.modules', function() {
     it('should expose the loaded module info on the instance injector', function() {
       var test1 = angular.module('test1', ['test2']).info({ version: '1.1' });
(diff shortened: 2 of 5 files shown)


gkalpak (commented):
My only concern is the following:

In case that our `supportTests` start returning false (so the tests are skipped), we might break the functionality but won't see any tests failures.

I would feel more comfortable if we had some way to essentially verify the `supportTests`. E.g. if we know that Chrome should support all features, have tests that verify the `if (/chrome/.test(navigator.userAgent)) support.* === true`. (Open to better ideas too :smiley:)

mgol:
@gkalpak I tend to agree. While in code shipped to browsers support tests are much, much better than UA-sniffing (so that unknown environments with proper support works), in tests it may be better to UA-sniff to make sure that our assertions are correct.

Note that currently we have the `classes` support test but all its uses are additionally wrapped in a Chrome UA-check; the worse of both worlds. :D So this PR in its current form is an improvement anyway.

How about we set all support values to `true` and only flip them to false in these few environments which we both test and know they wouldn't pass the tests? (mostly IE/Android Browser and Firefox just for the `classes` test) Such a solution would allow unknown environments to pass the tests as well provided they support new enough JS.

Narretz:
We could also create a "blacklist" of  browsers we know will not support classes (basically IE). That way, every new browser is automatically assuned to support classes

mgol:
It's not just IE, also old Safari/iOS (which we'll be testing for a while) and we might want to try running the tests on the Android Browser 4.0-4.3. For now it seems better IMO to just assert the results are what we expect them to be as we do in jQuery: https://github.com/jquery/jquery/blob/3.2.1/test/unit/support.js#L61-L235

mgol:
I'd like to land this PR as-is, the new tests verifying support tests results seem like a separate thing (I can submit a PR for that shortly).

gkalpak:
I am fine adding the support tests in a separate PR.

Narretz:
Me too!

PR #15970, opened 2017-05-07
docs(error/ctrlreg): Typo in the Controller Logic

ghost:
docs(error/ctrlreg): Typo in the Controller Logic

None of the current documentation causes for this error reference verifying the controller logic itself. However, this error could indicate that there is an error in the controller logic, similar to the code below (note the ill-placed semicolon on the first line):

```javascript
UserService.addNewGame(gameObject);
  .then(redirectToTeamSchedule);
```

Removing the semicolon from the first line of the controller logic alleviates the $controller:ctrlreg error.

docs update

Error: $controller:ctrlreg -- A controller with this name is not registered.
Solutions in the current documentation do not cover all possible reasons for the error.

docs update with additional detail as to where a code error may exist if a $controller:ctrlreg error is triggered

no

**Other information**:

### DIFF
--- a/docs/content/error/$controller/ctrlreg.ngdoc
+++ b/docs/content/error/$controller/ctrlreg.ngdoc
@@ -18,6 +18,6 @@ definition's controller property, or in the call to {@link ng.$controller `$cont
 2. You have not registered the controller (neither via {@link angular.Module#controller `Module.controller`}
 nor {@link ng.$controllerProvider#register `$controllerProvider.register()`}.
 3. You have a typo in the *registered* controller name.
-
+4. You have a typo within the controller logic itself.
 
 Please consult the {@link ng.$controller $controller} service api docs to learn more.


frederikprijck:
Doesn't this depends on how you're registering your controllers ?

Tried a comparable plunkr and got another error message: https://plnkr.co/edit/2pT57JXV3AI5GEUdYRyu?p=preview

gkalpak:
Saying "you might have a typo" is a very generic message and is not helpful imo. It is the same as saying "your JavaScript must be valid", but this should be considered a prerequisite.

Also, as @frederikprijck said, you should already get a different error because your JavaScript is invalid in the first place. But in any case, it is always possible to have a typo that leaves you with syntactically correct code, which behaves totally different that you expect. AngularJS should not try to be smart on what your error could be.

The error message says what is the problem as far as AngularJS is concerned (in this case the fact that the controller is not registered) and gives you a couple of hints on common causes (but definitely not an exhaustive list).

##
I am going to close this, because it doesn't add anything useful to the message.
But we appreciate your taking the time to submit a PR and would love to see more of this in the future :wink:

danzera:
OK - Thanks for the feedback. The error message just seemed to be misleading, but I’m pretty new to Angular, and full-stack dev in general, so I can’t say I’m an expert on error verification and where the responsibility lies.

I must have been mistaken about where Angular ends and JS begins.

gkalpak:
👍 
BTW, you might want to look into [eslint][1] :wink:

[1]: http://eslint.org/

danzera:
Thanks! I just switched over from Atom to VS Code last week and haven’t installed a linter yet. Not sure how I’ve gotten by this long without one. Guess it’s just forced me to pay closer attention, but can’t catch ‘em all sometimes.

Sorry for wasting your time.

PR #15975, opened 2017-05-09
Update ngRepeat.js

tsclaus:
docs(ngRepeatDirective): updates

Updates documentation to correctly refer to block property as `clone` instead of as `element`.


tsclaus:
I signed it!

PR #15983, opened 2017-05-11
fix(httpParamSerializerJQLike): First array element always non-indexed

odedniv:

Bug fix

Serializing an array of objects with `$httpParamSerializerJQLike` is ambiguous with serializing an object with numeric keys.

```javascript
params = { a: [{ b: 1 }, { c: 2 }] }
=> a[0][b]=1&a[1][c]=2
```

This *could* falsely be deserialized as: `{ a: { "0": { b: 1 }, "1": { c: 2 } } }` (and it absolutely will).

```javascript
params = { a: [{ b: 1 }, { c: 2 }] }
=> a[][b]=1&a[1][c]=2
```

This first element ensures `a` should be considered an array.

Unfortunately this is not backward compatibly (would actually break rack before it pulls the mentioned request), but this is somewhat mandatory as you simply cannot properly serialize some scenarios (see the tests in rack's PR for more scenarios). Should this be a configuration option somewhere?

**Other information**:

This follows these PR's in jQuery and rack: jquery/jquery#3666, rack/rack#1165

This PR is pretty much copied from the jQuery PR, and it should probably only be accepted after jQuery accepts their PR (since this method is a clone of theirs).

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -112,7 +112,7 @@ function $HttpParamSerializerJQLikeProvider() {
       function serialize(toSerialize, prefix, topLevel) {
         if (isArray(toSerialize)) {
           forEach(toSerialize, function(value, index) {
-            serialize(value, prefix + '[' + (isObject(value) ? index : '') + ']');
+            serialize(value, prefix + '[' + ((index > 0 && isObject(value)) ? index : '') + ']');
           });
         } else if (isObject(toSerialize) && !isDate(toSerialize)) {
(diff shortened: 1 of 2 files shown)


gkalpak:
We should definitely only consider this _after_ the jQuery PR has been accepted. It is not actionable atm.

Narretz:
The jquery PR https://github.com/jquery/jquery/pull/3666 has been closed because it's stale. For the time being, this should be closed, too.

Issue #15960, opened 2017-05-03
Remove support for nested objects/arrays in ng-class


jbedard:
189461f9bf6fda18ddbd16c42f2e959cf939c3da caused an issue with nested literal objects/arrays in ng-class (#15905), essentially because it allowed the expression inputs (a watch optimization) to be used properly in the case of literals with interceptorFns.

Instead of reverting 189461f9bf6fda18ddbd16c42f2e959cf939c3da, or changing ng-class to use deep watch, or declaring its interceptor as `$stateful` - I'd like to drop nested object/array support from `ng-class`. This will align `ng-class` with the [Angular version](https://github.com/angular/angular/blob/4.0.0/packages/common/src/directives/ng_class.ts#L59), should improve performance, should simplify the ng-class code a bit.

This would be a breaking change in 1.7.

Any strong objections? Strong use cases for nested objects/arrays in ng-class?

lgalfaso:
Putting a side the specific way that this should be resolved, there are a few points that should be covered:

1. Even when nested arrays/objects are no longer supported, simple objects/arrays are supported

2. 189461f9bf6fda18ddbd16c42f2e959cf939c3da caused
`ng-class="[root.obj1, root.obj2]"`
and
`$scope.root.styles2 = [root.obj1, root.obj2];`
`ng-class="root.styles2"`
to behave differently. This discrepancy must be avoided

3. Even if nested arrays/objects are not supported, then, if someone creates a stateless filter that expose the nested object/array as a single object then `ng-class="[root.obj1, root.obj2] | mergeThings"` must work (in some sense, this point at the previous one can be considered the same)

4. There should be a clear migration path, this may include the specific example from the bug and how this should be represented, e.g the non so obvious `ng-class="{[CLASS_STYLE_1]: root.obj1[CLASS_STYLE_1], [CLASS_STYLE_2]: root.obj2[CLASS_STYLE_2]}"`

OzzieOrca:
Angular 1.6.4 may not be representative of what this issue is trying to accomplish, but changing the example in [my initial comment](https://github.com/angular/angular.js/issues/15905#issuecomment-293126981) from
```
ng-class="{'has-error': (testForm.testInput | showErrors)}"
```
to
```
ng-class="(testForm.testInput | showErrors) && 'has-error'"
```
http://plnkr.co/edit/1kWniq?p=preview

or to
```
ng-class="{'has-error': (testForm.testInput | showErrors) && true}"
```
[...]

PR #15990, opened 2017-05-15
fix($parse): always re-evaluate filters within literals when an input is an object

### LINKED ISSUE #15964: Filters are not re-evaluated with non-simple input when present in an object/array literal
This is something not handled correctly by 25f008f541d68b09efd7b428b648c6d4899e6972 which started shallow-watching the inputs to object/array literals (even if those inputs are non-simple).  But in some cases those inputs might go through filters which change output based on non-shallow state.

Before 25f008f541d68b09efd7b428b648c6d4899e6972 such expressions would throw an infdig error so this is not a regression.  However once ng-class [started using this](https://github.com/angular/angular.js/commit/60394a9d91dad8932fa900af7c8529837f1d4557) it caused [one version](https://github.com/angular/angular.js/issues/15960#issuecomment-299109412) of #15905 (the other version would be nested objects, this is the objects-passed-to-filters version).

Here is a test showing this problem (can be pasted into parserSpec.js):
```javascript
        it('should reevaluate filters in literals with non-primitive', inject(function($parse) {
          var filterCalls = 0;
          $filterProvider.register('xy', valueFn(function(input) {
            filterCalls++;
            return input.x + input.y
          }));

          var watchCalls = 0;
          scope.$watch("{key: (value | xy)}", function() {
            watchCalls++;
          });

          scope.$apply("value = {x: 1, y: 1}");
          expect(filterCalls).toBe(1);
          expect(watchCalls).toBe(1);

          scope.$apply();
          expect(filterCalls).toBe(2);
          expect(watchCalls).toBe(1);

          scope.$apply("value.x = 2");
          expect(filterCalls).toBe(3);
          expect(watchCalls).toBe(2);
        }));
```

Note: interceptors and filters need the same fix in order to solve #15905

jbedard:
Fixes #15964 

This changes the [compareObjectIdentity](https://github.com/angular/angular.js/blob/v1.6.4/src/ng/parse.js#L1785) flag from being per-expression to per-input.  This allows inputs to filters to not compare object entity, but other inputs such as `- x` or `{x: x}` to only compare object entity.  This also expands this `compareObjectIdentity = true` beyond just object literals.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -622,15 +622,44 @@ function isStateless($filter, filterName) {
   return !fn.$stateful;
 }
 
-function findConstantAndWatchExpressions(ast, $filter) {
+// Detect nodes which could depend on non-shallow state of objects
+function isPure(node, parentIsPure) {
+  switch (node.type) {
+    // Computed members might invoke a stateful toString()
+    case AST.MemberExpression:
+      if (node.computed) {
+        return false;
+      }
+      break;
+
+    // Unary always convert to primative
+    case AST.UnaryExpression:
+      return true;
+
+    // The binary + operator can invoke a stateful toString().
+    case AST.BinaryExpression:
+      return node.operator !== '+';
+
+    // Functions / filters probably read state from within objects
+    case AST.CallExpression:
+      return false;
+  }
+
+  return (undefined === parentIsPure) || parentIsPure;
+}
+
+function findConstantAndWatchExpressions(ast, $filter, parentIsPure) {
   var allConstants;
   var argsToWatch;
   var isStatelessFilter;
+
+  var astIsPure = ast.isPure = isPure(ast, parentIsPure);
+
   switch (ast.type) {
   case AST.Program:
     allConstants = true;
     forEach(ast.body, function(expr) {
-      findConstantAndWatchExpressions(expr.expression, $filter);
+      findConstantAndWatchExpressions(expr.expression, $filter, astIsPure);
       allConstants = allConstants && expr.expression.constant;
     });
     ast.constant = allConstants;
@@ -640,26 +669,26 @@ function findConstantAndWatchExpressions(ast, $filter) {
     ast.toWatch = [];
     break;
   case AST.UnaryExpression:
-    findConstantAndWatchExpressions(ast.argument, $filter);
+    findConstantAndWatchExpressions(ast.argument, $filter, astIsPure);
     ast.constant = ast.argument.constant;
     ast.toWatch = ast.argument.toWatch;
     break;
   case AST.BinaryExpression:
-    findConstantAndWatchExpressions(ast.left, $filter);
-    findConstantAndWatchExpressions(ast.right, $filter);
+    findConstantAndWatchExpressions(ast.left, $filter, astIsPure);
+    findConstantAndWatchExpressions(ast.right, $filter, astIsPure);
     ast.constant = ast.left.constant && ast.right.constant;
     ast.toWatch = ast.left.toWatch.concat(ast.right.toWatch);
     break;
   case AST.LogicalExpression:
-    findConstantAndWatchExpressions(ast.left, $filter);
-    findConstantAndWatchExpressions(ast.right, $filter);
+    findConstantAndWatchExpressions(ast.left, $filter, astIsPure);
+    findConstantAndWatchExpressions(ast.right, $filter, astIsPure);
     ast.constant = ast.left.constant && ast.right.constant;
     ast.toWatch = ast.constant ? [] : [ast];
     break;
   case AST.ConditionalExpression:
-    findConstantAndWatchExpressions(ast.test, $filter);
-    findConstantAndWatchExpressions(ast.alternate, $filter);
-    findConstantAndWatchExpressions(ast.consequent, $filter);
+    findConstantAndWatchExpressions(ast.test, $filter, astIsPure);
+    findConstantAndWatchExpressions(ast.alternate, $filter, astIsPure);
+    findConstantAndWatchExpressions(ast.consequent, $filter, astIsPure);
     ast.constant = ast.test.constant && ast.alternate.constant && ast.consequent.constant;
     ast.toWatch = ast.constant ? [] : [ast];
     break;
@@ -668,9 +697,9 @@ function findConstantAndWatchExpressions(ast, $filter) {
     ast.toWatch = [ast];
     break;
   case AST.MemberExpression:
-    findConstantAndWatchExpressions(ast.object, $filter);
+    findConstantAndWatchExpressions(ast.object, $filter, astIsPure);
     if (ast.computed) {
-      findConstantAndWatchExpressions(ast.property, $filter);
+      findConstantAndWatchExpressions(ast.property, $filter, astIsPure);
     }
     ast.constant = ast.object.constant && (!ast.computed || ast.property.constant);
     ast.toWatch = [ast];
@@ -680,7 +709,7 @@ function findConstantAndWatchExpressions(ast, $filter) {
     allConstants = isStatelessFilter;
     argsToWatch = [];
     forEach(ast.arguments, function(expr) {
-      findConstantAndWatchExpressions(expr, $filter);
+      findConstantAndWatchExpressions(expr, $filter, astIsPure);
       allConstants = allConstants && expr.constant;
       if (!expr.constant) {
         argsToWatch.push.apply(argsToWatch, expr.toWatch);
@@ -690,8 +719,8 @@ function findConstantAndWatchExpressions(ast, $filter) {
     ast.toWatch = isStatelessFilter ? argsToWatch : [ast];
     break;
   case AST.AssignmentExpression:
-    findConstantAndWatchExpressions(ast.left, $filter);
-    findConstantAndWatchExpressions(ast.right, $filter);
+    findConstantAndWatchExpressions(ast.left, $filter, astIsPure);
+    findConstantAndWatchExpressions(ast.right, $filter, astIsPure);
     ast.constant = ast.left.constant && ast.right.constant;
     ast.toWatch = [ast];
     break;
@@ -699,7 +728,7 @@ function findConstantAndWatchExpressions(ast, $filter) {
     allConstants = true;
     argsToWatch = [];
     forEach(ast.elements, function(expr) {
-      findConstantAndWatchExpressions(expr, $filter);
+      findConstantAndWatchExpressions(expr, $filter, astIsPure);
       allConstants = allConstants && expr.constant;
       if (!expr.constant) {
         argsToWatch.push.apply(argsToWatch, expr.toWatch);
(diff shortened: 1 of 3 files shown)


src/ng/parse.js
```diff
@@ -622,15 +622,51 @@ function isStateless($filter, filterName) {
   return !fn.$stateful;
 }
 
-function findConstantAndWatchExpressions(ast, $filter) {
+// Detect nodes which could depend on non-shallow state of objects
+function isShallowState(stack) {
```
jbedard: This could probably use a better name/terminology...

src/ng/parse.js
```diff
@@ -781,7 +819,7 @@ ASTCompiler.prototype = {
       filters: {},
       fn: {vars: [], body: [], own: {}},
       assign: {vars: [], body: [], own: {}},
-      inputs: []
+      inputs: {}
```
jbedard: Probably a nicer way of doing this...

src/ng/parse.js
petebacondarwin: `isStackPure`?

test/ng/parseSpec.js
```diff
@@ -2973,6 +2973,111 @@ describe('parser', function() {
...
+          scope.$digest();
+          expect(filterCalls).toBe(2);
+          expect(watcherCalls).toBe(2);
+        }));
+
+        it('should not reevaluate literals containing filters with non-primitive input that does support valueOf()' +
+            ' when the instance changes but valueOf() does not', inject(function($parse) {
+          var filterCalls = 0;
+          $filterProvider.register('foo', valueFn(function(input) {
+            filterCalls++;
```
petebacondarwin: `filterCalls` is never tested, is it needed?

test/ng/parseSpec.js
```diff
@@ -2973,6 +2973,111 @@ describe('parser', function() {
...
+          scope.$apply("a = 1");
+          expect(filterCalls).toBe(1);
+          expect(watcherCalls).toBe(1);
+
+          scope.$apply("a = 2");
+          expect(filterCalls).toBe(2);
+          expect(watcherCalls).toBe(2);
+        }));
+
+        it('should always reevaluate filters with literal input containing non-primitives',
```
jbedard: One way to test this better (which I'll add) would be something like `[(obj | filterThatConvertsToPrimitive)]`

src/ng/parse.js
```diff
@@ -622,15 +622,51 @@ function isStateless($filter, filterName) {
   return !fn.$stateful;
 }
 
-function findConstantAndWatchExpressions(ast, $filter) {
+// Detect nodes which could depend on non-shallow state of objects
+function isShallowState(stack) {
+  for (var i = 0; i < stack.length; i++) {
```
petebacondarwin: Do we need to walk the stack every time? It seems that we are either pushing or popping items on and off this stack and then re-evaluating the shallowStateness of the stack. Can we not cache the state of each stack and then we only need to evaluate the pushed item?

jbedard: Yeah that's probably true.  Probably store the result of the switch statement in the stack? I'll try it out...

src/ng/parse.js
lgalfaso: By doing this (and later `Object.keys`) you are abusing the fact that `Object.keys` returns the keys in the order that were added to assert that expressions as `foo.bar();foo.baz()` will be evaluated in the right order. Even when this is somehow true, this is not part of the spec and we should not depend on this out-of-spec fact.

src/ng/parse.js
```diff
@@ -622,15 +622,51 @@ function isStateless($filter, filterName) {
...
+      // Functions / filters probably read state from within objects
+      case AST.CallExpression:
+        return false;
+    }
+  }
+
+  return true;
+}
+
+function findConstantAndWatchExpressions(ast, $filter, stack) {
```
lgalfaso: I am not following completely the logic here, but by just looking at the code it looks like `stack` can be replace with a simple boolean (that the first called call with `true` and the rest do `ast.isShallowState = parentShallowState && isShallowState([ast]);` (and of course this can be simplified).

src/ng/parse.js
jbedard: I forgot the order even mattered, good point 👍

src/ng/parse.js
jbedard: Yeah I think that's possible. I guess the idea of the `stack` was that it could also potentially be used for other purposes, but I'll try your suggestion.

petebacondarwin: Yeah, I think this is what I was getting at with https://github.com/angular/angular.js/pull/15990#discussion_r116576729

jbedard:
I think I've addressed all the comments. Also added a `ng-class` test for the exact issue someone reported.

I'm also open to other ideas around this and similar issues such as #15905 if anyone has alternative ideas.

Note that if the $filter solution in this PR is extended to interceptors it will also fix #15905 *except* [one-time literals containing objects](https://github.com/angular/angular.js/blob/v1.6.4/test/ng/directive/ngClassSpec.js#L612). That one failure is caused by [this condition](https://github.com/angular/angular.js/blob/v1.6.4/src/ng/parse.js#L1913-L1915) which I really wish I could find an alternative solution for, especially since that condition forces workarounds in multiple places where interceports are used.  The other thing that would fix that one ng-class test is [dropping nested object support in ng-class](https://github.com/angular/angular.js/issues/15960) 😒

src/ng/parse.js
```diff
@@ -622,15 +622,43 @@ function isStateless($filter, filterName) {
...
+
+  return parentPureness;
+}
+
+function findConstantAndWatchExpressions(ast, $filter, parentIsPure) {
   var allConstants;
   var argsToWatch;
   var isStatelessFilter;
+
+  ast.isPure = parentIsPure = isPure(ast, (undefined === parentIsPure) || parentIsPure);
```
petebacondarwin: `(undefined === parentIsPure) || parentIsPure` -  I would move this logic into the `isPure()` function.

src/ng/parse.js
```diff
@@ -622,15 +622,43 @@ function isStateless($filter, filterName) {
...
+  switch (node.type) {
+    // Computed members might invoke a stateful toString()
+    case AST.MemberExpression:
+      if (node.computed) {
+        return false;
+      }
+      break;
+
+    case AST.UnaryExpression:
+      return true;
```
petebacondarwin: In this case would you not defer to the `parentPureness` also?

src/ng/parse.js
```diff
@@ -839,13 +867,16 @@ ASTCompiler.prototype = {
 
   watchFns: function() {
     var result = [];
-    var fns = this.state.inputs;
+    var inputs = this.state.inputs;
     var self = this;
-    forEach(fns, function(name) {
-      result.push('var ' + name + '=' + self.generateFunction(name, 's'));
+    forEach(inputs, function(input) {
+      result.push('var ' + input.name + '=' + self.generateFunction(input.name, 's'));
```
petebacondarwin: is there any performance/code-size gain in setting and using `var name = input.name;` here?
Or does the minification process deal with that automatically?

test/ng/parseSpec.js
```diff
@@ -2973,6 +2973,172 @@ describe('parser', function() {
           expect(watcherCalls).toBe(1);
         }));
 
+        it('should not reevaluate filters in literals with non-primitive input that does support valueOf()',
```
petebacondarwin: can we use some describe blocks to simplify the test descriptions here:

```
describe('containing literals', function() {
  describe('with non-primitive inputs', function() {
    describe('that support valueOf()', function() {
      it('should not reevaluate filters, inputs do not change', ...);
      it('should reevaluate filters, when an input valueOf() result changes', ...);
      it('should not reevaluate filters, when the input changes but valueOf() does not ', ...);
    });
  });
  describe('with non-primitive inputs', function() {
  });
});
```

you get the idea...

petebacondarwin (commented):
This looks pretty good; a few minor comments.
My main question is about the purity of `UnaryExpressions`.
Otherwise happy to land this.

src/ng/parse.js
jbedard: The idea is to allow shallow-watching the `input` in things like `fn(!input)` where `!parentIsPure`, but the child is.

I've tried to exclude case which can invoke `toString` like `"" + input` because those could be non-pure (when `toString` reads inner state).

src/ng/parse.js
jbedard: I think performance/code-size has no (noticeable) gain or loss either way, so I'd vote for whatever is more readable/prettier. My preference is how it is (no local var) but I can change it if others would prefer that.

jbedard:
Updated based on PR feedback. Also added more tests and added a separate commit to reorganize the tests. How those tests are organized is debatable but I think what I did is similar to the suggestion in the comment.

petebacondarwin (approved):
LGTM!
Thanks @jbedard - please squash and merge (assuming Travis is happy).

jbedard:
@lgalfaso have a chance to look at the latest update?

src/ng/parse.js
```diff
@@ -622,15 +622,44 @@ function isStateless($filter, filterName) {
...
+
+  return (undefined === parentIsPure) || parentIsPure;
+}
+
+function findConstantAndWatchExpressions(ast, $filter, parentIsPure) {
   var allConstants;
   var argsToWatch;
   var isStatelessFilter;
+
+  ast.isPure = parentIsPure = isPure(ast, parentIsPure);
```
lgalfaso: [nit] I find it slightly odd to change the value of the argument. I would just use `ast.isPure` below and keep the argument as is.

lgalfaso (approved):
LGTM!

Just one minor comment.

src/ng/parse.js
jbedard: I think I agree with that.  Could also change the `parentIsPure` param to `parentAst`, but I'll go with your suggestion unless someone says otherwise.

jbedard:
Made the change and merged into master and v1.6.x

src/ng/parse.js
```diff
@@ -622,15 +622,44 @@ function isStateless($filter, filterName) {
...
+    // Computed members might invoke a stateful toString()
+    case AST.MemberExpression:
+      if (node.computed) {
+        return false;
+      }
+      break;
+
+    // Unary always convert to primative
+    case AST.UnaryExpression:
+      return true;
```
gkalpak: Can't unary expressions invoke a stateful `valueOf()`?

jbedard: Yeah I think you're right.

Seems like any numeric operator (everything except `!`, `&&` and `||`?) will use `valueOf()`. However this [will be handled](https://github.com/angular/angular.js/blob/b5118ac6a9e0a327b31094b3fdcdc0432b23ad2f/src/ng/parse.js#L1829) so I don't think it will lead to bugs, but this method is still wrong.

Now that I type that... `&&` and `||` also don't convert to primitives and might return an object so those are also incorrect, but [we watch the full logical expression](https://github.com/angular/angular.js/blob/b5118ac6a9e0a327b31094b3fdcdc0432b23ad2f/src/ng/parse.js#L686) so it won't lead to bugs, but this method is still wrong.

So I don't think there's a bug but it might be nice to make this function more correct?

Once again, always wait for @gkalpak to review your PRs...

gkalpak: First of all, I don't know what `&&` and `||` have to do with this (since they are not unary operators) :confused:

Regarding `getValueOf()`, I don't think it is enough to "save" us (in cases where `valueOf()` returns non-primitive values). I admit it is very unlikely that this will affect any real app, but from a theoretical point of view, I think expressions such as `'+obj'`, where `obj` has a `valueOf()` method that returns an object will not be handled correctly (e.g. if the returned object stays the same by reference, but one of its properties changes).

gkalpak: Hm...OK maybe that is not true, because unary operators will always evaluate to the same thing for objects whose `valueOf` returns a non-primitve value (`!` will return `false`, `+/-` will return `NaN`), so this might actually work correctly :grin:

gkalpak: > I do still think that numeric unary/binary operators are non-pure because valueOf might be stateful. Which is a bug in isPure, but the use of getValueOf later on prevents a bug in $parse. I'll probably still correct the isPure function...

That's also my conclusion :smiley:

PR #15997, opened 2017-05-19
chore(i18n): fix and update CLDR to v30.0.1

### LINKED ISSUE #15976: Update locales from closure-library
As instructed on https://github.com/angular/angular.js/pull/11925#issuecomment-116909287, I'm opening a new issue now that https://github.com/google/closure-library/blob/master/closure/goog/i18n/datetimesymbols.js is updated with the CLDR release 31 data (which includes the changes in the mentioned PR) and Angular needs to be updated accordingly.

I just checked the code on the master branch and the v1.6.x branch, and both seem outdated.

Thank you.

gkalpak:
Fix, i18n update.

Locale files based on CLDR 29 and broken `i18n/` scripts.

Locale files based on CLDR 30.0.1 and working `i18n/` scripts.

No.

**Other information**:
Fixes #15976.


Narretz:
Do you know why the scripts broke? Because they must have worked before.
Could you also add the instructions for updating the CLDR files to the contributing.md?

gkalpak:
I don't know why or when the scripts/tests broke. It is possible that someone run the node scripts directly from inside the `i18n/` directory, instead of running the `i18n/generate.sh` script (which runs the tests and the node scripts).

There is a [README.md][1] inside `i18n/` that explains what each script and directory is for.

[1]: https://github.com/angular/angular.js/tree/master/i18n

i18n/spec/closureI18nExtractorSpec.js
```diff
@@ -75,7 +75,6 @@ describe('findLocaleId', function() {
   it('should not find localeId if data is missing', function() {
     expect(findLocaleId('', 'num')).toBeUndefined();
     expect(findLocaleId('aa', 'datetime')).toBeUndefined();
-    expect(findLocaleId('aa', 'randomType')).toBeUndefined();
```
Narretz: Why is this assertion bad? Or, what caused it to be bad?

i18n/src/parser.js
```diff
@@ -33,7 +46,12 @@ function parsePattern(pattern) {
       positive = patternParts[0],
       negative = patternParts[1];
 
-  var positiveParts = positive.split(DECIMAL_SEP),
+  // The parsing logic below assumes that there will always be a DECIMAL_SEP in the pattern.
```
Narretz: I'm a bit confused by this. Does it mean if a locale doesn't have a DECIMAL_SEP, we add it after the last ZERO, which means it will be discarded later (because nothing comes after it)? So we only add it for the sake of parsing?

i18n/spec/closureI18nExtractorSpec.js
gkalpak: This throws an "unknown type in findLocaleId: randomType" error. This what the next test verifies.
I assume that at some point someone decided to change the implementation, so that unknown types cause an error (vs returning `undefined`), but didn't update this test.

Maybe we should run these tests on CI.

i18n/src/parser.js
gkalpak: Pretty much. We add it for the sake of parsing, because what comes after the DECIMAL_SEP determines the min/max fraction size (depending on the number of ZERO and DIGIT occurrences) and also the `posSuf`. For example, if we have `#,##0$`, we convert it to `#,##0.$`, which (when parsed) results in `minFrac: 0`, `maxFrac: 0`, `posSuf: $` - i.e. have 0 fraction digits and tput `$` after the number (`posSuf`).

Narretz: Ok, I see. Can you please a short summary of this to the comment? I think this will be helpful.

gkalpak: Nobody reads comments. When you want to understand the code, you look at the tests 😛 
<sub>(I'll update the comment.)</sub>

gkalpak: Done.

Narretz (approved):

PR #16001, opened 2017-05-20
Issue with priority of execution in filters

puneeth8994:
Need to use track by $index at the end in ng-repeat or else the filter considers $index as a value that should be filtered out. This seems to be common issue.

docs update

priority of execution is important while using filters, limit, and track by $index

**Other information**:

### DIFF
--- a/docs/content/error/filter/notarray.ngdoc
+++ b/docs/content/error/filter/notarray.ngdoc
@@ -49,3 +49,12 @@ You could as well convert the object to an array using a filter such as
   {{ item }}
 </div>
 ```
+
+This error might occur as a type error when track by $index is used (Where the value of $index is 0 for the first element which will be used for filtering instead of the array) such as
+```html
+<input ng-model="search">
+<div ng-repeat="item in items track by $index | filter:search">
+  {{ item }}
+</div>
+```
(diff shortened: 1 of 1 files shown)


puneeth8994:
I signed it!

Narretz:
In general, the docs say that the track by expression must always go at the end, so it's definitely not optional to do it this way. 

However, you get a different error message when you include the "as" part in the filter: https://docs.angularjs.org/error/$parse/syntax?p0=as&p1=is an unexpected token&p2=20&p3=$index | filter:q as results&p4=as results, see https://plnkr.co/edit/4tSsoInjMZkLJs0AUmtC?p=preview

That means if we put an explanation in the notarray error message, we are missing errors that trigger the parse syntax error.
Ideally, we would add this info to https://code.angularjs.org/snapshot/docs/error/ngRepeat/iexp, but since this is triggered before Angular parses the expression, this is difficult to do.

Narretz:
I don't feel like this makes the docs better, for the reasons pointed out above. I've opened a new issue to track a possible improvement to the ngRepeat expression parsing: https://github.com/angular/angular.js/issues/16023

PR #16003, opened 2017-05-21
docs(guide): component example update

sercaneraslan:
There is no need empty function on Components Controller anymore. Components are had a empty Controller by the default.


gkalpak:
I think the `HeroDetailController` is defined as a placeholder, for the logic that is added in the next section. That said, I don't think it is a bad idea to show that the `controller` property is optional (when there is no logic).

PR #16008, opened 2017-05-24
test(support): verify support tests results in all tested browsers

mgol:

New tests.

It's not tested whether browsers pass or fail specific support tests used to skip particular test blocks.

All browsers have verified support tests results.

No.

- ~~Docs have been added / updated (for bug fixes / features)~~

**Other information**:

### DIFF
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -283,16 +283,16 @@ describe('injector', function() {
 
 
     describe('es6', function() {
-      if (support.ES6Function) {
+      if (support.shorthandMethods) {
         // The functions are generated using `eval` as just having the ES6 syntax can break some browsers.
-        it('should be possible to annotate functions that are declared using ES6 syntax', function() {
+        it('should be possible to annotate shorthand methods', function() {
           // eslint-disable-next-line no-eval
           expect(annotate(eval('({ fn(x) { return; } })').fn)).toEqual(['x']);
         });
       }
 
 
-      if (support.fatArrow) {
+      if (support.fatArrows) {
         it('should create $inject for arrow functions', function() {
           // eslint-disable-next-line no-eval
           expect(annotate(eval('(a, b) => a'))).toEqual(['a', 'b']);
@@ -300,7 +300,7 @@ describe('injector', function() {
       }
 
 
-      if (support.fatArrow) {
+      if (support.fatArrows) {
         it('should create $inject for arrow functions with no parenthesis', function() {
           // eslint-disable-next-line no-eval
           expect(annotate(eval('a => a'))).toEqual(['a']);
@@ -308,7 +308,7 @@ describe('injector', function() {
       }
 
 
-      if (support.fatArrow) {
+      if (support.fatArrows) {
         it('should take args before first arrow', function() {
           // eslint-disable-next-line no-eval
           expect(annotate(eval('a => b => b'))).toEqual(['a']);
--- a/test/helpers/support.js
+++ b/test/helpers/support.js
@@ -1,0 +1,20 @@
+'use strict';
+
+var supportTests = {
+  classes: '/^class\\b/.test((class C {}).toString())',
+  fatArrows: 'a => a',
+  shorthandMethods: '({ fn(x) { return; } })'
+};
+
+var support = {};
+
+for (var prop in supportTests) {
+  if (supportTests.hasOwnProperty(prop)) {
+    try {
+      // eslint-disable-next-line no-eval
+      support[prop] = !!eval(supportTests[prop]);
+    } catch (e) {
+      support[prop] = false;
+    }
+  }
+}
--- a/test/helpers/supportSpec.js
+++ b/test/helpers/supportSpec.js
@@ -1,0 +1,97 @@
+'use strict';
+
+describe('support test results', function() {
+  var expected, version, testName;
+  var userAgent = window.navigator.userAgent;
+
+  // Support: iOS 8 only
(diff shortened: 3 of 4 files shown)


test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
```
gkalpak: I would expect this file to be in `test/helpers/`. Why here? It is confusing.

test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
...
+
+describe('support test results', function() {
+  var expected, version, testName;
+  var userAgent = window.navigator.userAgent;
+
+  // Support: iOS 8 only
+  if (/iPhone OS 10_1\d(?:_\d+)? /.test(userAgent)) {
+    // iOS 8 official simulators have broken user agent (containing something like `iPhone OS 10_12_5`,
+    // i.e. the macOS version in place of the iOS version so they'd fall into an incorrect bucket.
```
gkalpak: Missing closing `)`.

test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
...
+    // Fix the user agent there.
+    // NOTE: Make sure the above check doesn't catch the real iOS 10!
+    userAgent = userAgent.replace(/iPhone OS 10(?:_\d)+/, 'iPhone OS 8_1');
```
gkalpak: `_\d` --> `_\d+`

test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  if (/edge\//i.test(userAgent)) {
```
gkalpak: Why not convert `userAgent` to lowercase once and avoid all the `i` flags? :smiley:

test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
...
+    expected = {
+      classes: true,
+      fatArrows: true,
+      shorthandMethods: true
+    };
+  } else if (/(?:msie|trident)/i.test(userAgent)) {
```
gkalpak: `/(?:msie|trident)/` --> `/msie|trident/`

test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
...
+    };
+  } else if (/chrome/i.test(userAgent)) {
+    // Catches Chrome on Android as well (i.e. the default
+    // Android browser on Android >= 4.4).
+    expected = {
+      classes: true,
+      fatArrows: true,
+      shorthandMethods: true
+    };
+  } else if (/\b9\.\d+(\.\d+)* safari/i.test(userAgent)) {
```
gkalpak: `/\b9\.\d+(\.\d+)*` --> `/\b9(?:\.\d+)+`

test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
...
+    expected = {
+      classes: true,
+      fatArrows: false,
+      shorthandMethods: true
+    };
+  } else if (/\b\d+\.\d+(\.\d+)* safari/i.test(userAgent)) {
```
gkalpak: `/\b\d+\.\d+(\.\d+)*` --> `/\b\d+(?:\.\d+)+`

test/ng/supportSpec.js
```diff
@@ -0,0 +1,91 @@
...
+      fatArrows: false,
+      shorthandMethods: true
+    };
+  } else if (/iphone os/i.test(userAgent)) {
+    expected = {
+      classes: true,
+      fatArrows: true,
+      shorthandMethods: true
+    };
+  } else if (/android 4\.[0-3]/i.test(userAgent)) {
```
gkalpak: This will break when Android 4.10 is out :stuck_out_tongue:

gkalpak (changes requested):

test/ng/supportSpec.js
mgol: That was my plan. I must have clicked the wrong directory. ;)

test/ng/supportSpec.js
mgol: Android 4.x is dead, we're safe. :)

test/ng/supportSpec.js
mgol: Why in `test/helpers`? `test/helpers/` contains helpers for tests, not actual tests.

test/ng/supportSpec.js
mgol: That would have to be `lowerCasedUserAgent` to be precise. And maybe we'll need case sensitivity one day?

test/ng/supportSpec.js
gkalpak: It contains helpers for tests and their tests (when they exist) :smiley:
For lack of a better directory, keeping the pec file next to the source is better than putting it into an arbitrary directory 😛

mgol:
@gkalpak PR updated, tests now pass.

gkalpak:
Why did tests previously fail?
LGTM except for `supportSpec` being in `test/ng/` (instead of `test/helpers/`), as is [privateMocksSpec.js][1] for example.

[1]: https://github.com/mgol/angular.js/blob/492cb3537522da51ed6ea0f8895fc929ff8add17/test/helpers/privateMocksSpec.js

mgol:
The spec file moved (I missed `privateMocksSpec.js` previously).

The tests were failing as I didn't add results for Safari 8 so it was treated the same as Safari 10. :-)

gkalpak (approved):

PR #16010, opened 2017-05-25
fix($sanitize): use appropriate inert document strategy for Firefox and Safari

petebacondarwin:
Both Firefox and Safari are vulnerable to XSS if we use an inert document
created via `document.implementation.createHTMLDocument()`.

Now we check for those vulnerabilities and then use a DOMParser or XHR
strategy if needed.

Thanks to @cure53 for the heads up on this issue.

### DIFF
--- a/src/ngSanitize/sanitize.js
+++ b/src/ngSanitize/sanitize.js
@@ -313,16 +313,78 @@ function $SanitizeProvider() {
     return obj;
   }
 
-  var inertBodyElement = (function(window) {
-    var doc;
-    if (window.document && window.document.implementation) {
-      doc = window.document.implementation.createHTMLDocument('inert');
+  /**
+   * Create an inert document that contains the dirty HTML that needs sanitizing
+   * Depending upon browser support we use one of three strategies for doing this.
+   * Support: Safari 10.x -> XHR strategy
+   * Support: Firefox -> DomParser strategy
+   */
+  var getInertBodyElement /* function(html: string): HTMLBodyElement */ = (function(window, document) {
+    var inertDocument;
+    if (document && document.implementation) {
+      inertDocument = document.implementation.createHTMLDocument('inert');
     } else {
       throw $sanitizeMinErr('noinert', 'Can\'t create an inert html document');
     }
-    var docElement = doc.documentElement || doc.getDocumentElement();
-    return docElement.getElementsByTagName('body')[0];
-  })(window);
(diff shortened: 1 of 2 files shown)


cure53:
:+1:

src/ngSanitize/sanitize.js
```diff
@@ -313,16 +313,72 @@ function $SanitizeProvider() {
...
+      xhr.responseType = 'document';
+      xhr.open('GET', 'data:text/html;charset=utf-8,' + html, false);
+      xhr.send(null);
+      var body = xhr.response.body;
+      body.firstChild.remove();
+      return body;
+    }
+
+    function getInertBodyElement_DOMParser(html) {
+      html = '<remove></remove>' + html;
```
mgol: Why is that needed?

petebacondarwin: One moment :-)

petebacondarwin: I added a comment!

gkalpak (approved):

test/ngSanitize/sanitizeSpec.js
```diff
@@ -266,14 +268,25 @@ describe('HTML', function() {
     });
   });
 
+  // See https://github.com/cure53/DOMPurify/blob/a992d3a75031cb8bb032e5ea8399ba972bdf9a65/src/purify.js#L439-L449
```
mgol: Could you add support comments? Here:
```js
// Support: Safari 10.1+
```
and below:
```js
// Support: Firefox 53+
```

petebacondarwin: Since they are feature sniffing, I was not sure if we needed to mention the type and version of the browser... I expect that Firefox do not consider their part a bug, which means that it will never be resolved. But in any case, there may be other browsers out there that exhibit the same behaviour.

Is the aim of these comments to indicate when we can remove it?

mgol: Yes, it's mostly to know if/when to remove it. But I commented in a wrong place, the tests should stay forever as they are, such comments belong more in source.

petebacondarwin: I added comments here: https://github.com/angular/angular.js/pull/16010/files#diff-699beb0d8c48a8257d3d70331bf6138dR319

koto:
LGTM.

rjamet:
LGTM too.

petebacondarwin:
Cherry-picked to 1.6.x

PR #16013, opened 2017-05-25
avoid publishing new releases to deprecated bower

graingert:
N/A

bower install angular
* installs latest version of angularjs *

bower install angular
installs old version of angular

yes, no new bower releases.

**Other information**:

### DIFF
--- a/scripts/bower/publish.sh
+++ b/scripts/bower/publish.sh
@@ -74,8 +74,7 @@ function prepare {
   do
     echo "-- Updating version in bower-$repo to $NEW_VERSION"
     cd $TMP_DIR/bower-$repo
-    replaceJsonProp "bower.json" "version" ".*" "$NEW_VERSION"
-    replaceJsonProp "bower.json" "angular.*" ".*" "$NEW_VERSION"
+    rm -f "bower.json"
     replaceJsonProp "package.json" "version" ".*" "$NEW_VERSION"
     replaceJsonProp "package.json" "angular.*" ".*" "$NEW_VERSION"
 


graingert:
This will probably require some doc fixes for how people can transition to npm based angular, eg use `/node_modules/` rather than `/bower_components/`

graingert:
currently this PR stops all new releases to bower, however it might be better to release 1.7 and above only on npm.

gkalpak:
TBH, I don't see the benefit in doing so. Although not recommended for use in new projects, bower is still maintained and there are people using it. Given the next-to-zero overhead for us, I don't see why we should make the lives of people still using bower more difficult than they already are 😈

graingert:
@gkalpak perhaps to start this off, I could make a PR that updates the tutorial to use plain old npm

gkalpak:
Well, that would be nice, but...the tutorial uses `angular-seed` as a starting point, so we need to update [angular/angular-seed][1]. And there are currently three PRs open for that ([#385][2], [#389][3], [#390][4]) :flushed:

We should get one of those in (probably #390) and _then_ you could work on the tutorial :grin:

[1]: https://github.com/angular/angular-seed
[2]: https://github.com/angular/angular-seed/pull/385
[3]: https://github.com/angular/angular-seed/pull/389
[4]: https://github.com/angular/angular-seed/pull/390

graingert:
@gkalpak #390 is closed

gkalpak:
Wrong link. Try again :)

dcherman:
Is there a cost to supporting bower?  It seems needlessly breaking to stop publishing to it even if it is deprecated, especially since it's probably an automated process that is doing so.

Narretz:
I agree that stopping publishing to bower is a needless BC. I also agree that we can point out the deprecation where possible, but let's not track this in this PR.

PR #16015, opened 2017-05-26
fix($parse): respect the interceptor.$stateful flag

jbedard:
~~This actually depends on #16009 because that (unnecessary) `$stateful` flag was previously being ignored in the case of the failing test.~~

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1798,15 +1798,9 @@ function $ParseProvider() {
             var lexer = new Lexer($parseOptions);
             var parser = new Parser(lexer, $filter, $parseOptions);
             parsedExpression = parser.parse(exp);
-            if (parsedExpression.constant) {
-              parsedExpression.$$watchDelegate = constantWatchDelegate;
-            } else if (oneTime) {
-              parsedExpression.oneTime = true;
-              parsedExpression.$$watchDelegate = oneTimeWatchDelegate;
-            } else if (parsedExpression.inputs) {
-              parsedExpression.$$watchDelegate = inputsWatchDelegate;
-            }
-            cache[cacheKey] = parsedExpression;
+            parsedExpression.oneTime = !!oneTime;
+
+            cache[cacheKey] = addWatchDelegate(parsedExpression);
           }
           return addInterceptor(parsedExpression, interceptorFn);
 
@@ -1931,9 +1925,21 @@ function $ParseProvider() {
       return unwatch;
     }
 
+    function addWatchDelegate(parsedExpression) {
+      if (parsedExpression.constant) {
+        parsedExpression.$$watchDelegate = constantWatchDelegate;
+      } else if (parsedExpression.oneTime) {
+        parsedExpression.$$watchDelegate = oneTimeWatchDelegate;
+      } else if (parsedExpression.inputs) {
+        parsedExpression.$$watchDelegate = inputsWatchDelegate;
+      }
+
+      return parsedExpression;
+    }
+
     function addInterceptor(parsedExpression, interceptorFn) {
       if (!interceptorFn) return parsedExpression;
-      var watchDelegate = parsedExpression.$$watchDelegate;
+
       var useInputs = false;
 
       var isDone = parsedExpression.literal ? isAllDefined : isDefined;
@@ -1953,18 +1959,16 @@ function $ParseProvider() {
 
       var fn = parsedExpression.oneTime ? oneTimeInterceptedExpression : regularInterceptedExpression;
 
-      // Propogate the literal/oneTime attributes
+      // Propogate the literal/oneTime/constant attributes
       fn.literal = parsedExpression.literal;
       fn.oneTime = parsedExpression.oneTime;
+      fn.constant = parsedExpression.constant;
 
-      // Propagate or create inputs / $$watchDelegates
-      useInputs = !parsedExpression.inputs;
-      if (watchDelegate && watchDelegate !== inputsWatchDelegate) {
-        fn.$$watchDelegate = watchDelegate;
-        fn.inputs = parsedExpression.inputs;
-      } else if (!interceptorFn.$stateful) {
-        // Treat interceptor like filters - assume non-stateful by default and use the inputsWatchDelegate
-        fn.$$watchDelegate = inputsWatchDelegate;
+      // Treat the interceptor like filters.
+      // If it is not $stateful then only watch its inputs.
+      // If the expression itself has no inputs then use the full expression as an input.
+      if (!interceptorFn.$stateful) {
+        useInputs = !parsedExpression.inputs;
         fn.inputs = (parsedExpression.inputs ? parsedExpression.inputs : [parsedExpression]).map(function(e) {
               // Remove the isPure flag of inputs when it is not absolute because they are now wrapped in a
               // potentially non-pure interceptor function.
@@ -1975,7 +1979,7 @@ function $ParseProvider() {
             });
       }
 
-      return fn;
+      return addWatchDelegate(fn);
     }
   }];
 }
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1884,28 +1884,37 @@ function $ParseProvider() {
     function oneTimeWatchDelegate(scope, listener, objectEquality, parsedExpression, prettyPrintExpression) {
       var isDone = parsedExpression.literal ? isAllDefined : isDefined;
       var unwatch, lastValue;
-      if (parsedExpression.inputs) {
-        unwatch = inputsWatchDelegate(scope, oneTimeListener, objectEquality, parsedExpression, prettyPrintExpression);
-      } else {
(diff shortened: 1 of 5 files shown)


jbedard:
For the failing test (which depends on #16009): previously it only passed because the `inputsWatchDelegate` was incorrectly used due to [this block](https://github.com/angular/angular.js/blob/b5118ac6a9e0a327b31094b3fdcdc0432b23ad2f/src/ng/parse.js#L1959-L1962) assigning the `inputs` then the `inputsWatchDelegate` was [later used](https://github.com/angular/angular.js/blob/b5118ac6a9e0a327b31094b3fdcdc0432b23ad2f/src/ng/parse.js#L1890-L1891) because of those inputs.  This is incorrect because a $stateful filter/interceptor must always be invoked, not only if it's inputs change, like the added tests demonstrate.

src/ng/parse.js
```diff
@@ -1950,23 +1956,20 @@ function $ParseProvider() {
...
 
-      // Propogate the literal/oneTime attributes
+      // Propogate the literal/oneTime/constant attributes
       fn.literal = parsedExpression.literal;
       fn.oneTime = parsedExpression.oneTime;
+      fn.constant = parsedExpression.constant;
 
-      // Propagate or create inputs / $$watchDelegates
-      useInputs = !parsedExpression.inputs;
-      if (watchDelegate && watchDelegate !== inputsWatchDelegate) {
```
gkalpak: If we remove this, what happens with "custom" `$$watchDelegates` (such as the ones added by `$interpolate`)?

jbedard: This actually fixes a bug because that lost `$$watchDelegate` knows nothing about the interceptor and is built specifically for that interpolate fn.

For example watching `$parse($interpolate("{{foo}}"), function() { return "bar"; })` will always output the "foo" var even though it is wrapped in an interceptor that always returns "bar", yet executing that parsed-expression will always return "bar"...

gkalpak: Where is the test for this fix then? 😛

gkalpak: FWIW, I am still confused about what happens to `$$watchDelegates`. when combined with interceptors? Are they just ignored?

jbedard: Yeah they are just ignored, because the interceptor might have made that `$$watchDelegate` completely invalid. However the returned ("interecepted"?) fn might have some constant/oneTime/inputs properties that might cause a `$$watchDelegate` to be added.

jbedard: Here's an example, I'll try to add a test for this...
```
$parse($interpolate("a{{b}}"), function(x) { return x + 'c'; })
```
`$interpolate` adds a special `$$watchDelegate` for an efficient way to watch the interpolated-expression, but that delegate of course knows nothing about the `x + 'c'` interceptor.  Previously this line we're commenting on would blindly copy the delegate creating an expression which produces the correct value when executed (with the `'c'` concatenated at the end) but the wrong value when watched due to the use of a `$$watchDelegate` that knows nothing about the `x + 'c'`...

jbedard: I've added a test for this: "should not propogate $$watchDelegate to the interceptor wrapped expression"

Narretz:
Looks like a few tests failed. And there's also an eslint error

jbedard:
Fixed the lint issue. Not sure what the test errors were (don't think they failed before), we'll see if they fail again...

src/ng/parse.js
gkalpak: I see. And we are fine with that because we never use `$$watchDelegate` with interceptors in core, right?
I might add a comment (where?) about that, i.e. the you can't use `$$watchDelegate` with interceptors.

jbedard:
I think the remaining test failure is actually caused by #16021. I'll probably end up trying to fix both in this PR since they seem overlap a lot...

jbedard:
@lgalfaso would be great to get your review/opinion if you have a chance

src/ng/parse.js
```diff
@@ -1931,40 +1934,58 @@ function $ParseProvider() {
...
+      } else if (parsedExpression.inputs) {
+        parsedExpression.$$watchDelegate = inputsWatchDelegate;
+      }
+
+      return parsedExpression;
+    }
+
+    function chainInterceptors(first, second) {
+      function chainedInterceptor(value, scope, locals) {
+        return second(first(value, scope, locals), scope, locals);
```
lgalfaso: Are you sure that `locals` should be sent to both interceptors?

src/ng/parse.js
```diff
@@ -1931,40 +1934,58 @@ function $ParseProvider() {
...
 
-      var isDone = parsedExpression.literal ? isAllDefined : isDefined;
-
-      function regularInterceptedExpression(scope, locals, assign, inputs) {
-        var value = useInputs && inputs ? inputs[0] : parsedExpression(scope, locals, assign, inputs);
-        return interceptorFn(value, scope, locals);
+      // Extract any existing interceptors out of the parsedExpression
+      // to ensure the original parsedExpression is always the $$intercepted
+      if (parsedExpression.$$interceptor) {
+        interceptorFn = chainInterceptors(parsedExpression.$$interceptor, interceptorFn);
```
lgalfaso: I understand why this chaining is needed, just wonder if there is a cleaner solution.

lgalfaso (commented):
Overall, I think it is a good change and a nice cleanup. Just some minor comments.

src/ng/parse.js
jbedard: Good question. I'll look into it and try to add some tests...

jbedard: This is equivalent to today. The interceptor [is passed the scope and locals](https://github.com/angular/angular.js/blob/v1.6.5/src/ng/parse.js#L1963). Then if it is double-intercepted then it will [pass them through](https://github.com/angular/angular.js/blob/v1.6.5/src/ng/parse.js#L1962) until reaching the original expression.

I would be tempted to remove these args from the interceptor though, and only pass them to the original wrapped expression. Probably in a separate PR though...

src/ng/parse.js
jbedard: This is all internal so it can always be changed later. I'm open to any ideas though 👍

src/ng/parse.js
```diff
@@ -1884,28 +1884,37 @@ function $ParseProvider() {
...
-      if (parsedExpression.inputs) {
-        unwatch = inputsWatchDelegate(scope, oneTimeListener, objectEquality, parsedExpression, prettyPrintExpression);
-      } else {
-        unwatch = scope.$watch(oneTimeWatch, oneTimeListener, objectEquality);
-      }
+
+      var exp = parsedExpression.$$intercepted || parsedExpression;
+      var post = parsedExpression.$$interceptor || identity;
+
+      var useInputs = parsedExpression.inputs && !exp.inputs;
```
gkalpak: How could `parsedExpresion.inputs` be truthy and `parsedException.$$intercepted.inputs` be falsy?
Can you give an example scenario?

jbedard: It's basically detecting [this case](https://github.com/angular/angular.js/blob/v1.6.5/src/ng/parse.js#L1973). This is when the intercepted expression had no inputs, so the expression itself becomes the input.

For example
* `$parse("x", stringify)`: `x` has no inputs, but we don't want to execute `stringify(x)` per-digest so we set `x` itself as the input and `useInputs = true` to handle this
* `$parse("x | number", stringify)`: `x | number` already has `x` as an input so we just continue using that input (and `useInputs = false`)

jbedard: The fact that this line basically "detects" the `useInputs = true` from `addInterceptor` is something I don't like about this PR atm. It already existed before, but I guess now it is spread across two methods which seems worse to me. But it's all internal and could always be changed...

src/ng/parse.js
jbedard: See https://github.com/angular/angular.js/pull/16091 which might be nice to merge before this PR.

src/ng/parse.js
gkalpak: I can't say I can fully wrap my head around this, but the tests pass, so... :smiley:

gkalpak (approved):

src/ng/parse.js
jbedard: Which part? Anything I can explain more/better?  There's a couple weird things being changed in this PR, and I also get confused after not looking at it for a while ^_^

gkalpak: The `useInputs` bit still confuses me (but in tests we trust).
<sub>(At some point I need to grab my debugger and go through some usecases, but this doesn't have to hold this PR back.)</sub>

jbedard: The `useInputs` is also the part I don't like.

Note that I think it should really be renamed to something such as `theInputIsTheInterceptedExpression` (where normally the input is the input **to** the intercepted expression).

It is also just a performance thing, dropping all `useInputs == true` logic would still function correctly.

src/ng/parse.js
jbedard: This method is now simplified after #16091 merged.

PR #16026, opened 2017-06-01
perf(jqLite): Add passive event listeners

alejandroSuch:
perf(jqLite): Add passive event listeners

Changes in sqLite to add passive event listeners

Chrome warns with lots of messages telling us to add passive event listeners

Warnings disappear and there are less page blockings

No

- [] Tests for the changes have been added (for bug fixes / features)

**Other information**:

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -851,6 +851,21 @@ function specialMouseHandlerWrapper(target, event, handler) {
 // These functions chain results into a single
 // selector.
 //////////////////////////////////////////
+function passiveOptionSupported() {
+  var supportsPassiveOption = false;
+  try {
+    var opts = Object.defineProperty({}, 'passive', {
+      get: function() {
+        supportsPassiveOption = true;
+      }
+    });
+    window.addEventListener('test', null, opts);
+    // eslint-disable-next-line no-empty
+  } catch (e) {
+  }
+  return supportsPassiveOption;
+}
+
 forEach({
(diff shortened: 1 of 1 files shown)


dcherman:
From the MDN:

> passive: A Boolean indicating that the listener will never call preventDefault(). If it does, the user agent should ignore it and generate a console warning. See Improving scrolling performance with passive listeners to learn more.

You can't guarantee this for any listeners other than your own, so this would be a massive, gigantic breaking change to `.on`

gkalpak:
Idea: Make the default (passive or not) configurable and allow overwriting the default per case (e.g. `ng-click="..." ng-click-passive="true/false"`).
This definitely needs more thought and discussion.

dcherman:
You wouldn't be able to do that without a corresponding change in jQuery though unless you're willing to diverge from their API.

Also FWIW, I think it would be valid to allow this option to be usable with the `ng-` event handlers, but it would be a nightmare for plugin authors if you couldn't guarantee how your handlers will behave.  Would rather see this be an opt-in behavior rather than potentially an opt-out.

gkalpak:
Oh, you are right. I thought this was only about the `ng-<event>` directives, not jqLite's `.on()`. So, yeah, we have no intention of deviating from the jQuery API (unless there is a significant reason), because the idea of `jqLite` is to remain as compatible as possible with jQuery.

Closing as _won't fix_. We can re-consider if/when jQuery makes a similar change.

PR #16035, opened 2017-06-05
chore(jenkins): do not publish to code.angularjs.org snapshot

petebacondarwin:
While the firewall continues to block the update ports
we will not try to publish there.

### DIFF
--- a/scripts/code.angularjs.org/publish.sh
+++ b/scripts/code.angularjs.org/publish.sh
@@ -62,7 +62,11 @@ function _update_code() {
 
   for backend in "$@" ; do
     echo "-- Refreshing code.angularjs.org: backend=$backend"
-    curl http://$backend:8003/gitFetchSite.php
+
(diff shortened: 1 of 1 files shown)


Narretz:
Can you add the consequence of not pushing to code.angularjs.org to the commit message? And that this will be fixed by the Firebase migration? Otherwise LGTM

scripts/code.angularjs.org/publish.sh
```diff
@@ -62,7 +62,11 @@ function _update_code() {
 
   for backend in "$@" ; do
     echo "-- Refreshing code.angularjs.org: backend=$backend"
-    curl http://$backend:8003/gitFetchSite.php
+
+    # FIXME: We gave up publishing to code.angularjs.org because the GCE autonatically removes firewall
```
gkalpak: _autonatically_ --> _automatically_

PR #16036, opened 2017-06-06
fix(Angular): deprecate angular.merge

### LINKED ISSUE #12653: Angular.merge() fails on scope merge
When performing angular.merge() on an object containing a $scope it fails with 
"RangeError: Maximum call stack size exceeded" 

.Merge() is attempting to deep copy the scope. Which just doesn't work. Angular.copy() states that copying states is not supported. 

I was going to submit a PR but i'm not quite sure what the behavior should be. Should it just set a reference to the scope on the new object? (this is what I, as a user, would prefer) or skip it? Whatever it is I don't think it should throw an exception

CodePen:
http://codepen.io/anon/pen/eNwxYj

This has been handled with Dates and RegExp in the following issues:
 #12409 #11974 #11720 #10519

### LINKED ISSUE #14941: angular.merge fails with MediaStream objects
Bug

`angular.merge` transforms MediaStream objects in `src` into plain JavaScript Objects.

With the template: https://plnkr.co/edit/4cBUxjdB4kya0s5Wtbxd?p=preview 
Simpler: https://plnkr.co/edit/LGUAcCTbYyf4UTxSbvBC?p=preview

``` js
navigator.mediaDevices.getUserMedia({ audio: true }).then(function (stream) {
  var dst = angular.merge({}, { stream: stream });
  console.error('Is valid ?', stream === dst.stream); // false
})
```

`angular.merge` (and derived) should not alter MediaStream objects.

Allow writing simpler/cleaner code. The workaround is simply to manually set the stream objects.

Google Chrome - Version 51.0.2704.106 (64-bit)
Other browsers probably also affected.

MediaStream objects should be cloned properly (MDN: `MediaStream.clone()`) in the case of `angular.clone`, or copied as is in `angular.merge` and `angular.extend`.

### LINKED ISSUE #15180: BUG: angular.merge breaks some objects
I believe this is a bug. The angular merge function will break some objects. I found this issue because of another plugin called Angular-Chart. This plugin build charts (with the use of chartjs) in the angular way.
I posted the issue there, but in depth i believe this is an angular bug.
because the object generated with ctx.createLinearGradient function will be merged like it is an object, but should be treaded as value.
If you want to know the whole issue, you can find it here:
[https://github.com/jtblin/angular-chart.js/issues/510](https://github.com/jtblin/angular-chart.js/issues/510)

Shortly, the merge function should check if the item is a normal object or a special object to deside if going deeper or just putting the object like a value it the target is the correct way.

bellow how jquery does this check

```
var normalObject = {};
var gradient = ctx.createLinearGradient(0, 0, 250, 0);

typeof normalObject -> "object"
typeof gradient -> "object"

toString.call(normalObject) -> "[object Object]"
toString.call(gradient) -> "[object CanvasGradient]"
```

if more info is needed, please let me know

### LINKED ISSUE #15992: angular.merge should copy Blob like Object.assign
**I'm submitting a ...**

- [x] bug report
- [ ] feature request

**Current behavior:**

`angular.merge` doesn't copy `Blob` objects from source to destination.
The property is copied but the value is no longer a `Blob` object but an empty `Object`.

**Expected / new behavior:**

The expected behaviour would be to copy the `Blob` as is from source to destination.
`Object.assign` behaves this way (doesn't have this bug).

```
var aFileParts = ['<a id="a"><b id="b">hey!</b></a>']; // an array consisting of a single DOMString
var oMyBlob = new Blob(aFileParts, {type : 'text/html'}); // the blob
var obj1 = {blob: oMyBlob, foo: 1};
var obj2 = angular.merge({}, obj1, {foo: 2});
var obj3 = Object.assign({}, obj1, {foo: 2});
console.assert(obj2.blob instanceof Blob === obj3.blob instanceof Blob, 'angular.merge should copy Blobs like Object.assign');
```

**Angular version:** 1.6.1

**Browser:** [Chrome 57 Mac OS X El Capitan ]

**Anything else:**

Narretz:
This function has problems with special object types but since it's not used in core,
it is not worth implementing fixes for these cases.
A general purpose library like lodash (provides `merge`) should be used instead.

Closes #12653
Closes #14941
Closes #15180
Closes #15992

Deprecation

Unclear API support

Deprecation / docs update

No

- ~[ ] Tests for the changes have been added (for bug fixes / features)~

**Other information**:
Not sure if this should be marked fix or feat
~If we deprecate only in 1.7.0, then the known issue section can still go in 1.6.x~

### DIFF
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -421,6 +421,20 @@ function extend(dst) {
 * Unlike {@link angular.extend extend()}, `merge()` recursively descends into object properties of source
 * objects, performing a deep copy.
 *
+* @deprecated
+* sinceVersion="1.6.5"
+* This function is deprecated, but will not be removed in the 1.x lifecycle.
+* There are a edge cases (see {@link angular.merge#known-issues known issues}) that are not
+* supported by this function. We suggest
+* using [lodash's merge()](https://lodash.com/docs/4.17.4#merge) instead.
+*
+* @knownIssue
+* This is a list of (known) object types that cannot be / are incorrectly handled by this function:
+* - [`Blob`](https://developer.mozilla.org/docs/Web/API/Blob)
+* - [`MediaStream`](https://developer.mozilla.org/docs/Web/API/MediaStream)
+* - [`CanvasGradient`](https://developer.mozilla.org/docs/Web/API/CanvasGradient)
+* - AngularJS {@link $rootScope.Scope scopes};
+*
 * @param {Object} dst Destination object.
 * @param {...Object} src Source object(s).
 * @returns {Object} Reference to `dst`.
--- a/src/Angular.js
+++ b/src/Angular.js
@@ -424,12 +424,12 @@ function extend(dst) {
 * @deprecated
 * sinceVersion="1.6.5"
 * This function is deprecated, but will not be removed in the 1.x lifecycle.
-* There are a edge cases (see {@link angular.merge#known-issues known issues}) that are not
+* There are edge cases (see {@link angular.merge#known-issues known issues}) that are not
 * supported by this function. We suggest
 * using [lodash's merge()](https://lodash.com/docs/4.17.4#merge) instead.
 *
 * @knownIssue
-* This is a list of (known) object types that cannot be / are incorrectly handled by this function:
+* This is a list of (known) object types that are not handled correctly by this function:
 * - [`Blob`](https://developer.mozilla.org/docs/Web/API/Blob)
 * - [`MediaStream`](https://developer.mozilla.org/docs/Web/API/MediaStream)
 * - [`CanvasGradient`](https://developer.mozilla.org/docs/Web/API/CanvasGradient)


src/Angular.js
```diff
@@ -421,6 +421,20 @@ function extend(dst) {
 * Unlike {@link angular.extend extend()}, `merge()` recursively descends into object properties of source
 * objects, performing a deep copy.
 *
+* @deprecated
+* sinceVersion="1.6.5"
+* This function is deprecated, but will not be removed in the 1.x lifecycle.
+* There are a edge cases (see {@link angular.merge#known-issues known issues}) that are not
```
gkalpak: _a edge cases known issues_ --> _???_

src/Angular.js
```diff
@@ -421,6 +421,20 @@ function extend(dst) {
...
+* supported by this function. We suggest
+* using [lodash's merge()](https://lodash.com/docs/4.17.4#merge) instead.
+*
+* @knownIssue
+* This is a list of (known) object types that cannot be / are incorrectly handled by this function:
```
gkalpak: The `cannot be / are incorrectly` doesn't read that well. I would split it up (e.g. `cannot be handled or are handled incorrectly`) or change it to `cannot be correctly handled` or something.

src/Angular.js
Narretz: Get rid of the `a` I guess. Did you read over the `(see known issues)`?

gkalpak: Oh, right. I missed `(see ` :smiley: So, just remove `a` then.

petebacondarwin (approved):

PR #16040, opened 2017-06-07
test(ngMock): fix Firefox crashes on Travis

gkalpak:
This test keeps causing Firefox 47 (currently used on Travis) to crash and fail the build. The test passes locally (on Firefox 53). Lowering the loop count from 1000 to 100 seems to fix the issue.
(Note: The crach only affects the mocked implementation of `$interval` and does not happen locally.)

### DIFF
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -358,12 +358,12 @@ describe('ngMock', function() {
       $interval(function() { counterA++; });
       $interval(function() { counterB++; }, 0);
 
-      $interval.flush(1000);
-      expect(counterA).toBe(1000);
-      expect(counterB).toBe(1000);
-      $interval.flush(1000);
-      expect(counterA).toBe(2000);
-      expect(counterB).toBe(2000);
+      $interval.flush(100);
+      expect(counterA).toBe(100);
+      expect(counterB).toBe(100);
+      $interval.flush(100);
+      expect(counterA).toBe(200);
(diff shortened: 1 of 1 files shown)


mgol (approved):

Narretz:
If I remember correctly, browsers can't actually guarantee an interval of 1ms. Maybe we should instead set it to 10ms to relieve the pressure? Or do you think the condition in the test with 1000 loops is unrealistic and doesn't need to be tested?

petebacondarwin:
But this is a mock interval, right? We should have any reliance upon the real browser's ability to handler short intervals.

Narretz:
You are right, I read over the fact that this is only the mock implementation. But it looks like the "fake" interval of 1ms also causes the browsers in question problems - which could maybe be handled by using a slightly higher interval, which would also be more in line with what browsers do with real intervals.

gkalpak:
I haven't tried it but I suspect it is something about the loop count that throws Firefox off. Even if we increased the faked interval to 10ms, using `flush(10000)` would result in the same number of iterations (and probably the same crashing). Given that this only affects the mock implementation and that I can't reproduce locally with latest Firefox, I am fine going either way. Note though, that this is not specific to the 0-delay interval (e.g. you can cause Firefox 47 to crash on Travis by explicitly using `$interval(1)` in a test.)

Narretz:
Ok, in that case I'm fine with changing just the test

petebacondarwin:
OK.

PR #16042, opened 2017-06-08
Fix broken link - point to list of Pluralsight courses on AngularJS

sauravbhattacharya001:
AngularJS docs update: Currently, the link to Pluralsight at the bottom of the page https://docs.angularjs.org/guide/external-resources is broken. This will edit the link to a generic list of all AngularJs course on Pluralsight. This is an isolated UI change to a public facing document, therefore there are no breaking changes anticipated.

### DIFF
--- a/docs/content/guide/external-resources.ngdoc
+++ b/docs/content/guide/external-resources.ngdoc
@@ -139,7 +139,7 @@ You can find a larger list of AngularJS external libraries at [ngmodules.org](ht
   [CodeAcademy](http://www.codecademy.com/courses/javascript-advanced-en-2hJ3J/0/1),
   [CodeSchool](https://www.codeschool.com/courses/shaping-up-with-angular-js)
 * **Paid online:**
-  [Pluralsight (3 courses)](http://www.pluralsight.com/training/Courses/Find?highlight=true&searchTerm=angularjs),
+  [Pluralsight](https://app.pluralsight.com/library/search?q=angularjs),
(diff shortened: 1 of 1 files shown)


zalenix:
I signed it!

zalenix:
From the log, there were 2 types of errors:

```
launcher.sauce - Heartbeat to firefox 47 failed
```
and
```
PROXY 127.0.0.1:51852 failed to send half-close
```

Can someone look into why this might be happening? Also, is it worth considering running doc builds separately?

docs/content/guide/external-resources.ngdoc
```diff
@@ -139,7 +139,7 @@ You can find a larger list of AngularJS external libraries at [ngmodules.org](ht
   [CodeAcademy](http://www.codecademy.com/courses/javascript-advanced-en-2hJ3J/0/1),
   [CodeSchool](https://www.codeschool.com/courses/shaping-up-with-angular-js)
 * **Paid online:**
-  [Pluralsight (3 courses)](http://www.pluralsight.com/training/Courses/Find?highlight=true&searchTerm=angularjs),
+  [Pluralsight](https://app.pluralsight.com/library/search?q=angularjs),
```
gkalpak: This URL requires the user to sign in. I think we should use https://www.pluralsight.com/search?q=angularjs instead.

gkalpak (changes requested):

docs/content/guide/external-resources.ngdoc
zalenix: Thank you! I have made the change.

PR #16046, opened 2017-06-10
Improve readability of logging on Travis

Narretz:
chore: 
- it hides the dots for unit tests on Travis, which makes it easier to find the separate test tasks and results.
- avoids building package for unit tests (makes the unit test job slightly faster)
- re-arranges the grunt task list to make it easier to read

### DIFF
--- a/karma-shared.conf.js
+++ b/karma-shared.conf.js
@@ -10,8 +10,16 @@ module.exports = function(config, specificOptions) {
     browserDisconnectTimeout: 10000,
     browserDisconnectTolerance: 2,
     browserNoActivityTimeout: 30000,
-
-
+    reporters: ['spec'],
+    specReporter: {
+      maxLogLines: 5,             // limit number of lines logged per test
+      suppressErrorSummary: true, // do not print error summary
+      suppressFailed: false,      // do not print information about failed tests
+      suppressPassed: true,      // do not print information about passed tests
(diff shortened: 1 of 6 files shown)


karma-shared.conf.js
```diff
@@ -10,8 +10,16 @@ module.exports = function(config, specificOptions) {
     browserDisconnectTimeout: 10000,
     browserDisconnectTolerance: 2,
     browserNoActivityTimeout: 30000,
-
-
+    reporters: ['spec'],
+    specReporter: {
+      maxLogLines: 5,             // limit number of lines logged per test
+      suppressErrorSummary: true, // do not print error summary
+      suppressFailed: false,      // do not print information about failed tests
```
petebacondarwin: I guess that these two lines do not hide the actual failure information, which is crucial to understanding why the job fails?

Narretz: Exactly. "suppressFailed: true" would hide the failures. "suppressErrorSummary: true" is set because the summary is basically a list of all errors printed after all tests have run - not very useful if you only show the errors anyway.
We should probably increase maxLogLines to see more of failures with good stack traces.

petebacondarwin (approved):

petebacondarwin:
@Narretz do we care to cherry pick to 1.6?

PR #16059, opened 2017-06-16
fix($parse): do not shallow-watch inputs to one-time intercepted expressions

jbedard:
This was missed in aac5623247a86681cbe0e1c8179617b816394c1d

Depends on #15958, will be fixed in master by #16015 (I've added the test there as well).


Narretz:
LGTM. Please merge after Travis is green

PR #16061, opened 2017-06-17
fix(copy): support copying MediaStream object

### LINKED ISSUE #16055: angular.copy does not support MediaStream object
**I'm submitting a ...**

- [x] bug report
- [ ] feature request

**Current behavior:**

angular.copy does not work with (Local)MediaStream objects. This affects the watchability of objects that contain such an object.

**Expected / new behavior:**

https://plnkr.co/edit/B3LebuAQPoj6kRgS2eCG?p=preview

**Angular version:** 1.x.y

**Browser:** [all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

**Anything else:**

As always with these things, it's debatable if this needs to be fixed (this is probably not the last built.-in that can't be copied). Lodash's clone implementation works, but is also pretty  extensive. Maybe we could take a hint from there.

m-amr:
use MediaStream clone method to clone mediastream object

Closes #16055

Bug, fix Issue https://github.com/angular/angular.js/issues/16055

angular.copy does not work with (Local)MediaStream objects. This affects the watchability of objects that contain such an object.

copy MediaStream object using `MediaStream.clone()` based on documentation
https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/clone

No

**Other information**:

### DIFF
--- a/test/AngularSpec.js
+++ b/test/AngularSpec.js
@@ -603,6 +603,18 @@ describe('angular', function() {
       /* eslint-enable */
     });
 
+    it('should copy media stream objects', function() {
+      var source = new MediaStream();
+      var destination = copy(source);
+
+      expect(destination.id).toMatch(/^((\w+)-){4}(\w+)$/);
+      expect(typeof destination.active).toBe('boolean');
+      expect(source.id).not.toBe(destination.id);
+      expect(source.active).toBe(destination.active);
+
+
+    });
+
     it('should copy source until reaching a given max depth', function() {
       var source = {a1: 1, b1: {b2: {b3: 1}}, c1: [1, {c2: 1}], d1: {d2: 1}};
       var dest;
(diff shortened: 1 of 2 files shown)


test/AngularSpec.js
```diff
@@ -603,6 +603,18 @@ describe('angular', function() {
       /* eslint-enable */
     });
 
+    it('should copy media stream objects', function() {
+      var source = new MediaStream();
```
gkalpak: Not all browsers support `MediaStream` (which seems to cause Travis to fail).

gkalpak (changes requested):
Besides the failing tests, I don't think this is a good idea. See https://github.com/angular/angular.js/issues/16055#issuecomment-312623547.

PR #16064, opened 2017-06-21
fix($timeout/$interval): do not trigger a digest when cancelling a $timeout/$interval

### LINKED ISSUE #16057: Debounced ng-model-options in v1.6.* fires $digests on each keypress (worked fine in v1.5.*)
**I'm submitting a ...**

- [x] bug report
- [ ] feature request

**Current behavior:**

Video here: https://www.screencast.com/t/TFhK7wyaG !!!
Angular 1.6.* - <input  /> with ng-model-options="{debounce:1000}"  fires $digest cycles on each keypress.
Angular 1.5.* - works perfect!
https://codepen.io/anon/pen/BZQQMG?editors=1010
https://codepen.io/anon/pen/EXNNqN?editors=1010

**Expected / new behavior:**

`ng-model-options` should work in Angular 1.6.* as it worked in Angular 1.5.*.

**Angular version:** 1.6.* has bug (worked fine in 1.5.*)

**Browser:** [all]

**Anything else:**

jbedard:
Previously `.catch(noop)` was used on the rejected timeout/interval to prevent an unhandled rejection error. However this would schedule a deferred task to run the `noop`. If the cancelling was outside a digest this could cause a new digest such as with the ng-model `debounce` option.

Fixes #16057

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -351,7 +351,7 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
     state.pending = undefined;
     try {
       for (var i = 0, ii = pending.length; i < ii; ++i) {
-        state.pur = true;
+        markQStateExceptionHandled(state);
         promise = pending[i][0];
         fn = pending[i][state.status];
         try {
@@ -378,8 +378,8 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
     // eslint-disable-next-line no-unmodified-loop-condition
     while (!queueSize && checkQueue.length) {
       var toCheck = checkQueue.shift();
-      if (!toCheck.pur) {
-        toCheck.pur = true;
+      if (!isStateExceptionHandled(toCheck)) {
+        markQStateExceptionHandled(toCheck);
         var errorMessage = 'Possibly unhandled rejection: ' + toDebugString(toCheck.value);
         if (isError(toCheck.value)) {
           exceptionHandler(toCheck.value, errorMessage);
@@ -391,7 +391,7 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
   }
 
   function scheduleProcessQueue(state) {
-    if (errorOnUnhandledRejections && !state.pending && state.status === 2 && !state.pur) {
+    if (errorOnUnhandledRejections && !state.pending && state.status === 2 && !isStateExceptionHandled(state)) {
       if (queueSize === 0 && checkQueue.length === 0) {
         nextTick(processChecks);
       }
@@ -670,4 +670,14 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
   $Q.race = race;
 
   return $Q;
+}
+
+function isStateExceptionHandled(state) {
+  return !!state.pur;
+}
+function markQStateExceptionHandled(state) {
+  state.pur = true;
+}
+function markQExceptionHandled(q) {
+  markQStateExceptionHandled(q.$$state);
 }
--- a/test/ng/directive/ngModelOptionsSpec.js
+++ b/test/ng/directive/ngModelOptionsSpec.js
@@ -459,8 +459,14 @@ describe('ngModelOptions', function() {
           $rootScope.$watch(watchSpy);
 
           helper.changeInputValueTo('a');
+          $timeout.flush(2000);
           expect(watchSpy).not.toHaveBeenCalled();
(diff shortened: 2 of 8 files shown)


test/ngAnimate/animateCssSpec.js
```diff
@@ -1328,8 +1328,7 @@ describe('ngAnimate $animateCss', function() {
           animator.end();
 
           expect(element.data(ANIMATE_TIMER_KEY)).toBeUndefined();
-          $timeout.flush();
```
jbedard: This is an example of how this change can be "breaking" for some tests.

jbedard:
For now I have not removed the 2 `promise.catch(noop)` calls from resource.js. I think both are already cases where another promise will be resolved so I didn't think it was worth either a) exposing the new helper or b) accessing the internal `$$state.pur`.

test/ngAnimate/animateCssSpec.js
Narretz: So it's probably not a breaking change? Should we mention this in the review anyway? (But no under Breaking Changes)

Narretz:
Just saw that the unit tests failed with lots of compile test failures.

src/ng/q.js
```diff
@@ -378,8 +378,8 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
     // eslint-disable-next-line no-unmodified-loop-condition
     while (!queueSize && checkQueue.length) {
       var toCheck = checkQueue.shift();
-      if (!toCheck.pur) {
-        toCheck.pur = true;
+      if (isStateExceptionHandled(toCheck)) {
```
Narretz: this must be `!isStateExceptionHandled(toCheck)`

test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -459,11 +459,12 @@ describe('ngModelOptions', function() {
           $rootScope.$watch(watchSpy);
 
           helper.changeInputValueTo('a');
+          expect(function() {$browser.defer.flush();}).toThrowError('No deferred tasks to be flushed');
```
Narretz: This doesn't work correctly. It still flushes the task and triggers a digest. I haven't found out why exactly it doesn't work, but I suspect it's because of a mismatch between the real browser and the mock browser.

Narretz (changes requested):

test/ng/directive/ngModelOptionsSpec.js
Narretz: The assumption that there are no deferred tasks is wrong at this point, because there's obviously a pending timeout.
We just need to check that the watchFn doesn't get called. 
If you put in this:
```
          helper.changeInputValueTo('a');
          $timeout.flush(2000);
          helper.changeInputValueTo('b');
          $timeout.flush(2000);
```
the test will fail without the patch. The second timeout.flush is crucial - without it, the watchFn spy doesn't get called, but *with* it, it gets called twice (without the patch).

jbedard:
Updated. Thanks for pointing out the issues @Narretz

test/ng/directive/ngModelOptionsSpec.js
```diff
@@ -449,7 +449,7 @@ describe('ngModelOptions', function() {
         });
 
 
-        it('should not trigger digest while debouncing', function() {
+        it('should not trigger digest while debouncing', inject(function($browser) {
```
gkalpak: Why this?

jbedard: Because it was calling `$browser.defer.flush()` like the other tests (which is wrong in this case). I'll undo this...

jbedard: Fixed.

test/ngAnimate/animateCssSpec.js
Narretz: @jbedard Can you clarify how this can be "breaking"?

abobwhite:
Would this have affected `$timeout.cancel` from ever returning `true` in some cases? I have code that used to work fine but now the `cancel` function always returns `false` and the timeout is never cancelled.

jbedard:
I don't think this would have changed [the return value](https://github.com/angular/angular.js/commit/a222d0b452622624dc498ef0b9d3c43647fd4fbc#diff-7e6919ccdaa77645580e865e271080eeR91). Also doesn't look like `$browser.defer.cancel` changed anytime recently (but I didn't look too hard)...

PR #16075, opened 2017-06-29
fix($http) No error for json-like repsonses without "application/json" content-type

### LINKED ISSUE #16027: defaultHttpResponseTransform tries to parse as Json "text/plain" that isn't Json
**I'm submitting a ...**

- [x ] bug report
- [ ] feature request

**Current behavior:**

```
function defaultHttpResponseTransform(data, headers) {
  if (isString(data)) {
    // Strip json vulnerability protection prefix and trim whitespace
    var tempData = data.replace(JSON_PROTECTION_PREFIX, '').trim();

    if (tempData) {
      var contentType = headers('Content-Type');
      if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData)) {
        try {
          data = fromJson(tempData);
        } catch (e) {
          throw $httpMinErr('baddata', 'Data must be a valid JSON object. Received: "{0}". ' +
          'Parse error: "{1}"', data, e);
        }
      }
    }
  }

  return data;
}
```

In defaultHttpResponseTransform() if the response data passes the isJsonLike() function, the function tries to transform the data using fromJson().  This may fail if the data isn't actually Json.  If a server sends contentType "text/plain" with data "{a}" or "[a]" for example, these will result in angular throwing an error.

**Expected / new behavior:**

If the contentType is "text/plain", the defaultHttpResponseTransform should return the data without trying to change it.

**Angular version:** 1.6.4

katsos:
Don't throw error if http response has json-like data but content-type is not "application/json". Instead return data unprocessed.

Fixes https://github.com/angular/angular.js/issues/16027
This is not a breaking change.

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -137,11 +137,13 @@ function defaultHttpResponseTransform(data, headers) {
 
     if (tempData) {
       var contentType = headers('Content-Type');
-      if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData)) {
+      var hasJsonContentType = contentType && (contentType.indexOf(APPLICATION_JSON) === 0);
+
+      if (hasJsonContentType || isJsonLike(tempData)) {
         try {
           data = fromJson(tempData);
         } catch (e) {
-          if (!contentType || contentType && contentType.indexOf(APPLICATION_JSON)) {
+          if (!hasJsonContentType) {
             return data;
           }
           throw $httpMinErr('baddata', 'Data must be a valid JSON object. Received: "{0}". ' +
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -141,8 +141,8 @@ function defaultHttpResponseTransform(data, headers) {
         try {
           data = fromJson(tempData);
         } catch (e) {
-          if ((contentType && (contentType.indexOf(APPLICATION_JSON) !== 0))) {
-            return tempData;
+          if (!contentType || contentType && contentType.indexOf(APPLICATION_JSON)) {
+            return data;
           }
           throw $httpMinErr('baddata', 'Data must be a valid JSON object. Received: "{0}". ' +
           'Parse error: "{1}"', data, e);
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -141,6 +141,9 @@ function defaultHttpResponseTransform(data, headers) {
         try {
           data = fromJson(tempData);
         } catch (e) {
+          if ((contentType && (contentType.indexOf(APPLICATION_JSON) !== 0))) {
+            return tempData;
+          }
           throw $httpMinErr('baddata', 'Data must be a valid JSON object. Received: "{0}". ' +
           'Parse error: "{1}"', data, e);
         }
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1381,25 +1381,21 @@ describe('$http', function() {
           });
 
           it('should not throw an error if JSON is invalid but content-type is not application/json', function() {
-            var errCallback = jasmine.createSpy('error');
             $httpBackend.expect('GET', '/url').respond('{abcd}', {'Content-Type': 'text/plain'});
 
-            $http.get('/url').then(callback).catch(errCallback);
+            $http.get('/url').then(callback);
             $httpBackend.flush();
 
             expect(callback).toHaveBeenCalledOnce();
-            expect(errCallback).not.toHaveBeenCalled();
           });
 
           it('should not throw an error if JSON is invalid but content-type is not specified', function() {
-            var errCallback = jasmine.createSpy('error');
             $httpBackend.expect('GET', '/url').respond('{abcd}');
 
-            $http.get('/url').then(callback).catch(errCallback);
+            $http.get('/url').then(callback);
             $httpBackend.flush();
 
             expect(callback).toHaveBeenCalledOnce();
-            expect(errCallback).not.toHaveBeenCalled();
           });
 
           it('should return response unprocessed if JSON is invalid but content-type is not application/json', function() {
--- a/test/ng/httpSpec.js
+++ b/test/ng/httpSpec.js
@@ -1380,8 +1380,28 @@ describe('$http', function() {
             expect(errCallback.calls.mostRecent().args[0]).toEqualMinErr('$http', 'baddata');
           });
 
-        });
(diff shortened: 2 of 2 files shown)


src/ng/http.js
```diff
@@ -141,6 +141,9 @@ function defaultHttpResponseTransform(data, headers) {
         try {
           data = fromJson(tempData);
         } catch (e) {
+          if ((contentType && (contentType.indexOf(APPLICATION_JSON) !== 0))) {
```
gkalpak: Too much parenthesis (even for me :grin:).

src/ng/http.js
```diff
@@ -141,6 +141,9 @@ function defaultHttpResponseTransform(data, headers) {
...
+            return tempData;
```
gkalpak: Why return `tempData`? If we decide to return the original response, it should be `data`.

test/ng/httpSpec.js
```diff
@@ -1380,8 +1380,28 @@ describe('$http', function() {
...
+          });
 
+          it('should return response unprocessed if JSON is invalid but content-type is not application/json', function() {
+            var response = '{abcd}';
+            $httpBackend.expect('GET', '/url').respond(response, {'Content-Type': 'text/plain'});
+
+            $http.get('/url').then(callback);
+            $httpBackend.flush();
+
+            expect(callback.calls.mostRecent().args[0].data).toMatch(response);
```
gkalpak: Why not `.toBe()`?

test/ng/httpSpec.js
```diff
@@ -1380,8 +1380,28 @@ describe('$http', function() {
...
+          });
+
+        });
```
gkalpak: There is no test for not specified `Content-Type` header 😞

test/ng/httpSpec.js
katsos: `toBe()` looks for reference equality, `toMatch()` is responsible for matching strings and regexs. (according to jasmine documentation)

gkalpak: [`toMatch()`][1] is for regular expressions (you can pass a string, but it will be converted to a regular expression). It works in this case, but is unnecessary.

[`toBe()`][2] tests for strict equality (`===`), which works when comparing references to the same object instance or primitives (e.g. strings - as is the case here).

[`toEqual()`][3] can be used for testing loose equality (or with a custom comparator), but it is not necessary here.

[1]: https://github.com/jasmine/jasmine/blob/9cb2f06aa665870dbd5df5fd58a35863738961b3/src/core/matchers/toMatch.js#L14-L28
[2]: https://github.com/jasmine/jasmine/blob/9cb2f06aa665870dbd5df5fd58a35863738961b3/src/core/matchers/toMatch.js#L14-L28
[3]: https://github.com/jasmine/jasmine/blob/9cb2f06aa665870dbd5df5fd58a35863738961b3/src/core/matchers/toEqual.js#L10-L28

src/ng/http.js
```diff
@@ -141,6 +141,9 @@ function defaultHttpResponseTransform(data, headers) {
         try {
           data = fromJson(tempData);
         } catch (e) {
+          if (!contentType || contentType && contentType.indexOf(APPLICATION_JSON)) {
```
gkalpak: The `contentType && ` part is unnecessary (since we are already on the right-hand side of the `||`, where `contentType` is truthy. So, the above coondition would become `!contentType || contentType.indexOf(APPLICATION_JSON)`.

While the expression works as intended, it is a little cryptic/hard to read. Since you are essentially trying to negate the already checked `contentType && (contentType.indexOf(APPLICATION_JSON) === 0)`, I suggest you refactor it to something like:

```diff
 if (tempData) {
   var contentType = headers('Content-Type');
-  if ((contentType && (contentType.indexOf(APPLICATION_JSON) === 0)) || isJsonLike(tempData)) {
+  var hasJsonContentType = contentType && (contentType.indexOf(APPLICATION_JSON) === 0);
+  if (hasJsonContentType || isJsonLike(tempData)) {
     try {
       data = fromJson(tempData);
     } catch (e) {
+      if (!hasJsonContentType) {
+        return data;
+      }
       throw $httpMinErr(...);
     }
```

test/ng/httpSpec.js
```diff
@@ -1380,8 +1380,49 @@ describe('$http', function() {
             expect(errCallback.calls.mostRecent().args[0]).toEqualMinErr('$http', 'baddata');
           });
 
-        });
+          it('should not throw an error if JSON is invalid but content-type is not application/json', function() {
+            var errCallback = jasmine.createSpy('error');
+            $httpBackend.expect('GET', '/url').respond('{abcd}', {'Content-Type': 'text/plain'});
+
+            $http.get('/url').then(callback).catch(errCallback);
```
gkalpak: You don't really need `errCallback` (or the associated assertion) here.

test/ng/httpSpec.js
```diff
@@ -1380,8 +1380,49 @@ describe('$http', function() {
...
+
+            expect(callback).toHaveBeenCalledOnce();
+            expect(errCallback).not.toHaveBeenCalled();
+          });
 
+          it('should not throw an error if JSON is invalid but content-type is not specified', function() {
+            var errCallback = jasmine.createSpy('error');
+            $httpBackend.expect('GET', '/url').respond('{abcd}');
+
+            $http.get('/url').then(callback).catch(errCallback);
```
gkalpak: You don't really need `errCallback` (or the associated assertion) here.

test/ng/httpSpec.js
```diff
@@ -1380,8 +1380,49 @@ describe('$http', function() {
...
+            $httpBackend.flush();
+
+            expect(callback).toHaveBeenCalledOnce();
+            expect(errCallback).not.toHaveBeenCalled();
+          });
+
+          it('should return response unprocessed if JSON is invalid but content-type is not application/json', function() {
```
gkalpak: The "return response unprocessed" tests could be merged with their "not throw an error"counterparts, but I am fine keeping them as is too.

gkalpak (commented):
A few more suggestions :grin:

katsos:
@gkalpak about failing test, do I have to merge master with my branch?

gkalpak:
The failing tests look like flakes. (In any case, they don't seem related to this PR, so don't worry about them for now :smiley:)

katsos:
@gkalpak any progress with this one? thank you for your time!

gkalpak (approved):

PR #16086, opened 2017-07-06
Turkish Lira unicode character

Alaev:
I recommend adding the Turkish Lira Unicode character as it will be more informative than just TL and will also look great on recipes.

Adds Turkish lira Unicode character instead of TL

Shows TL instead of the Lira symbol

This feature will change the current TL to ₺

NO


Alaev:
I signed it!

PR #16091, opened 2017-07-11
refactor($parse): do not pass scope,locals to interceptor fns

jbedard:
After @lgalfaso's [comment](https://github.com/angular/angular.js/pull/16015/files#r126528737) I thought it would be a good idea to remove these params from interceptors. All the internal uses of interceptors are for doing things like data conversion/normalizing, never reading state out of the scope/locals. I think this is the intended use, and makes them like filters (which receive the input value + args).

Note that scope/locals still get passed through to the expression being wrapped/intercepted, it's just the interceptor-fn itself that now only receives the value.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1940,12 +1940,12 @@ function $ParseProvider() {
 
       function regularInterceptedExpression(scope, locals, assign, inputs) {
         var value = useInputs && inputs ? inputs[0] : parsedExpression(scope, locals, assign, inputs);
-        return interceptorFn(value, scope, locals);
+        return interceptorFn(value);
       }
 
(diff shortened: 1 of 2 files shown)


gkalpak (approved):

Narretz:
@jbedard can you please add the info from the PR to the commit message? This will be helpful for the next person who works on this code. While it's a refactor because it doesn't touch a public API, by including tests we change the expectations for how the internals should work.

jbedard:
Updated the commit message.

jbedard:
I'd also vote to only put this into 1.7 since it does change an internal but publicly accessible API

Narretz:
1.7 is fine with me

jbedard:
Commit msg look ok before merging?

Narretz:
LGTM

PR #16093, opened 2017-07-11
chore: deploy docs|code .angularjs.org to Firebase via Travis

### LINKED ISSUE #9674: Can the Angular docs on the website be made to deault to the latest stable release?
They seem to point to the latest git revision by default which is probably not what most people are using.

Narretz:
- code.angularjs.org and docs.angularjs.org are two separate Firebase projects
- both are automatically deployed via Travis config
- Travis is split up into 2 build stages: first, all tests are run, and if they pass, the deploy
stage runs a single job with both deployments (actual deployment depends on the state of the commit)
- docs. is deployed directly to Firebase hosting
- code. is uploaded to Firebase Google Cloud Storage and uses Firebase hosting rewrites to acces the
files
- jenkins builds still push the code builds to the code.angularjs.org Github repository

Further notes:

- the caching for the docs app is currently the default (3600)
- currently the secure firebase key is derived from my personal email account login. Should we create a key from a more internal google account instead?


scripts/code.angularjs.org-firebase/functions/index.js
```diff
@@ -0,0 +1,67 @@
...
+    return response.status(error.code).send(message);
+  });
+
+  function downloadAndSend() {
+    return bucket.file(downloadPath).download({
+      destination: `/tmp/${fileName}`
+    }).then(() => {
+      return response.status(200)
+        .set({
+          'Cache-Control': 'public, max-age=300, s-maxage=600'
```
Narretz: We can tweak the cache values later: s-maxage is the value for the Firebase CDN, max-age for the browser

scripts/code.angularjs.org-firebase/readme.firebase.code.md
```diff
@@ -0,0 +1,10 @@
+Firebase for code.angularjs.org
+===============================
+
+This folder contains the Google Firebase scripts for the code.angularjs.org setup.
+
+firebase.json contains the rewrite rules that route every subdirectory request to the cloud function
+in functions/index.js that serves the docs from the Firebase Google Cloud Storage bucket.
+
+The deployment to the Google Cloud Storage bucket happens automatically via Travis. See the travis.yml
+file in the repository root.
```
petebacondarwin: Could be worth a link to the other doc here, just in case people come across this file when they are looking for the other firebase configuration.

petebacondarwin (approved):
This is looking great @Narretz.
I have a left a few minor comments.

@IgorMinar - shall we generate secrets from a central account?

scripts/code.angularjs.org-firebase/functions/index.js
```diff
@@ -0,0 +1,67 @@
...
+const gcs = require('@google-cloud/storage')();
+const path = require('path');
+
+const gcsBucket = 'code-angularjs-org-338b8.appspot.com';
+const LOCAL_TMP_FOLDER = '/tmp/';
+
+function sendStoredFile(request, response) {
+  let filePathSegments = request.path.split('/').filter((segment) => {
+    return segment !== '';
+  });
```
gkalpak: Would `path.normalize()` also work?

scripts/code.angularjs.org-firebase/functions/index.js
```diff
@@ -0,0 +1,67 @@
...
+      downloadPath = path.join.apply(null, filePathSegments);
+      return downloadAndSend();
+    }
+
+    return Promise.reject(error);
+  }).catch(error => {
+    let message = 'General error';
+    if (error.code === 404) {
+      if (fileName.split('.').length === 1) {
+        message = 'Directory listing is not supported';
```
gkalpak: Could we host directly from Google storage (something like [this][1])?

[1]: https://cloud.google.com/storage/docs/hosting-static-website

scripts/code.angularjs.org-firebase/firebase.json
```diff
@@ -0,0 +1,21 @@
+{
+  "hosting": {
+    "public": "public",
+    "redirects": [
+      {
+        "source": "/:version/docs",
+        "destination": "/:version/docs/index.html",
+        "type": 301
```
Narretz: For some reason, redirect to `/:version/docs/` results in an infinite redirect loop in the browser 🤔

scripts/code.angularjs.org-firebase/functions/index.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+const functions = require('firebase-functions');
+const gcs = require('@google-cloud/storage')();
+const path = require('path');
+
+const gcsBucket = 'code-angularjs-org-338b8.appspot.com';
```
Narretz: There's an environment variable that we can use: process.env.GCLOUD_PROJECT

Narretz (commented):
I've addressed all your comments - only the redirect in code seems unfixable and directory listing / gcs static hosting can be handled later

petebacondarwin:
> @IgorMinar - shall we generate secrets from a central account?

He suggests that we go with the secrets that we have for now as we don't have a general solution to this yet.

petebacondarwin (approved):

PR #16096, opened 2017-07-11
feat($http): Added the ability to save a response from POST request with x-www-form-urlencoded 

XFree:
feature

if request method is POST and it contains Content-Type header "application / x-www-form-urlencoded" then this request will not be cached 

Added the ability to save a response from the server in the cache if the request was sent by the POST method with the Content-Type header "application / x-www-form-urlencoded"

no

**Other information**:

### DIFF
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1256,8 +1256,15 @@ function $HttpProvider() {
           cache,
           cachedResp,
           reqHeaders = config.headers,
-          isJsonp = lowercase(config.method) === 'jsonp',
-          url = config.url;
+          contentType = reqHeaders['Content-Type'],
+          lowercaseMethod = lowercase(config.method),
+          isJsonp = lowercaseMethod === 'jsonp',
+          isGet =  lowercaseMethod === 'get',
+          isPost = lowercaseMethod === 'post',
+          isPostUrlencoded = isPost && contentType && /application\/x-www-form-urlencoded/.test(contentType),
+          url = config.url,
+          // Save the source URL for using in the cache
+          cachedUrl = url;
 
       if (isJsonp) {
         // JSONP is a pretty sensitive operation where we're allowing a script to have full access to
@@ -1279,15 +1286,19 @@ function $HttpProvider() {
       promise.then(removePendingReq, removePendingReq);
 
       if ((config.cache || defaults.cache) && config.cache !== false &&
-          (config.method === 'GET' || config.method === 'JSONP')) {
+          (isGet || isJsonp || isPostUrlencoded)) {
         cache = isObject(config.cache) ? config.cache
             : isObject(/** @type {?} */ (defaults).cache)
               ? /** @type {?} */ (defaults).cache
               : defaultCache;
+        if (isPostUrlencoded) {
+          // Если это POST запрос, то к url прибавляем reqData
+          cachedUrl += '&' + reqData;
+        }
       }
 
       if (cache) {
-        cachedResp = cache.get(url);
+        cachedResp = cache.get(cachedUrl);
         if (isDefined(cachedResp)) {
           if (isPromiseLike(cachedResp)) {
             // cached request has already been sent, but there is no response yet
@@ -1302,7 +1313,7 @@ function $HttpProvider() {
           }
         } else {
           // put the promise for the non-transformed response into cache as a placeholder
-          cache.put(url, promise);
+          cache.put(cachedUrl, promise);
         }
       }
 
@@ -1357,10 +1368,10 @@ function $HttpProvider() {
       function done(status, response, headersString, statusText, xhrStatus) {
         if (cache) {
           if (isSuccess(status)) {
-            cache.put(url, [status, response, parseHeaders(headersString), statusText, xhrStatus]);
+            cache.put(cachedUrl, [status, response, parseHeaders(headersString), statusText, xhrStatus]);
           } else {
             // remove promise from the cache
-            cache.remove(url);
+            cache.remove(cachedUrl);
           }
         }
 
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1292,7 +1292,6 @@ function $HttpProvider() {
               ? /** @type {?} */ (defaults).cache
               : defaultCache;
         if (isPostUrlencoded) {
-          // Если это POST запрос, то к url прибавляем reqData
           cachedUrl = buildUrl(cachedUrl, reqData);
         }
       }
--- a/src/ng/http.js
+++ b/src/ng/http.js
@@ -1293,7 +1293,7 @@ function $HttpProvider() {
               : defaultCache;
         if (isPostUrlencoded) {
           // Если это POST запрос, то к url прибавляем reqData
-          cachedUrl += '&' + reqData;
+          cachedUrl = buildUrl(cachedUrl, reqData);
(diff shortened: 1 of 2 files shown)


Narretz:
Cache for POST requests has been requested before (https://github.com/angular/angular.js/issues/3031) but we've been hesitant because POST is usually not cachable in REST APIs. Your PR is even more specific, making only form-urlencoded POST cachable. Why?

XFree:
In older systems, a POST request with the header "application / x-www-form-urlencoded" is often used instead of the GET request. For a transparent transition from the old version of the API to the standard REST, you need to have a similar mechanism. If you make the decorator $ httpBackend or $ http, then you have to duplicate a lot of logic. This approach will replace the old resource with a new one without overwriting anything. This change will not affect the old functionality, because Must be explicitly included through the cache flag.

Narretz:
I see.
I still don't think that it makes a lot of sense to cache such a specific combination. Next thing you know someone wants another combination cached.

Instead of duplicating http shouldn't a light wrapper around it be enough? The wrapper would put the response in a custom cache and return it when appropriate.

XFree:
The wrapper significantly complicates the logic or does not take into account some cases. Perhaps it is worth the revising of the decoration mechanism when working with the cache. In the current implementation, checking for anything in the cache is only done if GET or JSONP methods are used.
In the document https://docs.angularjs.org/api/ng/service/$httpParamSerializerJQLike combinations are given as equivalent if I do not confuse.

src/ng/http.js
```diff
@@ -1279,15 +1286,18 @@ function $HttpProvider() {
...
 
       if ((config.cache || defaults.cache) && config.cache !== false &&
-          (config.method === 'GET' || config.method === 'JSONP')) {
+          (isGet || isJsonp || isPostUrlencoded)) {
         cache = isObject(config.cache) ? config.cache
             : isObject(/** @type {?} */ (defaults).cache)
               ? /** @type {?} */ (defaults).cache
               : defaultCache;
+        if (isPostUrlencoded) {
+          cachedUrl = buildUrl(cachedUrl, reqData);
```
gkalpak: FWIW, `reqData` is not necessarily a string.

gkalpak:
TBH, I still don't feel this is worth it. We are adding a workaround for some legacy system that uses POST requests incorrectly. Meanwhile, we are complicating the implementation, introducing more API surface (which we have to maintain) and more corner cases and make the API less straight forward for most users (which don't want this functionality).

The alternative is that we keep the current implementation (which suits most usecases) and people using legacy systems can't get their requests cached. It sounds like a minimal trade-off.

And if you are too concerned about it, you can create your own wrapper (as suggested above). It is just a few lines of code:

- Build the `cachedUrl` (as you do in this PR) and check it against your cache.
- If it matches, return the cached response.
- If it doesn't pass it to the original service.
- Cache the response when it comes back and return it to the user.

Narretz:
I know that a wrapper is not ideal, but as far as I am concerned, this PR is not mergable as it introduces another special case that will be problematic further down the road with regard to user expectations. It's also possible that someone has cache set to true for all requests, based in the assumption that POST requests aren't cached, and with this PR they would be if the header is right. Unlikely, but still.

What I could possibly see as a solution to this is to make the cache negotiation generally configurable similar to the serialization. The service would a) decide if a request can be cached and b) create the identifier for the cache entry (to fix https://github.com/angular/angular.js/issues/5028)

@gkalpak would you support such a solution? This would be implemented by community contributors

<sub>
> some legacy system that uses POST requests incorrectly

There are modern system that use POST requests with RPC apis. I don't think POST was created purely for use with REST apis.
</sub>

src/ng/http.js
XFree: In general, this is a string. Https://github.com/angular/angular.js/blob/master/src/ng/http.js#L1061
It's important for us to get a unique key. I can add a conversion to json, if for some reason there will be an an object

XFree:
@Narretz 
> There are modern system that use POST requests with RPC apis. I don't think POST was created purely for use with REST apis.
I described the situation I met. Namely, the transition from non-REST to REST. Of course, POST can be interpreted differently.
@gkalpak, The final plans of the service $ http, such a wrapper will be to be duplicated in each resource, as well as to duplicate some of the logic from the service itself. I am ready to change PR any way that it would be possible to make a normal decorator to solve any problems with the cache.

gkalpak:
> @gkalpak, The final plans of the service $ http, such a wrapper will be to be duplicated in each resource, as well as to duplicate some of the logic from the service itself. I am ready to change PR any way that it would be possible to make a normal decorator to solve any problems with the cache.

@XFree, I am not sure what you mean. What I propose with a wrapper service would solve any problems with cache in a few lines (via a decorator).

##
> There are modern system that use POST requests with RPC apis. I don't think POST was created purely for use with REST apis

@Narretz, I am not talking about REST APIs. But according to the [HTTP spec] some methods are intended to be used for side effects and some not:

> Implementors should be aware that the software represents the user in
   their interactions over the Internet, and should be careful to allow
   the user to be aware of any actions they might take which may have an
   unexpected significance to themselves or others.
>
> In particular, the convention has been established that the GET and
   HEAD methods SHOULD NOT have the significance of taking an action
   other than retrieval. These methods ought to be considered "safe".
   This allows user agents to represent other methods, such as POST, PUT
   and DELETE, in a special way, so that the user is made aware of the
   fact that a possibly unsafe action is being requested.
>
> Naturally, it is not possible to ensure that the server does not
   generate side-effects as a result of performing a GET request; in
   fact, some dynamic resources consider that a feature. The important
   distinction here is that the user did not request the side-effects,
   so therefore cannot be held accountable for them.

This doesn't mean that every server will behave the same way, but that is a general norm/guideline we should not go against by default imo. (And I understand we agree on that.)

##
> What I could possibly see as a solution to this is to make the cache negotiation generally configurable similar to the serialization.

I wouldn't necessarily be opposed to that, but I am not a big fan either :smiley:
It will be adding to the complexity and API surface and will only have minimal benefit (i.e. being able to easily cache POST requests) to few people (i.e. those with systems that use POST in an controversial way :grin:).

But if it can be implemented cleanly and without much overhead, I would definitely consider it.

[1]: https://tools.ietf.org/html/rfc2616#section-9.1.1

XFree:
@gkalpak, Can you give an example?
I could not make such a decorator, which would have turned out less when, than now, as well as all the options and settings were taken into account.

gkalpak:
@XFree, it all depends on the constraints you have and what you are willing to give up. Usually, in such cases you don't need to support all corner cases and features, so you can create a stripped down version that covers your needs.

A very simple version could look something like [this][1].

[1]: https://codepen.io/ExpertSystem/pen/MoZBpJ?editors=0012

XFree:
@gkalpak, That's a lot of code. I'll try to think over the mechanism that allows to decorate $ http properly to provide caching. Because There are POST search queries that do not change data and require caching.

gkalpak:
<sub>FWIW, the decorator is only 30 lines of code (including whitespace, closing brackets and everything).</sub>

Narretz:
I'm going to close this PR as we have decided that it's not going to be added in this form. @XFree if you want to work on a feature that makes cache negotiation configurable, you can open a new PR.

XFree:
@Narretz , ок.

PR #16098, opened 2017-07-11
fix() apostrophe in url

dmartres:
bug fix on $location service in case that url contains apostrophes

### DIFF
--- a/src/ng/browser.js
+++ b/src/ng/browser.js
@@ -132,7 +132,6 @@ function Browser(window, document, $log, $sniffer) {
     // setter
     if (url) {
       var sameState = lastHistoryState === state;
-
       // Don't change anything if previous and current URLs and states match. This also prevents
       // IE<10 from getting into redirect loop when in LocationHashbangInHtml5Url mode.
       // See https://github.com/angular/angular.js/commit/ffb2701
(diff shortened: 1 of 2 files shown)


gkalpak:
Thx for the PR, @dmartres!

From a quick look, I don't think the issue is specific to apostrophe. The main problem seems to be that we are comparing `self.url()` to `lastBrowserUrl`, but `lastBrowserUrl` can be set from different sources, not always with the same encoding/decoding. I think it boils down to two main cases:

- In some places, we set `lastBrowserUrl` to `location.href` (which is undecoded).
- In other places we set `lastBrowserUrl` to a value indirectly based on `$location.$$url`, which is passed through `decodeURIComponent`.

IMO, we should address the issue at its core and always use the same encoding/decoding when setting `lastBrowserUrl` (even if that will be a larger undertaking).

@dmartres, would you like to take a stab at it?

dmartres:
You're Right, but now I can't work on that :( . Should I report this as an issue?
Perhaps someone can take a look based on this discussion.

gkalpak:
Yes, please :smiley:

gkalpak:
Thx for opening the issue. Closing this.

PR #16102, opened 2017-07-12
feat($sanitize): Added support for the sftp protocol in $$ SanitizeUriProvider and linky filter

XFree:
… SanitizeUriProvider

Added support for the sftp protocol in $$ SanitizeUriProvider and linky filter

feature

sftp not suported

sftp suported

no

**Other information**:

### DIFF
--- a/src/ng/sanitizeUri.js
+++ b/src/ng/sanitizeUri.js
@@ -6,8 +6,8 @@
  * Private service to sanitize uris for links and images. Used by $compile and $sanitize.
  */
 function $$SanitizeUriProvider() {
-  var aHrefSanitizationWhitelist = /^\s*(https?|ftp|mailto|tel|file):/,
-    imgSrcSanitizationWhitelist = /^\s*((https?|ftp|file|blob):|data:image\/)/;
+  var aHrefSanitizationWhitelist = /^\s*(https?|s?ftp|mailto|tel|file):/,
+    imgSrcSanitizationWhitelist = /^\s*((https?|s?ftp|file|blob):|data:image\/)/;
 
   /**
    * @description
--- a/src/ng/sanitizeUri.js
+++ b/src/ng/sanitizeUri.js
@@ -7,7 +7,7 @@
  */
 function $$SanitizeUriProvider() {
   var aHrefSanitizationWhitelist = /^\s*(https?|s?ftp|mailto|tel|file):/,
-    imgSrcSanitizationWhitelist = /^\s*((https?|s?ftp|file|blob):|data:image\/)/;
+    imgSrcSanitizationWhitelist = /^\s*((https?|ftp|file|blob):|data:image\/)/;
 
   /**
    * @description
--- a/src/ngSanitize/filter/linky.js
+++ b/src/ngSanitize/filter/linky.js
@@ -6,7 +6,7 @@
  * @kind function
  *
  * @description
- * Finds links in text input and turns them into html links. Supports `http/https/ftp/mailto` and
+ * Finds links in text input and turns them into html links. Supports `http/https/ftp/sftp/mailto` and
  * plain email address links.
  *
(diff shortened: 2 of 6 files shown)


XFree:
I do not understand why the build is falling

Narretz:
Hi, the build error is unrelated to your PR. We are looking into it.

XFree:
@Narretz, Tell me, where are the rules for adding to the documentation about the fixes.

Narretz:
@XFree the docs are in the code files themselves. However, it looks like the default whitelist is not documented. The docs for linky are here: https://github.com/angular/angular.js/blob/master/src/ngSanitize/filter/linky.js

Btw, the scope for this should be `feat($sanitize) ...` as `$$SanitizeUriProvider` is a private service.

Narretz:
Could you please add `(s)ftp` tests to linky and info the linky docs that sftp is supported?

XFree:
Is the documentation generated automatically?
I did not really understand how to properly arrange the documentation.

Narretz:
@XFree the docs are created from the comments around the functions, e.g. https://github.com/angular/angular.js/blob/ecc09a45f02454e4e2478ad22cd0d179dccc7916/src/ngSanitize/filter/linky.js#L8-L10

XFree:
@Narretz , Oh. Thank you.

gkalpak (approved):

Narretz:
@mprobst / @rjamet could one of you have a quick look at this security related PR? It's very straightforward, I just wanna follow protocol. :)

rjamet:
Is there any browser with built-in support of sftp?... 

That looks safe, and so I'm okay with merging, but I think that might be better suited as an additional config knob that gets set in the application itself. 

SFTP matches the idea of fetching stuff at the end of a link, so I think it's reasonable. But if we allow sftp, then you could push that logic to ssh://, magnet://, onion:// and a more which would be surprising for devs who didn't explicitly allow those. In all of those cases, I'd rather see something like linkyProvider.allowProtocols("sftp://") / $compileProvider.allowProtocols("sftp://") in the config phase of the application: this way, you don't change the default, and if you need to use these protocols, then two lines are enough to enable them app-wide.

XFree:
@rjamet , In this case, when this need to do this for https, but this is beyond the scope of this PR.

src/ng/sanitizeUri.js
```diff
@@ -6,8 +6,8 @@
  * Private service to sanitize uris for links and images. Used by $compile and $sanitize.
  */
 function $$SanitizeUriProvider() {
-  var aHrefSanitizationWhitelist = /^\s*(https?|ftp|mailto|tel|file):/,
-    imgSrcSanitizationWhitelist = /^\s*((https?|ftp|file|blob):|data:image\/)/;
+  var aHrefSanitizationWhitelist = /^\s*(https?|s?ftp|mailto|tel|file):/,
+    imgSrcSanitizationWhitelist = /^\s*((https?|s?ftp|file|blob):|data:image\/)/;
```
rjamet: Oops, I missed this, sorry: do you need to update imgSrc too ? I can see how navigational links could work, but if browsers don't support the protocol, images and such wouldn't work at all (Chrome seems to complain with `Failed to load resource: net::ERR_UNKNOWN_URL_SCHEME` at least).

XFree: I'm fix it.

rjamet:
LGTM security-wise, I'll let the Angular team decide for the rest.

XFree:
I do not understand what the problems are with CI?

gkalpak:
`commitplease` was unable to validate the commit messages (didn't really looked into why). Given that this is the only test that failed, I think it is fine to ignore the CI failure.

XFree:
When will Pull request be merged? =)

petebacondarwin (approved):

PR #16109, opened 2017-07-13
feature($resource): Add Resource To Response For Error Interceptors

kmccullough:
Similar to how the resource object (which is to be populated later) is added to the response for a successful request and accessible to the response interceptor, this change attaches the resource object to the response for a failed request and is accessible by the responseError interceptor.

The resource object is not accessible from the responseError interceptor.

The resource object is accessible from the responseError interceptor.

No.

**Other information**:


kmccullough:
test this please

gkalpak (changes requested):
Could you please add some docs for that. I think the behavior of adding `resource` to the response is not documented (and it should be).
<sub>(If you don't feel like adding the docs, it's OK :smiley:)</sub>

LGTM otherwise 👍

kmccullough:
@gkalpak How's that?

gkalpak (approved):

Issue #16106, opened 2017-07-13
Outstanding Firebase migration tasks


Narretz:
- [x] fix the examples in the docs deployment (angular code files are missing).
  Pass the deployment mode of the app from the index-*.html file into the docs AngularJS app and change https://github.com/angular/angular.js/blob/828a2757b2e04a3bec12012c127812559fab78e3/docs/config/templates/examples/runnableExample.template.html so that instead of hardcoding deployments.default it sets the iframe src on runtime to the deployment in which the docs app currently is
- [x] Delete zipped angular build file before uploading new snapshot version
  - [ ] Check again, looks like the zips are not deleted
- [x] Don't run the build.sh in the deploy stage at all when travis is running from a PR
  Use env variable TRAVIS_PULL_REQUEST
  - [ ] use new "if" config for Travis stages
- [x] List directory contents in code.angularjs
- [x] deploy docs and code when the branch distTag is "latest" (not only on tag)
- [ ] migrate angularjs.org
- [ ] Switch domains

Narretz:
The rest of the tags are out of my hands.

PR #16110, opened 2017-07-14
make 1.6.x build steps like master

Narretz:

**Other information**:


.travis.yml
```diff
@@ -37,21 +37,63 @@ addons:
...
+    on_start: always # default: false
+
+jobs:
+  include:
+    - stage: deploy
+      env:
+        - JOB=deploy
+      before_script: skip
+      script:
+        - "./scripts/travis/build.sh"
```
gkalpak: Not directly related to this PR, but I noticed that the build script is run even on PRs (which will never get deployed).
Could we skip the extra work on PRs?

gkalpak (approved):
LGTM
<sub>(My one comment is not really related to this PR.)</sub>

.travis.yml
Narretz: Maybe we can skip the deploy stage on PRs via some config option
Or we can use the TRAVIS_PULL_REQUEST env variable to skip the build script. 
I've added it as a task here https://github.com/angular/angular.js/issues/16106

gkalpak: (Quite surprisingly, I didn't see a config option to skip that for PRs - or maybe I missed that. But yeah, `TRAVIS_PULL_REQUEST` would do the job.)

PR #16116, opened 2017-07-17
fix($parse): do not shallow-watch computed property keys

jbedard:
Fixes a case which b5118ac6a9e0a327b31094b3fdcdc0432b23ad2f missed. ES6 computed keys can implicitly invoke `.toString()` of an object so we can't shallow-watch it.

The other option is shallow-watching `String(key)` instead, but I'm a little hesitant to watch something slightly different then how it's used in the real expression.

### DIFF
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -3805,35 +3805,55 @@ describe('parser', function() {
 
         it('should watch ES6 object computed property changes', function() {
           var count = 0;
-          var values = [];
+          var lastValue;
 
           scope.$watch('{[a]: true}', function(val) {
             count++;
-            values.push(val);
-          }, true);
+            lastValue = val;
+          });
 
           scope.$digest();
           expect(count).toBe(1);
-          expect(values[0]).toEqual({'undefined': true});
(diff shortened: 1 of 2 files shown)


test/ng/parseSpec.js
```diff
@@ -3805,35 +3805,55 @@ describe('parser', function() {
...
+          scope.$watch('{[a]: true}', function(val) {
+            count++;
+            lastValue = val;
+          });
+
+          scope.a = {toString: function() { return this.b; }};
+          scope.a.b = 1;
+
+          //TODO: would be great if it didn't throw!
+          expect(function() { scope.$apply(); }).toThrowMinErr('$rootScope', 'infdig');
```
petebacondarwin: Hmm, doesn't this void the whole fix? Or did it throw before anyway?

petebacondarwin: OK, so let's have another PR that converts computed keys to strings inside the parser. Then we can watch the string.

jbedard: As discussed: this brings it back to how it was in 1.6.3 (or .4?), which for now is better then incorrectly shallow-watching the object which might contain state (like this test).

Issue #16078, opened 2017-06-30
Improve jqLiteRemoveClass performance


aresn:
I would like to propose a slight modification to jqLiteRemoveClass function to increase it's performance. 

Current implementation constantly reads the "class" value from DOM and writes it back to the DOM inside a for loop.  

Here is what I have created for our local branch of angular:

    function jqLiteRemoveClass(element, cssClasses) {
      if (cssClasses && element.setAttribute) {
        var currentClass = element.getAttribute('class');

        if (currentClass) {
          var cssClass;
          var curValue;
          curValue = currentClass = trim(currentClass);
          var classNames = cssClasses.split(' ');

          for (var i = 0, len = classNames.length; i < len; i++) {
            cssClass = trim(classNames[i]);
            currentClass = trim((' ' + currentClass + ' ')
              .replace(/[\n\t]/g, ' ')
              .replace(' ' + cssClass + ' ', ' '));
          }
          if (currentClass !== curValue) {
            element.setAttribute('class',currentClass);
          }
        }
      }
    }

Here is a link to a jsPerf that I have created that compares the two functions : 
https://jsperf.com/compare-angular-remove-class-with-improved-version/1

Narretz:
Hi, perf improvements are definitely welcome, although I wonder if you specifically experienced perf  problems related to removeClass in your apps?
And it does make sense to open this as a PR, so it's easier to see the change in the code.

PR #16131, opened 2017-07-26
perf(jqLite): avoid repeated add/removeAttribute in jqLiteRemoveClass

jbedard:
This does something similar to what was suggested in #16078, although keeps it a little more like `jqLiteAddClass` for now.

### DIFF
--- a/src/jqLite.js
+++ b/src/jqLite.js
@@ -422,13 +422,16 @@ function jqLiteRemoveClass(element, cssClasses) {
   if (cssClasses && element.setAttribute) {
     var existingClasses = (' ' + (element.getAttribute('class') || '') + ' ')
                             .replace(/[\n\t]/g, ' ');
+    var newClasses = existingClasses;
 
     forEach(cssClasses.split(' '), function(cssClass) {
       cssClass = trim(cssClass);
-      existingClasses = existingClasses.replace(' ' + cssClass + ' ', ' ');
+      newClasses = newClasses.replace(' ' + cssClass + ' ', ' ');
     });
 
-    element.setAttribute('class', trim(existingClasses));
+    if (newClasses !== existingClasses) {
+      element.setAttribute('class', trim(newClasses));
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


gkalpak (approved):
LGTM (as soon as Travis is happy)

src/jqLite.js
```diff
@@ -420,13 +420,15 @@ function jqLiteHasClass(element, selector) {
...
-      element.setAttribute('class', trim(
-          (' ' + (element.getAttribute('class') || '') + ' ')
-          .replace(/[\n\t]/g, ' ')
-          .replace(' ' + trim(cssClass) + ' ', ' '))
-      );
+      cssClass = trim(cssClass);
+      existingClasses = existingClasses.replace(' ' + cssClass + ' ', ' ');
     });
+
+    element.setAttribute('class', trim(existingClasses));
```
dcherman: Should you do a check to ensure that we don't set this unnecessarily which potentially causes unnecessary renders?

Ref: https://github.com/jquery/jquery/blob/master/src/attributes/classes.js#L41-L45

jbedard: I thought that should be done along with addClass in its own commit...

jbedard:
Passed now. However I decided to add one more commit like @dcherman suggested to only actually set the `class` attribute if it changed (for both `addClass` and `removeClass`).

gkalpak (approved):
LGTM
Do you think it is worth adding some tests to verify the new behavior (i.e. that we don't call `get/setAttribute()` more times than necessary)?

Narretz:
@jbedard can you please rebase this? The tests currently don't run because of a Travis problem at the branch point.

jbedard:
Rebased.

The tests had to be put in `if (_jqLiteMode)` since they are testing a JQLite implementation detail (I think jquery3 tests passed, but not jquery2).

mgol (approved):

test/jqLiteSpec.js
```diff
@@ -914,6 +914,37 @@ describe('jqLite', function() {
       });
 
 
+      //JQLite specific implementation/performance tests
```
mgol: A space before `JQLite`?

test/jqLiteSpec.js
```diff
@@ -1031,6 +1062,37 @@ describe('jqLite', function() {
         jqA.removeClass('foo baz noexistent');
         expect(a.className).toBe('bar');
       });
+
+
+      //JQLite specific implementation/performance tests
```
mgol: A space before `JQLite`?

test/jqLiteSpec.js
jbedard: NOTE: add space

PR #16141, opened 2017-08-02
test(*): fix tests involving `submit` on Chrome 60

gkalpak:
Test fix.

Some tests involving `submit` fail on Windows with Chrome 60.

All tests pass.

No

**Other information**:
On Chrome 60 (at least on Windows) the `submit` event when clicking on a submit button is not fired on the form element, unless it is already part of the DOM.

### DIFF
--- a/test/ng/directive/formSpec.js
+++ b/test/ng/directive/formSpec.js
@@ -386,6 +386,9 @@ describe('form', function() {
       doc = jqLite('<form ng-submit="submitMe()">' +
                      '<input type="submit" value="submit">' +
                    '</form>');
+      // Support: Chrome 60+ (on Windows)
+      // We need to add the form to the DOM in order for `submit` events to be properly fired.
+      window.document.body.appendChild(doc[0]);
 
       var assertPreventDefaultListener = function(e) {
         reloadPrevented = e.defaultPrevented || (e.returnValue === false);
@@ -420,15 +423,18 @@ describe('form', function() {
       inject(function($timeout) {
         doc = jqLite('<div>' +
                         '<form ng-submit="submitMe()">' +
-                          '<button ng-click="destroy()"></button>' +
+                          '<button type="submit" ng-click="destroy()"></button>' +
                         '</form>' +
                       '</div>');
+        // Support: Chrome 60+ (on Windows)
+        // We need to add the form to the DOM in order for `submit` events to be properly fired.
+        window.document.body.appendChild(doc[0]);
 
         var form = doc.find('form'),
             destroyed = false,
             nextTurn = false,
             submitted = false,
-            reloadPrevented;
+            reloadPrevented = 'never called';
 
         scope.destroy = function() {
           // yes, I know, scope methods should not do direct DOM manipulation, but I wanted to keep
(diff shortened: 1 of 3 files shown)


test/ng/directive/formSpec.js
```diff
@@ -420,15 +423,18 @@ describe('form', function() {
       inject(function($timeout) {
         doc = jqLite('<div>' +
                         '<form ng-submit="submitMe()">' +
-                          '<button ng-click="destroy()"></button>' +
+                          '<button type="submit" ng-click="destroy()"></button>' +
```
Narretz: Is this necessary? I think if there's no other button, then the first button is the submit button

test/ng/directive/formSpec.js
```diff
@@ -386,6 +386,9 @@ describe('form', function() {
       doc = jqLite('<form ng-submit="submitMe()">' +
                      '<input type="submit" value="submit">' +
                    '</form>');
+      // Support: Chrome 60+ (on Windows)
+      // We need to add the form to the DOM in order for `submit` events to be properly fired.
+      window.document.body.appendChild(doc[0]);
```
Narretz: Aren't all of these memory leaks? Or is there some global cleanup between tests?

test/ng/directive/formSpec.js
```diff
@@ -460,6 +466,7 @@ describe('form', function() {
         .runs(function() {
           expect(doc.html()).toBe('');
           expect(destroyed).toBe(true);
+          expect(reloadPrevented).toBe('never called');
```
Narretz: Is this unrelated to the other changes?

test/ng/directive/formSpec.js
gkalpak: We [empty body][1] before each test.

[1]: https://github.com/angular/angular.js/blob/3650723501ca5f50003192bc4f94c77cef711e1a/test/helpers/testabilityPatch.js#L23

test/ng/directive/formSpec.js
gkalpak: No, it is not necessary. Just making it more explicit for people that might not know.

test/ng/directive/formSpec.js
gkalpak: Yes :grin:

Narretz: The test results disagree with this expectation 
👀
⭕️ 

And based on the description I also would expect that this is true.

gkalpak: They agree locally (probably a windows thing) 😕

gkalpak: So, it turns out the cause is again that Chrome 60+ on Windows does not fire `submit` events if the form is not attached to the DOM. Changed the expectation to verify that the `submit` event is eithr never fired or (if fired) the reload is prevented.

Narretz (approved):
LGTM. We should obviously try to test on Chrome 60 on Travis some time soon.

Narretz:
Looks like chrome 60 is not yet available

gkalpak:
Yeah. Trying 59 :smiley:

gkalpak:
Also updated to latest version for Chrome (59) and Firefox (54) available on Saucelabs.
Merging...

PR #16149, opened 2017-08-06
test(ngOptions): fix flaky test on Firefox 54+ and Safari 9

gkalpak:
Reduce (remove?) test flaky-ness.

A particular `ngOptions` test is very flaky on Firefox 54+ (possibly due to a (buggy?) optimization).

The test is more reliable on all browsers.

No.

### DIFF
--- a/test/ng/directive/ngOptionsSpec.js
+++ b/test/ng/directive/ngOptionsSpec.js
@@ -2922,84 +2922,137 @@ describe('ngOptions', function() {
     });
 
 
-    it('should not re-set the `selected` property if it already has the correct value', function() {
-      scope.values = [{name: 'A'}, {name: 'B'}];
-      createMultiSelect();
+    // Support: Safari 9
+    // This test relies defining a getter/setter `selected` property on either `<option>` elements
+    // or their prototype. Some browsers (including Safari 9) are very flakey when the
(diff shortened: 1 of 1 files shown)


gkalpak:
In case anyone is wondering, this is what caused the issue (afaict):

- On Firefox and Chrome (didn't check other browsers), the `selected` property is implemented as a getter/setter property on the `HTMLOptionElement` prototype (not the `<option>` element itself).
- We were defining a getter/setter `selected` property on the `<option>` elements and spying on it.
- It seems that in certain cases (possibly when a specific optimization was triggered), Firefox would ignore our getter/setter spy and use the getter/setter on the prototype (potentially only for `get` operations).

Narretz:
As discussed, I think we should merge this and possibly skip the test altogether on Safari where it also sometimes sporadically fails.

Narretz (approved):

Narretz:
Let's merge it as is, and see if there are Safari failures later

gkalpak:
I disabled the test on Safari 9. Once Travis is green, I'll merge it.

Narretz:
There still time to revert!

gkalpak:
Too late, I squashed the commits 😛

Issue #15855, opened 2017-03-23
httpBackend's verify methods don't catch unexpected http calls


cmplank:
**I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] other (Please do not submit support requests here (see above))

**Current behavior:**
Starting in angular 1.6.0, when a unit test exercises an http call but doesn't specify an httpBackend expectation, neither the `httpBackend.verifyNoOutstandingExpectation()` nor the `httpBackend.verifyNoOutstandingRequest()` method throw an error. This problem still manifests in the latest snapshot.

**Expected / new behavior:**
Previously (on angular 1.5.11) each of the verify methods threw errors with the message "Unexpected request: GET <url> No more request expected" when an http method was called without an accompanying expectation.

**Minimal reproduction of the problem with instructions:**
Write a unit test that exercises a `$http` call but don't tell httpBackend to expect it. Call `httpBackend.verifyNoOutstandingExpectation()` and `httpBackend.verifyNoOutstandingRequest()` and neither of them will throw an error and your test will pass.

Plunk on angular snapshot: http://plnkr.co/edit/DZZ2Nl?p=preview
Change the angular version to 1.5.11 and the test will pass.

**Angular version:** 1.6.3
This is an issue in 1.6.0 through the most recent snapshot of angular. An interesting note is that it is not the version of angular-mocks that makes the difference as I would have expected, but rather the version of core angular.

**Browser:** [all]
Happens in node/karma/jasmine tests and in Chrome jasmine runner (see plunk)

**Anything else:**

Narretz:
It looks like this is related to the error handler in your `then` function. If you remove the empty error handler, verify throws. Maybe because of https://github.com/angular/angular.js/commit/c9dffde1cb167660120753181cb6d01dc1d1b3d0

Narretz:
@gkalpak could this be related to the changes in the $q error handling?

gkalpak:
Yes. When there is an unexpected request, `$httpBackend` throws, which causes the Promise returned by `$http.get()` to be rejected. Before 1.6, a thrown error was also passed to the `$exceptionHandler`, which rethrows by default in tests (hence you get an exception thrown). In 1.6. (due to e13eeabd7), throws errors are treated the same as regular rejection, i.e. just reject the promise and don't get passed to the `$exceptionHandler`. 

[...]

PR #16150, opened 2017-08-06
ugly solution for unexpected request fail on verify no outstanding request

marcin-wosinek:
Bug fix

https://github.com/angular/angular.js/issues/15855

same as before the regression

no

**Other information**:

### DIFF
--- a/src/ng/q.js
+++ b/src/ng/q.js
@@ -343,6 +343,17 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
     }
   });
 
+  function PassToExceptionHandlerError(e) {
+    this.message = e.message;
+    this.name = e.name;
+    if (e.line) this.line = e.line;
+    if (e.sourceId) this.sourceId = e.sourceId;
+    if (e.stack) this.stack = e.stack;
+    if (e.stackArray) this.stackArray = e.stackArray;
+  }
+
+  PassToExceptionHandlerError.prototype = Object.create(Error.prototype);
+
   function processQueue(state) {
     var fn, promise, pending;
 
@@ -364,6 +375,10 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
           }
         } catch (e) {
           rejectPromise(promise, e);
+          // This error is explicitly marked for being passed to the $exceptionHandler
+          if (e && e instanceof PassToExceptionHandlerError) {
+            exceptionHandler(e);
+          }
         }
       }
     } finally {
@@ -668,6 +683,7 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
   $Q.resolve = resolve;
   $Q.all = all;
   $Q.race = race;
+  $Q.$$PassToExceptionHandlerError = PassToExceptionHandlerError;
 
   return $Q;
 }
--- a/src/ngMock/angular-mocks.js
+++ b/src/ngMock/angular-mocks.js
@@ -1323,7 +1323,7 @@ angular.mock.dump = function(object) {
   ```
  */
 angular.mock.$httpBackendDecorator =
-  ['$rootScope', '$timeout', '$delegate', createHttpBackendMock];
+  ['$q', '$rootScope', '$timeout', '$delegate', createHttpBackendMock];
 
 /**
  * General factory function for $httpBackend mock.
@@ -1339,7 +1339,7 @@ angular.mock.$httpBackendDecorator =
  * @param {Object=} $browser Auto-flushing enabled if specified
  * @return {Object} Instance of $httpBackend mock
  */
-function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
+function createHttpBackendMock($q, $rootScope, $timeout, $delegate, $browser) {
   var definitions = [],
       expectations = [],
       responses = [],
@@ -1438,10 +1438,22 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
         return;
       }
     }
-    throw wasExpected ?
+
+    // In addition to be being converted to a rejection, this error also needs to be passed to
+    // the $exceptionHandler and be rethrown (so that the test fails).
+    var error = wasExpected ?
         new Error('No response defined !') :
         new Error('Unexpected request: ' + method + ' ' + url + '\n' +
                   (expectation ? 'Expected ' + expectation : 'No more request expected'));
+
+    // IE10+ and PhanthomJS do not set stack trace information, until the error is thrown
+    if (!error.stack) {
+      try {
+        throw error;
+      } catch (e) { /* empty */ }
+    }
+
+    throw new $q.$$PassToExceptionHandlerError(error);
   }
 
   /**
@@ -2706,7 +2718,7 @@ angular.module('ngMockE2E', ['ng']).config(['$provide', function($provide) {
  */
 angular.mock.e2e = {};
 angular.mock.e2e.$httpBackendDecorator =
-  ['$rootScope', '$timeout', '$delegate', '$browser', createHttpBackendMock];
+  ['$q', '$rootScope', '$timeout', '$delegate', '$browser', createHttpBackendMock];
 
 
 /**
--- a/test/ngMock/angular-mocksSpec.js
+++ b/test/ngMock/angular-mocksSpec.js
@@ -2429,7 +2429,7 @@ describe('ngMock', function() {
 
 describe('ngMockE2E', function() {
   describe('$httpBackend', function() {
-    var hb, realHttpBackend, realHttpBackendBrowser, callback;
+    var hb, realHttpBackend, realHttpBackendBrowser, $http, callback;
 
     beforeEach(function() {
       callback = jasmine.createSpy('callback');
@@ -2442,7 +2442,26 @@ describe('ngMockE2E', function() {
       module('ngMockE2E');
       inject(function($injector) {
         hb = $injector.get('$httpBackend');
+        $http = $injector.get('$http');
       });
+    });
+
+
+    it('should throw error when unexpected request - without error callback', function() {
+      expect(function() {
+        $http.get('/some').then(noop);
+
+        hb.verifyNoOutstandingRequest();
+      }).toThrowError('Unexpected request: GET /some\nNo more request expected');
+    });
+
+
+    it('should throw error when unexpected request - with error callback', function() {
+      expect(function() {
+        $http.get('/some').then(noop, noop);
+
+        hb.verifyNoOutstandingRequest();
+      }).toThrowError('Unexpected request: GET /some\nNo more request expected');
     });
 
 


Narretz:
Looks like this break the promises-aplus-tests.

src/ngMock/angular-mocks.js
```diff
@@ -1438,10 +1438,15 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
...
       }
     }
-    throw wasExpected ?
+    var error = wasExpected ?
         new Error('No response defined !') :
         new Error('Unexpected request: ' + method + ' ' + url + '\n' +
                   (expectation ? 'Expected ' + expectation : 'No more request expected'));
+
+    // TODO ugly!
+    error.rethrow = true;
```
gkalpak: Please rename to `$$rethrow` or `$$passToExceptionHandler`.

src/ngMock/angular-mocks.js
```diff
@@ -1438,10 +1438,15 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
         return;
       }
     }
-    throw wasExpected ?
+    var error = wasExpected ?
         new Error('No response defined !') :
         new Error('Unexpected request: ' + method + ' ' + url + '\n' +
                   (expectation ? 'Expected ' + expectation : 'No more request expected'));
+
+    // TODO ugly!
```
gkalpak: Remove the `ugly` comment 😛 
An actual comment would fine actually. Something like:

> In addition to be being converted to a rejection, this error also need to be passed to
> the `$exceptionHandler` and be rethrown (so that the test fails).

src/ng/q.js
```diff
@@ -364,6 +364,10 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
           }
         } catch (e) {
           rejectPromise(promise, e);
+          // TODO ugly!
+          if (e && e.rethrow === true) {
+            exceptionHandler(e);
```
gkalpak: A comment would be nice here too. E.g. something like:

> This error is explicitly marked for being passed to the `$exceptionHandler`.

gkalpak (changes requested):
Unfortunately, I couldn't come up with anything "less ugly" :smiley:
So, happy to go with this approach.

Thx for looking into it @marcin-wosinek 👍

marcin-wosinek:
Thanks for the feedback! looks that it fixed the issues on tests too:)

src/ng/q.js
```diff
@@ -364,6 +364,10 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
           }
         } catch (e) {
           rejectPromise(promise, e);
+          // This error is explicitly marked for being passed to the $exceptionHandler
+          if (e && e.$$passToExceptionHandle === true) {
```
gkalpak: `...Handler` (with an `r` in the end).

src/ngMock/angular-mocks.js
```diff
@@ -1438,10 +1438,16 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
         return;
       }
     }
-    throw wasExpected ?
+    var error = wasExpected ?
         new Error('No response defined !') :
         new Error('Unexpected request: ' + method + ' ' + url + '\n' +
                   (expectation ? 'Expected ' + expectation : 'No more request expected'));
+
+    // In addition to be being converted to a rejection, this error also need to be passed to
```
gkalpak: _need_ --> _needs_

Narretz:
We talked about this yesterday in the meeting and @petebacondarwin had another idea how to implement this, maybe he can outline it here

petebacondarwin:
The general concern was with potential collisions with user code by checking a "custom" property on the error objects. I accept that this is mitigated by the use of `$$` on the property. But still, I think it would be safer to use the other approach (suggested here: https://github.com/angular/angular.js/issues/15855#issuecomment-320526163) where we throw a "private" error class instead.

Since `ngMock` is in a separate closure to the `$q` service this error class would need to be exposed somewhere. We thought the safest place would be on `$q` itself (e.g. `$q.$$PassToExceptionHandlerError`).  For now we would keep this double dollared (`$$`) to indicate it is private but it could be made public if there was interest in it outside of `ngMock`.

@mgol suggested that an even cleaner approach would be to override the `$q` service in `ngMock` with the functionality described above, which would mean that production `$q` is not changed but that we have the desired effect when testing. We thought this was interesting but we concerned that the cost of changing `$q` to expose the necessary parts to support the override would be more complicating than just supporting it natively.

@mgol also pointed out that this "special" behaviour would not strictly be in keeping with the Promises A+ spec, even though it would pass the tests...

gkalpak:
> @mgol also pointed out that this "special" behaviour would not strictly be in keeping with the Promises A+ spec, even though it would pass the tests...

If it's passing the test... 😛 Why would it not be compiant with Promises/A+?
Our previous discussions about this "special" behavior of `$q` (i.e. passing stuff to the `$exceptionHandler`, which sometimes throws) were non-conclusive :grin:

mgol:
> If it's passing the test... 😛 Why would it not be compiant with Promises/A+?

Tests can almost never cover 100% of the spec. In this case, the spec doesn't allow to treat some rejections/errors differently.

gkalpak:
> In this case, the spec doesn't allow to treat some rejections/errors differently.

OOC, where does it say so?

mgol:
The spec describes what needs to happen when an exception is thrown in a
handler. It doesn't specify any requirements for those exceptions, hence
all the rules have to apply for every exception type.

gkalpak:
What the spec describes _does_ happen with all exceptions. But for some exceptions, additional stuff may happen.

But I do see your point. (I feel like we've had this conversation before 😛)

mgol:
@gkalpak Additional stuff may happen (we can then only discuss if it lies in the spirit of the spec) but that's not the point here. The spec mandates all exceptions need to be passed to the `catch` handlers and here the idea is to create a special exception class that avoids it. From the discussion at the last team meeting just re-throwing in the `exceptionHandler` wouldn't work as then it would be caught again by the `catch` handlers anyway. Or is there a way to make it work & not break the spec compliance?

gkalpak:
The promise will be [rejected as usual][1], but the error will also be [passed to the `$exceptionHandler`][2]. This is essentially the same that happened for all `Error` instances until not long ago. It will now happen for specific errors only (that will only be thrown during tests).

[1]: https://github.com/angular/angular.js/pull/16150/files#diff-8c0e145b42564a7c2838e221e0a0322aR366
[2]: https://github.com/angular/angular.js/pull/16150/files#diff-8c0e145b42564a7c2838e221e0a0322aR369

petebacondarwin:
Has anyone implemented our agreed solution here?

Narretz:
I don't think so. I think we wanted to wait for @gkalpak to return, as he reviewed this initially

gkalpak:
As per https://github.com/angular/angular.js/pull/16150#issuecomment-323097130, I am happy to go with the special Error class. @marcin-wosinek, would you be up to updating the PR with the new implementation (as described in https://github.com/angular/angular.js/pull/16150#issuecomment-322808943)?

marcin-wosinek:
yes, I'm about to update the implementation.

marcin-wosinek:
Looks it failed before due to some e2e glitch. It's ready to final review.

src/ng/q.js
```diff
@@ -343,6 +343,14 @@ function qFactory(nextTick, exceptionHandler, errorOnUnhandledRejections) {
     }
   });
 
+  function PassToExceptionHandlerError(message) {
+    this.name = '$$PassToExceptionHandlerError';
+    this.message = message;
+    this.stack = (new Error()).stack;
+  }
```
gkalpak: "Subclassing" Error is actually tricky. For example, different browsers use different properties (such as `line`, `sourceId`, `stackArray`). Other browsers do not populate the `stack` property until the error is actually thrown (so `(new Error).stack` will be empty). And finally, the stack trace will be a little confusing by showing a slightly different location:

```
# Expected
Error: Some error message
    at <where-the-error-was-actually-thrown>:<line>:<column>
    ...

# Actual
Error
    at PassToExceptionHandlerError (<file>:<line>:<column>)
    at <where-the-error-was-actually-thrown>:<line>:<column>
    ...
```

Here is what we do in `ngMock` to mitigate these issues:

https://github.com/angular/angular.js/blob/ecc09a45f02454e4e2478ad22cd0d179dccc7916/src/ngMock/angular-mocks.js#L3109-L3118

https://github.com/angular/angular.js/blob/ecc09a45f02454e4e2478ad22cd0d179dccc7916/src/ngMock/angular-mocks.js#L3122-L3128

https://github.com/angular/angular.js/blob/ecc09a45f02454e4e2478ad22cd0d179dccc7916/src/ngMock/angular-mocks.js#L3162

##
Another idea would be wrap the error in a non-error object and unwrap it in `$q`:

**src/ng/q.js**
```diff
+$Q.$$PassToExceptionHandlerWrapper = function PassToExceptionHandlerWrapper(error) { this.error = error; }
 ...
 function processQueue(state) {
   ...
   } catch (e) {
+    var passToExceptionHandler = e && e instanceof PassToExceptionHandlerWrapper;
+    e = passToExceptionHandler ? e.error : e;
     rejectPromise(promise, e);
+    if (passToExceptionHandler) {
+      exceptionHandler(e);
+    }
   }
```

**src/ngMock/angular-mocks.js**
```diff
+try {
+  // Throw the error to ensure the `stack` property is populated.
   throw wasExpected ?
       new Error('No response defined !') :
       new Error('Unexpected request: ' + method + ' ' + url + '\n' +
           (expectation ? 'Expected ' + expectation : 'No more request expected'));
+} catch (err) {
+  throw new $q.$$PassToExceptionHandlerWrapper(err);
+}
```

Then again, @mgol will say this violates the Promises/A+ spec and he will be right (since we will be rejecting with a different object than the one that was thrown) 😟 

##
Given all that, I think I might prefer the original "custom `$$passToExceptionHandler` property on the error" approach 😞 

What do others think?

gkalpak (changes requested):

src/ng/q.js
petebacondarwin: Are these issues with custom Errors still valid on the browser we support? I thought they were more IE9ish??

gkalpak: The "no stack until thrown" issue is IE10+ (and PhantomJS which is not officially supported).
I am not sure about the extra properties, but iirc they are Safari-specific (could be wrong though).
The "extra line in stack trace" issue is by design, but should be easy to work around.

Basically all the issues are "work-around-able". The question is, is it worth the extra complexity?
<sub>(I don't feel strongly about it, though. Happy to go either way.)</sub>

src/ngMock/angular-mocks.js
```diff
@@ -1438,10 +1438,15 @@ function createHttpBackendMock($rootScope, $timeout, $delegate, $browser) {
...
-    throw wasExpected ?
-        new Error('No response defined !') :
-        new Error('Unexpected request: ' + method + ' ' + url + '\n' +
+    var error = wasExpected ?
+        new $q.$$PassToExceptionHandlerError('No response defined !') :
+        new $q.$$PassToExceptionHandlerError('Unexpected request: ' + method + ' ' + url + '\n' +
                   (expectation ? 'Expected ' + expectation : 'No more request expected'));
+
+    // In addition to be being converted to a rejection, this error also needs to be passed to
+    // the $exceptionHandler and be rethrown (so that the test fails).
```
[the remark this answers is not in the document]
marcin-wosinek: sure

src/ng/q.js
marcin-wosinek: So there are 3 options:
1. Field on normal error (previous solution
2. Class inheriting from error (current code in PR, plus maybe changes to the way class defined)
3. Class wrapping error

Which way should be implemented?

marcin-wosinek:
master was updated with first suggestion from:
https://github.com/angular/angular.js/pull/16150#discussion_r137201542

the second suggestion is applied in:
https://github.com/marcin-wosinek/angular.js/tree/16150-b

the original proposition is in:
https://github.com/marcin-wosinek/angular.js/tree/16150-c

petebacondarwin:
@marcin-wosinek - thank you so much for you work and patience on this PR.

We had a long chat again today and have finally decided, after all to go with your original proposal (https://github.com/marcin-wosinek/angular.js/tree/16150-c). I will squash and merge the two commits from that branch, so there is no need to update this PR any further.

I am sorry it has taken so long to deal with this.

marcin-wosinek:
Cool; great to have it finally land:)

PR #16154, opened 2017-08-08
docs($compileProvider): more information on debugInfoEnabled

oBusk:

Improves documentation about what `$compileProvider.debugInfoEnabled(true)` does.

**Other information**:

Based on [compile.js](https://github.com/angular/angular.js/blob/b4d1e5e492df43c49c9e4215ea75d9155fd8ba88/src/ng/compile.js#L1819...L1851) source code.

Please feel free to improve the wording or come with feedback.

Related: #16155

### DIFF
--- a/src/ng/compile.js
+++ b/src/ng/compile.js
@@ -1389,8 +1389,8 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    * * `ng-binding` CSS class
    * * `ng-scope` and `ng-isolated-scope` CSS classes
    * * `$binding` data property containing an array of the binding expressions
-   * * `$scope`, `$isolateScope` or `$isolateScopeNoTemplate` data properties which are accessable via the
-   *   {@link angular.element#methods `scope()` or `isolateScope()` methods}.
+   * * Data properties used by the {@link angular.element#methods `scope()`/`isolateScope()` methods} to return
+   *   the element's scope.
    * * Placeholder comments will contain information about what directive and binding caused the placeholder.
    *   E.g. `<!-- ngIf: shouldShow() -->`
    *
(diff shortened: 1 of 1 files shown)


src/ng/compile.js
```diff
@@ -1387,7 +1387,11 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    * binding information and a reference to the current scope on to DOM elements.
    * If enabled, the compiler will add the following to DOM elements that have been bound to the scope
    * * `ng-binding` CSS class
+   * * `ng-scope` and `ng-isolated-scope` CSS classes
    * * `$binding` data property containing an array of the binding expressions
+   * * `scope()` and `isolateScope()` allowing access to scope from DOM Element
```
gkalpak: It doesn't add the methods. The methods are always present. It adds the `$scope` or `$isolateScope` (or `$isolateScopeNoTemplate` data properties (that can be used by the `.scope()` and `.isolateScope()` methods).

gkalpak (changes requested):

oBusk:
@gkalpak I've ammended according to your comments. Thanks for the feedback.

src/ng/compile.js
```diff
@@ -1387,7 +1387,12 @@ function $CompileProvider($provide, $$sanitizeUriProvider) {
    * binding information and a reference to the current scope on to DOM elements.
    * If enabled, the compiler will add the following to DOM elements that have been bound to the scope
    * * `ng-binding` CSS class
+   * * `ng-scope` and `ng-isolated-scope` CSS classes
    * * `$binding` data property containing an array of the binding expressions
+   * * `$scope`, `$isolateScope` or `$isolateScopeNoTemplate` data properties which are accessable via the
```
gkalpak: This is too much implementation details imo. Please change to something like:

> Data properties used by the {@link angular.element#methods `scope()`/`isolateScope()` methods} to return the element's scope.

gkalpak (changes requested):

src/ng/compile.js
oBusk: That is a very good point

PR #16161, opened 2017-08-10
fix($rootScope): fix potential memory leak when removing scope listeners 

### LINKED ISSUE #16135: Small Memory Leak in $rootScope.$on
**I'm submitting a ...**

- [x] bug report
- [ ] feature request

**Current behavior:**

Registering a new handler with `$rootScope.$on` [`push`es the handler to the end of a list](https://github.com/angular/angular.js/blob/master/src/ng/rootScope.js#L1169). Removing the handler [only `null`s out the array index that contains the handler](https://github.com/angular/angular.js/blob/master/src/ng/rootScope.js#L1183). The code never resizes the array to remove the `null` entries, leading to an array that grows without bounds.

Given:

* An event `'e'` such that there is always 1 registered listener on the event (preventing the handler list from being removed).
  * Note that this is easily achievable *even if all handlers are regularly removed.* All it takes is another call to `$on('e', handler)` to run before all handlers are de-registered.
* An application that regularly calls `$rootScope.$on('e'`, handler)` and responsibly de-registers the handler over the course of a session

...then the handler list for `'e'` will grow by 1 element every time `$rootScope.$on('e', handler)` is called. The end result is an array that looks like this: `[null, null, null, ... handler, handler]`.

I'm observing this happen in an app with the [$translateChangeSuccess event](https://github.com/angular-translate/angular-translate/blob/master/src/directive/translate.js#L320). The array grows to about 40 `null` entries over 8 navigations to different views/pages.

The leak is small, but avoidable.

**Expected / new behavior:**

The handler list does not grow unboundedly over the course of a session. Handler positions are either re-used (as in a [free list-based memory allocator](https://en.wikipedia.org/wiki/Free_list)), or handlers are registered in an object where handler positions are `delete`-ed once they are de-registered.

See above description.

**Angular version:** Observed in 1.4.1, but code still present in 1.6.

**Browser:**

All

**Anything else:**

See above for suggestions on how to fix.

Note that I am not heavily experienced with AngularJS, but I am heavily experienced in JavaScript. Hence, I am not able to provide you with a succinct reproduction of the issue, but the description and a quick glance of the code should confirm the problem.

jbedard:
This attempts to fix #16135, but is assuming that browsers will use some type of memory efficient array implementation for sparse arrays (which they appear to do, see [1](https://stackoverflow.com/questions/614126/why-is-array-push-sometimes-faster-than-arrayn-value/614255#614255), [2](https://www.safaribooksonline.com/library/view/javascript-the-definitive/9781449393854/ch07s03.html), [3](https://www.html5rocks.com/en/tutorials/speed/v8/#toc-topic-numbers)).  In the worst case this ~~should have no change~~ may cause some extra swapping of array implementations and/or not fix the bug, in the best case it will be more memory efficient in this edge case.

Also added a couple tests that show why we don't simply use `splice`.

Thanks @jvilk for investigating this `delete` solution.

### DIFF
--- a/src/ng/rootScope.js
+++ b/src/ng/rootScope.js
@@ -1207,7 +1207,10 @@ function $RootScopeProvider() {
         return function() {
           var indexOfListener = namedListeners.indexOf(listener);
           if (indexOfListener !== -1) {
-            namedListeners[indexOfListener] = null;
+            // Use delete in the hope of the browser deallocating the memory for the array entry,
+            // while not shifting the array indexes of other listeners.
+            // See issue https://github.com/angular/angular.js/issues/16135
+            delete namedListeners[indexOfListener];
             decrementListenerCount(self, 1, name);
           }
         };
--- a/test/ng/rootScopeSpec.js
+++ b/test/ng/rootScopeSpec.js
@@ -1903,6 +1903,90 @@ describe('Scope', function() {
         }));
 
 
+        it('should call next listener after removing the current listener via its own handler', inject(function($rootScope) {
+          var listener1 = jasmine.createSpy('listener1').and.callFake(function() { remove1(); });
+          var remove1 = $rootScope.$on('abc', listener1);
+
+          var listener2 = jasmine.createSpy('listener2');
+          var remove2 = $rootScope.$on('abc', listener2);
+
+          var listener3 = jasmine.createSpy('listener3');
+          var remove3 = $rootScope.$on('abc', listener3);
+
+          $rootScope.$broadcast('abc');
+          expect(listener1).toHaveBeenCalled();
+          expect(listener2).toHaveBeenCalled();
+          expect(listener3).toHaveBeenCalled();
+
+          listener1.calls.reset();
+          listener2.calls.reset();
+          listener3.calls.reset();
+
+          $rootScope.$broadcast('abc');
+          expect(listener1).not.toHaveBeenCalled();
+          expect(listener2).toHaveBeenCalled();
+          expect(listener3).toHaveBeenCalled();
+        }));
+
+
+        it('should call all subsequent listeners when a previous listener is removed via a handler', inject(function($rootScope) {
+          var listener1 = jasmine.createSpy();
+          var remove1 = $rootScope.$on('abc', listener1);
+
+          var listener2 = jasmine.createSpy().and.callFake(remove1);
+          var remove2 = $rootScope.$on('abc', listener2);
+
+          var listener3 = jasmine.createSpy();
+          var remove3 = $rootScope.$on('abc', listener3);
+
+          $rootScope.$broadcast('abc');
+          expect(listener1).toHaveBeenCalled();
+          expect(listener2).toHaveBeenCalled();
+          expect(listener3).toHaveBeenCalled();
+
+          listener1.calls.reset();
+          listener2.calls.reset();
+          listener3.calls.reset();
+
+          $rootScope.$broadcast('abc');
+          expect(listener1).not.toHaveBeenCalled();
+          expect(listener2).toHaveBeenCalled();
+          expect(listener3).toHaveBeenCalled();
+        }));
+
+
+        it('should not call listener when removed by previous', inject(function($rootScope) {
+          var listener1 = jasmine.createSpy('listener1');
+          var remove1 = $rootScope.$on('abc', listener1);
+
+          var listener2 = jasmine.createSpy('listener2').and.callFake(function() { remove3(); });
+          var remove2 = $rootScope.$on('abc', listener2);
+
+          var listener3 = jasmine.createSpy('listener3');
+          var remove3 = $rootScope.$on('abc', listener3);
+
+          var listener4 = jasmine.createSpy('listener4');
+          var remove4 = $rootScope.$on('abc', listener4);
+
+          $rootScope.$broadcast('abc');
+          expect(listener1).toHaveBeenCalled();
+          expect(listener2).toHaveBeenCalled();
+          expect(listener3).not.toHaveBeenCalled();
+          expect(listener4).toHaveBeenCalled();
+
+          listener1.calls.reset();
+          listener2.calls.reset();
+          listener3.calls.reset();
+          listener4.calls.reset();
+
+          $rootScope.$broadcast('abc');
+          expect(listener1).toHaveBeenCalled();
+          expect(listener2).toHaveBeenCalled();
+          expect(listener3).not.toHaveBeenCalled();
+          expect(listener4).toHaveBeenCalled();
+        }));
+
+
         it('should decrement ancestor $$listenerCount entries', inject(function($rootScope) {
           var child1 = $rootScope.$new(),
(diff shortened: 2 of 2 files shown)


Narretz:
The errors look weird. An incompatibility between yarn and dgeni-packages? Strange because I don't think we changed anything.

src/ng/rootScope.js
```diff
@@ -1180,7 +1180,10 @@ function $RootScopeProvider() {
         return function() {
           var indexOfListener = namedListeners.indexOf(listener);
           if (indexOfListener !== -1) {
-            namedListeners[indexOfListener] = null;
+            // Use delete in the hope of the browser deallocating the memory for the array entry,
```
gkalpak: > ...in the hope of the browser deallocating...

Why wouldn't it? :grin:

test/ng/rootScopeSpec.js
```diff
@@ -2316,6 +2316,68 @@ describe('Scope', function() {
...
+          var listener2 = jasmine.createSpy();
+          var remove2 = $rootScope.$on('abc', listener2);
+
+          var listener3 = jasmine.createSpy();
+          var remove3 = $rootScope.$on('abc', listener3);
+
+          $rootScope.$broadcast('abc');
+          expect(listener2).toHaveBeenCalled();
+          expect(listener3).toHaveBeenCalled();
+        }));
```
gkalpak: You could run a second round (it this and the subsequent tests) to make sure the listener will be called correctly and nothing will break after the initial run. E.g.:

```js
listener2.calls.reset();
listener3.calls.reset();

$rootScope.$broadcast('abc');
expect(listener2).toHaveBeenCalled();
expect(listener3).toHaveBeenCalled();
```

gkalpak (approved):
A minor suggestion, but LGTM 👍

src/ng/rootScope.js
jbedard: Because it is implementation dependent, not part of any spec etc.

Narretz:
@jbedard can you please rebase this? The tests currently don't run because of a Travis problem at the branch point.

test/ng/rootScopeSpec.js
jbedard: I've added this to all 3 tests

jbedard:
Rebased. Adjusted the tests a bit like @gkalpak suggested.

jbedard:
Another good read about array storage/optimizations: https://www.html5rocks.com/en/tutorials/speed/v8/#toc-topic-numbers

Again v8 specific, but this confirms what we've been talking about (and what @jvilk [measured](https://github.com/angular/angular.js/issues/16135#issuecomment-319464762)) with array vs dictionary versions of an array. Using `delete` could cause a switch to the dictionary implementation which forces memory reallocation/copying, but then (for sparse arrays) it will use less memory.

src/ng/rootScope.js
Narretz: I think there should be a more descriptive comment here, e.g. "depending on browser implementations, delete should deallocate the memory"

jbedard: I'm not sure if it "should" though. Is there anything that actually says it should? I think browsers (some? all?) do this just to be more efficient. I kind of like the "in the hope of" so I'm having trouble finding a better description...

Narretz: Did we ever find a conclusive answer to the actual implementation in specific browsers? Or is this based on benchmarks? Either info could be listed here and then maybe a note that we assume other browsers do it as well.

jbedard: Based on benchmarks Chrome and FF seem to both use a more memory efficient array implementation for sparse arrays (which I [noted here](https://github.com/angular/angular.js/issues/16135#issuecomment-334986679)). But I'm not sure about all browsers, or the details of when Chrome and FF do it, and I'm assuming there is no actual standard for this.

test/ng/rootScopeSpec.js
```diff
@@ -2316,6 +2316,99 @@ describe('Scope', function() {
...
+          expect(0 in $rootScope.$$listeners['abc']).toBe(true);
+
+          remove1();
+
+          expect($rootScope.$$listeners['abc'].length).toBe(2);
+          expect(0 in $rootScope.$$listeners['abc']).toBe(false);
+        }));
+
+
+        it('should call next listener when removing current', inject(function($rootScope) {
```
jbedard: I just realized there are two similar tests to this one where the current listener gets removed for [$emit](https://github.com/angular/angular.js/blob/d0d829899c4bcdf4b322d45245d7b811cd4cb855/test/ng/rootScopeSpec.js#L2438) and [$broadcast](https://github.com/angular/angular.js/blob/d0d829899c4bcdf4b322d45245d7b811cd4cb855/test/ng/rootScopeSpec.js#L2472) (https://github.com/angular/angular.js/commit/e6966e05f508d1d2633b9ff327fea912b12555ac) under `describe('$emit'`. Maybe those two should move here to replace this one? Or just keep this extra one?

jbedard:
Actually... I came up with an alternative that might be better, essentially doing the same as [$$digestWatchIndex](https://github.com/angular/angular.js/blob/v1.6.6/src/ng/rootScope.js#L840). Basically making the `$$listeners` iteration index global (per event type per scope) so it can be adjusted, if necessary, when a listener is removed. The only constraint is this means you can not iterator over `$$listeners` recursively which I _think_ is already a restriction due to the [prevoius array cleanup](https://github.com/angular/angular.js/blob/v1.6.6/src/ng/rootScope.js#L1337) that essentially had the same restriction.

See https://github.com/jbedard/angular.js/commit/68247ee4c13d6ef7fedb0aad09b8ce80d8a9fd61 - WDYT?

Narretz:
> The only constraint is this means you can not iterator over $$listeners recursively 

You mean iterate over $$listeners and modify while you iterate over them? That sounds like an acceptable tradeoff.

jbedard:
I think it would be recursive `$emit` or `$broadcast` calls of the same event name. I'm pretty sure this was already broken in some cases, but this (storing the iteration index on the array) would break it in all cases.

Narretz:
Since I don't really use events in AngularJs, I can't say if there are valid use cases for recursive emission and broadcasting. It's probably a BC for someone ...

gkalpak:
We decided to land this PR as is for 1.6.0 (to avoid any breaking changes and still avoid the memory leak).
For 1.7.0, we will implement something like jbedard/angular.js@68247ee (which avoids making iterations slower as the array grows in length) with an explicit check to prevent nested listener iterations and a breaking change notice.

Narretz:
SGTM

Narretz:
Don't forget to change the branch target to v1.6.x for this one

jbedard:
I've rebased this onto v1.6.x, also updated the first commit to include the changes from #16293.

gkalpak (approved):
Still LGTM :smiley:

jbedard:
I've merged this to v1.6.x and cherry-picked the first commit (extra tests, not specifically for this bug) into master.

PR #4694, opened 2013-10-29
feat($injector): ability to load new modules after bootstrapping

cleyshan:
Added new method to $injector called loadNewModules.
Added corresponding unit tests for new method.

$injector.loadNewModules can be called any time after bootstrap so that any new angular modules defined after bootstrap can still function correctly.  Please see https://groups.google.com/d/msg/angular/w0ZEBz02l8s/YK3H0eMmlxcJ for a detailed justification and explanation of this simple but powerful change.

PR #16224, opened 2017-09-13
feat($injector): ability to load new modules after bootstrapping

petebacondarwin:
The new method `$injector.loadNewModules(modules)` will add each of the
injectables to the injector and execute all of the run blocks for each
module passed to the method.

* The application developer is responsible for loading the code containing
the modules.
* Modules cannot be unloaded.
* Previously compiled HTML will not be affected by newly loaded directives,
filters and components.

**Other information**:

### DIFF
--- a/src/auto/injector.js
+++ b/src/auto/injector.js
@@ -337,7 +337,46 @@ function annotate(fn, strictDi, name) {
  *
  * @returns {Array.<string>} The names of the services which the function requires.
  */
-
+/**
+ * @ngdoc method
+ * @name $injector#loadNewModules
+ *
+ * @description
+ *
+ * **This is a dangerous API, which you use at your own risk!**
+ *
+ * Add the specified modules to the current injector.
+ *
+ * This method will add each of the injectables to the injector and execute all of the config and run
+ * blocks for each module passed to the method.
+ *
+ * If a module has already been loaded into the injector then it will not be loaded again.
+ *
+ * * The application developer is responsible for loading the code containing the modules; and for
+ * ensuring that lazy scripts are not downloaded and executed more often that desired.
+ * * Previously compiled HTML will not be affected by newly loaded directives, filters and components.
+ * * Modules cannot be unloaded.
+ *
+ * You can use {@link $injector#modules `$injector.modules`} to check whether a module has been loaded
+ * into the injector, which may indicate whether the script has been executed already.
+ *
+ * ## Example
+ *
+ * Here is an example of loading a bundle of modules, with a utility method called `getScript`:
+ *
+ * ```javascript
+ * app.factory('loadModule', function($injector) {
+ *   return function loadModule(moduleName, bundleUrl) {
+ *     return getScript(bundleUrl).then(function() { $injector.loadNewModules([moduleName]); });
+ *   };
+ * })
+ * ```
+ *
+ * @param {Array<String|Function|Array>=} mods an array of modules to load into the application.
+ *     Each item in the array should be the name of a predefined module or a (DI annotated)
+ *     function that will be invoked by the injector as a `config` block.
+ *     See: {@link angular.module modules}
+ */
 
 
 /**
@@ -700,6 +739,11 @@ function createInjector(modulesToLoad, strictDi) {
   instanceInjector = protoInstanceInjector.get('$injector');
   instanceInjector.strictDi = strictDi;
   forEach(runBlocks, function(fn) { if (fn) instanceInjector.invoke(fn); });
+
+  instanceInjector.loadNewModules = function(mods) {
+    forEach(loadModules(mods), function(fn) { if (fn) instanceInjector.invoke(fn); });
+  };
+
 
   return instanceInjector;
 
--- a/test/auto/injectorSpec.js
+++ b/test/auto/injectorSpec.js
@@ -19,7 +19,7 @@ describe('injector.modules', function() {
         .info({ version: '1.2' })
         .provider('test', ['$injector', function($injector) {
           providerInjector = $injector;
-          return { $get: function() {} };
+          return {$get: function() {}};
         }]);
       module('test1');
       // needed to ensure that the provider blocks are executed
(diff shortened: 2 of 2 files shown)


petebacondarwin:
The original commit came from #4694 and the author did sign CLA.

src/auto/injector.js
```diff
@@ -337,7 +337,53 @@ function annotate(fn, strictDi, name) {
...
+ *      $log.info('Script loaded : ' + name);
+ *      onReady(name);
+ *    });
+ *
+ *    function onReady(name) {
+ *      var moduleName = name + '.html.bundle';
+ *      $log.debug('AMD >> Loading Angular module [' + moduleName + '] ...');
+ *
+ *      if (config.modules) {
+ *        $injector.loadNewModules(config.modules);
```
gkalpak: Why this? Shouldn't the lazy-loaded module depend on these (as in `angular.module('someModule', [...dependent modules here...])`), in which case `loadNewModules` would automatically load them?

src/auto/injector.js
```diff
@@ -337,7 +337,53 @@ function annotate(fn, strictDi, name) {
...
+ *        $log.debug('AMD >> Children modules loaded >>>');
+ *        $log.debug(config.modules);
+ *      }
+ *
+ *      $injector.loadNewModules([moduleName]);
+ *      $log.debug('AMD >> Angular module [' + moduleName + '] loaded.');
+ *    }
+ * });
```
gkalpak: FWIW, I find the example unnecessarily complicated. I would expect something like:

```js
app.factory('loadModule', function($injector) {
  return function loadModule(moduleName, bundleUrl) {
    return getScript(bundleUrl).then(function() { $injector.loadNewModules([moduleName]); });
  };
})
```

test/auto/injectorSpec.js
```diff
@@ -152,6 +152,90 @@ describe('injector', function() {
...
+    it('should allow new modules to be added after injector creation and their runBlocks executed', function() {
+      var log = '';
+      angular.module('0', [], function() { log += '0'; }).run(function() { log += '1'; });
+      var injector = createInjector(['0']);
+      log += '2';
+      angular.module('a', [], function() { log += 'a'; }).run(function() { log += 'A'; });
+      angular.module('b', ['a'], function() { log += 'b'; }).run(function() { log += 'B'; });
+      injector.loadNewModules([
+          'b',
+          valueFn(function() { log += 'C'; }),
```
gkalpak: I would add some config fns here two :grin:

test/auto/injectorSpec.js
```diff
@@ -152,6 +152,90 @@ describe('injector', function() {
...
+          [valueFn(function() { log += 'D'; })]
+      ]);
+      expect(log).toEqual('012abABCD');
+    });
+
+
+    it('should be able to register a service from a new module', function() {
+      var injector = createInjector([]);
+      angular.module('a');
```
gkalpak: Did you mean `angular.module('a', [])`?

gkalpak (changes requested):
Some minor comments. LGTM overall.

src/auto/injector.js
```diff
@@ -337,7 +337,53 @@ function annotate(fn, strictDi, name) {
...
+ * @name $injector#loadNewModules
+ *
+ * @description
+ *
+ * **This is an dangerous API, which you use at your own risk!**
+ *
+ * Add the specified modules to the current injector.
+ *
+ * This method will add each of the injectables to the injector and execute all of the run blocks
+ * for each module passed to the method.
```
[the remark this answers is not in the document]
petebacondarwin: blockas -> blocks :-P

src/auto/injector.js
petebacondarwin: Not sure to be honest, this was a bit of a cut-n-paste job.

src/auto/injector.js
```diff
@@ -701,6 +747,11 @@ function createInjector(modulesToLoad, strictDi) {
   instanceInjector.strictDi = strictDi;
   forEach(runBlocks, function(fn) { if (fn) instanceInjector.invoke(fn); });
```
jbedard: Worth putting this in a method to avoid duplicating it a few lines down?

petebacondarwin:
I have reworked following the comments.

src/auto/injector.js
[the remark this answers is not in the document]
gkalpak: There were tests for them already.

test/auto/injectorSpec.js
```diff
@@ -152,6 +152,172 @@ describe('injector', function() {
...
+      angular.module('a', []);
+      angular.module('a').directive('aDirective', function() {
+        return {  template: 'test directive' };
+      });
+      injector.loadNewModules(['a']);
+      injector.invoke(function($compile, $rootScope) {
+        var scope = $rootScope.$new();
+        var html = '<div a-directive></div>';
+        var elem = angular.element(html);
+        $compile(elem)(scope);  // compile and link
```
gkalpak: Save up to three(3) lines with:

```js
var elem = $compile('<div a-directive></div>')($rootScope);
$rootScope.$digest();
```

🌿

test/auto/injectorSpec.js
```diff
@@ -152,6 +152,172 @@ describe('injector', function() {
...
+        expect(filter('test')).toEqual('test filtered');
+      });
+    });
+
+
+    it('should be able to register a directive from a new module', function() {
+      var injector = createInjector(['ng']);
+      angular.module('a', []);
+      angular.module('a').directive('aDirective', function() {
+        return {  template: 'test directive' };
```
gkalpak: Way too much whitespace inside object 😇

gkalpak (approved):
One real (docs related) comment and a handful of nitpicks :grin:
LGTM otherwise.

<sub>:100: for the added tests btw</sub>

src/auto/injector.js
```diff
@@ -337,7 +337,42 @@ function annotate(fn, strictDi, name) {
...
+ *
+ * ```javascript
+ *  app.factory('loadModule', function($injector) {
+ *  return function loadModule(moduleName, bundleUrl) {
+ *    return getScript(bundleUrl).then(function() { $injector.loadNewModules([moduleName]); });
+ *   };
+ * })
+ * ```
+ *
+ * @param {Array<String|Function|Array>=} mods an array of modules to load into the application.
```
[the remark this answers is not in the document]
petebacondarwin: I cut and pasted this from the `angular.bootstrap()` API code: https://github.com/angular/angular.js/blob/master/src/Angular.js#L1773

src/auto/injector.js
```diff
@@ -337,7 +337,42 @@ function annotate(fn, strictDi, name) {
...
+ *     Each item in the array should be the name of a predefined module or a (DI annotated)
+ *     function that will be invoked by the injector as a `config` block.
+ *     See: {@link angular.module modules}*
```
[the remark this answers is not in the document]
petebacondarwin: no

src/auto/injector.js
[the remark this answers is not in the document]
gkalpak: 😞  - I am not going to block the PR for that then  😛

test/auto/injectorSpec.js
```diff
@@ -152,6 +152,164 @@ describe('injector', function() {
...
+
+    it('should execute runBlocks and configBlocks in the correct order', function() {
+      var log = [];
+      angular.module('initial', [], function() { log.push(1); })
+        .config(function() { log.push(2); })
+        .run(function() { log.push(3); });
+      var injector = createInjector(['initial']);
+      log.push('created');
+
+      angular.module('a', [], function() { log.push(4);})
```
gkalpak: Missing whitespace after `;` 🔢

gkalpak (approved):
LGTM (except for a couple of indentation issues - including https://github.com/angular/angular.js/pull/16224#discussion_r139167402 :grin:)

src/auto/injector.js
```diff
@@ -337,7 +337,42 @@ function annotate(fn, strictDi, name) {
...
+ * * Previously compiled HTML will not be affected by newly loaded directives, filters and components.
+ *
+ * ## Example
+ *
+ * Here is an example of loading a bundle of modules, with a utility method called `getScript`:
+ *
+ * ```javascript
+ *  app.factory('loadModule', function($injector) {
+ *  return function loadModule(moduleName, bundleUrl) {
+ *    return getScript(bundleUrl).then(function() { $injector.loadNewModules([moduleName]); });
```
Narretz: You could use `fetch` in-place here for `getScript`. Only thing that might happen is that people use it and complain that it doesn't work in IE9 ...

src/auto/injector.js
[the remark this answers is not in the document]
Narretz: Then why does it say `mods` instead of `modules`? 😛

Narretz (commented):
I assume one of the common use cases is to load additional modules when the user navigates inside the application, i.e. before a route is loaded.

This means the application must download the module and load it, but only for the first time (as the modules cannot be unloaded).

~~I don't see a mechanism for checking if a module has been loaded already. We don't expose the list in the injector, either. Should we add a convenience method `hasModule()` ?
You could always check if a specific service is in the DI, but this won't work for modules that only load components etc.
It's not enough that the injector won't load a module twice either, it's the download that must be avoided. Correct caching could handle that ofc.~~

Georgios pointed out that we have https://docs.angularjs.org/api/auto/service/$injector#modules for that

gkalpak:
> I don't see a mechanism for checking if a module has been loaded already. We don't expose the list in the injector, either.

Doesn't [$injector.modules][1] cover that?

[1]: https://docs.angularjs.org/api/auto/service/$injector#modules

Narretz:
It does, thanks. I forgot we added that. Maybe we could add a note about the property to the docs.

src/auto/injector.js
petebacondarwin: @Narretz I think the difference is that `getScript` also executes the script.

petebacondarwin:
@Narretz do you mean add a link to the `modules` property docs in the docs for the `loadModules` method?

Narretz:
@petebacondarwin that's right

petebacondarwin:
Cherry picked to 1.6.x as 6e78fee73258bb0ae36414f9db2e8734273e481b

PR #16226, opened 2017-09-14
docs(guide/Providers): clarification about Service

AlexDEVpro:
Сlarification for completeness of information about Service.

Docs update.

Docs says that Service object isn't available in config phase, but it's available by adding "Provider" suffix to service name. Then $get function can be used then and it's caused extra call of this function if same service, for example, has also been injected in controller.

Docs says that Service object is available in config phase, but it not the right way and Provider should be used instead in case a configuration is required.

Nope.

### DIFF
--- a/docs/content/guide/providers.ngdoc
+++ b/docs/content/guide/providers.ngdoc
@@ -398,7 +398,7 @@ To wrap it up, let's summarize the most important points:
 <tr>
   <td>object available in config phase</td>
   <td class="error">no</td>
-  <td class="error">no</td>
+  <td class="error">no\*\*\*</td>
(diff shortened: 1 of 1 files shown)


AlexDEVpro:
I signed it!

gkalpak:
Injecting providers not registered via `.provider()` (as this docs change mentions) is both bad practice and not officially supported. I think adding this to the docs (even if explicitly mentioning it is wrong) is more confusing than helpful.

Thx for taking the time to submit a PR, though 👍

AlexDEVpro:
Ok, perhaps you are right.

PR #16237, opened 2017-09-21
wip: feat(ngModel): run formatters / setModelValue

Narretz:
feature

There's currently no way to manually run the model -> view pipeline / formatters

An API to run the whole pipeline / the formatters is introduced.

No

**Other information**:

The PR includes both a `$format` and a `$setModelValue` function. I think only one is needed, we just need to decide which one.

$format:

- PRO: does the most needed thing, very straightforward: run the formatters and updates the viewValue
- CON: developer must run $render manually
- CON: empty classes are set even though the DOM is not updated. Could be moved to $render()
- CON: no equivalent on the view -> model side (inconsistent API)

$setModelValue
- PRO: equivalent to $setViewValue, runs the whole pipeline
- CON: unintuitive that the function argument should / must be set to the current $modelValue, so the control does not get out of sync with the scope (same behavior as $setViewValue though)
- CON: only runs $render if the viewValue has actually changed.
- CON: always runs the validation, even though it might not be necessary

Both methods can handle the basic case, where an app developer wants to run the formatters the view -> model pipeline has been run, see https://github.com/angular/angular.js/issues/3407 or https://github.com/angular/angular.js/pull/5221

I personally tend to introduce `$format` as it has the smaller surface area and introduces fewer side effects. The full model -> view pipeline is not really needed for most cases.

### DIFF
--- a/src/ng/directive/ngModel.js
+++ b/src/ng/directive/ngModel.js
@@ -878,6 +878,154 @@ NgModelController.prototype = {
    */
   $overrideModelOptions: function(options) {
     this.$options = this.$options.createChild(options);
+  },
+
+  /**
+   * @ngdoc method
+   *
+   * @name  ngModel.NgModelController#$processModelValue
+
+   * @description
+   *
+   * Runs the model -> view pipeline on the current
+   * {@link ngModel.NgModelController#$modelValue $modelValue}.
+   *
+   * The following actions are performed by this method:
+   *
+   * - the `$modelValue` is run through the {@link ngModel.NgModelController#$formatters $formatters}
+   * and the result is set to the {@link ngModel.NgModelController#$viewValue $viewValue}
+   * - the `ng-empty` or `ng-not-empty` class is set on the element
+   * - if the `$viewValue` has changed:
+   *   - {@link ngModel.NgModelController#$render $render} is called on the control
+   *   - the {@link ngModel.NgModelController#$validators $validators} are run and
+   *   the validation status is set.
+   *
+   * This method is called by ngModel internally when the bound scope value changes.
+   * Application developers usually do not have to call this function themselves.
+   *
+   * This function can be used when the `$viewValue` or the rendered DOM value are not correctly
+   * formatted and the `$modelValue` must be run through the `$formatters` again.
+   *
+   * #### Example
+   *
+   * Consider a text input with an autocomplete list (for fruit), where the items are
+   * objects with a name and an id.
+   * A user enters `ap` and then selects `Apricot` from the list.
+   * Based on this, the autocomplete widget will call `$setViewValue({name: 'Apricot', id: 443})`,
+   * but the rendered value will still be `ap`.
+   * The widget can then call `ctrl.$processModelValue()` to run the model -> view
+   * pipeline again, which formats the object to the string `Apricot`,
+   * then updates the `$viewValue`, and finally renders it in the DOM.
+   *
+   * <example module="inputExample" name="ng-model-process">
+     <file name="index.html">
+      <div ng-controller="inputController" style="display: flex;">
+        <div style="margin-right: 30px;">
+          Search Fruit:
+          <basic-autocomplete items="items" on-select="selectedFruit = item"></basic-autocomplete>
+        </div>
+        <div>
+          Model:<br>
+          <pre>{{selectedFruit | json}}</pre>
+        </div>
+      </div>
+     </file>
+     <file name="app.js">
+      angular.module('inputExample', [])
+        .controller('inputController', function($scope) {
+          $scope.items = [
+            {name: 'Apricot', id: 443},
+            {name: 'Clementine', id: 972},
+            {name: 'Durian', id: 169},
+            {name: 'Jackfruit', id: 982},
+            {name: 'Strawberry', id: 863}
+          ];
+        })
+        .component('basicAutocomplete', {
+          bindings: {
+            items: '<',
+            onSelect: '&'
+          },
+          templateUrl: 'autocomplete.html',
+          controller: function($element, $scope) {
+            var that = this;
+            var ngModel;
+
+            that.$postLink = function() {
+              ngModel = $element.find('input').controller('ngModel');
+
+              ngModel.$formatters.push(function(value) {
+                return (value && value.name) || value;
+              });
+
+              ngModel.$parsers.push(function(value) {
+                var match = value;
+                for (var i = 0; i < that.items.length; i++) {
+                  if (that.items[i].name === value) {
+                    match = that.items[i];
+                    break;
+                  }
+                }
+
+                return match;
+              });
+            };
+
+            that.selectItem = function(item) {
+              ngModel.$setViewValue(item);
+              ngModel.$processModelValue();
+              that.onSelect({item: item});
+            };
+          }
+        });
+     </file>
+     <file name="autocomplete.html">
+       <div>
+         <input type="search" ng-model="$ctrl.searchTerm" />
+         <ul>
+           <li ng-repeat="item in $ctrl.items | filter:$ctrl.searchTerm">
+             <button ng-click="$ctrl.selectItem(item)">{{ item.name }}</button>
+           </li>
+         </ul>
+       </div>
+     </file>
+   * </example>
+   *
+   */
+  $processModelValue: function() {
+    var viewValue = this.$$format();
+
+    if (this.$viewValue !== viewValue) {
+      this.$$updateEmptyClasses(viewValue);
+      this.$viewValue = this.$$lastCommittedViewValue = viewValue;
+      this.$render();
+      // It is possible that model and view value have been updated during render
+      this.$$runValidators(this.$modelValue, this.$viewValue, noop);
+    }
+  },
+
+  /**
+   * This method is called internally to run the $formatters on the $modelValue
+   */
+  $$format: function() {
+    var formatters = this.$formatters,
+        idx = formatters.length;
+
+    var viewValue = this.$modelValue;
+    while (idx--) {
+      viewValue = formatters[idx](viewValue);
+    }
+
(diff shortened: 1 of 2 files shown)


src/ng/directive/ngModel.js
```diff
@@ -892,13 +892,16 @@ function setupModelWatcher(ctrl) {
   // 4. view should be changed back to 'a'
   ctrl.$$scope.$watch(function ngModelWatch(scope) {
     var modelValue = ctrl.$$ngModelGet(scope);
+    return modelValue;
+  }, ngModelWatchAction);
 
+  function ngModelWatchAction(modelValue) {
     // if scope model value and ngModel value are out of sync
     // TODO(perf): why not move this to the action fn?
```
jbedard: Can delete this now

src/ng/directive/ngModel.js
```diff
@@ -892,13 +892,16 @@ function setupModelWatcher(ctrl) {
...
     if (modelValue !== ctrl.$modelValue &&
```
jbedard: Can this be deleted now?

src/ng/directive/ngModel.js
```diff
@@ -921,7 +924,8 @@ function setupModelWatcher(ctrl) {
     }
 
     return modelValue;
```
jbedard: Can this be deleted now?

Narretz:
@jbedard I had to revert the code that was moved to the watchActionFn, because it caused test failures. Specifcally, it caused failures when the model was updated inside ngChange or the model setter - so I think the way it's done is necessary.

petebacondarwin:
What about something like `$processModelValue`, which is effectively the same as your `setModelValue` but doesn't take an argument, instead just using whatever is in th current `$modelValue`?

Narretz:
Hm, that sounds like a good in-between option. It' doesn't have a public equivalent for view -> model but $$parseAndValidate could be exposed if necessary.

gkalpak:
Generally I prefer to keep the API consistent (it makes it much easier to wrap your head around), but in this case `$processModelValue()` sounds like a good idea.

BTW, I agree it makes sense to tie `updateEmptyClasses()` with the rendering.

Narretz:
Î'll update the PR then. Regarding $$updateEmptyClasses(), I thought about it again, and think it's okay that this is set before $render() is called. It's also more of an issue if we had a $format() function, because that would not have called render.
But $processModelValue calls everything in the pipeline, including $render(), so it's not a big deal that $$updateEmptyClasses() is called right before it.
On the view -> model pipeline it's also only called when the viewValue is committed, so it's also not directly tied to the actual DOM / render value.

Narretz:
Updated.
I left $$format and $$setModelValue as internal functions, and added a PoC autocomplete example to illustrate the use case.

src/ng/directive/ngModel.js
```diff
@@ -894,30 +1029,14 @@ function setupModelWatcher(ctrl) {
...
-    // TODO(perf): why not move this to the action fn?
+    // This cannot be moved to the action function, because it would not catch the
+    // case where the model is changed in the ngChange function or the model setter
     if (modelValue !== ctrl.$modelValue &&
-       // checks for NaN is needed to allow setting the model to NaN when there's an asyncValidator
-        // eslint-disable-next-line no-self-compare
-       (ctrl.$modelValue === ctrl.$modelValue || modelValue === modelValue)
+      // checks for NaN is needed to allow setting the model to NaN when there's an asyncValidator
+      // eslint-disable-next-line no-self-compare
+     (ctrl.$modelValue === ctrl.$modelValue || modelValue === modelValue)
```
gkalpak: Nice try! But the indentation i still off 😛

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+
+  /**
+   * @ngdoc method
+   *
+   * @name  ngModel.NgModelController#$processModelValue
+
+   * @description
+   *
+   * Runs the model -> view pipeline on the current
+   * {@link ngModel.NgModelController#$modelValue $modelValue}
```
gkalpak: Missing `.` at the end.

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+   * - if the $viewValue has changed
+   *   - {@link ngModel.NgModelController#$render $render} is called on the control
+   *   - the {@link ngModel.NgModelController#$validators $validators} are run and
+   *   the validation status is set.
+   *
+   * This method is called by ngModel internally when the bound scope value changes.
+   * Application developers usually do not have to call this function themselves.
+   *
+   * This function can be used when the $viewValue or the rendered DOM value of the control should
+   * be updated after a user input.
```
gkalpak: It is not necessarily about user input. The way I understand it, it is for when app state has changed in a way that may affect formatting. (OK, I admit most of the time the state changes as a result of user input, but still... 😛)

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+   *
+   * For example, consider a text input with an autocomplete list (for fruit, where the items are
+   * objects with a name and an id.
```
gkalpak: Missing closing `)`.

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+   * A user enters `ap` and then selects `Apricot` from the list.
+   * Based on this, the autocomplete widget will call $setViewValue({name: 'Apricot', id: 443}),
```
gkalpak: It is better to wrap code in `\. (E.g. `$setViewValue(...)` here or `ctrl.$processModelValue()` and `$viewValue` below.)

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+   * but the rendered value will still be `ap`.
+   * The widget can then call ctrl.$processModelValue() to run the model -> view
+   * pipeline again, which includes a formatter that converts the object into the string `Apricot`
+   * which is set to the $viewValue and rendered in the DOM.
```
gkalpak: Consider avoiding two `which` clauses. E.g. something line:

>The widget can then call `ctrl.$processModelValue()` to run the model -> view pipeline, which will use a formatter to convert the object to the string `Apricot`, update the `$viewValue` and finally rendered it in the DOM.

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+                if (angular.isObject(value) && value.name) {
+                  return value.name;
+                }
+
+                return value;
+              });
+
+              ngModel.$parsers.push(function(value) {
+                if (angular.isString(value)) {
+                  return scope.items.find(function(item) {
```
gkalpak: This is not ES5-compatible :grin:

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+                ngModel.$setViewValue(item);
+                ngModel.$processModelValue();
+              };
+            }
+          };
+        });
+     </file>
+     <file name="index.html">
+       <div style="display: flex;">
+         <div margin-right: 30px;">
```
gkalpak: Forgot about `style` :smiley:

src/ng/directive/ngModel.js
```diff
@@ -878,6 +878,141 @@ NgModelController.prototype = {
...
+           Search Fruit:
+           <input ng-model="val" process-model />
+
+           <ul>
+             <li ng-repeat="item in items | filter:val"><button ng-click="select(item)">{{item.name}}</li>
```
gkalpak: Ehem...relying on a method set on the scope by a directive on a different element doesn't sound like a good practice :wink:
Maybe move all relevant code into the directive's template.

gkalpak (commented):
Left some docs-related comments.
Otherwise LGTM 👍

Narretz (commented):
I've fixed the typos and commented on the more tricky stuff

src/ng/directive/ngModel.js
gkalpak: Apparently if you want to escape \`, you need to place the `\` _before_, not _after_ the \` 😇

src/ng/directive/ngModel.js
gkalpak: I don't feel too strongly, but it would be good for the docs to show code that works in all supportd browsers. It would also be a first (afaik).

src/ng/directive/ngModel.js
gkalpak: I was having something like this in mind: https://plnkr.co/edit/AeOA4eMGGMONSTrLinH2?p=preview
(TBH, I find this whole usecase weird. It sounds like an unnecessary complicated way to create an autocomplete :smiley:)

Narretz: Well, that's less complex and shows better encapsulation. I'll take it. But for the record, I don't think a real world autocomplete would be successful with a hard-coded input element. At least not for me  ;)

You know the customer is always right, and apparently many devs implement autocomplete like this.

Narretz: Updated with a small tweak

src/ng/directive/ngModel.js
petebacondarwin: I guess we still need to return this because we need the digest to cycle again if the model value has actually changed...?

petebacondarwin (approved):

src/ng/directive/ngModel.js
jbedard: Yeah this comment is outdated now. This was from when this was moved to the listener fn...

PR #16240, opened 2017-09-23
docs fix

IvanProdaiko94:
#### What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)
fix

#### What is the current behavior? (You can also link to an open issue here)
`unicornLauncher` registered but then it `unicornLauncherProvider` injected into the `.config` fn.

#### What is the new behavior (if this is a feature change)?
`unicornLauncher` renamed to `unicornLauncherProvider`

### DIFF
--- a/docs/content/guide/providers.ngdoc
+++ b/docs/content/guide/providers.ngdoc
@@ -212,7 +212,7 @@ that need it. We can make it configurable like so:
 
 
 ```javascript
-myApp.provider('unicornLauncher', function UnicornLauncherProvider() {
+myApp.provider('unicornLauncherProvider', function UnicornLauncherProvider() {
   var useTinfoilShielding = false;
 
   this.useTinfoilShielding = function(value) {


docs/content/guide/providers.ngdoc
```diff
@@ -212,7 +212,7 @@ that need it. We can make it configurable like so:
 
 
 ```javascript
-myApp.provider('unicornLauncher', function UnicornLauncherProvider() {
+myApp.provider('unicornLauncherProvider', function UnicornLauncherProvider() {
```
gkalpak: This is not right. This is registering a provider for the `unicornLauncher` service. The service will be injectable as `unicornLauncher`. (The provider itself will be injectable as `unicornLauncherProvider` during the config phase, but that is irrelevant.)

gkalpak (changes requested):

gkalpak:
Thx for taking the time to submit a pull request, but this change is invalid.

The `.provider()` method takes the name of the service as its first argument (`unicornLauncher`) and the provider (who will provide that service) as the second argument.

The service will be injected by its name (`unicornLanncher`). During the config phase (before any services are instantiated), you can inject the providers themselves and configure them (hence the name "config phase"). Provider are injected by the name of the service they provide plus the `Provider` suffix. So, the provider of the `unicornLauncher` service can be injected (during the config phase, before the actual service instance is created by the provider) as `unicornLauncherProvider`.

I hope this clears it up. Looking forward to the next pull request :smiley:

PR #16260, opened 2017-10-06
feat($parse): enable optional access to the underlying AST

### LINKED ISSUE #16253: Proposal: expose the AST for expressions parsed through $parse.
**I'm submitting a ...**

- [ ] bug report
- [x] feature request

**Current behavior:**

Currently, there seems to be no way to access the AST of expressions parsed through $parse.
**Expected / new behavior:**

It could be useful to be able to access the AST of parsed expressions in order to avoid building a custom parser to handle advanced use cases in which it is required to work on the expression's AST. 

I believe that the format of the current AST is not going to change in the future, so exposing it should not be of concern for possibly breaking changes.

It would be a nice add to have it exposed through the `$parse`'s return value. The new behavior could be made opt-in, just in case keeping the AST "alive" after parsing and for each expression, affects the memory consumption.

Thank you for your time!

**AngularJS version:** 1.6.6

fpipita:
This PR adds a new private method to the `$parse` service, `$$getAst`,
which takes an Angular expression as its only argument and returns
the computed AST. This feature is not meant to be part of the public
API and might be subject to changes, so use it with caution.

Closes #16253

Feature.

https://github.com/angular/angular.js/issues/16253

The new opt-in behavior introduces a way to expose the ast.

No.

### DIFF
--- a/src/ng/parse.js
+++ b/src/ng/parse.js
@@ -1644,11 +1644,26 @@ Parser.prototype = {
   constructor: Parser,
 
   parse: function(text) {
-    var ast = this.ast.ast(text);
-    var fn = this.astCompiler.compile(ast);
-    fn.literal = isLiteral(ast);
-    fn.constant = isConstant(ast);
+    var ast = this.getAst(text);
+    var fn = this.astCompiler.compile(ast.ast);
+    fn.literal = isLiteral(ast.ast);
+    fn.constant = isConstant(ast.ast);
+    fn.oneTime = ast.oneTime;
     return fn;
+  },
+
+  getAst: function(exp) {
+    var oneTime = false;
+    exp = exp.trim();
+
+    if (exp.charAt(0) === ':' && exp.charAt(1) === ':') {
+      oneTime = true;
+      exp = exp.substring(2);
+    }
+    return {
+      ast: this.ast.ast(exp),
+      oneTime: oneTime
+    };
   }
 };
 
@@ -1771,10 +1786,11 @@ function $ParseProvider() {
           isIdentifierStart: isFunction(identStart) && identStart,
           isIdentifierContinue: isFunction(identContinue) && identContinue
         };
+    $parse.$$getAst = $$getAst;
     return $parse;
 
     function $parse(exp, interceptorFn) {
-      var parsedExpression, oneTime, cacheKey;
+      var parsedExpression, cacheKey;
 
       switch (typeof exp) {
         case 'string':
@@ -1784,14 +1800,9 @@ function $ParseProvider() {
           parsedExpression = cache[cacheKey];
 
           if (!parsedExpression) {
-            if (exp.charAt(0) === ':' && exp.charAt(1) === ':') {
-              oneTime = true;
-              exp = exp.substring(2);
-            }
             var lexer = new Lexer($parseOptions);
             var parser = new Parser(lexer, $filter, $parseOptions);
             parsedExpression = parser.parse(exp);
-            parsedExpression.oneTime = !!oneTime;
 
             cache[cacheKey] = addWatchDelegate(parsedExpression);
           }
@@ -1803,6 +1814,12 @@ function $ParseProvider() {
         default:
           return addInterceptor(noop, interceptorFn);
       }
+    }
+
+    function $$getAst(exp) {
+      var lexer = new Lexer($parseOptions);
+      var parser = new Parser(lexer, $filter, $parseOptions);
+      return parser.getAst(exp).ast;
     }
 
     function expressionInputDirtyCheck(newValue, oldValueOfValue, compareObjectIdentity) {
--- a/test/ng/parseSpec.js
+++ b/test/ng/parseSpec.js
@@ -4245,4 +4245,48 @@ describe('parser', function() {
       });
     });
   });
+
+  describe('hidden/unsupported features', function() {
+    describe('$$getAst()', function() {
+      it('should be a method exposed on the `$parse` service', inject(function($parse) {
+        expect(isFunction($parse.$$getAst)).toBeTruthy();
+      }));
+
+      it('should accept a string expression argument and return the corresponding AST', inject(function($parse) {
+        var ast = $parse.$$getAst('foo.bar');
+        expect(ast).toEqual({
+          type: 'Program',
+          body: [
+            {
+              type: 'ExpressionStatement',
+              expression: {
+                type: 'MemberExpression',
+                object: { type: 'Identifier', name: 'foo' },
+                property: { type: 'Identifier', name: 'bar' },
+                computed: false
+              }
+            }
+          ]
+        });
+      }));
+
+      it('should parse one time binding expressions', inject(function($parse) {
+        var ast = $parse.$$getAst('::foo.bar');
+        expect(ast).toEqual({
+          type: 'Program',
+          body: [
+            {
+              type: 'ExpressionStatement',
+              expression: {
+                type: 'MemberExpression',
+                object: { type: 'Identifier', name: 'foo' },
+                property: { type: 'Identifier', name: 'bar' },
+                computed: false
+              }
+            }
+          ]
+        });
+      }));
+    });
+  });
 });


petebacondarwin (approved):
Might be worth mentioning the potential memory issues with this in the documentation.

src/ng/parse.js
```diff
@@ -1763,6 +1768,24 @@ function $ParseProvider() {
     return this;
   };
 
+    /**
+   * @ngdoc method
```
gkalpak: Since this is supposed to be private API, we don't want it to appear in the public docs.

src/ng/parse.js
```diff
@@ -1763,6 +1768,24 @@ function $ParseProvider() {
...
+   * @name $parseProvider#$$exposeAst
+   * @description
+   *
+   * Configures the access to the generated AST for parsed expressions.
+   * This feature is not meant to be part of the public API, so use it at your own risk.
+   *
+   * @param {boolean} exposeAst if set to true, the AST of each parsed expression will be exposed through
+   * an `$$ast` property added to the $parse's return value.
```
gkalpak: _the $parse's return value_ --> _\`$parse\`'s return value_

src/ng/parse.js
```diff
@@ -1763,6 +1768,24 @@ function $ParseProvider() {
...
+   *
+   **/
+  this.$$exposeAst = function(exposeAst) {
+    if (isBoolean(exposeAst)) {
+      globalExposeAst = exposeAst;
+    }
+  };
```
gkalpak: It might be better to stick to the getter/setter model we use in other places:

```js
this.$$exposeAst = function(exposeAst) {
  if (isUndefined(exposeAst)) {
    // Getter
    return globalExposeAst;
  } else {
    // Setter
    globalExposeAst = !!exposeAst;   // or: (exposeAst === true);
    return this;
  }
}
```

src/ng/parse.js
```diff
@@ -1790,7 +1821,7 @@ function $ParseProvider() {
             }
             var lexer = new Lexer($parseOptions);
             var parser = new Parser(lexer, $filter, $parseOptions);
-            parsedExpression = parser.parse(exp);
+            parsedExpression = parser.parse(exp, finalExposeAst);
             parsedExpression.oneTime = !!oneTime;
 
             cache[cacheKey] = addWatchDelegate(parsedExpression);
```
gkalpak: One problem is that cached parsedExpressions will always follow the `$$exposeAst` from the first call. E.g.:

```js
$parse('user.name');   // --> No `$$ast` (expected)
$parse({expression: 'user.name', $$exposeAst: true});   // --> No `$$ast` (ooops)
```

Not sure what is the best way to handle that (without impacting uninterested users) 😕 

##
Also, theoretically, the `addInterceptor()` function should preserve the `$$ast` property, but we can assume nobody using `$$exposeAst` cares about interceptors at this point. Maybe at least add a comment somewhere (that adding an interceptor won't preserve the exposed `$$ast` (if any)).

src/ng/parse.js
```diff
@@ -1775,6 +1798,14 @@ function $ParseProvider() {
 
     function $parse(exp, interceptorFn) {
       var parsedExpression, oneTime, cacheKey;
+      var finalExposeAst = globalExposeAst;
```
gkalpak: The `final` prefix is kind of confusing imo. I would just call it `exposeAst` (or `localExposeAst`).

test/ng/parseSpec.js
```diff
@@ -4156,6 +4156,61 @@ describe('parser', function() {
         }));
       });
     });
+
+    describe('ast access', function() {
+      it('should be forbidden by default', inject(function($parse) {
+        var fn = $parse('foo.bar');
+        expect(fn.$$ast).toBeUndefined();
+      }));
```
gkalpak: Can you also add a test with `$parse({expression: '...'})`?

test/ng/parseSpec.js
```diff
@@ -4156,6 +4156,61 @@ describe('parser', function() {
...
+
+      it('should be configurable on a per call basis', inject(function($parse) {
+        var fn = $parse({expression: 'foo.bar', $$exposeAst: true});
+        expect(fn.$$ast).toBeDefined();
+      }));
```
gkalpak: Can you also add a test with `$$exposeAst: false`?

gkalpak (changes requested):
Looking good overall 👍 I left a couple of (mostly minor) comments.

src/ng/parse.js
```diff
@@ -1763,6 +1768,24 @@ function $ParseProvider() {
     return this;
   };
 
+    /**
```
gkalpak: Indentation isssue 😱 😱 😱 😱 😱 😱 😱

petebacondarwin:
Thinking more about this...

Would it be simpler just to expose a method `$parse.$$getAst(expression)` instead?  The lexer and ast builder will already be set up; and this would avoid all of these extra provider flags and mucking about with the parameters of a public API service.

fpipita:
@petebacondarwin your proposal looks actually cleaner. I'll rework the PR and follow the new approach.

fpipita:
Hi, I've reworked the PR and implemented the new behavior.

I didn't fixup the commits yet to prevent the comments by @gkalpak from being removed.

petebacondarwin (approved):
This is much better and cleaner from an API point of view. I am frustrated by the code duplication between the new `$$getAst` function and the `$parse` function.

I think that we can get the code size and complexity down a bit by refactoring the `Parser` to expose a new function `getAst()`. The cost of creating a parser is pretty small (it has to create an `ASTCompiler/ASTInterpreter` but their constructors are very simple).

Moreover we could move the parsing of the one-time binding flag (`::`) into a helper in the `Parser` which might save some more lines.

But these are internal refactorings that I would be happy to land in subsequent PRs.

src/ng/parse.js
```diff
@@ -1805,6 +1806,27 @@ function $ParseProvider() {
...
+      if (isOneTimeBinding(exp)) {
+        exp = exp.substring(2);
+      }
+      lexer = new Lexer($parseOptions);
+      ast = new AST(lexer, $parseOptions);
+      return ast.ast(exp);
+    }
+
+    function isOneTimeBinding(exp) {
+      if (typeof exp !== 'string') {
```
gkalpak: We don't need this check afaict.

petebacondarwin:
@gkalpak see https://github.com/angular/angular.js/issues/16253#issue-262697219

gkalpak:
Yes, I've seen that. I am just curious what are they going to do with the AST?

fpipita:
@petebacondarwin thank you for your feedback, I can further improve the PR with the refactorings. That duplication made me quite frustrated as well!

@gkalpak sure, I'll share what I'm doing :)

I've built a form field component which can be used like:

```html
<cm-form-field>
  <!-- it can be anything bound to a model through ng-model -->
  <input ng-model="foo.bar.description" />
</cm-form-field>
```

The component has two main responsibilities:
1. It abstracts away the underlaying ui framework (I'm using Bootstrap) by taking care of generating the required markup;
2. It takes care of adding a default label for the `input` field by using the convention that the label is named the same way as the [`_.startCase`](https://lodash.com/docs/4.17.4#startCase)'d name of identifier ng-model is bound to, so in the above example, the `input` field will get a default label of "Description";
3. It also adds the `name` attribute to the `input` field which by convention is exactly the same as the identifier's one;

It works in collaboration with an overloaded version of `ng-model`, which does the job of informing the `cmFormField`'s controller about the identifier's name.

To extract the identifier's name, `ng-model` needs to parse the expression it is bound to and, at the moment, I'm achieving it by using a regular expression. It works fairly well for simple use cases, but if I wanted to do anything more complex (like detecting a computed property's name), I fear things can get quickly messy :)

```html
<cm-form-field>
  <input ng-model="foo[$ctrl.bar]" />
</cm-form-field>
```

By using the AST, it would be easy to do it, instead.

src/ng/parse.js
fpipita: I've added this check to future-proof this utility function, just in case we wanted to use it in contexts where it is not safe to assume that the caller will pass it a string.

If we can assume that it'll always be called with a string then I'll remove the check.

gkalpak:
@fpipita, I see. Pretty interesting usecase. Thx for sharing :smiley:

fpipita:
@petebacondarwin I've reworked a bit the PR in order to remove the code duplication and I've added the changes as a distinct commit so it is easier to inspect them and eventually discard them.

I've also removed the two initial commits related to the first implementation, since everyone agreed about it looking quite bad.

I'm almost entirely happy with how it looks right now, except for the only bit I wasn't able to deduplicate, the `exp.trim()` call, which is still done twice, in both `$parse` and in the newly added `Parser.prototype.getAst`.

gkalpak (approved):
LGTM (although this PR description needs updating).
I wonder if we want to show this commit on the changelog (considering it is a "non-public" feature).

fpipita:
@gkalpak I've updated the PR description, thank you for pointing that out.

Do you think that the `refactor` tag would be appropriate for the type of changes? If not, which one would you suggest to use?

One last thing .. will you please let me know whether I can merge the two commits? :)

petebacondarwin:
Please do squash the two commits.

I think it is OK to include it as a `feat` commit. As long as the description makes it clear that it is a private feature and that people should not rely on it being stable. (There is little point in hiding it, since it is simple to just read through the commits if one really wants).

fpipita:
Hi, I've squashed the two commits and tried to make it clearer through the description that this is a private feature.

gkalpak:
Still LGTM (restarted Travis - the failures looked like flakes)

